PR #7193, opened 2022-11-30
OBB frustum culling algorithm and unit tests

otsakir:
Oriented bounding box (OBB) - frustum intersection algorithm

### DIFF
--- a/engine/dlib/src/dlib/intersection.cpp
+++ b/engine/dlib/src/dlib/intersection.cpp
@@ -98,4 +98,52 @@ bool TestFrustumSphere(const Frustum& frustum, const dmVMath::Vector4& pos, floa
     return true;
 }
 
+// Returns 'false' if the bounding box is off the frustum. Returning 'true' does not guarantee that the object intersects the frustum or is inside it.
+bool TestFrustumOBB(const Frustum& frustum, const dmVMath::Matrix4& world, dmVMath::Vector3& aabb_min, dmVMath::Vector3& aabb_max)
+{
+    // calculate coordinates of all 8 corners of the bounding cube. To find them we'll take all points P(Xi,Yj,Zk) i=aabb_min.x|aabb_max.x, j=aabb_min.y|aabb_max.y, k=aabb_min.z|aabb_max.z
+    dmVMath::Point3 point0 = dmVMath::Point3(aabb_min.getX(), aabb_min.getY(), aabb_min.getZ());
+    dmVMath::Point3 point1 = dmVMath::Point3(aabb_min.getX(), aabb_min.getY(), aabb_max.getZ());
+    dmVMath::Point3 point2 = dmVMath::Point3(aabb_min.getX(), aabb_max.getY(), aabb_min.getZ());
+    dmVMath::Point3 point3 = dmVMath::Point3(aabb_min.getX(), aabb_max.getY(), aabb_max.getZ());
+    dmVMath::Point3 point4 = dmVMath::Point3(aabb_max.getX(), aabb_min.getY(), aabb_min.getZ());
+    dmVMath::Point3 point5 = dmVMath::Point3(aabb_max.getX(), aabb_min.getY(), aabb_max.getZ());
+    dmVMath::Point3 point6 = dmVMath::Point3(aabb_max.getX(), aabb_max.getY(), aabb_min.getZ());
+    dmVMath::Point3 point7 = dmVMath::Point3(aabb_max.getX(), aabb_max.getY(), aabb_max.getZ());
+
+    dmVMath::Vector4 corner_points[8]; // corner points in world coords
+
+    corner_points[0] = world * point0;
+    corner_points[1] = world * point1;
+    corner_points[2] = world * point2;
+    corner_points[3] = world * point3;
+    corner_points[4] = world * point4;
+    corner_points[5] = world * point5;
+    corner_points[6] = world * point6;
+    corner_points[7] = world * point7;
+
+    // for any of the six frustum planes if the all corner points lie in the negative halfspace, do cull the object
+    for (int plane_i=0; plane_i<6; plane_i++)
+    {
+        // get plane normal/equation
+        dmVMath::Vector4 plane_normal = frustum.m_Planes[plane_i];
+        bool positive_found = false;
+        for (int corner_i=1; corner_i<8; corner_i++)
+        {
+            float dotproduct = Vectormath::Aos::dot(corner_points[corner_i], plane_normal);
+            if (dotproduct >= 0)
+            {
+                positive_found = true;
+                break; // no need to check for the rest of the points
+            }
+        }
+        if (! positive_found) // if all corners are in the negative halfspace for the plane, we're done
+        {
+            return false; // no intersection, do cull
+        }
+
+    }
+    return true; // probalby inside the frustum but not 100% sure
+}
+
 } // dmIntersection
--- a/engine/dlib/src/dlib/intersection.cpp
+++ b/engine/dlib/src/dlib/intersection.cpp
@@ -20,8 +20,7 @@ namespace dmIntersection
 // Keeping this private for now. Making sure that the W component is always 1
 static inline float DistanceToPlane(Plane plane, dmVMath::Vector4 pos)
 {
-    dmVMath::Vector4 r = Vectormath::Aos::mulPerElem(plane, pos); // nx * px + ny * py + nz * pz + d
-    return Vectormath::Aos::sum(r);
+    return Vectormath::Aos::dot(plane,pos); // nx * px + ny * py + nz * pz + d*pw , pw - should be 1
 }
 
 float DistanceToPlane(Plane plane, dmVMath::Point3 pos)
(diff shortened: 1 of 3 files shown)


engine/dlib/src/dlib/intersection.cpp
```diff
@@ -98,4 +98,52 @@ bool TestFrustumSphere(const Frustum& frustum, const dmVMath::Vector4& pos, floa
     return true;
 }
 
+// Returns 'false' if the bounding box is off the frustum. Returning 'true' does not guarantee that the object intersects the frustum or is inside it.
+bool TestFrustumOBB(const Frustum& frustum, const dmVMath::Matrix4& world, dmVMath::Vector3& aabb_min, dmVMath::Vector3& aabb_max)
```
JCash, line 102: As an optimisation, I also want you to support the "skip_near_far" boolean that you see in the other functions.

This allows us to skip planes that one sometimes doesn't care about.

engine/dlib/src/dlib/intersection.cpp
```diff
@@ -98,4 +98,52 @@ bool TestFrustumSphere(const Frustum& frustum, const dmVMath::Vector4& pos, floa
...
+
+    // for any of the six frustum planes if the all corner points lie in the negative halfspace, do cull the object
+    for (int plane_i=0; plane_i<6; plane_i++)
+    {
+        // get plane normal/equation
+        dmVMath::Vector4 plane_normal = frustum.m_Planes[plane_i];
+        bool positive_found = false;
+        for (int corner_i=1; corner_i<8; corner_i++)
+        {
+            float dotproduct = Vectormath::Aos::dot(corner_points[corner_i], plane_normal);
```
JCash, line 133: This dot product seems off.

The point is in world space, and since the normal is just a normal, what you'll get is the distance from origin, not the distance to the plane.

See the usage of `DistanceToPlane` for reference.

engine/dlib/src/dlib/intersection.cpp
```diff
@@ -98,4 +98,52 @@ bool TestFrustumSphere(const Frustum& frustum, const dmVMath::Vector4& pos, floa
...
+            if (dotproduct >= 0)
```
JCash, line 134: As I mentioned earlier, doing per-point checks leads to trouble.

We should use a SAT test (Separating axis theorem).
I.e. find the min/max extents (call them d1 and d2) of the box when projected onto the planes normal.
If:
* If the the sign of the two values are differing (`d1 * d2 < 0.0f`), then there is for sure an intersection
* If both d1 and d2 are positive, then the OBB is in front of the plane


In short: For each point in the OOB, we at the same time measure the distance to the plane for its _opposite_ point.



Instead, use the _range_ that

engine/dlib/src/dlib/intersection.cpp
```diff
@@ -98,4 +98,52 @@ bool TestFrustumSphere(const Frustum& frustum, const dmVMath::Vector4& pos, floa
...
+                break; // no need to check for the rest of the points
+            }
+        }
+        if (! positive_found) // if all corners are in the negative halfspace for the plane, we're done
+        {
+            return false; // no intersection, do cull
+        }
+
+    }
+    return true; // probalby inside the frustum but not 100% sure
```
JCash, line 146: "Probalby"

The comment is also interesting, since you don't elaborate on why it isn't 100% sure?

engine/dlib/src/test/test_intersection.cpp
```diff
@@ -105,6 +105,61 @@ TEST(dmVMath, TestFrustumSphere)
 
 }
 
+TEST(dmVMath, TestFrustumOBB)
+{
+    dmVMath::Matrix4 proj = dmVMath::Matrix4::orthographic(0.0f, FRUSTUM_WIDTH, 0.0f, FRUSTUM_HEIGHT, FRUSTUM_NEAR, FRUSTUM_FAR);
```
JCash, line 110: I would also like to see tests with a non orthographics projection matrix, as that's what most users will use with this function.

You can ofc choose constants that make it easier to reason about in the tests.

JCash (changes requested):
Overall I think there are a lot of calculations going on and I feel it should be possible to also simplify it a bit.

E.g. what if we for each plane in the frustum transform the plane into the space of the OOBB. Then we make it into a problem of AABB vs Plane intersections instead.
https://gdbooks.gitbooks.io/3dcollisions/content/Chapter2/static_aabb_plane.html

E:g. this approach to transform a plane using a matrix4 seems promising:
https://stackoverflow.com/a/48408289/468516

engine/dlib/src/dlib/intersection.cpp
JCash: Ah, good call, I missed the fact that the W argument is 1.
That's what the Point3 is for, and I got confused by the Vector4.

engine/dlib/src/dlib/intersection.cpp
```diff
@@ -98,4 +98,52 @@ bool TestFrustumSphere(const Frustum& frustum, const dmVMath::Vector4& pos, floa
...
+        // get plane normal/equation
+        dmVMath::Vector4 plane_normal = frustum.m_Planes[plane_i];
+        bool positive_found = false;
+        for (int corner_i=1; corner_i<8; corner_i++)
+        {
+            float dotproduct = Vectormath::Aos::dot(corner_points[corner_i], plane_normal);
+            if (dotproduct >= 0)
+            {
+                positive_found = true;
+                break; // no need to check for the rest of the points
```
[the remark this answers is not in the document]
JCash: NVM, I was under the impression that I had made the frustum normals point outwards.

engine/dlib/src/dlib/intersection.cpp
JCash: I still think the comment is "wrong", in that it gives the impression we are not sure what we are doing.
Instead, you could state what false positives _are_ allowed to happen.

engine/dlib/src/dlib/intersection.cpp
JCash, line 133: This is a good improvement to `DistanceToPlane(Plane plane, dmVMath::Vector4 pos)`.
Please update that function and use that here instead.

engine/dlib/src/dlib/intersection.cpp
otsakir: yes, It's kind of funny to read. And to write. I pondered a little on the return type (bool) and the name of the function which prepares the reader for a clean and strict piece of functionality.

Ok, I'll update the comment accordingly.

engine/dlib/src/dlib/intersection.cpp
```diff
@@ -98,4 +98,52 @@ bool TestFrustumSphere(const Frustum& frustum, const dmVMath::Vector4& pos, floa
...
+            }
+        }
+        if (! positive_found) // if all corners are in the negative halfspace for the plane, we're done
+        {
+            return false; // no intersection, do cull
+        }
```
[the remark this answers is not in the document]
otsakir: I'm not sure if i have to answer this or it's already settled.

If **all** corner points are in the negative halfspace of **any** one frustum plane (i.e. !positive_found) we don't care if they in front of another plane or not. They are for sure outside the frustum volume, so we can safely return 'false'.

engine/dlib/src/test/test_intersection.cpp
otsakir: Ok, i'll add another test for perspective projection. 

I'm not sure wdym by "constants" though.

otsakir: Done.

JCash (changes requested):

engine/dlib/src/test/test_intersection.cpp
```diff
@@ -113,80 +128,111 @@ TEST(dmVMath, TestFrustumOBB)
...
+    float BOX_SIDE = 1;
     dmVMath::Vector3 minPoint(0,0,0);
     dmVMath::Vector3 maxPoint(BOX_SIDE,BOX_SIDE,BOX_SIDE); // note, the box lies in the +Z axis
 
     dmVMath::Matrix4 trans;
 
     // place BB at (0,0,0), outside the frustum that starts at -FRUSTUM_NEAR
     trans = dmVMath::Matrix4::identity();
-    ASSERT_FALSE(dmIntersection::TestFrustumOBB(frustum, trans, minPoint, maxPoint));
+    ASSERT_FALSE(dmIntersection::TestFrustumOBB(frustum, trans, minPoint, maxPoint, false));
```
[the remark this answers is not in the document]
otsakir: yup, makes sense.

engine/dlib/src/dlib/intersection.cpp
```diff
@@ -123,14 +122,14 @@ bool TestFrustumOBB(const Frustum& frustum, const dmVMath::Matrix4& world, dmVMa
...
+    int max_planes = skip_near_far ? 4 : 6;
+    for (int plane_i=0; plane_i<max_planes; plane_i++)
     {
         // get plane normal/equation
-        dmVMath::Vector4 plane_normal = frustum.m_Planes[plane_i];
         bool positive_found = false;
         for (int corner_i=1; corner_i<8; corner_i++)
         {
-            float dotproduct = Vectormath::Aos::dot(corner_points[corner_i], plane_normal);
+            float dotproduct = DistanceToPlane(frustum.m_Planes[plane_i], corner_points[corner_i]);
```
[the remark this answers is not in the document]
otsakir: ok

engine/dlib/src/test/test_intersection.cpp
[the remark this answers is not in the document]
otsakir: done

JCash (approved):

PR #7195, opened 2022-12-01
Fixes html5 gamepad released state detection

britzl:
Fixes a regression introduced from standard gamepad support in HTML5 where the released state of gamepads wasn't properly detected.

### DIFF
--- a/engine/gameobject/src/gameobject/comp_script.cpp
+++ b/engine/gameobject/src/gameobject/comp_script.cpp
@@ -502,8 +502,8 @@ namespace dmGameObject
                 lua_settable(L, -3);
 
                 lua_pushliteral(L, "gamepad_buttons");
-                lua_createtable(L, dmHID::MAX_GAMEPAD_AXIS_COUNT, 0);
-                for (int i = 0; i < dmHID::MAX_GAMEPAD_AXIS_COUNT; ++i)
+                lua_createtable(L, dmHID::MAX_GAMEPAD_BUTTON_COUNT, 0);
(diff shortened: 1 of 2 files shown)


engine/glfw/js/library_glfw.js
```diff
@@ -573,12 +573,12 @@ var LibraryGLFW = {
             var gamepad = GLFW.lastGamepadState[joy];
 
             if (gamepad) {
-              if (!GLFW.joys[joy] || GLFW.joys[joy].id_string != gamepad.id) {
+              var gamepad_id = (gamepad.mapping == "standard") ? "Standard Gamepad" : gamepad.id;
```
britzl, line 576: The fix

engine/gameobject/src/gameobject/comp_script.cpp
```diff
@@ -502,8 +502,8 @@ namespace dmGameObject
                 lua_settable(L, -3);
 
                 lua_pushliteral(L, "gamepad_buttons");
-                lua_createtable(L, dmHID::MAX_GAMEPAD_AXIS_COUNT, 0);
-                for (int i = 0; i < dmHID::MAX_GAMEPAD_AXIS_COUNT; ++i)
+                lua_createtable(L, dmHID::MAX_GAMEPAD_BUTTON_COUNT, 0);
```
britzl, line 505: Unrelated but discovered while debugging.

Surely this must have been wrong? Both MAX_GAMEPAD_AXIS_COUNT and MAX_GAMEPAD_BUTTON_COUNT have the value 32 so it has so far not been a problem

JCash (approved):

PR #7196, opened 2022-12-01
Create atlas resource in runtime

### LINKED ISSUE #7192: Add resource.create_atlas function
**Task (REQUIRED):**
We already have resource.create_texture, but we need resource.create_atlas as well. Should be pretty self explanatory :) 

Part 4 of issue https://github.com/defold/defold/issues/5948

Jhonnyg:
Added new function *resource.create_atlas(path, args)* to create an atlas resource in runtime. The resource is almost the same as an atlas generated during build time except for a few caveats. Please consult the API reference for how to use this feature.

Fixes #7192 

* [x] Is this a new feature?


engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1194,194 +1571,24 @@ static int SetAtlas(lua_State* L)
...
     // Note: We do a separate pass over the lua state to validate the data in the args table,
     //       this is because we need to allocate dynamic memory and can't use luaL_check** functions
     //       since they longjmp away so we can't release the memory..
-    ValidateSetAtlasArgumentsFromLua(L, &num_geometries, &num_animations);
-
-    texture_set_ddf.m_Geometries.m_Data  = new dmGameSystemDDF::SpriteGeometry[num_geometries];
-    texture_set_ddf.m_Geometries.m_Count = num_geometries;
-    memset(texture_set_ddf.m_Geometries.m_Data, 0, sizeof(dmGameSystemDDF::SpriteGeometry) * num_geometries);
-
-    texture_set_ddf.m_Animations.m_Data  = new dmGameSystemDDF::TextureSetAnimation[num_animations];
```
Jhonnyg, line 1229: Moved all of this into a helper function since we want to do the same validation when creating the atlas as here

engine/gamesys/src/gamesys/test/resource/script_atlas.script
```diff
@@ -136,48 +197,69 @@ local function test_setting_valid(self)
...
+    -- invalid tests
+    test_get_atlas_invalid(self)
+
+    -- resource.set_atlas and resource.create_atlas should support the same arguments
+    -- so we can reuse the test functions here
+    test_atlas_invalid(self, resource.set_atlas, self.valid_tileset_path)
+    test_atlas_invalid(self, resource.create_atlas, self.valid_tileset_path)
+
+    -- do a specific invalid test for create_atlas
+    test_create_atlas_invalid(self)
```
Jhonnyg, line 263: Added tests for creating atlases as well to this test file instead of creating a whole new test. There is a lot of things in common with create_atlas and set_atlas with a bunch of helper functions so figured it would be better to reuse that instead of a new test

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -564,6 +583,22 @@ static void DestroyTextureImage(dmGraphics::TextureImage& texture_image)
     delete[] texture_image.m_Alternatives.m_Data;
 }
 
+static void CheckTextureResource(lua_State* L, int i, const char* field_name, const char** texture_path_out, dmGraphics::HTexture* texture_out)
+{
+    lua_getfield(L, i, field_name);
+    if (!lua_isstring(L, -1))
```
JCash, line 589: I think in general, for our "CheckXxx" functions we use "CheckHashOrString()" for getting the name of something.

I think it's likely that we might want to be able to receive a hash here?

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -564,6 +583,22 @@ static void DestroyTextureImage(dmGraphics::TextureImage& texture_image)
...
+    {
+        luaL_error(L, "Unable to get texture resource from field '%s', string expected", field_name);
```
JCash, line 591: This goes against the intended mechanism of DM_LUA_STACK_CHECK.
Either you let the caller handle the error, or you don't use DM_LUA_STACK_CHECK for this.

engine/gamesys/src/gamesys/scripts/script_resource.cpp
JCash: E.g. if the user wants to use a texture that is assigned to some resource (e.g. a sprite).
It doesn't even have to be a dynamically generated texture.

Looking at the code below, we use the string to set it into the texture set.
But, perhaps we can leave that string as 0, and instead set a texture name hash?

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -950,112 +972,467 @@ static dmGameSystemDDF::Playback GameObjectPlaybackToDDFPlayback(dmGameObject::P
...
+    assert(lua_gettop(L) == top);
+}
+
+// Creates a texture set from the lua stack, it is expected that the argument
+// table is on top of the stack and that all fields have valid data
+static void MakeTextureSetFromLua(lua_State* L, const char* texture_path, dmGraphics::HTexture texture, uint32_t num_geometries, uint8_t num_animations, dmGameSystemDDF::TextureSet* texture_set_ddf)
+{
+    int top = lua_gettop(L);
+
+    texture_set_ddf->m_Texture = texture_path;
```
JCash, line 1098: See note above regarding hash of texture resource path.

Apart from creating resources when loading from disc, what other uses does this string have?
Possibly this could be 0, and instead we set a hash value?

JCash (changes requested):

JCash (approved):

PR #7198, opened 2022-12-02
Added support for index buffers in vertex2.clj

JCash:
Same mechanic as was implemented previously in vertex.clj


vlaaad (approved):
LGTM! Is it basically copied from `editor.gl.vertex` ns?

PR #7199, opened 2022-12-02
Extract @note annotations to docs

britzl:


engine/script/src/script/script_doc_ddf.proto
```diff
@@ -64,6 +64,7 @@ message Info
     required string description = 4;
     required string path = 5;
     required string file = 6;
+    repeated string notes = 7;
```
Jhonnyg, line 67: The Element message has a repeated string notes, is that used for something else?

Jhonnyg (approved):
👍  just a small question regarding the proto files

engine/script/src/script/script_doc_ddf.proto
britzl: It is my understanding that Element is used for the documentation next to a function, struct or similar. Info on the other hand is used for top-of-file docs.

britzl: Or rather like on the namespace, see dlib/src/dlib/socket.h for an example.

Jhonnyg: Gotcha, what if you want to annotate something for a specific function? This means all @note texts will end up at the top with the description right? So if I want a specific note for some function argument I'd have to make sure to put in the correct place?

britzl: No no, we still track `@notes` for both and on the right level. Notes on a function will be shown on the function (once we add it to the reference API generator)

PR #7200, opened 2022-12-02
Make the search for resource links more efficient

### LINKED ISSUE #7176: Editor is slow when launching game and tabbing back to editor to read console
**Description**
The editor completely locks up when tabbing back to the editor after launching a large game which does a lot of logging to the console.

The editor should not lock up

vlaaad:
On big projects with a lot of resources, it might take a lot of time to show output in the console because for every potential resource link we have to filter through all tracked resources. We now try to be smarter about it, both by limiting the candidates for search only to ellipsis-prefixed ones and by building a suffix map for the added lines batch so we have to go through all the resources less frequently.

Fixes #7176

### DIFF
--- a/editor/src/clj/editor/console.clj
+++ b/editor/src/clj/editor/console.clj
@@ -23,7 +23,6 @@
             [editor.handler :as handler]
             [editor.resource :as resource]
             [editor.ui :as ui]
-            [editor.util :as util]
             [editor.workspace :as workspace])
   (:import [clojure.lang PersistentQueue]
            [editor.code.data Cursor CursorRange LayoutInfo Rect]
@@ -355,7 +354,6 @@
 
 (def ^:const line-sub-regions-pattern #"(?<=^|\s|[<\"'`])(\/[^\s>\"'`:]+)(?::?)(\d+)?")
 (def ^:private ^:const line-sub-regions-pattern-partial #"([^\s<>:]+):(\d+)")
-(def ^:private ^:const sub-region-strip-ellipsis-prefix-pattern #"^(?>\.{3})?(.*)")
 
 (def ^:private non-empty-string? (comp string? not-empty))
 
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/console.clj
```diff
@@ -413,26 +424,27 @@
                              (data/->Cursor row (count line)))
     :type type))
 
-(defn- make-line-regions [resource-map on-region-click! ^long row [type line]]
+;; todo fit in 4 args?
+(defn- make-line-regions [resource-map resource-suffix-map-delay on-region-click! row [type line]]
```
vlaaad, line 428: I had to remove the type-hint since they are supported only up to 4 args (i.e. it was no longer compiling)

editor/src/clj/editor/console.clj
```diff
@@ -442,37 +454,39 @@
...
+              resource-suffix-map-delay (make-resource-suffix-map-delay resource-map)
+              props (append-entries {:lines (if clear? [""] prev-lines)
+                                     :regions (if clear? [] prev-regions)}
+                                    entries resource-map resource-suffix-map-delay on-region-click!)]
+          (view/set-properties! view-node nil
+                                (cond-> (assoc props :document-width document-width)
+                                        was-scrolled-to-bottom? (assoc :scroll-y (data/scroll-to-bottom prev-layout (count (:lines props))))
+                                        clear? (assoc :cursor-ranges [data/document-start-cursor-range])
+                                        clear? (assoc :invalidated-row 0)
+                                        clear? (data/frame-cursor prev-layout)))))))
```
vlaaad, line 489: Mostly indentation due to wrapping in `with-auto-evaluation-context`,  but also I added a line that constructs `resource-suffix-map-delay`.

matgis (approved):
Looks great! I had some questions about the changes to the test, but apart from that it should be ready to merge.

editor/test/editor/console_test.clj
```diff
@@ -134,8 +134,11 @@
...
-                                              identity 10
-                                              "urce/in/the/project/the_resource.script:25: in function <urce/in/the/project/the_resource.script:24> </absolute/project/path/file.lua:65>")))))
+  (let [resource-map {"/absolute/project/path/file.lua" :WIN
+                      "/some/really/long/path/to/some/resource/in/the/project/the_resource.script" :WIN}]
+   (is (= 3 (count
+              (#'console/make-line-sub-regions
+                resource-map
+                (console/make-resource-suffix-map-delay resource-map)
+                identity 10
+                "...to/some/resource/in/the/project/the_resource.script:25: in function <...to/some/resource/in/the/project/the_resource.script:24> </absolute/project/path/file.lua:65>"))))))
```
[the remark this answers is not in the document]
vlaaad: The syntax changed, so the test was updated to reflect that. The change is: we need at least 50 characters (the size of the suffix key) after `...` for the suffix search to work. Currently, the ellipsis-prefixed paths are 56 characters long, but I wanted to have some wiggle room in case there were some minor differences in behavior between e.g. different platforms or architectures. The shortened paths in this test were less than 50 characters.

editor/src/clj/editor/console.clj
vlaaad: Agreed, I'll remove the todo.

PR #7204, opened 2022-12-06
Issue 6455 mesh frustum culling

otsakir:
Enables frustum culling checks for mesh component based on AABB metadata

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_mesh.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_mesh.cpp
@@ -28,6 +28,7 @@
 #include <dlib/dstrings.h>
 #include <dlib/transform.h>
 #include <dmsdk/dlib/vmath.h>
+#include <dmsdk/dlib/intersection.h>
 #include <gameobject/gameobject_ddf.h>
 #include <graphics/graphics.h>
 #include <render/render.h>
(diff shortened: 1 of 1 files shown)


engine/gamesys/src/gamesys/components/comp_mesh.cpp
```diff
@@ -739,6 +740,86 @@ namespace dmGameSystem
...
+                case dmBuffer::VALUE_TYPE_INT64:
+                    MetadataAabbToPoints((int64_t*)data, valueType, boundsMin, boundsMax);
+                break;
+                default:
+                    dmLogError("Invalid typeValue in buffer metadata");
+                    continue;
+
+            }
+
+            bool intersect = dmIntersection::TestFrustumOBB(frustum, component_p->m_World, boundsMin, boundsMax, false);
```
otsakir, line 815: Here skip_near_far is hardcoded to 'false'.

I'm not sure how to handle this. When would this be 'true' ?

An idea that came to mind is to "allow" AABB metadata with 4 instead of 6 numbers. If it's 4, then it's 2D culling. Set skip_near_far to 'true'. Otherwise set to 'false'.

engine/gamesys/src/gamesys/components/comp_mesh.cpp
```diff
@@ -739,6 +740,86 @@ namespace dmGameSystem
...
+                    MetadataAabbToPoints((int8_t*)data, valueType, boundsMin, boundsMax);
+                break;
+                case dmBuffer::VALUE_TYPE_INT16:
+                    MetadataAabbToPoints((int16_t*)data, valueType, boundsMin, boundsMax);
+                break;
+                case dmBuffer::VALUE_TYPE_INT32:
+                    MetadataAabbToPoints((int32_t*)data, valueType, boundsMin, boundsMax);
+                break;
+                case dmBuffer::VALUE_TYPE_INT64:
+                    MetadataAabbToPoints((int64_t*)data, valueType, boundsMin, boundsMax);
```
JCash, line 807: I think the extra code generated for something that we know the resulting data type for is unnessary.

E.g. this should work fine:
```
Vector3 bmin( (float)aabb[0], (float)aabb[1], (float)aabb[2] );
Vector3 bmax( (float)aabb[3], (float)aabb[4], (float)aabb[5] );
```

engine/gamesys/src/gamesys/components/comp_mesh.cpp
```diff
@@ -739,6 +740,86 @@ namespace dmGameSystem
...
+            dmVMath::Vector3 boundsMin, boundsMax;
+
+            void* data;
+            uint32_t count;
+            dmBuffer::ValueType valueType;
+            dmBuffer::Result r = dmBuffer::GetMetaData(br->m_Buffer, dmHashString64("AABB"), &data, &count, &valueType );
+            if (r == dmBuffer::RESULT_METADATA_MISSING) {
+                // no bounding box available - no culling
+                entry->m_Visibility = dmRender::VISIBILITY_FULL; // TODO - or shall we just "continue"
+                continue;
```
JCash, line 774: Redundant comment, we _are_ continuing?

engine/gamesys/src/gamesys/components/comp_mesh.cpp
```diff
@@ -739,6 +740,86 @@ namespace dmGameSystem
...
+            dmRender::RenderListEntry* entry = &params.m_Entries[i];
+            MeshComponent* component_p = (MeshComponent*)entry->m_UserData;
+
+            dmGameSystem::BufferResource* br = GetVerticesBuffer(component_p, component_p->m_Resource);
+            dmVMath::Vector3 boundsMin, boundsMax;
+
+            void* data;
+            uint32_t count;
+            dmBuffer::ValueType valueType;
+            dmBuffer::Result r = dmBuffer::GetMetaData(br->m_Buffer, dmHashString64("AABB"), &data, &count, &valueType );
```
JCash, line 770: No need to hash the string each loop.
Create a constant at file scope instead.

engine/gamesys/src/gamesys/components/comp_mesh.cpp
JCash: I don't think we should use 2D aabb.
I think using different functions is better.

Eventually, we probably want the user to set the flag from the render script, just like they do today when choosing the frustum.

JCash (changes requested):

engine/gamesys/src/gamesys/components/comp_mesh.cpp
otsakir: I'm not sure what you mean.

Shall we use skip_near_far=false for all types of meshes for now in comp_mesh.cpp ?

What about different functions ? Do you mean some sort of overloading in dmIntersection ? Like this ?

```

dmIntersection::TestFrustumOBB2D(frustum, component_p->m_World, boundsMin, boundsMax) {
     dmIntersection::TestFrustumOBB(frustum, component_p->m_World, boundsMin, boundsMax, true);
}

dmIntersection::TestFrustumOBB3D(frustum, component_p->m_World, boundsMin, boundsMax); {
     dmIntersection::TestFrustumOBB(frustum, component_p->m_World, boundsMin, boundsMax, false);
}
```

JCash: Not entirely sure what I meant with "different functions" here.
I think the current solution works: TestFrustumOBB and AABB with 6 floats.

engine/gamesys/src/gamesys/components/comp_mesh.cpp
JCash: No, but that's the "API", and what we expect, so we only need to support that.

otsakir: ok

engine/gamesys/src/gamesys/components/comp_mesh.cpp
otsakir: cool. 

No action item then. I'll the conversation open.

engine/gamesys/src/gamesys/components/comp_mesh.cpp
```diff
@@ -780,39 +772,8 @@ namespace dmGameSystem
                 continue;
             }
```
JCash, line 773: Since this data comes from the user, I would like us to make sure that the number of elements in the array is actually 6. Otherwise the usage below may go out of bounds.

JCash (changes requested):

engine/gamesys/src/gamesys/components/comp_mesh.cpp
otsakir: @JCash , how about also checking that the type is float ? 
Having an array of 6 8bit values will still get us out of bounds.

JCash: True, it's a good point. Lets check that too then.

otsakir: It's already done in the commit bellow (910ce). 
Closing this conversation.

otsakir:
@JCash , all comments have been addressed.

JCash (approved):

PR #7206, opened 2022-12-06
resource.set_texture transcode support

### LINKED ISSUE #7202: Transcode support for resource.create_texture and resource.set_texture
**Task (REQUIRED):**
There is currently no transcode support for creating / setting textures via resource functions, which is needed to enable better OTA sizes as well as avoiding image encoding in runtime.

Jhonnyg:
Added support for setting compressed and transcoded texture data in the resource.set_texture function. This can be used to decrease the runtime memory for dynamically created texture resources. The function now accepts binary data from .basis files and will pick a compressed GPU format for the client platform. 

Fixes #7202

* [x] Is this a new feature?


engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -589,15 +607,37 @@ static void DestroyTextureImage(dmGraphics::TextureImage& texture_image)
...
+ * - `resource.TEXTURE_FORMAT_RGBA_PVRTC_2BPPV1`
+ * - `resource.TEXTURE_FORMAT_RGBA_PVRTC_4BPPV1`
+ * - `resource.TEXTURE_FORMAT_RGB_ETC1`
+ * - `resource.TEXTURE_FORMAT_RGBA_ETC2`
+ * - `resource.TEXTURE_FORMAT_RGBA_ASTC_4x4`
+ * - `resource.TEXTURE_FORMAT_RGB_BC1`
+ * - `resource.TEXTURE_FORMAT_RGBA_BC3`
+ * - `resource.TEXTURE_FORMAT_R_BC4`
+ * - `resource.TEXTURE_FORMAT_RG_BC5`
+ * - `resource.TEXTURE_FORMAT_RGBA_BC7`
```
Jhonnyg, line 626: We expose these values to the script but don't say that we support them. If we expose them then the docs have to be honest about that too I think

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -673,7 +722,7 @@ static int CreateTexture(lua_State* L)
     void* resource = 0x0;
     dmResource::Result res = dmResource::CreateResource(g_ResourceModule.m_Factory, path_str, ddf_buffer.Begin(), ddf_buffer.Size(), &resource);
 
-    DestroyTextureImage(texture_image);
+    DestroyTextureImage(texture_image, buffer == 0);
```
Jhonnyg, line 725: If the script passed in a buffer we shouldn't delete the data since the buffer manages that

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1965,6 +2112,16 @@ static void LuaInit(lua_State* L)
 
 #undef SETGRAPHICSCONSTANT
 
+
+#define SETCOMPRESSIONTYPE(name) \
+    lua_pushnumber(L, (lua_Number) dmGraphics::TextureImage:: name); \
+    lua_setfield(L, -2, #name);
+
+    SETCOMPRESSIONTYPE(COMPRESSION_TYPE_DEFAULT);
+    SETCOMPRESSIONTYPE(COMPRESSION_TYPE_BASIS_UASTC);
```
Jhonnyg, line 2121: The ETC1 enum is not exposed here since we don't suport it

engine/tools/src/texconvert/main.cpp
```diff
@@ -0,0 +1,262 @@
...
+    printf("  --color-space <color-space> : Sets the color space, defaults to 'SRGB'. Supported values are:\n");
+    PRINT_ARG_LIST(g_cs_lut);
+    printf("  --compression-type          : Sets the compression type, defaults to 'DEFAULT'. Supported values are:\n");
+    PRINT_ARG_LIST(g_ct_lut);
+    printf("  --compression-level         : Sets the compression level, defaults to 'NORMAL'. Supported values are:\n");
+    PRINT_ARG_LIST(g_cl_lut);
+#undef PRINT_ARG_LIST
+}
+
+int main(int argc, char const *argv[])
```
Jhonnyg, line 238: I created this cmd line tool so that it's easy to use our internal encoder to convert between common image formats into basis files. We use the Texc library in bob as well, which we use in the TextureGenerator to generate test content for the engine and we *could* restructure that to do something similar like this tool, but the benefit here is:
* users can use the tool to produce runtime data with correct formats/versions we support in the engine and don't need to deal with java (class paths etc)
* the java tooling requires texture profiles to output compressed and transcoded formats
* this tool is more isolated which means it's easier to distribute in a game team

I guess we could move the functionality to the java class through waf_texc.py or something like that, but I personally would prefer to keep using this small tool instead of our pipeline

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -658,13 +703,22 @@ static int CreateTexture(lua_State* L)
...
 
+    dmBuffer::HBuffer buffer = 0;
+    if (dmScript::IsBuffer(L, 3))
+    {
+        // If a buffer has been passed in we can optionally support compression type
+        dmScript::LuaHBuffer* l_buffer = dmScript::CheckBuffer(L, 3);
+        buffer                         = l_buffer->m_Buffer;
+        compression_type               = (dmGraphics::TextureImage::CompressionType) CheckTableInteger(L, 2, "compression_type", (int) dmGraphics::TextureImage::COMPRESSION_TYPE_DEFAULT);
+    }
+
```
Jhonnyg, line 714: This is a new option for creating texture that perhaps should be moved out from this PR? 🤔 The reason the option was added is because we don't have a way of creating transcoded data in runtime via the buffer, so to create a compressed texture you first needed to create a dummy texture and then use set_texture to upload the compressed data. That's not optimal and so I think we do want to support passing a buffer to this function that can be used to create a texture with initial data without having to make the dummy texture first..

engine/gamesys/src/gamesys/test/resource/set_texture.script
```diff
@@ -77,21 +77,45 @@ local function test_fail_wrong_mipmap(self)
...
-	if self.update_counter == 0 then
-		test_success_simple(self)
-	elseif self.update_counter == 1 then
-		test_success_resize(self)
-	elseif self.update_counter == 2 then
-		test_fail_out_of_bounds(self)
-	elseif self.update_counter == 2 then
-		test_fail_wrong_mipmap(self)
+local function test_success_compressed(self)
+	local f     = io.open("src/gamesys/test/resource/blank.basis", "rb")
```
JCash, line 86: This won't work on Switch/PS4/PS5 tests (or Android/iOS later on for that matter).

You should use the `DM_HOSTFS` define that the build system sets, and pass that to the Lua script somehow.

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -2088,6 +2129,84 @@ static const luaL_reg Module_methods[] =
...
+ *
+ * @name resource.TEXTURE_FORMAT_RGBA_BC7
+ * @variable
+ */
+
+/*# BASIS_UASTC compression type
+ *
+ * @name resource.COMPRESSION_TYPE_BASIS_UASTC
+ * @variable
+ */
```
JCash, line 2208: Does the default texture compression need to be listed?

JCash (changes requested):

engine/gamesys/src/gamesys/scripts/script_resource.cpp
Jhonnyg: Not sure what you mean, do you mean that we should mark which one of these that is the default value in the docs?

PR #7208, opened 2022-12-07
Editor: Allow non-byte buffers in vertex2 and use IntBuffer for indices

matgis:
Additional fixes to changes introduced in #7198 in preparation for update Rive plugin.

### Technical changes
* Relax requirement on underlying `Buffer` in `vertex2/VertexBuffer` to allow us to use a `FloatBuffer`.
* Use `IntBuffer` for indices in `vertex2/VertexBuffer`.

### DIFF
--- a/editor/src/clj/editor/gl/vertex2.clj
+++ b/editor/src/clj/editor/gl/vertex2.clj
@@ -19,11 +19,10 @@
    [editor.gl.protocols :refer [GlBind]]
    [editor.gl.shader :as shader]
    [editor.scene-cache :as scene-cache]
-   [editor.types :as types]
    [internal.util :as util])
   (:import
    [com.jogamp.common.nio Buffers]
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/gl/vertex2.clj
```diff
@@ -83,7 +82,7 @@
   (position! [this position])
   (version [this]))
 
-(deftype VertexBuffer [vertex-description usage ^ByteBuffer buf ^{:unsynchronized-mutable true} version]
+(deftype VertexBuffer [vertex-description usage ^Buffer buf ^{:unsynchronized-mutable true} version]
```
matgis, line 85: The OpenGL methods require this to be a `Buffer`. If we relax the requirement here we can use a `FloatBuffer` as well as a `ByteBuffer`.

vlaaad (approved):
LGTM!

PR #7214, opened 2022-12-09
Create buffer in runtime

### LINKED ISSUE #5648: Create buffer resources at runtime
I have the following use-case: I want to create, at runtime, an arbitrary number of mesh components each with their own different mesh buffer. As it is right now, I can't (not sanely at least, see workaround), since buffer resources can only be declared at design time, so they are of a fixed number.

Two functions:
`resource.create_buffer(buffer)`: Given a buffer, creates a buffer resource and returns a hash handle to it that can be plugged into game object and component properties. The newly created resource will have a reference count of 1 and can be `release()`-d after being attached to a property.
`resource.release(resource_hash)` Decrements the reference count of a resource.

Using a script to generate 100 `.json` files, a really long dummy script filled with `go.property()` declarations and a really long `.go` file to make the associations. This way I could make a pool of empty buffer resources.

Not really relevant, but we're trying to do a sort of 3D tilemap and putting together the pieces inside a single buffer to render the mesh in one go seems like the thing to do.

xNDUrGM
AQwKfxb

Jhonnyg:
Added new function _resource.create_buffer_ to create buffer resources dynamically in runtime. The function takes a buffer object created by other buffer resources, from the C api or via the lua _buffer.create_ function.

Fixes #5648 

* [x] Is this a new feature?


engine/graphics/src/opengl/graphics_opengl_private.h
```diff
@@ -110,16 +110,17 @@ namespace dmGraphics
         uint32_t    m_ModificationVersion;
 
     };
-    // TODO: Why this one here!? Not used?
-    struct VertexBuffer
+
+    struct OpenglVertexAttribute
```
Jhonnyg, line 114: These are prefixed with OpenGL now so linking doesn't get screwed up if both vulkan and opengl are linked. I had some issues with this from constructors. We should probably rename all types to this format at some point

engine/graphics/src/dmsdk/graphics/graphics.h
```diff
@@ -397,6 +398,7 @@ namespace dmGraphics
         uint32_t        m_Size;
         Type            m_Type;
         bool            m_Normalize;
+        uint64_t        m_NameHash;
```
Jhonnyg, line 401: This was added so that we can map a buffer stream to a vertex declaration (in res_mesh) now that a buffer resource can be linked to a HBuffer directly that has been created without DDF data. It's a bit unfortunate that we exposed the VertexElement struct since this might be a breaking change. If we go this route, should we expose the DM_DECLARE_VERTEX_ELEMENT macro and make sure people use that as well?

engine/gamesys/proto/gamesys/buffer_ddf.proto
```diff
@@ -23,14 +23,14 @@ enum ValueType
 
 message StreamDesc
 {
-    required string name = 1;
+    required uint64 name_hash     = 1;
     required ValueType value_type = 2;
```
Jhonnyg, line 27: Engine only data so should be fine, we don't use the string name any more, only the hash

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1773,6 +1773,139 @@ static int SetSound(lua_State* L) {
...
+    int dummy_value_ptr = 0;
+    dmBufferDDF::StreamDesc buffer_stream_dummy_ddf = {};
+    buffer_stream_dummy_ddf.m_ValueType  = dmBufferDDF::VALUE_TYPE_INT8;
+    buffer_stream_dummy_ddf.m_ValueCount = 1;
+    buffer_stream_dummy_ddf.m_I.m_Data   = &dummy_value_ptr;
+    buffer_stream_dummy_ddf.m_I.m_Count  = 1;
+
+    dmBufferDDF::BufferDesc buffer_desc_dummy_ddf = {};
+    buffer_desc_dummy_ddf.m_Streams.m_Data = &buffer_stream_dummy_ddf;
+    buffer_desc_dummy_ddf.m_Streams.m_Count = 1;
```
Jhonnyg, line 1869: What's our alternatives here? We could add a m_Message pointer to the create/preload functions to the resource api, but do we really want that..

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1805,6 +1938,12 @@ static int GetBuffer(lua_State* L)
     void* resource = CheckResource(L, g_ResourceModule.m_Factory, path_hash, "bufferc");
 
     dmGameSystem::BufferResource* buffer_resource = (dmGameSystem::BufferResource*)resource;
+
+    if (!dmBuffer::IsBufferValid(buffer_resource->m_Buffer))
+    {
+        return luaL_error(L, "The buffer handle is invalid");
+    }
```
Jhonnyg, line 1945: if a reference to a buffer has been removed (e.g a LuaHBuffer object has GC'd), we should tell the user that

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -1620,12 +1622,21 @@ static uintptr_t GetExtProcAddress(const char* name, const char* extension_name,
...
-            GLint location = glGetAttribLocation(program, streams[i].m_Name);
+            int32_t location = -1;
+            for (int j = 0; j < program_ptr->m_Attributes.Size(); ++j)
+            {
+                if (program_ptr->m_Attributes[j].m_NameHash == streams[i].m_NameHash)
+                {
+                    location = program_ptr->m_Attributes[j].m_Location;
+                    break;
+                }
+            }
```
Jhonnyg, line 1638: The location for every attribute gets cached in the program now, which is one less GL call in the draw setup!

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/BufferBuilder.java
```diff
@@ -184,7 +185,7 @@ public void build(Task<Void> task) throws CompileExceptionError, IOException {
                 int streamValueCount = streamNode.get("count").asInt();
 
                 StreamDesc.Builder streamDescBuilder = StreamDesc.newBuilder();
-                streamDescBuilder.setName(streamName);
+                streamDescBuilder.setNameHash(MurmurHash.hash64(streamName));
```
JCash, line 188: Is there a corresponding change necessary in the Editor?

JCash (changes requested):
Only thing I see as an issue is the removal of the string from the stream declaration for buffers, since I think they help in debugging.

I do like that the hashing is done in the build step (as it should be)

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1773,6 +1773,139 @@ static int SetSound(lua_State* L) {
...
+/*# create a buffer resource
+ * This function creates a new buffer resource that can be used in the same way as any buffer created during build time.
+ * The function requires a valid buffer created from either [ref:buffer.create] or another pre-existing buffer resource.
+ * If the buffer has been created by a script, the new resource will take ownership of the buffer, meaning the buffer
+ * will not automatically be removed when the lua reference to the buffer is garbage collected. This behaviour can
+ * be overruled by specifying 'transfer_ownership = false' in the argument table.
+ *
+ * @name resource.create_buffer
+ *
+ * @param path [type:string] The path to the resource.
```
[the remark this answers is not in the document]
Jhonnyg: Yeah hm, good question.. Might have been an oversight from my part :( Should we change it?

engine/graphics/src/dmsdk/graphics/graphics.h
Jhonnyg: This is fixed since the stream declaration changes

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1773,6 +1773,140 @@ static int SetSound(lua_State* L) {
...
+    PreCreateResource(L, path_str, resource_ext, &canonical_path_hash);
+
+    luaL_checktype(L, 2, LUA_TTABLE);
+    lua_pushvalue(L, 2);
+
+    lua_getfield(L, -1, "buffer");
+    dmScript::LuaHBuffer* lua_buffer = dmScript::CheckBuffer(L, -1);
+    lua_pop(L, 1); // "buffer"
+
+    bool transfer_ownership = CheckFieldValue<bool>(L, -1, "transfer_ownership", true);
```
Jhonnyg, line 1850: Q: what do we want the default to be for this? I would think that the most common use case will be to create the buffer in a script and bind it to a resource that can be used in various places, where you probably want to always transfer the ownership. But maybe I'm wrong?

engine/gamesys/src/gamesys/components/comp_mesh.cpp
```diff
@@ -759,19 +775,20 @@ namespace dmGameSystem
...
+            }
+            else if (valueType != dmBuffer::VALUE_TYPE_FLOAT32 || count != 6)
             {
                 dmLogError("Invalid AABB metadata. Expected array of 6 floats.");
                 entry->m_Visibility = dmRender::VISIBILITY_FULL;
                 continue;
-            } else
-            if (r != dmBuffer::RESULT_OK)
+            }
+            else if (r != dmBuffer::RESULT_OK)
```
Jhonnyg, line 791: style fixes

engine/gamesys/src/gamesys/resources/res_mesh.cpp
```diff
@@ -26,124 +26,96 @@ namespace dmGameSystem
...
-            // case dmBufferDDF::VALUE_TYPE_UINT64:
-            // case dmBufferDDF::VALUE_TYPE_INT64:
+            // case dmBuffer::VALUE_TYPE_UINT64:
+            // case dmBuffer::VALUE_TYPE_INT64:
             default:
                 return dmGraphics::TYPE_BYTE;
         }
     }
 
-    static size_t StreamTypeToSize(dmBufferDDF::ValueType value_type)
```
Jhonnyg, line 71: not used anymore

engine/gamesys/src/gamesys/scripts/script_buffer.cpp
```diff
@@ -125,7 +125,7 @@ namespace dmGameSystem
         int                 m_BufferRef;// Holds a reference to the Lua object
     };
 
-    static inline dmBuffer::HBuffer UnpackLuaBuffer(dmScript::LuaHBuffer* lua_buffer)
+    dmBuffer::HBuffer UnpackLuaBuffer(dmScript::LuaHBuffer* lua_buffer)
```
Jhonnyg, line 128: used in script_resource now as well

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1819,19 +1819,21 @@ static int SetSound(lua_State* L) {
...
  * This function creates a new buffer resource that can be used in the same way as any buffer created during build time.
  * The function requires a valid buffer created from either [ref:buffer.create] or another pre-existing buffer resource.
- * By default, the new resource will take ownership of the buffer, meaning the buffer will not automatically be removed
+ * By default, the new resource will take ownership of the buffer lua reference, meaning the buffer will not automatically be removed
  * when the lua reference to the buffer is garbage collected. This behaviour can be overruled by specifying 'transfer_ownership = false'
- * in the argument table. Note that the path to the new resource must have the '.bufferc' extension,
- * meaning "/path/my_buffer" is not a valid path but "/path/my_buffer.bufferc" is. The path must also be unique, attempting to create
- * a buffer with the same name as an existing resource will raise an error.
+ * in the argument table. If the new buffer resource is created from a buffer object that is created by another resource,
+ * the buffer object will be copied and the new resource will effectively own a copy of the buffer instead.
```
JCash, line 1825: 👍

engine/gamesys/src/gamesys/scripts/script_resource.cpp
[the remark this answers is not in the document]
JCash: Nah, I think it's probably good to have the proper resource name for the resources.

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1927,6 +1942,48 @@ static int CreateBuffer(lua_State* L)
...
+        dmBuffer::HBuffer dst_buffer;
+        br = dmBuffer::Create(dst_count, streams_decl, stream_count, &dst_buffer);
+        free(streams_decl);
+
+        if (br != dmBuffer::RESULT_OK) {
+            return luaL_error(L, "Unable to create copy buffer: %s (%d).", dmBuffer::GetResultString(br), br);
+        }
+
+        // Copy supplied data to buffer
+        br = dmBuffer::Copy(dst_buffer, buffer);
```
JCash, line 1978: I honestly feel that we should have another Copy function that does all this work.
Would you mind adding such a function?

JCash (changes requested):
lgtm, but I think we want a new dmBuffer::Copy function

engine/gamesys/src/gamesys/scripts/script_resource.cpp
Jhonnyg: Yes I was thinking about that as well, I'll add it 👍

abadonna:
Do we need to call resource.release() to delete buffer resource when it's not needed any more?

Jhonnyg:
> Do we need to call resource.release() to delete buffer resource when it's not needed any more?

Calling a release on a resource will decrease the reference counter by 1, so it's up to you how you want to manage your resource. A reference will be owned by the collection the resource is create in and will automatically be deleted when the collection is destroyed, but if you want to remove a specific resource before that you can use resource.release.

PR #7229, opened 2022-12-19
Added overview of the engine input system

JCash:


engine/docs/INPUT_SYSTEM.md
```diff
@@ -0,0 +1,37 @@
...
+It may be a keyboard, mouse or a gamepad.
+
+Each platform has its own implementation of how the interaction is done with the hardware.
+
+### dmInput
+
+The input library uses the `dmHID` as the abstraction for the input.
+
+This library uses a `.gamepads` file for translating from the raw hardware gamepad information
+to how the Defold engine expects the data to be. E.g. which hardware button corresponds to which engine enum.
```
britzl, line 15: I don't think we should add manual line breaks to this kind of documentation. A markdown renderer should be able to generate flowing text without it. Also when translating from markdown to html there's less risk that a newline like this gets turned into a `</br>`. This section of text should be rendered as a single `<p>`aragraph and the browser should handle text flow.

engine/docs/INPUT_SYSTEM.md
```diff
@@ -0,0 +1,37 @@
...
+
+For a gamepad axis, the engine wants the final range to be in [-1,1].
+In order to get this behavior, the translation file may use scaling, negating and clamping to convert from hardware to engine values. The translation also supports deadxone values.
```
britzl, line 18: deadxone -> deadzone

britzl (changes requested):
Some typos and a comment about formatting

britzl (approved):

PR #7230, opened 2022-12-19
Editor: Added support for index buffers in vertex2.clj

JCash:
* Allow non-byte buffers in vertex2 and use IntBuffer for indices
* Use correct byte count when updating a vertex buffer
* Fix vertex2 count and offset when using a non-ByteBuffer


matgis (approved):
Looks good to me!

Issue #5447, opened 2021-01-04
Editor is very slow when one component instance or gui template is used in many places


notbadgun:
### Expected behaviour
I’m developing a physics game, and I make levels in Defold IDE. Each level is a collection with many pre-made game objects (obstacles, enemies, bonuses, walls). It’s a very convenient way to make levels because I can preview my levels right in the IDE and I don’t have to spend my time to make a special level editor.

How it looks:
![0bb150f22a0953e1ab65390010df5e031311d097](https://user-images.githubusercontent.com/13905665/103519438-9528df00-4e86-11eb-8857-4245f1d8a3d2.png)

My game will have 250+ levels. I made 60 levels, and it’s a pain to make more.

### Actual behaviour
IDE is very slow when there are many collections

### Steps to reproduce
For example, you can try to change something or to edit the main.script, i.e. even entering new text is lagging on my PC for 0.5-1 sec. PC has 8 core Ryzen 2700X.

**Demo project:**
[lot_of_collections1.zip](https://github.com/defold/defold/files/5763832/lot_of_collections1.zip)

**+1**
https://twitter.com/deulamco/status/1345740345900900352

<hr/>

<table>
<tr><td>Build time</td><td>2020-12-18T20:28:49.967999</td></tr>
<tr><td>Defold channel</td><td>editor-alpha</td></tr>
<tr><td>Defold editor sha</td><td>15990df1280c63c1744849d6456237c6411f593e</td></tr>
<tr><td>Defold engine sha</td><td>f7778a8f59ef2a8dda5d445f471368e8bd1cb1ac</td></tr>
<tr><td>Defold version</td><td>1.2.177</td></tr>
<tr><td>GPU</td><td>GeForce GTX 1080 Ti/PCIe/SSE2</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 NVIDIA 460.89</td></tr>
<tr><td>Java version</td><td>11.0.1+13</td></tr>
<tr><td>OS arch</td><td>amd64</td></tr>
<tr><td>OS name</td><td>Windows 10</td></tr>
<tr><td>OS version</td><td>10.0</td></tr>
<table>

subsoap:
I think this is part of a problem of non-localized state.

britzl:
From @matgis 

[...]

PR #7231, opened 2022-12-20
Improve editor performance when changing highly referenced nodes

vlaaad:
Technical changes:
- parallelize the computation of `basis-dependencies`;
- add caching layer for `basis-dependencies` for recent dependencies requests;
- remove `:cached` flag on some outputs;
- only return cached endpoints from `basis-dependencies` (and rename it reflect the new semantics).

Related to #5447

### DIFF
--- a/editor/src/clj/internal/graph.clj
+++ b/editor/src/clj/internal/graph.clj
@@ -19,8 +19,11 @@
             [internal.util :as util]
             [util.coll :refer [pair]])
   (:import [clojure.lang IPersistentSet]
+           [com.github.benmanes.caffeine.cache Cache Caffeine]
            [internal.graph.types Arc]
-           [java.util ArrayDeque HashSet]))
+           [java.util Set]
+           [java.util.concurrent ConcurrentHashMap ForkJoinPool TimeUnit]
+           [java.util.function Function]))
 
 ;; A brief braindump on Overrides.
 ;;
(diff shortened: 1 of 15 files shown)


editor/src/clj/internal/graph.clj
```diff
@@ -813,41 +816,71 @@
...
+    (.get basis-dependencies-cache
+          cache-key
+          (reify Function
+            (apply [_ _]
+              (loop [tasks (into [] endpoints->tasks-xf endpoints)]
+                (let [next-tasks
+                      (if (= 1 (count tasks))
+                        (into [] endpoints->tasks-xf ((nth tasks 0)))
+                        (into []
+                              (comp (mapcat deref) endpoints->tasks-xf)
```
matgis, line 879: Could this function composition be moved to the outer scope so it is not in the loop?

editor/src/clj/internal/system.clj
```diff
@@ -261,19 +261,19 @@
                                                       graph-after)
                                         system-after (if (and (has-history? system graph-id)
                                                               (meaningful-change? significantly-modified-graphs graph-id))
-                                                       (remember-change system graph-id graph-before graph-after (graph-id->outputs-modified graph-id #{}))
+                                                       (remember-change system graph-id graph-before graph-after (graph-id->cached-outputs-modified graph-id #{}))
```
matgis, line 264: It seems the `outputs-modified` from here are eventually used in a call to `ig/update-successors` inside the `is/time-warp` function. Is it OK for `ig/update-successors` to only consider the cached outputs and not all dependent outputs? Elsewhere we seem to involve non-cached outputs in `ig/update-successors`.

I saw that the field that holds the dependencies in `HistoryState` is named `cache-keys`, but it is being used with `ig/update-successors` and seems to have included non-cached outputs before this change. 😬

matgis (commented):
I have some concerns about us now only feeding cached outputs into `ig/update-successors` in the `time-warp` function during undo. Just wanted to make sure you were aware about this before approving.

editor/src/clj/internal/system.clj
vlaaad: I've been trying to understand how it works and it seems to it's safe to not update successors of uncached outputs. To my understanding, it might be safe to only update successors of `hydrated-outputs-to-refresh` and don't update cache-key successors at all, because successors for the graph are stored in the same graph that is preserved in the undo step, so undo swaps the graph and that restores successors. I added a test to validate that, but I would prefer if you had a look too to double-check I'm not making a mistake.

editor/src/clj/internal/graph.clj
```diff
@@ -829,58 +829,57 @@
...
+                                           (fn [endpoint]
+                                             (when (nil? (.putIfAbsent all-endpoints endpoint true))
+                                               (let [node-id (gt/endpoint-node-id endpoint)
+                                                     output (gt/endpoint-label endpoint)]
+                                                 (if-let [node (gt/node-by-id-at basis node-id)]
+                                                   (let [node-type (gt/node-type node)
+                                                         ^Set cached-outputs (.computeIfAbsent
+                                                                               node-type->cached-outputs
+                                                                               node-type
+                                                                               compute-cached-outputs-function)]
```
matgis, line 863: Now that `node-type->cached-outputs` is recreated in the `basis-dependencies-cache` cache miss function every time, it could be beneficial to instead make `in/cached-outputs` memoized  on the `deref`:ed node-type so we don't have to redo the work on every `basis-dependencies-cache` miss.

editor/test/internal/system_test.clj
```diff
@@ -488,6 +488,109 @@
...
+               (gt/endpoint v-sink :loud)}
+             (ts/graph-dependencies #{(gt/endpoint p-source :source-label)})))
+      ;; pipe2 isn't connected to anything
+      (is (= #{(gt/endpoint p-pipe2 :soft)}
+             (ts/graph-dependencies #{(gt/endpoint p-pipe2 :soft)})))
+      ;; cached dependencies still skip all pipes
+      (is (= #{(gt/endpoint p-source :_declared-properties)
+               (gt/endpoint v-sink :loud)}
+             (g/cached-dependencies (g/now) [(gt/endpoint p-source :source-label)]))))))
+
```
matgis, line 593: Great test!

matgis (approved):
Looks good to me! I found a small optimization opportunity, but I'm happy to merge without it.

PR #7236, opened 2022-12-21
Start language servers in the project folder

vlaaad:
We now start language servers in the project folder so it's possible to specify the launch command executable as a relative path.

Related to #3885

### DIFF
--- a/editor/src/clj/editor/lsp.clj
+++ b/editor/src/clj/editor/lsp.clj
@@ -438,7 +438,9 @@
   [lsp resource-node lines evaluation-context]
   (a/put! lsp (fn notify-lines-modified [state]
                 (let [resource (g/node-value resource-node :resource evaluation-context)]
-                  (sync-modified-lines-of-existing-node! state resource resource-node lines evaluation-context)))))
+                  (cond-> state
+                          (resource/file-resource? resource)
+                          (sync-modified-lines-of-existing-node! resource resource-node lines evaluation-context))))))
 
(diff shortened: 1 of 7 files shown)


editor/src/clj/editor/lsp.clj
```diff
@@ -438,7 +438,9 @@
   [lsp resource-node lines evaluation-context]
   (a/put! lsp (fn notify-lines-modified [state]
                 (let [resource (g/node-value resource-node :resource evaluation-context)]
-                  (sync-modified-lines-of-existing-node! state resource resource-node lines evaluation-context)))))
+                  (cond-> state
+                          (resource/file-resource? resource)
+                          (sync-modified-lines-of-existing-node! resource resource-node lines evaluation-context))))))
```
vlaaad, line 443: Now we don't try to sync/lint files from dependencies.

editor/src/clj/editor/lsp/server.clj
```diff
@@ -149,6 +152,9 @@
                          :else
                          (throw (ex-info "Invalid text document sync kind" {:value textDocumentSync})))})
 
+(defn- configuration-handler [{:keys [items]}]
+  (vec (repeat (count items) nil)))
```
vlaaad, line 156: Dummy configuration helper for lua-language-server that asks for configuration even though we don't define such capability.

matgis (approved):
Looks good to me!

PR #7238, opened 2022-12-22
Set texture float support

Jhonnyg:
Exposed floating point texture formats to the resource script API:

```
resource.TEXTURE_FORMAT_RGB16F
resource.TEXTURE_FORMAT_RGB32F
resource.TEXTURE_FORMAT_RGBA16F
resource.TEXTURE_FORMAT_RGBA32F
resource.TEXTURE_FORMAT_R16F
resource.TEXTURE_FORMAT_RG16F
resource.TEXTURE_FORMAT_R32F
resource.TEXTURE_FORMAT_RG32F
```
Note that the device running your game might not support all of these features, and the constants will only be exposed if it does. To check if the device can support a certain format, you can check if it is nil:

```
if resource.TEXTURE_FORMAT_RGBA16F ~= nil then
    -- it is safe to use this format
end
```


vlaaad (approved):
Editor changes are good!

JCash (approved):

PR #7243, opened 2022-12-28
Parallelize basis-dependencies calculation

vlaaad:
This change improves the performance of changing highly-referenced resources by around 30%.

Related to #5447


matgis (approved):
Looks good to me!

PR #7256, opened 2023-01-05
Fix for failing to pack macOS SDK

Lerg:
When there are more than one macOS SDK installed the script fails with garbage output. The fix is to limit the number of the "find" command results to one.


britzl (approved):
Looks good. Thanks!

PR #7268, opened 2023-01-11
Make bob extract plugin bin zips

vlaaad:
This changeset mirrors the editor behavior introduced in #7267

Related to #3885

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ExtenderUtil.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ExtenderUtil.java
@@ -15,6 +15,8 @@
 package com.dynamo.bob.pipeline;
 
 import static org.apache.commons.io.FilenameUtils.normalize;
+
+import com.dynamo.bob.bundle.BundleHelper;
 import org.apache.commons.io.filefilter.DirectoryFileFilter;
 import org.apache.commons.io.filefilter.RegexFileFilter;
 
@@ -37,6 +39,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
@@ -797,6 +800,13 @@ public class ExtenderUtil {
         }
     }
 
+    private static boolean isBinZip(String resourcePath) {
+        return resourcePath.endsWith("plugins/bin/common.zip") || Stream
+                .of(Platform.X86_64MacOS, Platform.X86_64Linux, Platform.X86_64Win32)
+                .flatMap(platform -> Stream.of(platform.getPair(), platform.getOs()))
+                .anyMatch(platformOption -> resourcePath.endsWith("plugins/bin/" + platformOption + ".zip"));
+    }
+
     public static void storeResources(File targetDirectory, List<IResource> resources) throws CompileExceptionError {
         for (IResource resource : resources) {
             String relativePath = resource.getPath();
@@ -814,7 +824,12 @@ public class ExtenderUtil {
             try {
                 byte[] data = resource.getContent();
                 FileUtils.writeByteArrayToFile(outputFile, data);
-                if (relativePath.startsWith("plugins/bin/") || relativePath.contains("/plugins/bin/")) {
+                // if file ends with plugins/bin/{platform}.zip, extract it too
+                if (isBinZip(relativePath)) {
+                    String fileName = outputFile.getName();
+                    File binDir = new File(outputFile.getParent(), fileName.substring(0, fileName.length() - 4));
+                    BundleHelper.unzip(new FileInputStream(outputFile), binDir.toPath());
+                } else if (relativePath.startsWith("plugins/bin/") || relativePath.contains("/plugins/bin/")) {
                     outputFile.setExecutable(true);
                 }
             } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


matgis (dismissed):
Looks good to me!

vlaaad:
I updated the code to be more like the editor's version, i.e. unzipping is recursive, so now it also marks unpacked plugins/bin files as executable.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ExtenderUtil.java
```diff
@@ -797,6 +803,13 @@ public static void storeResources(File targetDirectory, Map<String, IResource> r
         }
     }
 
+    private static boolean isBinZip(String resourcePath) {
+        return resourcePath.endsWith("plugins/bin/common.zip") || Stream
+                .of(Platform.X86_64MacOS, Platform.X86_64Linux, Platform.X86_64Win32)
+                .flatMap(platform -> Stream.of(platform.getPair(), platform.getOs()))
+                .anyMatch(platformOption -> resourcePath.endsWith("plugins/bin/" + platformOption + ".zip"));
```
JCash, line 810: Previously we've stuck with the platform name as the folder.
But I guess it makes sense to put them in the same folder then since they  themselves contain folder structures.

However, since they now are "generic", I'm not sure we should restrict us to the "bin" folder?
WHy not put them in the "plugins" folder, and then the zip file can contain files from any level in the folder structure.
E.g. what if the plugin contains .dll or .dylib that should go into the lib/ folder?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ExtenderUtil.java
```diff
@@ -797,6 +803,13 @@ public static void storeResources(File targetDirectory, Map<String, IResource> r
         }
     }
 
+    private static boolean isBinZip(String resourcePath) {
```
JCash, line 806: I'd favor a "isPLuginZip" instead.

JCash (changes requested):

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ExtenderUtil.java
vlaaad: The problem binzips are solving is that some binary executables require a lot of lua files alongside them on a disc to work. This messes with the editor and potentially makes bob compile and bundle those lua files to the game. `.dll` and `.dylib` files, however, don't mess with the editor, and as you said earlier, it's easier to work with files than with archives, so those can be stored in a lib folder as is, without being zipped.

With that said, do you think it still makes sense to make these archives more generic?

Btw, what do you mean by platform name? Is it only e.g. `macos` or `linux`, i.e. os name?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ExtenderUtil.java
```diff
@@ -803,49 +803,73 @@ public static void storeResources(File targetDirectory, Map<String, IResource> r
...
+    private static int getPluginZipSpecificity(String resourcePath) {
+        String baseName = FilenameUtils.getBaseName(resourcePath);
+        if (baseName.equals("common")) {
+            return 0;
+        } else if (baseName.contains("-")) {
+            return 2;
+        } else {
+            return 1;
+        }
+    }
```
JCash, line 822: 👍

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ExtenderUtil.java
```diff
@@ -803,49 +803,73 @@ public static void storeResources(File targetDirectory, Map<String, IResource> r
...
+            FileUtils.writeByteArrayToFile(outputFile, data);
+            String outputPath = outputFile.toString();
+            if (outputPath.contains("/plugins/bin/")) {
+                outputFile.setExecutable(true);
             }
+            return outputFile;
+        } catch (Exception e) {
+            throw new CompileExceptionError(resource, 0, e);
+        }
+    }
```
JCash, line 848: I wonder if we could somehow highlight these in the future.
e.g. 
1) show a list of unsigned/external executables that are downloaded from dependencies.
2) we store a cache with executable name/version and if it's a new version, the user has to accept "yes" in order to run it.

Might be overkill, but perhaps work talking about in the future.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ExtenderUtil.java
```diff
@@ -803,49 +803,73 @@ public static void storeResources(File targetDirectory, Map<String, IResource> r
...
 
+    public static void storeResources(File targetDirectory, List<IResource> resources) throws CompileExceptionError {
```
JCash, line 850: I don't see us excluding unzipping the unwanted platforms?
E.g. it seems odd to extract linux tools on a windows machine.

JCash (changes requested):
I wouldn't expect us to unpack desktop tools for a different host.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ExtenderUtil.java
vlaaad: Good catch, and we already don't unpack them in the editor.

JCash (approved):

PR #7276, opened 2023-01-13
Report defined globals in Lua configuration

vlaaad:
Related to #3885

### DIFF
--- a/editor/src/clj/editor/lua.clj
+++ b/editor/src/clj/editor/lua.clj
@@ -189,7 +189,7 @@
 (def defined-globals
   (into #{}
         (comp
-          (remove #(= :message (:type %)))
+          (remove #(#{:message :property} (:type %)))
           (map #(-> % :name (string/split #":" 2) first))
           (remove (conj base-globals "")))
         (get defold-docs "")))
--- a/editor/src/clj/editor/lua.clj
+++ b/editor/src/clj/editor/lua.clj
@@ -189,6 +189,7 @@
 (def defined-globals
   (into #{}
         (comp
+          (remove #(= :message (:type %)))
           (map #(-> % :name (string/split #":" 2) first))
           (remove (conj base-globals "")))
         (get defold-docs "")))
--- a/editor/src/clj/editor/lua.clj
+++ b/editor/src/clj/editor/lua.clj
@@ -182,7 +182,7 @@
 (def preinstalled-modules (into #{} (remove #{""} (keys defold-docs))))
 
 (def base-globals
-  (into #{}
+  (into #{"coroutine" "package" "string" "table" "math" "io" "file" "os" "debug"}
         (map #(.getName ^ScriptDoc$Element %))
         (get (load-sdoc (sdoc-path "base")) "")))
 
--- a/editor/src/clj/editor/lua.clj
+++ b/editor/src/clj/editor/lua.clj
@@ -190,8 +190,10 @@
   (into #{}
         (comp
           (remove #(#{:message :property} (:type %)))
-          (map #(-> % :name (string/split #":" 2) first))
-          (remove (conj base-globals "")))
+          (map :name)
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/lua.clj
```diff
@@ -181,6 +181,19 @@
...
+        (map #(.getName ^ScriptDoc$Element %))
+        (get (load-sdoc (sdoc-path "base")) "")))
+
+(def defined-globals
+  (into #{}
+        (comp
+          (remove #(= :message (:type %)))
+          (map #(-> % :name (string/split #":" 2) first))
+          (remove (conj base-globals "")))
+        (get defold-docs "")))
```
matgis, line 195: This appears to include some stuff from the Lua standard library, such as the `string` and `table` modules. Is that intentional?

I think this should report the same set of modules as we have in our API documentation, but currently this doesn't match. See here: https://defold.com/ref/stable/go/

Also, is is enough to just declare the modules like `go`? Don't we want it to declare its functions, like `go.set_position`, for example?

matgis (changes requested):
I don't think this is currently reporting the correct set of defined globals?

editor/src/clj/editor/lua.clj
vlaaad: Hmm good point, I wonder why we are getting them. 

Regarding functions — we don't need to define them, because luacheck does not try to validate that a function in a referenced module exists, only if a module itself exists. Surprising, yes, but it seems almost every lua linter works that way.

matgis (changes requested):
It's still not right, I'm afraid.

The list includes properties of game objects such as `position` and `rotation`, as well as callbacks that the user can choose to implement, like `on_reload`, `update` and `final`. I think you need to do a bit more sophisticated filtering of the `sdoc` files than you are currently doing.

vlaaad:
Argh, sorry, I'm rushing too fast with this. Yes, I needed to exclude properties. I think we should leave callback functions like `update`, `on_message` etc., otherwise, every script file will show warnings (as we currently show for `fixed_update` since it seems to be missing in the docs).
Screenshot 2023-01-17 at 14 18 21

vlaaad:
Okay, I carefully went through the globals we define and cross-checked with our [reference](https://defold.com/ref/stable/go/) and it seems we no longer have false positives.

matgis:
Why is that showing up as a warning, though? It's just a function definition. It it because it is not allowed to declare global functions or something?

matgis (approved):
I guess we can merge it like this?

matgis:
I don't think `client` and `master` exist as globals. They are just documented as methods on LuaSocket client or master object.

vlaaad:
Hmm, I saw e.g. [client:close](https://defold.com/ref/stable/socket/#client:close:) in the reference. Maybe I don't know Lua enough, is it not a top-level thing?

vlaaad:
I think Luacheck warns on defining new globals by default, yes.

vlaaad:
Btw, I found out that luacheck allows configuring globals with included fields — https://luacheck.readthedocs.io/en/stable/config.html#custom-sets-of-globals

The problem is that `lua-language-server` only accepts a flat list of strings for globals, and it uses the globals configuration in other places too. I'll create an issue on our fork.

editor/src/clj/editor/lua.clj
britzl: It is actually possibly to also define available functions on a module. Maybe something to add in a future update?

vlaaad:
Created an issue https://github.com/defold/lua-language-server/issues/8

editor/src/clj/editor/lua.clj
vlaaad: Yes, I created an issue for that https://github.com/defold/lua-language-server/issues/8

PR #7279, opened 2023-01-16
Use plugins zips instead of bin zips

vlaaad:
As discussed with Mathias, we might want to pack other stuff like libs into zip files. This changeset changes the plugins' zip archives extraction from:
```
${ext}/plugins/bin/${platform}.zip -> /build/plugins/${ext}/plugins/bin/${platform}/
```
to:
```
${ext}/plugins/${platform}.zip -> /build/plugins/${ext}/plugins/
```

Additionally, we now make sure to extract archives in their specificity order (x86_64-macos > macos > common) and register jars/libs from those archives.

Related to #3885


editor/src/clj/editor/workspace.clj
```diff
@@ -450,22 +457,25 @@ ordinary paths."
...
+    (some? (some #(string/ends-with? path %) (keys plugin-zip-names)))))
 
-(defn- unpack-bin-zip! [workspace resource]
+(defn- unpack-plugin-zip! [workspace resource]
   {:pre [(string/ends-with? (resource/proj-path resource) ".zip")]}
+  (unpack-resource! workspace resource)
   (let [proj-path (resource/proj-path resource)
         plugin-file (plugin-path workspace proj-path)
-        infix-path (subs proj-path 0 (- (count proj-path) 4))]
+        infix-path (resource/parent-proj-path proj-path)]
```
matgis, line 478: This appears to no longer use the name of the zip archive. Is this correct?

Previously it just stripped away the `.zip` extension, but now it strips away the entire file name.

editor/src/clj/editor/workspace.clj
```diff
@@ -475,16 +485,19 @@ ordinary paths."
...
-        (unpack-resource! workspace x)
-        (cond
-          (bin-zip? x) (unpack-bin-zip! workspace x)
-          (jar-file? x) (register-jar-file! workspace x)
-          (shared-library? x) (register-shared-library-file! workspace x))
-        (catch FileNotFoundException _
-          (throw (IOException. "\nExtension plugins needs updating.\nPlease restart editor for these changes to take effect!")))))))
+  (let [{plugin-zips true resources false} (->> changed
+                                                (eduction (filter is-plugin-file?))
+                                                (group-by plugin-zip?))]
```
matgis, line 490: Maybe use `filter` without the `eduction` for clarity?

editor/test/integration/engine/native_extensions_test.clj
```diff
@@ -43,12 +43,12 @@
...
 (deftest unpack-bin-zip-test
-  (testing "${ext}/plugins/bin/${platform}.zip is extracted to /build/plugins/${ext}/plugins/bin/platform/ folder"
+  (testing "${ext}/plugins/${platform}.zip is extracted to /build/plugins/${ext}/plugins/ folder"
    (with-clean-system
      (let [workspace (test-util/setup-workspace! world "test/resources/extension_project")
            _ (test-util/setup-project! workspace)
            root (g/node-value workspace :root)]
-       (is (.exists (io/file (str root "/ext_with_bin_zip/plugins/bin/x86_64-macos.zip"))))
+       (is (.exists (io/file (str root "/ext_with_bin_zip/plugins/x86_64-macos.zip"))))
        (is (not (.exists (io/file (str root "/ext_with_bin_zip/plugins/bin/x86_64-macos/lsp.editor_script")))))
```
matgis, line 52: Should this next path include `/bin/` now that you've stripped it from the line above?

In fact, this test looks suspect to me. It seems the `testing` description does not match what we check for? Are you sure this is correct? If so, I think it needs to be clarified somewhat with some comments.

matgis (changes requested):
I'm not sure I understand fully, but the `unpack-bin-zip-test` seems a bit suspect to me now.

Also, do we need to target this to `dev`, seeing as there are corresponding changes in Bob?

editor/src/clj/editor/workspace.clj
vlaaad: Yes, it's correct. As described in the commit and PR description, we use the following unpack rule:
```
${ext}/plugins/${platform}.zip -> /build/plugins/${ext}/plugins/
```
as suggested by @JCash. That way we can pack both `lib/${platform}` and `bin/${platform}` resources in a single archive.

editor/test/integration/engine/native_extensions_test.clj
vlaaad: No, the unchanged path is correct. Previously, we had `plugins/bin/x86_64-macos.zip` that contained a single file at the root (`/lsp.editor_script`). Now we have `plugins/x86_64-macos.zip` that contains a file at a different path (`/bin/x86_64-macos/lsp.editor_script`), so the result should be the same, hence the only change in test is the contents and location of the zip file.

vlaaad:
We don't need to target this to `dev` because this behavior is only relied upon by the editor. We have a separate PR for bob that targets dev that mirrors the behavior of this code, but it will still be unused since bob does not use unpacked binaries — the initial release for LSP is editor-only.

editor/test/integration/engine/native_extensions_test.clj
matgis: I think adding a comment detailing the file structure of the zip file would help.

matgis (approved):

editor/test/integration/engine/native_extensions_test.clj
vlaaad: Good point, I'll add it!

PR #7284, opened 2023-01-17
Correctly handle paths on win32

vlaaad:
Related to #3885

### DIFF
--- a/editor/src/clj/editor/fs.clj
+++ b/editor/src/clj/editor/fs.clj
@@ -292,12 +292,17 @@
   (cond-> (-> entry io/as-file .toPath .normalize .toAbsolutePath .toString)
           (not case-sensitive?) .toLowerCase))
 
+(def separator-char File/separatorChar)
+
 (defn below-directory?
   "Returns true if the file system entry is located below the directory."
   [^File entry ^File directory]
   (let [entry-path (comparable-path entry)
         directory-path (comparable-path directory)]
-    (and (= \/ (get entry-path (count directory-path)))
+    ;; Note: can't inline File/separatorChar here, otherwise during AOT this
+    ;; code will throw exception that no overload of clojure.lang.Util/equiv is
+    ;; found for char and Object
+    (and (= separator-char (get entry-path (count directory-path)))
          (string/starts-with? entry-path directory-path))))
 
 (declare copy-directory!)
--- a/editor/src/clj/editor/lsp/server.clj
+++ b/editor/src/clj/editor/lsp/server.clj
@@ -14,6 +14,7 @@
 
 (ns editor.lsp.server
   (:require [clojure.core.async :as a :refer [<! >!]]
+            [clojure.java.io :as io]
             [clojure.spec.alpha :as s]
             [clojure.string :as string]
             [dynamo.graph :as g]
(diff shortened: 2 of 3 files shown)


editor/src/clj/editor/fs.clj
```diff
@@ -292,12 +292,17 @@
...
           (not case-sensitive?) .toLowerCase))
 
+(def separator-char File/separatorChar)
+
 (defn below-directory?
   "Returns true if the file system entry is located below the directory."
   [^File entry ^File directory]
   (let [entry-path (comparable-path entry)
         directory-path (comparable-path directory)]
-    (and (= \/ (get entry-path (count directory-path)))
```
vlaaad, line 300: On windows, it was `\`, not `/`

editor/src/clj/editor/lsp/server.clj
```diff
@@ -67,13 +68,21 @@
...
   Map
   (launch [{:keys [command]} directory]
-    (.start
-      (doto (ProcessBuilder. ^"[Ljava.lang.String;" (into-array String command))
-        (.redirectError ProcessBuilder$Redirect/INHERIT)
-        (.directory directory)))))
+    {:pre [(vector? command)]}
+    ;; We need to resolve command in addition to setting ProcessBuilder's directory,
+    ;; since the latter only affects the current directory of the spawned process,
+    ;; and does not affect executable resolution
```
vlaaad, line 74: I have no idea how relative paths worked on macOS, but on Windows, I needed to specify the full path for the process to start.

editor/src/clj/editor/fs.clj
matgis: Good catch. Surprised to learn this wasn't caught by the editor tests, since there are pretty extensive tests for this function. Don't we run the tests on Windows? 😬

editor/src/clj/editor/lsp/server.clj
```diff
@@ -67,13 +68,21 @@
...
+    (let [resolved-command (update command 0 #(workspace/to-absolute-path (str directory) %))]
```
matgis, line 75: Are you aware `workspace/to-absolute-path` doesn't actually return an absolute path? It returns a project-relative path from a resource-relative path. I hate this.

matgis (commented):
The use of `workspace/to-absolute-path` makes me nervous., because I think it returns an "absolute path inside the project", i.e. a project-relative path.

If you need the absolute path, I recommend using the methods of the `File` class, and using `(io/file)` to join path segments. 

This stuff really needs to be cleaned up at some point. 😞

editor/src/clj/editor/lsp/server.clj
vlaaad: What? It returns an absolute path if the supplied path is relative, and the original path if the supplied path is absolute, which is what we want here:
```clj
(to-absolute-path "/path/to/project/root" "bin/foo.sh")
=> "/path/to/project/root/bin/foo.sh"
(to-absolute-path "/path/to/project/root" "/bin/foo.sh")
=> "/bin/foo.sh"
```

editor/src/clj/editor/fs.clj
vlaaad: Yeah, we have good tests for this, but they are not run on Windows :)

vlaaad:
@matgis I replaced the use of `to-absolute-path` with java `Path::resolve`

matgis (commented):
Looks good now!

vlaaad:
I traveled back in time and merged the PR after seeing your review 👍

Issue #5684, opened 2021-03-24
Add a checkbox to not open the resulting folder when finished bundling


subsoap:
Sometimes we need to rebundle from the editor over and over and it's annoying to have the same folder open constantly. A checkbox to open the resulting folder when bundling would be nice.

Adding an option to play a sound / show a toast in the editor settings when bundling finishes would also be useful.

britzl:
> Sometimes we need to rebundle from the editor over and over and it's annoying to have the same folder open constantly. A checkbox to open the resulting folder when bundling would be nice.

Oh yes, I totally agree. It annoys me as well :-)

PR #7305, opened 2023-01-20
Added option in preferences to prevent bundle target opening

AGulev:
Now it's possible to prevent default behaviour when bundle folder opens after bundling process is done.

Fix https://github.com/defold/defold/issues/5684

## PR checklist

* [ ] Code
* [ ] Documentation

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -2143,7 +2143,7 @@ If you do not specifically require different script states, consider changing th
                            (fn [successful?]
                              (when successful?
                                (if (some-> output-directory .isDirectory)
-                                 (when (prefs/get-prefs prefs "open-bundle-target-folder" false) (ui/open-file output-directory))
+                                 (when (prefs/get-prefs prefs "open-bundle-target-folder" true) (ui/open-file output-directory))
                                  (dialogs/make-info-dialog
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/app_view.clj
```diff
@@ -2143,7 +2143,7 @@ If you do not specifically require different script states, consider changing th
                            (fn [successful?]
                              (when successful?
                                (if (some-> output-directory .isDirectory)
-                                 (ui/open-file output-directory)
+                                 (when (prefs/get-prefs prefs "open-bundle-target-folder" false) (ui/open-file output-directory))
```
matgis, line 2146: The `false` argument here is being used when there is no preference set. It should be `true` here as well since the default is `true` in `prefs_dialog.clj` below.

Otherwise, users who haven't touched the preference will see it as checked, but it will behave as if it is unchecked.

matgis (changes requested):
It works, but we need so make the default `true` when getting the preference value.

editor/src/clj/editor/app_view.clj
AGulev: Got it, thx.
Fixed.

matgis (approved):
Looks good to merge!

PR #6902, opened 2022-08-22
Shader include support

Jhonnyg:
Added support for *#include* pragmas in shader files. This feature can be used to separate shader functions into modules that can be used across several shader files, which should help out with reusing code from your own project, or from asset extensions.

Fixes #6892 

## PR checklist (remove before merging)

* [x] Prepared the PR for release notes
	* [x] Proper description of feature or fix (see example x?)
	* [x] Added a #fixes ### if it fixes an issue
	* [x] Assigned issue to a project
* [-] Added documentation for public API changes
* [x] Is this a new feature?
	* [x] Added engine and/or editor unit tests
	* [x] Added or updated manual entry (https://github.com/defold/doc/pull/316)

PR #7313, opened 2023-01-24
Editor: Ensure the workspace root path is canonical

matgis:
Fix failing `custom-resources-cached` build-test on macOS due to a non-canonical workspace root path from temp directory messing up our new implementation of `resource/resolve-resource`.

Regression introduced in https://github.com/defold/defold/pull/6902.

### DIFF
--- a/editor/src/clj/editor/workspace.clj
+++ b/editor/src/clj/editor/workspace.clj
@@ -796,7 +796,7 @@ ordinary paths."
         (g/transact
           (g/make-nodes graph
             [workspace [Workspace
-                        :root project-path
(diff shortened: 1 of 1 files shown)


vlaaad (approved):
LGTM! 

Side note: shouldn't it be an error instead if we pass a non-canonical path into `make-workspace`?

matgis:
I think we can launch the editor from the command-line, and that point, the path might not be canonical either. I figured we might at well be liberal in what we accept and "correct" it.

vlaaad:
Very good gummisnodd!

PR #7316, opened 2023-01-24
Do not scan`build` folder

### LINKED ISSUE #7311: LSP lua scripts extracted to build/plugins are built by bob
**Describe the bug:**
Attempt to bundle the project with the LSP linter dependency (https://github.com/defold/lua-language-server/releases/download/v0.0.3/release.zip) fails because bob tries to build lua files needed for the language server binary that are extracted to `/build/plugins`.

**To Reproduce:**
1. Download and open the repro project [Mobile game lsp repro.zip](https://github.com/defold/defold/files/10487785/Mobile.game.lsp.repro.zip) (it's a mobile game template with a dependency on linting LSP)
2. Bundle
3. See a lot of lines like `Writing 64-bit bytecode without delta for build/plugins/lsp-lua-language-server/plugins/bin/x86_64-macos/locale/en-us/meta.lua`, then build fails because some lua requires are not found, and some syntax is incorrect (I assume the Lua language server uses newer version of Lua)

**Expected behavior:**
Bob builds only user code

**Defold version:**
  - Version [e.g. 1.4.1]
  - editor sha cb8b76d36177dbe7fefab1b4a2fa3f93fc1bd6bd

**Platforms:**
all

**Workaround:**
Add `build/plugins` to `.defignore` file.

**Additional context**
Related to #3885

AGulev:
Ignore `build` folder when scanning the root folder for files that should be built.

Fix https://github.com/defold/defold/issues/7311
## PR checklist

* [ ] Code
* [ ] Documentation

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
@@ -603,8 +603,10 @@ public class Bob {
         }
         project.mount(new ClassLoaderResourceScanner());
 
-        Set<String> skipDirs = new HashSet<String>(Arrays.asList(".git", project.getBuildDirectory(), ".internal"));
+        Set<String> skipDirs = new HashSet<String>(Arrays.asList(".git", project.getBuildDirectory(), ".internal", "build"));
+        TimeProfiler.start("findSources");
         project.findSources(sourceDirectory, skipDirs);
+        TimeProfiler.stop();
     }
 
     private static void validateChoices(String optionName, String value, List<String> validChoices) {
@@ -704,8 +706,10 @@ public class Bob {
 
         project.loadProjectFile();
 
+        TimeProfiler.start("setupProject");
         // resolves libraries and finds all sources
         setupProject(project, shouldResolveLibs, sourceDirectory);
+        TimeProfiler.stop();
 
         if (!cmd.hasOption("defoldsdk")) {
             project.setOption("defoldsdk", EngineVersion.sha1);
--- a/editor/src/clj/editor/pipeline/bob.clj
+++ b/editor/src/clj/editor/pipeline/bob.clj
@@ -53,7 +53,7 @@
   (catch Exception e
     (throw (ex-info "Failed to set verbose logging field in Bob." {} e))))
 
-(def skip-dirs #{".git" "build/default" ".internal"})
+(def skip-dirs #{".git" "build" ".internal" })
 (def html5-url-prefix "/html5")
 (def html5-mime-types {"js" "application/javascript",
                        "json" "application/json",


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -603,8 +603,10 @@ private static void setupProject(Project project, boolean resolveLibraries, Stri
         }
         project.mount(new ClassLoaderResourceScanner());
 
-        Set<String> skipDirs = new HashSet<String>(Arrays.asList(".git", project.getBuildDirectory(), ".internal"));
+        Set<String> skipDirs = new HashSet<String>(Arrays.asList(".git", project.getBuildDirectory(), ".internal", "build"));
```
AGulev, line 606: Actual fix: add both "build" and custom build folder may be specified with "-o"

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -603,8 +603,10 @@ private static void setupProject(Project project, boolean resolveLibraries, Stri
...
+        TimeProfiler.start("findSources");
         project.findSources(sourceDirectory, skipDirs);
+        TimeProfiler.stop();
```
AGulev, line 609: it is interesting to know how long it takes

britzl (dismissed):

vlaaad:
Not sure this PR will fix the issue because the editor does not use the Bob class, instead it only uses the Project class, therefore `skipDirs` will not be set by the editor.

Do you think this is the solution, e.g. as opposed to prepopulating `excludeFolders` in `createTasks`?

Note: `editor.pipeline.bob` ns defines its own `skip-dirs`:
```clj
(def skip-dirs #{".git" "build/default" ".internal"})
```

AGulev:
@vlaaadyes, the editor does exactly the same, but specify its own `skipDirs`
In the editor case, it's impossible to change build folder, so I think having it as :
```
(def skip-dirs #{".git" "build" ".internal" })
```
is fine

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -53,7 +53,7 @@
   (catch Exception e
     (throw (ex-info "Failed to set verbose logging field in Bob." {} e))))
 
-(def skip-dirs #{".git" "build/default" ".internal"})
+(def skip-dirs #{".git" "build" ".internal" })
```
AGulev, line 56: since it's impossible to change build folder from the editor we can replace "build/default" with "build"

vlaaad: LGTM! I can probably copy this change to `editor-dev` branch

AGulev: sure, but it will be just a half of the fix, because it is broken when build from bob (without the editor) as well

vlaaad: Yes, you are right

vlaaad (approved):

PR #7347, opened 2023-02-01
Improve GUI modification performance by 15%

vlaaad:
User-facing changes: the editor is more performant when modifying highly-referenced GUI nodes

Note: not sure how to make it easy to verify the changes, but when working on this PR I went through all GuiSceneNodes in a big project and compared their `:rt-pb-msg` and `(comp pb-msg->pb-rt-msg :pb-msg)` outputs and worked until there were no important differences. The remaining differences:
- spine GUI node's `:spine-skin` is no longer set to `""` if it was `nil`. Seems to work fine;
- spine GUI node was unnecessarily reordered after its children.

Related to #5447

### DIFF
--- a/editor/src/clj/editor/gui.clj
+++ b/editor/src/clj/editor/gui.clj
@@ -371,7 +371,6 @@
    [:node-outline :child-outlines]
    [:scene :child-scenes]
    [:node-msgs :node-msgs]
-   [:node-rt-msgs :node-rt-msgs]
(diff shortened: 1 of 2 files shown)


matgis (approved):
Seems fine! Presumably we can now remove the `node-rt-msgs` output in `extension-spine` as well?

vlaaad:
Yes, I'll do it today.

PR #7261, opened 2023-01-09
Improve the time it takes to save soon after opening the editor

matgis:
Improves #7175.
Will also need https://github.com/defold/extension-spine/pull/113 to be merged (done!).
The new version of `extension-spine` contains a runtime change, so this editor-only PR must target `dev`.

### User-facing changes
* Improved the time it takes to save the project soon after opening it in the editor.
* Breaking change: This change requires an updated version of `extension-spine` due to changes to the `GuiNode` interface in the editor.

### Technical changes
* We no longer encode the size of auto-sized images in sprites or gui elements that refer it in the protobuf messages we output for saving or building. The runtime will use the dimensions of the assigned image regardless of what we write to the protobuf binaries.
* We now check that GuiNodes registered using `gui/register-node-type-info!` implement all the required outputs when loading the plugin.

PR #7352, opened 2023-02-01
Editor: Use the node-specific manual-size default for gui shapes with :size-mode-auto

matgis:
Fixes a regression introduced by #7261.

### User-facing changes
* Auto-sized Gui shapes without a texture assigned retain their default size of 200 x 100 pixels.


editor/src/clj/editor/gui.clj
```diff
@@ -837,7 +837,7 @@
 
 (defn- strip-size-from-node-desc [node-desc]
   (-> node-desc
-      (assoc :size [0.0 0.0 0.0 0.0])
+      (dissoc :size)
```
matgis, line 840: Instead of explicitly setting the manual size to zero, we clear it out. This results in the property not being assigned when loading, so the default value of 200 x 100 pixels is used for the `manual-size` property.

vlaaad (approved):
YOLO

editor/src/clj/editor/gui.clj
```diff
@@ -835,9 +835,9 @@
 (defn- add-size-to-overridden-fields [overridden-fields]
   (vec (sort (conj overridden-fields size-prop-index))))
 
-(defn- strip-size-from-node-desc [node-desc]
+(defn- strip-size-from-shape-base-node-msg [node-desc]
   (-> node-desc
-      (assoc :size [0.0 0.0 0.0 0.0])
+      (assoc :size [200.0 100.0 0.0 1.0]) ; Default size for ShapeNode. The runtime uses this size if there is no texture applied.
```
matgis, line 840: The editor is fine with stripping out the `:size`, but it seems we still need to write it to the files for Bob and the runtime.

vlaaad (approved):
Works great!

PR #7354, opened 2023-02-01
Added Void Scrappers and Monkey Mart to editor splash screen

britzl:
Also stripped some metadata from existing images.


AGulev (approved):
❤️

PR #7358, opened 2023-02-02
Improve the editor's text layout performance

### LINKED ISSUE #7187: Editor works extremely slow when trying to open font with a big count of glyphs
The editor becomes unresponsible when open a font with big count of glyphs.
1. Open the attached project
2. Double click `KlokanTechNotoSansCJK-Bold.font` file
[SlowFontEditorRepo.zip](https://github.com/defold/defold/files/10104052/SlowFontEditorRepo.zip)

vlaaad:
User-facing changes: fonts with many glyphs will render faster, also the scene will not lag on zooming/panning. Time improvement: from 12s to 1ms.

Fixes #7187

### DIFF
--- a/editor/src/clj/editor/font.clj
+++ b/editor/src/clj/editor/font.clj
@@ -160,46 +160,119 @@
       :defold)
     :distance-field))
 
-(defn- measure-line [glyphs text-tracking line]
-  (let [w (reduce + (- text-tracking) (map (fn [c] (+ text-tracking (get-in glyphs [(int c) :advance] 0))) line))]
-    (if-let [last (get glyphs (and (last line) (int (last line))))]
+(defn- measure-line [glyphs text-tracking ^String line]
+  (let [w (transduce (comp
+                       (map #(:advance (glyphs (int %)) 0.0))
+                       (interpose text-tracking))
+                     +
+                     0.0
+                     line)
+        len (.length line)]
+    (if-let [last (get glyphs (and (pos? len) (int (.charAt line (dec len)))))]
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/font.clj
```diff
@@ -160,46 +160,119 @@
...
+                          (conj! lines "")
+                          (let [last-non-ws-index
+                                ;; Note: normally, this loop would be a subject to IOOB
+                                ;; exception, but here we have an invariant that
+                                ;; the accumulated string in sb is either empty or does
+                                ;; not start with a white space char
+                                (loop [i (dec line-length)]
+                                  (let [ch (.charAt sb i)]
+                                    (if (or (Character/isWhitespace ch)
+                                            (= zero-width-space ch))
```
vlaaad, line 200: This `or` is repeated a couple of times in the code. I was thinking of moving it to a standalone function, but then we would have char boxing, which would hurt the performance.

matgis (approved):
Looks like a great improvement!

In my testing, the result differs slightly from the original algorithm. Initially I considered this a big problem, but it turns out the original algorithm does not match the runtime result when building either.

I think we need to put the effort in to make it match the runtime at some point, but I think we can merge this improvement for now seeing as how it was already not matching the runtime result.

vlaaad:
Hmm, good point. I only tried to match the behavior of the old editor algorithm, deviating where it makes sense (it didn't strip trailing lines if they only contained whitespace), but we should really strive for compatibility with the engine here. Maybe you can make an issue for that and describe the differences there?

PR #7364, opened 2023-02-03
Fix text layout regression

vlaaad:
The issue was introduced during performance text layout optimization because I misunderstood how glyph advance, left-bearing, and width relate to each other. The old behavior checked if the last character's left-bearing+width go past the max line length limit, and if it didn't, increased the line by advance. I misunderstood it as if the advance is usually larger than left-bearing+width, while in reality, it's the opposite. Now I changed the new algorithm to match the old behavior where it uses the last character's left-bearing+width for comparison with the line max width.

Related to #7187


matgis (approved):
Ship it!

PR #7372, opened 2023-02-08
Improve resource node modification perf by ~3-5%

vlaaad:
This change removes the `reload-dependencies` output and inlines its only place of use. For highly-referenced scripts, the improvement is 5%. For GUI nodes, it's 3%.

Related to #5447


matgis (approved):
Looks good to me!

PR #7376, opened 2023-02-10
Add a dev tool to find dangling graph outputs

vlaaad:
This tool helps to find outputs that are not connected to anything so we can potentially remove them. Unfortunately, it seems we don't have dangling outputs that contribute to successors of highly referenced nodes. Besides `sha256`, that is. But it seems we need it in the graph for caching. Regardless, it highlights some outputs that we might want to remove in the future to not clutter the code (e.g. `breakpoint-rows`, `gui-scene`).

Related to #5447

### DIFF
--- a/editor/src/clj/dev.clj
+++ b/editor/src/clj/dev.clj
@@ -13,7 +13,8 @@
 ;; specific language governing permissions and limitations under the License.
 
 (ns dev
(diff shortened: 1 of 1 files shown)


matgis (approved):
LGTM!

PR #7381, opened 2023-02-13
Ignore errors thrown in __tostring

### LINKED ISSUE #7168: The editor doesn't show call stack in debbugger in some cases
On a huge project if I add a breakpoint in some core-heavy lua file, the editor things a bit and then nothing in the main view.

If I check menu options the debbugger works (I can even `step over`):
image

But I don't see debugger cursor in code, callstack and changes in the interface:
image
image

**To Reproduce (REQUIRED)**
1.  add breakpoint in `update()` function in a heavy script or gui_script function

  - 1.4.0 

**Platforms (REQUIRED):**
 - Platforms: Macos
 - OS: Macos

vlaaad:
If an exception is thrown in the `__tostring` method of a serialized Lua object, the editor's debugger will not receive the state at a breakpoint. We now silently ignore such errors and serialize the string as `???`.

User-facing changes: the debugger will serialize Lua objects even if they throw errors in `__tostring`.

Fixes #7168

### DIFF
--- a/engine/engine/content/builtins/scripts/edn.lua
+++ b/engine/engine/content/builtins/scripts/edn.lua
@@ -35,17 +35,20 @@ local escape_char_map = {
 }
 
 local function str(val)
-  local mt = debug.getmetatable(val)
-  if mt then 
-    if mt.__metatable == NodeProxy then
-      return string.format("%s%p", val, val)
-    elseif type(val) == "table" and mt.__tostring then
-      return string.format("%s %p", val, val)
+  local success, error_or_result = pcall(function()
+    local mt = debug.getmetatable(val)
+    if mt then
+      if mt.__metatable == NodeProxy then
+        return string.format("%s%p", val, val)
+      elseif type(val) == "table" and mt.__tostring then
+        return string.format("%s %p", val, val)
+      end
+      return tostring(val)
+    else
+      return tostring(val)
     end
-    return tostring(val)
-  else
-    return tostring(val)
-  end
+  end)
+  return success and error_or_result or '???'
 end
 
 local function escape_char(c)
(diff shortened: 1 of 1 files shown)


AGulev:
what kind of error it is? why is it happening?

vlaaad:
It is an error in `__tostring`, where nil is concatenated into a string. I found this place in a big project, it is something in custom `__tostring` that loops over table items, and after the fix when I managed to get the dump of variables on the stack, I found the offending object and it was a table with a single key `2`. 

There is a stack trace in the console output:
```
DEBUG:SCRIPT: stack traceback:
	builtins/scripts/edn.lua:171: in function '__concat'
	[redacted by vlaaad]:46: in function 'add_map_scalar_vector'
	[redacted by vlaaad]:122: in function <[redacted by vlaaad]:116>
	[C]: in function 'str'
	builtins/scripts/edn.lua:124: in function 'encode_table'
	builtins/scripts/edn.lua:132: in function 'encode_refs'
	builtins/scripts/edn.lua:155: in function <builtins/scripts/edn.lua:151>
	[C]: in function 'xpcall'
	builtins/scripts/edn.lua:166: in function <builtins/scripts/edn.lua:164>
	[C]: in function 'pcall'
	builtins/scripts/mobdebug.lua:1045: in function <builtins/scripts/mobdebug.lua:797>
```

AGulev:
hmmm
If it’s an error in the user’s script, it feels like we should show an error to the user and the user should fix it then.

I feel bad that we will add pcall (which is pretty slow) in such place and affect speed for all our users just because the user has errors in their metatables (which may “fire” in some other places)

That would be nice to compare performace for both: with and without pcall (maybe I’m over dramatizing)

vlaaad:
I measured the performance before/after the pcall change, the average time is:
- before pcall: 1.918561s
- after pcall: 2.019249s
This is an increase of 5%.

I agree we should show the error to the user, I'll add a print for it

AGulev (dismissed):
We still need to show an error to the end user. 
Other than that, it's fine.

AGulev (approved):
LGTM!

Issue #6604, opened 2022-05-14
Possibility to run build from the external code editor.


AGulev:
**Is your feature request related to a problem? Please describe (REQUIRED):**
When a developer uses an external code editor he has to switch back to the editor after every change he wanna test in the build.
That would be nice to have a possibility to ask editor read changes and build game without changing active window.

**Describe the solution you'd like (REQUIRED):**
The editor may receive a command from an external code editor using the editor's web server. 

**Describe alternatives you've considered (REQUIRED):**
none

britzl:
I think it is also worthwhile to investigate how to speed up a bob build.

AGulev:
>I think it is also worthwhile to investigate how to speed up a bob build.

yep, it's another issue. I'll create a ticket, now when we have a nice time report that should help to figure out what's going on there

mikatuo:
Is this the editor's web server? https://github.com/defold/defold/blob/dev/editor/src/clj/editor/boot_open_project.clj#L179-L184

mikatuo:
Hello,
Will you accept a PR for this task? Have you had a chance to think about the design and requirements?

How I see it:

Rather than adding new endpoints for each command ("Build", "Rebuild", "Hot Reload", etc) add new single endpoint that will be accepting different "commands".
Just to be safe it might be better to start a second web server to keep it separated from the existing one. This way it will not affect the existing web server in any way.
The first command that it will support might be "Build". Later "handlers" for new commands could be added.
I suggest to accept JSON since it will be much easier to work with. In example, the new endpoint could also be called from `.editor_script` files.
Another big question I have is can it be written in Java?

britzl:
> Will you accept a PR for this task?

Yes, absolutely! But we need a design discussion first! You have some suggestions that we need to discuss and agree upon before moving on with the implementation!

> Rather than adding new endpoints for each command ("Build", "Rebuild", "Hot Reload", etc) add new single endpoint that will be accepting different "commands"

There are pros and cons with this. With one endpoint per action it becomes obvious from the url what the endpoint does. And if an endpoint doesn't exists you'll get a 404.

A single endpoint will obviously result in fewer endpoints but it will be more prone to errors as you need to either POST some command or include the command as a query argument.

> Just to be safe it might be better to start a second web server. This way it will not affect the existing web server in any way.

I don't see the value in running another web server. More open ports and more CPU resources are required.

> Another big question I have is can it be written in Java?

[...]

PR #7382, opened 2023-02-13
Add external tool endpoints to the editor http server

matgis:
Implements #6604.

See the [documentation](https://github.com/defold/defold/blob/7ed47f3db3a24832522e16b0c5d3c5c82784376c/editor/doc/http-api.md).

### User-facing changes
* Add experimental `command` endpoint to editor http server. External tools can use this to trigger build, hot-reload, etc.
* Add experimental `console` endpoint to editor http server. External tools can use this to render console output similar to the editor.

#### Remaining work:
* We should make our port discoverable somehow. In `targets.clj`, we use an SSDP service to discover running engine targets on the network. Perhaps we could use a similar mechanism? Not sure if the editor should be the one broadcasting or listening, though.

### DIFF
--- a/editor/doc/http-api.md
+++ b/editor/doc/http-api.md
@@ -1,0 +1,98 @@
+Editor HTTP API
+===============
+
+External tools can connect to the running editor and interact with it using a REST API.
+
+### Caveats
+* This HTTP API is currently in an experimental state. It might change drastically or even be removed completely if we deem it necessary. But hopefully it shouldn't come to that.
+* Currently, the port is assigned randomly from a broad subrange. The editor will log the local endpoint url if started from a terminal.
+
+### Overview
+The editor can be interacted with by performing GET requests to the following URLs:
+```
+http://localhost:[port]/command
+http://localhost:[port]/console
+```
+
+### The Command Endpoint
+The command endpoint exposes a subset of the commands available from the editor menu bar. To trigger a command, perform a GET request in the format:
+```
+http://localhost:[port]/command/[command]
+```
+You can obtain a list of the available commands along with a short description in JSON format from `http://localhost:[port]/command`. At the time of writing, this returns:
+```json
+{
+  "asset-portal"       : "Open the Asset Portal in a web browser.",
+  "async-reload"       : "Reload all modified files from disk.",
+  "build"              : "Build and run the project.",
+  "build-html5"        : "Build the project for HTML5 and open it in a web browser.",
+  "debugger-break"     : "Break into the debugger.",
+  "debugger-continue"  : "Resume execution in the debugger.",
+  "debugger-detach"    : "Detach the debugger from the running project.",
+  "debugger-start"     : "Start the project with the debugger attached.",
+  "debugger-step-into" : "Step into the current expression in the debugger.",
+  "debugger-step-out"  : "Step out of the current expression in the debugger.",
+  "debugger-step-over" : "Step over the current expression in the debugger.",
+  "debugger-stop"      : "Stop the debugger and the running project.",
+  "documentation"      : "Open the Defold documentation in a web browser.",
+  "donate-page"        : "Open the Donate to Defold page in a web browser.",
+  "editor-logs"        : "Show the directory containing the editor logs.",
+  "engine-profiler"    : "Open the Engine Profiler in a web browser.",
+  "fetch-libraries"    : "Download the latest version of the project library dependencies.",
+  "hot-reload"         : "Hot-reload all modified files into the running project.",
+  "issues"             : "Open the Defold Issue Tracker in a web browser.",
+  "rebuild"            : "Rebuild and run the project.",
+  "rebundle"           : "Re-bundle the project using the previous Bundle dialog settings.",
+  "reload-extensions"  : "Reload editor extensions.",
+  "reload-stylesheets" : "Reload editor stylesheets.",
+  "report-issue"       : "Open the Report Issue page in a web browser.",
+  "report-suggestion"  : "Open the Report Suggestion page in a web browser.",
+  "show-build-errors"  : "Show the Build Errors tab.",
+  "show-console"       : "Show the Console tab.",
+  "show-curve-editor"  : "Show the Curve Editor tab.",
+  "support-forum"      : "Open the Defold Support Forum in a web browser.",
+  "toggle-pane-bottom" : "Toggle visibility of the bottom editor pane.",
+  "toggle-pane-left"   : "Toggle visibility of the left editor pane.",
+  "toggle-pane-right"  : "Toggle visibility of the right editor pane."
+}
+```
+
+#### Response Codes
+Commands are fire-and-forget. You will not know if a command succeeds or not, only if your request was accepted. As such, all command requests will return the same response codes. They are as follows:
+
+* `202 Accepted` - The request was accepted and we are processing it. We use this instead of `200 OK` since we won't be able to tell you when it is done or how it went.
+* `400 Bad Request` - The request was malformed.
+* `404 Not Found` - The string after `command/` is not a supported command.
+* `405 Method Not Allowed` - The command is supported, but it can't be executed right now (for example if a build is already in progress).
+* `500 Internal Server Error` - Something unexpected went wrong when executing the command. The error will be logged in the editor log.
+
+### The Console Endpoint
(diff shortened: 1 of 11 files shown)


editor/src/clj/editor/command_requests.clj
```diff
@@ -0,0 +1,270 @@
...
+(defonce ^:private supported-commands
+  ;; Notable exclusions:
+  ;; :save-all, :quit, anything that would open a modal dialog.
+  {:asset-portal
+   {:ui-handler :asset-portal
+    :help "Open the Asset Portal in a web browser."}
+
+   :async-reload
+   {:ui-handler :async-reload
+    :help "Reload all modified files from disk."}
```
vlaaad, line 37: I think we should not expose this command and make reloading an implementation detail to not push this complexity onto our users.

editor/src/clj/editor/command_requests.clj
```diff
@@ -0,0 +1,270 @@
...
+   {:ui-handler :step-out
+    :help "Step out of the current expression in the debugger."}
+
+   :debugger-step-over
+   {:ui-handler :step-over
+    :help "Step over the current expression in the debugger."}
+
+   :debugger-stop
+   {:ui-handler :stop-debugger
+    :help "Stop the debugger and the running project."}
```
vlaaad, line 77: I think commands such as these shouldn't trigger resource sync. On big projects, resource sync takes a while to run, so it will be annoying to unnecessarily wait e.g. 5 seconds to step over.

editor/src/clj/editor/command_requests.clj
```diff
@@ -0,0 +1,270 @@
...
+         (map? user-data)]}
+  (ui/run-now
+    (let [command-contexts (ui/node-contexts ui-node true)]
+      (ui/resolve-handler-ctx command-contexts command user-data))))
+
+(defn make-request-handler [ui-node render-reload-progress!]
+  {:pre [(ui/node? ui-node)
+         (fn? render-reload-progress!)]}
+  (fn request-handler [request]
+    (when (= "GET" (:method request))
```
vlaaad, line 231: I think we should not accept `GET` for command execution because `GET` implies read-only semantics that does not have side effects. Maybe `POST` will work better?

editor/doc/http-api.md
```diff
@@ -0,0 +1,98 @@
...
+#### Region Types
+Depending on the `type` of a region, it might have additional fields. Here is a non-exhaustive list of region types used in the editor.
+
+* `eval-expression` - An expression posted from the Evaluate Lua field in the debugger.
+* `eval-result` - The result returned by the `eval-expression`.
+* `eval-error` - A runtime error resulting from an `eval-expression`.
+* `extension-output` - Output from an editor script or extension.
+* `extension-error` - Errors from an editor script or extension.
+* `repeat` - Repeated console output that has been collapsed. Its `count` is the number of repeats.
+* `resource-reference` - A substring that matches an existing path in the project. It has a `proj-path-candidates` array of project-relative paths to matching resources, and can also have a zero-based `row` index into the resource if it is text-based.
```
vlaaad, line 96: Maybe we should rename the key to `project-path-candidates` to make it clearer?

vlaaad (changes requested):
Looks very exciting!

I think it would be better to implement 2 improvements before shipping to users:
- make side-effecting command execution semantically match HTTP methods, i.e. instead of `GET` require something like `POST`;
- make async reloading an implementation detail of command execution. I also think we should not perform a resource sync for every command, but instead only where it matters, because on big projects resource sync might be noticeable.

It feels magical to play with it from the CLI! I think in the future we will want to implement more features like responding with results instead of fire-and-forget requests. This is especially useful for e.g. debugger where we can respond with information like locals/scope/stack/current breakpoint etc... But as an MVP it looks very useful as is!

editor/doc/http-api.md
```diff
@@ -0,0 +1,98 @@
...
+  "debugger-start"     : "Start the project with the debugger attached.",
+  "debugger-step-into" : "Step into the current expression in the debugger.",
+  "debugger-step-out"  : "Step out of the current expression in the debugger.",
+  "debugger-step-over" : "Step over the current expression in the debugger.",
+  "debugger-stop"      : "Stop the debugger and the running project.",
+  "documentation"      : "Open the Defold documentation in a web browser.",
+  "donate-page"        : "Open the Donate to Defold page in a web browser.",
+  "editor-logs"        : "Show the directory containing the editor logs.",
+  "engine-profiler"    : "Open the Engine Profiler in a web browser.",
+  "fetch-libraries"    : "Download the latest version of the project library dependencies.",
```
mikatuo, line 41: I discovered recently that UPnP is dangerous and unsafe. Just wandering if exposing "fetch libraries" could be dangerous? When someone clicks on "Fetch Libraries" they are looking at their added extensions. If it will be possible to trigger it externally then could someone use UPnP vulnerabilities to "silently" add a native code into an opened project? For indie devs that should not be a concern, but big companies might become a target since the source code is available for everyone.

mikatuo:
Thank you so much for the changes! Defold Team :1st_place_medal:

editor/doc/http-api.md
mikatuo: Hmm. On the second thought everyone will be using source control so I'm probably being paranoid :)

editor/doc/http-api.md
```diff
@@ -0,0 +1,98 @@
+Editor HTTP API
+===============
+
+External tools can connect to the running editor and interact with it using a REST API.
+
+### Caveats
+* This HTTP API is currently in an experimental state. It might change drastically or even be removed completely if we deem it necessary. But hopefully it shouldn't come to that.
+* Currently, the port is assigned randomly from a broad subrange. The editor will log the local endpoint url if started from a terminal.
```
mikatuo, line 8: > The editor will log the local endpoint url if started from a terminal.

If VS Code started the editor from a terminal could it read the port from logs somehow?

britzl: Yes, that should be possible. I'm also thinking that maybe VSCode could set an env variable with a port for the editor to use. If the env variable is set the editor will use it, otherwise it will use a random port. Similar to the DM_SERVICE_PORT between editor and engine:

https://github.com/defold/defold/blob/dev/engine/docs/DEBUG_PORTS_AND_SERVICES.md

editor/src/clj/editor/command_requests.clj
matgis: I did not intend for this to be necessary for users to invoke this manually, but was imagining a use case where you had the editor open in a scene view on a second monitor while you ran a tool that generated content. That way, the tool could ask the editor to refresh the view without you having to tab into the editor.

Although TBH I'm not sure we refresh the scene view while the editor is in the background at the moment. 😅

editor/src/clj/editor/command_requests.clj
matgis: That's a great point. I'll add it as metadata to the actions!

editor/src/clj/editor/command_requests.clj
matgis: That makes sense. I'll change it to POST!

editor/doc/http-api.md
matgis: I'm torn about this. The term `proj-path` means a very specific thing in all places. It denotes a project-relative path with a leading slash, I.e. `/main/main.collection`. I think we should use the term in our APIs as well since it is useful and ubiquitous in the editor.

editor/doc/http-api.md
vlaaad: Which document?

matgis: The link britzl posted?

matgis: Although I'm a bit annoyed that the example checks for any service that contains the substring `"Defold"`. Where does that leave our "Defold Editor" service? 🤔

vlaaad: It's like the Windows 8 -> Windows 10 situation where they skipped windows 9 because of `iswin95or98=os.startsWith("Windows 9")` :D

editor/src/clj/editor/command_requests.clj
matgis: I decided to remove the reload command for now, since it is only available if you're have the "Load External Changes on App Focus" preference disabled, which feels wrong.

vlaaad (approved):
Ship it!

PR #7388, opened 2023-02-16
Identical names require the same values in templates

### LINKED ISSUE #7222: Resources in templates with the same name randomly overrides each other
When a resource (font, texture, spine model, particlefx) in a template and in the main GUI (or anoother template) has the same resource_name but a different resource in it, the editor should show an error to prevent undef. behavior.

**To Reproduce (REQUIRED)**
1. Create gui 
2. create gui for template
3. add any resource (e.g. font or spine model) to bothwith the same name but different resource references
4. add templete into the main gui 
5. run the game
6. one of the resources will be picked and shown

Editor should notify that user is triend to use two different resources with the same name 

  - any

[template-editor-bug.zip](https://github.com/defold/defold/files/10244414/template-editor-bug.zip)

vlaaad:
User-facing changes: when multiple templates used in the same GUI define resources with clashing names, we now require them to refer to the same values. For instance, suppose a GUI file named `A.gui` contains two templates: `B.gui` and `C.gui`. If both templates define a font named "main_font" and the actual fonts used in each template are not the same, `A.gui` will not know which one to use and will randomly choose one. To prevent this issue, we will now display a build error when such clashes occur.

Fixes #7222

### DIFF
--- a/editor/src/clj/editor/gui.clj
+++ b/editor/src/clj/editor/gui.clj
@@ -45,6 +45,7 @@
             [editor.scene-picking :as scene-picking]
             [editor.slice9 :as slice9]
             [editor.types :as types]
+            [editor.util :as eutil]
             [editor.validation :as validation]
             [editor.workspace :as workspace]
             [internal.graph.types :as gt]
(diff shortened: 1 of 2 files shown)


AGulev:
It’s interesting if this issue is reproducible in the Bob.

matgis (approved):
Looks good to me!

PR #7402, opened 2023-02-20
Set global logger level based on verbose option

britzl:
The `java.util.logging.Logger` instances used in the `AndroidBundler`, `HTML5Bundler` and several other places in the code logged messages with `INFO` level regardless if the `verbose` flag was set or not. This change uses the value of the verbose flag to configure the global Logger level throughout the code.


britzl:
At som point we should probably also look through all of the code and use Logger instances when it makes sense. I feel that there are some things we print using System.out.println() which we do not necessarily want to print all the time and instead we should use a Logger instance.

JCash (approved):

PR #7412, opened 2023-02-22
Clean out usage of posix header files

JCash:
Code cleanup task

### DIFF
--- a/engine/dlib/src/zip/miniz.h
+++ b/engine/dlib/src/zip/miniz.h
@@ -1,17 +1,3 @@
-// Copyright 2020-2023 The Defold Foundation
-// Copyright 2014-2020 King
-// Copyright 2009-2014 Ragnar Svensson, Christian Murray
-// Licensed under the Defold License version 1.0 (the "License"); you may not use
-// this file except in compliance with the License.
-// 
-// You may obtain a copy of the License, together with FAQs at
-// https://www.defold.com/license
-// 
-// Unless required by applicable law or agreed to in writing, software distributed
-// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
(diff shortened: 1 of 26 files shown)


engine/dlib/src/zip/miniz.h
```diff
@@ -1,17 +1,3 @@
...
-// this file except in compliance with the License.
-// 
-// You may obtain a copy of the License, together with FAQs at
-// https://www.defold.com/license
-// 
-// Unless required by applicable law or agreed to in writing, software distributed
-// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
-// CONDITIONS OF ANY KIND, either express or implied. See the License for the
-// specific language governing permissions and limitations under the License.
-
```
AGulev, line 14: Could you please add these files into [apply_license.py](https://github.com/defold/defold/pull/7412/files#diff-7517c4d65e6a229d0fc054796d485f61554e8c0a790d440ff865457e9546010e) to ignore it in the future?

JCash: Thx, I was looking to find that one...

engine/engine/src/engine.cpp
```diff
@@ -17,6 +17,7 @@
 #include "engine_private.h"
 
 #include <dmsdk/dlib/vmath.h>
+
 #include <sys/stat.h>
```
Jhonnyg, line 21: Should this still be here?

Jhonnyg (approved):
Looks good! Just a small q

engine/engine/src/engine.cpp
JCash: I have other changes in the private repo as well for this case. So it's a two-step process.

PR #7421, opened 2023-02-25
Implement generic handle container and convert buffer to use it

Jhonnyg:
Added a generalized handle container for mapping objects to opaque handles that can be used to expose an internal engine resource as just a number rather than an actual pointer to the resource. This first step replaces the implementation in the buffer to use this new system, and the next step will be to use the same container concept for exposing graphic resources to the gameobject world for passing handles between GO and render scripts.


engine/dlib/src/test/test_opaque_handle_container.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <jc_test/jc_test.h>
+
+TEST(dmOpaqueHandleContainer, Basic)
+{
+	dmOpaqueHandleContainer<int> simple;
+	ASSERT_EQ(simple.Capacity(), 0);
+	ASSERT_TRUE(simple.Full());
+
+	// Container has no storage
+	ASSERT_EQ(simple.Put(0), INVALID_OPAQUE_HANDLE);
```
JCash, line 27: Do we want a different behavior from array+hashtable, which assert on this operation?

engine/dlib/src/dlib/opaque_handle_container.h
```diff
@@ -0,0 +1,280 @@
...
+}
+
+template <typename T>
+HOpaqueHandle dmOpaqueHandleContainer<T>::Put(T* obj)
+{
+    uint32_t index = GetFirstEmptyIndex();
+    if (index == INVALID_OPAQUE_HANDLE)
+    {
+        return INVALID_OPAQUE_HANDLE;
+    }
```
JCash, line 241: We don't handle the case when the `m_Capacity==0xFFFF`...

engine/dlib/src/dlib/opaque_handle_container.h
```diff
@@ -0,0 +1,280 @@
...
+
+    m_Version++;
```
JCash, line 243: ...and `m_Version==0xFFFF` after this line...

engine/dlib/src/dlib/opaque_handle_container.h
```diff
@@ -0,0 +1,280 @@
...
+    if (m_Version == 0)
+    {
+        // Set to 1 to avoid potentially producing a 0 handle
+        m_Version = 1;
+    }
+
+    m_ObjectVersions[index] = m_Version;
+    m_Objects[index]        = obj;
+
+    return m_Version << 16 | index;
```
JCash, line 253: ... and we return `(0xFFFF << 16) | 0xFFFF` which is equal to `INVALID_OPAQUE_HANDLE`, right?

I know I probably wrote this code, but we should probably fix this right now.
Also, it might be good to mention that the container cannot handle more than 65534 or 65535 handles.

JCash (changes requested):
Overall good, but some minor fixes/tests needed

engine/dlib/src/dlib/opaque_handle_container.h
```diff
@@ -0,0 +1,280 @@
...
+    {
+        for (int i = 0; i < m_Capacity; ++i)
+        {
+            if (m_Objects[i] == 0)
+            {
+                return i;
+            }
+        }
+        return INVALID_OPAQUE_HANDLE;
+    }
```
JCash, line 155: Future idea:
In order to improve poerformance when finding the free slot, I think we can encode a linked list inside the `T** m_Objects`.
E.g. if we have a "uintptr_t m_FreeList" which points to the first free index, and where the values in the m_Objects are indices;
```
// popping the first free index
uintptr_t index = m_FreeList;
m_FreeList = (uintptr_t)m_Objects[index];
return index;
```

Jhonnyg: Yes, I thought about adding a freelist as well, I'll make a note about it in the code

engine/dlib/src/dlib/opaque_handle_container.h
Jhonnyg: Hm, put doesn't grow the container, even if the capacity is the maximum allowed we can still have free slots to put handles.. But maybe the comment wasn't meant regarding this function but more overall?

engine/dlib/src/dlib/opaque_handle_container.h
Jhonnyg: Yes, if both of them are 0xFFFF it will equate an invalid handle, so either we cap the version or the max size of the container so the max handle becomes 0xfffffffe or 0xfffeffff I suppose

PR #7424, opened 2023-02-26
Added support for trimmed images in atlases.

Lerg:
Texture sizes can be greatly reduced by trimming transparent pixels around the actual sprite image.

The engine itself already has support for arbitrary vertex positions for each image which was introduced with the Sprite Trim Mode feature. Therefore only the editor and bob were modified to add the trimmed images support.

This feature introduces new fields for images in the atlas outline pane. Namely `Source Position X`, `Source Position Y`, `Source Width`, `Source Height`.

`Source Position X` and `Source Position Y` indicate the position offset of the trimmed image inside the original image from the top left corner. `Source Width` and `Source Height` indicate the original image size and they are needed to correctly center the trimmed image inside the animation frame.

This new information about the images can be entered by hand or a special user script can be used to prepare such atlases.

## TexturePacker support

https://github.com/Lerg/defold-texturepacker-exporter

To ease the generation of such atlases a custom exporter was created for TexturePacker. The exporter generates atlas file for Defold and a JSON file for the Spritesheet Split editor script.

The exported supports storing atlas settings and animations settings such as margin, padding, FPS, playback and so on.

### Spritesheet Split editor script

https://github.com/Lerg/defold-editor-script-spritesheet

An editor script was created that takes a spritesheet image generated by TexturePacker with a JSON file containing information on each image inside the spritesheet and splits up the image into individual files suitable for the atlas.

Resolves #3629

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/TextureSetGenerator.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/TextureSetGenerator.java
@@ -29,6 +29,7 @@ import com.dynamo.gamesys.proto.Tile.Playback;
 import com.google.protobuf.ByteString;
 
 import javax.vecmath.Point2d;
+import javax.vecmath.Point4i;
 import javax.vecmath.Vector2d;
 
 import java.awt.Color;
@@ -43,7 +44,6 @@ import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
-import java.util.HashMap;
 import java.util.stream.Collectors;
 
 // For debugging image output
@@ -294,7 +294,7 @@ public class TextureSetGenerator {
      * 3. Shrink inner rects by previous extrusion
      * 4. Create vertex data for each frame (image) in each animation
      */
-    public static TextureSetResult calculateLayout(List<Rect> images, List<SpriteGeometry> imageHulls, int use_geometries,
+    public static TextureSetResult calculateLayout(List<Rect> images, List<SpriteGeometry> imageHulls, List<SpriteGeometry> imageHullsWithOffset, int use_geometries,
                                                 AnimIterator iterator,
                                                int margin, int innerPadding, int extrudeBorders,
                                                boolean rotate, boolean useTileGrid, Grid gridSize) {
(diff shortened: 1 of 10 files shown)


engine/gamesys/proto/gamesys/atlas_ddf.proto
```diff
@@ -13,6 +13,10 @@ message AtlasImage
 {
     required string image                           = 1 [(resource) = true];
     optional SpriteTrimmingMode sprite_trim_mode    = 2 [default = SPRITE_TRIM_MODE_OFF];
+    optional int32 source_position_x                = 3 [default = 0];
+    optional int32 source_position_y                = 4 [default = 0];
+    optional uint32 source_width                    = 5 [default = 0];
+    optional uint32 source_height                   = 6 [default = 0];
```
JCash, line 19: For any feature this large/important, we want to have a design phase at the beginning, in order to make sure we consider as many aspects as possible.

Currently, I feel that the file format change may not be what we want.
For instance, it looks like the texture packer result is extracted, into separate images(?) and then assembled again in our file format?
That's certainly something we need to think about.
E.g. an idea we've had is to store the actual packing info in the atlas file.

Please start a design document (e.g. google docs) where we can have discussions about the feature.

JCash (changes requested):
Needs a design document where we can consider file format changes and workflows.

engine/gamesys/proto/gamesys/atlas_ddf.proto
Lerg: This feature was designed and developed for a company and is now being used there internally, they were kind enough to share it with the community. Hence there was no public design document since it's basically an internal feature.

> For instance, it looks like the texture packer result is extracted, into separate images(?) and then assembled again in our file format?

TexturePacker exporter generates a ready atlas file for Defold, unfortunately it's not capable of exporting individual sprite images for Defold. Therefore I've created the spritesheet splitting script that just splits the TexturePacker's spritesheet into separate files. The generated JSON file is used just for that, the script does not modify the atlas file and the JSON file is deleted after the splitting.

> E.g. an idea we've had is to store the actual packing info in the atlas file.

Sorry, not understanding this. What other packing info? In this feature the packing info is already stored in the atlas file alone. The json file is not needed. These four new fields of images inside atlas are all that is needed for that feature.

I've created the design document file as you requested. https://docs.google.com/document/d/1fm5zHKLUZzMVYmLE2E7aFpb_2vBDxGB7qWhPHooL3Uw/edit?usp=sharing

We were not sure about Defold's roadmap in regard to atlas or texture improvements that's why we designed and developed this feature in this manner to meet the current project needs. We were designing the feature taking into account current Defold architecture with its pros and cons.
You can check the design file and we can discuss it and I can help with the Java side of things, but I would rather not touch the Clojure code, it's way to hard for me. The most important part of the feature's code is in [com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/TextureSetGenerator.java](https://github.com/defold/defold/pull/7424#diff-66aa1f9e9cb6b575e7454c7a8812104ae2b15f2e1b26d565b10510ad54d70393) anyway.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/TextureSetGenerator.java
```diff
@@ -374,33 +435,67 @@ public static BufferedImage layoutImages(LayoutResult layoutResult, Map<String,
...
+     * @param innerPadding atlas padding size
+     * @param extrudeBorders atlas extrude borders size
+     * @param rotate are images allowed to be rotated
+     * @param useTileGrid when using with tile grids
+     * @param gridSize grid size of the tile grid
+     * @return {@link TextureSetResult}
      */
-    public static TextureSetResult generate(List<BufferedImage> images, List<Integer> imageHullSizes, List<String> paths, AnimIterator iterator,
-            int margin, int innerPadding, int extrudeBorders, boolean rotate, boolean useTileGrid, Grid gridSize) {
+    public static TextureSetResult generate(List<BufferedImage> images, List<Integer> imageHullSizes, List<Point4i> imageSourceRects, List<String> paths, AnimIterator iterator,
```
paweljarosz, line 450: My small comment is that added parameters could be added as last ones, what do you think?

AGulev:
it's not something we going to merge
texture packer support will be introduced as a separate plugin

PR #7426, opened 2023-02-27
Improve Lua-related LSP configuration responses

vlaaad:
This changeset adds 2 improvements:
- for Lua language servers, the editor's LSP client now also reports globals defined in `.script_api` files, thus improving linting accuracy
- editor's LSP client now responds to `files.associations` configuration request with information about which file extensions correspond to which languages. This information might be used by language servers for filtering which files to lint when asked for workspace linting.

Related to #3885

### DIFF
--- a/editor/src/clj/editor/lsp/server.clj
+++ b/editor/src/clj/editor/lsp/server.clj
@@ -163,11 +163,29 @@
                          :else
                          (throw (ex-info "Invalid text document sync kind" {:value textDocumentSync})))})
 
-(def default-settings
-  {"Lua" {:diagnostics {:globals lua/defined-globals}}})
+(defn- configuration-handler [project]
+  (fn [{:keys [items]}]
+    (lsp.async/with-auto-evaluation-context evaluation-context
+      (mapv
+        (fn [{:keys [section]}]
+          (case section
+            "Lua"
+            (let [script-intelligence (g/node-value project :script-intelligence evaluation-context)
+                  completions (g/node-value script-intelligence :lua-completions evaluation-context)]
+              {:diagnostics {:globals (into lua/defined-globals
+                                            (lua/extract-globals-from-completions completions))}})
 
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/lsp/server.clj
```diff
@@ -163,11 +163,29 @@
...
+            (let [script-intelligence (g/node-value project :script-intelligence evaluation-context)
+                  completions (g/node-value script-intelligence :lua-completions evaluation-context)]
+              {:diagnostics {:globals (into lua/defined-globals
+                                            (lua/extract-globals-from-completions completions))}})
+
+            "files.associations"
+            (let [workspace (g/node-value project :workspace evaluation-context)
+                  resource-types (g/node-value workspace :resource-types evaluation-context)]
+              (into {}
+                    (keep (fn [[ext {:keys [textual? language]}]]
```
matgis, line 182: `:textual?` files include protobuf files such as `*.go` and `*.atlas` as well as settings files such as `game.project`. Do we want these to be included here? Or should we only include files that can be opened using the `:code` `:view`?

matgis (approved):
Looks good! I had a question about what gets included in `files.associations`, though?

editor/src/clj/editor/lsp/server.clj
vlaaad: Good question. So far, we only need information about files openable in the code view. When we will have workspace diagnostics though, it will be possible to use language servers to lint `.go` and `.atlas` files as a build step, but we currently don't define adequate language id for the protobuf text files (they are reported as `plaintext` while it should be something like [`textproto`](https://github.com/thejustinwalsh/textproto-grammar/blob/97efb452403712938d9e2165e4c003e4a0ef6ace/extension.js#L6) instead). Maybe we will allow editing these files as text in the future? Anyway, those files are excluded now because we don't report the fallback `plaintext` language.

PR #7431, opened 2023-03-02
Refactor textures as resource

Jhonnyg:
As a first step in representing textures as actual handles that can be passed around between the render scripts and the rest of the engine, we need to separate between the actual texture handles and the resources that hold them. A HTexture is currently a pointer to the asset created by the graphics adapter which is then stored in the resource system. Instead, a wrapper around the graphics handle is stored, which means that we can change the way we represent the handle.

Technical note: This is a breaking change in the public dmSDK part of the engine. Several resources now store TextureResource pointers instead of the HTexture objects, and any native extension that use these APIs will have to update their code to make sure the types are correct.

## PR checklist

* [ ] Code
* [ ] Documentation


engine/gamesys/src/gamesys/components/comp_mesh.cpp
```diff
@@ -504,10 +509,25 @@ namespace dmGameSystem
...
+                ro.m_Textures[i] = textures_component[i]->m_Texture;
+            }
+            else if (textures_resource[i])
+            {
+                ro.m_Textures[i] = textures_resource[i]->m_Texture;
+            }
+            else
+            {
+                ro.m_Textures[i] = 0;
+            }
```
JCash, line 524: No need since the ro.Init() was called above

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -645,7 +656,8 @@ namespace dmGameSystem
 
         for(uint32_t i = 0; i < MAX_TEXTURE_COUNT; ++i)
         {
-            ro.m_Textures[i] = GetTexture(component, component->m_Resource, i);
+            TextureResource* texture_res = GetTextureResource(component, component->m_Resource, i);
+            ro.m_Textures[i] = GetTexture(texture_res);
```
JCash, line 660: Imho, this code looks a little verbose, since the GetTexture only really gets the texture, but also has to check the resource first.

Perhaps have a `HTexture GetTexture(component, resource, i)`?

engine/gamesys/src/gamesys/resources/res_texture.h
```diff
@@ -17,6 +17,7 @@
 
 #include <resource/resource.h>
 #include <dmsdk/gamesys/resources/res_texture.h>
+#include <dmsdk/graphics/graphics.h>
```
JCash, line 20: This header is not used?
In order to keep code dependencies as low as possible, and build times low, we want to remove unused headers from headers (and ofc cpp files)

Looks like the `res_texture.h` isn't used either?

JCash (changes requested):
Looks good.
Just a small header fix plus some function name issue.

PR #7437, opened 2023-03-03
Fix bug when some of embeded GOs maybe ignored by component counter

### LINKED ISSUE #7434: Collection factories are broken in bundled builds
1.4.3 broke collection factories for us. In bundled builds collectionfactory.create() returns an empty table and nothing gets spawned. When running from the editor this doesn’t happen.

**To Reproduce (REQUIRED)**
We don't have a minimal repro case, but our code looks like this:
```lua
  local instance = collectionfactory.create(self.factory_url, nil, nil, {
    [hash("character")] = {
      -- some stuff
    }
  })
  pprint(instance) -- Empty table
```

The collection factory is not loaded dynamically, nor does it have a dyamic prototype

  - Version 1.4.3 (1.4.2 was fine)

**Platforms (REQUIRED):**
 - First noticed this on web, but then managed to repro on macOS too

AGulev:
In a case when embedded GOs in a few different collections are the same, component counter didn't count the second and the following GOs.

Fix https://github.com/defold/defold/issues/7434

## PR checklist

* [x] Code
* [ ] Documentation

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/CollectionBuilder.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/CollectionBuilder.java
@@ -68,7 +68,7 @@ public class CollectionBuilder extends ProtoBuilder<CollectionDesc.Builder> {
     }
 
     private void createGeneratedResources(Project project, CollectionDesc.Builder builder,
-        Map<Long, IResource> uniqueResources ) throws IOException, CompileExceptionError {
+        Map<Long, IResource> uniqueResources, Map<Long, IResource> allResources) throws IOException, CompileExceptionError {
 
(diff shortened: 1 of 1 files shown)


AGulev:
I'm not sure if test is needed here. I can't add it as unit test for collection, because I need an integration test where bob build the whole folder and keep state, because it's related to using of non-unique objects

britzl (approved):
I don't quite understand the fix, but if it works it works!

AGulev:
> I don't quite understand the fix, but if it works it works!

If we have two absolutely equal embedded resources in different collections, these resources will generate the same exact file (even if it's embedded, yes). One resource used in many collections with the same exact name, and it produces the same exact hash and file on disc. For the first collection this resource will be unique, bot for the second and the following `project.getGeneratedResource(hash, "go");` will return the existent resource.

PR #7444, opened 2023-03-06
Updated sprite size property documentation

britzl:
The sprite size property was marked as read-only even though it is possible to change the size of a sprite if the size-mode is set to manual.


britzl:
Also updated docs:

https://defold.com/manuals/sprite/#slice-9-texturing

PR #7463, opened 2023-03-13
Add warning when trying to load HTML5 build from local filesystem

### LINKED ISSUE #7447: Improvement to make it clear that html5 bundle should be run using a web server
**Task (REQUIRED):**
Sometimes people don't understand that bundle should be run using a web server
There are a few suggestions to improve the situation:

- [x] In debug (only in debug?) index.html add checking for file://  in the address, if it detected show a message with a link to manual
- [ ] Add a special warning text on a bundle window for html5 about it with a ref to the manual
- [ ] show message in console when bundle is done (in bob, and now we show it in the editor as well)
- [ ] create a bat/sh file which can help to run it

People should understand that the can't just double-click `index.html` and run the game

britzl:
A common mistake for beginners is to double-click on index.html when attempting to test an HTML5 bundle. This will not work since the browser is unable to load the rest of the files required to run the Defold game. This change adds a warning message and instructions on how to test an HTML5 build locally:

Screenshot 2023-03-13 at 15 47 05

Fixes #7447

### DIFF
--- a/engine/engine/content/builtins/manifests/web/engine_template.html
+++ b/engine/engine/content/builtins/manifests/web/engine_template.html
@@ -69,6 +69,11 @@
 </head>
 
 <body>
+	<div id="running-from-file-warning" style="display: none; margin: 3em;">
+		<h1>Running from local file ⚠️</h1>
+		<p>It seems like you have opened this file by double-clicking on it. In order to test your build in a browser <b>you need to load this file from a web server</b>. You can either upload this file and the rest of the files from a Defold HTML5 bundle to a web hosting service OR host them using a local web server on your home network.</p>
+		<p><a href="https://defold.com/manuals/html5/#testing-html5-build" target="_blank">Learn more about running a local web server in the Defold HTML5 manual</a>.</p>
+	</div>
 	<div id="app-container" class="canvas-app-container">
 		<div id="canvas-container" class="canvas-app-canvas-container">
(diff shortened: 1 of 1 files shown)


engine/engine/content/builtins/manifests/web/engine_template.html
```diff
@@ -69,6 +69,11 @@
 </head>
 
 <body>
+	<div id="running-from-file-warning" style="display: none; margin: 3em;">
```
AGulev, line 72: should we exclude it for release?

britzl: Possibly. Or have it as an option in game.project?

AGulev: hmmm maybe, but in this case user will never know we have something like this in index.html

britzl: One concern I have with removing this warning in a release build is that a beginner might export a release build and try to double-click on index.html and end up in the same situation as we're trying to resolve.

AGulev: true

AGulev (approved):

subsoap:
So point is maybe don't force this warning always / add easy way to disable it.

britzl:
> So point is maybe don't force this warning always / add easy way to disable it.

The warning is added to the default index.html. If you bundle your game into a single html file you won't be using the default index.html anyway.

PR #7465, opened 2023-03-14
Bugfix for using multiple rendertargets in WebGL2

Jhonnyg:
Fixed a regression for multiple rendertargets when running in a WebGL 2 context. We previously checked if the appropriate functions exist to support multiple rendertargets, but in WebGL 2 we can correctly assume that the functions exist since they are part of the baseline specification. We have also added a validation function for WebGL that checks if all the attachments of a rendertarget has the same size, which is a requirement as per the specification.

## PR checklist

* [ ] Code
* [ ] Documentation


engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2309,6 +2370,24 @@ static void LogFrameBufferError(GLenum status)
...
+        if (context->m_MultiTargetRenderingSupport)
+        {
+            max_color_attachments = MAX_BUFFER_COLOR_ATTACHMENTS;
+            check_attachments_same_size = true;
+
+            if (!ValidateFramebufferAttachments(max_color_attachments, buffer_type_flags, color_buffer_flags, creation_params))
+            {
+                dmLogError("All attachments must have the same size!");
+                return 0;
+            }
```
Jhonnyg, line 2387: This is a webgl caveat only, which is super annoying.. Maybe we should state that in the docs

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2430,7 +2509,13 @@ static void LogFrameBufferError(GLenum status)
         glBindFramebuffer(GL_FRAMEBUFFER, render_target == NULL ? glfwGetDefaultFramebuffer() : render_target->m_Id);
         CHECK_GL_ERROR;
 
-        if (render_target != NULL && PFN_glDrawBuffers != 0x0)
+    #if __EMSCRIPTEN__
+        #define DRAW_BUFFERS_FN glDrawBuffers
+    #else
+        #define DRAW_BUFFERS_FN PFN_glDrawBuffers
+    #endif
```
Jhonnyg, line 2516: We could maybe set the PFN_glDrawBuffers ptr to glDrawBuffers for emscripten and skip this define

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2279,6 +2280,66 @@ static void LogFrameBufferError(GLenum status)
         }
     }
 
+#if __EMSCRIPTEN__
+    // Only used for webgl currently
+    static bool ValidateFramebufferAttachments(uint8_t max_color_attachments, uint32_t buffer_type_flags, BufferType* color_buffer_flags, const TextureCreationParams creation_params[MAX_BUFFER_TYPE_COUNT])
+    {
+        bool first_set       = false;
```
JCash, line 2287: When is this boolean actually set?

JCash (changes requested):

PR #7467, opened 2023-03-14
Dont apply extrudeborders twice when calculating geometry uvs

Jhonnyg:
The rects already have been resized by the extrude borders value when calculating the geometry UVs, so we skip that in this PR. The previous code (pre-paged-atlas) used the non-clipped rects when doing the UV generation, so that's why the border extraction was needed. 

I.e old code:

```
List<Rect> rects = clipBorders(layout.getRectangles(), extrudeBorders);
if (imageHulls != null) {
    for (Rect rect : layout.getRectangles()) {
        SpriteGeometry geometry = imageHulls.get(rect.index);
        vertexData.left.addGeometries(createPolygonUVs(geometry, rect, layout.getWidth(), layout.getHeight(), extrudeBorders));
     }
}
```

New code:
```
layoutRects = clipBorders(layoutRects, extrudeBorders);

if (imageHulls != null) {
      for (Rect rect : layoutRects) {
          SpriteGeometry geometry = imageHulls.get(rect.index);
          vertexData.left.addGeometries(createPolygonUVs(geometry, rect, layoutWidth, layoutHeight, extrudeBorders));
      }
  }
```

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/TextureSetGenerator.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/TextureSetGenerator.java
@@ -232,18 +232,19 @@ public class TextureSetGenerator {
     }
 
     // From the vertices and layout, generate UV coordinates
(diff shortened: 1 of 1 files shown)


JCash (approved):
Lgtm

PR #7470, opened 2023-03-15
Improve manual Size Mode handling in the editor

### LINKED ISSUE #7464: Respect `manual` size mode for sprites in the editor
 `manual` size mode in the editor should be respected. Editor should set current size when user pick `Manual` and save it in sprite's proto file.

**To Reproduce (REQUIRED)**

1. Change sprite size to manual
2. for ex. sprite is 32x32
3. I see in width and height 32x32
4. I change image in atlas to 100x100 with the same name
5. return to the go with sprite sprite and now it’s 100x100

 I expect it to stay 32x32 

  - Version 1.4.3/1.4.4

matgis:
Fixes #7464

### User-facing changes
* Switching from auto to manual Size Mode on sprites or GUI objects will now set the manual size from the image size.
* Corrected the editor bounding box of sprites with manual Size Mode.

### DIFF
--- a/editor/src/clj/dev.clj
+++ b/editor/src/clj/dev.clj
@@ -164,6 +164,14 @@
 (defn console-view []
   (-> (view-of-type console/ConsoleNode) (g/targets-of :lines) ffirst))
 
+(defn node-values [node-id & labels]
+  (g/with-auto-evaluation-context evaluation-context
+    (into {}
+          (map (fn [label]
+                 (let [value (g/node-value node-id label evaluation-context)]
+                   (pair label value))))
+          labels)))
+
 (def node-type-key (comp :k g/node-type*))
 
 (defn- class-symbol [^Class class]
--- a/editor/src/clj/dynamo/graph.clj
+++ b/editor/src/clj/dynamo/graph.clj
@@ -160,8 +160,9 @@
          basis (is/basis system)
          id-generators (is/id-generators system)
          override-id-generator (is/override-id-generator system)
+         tx-data-context-map (or (:tx-data-context-map opts) {})
          metrics-collector (:metrics opts)
-         transaction-context (it/new-transaction-context basis id-generators override-id-generator metrics-collector)
+         transaction-context (it/new-transaction-context basis id-generators override-id-generator tx-data-context-map metrics-collector)
          tx-result (it/transact* transaction-context txs)]
      (when (and (not (:dry-run opts))
                 (= :ok (:status tx-result)))
--- a/editor/src/clj/editor/properties.clj
+++ b/editor/src/clj/editor/properties.clj
@@ -460,15 +460,29 @@
         new-value (from-fn new-value)]
     (set-value-txs evaluation-context node-id prop-kw key set-fn value new-value)))
 
-(defn- set-values [evaluation-context property values]
+(defn- resolve-set-operations
+  "Resolves the supplied property and sequence of new values to a vector of
+  individual [node-id prop-kw old-value new-value] vectors detailing the
+  low-level property set operations that will be performed as a result."
+  [property values]
+  (let [node-ids (:node-ids property)
+        prop-kws (:prop-kws property)
+        old-values (:values property)
+        new-values (if-some [from-fn (-> property :edit-type :from-type)]
+                     (mapv from-fn values)
+                     values)]
+    (mapv vector node-ids prop-kws old-values new-values)))
+
+(defn- edited-endpoints [set-operations]
+  (into #{}
+        (map (fn [[node-id prop-kw]]
+               (g/endpoint node-id prop-kw)))
+        set-operations))
+
+(defn- set-values [evaluation-context property set-operations]
   (let [key (:key property)
-        set-fn (get-in property [:edit-type :set-fn])
-        from-fn (get-in property [:edit-type :from-type] identity)]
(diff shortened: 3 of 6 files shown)


editor/src/clj/dev.clj
```diff
@@ -164,6 +164,14 @@
...
   (-> (view-of-type console/ConsoleNode) (g/targets-of :lines) ffirst))
 
+(defn node-values [node-id & labels]
+  (g/with-auto-evaluation-context evaluation-context
+    (into {}
+          (map (fn [label]
+                 (let [value (g/node-value node-id label evaluation-context)]
+                   (pair label value))))
+          labels)))
+
```
matgis, line 174: Unrelated: Added a helper to `dev` for something I tend to do a lot.

editor/src/clj/dynamo/graph.clj
```diff
@@ -160,8 +160,9 @@
          basis (is/basis system)
          id-generators (is/id-generators system)
          override-id-generator (is/override-id-generator system)
+         tx-data-context-map (or (:tx-data-context-map opts) {})
          metrics-collector (:metrics opts)
-         transaction-context (it/new-transaction-context basis id-generators override-id-generator metrics-collector)
+         transaction-context (it/new-transaction-context basis id-generators override-id-generator tx-data-context-map metrics-collector)
```
matgis, line 165: The `:tx-data-context` field inside the `transaction-context` hosts an `atom` that can be used to communicate state changes inside the transaction. Previously we were using it to keep track of the node-ids of `ResourceNodes` created as a result of a property setter referring to a resource we haven't picked up yet to make sure we do not create duplicates. For example, if several sprites would refer to a missing material resource, we want all of them to refer to the same defective resource node, not create duplicate defective resource nodes.

This `:tx-data-context` atom is accessible from the `evaluation-context` supplied to all property `set` functions.

You can now supply an initial state for the map in the `:tx-data-context`, and we use it to host the set of property endpoints edited by the user using the Property Editor. This enables use to check inside a property `set` function if we're being invoked as a result of the user editing a particular property with the help of the new `properties/user-edit?` function.

editor/src/clj/editor/properties.clj
```diff
@@ -460,15 +460,29 @@
...
+  individual [node-id prop-kw old-value new-value] vectors detailing the
+  low-level property set operations that will be performed as a result."
+  [property values]
+  (let [node-ids (:node-ids property)
+        prop-kws (:prop-kws property)
+        old-values (:values property)
+        new-values (if-some [from-fn (-> property :edit-type :from-type)]
+                     (mapv from-fn values)
+                     values)]
+    (mapv vector node-ids prop-kws old-values new-values)))
```
matgis, line 474: Refactoring: A `property` map in the Property Editor can refer to one or more properties on various nodes. For example, when I select a `EmbeddedComponentNode` representing a Sprite component in the Property Editor, it is also presenting the properties from the embedded `SpriteNode`.

We need this sequence of resolved "set-operations" in `set-values` below, but also to calculate the `edited-endpoints`.

editor/src/clj/editor/properties.clj
```diff
@@ -491,17 +505,35 @@
...
   ([property values]
    (set-values! property values (gensym)))
   ([property values op-seq]
    (when (not (read-only? property))
-     (let [evaluation-context (g/make-evaluation-context)]
+     (let [evaluation-context (g/make-evaluation-context)
+           set-operations (resolve-set-operations property values)
+           edited-endpoints (edited-endpoints set-operations)]
        (g/transact
+         {:tx-data-context-map {:edited-endpoints edited-endpoints}}
```
matgis, line 532: I the future we might want to do this from the `form-view` as well, but currently only the Property Editor is used to edit the properties that call the `properties/user-edit?` function.

editor/src/clj/editor/sprite.clj
```diff
@@ -375,12 +384,13 @@
...
-                                           (geom/make-aabb (Point3d. (- animation-width) (- animation-height) 0)
-                                                           (Point3d. animation-width animation-height 0)))
-                                         geom/empty-bounding-box)))
+  (output aabb AABB (g/fnk [size]
+                      (let [[^double width ^double height ^double depth] size
+                            half-width (* 0.5 width)
+                            half-height (* 0.5 height)
+                            half-depth (* 0.5 depth)]
+                        (geom/make-aabb (Point3d. (- half-width) (- half-height) (- half-depth))
+                                        (Point3d. half-width half-height half-depth)))))
```
matgis, line 393: Unrelated: I noticed the bounding boxes of sprites that use `:size-mode-manual` were not correct - the bounds were based on the texture size, causing box selection and framing to be wrong.

vlaaad (approved):
LGTM, but I have a question: is it possible to achieve the expected behavior by ensuring that during the loading we apply the size mode **before** applying the sizes? And in that case, what would be a better solution?

matgis:
Yes, I do believe that would work. But we'd also have to ensure the texture is assigned after the `:size-mode` in order to avoid loading the image dimensions from the referenced `png` files, which is rather costly.

This could be avoided for sprites, but in the case of GUI nodes we create override nodes, and there would already be an existing texture hooked up to the original node.

I also feel being explicit like this avoids the potential future pitfall if something messes up the order the properties are applied.

PR #7477, opened 2023-03-17
Added option to make DefoldActivity transparent on Android

Fr1ar:
It is now possible to make the main Activity transparency by adding a meta-data option in `AndroidManifest.xml`:

```xml
    <meta-data android:name="alpha.transparency" android:value="true" />
```

Note: By default the feature is disabled and does not negatively impact performance.

### DIFF
--- a/engine/glfw/java/com/dynamo/android/DefoldActivity.java
+++ b/engine/glfw/java/com/dynamo/android/DefoldActivity.java
@@ -17,6 +17,7 @@ package com.dynamo.android;
 
 import android.app.Activity;
 import android.app.NativeActivity;
+import android.content.pm.ApplicationInfo;
 import android.content.res.Configuration;
 import android.content.Context;
 import android.content.Intent;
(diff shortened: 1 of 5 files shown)


britzl:
Question: Do we also want this as a game.project option? We can automatically pre-process the AndroidManifest.xml and inject values from game.project.

engine/glfw/lib/android/android_jni.h
```diff
@@ -0,0 +1,27 @@
...
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#ifndef _ANDROID_JNI_H_
+#define _ANDROID_JNI_H_
+
+#include "internal.h"
+#include <jni.h>
+
+extern struct android_app* g_AndroidApp;
```
JCash, line 21: This belongs to a compile unit, e.g. android_jni.c

engine/glfw/java/com/dynamo/android/DefoldActivity.java
```diff
@@ -522,4 +523,39 @@ public String getGameControllerDeviceName(int deviceId) {
         }
         return name;
     }
+
+    /**
+     * Method to get meta-data value by key from AndroidManifest.xml
+     * @param key
+     * @return String
+     */
+    protected String getMetadata(String key) {
```
JCash, line 532: Neat function.
Will this require any extra permissions?

engine/glfw/lib/android/android_jni.c
```diff
@@ -0,0 +1,43 @@
...
+    JavaVM* vm = g_AndroidApp->activity->vm;
+    (*vm)->DetachCurrentThread(vm);
+}
+
+jmethodID JNIGetMethodID(JNIEnv* env, jobject instance, char* method, char* signature)
+{
+    if (instance == 0) return 0;
+    jclass clazz = (*env)->GetObjectClass(env, instance);
+    return (*env)->GetMethodID(env, clazz, method, signature);
+}
```
JCash, line 43: 👍

JCash (approved):
Looks good.
A small code change needed.
Also a question regarding any extra permissions needed to read the manifest?

engine/glfw/java/com/dynamo/android/DefoldActivity.java
Fr1ar: No, extra permissions are not required.

engine/glfw/lib/android/android_jni.h
Fr1ar: The `g_AndroidApp` variable is used in several places,
here:
https://github.com/Fr1ar/defold/blob/6329dbb495efd1ed983de429df126f54bbd20a8c/engine/glfw/lib/android/android_joystick.c#L289

and here:
https://github.com/Fr1ar/defold/blob/6329dbb495efd1ed983de429df126f54bbd20a8c/engine/glfw/lib/android/android_util.c#L30

so it must be declared in the header file.

Of course, it could be moved to a cpp file and wrapped around with `android_app* getAndroidApp()` function, but I don't think that's necessary.
So what should I do, leave it as it is or wrap it around?

Fr1ar:
> Also a question regarding any extra permissions needed to read the manifest?

No, extra permissions are not required.

Jhonnyg (approved):

PR #7480, opened 2023-03-17
Updated build toolchain with XCode 14.2, macOS 13.1 and iOS 16.2

### LINKED ISSUE #7452: Update Xcode to 14.2 and iOS to 16.1
**Task (REQUIRED):**
Some libraries already use Xcode 14.2. Which means it can't be built as a Defold extension. For example https://github.com/AGulev/DefVideoAds/tree/unity-4.6

Defold must also start supporting iOS 16.1 as that is a requirement going forward. (https://developer.apple.com/news/?id=z1erkhzr)

- [ ] Package Xcode tools and iOS SDK
    - [ ] Upload to AWS
- [ ] Update build_input.yml
- [ ] Update sdk.py
- [ ] Update extender setup-standalone-server.sh

Build for iOS 16.1 and use libs which were built using Xcode 14.2.

JCash:


.github/workflows/engine-nightly.yml
```diff
@@ -19,6 +19,7 @@ jobs:
     steps: [
       { name: 'Checkout', uses: actions/checkout@v3 },
       { name: 'Install Python', uses: actions/setup-python@v3, with: { python-version: 3.x, architecture: x64 } },
+      { name: 'XCode', uses: maxim-lobanov/setup-xcode@v1, with: { xcode-version: '14.2' } },
```
britzl, line 22: Is there no "official" setup action?

britzl: Ah, I see we've used it before as well. Ok!

JCash: DIdn't find one last time 🤷

britzl (dismissed):
Looks good. Only a note on indentation (mixed tabs and spaces)

AGulev (approved):

PR #7484, opened 2023-03-20
FIx for accidentally loading gltf models twice.

### LINKED ISSUE #7483: Fix issue with gltf model loader being loaded twice.
The gltf model loader was recently updated, and a second "finalize" step was accidentally run on the data.
It worked well on macOS, but most often would fail randomly on Win32.

JCash:


engine/modelc/scripts/test_model_importer.sh
```diff
@@ -26,6 +26,15 @@ LIBNAME=modelc_shared
...
     SUFFIX=.dylib
+elif [[ "$OSTYPE" == "cygwin" ]]; then
+    # POSIX compatibility layer and Linux environment emulation for Windows
+    SUFFIX=.dll
+elif [[ "$OSTYPE" == "msys" ]]; then
+    # Lightweight shell and GNU utilities compiled for Windows (part of MinGW)
+    SUFFIX=.dll
+elif [[ "$OSTYPE" == "win32" ]]; then
+    # I'm not sure this can happen.
+    SUFFIX=.dll
```
JCash, line 37: Make sure we can load/run the test script file on windows too

engine/modelc/src/modelimporter_gltf.cpp
```diff
@@ -40,6 +40,7 @@ namespace dmModelImporter
 struct GltfData
 {
     cgltf_data* m_Data;
+    uint8_t     m_Finalized : 1;
```
JCash, line 43: The load flow is multi step, and spread out, so I opted for this check.

britzl: I don't quite understand the fix. Is `LoadFinalizeGltf` called multiple times? If so, why is that?

engine/modelc/src/modelimporter_gltf.cpp
```diff
@@ -989,16 +988,13 @@ static void LoadScene(Scene* scene, cgltf_data* data)
...
-        LoadScene(scene, data->m_Data);
-        data->m_Finalized = 1;
-    }
+    LoadScene(scene, data->m_Data);
     return true;
 }
 
 static bool ValidateGltf(Scene* scene)
 {
+    printf("ValidateGltf\n");
```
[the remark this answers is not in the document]
JCash: no, we don't :(

engine/modelc/src/modelimporter_jni.cpp
```diff
@@ -979,7 +981,7 @@ JNIEXPORT jobject JNICALL Java_ModelImporter_LoadFromBufferInternal(JNIEnv* env,
         }
     }
 
-    if (!dmModelImporter::NeedsResolve(scene))
+    if (resolved && !dmModelImporter::NeedsResolve(scene))
```
britzl, line 984: What will the behaviour be if `resolved` is `false` and we proceed to call `dmModelImporter::CreateJavaScene()`?

JCash: In java, you will always have to call it at least once.
If all buffers were already resolved, then the user got a model and is good to go.

If the buffers aren't resolved, then the model is incomplete.

* From the editor side, we load the model once.
* If there are any missing buffers, we put those into a DataResolver map
* Then we call the model loader again, and this time with the data map so that it can find external buffer data.

The fix is  _only_ so that we only call Finalize exactly once.

The Finalize can be called either internally in modelimporter_gltf.cpp, if there was no need to load any extra buffers (all buffers were already resolved), and we can go ahead and return the result.
Or, it can be called here, if the jni has actually resolved all outstanding buffers.
And that only happends if there were outstanding buffers to begin with (i.e. the modelimporter_gltf.cpp didn't call Finalize)

britzl: Ah, thank you for the explanation!

PR #7485, opened 2023-03-20
Use workspace diagnostics during the build

vlaaad:
User-facing changes:
This changeset expands our LSP implementation by asking to diagnose the workspace on the build. Our fork of Sumneko's Lua language server was modified to respond to these diagnostic requests — you will need to use the [newest version](https://github.com/defold/lua-language-server/releases/tag/v0.0.4) to try it out.

Implementation notes:
- I cleaned up our `editor.app-view/async-build!` function. There were some issues with it. First, it had muddy semantics in terms of its results — it could return a map with results or errors, but it also could return `nil` when `editor.app-view/build-project!` threw an exception. It feels like it all worked by luck because `handle-build-results!` is used as a predicate of sorts to test whether the build was successful, and optionally show errors in the UI, and it happened that it returned the result if there was no `:error` key. And since the resulting map was `nil`, it was considered falsey and treated later as a failure. I streamlined it to always return a map with either successful results or error information. Second, it had a mix of positional and map args. Given how many args there were, I changed it to only accept kv-args to improve readability.
- The most complex part was designing request-response handling in the presence of asynchrony and the possibility of errors. In the end, I implemented it by tracking active requests in two places — at the LSP manager level, and on the individual language server level. All responses have to go through the LSP manager, and in the end, it allows us to atomically handle both server failures and responses when we have active requests.

Fixes #3885

### DIFF
--- a/editor/resources/templates/template.script
+++ b/editor/resources/templates/template.script
@@ -18,7 +18,8 @@ end
 
 function fixed_update(self, dt)
 	-- This function is called if 'Fixed Update Frequency' is enabled in the Engine section of game.project
-	-- Can be coupled with fixed updates of the physics simulation if 'Use Fixed Timestep' is enabled in Physics section of game.project
+	-- Can be coupled with fixed updates of the physics simulation if 'Use Fixed Timestep' is enabled in
+	-- Physics section of game.project
 	-- Add update code here
 	-- Learn more: https://defold.com/manuals/script/
 	-- Remove this function if not needed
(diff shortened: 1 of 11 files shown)


matgis (approved):
Seems to work great, and the updated `async-build!` interface is a nice improvement!

I did notice that we some minor issues reported from newly-created `script` files. In addition to a bunch of "unused argument" warnings (which I think we shouldn't do anything about - users can either disable the warning or rename the arguments), I also got these two:

```
/main/main.script
	Line 19: setting non-standard global variable 'fixed_update'
/main/main.script
	Line 21: line is too long (133 > 120)
```
* Is `fixed_update` missing from our API somehow?
* Should we update the comment in the `script` template so it doesn't exceed the line length limit?

I'm happy to merge this as-is, though!

Issue #7486, opened 2023-03-20
Changing component in-place hangs IDE


aglitchman:
**Describe the bug (REQUIRED)**
The latest version of the stable IDE becomes unresponsive (hangs) when I try to change a component to another.

**To Reproduce (REQUIRED)**
Take a look at the videos.

How it worked few days ago, i.e. it's the release version of Defold 1.4.3 from https://d.defold.com/stable/ 

https://user-images.githubusercontent.com/193258/226375581-4fa79d69-997b-4541-b6b6-4005982baacf.mp4

How it works in the current version, i.e. I updated the IDE to the latest by clicking "Update Available":

https://user-images.githubusercontent.com/193258/226375463-509c16b0-d5e2-4a60-bfbc-fd3b66e0550e.mp4

**Defold version (REQUIRED):**
  - Version 1.4.3
  - Editor SHA1: c74fd32e9552d8b441eed52a39d98055d03f7bfe

**Platforms (REQUIRED):**
 - OS: Windows 10

I tried two different PCs and the behaviour was similar.

**Minimal repro case project (OPTIONAL):**
[test_replace_model.zip](https://github.com/defold/defold/files/11019442/test_replace_model.zip)

JCash:
The latest dev build with the bug fix is here:
http://d.defold.com/archive/dev/64b5a2527649c0ce3416a3b88a96c785ba7ff496/dev/editor2/Defold-x86_64-win32.zip

Please double check if that version (with the memory crash fix) doesn't solve this for you

aglitchman:
> Please double check if that version (with the memory crash fix) doesn't solve this for you

This build partially fixed the problem that I PMed to you, i.e. the IDE viewport looks okay now in the demo project. But I tried to run the full project in this version and IDE _crashed_. I'll try to find the buggy model (or something like that) and will send you in PM.

The crash log: [hs_err_pid8776.log](https://github.com/defold/defold/files/11019746/hs_err_pid8776.log)


And the build still has this issue, i.e. it hangs, unfortunately.

aglitchman:
And it's not a model loader problem in this issue, i.e. try to swap sprite with sprite, for example. 

[test_replace_sprite.zip](https://github.com/defold/defold/files/11019805/test_replace_sprite.zip)

matgis:
A fix has been merged and you should see an editor update shortly. Thanks for reporting! ❤️

PR #7488, opened 2023-03-21
Fix editor freeze after editing certain properties

matgis:
Fixes https://github.com/defold/defold/issues/7486

### User-facing changes
* The editor no longer freezes after editing certain properties, such as the path of a referenced component.


vlaaad (approved):
`(repeat :lgtm)`

PR #7489, opened 2023-03-21
Fix: Input system can now map already connected gamepads

JCash:


engine/engine/src/engine.cpp
```diff
@@ -1005,6 +1002,11 @@ namespace dmEngine
         input_params.m_RepeatInterval = dmConfigFile::GetFloat(engine->m_Config, "input.repeat_interval", 0.2f);
         engine->m_InputContext = dmInput::NewContext(input_params);
 
+        dmHID::SetGamepadConnectivityCallback(engine->m_HidContext, dmInput::GamepadConnectivityCallback, engine->m_InputContext);
+
+        // Any connected devices are registered here.
+        dmHID::Init(engine->m_HidContext);
```
JCash, line 1008: Making sure the connected gamepads are sent after the callback was set

engine/engine/src/engine.cpp
```diff
@@ -1542,6 +1544,7 @@ namespace dmEngine
                     return;
                 }
 
+                dmInput::Update(engine->m_InputContext);
```
JCash, line 1547: I still don't fully understand the need to have the separation between the input/binding contexts, but to better handle any new connections I added an Update() function.

engine/hid/src/hid.h
```diff
@@ -65,7 +65,7 @@ namespace dmHID
         float m_X, m_Y, m_Z;
     };
 
-    typedef bool (* DMHIDGamepadFunc)(uint32_t, bool, void*);
+    typedef bool (* FHIDGamepadFunc)(uint32_t, bool, void*);
```
JCash, line 68: Naming conventions with `F` as in function (seen elsewhere in the engine)

engine/hid/src/native/hid_native.cpp
```diff
@@ -135,11 +135,15 @@ namespace dmHID
...
-                // The callback isn't ready to consume the event yet, we will have to try later
                 if (!context->m_GamepadConnectivityCallback(gamepad_index, connection_status, context->m_GamepadConnectivityUserdata))
                 {
+                    char buffer[128];
+                    GetGamepadDeviceName(context, gamepad, buffer, (uint32_t)sizeof(buffer));
+                    dmLogWarning("The connection for '%s' was ignored by the callback function!", buffer);
                     return;
                 }
+            } else {
+                dmLogWarning("There was no callback function set to handle the gamepad connection!");
```
JCash, line 146: Clearer messages on what's happening.

engine/input/src/input.cpp
```diff
@@ -356,6 +373,8 @@ namespace dmInput
 
     void DeleteBinding(HBinding binding)
     {
+        binding->m_Context->m_Binding = 0;
```
JCash, line 376: Since we only ever use one Binding, I see no big issue with us doing this connection.

engine/input/src/input.cpp
```diff
@@ -479,11 +498,11 @@ namespace dmInput
         action->m_HasGamepadPacket = 0;
     }
 
-    struct UpdateContext
+    struct UpdateActionContext
```
JCash, line 501: Making it clearer that it's about the actions/binding, and not the input context

engine/input/src/input.cpp
```diff
@@ -1201,7 +1211,19 @@ namespace dmInput
...
+    }
+
+    bool GamepadConnectivityCallback(uint32_t gamepad_index, bool connected, void* userdata)
+    {
+        Context* context = (Context*)userdata;
+        Binding* binding = (Binding*)context->m_Binding;
+        if (!binding)
+        {
+            context->m_UnmappedGamepads.Put(gamepad_index, connected);
+            return true; // Only return false if you wish to override the controller connection status!
```
JCash, line 1223: The input system is still a little bit of a mess.
SImply returning true here "solves" the issue, since the HID gamepad will then get connected, the SetupGamePads will create a binding for that gamepad, and the next UpdateActions will correctly map the content. But the code flow is a bit unintuitive.

So, I think getting a list of pending connections here is a good "documentation" strategy.

Jhonnyg (dismissed):
LGTM

PR #7491, opened 2023-03-21
Show more information in resource errors

vlaaad:
Related to https://github.com/defold/extension-spine/issues/123. Load errors might have useful information, so why not show it?


vlaaad:
Maybe a PR to the wrong branch

PR #7493, opened 2023-03-22
Change min versions: macos to 10.13 and ios to 11.0

AGulev:


AGulev:
Now when we migrated to XCode 14.2 it seems like we need to bump ios version to iOS 11 and MacOS version to 10.13:
image
https://developer.apple.com/support/xcode/

@JCash @britzl This PR change version to 10.9, but probably it should be higher 👆

britzl:
> This PR change version to 10.9, but probably it should be higher 👆

Yes, it looks like it should be increased to 10.13 and not 10.9? And the same on extender then?

AGulev:
> > This PR change version to 10.9, but probably it should be higher 👆
> 
> Yes, it looks like it should be increased to 10.13 and not 10.9? And the same on extender then?

I think so, I made these changes in this PR (also iOS)

britzl:
Also make sure to update the FAQ once we've merged this:

https://github.com/defold/doc/blob/master/docs/en/faq/faq.md

JCash:
I merged the stuff I did too, since I didn't know you were updating these things.

JCash (approved):
lgtm

AGulev:
> I merged the stuff I did too, since I didn't know you were updating these things.

Perfect! Thank you!

PR #7497, opened 2023-03-22
Made the type infos a dynamic struct that is initialized separately for each thread

JCash:


engine/modelc/src/modelimporter_jni.cpp
```diff
@@ -826,70 +840,71 @@ static void CreateSkins(JNIEnv* env, const dmModelImporter::Scene* scene, dmArra
...
+        skins[skin->m_Index] = CreateSkin(env, types, skin);
     }
 }
 
 
 static jobject CreateJavaScene(JNIEnv* env, const dmModelImporter::Scene* scene)
 {
-    InitializeJNITypes(env);
+    TypeInfos types;
+    InitializeJNITypes(env, &types);
```
JCash, line 851: Since this call may come from different threads, we need to handle the data accordingly.

The values of the class types actually differ too, so it's not enough to load them when the library is loaded.

britzl (approved):

PR #7500, opened 2023-03-23
Fix reloading of implicitly stateless nodes

### LINKED ISSUE #7397: after update .spinejson file changes do not get into build
In a large project after updating spinejson file and building project version of the file that gets to build is the one before update (from build cache?).

**To Reproduce (REQUIRED)**
1) build project
2) update spine json file (among whole lot of other files changes) (for example my changes in spinejson file added additional bone)
3) wait for `loading external changes to be done`
4) open corresponding .spinescene file in defold editor - there is version of .spinejson before changes (in my example no additional bone)
5) (in my example additional bone was added to .spinejson file) in corresponding .go file flag `create_go_bones` is true, and yet when i try to access bone game object in runtime i get error along the lines `no such bone` , access to other bones via same code is ok.

defold editor handles changes to spinejson file correctly (with existing build cache)

  - Version [e.g. 1.2.155]

**Platforms (REQUIRED):**
 - Windows
 - OS: Windows 11

On minimal example everything works fine, that is how i found out issue was not in spinejson file and may be in build cache

**Workaround (OPTIONAL):**
remove content of /build folder and rebuild everything

vlaaad:
Technical notes:
Implicitly stateless nodes are nodes that don't have the write and read functions, but have a load function, and are not explicitly marked as stateless. The problem was that `old-sha256` in such a case would be the same as `new-sha256`, since both are just reading and hashing the same resource file.

User-facing changes:
The Defold Editor now properly reloads changes to certain file types, such as `spinejson`, `md`, `ttf`, `fnt`, and `otf` files. Previously, if changes were made to these types of files and a build was performed, the editor would ignore the changes and use the previous version of the file instead. This update fixes this issue and ensures that changes to these file types are properly reflected in the editor.

Fixes #7397

### DIFF
--- a/editor/src/clj/editor/resource_update.clj
+++ b/editor/src/clj/editor/resource_update.clj
@@ -91,12 +91,10 @@
                          (catch Exception _
                            nil))
             new-sha256 (try
-                         (if (and read-fn write-fn)
+                         (when (and read-fn write-fn)
                            (let [source-value (read-fn new-resource)
                                  sanitized-content (write-fn source-value)]
-                             (DigestUtils/sha256Hex ^String sanitized-content))
-                           (with-open [input-stream (io/input-stream new-resource)]
-                             (DigestUtils/sha256Hex ^InputStream input-stream)))
+                             (DigestUtils/sha256Hex ^String sanitized-content)))
                          (catch Exception _
                            nil))]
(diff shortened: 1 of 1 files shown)


matgis (approved):
Ship it!

PR #7503, opened 2023-03-24
Exclude evaluation-context from handler ex-info

matgis:
### Technical changes
* Exclude the `evaluation-context` from `handler` exception info. Previously this resulted in huge amounts of error data being logged whenever a `handler` threw an exception.


vlaaad (approved):
LGTM!

PR #7504, opened 2023-03-24
Fixed root-level game object transforms inside referenced collections when building from the editor

### LINKED ISSUE #7126: Mesh diff transform in editor and in game.
### Expected behaviour
Mesh look same in editor and game
### Actual behaviour
Mesh not rotated in game
Error in console.
```
 Trying to get stream data outside of input DDF array.
 ```

### Steps to reproduce
[bad mesh.zip](https://github.com/defold/defold/files/9937722/bad.mesh.zip)

<hr/>

<table>
<tr><td>Build time</td><td>2022-11-02T11:44:12.649507</td></tr>
<tr><td>Defold editor sha</td><td>1dbf756283bc3aaa74552b8959569f051023b842</td></tr>
<tr><td>Defold engine sha</td><td>1dbf756283bc3aaa74552b8959569f051023b842</td></tr>
<tr><td>Defold version</td><td>1.4.0</td></tr>
<tr><td>GPU</td><td>AMD Radeon(TM) Graphics</td></tr>
<tr><td>GPU Driver</td><td>4.6.14756 Compatibility Profile Context 20.40.64 27.20.14064.3</td></tr>
<tr><td>Java version</td><td>11.0.15+10-LTS</td></tr>
<table>

### PR DESCRIPTION
**Pull request by matgis**
Fixes #7126.

### User-facing changes
* Fixed root-level game object transforms inside referenced collections when building from the editor

### Technical changes
* Introduced a `Pose` type in places where it is important to retain `position`, `rotation`, and `scale` properties.
* We no longer convert `position`, `rotation`, and `scale` into a `Matrix4` which we then decompose into `position`, `rotation`, and `scale` later in the build process. Doing so lost information and resulted in transforms that differ from how Bob was doing it, which was applying parent transforms to `position`, `rotation`, and `scale` in isolation.

### DIFF
--- a/editor/src/clj/editor/pose.clj
+++ b/editor/src/clj/editor/pose.clj
@@ -1,0 +1,255 @@
+(ns editor.pose
+  (:require [clojure.spec.alpha :as s]
+            [editor.math :as math])
+  (:import [javax.vecmath Matrix4d]))
+
+(set! *warn-on-reflection* true)
+(set! *unchecked-math* :warn-on-boxed)
+
+(s/def ::num number?)
+(s/def ::num3 (s/tuple ::num ::num ::num))
+(s/def ::num4 (s/tuple ::num ::num ::num ::num))
+(s/def ::translation ::num3)
+(s/def ::rotation ::num4)
+(s/def ::scale ::num3)
+(s/def ::euler-rotation ::num3)
+
+(defrecord Pose [translation rotation scale])
+
+(defn pose? [value]
+  (instance? Pose value))
+
+(def default-translation (vector-of :double 0.0 0.0 0.0))
+(def default-rotation (vector-of :double 0.0 0.0 0.0 1.0))
+(def default-scale (vector-of :double 1.0 1.0 1.0))
+(def ^Pose default (->Pose default-translation default-rotation default-scale))
+
+(def ^:private mat4-identity (doto (Matrix4d.) (.setIdentity)))
+(def ^:private num4-zero (vector-of :double 0.0 0.0 0.0 0.0))
+
+(defn- significant-translation? [value]
+  (if-some [[^double x ^double y ^double z] value]
+    (not (and (zero? x)
+              (zero? y)
+              (zero? z)))
+    false))
+
+(defn- significant-rotation? [value]
+  (if-some [[^double x ^double y ^double z ^double w] value]
+    (not (and (zero? x)
+              (zero? y)
+              (zero? z)
+              (== 1.0 w)))
+    false))
+
+(defn- significant-scale? [value]
+  (if-some [[^double x ^double y ^double z] value]
+    (not (and (== 1.0 x)
+              (== 1.0 y)
+              (== 1.0 z)))
+    false))
+
+(defn- add-vector [[^double ax ^double ay ^double az] [^double bx ^double by ^double bz]]
+  (vector-of
+    :double
+    (+ ax bx)
+    (+ ay by)
+    (+ az bz)))
+
+(defn- multiply-vector [[^double ax ^double ay ^double az] [^double bx ^double by ^double bz]]
+  (vector-of
+    :double
+    (* ax bx)
+    (* ay by)
+    (* az bz)))
+
+(defn- rotate-vector [[^double vx ^double vy ^double vz] [^double qx ^double qy ^double qz ^double qw]]
+  (vector-of
+    :double
+    (- (+ (* vx qw qw)
+          (* vy qz qw -2.0)
+          (* vz qy qw 2.0)
+          (* vx qx qx)
+          (* vy qx qy 2.0)
+          (* vz qx qz 2.0))
+       (* vx qy qy)
+       (* vx qz qz))
+    (- (+ (* vx qx qy 2.0)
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: editor/src/clj/editor/pose.clj [hunk 1]
```diff
@@ -0,0 +1,255 @@
...
+              (zero? z)
+              (== 1.0 w)))
+    false))
+
+(defn- significant-scale? [value]
+  (if-some [[^double x ^double y ^double z] value]
+    (not (and (== 1.0 x)
+              (== 1.0 y)
+              (== 1.0 z)))
+    false))
```
vlaaad, line 50: I think the `nil` check semantically should belong higher up the stack, e.g. in a place where some pose attribute might be `nil`, we do the check before calling these predicates.

editor/src/clj/editor/pose.clj
```diff
@@ -0,0 +1,255 @@
...
+                   (cond-> child-translation
+                           parent-scaled (multiply-vector parent-scale)
+                           parent-rotated (rotate-vector parent-rotation)
+                           parent-translated (add-vector parent-translation))))))
+
+(defn to-map [^Pose pose translation-key rotation-key scale-key]
+  {:pre [(pose? pose)]}
+  {translation-key (.-translation pose)
+   rotation-key (.-rotation pose)
+   scale-key (.-scale pose)})
```
vlaaad, line 248: Maybe we should instead change the code that uses `to-map` to expect `:translation`/`:rotation`/`:scale` keys? I worry that this fn enables us to introduce more inconsistencies in pose-related code. The pose record is already a map...

Or maybe the pose should use names `:position`/`:rotation`/`:scale3`?

editor/src/clj/editor/pose.clj
```diff
@@ -0,0 +1,255 @@
...
+  ^Pose [^double x ^double y ^double z]
+  (if (and (== 1.0 x) (== 1.0 y) (== 1.0 z))
+    default
+    (->Pose default-translation default-rotation (vector-of :double x y z))))
+
+(defn make
+  ^Pose [translation rotation scale]
+  {:pre [(or (nil? translation) (s/valid? ::translation translation))
+         (or (nil? rotation) (s/valid? ::rotation rotation))
+         (or (nil? scale) (s/valid? ::scale scale))]}
```
vlaaad, line 178: I think we should create and prioritize an issue about disabling asserts in production, and going through our asserts to separate dev-time contracts like these from user input validation. The spec library is somewhat slow...

editor/src/clj/editor/pose.clj
```diff
@@ -0,0 +1,255 @@
...
+       (* vy qx qx)
+       (* vy qz qz))
+    (- (+ (* vx qx qz 2.0)
+          (* vy qy qz 2.0)
+          (* vz qz qz)
+          (* vx qy qw -2.0)
+          (* vy qx qw 2.0)
+          (* vz qw qw))
+       (* vz qx qx)
+       (* vz qy qy))))
```
vlaaad, line 92: Did you write it yourself? I don't know how to verify this code...

vlaaad (approved):
LGTM! I have some comments, but it's all optional :)

editor/src/clj/editor/pose.clj
matgis: These inconsistencies are unfortunately placed upon us by the requirements of the existing file formats. Scale is named `:scale3` in collections because we used to have a uniform `:scale` that was a single `float`, but for game objects the non-uniform (i.e vector triplet) scale field is simply named `:scale`. Likewise, the translation field for both game objects and collections are named `:position`, even though it is in fact a translation along the (possibly rotated and scaled) axes of its parent. It is correctly (in my opinion) named `:translation` in the protobuf `Transform` message, which our Pose type use-case mirrors.

I would have preferred to call the `Pose` record `Transform` in the editor as well. Unfortunately the term `transform` is used to refer to `Matrix4` values everywhere in the editor, and both the Spine and Rive extensions already rely on the `transform` output of the base `SceneNode` they subclass to produce a `Matrix4`. Maybe we can clean this up in the future, but we'd have to do it onto the `dev` since it is a breaking change requiring projects to update their library dependencies.

For now, I prefer to do the map conversion like this. In fact nothing outside of the unit tests make assumptions about what the record fields are called, or even that the Pose is represented as a record. If we ever get around to migrating away from `javax.vecmath`, we might want to represent it differently.

editor/src/clj/editor/pose.clj
matgis: I looked at a couple of implementations of the formula, inlined a bunch of stuff, and reordered the coefficients in a way that made sense to me in prefix notation. I double-checked the math against both implementations I looked at to make sure I didn't mess it up, so I'm pretty confident it should be correct, but I guess time will tell! 😅

editor/src/clj/editor/pose.clj
matgis: Yeah, we probably should. Disabling asserts did not seem to make a lot of difference at the time I profiled it, but I've started to add more checks since then.

But I'm not really worried about this particular case, since we only create `Pose` instances when producing build targets.

PR #7517, opened 2023-03-29
Don't show submenu shortcuts for command options

### LINKED ISSUE #7442: Add Component menu in Collections displays that every objects Hotkey is the same
Sharing hotkeys is not very useful.

vlaaad:
User-facing changes:
Previously, sub-menus like `File -> New` and `Game Object -> Add Component` showed the same shortcut on every option item. This changeset turns this behavior off.

Technical notes:
I implemented the fix by only showing the key combo if the menu item does not define custom user data since this is what differentiates the options from each other. It was wrong to show the key combinations on such options because invoking a command using a key combination does not supply this user data anyway.

Fixes #7442

### DIFF
--- a/editor/src/clj/editor/ui.clj
+++ b/editor/src/clj/editor/ui.clj
@@ -1258,7 +1258,7 @@
     (user-data! menu-item ::menu-item-id id)
     (when command
       (.setId menu-item (name command)))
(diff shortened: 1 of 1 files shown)


matgis (approved):
Ship it!

PR #7520, opened 2023-03-30
Support 'oneof' blocks in our DDF layer

Jhonnyg:
This PR adds support for oneof blocks in protobuf in our engine runtime:

```
oneof values
{
    uint64 val_a = 1;
    int32  val_b = 2;
}
```

Caveats:
* serialization for repeated fields with oneofs doesn't produce the same string as SerializeAsString since we serialize all the fields because we don't know which of them has been set. This is only available if you do reflection on the data, but we don't have that. The values will still be correct, but the serialized string will contain a bit more data.
* Default values for oneof members will not work since we don't know which value to use..

## PR checklist

* [x] Code
* [-] Documentation
	* [-] Make sure that API documentation is updated in code comments
	* [-] Make sure that manuals are updated (in github.com/defold/doc)
* [-] Prepare pull request and affected issue for automatic release notes generator
	* [-] Pull request - Link the pull request to the issue(s) it is closing. Use on of the [approved closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
	* [-] Affected issue - Assign the issue to a project. Do not assign the pull request to a project if there is an issue which the pull request closes.
	* [-] Affected issue - Assign the "breaking change" label to the issue if introducing a breaking change.
	* [-] Affected issue - Assign the "skip release notes" is the issue should not be included in the generated release notes.


engine/ddf/src/ddf/ddf.h
```diff
@@ -51,6 +52,8 @@ namespace dmDDF
         Descriptor* m_MessageDescriptor;
         uint32_t    m_Offset;
         const char* m_DefaultValue;
+        uint8_t     m_OneOfIndex : 7;
+        uint8_t     m_OneOfSet   : 1;
```
Jhonnyg, line 56: We _could_ merge these or just use the 'm_OneOfSet' to distinguish if a field is part of a oneof definition

engine/ddf/src/ddf/ddf_save.cpp
```diff
@@ -38,6 +38,11 @@ namespace dmDDF
             FieldDescriptor* field_desc = &desc->m_Fields[i];
             Type type = (Type) field_desc->m_Type;
 
+            if (field_desc->m_OneOfIndex != DDF_NO_ONE_OF_INDEX && field_desc->m_OneOfSet == 0)
+            {
+                continue;
+            }
```
Jhonnyg, line 44: We skip serialization of the field if it has not been set

engine/ddf/src/ddf/ddf_load.cpp
```diff
@@ -111,6 +111,12 @@ namespace dmDDF
         for (int i = 0; i < desc->m_FieldCount; ++i)
         {
             const FieldDescriptor* f = &desc->m_Fields[i];
+            // We cannot support default values for oneof fields currently as we don't know which one to take
+            if (f->m_OneOfIndex != DDF_NO_ONE_OF_INDEX)
+            {
+                dmLogWarning("Default values for 'oneof' fields are not supported");
```
JCash, line 117: Is it possible to print file + message name + field name?

engine/ddf/src/ddfc.py
```diff
@@ -307,9 +330,25 @@ def to_cxx_descriptor(context, pp_cpp, pp_h, message_type, namespace_lst):
...
     for f in message_type.field:
-        tpl = (f.name, f.number, f.type, f.label)
+        one_of_index = 0
+
+        if f.HasField("oneof_index"):
+            oneof_decl = message_type.oneof_decl[f.oneof_index]
+
+            # indices start at zero, but we need to distinguish between which fields belong
+            # to a oneof block somehow
+            one_of_index = f.oneof_index + 1
```
JCash, line 343: Isn't that the "m_OneOfSet" member?
Since we're doing struct initialixation on line 377, sholdn't it be its own variable?
E.g. number, type and label also are bitfields and are set explicitly.

engine/ddf/src/ddfc.py
```diff
@@ -750,14 +793,13 @@ def get_field_type_name(self, f):
...
-                if x[0].name == 'ddf_namespace':
-                    namespace = x[1]
-
             includes = []
+
             for x in pf.options.ListFields():
                 if x[0].name == 'ddf_includes':
                     includes = x[1].split()
+                elif x[0].name == 'ddf_namespace':
+                    namespace = x[1]
```
JCash, line 802: 👍

engine/ddf/src/test/test_ddf.cpp
```diff
@@ -764,6 +764,148 @@ TEST(AlignmentTests, AlignField)
...
+    uint32_t msg_buf_size = msg_str.size();
+
+    DUMMY::TestDDF::OneOfMessage* message;
+    dmDDF::Result e = dmDDF::LoadMessage((void*) msg_buf, msg_buf_size, &DUMMY::TestDDF_OneOfMessage_DESCRIPTOR, (void**)&message);
+    ASSERT_EQ(dmDDF::RESULT_OK, e);
+    ASSERT_EQ(1337, message->m_OneOfFieldOne.m_IntVal);
+    ASSERT_EQ(99,   message->m_NotInOneof);
+    ASSERT_EQ(10,   message->m_NestedVal.m_NestedFieldName.m_NestedIntVal);
+    ASSERT_EQ(13,   message->m_NestedVal.m_NestedButNotInOneof);
+    ASSERT_STREQ(simple_string_valid.c_str(), message->m_OneOfFieldTwo.m_SimpleStringTwo);
```
JCash, line 797: I'm guessing it will work fine to use the descriptor to actually turn an "unknown" nested structure of OneOf's to figure out what the type of the values are, so that one can get the value properly (e.g. to convert to Lua)

engine/ddf/src/test/test_ddf_proto.proto
```diff
@@ -259,3 +259,94 @@ message TestFieldAlignment
...
+        bool bool_val = 2;
+    }
+
+    oneof one_of_field_string
+    {
+        int32 int_val_no_string = 3;
+        string string_val       = 4;
+    }
+}
+
```
JCash, line 352: I'm not sure we cover our use case here.
What we wanted was a nested (as in recursive type) structure, right?
We modeled it after the [json model](https://www.json.org/json-en.html) if I recall correctly-

E.g. can we do something like this:
```

message Member
{
    oneof value
    {
        int32  int_val    = 1;
        float  float_val  = 2;
        string string_val = 3;
        Object object_val = 4; // recursive type
    }
    
    string name = 5;
}

message Object
{
    repeated Member members = 1;
}
```

JCash (changes requested):
Small initialization fix.

ALso, test should reflect our use case a bit closer to verify our idea.

Jhonnyg:
Closing this in favor of this being merged: https://github.com/defold/defold/tree/protobuf-one-of-support-for-ddf-merge

PR #7539, opened 2023-04-03
Add an option to open a project from the editor

### LINKED ISSUE #5689: Add option to open a project from within the editor
hi guys,
The Editor must quit  to open another project?
thank you.

vlaaad:
User-facing changes: Added a `Open Project...` menu item in the File menu. It will launch another instance of the editor.

Technical notes: The code to launch the process is based on the code in the `updater.clj` file. A notable modification has been made to the file unpacking process. Instead of using a stable folder in the "editor support path" location, we now always unpack files to a temporary folder. It turned out that using the support path was a mistake because it is intended for files that persist between application and system restarts, whereas the unpacked files were always deleted when the editor was closed. This meant that the behavior of the unpacked files was similar to that of a temporary directory, but they were being stored in the wrong place. This shared storage location was causing issues with running multiple instances of the editor at the same time — shutdown of one instance of the editor broke another.

Fixes #5689

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -85,6 +85,7 @@
   (:import [clojure.lang ExceptionInfo]
            [com.defold.editor Editor]
            [com.defold.editor UIUtil]
+           [com.dynamo.bob Platform]
            [com.sun.javafx.scene NodeHelper]
            [java.io BufferedReader File IOException]
            [java.net URL]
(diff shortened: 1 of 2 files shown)


vlaaad:
@matgis I tested it on all 3 OSes (open one instance, open another, close first instance) and it worked.

matgis (changes requested):
Hmm. I don't think we want to change the unpack dir to a temp directory. If you examine the old code, it seems to only unpack the files to a temp directory in dev builds (where sha1 will be `null`).

The release version unpacks the resources once for a particular sha1 of the editor, then keeps the files around. That's probably why they reside in the user's support path and not in a temp directory.

I also found this old commit that seems to have been intended to address this issue in the past. Perhaps we only need to silently fail if the files are already unpacked? https://github.com/defold/defold/commit/9957f5964d37d8e63b465e2e5849e16ddc13f1d1

matgis (approved):
Huh. This seems wrong to me, since it means we have a longer startup time than we need.

I think we should add a task to fix this. In release builds, we shouldn't need to unpack on subsequent startups unless there's been an update. Perhaps we can mark the existing unpack dir for deletion when the user chooses to install an update?

In dev builds I think we should simply unpack to a random temp directory which we delete every time.

Anyway, I guess we can merge it like this and look into this later!

PR #7552, opened 2023-04-05
Gracefully handle the deletion of an edited table cell

### LINKED ISSUE #6924: Error when deleting form entry while editing it
When editing a form such as input bindings and adding a new entry but not filling out all data and immediately deleting the entry will cause a crash

**To Reproduce (REQUIRED)**

1. open game.input_binding file
2. add one action
image
3. when you want to insert the "name" for that action, leave it empty and without blur the textfield click on the minus icon
4. Observe error
image

No crash

  - Version 1.3.4

**Platforms (REQUIRED):**
 - Platforms: any
 - OS: Windows 11
 - Device: Laptop

vlaaad:
User-facing changes: deleting a row with a currently-edited table cell no longers throws an exception.

Technical notes: JavaFX strikes again... This looks like a helpful behavior they introduced in some update — committing a change of a cell that loses focus. Unfortunately, it happens to be executed in the order we didn't anticipate. The user clicks the "Remove" button -> we execute the `:remove-table-selection` handler -> we now get new items -> we set new items on a table view -> table view notices that we edit the table and commits the edit before replacing the cell -> we execute the `:commit-table-edit` handler on a new items vector, but the edit is applied as if to the old items vector. I.e. we track an index to change while editing, and it corresponds to the old vector. So, if we remove the last item that was edited, the index will be out of bounds by one. Applying the edit is `(assoc items index new-value)`, which will act as `conj` in case of off-by-one `assoc`. Just canceling the edit will prevent JavaFX from committing it which is a way out of this mess.

Fixes #6924

### DIFF
--- a/editor/src/clj/editor/cljfx_form_view.clj
+++ b/editor/src/clj/editor/cljfx_form_view.clj
@@ -816,11 +816,21 @@
                                                          on-value-changed
                                                          ui-state
                                                          state-path]}]
-  (let [indices-path (conj state-path :selected-indices)
-        selected-indices (set (get-in ui-state indices-path))
+  (let [{:keys [selected-indices edit] :as state} (get-in ui-state state-path)
+        selected-indices (set selected-indices)
         new-value (remove-indices selected-indices value)]
+    ;; When we remove a row that has been edited, JavaFX will commit the edit.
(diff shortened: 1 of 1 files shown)


matgis (approved):
Seems to work! LGTM!

PR #7553, opened 2023-04-05
Strip brackets from array uniform names during editor shader reflection

matgis:
### Technical changes
* Fixed an issue where array uniform names were incorrectly reported during shader reflection. This allows us to correctly set array uniforms when rendering. This will be used by `extension-rive` in the near future.

### DIFF
--- a/editor/src/clj/editor/gl/shader.clj
+++ b/editor/src/clj/editor/gl/shader.clj
@@ -544,7 +544,8 @@ of GLSL strings and returns an object that satisfies GlBind and GlEnable."
       (aget out-uniform-count 0))))
 
 (def ^:private uniform-info
-  (let [name-buffer-size 128
+  (let [name-array-suffix-pattern #"\[\d+\]$"
+        name-buffer-size 128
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/gl/shader.clj
```diff
@@ -544,7 +544,8 @@ of GLSL strings and returns an object that satisfies GlBind and GlEnable."
       (aget out-uniform-count 0))))
 
 (def ^:private uniform-info
-  (let [name-buffer-size 128
+  (let [name-array-suffix-pattern #"\[\d+\]$"
+        name-buffer-size 128
```
matgis, line 548: OpenGL was reporting array uniforms including a "[0]" suffix to the name. When we apply uniforms from materials, we want to address the uniform by name without this "[0]" suffix.

Here, we strip out the array suffix from the name we associate with the uniform in the editor. This behavior mirrors what we do in the runtime, since users can set uniforms from scripts as well.

vlaaad (approved):
Shipiiiiiit!

PR #7559, opened 2023-04-11
Refactor textures as handles

Jhonnyg:
Refactored how textures and rendertargets are exposed and used in the engine, instead of passing around raw pointers, we now represent them as generic asset handles. This is the first step to bring the render and gameobject world closer together, where graphical assets can be transparently shared between them. Meaning, a texture that has been created by a script (or any texture, e.g an atlas texture) can be passed to the render scripts where it could be bound to the render state for drawing. Similarly, a render target could be created in a render script, passed to a game object and used as a texture on a model or an atlas.

Fixes #7224 

## PR checklist

* [x] Code
* [-] Documentation
	* [-] Make sure that API documentation is updated in code comments
	* [-] Make sure that manuals are updated (in github.com/defold/doc)
* [-] Prepare pull request and affected issue for automatic release notes generator
	* [-] Affected issue - Assign the "skip release notes" is the issue should not be included in the generated release notes.


engine/graphics/src/dmsdk/graphics/graphics.h
```diff
@@ -34,16 +34,21 @@ namespace dmGraphics
      * @typedef
      * @name HContext
      */
-    typedef struct Context* HContext;
+    typedef void* HContext;
```
Jhonnyg, line 37: Maybe not 100% necessary, but It's really hard to debug when it's defined as a struct ptr and linking with multiple graphics adapters.

engine/graphics/src/null/graphics_null_private.h
```diff
@@ -78,10 +79,33 @@ namespace dmGraphics
...
+        }
+
+        union
+        {
+            RenderTarget m_RenderTarget;
+            Texture      m_Texture;
+        };
+
+        AssetType m_Type;
+    };
```
Jhonnyg, line 102: Right now we store all assets in the same container, so we carry around the type information with this shared struct. It's not really used anywhere, except for some validation here and there. This also means that the max number of shared assets is 65k since all of them will live in the same container.

A different approach that I'm thinking about is to represent the handle as a uint64_t instead and pack both the handle and the type into a "mixed handle-metadata" handle and store the handles in separate containers based on type. The benefits of this is that we don't need this extra indirection and don't need to mix handles in the same container. Should we allow exposing more assets like vertex buffers, shaders or uniform buffers in the future it might be nice to have this separation.. The drawback is of course that we "waste" the remaining bits in the handle, but maybe it's worth it?

engine/dlib/src/test/test_opaque_handle_container.cpp
```diff
@@ -19,52 +19,52 @@
 
 TEST(dmOpaqueHandleContainer, Basic)
 {
-	dmOpaqueHandleContainer<int> simple;
-	ASSERT_EQ(simple.Capacity(), 0);
-	ASSERT_TRUE(simple.Full());
+    dmOpaqueHandleContainer<int> simple;
+    ASSERT_EQ(simple.Capacity(), 0);
+    ASSERT_TRUE(simple.Full());
```
Jhonnyg, line 24: just a pet peeve with indentation.. nothing else should have changed in these tests

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -489,106 +506,114 @@ namespace dmGraphics
...
-        rt.m_ColorAttachmentCount = 1;
+        RenderTarget* rt           = new RenderTarget(DM_RENDERTARGET_BACKBUFFER_ID);
+        rt->m_RenderPass           = vulkan_context->m_MainRenderPass;
+        rt->m_Framebuffer          = vulkan_context->m_MainFrameBuffers[0];
+        rt->m_Extent               = vulkan_context->m_SwapChain->m_ImageExtent;
+        rt->m_ColorAttachmentCount = 1;
+
+        VulkanSharedAsset* asset = new VulkanSharedAsset();
+        asset->m_Type            = RESOURCE_TYPE_RENDER_TARGET;
+        asset->m_Asset           = rt;
```
Jhonnyg, line 542: It's a bit annoying to do more allocations now with the shared asset indirection that ends up in the container, maybe it's worth to pack the type into the handle 🤔 Eventually when (if) we change the render script API to generalize the texture functions to bind a texture or render target by handle we need to have the type information I think

engine/hid/src/native/hid_native.cpp
```diff
@@ -213,15 +213,18 @@ namespace dmHID
...
-        delete user_data;
-        g_HidContext->m_NativeContextUserData = 0;
+            for (int i = 0; i < user_data->m_GamepadDrivers.Size(); ++i)
+            {
+                user_data->m_GamepadDrivers[i]->m_Destroy(context, user_data->m_GamepadDrivers[i]);
+            }
+
+            delete user_data;
+            g_HidContext->m_NativeContextUserData = 0;
+        }
```
Jhonnyg, line 227: I think this is just an indentation fix

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
```diff
@@ -591,7 +591,7 @@ namespace dmGameSystem
             TextureResource* texture_res = texture_set_res->m_Texture;
             dmGraphics::HTexture texture = texture_res ? texture_res->m_Texture : 0;
```
JCash, line 592: An option is to store the m_Texture as a Texture and not a void*

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -378,9 +378,9 @@ static void LogFrameBufferError(GLenum status)
     DM_PFNGLCOMPRESSEDTEXSUBIMAGE3DPROC PFN_glCompressedTexSubImage3D = NULL;
 #endif
 
-    Context* g_Context = 0x0;
+    OpenGLContext* g_Context = 0x0;
```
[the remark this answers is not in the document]
JCash: Yeah, in general, we prefer to have the context passed in.
And in most cases it feels like an oversight, so I agree, a refactor at some point seems good.

JCash (changes requested):
lgtm.
I think we can avoid many naming changes.
Also, in some functions, we disregard the incoming context, and use a global variable instead. As discussed, we want less of that.

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -906,7 +906,7 @@ namespace dmGameSystem
             return texture_set_res->m_Texture->m_Texture;
         }
 
-        return (dmGraphics::HTexture) result;
+        return (dmGraphics::HTexture) (size_t) result;
```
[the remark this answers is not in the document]
Jhonnyg: I think because I changed the type of HTexture in the last stages of this feature, this shouldn't be needed anymore!

PR #7562, opened 2023-04-12
Clickable Lua preprocessor build errors

### LINKED ISSUE #7561: Clickable build errors from Lua preprocessor
Repo https://github.com/defold/extension-lua-preprocessor/archive/refs/heads/fix-error-messgae.zip

matgis:
Fixes #7561.

### User-facing changes
* Build errors from `extension-lua-preprocessor` now show more information and are clickable in the Build Errors view.

### DIFF
--- a/editor/src/clj/editor/code/script.clj
+++ b/editor/src/clj/editor/code/script.clj
@@ -37,7 +37,8 @@
             [editor.workspace :as workspace]
             [internal.util :as util]
             [schema.core :as s]
-            [service.log :as log])
+            [service.log :as log]
+            [util.coll :refer [pair]])
   (:import [com.dynamo.lua.proto Lua$LuaModule]
            [com.google.protobuf ByteString]))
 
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/code/script.clj
```diff
@@ -510,11 +514,20 @@
...
+          (log/error :message log-error-message :exception exception)
+          (if-some [[proj-path line-number] (try-parse-file-line exception-message)]
+            (let [project (project/get-project _node-id)
+                  exception-resource (workspace/resolve-resource resource proj-path)
+                  exception-node-id (project/get-resource-node project exception-resource)
+                  error-node-id (or exception-node-id _node-id)
+                  error-resource (if (nil? exception-node-id) resource exception-resource)
+                  error-user-data (r/make-code-error-user-data proj-path line-number)]
+              (g/->error error-node-id :modified-lines :fatal error-resource build-error-message error-user-data))
+            (g/->error _node-id :modified-lines :fatal resource build-error-message)))
```
matgis, line 530: Previously we were reporting the error on the `:build-targets` output, but it really is the `:modified-lines` property that is the source of these errors, so I changed it.

I don't think the label is used for anything in the error - it's mostly a mechanism for us to diagnose problems when developing.

editor/src/clj/editor/code/script.clj
```diff
@@ -496,6 +493,13 @@
               (util/distinct-by :name))
         (:script-properties lua-info)))
 
+(def ^:const file-line-pattern #"(?<=^|\s|[<\"'`])(\/[^\s>\"'`:]+)(?::?)(\d+)?")
```
vlaaad, line 496: That's hard to read, though I managed to understand it with regex101 :)

vlaaad (approved):
LGTM!

PR #7563, opened 2023-04-12
Copy HTML5 symbols to parent folder of application bundle

### LINKED ISSUE #7448: Make sure that users understand how much data will be downloaded for the end user in html5 bundle
**Task (REQUIRED):**
Sometimes user make their estimations based on bundle folder size which is wrong. We should find a way to notify users about how much data the end user will download. A few suggestions: 

- [ ] Add a special warning text on a bundle window for html5 about it with a ref to the manual
- [ ] Show a message in console with the estimated downloaded size of the bundle 
- [ ] An option to build only `wasm` or only `asm` 
- [ ] Remove `asm` version of the engine fully 
- [ ] save symbols outside of the main bundle folder

Better understanding how big the html5 Defold bundle is and how much the end user of the game will have to download.

britzl:
When bundling an HTML5 build with debug symbols enabled the symbols are no longer copied into the bundle folder. This will prevent accidental upload of symbols to cloud storage or HTML5 game distribution platform. Note that this in itself is not a problem, but it will make the build look larger than it is. This change is also in line with how the symbols are copied for other platforms.

Fixes #7448

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/HTML5Bundler.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/HTML5Bundler.java
@@ -306,14 +306,18 @@ public class HTML5Bundler implements IBundler {
         }
 
         // Copy debug symbols if they were generated
+        String symbolsName = "dmengine.js.symbols";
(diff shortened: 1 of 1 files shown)


AGulev (approved):
love it!

PR #7566, opened 2023-04-12
Show project template URLs on editor Welcome screen

### LINKED ISSUE #7537: Display download size in new project selection for templates/tutorials/samples
Once in awhile users and education administrators that want to add defold to the classroom find the templates/tutorials/sample projects are downloaded when chosen and may conflict with firewalls. `Host Unreachable`

## **∙ solution**

In editor adding "Download size: " text to each project in "New Project" menu. This will give users a visual indicator that they are downloading the project and also the download size of the project.

example:
launcher_downloadsize_1

Or displaying the url and download size.  Letting users know the project downloads are coming from github.

launcher_downloadsize_2

britzl:
The project templates that are used on the editor Welcome screen are downloaded from GitHub when used. This change shows the URL of each template project in the list. The URL is clickable and will download the template project if selected. This will help users understand that the templates are not stored locally with the editor but downloaded from the internet, and thus require an internet connection to use.

Fixes #7537

### DIFF
--- a/editor/src/clj/editor/welcome.clj
+++ b/editor/src/clj/editor/welcome.clj
@@ -46,7 +46,7 @@
            [javafx.event Event]
            [javafx.geometry Pos]
            [javafx.scene Node Parent Scene]
-           [javafx.scene.control Button ButtonBase Label ListView ProgressBar RadioButton TextArea TextField ToggleGroup]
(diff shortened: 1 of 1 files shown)


vlaaad (approved):
LGTM!

PR #7572, opened 2023-04-14
Various camera editor fixes

matgis:
Some fixes to @Jhonnyg camera frustum preview PR.


editor/scripts/camera-preview/generate_preview_data.py
```diff
@@ -35,7 +35,7 @@ def run_blender_script(path, args):
 
 def main():
     scene_input  = "preview_camera_scene.blend"
-    scene_output = "../../resources/camera-preview-edge-list.json"
+    scene_output = "../../resources/meshes/camera-preview-edge-list.json"
```
matgis, line 38: I moved the output to `resources/meshes`.

editor/src/clj/editor/camera.clj
```diff
@@ -60,26 +60,22 @@
...
-        c    (/ (- (+     near far)) (- far near))
-        d    (/ (- (* 2.0 near far)) (- far near))
-        m    (Matrix4d.)]
+        x (/ (* 2.0 near) (- xmax xmin))
+        y (/ (* 2.0 near) (- ymax ymin))
+        a (/ (+ xmin xmax) (- xmax xmin))
+        b (/ (+ ymin ymax) (- ymax ymin))
+        c (/ (- (+ near far)) (- far near))
+        d (/ (- (* 2.0 near far)) (- far near))
+        m (Matrix4d.)]
```
matgis, line 78: Broke the math out to a function we can share between `camera.clj` and `camera-editor.clj`.

editor/src/clj/editor/camera_editor.clj
```diff
@@ -34,19 +35,21 @@
            [javax.vecmath Matrix4d Vector3d Vector4d Quat4d]))
 
 (set! *warn-on-reflection* true)
+(set! *unchecked-math* :warn-on-boxed)
 
 (def camera-icon "icons/32/Icons_20-Camera.png")
 
-(defn- get-camera-mesh-lines-json [path]
+(defn- read-json-resource [path]
```
matgis, line 44: Renamed to reflect that it is actually a generic json-reading function.

editor/src/clj/editor/camera_editor.clj
```diff
@@ -34,19 +35,22 @@
...
-        camera-mesh-lines-v4 (map (fn [p]
-                                    (Vector4d. (get p 0) (get p 1) (get p 2) 1))
-                                  camera-edge-list)]
+
+(def ^:private camera-mesh-lines
+  (let [camera-name+edge-list (first (read-json-resource camera-mesh-json)) ; First mesh name + edge-list.
+        camera-edge-list (second camera-name+edge-list)
+        camera-mesh-lines-v4 (mapv (fn [p]
+                                     (Vector4d. (p 0) (p 1) (p 2) 1.0))
+                                   camera-edge-list)]
```
matgis, line 53: * Made the var private to this file.
* Renamed some stuff for clarity.
* Use `(p 0)` syntax to access Clojure vector elements.
* Use `mapv` instead of `map` to ensure the transformation happens immediately - not when first iterated over.

editor/src/clj/editor/camera_editor.clj
```diff
@@ -34,19 +35,22 @@
            [javax.vecmath Matrix4d Vector3d Vector4d Quat4d]))
 
 (set! *warn-on-reflection* true)
+(set! *unchecked-math* :warn-on-boxed)
```
matgis, line 38: Turned on boxed math warnings for the file.

Boxed math happens when you have a `Double` (wrapper deriving from Object) instead of a `double` (primitive value) involved in math.

Clojure collections can store any `Object`, but primitives need to be "boxed" inside a `Double` or a `Long` wrapper in order to be stored alongside other `Objects` inside Clojure collections.

This means you'll get a boxed value back whenever you extract a primitive from a collection. Function arguments will also be boxed unless you explicitly tag them as `^double` or `^long` (however, there's a limit to only four arguments for a function with primitive arguments).

Boxed math is orders-of-magnitude slower than primitive math, so we want to avoid it if we can.

Turning on `:warn-on-boxed` will print a warning when you reload the file for each line where boxed math is happening. I've gone over the code and addressed places where boxed math was occurring.

editor/src/clj/editor/camera_editor.clj
```diff
@@ -129,14 +133,14 @@
 
 (def outline-shader (shader/make-shader ::outline-shader outline-vertex-shader outline-fragment-shader))
 
-(defn- gen-outline-vertex [x y z w cr cg cb]
-  [(/ x w) (/ y w) (/ z w) cr cg cb 1])
+(defmacro ^:private gen-outline-vertex [x y z w cr cg cb]
+  `[(/ ~x ~w) (/ ~y ~w) (/ ~z ~w) ~cr ~cg ~cb 1.0])
```
matgis, line 137: Turning this into a macro avoids having to box the primitive arguments for the function call.

editor/src/clj/editor/camera_editor.clj
```diff
@@ -129,14 +133,14 @@
...
 
 (defn- conj-camera-mesh-vertices! [vbuf ^Matrix4d inv-view cr cg cb]
   (mapv (fn [^Vector4d p]
           (let [tp (math/transform-vector-v4 inv-view p)
                 camera-mesh-vx (gen-outline-vertex (.x tp) (.y tp) (.z tp) (.w tp) cr cg cb)]
-            (-> vbuf (conj! camera-mesh-vx))))
+            (conj! vbuf camera-mesh-vx)))
```
matgis, line 143: We typically only resort to the threading macro (`->`) when there are several steps.

editor/src/clj/editor/camera_editor.clj
```diff
@@ -160,9 +164,9 @@
         near-v2 (gen-outline-vertex (.x near-p2) (.y near-p2) (.z near-p2) (.w near-p2) cr cg cb)
         near-v3 (gen-outline-vertex (.x near-p3) (.y near-p3) (.z near-p3) (.w near-p3) cr cg cb)
 
-        cr-less (* cr 0.75)
-        cg-less (* cg 0.75)
-        cb-less (* cb 0.75)
+        cr-less (* ^double cr 0.55)
+        cg-less (* ^double cg 0.55)
+        cb-less (* ^double cb 0.55)
```
matgis, line 169: I really liked how the part of the frustum that was outside of the clipping volume was dimmed out, but felt it was a bit too subtle, so I dimmed it a bit more.

editor/src/clj/editor/camera_editor.clj
```diff
@@ -190,6 +194,10 @@
         (conj! camera-v) (conj! camera-near-v2)
         (conj! camera-v) (conj! camera-near-v3))))
 
+(def ^:private ^:const camera-frustum-vertices-count 32) ; This should correspond to the number of conj! in the last expression above.
+
+(def ^:private ^:const camera-preview-mesh-vertices-count (+ camera-frustum-vertices-count (count camera-mesh-lines)))
+
```
matgis, line 200: Made this into a shared constant, like you suggested.

editor/src/clj/editor/camera_editor.clj
```diff
@@ -205,115 +213,46 @@
...
-    (set! (. m m30) 0.0)
-    (set! (. m m31) 0.0)
-    (set! (. m m32) 0.0)
-    (set! (. m m33) 1.0)
-    m))
-
-(defn- camera-projection-matrix [is-orthographic near far fov]
-  (if (true? is-orthographic)
-    (camera-orthographic-projection-matrix near far fov)
-    (camera-perspective-projection-matrix near far fov)))
```
matgis, line 280: Moved the math to `camera.clj` as it is now shared with that file.

Note that the previous implementation took `fov` in radians. The shared implementation uses degrees, so we need to convert from radians to degress when we generate the `user-data` for our renderable.

We should probably do this at load-time and have the users edit the fov as degrees, but I feel a bit apprehensive about the potential floating-point conversion issues, so I'd like to do that as a separate PR later.

editor/src/clj/editor/camera_editor.clj
```diff
@@ -205,115 +213,46 @@
...
+(defn- camera-projection-matrix [is-orthographic near far fov-deg aspect-ratio]
+  ;; TODO: Derive aspect-ratio from display setting in game.project when auto-aspect-ratio is enabled.
+  (let [fov-y (double fov-deg)
+        fov-x (* fov-y (double aspect-ratio))]
+    (if (true? is-orthographic)
+      (camera/simple-orthographic-projection-matrix near far fov-x fov-y)
+      (camera/simple-perspective-projection-matrix near far fov-x fov-y))))
```
matgis, line 222: Use the `aspect-ratio` when calculating the horizontal fov. Still TODO: Support `auto-aspect-ratio`.

editor/src/clj/editor/camera_editor.clj
```diff
@@ -205,115 +213,46 @@
...
-              world-rotation (:world-rotation renderable)
-              ^Matrix4d proj-matrix (camera-projection-matrix (:is-orthographic user-data) (:near-z user-data) (:far-z user-data) (:fov user-data))
-              ^Matrix4d view-matrix (camera-view-matrix world-translation world-rotation)
-              ^Matrix4d inv-view-matrix (math/inverse view-matrix)
-              ^Matrix4d proj-view-matrix (doto (Matrix4d. proj-matrix) (.mul view-matrix))
-              ^Matrix4d inv-view-proj-matrix (math/inverse proj-view-matrix)]
-          (recur (rest renderables) (conj-camera-outline! vbuf world-translation inv-view-matrix inv-view-proj-matrix cr cg cb)))
-        (persistent! vbuf)))))
-
-(defn- render-frustum-outlines [^GL2 gl render-args renderables count]
```
matgis, line 304: Mostly indentation due to not needing the outer `let` to calculate the vertex count, as it is a constant now.

editor/src/clj/editor/camera_editor.clj
```diff
@@ -205,115 +213,46 @@
...
+  [renderables ^long renderable-count]
+  (loop [renderables renderables
+         vbuf (->color-vtx (* renderable-count camera-preview-mesh-vertices-count))]
+    (if-let [renderable (first renderables)]
+      (let [color (if (:selected renderable) colors/selected-outline-color colors/outline-color)
+            cr (get color 0)
+            cg (get color 1)
+            cb (get color 2)
+            {:keys [is-orthographic near-z far-z fov aspect-ratio]} (:user-data renderable)
```
matgis, line 233: I use [destructuring](https://clojure.org/guides/destructuring) here to unpack the `user-data`.

editor/src/clj/editor/camera_editor.clj
```diff
@@ -337,8 +277,8 @@
     :fov (:fov camera)
     :near-z (:near-z camera)
     :far-z (:far-z camera)
-    :auto-aspect-ratio (not (zero? (:auto-aspect-ratio camera)))
-    :orthographic-projection (not (zero? (:orthographic-projection camera)))
+    :auto-aspect-ratio (not= 0 (:auto-aspect-ratio camera))
+    :orthographic-projection (not= 0 (:orthographic-projection camera))
```
matgis, line 281: The use of `zero?` was causing a boxed math warning. This does not, since it can compare anything to anything. Thats's fine here, since this is only called at load-time.

vlaaad (approved):

editor/src/clj/editor/camera_editor.clj
matgis: I don't think the check happens when the value is boxed, but when a boxed value is unboxed for a primitive operation. For `zero?`, I would imagine it does a primitive comparison internally.

PR #7362, opened 2023-02-03
Add "texture_width" and "texture_height" to "resource.get_atlas" return

DreadPirateSergey:
## PR checklist

* [ ] Code
	* [ ] Add engine and/or editor unit tests.
	* [x] New and changed code follows the overall code style of existing code
	* [ ] Add comments where needed
* [ ] Documentation
	* [x] Make sure that API documentation is updated in code comments
	* [ ] Make sure that manuals are updated (in github.com/defold/doc)
* [ ] Prepare pull request and affected issue for automatic release notes generator
	* [ ] Pull request - Write a message that explains what this pull request does. What was the problem? How was it solved? What are the changes to APIs or the new APIs introduced? This message will be used in the generated release notes. Make sure it is well written and understandable for a user of Defold.
	* [ ] Pull request - Write a pull request title that in a sentence summarises what the pull request does. Do not include "Issue-1234 ..." in the title. This text will be used in the generated release notes.
	* [ ] Pull request - Link the pull request to the issue(s) it is closing. Use on of the [approved closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
	* [ ] Affected issue - Assign the issue to a project. Do not assign the pull request to a project if there is an issue which the pull request closes.
	* [ ] Affected issue - Assign the "breaking change" label to the issue if introducing a breaking change.
	* [ ] Affected issue - Assign the "skip release notes" is the issue should not be included in the generated release notes.

PR #7578, opened 2023-04-15
Resource get texture info

### LINKED ISSUE #7564: Functionality to get texture size of atlases
Is it possible to add some functionality to get texture size of atlas?

The problem is: Sometimes when working with shaders it is necessary to know the exact size of the atlas texture for calculations. For example, to draw outline of the same thickness for sprites located in different atlases. As long as one atlas is used, there are no problems, but if there are several atlases, and they are of different sizes (for example, 2048 * 2048 and 1024 * 1024), the algorithms will draw outline of different thicknesses.

Jhonnyg:
Added the new function `resource.get_texture_info` to get metadata from a texture resource or handle:

```
local my_texture_info = resource.get_texture_info(my_texture_path)

-- my_texture_info is a table that contains
{
     handle, -- the numeric handle of the texture
     width, -- width of the texture
     height, -- height of the texture
     depth, -- depth of the texture (1 for a 2D texture, 6 for a cube map and [1...slice-count] for an array texture)
     mipmaps, -- number of mipmaps the texture has
     type -- texture type
}
```

Fixes #7564
Fixes #7362

## PR checklist

* [x] Code
* [x] Documentation

### DIFF
--- a/engine/gamesys/src/gamesys/scripts/script_resource.cpp
+++ b/engine/gamesys/src/gamesys/scripts/script_resource.cpp
@@ -215,6 +215,7 @@ namespace dmGameSystem
 struct ResourceModule
 {
     dmResource::HFactory m_Factory;
+    dmGraphics::HContext m_GraphicsContext;
 } g_ResourceModule;
 
 static int ReportPathError(lua_State* L, dmResource::Result result, dmhash_t path_hash)
@@ -522,7 +523,7 @@ static dmGraphics::TextureImage::TextureFormat GraphicsTextureFormatToImageForma
     return (dmGraphics::TextureImage::TextureFormat) -1;
 }
 
-static dmGraphics::TextureImage::Type GraphicsTextureTypeToImageType(int texturetype)
+static dmGraphics::TextureImage::Type GraphicsTextureTypeToImageType(dmGraphics::TextureType texturetype)
 {
     switch(texturetype)
     {
@@ -714,11 +715,17 @@ static int CreateTexture(lua_State* L)
     dmGameObject::HCollection collection    = dmGameObject::GetCollection(sender_instance);
 
     luaL_checktype(L, 2, LUA_TTABLE);
-    uint32_t type        = (uint32_t) CheckTableInteger(L, 2, "type");
-    uint32_t width       = (uint32_t) CheckTableInteger(L, 2, "width");
-    uint32_t height      = (uint32_t) CheckTableInteger(L, 2, "height");
-    uint32_t format      = (uint32_t) CheckTableInteger(L, 2, "format");
-    uint32_t max_mipmaps = (uint32_t) CheckTableInteger(L, 2, "max_mipmaps", 0);
+    dmGraphics::TextureType type = (dmGraphics::TextureType) CheckTableInteger(L, 2, "type");
+    uint32_t width               = (uint32_t) CheckTableInteger(L, 2, "width");
+    uint32_t height              = (uint32_t) CheckTableInteger(L, 2, "height");
+    uint32_t format              = (uint32_t) CheckTableInteger(L, 2, "format");
+    uint32_t max_mipmaps         = (uint32_t) CheckTableInteger(L, 2, "max_mipmaps", 0);
+
+    // TODO: Texture arrays
+    if (!(type == dmGraphics::TEXTURE_TYPE_2D || type == dmGraphics::TEXTURE_TYPE_CUBE_MAP))
+    {
+        return luaL_error(L, "Unable to create texture, unsupported texture type '%s'.", dmGraphics::GetTextureTypeLiteral(type));
+    }
 
     dmGraphics::TextureImage::CompressionType compression_type = (dmGraphics::TextureImage::CompressionType) CheckTableInteger(L, 2, "compression_type", (int) dmGraphics::TextureImage::COMPRESSION_TYPE_DEFAULT);
 
@@ -755,7 +762,7 @@ static int CreateTexture(lua_State* L)
     if ((tex_type == dmGraphics::TextureImage::TYPE_CUBEMAP || tex_type == dmGraphics::TextureImage::TYPE_2D_ARRAY) && compression_type != dmGraphics::TextureImage::COMPRESSION_TYPE_DEFAULT)
     {
         return luaL_error(L, "Compression type %d requested for texture %s with type '%s', but this is currently not supported.",
-            (int) compression_type, path_str, dmGraphics::GetTextureTypeLiteral((dmGraphics::TextureType) type));
+            (int) compression_type, path_str, dmGraphics::GetTextureTypeLiteral(type));
     }
 
     MakeTextureImage(width, height, max_mipmaps, tex_bpp, tex_type, tex_format, compression_type, buffer, &texture_image);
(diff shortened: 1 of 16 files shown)


engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -2573,6 +2718,7 @@ static void LuaInit(lua_State* L)
 
     SETGRAPHICSCONSTANT(TEXTURE_TYPE_2D);
     SETGRAPHICSCONSTANT(TEXTURE_TYPE_CUBE_MAP);
+    SETGRAPHICSCONSTANT(TEXTURE_TYPE_2D_ARRAY);
```
Jhonnyg, line 2721: Added because even if we don't support creating dynamic array textures (yet), we can still get the info from a texture resource that has been created as an array texture!

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -500,25 +500,6 @@ static void LogFrameBufferError(GLenum status)
         return GL_FALSE;
     }
 
-    static Type ShaderDataTypeToGraphicsType(ShaderDesc::ShaderDataType shader_type)
```
Jhonnyg, line 503: unused

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -3635,6 +3644,27 @@ namespace dmGraphics
...
+        if (type == ASSET_TYPE_TEXTURE)
+        {
+            return GetAssetFromContainer<Texture>(context->m_AssetHandleContainer, asset_handle) != 0;
+        }
+        else if (type == ASSET_TYPE_RENDER_TARGET)
+        {
+            return GetAssetFromContainer<RenderTarget>(context->m_AssetHandleContainer, asset_handle) != 0;
+        }
+        return false;
+    }
```
Jhonnyg, line 3666: we should generalize this to just take a reference to the container, all the adapters will do this identically

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -714,11 +715,17 @@ static int CreateTexture(lua_State* L)
...
+    uint32_t width               = (uint32_t) CheckTableInteger(L, 2, "width");
+    uint32_t height              = (uint32_t) CheckTableInteger(L, 2, "height");
+    uint32_t format              = (uint32_t) CheckTableInteger(L, 2, "format");
+    uint32_t max_mipmaps         = (uint32_t) CheckTableInteger(L, 2, "max_mipmaps", 0);
+
+    // TODO: Texture arrays
+    if (!(type == dmGraphics::TEXTURE_TYPE_2D || type == dmGraphics::TEXTURE_TYPE_CUBE_MAP))
+    {
+        return luaL_error(L, "Unable to create texture, unsupported texture type '%s'.", dmGraphics::GetTextureTypeLiteral(type));
+    }
```
Jhonnyg, line 728: I want to do this in a separate PR, so just drop an error for now..

engine/graphics/src/graphics.h
```diff
@@ -216,33 +219,34 @@ namespace dmGraphics
...
-        const void* m_Data;
-        uint32_t m_DataSize;
-        uint16_t m_MipMap;
-        uint16_t m_Width;
-        uint16_t m_Height;
-        uint16_t m_Depth; // For array texture, this is slice count
+        TextureWrap   m_UWrap;
+        TextureWrap   m_VWrap;
+        const void*   m_Data;
+        uint32_t      m_DataSize;
```
Jhonnyg, line 238: Sorted on size

engine/render/src/render/render_script.cpp
```diff
@@ -1080,7 +1080,7 @@ namespace dmRender
             luaL_error(L, "Unable to create render target.");
         }
 
-        lua_pushinteger(L, render_target);
+        lua_pushnumber(L, render_target);
```
Jhonnyg, line 1083: win32 fix, integer is pushed as 32b and number as 64

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1003,6 +1016,131 @@ static int SetTexture(lua_State* L)
     return 0;
 }
 
+/*# get texture info
+ * Gets texture info from a texture resource path or a texture handle
+ *
+ * @name resource.get_texture_info
+ *
+ * @param path [type:hash|string|handle] The path to the resource or a texture handle
```
Jhonnyg, line 1024: Note to reviewer: the function supports both a path and a handle because I think it will make sense to query handles created from elsewhere, but now that I think about it maybe it should be in a different namespace?

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1003,6 +1016,131 @@ static int SetTexture(lua_State* L)
...
+    uint32_t texture_width               = dmGraphics::GetTextureWidth(texture_handle);
+    uint32_t texture_height              = dmGraphics::GetTextureHeight(texture_handle);
+    uint32_t texture_depth               = dmGraphics::GetTextureDepth(texture_handle);
+    uint32_t texture_mipmaps             = dmGraphics::GetTextureMipmapCount(texture_handle);
+    dmGraphics::TextureType texture_type = dmGraphics::GetTextureType(texture_handle);
+
+    lua_newtable(L);
+
+    lua_pushnumber(L, texture_handle);
+    lua_setfield(L, -2, "handle");
```
JCash, line 1123: Is this really safe for all possible 64 bit signed integers?
Some of those bit patterns will represent NaN's etc, are we guaranteed to get the same number back? And does it work the same way for both Lua 5.1 and LuaJIT?

I'm wondering if using it as a light user data instead removes those questions altogether:
https://www.lua.org/manual/5.1/manual.html#lua_pushlightuserdata

engine/graphics/src/graphics.h
```diff
@@ -216,33 +219,34 @@ namespace dmGraphics
...
-        bool m_SubUpdate;
         uint32_t m_X;
         uint32_t m_Y;
         uint32_t m_Z; // For array texture, this is the slice to set
+
+        uint16_t m_Width;
+        uint16_t m_Height;
+        uint16_t m_Depth; // For array texture, this is slice count
+        uint8_t  m_MipMap    : 7;
+        uint8_t  m_SubUpdate : 1;
```
JCash, line 249: What's the estimated upper limit of the depth? 255?
Then it could be:
```
uint16_t m_Depth : 8;
uint16_t m_MipMap : 7;
uint16_t m_SubUpdate : 1;
```

engine/graphics/src/graphics.h
JCash: Not all the way though.
The pointer is potentially larger than the enums.

engine/render/src/render/render_script.cpp
JCash: Same concern here then (re: light user data)

JCash (changes requested):
I'm a bit concerned with using our large handles as doubles.
Pleasy clarify (with a comment in code) that this is safe, or possibly use light user data

engine/graphics/src/graphics.h
Jhonnyg: The upper depth limit is dictated by the graphics card vendors for 3D/array textures, I don't think we have an absolute max for depth slices in our engine. We loosely support 16k textures so the depth should also adhere to that limit

engine/gamesys/src/gamesys/scripts/script_resource.cpp
Jhonnyg: Hmm, that's a good point. The reason I opted for the 64b handle was to store the type information in the handle itself so we know what it's supposed to represent. I.e if you create a texture via the resource namespace and pass it to another system (render script most likely) and call render.enable_texture with the handle we need to be able to distinguish between a render target and a texture. Alternatively we can represent the handle as a 32b value and store the type information in the graphics adapters, but it's a less 'neat' solution since we need another indirection just to store that extra bit (or maybe bits in the future). Maybe I'm wrong but with lightuserdata I'm not sure we solve this issue anyway since we don't serialize that type in the message system right? And they are 32b on win32 so I guess the only thing it helps with in that case is to preserve the bits in case lua does something strange with integers

engine/graphics/src/graphics.h
Jhonnyg: Fixed

Jhonnyg:
After some discussions with @JCash , we concluded that we can use doubles for the handles so we should be all good!

PR #7580, opened 2023-04-16
Use ANTLR instead of RegExp in LuaScanner

AGulev:
Here I decided to try an approach where I use manual step-in because we have expectations for each case. Maybe this approach isn't optimal, but a bit easier to think about it this way (similar to previous Implementation with regexps).

I don't know if it looks any better than the old one, I want you @britzl to take a look, what you think.

Also, I can try another approach, where I'll be using callbacks, collect data from them into a struct, and then analyze this data in `exitFunctioncall` callback. I think in such case I'll be able to provide better error message for each property.
This approach going to be a bit harder to follow because of a few callbacks where I fill data structure, but at the same time a bit more informative which mean we will be able to provide better error messages if we want so.

## PR checklist

* [x] Code
* [ ] Documentation

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaBuilderTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaBuilderTest.java
@@ -56,6 +56,7 @@ public class LuaBuilderTest extends AbstractProtoBuilderTest {
         src.append("go.property(\"url\", msg.url())\n");
         src.append("go.property(\"vec3\", vmath.vector3(1, 2, 3))\n");
         src.append("go.property(\"vec3empt\", vmath.vector3())\n");
+        src.append("go.property(\"vec3negative\", vmath.vector3(-1))\n");
         src.append("go.property(\"vec4\", vmath.vector4(4, 5, 6, 7))\n");
         src.append("go.property(\"quat\", vmath.quat(8, 9, 10, 11))\n");
         src.append("go.property(\"bool\", true)\n");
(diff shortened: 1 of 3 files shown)


com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
```diff
@@ -234,7 +234,7 @@ public void testPropsVec4() throws Exception {
     public void testPropsQuat() throws Exception {
         List<Property> properties = getPropertiesFromFile("test_props_quat.lua");
 
-        assertEquals(2, properties.size());
+        assertEquals(3, properties.size());
```
AGulev, line 237: Currently because of using RegExp we just ignore such cases and just bundle without errors :

>go.property("prop3", vmath.quat(1a, 2, 3, 4))

Using ANTLR - user will see the error
```
Only these types are available: number, hash, msg.url, vmath.vector3, vmath.vector4, vmath.quat, resource.*
```

So, the property is here, but with an error status, that's why I've fixed test  here.

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaBuilderTest.java
```diff
@@ -55,6 +55,7 @@ public void testProps() throws Exception {
         src.append("go.property(\"hash\", hash(\"hash\"))\n");
         src.append("go.property(\"url\", msg.url())\n");
         src.append("go.property(\"vec3\", vmath.vector3(1, 2, 3))\n");
+        src.append("go.property(\"vec3empt\", vmath.vector3())\n");
```
AGulev, line 58: Make sure that parsing of empty vectors works fine

britzl:
> I don't know if it looks any better than the old one, I want you @britzl to take a look, what you think.

I like the fact that we use antlr to parse the files. Definitely more robust, and easier to expand upon without regex hell.

👍

AGulev:
@britzl could you pls review it?

britzl (approved):

PR #7583, opened 2023-04-17
Support texture handles in render.enable_texture

Jhonnyg:
Render scripts now support enabling texture handles via the `render.enable_texture` function. This means that you can pass a texture handle from scripts to the renderer and use them as global textures when rendering:

```
-- in a regular gameobject script:
local my_texture_info = resource.get_texture_info("/my_atlas.a.texturesetc")
msg.post("@render:", "set_my_texture", { handle = my_texture_info.handle })

-- in a render script:
function update(self, dt)
    render.enable_texture(0, self.my_texture)
end

function on_message(self, message_id, message)
    if message_id == hash("set_my_texture") then
        self.my_texture = message.handle
    end
end
```

## PR checklist

* [x] Code
* [ ] Documentation

### DIFF
--- a/engine/graphics/src/graphics.h
+++ b/engine/graphics/src/graphics.h
@@ -17,9 +17,11 @@
 
 #include <stdint.h>
 #include <dmsdk/vectormath/cpp/vectormath_aos.h>
-
 #include <dmsdk/graphics/graphics.h>
+
 #include <dlib/hash.h>
+#include <dlib/opaque_handle_container.h>
+
 #include <ddf/ddf.h>
 #include <graphics/graphics_ddf.h>
 
@@ -632,7 +634,20 @@ namespace dmGraphics
     uint16_t    GetMipmapSize(uint16_t size_0, uint8_t mipmap);
     uint8_t     GetMipmapCount(uint16_t size);
 
-    bool IsAssetHandleValid(HContext context, HAssetHandle asset_handle);
+
+    // Asset handle helpers
+    const char* GetAssetTypeLiteral(AssetType type);
+    bool        IsAssetHandleValid(HContext context, HAssetHandle asset_handle);
+
+    static inline AssetType GetAssetType(HAssetHandle asset_handle)
+    {
+        return (AssetType) (asset_handle >> 32);
+    }
+
+    static inline HOpaqueHandle GetOpaqueHandle(HAssetHandle asset_handle)
+    {
+        return (HOpaqueHandle) asset_handle & 0xFFFFFFFF;
+    }
 
     /**
      * Get status of texture.
--- a/engine/graphics/src/graphics.h
+++ b/engine/graphics/src/graphics.h
@@ -212,13 +212,13 @@ namespace dmGraphics
     struct TextureParams
     {
         TextureParams()
-        : m_Format(TEXTURE_FORMAT_RGBA)
+        : m_Data(0x0)
+        , m_DataSize(0)
+        , m_Format(TEXTURE_FORMAT_RGBA)
         , m_MinFilter(TEXTURE_FILTER_LINEAR_MIPMAP_NEAREST)
         , m_MagFilter(TEXTURE_FILTER_LINEAR)
         , m_UWrap(TEXTURE_WRAP_CLAMP_TO_EDGE)
         , m_VWrap(TEXTURE_WRAP_CLAMP_TO_EDGE)
-        , m_Data(0x0)
-        , m_DataSize(0)
         , m_X(0)
         , m_Y(0)
         , m_Z(0)
@@ -229,13 +229,13 @@ namespace dmGraphics
         , m_SubUpdate(false)
         {}
 
+        const void*   m_Data;
+        uint32_t      m_DataSize;
         TextureFormat m_Format;
         TextureFilter m_MinFilter;
         TextureFilter m_MagFilter;
         TextureWrap   m_UWrap;
         TextureWrap   m_VWrap;
-        const void*   m_Data;
-        uint32_t      m_DataSize;
 
         // For sub texture updates
         uint32_t m_X;
--- a/engine/graphics/src/graphics.h
+++ b/engine/graphics/src/graphics.h
@@ -60,6 +60,9 @@ namespace dmGraphics
     typedef int (*EngineUpdate)(void* engine);
     typedef void (*EngineGetResult)(void* engine, int* run_action, int* exit_code, int* argc, char*** argv);
 
+    // Decorated asset handle with 32 bits meta | 32 bits opaque handle
+    typedef uint64_t HAssetHandle;
+
     static const HVertexProgram INVALID_VERTEX_PROGRAM_HANDLE = ~0u;
     static const HFragmentProgram INVALID_FRAGMENT_PROGRAM_HANDLE = ~0u;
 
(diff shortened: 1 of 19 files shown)


engine/render/src/render/render_script.cpp
```diff
@@ -1372,29 +1408,63 @@ namespace dmRender
...
+            }
 
+            // Texture can potentially still be zero if it's an attachment texture
             if(texture != 0)
             {
-                if (InsertCommand(i, Command(COMMAND_TYPE_ENABLE_TEXTURE, unit, (uintptr_t)texture)))
+                // JG: We are passing in the opaque handle here and not the asset handle, this is a workaround for 32 bit systems
+                //     because we are using uintptr_t that is defined as a uint32_t on those platforms. We reconstruct the actual handle
+                //     before actual passing in render_command.cpp that processes the command. And alternative would be to use 64b values for the operands?
+                if (InsertCommand(i, Command(COMMAND_TYPE_ENABLE_TEXTURE, unit, (uintptr_t) dmGraphics::GetOpaqueHandle(texture))))
```
Jhonnyg, line 1452: Another approach would be to go back to storing the type along the asset in the handle containers instead of void* as it is currently.. But for at least vulkan and console(s) it requires an extra pointer, which would be nice to avoid. As a future improvement we might want to think about representing the adapter structures in a more generic way somehow. Instead of having completely separate OpenGLTexture/VulkanTexture/NullTexture structs it would be nice to split and merge their representations. That way we can merge a lot of boilerplate code paths (e.g getting width/height from a texture doesn't need three separate functions in that case) as well as reduce divergence between them.. For now I don't think we need anything more complex to solve this particular issue, but It's worth to think about

engine/graphics/src/graphics.h
```diff
@@ -632,7 +634,26 @@ namespace dmGraphics
...
-    bool IsAssetHandleValid(HContext context, HAssetHandle asset_handle);
+
+    // Asset handle helpers
+    const char* GetAssetTypeLiteral(AssetType type);
+    bool        IsAssetHandleValid(HContext context, HAssetHandle asset_handle);
+
+    static inline HAssetHandle MakeAssetHandle(HOpaqueHandle opaque_handle, AssetType asset_type)
+    {
+        assert(asset_type != ASSET_TYPE_NONE && "Invalid asset type");
+        return ((uint64_t) asset_type) << 32 | opaque_handle;
```
JCash, line 645: I kind of expected a chane in the other PR, where we _do_ check that we don't use more than 53 bits?
We want to catch is as early as possible, right?

ALos in `IsAssetHandleValid()`

engine/render/src/render/render_script.cpp
```diff
@@ -1372,29 +1409,63 @@ namespace dmRender
...
+            if (!dmGraphics::IsAssetHandleValid(i->m_RenderContext->m_GraphicsContext, asset_handle))
+            {
+                char buf[128];
+                return luaL_error(L, "Texture handle '%s' is not valid.", AssetHandleToString(asset_handle, buf, sizeof(buf)));
+            }
+
+            if (asset_type == dmGraphics::ASSET_TYPE_RENDER_TARGET)
+            {
+                dmGraphics::BufferType buffer_type = dmGraphics::BUFFER_TYPE_COLOR0_BIT;
+                if (lua_isnumber(L, 3))
```
JCash, line 1429: No need to use double for the type I guess?
Integer should be fine?

engine/render/src/render/render_script.cpp
```diff
@@ -1309,21 +1351,16 @@ namespace dmRender
...
-        }
-        else
+        if (!lua_isnumber(L, 1))
         {
             return luaL_error(L, "Expected render target as the second argument to %s.set_render_target_size.", RENDER_SCRIPT_LIB_NAME);
         }
+
+        dmGraphics::HRenderTarget render_target = (dmGraphics::HRenderTarget) CheckAssetHandle(L, 1, i->m_RenderContext->m_GraphicsContext, dmGraphics::ASSET_TYPE_RENDER_TARGET);
+        uint32_t width = luaL_checknumber(L, 2);
+        uint32_t height = luaL_checknumber(L, 3);
```
JCash, line 1361: Integers?

engine/render/src/render/render_command.cpp
```diff
@@ -76,12 +76,12 @@ namespace dmRender
                 }
                 case COMMAND_TYPE_SET_RENDER_TARGET:
                 {
-                    dmGraphics::SetRenderTarget(context, (dmGraphics::HRenderTarget)c->m_Operands[0], c->m_Operands[1] );
+                    dmGraphics::SetRenderTarget(context, dmGraphics::MakeAssetHandle(c->m_Operands[0], dmGraphics::ASSET_TYPE_RENDER_TARGET), c->m_Operands[1] );
```
JCash, line 79: I would not have thought that we'd create a new handle for each command.
Is this what we want?

Also, for debugging, this handle will be different from the one sent from elsewhere, which makes it harder to debug?

engine/render/src/render/render_script.cpp
```diff
@@ -1199,7 +1234,7 @@ namespace dmRender
             lua_pop(L, 2);
         }
 
-        if (InsertCommand(i, Command(COMMAND_TYPE_SET_RENDER_TARGET, (uintptr_t)render_target, transient_buffer_types)))
+        if (InsertCommand(i, Command(COMMAND_TYPE_SET_RENDER_TARGET, (uintptr_t) dmGraphics::GetOpaqueHandle(render_target), transient_buffer_types)))
```
JCash, line 1237: Why do we need to convert from an asset handle to a raw handle, and then in the backend convert it back to an asset handle?

I would have thought that the asset handle was the handle to be used all the way?

engine/render/src/render/render_script.cpp
```diff
@@ -1372,29 +1409,63 @@ namespace dmRender
...
             }
-            return luaL_error(L, "Render target does not have a texture for the specified buffer type.");
+
+            char buf[128];
+            return luaL_error(L, "Texture handle '%s' is not valid.", AssetHandleToString(asset_handle, buf, sizeof(buf)));
         }
         else
         {
-            return luaL_error(L, "%s.enable_texture(unit, render_target, buffer_type) called with illegal parameters.", RENDER_SCRIPT_LIB_NAME);
+            return luaL_error(L, "%s.enable_texture(unit, render_target|handle, buffer_type) for unit %d called with illegal parameters.", RENDER_SCRIPT_LIB_NAME, unit);
```
JCash, line 1468: I guess the render target should be represented as a handle now?
I.e. should it only say "%s.enable_texture(unit, handle, buffer_type) ..." ?

Or am I missing something?

JCash (changes requested):
Some questions about the handles to/from
Some minor changes

engine/render/src/render/render_script.cpp
Jhonnyg: Sure, this is just moving old code I think but I can fix it 👍

engine/render/src/render/render_script.cpp
Jhonnyg: Aah yes, good catch

engine/render/src/render/render_script.cpp
Jhonnyg: Yes, it's because we use uintptr_t for the operands in the render commands which are defined as uint32 in 32b systems. So the handle will be truncated. I think maybe for this iteration of the feature we could use 64b operand types in the render commands and skip the conversion

engine/render/src/render/render_script.cpp
Jhonnyg: Yes (think this is also old moved code) I'll change it

engine/render/src/render/render_script.cpp
Jhonnyg: Ah, good old C! Yes I think that will be the way to go 👍

engine/render/src/render/render_script.cpp
JCash: Yeah `uint64_t` sounds like the way forward.

PR #7598, opened 2023-04-23
Added profiling scopes around the render dispatch begin/batch/end

JCash:
We want to know if these parts take time as well

### DIFF
--- a/engine/render/src/render/render.cpp
+++ b/engine/render/src/render/render.cpp
@@ -721,12 +721,17 @@ namespace dmRender
         params.m_Operation = RENDER_LIST_OPERATION_BEGIN;
         params.m_Context = context;
 
-        // All get begin operation first
-        for (uint32_t i=0;i!=context->m_RenderListDispatch.Size();i++)
         {
-            const RenderListDispatch& d = context->m_RenderListDispatch[i];
-            params.m_UserData = d.m_UserData;
-            d.m_DispatchFn(params);
+            DM_PROFILE("Dispatch_Begin");
+
+            // All get begin operation first
+            for (uint32_t i=0;i!=context->m_RenderListDispatch.Size();i++)
+            {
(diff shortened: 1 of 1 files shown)


engine/render/src/render/render.cpp
```diff
@@ -721,12 +721,17 @@ namespace dmRender
...
+            DM_PROFILE("Dispatch_Begin");
+
+            // All get begin operation first
+            for (uint32_t i=0;i!=context->m_RenderListDispatch.Size();i++)
+            {
+                const RenderListDispatch& d = context->m_RenderListDispatch[i];
+                params.m_UserData = d.m_UserData;
+                d.m_DispatchFn(params);
+            }
+
```
JCash, line 734: The change looks more than it is.
I've added the scope and indented the code.

engine/render/src/render/render.cpp
```diff
@@ -721,12 +721,17 @@ namespace dmRender
         params.m_Operation = RENDER_LIST_OPERATION_BEGIN;
         params.m_Context = context;
 
-        // All get begin operation first
-        for (uint32_t i=0;i!=context->m_RenderListDispatch.Size();i++)
         {
-            const RenderListDispatch& d = context->m_RenderListDispatch[i];
-            params.m_UserData = d.m_UserData;
-            d.m_DispatchFn(params);
+            DM_PROFILE("Dispatch_Begin");
```
Jhonnyg, line 725: I don't know exactly how the DM_PROFILE generates the scope name, but 'dispatch' is a generic name so maybe we should say what we are dispatching, something like RenderListDispatch?

Jhonnyg (approved):
Looks good, just a small Q

engine/render/src/render/render.cpp
JCash: The name is local, and it's displayed under the parent scope (currently "Render")

PR #7606, opened 2023-04-25
Individual materials for GUI nodes

### LINKED ISSUE #7455: Add option to set material on individual nodes
It is currently not possible to use different materials within a single gui. All box and pie nodes use the same material (set on the .gui component). Labels and particle fx use the material set on the .font or .particlefx. 

It should be possible to add multiple materials to a gui component, just like you can add Fonts, Layers, Layouts etc. And for each box and pie node it should be possible to select one of the materials. If no material is selected the default material on the .gui file is used. There should also be functions to get and set the material of a node (any node):

```
gui.set_material(node, hash("mymaterial")

local material = gui.get_material(node)
```

It should also be possible to swap materials on a gui, just like with textures/atlases:

```
go.property("some_material", resource.material(...))

go.set("#gui", "materials", self.some_material, { key = "mymaterial" })
```

The alternative is to use a big complex uber-shader and then select which shader code to run based on the z-value of a node.

Jhonnyg:
GUIs now has a list of materials that can be assigned to individual nodes, similar to textures, particles and fonts. If no material has been set on a node, the default GUI node material will be used.

Note:  Assigning separate materials will break batching in the same manner as the regular GO world!

Fixes #7455 

## PR checklist

* [ ] Code
* [ ] Documentation


editor/src/clj/editor/gui.clj
```diff
@@ -774,17 +786,27 @@
...
+  (property material g/Str
+            (default "")
+            (dynamic edit-type (g/fnk [material-names] (optional-gui-resource-choicebox material-names)))
+            (dynamic error (g/fnk [_node-id material-names material] (validate-material _node-id material-names material))))
+
   (output visual-base-node-msg g/Any produce-visual-base-node-msg)
   (output gui-scene g/Any (g/fnk [_node-id]
                                  (node->gui-scene _node-id)))
 
+  (output override-material-shader ShaderLifecycle (g/fnk [material-shaders material]
```
Jhonnyg, line 798: We should probably rename this output, the GuiNode already has a :material-shader so I don't think it can be called material-shader..

editor/src/clj/editor/gui.clj
```diff
@@ -1597,6 +1622,55 @@
...
+                         {:name name
+                          :material (resource/resource->proj-path material-resource)}))
+  (output material-shaders GuiResourceShaders :cached (g/fnk [override-material-shader name]
+                                                    ;; If the referenced material-resource is missing, we don't return an entry.
+                                                    (when (some? override-material-shader)
+                                                      {name override-material-shader})))
+  (output material-names GuiResourceNames (g/fnk [name] (sorted-set name)))
+  (output build-errors g/Any (g/fnk [_node-id name name-counts]
+                               (g/package-errors _node-id
+                                                 (prop-unique-id-error _node-id :name name name-counts "Name")))))
```
matgis, line 1672: We should add a call to `(validate-material-resource _node-id material material-shader)` here like we have for the default material specified in the `GuiSceneNode` to make it a build error to use materials with array textures.

editor/src/clj/editor/gui.clj
matgis: It looks like the old `material-shader` inputs and outputs were simply used to pass along the shader of the default material selected for the `GuiSceneNode`. We could get rid of it and instead pass that in as the "no material" entry associated with the empty string key in the `material-shaders` inputs & outputs you've introduced in this PR. That's very similar to how we do it for textures, and you already have code that falls back to the `""` entry in the `material-shaders` map elsewhere in this PR since you based it on how we do it for textures.

editor/src/clj/editor/gui.clj
```diff
@@ -2185,21 +2306,23 @@
 
 (defn- merge-rt-pb-msg [rt-pb-msg template-build-targets]
   (let [merge-fn! (fn [coll msg kw] (reduce conj! coll (map #(do [(:name %) %]) (get msg kw))))
-        [textures fonts particlefx-resources resources]
+        [textures materials fonts particlefx-resources resources]
         (loop [textures (transient {})
+        					  materials (transient {})
```
matgis, line 2311: Whoops! Wonky indentation!

editor/src/clj/editor/gui.clj
```diff
@@ -2388,6 +2513,7 @@
   (input layouts-node g/NodeID) ; for tests
   (input fonts-node g/NodeID) ; for tests
   (input textures-node g/NodeID) ; for tests
+  (input materials-node g/NodeID) ; for tests
```
matgis, line 2516: Well, did you write any tests? 😅

matgis (changes requested):
Big props for tackling the editor changes! The data dependencies in the Gui system are some of the most difficult to grasp in the entire editor code base. 😅

Before merging, we should add more validation for the selected material, and we could rename the new `override-material-shader` output if we included the default material in the `material-names` map under `""` like we do for the textures.

editor/src/clj/editor/gui.clj
Jhonnyg: Fixed

editor/src/clj/editor/gui.clj
```diff
@@ -1597,6 +1622,55 @@
...
+            (value (gu/passthrough material-resource))
+            (set (fn [evaluation-context self old-value new-value]
+                   (project/resource-setter
+                     evaluation-context self old-value new-value
+                     [:resource :material-resource]
+                     [:shader :override-material-shader]
+                     [:build-targets :dep-build-targets])))
+            (dynamic edit-type (g/constantly
+                                 {:type resource/Resource
+                                  :ext ["material"]})))
```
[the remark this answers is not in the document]
Jhonnyg: Fixed!

editor/src/clj/editor/gui.clj
```diff
@@ -1597,6 +1627,58 @@
...
+  (output resource-names GuiResourceNames (g/fnk [name] (sorted-set name)))
+  (input dep-build-targets g/Any)
+  (output dep-build-targets g/Any (gu/passthrough dep-build-targets))
+  (output pb-msg g/Any (g/fnk [name material-resource]
+                         {:name name
+                          :material (resource/resource->proj-path material-resource)}))
+  (output material-shaders GuiResourceShaders :cached (g/fnk [material-shader name]
+                                                    ;; If the referenced material-resource is missing, we don't return an entry.
+                                                    (when (some? material-shader)
+                                                      {name material-shader})))
```
matgis, line 1675: Style: Indentation looks off here?

editor/src/clj/editor/gui.clj
```diff
@@ -2424,6 +2548,16 @@
   (input texture-names GuiResourceNames :array)
   (output texture-names GuiResourceNames :cached (g/fnk [aux-texture-names texture-names] (into (sorted-set) cat (concat aux-texture-names texture-names))))
 
+  (input aux-material-shaders GuiResourceShaders :array)
+  (input material-shaders GuiResourceShaders :array)
+  (output material-shaders GuiResourceShaders :cached (g/fnk [aux-material-shaders material-shaders material-shader]
+                                                        (into {"" material-shader} (concat aux-material-shaders material-shaders))))
```
matgis, line 2554: Hmm, we may need to change this slightly. 🤔

I think with the current code, the default Gui material from the scene referencing a Template might replace the default Gui material inside the referenced scene, or something like that. TBH, I'm not sure what the old behavior was, but we need to check and make sure we still have the same behavior.

After we've established what the correct behavior is, we should be able to alter the outcome simply by changing the order of how values are merged into this map.

Jhonnyg: This is fixed now!

matgis (dismissed):
Editor changes look good! Great work!

engine/gui/src/gui.cpp
```diff
@@ -740,6 +742,28 @@ Result DeleteDynamicTexture(HScene scene, const dmhash_t texture_hash)
...
+        scene->m_Materials.Put(name_hash, (dmRender::HMaterial) material);
+
+        InternalNode* nodes = scene->m_Nodes.Begin();
+        for (uint32_t i = 0; i < scene->m_Nodes.Size(); ++i)
+        {
+            if (nodes[i].m_Node.m_MaterialNameHash == name_hash)
+            {
+                nodes[i].m_Node.m_Material = material;
+            }
+        }
```
JCash, line 762: I think rather than looping for every node, for each material, we could have a loop outside of the function (at setup) where we lookup the materials from the map instead?

JCash (dismissed):
Lgtm.
Some minor detail regarding setting the materials at creation time.

PR #7608, opened 2023-04-26
Add override inspector

### LINKED ISSUE #7127: GUI template editor customization, additional helpers
We share a lot of GUI components with a team also provided with GUI templates to it. There is an idea to adjust template settings in somehow to add additional notes to template in editor and also forbid/hide GUI several node fields to restrict editing

For example, If we have "Token", I want to nobody, who use it, no able to edit any settings exclude "position", "scale" and several other.
I wanna forbid to editing "size mode", "size" and other unrelated properties to this template in all nodes it included (every node can have their forbid property list)

It probably sounds weird, but I think it can help us have more structured and easy to use GUI templates with less bug issues. What do you think?

It is only editor feature

Not sure about concrete solution, but it will be good even if we able to edit in manually in *.gui file. For example add fields:
- note: "text" for additional info for template/node
- forbid_edit: property_name to remove it from editing in template included place

If property is already overrided - it should be shown with "revert" icon and able to edit

Keep it as is :)

vlaaad:
User-facing changes:
This changeset adds an override inspector view that can be requested using the Show Overrides context menu action. This action is available for items in the Outline View and for individual properties in the Properties View. When the action is executed, the editor will search through the uses of this item and show every place that changes the properties of the item. In this override inspector table, you can double-click on a result row to open the place where the properties are overridden.

Fixes #7127

### DIFF
--- a/editor/src/clj/editor/boot_open_project.clj
+++ b/editor/src/clj/editor/boot_open_project.clj
@@ -175,9 +175,8 @@
           scene-visibility     (scene-visibility/make-scene-visibility-node! *view-graph*)
           app-view             (app-view/make-app-view *view-graph* project stage menu-bar editor-tabs-split tool-tabs prefs)
           outline-view         (outline-view/make-outline-view *view-graph* project outline app-view)
-          properties-view      (properties-view/make-properties-view workspace project app-view *view-graph* (.lookup root "#properties"))
           asset-browser        (asset-browser/make-asset-browser *view-graph* workspace assets prefs)
-          open-resource        (partial app-view/open-resource app-view prefs workspace project)
+          open-resource        (partial #'app-view/open-resource app-view prefs workspace project)
           console-view         (console/make-console! *view-graph* workspace console-tab console-grid-pane open-resource)
           build-errors-view    (build-errors-view/make-build-errors-view (.lookup root "#build-errors-tree")
                                                                          (fn [resource selected-node-ids opts]
@@ -186,6 +185,7 @@
           search-results-view  (search-results-view/make-search-results-view! *view-graph*
                                                                               (.lookup root "#search-results-container")
                                                                               open-resource)
+          properties-view      (properties-view/make-properties-view workspace project app-view search-results-view *view-graph* (.lookup root "#properties"))
           changes-view         (changes-view/make-changes-view *view-graph* workspace prefs (.lookup root "#changes-container")
                                                                (fn [changes-view moved-files]
                                                                  (app-view/async-reload! app-view changes-view workspace moved-files)))
--- a/editor/src/clj/editor/scene.clj
+++ b/editor/src/clj/editor/scene.clj
@@ -1411,67 +1411,64 @@
                                (:grid opts) (:grid opts)
                                :else grid/Grid)
         tool-controller-type (get opts :tool-controller scene-tools/ToolController)]
-    (concat
-      (g/make-nodes view-graph
-                    [background      background/Background
-                     selection       [selection/SelectionController :select-fn (fn [selection op-seq]
-                                                                                 (g/transact
-                                                                                   (concat
-                                                                                     (g/operation-sequence op-seq)
-                                                                                     (g/operation-label "Select")
-                                                                                     (select-fn selection))))]
-                     camera          [c/CameraController :local-camera (or (:camera opts) (c/make-camera :orthographic identity {:fov-x 1000 :fov-y 1000}))]
-                     grid            grid-type
-                     tool-controller [tool-controller-type :prefs prefs]
-                     rulers          [rulers/Rulers]]
+    (g/make-nodes view-graph
+                  [background      background/Background
+                   selection       [selection/SelectionController :select-fn (fn [selection op-seq]
+                                                                               (g/transact
+                                                                                 (concat
+                                                                                   (g/operation-sequence op-seq)
+                                                                                   (g/operation-label "Select")
+                                                                                   (select-fn selection))))]
+                   camera          [c/CameraController :local-camera (or (:camera opts) (c/make-camera :orthographic identity {:fov-x 1000 :fov-y 1000}))]
+                   grid            grid-type
+                   tool-controller [tool-controller-type :prefs prefs]
+                   rulers          [rulers/Rulers]]
 
-                    (g/connect resource-node   :scene                         view-id         :scene)
+                  (g/connect resource-node   :scene                         view-id         :scene)
 
-                    (g/connect background      :renderable                    view-id         :aux-renderables)
+                  (g/connect background      :renderable                    view-id         :aux-renderables)
 
-                    (g/connect camera          :camera                        view-id         :camera)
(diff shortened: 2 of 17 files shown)


editor/src/clj/editor/boot_open_project.clj
```diff
@@ -175,9 +175,8 @@
           scene-visibility     (scene-visibility/make-scene-visibility-node! *view-graph*)
           app-view             (app-view/make-app-view *view-graph* project stage menu-bar editor-tabs-split tool-tabs prefs)
           outline-view         (outline-view/make-outline-view *view-graph* project outline app-view)
-          properties-view      (properties-view/make-properties-view workspace project app-view *view-graph* (.lookup root "#properties"))
           asset-browser        (asset-browser/make-asset-browser *view-graph* workspace assets prefs)
-          open-resource        (partial app-view/open-resource app-view prefs workspace project)
+          open-resource        (partial #'app-view/open-resource app-view prefs workspace project)
```
vlaaad, line 179: I used a var instead of a function reference here so it can be redefined at the REPL.

editor/src/clj/editor/scene.clj
```diff
@@ -1411,67 +1411,64 @@
...
+                  (g/connect view-id         :picking-selection             selection       :picking-selection)
+                  (g/connect view-id         :selection                     selection       :selection)
+
+                  (g/connect camera :camera rulers :camera)
+                  (g/connect rulers :renderables view-id :aux-renderables)
+                  (g/connect view-id :viewport rulers :viewport)
+                  (g/connect view-id :cursor-pos rulers :cursor-pos)
+
+                  (when-not (:manual-refresh? opts)
+                    (g/connect view-id :_node-id app-view-id :scene-view-ids)))))
```
vlaaad, line 1471: I removed `(concat ... (when-let [node-id (:select-node opts)] (select-fn [node-id])))`, and the `g/make-nodes` form is left intact. This is because it was the wrong place to perform the selection — this function is called when we construct a new view, but we might want selection on an already existing tab.

editor/styling/stylesheets/modules/_tool-tabs.scss
```diff
@@ -42,10 +42,6 @@
         }
     }
 
-    GridPane {
-        -fx-padding: 10;
-    }
-
```
vlaaad, line 48: I'm not sure why it was here, but this was messing with GridPanes used in the override inspector view. The only other visible difference I found after this change is in the "CSS Test" tool tab which is only used in dev.

editor/src/clj/editor/search_results_view.clj
```diff
@@ -307,3 +328,267 @@
...
+  (when (and (instance? MouseEvent e)
+             (= MouseEvent/MOUSE_PRESSED (.getEventType e)))
+    (let [^MouseEvent e e]
+      (when (even? (.getClickCount e))
+        (.consume e)
+        (when-let [tree (-> e
+                            ^TreeTableView .getSource
+                            .getSelectionModel
+                            ^TreeItem .getSelectedItem
+                            (some-> .getValue))]
```
matgis, line 453: In the past, we've had problems with tagging inside the threading macros - it was not being applied correctly in release builds and resulted in reflection. Maybe consider breaking this out into a separate function to avoid that?

editor/src/clj/editor/search_results_view.clj
```diff
@@ -307,3 +328,267 @@
...
+  (str ": " (string/join " " (:labels property ["X" "Y" "Z"]))))
+(defmethod property-column-suffix types/Vec4 [property]
+  (str ": " (string/join " " (:labels property ["X" "Y" "Z" "W"]))))
+
+(defmulti override-value-cell-view property->edit-type-dispatch-value)
+
+(defn- overridden-style-classes [property]
+  (if (contains? property :original-value)
+    ["overridden"]
+    []))
```
matgis, line 361: Would it work to return `nil` instead of an empty list here? Nil is also an empty collection, right?

editor/src/clj/editor/search_results_view.clj
```diff
@@ -307,3 +328,267 @@
...
+                                   :cell-value-factory (fxui/partial #'property-value property-keyword)
+                                   :cell-factory {:fx/cell-type fx.tree-table-cell/lifecycle
+                                                  :describe (fxui/partial #'value-cell open-resource-fn)}}))
+                           (:display-order state))
+            :root (->tree-item state)})]}]}}))
+
+(defn- make-override-tree [node-id properties {:keys [basis] :as evaluation-context}]
+  (let [property-pred (if (= :all properties)
+                        any?
+                        (set properties))]
```
matgis, line 506: Hey, you managed to find a use for `any?`! Good for you! 😄

editor/src/clj/editor/search_results_view.clj
```diff
@@ -307,3 +328,267 @@
...
+  Args:
+    search-results-view    node id of a search result view
+    node-id                root node id whose overrides are inspected
+    properties             either :all or a coll of property keywords to include
+                           in the override inspector output"
+  [search-results-view node-id properties]
+  (let [evaluation-context (g/make-evaluation-context)
+        parent (g/node-value search-results-view :search-results-container evaluation-context)
+        display-order (into {}
+                            (map-indexed (fn [i k] [k i]))
```
matgis, line 570: We have `util.coll/flipped-pair` for this exact purpose! 😄

matgis (approved):
Works and looks great! It's a wonderful addition, I think. ❤️ 

We should maybe consider including the original node as a parent to the results to make it easier to reference the original property values and navigate to the original scene?

matgis (approved):
I feel we can merge it like this, but I did discover a few annoyances:
* Choosing Show Overrides on a file, the user might expect to see all the overrides on objects inside that file, but we only show overrides on the root-level object. I feel this makes sense if you do it on the root node from the Outline pane, but kind of "leaks implementation details" in the case where you do it from the Assets pane for something like a Game Object or a Gui template. Maybe it would be better to disallow this for anything but `.script` files until we can implement a "show everything" solution?
* When I double-click an override to open it in the Scene View, it seems to open with a random framing, like extremely zoomed in. If I open the corresponding scene from the Assets pane or any other place, the contents of the scene are framed. Could we make it work like that from the Overrides list as well?

vlaaad:
Regarding 1st suggestion, I think a more generic solution would be to show an override option if the file has no child outlines, instead of filtering by file type. 
Regarding framing, it's framing and zooming in on the selected node, maybe if the node has no size the framing looks weird? When opening views for sized box nodes, I usually see the overridden box node. Do you suggest we select but don't frame here at all?

vlaaad:
I worry that outline calculation might be too slow if it's yet uncached.

matgis:
I guess the framing issue stems from script components not having a visual representation in the scene view. Perhaps we shouldn’t attempt framing in this case?

matgis:
We could look if `:overridable-properties` is among the `:tags` of the resource type before allowing Show Overrides on a resource, perhaps? It's currently only on script files.

PR #7616, opened 2023-04-28
Use `emptyStringIsFalse(true)` for mustache compiler

AGulev:
Now we have to create a new variable or replace value with custom logic if we want to use conditional property `{{#something}}` or `{{^something}}`, with this compiler flag empty value will be equal to false.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/BundleHelper.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/BundleHelper.java
@@ -155,7 +155,7 @@ public class BundleHelper {
             return "";
         }
         String s = new String(data);
-        Template template = Mustache.compiler().compile(s);
+        Template template = Mustache.compiler().emptyStringIsFalse(true).compile(s);
         StringWriter sw = new StringWriter();
         template.execute(propertiesMap, properties, sw);
         sw.flush();


britzl (commented):
Will this has some side-effect that we haven't anticipated on existing templates?

AGulev:
@britzl at the moment using conditional templates (`{{#something}}` or {`{^something}}`) on strings - doesn't make any sense because it always will be `true`. There is no reason to have it anywhere if you don't manually patch `bob.jar` like we do here https://github.com/defold/defold/blob/7ad957efd51aacb9746c5e814f59ba854eb67994/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/HTML5Bundler.java#L103-L105C1  (it's not something I want to fix now, it needs a separate cleanup and generalization of data for the templater to avoid such code  I have a few ideas)

But even with this flag, if I use {{something}} it will be an empty string, not `false`, so this change shouldn't break anything that works (if somebody uses string in condition it doesn't work right now anyway)

britzl (approved):

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/BundleHelper.java
```diff
@@ -155,7 +155,7 @@ static public String formatResource(Map<String, Map<String, Object>> propertiesM
             return "";
         }
         String s = new String(data);
-        Template template = Mustache.compiler().compile(s);
+        Template template = Mustache.compiler().emptyStringIsFalse(true).compile(s);
```
JCash, line 158: Please remember to comment on your PRs
What does this do? What was the problem? etc

PR #7625, opened 2023-05-02
Cubemap + WebGL float type fixes

Jhonnyg:
Fixed several issues with creating cubemap textures with float types. Creating cubemaps from scripts now also works correctly for all graphics adapters.

### DIFF
--- a/engine/graphics/src/opengl/graphics_opengl.cpp
+++ b/engine/graphics/src/opengl/graphics_opengl.cpp
@@ -1293,6 +1293,8 @@ static void LogFrameBufferError(GLenum status)
         }
 #endif
 
+        glEnable(GL_TEXTURE_CUBE_MAP_SEAMLESS);
+
         return WINDOW_RESULT_OK;
     }
 
--- a/engine/graphics/src/opengl/graphics_opengl.cpp
+++ b/engine/graphics/src/opengl/graphics_opengl.cpp
@@ -1267,10 +1267,14 @@ static void LogFrameBufferError(GLenum status)
             context->m_IndexBufferFormatSupport |= 1 << INDEXBUFFER_FORMAT_32;
         }
 
+    #if !defined(__EMSCRIPTEN__)
+        // Note: This is enabled automatically for WebGL2, and the defined value is not available **at all** on WebGL1,
+        //       so if we don't want to do crazy workarounds for WebGL let's just ignore it and move on.
         if (context->m_IsGles3Version || OpenGLIsExtensionSupported(context, "GL_ARB_seamless_cube_map"))
         {
             glEnable(GL_TEXTURE_CUBE_MAP_SEAMLESS);
         }
+    #endif
 #else
         context->m_IndexBufferFormatSupport |= 1 << INDEXBUFFER_FORMAT_32;
 
--- a/engine/graphics/src/opengl/graphics_opengl.cpp
+++ b/engine/graphics/src/opengl/graphics_opengl.cpp
@@ -1283,7 +1283,7 @@ static void LogFrameBufferError(GLenum status)
         CLEAR_GL_ERROR;
 #endif
 
-        // if (params->m_PrintDeviceInfo)
+        if (params->m_PrintDeviceInfo)
         {
             OpenGLPrintDeviceInfo(context);
         }
--- a/engine/graphics/src/opengl/graphics_opengl.cpp
+++ b/engine/graphics/src/opengl/graphics_opengl.cpp
@@ -1283,7 +1283,7 @@ static void LogFrameBufferError(GLenum status)
         CLEAR_GL_ERROR;
 #endif
 
-        if (params->m_PrintDeviceInfo)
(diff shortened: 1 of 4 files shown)


engine/graphics/src/opengl/graphics_opengl_defines.h
```diff
@@ -154,63 +154,63 @@
 
 #if defined(GL_RGB32F_EXT)
 #define DMGRAPHICS_TEXTURE_FORMAT_RGB32F                    (GL_RGB32F_EXT)
-#elif defined(GL_RGB32F) && !defined (__EMSCRIPTEN__)
+#elif defined(GL_RGB32F)
```
Jhonnyg, line 157: These are all core now on WebGL2

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2856,84 +2867,45 @@ static void LogFrameBufferError(GLenum status)
...
-                // Ok
-            } else if (bytes_per_row % 2 == 0) {
-                unpackAlignment = 2;
-            } else {
-                unpackAlignment = 1;
+    #ifdef __EMSCRIPTEN__
+        #define WEBGL1_WORKAROUND(fmt) \
+            if (!context->m_IsGles3Version) \
+            { \
+                gl_internal_format = fmt; \
```
Jhonnyg, line 2876: For backwards compatability we need to do a workaround for WebGL1 here to use the "old" formats. We can't rely on compile time constants since they differ between the context versions..

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -176,6 +176,7 @@
     #define glUnmapBufferARB glUnmapBufferOES
     #define GL_ARRAY_BUFFER_ARB GL_ARRAY_BUFFER
     #define GL_ELEMENT_ARRAY_BUFFER_ARB GL_ELEMENT_ARRAY_BUFFER
+    #define GL_TEXTURE_CUBE_MAP_SEAMLESS 0x884F
```
Jhonnyg, line 179: ES3 enum

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -1266,11 +1266,24 @@ static void LogFrameBufferError(GLenum status)
...
+        {
+            glEnable(GL_TEXTURE_CUBE_MAP_SEAMLESS);
+            CLEAR_GL_ERROR;
+        }
+    #endif
 #else
         context->m_IndexBufferFormatSupport |= 1 << INDEXBUFFER_FORMAT_32;
+
+        glEnable(GL_TEXTURE_CUBE_MAP_SEAMLESS);
+        CLEAR_GL_ERROR;
```
Jhonnyg, line 1283: We can't guarantee that enabling seamless should work, for "most" platforms it's on by default if we use a GL3+ context so for debug builds to work if this flag DOESNT work we need to clear the errors. It's fine if we don't get seamless sampling for the platforms that don't support it since it's at least not worse than it is now where we don't have use this flag at all..

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2953,52 +2930,119 @@ static void LogFrameBufferError(GLenum status)
...
+            dmLogError("Texture format %s is not a valid format.", GetTextureFormatLiteral(format));
             break;
         }
 
+    #undef ES2_ENUM_WORKAROUND
+    #undef EMSCRIPTEN_ES2_BACKWARDS_COMPAT
+    #undef ANDROID_ES2_BACKWARDS_COMPAT
+    }
+
+    static void OpenGLSetTexture(HTexture texture, const TextureParams& params)
```
Jhonnyg, line 2991: This looks like a "big" change, but it's just that I pulled out the type / format switch into a separate function

engine/graphics/src/opengl/graphics_opengl_defines.h
Jhonnyg: And for WebGL1 we have the fallback to the "correct" float types for that context.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -903,7 +904,7 @@ namespace dmGraphics
 
             selected_device = device;
             selected_queue_family = queue_family;
-            break;
+            break; // Why do we break here? x_x
```
JCash, line 907: I guess because we're happy with the device we've found?
Are we not?

JCash (dismissed):

engine/graphics/src/vulkan/graphics_vulkan.cpp
Jhonnyg: Ah, lol I'm a dummy. Yes we are, I thought we didn't loop through all devices but I was probably ~~drunk~~ tired when I read the code

PR #7642, opened 2023-05-10
Use `:textual?` flag for Search in Files

### LINKED ISSUE #7586: IDE: Exclude binary files from the search.
We have more and more files in the project (3400 files, 180 megabytes) and the Ctrl/Cmd+F project search function is getting slower and slower. Why? The IDE seems to search through all files, even binary ones.

For example, I entered "IHDR" (PNG header) and the IDE found this text in GLBs:

image

Improve the detection of binary files or exclude ".glb" from the search.

vlaaad:
Technical note: we now look at `:textual?` flag of a resource type to determine if a resource without save data might be textual.

User-facing changes:

We no longer search through `.glb` files when performing Search in Files.

Fixes #7586

### DIFF
--- a/editor/src/clj/editor/defold_project_search.clj
+++ b/editor/src/clj/editor/defold_project_search.clj
@@ -51,7 +51,9 @@
   [{:keys [content resource] :as save-data}]
   (or (when (some? content)
         (make-line-coll #(BufferedReader. (StringReader. content))))
-      (when (and (resource/exists? resource) (not (text-util/binary? resource)))
+      (when (and (resource/exists? resource)
+                 (resource/textual? resource)
+                 (not (text-util/binary? resource)))
         (make-line-coll #(io/reader resource)))))
 
 (defn compile-find-in-files-regex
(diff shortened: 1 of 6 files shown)


editor/src/clj/editor/defold_project_search.clj
```diff
@@ -51,7 +51,9 @@
   [{:keys [content resource] :as save-data}]
   (or (when (some? content)
         (make-line-coll #(BufferedReader. (StringReader. content))))
-      (when (and (resource/exists? resource) (not (text-util/binary? resource)))
+      (when (and (resource/exists? resource)
+                 (resource/textual? resource)
+                 (not (text-util/binary? resource)))
```
matgis, line 56: If `(:textual? resource-type)` is `true`, we don't need to check if `(text-util/binary? resource)`, right? If I understand it correctly, we only need to do that if `resource-type` is `nil`?

matgis (changes requested):
I don't think we need to call `text-util/binary?` for known resource types?

editor/src/clj/editor/defold_project_search.clj
vlaaad: Hmm, if there is no resource type, `resource/textual?` will return true, but I see what you mean, we shouldn't check the contents of resources with specified types...

matgis (approved):
Looks good! 👍

PR #7670, opened 2023-05-22
Fix issue with components counter in some cases

### LINKED ISSUE #7668: Issue with components counter
In some cases component counter doesn't take something into account which may crash the game.

**To Reproduce (REQUIRED)**
Bundle using bob.jar this repo and run it
https://github.com/AGulev/woodpecker-learns-accents

Should work fine

1.4.5

**Platforms (REQUIRED):**
 - Platforms: all
 - OS: all
 - Device: all

https://github.com/AGulev/woodpecker-learns-accents

**Workaround (OPTIONAL):**
add factory with go and needed component

AGulev:
Fix the issue that occurs when there are multiple instances of the same object containing a factory, which causes the component counter to break.

Fix https://github.com/defold/defold/issues/7668

## PR checklist

* [x] Code
* [ ] Documentation

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/ComponentsCounter.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/ComponentsCounter.java
@@ -82,7 +82,7 @@ public class ComponentsCounter {
                 componentName = componentName.substring(1);
             }
             Integer currentValue = components.getOrDefault(componentName, 0);
-            if (count < 0) {
+            if (count == DYNAMIC_VALUE) {
                 components.put(componentName, DYNAMIC_VALUE);
             } 
             else if (currentValue != DYNAMIC_VALUE) {
@@ -104,7 +104,12 @@ public class ComponentsCounter {
         public void add(Storage storage, Integer count) {
             Map<String, Integer> comps = storage.get();
             for (Map.Entry<String,Integer> entry : comps.entrySet()) {
-                add(entry.getKey(), count == DYNAMIC_VALUE ? DYNAMIC_VALUE : entry.getValue() * count);
+                Integer value = entry.getValue() * count;
+                // We can't multiply or sum DYNAMIC_VALUE, just set it
+                if (count == DYNAMIC_VALUE || entry.getValue() == DYNAMIC_VALUE) {
+                    value = DYNAMIC_VALUE;
+                }
+                add(entry.getKey(), value);
             }
         }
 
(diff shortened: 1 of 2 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/ComponentsCounter.java
```diff
@@ -82,7 +82,7 @@ public void add(String componentName, Integer count) {
                 componentName = componentName.substring(1);
             }
             Integer currentValue = components.getOrDefault(componentName, 0);
-            if (count == DYNAMIC_VALUE) {
+            if (count < 0) {
```
JCash, line 85: Please explain the issue and the reason why this fix works.

E.g. why the DYNAMIC_VALUE doesn't work as intended.
Does then the other uses of DYNAMIC_VALUE need updating?

AGulev: At lines 93, 97, and 104, there are different `add()` functions being used. One of them specifically handles the storage and count of objects (line 104).

This particular version of the `add()` function is called when the collection contains a few instances of an object, and the object has a factory associated with it. The factory's prototype is "GO with Sprite," and the storage will have the following structure:

Storage:
{
  sprite: DYNAMIC_VALUE (which is -1)
}

Now, if we call the `add()` function with the storage and the value 2:

>add(ourStorage, 2)

The result will be -2, which is different from the expected DYNAMIC_VALUE. Instead of setting the value to DYNAMIC_VALUE, the code currently subtracts -2 from the count of components. This behavior is the reason of the issue.

JCash: Does any of this need to be commented in code?

Also, if the value is -2, and we suddenly need to add +1, will it then still be -1?

AGulev: The value can be only -1 and greater 
we summarize positive values and if it's negative set DYNAMIC_VALUE
I'll add a comment into code (and maybe change it a bit to make logic a bit more obvious)

JCash: But "The result will be -2, which is different from the expected DYNAMIC_VALUE"
seem to contradict the code and your statements "The value can be only -1 and greater" and "DYNAMIC_VALUE (which is -1)"

Then the code should have worked?

AGulev: =) nope
The value in any storage can me -1 and greater.
But in this case there are a few instances of the same object, so we use function that do `storage * instance_count`
as result we have a value `<-1`

JCash (dismissed):

JCash (approved):

Issue #5392, opened 2020-12-09
Automatically install Android and iOS builds on devices


Lerg:
**Is your feature request related to a problem? Please describe (REQUIRED):**
Other engines provide an option to automatically install the built APK or APP files on Android or iOS devices.
Such option would speed up development and make it easier overall.

**Describe the solution you'd like (REQUIRED):**
Add a checkbox such as "Install and run on device" when building for iOS or Android. A default connected device is fine, but a dropdown with the list of devices is welcome too.

**Describe alternatives you've considered (REQUIRED):**
There is no alternative.

**Additional context (OPTIONAL):**
Unity does it for Android. Solar2D does it for both Android and iOS.

JCash:
I think this would be a good candidate for a "post bundle" event, that can invoke an editor script.

britzl:
Pre-step: Are we allowed to redistribute the tools in bob?
Yes: Bundle tools and use them to install and launch game on device

No - Step 1: Detect Android SDK and Xcode command line tools and use them to install and launch game
No - Step 2: Trigger installer if tools are not installed

vlaaad:
XCode command line tools are 700Mb. Solar2D [seems to be using](https://github.com/coronalabs/corona/blob/0e475c94440f3e2ce876df945dba0d16c6d3f9d6/platform/resources/ios_sendapp.sh#L128) the [ideviceinstaller](https://github.com/libimobiledevice/ideviceinstaller) executable for macOS installations though. Not sure if we can use it because the library uses GPL. It seems that Solar2D does not distribute it, but only checks if it exists and [provides install instructions](https://github.com/coronalabs/corona/blob/0e475c94440f3e2ce876df945dba0d16c6d3f9d6/platform/mac/TVOSAppBuildController.mm#LL1049C7-L1049C36).

adb is 13Mb though, and seems to be apache-licensed

JCash:
> It seems that Solar2D does not distribute it, but only checks if it exists and [provides install instructions](https://github.com/coronalabs/corona/blob/0e475c94440f3e2ce876df945dba0d16c6d3f9d6/platform/mac/TVOSAppBuildController.mm#LL1049C7-L1049C36).

This sounds like something a post-bundle editor script could do :)

vlaaad:
The plan:
- add new preferences:
  - Android installer path
  - Android installer args (should support substitutions like Code -> Open File preference)
  - iOS installer path
  - iOS installer args (should support substitutions like Code -> Open File preference)
[...]

PR #7673, opened 2023-05-22
Install and launch APKs after the Android bundle

vlaaad:
Technical notes: I refactored the `editor.process` ns because I didn't like the API it had. I based the API on a new [Clojure process ns](https://github.com/clojure/clojure/blob/master/src/clj/clojure/java/process.clj) that we might want to migrate to when it matures (I [reported some issues to them](https://ask.clojure.org/index.php/12942/feedback-on-the-new-clojure-java-process-namespace) when I was working on it).

User-facing changes:
If you have [ADB](https://developer.android.com/tools/adb) installed on your computer, you can now install the built APK on the connected device from the editor. The Android bundle dialog now has checkboxes for performing installation and/or launch after the bundle is completed. The preferences window now has a Tools pane where you can configure the path to ADB installation if the editor couldn't find it. The editor shows the installation process output in its console.

Related to #5392

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -19,10 +19,11 @@
             [cljfx.fx.text-flow :as fx.text-flow]
             [cljfx.fx.tooltip :as fx.tooltip]
             [cljfx.fx.v-box :as fx.v-box]
-            [clojure.java.io :as io]
             [clojure.edn :as edn]
+            [clojure.java.io :as io]
             [clojure.string :as string]
             [dynamo.graph :as g]
+            [editor.adb :as adb]
             [editor.build :as build]
             [editor.build-errors-view :as build-errors-view]
             [editor.bundle-dialog :as bundle-dialog]
@@ -78,17 +79,18 @@
             [internal.graph.types :as gt]
             [internal.util :refer [first-where]]
             [service.log :as log]
+            [service.smoke-log :as slog]
             [util.http-server :as http-server]
             [util.profiler :as profiler]
-            [util.thread-util :as thread-util]
-            [service.smoke-log :as slog])
+            [util.thread-util :as thread-util])
   (:import [clojure.lang ExceptionInfo]
            [com.defold.editor Editor]
            [com.defold.editor UIUtil]
            [com.dynamo.bob Platform]
            [com.sun.javafx.scene NodeHelper]
-           [java.io BufferedReader File IOException]
+           [java.io BufferedReader File IOException OutputStream PipedInputStream PipedOutputStream PrintWriter]
            [java.net URL]
+           [java.nio.charset StandardCharsets]
            [java.util Collection List]
            [javafx.beans.value ChangeListener]
            [javafx.collections ListChangeListener ObservableList]
@@ -728,7 +730,7 @@
     (future
       (error-reporting/catch-all!
         (extensions/execute-hook! project :on-target-launched hook-options)
-        (process/watchdog! process #(extensions/execute-hook! project :on-target-terminated hook-options))))))
+        (process/on-exit process #(extensions/execute-hook! project :on-target-terminated hook-options))))))
 
 (defn- target-cannot-swap-engine? [target]
   (and (some? target)
@@ -1207,6 +1209,12 @@ If you do not specifically require different script states, consider changing th
   (reset-console-stream! input-stream)
   (reset-remote-log-pump-thread! (start-log-pump! input-stream (make-remote-log-sink input-stream))))
 
+(defn- start-new-log-pipe!
+  ^PipedOutputStream []
+  (let [in (PipedInputStream.)]
+    (pipe-log-stream-to-console! in)
+    (PipedOutputStream. in)))
+
 (handler/defhandler :build-html5 :global
   (run [project prefs web-server build-errors-view changes-view main-stage tool-tab-pane]
     (let [main-scene (.getScene ^Stage main-stage)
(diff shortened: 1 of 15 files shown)


editor/src/clj/editor/bundle_dialog.clj
```diff
@@ -460,44 +525,45 @@
...
 
-                  (and (some? key-pass) (nil? keystore))
-                  [:fatal "Keystore must be set if key password is specified."])
+               (and (some? key-pass) (nil? keystore))
+               [:fatal "Keystore must be set if key password is specified."])
    :architecture (when-not (or architecture-32bit? architecture-64bit?)
                    [:fatal "At least one architecture must be selected."])
    :bundle-format (when-not bundle-format
-                   [:fatal "No bundle format selected."])})
+                    [:fatal "No bundle format selected."])})
```
vlaaad, line 557: Only indentation changes here.

editor/src/clj/editor/process.clj
```diff
@@ -13,31 +13,157 @@
...
+    (loop []
+      (let [size (.read in buf)]
+        (when-not (neg? size)
+          (do (.write out buf 0 size)
+              (.flush out)
+              (recur)))))))
+
+(defn await-exit-code
+  "Blocks until process exits and returns the exit code"
+  [^Process process]
```
matgis, line 118: Do we want to tag the return value as `^long` here?

editor/src/clj/editor/process.clj
```diff
@@ -13,31 +13,157 @@
...
+  (.waitFor process))
+
+(defn exit-ok?
+  "Checks if the exit code of the process indicates success"
+  [exit-code]
+  {:pre [(int? exit-code)]}
+  (zero? exit-code))
```
matgis, line 125: I don't think we need this function. It sort of feels like it should take a `Process` which is confusing. We can use `zero?` with `await-exit-code` directly whenever this would have been used.

editor/src/clj/editor/process.clj
```diff
@@ -13,31 +13,157 @@
...
-      (.directory pb directory))
+(defn- ->redirect
+  ^ProcessBuilder$Redirect [x]
+  (case x
+    :pipe ProcessBuilder$Redirect/PIPE
+    :inherit ProcessBuilder$Redirect/INHERIT
+    :discard ProcessBuilder$Redirect/DISCARD
+    x))
+
+(defn ^{:arglists '([& command+args] [opts & command+args])} start
```
matgis, line 32: Don't we want most of the functions in this file to have the`!` suffix? I'd think calling these from a `swap!` function would be bad?

editor/test/editor/process_test.clj
```diff
@@ -0,0 +1,40 @@
+(ns editor.process-test
+  (:require [clojure.test :refer :all]
+            [editor.process :as process])
+  (:import [java.io ByteArrayOutputStream]))
+
+(deftest start-test
+  (let [p (process/start "ls")
```
matgis, line 7: Does Windows have the `ls` command now? They used to have `dir` instead?

matgis (changes requested):
Some questions and a few minor concerns.

editor/src/clj/editor/app_view.clj
```diff
@@ -2290,6 +2300,32 @@ If you do not specifically require different script states, consider changing th
           show-search-results-tab! (partial show-search-results! main-scene tool-tab-pane)]
       (search-results-view/show-search-in-files-dialog! search-results-view project prefs show-search-results-tab!))))
 
+(defn- adb-post-bundle! [prefs output-directory project ^OutputStream out launch]
+  (g/with-auto-evaluation-context evaluation-context
+    (let [game-project (project/get-resource-node project "/game.project" evaluation-context)
+          package (game-project/get-setting game-project ["android" "package"] evaluation-context)
+          project-title (game-project/get-setting game-project ["project" "title"] evaluation-context)
+          apk-path (io/file output-directory project-title (str project-title ".apk"))]
```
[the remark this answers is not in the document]
vlaaad: If the package is empty, the build fails with "Android package name '' is not valid." error. If the project title is empty, the build does not fail, but the APK is not built either! I created an issue for this: https://github.com/defold/defold/issues/7677

editor/src/clj/editor/app_view.clj
```diff
@@ -2290,6 +2300,32 @@ If you do not specifically require different script states, consider changing th
...
+      ;; Do the actual work asynchronously because adb commands are blocking.
+      (future
+        ;; We should close the out when we are done here
+        (with-open [writer (PrintWriter. out true StandardCharsets/UTF_8)]
```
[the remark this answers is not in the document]
vlaaad: I don't think it can throw, but I'll add error reporting just in case

editor/src/clj/editor/app_view.clj
```diff
@@ -1207,6 +1209,12 @@ If you do not specifically require different script states, consider changing th
   (reset-console-stream! input-stream)
   (reset-remote-log-pump-thread! (start-log-pump! input-stream (make-remote-log-sink input-stream))))
 
+(defn- start-new-log-pipe!
+  ^PipedOutputStream []
+  (let [in (PipedInputStream.)]
+    (pipe-log-stream-to-console! in)
```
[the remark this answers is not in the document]
vlaaad: We don't need to undo this. While we should try to close the output stream when we are done with it, there is no harm in leaving it open, the worst thing that can happen is we block a log pump thread until a new input stream is configured to be shown in the console (e.g. when we build and run), and at that point, the blocked log pump thread will stop and be garbage collected.

editor/src/clj/editor/process.clj
vlaaad: Yeah, that was unnecessary.

editor/test/editor/process_test.clj
vlaaad: Hmm no, the tests will fail on windows. I'll make sure these tests are skipped on windows.

matgis (changes requested):
I just noticed that `adb_test.clj` and `process_test.clj` are missing the copyright header. Other than that it looks good to me!

matgis (approved):
LGTM!

PR #7690, opened 2023-05-29
Fixed an editor crash when loading a non-editable collection or game object with an embedded spine component referring to an out-of-date spinejson file

### LINKED ISSUE #7689: Editor crashes to desktop when loading a non-editable collection or game object with an embedded spine component referring to an out-of-date spinejson file
For non-editable collections and game objects, the `:sanitize-fn` decodes embedded strings into map representations of the embedded game objects and components. This process can fail if a plugin such as `extension-spine` throws an exception as it reads the embedded string. In this case, the `:sanitize-fn` gives up and returns the data in its non-sanitized form. Normally this would cause the `:load-fn` to throw an exception, but in the case of non-editable resources, the read protobuf map is simply assigned to a property. This eventually trips an assert at an inopportune time, which in our case terminates the editor process in the middle of loading.

matgis:
Fixes #7689

### User-facing changes
* Fixed an issue where the editor would crash when loading a project containing a non-editable collection or game object with an embedded Spine component referring to an out-of-date `.spinejson` file.

### Technical details
For non-editable collections and game objects, the `:sanitize-fn` decodes embedded strings into map representations of the embedded game objects and components. This process can fail if a plugin such as `extension-spine` throws an exception as it reads the embedded string. In this case, the `:sanitize-fn` gives up and returns the data in its non-sanitized form. Normally this would cause the `:load-fn` to throw an exception, but in the case of non-editable resources, the read protobuf map is simply assigned to a property. This eventually trips an assert at an inopportune time, which in our case terminates the editor process in the middle of loading.

This PR ensures that the read data has been properly decoded into map format before assigning the property in the `:load-fn` so that an exception is thrown in the right place for the resource to be flagged as defective.

### DIFF
--- a/editor/src/clj/editor/collection_non_editable.clj
+++ b/editor/src/clj/editor/collection_non_editable.clj
@@ -15,6 +15,7 @@
 (ns editor.collection-non-editable
   (:require [dynamo.graph :as g]
             [editor.build-target :as bt]
+            [editor.collection-string-data :as collection-string-data]
             [editor.collection-common :as collection-common]
             [editor.defold-project :as project]
             [editor.game-object-common :as game-object-common]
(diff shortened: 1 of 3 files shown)


vlaaad (approved):
Lgtm! It would be nice if your useful and descriptive comment in the PR description ended up in the commit message when it's merged into editor-dev, so that the intention of the change is preserved in history! Not sure if that happens when you squash.. it should?

matgis:
It didn't. But I put the details in the issue description. The comments in the code explain why we are doing this, though. At least that's something.

PR #7692, opened 2023-05-29
Implement Find References

### LINKED ISSUE #7691: Add Find References code navigation tool
Find References is an essential code navigation tool along with Goto Definition. We should provide it.

Implement Find References using LSP.

Create a bespoke implementation of Find References.

vlaaad:
UX note: I thought that showing a notification with a button to read more about LSPs is better than showing a greyed-out menu item with no clarification, so I decided to always make the LSP actions active.

User-facing changes: The editor now supports finding references using LSP, with the command assigned to a <kbd>Shift F12</kbd> shortcut. The Go to Definition and Find References commands are more discoverable now — we show them in the Edit menu bar. We now also point the user toward the LSP post on the forum in case they try to use these commands without a language server dependency.

Fixes #7691


matgis (approved):
Looks good to me!

PR #7728, opened 2023-06-07
Migrate to our managed sass dependency

### LINKED ISSUE #7727: Make editor manage its sass dependency
**Task (REQUIRED):**
To support the arm arch, we want to move the sass dependency to the editor. Eventually, we will update it to pick an arm version of sass.

The editor manages its sass dependency without relying on the engine.

vlaaad:
The [sass](https://github.com/sass/sassc) tool used by [lein-sass](https://github.com/stephanos/lein-sass) is deprecated in favor of [dart-sass](https://github.com/sass/dart-sass). [The latest release of Dart Sass](https://github.com/sass/dart-sass/releases/tag/1.62.1) has an arm build for macOS, so it will be possible to use it with an arm version of the Defold editor pipeline. This changeset replaces `lein-sass` plugin with our own that exposes the same API (`once` and `auto`), but uses Dart Sass underneath.

Dart Sass has a couple of minor differences from our old Sass that required changes to our scss files. One gotcha I found is that it omits CSS directives that have `null` values, so we have to use `string.unqoute("null")` instead for removing defaults (e.g. `-fx-shape: null` -> `-fx-shape: string.unquote("null")`).

With white space and empty lines ignored in the Idea diff viewer, I didn't notice any significant changes to the output CSS files. The output has slightly different formatting, and it removes trailing `.0`s from numbers, which I think should not result in a different styling.

Fixes #7727


editor/styling/stylesheets/_editor.scss
```diff
@@ -67,7 +67,7 @@ Links:
   -fx-text-color-base: -df-defold-blue-lighter;
   -fx-text-fill: -df-defold-blue-lighter; }
 
-&.override-inspector-check-box {
+.override-inspector-check-box {
```
vlaaad, line 70: There was a top-level `&`, which is wrong since `&` refers to a parent, and on the top level no parent exists. Dart Sass reports it as error.

editor/styling/stylesheets/_editor.scss
```diff
@@ -145,19 +145,7 @@ Links:
...
-                L  7  5
-                L 10  8
-                L  8 10
-                L  5  7
-                L  2 10
-                L  0  8
-                L  3  5
-                L  0  2
-                Z";
+    -fx-shape: "M 2 0 L  5  3 L  8  0 L 10  2 L  7  5 L 10  8 L  8 10 L  5  7 L  2 10 L  0  8 L  3  5 L  0  2 Z";
```
vlaaad, line 148: It was nicer before, but Dart Sass doesn't like it :/

editor/styling/stylesheets/dialogs.sass
```diff
@@ -1,3 +1,3 @@
-@import 'base/_typography';
-@import 'base/_palette';
-@import '_dialogs';
+@import 'base/_typography'
+@import 'base/_palette'
+@import '_dialogs'
```
vlaaad, line 3: Dart Sass doesn't like semicolons after imports.

editor/styling/stylesheets/_editor.scss
matgis: Maybe we should remove the formatting whitespace (i.e. the double-spaces) altogether?

matgis (approved):
Looks good!

editor/tasks/leiningen/sass.clj
```diff
@@ -0,0 +1,86 @@
...
+              platform
+              ext)]
+    (extract (http-cache/download url) packages-path ext)
+    (str (io/file packages-path "dart-sass" "sass"))))
+
+(defn sass [project & [mode]]
+  {:pre [(#{"once" "auto"} mode)]}
+  (let [{:keys [target-path sass]} project
+        {:keys [src output-directory]} sass
+        packages-path (io/file target-path "packages")
```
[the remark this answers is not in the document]
vlaaad: I'm not sure we would want that. Looks like unnecessary coupling — why put the editor's dependency into engine packages?

PR #7765, opened 2023-06-27
Check node existence before querying it on load

### LINKED ISSUE #7174: Crash on opening project with broken GUI template

**To Reproduce (REQUIRED)**
1. create empty project
2. add 2 gui files: main.gui and part.gui
3. add part.gui as template into main.gui
4. close defold, open main.gui in text editor, edit line `template: "/main/part.gui"` to be `template: "/main/part1.gui"`
5. save changes, attempt to open project with defold, it will crash on start

Defold starts, somehow indicating that there is issue in gui file (for example like when gui node has wrong texture)

  - Version 1.3.7

**Platforms (REQUIRED):**
 Windows 10

vlaaad:
Technical notes:
The problem is that `editor.defold-project/connect-resource-node` might return a node id for a node that does not yet exist in the system in case where it has to create a new node. We cannot query such nodes using `g/node-value` because they don't exist yet. Users of `connect-resource-node` first need to check that the node already exists if they want to use `g/node-value` on it.

User-facing changes:
Projects that previously couldn't be opened due to broken resource references in files now can be opened.

Fixes #7174

### DIFF
--- a/editor/src/clj/editor/collection.clj
+++ b/editor/src/clj/editor/collection.clj
@@ -335,7 +335,9 @@
                              project (project/get-project basis self)]
                          (if (some-> new-resource resource/editable?)
                            ;; This is an editable source resource. Create an override node and make connections to enable full editing.
-                           (let [{connect-tx-data :tx-data go-node :node-id} (project/connect-resource-node evaluation-context project new-resource self [])]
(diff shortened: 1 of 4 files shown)


matgis (approved):
Looks good to me!

PR #7769, opened 2023-06-29
Remove :injections from editor tests

matgis:
### Technical changes
* Remove unneeded `:injections` section from `:test` profile.
* Remove unsupported `--illegal-access=warn` JVM parameter.


vlaaad (dismissed):
Ship it!

PR #7787, opened 2023-07-11
Add conditional breakpoints

### LINKED ISSUE #7274: Conditional breakpoints in the debugger
Add support for conditional breakpoints which are only hit when the provided condition (Lua code) evaluates to true. Example from VSCode:

https://code.visualstudio.com/docs/editor/debugging#_conditional-breakpoints

vlaaad:
User-facing changes:
You can set a Lua condition that must evaluate to true for the breakpoint to be triggered. You can do it using new breakpoint popup that you can open by right-clicking on the breakpoint code editor area or pressing <kbd>Alt F9</kbd>.

Technical changes:
Conditional breakpoints are implemented by expanding the `SETB` mobdebug command from `SETB file line` to `SETB file line condition?`. If the condition fails to compile, the breakpoint will trigger every time. If the condition throws an error during its check, the breakpoint will not trigger though.

Fixes #7274


editor/src/clj/editor/code/view.clj
```diff
@@ -2657,7 +2823,7 @@
                                              :font-size (.getValue font-size-property)
                                              :font-name  (.getValue font-name-property)
                                              :grammar grammar
-                                             :gutter-view (CodeEditorGutterView.)
+                                             :gutter-view (->CodeEditorGutterView)
```
vlaaad, line 2826: Using `(->CodeEditorGutterView)` instead of `(CodeEditorGutterView.)` allows us to skip re-evaluating the `make-view!` defn after we redefine `CodeEditorGutterView` class.

editor/src/clj/editor/debug_view.clj
```diff
@@ -414,8 +414,8 @@
      (when (or (seq added) (seq removed))
        (mobdebug/with-suspended-session debug-session
          (fn [debug-session]
-           (run! #(set-breakpoint! debug-session %) added)
-           (run! #(remove-breakpoint! debug-session %) removed)))))))
+           (run! #(remove-breakpoint! debug-session %) removed)
+           (run! #(set-breakpoint! debug-session %) added)))))))
```
vlaaad, line 418: Need to remove before adding e.g. when we modify a breakpoint condition: we want to remove-then-set updates (as expected), but set-then-remove removes (unexpected)

AGulev (dismissed):
Lua part looks good - shiiiiip it!

editor/src/clj/editor/code/view.clj
```diff
@@ -2485,11 +2505,34 @@
...
             (let [y (data/row->y layout source-line-index)]
-              (when (contains? breakpoint-rows source-line-index)
+              (when (contains? breakpoint-row->condition source-line-index)
                 (.setFill gc gutter-breakpoint-color)
                 (.fillOval gc
-                           (+ (.x line-numbers-rect) (.w line-numbers-rect) 3.0)
-                           (+ y 3.0) indicator-diameter indicator-diameter))
+                           (+ (.x line-numbers-rect) (.w line-numbers-rect) indicator-offset)
+                           (+ y indicator-offset) indicator-diameter indicator-diameter)
+                (when (breakpoint-row->condition source-line-index)
```
matgis, line 2513: This lookup could maybe be merged with the `contains?` check above so we don't traverse the map twice?

editor/src/clj/editor/code/view.clj
```diff
@@ -2641,6 +2684,129 @@
...
+           :children
+           [{:fx/type fx.region/lifecycle
+             :style-class "breakpoint-editor-background"}
+            {:fx/type fx.v-box/lifecycle
+             :style-class "breakpoint-editor-content"
+             :spacing padding
+             :children
+             [{:fx/type fx.label/lifecycle
+               :style-class ["label" "breakpoint-editor-label" "breakpoint-editor-header"]
+               :text (format "Breakpoint on line %d" (inc ^long (data/breakpoint-row edited-breakpoint)))}
```
matgis, line 2739: `data/breakpoint-row` just `asserts` that the region is of the expected type before returning the `row`, but since we know this will be a breakpoint region we can use `data/CursorRange->line-number` instead.

```suggestion
               :text (format "Breakpoint on line %d" (data/CursorRange->line-number edited-breakpoint))}
```

editor/src/clj/editor/code/view.clj
```diff
@@ -2641,6 +2684,129 @@
...
+(defn- create-breakpoint-editor! [view-node canvas ^Tab tab]
+  (let [state (atom nil)
+        timer (ui/->timer
+                10
+                "breakpoint-code-editor-timer"
+                (fn [_ _ _]
+                  (when (and (.isSelected tab) (not (ui/ui-disabled?)))
+                    (g/with-auto-evaluation-context evaluation-context
+                      (reset! state
+                        (if-let [edited-breakpoint (g/node-value view-node :edited-breakpoint evaluation-context)]
```
matgis, line 2772: We could use `when-some` instead of `if-let` here, since we're returning `nil` when false.

editor/src/clj/editor/keymap.clj
```diff
@@ -36,6 +36,7 @@
            ["Alt+Down" :end-of-line]
            ["Alt+Down" :move-down]
            ["Alt+Left" :prev-word]
+           ["Alt+F9" :edit-breakpoint]
```
matgis, line 39: I think we should add a menu item for this, since we already have one for Toggle Breakpoint. This would help users discover the added keyboard shortcut.

matgis (dismissed):
Looks good to me! We could maybe consider adding a menu item for Edit Breakpoint?

editor/src/clj/editor/code/view.clj
vlaaad: Hmm yes, I will use `when-let` instead of `if-let`. 

Btw, I'm usually hesitant to use `when-some` because I assume it will do more work than `when-let`. `(when-let [x ...] ...)` expands to `(if x ...)`, which checks if x is not nil or false in `if`, but `(when-some [x ...] ...)` expands to `(if (nil? x) nil ...)`, which will first check if `x` is not nil in `nil?`, and then it will check if a boolean returned by that check is not nil or false in `if`. 

Because of that assumption, I default to `if-let` and `when-let`. Maybe I'm thinking too much into this?

matgis: That's true. And while I don't think it would make much of a difference most of the time, it is something to keep in mind. I tend to prefer `if-some` and `when-some` because I think it is somewhat clearer, but maybe I should start using `if-let` and `when-let` unless the bound values is a boolean. 🤔

PR #7788, opened 2023-07-11
Make sure that unloading of a gui component removes its dynamic textures

### LINKED ISSUE #7743: Resource profiler: add information about resources created from gui.new_texture()
Our game uses a lot of textures created by `gui.new_texture()` and we need to know an impact on memory usage

I guess the good place to see information about is the Resource Profiler (with #7742)

As alternative we can create a wrapper code and do the calculations by it

AGulev:
Make sure that unloading a GUI component removes any dynamic textures associated with it to free up memory and prevent memory leaks.

Added `GuiDynamicTexturesSizeMb` counter to the profiler.

Fix https://github.com/defold/defold/issues/7743

## PR checklist

* [x] Code
* [ ] Documentation

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_gui.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_gui.cpp
@@ -2338,6 +2338,7 @@ namespace dmGameSystem
             dmLogError("Error when finalizing gui component: %d.", result);
         }
         dmGui::ClearTextures(gui_component->m_Scene);
+        dmGui::ForceRemoveDynamicTextures(gui_component->m_Scene, &DeleteTexture);
         dmGui::ClearFonts(gui_component->m_Scene);
         dmGui::ClearNodes(gui_component->m_Scene);
         dmGui::ClearLayouts(gui_component->m_Scene);
--- a/engine/gamesys/src/gamesys/components/comp_gui.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_gui.cpp
@@ -807,7 +807,7 @@ namespace dmGameSystem
     {
         GuiComponent* gui_component = (GuiComponent*)*params.m_UserData;
         dmGui::Result result = dmGui::FinalScene(gui_component->m_Scene);
-        dmGui::ForceRemoveDynamicTextures(gui_component->m_Scene, &DeleteTexture);
+        dmGui::ClearDynamicTextures(gui_component->m_Scene, &DeleteTexture);
         if (result != dmGui::RESULT_OK)
         {
             // TODO: Translate result
@@ -2338,7 +2338,7 @@ namespace dmGameSystem
             dmLogError("Error when finalizing gui component: %d.", result);
         }
         dmGui::ClearTextures(gui_component->m_Scene);
-        dmGui::ForceRemoveDynamicTextures(gui_component->m_Scene, &DeleteTexture);
+        dmGui::ClearDynamicTextures(gui_component->m_Scene, &DeleteTexture);
         dmGui::ClearFonts(gui_component->m_Scene);
         dmGui::ClearNodes(gui_component->m_Scene);
         dmGui::ClearLayouts(gui_component->m_Scene);
--- a/engine/gamesys/src/gamesys/components/comp_gui.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_gui.cpp
@@ -807,7 +807,7 @@ namespace dmGameSystem
     {
         GuiComponent* gui_component = (GuiComponent*)*params.m_UserData;
         dmGui::Result result = dmGui::FinalScene(gui_component->m_Scene);
-        dmGui::ClearDynamicTextures(gui_component->m_Scene, &DeleteTexture);
+        dmGui::DeleteDynamicTextures(gui_component->m_Scene, &DeleteTexture);
         if (result != dmGui::RESULT_OK)
         {
             // TODO: Translate result
@@ -2338,7 +2338,7 @@ namespace dmGameSystem
             dmLogError("Error when finalizing gui component: %d.", result);
         }
         dmGui::ClearTextures(gui_component->m_Scene);
-        dmGui::ClearDynamicTextures(gui_component->m_Scene, &DeleteTexture);
+        dmGui::DeleteDynamicTextures(gui_component->m_Scene, &DeleteTexture);
         dmGui::ClearFonts(gui_component->m_Scene);
         dmGui::ClearNodes(gui_component->m_Scene);
         dmGui::ClearLayouts(gui_component->m_Scene);
--- a/engine/gamesys/src/gamesys/components/comp_gui.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_gui.cpp
@@ -806,8 +806,7 @@ namespace dmGameSystem
     static dmGameObject::CreateResult CompGuiFinal(const dmGameObject::ComponentFinalParams& params)
     {
         GuiComponent* gui_component = (GuiComponent*)*params.m_UserData;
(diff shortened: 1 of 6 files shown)


engine/profiler/src/profile_render.cpp
```diff
@@ -86,7 +86,7 @@ namespace dmProfileRender
     static const int SAMPLE_FRAMES_COUNT_WIDTH = 3 * CHARACTER_WIDTH;
 
     static const int COUNTERS_NAME_WIDTH  = 32 * CHARACTER_WIDTH;
-    static const int COUNTERS_COUNT_WIDTH = 8 * CHARACTER_WIDTH;
+    static const int COUNTERS_COUNT_WIDTH = 12 * CHARACTER_WIDTH;
```
AGulev, line 89: image
Make sure we have enough space

engine/gui/src/gui.cpp
```diff
@@ -49,6 +49,7 @@ DM_PROPERTY_U32(rmtp_GuiStaticTextures, 0, FrameReset, "", &rmtp_Gui);
 DM_PROPERTY_U32(rmtp_GuiDynamicTextures, 0, FrameReset, "", &rmtp_Gui);
 DM_PROPERTY_U32(rmtp_GuiTextures, 0, FrameReset, "", &rmtp_Gui);
 DM_PROPERTY_U32(rmtp_GuiParticlefx, 0, FrameReset, "", &rmtp_Gui);
+DM_PROPERTY_U32(rmtp_GuiDynamicTexturesSize, 0, NoFlags, "size of dynamic tex in bytes", &rmtp_Gui);
```
AGulev, line 52: The counter without `FrameReset`. We don't check it every frame

engine/gui/src/gui.cpp
```diff
@@ -612,6 +613,9 @@ Result DeleteDynamicTexture(HScene scene, const dmhash_t texture_hash)
             t->m_Buffer = 0;
         }
 
+        uint32_t expected_buffer_size = t->m_Width * t->m_Height * dmImage::BytesPerPixel(t->m_Type);
+        DM_PROPERTY_ADD_U32(rmtp_GuiDynamicTexturesSize, -expected_buffer_size);
```
AGulev, line 617: Add and remove values in the exact moment we create, update or remove texture

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -806,6 +807,7 @@ namespace dmGameSystem
     {
         GuiComponent* gui_component = (GuiComponent*)*params.m_UserData;
         dmGui::Result result = dmGui::FinalScene(gui_component->m_Scene);
+        dmGui::ForceRemoveDynamicTextures(gui_component->m_Scene, &DeleteTexture);
```
AGulev, line 810: I thought about it and not sure if we should do it in `CompGuiOnReload` as well?

britzl: Yes, seems reasonable to also do it in reload.

AGulev: It makes sense. Dine.

AGulev:
Ready for review

engine/gui/src/gui.h
```diff
@@ -751,6 +751,12 @@ namespace dmGui
      */
     void RenderScene(HScene scene, const RenderSceneParams& params, void* context);
 
+    /**
+     * Force deletion of the dynamic textures of the scene
+     * @param scene Scene for which to remove the dynamic textures
```
britzl, line 756: Missing param m_DeleteTexture

engine/gui/src/gui.h
```diff
@@ -751,6 +751,12 @@ namespace dmGui
...
+     */
+    void ForceRemoveDynamicTextures(HScene scene, DeleteTexture m_DeleteTexture);
```
britzl, line 758: The description says "force deletion" while the function name is "ForceRemove". I think "Delete" is more appropriate than "Remove".

I also wonder about "Force". Can't this be omitted from the name? DeleteDynamicTextures() seems descriptive enough of what the function does.

britzl (changes requested):
Some questions about naming, but otherwise good!

engine/gui/src/gui.h
AGulev: Thx. Also it should be `deleteTexture`

engine/gui/src/gui.h
AGulev: We already have `ClearTextures()`, so I think `ClearDynamicTextures` would be better

britzl: Hmm, but we are deleting the textures are we not? ClearTextures() seems to only cancel animations and null the pointers/handles. I don't think ClearDynamicTextures() properly conveys that the textures are deleted.

britzl (dismissed):
I still have a question about clear vs delete, but otherwise good.

engine/gui/src/gui.h
AGulev: Now it's `DeleteDynamicTextures()`

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -806,6 +807,7 @@ namespace dmGameSystem
     {
         GuiComponent* gui_component = (GuiComponent*)*params.m_UserData;
         dmGui::Result result = dmGui::FinalScene(gui_component->m_Scene);
+        dmGui::DeleteDynamicTextures(gui_component->m_Scene, &DeleteTexture);
```
JCash, line 810: After "Final" I expect the scene to be invalid (i.e. deallocated).

And I think deleting of the textures should be handled by the FinalScene.

Since the pattern is that dmGui takes some Params struct, I suggest it is the pattern we should use here as well.

engine/gui/src/gui.h
```diff
@@ -751,6 +751,13 @@ namespace dmGui
      */
     void RenderScene(HScene scene, const RenderSceneParams& params, void* context);
 
+    /**
+     * Delete all dynamic textures from the scene.
+     * @param scene Scene for which to remove the dynamic textures
+     * @param deleteTexture DeleteTexture Callback to delete a texture
+     */
+    void DeleteDynamicTextures(HScene scene, DeleteTexture deleteTexture);
```
JCash, line 759: Our arguments are using `snake_case` -> `delete_texture`

engine/profiler/src/profile_render.cpp
JCash: An option is to use `GuiDynamicTextureSizeKb`and divide by 1024. Or use Mb even

JCash (changes requested):
Some implementation details.

engine/profiler/src/profile_render.cpp
AGulev: I like it! Done!

JCash (changes requested):
Some minor changes and a question

engine/gui/src/gui.cpp
```diff
@@ -2104,7 +2086,7 @@ namespace dmGui
         return RunScript(scene, SCRIPT_FUNCTION_INIT, LUA_NOREF, 0x0);
     }
 
-    Result FinalScene(HScene scene)
+    Result FinalScene(HScene scene, DeleteTexture delete_texture)
```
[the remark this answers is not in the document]
AGulev: I just didn't want to create a structure for just one parameter.
for example we have `UpdateScene(HScene scene, float dt)`
`DispatchMessage(HScene scene, dmMessage::Message* message);`
So if we will need more parameters we will create a struct for it

britzl (approved):

PR #7828, opened 2023-07-24
Add `New window` option in Dock

### LINKED ISSUE #4863: Add a menu option "Open new window" to launch icon on macOS
Add a menu option "Open new window" to launch icon on MAC OS, which would open a new editor window where user can open, for example a demo project.

AGulev:
New option `New window` for task bar menu added. It opens a new instance of the editor.

Fixes https://github.com/defold/defold/issues/4863

## PR checklist

* [ ] Code
* [ ] Documentation

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -1484,6 +1484,8 @@ If you do not specifically require different script states, consider changing th
 (handler/defhandler :reload-stylesheet :global
   (run [] (ui/reload-root-styles!)))
 
+; This function re-implemented in Java in Start.java->newWindow(). If you change something
+; please, change it there as well.
 (handler/defhandler :open-project :global
   (active? [] (and (system/defold-resourcespath) (system/defold-launcherpath)))
   (run [] (let [resources-path (system/defold-resourcespath)
--- a/editor/src/java/com/defold/editor/Start.java
+++ b/editor/src/java/com/defold/editor/Start.java
@@ -30,6 +30,7 @@ import org.slf4j.LoggerFactory;
 import org.slf4j.bridge.SLF4JBridgeHandler;
 
 import java.awt.image.BufferedImage;
+import java.awt.Desktop;
 import java.io.File;
 import java.lang.reflect.Method;
 import java.nio.file.Path;
@@ -117,19 +118,7 @@ public class Start extends Application {
     @Override
     public void start(Stage primaryStage) throws Exception {
         try {
-            /*
-              Note
-              Don't remove
-
-              Background
-              Before the mysterious line below Command-H on OSX would open a generic Java about dialog instead of hiding the application.
-              The hypothosis is that awt must be initialized before JavaFX and in particular on the main thread as we're pooling stuff using
-              a threadpool.
-              Something even more mysterious is that if the construction of the buffered image is moved to "static void main(.." we get a null pointer in
-              clojure.java.io/resource..
-            */
-
-            new BufferedImage(1, 1, BufferedImage.TYPE_3BYTE_BGR);
+            Desktop.getDesktop().setAboutHandler(null);
 
             // Clean up old packages as they consume a lot of hard drive space.
             // NOTE! This is a temp hack to give some hard drive space back to users.
--- a/editor/src/java/com/defold/editor/Start.java
+++ b/editor/src/java/com/defold/editor/Start.java
@@ -188,22 +188,26 @@ public class Start extends Application {
             throw t;
         }
 
-        if (Taskbar.isTaskbarSupported() && isPathsAvailable()) {
-            Taskbar taskbar = Taskbar.getTaskbar();
-            PopupMenu dockMenu = new PopupMenu();
-            MenuItem dockMenuItem = new MenuItem("New window");
-            dockMenuItem.addActionListener(new ActionListener() {
-                @Override
-                public void actionPerformed(ActionEvent e) {
-                     try {
-                        newWindow();
-                    } catch (Throwable t) {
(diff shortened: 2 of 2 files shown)


editor/src/java/com/defold/editor/Start.java
```diff
@@ -148,6 +184,24 @@ public void start(Stage primaryStage) throws Exception {
...
+                     try {
+                        newWindow();
+                    } catch (Throwable t) {
+                        logger.error("Can't open new window", t);
+                    }
+                }
+            });
+            dockMenu.add(dockMenuItem);
+            taskbar.setMenu(dockMenu);
+        }
```
AGulev, line 204: Screenshot 2023-07-24 at 09 18 35

editor/src/java/com/defold/editor/Start.java
```diff
@@ -116,14 +125,41 @@ private void prunePackages() {
...
     @Override
     public void start(Stage primaryStage) throws Exception {
         try {
             Desktop.getDesktop().setAboutHandler(null);
         } catch (final UnsupportedOperationException e) {
-            System.out.println("The os does not support: 'desktop.setAboutHandler'");
+            logger.error("The os does not support: 'desktop.setAboutHandler'", e);
         } catch (final SecurityException e) {
-            System.out.println("There was a security exception for: 'desktop.setAboutHandler'");
+            logger.error("There was a security exception for: 'desktop.setAboutHandler'", e);
```
AGulev, line 162: we have logger here. I missed it.

editor/src/clj/editor/app_view.clj
```diff
@@ -1484,6 +1484,8 @@ If you do not specifically require different script states, consider changing th
 (handler/defhandler :reload-stylesheet :global
   (run [] (ui/reload-root-styles!)))
 
+; This function re-implemented in Java in Start.java->newWindow(). If you change something
+; please, change it there as well.
```
AGulev, line 1488: I would prefer to reuse java code in Clojure. But I think it's not something we want most of the time. So I just added this comment here to make sure we keep it sync.

AGulev:
Tested on win/mac/linux

britzl (approved):

JCash:
Do we have race condition for resources between the instances?
E.g.:
1) Opening the same project twice
2) "Global" files such as settings etc.
3) Are editor scripts affected?

AGulev:
@JCash I don't know, but we already can create a new project from `File->Open Project...`

JCash:
Ok, waiting for the first issue then :) 🤷

AGulev:
> Ok, waiting for the first issue then :) 🤷

I expect these things were investigated when the initial functionality was implemented.
I think adding one more place for doing the same, users can already do from another place for a couple of months shouldn't affect anything. But I'll wait for review from somebody from the editor team.

editor/src/java/com/defold/editor/Start.java
```diff
@@ -151,6 +187,30 @@ public void start(Stage primaryStage) throws Exception {
             logger.error("failed start", t);
             throw t;
         }
+
+        try {
+            if (Taskbar.isTaskbarSupported() && Taskbar.getTaskbar().isSupported(Taskbar.Feature.MENU) && isPathsAvailable()) {
+                Taskbar taskbar = Taskbar.getTaskbar();
+                PopupMenu dockMenu = new PopupMenu();
+                MenuItem dockMenuItem = new MenuItem("New window");
```
matgis, line 195: I think it should be "New Window", with a capitalized "W", to match the capitalization of the other entries in the menu?

editor/src/java/com/defold/editor/Start.java
```diff
@@ -151,6 +187,30 @@ public void start(Stage primaryStage) throws Exception {
...
+                        } catch (Throwable t) {
+                            logger.error("Can't open new window", t);
+                        }
+                    }
+                });
+                dockMenu.add(dockMenuItem);
+                taskbar.setMenu(dockMenu);
+            }
+        } catch (final UnsupportedOperationException e) {
+            logger.error("The OS does not support taskbar menu", e);
```
matgis, line 210: Can this even happen after the checks we do? If so, it should probably be logged as info, not an error.

editor/src/java/com/defold/editor/Start.java
```diff
@@ -151,6 +187,30 @@ public void start(Stage primaryStage) throws Exception {
...
+        try {
+            if (Taskbar.isTaskbarSupported() && Taskbar.getTaskbar().isSupported(Taskbar.Feature.MENU) && isPathsAvailable()) {
+                Taskbar taskbar = Taskbar.getTaskbar();
+                PopupMenu dockMenu = new PopupMenu();
+                MenuItem dockMenuItem = new MenuItem("New window");
+                dockMenuItem.addActionListener(new ActionListener() {
+                    @Override
+                    public void actionPerformed(ActionEvent e) {
+                        try {
+                            newWindow();
```
matgis, line 200: We shouldn't re-implement the functionality in Java, rather, we should either invoke the handler directly (although I can't find a convenient way to do that? I think is should be done how we do it from `editor.keymap/execute-command`).

Or we could break out the code in the `:open-project` `defhandler` into a Clojure function that we place next to the handler and call the Clojure function from here. Although to be honest I don't see why this code has to be in Java at all. Couldn't we do this from `boot-open-project/load-stage!` where we do all the other Window and Menu setup?

matgis (changes requested):
Unless there is some reason for needing to do this before we do anything else, I think we should move the code that adds the context menu item into `boot-open-project/load-stage!` where we do the rest of the Window and Menu setup. That also makes it easy to invoke the existing Clojure code.

PR #7832, opened 2023-07-25
Release script for Steam uploads

britzl:
This PR adds support for uploading the editor to Steam using a separate build step.

### DIFF
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -1683,8 +1683,8 @@ class Configuration(object):
             release_to_github.release(self, tag_name, release_sha1, releases[0], release_name=release_name, body=body, prerelease=prerelease, editor_only=is_editor_branch)
         
         # Release to steam for stable only
-        # if tag_name and (self.channel == 'stable'):
-        #     release_to_steam.release(self, tag_name, release_sha1, releases[0])
+        # if tag_name and (self.channel == 'editor-alpha'):
+        #     self.release_to_steam()
 
     # E.g. use with ./scripts/build.py release_to_github --github-token=$CITOKEN --channel=editor-alpha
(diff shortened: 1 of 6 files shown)


scripts/build.py
```diff
@@ -1681,6 +1681,10 @@ def release(self):
             body = self._get_github_release_body()
             release_name = 'v%s - %s' % (self.version, engine_channel or self.channel)
             release_to_github.release(self, tag_name, release_sha1, releases[0], release_name=release_name, body=body, prerelease=prerelease, editor_only=is_editor_branch)
+        
+        # Release to steam for stable only
+        # if tag_name and (self.channel == 'editor-alpha'):
+        #     self.release_to_steam()
```
AGulev, line 1687: I thought about once a month release (push to master) but now I think maybe you are right and do it the same way as we do for internal updates - makes sense

AGulev (dismissed):
I thought you're going to use a GitHub action, but you wrote everything yourself O_o

(pay attention that build failed)

scripts/build.py
britzl: My thinking is that we automatically release to Steam when we release a normal stable release (this is a relatively recent change that we automatically release to stable). Although I've decided to disable it for now since I haven't had time to test the release flow manually on a Linux machine yet. So the idea for now is that we do the release to steam as a manual step and the end of the release flow.

britzl:
> I thought you're going to use a GitHub action, but you wrote everything yourself O_o
> 
> (pay attention that build failed)

I skipped using a GitHub action since I want to be able to also do manual releases to Steam using a python script.

AGulev (approved):

PR #7866, opened 2023-08-07
Custom vertex format support for ParticleFX components

### LINKED ISSUE #7818: Custom vertex formats (particlefx support)
Same as #7427  but with support for particle effect emitters (non-gui)

Jhonnyg:
Particle FX nodes can now utilise the custom vertex formats previously added to sprites. This enables you to pass custom data from the editor into shader attributes when rendering particles.

Fixes #7818 

## PR checklist

* [x] Code
* [ ] Documentation


editor/src/clj/editor/gl/vertex.clj
```diff
@@ -327,21 +327,6 @@ the `do-gl` macro from `editor.gl`."
...
-  [layout ^ByteBuffer buffer]
-  (assert layout)
-  (assert (= 0 (mod (.limit buffer) (:vertex-size layout))))
-  (let [^ByteBuffer buffer (.duplicate buffer)
-        limit         (.limit buffer)
-        count         (mod limit (:vertex-size layout))
-        buffer-starts (buffer-starts limit layout)
-        slices        (b/slice buffer (map min (repeat limit) buffer-starts))]
-    (->PersistentVertexBuffer layout count buffer slices (AtomicLong. count) not-allowed not-allowed)))
-
```
Jhonnyg, line 344: No used anywhere now

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
```diff
@@ -105,23 +104,18 @@ namespace dmGameSystem
...
         memset(world->m_ConstantBuffers.Begin(), 0, sizeof(dmRender::HNamedConstantBuffer)*max_emitter_count);
 
-        uint32_t buffer_size = dmParticle::GetVertexBufferSize(ctx->m_MaxParticleCount, dmParticle::PARTICLE_GO);
+        // position   : 3
+        // color      : 4
+        // texcoord0  : 2
+        // page_index : 1
+        const uint32_t default_vx_size = sizeof(float) * (3 + 4 + 2 + 1);
+        const uint32_t buffer_size     = ctx->m_MaxParticleCount * 6 * default_vx_size;
+        world->m_VertexBufferData.SetCapacity(ctx->m_MaxParticleCount * 6 * default_vx_size);
```
Jhonnyg, line 113: We don't have any particle "formats" now, the size of the vx buffer is only dependent on the vertex size now.

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
```diff
@@ -254,48 +248,118 @@ namespace dmGameSystem
...
+        infos->m_VertexStride = dmGraphics::GetVertexDeclarationStride(dmRender::GetVertexDeclaration(material));
+
+        for (int i = 0; i < infos->m_NumInfos; ++i)
+        {
+            dmParticle::ParticleVertexAttributeInfo& info = infos->m_Infos[i];
+            infos->m_Infos[i].m_NameHash = material_attributes[i].m_NameHash;
+            dmRender::GetMaterialProgramAttributeValues(material, i, (const uint8_t**) &info.m_ValuePtr, &info.m_ValueByteSize);
+        }
+    }
+
```
Jhonnyg, line 292: Future improvement: all of these functions are almost identical to comp_sprite, would be nice to merge them since we will follow this pattern later for other components.

engine/graphics/src/wscript
```diff
@@ -27,6 +27,15 @@ def build(bld):
...
                     target          = 'graphics_proto')
 
+    bld.add_group()
+
+    obj = bld.stlib(features        = obj.features + ['skip_asan'],
+                    includes        = obj.includes,
+                    source          = bld.path.ant_glob(['graphics_proto.cpp']) + bld.path.parent.ant_glob('proto/graphics/*'),
+                    protoc_includes = obj.protoc_includes,
+                    use             = obj.use,
+                    target          = 'graphics_proto_noasan')
```
Jhonnyg, line 37: Particle now depends on graphics being built first, since we use data structures from that library

editor/src/clj/editor/particle_lib.clj
```diff
@@ -133,7 +117,7 @@
     {:context context
      :prototype prototype
      :instances instances
-     :raw-vbufs (make-raw-vbufs (:emitters prototype-msg))
+     :raw-vbufs []
```
Jhonnyg, line 120: These are now created on the fly, since we don't actually know what to create before it's used. As in, the size of the vbuffer depend on the size of a vertex now instead of a fixed vertex size, as it was before.

editor/src/clj/editor/particle_lib.clj
```diff
@@ -177,20 +161,72 @@
...
+    :semantic-type-position 2
+    :semantic-type-texcoord 3
+    :semantic-type-page-index 4
+    :semantic-type-color 5))
+
+(defn- coordinate-space->int [coordinate-space]
+  (case coordinate-space
+    :coordinate-space-world 1
+    :coordinate-space-local 2
+    2))
```
Jhonnyg, line 180: Should we use java enums or something like that here instead?

engine/particle/src/particle.cpp
```diff
@@ -45,18 +45,21 @@ namespace dmParticle
...
+    const char* MAX_PARTICLE_COUNT_KEY = "particle_fx.max_particle_count";
 
     /// Used for degree to radian conversion
     const float DEG_RAD = (float) (M_PI / 180.0);
 
     /// Simulate motion blur at 60 fps with a 180 deg shutter
     const static float STRETCH_SCALING = (1.0f/60.0f) * 0.5f;
 
+    // JG: Would be nice to share these somewhere more locally, like a constant hash directory in dlib or something..
+    static const dmhash_t VERTEX_STREAM_COLOR = dmHashString64("color");
```
Jhonnyg, line 61: I need to have this value somewhere since we need to distinguish between a "built-in" color stream from the engine and a custom stream. Or in short, we can't just rely on the semantic type to output a color, we need the name of the stream as well. (See how we write vertices later on..)

engine/particle/src/particle.cpp
```diff
@@ -1016,12 +1030,79 @@ namespace dmParticle
...
+        for (int i = 0; i < attribute_infos.m_NumInfos; ++i)
+        {
+            if (attribute_infos.m_Infos[i].m_SemanticType == dmGraphics::VertexAttribute::SEMANTIC_TYPE_POSITION &&
+                attribute_infos.m_Infos[i].m_CoordinateSpace == dmGraphics::COORDINATE_SPACE_LOCAL)
+            {
+                return true;
+            }
+        }
+        return false;
+    }
```
Jhonnyg, line 1044: A small performance thing, we can skip calculating local positions in the update function if we don't need it.

engine/particle/src/particle.cpp
```diff
@@ -1016,12 +1030,79 @@ namespace dmParticle
...
+                {
+                    if (info.m_NameHash == VERTEX_STREAM_COLOR)
+                    {
+                        memcpy(write_ptr, &color, info.m_ValueByteSize);
+                    }
+                    else
+                    {
+                        memcpy(write_ptr, info.m_ValuePtr, info.m_ValueByteSize);
+                    }
+                } break;
```
Jhonnyg, line 1080: As previously mentioned, we need to know if the stream has the name "color" or not, if that is the case we output out the particle color and otherwise we put the binary value from the actual attribute.

engine/particle/src/particle.cpp
```diff
@@ -1215,82 +1305,38 @@ namespace dmParticle
...
         {
             if (emitter->m_RenderWarning == 0)
             {
-                const char* config_key = MAX_PARTICLE_COUNT_KEY;
-
-                if (format == PARTICLE_GUI)
-                    config_key = "gui.max_particle_count";
-
-                dmLogWarning("Maximum number of particles (%d) exceeded, particles will not be rendered. Change \"%s\" in the config file.", context->m_MaxParticleCount, config_key);
+                res = GENERATE_VERTEX_DATA_MAX_PARTICLES_EXCEEDED;
```
Jhonnyg, line 1333: Each component has to deal with the error by themselves.

engine/particle/src/particle.cpp
```diff
@@ -2079,19 +2133,57 @@ namespace dmParticle
...
+        uint32_t ix = context->m_AttributeDataPtrIndex;
+        context->m_AttributeDataPtrIndex++;
+
+        if (ix >= context->m_AttributeDataPtrs.Capacity())
+        {
+            void* new_ptr = malloc(ATTRIBUTE_WRAPPER_SIZE);
+            context->m_AttributeDataPtrs.OffsetCapacity(1);
+            context->m_AttributeDataPtrs.Push(new_ptr);
+            return new_ptr;
+        }
```
Jhonnyg, line 2160: We need to retain the same pointers between every simulation step, which other expanding array solutions will not guarantee, and I don't think we want to preallocate a large array for this since the scratch buffer approach is ONLY for the editor.

matgis (changes requested):
Found some minor niggles on the editor side, but they shouldn't be a lot of effort to fix!

engine/particle/src/particle.cpp
JCash: Isn't this true for the other cases?
If the user set a POSITION or TEXCOORD attribute with name "extra", it will get the predefined position/uv values.
Or isn't is allowed, and should that a build error?

Also, do we want to look at the type?
E.g. if the user really wants to use a single uint32_t instead of float4 for color for bandwidth.

Later, I also feel this "write logic" should be standardized/centralized somehow.

engine/particle/src/particle.cpp
```diff
@@ -2079,19 +2133,57 @@ namespace dmParticle
         stats->m_Time = i->m_PlayTime;
     }
 
-    uint32_t GetVertexBufferSize(uint32_t particle_count, ParticleVertexFormat vertex_format)
+    uint32_t GetVertexBufferSize(uint32_t particle_count, uint32_t vertex_size)
+    {
+        return particle_count * 6 * vertex_size;
+    }
+
+    uint32_t GetMaxVertexBufferSize(HParticleContext context, uint32_t vertex_size)
```
JCash, line 2141: I noticed this function before, and it is redundant imho?

engine/particle/src/particle.cpp
```diff
@@ -2079,19 +2133,57 @@ namespace dmParticle
...
+    void* WriteAttributeToScratchBuffer(HParticleContext context, void* bytes, uint32_t byte_count)
+    {
+        if (byte_count == 0)
+        {
+            return 0;
+        }
+
+        void* wrapper = AcquireAttributeDataPtr(context);
+
+        assert(byte_count < ATTRIBUTE_WRAPPER_SIZE);
```
JCash, line 2175: I think an assert here will take down the editor.

JCash (changes requested):
Mainly the assert in the editor only C++ code needs attention.

Also some questions that may warrant some clarifications.

editor/src/clj/editor/particle_lib.clj
Jhonnyg: We can use the protobuf enums

editor/src/clj/editor/particle_lib.clj
```diff
@@ -14,31 +14,25 @@
...
-           [com.defold.libs ParticleLibrary ParticleLibrary$Vector3 ParticleLibrary$Vector4 ParticleLibrary$Quat
-            ParticleLibrary$AnimationData ParticleLibrary$FetchAnimationCallback
-            ParticleLibrary$FetchAnimationResult ParticleLibrary$AnimPlayback
-            ParticleLibrary$Stats ParticleLibrary$InstanceStats
-            ParticleLibrary$RenderInstanceCallback]
+           [com.defold.libs ParticleLibrary ParticleLibrary$ParticleVertexAttributeInfos ParticleLibrary$ParticleVertexAttributeInfo ParticleLibrary$Vector3 ParticleLibrary$Vector4 ParticleLibrary$Quat
+                            ParticleLibrary$AnimationData ParticleLibrary$FetchAnimationCallback
+                            ParticleLibrary$FetchAnimationResult ParticleLibrary$AnimPlayback
+                            ParticleLibrary$Stats ParticleLibrary$InstanceStats
+                            ParticleLibrary$RenderInstanceCallback]
```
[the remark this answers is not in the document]
Jhonnyg: It's weird, I didn't write it like that! Some automatic function did, but I don't know which one 🤔

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -1060,7 +1081,7 @@ namespace dmGameSystem
         dmGui::NodeType node_type = dmGui::GetNodeType(scene, first_node);
         assert(node_type == dmGui::NODE_TYPE_PARTICLEFX);
 
-        uint32_t vb_max_size = dmParticle::GetMaxVertexBufferSize(gui_world->m_ParticleContext, dmParticle::PARTICLE_GUI) - gui_world->m_RenderedParticlesSize;
+        uint32_t vb_max_size = dmParticle::GetMaxVertexBufferSize(gui_world->m_ParticleContext, sizeof(ParticleGuiVertex)) - gui_world->m_RenderedParticlesSize;
```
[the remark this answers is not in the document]
Jhonnyg: > Not sure what the best way is to define the particle budget now.
> Previously it was bytes, but now that we can support multiple materials with different vertex formats, it might be relevant to think of #particles.
> 🤔

What budget are you referring to here? Both the GUI and the particle settings specify particle count (and components)

image

> The usage of the sizeof(ParticleGuiVertex) seems a little strange now. Perhaps add a comment or two to show that the size is an estimate, and that the code is works ok?
> 
> E.g. the vertex count relies on that size

But for GUI the size isn't an estimate, it _is_ what is allocated regardless of what material is used 🤔 this is because we don't support any custom vertex formats in GUI right now, so we don't use the configuration from the material.

editor/src/clj/editor/particlefx.clj
```diff
@@ -386,18 +398,30 @@
...
-        (let [context (:context pfx-sim)
-              vbuf (plib/get-emitter-vertex-data pfx-sim emitter-index)]
+      (when-let [pfx-sim-atom (when (and emitter-sim-data pfx-sim-request-id)
+                                (:pfx-sim (scene-cache/lookup-object ::pfx-sim pfx-sim-request-id nil)))]
+        (let [pfx-sim @pfx-sim-atom
+              raw-vbuf (plib/gen-emitter-vertex-data pfx-sim emitter-index color max-particle-count vertex-description shader-bound-attributes vertex-attribute-bytes)
+              context (:context pfx-sim)
+              vbuf (vtx/wrap-vertex-buffer vertex-description :static raw-vbuf)
+              all-raw-vbufs (vset (:raw-vbufs pfx-sim) emitter-index raw-vbuf)]
+          (swap! pfx-sim-atom assoc :raw-vbufs all-raw-vbufs)
```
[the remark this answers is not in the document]
Jhonnyg: Yeah, it makes things more complicated for sure... But hm, I'll add a note about this for now and it will have to be a headache for the future I guess 🤷

engine/particle/src/particle.cpp
Jhonnyg: Yeah that's true I suppose. The problem I ran into was that the "color" semantic is the only one that has any value in the editor since we decorate the property with a color widget. So the way we had originally setup the semantic types means that I can't decorate an attribute with a color widget, but maybe that's fine and should be a separate "edit type" value or something.. But yes, if we are to conform these vertex writing functions into a shared location I think we should produce the built-in data based on the semantic types.

> Also, do we want to look at the type?
> E.g. if the user really wants to use a single uint32_t instead of float4 for color for bandwidth.

Do you mean with the built-in semantic types? Because the data produced from the content pipeline is narrowed or expanded to the correct type, but the built-in types are not. For the built-in types we need to do data conversion between the floats (it's all float streams I think?) into whatever format. I think we should do it, but I'd rather try to get this PR in first and then take a stab at merging the vertex writing function in a separate PR.

PR #7887, opened 2023-08-15
Separate glyph data from fonts

### LINKED ISSUE #7360: Double font resource with different materials included to build
For example I have two similar font resource (all .font settings are equal), but they have different materials.

If I will build the game, it will include both of this fonts. Seems it should use only one of them.

My case: I'm using world gui material, so I have to use the font twice. But I don't want to include the resource itself several times

**To Reproduce (REQUIRED)**
- Load the example ([issue-double-font-resource.zip](https://github.com/defold/defold/files/10573770/issue-double-font-resource.zip))
- Build with build report
- See the result

The font resources with equal params should be included only once

- Any

**Platforms (REQUIRED):**
- Any

[issue-double-font-resource.zip](https://github.com/defold/defold/files/10573770/issue-double-font-resource.zip)

**Screenshots (OPTIONAL):**
Screenshot 2023-02-02 at 23 53 30

Jhonnyg:
Improved the font resource memory footprint by separating glyph data from common font settings such as material and shadow parameters that are generally not needed for the glyph data. Building a font file will not result in two resources on disk, one '.fontc' and one '.glyph_bankc' respectively. The glyph bank may be shared between two or more font files depending on font settings.

Fixes #7360

## PR checklist

* [x] Code
* [ ] Documentation


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ArchiveBuilder.java
```diff
@@ -371,6 +371,16 @@ public static void main(String[] args) throws IOException, NoSuchAlgorithmExcept
...
 
+                if (args[i].endsWith("fontc")) {
+                    File glyphBankFile = new File(args[i].replace("fontc", "glyph_bankc"));
+
+                    if (!glyphBankFile.isFile()) {
+                        printUsageAndTerminate("file does not exist: " + glyphBankFile.getAbsolutePath());
+                    }
+
+                    inputs.add(glyphBankFile);
+                }
```
Jhonnyg, line 382: Not sure if there's a better way to solve this.. we need to be able to include source files that generate more than one build target, which we currently don't support.

engine/render/src/waf_render.py
```diff
@@ -37,8 +37,10 @@ def transform_material(task, msg):
 def font_file(self, node):
     classpath = [self.env['DYNAMO_HOME'] + '/share/java/bob-light.jar']
     fontmap = self.create_task('fontmap')
+
     fontmap.env['CLASSPATH'] = os.pathsep.join(classpath)
     fontmap.env['CONTENT_ROOT'] = fontmap.generator.content_root
+    fontmap.env['PROJECT_ROOT'] = fontmap.generator.project_root
```
Jhonnyg, line 43: This feels a bit like a hack, but I don't want to break my back into fixing this the "right way" according to waf. I'd rather replace content building in waf with something simpler like a shell script..

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/FontBuilderTest.java
```diff
@@ -71,9 +92,38 @@ public void testFNTSubDir() throws Exception {
...
+        srcThree.append("material: \"/test2.material\"\n");
+        srcThree.append("size: 16\n");
+        srcThree.append("shadow_x: 1337.0\n");
+
+        FontMap fontMapOne   = getFontMap(build("/test1.font", srcOne.toString()));
+        FontMap fontMapTwo   = getFontMap(build("/test2.font", srcTwo.toString()));
+        FontMap fontMapThree = getFontMap(build("/test3.font", srcThree.toString()));
+
+        assertEquals(fontMapOne.getGlyphBank(), fontMapTwo.getGlyphBank());
+        assertEquals(fontMapOne.getGlyphBank(), fontMapThree.getGlyphBank());
```
Jhonnyg, line 124: image

matgis (dismissed):
Editor and protobuf changes look good to me!

engine/engine/content/wscript
```diff
@@ -40,6 +40,7 @@ def build(bld):
     builtins = bld(features='barchive',
                    source_root='default/content',
                    content_root='../content',
+                   project_root='content',
```
JCash, line 43: Having two variables pointing out the same path seems a bit odd.

What was the issue with using "content_root"? That it starts with "../" ?

engine/gamesys/src/gamesys/resources/res_glyph_bank.h
```diff
@@ -0,0 +1,37 @@
+// Copyright 2020-2023 The Defold Foundation
```
JCash, line 1: For most (all?) other res_*.h, we have them in the dmsdk folder, since some extensions might need them.
I think this one is ok to be private (at least for now) since even res_font_map.h doesn't refer to this data, and we have no current use case for making it public.
As long as it builds when using the extender, I'm happy.

engine/render/src/waf_render.py
JCash: Well, it has less to do with Waf, and more to do with what path we want uin our tool chain.
Currently, the cases I've seen is that the content_path and project_path point to the same path, only with different formats (relative vs absolute).

scripts/unpack_ddf.py
```diff
@@ -51,6 +52,8 @@
 BUILDERS['.collectionc']    = gameobject.gameobject_ddf_pb2.CollectionDesc
 BUILDERS['.luac']           = gameobject.lua_ddf_pb2.LuaModule
 BUILDERS['.materialc']      = render.material_ddf_pb2.MaterialDesc
+BUILDERS['.fontc']          = render.font_ddf_pb2.FontMap
+BUILDERS['.glyph_bankc']    = render.font_ddf_pb2.GlyphBank
```
JCash, line 56: ❤️

JCash (changes requested):
LGTM

However, I'd like some cleanup regarding the extra project path to the tool chain

PR #7890, opened 2023-08-16
Add filtering to the editor Console

### LINKED ISSUE #6145: Filter for console in the Editor
Big projects usually have a lot of logs. 
It would be useful to have not only search for console but also filter, to  show only need entries if filter activated (has a value in it)

vlaaad:
User-facing changes:
This changeset adds a Filter function to the editor console. The editor supports 2 types of filters:
- inclusive filters, e.g. only show lines with text "my-app-subsystem" with `my-app-subsystem` filter
- exclusive filters, e.g. exclude lines with text "chatty-subsystem" with `!chatty-subsystem` filter

Additionally, you can use multiple filters simultaneously.

Fixes #6145


editor/src/clj/editor/console.clj
```diff
@@ -65,33 +90,189 @@
...
+            (let [exclusion (subs exclusion-with-prefix 1)]
+              (fn exclusion-pred [entry]
+                (not (string/includes? (entry 1) exclusion)))))
+          (combine-preds [pred-combinator preds]
+            (case (count preds)
+              0 nil
+              1 (first preds)
+              (apply pred-combinator preds)))]
+    (let [[exclusions inclusions] (coll/separate-by #(and (<= 2 (count %))
+                                                          (string/starts-with? % "!"))
```
matgis, line 124: What role does `(<= 2 (count %)` play here? Isn't the `starts-with?` check sufficient?

editor/src/clj/editor/console.clj
```diff
@@ -65,33 +90,189 @@
...
 ;; Tool Bar
 ;; -----------------------------------------------------------------------------
+(def ^:private ext-with-button-props
+  (fx/make-ext-with-props fx.button/props))
+
+(defn filter-console-list-cell-view [[i [text selected] :as in]]
+  (if in
+    {:style-class "console-filter-popup-list-cell"
+     ;; somehow this line makes the check-box's label to limit its width, so it
+     ;; does no produce horizontal scrollbar, but instead truncates the text
```
matgis, line 158: Typo: "does not produce a"?

editor/src/clj/editor/console.clj
```diff
@@ -65,33 +90,189 @@
...
+                           :mnemonic-parsing false
+                           :on-selected-changed {:event-type :select :index i}
+                           :text text}
+                          {:fx/type fx.button/lifecycle
+                           :focus-traversable false
+                           :style-class "console-filter-popup-list-cell-remove-button"
+                           :graphic {:fx/type fx.region/lifecycle
+                                     :style-class "cross"}
+                           :on-action {:event-type :delete :index i}}]}}
+    {}))
```
matgis, line 179: Should we invert the `if` test so we can have this at the top?

matgis (approved):
Very nicely done! Ship it!

editor/src/clj/editor/console.clj
vlaaad: I wanted to ensure that there are at least 2 characters so it's not a single exclamation mark. After applying `(subs "!" 1)` we will get an empty string which will make us exclude every line since every line contains an empty string %)

matgis: Ah, makes sense!

vlaaad: Yes, which is weird, but less weird than excluding everything with 😄

PR #7894, opened 2023-08-17
Vulkan - Keep track of number of descriptors allocated from the pool

Jhonnyg:
This PR fixes a long running issue with incorrect tracking of number of descriptors created for the vulkan renderer.

### DIFF
--- a/engine/graphics/src/vulkan/graphics_vulkan_device.cpp
+++ b/engine/graphics/src/vulkan/graphics_vulkan_device.cpp
@@ -3,16 +3,17 @@
 // Copyright 2009-2014 Ragnar Svensson, Christian Murray
 // Licensed under the Defold License version 1.0 (the "License"); you may not use
 // this file except in compliance with the License.
-// 
+//
 // You may obtain a copy of the License, together with FAQs at
 // https://www.defold.com/license
-// 
+//
 // Unless required by applicable law or agreed to in writing, software distributed
 // under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
 #include <dlib/math.h>
+#include <dlib/log.h>
 
 #include "graphics_vulkan_defines.h"
 #include "graphics_vulkan_private.h"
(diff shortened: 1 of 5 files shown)


engine/graphics/src/vulkan/graphics_vulkan_device.cpp
```diff
@@ -74,36 +75,97 @@ namespace dmGraphics
...
+        {
+            allocator->m_DescriptorSetMax += allocator->m_DescriptorsPerPool;
+            AllocateDescriptorSets(allocator);
+        }
+
+    #ifdef _DEBUG
+        if (allocator->m_DescriptorsPerPool * allocator->m_DescriptorPools.Size() > 16768)
+        {
+            dmLogOnceWarning("Vulkan: There are more than 16768 descriptors (%d) in flight, this might be a performance issue.", m_DescriptorPools.Size());
+        }
```
Jhonnyg, line 130: The limit is basically taken from thin air, but I think we want to be notified when someone reaches this value as there will be incentive to actually do some of these improvements in our vulkan renderer

engine/graphics/src/vulkan/graphics_vulkan_device.cpp
```diff
@@ -74,36 +75,97 @@ namespace dmGraphics
...
+        DescriptorPool& pool = allocator->m_DescriptorPools[allocator->m_DescriptorPoolIndex];
+
+        if ((allocator->m_DescriptorsPerPool - pool.m_DescriptorCount) < num_descriptors)
+        {
+            allocator->m_DescriptorPoolIndex++;
+            if (allocator->m_DescriptorPoolIndex >= allocator->m_DescriptorPools.Size())
+            {
+                res = AllocateDescriptorPool(allocator, vk_device);
+            }
+        }
```
Jhonnyg, line 118: We now keep adding pools if we need to (as suggested by samsungs documentation). Previously we resized the pool by removing the old pool and replacing it with a new pool, but according to samsung this is the preferred way.

engine/graphics/src/vulkan/graphics_vulkan_device.cpp
JCash: Yeah, it does look a bit curious, as it's not really related to any power of two (16384, 32768 etc)

JCash (dismissed):

engine/graphics/src/vulkan/graphics_vulkan_device.cpp
Jhonnyg: Lol, I think I meant to use 16384 here.. 😶‍🌫️

PR #7897, opened 2023-08-18
Sign Windows editor using EV Code Signing certificate

britzl:
The Windows version of the editor is now signed using an EV Code Signing certificate to automatically gain trusted status on Microsoft Defender SmartScreen Reputation filter.


.github/workflows/editor-only.yml
```diff
@@ -15,8 +15,7 @@ env:
   DM_PACKAGES_URL: ${{ secrets.DM_PACKAGES_URL }}
   DM_ARCHIVE_DOMAIN: ${{ secrets.DM_ARCHIVE_DOMAIN }}
   DM_RELEASE_REPOSITORY: ${{ secrets.DM_RELEASE_REPOSITORY }}
-  WINDOWS_CERTIFICATE: ${{ secrets.WINDOWS_CERTIFICATE }}
-  WINDOWS_CERTIFICATE_PASS: ${{ secrets.WINDOWS_CERTIFICATE_PASS }}
+  GCLOUD_EDITOR_SERVICE_KEY: ${{ secrets.GCLOUD_EDITOR_SERVICE_KEY }}
```
britzl, line 18: Exported service account key, json format, base 64 encoded

.github/workflows/main-ci.yml
```diff
@@ -310,7 +309,7 @@ jobs:
 
   sign-editor-windows:
     needs: [build-editor]
-    runs-on: windows-latest
+    runs-on: ubuntu-20.04
```
britzl, line 312: We can run the windows signing process on a Linux machine now that we use jsign

build_tools/run.py
```diff
@@ -32,11 +32,16 @@ def _to_str(x):
     return x
 
 def _exec_command(arg_list, **kwargs):
+    silent = False
```
britzl, line 35: We need to be able to sometimes run a command without printing the output to the console. Example is when printing the authentication token.

scripts/ev_code_signing.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
britzl, line 1: Not used on a day to day basis, but I kept it here so that it is easy to test code signing without the full build system.

build_tools/run.py
AGulev: is it because it prints secrest?

editor/scripts/bundle.py
```diff
@@ -466,13 +491,13 @@ def sign(options):
...
             for lib in find_files(os.path.join(jdk_path, "lib"), "*.dylib"):
-                sign_files('macos', options, lib)
-            sign_files('macos', options, os.path.join(jdk_path, "lib", "jspawnhelper"))
-            sign_files('macos', options, os.path.join(sign_dir, "Defold.app"))
+                sign_file('macos', options, lib)
+            sign_file('macos', options, os.path.join(jdk_path, "lib", "jspawnhelper"))
+            sign_file('macos', options, os.path.join(sign_dir, "Defold.app"))
         elif 'win32' in platform:
-            sign_files('win32', options, os.path.join(sign_dir, "Defold", "Defold.exe"))
+            sign_file('win32', options, os.path.join(sign_dir, "Defold", "Defold.exe"))
```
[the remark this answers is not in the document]
britzl: Maybe. We've never done it before, but perhaps that is something we want to start doing?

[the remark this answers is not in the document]
AGulev: I think it's something to think about as a separte issue. And ofc. only prebuilt exe because we are sure it's safe

build_tools/run.py
britzl: Correct. We start by getting the service account key from a GH secret and activate the service account. This part is fine, but in the next step we call `gcloud auth print-access-token` and capture the output (short lived access token) and use that in the subsequent signing step. We want to capture the token but not print it to stdout.

AGulev:
@britzl Ithought we shouldn't see this window at all
aloso publisher looks strange
IMG_6852

AGulev:
@britzl it seems like you use Root sertificate instead of ours

britzl:
That doesn't look right... Let me check the certificate chain.

ci/gcloud_certfile.cer
```diff
@@ -0,0 +1,44 @@
+-----BEGIN CERTIFICATE-----
```
britzl, line 1: This is the public cert. The private key is in Google Cloud on the HSM.

AGulev (approved):
🔥
That would be nice, to have some kind of instruction or/shortlist in our internal documentation to be able to update the cert or repeat the process in case something will happen

britzl:
> That would be nice, to have some kind of instruction or/shortlist in our internal documentation to be able to update the cert or repeat the process in case something will happen

There's a document on Google Drive named "GlobalSign and EV Cert in the cloud".

Issue #7896, opened 2023-08-18
Scrolling the asset list is slow in the editor


britzl:
**Describe the bug (REQUIRED)**
Scrolling in the asset list has become a bit sluggish recently

**To Reproduce (REQUIRED)**
1. Open a project
2. Scroll the asset list
3. Notice that it is a bit sluggish

**Expected behavior (REQUIRED)**
It should not be slow to scroll the list regardless of the amount of entries

**Defold version (REQUIRED):**
  - Version 1.5.0

**Platforms (REQUIRED):**
 - Platforms: All

**Logs (OPTIONAL):**
From @vlaaad:

JavaFX setGraphic and setText are hot spots when scrolling. Looks like a JavaFX performance issue.

![image](https://github.com/defold/defold/assets/1300688/5ab6d9ec-44ff-441c-b3db-8c57a89ad3b6)

* I’ve been trying various javafx versions, it seems the problem starts with javafx 21-ea+21 , released Jun 12, 2023
* previous version, 21-ea+17, was released on May 13, 2023
* in the commits between those dates there are a couple of commits that touch VirtualFlow used in TreeView:
    * [1a0f6c7f5e131d0a40073ccfe21bb6dd0d6b2da3](https://github.com/openjdk/jfx/commit/1a0f6c7f5e131d0a40073ccfe21bb6dd0d6b2da3#diff- 
   * 7481b3b029b6b471ea64e4d51527257519ef1bd61a437b4047457bb2af6790fb)
[9ad0e908088373aaa7220fc98ef5625d8f9ff6a4](https://github.com/openjdk/jfx/commit/9ad0e908088373aaa7220fc98ef5625d8f9ff6a4#diff-7481b3b029b6b471ea64e4d51527257519ef1bd61a437b4047457bb2af6790fb)
* ironically, the second commit is about “potential performance improvements” in VirtualFlow
* there are VirtualFlow methods in a call stack when I scroll, hinting it might be related
* the methods in the call stack are not the changed methods, but perhaps the changes affect the methods, need to check
* we can probably roll back to javafx 21-ea+17, though I haven’t checked if it contains macos drag-n-drop fix

**Workaround (OPTIONAL):**
Downgrade to JavaFX 21-ea+17

PR #7907, opened 2023-08-22
Improve resource icon styling performance

vlaaad:
Technical notes:
While profiling, I noticed that there is a hotspot in checking if a CSS style applies. Reducing the amount of style to check against helped with the hotspot.

Related to #7896

### DIFF
--- a/editor/src/clj/editor/resource.clj
+++ b/editor/src/clj/editor/resource.clj
@@ -433,20 +433,62 @@
         (io/copy in f))
       (.getAbsolutePath f))))
 
+(def ^:private ext->style-class-suffix
+  {;; Script files
+   "fp" "script"
+   "gui_script" "script"
+   "lua" "script"
+   "render_script" "script"
+   "script" "script"
+   "vp" "script"
+   "glsl" "script"
+
+   ;; Design files
+   "atlas" "design"
+   "collection" "design"
+   "collisionobject" "design"
+   "cubemap" "design"
+   "dae" "design"
+   "font" "design"
+   "go" "design"
+   "gui" "design"
+   "label" "design"
+   "model" "design"
+   "particlefx" "design"
+   "spinemodel" "design"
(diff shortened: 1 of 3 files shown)


editor/src/clj/editor/resource.clj
```diff
@@ -433,20 +433,62 @@
...
+   "display_profiles" "property"
+   "factory" "property"
+   "gamepads" "property"
+   "input_binding" "property"
+   "material" "property"
+   "project" "property"
+   "render" "property"
+   "sound" "property"
+   "texture_profiles" "property"})
+
```
matgis, line 478: Wouldn't it be nicer to specify this as
```
{"design" ["atlas"
           "collection"
           "collisionobject"
           ...]
 "property" ["animationset"
             "camera"
             "collectionfactory"
             ...]
 ...}
```
and programmatically map that into this structure?

We should also consider applying the `"resource-ext-"` prefix to the values here to avoid creating garbage strings for each resource later.

editor/styling/stylesheets/_dialogs.scss
```diff
@@ -295,28 +295,14 @@ $spacing-largest: 40px;
         @include icon-extension-template(-df-folder, -df-folder-active)
     }
 
-    &.resource-ext- {
-        /* Script files */
-        &fp, &gui_script, &lua,
-        &render_script, &script, &vp {
+    &.resource-ext {
+        &-script {
```
matgis, line 299: It's not really a `resource-ext-` anymore. Should we rename the prefix to `resource-kind-` or something like that?

matgis (approved):
Nice find! I think it could benefit from some changes, but I'm fine with merging it like this.

editor/styling/stylesheets/_dialogs.scss
vlaaad: Yep!

PR #7908, opened 2023-08-22
Property editor fixes

matgis:
### User-facing changes
* Fixed some situations where the Property Editor would take longer to redraw after editing a property.
* Fixed a situation where a property could be reverted to its previous value shortly after an edit.
* Fixed a situation where duplicate undo steps could be created after committing a property change with the Enter key.

### Technical changes
* Property editor controls are no longer re-created after an edit if there is a property with an `:edit-type` that has a `:clear-fn` present.
* The `on-edit!` callback is no longer called when the value of an `Editable` control is set programmatically.
* The `auto-commit` callback is no longer called after focus leaves a text field whose edits have already been committed using the Enter key.


vlaaad (approved):
LGTM!

PR #7909, opened 2023-08-22
Added the ability to redirect urls of published artifacts to move_s3_release.py

matgis:
This PR adds a new `redirect_release` command to the `move_s3_release.py` script. It takes the same arguments as the existing `move_release` command, but only updates the archive redirects so that they point to the requested channel.

This is useful if a build non-persisted build has been triggered from the same commit hash as a build from the `stable` channel, thereby potentially overwriting the redirects so that it points to the temporary build artifacts.

We recently ran into this when a `beta` build was triggered from the same commit as the `stable` build. Due to scheduling, the `beta` build completed the Windows engine job after the `stable` Windows engine, redirecting the archive URL to the `beta` artifact, which was then deleted to free up disk space a bit later.

To fix such a situation in the future, you would simply run:
```
./scripts/move_s3_release.py --sha1=<sha1> --to-channel=stable redirect_release
```
And then use the CloudFront dashboard to invalidate the cached redirects. We output a list of what needs to be invalidated after the script completes.

### DIFF
--- a/scripts/move_s3_release.py
+++ b/scripts/move_s3_release.py
@@ -26,42 +26,49 @@ CDN_UPLOAD_URL="s3://d.defold.com/archive"
 if __name__ == '__main__':
     boto_path = os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(__file__)), '../packages/boto-2.28.0-py2.7.egg'))
     sys.path.insert(0, boto_path)
-    usage = '''usage: %prog [options] command(s)
+    usage = '''usage: %prog [options] command
 
 Commands:
-move_release - Move a release on S3 to a new channel
-
-Multiple commands can be specified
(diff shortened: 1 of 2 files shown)


scripts/move_s3_release.py
```diff
@@ -26,42 +26,49 @@
...
     default_archive_path = CDN_UPLOAD_URL
     parser.add_option('--archive-path', dest='archive_path',
                       default = default_archive_path,
                       help = 'S3 path where archives are uploaded')
 
-    options, args = parser.parse_args()
+    options, commands = parser.parse_args()
 
-    if not args:
+    if len(commands) != 1:
```
matgis, line 52: We no longer support multiple command, but I don't think that made sense for the script earlier either?

britzl (approved):

Issue #7781, opened 2023-07-07
arm64-macos support for the Editor


AGulev:
**Task (REQUIRED):**

- [x] Make it possible to run the editor using JDK arm64 macOS
- [x] Bundle a separate editor for arm64 macOS so user can pick needed version to download

vlaaad:
Site PR: https://github.com/defold/defold.github.io/pull/74

vlaaad:
Done!

PR #7914, opened 2023-08-24
Make editor load natives only for its platform

vlaaad:
Technical notes:
With both x64 and arm64 macOS platforms, the editor needs to differentiate between native libraries it might need to load. It can't blindly load all `*.dylib` files anymore.

Related to #7781


AGulev:
@vlaaad please merge it into beta as well, because I'm not sure if new spine with arm64-osx will work on intel mac

vlaaad:
@AGulev Yes, I intend to merge it into dev and beta

matgis (approved):
Looks good to me!

PR #7916, opened 2023-08-24
Calling go.get on a non-array constant property should give a single value

Jhonnyg:


engine/gameobject/src/gameobject/gameobject_script.cpp
```diff
@@ -762,7 +762,7 @@ namespace dmGameObject
         dmGameObject::PropertyDesc property_desc;
         dmGameObject::PropertyResult result = dmGameObject::GetProperty(target_instance, target.m_Fragment, property_id, property_options, property_desc);
 
-        if (result == dmGameObject::PROPERTY_RESULT_OK && !index_requested && property_desc.m_ValueType == dmGameObject::PROP_VALUE_ARRAY && property_desc.m_ArrayLength > 0)
+        if (result == dmGameObject::PROPERTY_RESULT_OK && !index_requested && property_desc.m_ValueType == dmGameObject::PROP_VALUE_ARRAY && property_desc.m_ArrayLength > 1)
```
Jhonnyg, line 765: Main cause of the issue. We treat a constant with one element as an array with one element, but to avoid regressions we return all constants / arrays with one element as a single element. Meaning, if you have specified a constant with a single value (e.g uniform my_vec[1]) calling a go.get(.., "my_vec") will yield a single vmath.vector4 and not an entire table.

We _could_ potentially return it as a table with a single value, but that requires a fair amount of refactoring and I'm not sure it's worth the effort right now? A one-element array is a bit weird, the only use case I can think of is if you have an array size that is set from a macro:

```
#define LIGHT_COUNT 1
uniform light_data[LIGHT_COUNT];
```

AGulev: I think it's good enough for now and that's how it was before (I added hash tables and removed array lengths at all because we didn't use is).

Maybe it worth to add a task for it, just to track it and maybe make refactoring in the future.

AGulev (approved):

PR #7932, opened 2023-08-30
Slightly improved editor scene view interactions for Guis with a lot of templates

matgis:
### User-facing changes
* Slightly improved performance of scene view interactions for Guis that include a lot of templates.
* Resource name clashes among templates referenced from a Gui will no longer highlight the root node of the referencing Gui with an error icon.

### Technical changes
* Eliminated the data dependency from referenced template scene `build-targets` and `GuiSceneNode.node-outline`.
* Support disabling `*check-schemas*` post compile-time. Used in some performance tests.
* Added development helpers for interacting with the YourKit Java profiler.


editor/src/clj/internal/node.clj
```diff
@@ -1507,7 +1507,8 @@
 (defn- schema-check-result-form [description label node-id-sym label-sym evaluation-context-sym result-sym forms]
   (if *check-schemas*
     `(do
-       (schema-check-result ~node-id-sym ~label-sym ~evaluation-context-sym ~(deduce-output-type-form description label) ~result-sym)
+       (when ~`*check-schemas* ; Inner check to support disabling the schema check post compile-time.
+         (schema-check-result ~node-id-sym ~label-sym ~evaluation-context-sym ~(deduce-output-type-form description label) ~result-sym))
```
vlaaad, line 1511: Hmm, this change worries me, is it necessary? This code will be injected into every output, increasing the code size and potentially affecting jit. I also don't see `*check-schemas*` being rebound anywhere...

editor/sidecar/aot.clj
```diff
@@ -22,7 +22,8 @@
 
 (def excluded-sources-by-build-type
   {"release" ["src/clj/dev.clj"
-              "src/clj/prof.clj"]})
+              "src/clj/prof.clj"
+              "src/clj/yourkit.clj"]})
```
vlaaad, line 26: Hmm, I don't like that we keep dev-time code in the source for production code and then exclude it where necessary. I think a simpler approach would be to keep it separately, e.g. in `src/yourkit`, and then include it on the classpath during the development. This is done e.g. for Reveal with `:reveal` profile in `project.clj`.

vlaaad (commented):
GUI changes look good, but I don't understand the intention of the `*check-schemas*` change.

I also think we should not keep the dev-time code in `src/clj`, maybe you could move it out?

editor/sidecar/aot.clj
matgis: Makes sense. I'll move these files to `src/dev` and add it as a `source-path` in the `:dev` profile.

editor/src/clj/internal/node.clj
matgis: I'm merely wrapping the existing call in `(when *check-schemas* ...)`. Do you think it will have much of an impact? This is called at compile-time to generate the forms, so the additional `(when *check-schemas* ...)` check only applies to `:dev` builds that have `*check-schemas*` enabled. If `*check-schemas*` is `false` at compile time, which it is in `:release` builds, the `schema-check-result-form` function returns no forms at all.

`*check-schemas*` is being rebound here in `perf_test.clj`: https://github.com/defold/defold/blob/editor-dev/editor/test/integration/perf_test.clj#L47

vlaaad (approved):
LGTM!

PR #7934, opened 2023-08-30
Less aggressive rounding of numeric values in the editor

### LINKED ISSUE #6414: Editor rounds values when moving objects
The editor rounds values when moving objects. 

**To Reproduce (REQUIRED)**

https://user-images.githubusercontent.com/193258/154264998-9bf5deec-2e9a-4efb-b978-4c2500fbe067.mp4

The editor should not affect the values after the decimal point, i.e.:
- `0.454545 + 1` should be `1.454545`, but not `1.455`.
- `0.454545 - 1` should be `-0.545455`, but not `-0.545`.

  - Version 1.2.192

**Platforms (REQUIRED):**
 - OS: Windows 10

matgis:
Fixes #6414.

### User-facing changes
* Less aggressive rounding of numeric values in the editor.

### DIFF
--- a/editor/src/clj/editor/field_expression.clj
+++ b/editor/src/clj/editor/field_expression.clj
@@ -45,7 +45,7 @@
 
 (defn to-double [s]
  (try
-   (evaluate-expression #(Double/parseDouble %) 0.01 s)
+   (evaluate-expression #(Double/parseDouble %) math/precision-general s)
    (catch Throwable _
      nil)))
 
--- a/editor/src/clj/editor/gui.clj
+++ b/editor/src/clj/editor/gui.clj
@@ -336,7 +336,7 @@
   (-> (doto (Quat4d.)
         (math/clj->vecmath clj-quat))
       math/quat->euler
-      properties/round-vec
+      properties/round-vec-coarse
       (conj 1.0)))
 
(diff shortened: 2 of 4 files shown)


editor/src/clj/editor/field_expression.clj
```diff
@@ -45,7 +45,7 @@
 
 (defn to-double [s]
  (try
-   (evaluate-expression #(Double/parseDouble %) 0.01 s)
+   (evaluate-expression #(Double/parseDouble %) math/precision-general s)
```
matgis, line 48: Use the same rounding as everything else for the results of expressions typed into numeric fields.

editor/src/clj/editor/gui.clj
```diff
@@ -336,7 +336,7 @@
   (-> (doto (Quat4d.)
         (math/clj->vecmath clj-quat))
       math/quat->euler
-      properties/round-vec
+      properties/round-vec-coarse
```
matgis, line 339: Use coarse rounding when converting quaternions to euler angles for better value stability.

Prior to this PR, we used coarse rounding (three significant decimal places) everywhere. With this change, we use moderate rounding (six significant decimal spaces) for everything but the quaternion to euler angle conversion.

vlaaad (approved):
LGTM!

PR #7940, opened 2023-09-01
Fix native extension test on arm64-macos

vlaaad:
Related to #7781


editor/test/integration/engine/native_extensions_test.clj
```diff
@@ -48,17 +49,17 @@
      (let [workspace (test-util/setup-workspace! world "test/resources/extension_project")
```
matgis, line 49: We should change this to use `setup-scratch-workspace!` instead so we don't get false positives from the test when running locally.

matgis (changes requested):
We should ensure we don't get false positives from the test when running locally by using a scratch workspace.

matgis (approved):
Ship it!

PR #7946, opened 2023-09-01
Made editor build tests significantly faster

matgis:
### User-facing changes
* The editor will give up if it is unable to establish a connection to a referenced library URL after 2 seconds.

### Technical changes
* `integration.build-test` run time decreased by 2m28s.
* Decreased connection timeout when fetching libraries.
* Removed invalid library dependencies from build test projects.
* Updated `.idea/.gitignore` and deleted `.idea/compiler.xml` from source control. This file is managed by Cursive and was showing changes from dependencies in the `:user` profile.


editor/test/resources/buildtime_conversion/game.project
```diff
@@ -1,6 +1,5 @@
 [project]
 title = Game Project Properties
-dependencies = http://test.com/test.zip
```
matgis, line 3: Any invalid dependencies were stalling for 75 seconds each until they finally gave up and threw a `SocketTimeoutException`.

editor/test/resources/game_project_properties/game.project
```diff
@@ -1,6 +1,6 @@
 [project]
 title = Game Project Properties
-dependencies = http://test.com/test.zip
+dependencies = http://test.com/not-responding.zip
```
matgis, line 3: We have a test that verifies that non-responding dependencies are not included in the build output, so we want to keep a URL here that can be resolved but will eventually time out. There doesn't appear to be an "official" test URL that guarantees this behavior, but I changed the zip file name for clarity.

vlaaad (approved):
LGTM!

PR #7955, opened 2023-09-04
Fixed typos in annotations of lua interfaces

astrochili:
Fixed typos in cpp files with base documantation for lua api. It's also required to generate more valid lua annotations.

### DIFF
--- a/engine/crash/src/script_crash.cpp
+++ b/engine/crash/src/script_crash.cpp
@@ -84,7 +84,7 @@ namespace dmCrash
      * load, so loading is one-shot.
      *
      * @name crash.load_previous
-     * @return handle [type:number] handle to the loaded dump, or nil if no dump was found
+     * @return handle [type:number|nil] handle to the loaded dump, or `nil` if no dump was found
      */
     static int Crash_LoadPrevious(lua_State* L)
     {
@@ -221,7 +221,7 @@ namespace dmCrash
      * @name crash.get_sys_field
      * @param handle [type:number] crash dump handle
      * @param index [type:number] system field enum. Must be less than [ref:crash.SYSFIELD_MAX]
-     * @return value [type:string] value recorded in the crash dump, or nil if it didn't exist
+     * @return value [type:string|nil] value recorded in the crash dump, or `nil` if it didn't exist
      */
     static int Crash_GetSysField(lua_State* L)
     {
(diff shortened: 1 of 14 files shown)


engine/crash/src/script_crash.cpp
```diff
@@ -84,7 +84,7 @@ namespace dmCrash
      * load, so loading is one-shot.
      *
      * @name crash.load_previous
-     * @return handle [type:number] handle to the loaded dump, or nil if no dump was found
+     * @return [handle] [type:number] handle to the loaded dump, or `nil` if no dump was found
```
britzl, line 87: I sometimes see our documentation use [] on the return value name, and sometimes we don't. What's the difference?

britzl (commented):
Question about return value syntax

engine/crash/src/script_crash.cpp
astrochili: As I understand from the documentation examples, as well as from the Lua annotation generation algorithms that exists today, `[]` means that the value is optional and can be `nil`. I put brackets where it is explicitly written in the description of the return value that `nil` can be returned.

britzl: It is true that [] is used for @param when the parameter is optional. But I don't think we should use it for return values, even if the return value sometimes can be nil.

astrochili: Hm, but why not? If we designate the return value as optional, the [Lua Language Server](https://github.com/defold/lua-language-server) will also be able to work more correctly and accept the return value as optional.

Here are examples already:
- https://github.com/defold/defold/blob/8f40197b0cf36995ac4b664db0b73943417c437d/engine/script/src/luasocket/luasocket.doc_h#L222-L224
- https://github.com/defold/defold/blob/8f40197b0cf36995ac4b664db0b73943417c437d/engine/gui/src/gui_script.cpp#L5125
- https://github.com/defold/defold/blob/8f40197b0cf36995ac4b664db0b73943417c437d/engine/gameobject/src/gameobject/gameobject_script.cpp#L2954C10-L2954C10

astrochili: I just haven't figured out what this format is for yet, except for generating documentation for the webite and for tooltips in the editor.

astrochili: Another option, if prefferable, I can remove `[]` from `[consume]` and `[retD+1]`/`[retD+2]`/`[retN]` and add an additional return type `[...|nil]`

britzl: We've discussed this a bit more, and our standpoint is that return values are not optional in the same sense as function arguments. The engine will always return values, although sometimes these values are nil (an actual Lua type).

So in our mind the correct way would be to add it to the return type definition. Example:

```
@return foobar [type:string|nil] Return foobar or nil if not available
```

astrochili: Roger that! Updated the fork, please check it out.

britzl (approved):
Good job! Thank you!

PR #8038, opened 2023-09-24
Added support for multiple animated nodes in a .gltf file

JCash:
Previously, we assumed that the first model in the scene was the one that was used for skinning.
However, that didn't cover all the use cases, and instead we now render all the models in the scene.

Additionally, we now also list all the animations from a .gltf/.glb file when it's selected as the animation set for a `.model` component. Previously, it always only selected the longest animation in the file.
The previous behavior for `.animationset`'s is not changed, any animation files used in an animation set will only pick one animation (the longest), just as before.

Also, if there is a mesh in the model, that isn't skinned, but is a child to a bone, then that model will not be skinned at runtime, but get assigned a non-skinned material (if it didn't already have one)

Added `model.set_mesh_enabled()` and `model.get_mesh_enabled()` to control visibility of meshes within a scene.

## PR checklist

* [x] Code
* [ ] Documentation


engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -651,6 +667,25 @@ namespace dmGameSystem
         }
     }
 
+    // static void OutputVector4(const dmVMath::Vector4& v)
```
JCash, line 670: for debugging every now and then.

Jhonnyg (dismissed):
I think it looks good. Do we want a unit test for the case you originally had issues with?

PR #8054, opened 2023-09-27
Added atlas image rename functionality

JCash:
We've added the property `Rename patterns` to the Atlas resource type.
This will facilitate easier coupling of animations between multiple atlases, and is a first step towards multiple texture support for Sprites.

The property is a list of `,` separated patterns, where each pattern is a `=`separated string of the format `<src>=<dst>`.
Examples of Atlas imaged renaming
* Pattern: `_normal=` - An image of the name `hat_normal` -> `hat`.
* Pattern: `hat=cat,_normal=` - An image of the name `hat_normal` -> `cat`.

## PR checklist

* [x] Code
* [x] Documentation


JCash:
Added doc PR:
https://github.com/defold/defold.github.io/pull/88

editor/src/clj/editor/atlas.clj
```diff
@@ -217,7 +217,12 @@
   (inherits outline/OutlineNode)
 
   (property id g/Str
-            (value (g/fnk [maybe-image-resource] (some-> maybe-image-resource resource/proj-path path->id)))
+            (value (g/fnk [maybe-image-resource rename-patterns] 
+                          (let [rename-patterns (if (nil? rename-patterns) "" rename-patterns)
+                                id (some-> maybe-image-resource resource/proj-path path->id)]
+                            (if (nil? id)
+                              ""
+                              (AtlasUtil/replaceStrings rename-patterns id)))))
```
JCash, line 225: At first, I wrote a nice Clojure thing, but then also opted out of it, since I didn't want to have the logic in two places.

editor/src/clj/editor/atlas.clj
```diff
@@ -699,6 +712,7 @@
   (property max-page-size types/Vec2
             (dynamic edit-type (g/constantly {:type types/Vec2 :labels ["W" "H"]}))
             (dynamic error (g/fnk [_node-id max-page-size] (validate-max-page-size _node-id max-page-size))))
+  (property rename-patterns g/Str)
```
JCash, line 715: I didn't want to go down the route of creating a dynamic array right now.
It would be great to have it at some point though.

editor/test/integration/atlas_test.clj
```diff
@@ -47,6 +47,18 @@
...
+    (let [workspace (test-util/setup-scratch-workspace! world "test/resources/image_project")
+          project (test-util/setup-project! workspace)
+          atlas (project/get-resource-node project "/main/rename.atlas")
+          ddf-texture-set (g/node-value atlas :texture-set)
+          animation-ids-in-ddf (mapv :id (:animations ddf-texture-set))]
+      (is (= (set ["ball"
+                   "diamond_dogs"
+                   "test_anim"])
+             (set animation-ids-in-ddf))))))
+
```
matgis, line 61:
```suggestion
(deftest rename-anim
  (test-util/with-loaded-project "test/resources/image_project"
    (let [atlas (project/get-resource-node project "/main/rename.atlas")
          ddf-texture-set (g/node-value atlas :texture-set)
          animation-ids-in-ddf (into #{}
                                     (map :id)
                                     (:animations ddf-texture-set))]
      (is (= #{"ball"
               "diamond_dogs"
               "test_anim"}
             animation-ids-in-ddf)))))

```
* `test-util/with-loaded-project` will set up a `workspace` and a `project` for you. By default it will load the shared test project, but you can specify the desired project path. Scratch workspaces are used when you need to modify the project disk state in your test, but we don't need it here.
* You can map the `animation-ids-in-ddf` directly into a `set` using `(into <target-coll> <transducer> <source-coll>)`. Here `(map :id)` is used without an input collection argument, which returns a [transducer](https://clojure.org/reference/transducers) (kind of a recipe of one or more transformations) instead of a collection.
* You can use the [set literal](https://clojure.org/reference/data_structures#Sets) syntax for the expected result.

editor/src/clj/editor/atlas.clj
matgis: I think it is good that we share the code. A potential improvement would be to only do the tokenization & preparations once in the `AtlasNode` whenever the `rename-patterns` property changes so we don't have to repeat it for every image. But I think we can do that at a later stage.

editor/src/clj/editor/atlas.clj
```diff
@@ -217,7 +217,12 @@
   (inherits outline/OutlineNode)
 
   (property id g/Str
-            (value (g/fnk [maybe-image-resource] (some-> maybe-image-resource resource/proj-path path->id)))
+            (value (g/fnk [maybe-image-resource rename-patterns] 
+                          (let [rename-patterns (if (nil? rename-patterns) "" rename-patterns)
+                                id (some-> maybe-image-resource resource/proj-path path->id)]
+                            (if (nil? id)
+                              ""
```
matgis, line 224: The old code would produce `nil` if the `(some-> maybe-image-resource ...)` expression returned `nil`. Perhaps we shouldn't change it?

editor/src/clj/editor/atlas.clj
```diff
@@ -52,7 +52,7 @@
            [com.dynamo.gamesys.proto AtlasProto$Atlas AtlasProto$AtlasImage]
            [com.dynamo.gamesys.proto TextureSetProto$TextureSet]
            [com.dynamo.gamesys.proto Tile$Playback Tile$SpriteTrimmingMode]
-           [com.dynamo.bob.pipeline ShaderUtil$Common ShaderUtil$VariantTextureArrayFallback]
+           [com.dynamo.bob.pipeline ShaderUtil$Common ShaderUtil$VariantTextureArrayFallback AtlasUtil]
```
matgis, line 55: Let's put `AtlasUtil` at the front of the list. We try to keep these sorted!

engine/gamesys/proto/gamesys/atlas_ddf.proto
```diff
@@ -34,4 +34,6 @@ message Atlas
     optional uint32 inner_padding        = 5 [default = 0];
     optional uint32 max_page_width       = 6 [default = 0];
     optional uint32 max_page_height      = 7 [default = 0];
+    // A list of comma separated patterns to rename animations: E.g. "_nrm=,_normal="
+    optional string rename_patterns      = 8;
```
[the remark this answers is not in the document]
matgis: I think having it as a string in the file format is fine. It should be fairly uncommon to need more than one replacement, but it sure is nice to have the ability!

matgis (changes requested):
Editor changes look mostly fine, although the test does some unnecessary cloning of the project that I think we should avoid.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/AtlasUtil.java
```diff
@@ -169,19 +165,58 @@ private interface PathTransformer {
         String transform(String path);
     }
 
+    private static String[] getReplaceTokens(String pattern) {
+        // "=" -> 0 tokens
+        // "abc=" -> 1 token:  ["abc"]
+        // "=abc" -> 2 tokens: ["", "abc"]
+        // "abc==" -> 1 token: ["abc"]
+        // "abc=def=" -> 2 tokens: ["abc", "def"]
+        String tokens[] = pattern.split("=");
```
britzl, line 174: What happens if the string is malformed?

"hat=cat,foo=,bar"

* In the first case we get token[0] = "hat" and token[1] = "cat" - any occurrence of "hat" is replaced with "cat" 👍 
* In the second case we get token[0] = "foo" and token[1] = "" - any occurence of "foo" is replaced with "" 👍 
* In the third case we get token[0] = "" and token[1] = "" - will have no effect 👍 (it will actually early out on line 186, although I don't think it is required)

I guess there is no real problem here, but maybe we want to check that the string contains an equal character?

britzl (dismissed):

matgis (commented):
I noticed patterns that do not include `=` will be replaced with empty strings. E.g. `_red,_button` will strip both `_red` and `_button` from the names.

This seems fine to me, but if we don't want this behavior we should add some kind of validation to the property since it will end up in the project files.

matgis (dismissed):
Looks good to me!

matgis (dismissed):
LGTM!

Issue #7699, opened 2023-05-30
LSP: add autocompletion support


vlaaad:
**Task (REQUIRED):**
We should support completions from language servers.

**Expected outcome (REQUIRED):**
The editor shows suggestions both from language servers and built-in completions

AGulev:
related https://github.com/defold/defold/issues/6278

AGulev:
related https://github.com/defold/defold/issues/5480

AGulev:
related https://github.com/defold/defold/issues/4073 ?

PR #8057, opened 2023-09-28
Improve script_api documentation popup

vlaaad:
User-facing changes:
The editor documentation popup now shows args, returns, and examples from `*.script_api` files.

Related to #7699


matgis (approved):
LGTM!

PR #8083, opened 2023-10-04
Support uniform buffers for Vulkan

Jhonnyg:
This PR adds:
* Support for uniform buffers for the Vulkan backend. 
* If a shader has specified a #version it will not be transformed from ES2 to ES3 (SPIR-V only)

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderUtil.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderUtil.java
@@ -42,6 +42,7 @@ public class ShaderUtil {
         public static final int     MAX_ARRAY_SAMPLERS             = 8;
         public static final String  glSampler2DArrayRegex          = "(.+)sampler2DArray\\s+(\\w+);";
         public static final Pattern regexUniformKeywordPattern     = Pattern.compile("((?<keyword>uniform)\\s+|(?<layout>layout\\s*\\(.*\\n*.*\\)\\s*)\\s+|(?<precision>lowp|mediump|highp)\\s+)*(?<type>\\S+)\\s+(?<identifier>\\S+)\\s*(?<any>.*)\\s*;");
+        public static final Pattern regexUniformBlockBeginKeywordPattern     = Pattern.compile("((?<keyword>uniform)\\s+|(?<layout>layout\\s*\\(.*\\n*.*\\)\\s*)\\s+)*(?<type>\\S+)\\s*");
         public static String        includeDirectiveReplaceBaseStr = "[^\\S\r\n]?\\s*\\#include\\s+(?:<%s>|\"%s\")";
         public static String        includeDirectiveBaseStr        = "^\\s*\\#include\\s+(?:<(?<pathbrackets>[^\"<>|\b]+)>|\"(?<pathquotes>[^\"<>|\b]+)\")\\s*(?://.*)?$";
         public static final Pattern includeDirectivePattern        = Pattern.compile(includeDirectiveBaseStr);
@@ -54,11 +55,11 @@ public class ShaderUtil {
         }
 
         public static String stripComments(String source)
-        {  
+        {
             CharStream stream = CharStreams.fromString(source);
             GLSLLexer lexer = new GLSLLexer(stream);
             CommonTokenStream tokens = new CommonTokenStream(lexer, GLSLLexer.COMMENTS);
-            // Get all tokens from lexer until EOF 
+            // Get all tokens from lexer until EOF
             tokens.fill();
             TokenStreamRewriter rewriter = new TokenStreamRewriter(tokens);
             // Iterate over the tokens and remove comments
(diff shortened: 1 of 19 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderUtil.java
```diff
@@ -39,13 +39,14 @@
...
-        public static String        includeDirectiveReplaceBaseStr = "[^\\S\r\n]?\\s*\\#include\\s+(?:<%s>|\"%s\")";
-        public static String        includeDirectiveBaseStr        = "^\\s*\\#include\\s+(?:<(?<pathbrackets>[^\"<>|\b]+)>|\"(?<pathquotes>[^\"<>|\b]+)\")\\s*(?://.*)?$";
-        public static final Pattern includeDirectivePattern        = Pattern.compile(includeDirectiveBaseStr);
-        public static final Pattern arrayArraySamplerPattern       = Pattern.compile("^\\s*uniform(?<qualifier>.*)sampler2DArray\\s+(?<uniform>\\w+);$");
+        public static final int     MAX_ARRAY_SAMPLERS                   = 8;
+        public static final String  glSampler2DArrayRegex                = "(.+)sampler2DArray\\s+(\\w+);";
+        public static final Pattern regexUniformKeywordPattern           = Pattern.compile("((?<keyword>uniform)\\s+|(?<layout>layout\\s*\\(.*\\n*.*\\)\\s*)\\s+|(?<precision>lowp|mediump|highp)\\s+)*(?<type>\\S+)\\s+(?<identifier>\\S+)\\s*(?<any>.*)\\s*;");
+        public static final Pattern regexUniformBlockBeginKeywordPattern = Pattern.compile("((?<keyword>uniform)\\s+|(?<layout>layout\\s*\\(.*\\n*.*\\)\\s*)\\s+)*(?<type>\\S+)\\s*");
```
Jhonnyg, line 45: This is the addition here. Previously we didn't touch uniform buffers if they existed in the code, but we actually have to detect them and 'fix' them if they are missing a set identifier.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderUtil.java
```diff
@@ -391,6 +393,8 @@ public static enum ShaderType {
         private static final String dmEngineGeneratedRep = "_DMENGINE_GENERATED_";
 
         private static final String glUBRep                     = dmEngineGeneratedRep + "UB_";
+        private static final String glUBRepVs                   = glUBRep + "VS_";
+        private static final String glUBRepFs                   = glUBRep + "FS_";
```
Jhonnyg, line 397: Easier to debug with different names.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -485,14 +485,19 @@ namespace dmGraphics
...
         // The m_Framebuffer construct will be rotated sequentially
         // with the framebuffer objects created per swap chain.
-        RenderTarget* rt           = new RenderTarget(DM_RENDERTARGET_BACKBUFFER_ID);
+
+        RenderTarget* rt = GetAssetFromContainer<RenderTarget>(context->m_AssetHandleContainer, context->m_MainRenderTarget);
+        if (rt == 0x0)
+        {
+            rt                          = new RenderTarget(DM_RENDERTARGET_BACKBUFFER_ID);
+            context->m_MainRenderTarget = StoreAssetInContainer(context->m_AssetHandleContainer, rt, ASSET_TYPE_RENDER_TARGET);
+        }
```
Jhonnyg, line 494: This fixes the resize issue

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -1945,6 +1952,11 @@ namespace dmGraphics
         const uint32_t num_samplers        = program_ptr->m_VertexModule->m_TextureSamplerCount + program_ptr->m_FragmentModule->m_TextureSamplerCount;
         const uint32_t num_descriptors     = num_uniform_buffers + num_samplers;
 
+        if (num_descriptors == 0)
+        {
+            return VK_SUCCESS;
+        }
```
Jhonnyg, line 1958: Fixes an edge case where no uniforms at all are specified in the shaders.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -2572,22 +2617,50 @@ namespace dmGraphics
...
+                        UniformBlockMember& member = module->m_Uniforms[i].m_BlockMembers[j];
+                        *type = ShaderDataTypeToGraphicsType(member.m_Type);
+                        *size = member.m_ElementCount;
+                        return (uint32_t) dmStrlCpy(buffer, member.m_Name, buffer_size);
+                    }
+
+                    search_index++;
+                }
+            }
+        }
```
Jhonnyg, line 2660: The API for querying locations are based on OpenGL where all uniforms are listed as an increasing index. This might change when we have uniform buffer support for OpenGL.

engine/render/src/render/material.cpp
```diff
@@ -477,30 +477,6 @@ namespace dmRender
         return 0xFFFFFFFF;
     }
 
-    uint32_t ApplyTextureAndSampler(dmRender::HRenderContext render_context, dmGraphics::HTexture texture, HSampler sampler, uint8_t unit)
```
Jhonnyg, line 480: This isn't used anywhere, and not available in dmsdk

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderUtil.java
```diff
@@ -510,13 +534,31 @@ public static Result transform(String input, ShaderType shaderType, String targe
...
+                                String type   = uniforBlockBeginMatcher.group("type");
+                                String any    = uniforBlockBeginMatcher.group("any");
+
+                                boolean blockScopeBegin = line.contains("{");
+
+                                if (layout == null) {
+                                    layout = "layout(set=" + layoutSet + ")";
+                                }
+
+                                line = layout + " " + keyword + " " + type + (blockScopeBegin ? "{" : "");
```
Jhonnyg, line 559: The difference between this and the line above is that this handles uniform buffers and the above handles uniforms that are *NOT* in a buffer. The difference is more or less that in this case we don't match the ending ';' character, and we don't care about the identifier (which we might want to do later with pure GLSL).

PR #8085, opened 2023-10-04
Add quick help page

### LINKED ISSUE #5990: IDE: Display quick help if no files are opened
I like a quick help page that VS Code shows when there are no files opened. Defold IDE can display the same quick help to be not so empty and help new users navigate.

See the example below (I copy-pasted it from the VS Code).

image

quangson91:
Added a quick help page when no files are open.

Fixes #5990 

Technical changes:

**Detail**
Add quick-help-view beside the editor-tabs-split.
When there is no tabs -> show the quick help view.

Demo:

https://github.com/defold/defold/assets/6185205/30a1c9a4-25b7-4eec-8541-da97f1698758

---
**UPDATE**

Change logo style

image


editor/src/clj/editor/app_view.clj
```diff
@@ -294,6 +296,81 @@
                                 (get sub-selections-by-resource-node active-resource-node)))
   (output refresh-tab-panes g/Any :cached (g/fnk [^SplitPane editor-tabs-split open-views open-dirty-views]
                                             (let [tab-panes (.getItems editor-tabs-split)]
+
+                                              ;TODO: Extract this logic to another method.
+                                              (let [total-tabs (reduce + (map #(-> ^TabPane % (.getTabs) (.size)) tab-panes))
```
quangson91, line 301: @matgis @britzl 
I really want to extract this code to another function.

I have try create a function (in this file) like this:

```
(defn- update-quick-help-pane [arg]
; implement here
)
```

But I don't know how to call this method from here.
Could you please suggest?

Thx!

editor/src/clj/editor/app_view.clj
```diff
@@ -294,6 +296,81 @@
...
+                                                    is-valid (and quick-help-box box-items)]
+                                                (if is-valid (.setVisible quick-help-box is-empty))
+                                                (if (and is-valid (zero? (.size (.getChildren box-items))))
+                                                ;(if (and is-valid)
+                                                  (do
+                                                    (.setVisible quick-help-box is-empty)
+
+                                                    ;TODO: Do nothing if the box-items has children
+                                                    ;TODO: Move the items to the other place where we can define shortcut for each OS
+                                                    (let [items [{:label "Re-Open Closed File" :keys ["Ctrl" "Shift" "T"]}
```
quangson91, line 314: I also want to extract these config.
Where is the good place to put this one?
I think we also need dynamic config (by platform). Because each platform may have different shortcut.

editor/styling/stylesheets/_editor.scss
```diff
@@ -201,3 +201,11 @@ Links:
 #sponsor-push > .hyperlink:hover {
   -fx-text-fill: #89d2ff;
 }
+
+.key-button {
```
quangson91, line 205: I tried put these config directly in the editor.css.
Then later, I copy the config here.

I would like to know, what is the process when build to generate css from scss.

editor/resources/quick-help-view.fxml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import javafx.geometry.Insets?>
+<?import javafx.scene.image.Image?>
+<?import javafx.scene.image.ImageView?>
+<?import javafx.scene.layout.*?>
+<VBox id="quick-help-box" alignment="CENTER" spacing="10" visible="false" AnchorPane.bottomAnchor="0.0" AnchorPane.leftAnchor="0.0" AnchorPane.rightAnchor="0.0" AnchorPane.topAnchor="0.0" xmlns="http://javafx.com/javafx/17.0.2-ea" xmlns:fx="http://javafx.com/fxml/1">
+    <ImageView fitHeight="150.0" fitWidth="200.0" pickOnBounds="true" preserveRatio="true">
+        <Image url="@logo-ver-outline-dark-320.png" />
```
matgis: I think we should use the SVG version of the logo. Perhaps the white one with some transparency could work? You can get it in SVG format from here: https://defold.com/logo-and-trademark/

editor/src/clj/editor/app_view.clj
matgis: Users can override the keymap using a config file, so we can't hardcode these.

The `AppView` node you're editing has access to the keymap. Add `keymap-config` as an input dependency to the surrounding `g/fnk`. This ensures the view will be updated if the keymap changes when we get around to making a keymap editor in the future.
```
(g/fnk [^SplitPane editor-tabs-split open-views open-dirty-views keymap-config]
  ...)
```
You can then do something like:
```
(let [command->key-combo
      (into {}
            (mapcat (fn [[key-combo command-infos]]
                      (map (fn [command-info]
                             [(:command command-info) key-combo])
                           command-infos)))
            keymap)]
  ;; You can now get the key-combo info for each command, i.e.
  (command->key-combo :reopen-recent-file)
  (command->key-combo :open-asset)
  ...)
```
The key-combo is a map that looks like this:
```
{:key "T",
 :alt-down? false,
 :control-down? false,
 :meta-down? true,
 :shift-down? true,
 :shortcut-down? false}
```
`:shortcut-down?` will be `false`, and should be ignored. On macOS, **Cmd** (`:meta-down?`) is used instead of **Ctrl** for basic shortcuts, but **Ctrl** may be used for some advanced shortcuts. You can assume `:meta-down?` will be `false` on non-mac platforms.

aglitchman:
I thought my issue is one of those that will never be done. Because it is not so essential, but rather for UI beautification.

@quangson91 You are doing very well! 

P.S. I would 2x the size of the logo and increase the line spacing.

editor/src/clj/editor/app_view.clj
quangson91: I used your suggestion. For now, the key-map can be dynamic display by OS.

editor/resources/quick-help-view.fxml
quangson91: Does javafx support display svg natively?

For now, I change to use the Defold logo `logo-ver-outline-white.png`, and it looks pretty good.

matgis: I think it does. Did you try it?

quangson91: Yes, I already tried. It display nothing when we use svg format.

image

matgis (changes requested):
Nice work!

I think it works great. I found a few minor code style issues. I also tweaked the labels, and the list of actions we display. But somehow the logo shows up as opaque white now? I think we wanted it to be more subtly blended, right? It is kind of blinding right now.

I also provided a few Clojure tips & suggestions, but you don't have to do them if you disagree. 😅

quangson91:
@matgis 
Thank you for the feedback. I will fix those issues.

For the logo, I missed update the style in the _editor.scss.

I will update it.

```
#quick-help-logo {
    -fx-opacity: 0.1;
}

```

editor/src/clj/editor/app_view.clj
```diff
@@ -240,6 +242,88 @@
...
       escaped-resource-name)))
 
+(defn- create-key-info [label key-combo]
+  (let [keys (remove nil? (list
+                            (if (:meta-down? key-combo) "Meta")
+                            (if (:control-down? key-combo) "Ctrl")
+                            (if (:alt-down? key-combo) "Alt")
+                            (if (:shift-down? key-combo) "Shift")
+                            (str (:key key-combo))))]
+    {:label label :keys (into [] keys)}))
```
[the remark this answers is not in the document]
quangson91: I am learning Clojure. So really love the detail.

editor/resources/quick-help-view.fxml
matgis: Hmm, yes. It seems it supports SVG shape syntax using the `<SVGPath>` element, but not the full SVG spec. Oh well.

editor/resources/quick-help-view.fxml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import javafx.geometry.Insets?>
+<?import javafx.scene.image.Image?>
+<?import javafx.scene.image.ImageView?>
+<?import javafx.scene.layout.*?>
+<VBox id="quick-help-box" alignment="CENTER" spacing="10" visible="false" AnchorPane.bottomAnchor="0.0" AnchorPane.leftAnchor="0.0" AnchorPane.rightAnchor="0.0" AnchorPane.topAnchor="0.0" xmlns="http://javafx.com/javafx/17.0.2-ea">
+    <ImageView id="quick-help-logo" fitWidth="300.0" pickOnBounds="true" preserveRatio="true">
+        <Image url="@logo-ver-outline-white.png" />
```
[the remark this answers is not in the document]
britzl: Note that there is also a version of the logo for use blended into a dark background:

https://github.com/defold/assets-defold/blob/master/assets-defold/logo/logo-ver-outline-bright.png

editor/src/clj/editor/app_view.clj
```diff
@@ -96,12 +96,14 @@
...
+           [javafx.geometry HPos Orientation Pos]
            [javafx.scene Parent Scene]
-           [javafx.scene.control MenuBar SplitPane Tab TabPane TabPane$TabClosingPolicy TabPane$TabDragPolicy Tooltip]
+           [javafx.scene.text Font]
+           [javafx.scene.paint Color]
+           [javafx.scene.control MenuBar SplitPane Tab TabPane TabPane$TabClosingPolicy TabPane$TabDragPolicy Tooltip Label]
            [javafx.scene.image Image ImageView]
            [javafx.scene.input Clipboard ClipboardContent]
-           [javafx.scene.layout AnchorPane StackPane]
+           [javafx.scene.layout AnchorPane GridPane HBox StackPane VBox Region]
```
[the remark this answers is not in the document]
quangson91: Yup. I removed.

editor/src/clj/editor/app_view.clj
```diff
@@ -240,6 +242,88 @@
...
+                            (str (:key key-combo))))]
+    {:label label :keys (into [] keys)}))
+(defn- update-quick-help-pane [^SplitPane editor-tabs-split keymap]
+  (let [
+        tab-panes (.getItems editor-tabs-split)
+        is-empty (not (some (complement empty?) (map #(-> ^TabPane % (.getTabs)) tab-panes)))
+        quick-help-box (.lookup (.getParent editor-tabs-split) "#quick-help-box")
+        ^GridPane box-items (.lookup (.getParent editor-tabs-split) "#quick-help-items")]
+
+    ;Only make quick help visible when there is no-tabs.
```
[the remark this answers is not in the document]
quangson91: I got it. I am also update in other places.

editor/resources/quick-help-view.fxml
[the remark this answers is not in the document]
quangson91: I updated the _editor.scss

editor/styling/stylesheets/_editor.scss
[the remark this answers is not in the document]
quangson91: Yup. I changed to `#quick-help-items > .key-button`

quangson91:
@matgis 
Thank you for the very detail.
You guys in the Defold team is so kind.

I just update source code by your suggestions.
Please take a looK!

vlaaad:
I have a couple of notes:
For @quangson91:
- First of all, we are very grateful for your contributions ❤️
- Instead of creating strings like `"Cmd"` and `"Ctrl"` in `create-key-info`, rely on JavaFX's [KeyCombination::getDisplayText](https://openjfx.io/javadoc/14/javafx.graphics/javafx/scene/input/KeyCombination.html#getDisplayText()) method. This will give us nicely-looking shortcuts on macOS like <kbd>⌃⇧→</kbd> — that looks the same as in the menu bar.
- We cannot expect any shortcuts to be present — our users can create custom keymaps that don't have shortcuts for commands we want to show. We must check that the shortcut exists before invoking `create-key-info`

For @matgis:
I hope we will migrate the rest of the editor to [cljfx](github.com/cljfx/cljfx) eventually, and that makes me hesitant to introduce more code that uses raw JavaFX, fxml, and mutations of JavaFX view graph. On the other hand, I understand that this is another layer of complexity that might be too much for external contributors. Do you think we are okay with supporting the added raw JavaFX/FXML/mutability code?

matgis:
@vlaaad Yes, I'm fine with maintaining the raw JavaFX calls for this sort of thing. It is a simple view. We can redo it in [cljfx](https://github.com/defold/defold/pull/github.com/cljfx/cljfx) in a future PR.

I think we have to accept the fact that we currently have a mix of raw JavaFX and cljfx views, so contributors will look at what is already there when they contribute.

matgis (changes requested):
@vlaaad has a good point about the possibility of key commands not being bound if the user has a custom keymap.

You could use [keep](https://clojuredocs.org/clojure.core/keep) when you generate the list of `items` to only include commands that have a shortcut assigned. Something like this:

```clj
(keep (fn [[command label]]
        (when-some [key-combo (command->key-combo command)]
          (create-key-info label key-combo)))
      [[:open-asset "Open Asset"]
       [:reopen-recent-file "Re-Open Closed File"]
       ,,,])
```

Also, could you please add a newline between `(defn- update-quick-help-pane ...)` and `(g/defnode AppView ...)`? 🙂

vlaaad:
`key-combo->display-text` uses the same method of the KeyCombination class that I suggest using directly. You have access to KeyCombination instances in `update-quick-help-pane`: the keymap contains them in a `:key-combo` values:
```clj
(g/node-value (dev/app-view) :keymap)
=> {{:key "Up"
     :alt-down? false
     :control-down? false
     :meta-down? true
     :shift-down? true
     :shortcut-down? false} [{:command :select-beginning-of-file
                              :shortcut "Shift+Meta+Up"
                              :key-combo #object[javafx.scene.input.KeyCodeCombination 0x5d8a6b6f "Shift+Meta+Up"]}]
    ...}
```
In this output it's shown as `#object[javafx.scene.input.KeyCodeCombination 0x5d8a6b6f "Shift+Meta+Up"]`.

You need to modify how you create `command->key-combo` so that it has instances of KeyCombination as values.

quangson91:
@vlaaad  I updated in the last commit: https://github.com/defold/defold/pull/8085/commits/62041b37c99fb361e4fcb3704f9b6d75508f55e9

Could you please take a look?

The small problem is `getDisplayText` itself return the `String`. So it really hard to create a separate keys.
I mean, it returns strings like: "Ctrl+Shift+A", but it would better if we can get list of ("Ctrl", "Shift", "A").
So we can represents in the UI better.

Split by "+" sometimes work on linux/windows. But doesn't work on macOS.

So if you have any suggestion please tell me.

For  now, the UI is look like this one:
image

vlaaad:
Hi @quangson91!

I'm sorry that I asked you for changes so many times, now I see that my suggestion to use `.getDisplayText` made the view worse, and it looked better with how you implemented it earlier, especially on macOS where proper view would additionally require us to use a different font:
Screenshot 2023-10-06 at 09 38 31

Would you kindly revert your last commit? Then I think we will be happy to merge it!

quangson91:
Hi @vlaaad 
I just updated.
The commit is revert & borrow some code from displayText.
That mean, the display key is different between mac & non-mac os.

Here is the result (I just fake is-macos = false to test; but please check on your mac, since the key using the default system-font).

Screenshot_20231006_152925
Screenshot_20231006_152954

If there is thing that need to fix, please tell me.

Thanks!

matgis:
Well done! ❤️

It looks good on my Mac. I'm happy to merge this in its current state.

image

matgis (approved):

vlaaad:
Yes, looks great! Thank you @quangson91! Merge it @matgis 👍

matgis:
Merged! Thanks for contributing to Defold! ❤️

vlaaad:
@matgis We should also cherry-pick it into editor-dev I think

matgis:
I agree! Cherry-picked into `editor-dev`.

mikatuo:
@quangson91 I opened Defold editor recently and could not understand why it looked so much nicer than usual and if the logo with keys was there before. Thank you!

PR #8093, opened 2023-10-05
Add TEXTURE_FORMAT_RGBA32UI support for vulkan

Jhonnyg:
Some extensions require this format to work on Vulkan. It should not be adopted engine-wide yet, but since the texture formats are exposed in dmsdk we need to make that API change as well..

### DIFF
--- a/engine/graphics/src/vulkan/graphics_vulkan.cpp
+++ b/engine/graphics/src/vulkan/graphics_vulkan.cpp
@@ -792,6 +792,7 @@ namespace dmGraphics
                                             TEXTURE_FORMAT_LUMINANCE_ALPHA,
                                             TEXTURE_FORMAT_RGBA,
 
+                                            // Float formats
                                             TEXTURE_FORMAT_RGB16F,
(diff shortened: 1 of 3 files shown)


engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -1469,7 +1473,12 @@ namespace dmGraphics
     {
         assert(size > 0);
         DeviceBuffer* buffer = new DeviceBuffer(VK_BUFFER_USAGE_INDEX_BUFFER_BIT);
-        DeviceBufferUploadHelper((VulkanContext*) context, data, size, 0, (DeviceBuffer*) buffer);
+
+        if (size > 0)
+        {
+            DeviceBufferUploadHelper((VulkanContext*) context, data, size, 0, buffer);
+        }
```
Jhonnyg, line 1480: We get a hard crash if we try to upload a zero buffer. The vertex buffer equivalent function does this as well

PR #8105, opened 2023-10-07
Make suggest code editor autocomplete

### LINKED ISSUE #8016: Improve UX for configuring external code editor

quangson91:
Fixes #8016 

**Implement**
- Define list editor (name + path possible for each OS): https://github.com/defold/defold/pull/8105/files#diff-0da8130ad6435081bda29bc4ebabbf13ab067345a1e312ac96c7ac9c77209ef6R57

- Filter exist path then display in the dropdown box.

At the end of dropdown -> Add "Brower..."; So when user click -> open file picker to select file.
After the file (editor) is selected -> Put that item above "Browser..." and selected that item.

If user "Browser..." file again -> that replace previous "Browser... item" with the new one.

In the dropdown, the first item is "Defold Editor" that mean edit insider Defold editor itself.

**TODO**

Asking for help to add more editor here: https://github.com/defold/defold/pull/8105/files#diff-0da8130ad6435081bda29bc4ebabbf13ab067345a1e312ac96c7ac9c77209ef6R57

**Demo**

https://github.com/defold/defold/assets/6185205/c2d70232-0ea2-49b9-b61d-776cf2921c25

### DIFF
--- a/editor/src/clj/editor/prefs_dialog.clj
+++ b/editor/src/clj/editor/prefs_dialog.clj
@@ -56,18 +56,25 @@
   (filter #(.exists (io/file %)) files))
 
 (defn- suggest-code-editor-macos []
-  (let [files (list "/Applications/Visual Studio Code.app/Contents/MacOS/Electron")]
+  (let [files (list "/Applications/Visual Studio Code.app/Contents/MacOS/Electron"
+                    "/Applications/Sublime Text.app/Contents/SharedSupport/bin/subl"
+                    "/Applications/TextMate.app/Contents/Resources/mate")]
     (filter-exist-files files)))
 
 (defn- suggest-code-editor-win []
   (let [files (list "C:/Program Files/Microsoft VS Code/Code.exe"
-                    "C:/Program Files (x86)/Microsoft VS Code/Code.exe")]
+                    "C:/Program Files (x86)/Microsoft VS Code/Code.exe"
+                    "C:/Program Files/Sublime Text/sublime_text.exe"
+                    "C:/Program Files (x86)/Sublime Text/sublime_text.exe"
+                    "C:/Program Files/Notepad++/notepad++.exe")]
     (filter-exist-files files)))
 
 (defn- suggest-code-editor-linux []
   (let [files (list "/usr/bin/code"
                     "/usr/share/code"
-                    "/snap/code/current")]
+                    "/snap/code/current"
+                    "/usr/bin/subl"
+                    "/snap/sublime-text/current/opt/sublime_text")]
     (filter-exist-files files)))
 
 (defn- suggest-code-editor []
--- a/editor/src/clj/editor/prefs_dialog.clj
+++ b/editor/src/clj/editor/prefs_dialog.clj
@@ -13,15 +13,17 @@
 ;; specific language governing permissions and limitations under the License.
 
 (ns editor.prefs-dialog
-  (:require [editor.engine :as engine]
+  (:require [clojure.java.io :as io]
+            [editor.engine :as engine]
             [editor.engine.native-extensions :as native-extensions]
             [editor.prefs :as prefs]
             [editor.system :as system]
             [editor.ui :as ui])
   (:import [com.defold.control LongField]
-           [javafx.geometry VPos]
+           [com.sun.javafx PlatformUtil]
+           [javafx.geometry Side VPos]
            [javafx.scene Parent Scene]
-           [javafx.scene.control ColorPicker CheckBox ChoiceBox Label TextArea TextField Tab TabPane]
+           [javafx.scene.control ColorPicker CheckBox ChoiceBox Label TextArea TextField Tab TabPane ContextMenu MenuItem]
            [javafx.scene.layout ColumnConstraints GridPane Priority]
            [javafx.scene.input KeyCode KeyEvent]
            [javafx.util StringConverter]))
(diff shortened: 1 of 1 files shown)


britzl:
> * So it would be great to know what is editor that Defold user want to add

Suggestions:

* Sublime Text (Windows, macOS, Linux)
* Notepad++ (Windows)
* Emacs? Vim? Nano?
* TextMate (macOS)

quangson91:
@britzl  I just updated the PR.
Instead of autocomplete typing, now I show the posibility editor, so user can select.
User can also able choose file by using file picker.

Here is the demo
https://github.com/defold/defold/assets/6185205/6b1ecec1-76cc-42f0-bb02-48ac24de61f2

matgis:
I like how it only shows you editors that are installed on your computer!

However, this doesn't quite work how I'd personally like it to. Here are some suggestions:

* The **Open File** and **Open File at Line** fields only really make sense to show if the user has chosen to configure a custom editor. I think instead of **Browse...** (currently "Borrow...") being an entry in the dropdown, I think it should be a **Custom...** entry at the bottom of the dropdown that shows the **Executable** path browse field along with the **Open File** and **Open File at Line** fields. And we would hide all of these custom configuration fields otherwise.
* If we're going to include Defold Editor in the list, we should change the label to say **Code Editor** instead of **Custom Editor**.
* The list of selectable editors in the dropdown should be sorted by name.

I also think it would be cool if `app-view` would call functions in `prefs-dialog.clj` to get the executable command lines instead of reading the preferences directly. That would enable us to keep the preferences for the Custom Editor configuration even if the user switches back and forth between a Custom or a Supported text editor.

You might need to make some changes to the way the preference dialog works to achieve only showing certain preference fields depending on the dropdown setting.

If you feel like really getting your hands dirty, you could opt to re-implement the Preferences dialog using [cljfx](https://github.com/cljfx/cljfx), like we've done for most of the dialogs already in `editor/dialogs.clj`. I appreciate this might be a larger task that might be uncomfortable to tackle as a new contributor though, so I understand completely if you choose to simply modifying the existing JavaFX code.

We will get around to migrating everything over to [cljfx](https://github.com/cljfx/cljfx) eventually, and really some minor additions to the JavaFX-based preference dialog won't make the task any more difficult for us! 😅

astrochili:
> However, this doesn't quite work how I'd personally like it to. Here are some suggestions:

Huh, what a clear match, suggested the same in [this comment](https://github.com/defold/defold/issues/8016#issuecomment-1752548618) 🤓.

matgis:
> The executable path field should be visible for any known editor (without Defold Editor), because it's often impossible to determine the path of the installed editor automatically and correctly.

This makes me a bit uncomfortable. First, it would mean we can't filter the list of external editors based on what is installed. It also suggests we need to maintain preferences for the executable paths for each external editor. I would say for Windows and Mac users it would likely be enough to look for the editors at the default install path and show them, since that will be the same for the vast majority of users, but I'm not sure about the situation on Linux.

Advanced users can always configure a custom editor if they have a non-standard install, but maybe I'm sacrificing configurability for relatively small gains in usability for the common case?

astrochili:
> This makes me a bit uncomfortable. First, it would mean we can't filter the list of external editors based on what is installed. It also suggests we need to maintain preferences for the executable paths for each external editor. I would say for Windows and Mac users it would likely be enough to look for the editors at the default install path and show them, since that will be the same for the vast majority of users, but I'm not sure about the situation on Linux.

Filtering the list of editors is probably a very rash approach, since there is no guarantee that the list will be adequately compiled based on some standard installation paths. Users who install the code editor in a non-standard path will lose the ability to use the pre-defined `Open File` and `Open File at Line` values simply because they installed the code editor on a different disc or in a different folder. Even on macOS, a user can can install the application not just in `/Applciations/` but in `/%username%/Applciations/`.

So if the user installed VS Code on the `D` disc in the `Programs` folder, then why should he select `Custom` and manually configure the parameters (which he doesn't know without opening the manual) when VS Code is already in the list?

> Advanced users can always configure a custom editor if they have a non-standard install, but maybe I'm sacrificing configurability for relatively small gains in usability for the common case?

If the default path is valid and pre-filled in the path field, it's unlikely to reduce usability for the average user.

I'm just not sure that users installing a code editor in a different path from the standard one is a sign of an advanced user. Everyone who uses the Defold engine in one way or another is an advanced user. And at the same time, knowledge about the parameters of magic fields was often not clear to advanced users as well.

matgis:
OK, I'm convinced. Let's make the **Executable Path** field visible and editable for all Code Editors, but hide the **Open File** and **Open File at Line** fields for everything but the Custom configuration.

And we will need to show the full list of available Code Editors in the dropdown regardless of if we find an executable at the default path or not.

The **Executable Path** field should be pre-populated with the default executable path for the selected Code Editor if the user has not customized it.

It would be nice if we applied the `field-error` CSS class to the **Executable Path** field if the path does not exist.

Regarding preferences, I guess we should store the executable path for each Code Editor in the user preferences, if they choose to customize it.

quangson91:
@matgis I think it looks great.
I will continue other PR.
for this PR I will try to learn cljfx and implement it.

But if this issue is urgent, then please implement it, so I can learn from that PR as well.

matgis:
Don't worry. It is not urgent, just a nice feature! 🙂

Also, please feel free to use [cljfx](https://github.com/cljfx/cljfx) or not if you prefer. This is an isolated part of the code base so it is not a big deal if you decide to use JavaFX directly. We can port it to [cljfx](https://github.com/cljfx/cljfx) later, and we're grateful for your contributions regardless! ❤️

quangson91:
@quangson91 I closed this PR because I want to create a new one.
Since it is better to write the new feature in cljfx.

britzl:
Ah, ok, thank you for clarifying it. Will you attempt to do more PRs to the Defold editor?

quangson91:
@britzl Yes, I am.
I would like to do some PRs each month.

britzl:
> @britzl Yes, I am. I would like to do some PRs each month.

I'm happy to hear that! Let us know if you need any help or guidance!

PR #8108, opened 2023-10-07
'default.render_script' refactoring reducing of the memory allocation and using of the Defold best practices.

### LINKED ISSUE #3198: Render script refactoring (DEF-3832)
Would be great to refactor our default render script to avoid memory allocations in update function.
This is not really optimisation, it's just for consistency between Defold idiology and code that we have inside the engine as default.

AGulev:
Now update() allocates 0 bytes and all the allocations and recalculations happens only by receiving external messages.

Fix https://github.com/defold/defold/issues/3198
## PR checklist

* [x] Code
* [ ] Documentation


AGulev:
I found that `default.render_script` allocates 1.171875 Kb per frame in `update()` function, which is 4 218,75Kb in a minute.
New version doesn't allocate anything in update function.

Here is a small test of the render script:
https://github.com/AGulev/test_default_defold_render_script

engine/engine/content/builtins/render/default.render_script
```diff
@@ -3,92 +3,177 @@
...
 end
 
-function init(self)
-    self.tile_pred = render.predicate({"tile"})
-    self.gui_pred = render.predicate({"gui"})
-    self.text_pred = render.predicate({"text"})
-    self.particle_pred = render.predicate({"particle"})
+local function update_camera(camera, state)
+    camera.proj = camera.projection_fn(camera, state)
+    camera.frustum.frustrum = camera.proj * camera.view
```
britzl, line 71: frustrum -> frustum

engine/engine/content/builtins/render/default.render_script
```diff
@@ -97,45 +182,42 @@ function update(self)
     render.set_blend_func(render.BLEND_SRC_ALPHA, render.BLEND_ONE_MINUS_SRC_ALPHA)
     render.disable_state(render.STATE_CULL_FACE)
 
-    render.draw(self.tile_pred, {frustum = frustum})
-    render.draw(self.particle_pred, {frustum = frustum})
+    render.draw(predicates.tile, camera_world.frustum)
+    render.draw(predicates.particle, camera_world.frustum)
```
britzl, line 186: Should we also draw models?

engine/engine/content/builtins/render/default.render_script
```diff
@@ -3,92 +3,177 @@
...
+    for _, predicate_name in pairs(arg) do
+        predicates[predicate_name] = render.predicate({predicate_name})
+    end
+    return predicates
+end
 
-    self.view = vmath.matrix4()
+local function create_camera(state, name, is_main_camera)
+    local camera = {}
+    camera.frustum = {}
```
britzl, line 114: What is the point of having camera.frustum.frustum and not only camera.frustum? Are you thinking that we might want to store other options/settings for the frustum in camera.frustum table?

AGulev: we should send a table {frustum = ...}
so, ofc we can send the whole `camera` with many other fields, but I would prefer to have it isolated

britzl: Ah, yes, ofc, so that we create the table only once. Makes sense!

engine/engine/content/builtins/render/default.render_script
AGulev: Not sure, what do you think? Maybe we can do it as a separate step lately when we decide how we wanna work with camera and render_script modularity

engine/engine/content/builtins/render/default.render_script
AGulev: nice catch! thx!

britzl (approved):
Looks good! Remember to check documentation for the render script. Not sure if we need to change anything, but might be worth checking anyway.

engine/engine/content/builtins/render/default.render_script
aglitchman: > Not sure, what do you think? Maybe we can do it as a separate step lately when we decide how we wanna work with camera and render_script modularity

I think you should draw models as well! So that the standard material for models will work on the standard render script. Plus, it will be possible to simplify 3d template by removing its version of render script from it. This will make the engine much more 3D friendly.

Since the PR was accepted, should I write a feature request?

AGulev: >Since the PR was accepted, should I write a feature request?

yes, pls
It's easier to do one thing at a time, and solve one problem at a time. That's why I didn't want to add it as part of this PR

aglitchman: https://github.com/defold/defold/issues/8158

PR #8110, opened 2023-10-08
Use exp4j to evaluate math string

### LINKED ISSUE #7816: Calculate numeric input in the editor
I'd like to be able to use calculations for numeric input in the editor.

image

So with the image above I'd like to input `1920*3/4` and it evaluate to `1440` when I press enter or focus leaves the input box.

It would be great to work with more complicated calculations like `(100-20)*(4+3/5)`.

**Bonus idea:** It would be good if it remembers the calculation so if I click on the input `1440` from my first example then it changes back to the calculation `1920*3/4`. This seems a lot harder to implement but could be super helpful.

quangson91:
You can now evaluate complex mathematical expressions such as `4 * (2 ^ 16)` in numeric text fields and they will be set to the result.

Fixes #7816.

### Technical changes
* Use `exp4j` to evaluate strings from numeric text fields.

***Demo***
https://github.com/defold/defold/assets/6185205/836a46be-2a9d-4494-92a5-4e82e1931ef7

### DIFF
--- a/editor/src/clj/editor/field_expression.clj
+++ b/editor/src/clj/editor/field_expression.clj
@@ -13,8 +13,7 @@
 ;; specific language governing permissions and limitations under the License.
 
 (ns editor.field-expression
-  (:require [clojure.string :as string]
-            [editor.math :as math])
+  (:require [editor.math :as math])
   (:import [java.math MathContext]
            [java.text DecimalFormat DecimalFormatSymbols]
            [java.util Locale]
@@ -22,46 +21,33 @@
 
 (set! *warn-on-reflection* true)
 
-(defn- parsable-number-format [text]
-  (-> text
-      (string/replace \, \.)))
-
-(defn- evaluate-expression [parse-fn precision text]
-  (if-let [matches (re-find #"(.+?)\s*((?<![Ee])[\+\-*/])\s*(.+)" text)]
-    (let [a (parse-fn (parsable-number-format (matches 1)))
-          b (parse-fn (parsable-number-format (matches 3)))
-          op (resolve (symbol (matches 2)))]
-      (math/round-with-precision (op a b) precision))
-    (parse-fn (parsable-number-format text))))
-
-(defn- to-double-safe [s]
+(defn- evaluate-expression [text]
   (try
-    (let [value (-> (ExpressionBuilder. s) .build .evaluate)]
-      (math/round-with-precision value math/precision-general))
-    (catch Throwable _
-      nil)))
-
-(defn- to-int-legacy [s]
-  (try
-    (int (evaluate-expression #(Integer/parseInt %) 1.0 s))
-    (catch Throwable _
-      nil)))
-
-(defn- to-double-legacy [s]
-  (try
-    (evaluate-expression #(Double/parseDouble %) math/precision-general s)
+    (-> (ExpressionBuilder. text) .build .evaluate)
     (catch Throwable _
       nil)))
 
 (defn to-int [s]
-  (or (to-int-legacy s) (int (to-double-safe s))))
+  (try
+    (Integer/parseInt s)
+    (catch Throwable _
+      (try
+        (-> s evaluate-expression int)
+        (catch Throwable _
+          nil)))))
 
 (defn format-int
   ^String [n]
   (if (nil? n) "" (str n)))
 
 (defn to-double [s]
-  (or (to-double-legacy s) (to-double-safe s)))
+  (try
+    (Double/parseDouble s)
+    (catch Throwable _
+      (try
+        (math/round-with-precision (evaluate-expression s) math/precision-general)
+        (catch Throwable _
+          nil)))))
 
 (def ^:private ^DecimalFormat double-decimal-format
   (doto (DecimalFormat. "0"
--- a/editor/src/clj/editor/field_expression.clj
+++ b/editor/src/clj/editor/field_expression.clj
@@ -17,7 +17,8 @@
             [editor.math :as math])
   (:import (java.math MathContext)
            (java.text DecimalFormat DecimalFormatSymbols)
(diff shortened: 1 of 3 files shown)


editor/src/clj/editor/field_expression.clj
```diff
@@ -17,7 +17,8 @@
             [editor.math :as math])
   (:import (java.math MathContext)
            (java.text DecimalFormat DecimalFormatSymbols)
-           (java.util Locale)))
+           (java.util Locale)
+           [net.objecthunter.exp4j ExpressionBuilder]))
```
matgis, line 21: Style: There's a mix of brackets here. Many times you can use **Code > Optimize Imports** in IntelliJ to ensure the requires and imports are consistent and also sorted, which helps with merge conflicts between branches that each add imports to the same file. Unfortunately IntelliJ will sometimes not understand that something is being used and will strip out a needed import, and `ExpressionBuilder` seems to be a victim of that particular bug. 😞

I'll let you decide if you want to change the `()`-wrapped imports to `[]`, change your added import to `()`, or do nothing about it, but at least I got to tell you about the **Optimized Imports** feature. It really works most of the time and can be a great way to keep the imports organized! 🙂

editor/test/editor/field_expression_test.clj
```diff
@@ -114,7 +114,6 @@
     (is (= 1.1 (to-double "1.1")))
     (is (= -2.0 (to-double "-2")))
     (is (= -2.5 (to-double "-2.5")))
-    (is (= -1.5 (to-double "-1,5")))
```
matgis, line 117: Hmm, yes. I think it is probably fine to not allow commas in separators, so it should be fine to remove this test. 🤔

We added support for this quite some time ago, but I can't remember if we did it on a whim or if our users requested it. I suppose we can silently remove support for commas and await the response from the community. 😅

editor/src/clj/editor/field_expression.clj
```diff
@@ -33,21 +34,33 @@
...
 (defn format-int
   ^String [n]
   (if (nil? n) "" (str n)))
 
 (defn to-double [s]
- (try
-   (evaluate-expression #(Double/parseDouble %) math/precision-general s)
-   (catch Throwable _
-     nil)))
+  (or (to-double-safe s) (to-double-legacy s)))
```
matgis, line 63: I agree we should probably get rid of the legacy way of doing it, but perhaps we should only change the order we attempt this so that we try to call `Double/parseDouble` or `Integer/parseInt` first, and if they throw an exception we fall back on the `ExpressionBuilder`?

I think that would allow us to support the special sentinel values like `NaN` and `Infinity` for doubles. People can do whatever they want in their game projects, so potentially `Infinity` could have a special meaning in their game logic and they need to be able to have it as a value. I think it is fine to not allow them in arithmetic expressions, though.

editor/src/clj/editor/field_expression.clj
```diff
@@ -33,21 +34,33 @@
       (math/round-with-precision (op a b) precision))
     (parse-fn (parsable-number-format text))))
 
-(defn to-int [s]
+(defn- to-double-safe [s]
+  (try
+    (-> (ExpressionBuilder. s) .build .evaluate)
```
matgis, line 39: I think we still want to call `math/round-with-precision` on the resulting value after evaluating the expression. 🤔

This gets rid of the annoying number of decimals you tend to get from the resulting values of double arithmetic.

matgis (changes requested):
This is nice! ❤️

I put some suggestions about how we could still support sentinel values such as `Infinity` and `NaN` outside of expressions.

editor/src/clj/editor/field_expression.clj
quangson91: I updated.

editor/src/clj/editor/field_expression.clj
quangson91: Sure, just added roud-with-precision

editor/src/clj/editor/field_expression.clj
quangson91: Yup, I just use `Optimized Imports` features.

Thank you!

matgis (changes requested):
OK, now it works! Let's clean it up a bit to get rid of the legacy code!

The `to-int` and `to-double` functions will both need to evaluate expressions, but they will use different methods of parsing non-expression values and rounding the result.

We could use a `try`-`catch` expression and call the built-in `parse` methods:
```clj
(defn to-int [s]
  (try
    (Integer/parseInt s)
    (catch Throwable _
      (try
        (-> s evaluate-expression int)
        (catch Throwable _
          nil)))))
```
The `parseInt` method here will throw an exception if the input string cannot be parsed as an integer. We can then try to evaluate it as an expression instead inside the `catch` clause.
But note that we also need to catch exceptions from the expression parser in case of expression syntax errors. In this case we should return `nil`.

If successful, the expression parser will return a `double` value. We need to either round it with `math/round-with-precision` or cast it to an `int`. Note that it is possible the resulting `double` returned from the expression parser cannot fit into an `int`. If so, we should return `nil` from `to-int`.

I would personally repurpose the old `evaluate-expression` function to take a string and evaluate it using the `ExpressionBuilder` so it can be called from both `to-int` and `to-double`, but you can do what you like.

We can get rid of `parsable-number-format`, `to-int-legacy` and `to-double-legacy` and their corresponding `safe` variants. We only need to call `Integer/parseInt` or `Double/parseDouble` on the input string in `to-int` and `to-double`, and use the expression parser if they fail.

quangson91:
@matgis Thank you for the feedback.
It is much cleaner now.

matgis (approved):
Works great! I'm going to update the PR description to prepare it for our release notes script and then merge it.

Thanks for contributing to Defold! ❤️

PR #8112, opened 2023-10-08
Set min/max size for quick help to zero

### LINKED ISSUE #8111: Remove editor console max height that was added in
Editor use to have no max height for the console. Now I can't drag it to the top of the screen. I prefer to be able to see a lot of data at once so this change slightly disrupts workflow. 

Revert the limit the max height for the console.

Add an option to remove it

editor_window

quangson91:
Fixes #8111

Change min/max height of quick-help to zero. So the console log can be resize to maximum height.

**Demo:**

https://github.com/defold/defold/assets/6185205/6e384e0a-a264-4f27-86f0-6a21cb991b85

### DIFF
--- a/editor/resources/quick-help-view.fxml
+++ b/editor/resources/quick-help-view.fxml
@@ -4,7 +4,10 @@
 <?import javafx.scene.image.Image?>
 <?import javafx.scene.image.ImageView?>
 <?import javafx.scene.layout.*?>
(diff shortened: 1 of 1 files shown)


britzl (approved):
Looks good. I noticed that since this is an editor only change it should be made on the `editor-dev` branch and not the `dev` branch, but since the previous addition also was made to `dev` we need to do the same here.

vlaaad:
@britzl this commit should be cherry-picked into `editor-dev` so it reaches our users faster.

vlaaad:
Ah it seems @AGulev already did this.

AGulev:
@vlaaad @britzl I did it yesterday, it is in `editor-dev` now

PR #8142, opened 2023-10-13
Model root name fix when built from editor

JCash:


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ModelUtil.java
```diff
@@ -753,6 +753,9 @@ public static ArrayList<ModelImporter.Bone> loadSkeleton(Scene scene) {
         // get the first skeleton
         ModelImporter.Skin skin = scene.skins[0];
         for (Bone bone : skin.bones) {
+            if (bone.index == 0) {
+                bone.name = "root";
+            }
```
JCash, line 758: I forgot to update this function, which is used by the editor.

PR #8174, opened 2023-10-23
Improved handling of script `init()` errors 

### LINKED ISSUE #7978: Improve handling of script init errors
If a game object has multiple script components and there is an error in the init() of one script the rest of the script init() functions will not run. The update() function of all scripts will still run, even if some scripts never got a chance to initialise.

**To Reproduce (REQUIRED)**
1. Create a game object
2. Add two or more scripts
3. Introduce an error in the first script, for instance `print(a + b)`
4. Add print() calls in init() and update() of all scripts
5. Build and run
6. Notice that init() of scripts after the failing script will not run. Notice that update() is still called for all scripts.

We always say that developers should not rely on order of scripts, but in this case an error in one script will directly impact the other scripts.

I expect one of these things:

1. The init() function of all scrips should run, even if there is an error in one of them
2. No scripts on the game object should execute any lifecycle function if one script fails to init. The init() error of one script should invalidate the entire game object.

I think the least disruptive is to implement 1).

  - 1.5.0

**Platforms (REQUIRED):**
 - All

AGulev:
We've improved handling of script `init()` errors.
For the initial setup
```
-go
--script1.script
----init()
----update()
--script2.script
----init()
----update()
```
if error happens in the `init()` function for the `script1.script` **old behavior** was
- the engine calls `script1.script->update()`
- the engine **doesn't** call `script2.script->init()`
- the engine calls `script2.script->update()`

As a result, the reason of the problem is an error in `script1.script->init()` which we see only once, but `script1.script->update()` and `script2.script->update()` may produce a ton of errors because scripts weren't initialized.

With this improvement, if error happens in the `script1.script->init()` function the engine behaves the following way:
- the engine **doesn't** call `script1.script->update()`
- the engine calls `script2.script->init()` and if this function doesn't have Lua errors:
  - the engine calls `script2.script->update()`

It makes script components more independent.
Also, it prevents situations when errors in `update()` function shadow the real reason of the problem in `init()` function.

Fix https://github.com/defold/defold/issues/7978
## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -1693,11 +1695,11 @@ namespace dmGameObject
                 CreateResult result = component_type->m_InitFunction(params);
                 if (result != CREATE_RESULT_OK)
                 {
-                    return false;
+                    init_result = false;
                 }
             }
         }
-        return true;
+        return init_result;
```
AGulev, line 1702: Make initialization of components independent

engine/gameobject/src/gameobject/comp_script.cpp
```diff
@@ -182,12 +182,11 @@ namespace dmGameObject
         ScriptResult ret = RunScript(GetLuaState(params.m_Context), script_instance->m_Script, SCRIPT_FUNCTION_INIT, script_instance, run_params);
         if (ret == SCRIPT_RESULT_FAILED)
         {
+            script_instance->m_Inited = 0;
```
britzl, line 185: m_Initialized instead?

britzl (commented):
I think the changes are good. It definitely makes for a much more predictable and reasonable behaviour when a script error happens on init!

britzl (dismissed):

engine/gameobject/src/gameobject/gameobject_script.h
```diff
@@ -79,8 +79,9 @@ namespace dmGameObject
         int         m_ContextTableReference;
         uint16_t    m_ComponentIndex;
         HProperties m_Properties;
-        uint16_t    m_Update : 1;
-        uint16_t    m_Padding : 15;
+        uint8_t    m_Update       : 1;
+        uint8_t    m_Initialized  : 1;
+        uint8_t    m_Padding      : 6;
```
JCash, line 84: 👍

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -2190,8 +2192,12 @@ namespace dmGameSystem
 
     static dmGameObject::CreateResult CompGuiAddToUpdate(const dmGameObject::ComponentAddToUpdateParams& params) {
         GuiComponent* gui_component = (GuiComponent*)*params.m_UserData;
-        gui_component->m_AddedToUpdate = true;
-        return dmGameObject::CREATE_RESULT_OK;
+        if (gui_component->m_Initialized) {
+            gui_component->m_AddedToUpdate = true;
+            return dmGameObject::CREATE_RESULT_OK;
+        }
+        gui_component->m_AddedToUpdate = false;
```
JCash, line 2199: Isn't this false already?
I generally want us to avoid initializing the same value multiple times (every little code snippet adds up...).
Same could be asked with the "Initialized = 0" above. Presumably they were already 0?

JCash (changes requested):
LGTM, but could probably avoid setting some values that already have a default.

PR #8178, opened 2023-10-24
Added swift compilation commands to build.yml

britzl:
Added swift compilation commands to build.yml


share/extender/build_input.yml
```diff
@@ -164,13 +164,21 @@ platforms:
...
+    emitSwiftModuleCmd: '{{env.PLATFORMSDK_DIR}}/XcodeDefault{{env.XCODE_VERSION}}.xctoolchain/usr/bin/swiftc         -target arm-apple-darwin19 -module-name {{moduleName}} {{#swiftSourceFiles}}{{{.}}} {{/swiftSourceFiles}} -sdk {{env.SYSROOT}} -swift-version {{swiftVersion}} {{#includes}}-I{{{.}}} {{/includes}} {{#platformIncludes}}-I{{.}} {{/platformIncludes}} {{#ext.includes}}-I{{{.}}} {{/ext.includes}} -DCOCOAPODS -DSWIFT_PACKAGE -enable-bare-slash-regex -enforce-exclusivity=checked -emit-object -emit-module -emit-module-path {{swiftModulePath}}'
+    compileSwiftCmd:    '{{env.PLATFORMSDK_DIR}}/XcodeDefault{{env.XCODE_VERSION}}.xctoolchain/usr/bin/swift-frontend -target arm-apple-darwin19 -module-name {{moduleName}} {{#swiftSourceFiles}}{{{.}}} {{/swiftSourceFiles}} -sdk {{env.SYSROOT}} -swift-version {{swiftVersion}} {{#includes}}-I{{{.}}} {{/includes}} {{#platformIncludes}}-I{{.}} {{/platformIncludes}} {{#ext.includes}}-I{{{.}}} {{/ext.includes}} -DCOCOAPODS -DSWIFT_PACKAGE -enable-bare-slash-regex -enforce-exclusivity=checked -emit-object -import-underlying-module -enable-objc-interop -primary-file {{swiftPrimarySourceFile}} -o{{tgt}}'
+
```
JCash, line 170: I'd like us to keep the effort of keeping common command lines in the "osx" config, and instead extra flags in e.g. "emitSwiftHeaderFlags" in the context.

See how it works with "compileCmd" and "flags"

Same should work with iOS (we just haven't cleaned up the entire build_inut.yml yet)

JCash (changes requested):

share/extender/build_input.yml
britzl: Ok, I've moved and split things up a bit. Does it look ok, or do you also think we should move the commands all the way up to common, even if they don't apply to all platforms?

JCash: We can do that later. But perhaps yes, to keep the code tight.
Possibly, we'll introduce some other mechanic for languages, like a Swift-only context.

JCash (approved):

PR #8181, opened 2023-10-25
Compute shader creation in render and graphics libraries

Jhonnyg:


engine/graphics/src/null/glsl_uniform_parser.cpp
```diff
@@ -164,16 +164,36 @@ namespace dmGraphics
...
+            word_start = line_start;
+            word_end   = line_start;
+
+        #if 0
+            char line_buf[256];
+            size_t line_len = line_end - line_start;
+            memcpy(line_buf, line_start, line_len);
+            line_buf[line_len] = '\0';
+            printf("LINE: %s", line_buf);
+        #endif
```
Jhonnyg, line 186: For debugging.

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -1941,6 +1942,23 @@ static void LogFrameBufferError(GLenum status)
...
+    static HProgram OpenGLNewProgramFromCompute(HContext context, HComputeProgram compute_program)
+    {
+        assert(0 && "Not implemented!");
+        return 0;
+    }
+
+    static void OpenGLDeleteComputeProgram(HComputeProgram prog)
+    {
+        assert(0 && "Not implemented!");
+    }
```
Jhonnyg, line 1960: None of these functions should be called from anywhere in the engine yet. The renderer checks for context support in the compute.cpp code, so for now this is fine.

engine/graphics/src/test/test_graphics.cpp
```diff
@@ -484,6 +484,32 @@ TEST_F(dmGraphicsTest, TestProgram)
     dmGraphics::DeleteFragmentProgram(fp);
 }
 
+TEST_F(dmGraphicsTest, TestComputeProgram)
+{
+    const char* compute_data = ""
+        "#version 430 core\n"
+        "layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;\n"
```
Jhonnyg, line 491: Later on we should probably parse this string to get the work units, but it's not needed right now.

PR #8191, opened 2023-10-29
Create constants with same name once for vulkan

Jhonnyg:
## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/engine/render/src/render/program_utils.cpp
+++ b/engine/render/src/render/program_utils.cpp
@@ -73,7 +73,7 @@ namespace dmRender
         dmVMath::Vector4* default_values = 0;
         uint32_t sampler_index = 0;
 
-        bool program_language_glsl = IsContextLanguageGlsl(dmGraphics::GetShaderProgramLanguage(graphics_context));
+        bool program_language_glsl = IsContextLanguageGlsl(dmGraphics::GetProgramLanguage(program));
 
         for (uint32_t i = 0; i < total_constants_count; ++i)
         {
--- a/engine/render/src/render/program_utils.cpp
+++ b/engine/render/src/render/program_utils.cpp
@@ -53,7 +53,16 @@ namespace dmRender
         *samplers_count_out = samplers_count;
     }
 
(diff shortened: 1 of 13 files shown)


engine/render/src/render/program_utils.cpp
```diff
@@ -82,19 +93,36 @@ namespace dmRender
...
             assert(name_str_length > 0);
 
-            // For uniform arrays, OpenGL returns the name as "uniform[0]",
-            // but we want to identify it as the base name instead.
-            for (int j = 0; j < name_str_length; ++j)
+            if (program_language_glsl)
             {
-                if (buffer[j] == '[')
+                // For uniform arrays, OpenGL returns the name as "uniform[0]",
+                // but we want to identify it as the base name instead.
```
JCash, line 99: Is this something we want handled by the content pipeline?

And, alternatively, we can implement a dmGraphics::GetUniformNameHash()

engine/render/src/render/program_utils.cpp
```diff
@@ -64,6 +73,8 @@ namespace dmRender
         dmVMath::Vector4* default_values = 0;
         uint32_t sampler_index = 0;
 
+        bool program_language_glsl = IsContextLanguageGlsl(dmGraphics::GetShaderProgramLanguage(graphics_context));
```
JCash, line 76: Will there every be a case where we have two different language types at the same time? E.g. will compute always be the same as the vertex/fragment shaders? And same versions?

E.g. rather than have a duplicate function with an extra context, I think it might be better to store the language in the HProgram instead?

JCash (dismissed):
I don't have strong opinions, but it feels like the language could be owned by the program, and not the context.

engine/render/src/render/program_utils.cpp
Jhonnyg: Yes that would be ideal. Right now I don't know that we can generate proper reflection data for all opengl programs, I think I need to do some research.. I would assume SPIR-V would have something for this, but I'm not sure.

engine/render/src/render/program_utils.cpp
Jhonnyg: Yes, that's probably a better idea.. Right now we only support a single shader language per context, but we need to fix that for other types of shaders (i.e compute), and especially since we can specify the version directly in a shader.

PR #8215, opened 2023-11-04
Use 'am' instead of 'monkey' to perform an Android app launch

### LINKED ISSUE #8223: Bundling and launching for Android automatically enables the accelerometer rotation
When bundling and launching an app for Android from the editor, the accelerometer rotation on the target device gets enabled. This issue is related to [monkey](https://stackoverflow.com/questions/56684778).

**To Reproduce (REQUIRED)**
1. Go to Project -> Bundle -> Android Application
2. Check "Launch installed app"
3. Click on "Create Bundle"

The app launches on the device, but you can see that the auto-rotate is enabled on its own.

  - Version: 1.6.1

**Platforms (REQUIRED):**
 - Platforms: Android

**Workaround (OPTIONAL):**
It is possible to obtain the state of the accelerometer rotation before launching and restore it afterwards. As an alternative to the monkey tool, launching through `adb shell am` with the activity name and package name is better to avoid the issue altogether.

Ultimanidev:
User-facing changes:
Android accelerometer rotation (i.e. "Auto rotate") is no longer turned on when launching the bundled Android app.

Fixes #8223

### DIFF
--- a/editor/src/clj/editor/adb.clj
+++ b/editor/src/clj/editor/adb.clj
@@ -97,9 +97,9 @@ If it's already installed, configure its path in the Preferences' Tools pane." {
 
 (defn launch!
   "Perform an application launch and block until the launch is done
-
+  
   Returns nil on success, throws exception on failure
-
+  
   Args:
     adb-path    path to adb command, can be obtained with [[get-adb-path]]
     device      device to install on, can be obtained with [[list-devices!]]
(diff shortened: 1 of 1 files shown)


vlaaad:
Can you please create an issue first that describes a problem? If I understand it correctly, will it solve the problem if we don't use the monkey script to launch the app? It shouldn't be a problem to get the activity name to launch the app...

vlaaad:
You probably should rebase/cherry-pick your commits on top of a branch from editor-dev :D

vlaaad (approved):
Works great!

vlaaad:
Many thanks for the contribution @ultimanidev!

PR #8226, opened 2023-11-09
Open engine log file by double-clicking the error

vlaaad:
User-facing changes:
Double-clicking on the build error "For the full log, see /file/path/log.txt" will now open the error log file.


matgis (approved):
Looks good to me!

PR #8242, opened 2023-11-15
Use the WebGL 1.0 specific constant for the half float format.

### LINKED ISSUE #8241: WebGL 1.0 should use the HALF_FLOAT_OES constant for half float formats
Ref: https://developer.mozilla.org/en-US/docs/Web/API/OES_texture_half_float

To upload half float textures on WebGL 1.0 the renderer should use the HALF_FLOAT_OES (value is 0x8D61) constant as a texture type. Otherwise we will get an error about the invalid texture format. 

We should be able to use half float textures on WebGL 1.0.

  - Version: 1.6.3 `dev` branch

**Platforms (REQUIRED):**
 - Platforms: HTML5

aglitchman:
WebGL 1.0 supports loading half float textures. But it requires the use of HALF_FLOAT_OES constant, which has a different value from the HALF_FLOAT constant from WebGL 2.0 / OpenGL ES 3.0. This fix enables to use half float texture format on WebGL 1.0 devices.

Fixes #8241

## PR checklist

* [ ] Code
* [ ] Documentation

### DIFF
--- a/engine/graphics/src/opengl/graphics_opengl.cpp
+++ b/engine/graphics/src/opengl/graphics_opengl.cpp
@@ -159,7 +159,6 @@
     #if defined GL_ES_VERSION_2_0
         #undef GL_ARRAY_BUFFER_ARB
         #undef GL_ELEMENT_ARRAY_BUFFER_ARB
-        #define GL_HALF_FLOAT_OES 0x8D61
     #endif
 #else
     #error "Platform not supported."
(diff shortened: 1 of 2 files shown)


engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -159,6 +159,7 @@
     #if defined GL_ES_VERSION_2_0
         #undef GL_ARRAY_BUFFER_ARB
         #undef GL_ELEMENT_ARRAY_BUFFER_ARB
+        #define GL_HALF_FLOAT_OES 0x8D61
```
aglitchman, line 162: In fact, for WebGL we should use header files from ES 2.0/3.0, because it is based on it. But someone decided to use files from desktop GL. That's strange. And there is no GL_HALF_FLOAT_OES in desktop GL. So I added this constant manually as #define.

PS If the desktop include above is replaced by include gl2ext.h, for example, the webgl renderer stops working and pulls a huge amount of patching and testing.

aglitchman:
I made a build of the engine myself and tested half float texture on WebGL 1.0 device - works great.

The only thing for WebGL 1.0 is that half float texture must be POT.

engine/graphics/src/opengl/graphics_opengl.cpp
Jhonnyg: I think you should put this into graphics_opengl_defines.h and follow the style of the other defines

Jhonnyg (changes requested):
If this requires POT textures, we should probably output an error if someone is trying to use a wrong combination, otherwise we might just get a black screen and no info on why it is happening

engine/graphics/src/opengl/graphics_opengl.cpp
aglitchman: Okay 🙌🏻 Done.

aglitchman:
> If this requires POT textures, we should probably output an error if someone is trying to use a wrong combination, otherwise we might just get a black screen and no info on why it is happening

Ok 🙌🏻  I think it should be a new issue and a new PR (like "The engine should warn that the texture should be POT").

aglitchman:
Looks to me like everything is ok now and you can do the merge.

PR #8243, opened 2023-11-16
Linux: Add "Create Desktop Entry" command

### LINKED ISSUE #8095: Easier way to install Linux shortcut
Users have to create their own .desktop file on Linux to create a desktop shortcut. We could provide that.

Add an entry to the Help menu to create a .desktop file, just like JetBrains does:
Screenshot from 2023-10-05 15-36-44

* Add to the package an icon image file,  .desktop file and a shell script to setup this for the user. No editor code changes.

Related forum posts:
https://forum.defold.com/t/shortcut-for-linux/6035
https://forum.defold.com/t/start-linux-client-with-a-double-click/3691/5

I would love some guidance to get me started to make this happen

vlaaad:
User-facing changes:
Linux: add "Help" -> "Create Desktop Entry" command that creates a desktop entry so it's easier to launch Defold afterwards.

Technical notes:
The original changes were fine, but I decided to add a couple of changes, namely that there is now a feedback to executing this command — a dialog with result. Also I made sure we run `which xdg-desktop-menu` only once to check if the command is available.
Initially, I also wanted to create a temp `.desktop` file, but it doesn't have a stable name, i.e. the file created is something like `defold-editor123123123.desktop`, which leads to duplicate entries being created on repeated invokations of the command, while using a stable name makes the action idempotent.

Fixes #8095


editor/src/clj/editor/app_view.clj
```diff
@@ -2619,3 +2623,52 @@ If you do not specifically require different script states, consider changing th
   (enabled? [] (disk-availability/available?))
   (run [app-view changes-view prefs workspace project]
        (ensure-exists-and-open-for-editing! shared-editor-settings/project-shared-editor-settings-proj-path app-view changes-view prefs project)))
+
+(defn get-linux-desktop-entry [launcher-path install-dir]
```
ivanquirino, line 2627: This function could also be made private, I guess there won't be used anywhere else

editor/src/clj/editor/app_view.clj
```diff
@@ -2619,3 +2623,52 @@ If you do not specifically require different script states, consider changing th
...
+       "Terminal=false\n"
+       "Type=Application\n"
+       "StartupWMClass=com.defold.editor.Start\n"
+       "Categories=Games;Development;Editor;\n"
+       "StartupNotify=true\n"
+       "Exec=" launcher-path "\n"
+       "Icon=" install-dir "/logo_blue.png\n"))
+
+(def ^:private xdg-desktop-menu-path
+  (delay
```
ivanquirino, line 2640: Nice, haven't thought about that

editor/src/clj/editor/app_view.clj
```diff
@@ -2619,3 +2623,52 @@ If you do not specifically require different script states, consider changing th
...
+(handler/defhandler :create-desktop-entry :global
+  (active? [] (some? @xdg-desktop-menu-path))
+  (enabled? [] (and (system/defold-resourcespath) (system/defold-launcherpath)))
+  (run []
+       (let [xdg-desktop-menu @xdg-desktop-menu-path
+             install-dir (.getCanonicalFile (io/file (system/defold-resourcespath)))
+             launcher-path (.getCanonicalFile (io/file (system/defold-launcherpath)))
+             desktop-entry (get-linux-desktop-entry launcher-path install-dir)
+             desktop-entry-file (io/file install-dir "defold-editor.desktop")]
+         (try
```
ivanquirino, line 2655: Great, also haven't thought about exceptions and the user feedback for this.

ivanquirino (approved):
It was really nice contributing to a tool I use. 😄

matgis (approved):
Looks good to me!

editor/src/clj/editor/app_view.clj
vlaaad: `process/exec!` will throw if the exit code is not zero, we should probably document it.

PR #8248, opened 2023-11-20
Unpack once per editor sha

### LINKED ISSUE #7543: Improve editor assets unpacking
**Task (REQUIRED):**
Currently, we unpack files on every editor start and then delete them on the editor shutdown. We should only unpack once per editor (engine?) SHA, while also making sure to not pollute the hard drive too much.

We generally unpack assets once per editor or engine update.

vlaaad:
User-facing changes:
The editor will now unpack its files only once per editor version.

Technical notes:
We now write the `editor-sha.txt` file with the SHA of the editor when unpacking. We also check if the SHA in the file matches the editor's SHA and don't unpack anything in that case. This allows us to unpack the resources only once per editor version.
Another change is that we now update the lastModifiedTime of the unpack directory every day while the editor is running, so that it won't get deleted by another editor checking for old directories to delete.

Fixes #7543

### DIFF
--- a/editor/src/java/com/defold/libs/ResourceUnpacker.java
+++ b/editor/src/java/com/defold/libs/ResourceUnpacker.java
@@ -19,9 +19,13 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.*;
+import java.nio.file.attribute.FileTime;
+import java.time.Instant;
 import java.util.Date;
 import java.util.Collections;
 import java.util.Iterator;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
 
 import com.defold.editor.Editor;
@@ -41,7 +45,6 @@ public class ResourceUnpacker {
     private static volatile boolean isInitialized = false;
     private static Object lock = new Object();
     private static Logger logger = LoggerFactory.getLogger(ResourceUnpacker.class);
-    private static Path unpackedLibDir;
 
     // unpack dirs should be automatically deleted using a shutdown hook but
     // the shutdown hook will not run if the editor doesn't shut down gracefully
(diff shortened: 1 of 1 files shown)


editor/src/java/com/defold/libs/ResourceUnpacker.java
```diff
@@ -59,13 +62,19 @@ private static void deleteOldUnpackDirs(Path unpackPath) throws IOException {
...
+                    try {
+                        long creationTime = Files.getLastModifiedTime(unpackDir).toMillis();
+                        if (creationTime < oneWeekAgo) {
+                            logger.info("deleting unpack dir {}", unpackDir);
+                            FileUtils.deleteQuietly(unpackDir.toFile());
+                        }
+                    } catch (IOException e) {
+                        logger.warn("Couldn't get lastModifiedTime of {}", unpackDir, e);
+                    }
+                });
```
vlaaad, line 77: mostly a refactoring from Files to Paths API, though there is one useful addition — we no longer delete the unpack path itself.

editor/src/java/com/defold/libs/ResourceUnpacker.java
```diff
@@ -85,15 +94,37 @@ public static void unpackResources() throws IOException, URISyntaxException {
...
+                }
+                if (unpackPath.getParent().startsWith(Editor.getSupportPath())) {
+                    // Prevent from deletion by deleteOldUnpackDirs
+                    Executors.newSingleThreadScheduledExecutor().scheduleAtFixedRate(() -> {
+                        try {
+                            Files.setLastModifiedTime(unpackPath, FileTime.from(Instant.now()));
+                        } catch (IOException e) {
+                            logger.warn("Couldn't set lastModifiedTime for {}", unpackPath, e);
+                        }
+                    }, 0, 1, TimeUnit.DAYS);
```
vlaaad, line 124: I tested this with a 1 second scheduler though :D

matgis (approved):
Looks good to me!

PR #8255, opened 2023-11-23
Custom vertex format support for Models

Jhonnyg:
We have added initial support for custom vertex formats for models. 

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


editor/src/clj/editor/model.clj
```diff
@@ -370,66 +413,69 @@
...
+                                                                           [texture-binding-prop-key
+                                                                            {:node-id material-binding-node-id
+                                                                             :label sampler
+                                                                             :value nil
+                                                                             :type resource/Resource
+                                                                             :edit-type {:type resource/Resource
+                                                                                         :ext (conj image/exts "cubemap")
+                                                                                         :set-fn (fn [_ _ _ new] (create-texture-binding-tx material-binding-node-id sampler new))}}])))))
+                                                                 (sort-by key all-sampler-name+orders))]
+                          (vec (concat combined-material-properties material-attribute-properties)))
```
matgis, line 478: `(into combined-material-properties material-attribute-properties)` works, since `combined-material-properties` is already a vector we can append `material-attribute-properties` `into`.

editor/src/clj/editor/model_scene.clj
```diff
@@ -88,79 +93,84 @@
...
+                    normals)]
+      res)))
+
+(defn- mesh->attribute-data [mesh ^Matrix4d world-transform ^Matrix3f normal-transform vertex-attribute-bytes vertex-space]
+  (let [^ints indices (:indices mesh)
+        ^floats positions (:positions mesh)
+        ^floats texcoords (:texcoord0 mesh)
+        ^floats normals   (:normals mesh)
+        mesh-data-out
+        (reduce (fn [out-data vi]
```
matgis, line 113: You could avoid some boxed math by annotating `^long vi` here.

editor/src/clj/editor/graphics.clj
```diff
@@ -381,3 +406,115 @@
...
+(defn- attribute-data->world-position-v3 [data]
+  (let [local-positions (:position-data data)
+        world-transform (:world-transform data)]
+    (geom/transf-p world-transform local-positions)))
+(defn- attribute-data->world-position-v4 [data]
+  (let [local-positions (:position-data data)
+        world-transform (:world-transform data)]
+    (geom/transf-p4 world-transform local-positions)))
+
+(defn put-attributes [^VertexBuffer vbuf attribute-data-arrays]
```
matgis, line 427: This should probably be named `put-attributes!` since it is doing mutation?

editor/src/clj/editor/model_scene.clj
```diff
@@ -88,79 +93,84 @@
...
+                            (filter #(= (:name %) "position"))
+                            first)]
+    (if position-entry
+      material-attribute-infos
+      (conj material-attribute-infos {:name "position"
+                                      :name-key (vtx/attribute-name->key "position")
+                                      :semantic-type :semantic-type-position
+                                      ;; TODO: This is a workaround because we set the manufactured "position"
+                                      ;;       to coordinate-space-world which means that we will produce wrong vertices
+                                      ;;       based on the vertex space setting on the material..
```
matgis, line 169: Can you add some info about what a proper fix looks like?

editor/src/clj/editor/model_scene.clj
```diff
@@ -415,17 +418,15 @@
...
 
 (defn- update-vb [^GL2 gl ^VertexBuffer vb data]
-  (let [{:keys [mesh world-transform vertex-space scratch]} data
+  (let [{:keys [mesh world-transform vertex-space vertex-attribute-bytes]} data
         world-transform (doto (Matrix4d.) (math/clj->vecmath world-transform))]
     (-> vb
-      (vtx/clear!)
-      (mesh->vb! world-transform vertex-space mesh scratch)
-      (vtx/flip!))))
+      (mesh->vb! world-transform vertex-space vertex-attribute-bytes mesh))))
```
matgis, line 424: Style: We could get rid of the threading macro now that we only have one call.

matgis (changes requested):
Editor code looks mostly good, but I think we may strip vertex attribute overrides from the project files when saving something that refers to a defective material?

JCash (dismissed):

editor/src/clj/editor/graphics.clj
```diff
@@ -332,35 +339,53 @@
...
+                  assumed-attribute-info {:element-count element-count
+                                          :name-key name-key}
+                  property-type (attribute-property-type assumed-attribute-info)]]
+        [(attribute-key->property-key name-key)
+         {:node-id _node-id
+          :value (attribute-value values property-type nil element-count)
+          :label (properties/keyword->name name-key)
+          :type property-type
+          :edit-type (attribute-edit-type assumed-attribute-info property-type)
+          :original-value []}]))))
```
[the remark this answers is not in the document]
vlaaad: Hmm, I wrote this `for`, but I don't quite understand what you suggest here and what would be the benefit of using `keep` over `for` here — in both cases, we create a lazy seq of 2-element tuples. I don't see where we return single elements.

editor/src/clj/editor/model.clj
```diff
@@ -319,16 +345,21 @@
   (property vertex-attribute-overrides g/Any
             (default {})
             (dynamic visible (g/constantly false)))
+  (property vertex-attribute-overrides-value-sources g/Any
+            (default {})
+            (dynamic visible (g/constantly false)))
```
matgis, line 350: Rename to `vertex-attribute-override-value-sources`?

TBH, I think it would be better to refactor our code a bit and add the `value-source` information to the entries in `vertex-attribute-overrides`. That would allow us to generalize the code in `graphics` to retain any orphaned values across all the resource types that make use of vertex attribute overrides.

editor/src/clj/editor/model.clj
```diff
@@ -298,6 +314,16 @@
       texture-binding-infos
       renames)))
 
+(g/defnk produce-attribute-value-sources [material-attribute-infos vertex-attribute-overrides vertex-attribute-overrides-value-sources]
```
matgis, line 317: Rename to match output: `produce-vertex-attribute-value-sources`?

matgis (changes requested):
I think it would be better to refactor our code a bit and add the `value-source` information to the entries in `vertex-attribute-overrides` rather than having them map directly to the double values for each attribute. That would allow us to generalize the code in `graphics` to retain any orphaned values across all the resource types that make use of vertex attribute overrides.

matgis (dismissed):
Seems to work great! Nice work! Should be ready to merge once you've hidden the Normal attribute from the Property Editor.

PR #8259, opened 2023-11-24
Fix memory leak #7685

### LINKED ISSUE #7685: Small memory leak in json_initialize_config
it is leaking 1 Kb for each call
image

afrokick:
Possible fix.

There is a static variable for decoding config, because we can re-use it for decoding.

### DIFF
--- a/engine/script/src/luacjson/lua_cjson.c
+++ b/engine/script/src/luacjson/lua_cjson.c
@@ -1547,6 +1547,9 @@ static int json_decode(lua_State *l)
 // Defold: Expose needed functions to our script_json module
 ////////////////////////////////////////////////////////////
 
+static json_config_t cachedDecodeConfig;
+json_initialize_config(&cachedDecodeConfig);
+
 // The major difference between this function and the internal
 // cjson json_encode function is that we don't use the config
 // object created by cjson. Instead we initialize and create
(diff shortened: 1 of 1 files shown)


britzl:
Nice find! It looks like every time we do an encode that we also initialize a string buffer to hold the encoded string:

https://github.com/defold/defold/blob/dev/engine/script/src/luacjson/lua_cjson.c#L1561

https://github.com/defold/defold/blob/dev/engine/script/src/luacjson/lua_cjson.c#L479-L481

The problem is that we expect the user to delete this string, but it seems like it doesn't happen in this case? 

https://github.com/defold/defold/blob/dev/engine/script/src/luacjson/lua_cjson.c#L959

It is strange though because it looks like we do free it here in the Lua API at least:

https://github.com/defold/defold/blob/dev/engine/script/src/script_json.cpp#L157

Are you using LuaToJson() from C++ and forgot to free the string?

engine/script/src/luacjson/lua_cjson.c
```diff
@@ -1547,19 +1547,20 @@ static int json_decode(lua_State *l)
 // Defold: Expose needed functions to our script_json module
 ////////////////////////////////////////////////////////////
 
+static json_config_t cachedDecodeConfig;
```
aglitchman, line 1550: Code style is different. Variables should have `snake_case` naming.

AGulev:
Please follow the PR checklist. One of the points is a linking with the issue you are solving. It will help with the context of the issue for a reviewer.

* [ ] Code
* [ ] Documentation

PR #8269, opened 2023-11-27
Fix typo in profiler documentation (defold/defold.github.io#104)

Daniel-EST:
Fix a typo in the profiler API documentation related to the available platforms at the function `profiler.get_memory_usage()`

Fixes defold/defold.github.io#104


britzl (approved):
Thanks!

PR #8290, opened 2023-12-05
Bugfix for sprite page indices

Jhonnyg:


engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -705,7 +705,15 @@ namespace dmGameSystem
                 case dmGraphics::VertexAttribute::SEMANTIC_TYPE_PAGE_INDEX:
                 {
                     uint32_t unit = num_page_indices++;
-                    memcpy(write_ptr, &page_indices[unit], info->m_ValueByteSize);
+                    float page_index = (float) page_indices[unit];
+
+                    if (info->m_Attribute->m_DataType != dmGraphics::VertexAttribute::TYPE_FLOAT)
```
JCash, line 710: Do we want the check on only this attribute?

JCash (approved):

engine/gamesys/src/gamesys/components/comp_sprite.cpp
Jhonnyg: For now I'd say yet, I just want to fix the immediate issue. We really need to come back and fix the data conversion soon :o

PR #8293, opened 2023-12-06
Compute shader support, part deux

Jhonnyg:
This is the second part of the compute shader support, this PR includes engine resources for the compute program and shaders needed to build the compute program. At this point we cannot still reference the compute programs anywhere in the engine (which is the next step - to formalize a user-facing API around this work).


engine/gamesys/src/gamesys/gamesys.cpp
```diff
@@ -104,6 +106,8 @@ namespace dmGameSystem
         REGISTER_RESOURCE_TYPE("meshc", graphics_context, ResMeshPreload, ResMeshCreate, 0, ResMeshDestroy, ResMeshRecreate);
         REGISTER_RESOURCE_TYPE("modelc", graphics_context, ResModelPreload, ResModelCreate, 0, ResModelDestroy, ResModelRecreate);
         REGISTER_RESOURCE_TYPE("materialc", render_context, ResMaterialPreload, ResMaterialCreate, 0, ResMaterialDestroy, ResMaterialRecreate);
+        REGISTER_RESOURCE_TYPE("compute_programc", render_context, ResComputeProgramPreload, ResComputeProgramCreate, 0, ResComputeProgramDestroy, ResComputeProgramRecreate);
+        REGISTER_RESOURCE_TYPE("computec", graphics_context, ResComputeShaderPreload, ResComputeShaderCreate, 0, ResComputeShaderDestroy, ResComputeShaderRecreate);
```
Jhonnyg, line 110: Note for reviewer - what should we call these?

A "compute program" is analogous to a fragment/vertex shader and a "compute" is the linked compute program that contains all the resource bind points and is analogous to a material.

engine/graphics/src/dmsdk/graphics/graphics.h
```diff
@@ -396,6 +397,7 @@ namespace dmGraphics
         TYPE_FLOAT_VEC3       = 13,
         TYPE_FLOAT_MAT2       = 14,
         TYPE_FLOAT_MAT3       = 15,
+        TYPE_IMAGE_2D         = 16,
```
Jhonnyg, line 400: An "image" doesn't have a sampler and is one of the most common ways of passing data into a compute shader or as a data storage for writing.

engine/graphics/src/graphics.cpp
```diff
@@ -428,6 +431,30 @@ namespace dmGraphics
...
+            case ShaderDesc::SHADER_TYPE_SAMPLER2D:       return TYPE_SAMPLER_2D;
+            case ShaderDesc::SHADER_TYPE_SAMPLER_CUBE:    return TYPE_SAMPLER_CUBE;
+            case ShaderDesc::SHADER_TYPE_SAMPLER2D_ARRAY: return TYPE_SAMPLER_2D_ARRAY;
+            case ShaderDesc::SHADER_TYPE_IMAGE2D:         return TYPE_IMAGE_2D;
+            default: break;
+        }
+
+        // Not supported
+        return (Type) 0xffffffff;
+    }
```
Jhonnyg, line 456: I'm wondering if we should simply remove all the types from the shaderdesc.. feels like we should expose everything we use via the common dmGraphics::Type enum instead?

engine/graphics/src/graphics_adapter.h
```diff
@@ -100,11 +100,13 @@ namespace dmGraphics
...
     typedef void (*DeleteFragmentProgramFn)(HFragmentProgram prog);
+    typedef ShaderDesc::Language (*GetShaderProgramLanguageFn)(HContext context, ShaderDesc::ShaderClass shader_class);
     typedef ShaderDesc::Language (*GetProgramLanguageFn)(HProgram program);
-    typedef ShaderDesc::Language (*GetShaderProgramLanguageFn)(HContext context);
     typedef void (*EnableProgramFn)(HContext context, HProgram program);
     typedef void (*DisableProgramFn)(HContext context);
-    typedef bool (*ReloadProgramFn)(HContext context, HProgram program, HVertexProgram vert_program, HFragmentProgram frag_program);
+    typedef bool (*ReloadProgramGraphicsFn)(HContext context, HProgram program, HVertexProgram vert_program, HFragmentProgram frag_program);
+    typedef bool (*ReloadComputeProgramFn)(HComputeProgram prog, ShaderDesc::Shader* ddf);
+    typedef bool (*ReloadProgramComputeFn)(HContext context, HProgram program, HComputeProgram compute_program);
```
Jhonnyg, line 109: In all of the graphics adapters there is actually almost no distinction between how we create the shader modules, so we are now adding more and more functions that do the same work under the hood.. maybe we should try to merge some of the API into fewer functions? 

If I could do it the way I want to, I would remove HFragmentProgram, HVertexProgram and HComputeProgram, and just represent them as a "HShader" or "HShaderModule" or something like that.

engine/graphics/src/test/test_app_graphics.cpp
```diff
@@ -0,0 +1,497 @@
...
+        dmGraphics::EnableTexture(engine->m_GraphicsContext, 0, 0, sub_pass_0_color);
+
+        dmGraphics::EnableVertexDeclaration(engine->m_GraphicsContext, m_VertexDeclaration, m_VertexBuffer);
+        dmGraphics::Draw(engine->m_GraphicsContext, dmGraphics::PRIMITIVE_TRIANGLES, 0, 6);
+
+        dmGraphics::SetRenderTarget(engine->m_GraphicsContext, 0, 0);
+    }
+};
+
+struct ComputeTest : ITest
```
JCash, line 348: I added the test_app_vulkan.cpp, so that I could more easily debug stuff on Nintendo Switch.

I think generalising it is good, as that was the goal. (e.g. support OpenGL + PS5 too)
I just want us to not lose targets along the way.
E.g. will this new test app run on Switch when we merge it into the dswitch repo?

engine/graphics/src/test/test_app_graphics_package_assets.py
```diff
@@ -51,6 +60,10 @@ def write_header(header_path, assets):
...
-		[get_buffer_str("vertex_program", "test_app_vulkan.vs", "vert"),
-		 get_buffer_str("fragment_program", "test_app_vulkan.fs", "frag")])
\ No newline at end of file
+		"test_app_graphics_assets.h",
+		[to_glsl("glsl_vertex_program", "test_app_graphics.vs"),
+		 to_glsl("glsl_fragment_program", "test_app_graphics.fs"),
+		 to_glsl("glsl_compute_program", "test_app_graphics.compute"),
+		 to_spirv("spirv_vertex_program", "test_app_graphics.vs", "vert"),
+		 to_spirv("spirv_fragment_program", "test_app_graphics.fs", "frag"),
+		 to_spirv("spirv_compute_program", "test_app_graphics.compute", "compute")])
```
JCash, line 69: Are these assets cross platform?
Can I run them on Nintendo Switch?

engine/graphics/src/wscript
```diff
@@ -48,10 +48,16 @@ def build(bld):
...
 
     if platform_supports_feature(bld.env.PLATFORM, 'opengl', {}):
+
+        opengl_defines = []
+        if platform_supports_feature(bld.env.PLATFORM, 'compute', {}):
+            opengl_defines = ['DM_HAVE_PLATFORM_COMPUTE_SUPPORT']
+
         obj = bld.stlib(features        = 'cxx',
-                        includes        = ['.', 'opengl'] + proto_cpp_includes,
+                        includes        = ['.', 'opengl', 'opengl/win32'] + proto_cpp_includes,
```
JCash, line 57: Why should non-win32 platforms include a win32 folder?
Then it feels like something is a bit off?

JCash (changes requested):
lgtm
Some questions about Nintendo Switch support.
Also a minor fix needed for the checking of compute support

engine/gamesys/src/gamesys/gamesys.cpp
Jhonnyg: ah yes! .cp is good! I'll try to keep everything similar to what we already have with vx/frag programs (even though I'd like to call them shaders since it's common nomenclature, but oh well)

engine/graphics/src/test/test_app_graphics.cpp
Jhonnyg: I think it should, when Vulkan is being linked it will always gets picked first, so we shouldn't need any changes to this (but if we do, I'll make sure it still runs).

The changes I've made just enabled me to test the graphics API while building new features, and as such I'd like to be able to test any API for consitency. The alternative is to make separate API tests, but I think this is a better approach.

engine/graphics/src/test/test_app_graphics_package_assets.py
Jhonnyg: Yes, all of these will be compiled into .spv counterparts, and they should be able to run on any vulkan driver.

engine/graphics/src/graphics.cpp
Jhonnyg: Putting a pin on this, I'll fix it in another PR I think.

PR #8294, opened 2023-12-06
Improved save-data tests

matgis:
Rewrote the save tests related to project data persistence from scratch, including meta-tests that verify that we have non-default values for every property in our project files.

### Technical changes
* Editor plugins that register resource types and GUI node types no longer register duplicate instances, so are much nicer to work with when you reload projects, etc.
* You now get a lot nicer errors from failing save-data tests. There's a value diff, a line-diff with surrounding context, detailed file paths, and friendly error messages.
* We now have full coverage of every property that our project files can hold, and tests in place to ensure we don't degrade our coverage as we add new fields. 
* As a result, you will now fail certain editor tests if you add a field to a `.proto` message definition used anywhere in the project files. At this point, you must either:
  * Give a reason why the field should be ignored by the save-data tests by editing `pb-ignored-fields` in `editor/test/integration/save_data_test.clj`.
  * Open up the `editor/test/resources/save_data_project` in the editor and set the new field to a non-default value in any of the files at root-level of the project.
  * Annotate the field with `[(runtime_only)=true]` in the `.proto` file. Note that this should strictly be reserved for fields that has only ever been part of the compiled binary output. Never do this instead of adding a `:deprecated` or `:unimplemented` ignore rule to `pb-ignored-fields` in `editor/test/integration/save_data_test.clj`, not even as a temporary hack, because we will quickly lose control of changes that need to be covered by migration tests, etc.
* The old save-data test have been ported to the new system and save-data-project. However, the "Literal Equality" test that used to check that certain project files would still be openable in the old editor has been removed. We haven't cared about that for ages.


editor/test/integration/reload_unchanged_test.clj
```diff
@@ -180,10 +147,8 @@
          (save-datas->proj-path-set (project/dirty-save-data project)))))
 
 (deftest keep-existing-nodes-no-false-positives-test
-  (with-clean-system
-    (let [workspace (test-util/setup-scratch-workspace! world "test/resources/reload_unchanged_project")
-          project (test-util/setup-project! workspace)
-          project-graph (g/node-id->graph-id project)
+  (test-util/with-scratch-project project-path
+    (let [project-graph (g/node-id->graph-id project)
```
matgis, line 151: We have `test-util/with-scratch-project` for that!

editor/test/integration/save_data_test.clj
```diff
@@ -0,0 +1,1372 @@
...
+      (check-save-data-disk-equivalence! save-data))))
+
+(deftest save-value-is-equivalent-to-source-value-test
+  "This test is intended to verify that the saved data contains all the same
+  information as the read data. These tests bypass the dirty check in order to
+  verify that the information saved is equivalent to the data loaded. Failures
+  might signal that we've forgotten to read data from the file, or somehow we're
+  not writing all properties to disk."
+  (testing "Saved data should be equivalent to read data."
+    (check-project-save-data-disk-equivalence! project/all-save-data)))
```
matgis, line 1314: This test covers what `save-test/save-all` used to cover.

editor/test/integration/save_data_test.clj
```diff
@@ -0,0 +1,1372 @@
...
+  files, because the embedded components are written as a string literal.
+
+  If this test fails, you need to ensure the loaded data is migrated to the new
+  format by adding a :sanitize-fn when registering your resource type (Example:
+  `collision_object.clj`). Non-embedded components do not have this issue as
+  long as your added protobuf field has a default value. But more drastic file
+  format changes have happened in the past, and you can find other examples of
+  :sanitize-fn usage in non-component resource types."
+  (testing "The project should not have unsaved changes immediately after loading."
+    (check-project-save-data-disk-equivalence! project/dirty-save-data)))
```
matgis, line 1330: This test covers the first part of what `save-test/save-dirty` used to cover.

editor/test/integration/save_data_test.clj
```diff
@@ -0,0 +1,1372 @@
...
+        (clear-cached-save-data!)
+        (doseq [resource checked-resources]
+          (let [proj-path (resource/proj-path resource)]
+            (testing (format "File `%s` should not have unsaved changes after saving." proj-path)
+              (let [node-id (test-util/resource-node project resource)
+                    save-data (g/valid-node-value node-id :save-data)]
+                (is (not (:dirty? save-data))
+                    "Unsaved changes detected after saving.")
+                (when (:dirty? save-data)
+                  (check-save-data-disk-equivalence! save-data))))))))))
```
matgis, line 1372: This test covers the remaining part of what `save-test/save-dirty` used to cover.

editor/test/integration/save_test.clj
```diff
@@ -31,298 +29,28 @@
...
-                            (some? only-in-save-pb-msg))
-                    (println "When comparing" (resource/proj-path resource))
-                    (prn 'disk only-in-disk-pb-msg)
-                    (prn 'save only-in-save-pb-msg)))
-
-                ;; We don't have a read-fn. Compare the strings.
-                (let [disk-string (slurp resource)]
-                  (is (= disk-string save-string)))))))))))
-
-(deftest save-all-literal-equality
```
matgis, line 125: We have no replacement for this in the new save-data tests, because we're no longer interested in our saved files being compatible with the old editor. It used to serve some purpose in terms of covering some additional properties in the project files, but now we have full coverage without it. The content remains in the test project, since other tests interact with it, and we might as well ensure we can still load it. But you shouldn't need to ever update the file format of these legacy files anymore.

editor/test/support/test_support.clj
```diff
@@ -37,19 +37,6 @@
...
-     (valid-node-value node-id label evaluation-context)))
-  ([node-id label evaluation-context]
-   (let [value (g/node-value node-id label evaluation-context)]
-     (if (g/error? value)
-       (throw (ex-info "Evaluation produced an ErrorValue."
-                       {:node-type (g/node-type-kw (:basis evaluation-context) node-id)
-                        :label label
-                        :error-value value}))
-       value))))
-
```
matgis, line 52: Moved to `g/valid-node-value`.

editor/test/util/diff_test.clj
```diff
@@ -0,0 +1,105 @@
...
+            :right-lines ["2"
+                          "3"
+                          ""]
+            :edits [{:type :delete
+                     :left {:begin 0 :end 1}
+                     :right {:begin 0 :end 0}}
+                    {:type :nop
+                     :left {:begin 1 :end 3}
+                     :right {:begin 0 :end 2}}]}
+           (diff/find-edits "1\n2\n3\n" "2\n3\n")))))
```
matgis, line 105: Moved here from `diff-view-test` with some alterations. We cover all the old stuff, though!

vlaaad (approved):
Monumental work, great documentation. Ship it!

editor/test/integration/save_data_test.clj
```diff
@@ -0,0 +1,1372 @@
...
+    (clear-cached-save-data!)
+    (doseq [save-data (project->save-datas project)]
+      (check-save-data-disk-equivalence! save-data))))
+
+(deftest save-value-is-equivalent-to-source-value-test
+  "This test is intended to verify that the saved data contains all the same
+  information as the read data. These tests bypass the dirty check in order to
+  verify that the information saved is equivalent to the data loaded. Failures
+  might signal that we've forgotten to read data from the file, or somehow we're
+  not writing all properties to disk."
```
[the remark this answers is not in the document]
matgis: Ah, you're right! I'll convert them to regular comments!

PR #8298, opened 2023-12-07
Fix for lost material and texture assignments in pasted Model component

matgis:
Fix an issue where pasting an embedded Model component would lose the material and texture bindings.

### Technical changes
* If a node is connected to a `:copied-nodes` input of a copied node, it will be traversed in the copied graph fragment.


vlaaad (approved):
LGTM!

PR #8330, opened 2023-12-18
Load images in runtime as buffers

### LINKED ISSUE #8329: Add option to return buffer from image.load

Jhonnyg:
The `image` script module now has a new function to return a buffer object. The new function is called `image.load_buffer` and can be used to dynamically create texture resources in runtime:

```lua
local img = image.load_buffer(compressed_data)
-- img.buffer now contains an actual buffer object that can be passed into resource.create_texture
local tparams = {
	width  = img.width,
	height = img.height,
	type   = resource.TEXTURE_TYPE_2D,
	format = resource.TEXTURE_FORMAT_RGBA
}
local my_texture_id = resource.create_texture("/my_custom_texture.texturec", tparams, img.buffer)
```

Fixes #8329 

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


engine/gamesys/src/gamesys/scripts/script_image.cpp
```diff
@@ -70,7 +72,15 @@ namespace dmScript
...
-    * @param [premult] [type:boolean] optional flag if alpha should be premultiplied. Defaults to `false`
+    * @param options [type:table] A table containing options about how to load the image. Supported entries:
+    *
+    * `premultiply_alpha`
+    * : [type:boolean] if true, premultiply the alpha channel
+    *
+    * `buffer`
+    * : [type:boolean] if true, the 'buffer' parameter of the returned table will be a buffer object (see [ref:buffer.create] for more info)
+    * that can be used for creating texture or atlas resources.
+    *
```
Jhonnyg, line 83: What should we call this parameter?

britzl: I think `buffer` is fine. Do we want more formats in the future? If that is the case then perhaps `format` is better, with values such as `bytes` (ie string) and `buffer`. But perhaps `format` implies image format more than return value type? Perhaps `type` instead?

engine/gamesys/src/gamesys/scripts/script_image.cpp
```diff
@@ -93,6 +103,24 @@ namespace dmScript
...
+    * local imgurl = "http://www.site.com/image.png"
+    * http.request(imgurl, "GET", function(self, id, response)
+    *         local img = image.load(response.response, { buffer = true })
+    *         local tparams = {
+    *             width  = img.width,
+    *             height = img.height,
+    *             type   = resource.TEXTURE_TYPE_2D,
+    *             format = resource.TEXTURE_FORMAT_RGBA }
+    *
+    *         local my_texture_id = resource.create_texture("/my_custom_texture.texturec", tparams, img.buffer)
```
JCash, line 119: One can also imagine the api allowing a `string`, and we'll create a temporary buffer on the fly.
Which might be handy in the future if someone want to downkload a resource, and add it directlty into the resource system (without "loading" it) (e.g. we have a new api which we're using for the Rive extension, where we simply insert new precompiled resources)

engine/gamesys/src/gamesys/scripts/script_image.cpp
```diff
@@ -102,8 +130,29 @@ namespace dmScript
...
+                lua_getfield(L, -1, "buffer");
+                buffer_output = lua_isnil(L, -1) ? false : lua_toboolean(L, -1);
+                lua_pop(L, 1);
+
+                lua_pop(L, 1);
+            }
+            else
+            {
+                // DEPRECATED
+                premult = lua_toboolean(L, 2);
```
JCash, line 154: Ideally, we should find these places in the code and log a dmLogWarningOnce message about deprecation and the suggested solution. MIght help with migration later on when we want to finally remove it.

engine/gamesys/src/gamesys/scripts/script_image.cpp
JCash: Another api would be to create a buffer directly from a string?

engine/gamesys/src/gamesys/scripts/script_image.cpp
```diff
@@ -12,23 +12,25 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
+#include <assert.h>
```
JCash, line 15: Unused?
And if it _is_ used, we're in scripting domain, where lua_error would be preferred?

engine/gamesys/src/gamesys/scripts/script_image.cpp
JCash: Was an `image.load_buffer()` also considered?

While adding an extra argument to the options table is likely the least invasive change we could make, and also saves code size. But I also think it's a slightly odd api.

And it sets a precedence that we can return different (unrelated) types from the same function, which isn't usually how we code in general.

JCash (dismissed):
I tentatively approve, but I wonder if any other apis were considered?

PR #8344, opened 2023-12-21
Buffered render buffers support

Jhonnyg:
Fixed an issue for certain graphics adapter where vertex and index buffers are overwritten during the same frame.

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


engine/graphics/src/graphics.h
```diff
@@ -696,6 +696,12 @@ namespace dmGraphics
     void ReadPixels(HContext context, void* buffer, uint32_t buffer_size);
 
     uint32_t GetTypeSize(dmGraphics::Type type);
+
+    // Both experimental + tests only:
+    void* MapVertexBuffer(HContext context, HVertexBuffer buffer, BufferAccess access);
+    bool  UnmapVertexBuffer(HContext context, HVertexBuffer buffer);
+    void* MapIndexBuffer(HContext context, HIndexBuffer buffer, BufferAccess access);
+    bool  UnmapIndexBuffer(HContext context, HIndexBuffer buffer);
```
Jhonnyg, line 704: used in render tests now

engine/render/src/render/buffer.cpp
```diff
@@ -0,0 +1,146 @@
...
+            return;
+
+        for (int i = 0; i < buffer->m_Buffers.Size(); ++i)
+        {
+            DeleteRenderBuffer(buffer->m_Buffers[i], buffer->m_Type);
+        }
+        delete buffer;
+    }
+
+    HRenderBuffer AdvanceRenderBuffer(HRenderContext render_context, HBufferedRenderBuffer buffer)
```
JCash, line 76: I think the pattern of NewRenderBuffer() + AdvanceRenderBuffer() is a bit unusua imho.
I'd expect the buffer to either have one item already after New(), or that Advance just skip to the next index.

Perhaps something name it like AddRenderBuffer() ?

engine/render/src/render/buffer.cpp
```diff
@@ -0,0 +1,146 @@
...
+    {
+        if (!buffer)
+            return 0;
+
+        if (buffer->m_Buffers.Empty())
+        {
+            buffer->m_Buffers.OffsetCapacity(1);
+            CreateAndPush(render_context, buffer);
+        }
```
JCash, line 85: If NewRenderBuffer() already added a buffer, we wouldn't have to do this extra step here, which imo makes it harder to read.

Perhaps shift it to make it only only depend on the possibility to grow:
```
HRenderBuffer AddRenderBuffer(...)
{
    if (render_context->m_MultiBufferingRequired)
    {
        if (buffer->m_Buffers.Full())
        {
            buffer->m_Buffers.OffsetCapacity(4); // The pointers are small in comparison, the allocation itself is likely more costly
        }
        CreateAndPush(render_context, buffer);
        buffer->m_BufferIndex++;
    }
    return buffer->m_Buffers[buffer->m_BufferIndex];
}
```

engine/render/src/render/buffer.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+        if (new_buffer_count == n)
+            return;
+
+        for (int i = 0; i < (n - new_buffer_count); ++i)
+        {
+            PopAndDelete(render_context, buffer);
+        }
+
+        buffer->m_Buffers.SetCapacity(new_buffer_count);
```
JCash, line 137: While technically correct, will it lead to constant reallocations?
This buffer doesn't cost much, and if it grew to a certain size during one frame, it is very likely to be the case the next frame as well.

Also, I feel looping backwards is slightly harder to read at a glance.
Perhaps something like this:
```
void TrimBuffer(HRenderContext render_context, HBufferedRenderBuffer buffer)
{
    uint32_t n = buffer->m_Buffers.Size();
    uint32_t new_buffer_count = buffer->m_BufferIndex+1;
    for (int i = new_buffer_count; i < n; ++i)
    {
        DeleteRenderBuffer(buffer->m_Buffers[i], buffer->m_Type);
    }
    buffer->m_Buffers.SetSize(new_buffer_count);
}
```

engine/render/src/render/buffer.cpp
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    void RewindBuffer(HRenderContext render_context, HBufferedRenderBuffer buffer)
+    {
+        if (!buffer)
+            return;
```
JCash, line 143: What is the use case here?
Will the HRenderedBuffer become null in a component, and we still try to render it?
I wonder if it's better to handle it at the component level? E.g. disabling the component world in that case?

JCash (changes requested):
I think the "Advance" pattern/functionality is a bit confusing, and I think it can me made clearer.

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
```diff
@@ -113,7 +113,10 @@ namespace dmGameSystem
         const uint32_t buffer_size     = ctx->m_MaxParticleCount * 6 * default_vx_size;
         world->m_VertexBufferData.SetCapacity(buffer_size);
 
-        world->m_VertexBuffer = dmGraphics::NewVertexBuffer(dmRender::GetGraphicsContext(ctx->m_RenderContext), buffer_size, 0x0, dmGraphics::BUFFER_USAGE_STREAM_DRAW);
+        world->m_VertexBuffer = dmRender::NewBufferedRenderBuffer(ctx->m_RenderContext, dmRender::RENDER_BUFFER_TYPE_VERTEX_BUFFER);
+        dmRender::AdvanceRenderBuffer(ctx->m_RenderContext, world->m_VertexBuffer);
```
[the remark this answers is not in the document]
Jhonnyg: Hm yes that is true, I think this is a casing of how to name things in the API. Right now it's more of a container than a buffer with container-like capabilities, which is probably what it should be.

engine/render/src/render/buffer.cpp
Jhonnyg: Yes, I was running into this in my head as well when developing, I couldn't find a good name for what the function actually did, since I want a separation of which of the internal buffers are "current" and a way of adding new buffers. But AddRenderBuffer is a good choice I think :)

engine/render/src/render/buffer.cpp
Jhonnyg: It's just a shortcut to avoid having to constantly check the buffer for existance in the components since they create the buffers at different times (not necessary at world init, e.g sprites). 

I don't have a strong opinion. Maybe it's better to always create a render buffer when creating a component world? The default single buffer that is created will not allocate any memory anyway so it might as well always be present? 🤔

PR #8348, opened 2023-12-23
Migrate OSX to GLFW3+

### LINKED ISSUE #7970: Migrate to GLFW3
**Task (REQUIRED):**
Our current GLFW version is over ten years old and is thus severely outdated. As a first step of refacturing this system, we should upgrade to the latest GLFW3 for desktop platforms and keep modifications out of it. In the long run we should remove building glfw altogether from the core engine and just store it as a package.

It's obvious isn't it!

Jhonnyg:
We have removed our old modified GLFW version on OSX in favour of the latest publicly released GLFW 3.4 version. This will enable us to improve our desktop feature set, as well as help fix many legacy OS issues across the board. When we consider that this work is stable enough on OSX, the rest of the desktop platforms will get updated as well.

Fixes #7970


engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -543,7 +569,7 @@ static void LogFrameBufferError(GLenum status)
 
     static bool OpenGLIsSupported()
     {
-        return (glfwInit() == GL_TRUE);
+        return true;
     }
```
Jhonnyg, line 573: If we got this far, we already know we have opengl support

share/ext/glfw/build_glfw.sh
```diff
@@ -38,6 +38,9 @@ CMAKE_FLAGS="-DCMAKE_BUILD_TYPE=Release ${CMAKE_FLAGS}"
 CMAKE_FLAGS="-DGLFW_BUILD_EXAMPLES=OFF ${CMAKE_FLAGS}"
 CMAKE_FLAGS="-DGLFW_BUILD_TESTS=OFF ${CMAKE_FLAGS}"
 CMAKE_FLAGS="-DGLFW_BUILD_DOCS=OFF ${CMAKE_FLAGS}"
+CMAKE_FLAGS="-DGLFW_VULKAN_STATIC=OFF ${CMAKE_FLAGS}"
+
+CMAKE_FLAGS_VULKAN="-DGLFW_VULKAN_STATIC=ON"
```
Jhonnyg, line 43: We build two libraries, one with vulkan support and one without

Jhonnyg: This is also why the packages are a lot bigger. Not sure if we can do anything about it since we want vulkan as opt-in/opt-out

engine/engine/wscript
```diff
@@ -127,7 +136,6 @@ def configure(conf):
     conf.env['STLIB_HID_NULL'] = 'hid_null'
     conf.env['STLIB_INPUT'] = 'input'
     conf.env['STLIB_LIVEUPDATE'] = 'liveupdate'
-    conf.env['STLIB_PLATFORM_NULL'] = 'platform_null'
```
Jhonnyg, line 130: duplicate

engine/graphics/src/graphics.cpp
```diff
@@ -72,7 +72,8 @@ namespace dmGraphics
 
         while(next)
         {
-            if (next->m_Priority < selected->m_Priority && next->m_IsSupportedCb())
+            bool is_supported = next->m_IsSupportedCb();
+            if (next->m_Priority < selected->m_Priority && is_supported)
             {
```
Jhonnyg, line 77: If we only link with one adapter we never actually do this check..

engine/hid/src/hid_null.cpp
```diff
@@ -58,6 +60,7 @@ namespace dmHID
 
     void Update(HContext context)
     {
+        dmPlatform::PollEvents(context->m_Window);
```
Jhonnyg, line 63: Old glfw automatically polls for events when doing a swapbuffers, so the hid_null doesn't respond to any mouse events (like being able to close the window).

vlaaad (dismissed):
The editor changes look good to me!

engine/platform/src/platform_window.h
```diff
@@ -55,6 +55,15 @@ namespace dmPlatform
         DEVICE_STATE_KEYBOARD_NUMBER_PAD = 5,
         DEVICE_STATE_KEYBOARD_EMAIL      = 6,
         DEVICE_STATE_KEYBOARD_PASSWORD   = 7,
+        DEVICE_STATE_KEYBOARD_RESET      = 8,
+        DEVICE_STATE_JOYSTICK_PRESENT    = 9,
```
AGulev, line 59: this JOYSTICK and GAMEPAD in the same header are so annoying ...

AGulev (dismissed):
👍

engine/graphics/src/opengl/graphics_opengl.cpp
JCash: Then I guess this function is more on the deprecated side of things?
If we do checks before this step, then we'll likely never really call the "IsSupported()" function?

engine/hid/src/native/hid_native.cpp
```diff
@@ -228,7 +225,7 @@ namespace dmHID
                 Keyboard* keyboard = &context->m_Keyboards[k];
                 keyboard->m_Connected = 1; // TODO: Actually detect if the keyboard is present
 
-                for (uint32_t i = 0; i < MAX_KEY_COUNT; ++i)
+                for (uint32_t i = 32; i < MAX_KEY_COUNT; ++i)
```
JCash, line 228: Hmm, why we suddenly don't store things in the first integer (m_Keys[0]) ?

engine/platform/src/platform_window.h
```diff
@@ -140,23 +162,42 @@ namespace dmPlatform
...
     void           PollEvents(HWindow window);
+    void           SwapBuffers(HWindow window);
+
+    void*          AcquireAuxContext(HWindow window);
+    void           UnacquireAuxContext(HWindow window, void* aux_context);
+
+    // Platform specifics
+    int32_t        AndroidVerifySurface(HWindow window);
+    void           AndroidBeginFrame(HWindow window);
+    void           iOSSetViewTypeOpenGL(HWindow window);
```
JCash, line 200: Another option is to put the platform specific things in a platform specific header, and only include that. E.g.:
```
#if defined(ANDROID)
    #include <platform_android.h>
#endif
```

etc, just like you did for opengl etc

JCash (dismissed):

engine/graphics/src/opengl/graphics_opengl.cpp
Jhonnyg: No, we still need it for other adapters (vulkan + dx12 afaik).

engine/hid/src/native/hid_native.cpp
Jhonnyg: Eeeh yeah I don't remember actually. I've reverted it

PR #8353, opened 2023-12-30
Added options to control how json data is encoded and decoded

### LINKED ISSUE #8317: add support for using `nil` for `json.decode` instead of `json.null`
When I decode JSON using `json.decode`, my sparse tables have `json.null` in them- which is a `userdata` value and incompatible with numerous other APIs, including `sys.save`. Example JSON data:
```                "crew": [
                    {
                        "tier": 1,
                        "quest": 6,
                        "card_key": "jealous_stepsisters"
                    },
                    null,
                    null,
                    {
                        "tier": 1,
                        "card_key": "tink"
                    }
                ],
```

An option to `json.decode` to simple use `nil` when encountering nulls in the JSON. E.g. for the above data, I would like the resulting table to simply have [1] and [4] filled, without having any placeholder (userdata) value in [2] and [3].

1) using a Lua-only JSON library. I'm doing this for now, but this adds redundant code and is presumably much less performant.

2) scanning the result from `json.decode` and nilling out the userdatas. This is my long-term plan but is also obviously more complex and less performant overall. (the above is a tiny subset of the data, I'm deserializing 50k+ of data at a time)

britzl:
It is now possible to configure the behaviour when encoding and decoding JSON data.

When encoding to JSON it is possible to control wether an empty Lua table should encode as a JSON object or array. The default behaviour when encoding an empty Lua table is to encode to a JSON object.

When decoding from JSON it is now possible to control wether a JSON null value should decode to `json-null` or `nil`. The default behaviour when decoding a JSON null value is to decode to nil.

```lua
local t = {}
json.encode(t, { encode_empty_table_as_object = true }) -- {}
json.encode(t, { encode_empty_table_as_object = false }) -- []
json.encode(t) -- {}


local s = "[ 1, 2, null, 3 ]"
json.decode(s, { decode_null_as_userdata = true}) -- { 1, 2, json.null, 4 }
json.decode(s, { decode_null_as_userdata = false}) -- { 1, 2, nil, 4 }
json.decode(s) -- { 1, 2, nil, 4 }
```

Fixes #8317

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/engine/script/src/test/test_json.lua
+++ b/engine/script/src/test/test_json.lua
@@ -57,7 +57,9 @@ function test_json_invalid_primitive()
 end
 
 function test_json_valid_primitive()
-    assert(json.decode("null") == json.null)
+    assert(json.decode("null") == nil)
+    assert(json.decode("null", { decode_null_as_userdata = false }) == nil)
+    assert(json.decode("null", { decode_null_as_userdata = true }) == json.null)
     assert(json.decode("true") == true)
     assert(json.decode("false") == false)
 
@@ -161,6 +163,8 @@ end
 function test_json_encode()
     -- empty table becomes empty dict
     assert(json.encode({}) == "{}")
+    assert(json.encode({}, { encode_empty_table_as_object = true }) == "{}")
+    assert(json.encode({}, { encode_empty_table_as_object = false }) == "[]")
     -- table with nested table(s) becomes array of tables
     assert(json.encode({{}}) == "[{}]")
     -- table with elements becomes array of elements
@@ -169,7 +173,7 @@ function test_json_encode()
     assert(json.encode(nil) == "null")
     -- however, a table with nil becomes empty table
     assert(json.encode({nil}) == "{}")
-    -- interleaved nil becomse a null element (note: not in quotes)
+    -- interleaved nil becomes a null element (note: not in quotes)
     assert(json.encode({1,nil,2}) == "[1,null,2]")
 
     local v3_enc = json.encode(vmath.vector3()) -- user data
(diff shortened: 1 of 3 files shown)


britzl:
Question is if this should be configurable as an option to json.decode()?

```
json.decode(s, { decode_null_as_userdata = false }
```

JCash:
I think the release notes should specify which one (true/false) is the default?

engine/script/src/test/test_json.lua
```diff
@@ -57,7 +57,9 @@ function test_json_invalid_primitive()
 end
 
 function test_json_valid_primitive()
-    assert(json.decode("null") == json.null)
+    assert(json.decode("null") == nil)
+    assert(json.decode("null", { decode_null_as_userdata = false }) == nil)
+    assert(json.decode("null", { decode_null_as_userdata = true }) == json.null)
```
JCash, line 62: Can we add a test for testing the regression?
I.e that the "key":null shouldn't become "key":json.null ?

JCash (dismissed):
lgtm
Could add another test since we didn't catch the regression.

engine/script/src/test/test_json.lua
britzl: Added!

britzl:
> I think the release notes should specify which one (true/false) is the default?

Good point. Added to code comment and to PR notes.

JCash (approved):

PR #8361, opened 2024-01-04
Fix for shutting down OpenAL in a more robust manner

JCash:
Instead of waiting for the OpenAL buffers to be finished processing, we instead stop the source and delete the buffers.
This seems to be the way it's done in 

### Technical changes

By looking at the public repo for [openal-soft](https://github.com/kcat/openal-soft)
they use a simpler approach:
https://github.com/kcat/openal-soft/blob/6675317107257c2cc16c947b359d557821d85bf2/examples/alstream.c#L115-L116

https://github.com/kcat/openal-soft/blob/6675317107257c2cc16c947b359d557821d85bf2/examples/common/alhelpers.c#L90-L104

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/engine/sound/src/devices/device_openal.cpp
+++ b/engine/sound/src/devices/device_openal.cpp
@@ -36,7 +36,8 @@ namespace dmDeviceOpenAL
     {
         ALCdevice*      m_Device;
         ALCcontext*     m_Context;
-        dmArray<ALuint> m_Buffers;
+        dmArray<ALuint> m_AllBuffers;   // All original buffers
+        dmArray<ALuint> m_Buffers;      // The pool
         ALuint          m_Source;
         uint32_t        m_MixRate;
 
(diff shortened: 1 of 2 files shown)


engine/sound/src/devices/device_openal.cpp
```diff
@@ -120,44 +128,12 @@ namespace dmDeviceOpenAL
...
-                break;
-            }
-        }
+        CheckAndPrintError();
 
         alDeleteSources(1, &openal->m_Source);
+        CheckAndPrintError();
 
-        alDeleteBuffers(openal->m_Buffers.Size(), openal->m_Buffers.Begin());
+        alDeleteBuffers(openal->m_AllBuffers.Size(), openal->m_AllBuffers.Begin());
```
JCash, line 136: We've had our old iterative solution since 2014 (8d01fa7c839), but I'm not sure why it was written that way to begin with.

AGulev: I remember some issues on iOS. I think new solution should be carefully tested on iOS (esp. with going background/foreground, closing app, and restart)

britzl: Yeah, this is my main concern as well. But the old solution came in place 10 years ago. A lot has happened since.

britzl: I do like that the new solution is so clean and without that loop though!

britzl (approved):

Issue #8261, opened 2023-11-24
Decrease memory use in large projects


matgis:
**Task (REQUIRED):**
Users have reported no longer being able to build their large project in the editor on computers with limited RAM installed. We'd like to decrease the amount of memory required by the project.

**Expected outcome (REQUIRED):**
The project uses less memory and can be successfully built on these RAM-constrained computers.

PR #8375, opened 2024-01-08
Slightly improve editor memory usage for large projects

matgis:
Slightly improve editor memory usage in projects with a large number of node endpoints. This typically corresponds to the number of files in the project, but varies significantly by file type.

Improves #8261

### Technical changes
* Changed the hash map implementation inside the `WeakInterner` to use a weak addressing scheme instead of linked-list buckets. This saves having to store a reference to the next `Entry` inside each `Entry` in the hash map.
* Updated the hash map implementation inside the `WeakInterner` to use prime number hash table sizes. This is necessary to achieve optimal distribution of values, which is critical when using a weak addressing scheme.
* Updated the `hashCode` and `hashEq` implementations for `Endpoint` to produce better hash values, which also helps with the value distribution.
* Store the `node-id` as a primitive `long` instead of a boxed `Long` in the `Endpoint`. This does not appear to save any memory at present, but does have an effect on the performance of `hashCode`, `hashEq`, and equality comparisons between `Endpoints`, measurably improving lookup times.
* Added functions to the `dev` module to get statistics from the `WeakInterner` used to intern `Endpoints`.

### DIFF
--- a/editor/src/clj/editor/gui.clj
+++ b/editor/src/clj/editor/gui.clj
@@ -3087,10 +3087,8 @@
         child-indices (g/node-value parent :child-indices)
         before? (partial > node-index)
         after? (partial < node-index)
-        ascending-order #(compare %1 %2)
-        descending-order #(compare %2 %1)
         neighbour (first (sort-by second
-                                  (if (= offset -1) descending-order ascending-order)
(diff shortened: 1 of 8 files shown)


vlaaad (approved):
Ship it!

PR #8396, opened 2024-01-12
Cache project build results even if the engine build fails

### LINKED ISSUE #8395: Editor should cache project build results even if the engine build fails
Currently the editor throws away the perfectly good results from the project build if the native extension build of the engine reports an error, causing rebuilds to take longer than they need to. The engine failing to build shouldn't stop us from caching the project build results.

matgis:
The editor will now cache project build results independently of any native extension engine build errors that may occur.

Fixes #8395


vlaaad (approved):
LGTM!

PR #8416, opened 2024-01-17
New API functions: sys.load_buffer and sys.load_buffer_async

### LINKED ISSUE #3637: Async sys.load_resource (DEF-2670)
Description:
Right now we can use `sys.load_resource()` to load resources from the game archive. This blocks the main thread. We would like to have a way to do that in async way and receive a message/callback upon completion.

Use case:
Loading a lot of images in one gui screen at the same time slows down the animations.

Business need:
Better user experience.

Jhonnyg:
Two new script functions has been added:

```lua
function async_callback(self, request_id, status, buf)
    -- do something with data here
end

local buf = sys.load_buffer(path)
local request_id = sys.load_buffer_async(path, callback)
```

These functions can be used to read data into a buffer object from either via the resource system, or direct file locations if the resource could not be found on any of the mounted resource locations. Meaning, the path supplied is not necessary a custom resource location, it can be any file on disk. 

Fixes #3637

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


engine/dlib/src/dlib/opaque_handle_container.h
```diff
@@ -90,6 +90,21 @@ class dmOpaqueHandleContainer
...
+     * range of the object list. If the handle is invalid, a NULL pointer will be returned.
+     *
+     * Note: this is indented for internal use only and relies on the fact that the backing
+     *       data structure for the container is a continuous array.
+     *
+     * @name GetByIndex
+     * @param handle [type:uint32_t] The index to query the container with
+     * @return pointer [type:T*] Return the object associated with the handle If it is valid, NULL otherwise
+     */
+    T* GetByIndex(uint32_t handle);
```
Jhonnyg, line 106: Alternatively we could add an iterator or some sort of callback instead that could be invoked per element

engine/engine/src/engine.cpp
```diff
@@ -1535,16 +1535,32 @@ namespace dmEngine
...
                         dmScript::Update(engine->m_SharedScriptContext);
-                    } else {
-                        if (engine->m_GOScriptContext) {
+                    }
+                     else
+                     {
+                        if (engine->m_GOScriptContext)
+                        {
+                            script_lib_context.m_LuaState = dmScript::GetLuaState(engine->m_GOScriptContext);
+                            dmGameSystem::UpdateScriptLibs(script_lib_context);
```
Jhonnyg, line 1553: We need to call this to keep track of all active requests. The load queue has a limited amount of pending requests, so we call the update function to dispatch initalized but not sent requests (preload queue does something similar)

engine/gamesys/src/gamesys/scripts/script_sys_gamesys.cpp
```diff
@@ -0,0 +1,476 @@
...
+        assert(top == lua_gettop(L));
+
+        g_SysModule.m_Factory           = context.m_Factory;
+        g_SysModule.m_LoadQueue         = dmLoadQueue::CreateQueue(context.m_Factory);
+        g_SysModule.m_LoadRequestsMutex = dmMutex::New();
+    }
+
+    bool ScriptSysGameSysUpdate(const ScriptLibContext& context)
+    {
+        dmMutex::ScopedLock lk(g_SysModule.m_LoadRequestsMutex);
```
Jhonnyg, line 414: I wonder if we can get stalls here if the load function has taken the mutex any doing a massive load from a http request or some such scenario? In the update we need to look at the requests in the list of requests, but the load function will touch the data, so we need to solve this somehow.

engine/gamesys/src/gamesys/scripts/script_sys_gamesys.cpp
```diff
@@ -0,0 +1,476 @@
...
+        dmLoadQueue::FreeLoad(g_SysModule.m_LoadQueue, request->m_Request);
+        bool result = true;
+
+        if (dmScript::IsCallbackValid(request->m_CallbackInfo))
+        {
+            lua_State* L = dmScript::GetCallbackLuaContext(request->m_CallbackInfo);
+            DM_LUA_STACK_CHECK(L, 0);
+
+            // callback has the format:
+            // function(self, request_id, status, buffer)
```
Jhonnyg, line 84: What do we want to return here? The request id can be used to identify a specific request that was issued, even if multiple requests for the same path was issued.

britzl: Not sure what exactly is needed, but I suppose if we want to give ourselves a bit more flexibility then we could do what we do with `http.request()` and return a `result` object containing status, buffer and whatever else we may need.

engine/gamesys/src/gamesys/scripts/script_sys_gamesys.cpp
```diff
@@ -0,0 +1,476 @@
...
+     * local my_request = sys.load_buffer_async("/assets/my_level_data.bin", my_callback)
+     * ```
+     *
+     * Load binary data from non-custom resource files on disk:
+     *
+     * ```lua
+     * function my_callback(self, request_id, status, buf)
+     *   if status ~= sys.REQUEST_STATUS_FINISHED then
+     *     -- uh oh! File could not be found, do something graceful
+     *   elif request_id == self.first_asset then
```
britzl, line 304: elseif

britzl (changes requested):
Wrong syntax in example. And a comment on the callback function arguments.

engine/dlib/src/dlib/opaque_handle_container.h
JCash: Please rename it "index" from "handle", as the api becomes a bit confusing.

engine/engine/src/engine.cpp
```diff
@@ -1535,16 +1535,32 @@ namespace dmEngine
...
                             dmScript::Update(engine->m_GOScriptContext);
                         }
-                        if (engine->m_RenderScriptContext) {
+                        if (engine->m_RenderScriptContext)
```
JCash, line 1556: I assume this means we cannot load from the render script?

engine/gamesys/src/gamesys/scripts/script_sys_gamesys.cpp
```diff
@@ -0,0 +1,476 @@
...
+        dmScript::LuaCallbackInfo* callback_info = dmScript::CreateCallback(dmScript::GetMainThread(L), 2);
+
+        if (callback_info == 0x0)
+        {
+            return luaL_error(L, "sys.load_buffer_async failed to create callback");
+        }
+
+        dmhash_t path_hash = dmHashString64(path);
+        {
+            dmMutex::ScopedLock lk(g_SysModule.m_LoadRequestsMutex);
```
JCash, line 332: Unless you need to interact with the lock, use the define `DM_MUTEX_SCOPED_LOCK`

engine/gamesys/src/gamesys/scripts/script_sys_gamesys.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+            // We currently don't do anything about duplicated requests here (should we?)
```
JCash, line 334: I don't think we should care. The user can easily prevent it, and we can avoid redundant code. Every bit counts.

engine/gamesys/src/gamesys/scripts/script_sys_gamesys.cpp
JCash: You can use `dmMutex::TryLock()` if you want?

engine/gamesys/src/gamesys/scripts/script_sys_gamesys.cpp
```diff
@@ -0,0 +1,476 @@
...
+                    result &= HandleRequestCompleted(request);
+                }
+                else
+                {
+                    void* buf;
+                    uint32_t size;
+                    dmLoadQueue::LoadResult load_result;
+                    if (dmLoadQueue::EndLoad(g_SysModule.m_LoadQueue, request->m_Request, &buf, &size, &load_result) == dmLoadQueue::RESULT_OK)
+                    {
+                        HandleRequestCompleted(request);
```
JCash, line 442: Shouldn't result be updated here?

engine/gamesys/src/gamesys/scripts/script_sys_gamesys.cpp
```diff
@@ -0,0 +1,476 @@
...
+                    }
+                }
+            }
+        }
+
+        return result;
```
JCash, line 448: The result is only checked in the unit tests?

engine/resource/src/resource.h
```diff
@@ -406,6 +408,16 @@ namespace dmResource
...
+    struct PreloadHintInfo
+    {
+        HPreloader m_Preloader;
+        int32_t m_Parent;
+    };
+
+    // load with default internal buffer and its management, returns buffer ptr in 'buffer'
+    Result LoadResource(HFactory factory, const char* path, const char* original_name, void** buffer, uint32_t* resource_size);
+    // load with own buffer
+    Result DoLoadResource(HFactory factory, const char* path, const char* original_name, uint32_t* resource_size, LoadBufferType* buffer);
```
JCash, line 420: The naming "Do..." suggests that it's a private function. (especially since one has to look at the code in the source file to figure out the difference)

Please rename it to e.g. LoadResourceFromBuffer

engine/gamesys/src/gamesys/scripts/script_sys_gamesys.cpp
```diff
@@ -0,0 +1,476 @@
...
+                }
+            }
+
+            if (g_SysModule.m_LoadRequests.Full())
+            {
+                g_SysModule.m_LoadRequests.Allocate(4);
+            }
+
+            LuaRequest* request     = new LuaRequest();
+            request->m_Path         = path;
```
JCash, line 351: strdup(path)

https://www.lua.org/manual/5.1/manual.html#lua_tolstring
```
there is no guarantee that the pointer returned by [lua_tolstring](https://www.lua.org/manual/5.1/manual.html#lua_tolstring) will be valid after the corresponding value is removed from the stack.
```

JCash (changes requested):
Some minor fixes needed

engine/engine/src/engine.cpp
Jhonnyg: No, we don't have access to buffer.* functions either, so not sure how that would work

engine/gamesys/src/gamesys/scripts/script_sys_gamesys.cpp
Jhonnyg: Yes, I refactored this to call a function from the test to check the latest result instead.

Issue #5301, opened 2020-10-27
Create and edit markdown and text files within the Defold editor.


bryanrieger:
**Is your feature request related to a problem? Please describe (REQUIRED):**
I often use markdown and plain text files for docs/notes/logs in my Defold projects, but there is no way to create or edit these files directly within Defold.

**Describe the solution you'd like (REQUIRED):**
I'd like to be able to create and edit markdown and plain text files directly within the Defold editor, much like you can many other asset types. Ideally, you should be able to add markdown or plain text files to a project by right-clicking in the ‘Assets’ panel and opening the ‘New…’ menu and selecting 'Markdown' or 'Text' file. You would edit a markdown or plain text file just like any other file by double-clicking it to open from the Assets panel. As Markdown is currently rendered for documentation purposes, it would be necessary to make the document editable (as plain text) through a tab/toggle/etc. The default state should probably remain rendered and non-editable for documentation purposes.

**Describe alternatives you've considered (REQUIRED):**
Currently I keep a text editor such as VSCode or BBEdit open in order to create and edit markdown and plain text files within Defold projects.

**Additional context (OPTIONAL):**
This is definitely a low-priority request, but would make creating and maintaining project documentation a little simpler.

subsoap:
I often create files such as text txt files outside of the editor. This would be useful to do in editor.

jbp4444:
I wonder if a simpler approach might be to have an "open md as text" option -- e.g. a checkbox in the Preferences-->Code panel. Unchecked: the file opens as markdown, but is view only (the way it currently works); if Checked: the file is opened as a plain text file (you can read/write/edit it, but there is no markdown rendering).

vovs03:
# 💚 Plus one in support

> I also join in wanting to have this functionality directly in the Defold-editor. Of course, I understand that this may not be a priority, but I would like _to use_ 💚 **this option** [**Edit Markdown-file**].

PR #8422, opened 2024-01-18
Allow creating any file with any extension

vlaaad:
User-facing changes:
This changeset adds a new "File" option to the New File selector. When the option is selected, the user then can create any file with any extension. We still use file templates if the user types a registered file extension.

Technical notes:
I can't believe we didn't have it.

Related to #5301

### DIFF
--- a/editor/src/clj/editor/asset_browser.clj
+++ b/editor/src/clj/editor/asset_browser.clj
@@ -107,6 +107,10 @@
     :command :new-file
     :expand? true
     :icon "icons/64/Icons_29-AT-Unknown.png"}
+   {:label "New File"
+    :command :new-file
+    :user-data {:any-file true}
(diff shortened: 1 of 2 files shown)


matgis (changes requested):
New feature works great, but it is now possible to name both New File and New Folder ".",  "..". This results in weirdness or an error, so it shouldn't be possible. You also shouldn't be able to create files or folders outside the project by typing in something like "../../new".

vlaaad:
We filter out slashes, so it should be impossible to create files or folders outside of the project:
Screenshot 2024-01-19 at 09 22 49

vlaaad:
Ah, you can type 2 dots and it will attempt to create a parent dir...

matgis (approved):
Ship it!

PR #8424, opened 2024-01-18
Render target as a resource

### LINKED ISSUE #8312: Render target as a resource
Currently there is no connection between render targets and anything in the game world, having a render target resource that can be rendered as a camera target or render script and assigned to game objects would make certain render setups easier to implement.

A .render_target resource with fields:

```
width/height
attachment type (2D, Cube, Array)
attachment count
attachment format (rgba8, float and so on)
depth/stencil enabled
depth format (16,24,32)
```

As a first step, we don't have to implement the entire chain of resource file + editor support. We can add a new resource.create_render_target() function together with a way of retrieving texture handles from the resource (perhaps via resource.get_texture_info?) so that it can be passed into a render script and used in render.set_render_target and so on.

We can create render targets in the render scripts now, but there is no way to attach an existing texture to it, or get a texture handle from a render target. While either of those options would be easier to implement, having an RT as a resource has more benefits from an authoring standpoint, e.g format validation, assigning global texture assets, editor preview, effect authoring and so on.

Unity: https://docs.unity3d.com/Manual/class-RenderTexture.html

Jhonnyg:
A new resource type has been added that creates render targets that can be used as a regular texture in the engine.
image

The .render file has also been refactored to support this new render resource type. You can now add .render_target files into the render prototype resource. In the render_script, you can use these render targets in the same way as before, except that you can now pass in the render resource name when binding the render target:

image

```lua
-- in a render script:
render.set_render_target("my_rt")
```

A new function has also been added to .script files to get all the metadata about a render target resource:
`resource.get_render_target_info(path)`

Fixes #8312


engine/gamesys/src/gamesys/resources/res_render_prototype.cpp
```diff
@@ -53,28 +60,106 @@ namespace dmGameSystem
...
+                    }
+                    else if (res_type == res_type_render_target)
+                    {
+                        render_resource_type = dmRender::RENDER_RESOURCE_TYPE_RENDER_TARGET;
+                    }
+                    else
+                    {
+                        result = dmResource::RESULT_NOT_SUPPORTED;
+                        break;
+                    }
```
Jhonnyg, line 123: Is it good enough to check the extension in the name instead of checking the actual type information?

engine/gamesys/src/gamesys/resources/res_render_prototype.cpp
```diff
@@ -53,28 +60,106 @@ namespace dmGameSystem
...
+                        prototype->m_Instance,
+                        render_resource_metas[i].m_Name,
+                        render_resource_val,
+                        render_resource_metas[i].m_Type);
                 }
             }
+            else if (!prototype->m_RenderResources.Full())
+            {
+                result = dmResource::RESULT_OUT_OF_RESOURCES;
+            }
```
Jhonnyg, line 162: This is from the old code, but I feel it's a bit odd to return this result? The reason the fail condition has happened is because some resource couldn't be loaded afaik

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -2855,6 +2999,13 @@ static void LuaInit(lua_State* L, dmGraphics::HContext graphics_context)
     SETGRAPHICS_ENUM(TEXTURE_TYPE_2D);
     SETGRAPHICS_ENUM(TEXTURE_TYPE_CUBE_MAP);
     SETGRAPHICS_ENUM(TEXTURE_TYPE_2D_ARRAY);
+
+    SETGRAPHICS_ENUM(BUFFER_TYPE_COLOR0_BIT);
+    SETGRAPHICS_ENUM(BUFFER_TYPE_COLOR1_BIT);
+    SETGRAPHICS_ENUM(BUFFER_TYPE_COLOR2_BIT);
+    SETGRAPHICS_ENUM(BUFFER_TYPE_COLOR3_BIT);
+    SETGRAPHICS_ENUM(BUFFER_TYPE_DEPTH_BIT);
+    SETGRAPHICS_ENUM(BUFFER_TYPE_STENCIL_BIT);
```
Jhonnyg, line 3008: It's a bit annoying that we don't have access to the render namespace here..

engine/graphics/src/dmsdk/graphics/graphics.h
```diff
@@ -36,6 +36,16 @@ namespace dmGraphics
...
 
+    /*#
+     * Asset handle
+     * A generic asset handle with 21 bits of metadata and a 32 bits opaque handle
+     * Note that we can only use a total of 53 bits out of the 64 due to the handles are exposed
+     * to the users via lua: http://lua-users.org/wiki/NumbersTutorial
+     * @typedef
+     * @name HAssetHandle
+     */
+    typedef uint64_t HAssetHandle;
```
Jhonnyg, line 47: Moved here because res_texture.h is using this for the backing texture handle.

editor/test/integration/save_data_test.clj
```diff
@@ -470,7 +470,15 @@
...
-    {"texture" :unimplemented}}}) ; Default texture resources not supported yet.
+    {"texture" :unimplemented}} ; Default texture resources not supported yet.
+
+   'dmRenderDDF.RenderTargetDesc.DepthStencilAttachment
+   {:default
+    {"format" :unimplemented}} ; Non-default depth/stencil format not supported yet.
+
+   'dmRenderDDF.RenderPrototypeDesc
+   {:default
+    {"materials" :deprecated}}})
```
matgis, line 481: We try to keep this list sorted, so `dmRenderDDF.RenderPrototypeDesc` should really come before `dmRenderDDF.RenderTargetDesc.DepthStencilAttachment`.

editor/src/clj/editor/render_target.clj
```diff
@@ -0,0 +1,129 @@
...
+  (:require [editor.build-target :as bt]
+            [dynamo.graph :as g]
+            [editor.graph-util :as gu]
+            [editor.gl.texture :as texture]
+            [editor.protobuf :as protobuf]
+            [editor.protobuf-forms :as protobuf-forms]
+            [editor.resource-node :as resource-node]
+            [editor.workspace :as workspace])
+  (:import [com.dynamo.graphics.proto Graphics$TextureImage$Type Graphics$TextureImage$TextureFormat]
+           [com.dynamo.render.proto RenderTarget$RenderTargetDesc]))
```
matgis, line 25: **Code > Optimize Imports** will sort this list for you!

editor/src/clj/editor/render_pb.clj
```diff
@@ -70,43 +70,43 @@
...
-(g/defnk produce-pb-msg [script-resource named-materials]
+(g/defnk produce-pb-msg [script-resource named-render-resources]
   {:script (resource/resource->proj-path script-resource)
-   :materials (mapv (fn [{:keys [name material]}]
+   :render-resources (mapv (fn [{:keys [name path]}]
                       {:name name
-                       :material (resource/resource->proj-path material)})
-                    named-materials)})
+                       :path (resource/resource->proj-path path)})
+                    named-render-resources)})
```
matgis, line 109: Indentation looks messed up here. Select lines, then **Code > Auto-Indent Lines** should fix it!

editor/src/clj/editor/render_pb.clj
```diff
@@ -70,43 +70,43 @@
...
-                               (make-named-material-node graph-id node-id name material)))))))
-
-(g/defnk produce-form-data [_node-id script-resource named-materials]
+    [:named-render-resources] (let [graph-id (g/node-id->graph-id node-id)]
+                              (g/transact
+                                (concat
+                                  (for [[named-render-resource-id _] (g/sources-of node-id :named-render-resources)]
+                                    (g/delete-node named-render-resource-id))
+                                  (for [{:keys [name path]} value]
+                                    (make-named-render-resource-node graph-id node-id name path)))))))
```
matgis, line 95: Indentation looks messed up here. Select lines, then **Code > Auto-Indent Lines** should fix it!

editor/src/clj/editor/render_pb.clj
```diff
@@ -27,40 +27,40 @@
...
+  (output named-render-resource g/Any (g/fnk [_node-id name render-resource]
                                  {:name name
-                                  :material material}))
-  (output dep-build-targets g/Any (gu/passthrough dep-build-targets)))
+                                  :path render-resource})))
 
-(defn- make-named-material-node
-  [graph-id render-node name material-resource]
+(defn- make-named-render-resource-node
+  [graph-id render-node name render-resource-path]
```
matgis, line 57: We use `path` only to refer to string paths. This should be a `Resource`. So I guess that makes this a `render-resource-resource` 😅.

editor/src/clj/editor/render_target.clj
```diff
@@ -0,0 +1,129 @@
...
+                          :label "width"
+                          :type :number}
+                         {:path [:height]
+                          :label "height"
+                          :type :number}
+                         {:path [:format]
+                          :label "format"
+                          :type :choicebox
+                          :options (protobuf-forms/make-options (protobuf/enum-values Graphics$TextureImage$TextureFormat))
+                          :default (ffirst (protobuf/enum-values Graphics$TextureImage$TextureFormat))}]}
```
matgis, line 52: Is `TEXTURE_FORMAT_LUMINANCE` really the most sensible default? 🤔

editor/src/clj/editor/render_target.clj
```diff
@@ -0,0 +1,129 @@
...
+              {:path [:depth-stencil-attachment-width]
+               :label "Depth/Stencil Width"
+               :type :number}
+              {:path [:depth-stencil-attachment-height]
+               :label "Depth/Stencil Height"
+               :type :number}
+              {:path [:depth-stencil-attachment-texture-storage]
+               :label "Depth Buffer Texture Storage"
+               :type :boolean}]}]})
```
matgis, line 61: This doesn't seem to match the updated structure of `RenderTarget$RenderTargetDesc`?

Also, add line breaks between top-level forms.

editor/src/clj/editor/render_target.clj
```diff
@@ -0,0 +1,129 @@
...
+                          :set set-form-op
+                          :clear clear-form-op}))))
+
+(g/defnk produce-pb-msg
+  [color-attachments depth-stencil-attachment-width depth-stencil-attachment-height depth-stencil-attachment-format depth-stencil-attachment-texture-storage]
+  {:color-attachments color-attachments
+   :depth-stencil-attachment {:width depth-stencil-attachment-width
+                              :height depth-stencil-attachment-height
+                              :format depth-stencil-attachment-format
+                              :texture-storage depth-stencil-attachment-texture-storage}})
```
matgis, line 77: This doesn't seem to match the updated structure of `RenderTarget$RenderTargetDesc`. How can this possibly work? 🤔

editor/src/clj/editor/render_target.clj
```diff
@@ -0,0 +1,129 @@
...
+      (g/set-property self :depth-stencil-attachment-texture-storage (:texture-storage depth-stencil-attachment)))))
+
+(defn register-resource-types
+  [workspace]
+  (resource-node/register-ddf-resource-type workspace
+                                            :ext "render_target"
+                                            :node-type RenderTargetNode
+                                            :ddf-type RenderTarget$RenderTargetDesc
+                                            :load-fn load-render-target
+                                            :icon atlas-icon
```
matgis, line 126: Should we really use the `atlas-icon` for render targets? `icons/32/Icons_36-Texture.png` seems like it might be more suitable?

editor/test/resources/save_data_project/checked.render
```diff
@@ -3,3 +3,7 @@ materials {
   name: "render_material"
   material: "/checked.material"
 }
+render_resources {
+  name: "render_target"
+  path: "/checked.render_target"
+}
```
matgis, line 9: Did you hand-edit this? `materials` should no longer be here, right?

editor/test/resources/save_data_project/checked.render_target
```diff
@@ -0,0 +1,14 @@
...
+    format: TEXTURE_FORMAT_LUMINANCE
+}
+
+depth_stencil_attachment
+{
+    width: 128
+    height: 128
+    format: DEPTH_STENCIL_FORMAT_D32F_S8U
+    texture_storage: true
+}
```
matgis, line 14: Don't hand-edit these! They should be saved by the editor.

editor/test/resources/save_data_project/silently_migrated/legacy_render_prototype.render
```diff
@@ -0,0 +1,5 @@
+script: "/builtins/render/default.render_script"
+materials {
+  name: "test"
+  material: "/builtins/materials/sprite.material"
+}
```
matgis, line 5: This also looks suspiciously hand-edited seeing as there is no newline at the end. 🤔

matgis (changes requested):
The editor code doesn't seem to match the updated structure of `RenderTarget$RenderTargetDesc`. How can this work? Did you try building and saving from the editor?

editor/test/resources/save_data_project/checked.render_target
Jhonnyg: Well, how am I supposed to know that? :)

editor/src/clj/editor/render_target.clj
Jhonnyg: I'm going with RGBA then!

editor/src/clj/editor/render_target.clj
matgis: Oh, never mind. I just got confused... 😵‍💫
This seems fine after all.

editor/src/clj/editor/render_target.clj
matgis: Oh, never mind. I just got confused... 😵‍💫
This seems fine after all.

editor/src/clj/editor/render_target.clj
```diff
@@ -49,7 +49,7 @@
                           :label "format"
                           :type :choicebox
                           :options (protobuf-forms/make-options (protobuf/enum-values Graphics$TextureImage$TextureFormat))
-                          :default (ffirst (protobuf/enum-values Graphics$TextureImage$TextureFormat))}]}
+                          :default Graphics$TextureImage$TextureFormat/TEXTURE_FORMAT_RGBA}]}
```
matgis, line 52: Does this work? In `material.clj` we specify the default as the keyword value, i.e. `:semantic-type-none` for `Graphics$VertexAttribute$SemanticType`.

engine/gamesys/src/gamesys/resources/res_texture.cpp
```diff
@@ -353,7 +355,7 @@ namespace dmGameSystem
         {
             TextureResource* texture_res = new TextureResource();
             texture_res->m_Texture = texture;
-            params.m_Resource->m_Resource = (void*) texture_res;
+            params.m_Resource->m_Resource = texture_res;
```
JCash, line 358: Huh, I don't see a change here. I expected _something_ to have changed, since the type of m_Texture has changed?
Are we assigning a HTexture to the HAssetHandle ?
Perhaps I'm not used to having handles inherit from each other (I think Windows.h does it a lot though, so I guess it's common).

JCash (changes requested):
My main concern is the GetTexture() being called each frame for all components.
It is an extra indirection that I feel is likely avoidable?
See possible suggestions in PR.

Also, I'm still coming to terms as to using Handle overrides.
One thing is for storing in a list, but it's another thing imho when being used in a certain case. E.g. using a HTexture in the TextureResource.

I don't want the code to become too generic when it comes to using the types (i..e handles)

Small note, the graphics.h doesn't mention that HTexture/HRenderTarget are valid HAssetHandles.

engine/gamesys/src/gamesys/resources/res_render_prototype.cpp
Jhonnyg: I think the lines got messed up here for some reason, I was referring to getting the type from the extension and comparing that to the fixed type pointers we get above instead of just comparing to the extension string

engine/graphics/src/dmsdk/graphics/graphics.h
Jhonnyg: I moved it back as we don't do this any longer.

engine/gamesys/src/waf_gamesys.py
```diff
@@ -330,9 +330,25 @@ def transform_collectionfactory(task, msg):
...
+        entry.name = m.name
+        entry.path = m.material
+        msg.render_resources.append(entry)
+
+    for r in msg.render_resources:
+        r.path = r.path.replace('.material', '.materialc')
+        r.path = r.path.replace('.render_target', '.render_targetc')
+
+    msg.materials.clear()
+    return msg
```
Jhonnyg, line 348: This is the same migration as we do in bob/editor. 

Side-note: Why do we even have this python file that generates all this content and not use bob-light? This is probably faster, but it's yet another pipeline that needs to be kept up-to-date..

matgis (dismissed):
Looks good to me! Nice work!

engine/gamesys/src/gamesys/resources/res_render_target.cpp
```diff
@@ -159,24 +182,22 @@ namespace dmGameSystem
...
-        dmRender::HRenderContext render_context = (dmRender::HRenderContext) params.m_Context;
-
-        rt_resource->m_Texture = dmGraphics::NewRenderTarget(
-            dmRender::GetGraphicsContext(render_context),
-            buffer_type_flags,
-            rt_params);
+        dmRender::HRenderContext render_context  = (dmRender::HRenderContext) params.m_Context;
+        rt_resource->m_RenderTarget              = dmGraphics::NewRenderTarget(dmRender::GetGraphicsContext(render_context), buffer_type_flags, rt_params);
+        
+        //rt_resource->m_TextureResource.m_Texture = dmGraphics::GetRenderTargetTexture(rt_resource->m_RenderTarget, dmGraphics::BUFFER_TYPE_COLOR0_BIT);
```
JCash: old code

engine/gamesys/src/gamesys/resources/res_mesh.cpp
```diff
@@ -204,8 +207,8 @@ namespace dmGameSystem
 
                 if (render_res_type == dmRender::RENDER_RESOURCE_TYPE_RENDER_TARGET)
                 {
-                    RenderTargetResource* rt_res = (RenderTargetResource*) texture_res;
-                    textures[i] = rt_res->m_TextureResource;
+                    render_targets[i] = (RenderTargetResource*) texture_res;
+                    textures[i]       = render_targets[i]->m_TextureResource;
```
JCash, line 211: 👍

JCash (dismissed):

PR #8434, opened 2024-01-22
Add `editor.create_directory()` and code completion to editor scripts

### LINKED ISSUE #8425: Add `editor.mkdir` function
**Context:**
Our users want to be able to create a folder using the editor script API. Currently, the only way to do it is by returning a shell action, but running any lua logic like `io.open():write()` is then impossible. Related to #6810.

**Task:**
Add the `editor.mkdir()` function.

**Expected outcome:**
The users can create a function and write to a file in the folder with the same command. Also, it would be nice if we had editor scripts related completions in the code editor.

vlaaad:
User-facing changes:

This changeset adds a new function to editor scripts that allows the user to create a new directory. The function expects a resource path, i.e. a `/`-prefixed path from the root of the project, for example:
```lua
editor.create_directory("/assets/generated")
local file = io.open("assets/generated/build.json")
file:write(...)
file:flush()
file:close()
```

Additionally, the editor now provides code completions for functions in the `editor` module.

Fixes #8425

### DIFF
--- a/editor/src/clj/editor/editor_extensions.clj
+++ b/editor/src/clj/editor/editor_extensions.clj
@@ -23,6 +23,7 @@
             [editor.console :as console]
             [editor.defold-project :as project]
             [editor.error-reporting :as error-reporting]
+            [editor.fs :as fs]
             [editor.graph-util :as gu]
             [editor.handler :as handler]
             [editor.lsp :as lsp]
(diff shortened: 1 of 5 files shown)


vlaaad:
[defold/doc PR](https://github.com/defold/doc/pull/376)

matgis (approved):
LGTM!

PR #8450, opened 2024-01-25
If the `hash` string is unknown, show its value instead.

AGulev:
Instead of showing just `<unknown>` when the corresponding string value is unknown, print its numeric value. For example:
>`<unknown:2256903789829718645>`

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/engine/dlib/src/dlib/hash.cpp
+++ b/engine/dlib/src/dlib/hash.cpp
@@ -669,7 +669,7 @@ DM_DLLEXPORT const char* dmHashReverseSafe64(uint64_t hash)
     if (s == 0)
     {
         char tmp[64];
-        dmSnPrintf(tmp, sizeof(tmp), "<unknown:%llu>", hash);
+        dmSnPrintf(tmp, sizeof(tmp), "<unknown:%llu>", (unsigned long long)hash);
         return tmp;
     }
     return s;
--- a/engine/dlib/src/dlib/hash.cpp
+++ b/engine/dlib/src/dlib/hash.cpp
@@ -21,6 +21,7 @@
 #include "index_pool.h"
 #include "align.h"
 #include <dlib/mutex.h>
+#include <dlib/dstrings.h>
 #include <dlib/hashtable.h>
 
 struct ReverseHashEntry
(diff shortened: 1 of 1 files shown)


engine/dlib/src/dlib/hash.cpp
```diff
@@ -665,12 +666,24 @@ DM_DLLEXPORT void dmHashReverseErase64(uint64_t hash)
 DM_DLLEXPORT const char* dmHashReverseSafe64(uint64_t hash)
 {
     const char* s = (const char*)dmHashReverse64(hash, 0);
-    return s != 0 ? s : "<unknown>";
+    if (s == 0)
+    {
+        char tmp[64];
+        dmSnPrintf(tmp, sizeof(tmp), "<unknown:%llu>", (unsigned long long)hash);
```
JCash, line 672: What is the output used for?
Debugging only?
Or as actual strings in a release build? (Do we want to encourage that?)

And, what's the best format here?
signed long long, unsigned long long, hex?

Are we using the hashes elsewhere? E.g. in an editor/bob script, what would be the output format?

JCash (changes requested):
I couldn't find a design, or any ticket relating to this?
I feel this is a case of helpful thing, that might be hard to roll back later on?

engine/dlib/src/dlib/hash.cpp
AGulev: you right

What I want is to have an ability to get a value from a hashmap by this value in debug build
I expect hash to be the same in release and in debug build
So if I have the hash value from the release build, it should be possible to get the corresponding string value in the debug build
---
Talking about downsides. Yes, people already try to extract strings from the hash value, and this fix will not make it worse. Also, a string from debug will be different from a string from release, so they can’t abuse it

AGulev: One more thing is: 
I have a lot of messages in a released game like:
>“DEFOLD:ERROR:GAMEOBJECT: Instance ‘<unknown>’ could not be found when dispatching message ‘<unknown>’ sent from game:<unknown>#<unknown>\n”

It makes it impossible to figure out what’s going on. I hope having such a numeric value will help with "symbolication" of this message to make it easier to understand what's going on there

JCash (approved):

PR #8456, opened 2024-01-28
New function `physics.update_mass(url, mass)`

### LINKED ISSUE #3149: Change mass of collision object at runtime (DEF-3930)
Requested on the forum:

https://forum.defold.com/t/how-to-change-mass-of-collision-object-def-3930/51340

AGulev:
Added new function `physics.update_mass(url, mass)` which updates the mass of a dynamic 2D collision object in the physics world.

The function recalculates the density of each shape based on the total area of all shapes and the specified mass, then updates the mass of the body accordingly.

Fix https://github.com/defold/defold/issues/3149

## PR checklist

* [x] Code
* [x] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_collision_object.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_collision_object.cpp
@@ -847,8 +847,6 @@ namespace dmGameSystem
 
     static void RayCastCallback(const dmPhysics::RayCastResponse& response, const dmPhysics::RayCastRequest& request, void* user_data)
     {
-        dmhash_t message_id;
-
         dmGameObject::HInstance instance = (dmGameObject::HInstance)request.m_UserData;
         dmMessage::URL receiver;
         receiver.m_Socket = dmGameObject::GetMessageSocket(dmGameObject::GetCollection(instance));
@@ -2097,6 +2095,24 @@ namespace dmGameSystem
             dmPhysics::SetMaskBit2D(component->m_Object2D, groupbit, boolvalue);
         }
         return true;
+    }
+
+    void UpdateMass(void* _world, void* _component, float mass)
+    {
+        CollisionWorld* world = (CollisionWorld*)_world;
+        CollisionComponent* component = (CollisionComponent*)_component;
+
+        if (world->m_3D)
+        {
+            dmLogError("The Update Mass function has not been implemented for 3D yet");
+        }
+        else
+        {
+            if(!dmPhysics::UpdateMass2D(component->m_Object2D, mass))
+            {
+                dmLogError("The Update Mass function can be used only for Dynamic objects with shape area > 0");
+            }
+        }
     }
 
     static bool CompCollisionIterPropertiesGetNext(dmGameObject::SceneNodePropertyIterator* pit)
--- a/engine/gamesys/src/gamesys/scripts/script_physics.cpp
+++ b/engine/gamesys/src/gamesys/scripts/script_physics.cpp
@@ -1639,6 +1639,49 @@ namespace dmGameSystem
         return 0;
     }
 
+     /*# updates the mass of a dynamic 2D collision object in the physics world.
+     *
+     * The function recalculates the density of each shape based on the total area of all shapes and the specified mass, then updates the mass of the body accordingly.
+     *
+     * Note: Currently only supported in 2D physics.
+     *
+     * @name physics.update_mass
+     *
+     * @param collisionobject [type:string|hash|url] the collision object whose mass needs to be updated.
+     * @param mass [type:number] the new mass value to set for the collision object.
+     *
+     * @examples
+     *
+     * ```lua
+     *  physics.update_mass("#collisionobject", 14)
+     * ```
+     */
+    static int Physics_UpdateMass(lua_State* L)
+    {
+        DM_LUA_STACK_CHECK(L, 0);
+        int top = lua_gettop(L);
+
+        dmScript::GetGlobal(L, PHYSICS_CONTEXT_HASH);
+        PhysicsScriptContext* context = (PhysicsScriptContext*)lua_touserdata(L, -1);
+        lua_pop(L, 1);
+
+        dmGameObject::HInstance sender_instance = CheckGoInstance(L);
+        dmGameObject::HCollection collection = dmGameObject::GetCollection(sender_instance);
+
+        void* world = dmGameObject::GetWorld(collection, context->m_ComponentIndex);
+        if (world == 0x0)
+        {
+            return DM_LUA_ERROR("Physics world doesn't exist. Make sure you have at least one physics component in collection.");
+        }
+        void* comp = 0x0;
+        GetCollisionObject(L, 1, collection, &comp, &world);
+
+        float mass = luaL_checknumber(L, 2);
+        dmGameSystem::UpdateMass(world, comp, mass);
+
+        return 0;
+    }
+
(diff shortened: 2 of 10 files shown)


engine/gamesys/src/gamesys/components/comp_collision_object.cpp
```diff
@@ -847,8 +847,6 @@ namespace dmGameSystem
 
     static void RayCastCallback(const dmPhysics::RayCastResponse& response, const dmPhysics::RayCastRequest& request, void* user_data)
     {
-        dmhash_t message_id;
```
AGulev, line 850: fix compiler warning (never used var)

engine/gamesys/src/gamesys/components/comp_collision_object.cpp
```diff
@@ -2099,6 +2097,24 @@ namespace dmGameSystem
...
     }
 
+    void UpdateMass(void* _world, void* _component, float mass)
+    {
+        CollisionWorld* world = (CollisionWorld*)_world;
+        CollisionComponent* component = (CollisionComponent*)_component;
+
+        if (world->m_3D)
+        {
+            dmLogError("The Update Mass function has not been implemented for 3D yet");
```
AGulev, line 2107: For now it's 2D only function

engine/physics/src/physics/physics.h
```diff
@@ -1080,16 +1080,16 @@ namespace dmPhysics
...
-	void SetMaskBit2D(HCollisionObject2D collision_object, uint16_t groupbit, bool boolvalue);
-
-	uint16_t GetGroup3D(HCollisionObject3D collision_object);
-	void SetGroup3D(HWorld3D world, HCollisionObject3D collision_object, uint16_t groupbit);
-	bool GetMaskBit3D(HCollisionObject3D collision_object, uint16_t groupbit);
-	void SetMaskBit3D(HWorld3D world, HCollisionObject3D collision_object, uint16_t groupbit, bool boolvalue);
-
+    bool GetMaskBit2D(HCollisionObject2D collision_object, uint16_t groupbit);
+    void SetMaskBit2D(HCollisionObject2D collision_object, uint16_t groupbit, bool boolvalue);
+    bool UpdateMass2D(HCollisionObject2D collision_object, float mass);
```
AGulev, line 1087: I added only this line, the rest is a fix for tabs

engine/physics/src/physics/physics_2d.cpp
```diff
@@ -1234,56 +1234,84 @@ namespace dmPhysics
...
+                    filter.maskBits |= groupbit;
+                else
+                    filter.maskBits &= ~groupbit;
+                fixture->SetFilterData(filter, 0);
+            }
+            fixture = fixture->GetNext();
+        }
+    }
+
+    bool UpdateMass2D(HCollisionObject2D collision_object, float mass) {
```
AGulev, line 1277: I added only this function, the rest is just a fix of tabs

engine/gamesys/src/gamesys/scripts/script_physics.cpp
```diff
@@ -1639,6 +1639,49 @@ namespace dmGameSystem
...
+     * @examples
+     *
+     * ```lua
+     *  physics.update_mass("#collisionobject", 14)
+     * ```
+     */
+    static int Physics_UpdateMass(lua_State* L)
+    {
+        DM_LUA_STACK_CHECK(L, 0);
+        int top = lua_gettop(L);
```
britzl, line 1662: Unused?

AGulev: no =) 
Thank you!

britzl (dismissed):

engine/gamesys/src/gamesys/components/comp_collision_object.cpp
```diff
@@ -2099,6 +2097,24 @@ namespace dmGameSystem
         return true;
     }
 
+    void UpdateMass(void* _world, void* _component, float mass)
```
JCash, line 2100: I would still like it to be called "SetMass" like any other setter/getter.

engine/physics/src/physics/physics_2d.cpp
```diff
@@ -1234,56 +1234,84 @@ namespace dmPhysics
...
+            fixture = fixture->GetNext();
+
+        }
+        fixture = body->GetFixtureList();
+        if (total_area <= 0.0f) {
+            return false;
+        }
+        float new_density = mass / total_area;
+        while (fixture) {
+            fixture->SetDensity(new_density);
```
JCash, line 1298: Is density the same for all fixtures _before_ updating the mass?
Because, now they are.
Also, I can totally see us changing this in the future, where each shape should have it's own mass.

JCash (approved):

engine/physics/src/physics/physics_2d.cpp
AGulev: There is no such option to set mass per fixture. The only thing we can add in the future is to set density manually per fixture and then calculate mass based on it (or vice-versa). But it looks like some advanced level of setup. When I checked the other box2d tools/engines - nobody setup it like this.

Before - yes, they are set equal during creating, but with wrong collision object mass per shape (we will fix later).

engine/gamesys/src/gamesys/components/comp_collision_object.cpp
AGulev: I explained it in the DD:
>Using of `set_mass` feels like wrong expectations
>Current function is more like `update_mass`
because it’s not replacing one value with another that will be taken by a system later
>But this is an active action where we iterate over all collision body fixtures calculating density and after that reseting mass of the collision body to recalculate the rest of parameters based on fixtures with new density

If I change the other APIs we have then with `set_*` we just change value in a structure but in this case we start recalculation process

AGulev: there will de another function later, it will be something like `set_mass_data` and `get_mass_data` - these two are getter/seeter functions. Because they just set and force using of values from it without any calculations and ignoring physic laws

AGulev: more info about it here: https://docs.google.com/document/d/1BJDYFRNSAiRdKhUQeaH6MkwH0ok_XW851R6-Rne7zQ8/edit

engine/physics/src/physics/physics_2d.cpp
AGulev: No, density is different for each shape in current implementation. It was more like the same mass for each piece with different density. But it is still one collision object, why it should have different density unless you manually decided to have it different (which isn't possible right now)?

I mean, you have a collision object with 2 shapes and mass 2kg. The first shape is square 100x100 and the second 1x1. Why should they have the same mass if the first one has 10k bigger area?

engine/gamesys/src/gamesys/components/comp_collision_object.cpp
JCash: Sorry I didn't see that part of the DD. And I don't wholly agree with that :)
I can agree that it might be a way to communicate that some extra work is being done.
But I also think that "Set" functions are allowed to calculate things, just as "Get" functions are allowed to calcualte things.
But, let's go with this.

PR #8474, opened 2024-01-30
Bugfix for particles, models and tilegrids when running on non-opengl graphics adapters

Jhonnyg:


engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1204,10 +1204,12 @@ namespace dmGameSystem
                     total_count += world->m_VertexBufferVertexCounts[batch_index];
                 }
 
-                world->m_DispatchCount++;
-
-                DM_PROPERTY_ADD_U32(rmtp_ModelVertexCount, total_count);
-                DM_PROPERTY_ADD_U32(rmtp_ModelVertexSize, total_count * sizeof(dmRig::RigModelVertex));
+                if (total_count > 0)
+                {
+                    world->m_DispatchCount++;
```
JCash, line 1209: I cannot see where this variable is currently used in the model component? 🤔

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1204,10 +1204,12 @@ namespace dmGameSystem
...
+                    DM_PROPERTY_ADD_U32(rmtp_ModelVertexCount, total_count);
+                    DM_PROPERTY_ADD_U32(rmtp_ModelVertexSize, total_count * sizeof(dmRig::RigModelVertex));
```
JCash, line 1211: Note to self: We should keep a proper variable to collect these, as the vertex formats will differ.

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1204,10 +1204,12 @@ namespace dmGameSystem
                     total_count += world->m_VertexBufferVertexCounts[batch_index];
                 }
 
-                world->m_DispatchCount++;
-
-                DM_PROPERTY_ADD_U32(rmtp_ModelVertexCount, total_count);
-                DM_PROPERTY_ADD_U32(rmtp_ModelVertexSize, total_count * sizeof(dmRig::RigModelVertex));
+                if (total_count > 0)
```
JCash, line 1207: Probably not for this PR, but I thought I should bring it up:

I noticed that the "batch index" is predetermined by the `m_MinorOrder`in the CompModelRender(), but is that really the best way to do it?

Shouldn't the vertex buffer index be more dynamic, and be counted upwards, as the `RENDER_LIST_OPERATION_BATCH` calls are coming in?
It feels a bit unintuitive 🤔

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
```diff
@@ -408,24 +408,25 @@ namespace dmGameSystem
...
+                break;
+            case dmRender::RENDER_LIST_OPERATION_BATCH:
+                RenderBatch(pfx_world, params.m_Context, params.m_Buf, params.m_Begin, params.m_End);
+                break;
+            case dmRender::RENDER_LIST_OPERATION_END:
+                if (pfx_world->m_VertexBufferData.Size() > 0)
+                {
+                    dmRender::SetBufferData(params.m_Context, pfx_world->m_VertexBuffer, pfx_world->m_VertexBufferData.Size(), pfx_world->m_VertexBufferData.Begin(), dmGraphics::BUFFER_USAGE_STREAM_DRAW);
+                    DM_PROPERTY_ADD_U32(rmtp_ParticleVertexCount, pfx_world->m_VerticesWritten);
+                    DM_PROPERTY_ADD_U32(rmtp_ParticleVertexSize, pfx_world->m_VertexBufferData.Size());
```
JCash, line 425: These statements already handle a vertex count of 0.

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
```diff
@@ -408,24 +408,25 @@ namespace dmGameSystem
...
+                    pfx_world->m_DispatchCount++;
```
JCash, line 426: (Not saying this is part of this PR)

Q: Is this working the way we want it to? (perhaps we've postponed it to a MVP2 or something?)
(Assuming we want the same logic in all components, so think "model component" as a current example, as it supports vertex formats.)

If I have 2 components with two materials, using the same predicate tag but different vertex formats.
During one `render.draw(pred)` I would get two calls to `RENDER_LIST_OPERATION_BATCH`, each requiring a different vertex buffer. Currently, this doesn't seem to match the code, which uses the "dispatch count", which essentially counts numbers of `render.draw()` calls?

To me it seems we need to manage buffers depending on `RENDER_LIST_OPERATION_BATCH` ?

Or is my logic way off here?

JCash (dismissed):
Fixes the immediate issue, but it also feels like there's more to the story.
See questions.

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
Jhonnyg: In the case where you have two materials for one draw call, we can reuse the same vertex buffer regardless of how the vertex format looks like. The memory overwrite issue happens when you do multiple render.draw executions with the same component type:

```
render.draw(pred_1):
batch_begin (reset write pointers)
batch_1 (put vertices into vx buffer on cpu)
...
batch_x
batch_end (write vertices into vx buffer on GPU)

render.draw(pred_2):
batch_begin (again, reset write pointers)
batch_1 -> batch_x (same as before)
batch_end (write vertices on GPU with the SAME vx buffer handle as before)
```
The second batch_end call will start copying over the data to the memory mapped GPU buffer during the same frame, meaning we can get random results depending on when the data is read or written.

In the case of models, we do things a bit differently but it should essentially be the same thing (I'll have to double check why we don't advance the render buffers for models, might be something that's overlooked..)

engine/gamesys/src/gamesys/components/comp_model.cpp
Jhonnyg: We would have to deal with that based on local vs world space, local spaced models uses the "baked" vertex buffers already afaik. But yes, it would be great if we simplified this a bit

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
JCash: "we can reuse the same vertex buffer "

Really? Isn't the vertex buffer tied to a single vertex format? And, once the first batch is submitted, it is in flight and we don't want to alter it anymore?
I am probably missing something 🤔

JCash: Ah, I'm talking about a different scenario:

```
render.draw(pred):
batch_begin (reset write pointers)
batch_1 (material 1, vertex format 1)
batch_2 (material 2, vertex format 2)
...
batch_x
batch_end
```

This to me seems like we don't handle well at the current moment?

JCash (approved):

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
Jhonnyg: We do yes, we can reuse the same vertex buffer as long as we calculate all the offsets correct (they must start at an even vertex, which depends on the vertex declaration stride).

PR #8476, opened 2024-01-30
Use a shared dmJobThread worker for gamesys and graphics

Jhonnyg:
Changes:
* moved the liveupdate job thread to dlib
* gamesys and graphics uses a shared worker pool now
* null graphics can use a worker to set data (used for tests only right now)


engine/dlib/src/dlib/job_thread.cpp
```diff
@@ -69,7 +69,7 @@ struct JobThreadContext
 struct JobContext
 {
 #if defined(DM_HAS_THREADS)
-    dmThread::Thread    m_Thread;
+    dmArray<dmThread::Thread> m_Threads;
```
Jhonnyg, line 72: the jobthread can use multiple workers now

engine/dlib/src/dlib/job_thread.cpp
```diff
@@ -207,8 +219,18 @@ void Update(HContext context)
...
 }
 
+bool PlatformHasThreadSupport()
+{
+#if defined(DM_HAS_THREADS)
+    return true;
+#else
+    return false;
+#endif
+}
```
Jhonnyg, line 234: Maybe this should be moved somewhere? I use it to figure out if we should initialize all of the async job state in the graphics adapters

engine/dlib/src/dlib/job_thread.h
```diff
@@ -3,24 +3,36 @@
...
     typedef void (*FCallback)(void* context, void* data, int result);
 
-    HContext    Create(const char* thread_name);
-    void        Destroy(HContext context);
-    void        Update(HContext context); // Flushes any items and calls PostProcess
-    void        PushJob(HContext context, FProcess process, FCallback callback, void* user_context, void* data);
+    struct JobThreadCreationParams
+    {
+        const char* m_ThreadName;
+    };
```
Jhonnyg, line 29: Just one param now, but in the future we might want to add stuff like tags / priority to push jobs to specific workers (like a specific graphics worker, a specific IO worker) that can take priority for certain jobs, or perhaps stack size and whatnot.

engine/dlib/src/test/test_job_thread.cpp
```diff
@@ -0,0 +1,87 @@
...
+    uint8_t contexts[8];
+    uint8_t datas[8];
+    for (int i = 0; i < DM_ARRAY_SIZE(contexts); ++i)
+    {
+        contexts[i] = i;
+        datas[i] = 0;
+        dmJobThread::PushJob(ctx, process, callback, (void*) &contexts[i], (void*) &datas[i]);
+    }
+
+    uint64_t stop_time = dmTime::GetTime() + 1*1e6; // 1 second
```
Jhonnyg, line 63: It apparently needs a somewhat "high" value here for CI..

engine/engine/src/engine.cpp
```diff
@@ -852,6 +852,10 @@ namespace dmEngine
             return false;
         }
 
+        dmJobThread::JobThreadCreationParams job_thread_create_param;
+        job_thread_create_param.m_ThreadName = "DefoldJobThread1";
+        engine->m_JobThreadContext           = dmJobThread::Create(1, &job_thread_create_param);
```
Jhonnyg, line 857: Right now we create a single worker thread, but this could be scaled up if we want. People recommend something like "number of cores + 1" or similar, but we have a bunch of other threads as well so I don't know what we want here.

engine/graphics/src/null/graphics_null.cpp
```diff
@@ -75,6 +76,7 @@ namespace dmGraphics
         m_Height                  = params.m_Height;
         m_Window                  = params.m_Window;
         m_PrintDeviceInfo         = params.m_PrintDeviceInfo;
+        m_JobThread               = params.m_JobThread;
```
Jhonnyg, line 79: The null adapter can now use threads if we want to. This is mostly used for testing right now, the reason being I wanted to make sure the job system API works. But maybe we don't want to do this change at all? I.e skip the test(s) in test_graphics and instead do some sort of runtime validation as we do for OpenGL?

engine/liveupdate/src/liveupdate.cpp
```diff
@@ -946,7 +946,10 @@ namespace dmLiveUpdate
 
         dmLogInfo("Liveupdate folder located at: %s", g_LiveUpdate.m_AppSupportPath);
 
-        g_LiveUpdate.m_JobThread = dmJobThread::Create("liveupdate_jobs");
+        dmJobThread::JobThreadCreationParams job_thread_create_param;
+        job_thread_create_param.m_ThreadName = "liveupdate_jobs";
+
+        g_LiveUpdate.m_JobThread = dmJobThread::Create(1, &job_thread_create_param);
```
Jhonnyg, line 952: Liveupdate has its own thread right now, but we could potentially use the main engine workers for this as well and process more jobs in parallell. It felt like a bigger change right now since I don't know this system very well

engine/platform/src/platform_window_glfw.cpp
```diff
@@ -362,20 +362,45 @@ namespace dmPlatform
         return -1;
     }
 
+    #ifndef __EMSCRIPTEN__
+        #define GLFW_AUX_CONTEXT_SUPPORTED
+    #endif
+    static inline int32_t QueryAuxContextImpl()
+    {
+    #if defined(GLFW_AUX_CONTEXT_SUPPORTED)
+        return glfwQueryAuxContext();
```
Jhonnyg, line 371: I guess we could add this function to the GLFW js library, but I don't want to keep adding things to it since it's an external lib nowadays.

engine/resource/src/async/load_queue_sync.cpp
```diff
@@ -73,17 +73,7 @@ namespace dmLoadQueue
...
-            load_result->m_LoadResult = request->m_PreloadInfo.m_LoadResourceFunction(queue->m_Factory, request->m_CanonicalPath, request->m_Name, size, &queue->m_LoadBuffer, request->m_PreloadInfo.m_Context);
-            *buf = queue->m_LoadBuffer.Begin();
-            assert(queue->m_LoadBuffer.Size() == *size);
-        }
-        else
-        {
-            load_result->m_LoadResult = dmResource::LoadResource(queue->m_Factory, request->m_CanonicalPath, request->m_Name, buf, size);
-        }
-
+        load_result->m_LoadResult    = dmResource::LoadResource(queue->m_Factory, request->m_CanonicalPath, request->m_Name, buf, size);
```
Jhonnyg, line 76: rolled back some of the changes from #8416 since we are not using the "load queue" anymore in sys.load_buffer_async

engine/dlib/src/test/test_job_thread.cpp
```diff
@@ -0,0 +1,87 @@
...
+    dmJobThread::Destroy(ctx);
+
+    ASSERT_TRUE(tests_done);
+}
+
+int main(int argc, char **argv)
+{
+    jc_test_init(&argc, argv);
+    return jc_test_run_all();
+}
```
JCash, line 87: In case you didn't know, you can test with "--with-tsan" locally, to help catching potential issues early on
(we do it nightly on CI as well)

engine/engine/src/engine.cpp
JCash: Yeah, threads are a resource.
E.g. Switch only has a few.
I'd say a game project config value should fit most projects.

engine/engine/src/engine.cpp
```diff
@@ -1531,6 +1537,12 @@ namespace dmEngine
                         return;
                     }
                 }
+
+                {
+                    DM_PROFILE("Job Thread");
```
JCash, line 1542: I'd say you can put it in the system itself? (profile.h is in dlib)
Also for each worker threads.

engine/dlib/src/dlib/job_thread.h
```diff
@@ -3,24 +3,36 @@
...
+
+    HContext Create(uint8_t thread_count, const JobThreadCreationParams* create_params);
```
JCash, line 31: For platforms that _don't_ support threads, it makes more sense to put that in the creation params as well, I think

engine/dlib/src/wscript
```diff
@@ -391,6 +392,7 @@ def build(bld):
     bld.install_files('${PREFIX}/ext/include/basis/encoder', bld.path.ant_glob('basis/encoder/*.h'))
     bld.install_files('${PREFIX}/ext/include/basis/transcoder', bld.path.ant_glob('basis/transcoder/*.h'))
     bld.install_files('${PREFIX}/ext/include/stb/', bld.path.ant_glob('stb/*.h'))
+    bld.install_files('${PREFIX}/ext/include/jc/', bld.path.ant_glob('jc/*.h'))
```
JCash, line 395: also update "apply_license.py" as it has moved

JCash (dismissed):
lgtm.
Some minor cleanup needed (e.g apply_license.py, and profile scopes)

PR #8497, opened 2024-02-05
Vertex declaration simplifcation

Jhonnyg:
This PR simplifies how vertex declarations are used in the various graphics adapters.
Instead of each adapter having their own structs to define this, we use the same representation across all adapters.

Pros: Simpler code, less to maintain, 
Cons: Slight increase of memory per adapter since we have to store some adapter specific data in the struct. In the future we can probably fix this by a second layer of separation, but this should be good enough for now.

TODO (not this PR):
* Update rive to use the new setup (vertex step enum is now in dmsdk)
* Update consoles (remove specific vx decl functions in PS4/PS5)

### DIFF
--- a/engine/graphics/src/dmsdk/graphics/graphics.h
+++ b/engine/graphics/src/dmsdk/graphics/graphics.h
@@ -439,10 +439,17 @@ namespace dmGraphics
         BLEND_FACTOR_ONE_MINUS_CONSTANT_ALPHA = 14,
     };
 
+    /*#
+     * Vertex step function. Dictates how the data for a vertex attribute should be read in a vertex shader.
+     * @enum
+     * @name VertexStepFunction
+     * @member VERTEX_STEP_FUNCTION_VERTEX
+     * @member VERTEX_STEP_FUNCTION_INSTANCE
+     */
     enum VertexStepFunction
     {
-        VERTEX_STEP_VERTEX,
-        VERTEX_STEP_INSTANCE,
+        VERTEX_STEP_FUNCTION_VERTEX,
+        VERTEX_STEP_FUNCTION_INSTANCE,
     };
 
     /*#
--- a/engine/graphics/src/dmsdk/graphics/graphics.h
+++ b/engine/graphics/src/dmsdk/graphics/graphics.h
@@ -439,6 +439,12 @@ namespace dmGraphics
         BLEND_FACTOR_ONE_MINUS_CONSTANT_ALPHA = 14,
     };
 
+    enum VertexStepFunction
+    {
+        VERTEX_STEP_VERTEX,
+        VERTEX_STEP_INSTANCE,
+    };
+
     /*#
      * Create new vertex stream declaration. A stream declaration contains a list of vertex streams
      * that should be used to create a vertex declaration from.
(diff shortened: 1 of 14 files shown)


engine/graphics/src/dmsdk/graphics/graphics.h
```diff
@@ -439,6 +439,19 @@ namespace dmGraphics
...
+     * @enum
+     * @name VertexStepFunction
+     * @member VERTEX_STEP_FUNCTION_VERTEX
+     * @member VERTEX_STEP_FUNCTION_INSTANCE
+     */
+    enum VertexStepFunction
+    {
+        VERTEX_STEP_FUNCTION_VERTEX,
+        VERTEX_STEP_FUNCTION_INSTANCE,
+    };
```
Jhonnyg, line 453: We already support instancing in the vulkan API because of extensions (and we need it soon for engine wide support anyway)

engine/graphics/src/null/graphics_null.cpp
```diff
@@ -953,6 +939,7 @@ namespace dmGraphics
             case TYPE_SAMPLER_2D:       return 1;
             case TYPE_SAMPLER_CUBE:     return 1;
             case TYPE_SAMPLER_2D_ARRAY: return 1;
+            case TYPE_IMAGE_2D:         return 1;
```
Jhonnyg, line 942: warning fix

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -1128,7 +1128,6 @@ static void LogFrameBufferError(GLenum status)
         // Hardcoded values for iOS and Android for now. The value is a hint, max number of vertices will still work with performance penalty
         // The below seems to be the reported sweet spot for modern or semi-modern hardware
         context->m_MaxElementVertices = 1024*1024;
-        context->m_MaxElementIndices = 1024*1024;
```
Jhonnyg, line 1131: these were not used

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -1488,11 +1486,6 @@ static void LogFrameBufferError(GLenum status)
         return (((OpenGLContext*) context)->m_IndexBufferFormatSupport & (1 << format)) != 0;
     }
 
-    static uint32_t OpenGLGetMaxElementIndices(HContext context)
```
Jhonnyg, line 1491: not used

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -1512,16 +1505,14 @@ static void LogFrameBufferError(GLenum status)
         memset(vd, 0, sizeof(VertexDeclaration));
 
         vd->m_Stride = 0;
-
         for (uint32_t i=0; i<stream_declaration->m_StreamCount; i++)
         {
-            vd->m_Streams[i].m_NameHash      = stream_declaration->m_Streams[i].m_NameHash;
-            vd->m_Streams[i].m_LogicalIndex  = i;
```
Jhonnyg, line 1519: not used

PR #8514, opened 2024-02-07
Fix thread memory read/write issues

Jhonnyg:


engine/graphics/src/test/test_graphics.cpp
```diff
@@ -718,34 +716,32 @@ TEST_F(dmGraphicsTest, TestTextureAsyncDelete)
     params.m_Height = HEIGHT;
     params.m_Format = dmGraphics::TEXTURE_FORMAT_LUMINANCE;
 
+    const uint32_t TEXTURE_COUNT = 64;
```
Jhonnyg, line 719: We now try to flood the thread(s) with more work to _really_ make sure this process works

engine/graphics/src/null/graphics_null.cpp
```diff
@@ -1324,12 +1324,9 @@ namespace dmGraphics
...
     }
 
-    static int DoDeleteTexture(void* _context, void* _texture)
+    static int DoDeleteTexture(void* _h_texture, void* _texture)
     {
-        NullContext* context = (NullContext*)_context;
-        HTexture texture = (HTexture) _texture;
-        Texture* tex = GetAssetFromContainer<Texture>(context->m_AssetHandleContainer, texture);
-
+        Texture* tex = (Texture*) _texture;
```
Jhonnyg, line 1329: We don't touch the asset handle container on the texture thread any longer. We do that in the complete function

Issue #8394, opened 2024-01-12
Use async engine build in the editor


vlaaad:
**Task (REQUIRED):**
Similarly to bob, we should use async API to build the engine in the editor.

**Expected outcome (REQUIRED):**
The editor builds the engine using asynchronous HTTP requests.

PR #8519, opened 2024-02-08
Update the Extender Client

vlaaad:
Technical notes:
While implementing the ExtenderClient support in the editor, I noticed that it defines unnecessary methods that its users must implement. This changeset removes the implementations of the methods.

Related to #8394
Related to https://github.com/defold/extender/issues/343


matgis (approved):
Looks good to me!

JCash (approved):

PR #8520, opened 2024-02-08
Increased length of debug output of an URL to 512 characters

JCash:
## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


AGulev (dismissed):

engine/dlib/src/dlib/hash.cpp
```diff
@@ -668,7 +668,7 @@ DM_DLLEXPORT const char* dmHashReverseSafe64(uint64_t hash)
     const char* s = (const char*)dmHashReverse64(hash, 0);
     if (s == 0)
     {
-        char tmp[64];
```
JCash, line 671: I'm afraid I missed this in the code review:

* Returning stack allocated variables is not good, as they may be invalid when receiving it on the other end.

* The function is not thread safe now, since two different threads will write to the same memory buffer.

I don't have a good way to deal with it as the current api is.
I would introduce a new function for your use case, where to store the variable:

`const char* dmHashReverseSafe64(hash, buffer, buffersize);`

engine/script/src/script_msg.cpp
```diff
@@ -80,13 +80,13 @@ namespace dmScript
         dmStrlCat(buffer, ":", buffer_size);
         if (url->m_Path != 0)
         {
-            dmSnPrintf(tmp, sizeof(tmp), "%s", dmHashReverseSafe64(url->m_Path));
+            dmStrlCpy(tmp, dmHashReverseSafe64(url->m_Path), sizeof(tmp));
```
JCash, line 83: Replacing a more expensive formatting function with a string copy instead.

engine/script/src/test/test_script_msg.cpp
```diff
@@ -388,23 +388,46 @@ TEST_F(ScriptMsgTest, TestURLToString)
...
     ASSERT_EQ(dmMessage::RESULT_OK, dmMessage::NewSocket("very_very_very_very_very_very_very_very_very_very_very_very_socket", &overflow_socket));
     ASSERT_TRUE(dmScriptTest::RunString(L,
-        "local url = msg.url()\n"
-        "print(tostring(url))\n"
-        "url = msg.url(\"socket\", \"path\", \"test\")\n"
-        "print(tostring(url))\n"
-        "-- overflow\n"
-        "url = msg.url(\"very_very_very_very_very_very_very_very_very_very_very_very_socket\", \"path\", \"test\")\n"
-        "print(tostring(url))\n"
+        "test_msg_fn = function (url, expected)\n"
```
JCash, line 391: I'm still not certain why my tests failed, other than that I got trashed strings.
This was also backed up by the fact that the CI failed/succeeded randomly for the same static testdata.

So it was definitely misbehaving, and I think we actually need to change the reversehash back to the way it was, and handle it locally where the feature was wanted.

JCash:
@AGulev What was your use case for that previous change? Was it only for printing from scripts?
Then I suggest we move that change there.

AGulev:
> @AGulev What was your use case for that previous change? Was it only for printing from scripts? Then I suggest we move that change there.

I explained it here
https://github.com/defold/defold/pull/8450

A lot of places where we use hashes in release produce this type of errors:
```
"DEFOLD:ERROR:GAMEOBJECT: Component '<unknown>#<unknown>' could not be found when dispatching message '<unknown>' sent from gamePage:<unknown>#<unknown>\n"
```

```
"DEFOLD:ERROR:GAMESYS: The collection /collections/_loader/spinner/spinner.collectionc could not be enabled since it is already. Message '<unknown>' sent from loader:<unknown>#<unknown> to loader:<unknown>#<unknown>.\n"
```

```
"DEFOLD:ERROR:GAMESYS: A sound could not be played, error: -3. Message '<unknown>' sent from upgrades_scene:<unknown>#<unknown> to main:<unknown>#<unknown>."
```

etc
it makes it impossible to fix in most of the cases

engine/dlib/src/dlib/hash.cpp
```diff
@@ -666,24 +666,31 @@ DM_DLLEXPORT void dmHashReverseErase64(uint64_t hash)
...
 {
     const char* s = (const char*)dmHashReverse64(hash, 0);
-    if (s == 0)
-    {
-        char tmp[64];
-        dmSnPrintf(tmp, sizeof(tmp), "<unknown:%llu>", (unsigned long long)hash);
-        return tmp;
-    }
-    return s;
+    return s != 0 ? s : "<unknown>";
```
JCash, line 669: In practice, the "s" lifespan is valid until the game (and the reverse table) is shut down.

engine/dlib/src/dlib/hash.cpp
```diff
@@ -666,24 +666,31 @@ DM_DLLEXPORT void dmHashReverseErase64(uint64_t hash)
...
 {
     const char* s = (const char*)dmHashReverse32(hash, 0);
-    if (s == 0)
-    {
-        char tmp[32];
-        dmSnPrintf(tmp, sizeof(tmp), "<unknown:%u>", hash);
-        return tmp;
-    }
-    return s;
+    return s != 0 ? s : "<unknown>";
```
JCash, line 675: Revert to old behavior (it is a public api)

engine/dlib/src/dlib/hash.cpp
```diff
@@ -666,24 +666,31 @@ DM_DLLEXPORT void dmHashReverseErase64(uint64_t hash)
...
 }
 
+DM_DLLEXPORT const char* dmHashReverseSafeBuffer64(uint64_t hash, char* buffer, uint32_t buffer_size)
```
JCash, line 678: Introducing two new thread safe functions.

engine/dlib/src/test/test_dlib.cpp
```diff
@@ -181,10 +181,45 @@ TEST_F(dlib, HashToString64)
...
+
+        const char* reverse_safe = (const char*) dmHashReverseSafe64(iter->second.first);
+        ASSERT_NE((void*) 0, reverse_safe);
+        ASSERT_STREQ(iter->first.c_str(), reverse_safe);
+
         ++iter;
     }
 }
 
+TEST_F(dlib, ReverseHashSafe)
```
JCash, line 193: Added explicit tests for these reverse functions

engine/dlib/src/test/test_dlib.cpp
```diff
@@ -181,10 +181,45 @@ TEST_F(dlib, HashToString64)
...
+        char tmp[256];
+        ASSERT_STREQ(test_string, dmHashReverseSafeBuffer64(h, tmp, sizeof(tmp)));
+        ASSERT_STREQ(test_string, tmp);
+    }
+
+    {
+        dmHashEnableReverseHash(false);
+        const char* test_string = "TestString2";
+        uint64_t h64 = dmHashString64(test_string);
+        uint64_t h32 = dmHashString64(test_string);
```
JCash, line 208: disable reverse hashing for this hash

engine/script/src/script_msg.cpp
```diff
@@ -72,21 +72,21 @@ namespace dmScript
 
         if( !socketname )
         {
-            dmSnPrintf(tmp, sizeof(tmp), "%s", dmHashReverseSafe64(url->m_Socket));
+            dmHashReverseSafeBuffer64(url->m_Socket, tmp, sizeof(tmp));
```
JCash, line 75: Use the new function to copy directly into this temp variable

JCash: As for any other cases, we need to use the new function to do it.

engine/dlib/src/dlib/hash.cpp
```diff
@@ -666,24 +666,31 @@ DM_DLLEXPORT void dmHashReverseErase64(uint64_t hash)
...
+{
+    const char* s = (const char*)dmHashReverse64(hash, 0);
+    if (s)
+        dmStrlCpy(buffer, s, buffer_size);
+    else
+        dmSnPrintf(buffer, buffer_size, "<unknown:%llu>", hash);
```
AGulev, line 684: hmm
tests fails on `:%llu`

maybe we should use `PRIu64 ` from `<inttypes.h>` ?
like
```
dmSnPrintf(buffer, buffer_size, "<unknown:%" PRIu64 ">", hash);
```
?

JCash: We have the DM_HASH_FMT for these cases, and unless you object, I can use that instead.

Currently, it's hex format, but I can add an integer version.
Although, it'd be good if ther are consistent throughout.

engine/dlib/src/dlib/hash.cpp
```diff
@@ -3,10 +3,10 @@
 // Copyright 2009-2014 Ragnar Svensson, Christian Murray
 // Licensed under the Defold License version 1.0 (the "License"); you may not use
 // this file except in compliance with the License.
-// 
+//
```
JCash, line 6: I guess we should fix our update-script so we don't have to deal with this another full year.

engine/dlib/src/dlib/hash.cpp
AGulev: yes, sure. I think it doesn't matter what exactly to convert into a hash in a NE  (trying to get string in Debug version)

AGulev (approved):
😳👏

PR #8522, opened 2024-02-09
Graph successor memory optimizations

matgis:
Large projects with a lot of connections will now use significantly less memory.

### Technical changes
* Excluded empty entries from the `:successors` map.
* We now store successor `Endpoints` in tight-fitting Java arrays in the `:successors` map.
* Added `mem` module for performing memory measurements. Its interface is based on `clj-memory-meter`, but it exposes more options to the underlying `JAMM` library to allow you to exclude classes from the measurement.
* Removed `clj-memory-meter` as a dev dependency.
* Added `:jamm` Leiningen profile to load the `JAMM` library as a Java agent and enable you to perform memory measurements using the `mem` module.
* Added `:no-schemas` Leiningen profile so you can conveniently disable schema checks without hacking `*check-schemas*` in `internal.node`.

### DIFF
--- a/editor/project.clj
+++ b/editor/project.clj
@@ -196,6 +196,7 @@
                                            (.replace (System/getProperty "user.home") \\ \/)
                                            "/.m2/repository/com/github/jbellis/jamm/0.4.0/jamm-0.4.0.jar")]}
                       :no-asserts {:global-vars {*assert* false}}
+                      :no-schemas {:jvm-opts ["-Ddefold.schema.check.disable=true"]}
                       :x86_64-linux {:dependencies [[org.openjfx/javafx-base "22-ea+11" :classifier "linux" :exclusions [org.openjfx/javafx-base]]
                                                     [org.openjfx/javafx-controls "22-ea+11" :classifier "linux" :exclusions [org.openjfx/javafx-controls org.openjfx/javafx-graphics]]
                                                     [org.openjfx/javafx-graphics "22-ea+11" :classifier "linux" :exclusions [org.openjfx/javafx-graphics org.openjfx/javafx-base]]
--- a/editor/project.clj
+++ b/editor/project.clj
@@ -191,6 +191,10 @@
                                :injections [(require 'editor.reveal)]
                                :dependencies [[vlaaad/reveal "1.3.280"]]}
                       :metrics {:jvm-opts ["-Ddefold.metrics=true"]}
+                      :jamm {:dependencies [[com.github.jbellis/jamm "0.4.0"]]
+                             :jvm-opts [~(str "-javaagent:"
+                                           (.replace (System/getProperty "user.home") \\ \/)
+                                           "/.m2/repository/com/github/jbellis/jamm/0.4.0/jamm-0.4.0.jar")]}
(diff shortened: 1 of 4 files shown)


editor/project.clj
```diff
@@ -191,7 +191,12 @@
                                :injections [(require 'editor.reveal)]
                                :dependencies [[vlaaad/reveal "1.3.280"]]}
                       :metrics {:jvm-opts ["-Ddefold.metrics=true"]}
+                      :jamm {:dependencies [[com.github.jbellis/jamm "0.4.0"]]
+                             :jvm-opts [~(str "-javaagent:"
+                                           (.replace (System/getProperty "user.home") \\ \/)
+                                           "/.m2/repository/com/github/jbellis/jamm/0.4.0/jamm-0.4.0.jar")]}
```
matgis, line 197: Maybe there is a better way to do this? 🤔 We're kind of limited in what we can use from `project.clj`, though. Still, this should work across all platforms, I believe.

editor/project.clj
```diff
@@ -233,7 +238,6 @@
                                                     [org.openjfx/javafx-fxml "22-ea+11"]
                                                     [org.openjfx/javafx-swing "22-ea+11"]
                                                     [com.clojure-goes-fast/clj-async-profiler "0.5.1"]
-                                                    [com.clojure-goes-fast/clj-memory-meter "0.1.2"]
```
matgis, line 236: The `:jamm` profile now has a direct dependency to the underlying `JAMM` library `clj-memory-meter` was using.

editor/src/clj/internal/graph.clj
```diff
@@ -1158,7 +1159,7 @@
                                                        ([]
                                                         (pair {} []))
                                                        ([[new-acc remove-acc] [node-id labels old-node-successors]]
-                                                        (let [new-node-successors (if-some [node (gt/node-by-id-at basis node-id)]
+                                                        (let [new-node-successors (when-some [node (gt/node-by-id-at basis node-id)]
```
matgis, line 1162: There was as `nil` in the `false`-case branch.

editor/src/dev/mem.clj
```diff
@@ -0,0 +1,212 @@
...
+
+    :debug
+    If true, print the object layout tree to stdout. Can also be set to a number
+    to limit the nesting level being printed.
+
+    :ignore-known-singletons
+    If false, include known singletons such as Enum, Class, ClassLoader and
+    AccessControlContext in the measurement.
+
+    :ignore-non-strong-references
```
vlaaad, line 188: Nitpick: why not call it `:ignore-weak-references`?

vlaaad (approved):
Looks great!

editor/src/dev/mem.clj
matgis: I think because `WeakReference` is specific type of weak reference, but there are several others like `PhantomReference` and `SoftReference` as well.

vlaaad: Ah, it mirrors the naming and semantics from JAMM's Filters, got it!

PR #8533, opened 2024-02-13
Make welcome screen resizable

vlaaad:


matgis (changes requested):
If we are to make it resizable we should put a limit on how small you can make the window. Currently you can make it so small UI elements scrunch up in an ugly manner and start to overlap. Also, we should make the project titles crop to ellipsis (...) rather than spanning multiple lines if there's not enough room.

vlaaad:
With a minimum size limit, I think there is no need for using ellipsis in project titles

matgis (approved):
Ship it!

PR #8596, opened 2024-02-26
More flexibility when calculating build target content hash

matgis:
### Technical changes
* Added Java interface `com.defold.util.IDigestable`. The editor will allow instances of classes that implement this interface inside build targets. The data written to the `Writer` in the `digest` method will become part of the calculated SHA1 used for the `:content-hash` of the build target.
* Reinstated the ability to ignore entries in build target maps by prefixing their keys with `:digest-ignored/`.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/defold/util/IDigestable.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/defold/util/IDigestable.java
@@ -1,0 +1,35 @@
+// Copyright 2020-2024 The Defold Foundation
+// Copyright 2014-2020 King
+// Copyright 2009-2014 Ragnar Svensson, Christian Murray
+// Licensed under the Defold License version 1.0 (the "License"); you may not use
+// this file except in compliance with the License.
+//
+// You may obtain a copy of the License, together with FAQs at
+// https://www.defold.com/license
+//
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+package com.defold.util;
+
+import java.io.Writer;
+
+/**
+ * Interface used to identify equivalent objects among instances of the same
+ * type. Used during the build process to merge equivalent build outputs into a
+ * single shared resource.
+ */
+public interface IDigestable {
+    /**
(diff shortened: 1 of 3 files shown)


editor/project.clj
```diff
@@ -247,6 +247,7 @@
                                 :source-paths      ["src/dev"]
                                 :repl-options      {:init-ns user}
                                 :proto-paths       ["test/proto"]
+                                :java-source-paths ["test-java"]
```
matgis, line 250: I wanted a Java class with the new `ConsistentHashCode` attribute for the editor tests, but this is the first time we'd wanted to include a Java class in our tests. Ideally, we'd probably like to split test source roots into `test/clj` and `test/java` like we do for `src`, but I didn't want to change the directory structure in this PR.

com.dynamo.cr/com.dynamo.cr.bob/src/com/defold/annotation/ConsistentHashCode.java
```diff
@@ -0,0 +1,28 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Indicates that the value returned by the {@link Object#hashCode() hashCode} method of the annotated class is guaranteed to be consistent between application runs.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface ConsistentHashCode {
+}
```
matgis, line 28: We use a Java attribute rather than a marker interface so that subclasses will not erroneously inherit this property.

vlaaad (changes requested):
I think the part about ignoring keys is good. I think we should use the content for SHA1 hashing and not the hashCode, which is already a hash, but with a lower collision protection. Maybe something like a `Digestable` interface that should write the content to the Writer would do it?

vlaaad (approved):
LGTM!

com.dynamo.cr/com.dynamo.cr.bob/src/com/defold/util/IDigestable.java
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Writes some unique representation of the object to the provided Writer.
+     * Anything written to the Writer will be hashed, along with the class name
+     * of the object, and used to identify equivalent objects among instances
+     * of the same type. The written data is expected to be consistent between
+     * runs of the application.
+     * @param writer The java.io.Writer to write to.
+     * @throws Exception In case of an error.
+     */
+    void digest(Writer writer) throws Exception;
```
JCash, line 34: 👍

PR #8597, opened 2024-02-26
Gamesys vertex attribute consolidation

Jhonnyg:
This PR consolidates some of the vertex attribute writing that happens in the engine for sprites, models and particles. 

TODO for the future:
* Use the shared vertex writer for the particle fx
* Use the shared vertex writer for models


JCash (approved):
I like it a lot!

PR #8601, opened 2024-02-27
Use arm64 dart-sass on Apple silicon

matgis:
### Technical changes
* Use the `arm64` variant of `dart-sass` on Apple silicon, so contributors will not have to install Rosetta 2.


vlaaad (approved):
Shipit!

PR #8604, opened 2024-02-28
Get progress in callback for HTTP requests

### LINKED ISSUE #6486: Report progress for http.request calls
There is currently no way of knowing the status and progress of an ongoing http request while sending and/or receiving large amounts of data.

An API to get status of ongoing requests. Probably a new option to http.request():

```
local function handle_response(self, id, response)
    if response.state == http.STATE_SSLHANDSHAKE then
        print("Negotiating connection")
    elseif response.state == http.STATE_SEND then
        print("Number of bytes sent:", response.bytes_sent)
    elseif response.state == http.STATE_RECEIVE then
        print("Number of bytes received:", response.bytes_received)
        print(response.headers["content-length"])
    elseif response.state == http.STATE_DONE then
        print("Done!")
    end
end

local url = ""
local method = "GET"
local headers = {}
local post_data = nil
local options = {
    report_progress = true
}
http.request(url, method, handle_response, headers, post_data, options)
```

Implement HTTP GET and POST using sockets

Jhonnyg:
Added a new parameter to the option table for the `http.request` function called 'report_progress'. When `report_progress `is true, the amount of bytes sent and/or received for a request will be passed into the callback function:

```lua
http.request("http://my-address", "GET",
	function(self, id, response)
		if response.bytes_total ~= nil then
			update_my_progress_bar(self, response.bytes_received / response.bytes_total)
		else
			-- handle response like normal
		end
	end,
	nil, nil, { report_progress = true })
```

Fixes #6486


engine/dlib/src/dlib/http_client.cpp
```diff
@@ -855,8 +857,6 @@ if (sock_res != dmSocket::RESULT_OK)\
                         r = RESULT_OK;
                         break;
                     }
-
-                    ++chunk_number;
```
Jhonnyg, line 859: unused

engine/gamesys/src/gamesys/test/test_gamesys.h
```diff
@@ -34,9 +37,25 @@
 
 #include <testmain/testmain.h>
 
-#define JC_TEST_IMPLEMENTATION
 #include <jc_test/jc_test.h>
```
Jhonnyg, line 40: I'm using the test_gamesys.h in the HTTP gamesys tests

engine/gamesys/src/gamesys/test/test_gamesys_http.cpp
```diff
@@ -0,0 +1,124 @@
...
+        dmTime::Sleep(30*1000);
+    }
+
+    ASSERT_TRUE(tests_done);
+    ASSERT_TRUE(dmGameObject::Final(m_Collection));
+
+    dmGameSystem::FinalizeScriptLibs(scriptlibcontext);
+}
+
+int main(int argc, char **argv)
```
Jhonnyg, line 86: This file is here for two reasons:
* I don't want to mess the current HTTP tests up that was moved into gamesys
* I don't want to put this into the main gamesys test because it relies on a http server running, and the gamesys test is getting a bit out of hand size wise

engine/script/src/test/wscript
```diff
@@ -18,41 +18,29 @@ def build(bld):
     bld.stlib(features = 'c cxx',
               includes = '..',
               use = 'DLIB LUA script',
-              target = 'test_script_private',
-              source = 'test_script_private.cpp')
+              target = 'test_script',
+              source = 'test_script.cpp test_script_private.cpp')
```
Jhonnyg, line 22: I reuse this in the gamesys http tests now

britzl (dismissed):
Lua API wise it looks good I think. Backwards compatible and easy to use. I'll let @JCash comment on the other code changes.

JCash (dismissed):

PR #8605, opened 2024-02-28
Added Linux layouts for 2 gamepads

Domarius:
I've added layouts for Linux, for the two gamepads:
- 8BitDo Pro 2
- Logitech Logitech Dual Action

Domarius


AGulev (approved):
LGTM

PR #8608, opened 2024-02-29
Use ptr to resource instead of HTexture when getting animation callback

Jhonnyg:
Fixes an issue with web builds for particle effects and gui components, where a texture handle (64-bit) got truncated into a 32-bit value since it is cast to a void* which is 32-bit on web platform(s) due to emscripten.

### DIFF
--- a/engine/engine/src/engine.cpp
+++ b/engine/engine/src/engine.cpp
@@ -415,6 +415,7 @@ namespace dmEngine
         {
             return dmGraphics::TEXTURE_FILTER_LINEAR_MIPMAP_LINEAR;
         }
+        return (dmGraphics::TextureFilter) -1;
     }
 
     dmGraphics::TextureFilter ConvertMagTextureFilter(const char* filter)
--- a/engine/gamesys/src/gamesys/scripts/script_physics.cpp
+++ b/engine/gamesys/src/gamesys/scripts/script_physics.cpp
@@ -1392,7 +1392,10 @@ namespace dmGameSystem
                 lua_setfield(L, -2, "diameter");
                 lua_pushnumber(L, shape_info.m_CapsuleDiameterHeight[1]);
                 lua_setfield(L, -2, "height");
-            break;
+                break;
+            case dmPhysicsDDF::CollisionShape::TYPE_HULL:
+                break;
+            default:break;
         }
 
         return 1;
--- a/engine/gui/src/dmsdk/gui/gui.h
+++ b/engine/gui/src/dmsdk/gui/gui.h
@@ -58,6 +58,14 @@ namespace dmGui
     typedef struct Script* HScript;
 
     /*#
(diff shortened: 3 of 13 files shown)


engine/engine/src/engine.cpp
```diff
@@ -415,6 +415,7 @@ namespace dmEngine
         {
             return dmGraphics::TEXTURE_FILTER_LINEAR_MIPMAP_LINEAR;
         }
+        return (dmGraphics::TextureFilter) -1;
```
Jhonnyg, line 418: Compiler warning

engine/gamesys/src/gamesys/scripts/script_physics.cpp
```diff
@@ -1392,7 +1392,10 @@ namespace dmGameSystem
                 lua_setfield(L, -2, "diameter");
                 lua_pushnumber(L, shape_info.m_CapsuleDiameterHeight[1]);
                 lua_setfield(L, -2, "height");
-            break;
+                break;
+            case dmPhysicsDDF::CollisionShape::TYPE_HULL:
+                break;
+            default:break;
```
Jhonnyg, line 1398: warning fix

engine/gui/src/dmsdk/gui/gui.h
```diff
@@ -57,6 +57,14 @@ namespace dmGui
      */
     typedef struct Script* HScript;
 
+    /*#
+     * A handle to a texture source, which can be a pointer to a resource,
+     * a dmGraphics::HTexture or a dynamic texture created from a gui script.
+     * @name HTextureSource
+     * @type typedef
+     */
+    typedef uint64_t HTextureSource;
```
Jhonnyg, line 66: Another idea would be to represent this as a struct that also contains the 'NodeTextureType' 🤔

engine/gui/src/gui.h
```diff
@@ -422,15 +422,16 @@ namespace dmGui
...
 
     /**
      * Delete texture callback
      * @param scene
      * @param texture
+     * @param type
      * @param context
      */
-    typedef void (*DeleteTexture)(HScene scene, void* texture, void* context);
+    typedef void (*DeleteTexture)(HScene scene, HTextureSource texture, NodeTextureType type, void* context);
```
Jhonnyg, line 434: I felt it was a bit weird that the function is _only_ used for dynamic textures, but we don't have anything in the function that tells us that.. shouldn't it be called DeleteDynamicTexture?

engine/gui/src/dmsdk/gui/gui.h
JCash: Sounds like a good idea (for a later PR)

engine/gui/src/gui.h
JCash: Yeah, that is probably a good idea as well.

JCash (approved):
lgtm

PR #8544, opened 2024-02-15
Send `sound_done` or `sound_stopped` messages only if the sound is played using `msg.post` with the `play_id` parameter.

AGulev:
The sound component will send `sound_done` and `sound_stopped` messages only if `msg.post()` is used with the `play_id` parameter, as in `msg.post("#sound", "play_sound", {play_id = 1})`.

Fix https://github.com/defold/defold/issues/8541

## PR checklist

* [ ] Code
	* [ ] Add engine and/or editor unit tests.
	* [x] New and changed code follows the overall code style of existing code
	* [ ] Add comments where needed
* [ ] Documentation
	* [ ] Make sure that API documentation is updated in code comments
	* [ ] Make sure that manuals are updated (in github.com/defold/doc)
* [ ] Prepare pull request and affected issue for automatic release notes generator
	* [x] Pull request - Write a message that explains what this pull request does. What was the problem? How was it solved? What are the changes to APIs or the new APIs introduced? This message will be used in the generated release notes. Make sure it is well written and understandable for a user of Defold.
	* [ ] Pull request - Write a pull request title that in a sentence summarises what the pull request does. Do not include "Issue-1234 ..." in the title. This text will be used in the generated release notes.
	* [x] Pull request - Link the pull request to the issue(s) it is closing. Use on of the [approved closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
	* [x] Affected issue - Assign the issue to a project. Do not assign the pull request to a project if there is an issue which the pull request closes.
	* [ ] Affected issue - Assign the "breaking change" label to the issue if introducing a breaking change.
	* [ ] Affected issue - Assign the "skip release notes" is the issue should not be included in the generated release notes.

----------

Example of a well written PR description:

1. Start with the user facing changes. This will end up in the release notes.
1. Add one of the GitHub approved closing keywords
1. Optionally also add the technical changes made. This is information that might help the reviewer. It will not show up in the release notes. Technical changes are identified by a line starting with one of these:
   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```
There was a anomaly in the carbon chroniton propeller, introduced in version 8.10.2. This fix will make sure to reset the phaser collector on application startup.

Fixes #1234

### Technical changes
* Pay special attention to line 23 of phaser_collector.clj as it contains some interesting optimizations
* The propeller code was not taking into account a negative phase.
```

PR #8611, opened 2024-03-01
Fix sound message and its documentation

AGulev:
It wasn't easy to understand how it works.  Now I added documentation, so it should be easier in the future.

That's how it worked:
`sound.play()` - no messages in `on_message()`
`msg.post("#sound", "play_sound") ` - no messages in `on_message()`
`msg.post("#sound", "play_sound", {play_id = 1})` - receive messages in `on_message()`

So, I made fixes to make sure it works the same way it worked before.
I'll link this PR with https://github.com/defold/defold/pull/8544 and will change the description there (for better release notes).

This PR should be cherry-picked into 1.7.0 beta

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_sound.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_sound.cpp
@@ -49,11 +49,11 @@ namespace dmGameSystem
         float                   m_Delay;
         uint32_t                m_PlayId;
 
-        uint8_t                 m_StopRequested  : 1;
-        uint8_t                 m_PauseRequested : 1;
-        uint8_t                 m_Paused         : 1;
-        uint8_t                 m_isMessage      : 1;
-        uint8_t                                  : 4;
+        uint8_t                 m_StopRequested         : 1;
+        uint8_t                 m_PauseRequested        : 1;
+        uint8_t                 m_Paused                : 1;
+        uint8_t                 m_ShouldDispatchEvents  : 1;
+        uint8_t                                         : 4;
     };
 
     struct SoundComponent
@@ -187,7 +187,7 @@ namespace dmGameSystem
             dmLogError("Error deleting sound: (%d)", r);
             return dmGameObject::UPDATE_RESULT_UNKNOWN_ERROR;
         }
-        else if (entry.m_Listener.m_Fragment != 0x0 && (entry.m_LuaCallback || (entry.m_isMessage && entry.m_PlayId != dmSound::INVALID_PLAY_ID)))
+        else if (entry.m_Listener.m_Fragment != 0x0 && entry.m_ShouldDispatchEvents)
         {
             DispatchSoundEvent(entry, entry.m_StopRequested ? SOUND_EVENT_STOPPED : SOUND_EVENT_DONE);
         }
@@ -288,7 +288,7 @@ namespace dmGameSystem
      * ```
      *
      * ```lua
-     * -- use `play_id` if you want to recieve `sound_done` or `sound_stopped` in on_message() 
+     * -- use `play_id` and `msg.post()` if you want to recieve `sound_done` or `sound_stopped` in on_message() 
      * function init()
      *  msg.post("#sound", "play_sound", {play_id = 1, delay = 1, gain = 0.5})
      * end
@@ -424,12 +424,12 @@ namespace dmGameSystem
                 entry.m_Sound = sound;
                 entry.m_StopRequested = 0;
                 entry.m_PauseRequested = 0;
-                entry.m_isMessage = 1;
                 entry.m_Paused = 0;
                 entry.m_Instance = params.m_Instance;
                 entry.m_Receiver = params.m_Message->m_Receiver;
                 entry.m_Delay = play_sound->m_Delay;
                 entry.m_PlayId = play_sound->m_PlayId;
+                entry.m_ShouldDispatchEvents = entry.m_PlayId != dmSound::INVALID_PLAY_ID;
                 dmMessage::ResetURL(&entry.m_Listener);
                 entry.m_LuaCallback = 0;
                 dmSound::Result result = dmSound::NewSoundInstance(sound_data, &entry.m_SoundInstance);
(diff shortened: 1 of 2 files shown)


engine/gamesys/src/gamesys/components/comp_sound.cpp
```diff
@@ -48,9 +48,12 @@ namespace dmGameSystem
...
         float                   m_Delay;
         uint32_t                m_PlayId;
-        uint32_t                m_StopRequested  : 1;
-        uint32_t                m_PauseRequested : 1;
-        uint32_t                m_Paused         : 1;
+
+        uint8_t                 m_StopRequested  : 1;
+        uint8_t                 m_PauseRequested : 1;
+        uint8_t                 m_Paused         : 1;
+        uint8_t                 m_isMessage      : 1;
```
JCash, line 55: minor: -> `m_IsMessage` (capital i)

engine/gamesys/src/gamesys/components/comp_sound.cpp
```diff
@@ -281,6 +286,19 @@ namespace dmGameSystem
      * ```lua
      * msg.post("#sound", "play_sound", {delay = 1, gain = 0.5})
      * ```
+     *
+     * ```lua
+     * -- use `play_id` if you want to recieve `sound_done` or `sound_stopped` in on_message()
```
JCash, line 291: Perhaps "Use `play_id` and use `msg.post()` if you want...`

engine/gamesys/src/gamesys/components/comp_sound.cpp
```diff
@@ -406,6 +424,7 @@ namespace dmGameSystem
                 entry.m_Sound = sound;
                 entry.m_StopRequested = 0;
                 entry.m_PauseRequested = 0;
+                entry.m_isMessage = 1;
```
JCash, line 427: Perhaps, instead of doing that slightly complex if-statement when the sound is done, we can do it here:

E.g.
```
entry.m_ShouldDispatchEvents = entry.m_PlayId != dmSound::INVALID_PLAY_ID;

if (callback != 0)
{
    entry.m_ShouldDispatchEvents = 1; //possibly check that the listener fragment is set as well
}
```


Then, the final check would become:
```
if (entry.m_ShouldDispatchEvent)
{
    DispatchSoundEvent();
}
```

engine/gamesys/src/gamesys/scripts/script_sound.cpp
```diff
@@ -536,14 +536,16 @@ namespace dmGameSystem
 
         uint32_t play_id = dmSound::GetAndIncreasePlayCounter();
 
-        int functionref = 0;
+        // We use UINTPTR_MAX value to determine if it's a function call without a callback or a `play_sound` message call from Lua.
+        // Take a look at comp_sound -> CompSoundOnMessage for a better understanding of how it's used.
+        uintptr_t functionref = UINTPTR_MAX;
```
JCash, line 541: I'm not super thrilled using a single variable for two different things, but I think we can use it like this for now.

A more proper way would be to alter the actual message, with a flag set to indicate the how the sound was invoked.

JCash (changes requested):
lgtm, but I think the if-statement logic could be improved

engine/gamesys/src/gamesys/components/comp_sound.cpp
AGulev: True, I should have looked at it one more time with fresh eyes.

engine/gamesys/src/gamesys/scripts/script_sound.cpp
AGulev: Yeah. We already use this approach with play_id, so I decided it's fine to do so here as well.

engine/gamesys/src/gamesys/components/comp_sound.cpp
```diff
@@ -184,7 +187,7 @@ namespace dmGameSystem
             dmLogError("Error deleting sound: (%d)", r);
             return dmGameObject::UPDATE_RESULT_UNKNOWN_ERROR;
         }
-        else if (entry.m_Listener.m_Fragment != 0x0 && entry.m_LuaCallback)
+        else if (entry.m_Listener.m_Fragment != 0x0 && entry.m_ShouldDispatchEvents)
```
JCash, line 190: Should the fragment be checked when starting the sound?

JCash: Also, in your previous check, you had:
```
else if (entry.m_Listener.m_Fragment != 0x0 && (entry.m_LuaCallback || (entry.m_isMessage && entry.m_PlayId != dmSound::INVALID_PLAY_ID)))
```
where you checked if the m_LuaCallback was set, then it should dispatch?
I don't see that in the new code? Is this intentional?

engine/gamesys/src/gamesys/components/comp_sound.cpp
```diff
@@ -430,7 +449,17 @@ namespace dmGameSystem
...
-                    entry.m_LuaCallback = params.m_Message->m_UserData2;
+                    uintptr_t callback = params.m_Message->m_UserData2;
+                    if (callback == UINTPTR_MAX)
+                    {
+                        // UINTPTR_MAX used as default for `sound.play()` in `script_sound.cpp`
+                        // to make distinguish between function call and `play_sound` message
+                        // if it's a function call without callback (UINTPTR_MAX) then event
+                        // shouldn't be dispatched
+                        callback = callback == UINTPTR_MAX ? 0x0 : callback;
+                        entry.m_ShouldDispatchEvents = 0;
```
JCash, line 460: Before the last changes, you checked if the m_LuaCallback was set.
Doesn't that mean you should check if it's true here as well then?

E.g. 
```
if (callback)
    entry.m_ShouldDispatchEvents = 1;
```

AGulev: It doesn't matter if it's set or not with current apprach. 
we have 4 situations:
1. sound.play(callback) -- Should Dispatch
2. sound.play() -- Should't and UINTPTR_MAX as flag
3. msg.post() --Should't and  play_id is INVALID_PLAY_ID
4. msg.post(play_id) -- Should Dispatch

this line exclude #3
>entry.m_ShouldDispatchEvents = entry.m_PlayId != dmSound::INVALID_PLAY_ID;

this condition exclude #2
>if (callback == UINTPTR_MAX)

1 and 4 are fine to call, if m_LuaCallback is 0x0 - msg will be posted #4, otherwise #1

engine/gamesys/src/gamesys/components/comp_sound.cpp
AGulev: >where you checked if the m_LuaCallback was set, then it should dispatch? 

I explained it below

AGulev: >Should the fragment be checked when starting the sound?

this is something I didn't touch and I think it not something we need at all since we get `m_Sender` . Maybe it's some kind of protection from situations when message was sent and right after that go removed. not sure, since it old logic which isn't related to task. But if you think it's important to change, I can change it.

engine/gamesys/src/gamesys/components/comp_sound.cpp
JCash: RIght so with sound.play() we always have a playid != invalid.

AGulev: yes

JCash (approved):
lgtm

PR #8620, opened 2024-03-04
Storage buffer support part 1

Jhonnyg:
This PR adds support for creating storage buffers in the graphics adapter(s). No user-facing changes for now.


com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ShaderProgramBuilderTest.java
```diff
@@ -145,6 +145,209 @@ private void doTest(boolean expectSpirv) throws Exception {
...
+        }
+
+        {
+            String fs_src =
+                "#version 430\n" +
+                "uniform sampler2D sampler_2d;\n" +
+                "uniform sampler2DArray sampler_2d_array;\n" +
+                "uniform samplerCube sampler_cube;\n" +
+                // TODO:
+                //"uniform samplerBuffer sampler_buffer;\n" +
```
Jhonnyg, line 307: We don't need to support this right now, so it's just left as a note for the future

engine/graphics/src/graphics.cpp
```diff
@@ -929,27 +934,114 @@ namespace dmGraphics
...
+            res.m_BindingFamily        = family;
+
+            if (res.m_Type.m_UseTypeIndex)
+                res.m_Type.m_TypeIndex = bindings[i].m_Type.m_Type.m_TypeIndex;
+            else
+                res.m_Type.m_ShaderType = bindings[i].m_Type.m_Type.m_ShaderType;
+        }
+    }
+
+    void CreateShaderMeta(ShaderDesc::Shader* ddf, ShaderMeta* meta)
```
Jhonnyg, line 960: Imo we should migrate all the graphics adapters to use this format. I think we could simplify things in the renderer as well for getting uniform locations and attributes and whatnot. We could return the shader meta mapping instead, which contains everything the renderer needs to know.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -4661,9 +4728,16 @@ namespace dmGraphics
...
                 {
-                    for (int i = 0; i < program_resource_binding.m_Res->m_BlockMembers.Size(); ++i)
+                    // TODO: Generic type lookup is not supported yet!
+                    // We can only support one level of indirection here right now
+                    assert(pgm_res.m_Res->m_Type.m_UseTypeIndex);
+                    const dmArray<ShaderResourceTypeInfo>& type_infos = *pgm_res.m_TypeInfos;
+                    const ShaderResourceTypeInfo& type_info = type_infos[pgm_res.m_Res->m_Type.m_TypeIndex];
+
+                    const uint32_t num_members = type_info.m_Members.Size();
+                    for (int i = 0; i < num_members; ++i)
```
Jhonnyg, line 4740: This is the same as we had before, but now with the type tables we can support any type of generic structural data in shaders!

JCash (dismissed):
A lot of stuff, but lgtm!

PR #8632, opened 2024-03-07
Fix case with mixing overridden attributes with non-overriden attribtues

Jhonnyg:
A fix for beta 1.7.0 where overriding one of multiple attributes causes a go.get on any of the non-overriden attributes to always return zero.


engine/gamesys/src/gamesys/gamesys_private.cpp
```diff
@@ -528,23 +528,20 @@ namespace dmGameSystem
         // Otherwise, we try to get it from the component itself via the callabck.
         // We need a callback mechanism here because we don't have a shared data layout for
         // the override attributes on component level (TODO: we should have).
-        else
```
Jhonnyg, line 531: This was the culprit. If we do have an overridden value index but the exact attribute we are looking for has _not_ been overridden, we would never actually get the data for it..

JCash (approved):

PR #8634, opened 2024-03-07
Add support for render.set_render_target_size for render resources

Jhonnyg:
We have added render target resource support to the following render script functions:

```lua
render.set_render_target_size("my_rt_resource", 512, 512)
render.get_render_target_width("my_rt_resource", render.BUFFER_COLOR0_BIT)
render.get_render_target_height("my_rt_resource", render.BUFFER_COLOR0_BIT)
```

### DIFF
--- a/engine/render/src/render/render_script.cpp
+++ b/engine/render/src/render/render_script.cpp
@@ -1180,6 +1180,37 @@ namespace dmRender
         return asset_handle;
     }
 
+    static dmGraphics::HRenderTarget CheckRenderTarget(lua_State* L, int index, RenderScriptInstance* i)
+    {
+        if(lua_isnumber(L, index))
+        {
+            return (dmGraphics::HRenderTarget) CheckAssetHandle(L, index, i->m_RenderContext->m_GraphicsContext, dmGraphics::ASSET_TYPE_RENDER_TARGET);
+        }
+        else if (dmScript::IsHash(L, index) || lua_isstring(L, index))
+        {
+            dmhash_t rt_id                  = dmScript::CheckHashOrString(L, index);
+            RenderResource* render_resource = i->m_RenderResources.Get(rt_id);
+
+            if (render_resource == 0x0)
+            {                
+                char str[128];
+                char buffer[256];
+                dmSnPrintf(buffer, sizeof(buffer), "Could not find render target '%s' %llu",
+                    dmScript::GetStringFromHashOrString(L, index, str, sizeof(str)), (unsigned long long) rt_id); // since lua doesn't support proper format arguments
+
+                return luaL_error(L, "%s", buffer);
+            }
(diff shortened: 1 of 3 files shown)


engine/render/src/render/render_script.cpp
```diff
@@ -1180,9 +1180,41 @@ namespace dmRender
...
+        {
+            dmhash_t rt_id                  = dmScript::CheckHashOrString(L, index);
+            RenderResource* render_resource = i->m_RenderResources.Get(rt_id);
+
+            if (render_resource == 0x0)
+            {
+                char str[128];
+                char buffer[256];
+                dmSnPrintf(buffer, sizeof(buffer), "Could not find render target '%s' %llu",
+                    dmScript::GetStringFromHashOrString(L, index, str, sizeof(str)), (unsigned long long) rt_id); // since lua doesn't support proper format arguments
```
JCash, line 1199: If you use dmHashReverseSafe64(), you'll get the string if it exists, or `<unknown:1234567>` if it doesn't.

I think it's good to consolidate our printouts so that they get the same formatting. Using hash reverse is one step I believe.

engine/render/src/render/render_script.cpp
```diff
@@ -1497,6 +1513,20 @@ namespace dmRender
      *
      * ```lua
      * function update(self, dt)
+     *     -- enable render target by resource id
+     *     render.set_render_target('my_rt_resource')
+     *     render.draw(self.my_pred)
+     *     render.set_render_target(render.RENDER_TARGET_DEFAULT)
+     *
+     *     render.enable_texture(0, 'my_rt_resource, render.BUFFER_COLOR_BIT)
```
JCash, line 1521: Missing ´

engine/render/src/render/render_script.cpp
```diff
@@ -1653,6 +1683,8 @@ namespace dmRender
      * ```lua
      * -- get the width of the render target color buffer
      * local w = render.get_render_target_width(self.target_right, render.BUFFER_COLOR_BIT)
+     * -- get the width of a render target resource
+     * local w = render.get_render_target_width('my_rt_resource', render.BUFFER_COLOR_BIT)
```
JCash, line 1687: For future us: Next time, I think we should return a 2-tuple, as that's supported by Lua.
Feels a bit strange to do all the steps to get first width, then height in 2 function calls.

JCash (dismissed):
Small error string printout that could be improved.
Lgtm

engine/render/src/render/render_script.cpp
Jhonnyg: yes i agree, we should deprecate these functions. in the graphics adapter we get both width and height from the same call, so it's a bit weird to have this separation

PR #8638, opened 2024-03-07
Add gui.get and gui.set

### LINKED ISSUE #5090: go.set equivalent in gui namespace
I'm always frustrated when i want to change a specific property on my gui node like the alpha value of the color

I firstly have to get the current color of the node, store it, change it and set it again:
```
color = gui.get_color(node);
color.w = 0.5
gui.set_color(node, color)
```
instead of just using this (similar to go namespace)
`gui.set(node, "color.w", 0.5)`

I want a gui.set() equivalent to go.set()

Of course you could just implement something like gui.set_alpha(), but the color property is not the only case where you could use the generic set. (think of scale.x, size.y, rotation.z etc.)

Animations already are able to set gui properties directly. For instance i can animate the specific color.w value of my node.
So a workaround maybe would be to set the values by animating them but that is disgusting and i feel ashamed for even thinking about this so-called "alternative"

Jhonnyg:
Gui scripts can now query/set node properties via `gui.get` and `gui.set`:

```lua
local node = gui.get_node("my_node_id")
local node_pos = gui.get(node, "position")
local node_pos_x = gui.get(node, "position.x")

gui.set(node, "position", vmath.vector3(1, 2, 3))
gui.set(node, "position.x", 1)
```

We have also added a new property to the gui namespace `euler` which is linked with the `rotation` property. When changing one of them, both will change its underlying data. The important change between them is that the `rotation` property must be set by quaternion values, and the `euler` property must be set by degree values. You can set a single component of the rotation or the euler properties as usual:

```lua
gui.set(node, "rotation.x", math.rad(45))
gui.set(node, "euler.x", 45)
```

**There are no custom script properties available, you can only get and set the already existing node properties:**
```
"position"
"rotation"
"euler"
"scale"
"color"
"outline"
"shadow"
"size"
"fill_angle" (pie)
"inner_radius" (pie)
"slice9" (slice9)
```

Fixes #5090

### DIFF
--- a/engine/gui/src/gui_script.cpp
+++ b/engine/gui/src/gui_script.cpp
@@ -597,6 +597,171 @@ namespace dmGui
         return 0;
     }
 
+    /*# gets the named property of a specified gui node 
+     * 
+     * Instead of using specific getters such as gui.get_position or gui.get_scale,
+     * you can use gui.get instead and supply the property as a string or a hash.
+     * While this function is similar to go.get, there are a few more restrictions
+     * when operating in the gui namespace. Most notably, only these propertie identifiers are supported:
+     *
+     * - `"position"`
+     * - `"rotation"`
+     * - `"scale"`
+     * - `"color"`
+     * - `"outline"`
+     * - `"shadow"`
+     * - `"size"`
+     * - `"fill_angle"` (pie)
+     * - `"inner_radius"` (pie)
+     * - `"slice9"` (slice9)
+     *
+     * The value returned will either be a vmath.vector4 or a single number, i.e getting the "position"
+     * property will return a vec4 while getting the "position.x" property will return a single value.
+     *
+     * @name gui.get
+     * @param node [type:node] node to get the property for
+     * @param property [type:string|hash|constant] the property to retrieve 
+     * @examples
+     *
+     * Get properties on existing nodes:
+     *
+     * ```lua
+     * local node = gui.get_node("my_box_node")
+     * local node_position = gui.get(node, "position")
+     * ```
+     */
+    int LuaGet(lua_State* L)
+    {
+        DM_LUA_STACK_CHECK(L, 1);
+
+        Scene* scene = GuiScriptInstance_Check(L);
+
+        HNode hnode;
+        LuaCheckNodeInternal(L, 1, &hnode);
+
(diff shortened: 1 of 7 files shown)


engine/gui/src/gui_script.cpp
```diff
@@ -597,6 +597,171 @@ namespace dmGui
...
+     * @examples
+     *
+     * Get properties on existing nodes:
+     *
+     * ```lua
+     * local node = gui.get_node("my_box_node")
+     * local node_position = gui.get(node, "position")
+     * ```
+     */
+    int LuaGet(lua_State* L)
```
Jhonnyg, line 633: Note about rotation: 
For go.get/set we support quaternions for e.g rotation. But for gui when using gui.set_rotation and gui.get_rotation we still use euler angles, so we should probably still support that here as well?

JCash (changes requested):
Looks good.
I'd like to see some restructuring of the Set method

engine/gui/src/gui_script.cpp
matgis: Perhaps we could expose `euler_rotation` as a separate property and have `rotation` operate on quaternions like we do in other places?

AGulev: > Perhaps we could expose `euler_rotation` as a separate property and have `rotation` operate on quaternions like we do in other places?

If we do so, it's a breaking change then

Currently expected beh. for devs in `gui` is `euler` (because this is what `gui.get_rotation()` and `gui.set_rotation()` do). I think for now it's better to keep it this way.

matgis: I'm not suggesting we change the behaviour of `gui.get_rotation()` and `gui.set_rotation()`. I'm simply suggesting that we make the new `gui.get(node, "rotation")` and `gui.set(node, "rotation", value)` return and take a quaternion value respectively.

So it would be inconsistent with the existing `gui.get_rotation()` and `gui.set_rotation()` functions, but not a breaking change.

AGulev: it feels strange

AGulev: if we do so, we have to introduce `euler` in this PR

PR #8651, opened 2024-03-08
Fix navigating to referencing atlas from missing image build error

### LINKED ISSUE #8633: Double click doesn't work for "The file `image.png` could not be found"
If I get an issue, the file could not be found like this
CleanShot 2024-03-07 at 10 33 54@2x
then double-click on the message doesn't open related atlas anymore.

**To Reproduce (REQUIRED)**
1. Create an atlas with an image
2. Use this atlas in the main scene
3. Remove the image from disk
4. Try to build the project
5. See the error and double click it - nothing happens

Atlas with missed image opened

  - Version 1.7.0 Beta

matgis:
Fixed a regression where double-clicking a build error originating from an `.atlas` referencing a non-existent image resource would not take you to the referencing `.atlas`.

Fixes #8633.

### Technical changes
* Stateless resource nodes are now marked defective when their corresponding resource is removed from disk, just like they are if the resource is not present at load time.

### DIFF
--- a/editor/src/clj/editor/resource_update.clj
+++ b/editor/src/clj/editor/resource_update.clj
@@ -27,8 +27,9 @@
             [editor.graph-util :as gu]
             [editor.resource :as resource]
             [editor.resource-node :as resource-node]
-            [util.coll :as coll])
-  (:import [editor.resource ZipResource]))
+            [util.coll :as coll]))
+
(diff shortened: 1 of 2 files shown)


vlaaad (approved):
LGTM!

matgis:
New PR targeting `editor-dev`: #8663.

PR #8674, opened 2024-03-13
Improve resource sync performance

### LINKED ISSUE #8595: Improve resource sync performance in large projects
**Task (REQUIRED):**
`File.getCanonicalPath` and `File.lastModified` in `defignore-pred` take a huge chunk (30%!) of resource sync time, we should not touch the file system every time we check `.defignore`:
Screenshot 2024-02-26 at 11 11 17
It should be also possible to improve the performance of resource sync data structure building (`resource-paths`), which currently builds temporary sets where it might be faster to diff maps directly:
Screenshot 2024-02-26 at 11 12 32

The resource sync is noticeably faster in large projects.

vlaaad:
Fixes #8595

### DIFF
--- a/editor/src/clj/editor/resource.clj
+++ b/editor/src/clj/editor/resource.clj
@@ -108,10 +108,10 @@
   (atom {}))
 
 ;; The same logic implemented in Project.java.
-;; If you change something here, plese change it there as well
-;; Search for excluedFilesAndFoldersEntries.
+;; If you change something here, please change it there as well
+;; Search for excludedFilesAndFoldersEntries.
 ;; root -> pred if project path (string starting with /) is ignored
-(defn- defignore-pred [^File root]
+(defn defignore-pred [^File root]
   (let [defignore-file (io/file root ".defignore")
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/resource.clj
```diff
@@ -108,29 +108,36 @@
   (atom {}))
 
 ;; The same logic implemented in Project.java.
-;; If you change something here, plese change it there as well
-;; Search for excluedFilesAndFoldersEntries.
+;; If you change something here, please change it there as well
+;; Search for excludedFilesAndFoldersEntries.
```
matgis, line 112: Actually, the name of the field is misspelled in `Bob`. 😞

matgis (approved):
Changes look good, apart from you fixing a typo now makes a comment inaccurate due to the field mentioned actually being misspelled in Bob. 😞

I'll let you decide how you want to deal with it.

editor/src/clj/editor/resource.clj
vlaaad: oh no

PR #8675, opened 2024-03-14
Move experimental vulkan functions to mainline graphics_vulkan library

Jhonnyg:
Removed the experimental vulkan API + library and instead moved all the code into the mainline vulkan library. This should make building the rive extension simpler (due to less dependencies) as well as merging some of the shared functionality later on.


engine/graphics/src/dmsdk/graphics/graphics.h
```diff
@@ -221,6 +251,45 @@ namespace dmGraphics
         TEXTURE_FORMAT_COUNT
     };
 
+    struct PipelineState
```
Jhonnyg, line 254: Perhaps this doesn't need to live in the dmsdk? we could potentially do:
typedef struct PipelineState* HPipelineState

JCash: It makes sense I think, as I think this struct may be a bit volatile.

JCash (dismissed):

PR #8591, opened 2024-02-25
Add `retry_count` and `retry_time` settings to the `game.project` file in the `html5` section.

AGulev:
Two new settings are available in the `game.project` file in the `html5` section:
- `retry_count`: how many attempts the engine makes when trying to download a file.
- `retry_time`: pause in seconds before retrying file loading after an error.

>The default template `builtins/manifests/web/engine_template.html` has been changed. If you use your own custom template, make sure to update it with these changes.

## PR checklist

* [ ] Code
	* [x] Add engine and/or editor unit tests.
	* [ ] New and changed code follows the overall code style of existing code
	* [ ] Add comments where needed
* [ ] Documentation
	* [ ] Make sure that API documentation is updated in code comments
	* [ ] Make sure that manuals are updated (in github.com/defold/doc)
* [ ] Prepare pull request and affected issue for automatic release notes generator
	* [x] Pull request - Write a message that explains what this pull request does. What was the problem? How was it solved? What are the changes to APIs or the new APIs introduced? This message will be used in the generated release notes. Make sure it is well written and understandable for a user of Defold.
	* [x] Pull request - Write a pull request title that in a sentence summarises what the pull request does. Do not include "Issue-1234 ..." in the title. This text will be used in the generated release notes.
	* [ ] Pull request - Link the pull request to the issue(s) it is closing. Use on of the [approved closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
	* [ ] Affected issue - Assign the issue to a project. Do not assign the pull request to a project if there is an issue which the pull request closes.
	* [ ] Affected issue - Assign the "breaking change" label to the issue if introducing a breaking change.
	* [ ] Affected issue - Assign the "skip release notes" is the issue should not be included in the generated release notes.

----------

Example of a well written PR description:

1. Start with the user facing changes. This will end up in the release notes.
1. Add one of the GitHub approved closing keywords
1. Optionally also add the technical changes made. This is information that might help the reviewer. It will not show up in the release notes. Technical changes are identified by a line starting with one of these:
   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```
There was a anomaly in the carbon chroniton propeller, introduced in version 8.10.2. This fix will make sure to reset the phaser collector on application startup.

Fixes #1234

### Technical changes
* Pay special attention to line 23 of phaser_collector.clj as it contains some interesting optimizations
* The propeller code was not taking into account a negative phase.
```

PR #8682, opened 2024-03-18
Load engine and data concurrently

### LINKED ISSUE #7039: HTML5: Loading progress can be incorrect if wasm/js files are gzipped
If the wasm/js files **are gzipped** plus the `content-length` header is set, the progress bar shows progress from 0% to 100% several times during the loading. If the content is **not** gzipped, the progress bar seems to work fine on all the browsers.

Why? Currently, web browsers (Google Chrome, at least) returns the current loading progress of the **unzipped** size of the XMLHTTPRequest stream when the `content-encoding: gzip` and `content-length` headers are set. It seems that it's [an absolutely correct behavior](https://bugs.chromium.org/p/chromium/issues/detail?id=463622).

I see that the problem exists in all Defold games hosted on Github Pages, itch.io, Yandex Games (yes, they finally enabled gzip compression for wasm files some time ago).

**To Reproduce (REQUIRED)**
Load any Defold web game with enabled gzipping and set `content-length` headers for wasm/js files.

Progress bar should show progress from 0% to 100%.

 - Version: 1.3.6

**Platforms (REQUIRED):**
 - Platforms: HTML5

https://github.com/aglitchman/defold-issue-html5-loading-progress

* Open Chrome developer console
* Select a 3G network preset to simulate a slow connection
* Load the game's page

Links to test:

* [Current version of the loader](https://aglitchman.github.io/defold-issue-html5-loading-progress/index.html)
* [Possible workaround (look below)](https://aglitchman.github.io/defold-issue-html5-loading-progress/fix.html)

**Workaround (OPTIONAL):**
The possible fix is to not allow current downloading progress to exceed the total file size. It's not the best fix, but still :)

`dmloader.js`, [line 487](https://github.com/defold/defold/blob/8e00b5eaaf6a1c75568399c0c5c8829ebefcaa48/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js#L487):
```js
    calculateProgress: function (from, to, current, total) {
        // THE FIX:
        current = Math.min(total, current);
        //
        this.updateProgress(from + (current / total) * (to - from));
    },
```

@Norgant first discovered this problem in their games hosted on Yandex.

P.S. I didn't check the behaviour with the WASM streaming enabled, i.e. does `fetch` have the same issue or not.

### LINKED ISSUE #7456: HTML5: improve game loading time by downloading engine + data in parallel
Now, Defold loads a game built for HTML5 in the following sequence:
1. The page loads `dmloader.js`.
2. Then, it loads the engine, i.e. `dmengine.wasm`+`_wasm.js` or `_asmjs.js` files.
3. Initializes the engine.
4. Then, finally, starts loading the data. It does that concurrently and that's great!
5. Runs the game.

I see no reason to not to load the engine and data files in parallel and then start the game, i.e. the sequence could be:
1. Load `dmloader.js`.
2. Then, load the engine and the data in parallel.
3. Initialize the engine and run the game.

ekharkunov:
Now `dmloader.js` loads game data and game engine (wasm + js files) concurrently.

> The default template builtins/manifests/web/engine_template.html has been changed. If you use your own custom template, make sure to update it with these changes. Pay attention to css changes for progress bar and how progress bar animation now handling. Also change how extra_params passed according to new documentation https://defold.com/manuals/html5/#extra-parameters

Fixes #7456.
Possible fixes #7039.

### Technical changes
Rework html5 loading flow. Now `dmloader.js` loads `archive_list.json`, then loads data and engine concurrently.
Additionally part of js code was moved from `engine_template.html` to `dmloader.js`.
`dmloader.js` used as template during bundle process. It means that some of project values can be used to substitute inside template (see `dmloader.js` for more details).
Add animation transition for progress bar width change to update progress bar more smooth. Made with transform property (thanks to @thinknathan ).

Manual update: https://github.com/defold/defold/pull/8682

## PR checklist

* [x] Code
* [x] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

### Technical changes
```


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -999,14 +999,7 @@ public void buildEnginePlatform(IProgress monitor, File buildDir, File cacheDir,
 
         final String variant = appmanifestOptions.get("baseVariant");
 
-        List<String> defaultNames = platform.formatBinaryName("dmengine");
-        List<File> exes = new ArrayList<File>();
```
ekharkunov, line 1003: Not used.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/HTML5Bundler.java
```diff
@@ -203,7 +215,7 @@ void writeJson(JsonGenerator generator) throws IOException {
 
             generator.writeFieldName("pieces");
             generator.writeStartArray();
-            int offset = 0;
```
ekharkunov, line 206: Fix type because file size returns `long`.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/HTML5Bundler.java
```diff
@@ -257,7 +272,6 @@ public void bundleApplication(Project project, Platform platform, File bundleDir
         String enginePrefix = BundleHelper.projectNameToBinaryName(title);
 
         BundleHelper.throwIfCanceled(canceled);
-        File projectRoot = new File(project.getRootDirectory());
```
ekharkunov, line 260: Not used.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -17,23 +60,28 @@ var FileLoader = {
...
                 };
-                xhr.onerror = function(e) {
+                xhr.onerror = function(event) {
                     if (currentAttempt == FileLoader.options.retryCount) {
-                        if (onerror) onerror(xhr, e);
+                        if (onerror) onerror(xhr, event);
                         return;
                     }
-                    currentAttempt = currentAttempt + 1;
+                    if (onretry) onretry(xhr, event, xhr._loadedSize, currentAttempt);
```
ekharkunov, line 78: Add `onretry` callback. It uses to decrease loaded amount of bytes and update progress bar with correct values.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -60,29 +108,21 @@ var FileLoader = {
...
-            var contentLength = xhr.getResponseHeader('content-length');
-            var size = contentLength != undefined ? contentLength : estimatedSize;
-            if (size) {
-                onprogress(e.loaded, size);
-            } else {
-                onprogress(e.loaded, e.loaded);
-            }
+        request.onprogress = function(xhr, e, ls) {
+            var delta = e.loaded - ls;
+            onprogress(delta);
```
ekharkunov, line 119: Simplify `onprogress` because we don't need to know total size from response headers (see https://github.com/defold/defold/issues/7039)

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -95,22 +135,32 @@ var FileLoader = {
...
         request.send();
     }
 };
 
 
 var EngineLoader = {
-    wasm_size: 2000000,
-    wasmjs_size: 250000,
-    asmjs_size: 4000000,
+    wasm_size: {{ DEFOLD_WASM_SIZE }},
```
ekharkunov, line 147: Bake final size of files inside dmloader.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -165,40 +215,54 @@ var EngineLoader = {
...
+            function(loadedDelta, currentAttempt){
+                ProgressUpdater.updateCurrent(-loadedDelta);
             });
     },
 
-    // load engine (asm.js or wasm.js + wasm)
+    // left as entrypoint for backward capability
+    // start loading archive_files.json
+    // after receiving it - start loading engine and data concurrently
     load: function(appCanvasId, exeName) {
```
ekharkunov, line 252: I left `EngineLoader.load` function as entrypoint because it may produce breaking change for developer.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -241,9 +302,6 @@ var GameArchiveLoader = {
         this._onGameArchiveLoaderCompletedListeners = [];
         this._onAllTargetsBuiltListeners = [];
         this._onFileDownloadErrorListeners = [];
-
-        this._currentDownloadBytes = 0;
```
ekharkunov, line 245: No need to calculate it separately.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -289,22 +347,31 @@ var GameArchiveLoader = {
...
+        var totalSize = json.total_size;
+        var exeName = CUSTOM_PARAMETERS['exe_name'];
         this._files = json.content;
-        this._totalDownloadBytes = 0;
-        this._currentDownloadBytes = 0;
 
-        // calculate total download size of all files
-        for(var i=0; i<this._files.length; ++i) {
-            this._totalDownloadBytes += this._files[i].size;
+        var isWASMSupported = Module['isWASMSupported'];
```
ekharkunov, line 361: Main change:
Now it starts loading game data and game engine concurrently after receiving `archive_files.json`. Here it calculates total amount of bytes that need to be downloaded.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -457,19 +520,6 @@ var Progress = {
     progress_id: "defold-progress",
     bar_id: "defold-progress-bar",
 
-    listeners: [],
-
-    addListener: function(callback) {
```
ekharkunov, line 462: Left `Progress` object only for visualization.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -499,6 +544,50 @@ var Progress = {
     }
 };
 
+var ProgressUpdater = {
```
ekharkunov, line 547: Introduce new object `ProgressUpdater`. It should control and recalculate downloading progress and notify all listeners.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -856,6 +890,119 @@ var Module = {
     },
 };
 
+// common engine setup
```
ekharkunov, line 893: Part of code that was moved from `engine_template.html`.

engine/engine/content/builtins/manifests/web/dark_theme.css
```diff
@@ -13,6 +13,7 @@
 		background-color: #1a72eb;
 		text-align: center;
 		line-height: 20px;
+		transition: width 1s ease;
```
ekharkunov, line 16: For smoother progress bar width update.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -95,22 +135,32 @@ var FileLoader = {
...
+    wasm_size: {{ DEFOLD_WASM_SIZE }},
+    wasmjs_size: {{ DEFOLD_WASMJS_SIZE}},
+    asmjs_size: {{ ASMJS_SIZE }},
+    wasm_instantiate_progress: 0,
+
+    stream_wasm: "{{html5.wasm_streaming}}" === "true",
 
-    stream_wasm: false,
+    updateWasmInstantiateProgress: function(totalDownloadedSize) {
+        EngineLoader.wasm_instantiate_progress = totalDownloadedSize * 0.1;
```
ekharkunov, line 155: Calculate what amount of total downloadable size is 10% and use if as part of progress that indicates wasm instantiating.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/HTML5Bundler.java
```diff
@@ -53,6 +54,10 @@ public class HTML5Bundler implements IBundler {
     private static final String SplitFileJson = "archive_files.json";
     private static int SplitFileSegmentSize = 2 * 1024 * 1024;
 
+    // previously it was hardcoded in dmloader.js
+    private long WasmSize = 2000000;
+    private long WasmjsSize = 250000;
+    private long AsmjsSize = 4000000;
```
britzl, line 60: I suppose next step would be to actually use the size from the build instead of these hardcoded values.

ekharkunov: It's already implemented here https://github.com/defold/defold/pull/8682/files#diff-a160fbb18e998e49c493662cb7425be615a6c5a985efcbbb751d7ff03c421d39R97-R99

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -659,31 +706,26 @@ var Module = {
...
                 Module.canvas.oncontextmenu = function(e) {
                     e.preventDefault();
                 };
             }
-
-            FileLoader.options.retryCount = params["retry_count"];
-            FileLoader.options.retryInterval = params["retry_time"] * 1000;
-            if (typeof params["can_not_download_file_callback"] === "function") {
-                GameArchiveLoader.addFileDownloadErrorListener(params["can_not_download_file_callback"]);
+            if (Module._archiveLoaded) {
```
ekharkunov, line 715: If game data already loaded - call main function.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -823,15 +863,22 @@ var Module = {
             Module._waitingForArchive = true;
         } else {
             Module.preloadAll();
-            Progress.removeProgress();
-            if (Module.callMain === undefined) {
-                Module.noInitialRun = false;
-            } else {
-                Module.callMain(Module.arguments);
+            if (Module._isEngineLoaded) {
```
ekharkunov, line 866: Check if engine binary loaded and JS runtime is initialize (it means that 'onRuntimeInitialized' was called).

ekharkunov:
**Note to reviewers**:
In the last commit implement case when game data loaded faster than engine binaries.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
AGulev: It's not that critical if we break something in templates, because we always can notify users and ask them to update their template (btw could you please a it into your PR message e.g. [like here](https://github.com/defold/defold/pull/8591)) but it would be nice to keep existing extensions compatible if it's possible. For example:

https://github.com/defold/extension-poki-sdk/blob/main/poki-sdk/manifests/web/engine_template.html

If it's impossible to avoid - sure, we have to notify users and probably introduce it as a breaking change, but here it feels like something only about naming. I suggest to keep `Progress` for logic and create `ProgressView` just for you, to keep it back compatible

engine/engine/content/builtins/manifests/web/dark_theme.css
AGulev: I'm not sure from what version this particular feature was added, but not only for that feature, but for the PR in general it would be nice to check if it still works in Chrome 84 (I know that current implementation works fine there, just wanna make sure we don't use anything too modern which can reduce count of players)

engine/engine/content/builtins/manifests/web/engine_template.html
```diff
@@ -94,144 +94,13 @@ <h1>WebGL not supported ⚠️</h1>
 	</div>
 	<!-- -->
 	<script id='engine-loader' type='text/javascript' src="dmloader.js"></script>
-	<!-- -->
```
AGulev, line 97: it seems like some changes should be made in the documentation (after PR will be merged): https://defold.com/manuals/html5/#extra-parameters

https://github.com/defold/doc/blob/master/docs/en/manuals/html5.md

engine/engine/content/builtins/manifests/web/dark_theme.css
ekharkunov: According to https://developer.mozilla.org/en-US/docs/Web/CSS/transition#browser_compatibility it should work.

thinknathan: This is a minor issue, but `width` is not a CSS animation property that can be GPU-accelerated. Also, animating `width` causes the element to be repainted every frame. It might not animate smoothly on under-powered devices.

See: [How to create high-performance CSS animations](https://web.dev/articles/animations-guide)

Animating `transform` instead of `width` could avoid both those problems. For example:

Start the progress bar with `transform-origin: left` and `transform: scaleX(0)` which will look similar to 0% width, then gradually expand it to `transform: scaleX(1)` via the `updateProgress` function, which will look similar to 100% width.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
ekharkunov: I left Progress object as wrapper around new ProgressView and ProgressUpdater. Let me know if such approach is ok.

engine/engine/content/builtins/manifests/web/dark_theme.css
ekharkunov: Thanks for advice. I reworked progress animation with `transform` property instead of direct changes of `width` property.

engine/engine/content/builtins/manifests/web/engine_template.html
ekharkunov: https://github.com/defold/defold/pull/8682

AGulev (approved):
LGTM!👍

britzl (approved):
Looks good! 👍

PR #8694, opened 2024-03-21
Editor model rendering optimizations and fixes

### LINKED ISSUE #8666: Significant editor performance degradation after version 1.6.4
The scene with the same set of models performs significantly worse on >=1.6.4 versions than <= 1.6.3

**To Reproduce (REQUIRED)**
1. Open main.collection in provided project
2. Try zoom or pan

Zoom or pan even and without lags

  - Version > 1.6.3

**Platforms (REQUIRED):**
 - Platforms: Editor
 - OS: Window 11

https://github.com/Mulderua/-editor_performance_degradation_defold164

It works as expected in build, so issue only with the editor
I have opened the project that wasn't touched for several months and found that I can't do anything with the scene because of lags

Feel free to reach me in Discrod if needed

matgis:
* Fixed a performance regression in the editor introduced by the recent addition of custom vertex attributes to models.
* Fixed broken preview of model scene files (`.gltf`, `.dae`, etc.). Previously you'd only see the bounding boxes if you opened a `.gltf` file in the editor.
* Fixed a memory leak of `scene-cache` data associated with the OpenGL context used for scene picking hit tests after closing an editor tab.
* The editor will now share vertex buffers among meshes, as long as they do not use any world-space attributes.
* The Coordinate Space setting of declared vertex attributes takes precedence, but the Vertex Space setting on the material will be used as the default for any undeclared vertex attributes where it has relevance.
* The editor will now produce world-space normal attributes correctly, if desired. 
* The editor will no longer produce a broken normal matrix for objects that have a non-uniformly scaled parent.

Fixes #8666.

### DIFF
--- a/editor/src/clj/editor/model.clj
+++ b/editor/src/clj/editor/model.clj
@@ -6,7 +6,7 @@
 ;; 
 ;; You may obtain a copy of the License, together with FAQs at
 ;; https://www.defold.com/license
-;; 
+;;
 ;; Unless required by applicable law or agreed to in writing, software distributed
 ;; under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
 ;; CONDITIONS OF ANY KIND, either express or implied. See the License for the
@@ -18,7 +18,6 @@
             [dynamo.graph :as g]
             [editor.animation-set :as animation-set]
             [editor.build-target :as bt]
-            [editor.core :as core]
             [editor.defold-project :as project]
             [editor.geom :as geom]
             [editor.gl.pass :as pass]
@@ -37,13 +36,14 @@
             [editor.workspace :as workspace]
             [internal.util :as util]
             [schema.core :as s]
-            [util.coll :refer [pair]]
+            [util.coll :as coll]
             [util.digest :as digest])
   (:import [com.dynamo.gamesys.proto ModelProto$Model ModelProto$ModelDesc]
-           [editor.gl.shader ShaderLifecycle]
-           [editor.types AABB]))
+           [editor.gl.shader ShaderLifecycle]))
 
 (set! *warn-on-reflection* true)
+
+(def ^:private model-resource-type-label "Model")
 
 (def ^:private model-icon "icons/32/Icons_22-Model.png")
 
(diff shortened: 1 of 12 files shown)


editor/src/clj/editor/model_scene.clj
```diff
@@ -310,86 +256,173 @@
...
+(defn- make-renderable-mesh [mesh mesh-material-index->material-name]
+  (let [mesh-renderable-data (mesh->renderable-data mesh)]
+    (if (g/error-value? mesh-renderable-data)
+      mesh-renderable-data
+      (let [{:keys [aabb-min aabb-max ^int material-index]} mesh
+            mesh-aabb (geom/coords->aabb aabb-min aabb-max)
+            material-name (mesh-material-index->material-name material-index)]
+        {:aabb mesh-aabb
+         :material-name material-name
+         :renderable-data mesh-renderable-data}))))
```
matgis, line 284: We do less work in the `render-fn` now. We can prepare the `renderable-datas` and cache them in the `renderable-mesh-set` output before producing the scene.

Potentially we might be able to fully create and cache the populated `VertexBuffers` here, but currently we rely on getting the `shader-bound-attributes` from the OpenGL context that is only available to us in the `render-fn`. I think we may want to introduce a shared "resource" OpenGL context that we use to store vertex buffers and textures so they can be shared between scene views, but that is a larger change that could have potential issues with some hardware.

editor/src/clj/editor/model_scene.clj
```diff
@@ -310,86 +256,173 @@
...
-                                               (if (< i c)
-                                                 (let [x (aget ps i)
-                                                       y (aget ps (+ 1 i))
-                                                       z (aget ps (+ 2 i))
-                                                       p (Point3d. x y z)
-                                                       _ (.transform local-transform p)]
-                                                   (recur (+ i 3) (geom/aabb-incorporate aabb p)))
-                                                 aabb))]
-                                    (recur aabb (next meshes)))
-                                  aabb))))
```
matgis, line 383: We now use the `aabb-min` and `aabb-max` values from the importer! 🌈

editor/src/clj/editor/model_scene.clj
```diff
@@ -400,14 +433,15 @@
                                     :icon mesh-icon
                                     :view-types [:scene :text]))
 
-(defn- update-vb [^GL2 gl ^VertexBuffer vb data]
-  (let [{:keys [mesh world-transform vertex-space vertex-attribute-bytes]} data
-        world-transform (doto (Matrix4d.) (math/clj->vecmath world-transform))]
-    (mesh->vb! vb world-transform vertex-space vertex-attribute-bytes mesh)))
+(defn- update-vb [^GL2 _gl ^VertexBuffer vb data]
+  (let [{:keys [mesh-renderable-data ^Matrix4d world-transform ^Matrix4d normal-transform vertex-attribute-bytes]} data]
+    (mesh->vb! vb world-transform normal-transform vertex-attribute-bytes mesh-renderable-data)
```
matgis, line 438: Added `normal-transform`, changed final argument to `renderable-data` for the mesh.

editor/src/clj/editor/model_scene.clj
```diff
@@ -400,14 +433,15 @@
...
+    vb))
 
 (defn- make-vb [^GL2 gl data]
-  (let [{:keys [mesh vertex-description]} data
-        vbuf (vtx/make-vertex-buffer vertex-description :dynamic (alength ^ints (:indices mesh)))]
+  (let [{:keys [mesh-renderable-data vertex-description]} data
+        position-data (:position-data mesh-renderable-data)
+        vbuf (vtx/make-vertex-buffer vertex-description :dynamic (count position-data))]
```
matgis, line 444: The `:indices` have been used to construct a non-indexed triangle list already in the  `renderable-mesh-set` output, so at this point we can just count the number of positions.

editor/src/clj/editor/model_scene.clj
```diff
@@ -310,86 +256,173 @@
...
+     :children model-scenes}))
+
+(g/defnk produce-scene [_node-id renderable-mesh-set]
+  (make-scene renderable-mesh-set _node-id))
+
+(defn- augment-mesh-scene [mesh-scene old-node-id new-node-id new-node-outline-key material-name->material-scene-info]
+  (let [mesh-renderable (:renderable mesh-scene)
+        material-name (:material-name mesh-renderable)
+        material-scene-info (material-name->material-scene-info material-name)
+        claimed-scene (scene/claim-child-scene old-node-id new-node-id new-node-outline-key mesh-scene)]
```
matgis, line 369: An issue with the old code was that the Model using a particular `model-scene` did not claim ownership of the scene-data produced by its `scene` output. This resulted in meshes that had different requirements in terms of materials and transforms sharing the same `request-id` for the `scene-cache`. There were also issues related to picking and selection, since the renderables could not be uniquely identified to belong to the referencing Model.

vlaaad (approved):
Did I read it? Yes. Did I get it? Well... :D

matgis:
Haha, OK... I'll ask Jhonny to have a look as well. 😅

editor/src/clj/editor/scene.clj
```diff
@@ -870,6 +870,7 @@
       (.destroy drawable))
     (when-let [^GLAutoDrawable picking-drawable (g/node-value node-id :picking-drawable)]
       (gl/with-drawable-as-current picking-drawable
+        (scene-cache/drop-context! gl)
```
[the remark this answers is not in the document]
Jhonnyg: I didn't know we created a secondary context for the picking, why do we need that?

Jhonnyg (approved):

editor/src/clj/editor/model.clj
```diff
@@ -218,51 +218,11 @@
       explicit-textures
       samplers)))
 
-(g/defnk produce-scene [_node-id scene mesh-material-ids material-scene-infos]
-  (if (not scene)
+(g/defnk produce-scene [_node-id scene material-name->material-scene-info]
+  (if scene
+    (model-scene/augment-scene scene _node-id model-resource-type-label material-name->material-scene-info)
```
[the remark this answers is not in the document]
matgis: Well, it didn't before this PR. 😄

I moved the "augmentation" code to `model_scene.clj` so `model.clj` does not have to rely on it being structured in a particular way.

editor/src/clj/editor/scene.clj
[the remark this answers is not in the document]
matgis: I think the OpenGL context is associated with the `GLAutoDrawable`. Maybe we do it this way so we can perform picking without having to discard the color buffer? 🤔

PR #8696, opened 2024-03-21
Fixed property schema validation in the editor

matgis:
### Technical changes
* Schema validation errors from node properties are now reported correctly instead of throwing an unhelpful exception.


vlaaad (approved):
Shipit!

PR #8698, opened 2024-03-22
Fix flickering on vulkan when drawing to a render target

Jhonnyg:
Fixed an issue where flickering was happening on the Vulkan and MoltenVK renderers when using a render target.


engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -1094,6 +1094,11 @@ namespace dmGraphics
 
         delete[] device_list;
 
+        if (created_width != context->m_Width || created_height != context->m_Height)
+        {
+            dmPlatform::SetWindowSize(context->m_Window, created_width, created_height);
+        }
```
Jhonnyg, line 1100: If the swapchain has been created at a different resolution we need to make sure the GLFW layer knows about the new updated resoultion.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -1342,7 +1347,10 @@ namespace dmGraphics
         vk_present_info.pResults           = 0;
 
         res = vkQueuePresentKHR(context->m_LogicalDevice.m_PresentQueue, &vk_present_info);
-        CHECK_VK_ERROR(res);
+        if (res != VK_SUCCESS && res != VK_SUBOPTIMAL_KHR)
+        {
+            CHECK_VK_ERROR(res);
+        }
```
Jhonnyg, line 1353: This is a fix / workaround for android where the swap chain capabilities can have a presentation transform. For now we skip the transform completely by setting the currentTransform = identity, but that causes the presentation function to return a suboptimal result, which we don't care about right now.

engine/graphics/src/vulkan/graphics_vulkan_swap_chain.cpp
```diff
@@ -153,6 +153,8 @@ namespace dmGraphics
                 capabilities.m_SurfaceCapabilities.maxImageCount);
         }
 
+        swap_chain_image_count = dmMath::Min(swap_chain_image_count, (uint32_t) DM_MAX_FRAMES_IN_FLIGHT);
```
Jhonnyg, line 156: We can only support a fixed amount of frame resources (and don't want to)

engine/graphics/src/vulkan/graphics_vulkan.cpp
JCash: Would this be good info as a code comment, or is it easy to figure out?

JCash (dismissed):
lgtm
only small note about possible code comment

engine/graphics/src/vulkan/graphics_vulkan.cpp
Jhonnyg: Yeah it's probably a good code comment, I will forget it otherwise

PR #8709, opened 2024-03-25
Allow renaming multiple files with the same name at once

### LINKED ISSUE #6690: Allow renaming multiple files at once
It's common to have multiple related files with the same name but different extensions, e.g. `foo.gui`+`foo.gui_script`. When making new guis, it's also common to copy some base gui+script set and rename it to a new one. Allowing renaming these files together would speed up this process

Allow selecting multiple files with the same name and different extensions and rename them together.

We could suggest selecting a name for files with the same names but different extensions when pasting it.

vlaaad:
Technical notes:
I relaxed the rename requirement from "either 1 folder or 1 file" to "either 1 folder or a non-empty collection of files with the same names and different extensions"

Fixes #6690

### DIFF
--- a/editor/src/clj/editor/asset_browser.clj
+++ b/editor/src/clj/editor/asset_browser.clj
@@ -28,7 +28,8 @@
             [editor.resource :as resource]
             [editor.resource-watch :as resource-watch]
             [editor.ui :as ui]
-            [editor.workspace :as workspace])
+            [editor.workspace :as workspace]
+            [util.coll :refer [pair]])
   (:import [com.defold.control TreeCell]
            [editor.resource FileResource]
            [java.io File]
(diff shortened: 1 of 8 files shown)


editor/src/clj/editor/asset_browser.clj
```diff
@@ -357,32 +358,55 @@
...
+          (let [resolved-rename-pairs (into resolved-conflicts case-changes)]
+            (when (seq resolved-rename-pairs)
+              (workspace/resource-sync!
+                workspace
+                (into []
+                      (mapcat (fn [[src-file dst-file]]
+                                (let [src-files (vec (file-seq src-file))]
+                                  (fs/move! src-file dst-file)
+                                  (moved-files src-file dst-file src-files))))
+                      resolved-rename-pairs)))))))))
```
vlaaad, line 409: I moved `rename?` and `rename` around so I can assert that `rename?` holds when `rename` is called.

editor/src/clj/editor/dialogs.clj
```diff
@@ -787,21 +787,37 @@
 
 (defn- sanitize-common [name]
   (-> name
-      string/trim
       (string/replace #"[/\\]" "") ; strip path separators
       (string/replace #"[\"']" "") ; strip quotes
-      (string/replace #"[<>:|?*]" ""))) ; Additional Windows forbidden characters
+      (string/replace #"[<>:|?*]" "") ; Additional Windows forbidden characters
+      string/trim))
```
vlaaad, line 793: We actually want to trim last, otherwise we still might end up with white space on the edges of a name after we remove the forbidden characters, e.g. for cases like `aaa  '`.

editor/src/clj/editor/dialogs.clj
```diff
@@ -787,21 +787,37 @@
...
 
 (defn sanitize-file-name [extension name]
   (let [name (sanitize-common name)]
     (cond-> name
-            ; disallow dots when extension is expected
-            (seq extension) (string/replace #"\." "")
```
vlaaad, line 799: We actually want to allow dots in file names when renaming files, so that files named like `type-definition.d.tl` can be renamed without losing their meaningful dots.

editor/src/clj/editor/asset_browser.clj
vlaaad: The `rename` logic was changed to allow renaming multiple files at once, but the intent and behavior largely stayed the same.

matgis (approved):
Looks good to me!

PR #8710, opened 2024-03-25
Produce local and world spaced models with correct custom attribute values

### LINKED ISSUE #8693: Model missing custom vertex attributes in World vertex space
Custom vertex attribute color tint results in black mesh when Vertex Space is set to World in the material.

**To Reproduce (REQUIRED)**
1. Open the [attached project](https://github.com/defold/defold/files/14697356/dev-models.zip) in the editor.
2. Open `main/main.collection`.
3. Select `both.collection` from the Outline View and press `F` to frame it in the Scene View. This is how the scene should look in-engine.
4. Build and run the project.

All four boxes should be tinted a light grey colour and look like it does in the editor.

**Defold version (REQUIRED)**
  - 1.7.0

**Platforms (REQUIRED)**
 - Platforms: macOS
 - OS: Sonoma 14.3.1 (23D60)
 - Device: MacBook Pro 16-inch, 2021

**Screenshots (OPTIONAL):**
Expected:
Image

Actual:
Image

Jhonnyg:
Fixed an issue where models are using custom attributes with and without mesh colors available in the .glb file:
* When a model is using a mesh without color data, it should produce the overridden data from the custom vertex attributes
* When a model is using a mesh with color data, the color data should always take precedence over the custom vertex format

Fixes #8693

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_model.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_model.cpp
@@ -526,7 +526,10 @@ namespace dmGameSystem
                     &attribute_infos);
 
         uint8_t* scratch_attribute_vertex = (uint8_t*) malloc(dmGraphics::GetVertexDeclarationStride(material_vx_decl));
-        uint32_t custom_vertex_size       = 0;
+
+        dmGraphics::VertexAttributeInfos non_default_attribute;
+        non_default_attribute.m_VertexStride = 0;
+        non_default_attribute.m_NumInfos     = 0;
 
         for (int i = 0; i < material_infos.m_NumInfos; ++i)
         {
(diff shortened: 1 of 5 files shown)


engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -542,21 +545,31 @@ namespace dmGameSystem
                     dmGraphics::GetGraphicsType(attr_model.m_DataType),
                     attr_model.m_Normalize);
 
-                uint8_t* data_write_ptr = scratch_attribute_vertex + custom_vertex_size;
-                memcpy(data_write_ptr, attr_model.m_ValuePtr, attr_model.m_ValueByteSize);
-                custom_vertex_size += attr_model.m_ValueByteSize;
+                uint32_t attribute_index = non_default_attribute.m_NumInfos++;
+                non_default_attribute.m_Infos[attribute_index]                 = attr_model;
+                non_default_attribute.m_Infos[attribute_index].m_ElementCount  = attr_material.m_ElementCount;
+                non_default_attribute.m_VertexStride                          += attr_model.m_ValueByteSize;
```
Jhonnyg, line 551: The old setup doesn't work when the model has per-vertex attribute data but is not a "default" attribute, for example in the case where the material has an attribute with a color semantic that differs from the default color semantic attribute type. We need to cover both cases when the model has color data and when it doesn't.

JCash (approved):
lgtm, however the wording in the PR comment about color precedence is a little confusing.
I'll comment in the PR.

JCash:
Regarding the PR documentation & expectations here.

"When a model is using a mesh with color data, the color data should always take precedence over the custom vertex format"

I assume that it _is_ possible for a user to override the color in the component in the editor? And in such case the meshes color data _doesn't_ take precedence?
If so, I feel the PR documentation is a bit unclear?

PR #8713, opened 2024-03-25
Update github workflows

ekharkunov:
### Technical changes
PR updates github actions version. It remove (almost completely) warnings about nodejs version.
Update Python installation to 3.12. It's imposible to get rid of `Install Python` step now because in that case it select wrong version of Python interpreter. Add dependabot configuration.
Use macos-13 runner image because we need to use xcode 15.1 according to https://github.com/defold/defold/blob/dev/build_tools/sdk.py#L41

Set boto branch as target to show only GH workflow difference.

## PR checklist

* [x] Code
* [x] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/.github/workflows/editor-only.yml
+++ b/.github/workflows/editor-only.yml
@@ -26,11 +26,16 @@ jobs:
   build-editor:
     runs-on: ubuntu-20.04
     steps: [
-      { name: 'Checkout', uses: actions/checkout@v3 },
-      { name: 'Fetch tags', run: 'git fetch --depth=1 origin +refs/tags/*:refs/tags/*' },
-      { name: 'Install Python', uses: actions/setup-python@v3, with: { python-version: 3.11, architecture: x64 } },
-      { name: 'Install Java', uses: actions/setup-java@v3, with: { java-version: '17.0.5+8', distribution: 'temurin'} },
-      { name: 'Install Leiningen', uses: DeLaGuardo/setup-clojure@master, with: { lein: 2.8.3 } },
+      {
+        name: 'Checkout',
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11,
+        with: {
+          fetch-tags: true
+        }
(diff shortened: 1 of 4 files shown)


.github/workflows/editor-only.yml
```diff
@@ -26,11 +26,16 @@ jobs:
...
     runs-on: ubuntu-20.04
     steps: [
-      { name: 'Checkout', uses: actions/checkout@v3 },
-      { name: 'Fetch tags', run: 'git fetch --depth=1 origin +refs/tags/*:refs/tags/*' },
-      { name: 'Install Python', uses: actions/setup-python@v3, with: { python-version: 3.11, architecture: x64 } },
-      { name: 'Install Java', uses: actions/setup-java@v3, with: { java-version: '17.0.5+8', distribution: 'temurin'} },
-      { name: 'Install Leiningen', uses: DeLaGuardo/setup-clojure@master, with: { lein: 2.8.3 } },
+      {
+        name: 'Checkout',
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11,
```
britzl, line 31: Isn't it possible to use `actions/checkout@v4.1.1` instead of the sha1? I know it is the same but if we use a tag I know at a glance which version is used.

ekharkunov: According to https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions it's recommended to use hash instead of tag name. And sometimes maintainers like to move tags and it can lead to inconsistent between two builds.
I can add a comment with tag name. Kind of dependabot can rewrite comment when propose new action update.

britzl: I was not aware of this! Thanks. 👍

JCash (approved):

PR #8727, opened 2024-03-27
Pass all known model scene attributes to shaders in editor viewport

### LINKED ISSUE #7407: IDE: Pass color stream to display 3D models properly
It would be great if IDE passed color stream to be able to preview 3D models with vertex colors properly.

Currently, it doesn't do that and this is how IDE displays models with vertex colors:
image

How they look in the game:
image

The demo project:
[issue_model_color.zip](https://github.com/defold/defold/files/10787102/issue_model_color.zip)

The same drawback described there - https://forum.defold.com/t/new-gltf-model-importer-beta-test/71436/60?u=aglitchman

matgis:
The scene view will now pass all currently supported vertex attributes to the shader when rendering models.

Fixes #7407

### Technical changes
* Vertex attributes that are present in the Mesh data will take precedence over attribute values from the Material or the Model referencing the Mesh when rendered in the editor viewport.

-----

Previously submitted as #8724. This is a new PR with different branch name targeting `dev`.

### DIFF
--- a/editor/src/clj/editor/graphics.clj
+++ b/editor/src/clj/editor/graphics.clj
@@ -18,6 +18,7 @@
             [editor.gl.shader :as shader]
             [editor.gl.vertex2 :as vtx]
             [editor.properties :as properties]
+            [editor.protobuf :as protobuf]
             [editor.types :as types]
             [editor.util :as eutil]
             [editor.validation :as validation]
@@ -26,7 +27,8 @@
             [util.coll :refer [pair]]
             [util.murmur :as murmur]
             [util.num :as num])
-  (:import [com.google.protobuf ByteString]
+  (:import [com.dynamo.graphics.proto Graphics$VertexAttribute$SemanticType]
+           [com.google.protobuf ByteString]
            [com.jogamp.opengl GL2]
            [editor.gl.vertex2 VertexBuffer]
            [java.nio ByteBuffer]))
@@ -512,15 +514,20 @@
         world-transform (:world-transform renderable-data)]
     (geom/transf-p4 world-transform local-positions)))
 
-(defn- renderable-data->world-normal-v3 [renderable-data]
-  (let [local-normals (:normal-data renderable-data)
+(defn- renderable-data->world-direction-v3 [renderable-data->local-directions renderable-data]
+  (let [local-directions (renderable-data->local-directions renderable-data)
         normal-transform (:normal-transform renderable-data)]
-    (geom/transf-n normal-transform local-normals)))
+    (geom/transf-n normal-transform local-directions)))
 
-(defn- renderable-data->world-normal-v4 [renderable-data]
-  (let [local-normals (:normal-data renderable-data)
+(defn- renderable-data->world-direction-v4 [renderable-data->local-directions renderable-data]
+  (let [local-directions (renderable-data->local-directions renderable-data)
         normal-transform (:normal-transform renderable-data)]
-    (geom/transf-n4 normal-transform local-normals)))
+    (geom/transf-n4 normal-transform local-directions)))
+
+(def ^:private renderable-data->world-normal-v3 (partial renderable-data->world-direction-v3 :normal-data))
+(def ^:private renderable-data->world-normal-v4 (partial renderable-data->world-direction-v4 :normal-data))
+(def ^:private renderable-data->world-tangent-v3 (partial renderable-data->world-direction-v3 :tangent-data))
+(def ^:private renderable-data->world-tangent-v4 (partial renderable-data->world-direction-v4 :tangent-data))
 
 (defn put-attributes! [^VertexBuffer vbuf renderable-datas]
   (let [vertex-description (.vertex-description vbuf)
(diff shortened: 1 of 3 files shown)


editor/src/clj/editor/graphics.clj
```diff
@@ -555,15 +562,45 @@
                   attribute-byte-offset
                   renderable-datas))
 
-        texcoord-index-vol (volatile! -1)
-        page-index-vol (volatile! -1)]
```
matgis, line 559: These volatiles were previously used to keep track of how many times we've encountered an attribute with `:semantic-type-texcoord`, or `:semantic-type-page-index`, respectively. We now track this for all `semantic-types`, along with the `:attribute-byte-offset`, in the `reduce-info` `hashmap` we propagate as the state argument to the reduction function as we `reduce` over the `attributes`.

editor/src/clj/editor/graphics.clj
```diff
@@ -581,51 +618,77 @@
...
-                            4 renderable-data->world-position-v4)]
-                      (put-renderables! attribute-byte-offset renderable-data->world-position put-attribute-doubles!))
-                    (put-renderables! attribute-byte-offset :position-data put-attribute-doubles!))
-
-                  :semantic-type-texcoord
-                  (let [i (vswap! texcoord-index-vol inc)]
-                    (put-renderables! attribute-byte-offset #(get-in % [:texcoord-datas i :uv-data]) put-attribute-doubles!))
-
-                  :semantic-type-page-index
-                  (let [i (vswap! page-index-vol inc)]
```
matgis, line 600: Big diff here, mostly due to indentation. We now check if `(mesh-data-exists? semantic-type channel)` to determine if we should use the attribute data from the Mesh or from the Material/Model Overrides. If there is data in the Mesh, it takes precedence for now.

Previously, the `semantic-type` determined if we should take the data from the Mesh or from the Material/Model Overrides. `:semantic-type-position`, `:semantic-type-normal`, `:semantic-type-texcoord`, and `:semantic-type-page-index` would always be taken from the Mesh, and any other attributes from the Material/Model Overrides.

This also applies to Sprites, but you get the idea. For sprites, the editor provides `:semantic-type-position`, `:semantic-type-texcoord`, and `:semantic-type-page-index` as "Mesh" data.

editor/src/clj/editor/model_scene.clj
```diff
@@ -119,47 +120,56 @@
...
+              (= component-count (count (peek vectors))))
+      vectors)))
+
 (defn mesh->renderable-data [mesh]
-  (let [positions (doubles->floats (:positions mesh))
-        normals (doubles->floats (:normals mesh))
-        texcoord0s (doubles->floats (:texcoord0 mesh))
+  (let [positions (flat->vectors 3 (:positions mesh))
+        normals (flat->vectors 3 (:normals mesh))
+        tangents (flat->vectors 3 (:tangents mesh))
```
matgis, line 136: Apparently there are only three components to the `tangents` we import from the meshes, and there are no `binormals`. So how are you supposed to reconstruct the tangent space? 🤔

Either we need the `binormals`, or we could have a handedness scalar [`-1.0` or `1.0`] in `tangent.w` to inverse it for purposes of the cross-product with the `normal`.

For now, I'll just pass it on to the shader, I guess... 🤷

editor/src/clj/util/coll.clj
```diff
@@ -132,6 +133,41 @@
...
+              (rf result finished))
+            result))))))
+  ([primitive-type ^long partition-length coll]
+   (partition-all-primitives primitive-type partition-length partition-length coll))
+  ([primitive-type ^long partition-length ^long step coll]
+   (lazy-seq
+     (when-let [in-progress (seq coll)]
+       (let [finished (apply vector-of primitive-type (take partition-length in-progress))]
+         (cons finished (partition-all-primitives primitive-type partition-length step (nthrest in-progress step))))))))
+
```
matgis, line 170: Based on `core.partition-all`. The main difference is that we `(apply vector-of primitive-type ...)` to the `in-progress` items when we've finished a partition.

editor/src/clj/util/coll.clj
```diff
@@ -132,6 +133,41 @@
          (map (pair-fn key-fn value-fn))
          coll)))
 
+(defn partition-all-primitives
+  "Returns a lazy sequence of primitive vectors. Like core.partition-all, but
+  creates a new vector of a single primitive-type for each partition. The
+  primitive-type must be one of :int :long :float :double :byte :short :char, or
+  :boolean. Returns a stateful transducer when no collection is provided."
+  {:added "1.2"
```
vlaaad, line 141: What do you mean by `:added "1.2"`?

editor/src/clj/util/coll.clj
vlaaad: Why do we want to use `vector-of` instead of plain vectors? I see that `Vec` stores data in primitive array, but will box the primitive on every lookup. I guess we mainly store the renderable data, and rarely access it?

vlaaad (dismissed):
LGTM!

editor/src/clj/editor/model_scene.clj
Jhonnyg: Yeah, it's a good question. Afaik blender exports handedness in the w component, which would be the simplest fix for it. I wonder if this test shows handedness issues: https://github.com/KhronosGroup/glTF-Sample-Models/blob/main/2.0/NormalTangentTest/README.md

In any case, we don't have to fix it now but I'll add an issue for it

Jhonnyg (dismissed):

editor/src/clj/util/coll.clj
matgis: I measured it, and it doesn't appear to make a difference in this case. However, I also discovered that `reduce` is slower over a `(vector-of :double)` compared to a regular `vector` by a factor of 3-4 when I experimented with interleaving x, y, z values in a flat double vector. Again, not relevant to this case, but perhaps something to be aware of in the future.

```
From object-vector...
WARNING: Final GC required 3.340659678476196 % of runtime
Evaluation count : 359038080 in 60 samples of 5983968 calls.
             Execution time mean : 165,561063 ns
    Execution time std-deviation : 0,985169 ns
   Execution time lower quantile : 164,176282 ns ( 2,5%)
   Execution time upper quantile : 168,523537 ns (97,5%)
                   Overhead used : 1,993035 ns

Found 7 outliers in 60 samples (11,6667 %)
	low-severe	 1 (1,6667 %)
	low-mild	 3 (5,0000 %)
	high-mild	 3 (5,0000 %)
 Variance from outliers : 1,6389 % Variance is slightly inflated by outliers

From primitive-vector...
Evaluation count : 359670360 in 60 samples of 5994506 calls.
             Execution time mean : 167,817901 ns
    Execution time std-deviation : 11,173209 ns
   Execution time lower quantile : 162,723047 ns ( 2,5%)
   Execution time upper quantile : 196,444340 ns (97,5%)
                   Overhead used : 1,993035 ns

Found 4 outliers in 60 samples (6,6667 %)
	low-severe	 2 (3,3333 %)
	low-mild	 2 (3,3333 %)
 Variance from outliers : 50,0984 % Variance is severely inflated by outliers
```
Interestingly, I got the GC warning from the regular `vector` case but not the `(vector-of :double)` case. Likely just a coincidence that it triggered a GC that time, because it did not happen consistently.

PR #8732, opened 2024-03-28
Added resource.create_texture_async()

### LINKED ISSUE #8350: Async `resource.create_texture()`
No

Now Defold supports a few ways to load textures from external source:
- `resource.create_texture()`
- `gui.new_texture()`

The both ways create textures synchronously and require to have texture data loaded. This approach works fine if the game asks  small textures dynamically. But it leads to performance problems if the game wants to load big texture.

Async version of `resource.create_texture()` with support of the texture file source under external filesystem (in game data folder) or under game filesystem. The main idea is to keep all works done in async manner from file reading to texture loading. Also possibility to cancel operation would be great.

No alternatives because there are no async file load API (`sys.load_resource` and `io.file`) and async texture creation API

Jhonnyg:
Added a new resource function to create textures asynchronously: `resource.create_texture_async(path, tparams, [buffer], [callback])`:

* path - the path to the resource to create (/main/my_texture.texturec)
* tparams - texture params, width, height, format and so on
* [buffer] - optional buffer that contains the texture data. if the buffer isn't provided, the texture will be created with blank data
* [callback] - optional callback that will be called when the upload has completed. The callback has the following setup:

`function my_callback(self, request_id, result)`
* self - the script where the request was issued
* request_id - the unique request id that is returned from resource.create_texture_async
* result - a table containing:
                'path' - the resource path of the texture that was updated

This function will create a new texture resource and upload the data into it in a worker thread. The function will return both a resource path and a request id for the async upload, so that specific requests can be tracked during the lifetime of the app or game. The returned resource path can be used immediately (i.e you can pass it to components with go.set), the initial texture will be a blank 1x1 texture which will later be replaced by the actual texture data. 

```lua
function my_callback(self, request_id, result)
    -- texture has been updated, do something with it
   self.requests[request_id].loaded = true
   go.set("#model", "texture0", result.path)
end

local t_path, t_request_id = resource.create_texture_async("/main/my_texture.texturec", t_params, my_buffer, my_callback)
self.requests[t_request_id] = { path = t_path, loaded = false }
```

Fixes #8350


engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -221,10 +221,41 @@ namespace dmGameSystem
...
+    dmGraphics::TextureFormat                 m_Format;
+    dmGraphics::TextureImage::Type            m_TextureType;
+    dmGraphics::TextureImage::TextureFormat   m_TextureFormat;
+    dmGraphics::TextureImage::CompressionType m_CompressionType;
+    dmBuffer::HBuffer                         m_Buffer;
+    uint32_t                                  m_Width;
+    uint32_t                                  m_Height;
+    uint32_t                                  m_MaxMipMaps;
+    uint32_t                                  m_TextureBpp;
+};
```
Jhonnyg, line 252: Q: Do we want to create a script_resource_texture.cpp and move the texture functions and structs into it? this file is getting kinda big..

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -647,6 +680,138 @@ static void CheckTextureResource(lua_State* L, int i, const char* field_name, dm
...
+        }
+
+        dmScript::DestroyCallback(request->m_CallbackInfo);
+    }
+
+    // If we used a temporary upload buffer, we need to delete it.
+    if (request->m_UseUploadBuffer)
+    {
+        dmBuffer::Destroy(request->m_Buffer);
+    }
```
Jhonnyg, line 805: This is so we can support passing in a nil buffer to create a blank texture async

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -779,101 +944,285 @@ static void CheckTextureResource(lua_State* L, int i, const char* field_name, dm
...
+ * - `resource.TEXTURE_FORMAT_RG_BC5`
+ * - `resource.TEXTURE_FORMAT_RGBA_BC7`
+ * - `resource.TEXTURE_FORMAT_RGB16F`
+ * - `resource.TEXTURE_FORMAT_RGB32F`
+ * - `resource.TEXTURE_FORMAT_RGBA16F`
+ * - `resource.TEXTURE_FORMAT_RGBA32F`
+ * - `resource.TEXTURE_FORMAT_R16F`
+ * - `resource.TEXTURE_FORMAT_RG16F`
+ * - `resource.TEXTURE_FORMAT_R32F`
+ * - `resource.TEXTURE_FORMAT_RG32F`
```
Jhonnyg, line 1033: Can we paste or include docstrings from other parts of the engine somehow? This is a copy-paste from resource.create_texture, so if we decide to chagne something we need to keep these in sync, which is brittle..

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -779,101 +944,285 @@ static void CheckTextureResource(lua_State* L, int i, const char* field_name, dm
...
 
-    DestroyTextureImage(texture_image, buffer == 0);
+    dmGraphics::TextureParams texture_params;
+    texture_params.m_Width  = create_params.m_Width;
+    texture_params.m_Height = create_params.m_Height;
+    texture_params.m_Format = create_params.m_Format;
 
-    if (res != dmResource::RESULT_OK)
+    dmBuffer::GetBytes(request->m_Buffer, (void**)&texture_params.m_Data, &texture_params.m_DataSize);
+    dmGraphics::SetTextureAsync(texture_dst, texture_params);
```
Jhonnyg, line 1212: We sidestep the resource system here for the update mainly for simplicity, but should we consider being able to update a texture resource without explicit syncronization? the texture resource uploads data via the same SetTextureAsync function here but does an explicit sync. This causes the app to hang/stall until the upload is complete, which is the point of this function to avoid such behaviour..

engine/graphics/src/null/graphics_null.cpp
```diff
@@ -1356,12 +1363,22 @@ namespace dmGraphics
             tex->m_OriginalHeight = params.m_OriginalHeight;
         }
 
-        return StoreAssetInContainer(context->m_AssetHandleContainer, tex, ASSET_TYPE_TEXTURE);
+        ASSET_CONTAINER_LOCK_BEGIN(context);
+
+        HTexture handle = StoreAssetInContainer(context->m_AssetHandleContainer, tex, ASSET_TYPE_TEXTURE);
+
+        ASSET_CONTAINER_LOCK_END(context);
```
Jhonnyg, line 1370: I used this pattern instead of the scoped lock, since we only need the lock in specific tests, i.e we don't use the async texture loading by default when using the null graphics library.

engine/graphics/src/null/graphics_null.cpp
```diff
@@ -1712,6 +1735,18 @@ namespace dmGraphics
...
+
+        {
+            ASSET_CONTAINER_LOCK_BEGIN(context);
+            Texture* tex = GetAssetFromContainer<Texture>(context->m_AssetHandleContainer, ap.m_Texture);
+            if (tex)
+            {
+                SetTexture(ap.m_Texture, ap.m_Params);
+                tex->m_DataState &= ~(1<<ap.m_Params.m_MipMap);
+            }
+            ASSET_CONTAINER_LOCK_END(context);
```
Jhonnyg, line 1747: This is why we need the lock now, since we do the actual upload in the thread. So we need to protect the asset container so other threads don't start fiddling with its memory

engine/graphics/src/opengl/graphics_opengl_private.h
```diff
@@ -78,6 +78,7 @@ namespace dmGraphics
         char*                   m_ExtensionsString;
         void*                   m_AuxContext;
         volatile bool           m_AuxContextJobPending;
+        volatile bool           m_DeleteContextRequested;
```
Jhonnyg, line 81: maybe we should use an atomic instead for these?

JCash (changes requested):
Overall good.

Requested change for moving the IncRef above the SetTextureAsync call site.

Also, a question about polling versus callbacks.

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -647,6 +680,138 @@ static void CheckTextureResource(lua_State* L, int i, const char* field_name, dm
...
+
+    dmGraphics::TextureImage::CompressionType compression_type = (dmGraphics::TextureImage::CompressionType) CheckTableInteger(L, 2, "compression_type", (int) dmGraphics::TextureImage::COMPRESSION_TYPE_DEFAULT);
+
+    dmBuffer::HBuffer buffer = 0;
+    if (lua_gettop(L) > 2 && !lua_isnil(L, 3))
+    {
+        // TODO: Support creating texture from string
+        dmScript::LuaHBuffer* lua_buffer = dmScript::CheckBuffer(L, 3);
+        buffer                           = dmGameSystem::UnpackLuaBuffer(lua_buffer);
+    }
```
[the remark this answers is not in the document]
Jhonnyg: Yeah, it's further down:
request->m_BufferRef = dmScript::Ref(L, LUA_REGISTRYINDEX);

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -3137,6 +3485,20 @@ void ScriptResourceRegister(const ScriptLibContext& context)
...
+void ScriptResourceUpdate(const ScriptLibContext& context)
+{
+    uint32_t request_count = g_ResourceModule.m_LoadRequests.Capacity();
+    for (int i = 0; i < request_count; ++i)
+    {
+        SetTextureAsyncRequest* request = g_ResourceModule.m_LoadRequests.GetByIndex(i);
+
+        if (request && dmGraphics::GetTextureStatusFlags(request->m_Texture) == dmGraphics::TEXTURE_STATUS_OK)
+        {
+            HandleRequestCompleted(request);
```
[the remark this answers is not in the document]
Jhonnyg: Yeah, I guess that could work hm.. we need to modify the graphics API a bit to support passing in a "complete" callback. I'll see if I can remedy that

engine/graphics/src/opengl/graphics_opengl_private.h
Jhonnyg: Changed these to atomics, should be easy to merge if we get conflicts 👍

engine/gamesys/src/gamesys/scripts/script_resource.cpp
Jhonnyg: I think i'll do this separately

PR #8743, opened 2024-03-31
Instancing support for models

### LINKED ISSUE #7314: Geometry instance support
To increase performance for more advanced 3D features such as foliage, crowds and props (and perhaps even 2D, e.g particles) you typically want to use hardware instancing to avoid issuing unnecessary draw calls as well as passing more vertices to the GPU before rendering.

Control what objects or materials that should be issued as instanced calls.

Only alternative is what we have now, i.e draw models one-by-one or via the built-in batching that happens per component type.

### LINKED ISSUE #9398: Error caused if Vertex Attribute Count is too high, followed by UX failures
### Expected behaviour
Entering an invalid "Count" for a Vertex Attribute should not crash the material editor/outline

### Actual behaviour
Entering "5" for a Vertex Attribute "count" will cause an error, followed by missing UX elements e.g. stuff in the Outline fails to load until Defold is restarted.

### Steps to reproduce
image

<hr/>

<table>
<tr><td>Build time</td><td>2024-06-25T16:26:09.046426</td></tr>
<tr><td>Defold editor sha</td><td>3642fa5b700b8d6e74e3f0595e75cd50eff54459</td></tr>
<tr><td>Defold engine sha</td><td>d6882f432beca85d460ec42497888157c356d058</td></tr>
<tr><td>Defold version</td><td>1.9.0</td></tr>
<tr><td>Error</td><td><a href='https://sentry.io/organizations/defold/issues/?query=id%3A"56940ccba26842a394f346fd2b4b6a7c"'>56940ccba26842a394f346fd2b4b6a7c</a></td></tr>
<tr><td>GPU</td><td>NVIDIA GeForce RTX 3060/PCIe/SSE2</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 NVIDIA 535.98</td></tr>
<table>

Jhonnyg:
Model components can now use hardware instancing! This means that models that share the same mesh but has its own transform and normal matrices can now be rendered with a single draw call instead of individual draw calls. This is useful for repeated geometry, such as trees, foliage, static props such as trash cans, boxes or even complete buildings. 

The simplest route to start using instancing is to add an attribute in the vertex shader that holds the `world matrix`, or a `normal matrix`:

```glsl
attribute mat4 mtx_world;
attribute mat4 mtx_normal;
```

When any attribute is marked as an instanced attribute, instanced draw calls will automatically be created. This also works for custom vertex attributes. Here's an example of using an custom instanced attribute for storing an individual tint attribute per model:

image

The color attribute in this case has its `vertex step function` marked as `instance` in the material, which means that this attribute will be repeated per instance instead of per vertex:

image

This enables us to draw all boxes with just one draw call instead of 4, even with custom vertex data per model!

To read more about instancing in Defold, please consult the [Manual entry](https://defold.com/manuals/material/#instancing).

A simple example that shows how to use the new functionality is available here:
[instancing-test.zip](https://github.com/user-attachments/files/16597305/instancing-test.zip)

Fixes #7314 
Fixes #9398


engine/graphics/proto/graphics/graphics_ddf.proto
```diff
@@ -36,15 +42,28 @@ message VertexAttribute
...
+    enum ShaderType
+    {
+        SHADER_TYPE_NUMBER = 1;
+        SHADER_TYPE_VEC2   = 2;
+        SHADER_TYPE_VEC3   = 3;
+        SHADER_TYPE_VEC4   = 4;
+        SHADER_TYPE_MAT2   = 5;
+        SHADER_TYPE_MAT3   = 6;
+        SHADER_TYPE_MAT4   = 7;
+    }
```
Jhonnyg, line 54: We have a special type for these here instead of the "ShaderDataType" that also contains these values. Not sure we can merge them, since the shaderdatatype has things like uint, float and so forth, but we already have that covered that with the "DataType" enum

editor/src/clj/editor/gl.clj
```diff
@@ -158,6 +158,7 @@
 (defmacro gl-bind-buffer [gl type name]                                  `(.glBindBuffer ~gl ~type ~name))
 (defmacro gl-buffer-data [gl type size data usage]                       `(.glBufferData ~gl ~type ~size ~data ~usage))
 (defmacro gl-vertex-attrib-pointer [gl idx size type norm stride offset] `(.glVertexAttribPointer ~gl ~idx ~size ~type ~norm ~stride ~offset))
+(defmacro gl-vertex-attrib-divisor [gl idx divisor]                      `(.glVertexAttribDivisor ~gl ~idx ~divisor))
```
Jhonnyg, line 161: This will be needed eventually once we implement instancing in the editor.

editor/src/clj/editor/graphics.clj
```diff
@@ -301,30 +332,35 @@
...
     ;; We still want to remove the default empty :name-hash string, though.
-    (-> attribute
+    (-> (if (some? attribute-vector-type)
+          (assoc attribute :vector-type attribute-vector-type)
+          attribute)
         (dissoc :name-hash :double-values :long-values :binary-values)
-        (assoc attribute-value-keyword {:v attribute-values}))))
+        (assoc attribute-value-keyword {:v attribute-values})
+        ;; element-count is deprecated in favor of vector-type
+        (dissoc :element-count))))
```
Jhonnyg, line 348: Maybe this can be written in some cleaner way? i.e instead of the if-cond

editor/src/clj/editor/material.clj
```diff
@@ -244,10 +245,11 @@
...
                     :default :type-float}
-                   {:path [:element-count]
-                    :label "Count"
-                    :type :integer
-                    :default default-element-count}
+                   {:path [:vector-type]
+                    :label "Vector Type"
+                    :type :choicebox
+                    :options (protobuf-forms/make-options vector-type-values)
+                    :default default-vector-type}
```
Jhonnyg, line 252: Q: If we have a vec4 in the shader and reconfigure it here as a mat4 - what should we do?
* Build error
* Filter out the impossible values from the dropdown?
* Both?

editor/src/clj/editor/model_scene.clj
```diff
@@ -172,46 +172,70 @@
                 (some some? texcoord-datas) (assoc :texcoord-datas texcoord-datas)))
       (error-values/error-fatal "Failed to produce vertex buffers from mesh set. The scene might contain invalid data."))))
 
-(defn mesh->vb! [^VertexBuffer vbuf ^Matrix4d world-transform ^Matrix4d normal-transform vertex-attribute-bytes mesh-renderable-data]
+(defn- matrix4->bytes [^Matrix4d mtx]
+  (let [matrix-array (math/vecmath->clj (doto (Matrix4d. mtx) (.transpose)))
+        byte-array (byte-array (* 4 16))
+        byte-buffer (vtx/wrap-buf byte-array)]
+    (vtx/buf-push! byte-buffer :float false matrix-array)
+    byte-array))
```
Jhonnyg, line 180: This function currently doesn't take the data type into account - does it make sense to convert a world matrix to a "byte matrix"?

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -159,17 +172,24 @@ namespace dmGameSystem
...
+        dmGraphics::AddVertexStream(stream_declaration_vertex, "position",  3, dmGraphics::TYPE_FLOAT, false);
+        dmGraphics::AddVertexStream(stream_declaration_vertex, "normal",    3, dmGraphics::TYPE_FLOAT, false);
+        dmGraphics::AddVertexStream(stream_declaration_vertex, "tangent",   4, dmGraphics::TYPE_FLOAT, false);
+        dmGraphics::AddVertexStream(stream_declaration_vertex, "color",     4, dmGraphics::TYPE_FLOAT, false);
+        dmGraphics::AddVertexStream(stream_declaration_vertex, "texcoord0", 2, dmGraphics::TYPE_FLOAT, false);
+        dmGraphics::AddVertexStream(stream_declaration_vertex, "texcoord1", 2, dmGraphics::TYPE_FLOAT, false);
+
+        dmGraphics::HVertexStreamDeclaration stream_declaration_instance = dmGraphics::NewVertexStreamDeclaration(graphics_context, dmGraphics::VERTEX_STEP_FUNCTION_INSTANCE);
+        dmGraphics::AddVertexStream(stream_declaration_instance, "mtx_world",  16, dmGraphics::TYPE_FLOAT, false);
+        dmGraphics::AddVertexStream(stream_declaration_instance, "mtx_normal", 16, dmGraphics::TYPE_FLOAT, false);
```
Jhonnyg, line 185: If someone is using a "standard" instancing setup with either a world or a normal transform matrix, we can use the fast route by not evaluating each attribute and instead use a fixed data layout + vertex declaration.

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -371,6 +415,45 @@ namespace dmGameSystem
         dmHashUpdateBuffer32(state, material->m_Textures, sizeof(dmGameSystem::TextureResource*)*material->m_NumTextures);
     }
 
+    static void HashRenderItem(HashState32* state, ModelComponent* component, const MeshRenderItem& item)
```
Jhonnyg, line 418: I think we can reuse the render item hashes to solve this issue:
https://github.com/orgs/defold/projects/50?pane=issue&itemId=69469210

engine/graphics/src/graphics.h
```diff
@@ -341,6 +343,27 @@ namespace dmGraphics
         uint32_t            m_StructSize;
     };
 
+    struct VertexDeclaration
```
Jhonnyg, line 346: We have merged the representation for vertex declarations into a shared struct. If it works for opengl and vulkan, it will work for all adapter types

editor/src/clj/editor/particle_lib.clj
```diff
@@ -186,6 +186,21 @@
...
 
+(defn- vector-type->element-count [vector-type]
+  (case vector-type
+    :vector-type-scalar 1
+    :vector-type-vec2 2
+    :vector-type-vec3 3
+    :vector-type-vec4 4
+    :vector-type-mat2 4
+    :vector-type-mat3 9
+    :vector-type-mat4 16))
```
matgis, line 197: This looks exactly like `graphics/attribute-vector-type->component-count`. Maybe rename that to `graphics/vector-type->component-count` and use it instead?

editor/test/integration/save_data_test.clj
```diff
@@ -670,7 +674,17 @@
...
+      (let [legacy-element-count-material (project/get-resource-node project "/silently_migrated/legacy_vertex_attribute_element_count.material")
+            legacy-attributes (g/node-value legacy-element-count-material :attributes)
+            filter-attribute-by-name (fn [attributes name]
+                                       (first (filterv #(= name (:name %)) attributes)))]
+        (is (= :vector-type-scalar (:vector-type (filter-attribute-by-name legacy-attributes "legacy_count_1"))))
+        (is (= :vector-type-vec2 (:vector-type (filter-attribute-by-name legacy-attributes "legacy_count_2"))))
+        (is (= :vector-type-vec3 (:vector-type (filter-attribute-by-name legacy-attributes "legacy_count_3"))))
+        (is (= :vector-type-vec4 (:vector-type (filter-attribute-by-name legacy-attributes "legacy_count_4"))))
+        (is (= :vector-type-mat3 (:vector-type (filter-attribute-by-name legacy-attributes "legacy_count_9"))))
+        (is (= :vector-type-mat4 (:vector-type (filter-attribute-by-name legacy-attributes "legacy_count_16"))))))
```
matgis, line 687: It seems like we could just create a map of attribute name to vector-type and check the entire map?
```suggestion
            vector-types-by-attribute-name (into (sorted-map)
                                                 (map (juxt :name :vector-type))
                                                 legacy-attributes)]
        (is (= {"legacy_count_1" :vector-type-scalar
                "legacy_count_2" :vector-type-vec2
                "legacy_count_3" :vector-type-vec3
                "legacy_count_4" :vector-type-vec4
                "legacy_count_9" :vector-type-mat3
                "legacy_count_16" :vector-type-mat4}
               vector-types-by-attribute-name))))
```

editor/test/resources/save_data_project/checked.material
```diff
@@ -159,78 +158,196 @@ attributes {
...
+    v: 0.0
+    v: 0.0
+    v: 0.0
+    v: 1.0
+    v: 0.0
+    v: 0.0
+    v: 0.0
+    v: 0.0
+    v: 1.0
+  }
```
matgis, line 257: Maybe we shouldn't write attributes to the project files when they have a semantic type that will be supplied by the engine / editor?

I guess we could strip this out later if we wanted to. Just feels a bit excessive for these new matrix attributes.

matgis (changes requested):
I've reviewed the editor changes and found some issues when testing. In particular, the `mtx_normal` attribute is wrong in the editor, and `mat3` attributes don't work correctly in either the editor or the engine.

osov:
> To read more about instancing in Defold, please consult the [Manual entry](https://www.youtube.com/watch?v=dQw4w9WgXcQ).
> 
Url not correct ?

JCash (dismissed):

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/GraphicsUtil.java
```diff
@@ -175,6 +175,47 @@ private static ByteString makeBinaryValues(VertexAttribute attr, VertexAttribute
...
+                        attributeBuilder.setVectorType(VertexAttribute.VectorType.VECTOR_TYPE_MAT2);
+                    } else {
+                        attributeBuilder.setVectorType(VertexAttribute.VectorType.VECTOR_TYPE_VEC4);
+                    }
+                    break;
+                case 9:
+                    attributeBuilder.setVectorType(VertexAttribute.VectorType.VECTOR_TYPE_MAT3);
+                    break;
+                case 16:
+                    attributeBuilder.setVectorType(VertexAttribute.VectorType.VECTOR_TYPE_MAT3);
```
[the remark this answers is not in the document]
Jhonnyg: Added a test 👍

editor/src/clj/editor/gl/vertex2.clj
```diff
@@ -231,7 +246,8 @@
         suffix           (keyword (or suffix "float"))
         num-components   (type-component-counts prefix)
         attribute-name   (name nm)
-        attribute-key    (attribute-name->key attribute-name)]
+        attribute-key    (attribute-name->key attribute-name)
+        semantic-type    (attribute-key->semantic-type attribute-key)] ; TODO: Typically determined by vertex-space setting of material.
```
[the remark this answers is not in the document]
Jhonnyg: Yeah I'm not sure I understand this either. Perhaps the todo is that we _wanted_ to include the coordinate system in this mess somehow?

editor/src/clj/editor/material.clj
```diff
@@ -228,9 +226,12 @@
        :columns (let [semantic-type-values (protobuf/enum-values Graphics$VertexAttribute$SemanticType)
                       data-type-values (protobuf/enum-values Graphics$VertexAttribute$DataType)
                       coordinate-space-values (protobuf/enum-values Graphics$CoordinateSpace)
+                      vector-type-values (protobuf/enum-values Graphics$VertexAttribute$VectorType)
+                      vertex-step-function (protobuf/enum-values Graphics$VertexStepFunction)
                       default-semantic-type :semantic-type-none
-                      default-element-count 3
-                      default-values (graphics/resize-doubles (vector-of :double) default-semantic-type default-element-count)]
+                      default-vector-type :vector-type-vec4
+                      default-values (graphics/resize-doubles (vector-of :double) default-semantic-type default-vector-type)
```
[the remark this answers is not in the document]
Jhonnyg: What we currently have is exactly how editing matrix uniforms work afaik. If you add a "user matrix" in the constants table it will default to zeroes and only show the top row of the matrix. That top row is editable so I think, *for now*, we should have the same behaviour and really really make sure to fix this soon :)

editor/src/clj/editor/graphics.clj
```diff
@@ -426,24 +481,30 @@
...
-(defn- attribute-expected-element-count [{:keys [element-count semantic-type] :as _attribute}]
-  {:pre [(integer? element-count)]}
+    (case (:vector-type attribute)
+      :vector-type-scalar g/Num
+      :vector-type-vec2 types/Vec2
+      :vector-type-vec3 types/Vec3
+      :vector-type-vec4 types/Vec4
+      :vector-type-mat2 types/Vec4
+      :vector-type-mat3 types/Vec4
+      :vector-type-mat4 types/Vec4)))
```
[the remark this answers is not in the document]
Jhonnyg: We didn't have matrix types before for attributes and since we can edit them as vec4's in the material editor I figured we'd want the same functionality for custom attributes.

editor/src/clj/editor/model_scene.clj
Jhonnyg: I think we need to do an actual type conversion from mat4 into mat3 (or mat2) here.

editor/test/resources/save_data_project/checked.material
Jhonnyg: Righ, yes that makes sense 👍 same goes for all vector types I suppose

editor/src/clj/editor/graphics.clj
```diff
@@ -592,6 +653,15 @@
...
 (def ^:private renderable-data->world-tangent-v3 (partial renderable-data->world-direction-v3 :tangent-data))
 
+(defn- matrix4+attribute->flat-array [^Matrix4d matrix attribute]
+  (let [vector-component-count (vector-type->component-count (:vector-type attribute))
+        matrix-flat-array (math/vecmath->clj (doto (Matrix4d. matrix) (.transpose)))
+        matrix-row-column-count (vtx/vertex-attribute->row-column-count attribute)]
+    (vec (flatten (if matrix-row-column-count
+                    (map #(take matrix-row-column-count %)
+                         (take matrix-row-column-count (partition 4 matrix-flat-array)))
+                    (take vector-component-count matrix-flat-array))))))
```
Jhonnyg, line 663: Maybe this can be written more efficiently? I couldn't find a nice way in the matrix class to grab a sub-matrix

PR #8754, opened 2024-04-03
Support environment-variable replacement for usernames in dependency URLs

matgis:
The editor can now inject environment variables into both the username and password of library dependency URLs that use authentication.

### DIFF
--- a/editor/src/clj/editor/library.clj
+++ b/editor/src/clj/editor/library.clj
@@ -99,17 +99,25 @@
     (io/copy is f)
     f))
 
+(defn- replace-user-info-env-variable [^String user-info-token]
+  (if (and (some? user-info-token)
+           (str/starts-with? user-info-token "__")
+           (str/ends-with? user-info-token "__"))
+    (let [env-key (subs user-info-token 2 (- (count user-info-token) 2))
+          env-value (System/getenv env-key)]
+      (or env-value user-info-token))
+    user-info-token))
+
 (defn- make-basic-auth-headers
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/library.clj
```diff
@@ -99,17 +99,25 @@
...
-      {"Authorization" (format "Basic %s" (str->b64 user-info))})))
-  
+  (let [[username password] (str/split user-info #":")
+        username (replace-user-info-env-variable username)
+        password (replace-user-info-env-variable password)
+        user-info (cond-> username
+                          password (str ":" password))
+        user-info-b64 (str->b64 user-info)
+        authorization (format "Basic %s" user-info-b64)]
+    {"Authorization" authorization}))
```
matgis, line 120: Previously only the password part of the URL supported environment variable replacement, but according to the docs we should also support it for the username: https://defold.com/manuals/libraries/#basic-access-authentication

editor/src/dev/dev.clj
```diff
@@ -957,7 +957,7 @@
                                                       num-strs)
                             first-col-width-fmt (str \% first-col-width \s)
                             rest-col-width-fmt (str \% rest-col-width \s)
-                            fmt-col (fn [index num-str]
+                            fmt-col (fn [^long index num-str]
```
matgis, line 960: Unrelated: Fixed boxed math warning.

vlaaad (approved):
LGTM!

PR #8769, opened 2024-04-08
Fixed broken gizmo manipulations for Model component

### LINKED ISSUE #8756: Wrong axis translation after model rotation
[Запись экрана от 04.04.2024 11:37:22.webm](https://github.com/defold/defold/assets/21050233/524359d2-fd53-4d42-944e-8dd7a4048754)

**To Reproduce (REQUIRED)**
1. Open Basic 3d Project
2. Rotate cube to Y=90
3. Translate cube to X or Z coordinates

Object need translated to where arrow is look

  - Version 1.7.0
  - 
**Platforms (REQUIRED):**
 - OS: Ubuntu, Mac OS

matgis:
* Fixed a regression where the move gizmo would translate rotated Models along the wrong axis.
* Fixed a regression where the manipulator gizmo would not be properly aligned to Model pivot points.

Fixes #8756

### DIFF
--- a/editor/src/clj/editor/model_scene.clj
+++ b/editor/src/clj/editor/model_scene.clj
@@ -236,12 +236,11 @@
 
 (defn- render-outline [^GL2 gl render-args renderables rcount]
   (assert (= 1 rcount) "Batching is disabled in the editor for simplicity.")
-  (let [pass (:pass render-args)
-        renderable (first renderables)
-        node-id (:node-id renderable)
-        request-id [::outline node-id]]
-    (assert (= pass/outline pass))
-    (render/render-aabb-outline gl render-args request-id renderables rcount)))
+  (when (= pass/outline (:pass render-args))
+    (let [renderable (first renderables)
+          node-id (:node-id renderable)
+          request-id [::outline node-id]]
+      (render/render-aabb-outline gl render-args request-id renderables rcount))))
 
 (g/defnk produce-mesh-set-build-target [_node-id resource content]
   (rig/make-mesh-set-build-target (resource/workspace resource) _node-id (:mesh-set content)))
@@ -324,8 +323,7 @@
 
 (defn- make-mesh-scene [renderable-mesh model-scene-resource-node-id]
   (let [{:keys [aabb material-name renderable-data]} renderable-mesh]
-    {:node-id model-scene-resource-node-id
-     :aabb aabb
+    {:aabb aabb
      :renderable {:render-fn render-mesh
                   :tags #{:model}
                   :batch-key nil ; Batching is disabled in the editor for simplicity.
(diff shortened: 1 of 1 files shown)


AGulev:
Does the fix work for both modes? 
CleanShot 2024-04-08 at 18 03 25@2x

editor/src/clj/editor/model_scene.clj
```diff
@@ -236,12 +236,11 @@
...
-        renderable (first renderables)
-        node-id (:node-id renderable)
-        request-id [::outline node-id]]
-    (assert (= pass/outline pass))
-    (render/render-aabb-outline gl render-args request-id renderables rcount)))
+  (when (= pass/outline (:pass render-args))
+    (let [renderable (first renderables)
+          node-id (:node-id renderable)
+          request-id [::outline node-id]]
+      (render/render-aabb-outline gl render-args request-id renderables rcount))))
```
matgis, line 243: Mostly indentation. Wrapped in `(when (= pass/outline (:pass render-args)) ...)` instead of asserting on the pass. We needed to add the renderable to `pass/selection` or `pass/opaque-selection` for it to be considered selectable.

editor/src/clj/editor/model_scene.clj
```diff
@@ -355,8 +352,8 @@
      :renderable {:render-fn render-outline
                   :tags #{:model :outline}
                   :batch-key nil ; Batching is disabled in the editor for simplicity.
-                  :select-batch-key model-scene-resource-node-id
-                  :passes [pass/outline]}
+                  :select-batch-key :not-rendered ; The render-fn only does anything during the outline pass.
+                  :passes [pass/outline pass/opaque-selection]} ; Include in a selection pass to ensure it can be selected and manipulated.
```
matgis, line 356: It seems the correct way to do this is to have the root `scene` include the `:node-id` of the `ResourceNode` (i.e. the `ModelSceneNode` in our case), and include a `:renderable` that is assigned to either `pass/selection` or `pass/opaque-selection`. It doesn't have to render anything, but this ensures it can be included in the `:selected-renderables` output of the `SceneRenderer`.

None of the `:children` scenes should specify a `:node-id`, unless they are actually from different nodes in the graph that can be transformed independently using the gizmo.

Jhonnyg (approved):

editor/src/clj/editor/model_scene.clj
```diff
@@ -355,8 +352,8 @@
      :renderable {:render-fn render-outline
                   :tags #{:model :outline}
                   :batch-key nil ; Batching is disabled in the editor for simplicity.
-                  :select-batch-key model-scene-resource-node-id
-                  :passes [pass/outline]}
+                  :select-batch-key :not-rendered ; The render-fn only does anything during the outline pass.
```
vlaaad, line 355: A bit confusing, at first I thought this is a specially recognized key, but it's not :)

vlaaad (approved):
LGTM!

PR #8772, opened 2024-04-08
Added diff functions to the dev module

matgis:
### Technical changes
* Added `dev/diff-values!` to print a structural diff between two values to stdout.
* Added `dev/diff-text!` to print a line diff of two values to stdout.

### DIFF
--- a/editor/src/dev/dev.clj
+++ b/editor/src/dev/dev.clj
@@ -20,6 +20,7 @@
             [editor.buffers :as buffers]
             [editor.changes-view :as changes-view]
             [editor.code.data :as code.data]
+            [editor.code.util :as code.util]
             [editor.collection :as collection]
             [editor.console :as console]
             [editor.curve-view :as curve-view]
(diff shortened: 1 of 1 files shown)


editor/src/dev/dev.clj
```diff
@@ -903,6 +907,10 @@
              (namespaced-class-symbol CursorRange)
              (partial object-data-pprint-handler nil code.data/cursor-range-print-data)
 
+             (namespaced-class-symbol RenderPass)
+             (fn render-pass-pprint-handler [printer ^RenderPass render-pass]
+               (fmt-doc printer (symbol "pass" (.nm render-pass))))
+
```
matgis, line 913: Pretty-print `RenderPass` values as names like `pass/opaque-selection` instead of a map of all its internals. We have a limited global set of passes, and this makes it a lot easier to inspect scene data, where these can be used as map keys.

vlaaad (approved):
Shipit!

PR #8837, opened 2024-04-23
Allow setting Image property when multi-selecting

### LINKED ISSUE #8721: Not possible to set Image property for multiple Sprite components
When selecting two or more Sprite components properties in Outline, the Image property disappears from the Properties pane.

**To Reproduce (REQUIRED)**
1. Add two or more sprite components
2. Select two or more sprite components (while holding Ctrl or Shift)
3. The Image property will disappear (as the only one property that disappears)

Image property shouldn't disappear and we should be able to set Image property for multiple sprites at once, as well as other properties.

  - 1.7.1

**Platforms (REQUIRED):**
 - Windows 10

**Screenshots (OPTIONAL):**
multiple_image_property

vlaaad:
Fixes #8721

### DIFF
--- a/editor/src/clj/editor/properties.clj
+++ b/editor/src/clj/editor/properties.clj
@@ -22,9 +22,11 @@
             [editor.protobuf :as protobuf]
             [editor.resource :as resource]
             [editor.types :as t]
-            [editor.util :as util]
+            [editor.util :as eutil]
             [editor.workspace :as workspace]
+            [internal.util :as util]
             [schema.core :as s]
+            [util.coll :refer [pair]]
             [util.id-vec :as iv]
             [util.murmur :as murmur])
   (:import [java.util StringTokenizer]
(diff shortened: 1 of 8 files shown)


editor/src/clj/editor/properties.clj
```diff
@@ -406,36 +408,46 @@
...
+                                             :errors (mapv (fn [{:keys [value error]}]
+                                                             (or error
+                                                                 (when (g/error? value) value))) v)
+                                             :edit-type (property-edit-type (first v))
+                                             :label (:label (first v))
+                                             :read-only? (reduce (fn [res read-only] (or res read-only)) false (map #(get % :read-only? false) v))}
+                                       default-vals (mapv :original-value (filter #(contains? % :original-value) v))
+                                       prop (if (empty? default-vals) prop (assoc prop :original-values default-vals))]
+                                   (pair k prop)))))
+                        visible-prop-colls)]
```
vlaaad, line 450: No change in logic, but I refactored this block to use transducers instead of lazy seqs when investigating the issue. The actual issue was that this block filters out properties with non-equal edit types, but we used anonymous functions for `:set-fn`/`:clear-fn` values in edit type.

editor/src/clj/editor/properties.clj
```diff
@@ -406,36 +408,46 @@
...
-                                        default-vals (mapv :original-value (filter #(contains? % :original-value) v))
-                                        prop (if (empty? default-vals) prop (assoc prop :original-values default-vals))]
-                                    [k prop]))
-                                common-props))]
+                                             :errors (mapv (fn [{:keys [value error]}]
+                                                             (or error
+                                                                 (when (g/error? value) value))) v)
+                                             :edit-type (property-edit-type (first v))
+                                             :label (:label (first v))
+                                             :read-only? (reduce (fn [res read-only] (or res read-only)) false (map #(get % :read-only? false) v))}
```
matgis, line 446: A candidate for `transduce`, maybe?

editor/src/clj/editor/sprite.clj
```diff
@@ -467,8 +474,7 @@
                        :error (validation/prop-error :info _node-id :texture validation/prop-nil? nil label)
                        :edit-type {:type resource/Resource
                                    :ext extension
-                                   :set-fn (fn [_ _ _ new]
-                                             (create-texture-binding-tx _node-id sampler-name new))}}])))))
+                                   :set-fn (fxui/partial set-texture-biding-id sampler-name)}}])))))
```
matgis, line 477: Perhaps we should consider moving `fxui/partial` to the `util.fn` module? It seems generally useful!

matgis (changes requested):
I think you left a debug `tap>` in by mistake? Looks good otherwise!

editor/src/clj/editor/properties.clj
```diff
@@ -406,36 +408,46 @@
...
+        coalesced (into {}
+                        (comp
+                          ; Filter out properties not common to *all* property sets
+                          ; Heuristic is to compare count and also type
+                          (filter
+                            (fn [e]
+                              (let [ret (and (= node-count (count (val e)))
+                                             (apply = (map property-edit-type (val e))))]
+                                (when-not ret
+                                  (tap> (map property-edit-type (val e))))
```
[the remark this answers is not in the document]
vlaaad: Duh, good catch!

matgis (approved):
Ship it!

PR #8839, opened 2024-04-23
Limit the number of rendered errors

### LINKED ISSUE #8678: Can't open an atlas with many broken dependencies in it.
I have atlas.atlas which has ~5k png links which aren’t on disk (I rename all the files on disk). I can’t open this atlas at all. I gave up and force close the editor after 10 minutes of freezing (not sure what’s going on there, maybe it tries to render all the errors on the screen, as I remember we print in right in the atlas view). I initially though it’s a regression bug, but it’s the same in 1.6.0.

**To Reproduce (REQUIRED)**
1. Open the `BigAtlases.project`
2.  Try to open `atlas.atlas`
3. The editor freeze for a long time and can't open it

Atlas that has broken dependencies (links to PNGs which don't exist on disk) opens fast and showing errors it has.

  - Version 1.7.0

**Platforms (REQUIRED):**
 - Platforms: Mac
 - OS: MacOS
 - Device: Macbook pro M1 Max

[BigAtlases.zip](https://github.com/defold/defold/files/14612216/BigAtlases.zip)

**Workaround (OPTIONAL):**
Manually cleanup the atlas in an external text editor

vlaaad:
Fixes #8678

### DIFF
--- a/editor/src/clj/editor/scene.clj
+++ b/editor/src/clj/editor/scene.clj
@@ -73,27 +73,35 @@
   (let [{:keys [resource] :as resource-node} (and node-id (g/node-by-id node-id))]
     (and resource (resource/resource-name resource))))
 
-(defn- root-causes
-  [error]
-  (->> (tree-seq :causes :causes error)
-       (filter :message)
-       (remove :causes)))
-
 (defn- render-error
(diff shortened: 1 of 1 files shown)


matgis (approved):
Looks good to me!

PR #8849, opened 2024-04-25
Annotations update

astrochili:
- Fixed the `b2d.get_body` url type — added `string` and `hash`.
- Added the `body` param to all the `b2d.body` methods, because that's the only way it works.
- Splitted the @return structure for `resource.create_texture_async`. Because it breaks the generation of the annotation structure now, which causes a bug both on the web page and in the Lua annotations for the language server. Maybe we need to fix the documentation meta file generation, but for now this is a working solution.


AGulev (approved):

AGulev:
@astrochili this is good one, thank you!

PR #8855, opened 2024-04-25
Increased editor UI contrast

### LINKED ISSUE #8384: Increase the editor UI contrast
Somehow, the editor contrast decreased over the last couple of years:
05-liveupdate-settings
image (1)

**Task (REQUIRED):**
We should make a pass on the editor CSS and improve the readability.

The text throughout the editor is more readable.

britzl:
The editor color palette has been reverted back to what it looked like a few years ago with a higher contrast between text and other ui elements and the background.

Fixes #8384

LEFT = CURRENT LOW CONTRAST PALETTE, RIGHT = UPDATED HIGH CONTRAST PALETTE IN THIS PR

Screenshot 2024-04-26 at 09 15 19
Screenshot 2024-04-26 at 09 14 18
Screenshot 2024-04-26 at 09 13 46
Screenshot 2024-04-26 at 08 17 15
Screenshot 2024-04-26 at 08 19 20
Screenshot 2024-04-26 at 09 40 47
Screenshot 2024-04-26 at 09 40 35

I was unable to build an editor from 2020, but there's a bunch of reference images in the documentation from that time:

https://github.com/defold/doc/tree/ce0109027b77aea1f05485b09e6e39bd1c98aa4e/docs/en/manuals/images

### DIFF
--- a/editor/resources/splash.css
+++ b/editor/resources/splash.css
@@ -66,20 +66,20 @@ Label.emphasis {
 }
 
 * {
-  -df-background-darker: derive(#212428, -13%);
-  -df-background-dark: derive(#212428, -6%);
+  -df-background-darker: derive(#212428, -40%);
   -df-background: #212428;
-  -df-background-light: derive(#212428, 6%);
-  -df-background-lighter: derive(#212428, 13%);
-  -df-component-darker: derive(#52575c, -30%);
-  -df-component-dark: derive(#52575c, -15%);
-  -df-component: #52575c;
-  -df-component-light: derive(#52575c, 15%);
-  -df-component-lighter: derive(#52575c, 30%);
-  -df-text-dark: derive(#a2b0be, -30%);
+  -df-background-light: #2c2e33;
+  -df-background-lighter: #3B3F43;
+  -df-background-input: #292a2f;
+  -df-component-darker: #212428;
+  -df-component-dark: #393c41;
+  -df-component-light: #45474B;
+  -df-component-lighter: #52575c;
+  -df-component-icon: #8f9296;
+  -df-text-dark: #848484;
   -df-text: #a2b0be;
-  -df-text-light: derive(#a2b0be, 85%);
-  -df-text-selected: derive(#a2b0be, 85%);
+  -df-text-light: #f0f2f6;
+  -df-text-selected: #f0f2f6;
   -df-defold-red-dark: derive(#e32f44, -10%);
   -df-defold-red: #e32f44;
   -df-defold-red-light: derive(#e32f44, 10%);
@@ -179,7 +179,7 @@ Label.emphasis {
 }
 
 .progress-indicator {
-  -fx-progress-color: -df-component;
+  -fx-progress-color: -df-component-lighter;
 }
 
 @font-face {
@@ -243,7 +243,7 @@ Label.emphasis {
 }
 
 .button:hover {
-  -fx-background-color: -df-component;
+  -fx-background-color: -df-component-lighter;
 }
 
 .button:armed {
--- a/editor/resources/splash.css
+++ b/editor/resources/splash.css
@@ -66,20 +66,20 @@ Label.emphasis {
 }
 
 * {
-  -df-background-darker: derive(#212428, -10%);
-  -df-background-dark: derive(#212428, -5%);
+  -df-background-darker: derive(#212428, -13%);
+  -df-background-dark: derive(#212428, -6%);
   -df-background: #212428;
-  -df-background-light: derive(#212428, 10%);
-  -df-background-lighter: derive(#212428, 20%);
-  -df-component-darker: derive(#464c55, -20%);
-  -df-component-dark: derive(#464c55, -10%);
-  -df-component: #464c55;
-  -df-component-light: derive(#464c55, 10%);
-  -df-component-lighter: derive(#464c55, 20%);
-  -df-text-darker: derive(#b4bac1, -20%);
-  -df-text-dark: derive(#b4bac1, -40%);
-  -df-text: #b4bac1;
-  -df-text-selected: derive(#b4bac1, 20%);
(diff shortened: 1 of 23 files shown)


AGulev:
@britzl could you pls rebase? Because now PR contains unrelated files

britzl:
> @britzl could you pls rebase? Because now PR contains unrelated files

Oh, thanks, missed that! Now merging to editor-dev.

AGulev:
Is it really higher contrast than it was? 
I mean it's become a bit blue, yes, but what contrast is the same, I think
Left is my experement (I think it's too much, yes). Right is how it was. In the middle it is your variant:
image

AGulev:
UPD: when I look at it - yes, it is brighter. It doesn't feel like that from the first view because background is different as well.

AGulev (approved):
It's hard to say if it's better without using it. It's different and feels unusual because of the fact I get used to having old colors. From the first view, the text doesn't feel brighter.

We should decide if we should just change it or think about having two different themes with a switcher if it's possible, which adds an ability to add light to them in the future.  

I approve this PR, but I think you should decide with Mats if we should ship it.

britzl:
The difference between what we have now and my revert back to the old color values is subtle, but there is a difference for sure. The contrast is increased between the background and components (buttons, input fields, sliders etc). Perhaps the contrast can be increased further, perhaps by 10-20% or so.

matgis (changes requested):
Wow, that's a lot of changes to the scss files. 😬

Just looking at the visual changes, I'd say there are some things that are better, some that made things worse, and some that are messed up.

* There is more contrast in editable fields, due to the higher text intensity and darker field backgrounds. However, the field labels now appear to have slightly less contrast against the background. Combined with the higher contrast field values, the labels are further diminished against the background due to their proximity to a higher-contrast element. I'm not sure if I find the end result to be less or more readable. I'm torn if I think we should make this change. 🤔
* The increased intensity of the text on highlighted and hovered items in the Assets and Outline panels, as well as the editor tabs and context menus is a clear improvement to me. I think we should do it.
* In form views, the intensity of text and field backgrounds is now more inconsistent than it was before. It was not fully consistent before either, but now there is a more pronounced difference between the field backgrounds of text fields, drop-down and table views in that both the backgrounds and the text intensities can have various values. I think we should make it consistent if we want to to make this change. I find the overall impression of editing a material slightly "messier" visually now.
* ⚠️ The section dividers in context menus straight up disappeared with this change. We need them back!
* ⚠️ The floating toolbar in scene views became very dark with this change. I don't think it was ever dark like that It's hard to see what the selected tool is. It's especially bad if you open the Visibility Filters popup, since the toggle button backgrounds now blend into the popup background so they no longer read as toggles. I think we need to restore the floating tool bar colors to what they were before.
* The contrast between the side panel and the main panel on the Welcome screen feels a bit too intense to me, compared to the other elements. I think the Welcome screen looked better before this change, but I concede the added contrast makes it more readable. Perhaps I just haven't acclimatized to the change yet. 🤔

britzl:
> * The section dividers in context menus straight up disappeared with this change. We need them back

I missed that!

Screenshot 2024-04-29 at 23 28 38

britzl:
> * The floating toolbar in scene views became very dark with this change

Good point. I issed that. Here's an update: 

Screenshot 2024-04-29 at 23 55 20

britzl:
> * There is more contrast in editable fields, due to the higher text intensity and darker field backgrounds. However, the field labels now appear to have slightly less contrast against the background. Combined with the higher contrast field values, the labels are further diminished against the background due to their proximity to a higher-contrast element. I'm not sure if I find the end result to be less or more readable. I'm torn if I think we should make this change.

Well, I went back to what we had in 2020 before we changed the styles. And in 2020 the field labels were slightly lower in contrast than the main field labels.

fieldvalues_current

I like the lower contrast field values. They are only used as hints of what's supposed to go into each field, and as such I think they should have a lower contrast than the main value.

britzl:
> * I think we should make it consistent if we want to to make this change. I find the overall impression of editing a material slightly "messier" visually now.

How is it messier? It looks almost exactly the same. The difference is that the text in the input fields are brighter. Otherwise they're the same, no?

Screenshot 2024-04-30 at 07 38 45

britzl:
> * The contrast between the side panel and the main panel on the Welcome screen feels a bit too intense to me, compared to the other elements. I think the Welcome screen looked better before this change, but I concede the added contrast makes it more readable. Perhaps I just haven't acclimatized to the change yet.

I don't have a super strong opinion on the Welcome Screen. Again, I went back to how it looked in 2020, before changes to the colors were made.

matgis (changes requested):
You fixed the toolbar in the Scene View, but the Visibility Filters popup that appears when you click the eye icon is still dark.

matgis:
> How is it messier? It looks almost exactly the same. The difference is that the text in the input fields are brighter. Otherwise they're the same, no?

There are now three distinct intensities of controls, whereas before it was two. It doesn't matter. We can do a pass on it later.

britzl:
> You fixed the toolbar in the Scene View, but the Visibility Filters popup that appears when you click the eye icon is still dark.

Thanks. Fixed:

Screenshot 2024-04-30 at 11 15 46

matgis (approved):
Go ahead and merge!

britzl:
> > How is it messier? It looks almost exactly the same. The difference is that the text in the input fields are brighter. Otherwise they're the same, no?
> 
> There are now three distinct intensities of controls, whereas before it was two. It doesn't matter. We can do a pass on it later.

True. I probably prefer what we have currently, but I didn't want to bring my own opinion into this change but rather revert to the original colours.

But I agree that we should make another pass over it at some point to perhaps increase contrast a bit more in some places, but also maybe tweak the cljfx form styling, which is a bit messy, especially when it comes to editable vs non editable fields and the table views.

PR #8525, opened 2024-02-10
Add camera focus message's wrappers.

ekharkunov:
Now it possible to activate/deactivate camera component via `camera.acquire_focus`/ `camera.release_focus`.
```
camera.acquire_focus("/observer#main_camera")
```
```
camera.release_focus("/observer#main_camera")
```

Fixes #3440.

## PR checklist

* [ ] Code
	* [ ] Add engine and/or editor unit tests.
	* [x] New and changed code follows the overall code style of existing code
	* [x] Add comments where needed
* [x] Documentation
	* [x] Make sure that API documentation is updated in code comments
	* [x] Make sure that manuals are updated (in github.com/defold/doc)
* [x] Prepare pull request and affected issue for automatic release notes generator
	* [x] Pull request - Write a message that explains what this pull request does. What was the problem? How was it solved? What are the changes to APIs or the new APIs introduced? This message will be used in the generated release notes. Make sure it is well written and understandable for a user of Defold.
	* [x] Pull request - Write a pull request title that in a sentence summarises what the pull request does. Do not include "Issue-1234 ..." in the title. This text will be used in the generated release notes.
	* [x] Pull request - Link the pull request to the issue(s) it is closing. Use on of the [approved closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
	* [x] Affected issue - Assign the issue to a project. Do not assign the pull request to a project if there is an issue which the pull request closes.
	* [x] Affected issue - Assign the "breaking change" label to the issue if introducing a breaking change.
	* [x] Affected issue - Assign the "skip release notes" is the issue should not be included in the generated release notes.

----------

Example of a well written PR description:

1. Start with the user facing changes. This will end up in the release notes.
1. Add one of the GitHub approved closing keywords
1. Optionally also add the technical changes made. This is information that might help the reviewer. It will not show up in the release notes. Technical changes are identified by a line starting with one of these:
   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```
There was a anomaly in the carbon chroniton propeller, introduced in version 8.10.2. This fix will make sure to reset the phaser collector on application startup.

Fixes #1234

### Technical changes
* Pay special attention to line 23 of phaser_collector.clj as it contains some interesting optimizations
* The propeller code was not taking into account a negative phase.
```

PR #8868, opened 2024-04-29
Camera component refactor + improved camera module

### LINKED ISSUE #3227: Make camera component useful (DEF-2984)
The camera component is not really useful today, which is why Blossom 2 have had to create our own camera system in order to do something reasonable with it.

Minimum what is needed to make it useful:

- [x] * DEF-2985. Camera properties should be properties that are available via go.get/set
- [x] * [DEF-2985. Camera properties should include both view and projection matrices as well as inverted versions of them.](https://github.com/defold/defold/issues/7761)
- [x] ~~* [DEF-2986. Render script should be able to get properties from any camera, by url]~~(https://github.com/defold/defold/issues/3717)
- [ ] #7763
- [x] * [DEF-2988. Editor should have a box where you can see what the camera sees.](https://github.com/defold/defold/issues/3226)
- [x] * Update documentation that says the fov is horizontal (it is actually vertical fov)

Jhonnyg:
The camera script module and the camera component has been updated:

* The camera component will automatically acquire focus when it is created. There is no more need to send the "acquire_camera_focus" message to the component. To disable the camera, use go.set("#camera", "disable") instead.

`NOTE:
This is a subtle change in behaviour. Previously you had to send a "acquire_camera_focus" message to the component for a camera to be updated. With the new code path, all cameras will be enabled and updated automatically until you actively send an "unacquire_camera_focus" message. Since the old behaviour was so cumbersome to work with, we **do not consider this as a breaking change** - instead we consider it to be a bugfix or simply as a workflow improvement.`

* All scripts can now access the `camera` namespace. There are new functions in the module that can now be used:
```lua
-- returns a list of all currently available cameras. each entry in the table is a camera URL (see msg.url)
local all_cameras = camera.get_cameras()
-- get/set properties. note that not all properties can be set
camera.get_projection(camera_url) -- read-only
camera.get_view(camera_url)-- read-only
camera.get_aspect_ratio(camera_url)
camera.set_aspect_ratio(camera_url)
camera.get_fov(camera_url)
camera.set_fov(camera_url)
camera.get_near_z(camera_url)
camera.set_near_z(camera_url)
camera.get_far_z(camera_url)
camera.set_far_z(camera_url)
camera.get_orthographic_zoom(camera_url)
camera.set_orthographic_zoom(camera_url)
```

* The render API has a new function `render.set_camera(camera_url, [options])` - this can be used in a render script to automatically use the view and projection matrices from the camera in the next draw call:
```lua
-- a typical camera based render loop can now look like this
for k,v in pairs(camera.get_cameras()) do
    -- note that the view and projection matrices previously set up
    -- using render.set_view and render.set_projection will be wiped by this call
    -- the 'use_frustum' option will use the cameras view projection matrix for frustum culling
    -- this will take precedence over any frustum passed in to the render.draw function
    render.set_camera(v, { use_frustum = true })
    -- set other drawing states
    render.draw(pred_scene)
end
-- unbind the camera for custom rendering
render.set_camera()
render.set_view(...)
render.set_projection(...)
render.draw(pred_gui)
```

The next iteration of improving the camera component involves:
* adding a viewport property to the camera that will automatically be set when calling render.set_camera
* adding coordinate system conversion functions (screen->world, world->screen, project/unproject and so on)

Fixes #3227


engine/gamesys/src/gamesys/components/comp_camera.cpp
```diff
@@ -124,6 +118,50 @@ namespace dmGameSystem
...
+    static inline dmMessage::URL CameraToURL(const CameraComponent* camera)
+    {
+        dmMessage::URL camera_url;
+        camera_url.m_Socket = dmGameObject::GetMessageSocket(dmGameObject::GetCollection(camera->m_Instance));
+        camera_url.m_Path   = dmGameObject::GetIdentifier(camera->m_Instance);
+        dmGameObject::GetComponentId(camera->m_Instance, camera->m_ComponentIndex, &camera_url.m_Fragment);
+        return camera_url;
+    }
+
+    static void CameraStackPop(CameraWorld* world, CameraComponent* camera)
```
Jhonnyg, line 130: We still use a camera stack because we keep track of a "main" camera based on the order for the stack. If the current main camera gets removed we need to know which was the previous main camera so we can put that into the camera.MAIN property

engine/gamesys/src/gamesys/components/comp_camera.cpp
```diff
@@ -190,63 +229,54 @@ namespace dmGameSystem
...
+        if (result != dmMessage::RESULT_OK)
+        {
+            dmLogError("The socket '%s' could not be found.", dmRender::RENDER_SOCKET_NAME);
+            return false;
+        }
+
+        dmMessage::Post(&camera_url, &receiver, dmGameSystemDDF::SetViewProjection::m_DDFDescriptor->m_NameHash,
+            0, (uintptr_t)dmGameSystemDDF::SetViewProjection::m_DDFDescriptor, &set_view_projection, sizeof(dmGameSystemDDF::SetViewProjection), 0);
+        return true;
+    }
```
Jhonnyg, line 257: This is here for backwards compatibility only, every camera is already available from the render script

engine/gamesys/src/gamesys/components/comp_camera.cpp
```diff
@@ -258,59 +288,26 @@ namespace dmGameSystem
...
+
+            dmGamesysDDF::SetCamera* ddf         = (dmGamesysDDF::SetCamera*)params.m_Message->m_Data;
+            camera_data.m_AspectRatio            = ddf->m_AspectRatio;
+            camera_data.m_Fov                    = ddf->m_Fov;
+            camera_data.m_NearZ                  = ddf->m_NearZ;
+            camera_data.m_FarZ                   = ddf->m_FarZ;
+            camera_data.m_OrthographicProjection = ddf->m_OrthographicProjection;
+            camera_data.m_OrthographicZoom       = ddf->m_OrthographicZoom;
+
+            dmRender::SetRenderCameraData(render_context, camera->m_RenderCamera, camera_data);
```
Jhonnyg, line 302: We could potentially return the pointer to the data to avoid calling get + set, but this is a bit cleaner imo

engine/render/src/render/render_script.cpp
```diff
@@ -2992,13 +2992,50 @@ namespace dmRender
...
+     * @param camera_id [type:url|handle|nil] camera id to use, or nil to reset
+     *
+     * @examples
+     *
+     * Set the current camera to be used for rendering
+     *
+     * ```lua
+     * render.set_camera("main:/my_go#camera")
+     * render.draw(self.my_pred)
+     * render.set_camera(nil)
```
Jhonnyg, line 3021: Maybe we should have a render.CAMERA_NONE or something that is nil, similar to render.RENDER_TARGET_DEFAULT

engine/render/src/render/render_script_camera.cpp
```diff
@@ -0,0 +1,314 @@
...
+     * Camera functions, messages and constants.
+     *
+     * @document
+     * @name Camera
+     * @namespace camera
+     */
+    #define RENDER_SCRIPT_CAMERA_MAIN
+    #define RENDER_SCRIPT_CAMERA_LIB_NAME           "camera"
+    #define RENDER_SCRIPT_CAMERA_MAIN_NAME          "CameraMain"
+    #define RENDER_SCRIPT_CAMERA_MAIN_PROPERTY_NAME "MAIN"
```
Jhonnyg, line 32: Should we call it camera.MAIN or camera.CAMERA_MAIN?

engine/render/src/render/render_script_camera.cpp
```diff
@@ -0,0 +1,314 @@
...
+        lua_settable(L, -3);
+
+        // TODO: Since we don't have any way of changing this (yet), we don't expose it.
+        //       The idea is to use a normalized viewport vector that can be set though the API
+        //       and via the editor, but it is the next part of this feature.
+        /*
+        lua_pushstring(L, "viewport");
+        dmScript::PushVector4(L, camera->m_Data.m_Viewport);
+        lua_settable(L, -3);
+        */
```
Jhonnyg, line 192: This should be a separate task since it involves proto and editor changes

engine/gamesys/src/gamesys/components/comp_camera.cpp
```diff
@@ -36,84 +36,78 @@ namespace dmGameSystem
...
+        uint16_t                m_ComponentIndex;
+        uint8_t                 m_AddedToUpdate : 1;
     };
 
     struct CameraWorld
     {
-        dmArray<CameraComponent> m_Cameras;
-        dmArray<CameraComponent*> m_FocusStack;
+        dmArray<CameraComponent>  m_Cameras;
+        dmArray<CameraComponent*> m_CameraStack;
```
Jhonnyg, line 56: We still use a camera stack because we keep track of a "main" camera based on the order for the stack. If the current main camera gets removed we need to know which was the previous main camera so we can put that into the camera.MAIN property

engine/gamesys/src/gamesys/components/comp_camera.cpp
```diff
@@ -36,84 +36,78 @@ namespace dmGameSystem
...
-    static const dmhash_t CAMERA_PROP_PROJECTION = dmHashString64("projection");
-    static const dmhash_t CAMERA_PROP_VIEW = dmHashString64("view");
-    static const dmhash_t CAMERA_PROP_ASPECT_RATIO = dmHashString64("aspect_ratio");
+    static const dmhash_t CAMERA_PROP_PROJECTION        = dmHashString64("projection");
+    static const dmhash_t CAMERA_PROP_VIEW              = dmHashString64("view");
+    static const dmhash_t CAMERA_PROP_ASPECT_RATIO      = dmHashString64("aspect_ratio");
 
 
-    void CompCameraUpdateViewProjection(CameraComponent* camera, dmRender::RenderContext* render_context)
+    static void CompCameraUpdateViewProjection(CameraComponent* camera, dmRender::RenderContext* render_context)
```
Jhonnyg, line 68: Perhaps this should be a dmRender function? Since the renderer owns the render camera and its data now, perhaps we should expose this?

dmRender::UpdateRenderCamera(context, camera, world_position)

engine/render/src/render/render_script_camera.cpp
```diff
@@ -0,0 +1,314 @@
...
+    {
+        {"__newindex",  RenderScriptCamera_newindex},
+        {"__index", RenderScriptCamera_index},
+        {0, 0}
+    };
+
+    static const luaL_reg RenderScriptCamera_Methods[] =
+    {
+        {"get_cameras", RenderScriptCamera_GetCameras},
+        {"get_info",    RenderScriptCamera_GetInfo},
```
Jhonnyg, line 276: Q: There is currently no "set_info" or similar to modify a camera (outside of properties), and I think perhaps that it's something people expect? If we decide to do that now I think we should move the update logic into the camera.cpp file instead of passing in explicit view and projection matrices.

engine/render/src/render/render.cpp
```diff
@@ -802,12 +805,22 @@ namespace dmRender
...
+            context->m_ViewProj   = context->m_CurrentRenderCamera->m_ViewProjection;
+        }
+
         // Cleared once per frame
         if (context->m_RenderListRanges.Empty())
         {
             SortRenderList(context);
         }
 
+        // TODO: There is no connection between render cameras and frustum currently
```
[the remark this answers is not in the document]
AGulev: I like option 3 here

engine/render/src/render/render_script_camera.cpp
```diff
@@ -0,0 +1,314 @@
...
+        PUSH_BOOL("auto_aspect_ratio",       m_AutoAspectRatio);
+    #undef PUSH_BOOL
+
+        return 1;
+    }
+
+    // This is called when getting a property on the camera module.
+    // If the 'MAIN' property is called, we return the URL to the current main camera.
+    // Otherwise, we just return the tble entry.
+    static int RenderScriptCamera_index(lua_State* L)
```
AGulev, line 224: Tbh I don't like this `camera.MAIN` property at all. I would prefer to have it like a separate function to make it clear and in alight with the rest engine's API

AGulev:
When we deprecate something we hide it from docs, should we hide `acquire_camera_focus ` and `release_camera_focus ` ?

engine/render/src/render/render_script_camera.cpp
AGulev: We have https://defold.com/ref/stable/camera/#set_camera message. And now it's a bit confusing and name clash with a function that does not the same what function does. Maybe we should deprecate it and replace with a new message `set_info`?

britzl: We do not have a single set_info() function in the current APIs, but we do have a bunch of different get_info() functions, such as sys.get_sys_info(), timer.get_info(), render.get_render_target_info() and so on.

I prefer if we either provide functions to set the various properties of the camera (set_view(), set_projection() and so on) OR as component properties which we can then set AND get.

engine/render/src/render/render_script_camera.cpp
britzl: Yes, this is the thing that stands out the most with this solution. It feels like something quite different from what we have elsewhere.

britzl:
> When we deprecate something we hide it from docs, should we hide `acquire_camera_focus ` and `release_camera_focus ` ?

Yes!

britzl (changes requested):
I've left some comments that I think we need to discuss

ekharkunov:
> The camera component will automatically acquire focus when it is created. There is no more need to send the "acquire_camera_focus" message to the component.
I've added function `camera.acquire_focus` here https://github.com/defold/defold/pull/8525. I guess it also should be deprecated.

engine/render/src/render/render_script_camera.cpp
JCash: I'm with Björn here. I prefer separate functions for better clarity.
And probably, as a first step, component properties, as they can be get/set and animated.

JCash (changes requested):
Looks good, but some API concerns that would be good to look into.

engine/render/src/render/camera.cpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    void SetRenderCameraMainCamera(HRenderContext render_context, HRenderCamera camera)
+    {
+        RenderCamera* c = render_context->m_RenderCameras.Get(camera);
+        if (c)
+        {
+            for (int i = 0; i < render_context->m_RenderCameras.Capacity(); ++i)
+            {
+                RenderCamera* c_other = render_context->m_RenderCameras.GetByIndex(i);
```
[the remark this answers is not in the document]
Jhonnyg: Yes I've thought about that as well, I think it was initially not intended to be iterated through like an array, but I have come across more and more of these occasions now that I need to do exactly that. But yes, it's a separate task

engine/render/src/render/render_private.h
```diff
@@ -261,18 +273,16 @@ namespace dmRender
 
         dmHashTable32<MaterialTagList>  m_MaterialTagLists;
 
-        HFontMap                    m_SystemFontMap;
+        dmOpaqueHandleContainer<RenderCamera> m_RenderCameras;
+        RenderCamera*                         m_CurrentRenderCamera; // When != 0, the renderer will use the matrices from this camera.
```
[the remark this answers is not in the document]
Jhonnyg: No, this is the camera that should be used during a draw call (if set)

engine/render/src/render/render_script_camera.cpp
```diff
@@ -0,0 +1,314 @@
...
+    */
+    static int RenderScriptCamera_GetInfo(lua_State* L)
+    {
+        DM_LUA_STACK_CHECK(L, 1);
+        RenderCamera* camera = CheckRenderCamera(L, 1, g_RenderScriptCameraModule.m_RenderContext);
+
+        lua_newtable(L);
+        lua_pushstring(L, "url");
+        dmScript::PushURL(L, camera->m_URL);
+        lua_settable(L, -3);
```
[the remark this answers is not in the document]
Jhonnyg: Aah, yeah. I keep forgetting these nuances of lua..

engine/render/src/render/camera.cpp
```diff
@@ -0,0 +1,126 @@
...
+                    continue;
+                }
+                if (c_other == c)
+                {
+                    c_other->m_IsMainCamera = 1;
+                    RenderScriptCameraSetMainCamera(c_other->m_URL);
+                }
+                else
+                {
+                    c_other->m_IsMainCamera = 0;
```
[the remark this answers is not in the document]
Jhonnyg: I think the flag was there to simplify the get_info() function, but we should be able to compare pointers or URLs instead

engine/render/src/render/render.h
```diff
@@ -328,6 +340,19 @@ namespace dmRender
...
+     * A render camera is a wrapper around common camera properties such as fov, near and far planes, viewport and aspect ratio.
+     * Within the engine, the render cameras are "owned" by the renderer, but they can be manipulated elsewhere.
+     * The render script can set current camera used for rendering by using render.set_camera(...), which will automatically
+     * take precedence over the view and projection matrices set by render.set_view() and render.set_projection().
+     */
+    HRenderCamera                   NewRenderCamera(HRenderContext context);
+    void                            DeleteRenderCamera(HRenderContext context, HRenderCamera camera);
+    void                            SetRenderCameraURL(HRenderContext render_context, HRenderCamera camera, const dmMessage::URL& camera_url);
+    void                            SetRenderCameraData(HRenderContext render_context, HRenderCamera camera, RenderCameraData data);
+    void                            SetRenderCameraMatrices(HRenderContext render_context, HRenderCamera camera, const dmVMath::Matrix4 view, const dmVMath::Matrix4 projection);
```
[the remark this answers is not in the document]
Jhonnyg: I think these functions will be removed anyway if we move the update code into dmRender. These are just here because the camera component does the update currently.

engine/render/src/render/render.cpp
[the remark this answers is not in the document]
Jhonnyg: Fixed

engine/render/src/render/render_script_camera.cpp
Jhonnyg: Hmm, I think perhaps we should remove the concept of having a main camera altogether. The point of having a "main" camera is to organize your rendering in a multi-camera setting, where you want one camera to be a "shadow" camera and some other camera to be a "scene" camera, and another to be a "postprocessing" camera or something like that. I think instead of a specific "main" camera like this, I think you would want to tag a camera with some made-up tag, such as "scene", "shadow", "postprocessing".. For now lets skip the main camera setup

engine/render/src/render/render_script_camera.cpp
Jhonnyg: Alright, I'll add functions to get the same properties as we already have in the camera (the data is already owned by the renderer now and not the camera component). Setting view and projection is a different story since we evaluate them on every update. So if we want to support setting custom view and projection matrices we need to keep track of the overrides and some way of going "back" to using the update-based matrices. For now I'll skip that however

AGulev (approved):
🔥

PR #8870, opened 2024-04-30
Add error reason for fwrite in sys.save

### LINKED ISSUE #8856: Output errno when fwrite fails in Sys_Save (and errno is supported)

Jhonnyg:
Fixed an issue where sys.save doesn't output the reason why it couldn't write to a file.

Fixes #8856

### DIFF
--- a/engine/script/src/script_sys.cpp
+++ b/engine/script/src/script_sys.cpp
@@ -161,13 +161,14 @@ union SaveLoadBuffer
         if (!file)
         {
             Sys_FreeTableSerializationBuffer(buffer);
-            #if !defined(DM_NO_ERRNO)
-                char errmsg[128] = {};
-                dmStrError(errmsg, sizeof(errmsg), errno);
-                return luaL_error(L, "Could not open the file %s, reason: %s.", tmp_filename, errmsg);
-            #else
-                return luaL_error(L, "Could not open the file %s", tmp_filename);
-            #endif
+
+        #if defined(DM_NO_ERRNO)
+            return luaL_error(L, "Could not open the file %s", tmp_filename);
+        #else
+            char errmsg[128] = {};
+            dmStrError(errmsg, sizeof(errmsg), errno);
+            return luaL_error(L, "Could not open the file %s, reason: %s.", tmp_filename, errmsg);
+        #endif
         }
 
         bool result = fwrite(buffer, 1, n_used, file) == n_used;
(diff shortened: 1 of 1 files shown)


engine/script/src/script_sys.cpp
```diff
@@ -161,13 +161,14 @@ union SaveLoadBuffer
...
-                return luaL_error(L, "Could not open the file %s", tmp_filename);
-            #endif
+
+        #if defined(DM_NO_ERRNO)
+            return luaL_error(L, "Could not open the file %s", tmp_filename);
+        #else
+            char errmsg[128] = {};
+            dmStrError(errmsg, sizeof(errmsg), errno);
+            return luaL_error(L, "Could not open the file %s, reason: %s.", tmp_filename, errmsg);
+        #endif
```
Jhonnyg, line 171: I flipped the double negativeness of the ifdef

engine/script/src/script_sys.cpp
```diff
@@ -177,7 +178,14 @@ union SaveLoadBuffer
         if (!result)
         {
             dmSys::Unlink(tmp_filename);
+
+        #if defined(DM_NO_ERRNO)
             return luaL_error(L, "Could not write to the file %s.", filename);
+        #else
+            char errmsg[128] = {};
+            dmStrError(errmsg, sizeof(errmsg), errno);
+            return luaL_error(L, "Could not write to the file %s, reason: %s.", filename, errmsg);
```
Jhonnyg, line 187: According to unix and windows manuals, fwrite _should_ set errno when it fails, so this should be enough..

engine/script/src/script_sys.cpp
JCash: `dmStrError()` does the same thing when DM_NO_ERRNO is defined (See dstrings.cpp:236)
That way you don't have to use another `ifdef`
I would recommend removing the "if defined(DM_NO_ERRNO)"

JCash (changes requested):
Approved, but with some minor change request.

engine/script/src/script_sys.cpp
Jhonnyg: Not sure I understand, in this case and after fopen a few lines above, we want to pass in the errno into the function to actually parse it, and if I remove the ifdef errno would be undefined for the platforms where it isn't supported, right?

PR #8871, opened 2024-04-30
Support tangent space handedness

### LINKED ISSUE #8733: Support tangent vector handedness
**Task (REQUIRED):**
Right now we export tangents from modelc as vec3s and then derive bitangents from the normal and the tangent, meaning we have no way of knowing if the tangent coordinate system is correct or not. A typical solution, if we don't want to export bitangents, is to pack the handedness direction in the .w component of the tangent vector, which can then be used to multiple the bitangent vector with after the derivation in the shader. Blender exports tangents in this way (and I think most do), so it should be a simple improvement.

Describe the expected outcome of the task (new issue, documentation, code change etc)

Jhonnyg:
Tangents are now produced as vec4 in the engine, where the fourth component is the handedness of the tangent vector. This can then be used to calculate the correct tangent space vectors in a shader:

`vec3 bitangent = cross(normal, tangent.xyz * tangent.w);`

Fixes #8733


editor/src/clj/editor/graphics.clj
```diff
@@ -314,7 +314,12 @@
           :semantic-type :semantic-type-normal
           :coordinate-space :coordinate-space-world
           :data-type :type-float
-          :element-count 3}]))
+          :element-count 3}
+         {:name "tangent"
+          :semantic-type :semantic-type-tangent
+          :coordinate-space :coordinate-space-world
+          :data-type :type-float
+          :element-count 4}]))
```
matgis, line 322: When using world-space meshes, the tangents in `:tangent-data` will be transformed by the `renderable-data->world-direction-v4` function. Currently, this writes a zero for the W component when it calls the `geom/transf-n4` function. I think we need to update `geom/transf-n4` to pass along the unaltered W component.

It seems to me, we would need to do something similar on the engine side since we presumably transform the tangents using the normal matrix somewhere there as well, but I'm not seeing it in this PR? 🤔

matgis (changes requested):
I think we need to pass on the unaltered W component of the tangent when transforming to world space? 🤔

JCash (dismissed):

editor/src/clj/editor/graphics.clj
Jhonnyg: The engine preserves the handedness in Rig.cpp in the "GenerateNormalData" function.

matgis (approved):
LGTM!

PR #8879, opened 2024-05-02
[Web] Fix sound play while AudioContext is suspended

### LINKED ISSUE #8238: Sounds in html5 not started until first user touch

**ERROR:SOUND: Out of sound data instance slots (256). Increase the project setting 'sound.max_sound_instances'\n"**

Browser block sound play until first user touch. So sounds not started and player reach limit.

This error not block player from playing in my case. But if someone use cb of sounds for logic it can break there games.

This example from my game. I play step sounds when player move.

**This is correct. I make touch and you can hear sounds.**

https://github.com/defold/defold/assets/2655263/3105c490-8453-48a8-a434-dbc35f8531bb

**This is error.**

https://github.com/defold/defold/assets/2655263/84357247-64f6-42c2-828a-75a3433ed960

Also you can hear in video that all sounds start playing on first touch.

**To Reproduce (REQUIRED)**
1. Start web build
2. Some sounds
3. Sounds not finished until user make first touch(browser not allow to play sounds untils first click or touch)
4. You get error when reach limit.

Not sure what is best case for that scenario. 
If it possible sounds should started and finished. If not mb we need sound.can_play() method

  - Version 1.6.1

**Platforms (REQUIRED):**
 - Platforms: HTML5

ekharkunov:
Fixes #8238 

### Technical changes
If AudioContext is suspended we calculate play offset but don't create audio buffers and audio sources on JS side.

## PR checklist

* [x] Code
* [x] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/engine/sound/src/js/library_sound.js
+++ b/engine/sound/src/js/library_sound.js
@@ -58,6 +58,9 @@ var LibrarySoundDevice =
                     var len = sample_count / this.sampleRate;
                     // use real buffer length next time.
                     this.bufferDuration = len;
+                    // only append overall length of audio buffer in suspended stay
+                    // it helps prevent sound instance consume on the engine side
+                    // because from engine point of view - sound plays.
                     if (!this._isContextRunning()) {
                         this.suspendedBufferedTo += len;
                         return;
@@ -89,6 +92,10 @@ var LibrarySoundDevice =
                             return 1;
                         ahead = this.bufferedTo - shared.audioCtx.currentTime;
                     } else {
+                        // if audio context in suspended or closed state we simulate audio play
+                        // by calculating play time
+                        // when audio context become active - start filling audio buffer from the beginning
+                        // and start using audioCtx.currentTime
                         ahead = this.suspendedBufferedTo - this._getCurrentSuspendedTime();
                     }
                     var inqueue = Math.ceil(ahead / this.bufferDuration);
--- a/engine/sound/src/js/library_sound.js
+++ b/engine/sound/src/js/library_sound.js
@@ -19,7 +19,7 @@ var LibrarySoundDevice =
             };
             window._dmJSDeviceShared = shared;
         }
-        
+
         var id = shared.count++;
         var device;
 
@@ -39,8 +39,29 @@ var LibrarySoundDevice =
                 sampleRate: shared.audioCtx.sampleRate,
                 bufferedTo: 0,
                 bufferDuration: 0,
+                creatingTime: Date.now() / 1000,
+                lastTimeInSuspendedState: Date.now() / 1000,
+                suspendedBufferedTo: 0,
                 // functions
+                _isContextRunning: function() {
+                    var audioCtx = window._dmJSDeviceShared.audioCtx;
+                    return audioCtx !== undefined && audioCtx.state == "running"
+                },
+                _getCurrentSuspendedTime: function() {
+                    if (!this._isContextRunning()) {
+                        this.lastTimeInSuspendedState = Date.now() / 1000;
+                        return this.lastTimeInSuspendedState - this.creatingTime;
+                    }
+                    return 0;
+                },
                 _queue: function(samples, sample_count) {
+                    var len = sample_count / this.sampleRate;
+                    // use real buffer length next time.
+                    this.bufferDuration = len;
+                    if (!this._isContextRunning()) {
+                        this.suspendedBufferedTo += len;
+                        return;
+                    }
                     var buf = shared.audioCtx.createBuffer(2, sample_count, this.sampleRate);
                     var c0 = buf.getChannelData(0);
                     var c1 = buf.getChannelData(1);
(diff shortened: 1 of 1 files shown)


AGulev:
What will happen if many sounds started without having context and then later game got the context. Will all of them play at the same time?

ekharkunov:
> What will happen if many sounds started without having context and then later game got the context. Will all of them play at the same time?

I've test in following setup:
* Play 30 seconds sound in `init` function of bootstrap collection.
* Print all messages from play callback
So when I load a page - console has log with 'sound_done' message immediately. So sound is removed after that. As result - when we start playing lots of sound without active audio context - they are all completed and when user click on canvas - no sounds will be played.

AGulev:
>So when I load a page - console has log with 'sound_done' message immediately. So sound is removed after that. 

Good. But in case when I run music that would probably mean I will never hear it because instead playing it will end immediately.

>As result - when we start playing lots of sound without active audio context - they are all completed and when user click on canvas - no sounds will be played.

What will happen if I play sound with looping option?

-----
What I thought about, maybe it's possible to consume sound buffers silently like if sound device exist. So we kinda play sounds but without playing it. And when the engine gets sound context it will continue playing from the place it is at now (I didn't investigate if it's possible to implement this way, but it would help to avoid all the issues with immediate callback and music that we start once in the beginning and will never hear because of that)

ekharkunov:
> What I thought about, maybe it's possible to consume sound buffers silently like if sound device exist. So we kinda play sounds but without playing it. And when the engine gets sound context it will continue playing from the place it is at now (I didn't investigate if it's possible to implement this way, but it would help to avoid all the issues with immediate callback and music that we start once in the beginning and will never hear because of that)

Almost can be done. What I found:
* if prevent buffer from queuing here https://github.com/defold/defold/blob/dev/engine/sound/src/sound.cpp#L1476 - sound played much more faster than it should be. If happend because we based on `free_slots` that returned from audio device (https://github.com/defold/defold/blob/dev/engine/sound/src/sound.cpp#L1459)
* if prevent buffer creating and playing on JS side - it almost works as needed because it based only on buffer length and `currentTime` from AudioContext https://github.com/defold/defold/blob/dev/engine/sound/src/js/library_sound.js#L66. But it's not working completely right because AudioContext.currentTime isn't changed while AudioContext stay in 'suspended' state.
It needs to be somehow simulate running timer while AudioContext is suspended and than use that elapsed time during `_freeBufferSlots` calculation.

AGulev:
> * if prevent buffer creating and playing on JS side - it almost works as needed because it based only on buffer length and `currentTime` from AudioContext https://github.com/defold/defold/blob/dev/engine/sound/src/js/library_sound.js#L66. But it's not working completely right because AudioContext.currentTime isn't changed while AudioContext stay in 'suspended' state.
>   It needs to be somehow simulate running timer while AudioContext is suspended and than use that elapsed time during `_freeBufferSlots` calculation.

What if we get 
>var initTime = Date.now() / 1000;

When we request to play sound without having a context. 
And then use 
> currentTime = (Date.now() / 1000) -  initTime

instead of `shared.audioCtx.currentTime` if context doesn't exist and then when context created save latest "synthetic" current time before creation and continue to use it as:
>currentTime  = currentTime  + shared.audioCtx.currentTime;

not sure if it will work, but if so, then we avoid any actions from user side, because for the engine it will be like engine plays sound without playing until it can actually play it

ekharkunov:
> > * if prevent buffer creating and playing on JS side - it almost works as needed because it based only on buffer length and `currentTime` from AudioContext https://github.com/defold/defold/blob/dev/engine/sound/src/js/library_sound.js#L66. But it's not working completely right because AudioContext.currentTime isn't changed while AudioContext stay in 'suspended' state.
> >   It needs to be somehow simulate running timer while AudioContext is suspended and than use that elapsed time during `_freeBufferSlots` calculation.
> 
> What if we get
> 
> > var initTime = Date.now() / 1000;
> 
> When we request to play sound without having a context for the first time and then use
> 
> > currentTime = (Date.now() / 1000) -  initTime
> 
> instead of `shared.audioCtx.currentTime` if context doesn't exist and then when context created save latest "synthetic" current time before creation and continue to use it as:
> 
> > currentTime  = currentTime  + shared.audioCtx.currentTime;
> 
> not sure if it will work, but if so, then we avoid any actions from user side, because for the engine it will be like engine plays sound without playing until it can actually play it

I'm trying several hours to implement in such way. But can't resolve issue to start sound from the middle. Now it correctly calculated and send 'sound_done' message but sound plays with the shift.

AGulev (dismissed):
I think this implementation is the best for the end user, because it doesn't require to think about the sound context and make it possible to keep logic as is.

But taking into account this is a bit of a hack, I would prefer to have a couple of sentences comment right in code with link to this PR where we had conversation about implementation, so the next time it will be touched, we know what's going on and why

AGulev (approved):
👍

PR #8891, opened 2024-05-06
Improve support for material constant types

Jhonnyg:
Currently Defold only supports setting and getting material constants via `go.set / go.get` - `vec4` and `mat4`. We can now support a wider range of types with this system:

Numbers and vector types: `float, vec2, vec3, vec4`
Matrix types: `mat2, mat3, mat4`

Example:
```glsl
// my_shader.fp
uniform vec2 mouse_pos;
uniform mat3 camera_rotation;
```

```
go.set("#model", "mouse_pos", vmath.vector3(self.mouse_x, self.mouse_y, 0))
go.set("#model" "camera_rotation", go.get_rotation(...))
```


Jhonnyg:
This has grown too stale to continue working on, I'm just gonna close it.

PR #8914, opened 2024-05-13
Bundle lua language server into the editor

### LINKED ISSUE #7704: LSP: bundle lua language server into the editor
**Task (REQUIRED):**
Lua language server is useful, we should provide it out of the box.

Lua language server does not require a dependency and works in the editor out of the box.

vlaaad:
This changeset includes the Lua language server into the editor distribution, so it's enabled in all projects. Yay!

Fixes #7704


matgis (approved):
Looks good to me! Just a question about the logging of unpacked resources. Presumably this was removed because it became quite spammy from all the additional files we unpack now?

vlaaad:
Yes exactly, too much spam in the logs.

PR #8920, opened 2024-05-14
Properly fixed overrides in gitattribute for protobuffer text files

redoak:
Replaced use of alias text-proto in gitattribute linguist-language, since it doesn't work.

This is a follow-up to my PR #8901 from last week.

I have tested this in my own repo. FYI, for me it sometimes takes a few reloads of the Github file browser to show the new syntax highlighting.

skip release notes


britzl:
> Replaced use of alias text-proto in gitattribute linguist-language, since it doesn't work.

Ok, unfortunate, but good to hear that it works now. I see now that `Protocol-Buffer-Text-Format` is indeed the correct name:

https://github.com/github-linguist/linguist/blob/master/lib/linguist/languages.yml#L5492

> skip release notes

Assign the label `skip release notes` to exclude this from generated release notes.

redoak:
> Assign the label `skip release notes` to exclude this from generated release notes.

I don't appear to have permissions to add labels. Github docs says it requires triage access. Perhaps it's a good thing that not just anyone can come in and add or remove a million labels though.

britzl:
> > Assign the label `skip release notes` to exclude this from generated release notes.
> 
> I don't appear to have permissions to add labels. Github docs says it requires triage access. Perhaps it's a good thing that not just anyone can come in and add or remove a million labels though.

Good point!

PR #8922, opened 2024-05-14
Add Defold YAML file types to gitattributes

redoak:
In gitattribute, set file types appmanifest, manifest, and script_api to use YAML linguist-language.

I have tested this in my own repo.

skip release notes


britzl (approved):
Thanks

PR #8932, opened 2024-05-15
Make the ASCII character set optional in fonts

### LINKED ISSUE #8070: Fonts: make the ASCII character set optional
Usually we use several fonts in our games to nicely display different numerical information (damage value, speed dial and so on). And Defold includes **all ASCII characters in fonts**. However, for this case, it doesn’t make sense to store unused glyphs in the fonts . Especially if it’s a web game where the download size matters.

Add an option to not include ASCII characters, only those specified in Extra Chars. To avoid using third party tools as in the link below.

https://forum.defold.com/t/big-list-of-defold-pro-tips/1519/111?u=aglitchman

AGulev:
The `Extra Characters` field has been deprecated and replaced with a new field called `Characters`. This new field combines the default ASCII with any previously defined `Extra Characters`. Using this new field, developers can specify exactly which symbols the font should include, e.g. numbers only, etc.

Fix https://github.com/defold/defold/issues/8070
## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/font/Fontc.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/font/Fontc.java
@@ -52,6 +52,8 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.HashSet;
+import java.util.Set;
 
 import javax.imageio.ImageIO;
 
@@ -74,6 +76,7 @@ import com.dynamo.render.proto.Font.FontTextureFormat;
 import com.dynamo.render.proto.Font.FontRenderMode;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.TextFormat;
+import org.apache.commons.lang3.StringUtils;
 
 class OrderComparator implements Comparator<Fontc.Glyph> {
     @Override
@@ -117,6 +120,16 @@ class BlendComposite implements Composite {
 }
 
 public class Fontc {
+
+    public static final char[] ASCII_7BIT;
+    static {
+        int start = 32;
+        int end = 126;
+        ASCII_7BIT = new char[end - start + 1];
+        for (int i = start; i <= end; i++) {
+            ASCII_7BIT[i - start] = (char) i;
+        }
+    }
 
     public enum InputFontFormat {
         FORMAT_TRUETYPE,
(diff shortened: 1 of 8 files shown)


AGulev:
PR isn't finished. Editor changes needed

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/font/Fontc.java
```diff
@@ -118,6 +121,16 @@ public CompositeContext createContext(ColorModel arg0, ColorModel arg1, Renderin
...
 
+    public static final char[] ASCII_7BIT;
+    static {
+        int start = 32;
+        int end = 126;
+        ASCII_7BIT = new char[end - start + 1];
+        for (int i = start; i <= end; i++) {
+            ASCII_7BIT[i - start] = (char) i;
+        }
+    }
```
AGulev, line 132: It would be better to have only one place where default symbols specified

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/font/Fontc.java
```diff
@@ -231,17 +244,26 @@ public void TTFBuilder(InputStream fontStream) throws FontFormatException, IOExc
...
-            // 7-bit ASCII. Note inclusive range [32,126]
-            for (int i = 32; i <= 126; ++i) {
-                characters.add(i);
+            String chars = fontDesc.getCharacters();
+            if (!StringUtils.isEmpty(chars)) {
+                for (int i = 0; i < chars.length(); i++) {
+                    char c = chars.charAt(i);
+                    characters.add((int) c);
+                }
             }
```
AGulev, line 253: logic that can read both Old and New path. If new chars field is filled, then use this one, else use old path

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/font/Fontc.java
```diff
@@ -231,17 +244,26 @@ public void TTFBuilder(InputStream fontStream) throws FontFormatException, IOExc
...
+                }
+                String extraCharacters = fontDesc.getExtraCharacters();
+                for (int i = 0; i < extraCharacters.length(); i++) {
+                    char c = extraCharacters.charAt(i);
+                    characters.add((int) c);
+                }
             }
         }
+        Set<Integer> deDup = new HashSet<Integer>(characters);
+        characters = new ArrayList<Integer>(deDup);
```
AGulev, line 266: a trick for deduplication

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/font/Fontc.java
```diff
@@ -231,17 +244,26 @@ public void TTFBuilder(InputStream fontStream) throws FontFormatException, IOExc
...
+                    characters.add((int) c);
+                }
+                String extraCharacters = fontDesc.getExtraCharacters();
+                for (int i = 0; i < extraCharacters.length(); i++) {
+                    char c = extraCharacters.charAt(i);
+                    characters.add((int) c);
+                }
             }
         }
+        Set<Integer> deDup = new HashSet<Integer>(characters);
```
matgis, line 265: Good thinking with the de-duplication!

However, you probably want to use a `SortedSet` here to ensure you get consistent ordering between builds?

AGulev: It makes sense. Thank you! I replaced it with a `TreeSet`

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/font/Fontc.java
britzl: Do we use TreeSet because we need a sorted list? Otherwise HashSet is probably faster. Although I'm not sure it will have much impact on performance anyway.

britzl (approved):
Looks good. I had a question and suggestion, but nothing to prevent this from being merged

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/font/Fontc.java
AGulev: yes, Mats made a assumption that it would be nice to have consistent output between builds, and I agree with him:

>However, you probably want to use a SortedSet here to ensure you get consistent ordering between builds?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/font/Fontc.java
AGulev: I know that we usually don't do static initialization like that, but in this case, having it like this explains the logic behind this string. If you don't mind, I keep it this way.

PR #8939, opened 2024-05-16
Add extension-teal tests to the editor

matgis:
### Technical changes
* Added editor tests for `extension-teal`.
* In `dev` builds of the editor, JVM properties will be injected into URL setting values.
* Dependency URLs for official extensions are now injected as JVM properties in the format `defold.extension.teal.url=https://...` defined in `project.clj` into editor test projects.
* In `dev` builds of the editor JVM properties in the format `defold.extension.teal.path=/Users/.../Repos/extension-teal` will cause the editor to use the contents of the specified directory in place of the corresponding `defold.extension.*.url` when fetching libraries.
* When registering resource types, you can now specify `:built-pb-class <ProtobufMessageClass>` to aid test functions such as `test-util/make-build-output-infos-by-path` to deserialize built binaries into Clojure maps.


editor/src/clj/editor/protobuf.clj
```diff
@@ -55,6 +55,9 @@ Macros currently mean no foreseeable performance gain however."
       (.getBytes StandardCharsets/UTF_8)
       (TextFormat/escapeBytes)))
 
+(defn pb-class? [value]
+  (isa? value Message))
```
vlaaad, line 59: `isa?` might be too open, i.e. it will return true both for subclasses of Message and values that are instances of the class. Maybe use something like `(.isAssignableFrom Message value)`?

vlaaad (dismissed):
Great addition to the test suite! 🔥

editor/src/clj/editor/protobuf.clj
matgis: Good point! We also do not want to allow the `Message` class itself here. I'll fix!

vlaaad (approved):
LGTM!

PR #8942, opened 2024-05-17
Editor: Unversion vcs.xml from .idea since it contains local settings

matgis:
### Technical changes
* This removes `vcs.xml` from the `.idea` directory and adds it into `.gitignore`, since it contains version-control-related settings derived from the local environment. Idea will re-write this file when you reopen and save the project. Having `vcs.xml` untracked lets users list additional `:source-paths` in `~/.lein/profiles.clj` without worrying about their VCS configuration.


vlaaad (approved):
LGTM!

PR #8956, opened 2024-05-19
Added Android bundler option to extract native libs on install

### LINKED ISSUE #8955: Android bundler should respect the extractNativeLibs manifest option
The AndroidManifest.xml can have an application attribute `android:extractNativeLibs` which controls wether native libs are extracted from the APK when installed. If the native libs are not extracted it is expected that the native libs remain in the APK uncompressed.

We need a game.project setting to control `extractNativeLibs` and if not set, the Android bundler must make sure that any libs are uncompressed when the APK is created.

britzl:
The Android bundler will now respect the `android:extractNativeLibs` attribute on the application level in AndroidManifest.xml. The option is configurable from `game.project` in the Android section.

This attribute indicates whether the package installer extracts native libraries from the APK to the file system. If set to "false", your native libraries are stored uncompressed in the APK. Although your APK might be larger, your application loads faster because the libraries load directly from the APK at runtime.

### IMPORTANT
If you are using a custom AndroidManifest.xml and would like to use this new option please make sure to add the `android:extractNativeLibs` attribute to the `<application>` tag:

```lua
<application
        android:extractNativeLibs="{{android.extract_native_libs}}"
```

Fixes #8955


engine/engine/content/builtins/manifests/android/AndroidManifest.xml
```diff
@@ -13,6 +13,7 @@
         {{#has-icons?}}
         android:icon="@drawable/icon"
         {{/has-icons?}}
+        android:extractNativeLibs="{{android.extract_native_libs}}"
```
AGulev, line 16: For all the changes in `AndroidManifest.xml` it's useful to add a warning that now users have to update their custom Android Manifests.

editor/resources/meta.edn
```diff
@@ -720,6 +720,10 @@
    :label "ProGuard config",
    :default "",
    :path ["android" "proguard"]}
+  {:type :boolean,
+   :help "This attribute indicates whether the package installer extracts native libraries from the APK to the file system",
+   :default true,
+   :path ["android" "extract_native_libs"]}
```
AGulev, line 726: It worth adding it into documentation with a link to the official Android documentation

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/AndroidBundler.java
```diff
@@ -682,13 +682,21 @@ private static File createAppBundleBaseZip(Project project, File outDir, File ap
...
             File aabDir = new File(outDir, "aab");
             File baseConfig = new File(aabDir, "BundleConfig.json");
             try (BufferedWriter writer = new BufferedWriter(new FileWriter(baseConfig))) {
-                writer.write("{\"compression\":{\"uncompressedGlob\": [\"assets/game.arcd\"]}}");
+                String uncompressedGlob = "\"assets/game.arcd\"";
+                if (!extractNativeLibs) {
+                    uncompressedGlob += ",";
+                    uncompressedGlob += "\"lib/**/*.so\"";
+                }
+                writer.write("{\"compression\":{\"uncompressedGlob\": [" + uncompressedGlob + "]}}");
```
AGulev, line 699: At some point it will be useful to give access to that file as well as we do for manifest files, but it's good for now.

AGulev (dismissed):

AGulev:
Make sure you fixed the Editor's test before merge

editor/resources/meta.edn
AGulev: It's nice option to have, but I don't see a situation when I want to use it tbh

britzl: Documentation PR here: https://github.com/defold/doc/pull/441

AGulev: Ah, yes. I forgot

AGulev:
The edtor test check that all the values aren't default, so you need to fix the game.project file in failed test

AGulev (approved):

PR #8959, opened 2024-05-20
Don't suggest completion popup unnecessarily

### LINKED ISSUE #8946: Editor autocomplete suggests global variable _G
**Describe the bug**
In some situations, the editor suggests the global variable _G (and other items) when it would be better to suggest nothing.

**To Reproduce**
1. In the code editor type an opening curly brace or comma
2. Notice that a list of suggestions is given, with _G at the top
3. Hit return
4. _G has been inserted

**Expected behavior**
No autocomplete occurs. I can imagine the autocomplete might be more useful after a comma in other situations, maybe a function call.

**Defold version:**
 1.8.0 / 1.8.1 beta

**Platforms:**
 - Platforms: macOS
 - Device: Macbook Pro

**Screenshots :**

image
image
image

vlaaad:
The editor no longer shows the completion popup after typing `{` and `,`.

Fixes #8946

### DIFF
--- a/editor/src/clj/editor/code/script.clj
+++ b/editor/src/clj/editor/code/script.clj
@@ -59,6 +59,7 @@
                        :field #{"."}
                        :module #{"."}}
    :completion-trigger-characters #{"."}
+   :ignored-completion-trigger-characters #{"{" ","}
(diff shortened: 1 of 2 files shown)


matgis (approved):
Ship it!

PR #8986, opened 2024-05-27
Allow adding multiple components at once

### LINKED ISSUE #6638: Editor should allow adding multiple components at once
2022-05-26 14_38_54-Select Component File

I have to add each of these one at a time...

Add Component File dialog should enable selecting multiple component files at once to add.

AGulev:
Now it's possible to add multiple components to a GameObject at once.

Fix https://github.com/defold/defold/issues/6638
## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/editor/src/clj/editor/game_object.clj
+++ b/editor/src/clj/editor/game_object.clj
@@ -482,7 +482,7 @@
                                  ext))
                              (workspace/get-resource-type-map workspace))]
     (when-let [resources (resource-dialog/make workspace project {:ext component-exts :title "Select Component File" :selection :multiple})]
-      (doseq [resource (if (map? resources) (vals resources) resources)]
+      (doseq [resource resources]
         (add-referenced-component! go-id resource select-fn)))))
 
 (defn- selection->game-object [selection]
--- a/editor/src/clj/editor/game_object.clj
+++ b/editor/src/clj/editor/game_object.clj
@@ -481,8 +481,9 @@
                                          (contains? tags :embeddable))
                                  ext))
                              (workspace/get-resource-type-map workspace))]
-    (when-let [resource (first (resource-dialog/make workspace project {:ext component-exts :title "Select Component File"}))]
-      (add-referenced-component! go-id resource select-fn))))
+    (when-let [resources (resource-dialog/make workspace project {:ext component-exts :title "Select Component File" :selection :multiple})]
+      (doseq [resource (if (map? resources) (vals resources) resources)]
+        (add-referenced-component! go-id resource select-fn)))))
 
 (defn- selection->game-object [selection]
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/game_object.clj
```diff
@@ -481,8 +481,9 @@
                                          (contains? tags :embeddable))
                                  ext))
                              (workspace/get-resource-type-map workspace))]
-    (when-let [resource (first (resource-dialog/make workspace project {:ext component-exts :title "Select Component File"}))]
-      (add-referenced-component! go-id resource select-fn))))
+    (when-let [resources (resource-dialog/make workspace project {:ext component-exts :title "Select Component File" :selection :multiple})]
+      (doseq [resource (if (map? resources) (vals resources) resources)]
+        (add-referenced-component! go-id resource select-fn)))))
```
AGulev, line 486: It feels strange we didn't do it before.
Is it because in such case each adding will be a separate action and we will undo it one by one? Is it really a problem?

matgis: I think it may just be that we added the ability to select multiple items in the dialog after this code was written. 🤔

editor/src/clj/editor/game_object.clj
```diff
@@ -481,8 +481,9 @@
                                          (contains? tags :embeddable))
                                  ext))
                              (workspace/get-resource-type-map workspace))]
-    (when-let [resource (first (resource-dialog/make workspace project {:ext component-exts :title "Select Component File"}))]
-      (add-referenced-component! go-id resource select-fn))))
+    (when-let [resources (resource-dialog/make workspace project {:ext component-exts :title "Select Component File" :selection :multiple})]
+      (doseq [resource (if (map? resources) (vals resources) resources)]
```
matgis, line 485: Hmm, I don't think the value returned by `resource-dialog/make` can be a map? 🤔

The docstring says it can return either a list of selected items, or `nil` if they clicked Cancel. Am I missing something?

I think we should remove the `map?` check if it can't be a map, since is may be a bit confusing.

matgis (changes requested):
Slight weirdness around the return value of `resource-dialog/make`, but looks good otherwise!

editor/src/clj/editor/game_object.clj
AGulev: I was fighting with this bug for an hour or so yesterday. But now, when I run it, it works just fine. Maybe I broke something else and then, when I fixed it, I forgot to reload it. (Mental note: if something goes unexpectedly wrong, just re-run the editor.)

thank you!

matgis (approved):
LGTM! ❤️

PR #8990, opened 2024-05-27
Editor: Use explicit list of characters in .font files

matgis:
Editor part of #8070.

### DIFF
--- a/editor/src/clj/editor/font.clj
+++ b/editor/src/clj/editor/font.clj
@@ -494,7 +494,7 @@
                                   :passes [pass/transparent]}))))
 
 (g/defnk produce-pb-msg [pb font material size antialias alpha outline-alpha outline-width
-                         shadow-alpha shadow-blur shadow-x shadow-y extra-characters output-format
+                         shadow-alpha shadow-blur shadow-x shadow-y characters output-format
                          all-chars cache-width cache-height render-mode]
   ;; The reason we use the originally loaded pb is to retain any default values
   ;; This is important for the saved file to not contain more data than it would
@@ -511,7 +511,7 @@
           :shadow-blur shadow-blur
           :shadow-x shadow-x
           :shadow-y shadow-y
-          :extra-characters extra-characters
+          :characters characters
           :output-format output-format
           :all-chars all-chars
           :cache-width cache-width
@@ -741,14 +741,23 @@
             (dynamic visible output-format-defold-or-distance-field?))
   (property shadow-y g/Num
             (dynamic visible output-format-defold-or-distance-field?))
-  (property extra-characters g/Str
-            (dynamic visible output-format-defold-or-distance-field?))
-  (property all-chars g/Bool
-            (dynamic visible output-format-defold-or-distance-field?))
   (property cache-width g/Int
             (dynamic error (validation/prop-error-fnk :fatal validation/prop-negative? cache-width)))
   (property cache-height g/Int
             (dynamic error (validation/prop-error-fnk :fatal validation/prop-negative? cache-height)))
+  (property characters g/Str (default (protobuf/default Font$FontDesc :characters))
+            (dynamic visible output-format-defold-or-distance-field?)
+            (dynamic read-only? (gu/passthrough all-chars))
+            (set (fn [evaluation-context self _old-value new-value]
+                   ;; It is illegal to have no characters in the font, but it is
+                   ;; also useful to be able to reset the field to the default
+                   ;; set of characters. If the user clears the field, we reset
+                   ;; to the printable ASCII characters.
+                   (when (and (= "" new-value)
+                              (properties/user-edit? self :characters evaluation-context))
+                     (g/set-property self :characters fontc/default-characters-string)))))
+  (property all-chars g/Bool
+            (dynamic visible output-format-defold-or-distance-field?))
 
   (property pb g/Any (dynamic visible (g/constantly false)))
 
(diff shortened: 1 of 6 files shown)


editor/src/clj/editor/pipeline/fontc.clj
```diff
@@ -16,22 +16,24 @@
...
+           [java.awt BasicStroke Canvas Color Composite CompositeContext Font FontMetrics Graphics2D RenderingHints Shape Transparency]
            [java.awt.color ColorSpace]
-           [java.io InputStream FileNotFoundException IOException]
+           [java.awt.font FontRenderContext GlyphVector]
+           [java.awt.geom AffineTransform FlatteningPathIterator PathIterator]
+           [java.awt.image BufferedImage ComponentColorModel ConvolveOp DataBuffer DataBufferByte Kernel Raster WritableRaster]
+           [java.io FileNotFoundException IOException InputStream]
            [java.nio.file Paths]
+           [java.util Arrays]
+           [javax.imageio ImageIO]
```
matgis, line 29: I ran **Code > Optimize Imports** on these to simplify the merge with the changes on my other branch.

editor/src/clj/editor/pipeline/fontc.clj
```diff
@@ -392,7 +394,11 @@
 (defn- ttf-semi-glyphs [font-desc ^Font font antialias]
   (let [prospect-codepoints (if (:all-chars font-desc)
                               (range 0x10ffff)
-                              (concat (range 32 127) (map int (:extra-characters font-desc))))
+                              (sort
+                                (eduction
+                                  (map int)
+                                  (distinct)
+                                  (:characters font-desc))))
```
matgis, line 401: `eduction` lets us combine several transformations like `(map int)` and `(distinct)` on the sequence `(:characters font-desc)` without generating intermediate collections.

editor/src/clj/editor/pipeline/fontc.clj
```diff
@@ -597,8 +603,8 @@
             (aset image out-shadow-offset (unchecked-byte df-outline-channel)))))
       (when (and (> shadow-alpha 0.0) (> shadow-blur 0) (> channel-count 1))
         (let [image-byte-length (* width height channel-count)
-              shadow-image' (byte-array image-byte-length)
-              shadow-image (do (System/arraycopy image 0 shadow-image' 0 image-byte-length) shadow-image')
+              shadow-image (byte-array image-byte-length)
+              _ (System/arraycopy image 0 shadow-image 0 image-byte-length)
```
matgis, line 607: Unwrapped the `(do ... shadow-image')` expression here, since Cursive couldn't infer out the correct `DataBufferByte` constructor from the types on the next line. Instead, I do the `arrayCopy` call inline and ignore its return value.

editor/test/integration/save_data_test.clj
```diff
@@ -612,6 +616,11 @@
         (is (= [2.0 2.0 2.0] (g/node-value embedded-go :scale)))
         (is (= [2.0 2.0 2.0] (g/node-value referenced-go :scale)))))
 
+    (testing "font"
+      (let [extra-characters-font (project/get-resource-node project "/silently_migrated/extra_characters.font")]
+        (is (= " !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~åäö"
+               (g/node-value extra-characters-font :characters)))))
+
```
matgis, line 623: Test that the silent migration moves any `:extra-characters` into `:characters`. The `åäö` came from `:extra-characters`.

editor/test/resources/save_data_project/silently_migrated/extra_characters.font
```diff
@@ -0,0 +1,17 @@
...
+size: 15
+antialias: 1
+alpha: 1.0
+outline_alpha: 0.0
+outline_width: 0.0
+shadow_alpha: 0.0
+shadow_blur: 0
+shadow_x: 0.0
+shadow_y: 0.0
+extra_characters: "\303\245\303\244\303\266"
```
matgis, line 12: `åäö` encodes like this in protobuf text format.

AGulev (approved):
Is it even real to know about all the functions like  `eduction` and `(pos? (count characters))`? =)
When I read it for me it's like "What" "Whaaat" =)

matgis:
If it makes you feel better, I only learned about `eduction` about a year or so ago. 😅

It's not something you really have to know, but it can be more efficient than the alternatives in certain cases.

editor/src/clj/editor/font.clj
```diff
@@ -741,14 +741,23 @@
...
+            (dynamic visible output-format-defold-or-distance-field?)
+            (dynamic read-only? (gu/passthrough all-chars))
+            (set (fn [evaluation-context self _old-value new-value]
+                   ;; It is illegal to have no characters in the font, but it is
+                   ;; also useful to be able to reset the field to the default
+                   ;; set of characters. If the user clears the field, we reset
+                   ;; to the printable ASCII characters.
+                   (when (and (= "" new-value)
+                              (properties/user-edit? self :characters evaluation-context))
+                     (g/set-property self :characters fontc/default-characters-string)))))
```
vlaaad, line 758: Should we also dedupe the characters here?

vlaaad (approved):
LGTM! Maybe we should dedupe the characters when setting them?

editor/src/clj/editor/font.clj
matgis: Since this is a user-facing field, I don't think we shouldn't interfere with their content. I figure they may want to paste in lines of dialogue, or things of similar nature, and it would be weird if we mangled them here.

PR #8995, opened 2024-05-28
Fixed black screen on iOS launch

### LINKED ISSUE #6142: Black screen blinking on iOS with a custom LaunchScreen.storyboard
Launch screen `blinks to the black screen and then` fades to the game screen with the ‘clear_color’ background.

**To Reproduce (REQUIRED)**
Bundle the repro project to iOS device and run.

Launch screen fades to the game screen with the ‘clear_color’ background.

  - Defold 1.2.186

**Platforms (REQUIRED):**
 - Platforms: iOS
 - OS: iOS 15
 - Device: iPhone 12 mini

[ios-launch-screen-black-blink.zip](https://github.com/defold/defold/files/7429106/ios-launch-screen-black-blink.zip)

**Workaround (OPTIONAL):**
Use the black launch screen only?

**Screenshots (OPTIONAL):**
https://user-images.githubusercontent.com/4752473/139141693-fe007b56-3d7c-4f97-b408-5ab3a245a8b9.MP4

What If we add the backgroundColor with clear_color value to the [ViewController.m](https://github.com/defold/defold/blob/3e5fda5b774763bf1390b32ad7246024a8a3dc25/engine/glfw/lib/ios/app/ViewController.m) view? This will indirectly fix the situation if this blink is the black color of the UIWindow, but I don't now exactly what is it.

nog1potje:
When launching an iOS game made in defold, a brief black screen could be observed in the transition from the launch screen to the game. This fix covers up that black screen with the contents of the launch screen.

Fixes #6142

- WARNING: as of now this fix is incomplete, as it only works if the user uses a file called LaunchScreen. Please help provide guidance on how to access user set variables.

### Technical changes
- Load the launch screen storyboard and put it in front of the main view as soon as didFinishLaunchingWithOptions is called by the OS
- Remove this cover when the engine is initialized


external/glfw/lib/ios/app/AppDelegate.m
```diff
@@ -182,6 +194,16 @@ - (void)appUpdate
     {
         ShutdownEngine(true);
     }
+
+    // Cleanup the placeholder laucnh screen view once the engine is initialized
```
britzl, line 198: laucnh -> launch

nog1potje:
Does this class have access to dmConfigFile? There's no other usage within the glfw library that I could use as an example. Also, how do I compile the glfw library to test my code? The makefile does not seem to contain a configuration for iOS. I've sent a friend request through discord if that's ok.

britzl:
> Does this class have access to dmConfigFile? There's no other usage within the glfw library that I could use as an example. Also, how do I compile the glfw library to test my code? The makefile does not seem to contain a configuration for iOS. I've sent a friend request through discord if that's ok.

Ah, I did not notice that t he fix was in the glfw layer. Ok, we need some additional input then. @Jhonnyg can you assist?

external/glfw/lib/ios/app/AppDelegate.m
```diff
@@ -84,6 +84,18 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
     window = [[UIWindow alloc] initWithFrame:bounds];
     window.rootViewController = [[[ViewController alloc] init] autorelease];
+
+    // Load the LaunchScreen storyboard
+    UIStoryboard *storyboard = [UIStoryboard storyboardWithName:@"LaunchScreen" bundle:nil];
```
JCash, line 89: Agreed, we need to look into the plist properties after the `launch-screen` property before accepting this PR.

JCash (changes requested):
Overall good, but needs a solution for getting the launch screen name

nog1potje:
I figured out how to access the launch screen through iOS functionality, no need to access the defold settings. Commit incoming

nog1potje:
Done, the launch screen is now retrieved from the info.plist dictionary, and will be displayed if found. This should ensure it's always the same visuals as those displayed during load.

britzl:
The glfw packages are precompiled using the `build_external ` build command and then copied to `packages`. This PR should preferably also include the pre-built glfw libraries.

https://github.com/defold/defold/tree/dev/external#external

nog1potje:
Sounds good, I'll get on it

britzl:
> Sounds good, I'll get on it

Cool, let us know if you need help with the setup.

nog1potje:
glfw successfully compiled and I've committed the file, but my local editor instance won't bundle iOS apps, not sure why. I'm getting com.defold.extender.ExtenderException: The given sdk does not exist: 393881b328f0763b0ba1197764a322a6771869de However, this happens for me on the base dev branch as well. Can someone help test it?

JCash:
I think the issue is that the editor tries to build with an archived release, but since you're actually using the local build, it _won't_ be an archived release.

What are your build steps for the editor?
It should be like:

First, for the engine:
* build iOS
* build Bob

And then

```
cd editor
lein init
lein run
```

JCash (dismissed):

nog1potje:
Thanks for the approval! I haven't been able to put any time in this week unfortunately. Hopefully next week. Also, why the conflict? Are there any other changes to the ios glfw library?

britzl:
@nog1potje the comment in the initial PR about _"WARNING: as of now this fix is incomplete, as it only works if the user uses a file called LaunchScreen. Please help provide guidance on how to access user set variables."_ is no longer accurate right?

nog1potje:
Yes, that warning can now be ignored

britzl (approved):

PR #9002, opened 2024-05-29
Multi-touch improvements for Android devices

### LINKED ISSUE #7216: Multi touch input getting lost on Android
This bug seems to have been flagged as fixed several times, however, I am experiencing touch up and touch down event losses on android.

I have two gui boxes which are acting as left and right "paddles". When the user presses the left paddle at the same time as releasing the right paddle, the touch up and touch down events are lost. This happens vise versa also.

As far as I can tell, it happens when the actions are executed on the same frame.

I can't reproduce the problem using arrow keys, it only happens with multi touch.

I am still fairly new to Defold, so I'm sorry if I've missed anything glaringly obvious.

I would love to continue using Defold for my project, but am afraid this will be a show stopper if there is no workaround/fix.

I'm testing by bundling the project and installing on a Samsung A53 5G.

I'm also using Monarch with screen proxies, but can't see that would be a problem.

Defold Version 1.4.0

Many thanks in advance

Jhonnyg:
Fixed an issue where multi-touch on Android caused some events to not be processed, leading to "ghost" input events that gets dispatched to the `on_input` function forever.

Fixes #7216


external/glfw/lib/android/android_init.c
```diff
@@ -837,16 +868,28 @@ int32_t glfwAndroidHandleInput(struct android_app* app, AInputEvent* event)
...
+        {
+            g_MaxAppInputEvents = size_increase;
+            g_AppInputEvents = malloc(sizeof(struct InputEvent) * g_MaxAppInputEvents);
+            memset(g_AppInputEvents, 0, sizeof(struct InputEvent) * g_MaxAppInputEvents);
+        }
+        else
+        {
+            g_MaxAppInputEvents += size_increase;
+            g_AppInputEvents = realloc(g_AppInputEvents, sizeof(struct InputEvent) * g_MaxAppInputEvents);
+        }
```
JCash, line 887: No need for the check.
You can use `realloc`with a null pointer, so only the else statement is really needed.

Also, the first case zeroes out memory, and your second case doesn't.
I personally would skip the memset in favor of less code, as we're not likely to access out-of-bounds anyways, and if we did it would fail anyways.

JCash (changes requested):
lgtm.
I'd prefer to use realloc all the time though.

JCash (dismissed):

PR #9021, opened 2024-06-02
Add option to build script that generates compile_commands.json file(s)

### LINKED ISSUE #9020: Make it possible for build scripts to generate compile_commands.json
When developing the engine part of Defold using Sublime Text, it can be a bit annoying to not have good auto complete/"intellisense-ish". The "[EasyClangComplete](https://niosus.github.io/EasyClangComplete/)" plugin adds some support for this, but the developer would need to setup include paths and build options themselves, OR use a "[compile_commands.json](https://clang.llvm.org/docs/JSONCompilationDatabase.html)" file generated by the build system - the Defold build system currently doesn't output such file.

The build system for Defold should be able to output a standard `compile_commands.json` that can be read by code editors/tools.

TBH you could use some fancier dev environment than Sublime Text, I guess, but that doesn't sound fun?

andsve:
Adds a new `--generate-compile-commands` option to `build.py` that generates `compile_commands.json` files per engine lib/sub project and concatenates them into one "global" json that is written to the root.

Fixes #9020 

Before (using ECC, but without compile_commands.json):
image
image

After:
image
image


Jhonnyg (approved):
Perfect! A future improvement would be if we could add the generation into the submodules script (or a separate script altogether) that does the merge step, so that we can utilize the iterativeness of the submodules approach. But otherwise, looks good!

PR #9023, opened 2024-06-03
Dmsdk graphics glfw wrapper fixes

Jhonnyg:
Fixes an issue with missing symbols in DMSDK due to the recent glfw3 upgrade.

### DIFF
--- a/engine/graphics/src/graphics.cpp
+++ b/engine/graphics/src/graphics.cpp
@@ -35,11 +35,16 @@ namespace dmGraphics
     static GraphicsAdapter*             g_adapter = 0;
     static GraphicsAdapterFunctionTable g_functions;
 
-    void RegisterGraphicsAdapter(GraphicsAdapter* adapter, GraphicsAdapterIsSupportedCb is_supported_cb, GraphicsAdapterRegisterFunctionsCb register_functions_cb, int8_t priority)
+    void RegisterGraphicsAdapter(GraphicsAdapter* adapter,
+        GraphicsAdapterIsSupportedCb              is_supported_cb,
+        GraphicsAdapterRegisterFunctionsCb        register_functions_cb,
+        GraphicsAdapterGetContextCb               get_context_cb,
+        int8_t                                    priority)
     {
         adapter->m_Next          = g_adapter_list;
         adapter->m_IsSupportedCb = is_supported_cb;
         adapter->m_RegisterCb    = register_functions_cb;
+        adapter->m_GetContextCb  = get_context_cb;
         adapter->m_Priority      = priority;
         g_adapter_list           = adapter;
     }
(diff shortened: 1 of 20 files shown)


engine/graphics/src/graphics.cpp
```diff
@@ -271,6 +276,11 @@ namespace dmGraphics
         return g_functions.m_NewContext(params);
     }
 
+    HContext GetInstalledContext()
+    {
+        assert(g_adapter && "No graphics adapter installed");
+        return g_adapter->m_GetContextCb();
+    }
```
Jhonnyg, line 283: This is currently only available internally (via graphics_private.h) to be able to get the current window handle. I decided to put a callback on the adapter pointer to avoid having to add a new function signature on the adapter function table, but with the benefit that each adapter will get compile errors if no callback is passed in.

engine/graphics/src/graphics_glfw_wrappers.cpp
```diff
@@ -12,25 +12,71 @@
...
-    GLXContext GetNativeX11GLXContext()     { return glfwGetX11GLXContext(); }
+#if defined(DM_PLATFORM_IOS)
+    id GetNativeiOSUIWindow()               { return dmPlatform::GetiOSUIWindow(); }
+    id GetNativeiOSUIView()                 { return dmPlatform::GetiOSUIView(); }
+    id GetNativeiOSEAGLContext()            { return dmPlatform::GetiOSEAGLContext(); }
+#else
+    id GetNativeiOSUIWindow()               { return 0; }
+    id GetNativeiOSUIView()                 { return 0; }
+    id GetNativeiOSEAGLContext()            { return 0; }
+#endif
```
Jhonnyg, line 41: This setup avoids having to implement all platform functions for all platforms

engine/platform/src/platform_window_osx.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include "platform_window.h"
+
+#include <objc/objc.h>
+
+namespace dmPlatform
+{
+    id GetOSXNSWindow(HWindow window);
+    id GetOSXNSView(HWindow window);
+    id GetOSXNSOpenGLContext(HWindow window);
```
Jhonnyg, line 26: Due to GLFW3, these functions require a window handle

engine/graphics/src/graphics_glfw_wrappers.cpp
```diff
@@ -12,25 +12,71 @@
...
+    #include <platform/platform_window_ios.h>
+#elif defined(ANDROID)
+    #include <platform/platform_window_android.h>
+#elif defined(__MACH__)
+    #include <platform/platform_window_osx.h>
+#elif defined(_MSC_VER)
+    #include <platform/platform_window_win32.h>
+#elif defined(__linux__) && !defined(ANDROID)
+    #include <platform/platform_window_linux.h>
+#endif
```
Jhonnyg, line 29: Do we expose DM_PLATFORM_ defines for all platforms? I couldn't get it to work properly on os 🤔

JCash: According to waf_dynamo, no. There we only do ios/macos.
On the extender server though (build_input.yml) we do.

JCash (dismissed):
A bit unclear what the "fix" was.

PR #9026, opened 2024-06-04
Ensure build targets from embedded objects inside editable resources fuse with equivalents from non-editable resources

### LINKED ISSUE #9025: Equivalent embedded data in editable and non-editable resource cause exception on build
If an embedded game object or component inside a game object or collection located below a directory marked non-editable matches an equivalent embedded object in a regular game object or collection, building the project throws an exception.

**To Reproduce (REQUIRED)**
1. Create a game object file and add an embedded component to it.
2. Make a copy of the game object file below a directory marked non-editable in `game.project`.
3. Reference both game objects from the main collection.
4. Build the project.

The project builds successfully. In the build output, the embedded component becomes a single binary referenced by both game objects.

1.8.1

matgis:
Fixed exception when building a project that contains equivalent embedded objects in both an editable and a non-editable resource.

Fixes #9025

### DIFF
--- a/editor/src/clj/editor/pipeline.clj
+++ b/editor/src/clj/editor/pipeline.clj
@@ -109,10 +109,23 @@
 
 (defn- make-dep-resources
   [deps build-targets-by-content-hash]
+  ;; Create a map that resolves an original BuildResource into a fused
+  ;; BuildResource. Build target fusion is based on the content-hash values of
+  ;; the build targets. In order to fuse build targets from both editable and
+  ;; non-editable BuildResources, we make sure to add their counterpart to the
+  ;; resulting map alongside the original BuildResource.
   (into {}
-        (map (fn [{:keys [content-hash resource] :as _build-target}]
-               (assert (bt/content-hash? content-hash))
-               [resource (:resource (get build-targets-by-content-hash content-hash))]))
+        (mapcat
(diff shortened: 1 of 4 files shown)


editor/src/clj/editor/pipeline.clj
```diff
@@ -109,10 +109,23 @@
...
+          (fn [{:keys [content-hash] :as build-target}]
+            (assert (bt/content-hash? content-hash))
+            (let [original-build-resource (:resource build-target)
+                  counterpart-build-resource (workspace/counterpart-build-resource original-build-resource)
+                  fused-build-target (get build-targets-by-content-hash content-hash)
+                  fused-build-resource (:resource fused-build-target)]
+              (cond-> [(pair original-build-resource fused-build-resource)]
+
+                      counterpart-build-resource
+                      (conj (pair counterpart-build-resource fused-build-resource))))))
```
matgis, line 128: This was the source of the exception. The build targets from both embedded resources will have the same `content-hash`, but in `dep-resources`, only the `:resource` of whichever of them got put in `build-targets-by-content-hash` last would resolve to the matching build target.

vlaaad (approved):
LGTM! I also like that our new code is, in general, better documented 👍

PR #9031, opened 2024-06-04
Add child-seqs to outline nodes for copy+paste handling

### LINKED ISSUE #6682: Allow copy/pasting layers between guis
It's common to have the same layers structure in different guis, but there is no way to copy/paste layers from other guis to the new one. Currently, for every new GUI file we have to manually create layers and name them one by one, which slows down a developer.

Allow copying layers between guis.

It's possible to have a template GUI file with common layers/textures/fonts etc. that is copy-pasted when we want to create a new GUI. 

Jhonnyg:
The editor now has support for copy/paste/cut of all resources nodes in the outline view except for "layouts". Layouts are unique and only one of each type can exist in a GUI at any time.

Fixes #6682

### DIFF
--- a/editor/src/clj/editor/gui.clj
+++ b/editor/src/clj/editor/gui.clj
@@ -425,7 +425,8 @@
         (g/set-property source :child-index next-index)
         (attach-gui-node node-tree target source type)))))
 
-(defn- gen-outline-node-tx-attach-fn
+;; SDK api
+(defn gen-outline-node-tx-attach-fn
   ([attach-fn]
    (gen-outline-node-tx-attach-fn attach-fn :names))
   ([attach-fn target-name-key]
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/gui.clj
```diff
@@ -425,6 +425,17 @@
         (g/set-property source :child-index next-index)
         (attach-gui-node node-tree target source type)))))
 
+(defn- gen-outline-node-tx-attach-fn
```
matgis, line 428: Should we make this public and mark it `;; SDK api` so we can also make the corresponding changes to the `SpineScenesNode` in `extension-spine`?

https://github.com/defold/extension-spine/blob/main/defold-spine/editor/src/spineguiext.clj#L309

matgis (approved):
Nice job! Works great! ❤️

editor/src/clj/editor/gui.clj
Jhonnyg: Good idea! I did that 👍

PR #9057, opened 2024-06-10
Add attachment texture resource path into resource.get_render_target_info

Jhonnyg:
Added a new `texture` field in the data returned from `resource.get_render_target_info`:
```lua
local rt_info = resource.get_render_target_info("/my_rt.render_targetc")
local t_info = resource.get_texture_info(rt_info.attachments[1].texture)
```

This can then be be used to change the backing texture of an atlas, which should make it easier to achieve custom rendering for any type that is using atlases:

```lua
go.property("rt_atlas", resource.atlas())
go.property("rt", resource.render_target())

local function render_target_to_atlas(self)
	-- get the render target info. This contains all the attachment texture resources
	local rt_info = resource.get_render_target_info(self.rt)
	-- get the atlas info, we can create this from scratch, but it's easier
	-- to use the data that is already there and then just modify it
	local atlas_info = resource.get_atlas(self.rt_atlas)
	-- update the atlas backing texture to use our render target attachment
	atlas_info.texture = rt_info.attachments[1].texture
	-- set / update the atlas with the new atlas info
	resource.set_atlas(self.rt_atlas, atlas_info)
end
```

Scripts can now also pick up render targets as script properties as well:
`go.property("my_rt", resource.render_target())`


editor/src/clj/editor/code/script_compilation.clj
```diff
@@ -68,12 +68,13 @@
...
-   "buffer"      (constantly "buffer")
-   "texture"     (constantly (conj image/exts "cubemap" "render_target"))
-   "tile_source" (constantly "tilesource")})
+  {"atlas"        #(workspace/resource-kind-extensions % :atlas)
+   "font"          (constantly "font")
+   "material"      (constantly "material")
+   "buffer"        (constantly "buffer")
+   "texture"       (constantly (conj image/exts "cubemap" "render_target"))
+   "tile_source"   (constantly "tilesource")
+   "render_target" (constantly "render_target")})
```
Jhonnyg, line 77: it's already available in the "texture", but since BOB allows this functionality out of the gate, the editor should to. Also, it serves different use cases

vlaaad (dismissed):
Editor changes look good!

AGulev:
is it still for 1.9.0?

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1783,6 +1826,21 @@ static int GetRenderTargetInfo(lua_State* L)
             lua_pushinteger(L, color_buffer_flags[i]);
             lua_setfield(L, -2, "buffer_type");
```
JCash, line 1827: nitpick: for readability, perhaps use "lua_pushinteger(L, buffer_type);" for readability/consistency

JCash (changes requested):
LGTM
Seems like a memory overwrite when recreating the resource though.
I'm not sure we have extensive recreate-tests for resources, but we should strive for that in the future at least.

PR #9066, opened 2024-06-12
Allow pasting objects into itself

### LINKED ISSUE #9065: Object tries to insert into itself at cmd+c, cmd+v
Previously you could just do cmd+c, cmd+v in the Asset Browser, but now you have to select a folder, otherwise it says “Cannot paste folder ‘’ into itself”. That is, if you want to copy a script, it copies it into itself as a folder. 

1. Select the object in the Asset Browser
2. Copy and paste it using hotkeys
3. You will get an error that you cannot paste a folder into itself.

  - Version 1.8.1

AGulev:
We process pasting into itself later in this flow by renaming a new file/folder. So here I shouldn't check this case at all

Fix https://github.com/defold/defold/issues/9065

### DIFF
--- a/editor/src/clj/editor/asset_browser.clj
+++ b/editor/src/clj/editor/asset_browser.clj
@@ -345,11 +345,11 @@
        (let [tree-view (g/node-value asset-browser :tree-view)
              resource (first selection)
              src-files (.getFiles (Clipboard/getSystemClipboard))
-             dest-path (resource/abs-path resource)]
-         (if-let [conflicting-file (some #(when (and (.startsWith (.toPath (io/file dest-path))
-                                                                  (.toPath ^File %))
-                                                     (not= dest-path (.getPath ^File %)))
-                                            %)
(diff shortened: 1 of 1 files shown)


vlaaad (changes requested):
I found an issue with copying files. If I select a file, press <kbd>Cmd+C</kbd>, then select another file in the same folder that has the same prefix, then press <kbd>Cmd+V</kbd>, I get a notification that I can't paste a folder (it's not a folder!) into itself. 

A completely realistic example of these files are `foo.gui` and `foo.gui_script`. I can't create a copy `foo.gui` when I'm pasting it onto `foo.gui_script`.

I think we should add a `(.isDirectory ^File %)` check when searching for a conflicting file.

AGulev:
@vlaaad yes! Totally makes sense! Thank you. I've fixed that.

vlaaad (approved):
Shipit!

vlaaad:
BTW, you might want to cherry-pick the change to `edtor-dev` :)

PR #9073, opened 2024-06-13
Fix broken input

Jhonnyg:


engine/hid/src/hid.cpp
```diff
@@ -496,9 +496,11 @@ namespace dmHID
                 dmPlatform::PLATFORM_KEY_BACK,
             };
 
-            int key_index = key_index - HID_SPECIAL_START;
+            key_index = key_index - HID_SPECIAL_START;
```
Jhonnyg, line 499: This is/was the issue: the "key_index" symbol already existed, meaning key_index would be undefined in this case.

AGulev: ouch!

AGulev (approved):
🔥

PR #9081, opened 2024-06-14
Overhauled editor save system

### LINKED ISSUE #6410: Do not put default protobuf values in the resource files.

For now all the game resources contains both default and custom values in resources (collection, go etc)
It's not necessary to have it in files because in protobuf default values specified in `proto` files.

That would make format clearer and resources smaller in repo. Also it will simplify using of external editors that create the output files in Defold proto format.

### LINKED ISSUE #7175: Improve performance of first save in a large project
**Task (REQUIRED):**
Saving for the first time in a large project take a long time. We need to profile and investigate what's going on and see if we can improve the speed somehow.

Saving for the first time should be faster.

matgis:
Overhauled the systems that are responsible for loading and saving project files in the editor.

Default values and non-overridden values originating from referenced resources such as Gui template scenes will no longer be written to the project files.

The updated sparse file formats will be used for any files that are saved from now on. If you want to upgrade all the project files to the new sparse file formats at once, select **Upgrade File Formats...** from the **File** menu.

Fixes #6410
Fixes #7175

If you've built an extension that includes experimental Clojure-based editor plugins (maybe based on the the [extension-simpledata](https://github.com/defold/extension-simpledata) project on GitHub?), you can port it to the new save system by making changes corresponding to [our own changes](https://github.com/defold/extension-simpledata/pull/12/commits/8c0b529419d922b44e79c2145860b342d5f8c05c) to `extension-simpledata`. In summary:

* You should use `editor.graph-util/set-properties-from-pb-map` in the `:load-fn` of your Protobuf-based resource types to make sure we only return transaction steps that set properties that are present in the `data` we read from disk.
* Every `property` in your Protobuf-based `defnodes` should include a `default` value from the specific Protobuf field it is backed by to ensure the node works correctly when we don't have a value for the field in the `data` we read from disk.
* You should use `protobuf/make-map-without-defaults` when producing the `save-value` output in your Protobuf-based `defnodes` to ensure we don't write defaults to disk when saving.

#### Save values and detecting unsaved changes

After this change, the process of loading the project is split up into multiple phases. Please refer to the updated [Resource Sync](https://github.com/defold/defold/blob/issue-6410-editor-save-value/editor/README_SYSTEMS_OVERVIEW.md#resource-sync) and [Save Data](https://github.com/defold/defold/blob/issue-6410-editor-save-value/editor/README_SYSTEMS_OVERVIEW.md#save-data) sections of the [Systems Overview](https://github.com/defold/defold/blob/issue-6410-editor-save-value/editor/README_SYSTEMS_OVERVIEW.md) for details.

Previously, the `save-data` output for every `ResourceNode` contained a `:content` field, which was a string. This string would be compared against the string resulting from reading the current file contents from disk, applying the `:read-fn` to it to obtain a Clojure map, them applying the `:write-fn` to the Clojure map to get a string in the latest file format, and comparing that string to our `:content` string to figure out if we were `:dirty` and should be written to disk. Even with a lot of caching, this was obviously quite inefficient. The editor would have to do a lot of work to figure out if a file was in need of saving or not. This also meant the editor could mistakenly overwrite externally modified files with old data unless we made sure to incorporate the external changes into the graph before saving.

In the new system, we've split up the load process into several phases, and are able to compare the sanitized `source-value` returned by the `:read-fn` at the time a resource was loaded directly against the value produced by the `save-value` output. Doing it this way also saves us from having to scan the file system and incorporate external changes into the graph every time we save, because we're no longer comparing against the on-disk state we might not have seen.

Splitting things up also means we can apply the `:dependencies-fn` to the sanitized `source-value` returned by the `:read-fn` instead of reading the data from disk twice.

We can also do something a bit sneaky, but hugely beneficial to initial save times. Given that the `source-value` returned by the `:read-fn` must exactly match the value returned by the `save-value` output unless the user has made changes (enforced by [no-unsaved-changes-after-load-test](https://github.com/defold/defold/blob/issue-6410-editor-save-value/editor/test/integration/save_data_test.clj#L1323)), we can sneak the loaded `save-value` and its corresponding non-dirty `save-data` into the system cache after loading, so these outputs are immediately cached for all save-tracked project files. All the save-related outputs will remain in the cache until they are explicitly cleared out or invalidated, since they are retained outputs. For file formats that require immediate migration, we provide `g/flag-nodes-as-migrated!` to stop save-related outputs for specific nodes from being sneaked into the cache this way.

In addition, we no longer keep embedded resource data as strings in the graph. Embedded game objects and components are now loaded and kept as Clojure maps. There's a `:string-encode-fn` option for Protobuf-based `resource-types` that allows us to encode the data into strings when the `:write-fn` is applied.

While having the save state be represented by the actual text of the output file was inefficient, it made comparing the output against the input simple. Now that we're comparing structured Clojure maps instead of output file format strings, we need to be strict about maintaining data equality.

In particular, the editor would earlier play a bit fast and loose with regard to numeric types. Numbers in Protobuf-based resources would match their declared field types when read, but become replaced with `doubles` when edited. Since the `float` value we read from a string like `"0.1"` in the file does not equal the `double` value parsed from the text `"0.1"` in the editor field, we would see erroneous unsaved changes after the user restored it its original value. To fight this, all editor widgets and transform tools have been updated to respect and preserve the original value types.

Since we no longer have a string representation of the file contents conveniently available for the in-memory state of files in the project, we needed a new solution for project-wide text search. Each `resource-type` now has a custom search implementation that is tailor-made to the specifics of the data format. For example, code resources search their vector of string lines, whereas Protobuf-based resources search their Clojure map representation.

When registering a `resource-type`, you specify a `:search-fn` and a `:search-value-fn`. The `:search-value-fn` is used to produce a `search-value` from a `ResourceNode`, and defaults to returning the `save-value` output. The `:search-fn` is a multi-arity function used to search the resource for a matching substring. The first arity should accept a search string and return a value that can be used by the second arity to perform the search (typically a regex `Pattern`). The second arity should accept a `search-value` returned from the `:search-value-fn` and the value returned by the first arity, and return a sequence of `match-infos`. Each `match-info` should have a `:match-type` and details about the match unique to that `:match-type`. This can then be used to select the matching sub-region when the match is opened in an editor view.

#### Sparse file formats

As stated in the user-facing changes, `.gui` scenes files that bring in other `.gui` files as `TemplateNodes` will no longer include the non-overridden property values from the referenced `.gui` scenes. However, because the `id` and `parent` fields are overridden, the referencing `.gui` file will still contain node hierarchies that mirror the structure of the referenced `.gui` file. It's possible we could get rid of this requirement in the future. For now, at least setting properties in a referenced `.gui` file won't modify the referencing `.gui` files anymore.

By default, we no longer include default values from `optional` fields when reading and writing Protobuf-based file formats. A notable exception to this is resource fields. Any field designated `[(resource) = true]` in a `.proto` file will be included in the project files, even though it is assigned the default value. We retain resource paths so that users can perform search and replace operations on paths in project files using an external text editor.

We also explicitly include some specific fields in cases where we've not yet decided how to handle the absence of a value (for example, the `scale_along_z` field in `.collection` files).

When a Protobuf-based file format is read, its `:read-fn` will strip out any `optional` field values that remain at their default. We then use `protobuf/make-map-without-defaults` in the corresponding `save-value` output to do the same when we produce the `save-value`.

You can specify `:read-defaults true` when registering a Protobuf `resource-type` to make the `:read-fn` include default values for all `optional` fields. If you chose to do this, you should pair it with `protobuf/make-map-with-defaults` in your `save-value` output.

All the Protobuf-based node types have been updated to support the absence of default values in the value returned by the `:read-fn`. This involves a combination of specifying defaults for node properties where necessary, and making use of the new `gu/set-properties-from-pb-map` helper to emit `g/set-property` `transaction-steps` for the fields that are present in the value returned by the `:read-fn`.

It is imperative that the default value specified for a node property corresponds to the default of the Protobuf field. The property will not be written to disk when it is set to the default value, so the `:load-fn` will never see it. Default property values for nodes that represent Protobuf file formats should be based on `(protobuf/default PbClass :pb-field)` for things to work as expected in the absence of read values.

#### Protobuf changes

Some changes have been made to the `.proto` files to strip out common values from the file formats. In some cases, fields that were previously `required` have been changes to `optional`. For example, every field used to be `required` in `.label` files, but it doesn't make sense for every `.label` file to specify values for specialized properties like `leading` and `tracking`. Changing a field from `required` to `optional` is safe in our case, since we do not guarantee newer revisions of our file formats to be compatible with old releases of Defold.

The Protobuf compiler supports default values for primitive fields, but not for message-type fields. In order to support a default value of `[1.0 1.0 1.0]` for scaling or `[0.0 0.0 0.0 1.0]` for opaque black color, several new `Vector` types that vary the field defaults have been introduced in [ddf_math.proto](https://github.com/defold/defold/blob/issue-6410-editor-save-value/engine/ddf/src/ddf/ddf_math.proto) and used in various places in the `.proto` files.

##### Somewhat controversial Protobuf choices

The `y` field of [SplinePoint](https://github.com/defold/defold/blob/issue-6410-editor-save-value/engine/particle/proto/particle/particle_ddf.proto#L35) was kept as `required` while all the other fields were changed to `optional`. The `y` value represents the value of a point on a particular `x` (normalized time) along the curve. The `SplinePoint` message is typically used in a repeated `points` field, which covers multiple curve points, but also the case where a single value is used for the duration of a curve (i.e. "not using a curve" in editor terms). When not using a curve, a value of `0.0` looks like this:
```
particle_properties {
  key: PARTICLE_KEY_RED
  points {
    y: 0.0
  }
}
```
If the `y` field was not required, the `points` would look strange for value `0.0`:
```
  points {
  }
```
It is a repeated field, so this is showing a list with a single `SplinePoint` that has defaults for every field.

You could make an argument for having the `x` field be `required` as well for symmetry, but then we'd have a lot of redundant `x: 0.0` lines in the `.particlefx` files. I'm not sure what is preferable, to be honest.

### Companion Pull Requests

The following PRs update our Editor extensions to use sparse file formats:

* https://github.com/defold/extension-rive/pull/102
* https://github.com/defold/extension-simpledata/pull/12
* https://github.com/defold/extension-spine/pull/178
* https://github.com/defold/extension-texturepacker/pull/9


editor/src/clj/editor/build_target.clj
```diff
@@ -24,24 +24,22 @@
...
+     :build-fn (:build-fn build-target)
+     :user-data (:user-data build-target)}))
 
 (defn content-hash
-  ^String [build-target]
+  ^String [build-target opts]
   (let [content-hash-components (content-hash-components build-target)]
     (try
-      (digestable/sha1-hash content-hash-components)
+      (digestable/sha1-hash content-hash-components opts)
```
matgis, line 42: The `opts` are piped through the whole digest process. We supply a `:node-id->persistent-value` map and use it to resolve any resource `node-ids` we encounter in the data to their corresponding `proj-paths`.

editor/src/clj/editor/cljfx_form_view.clj
```diff
@@ -608,10 +638,10 @@
                              :fit-size 16}]}]}))
 
 (defmethod handle-event :add-list-items [{:keys [value on-value-changed fx/event]}]
-  {:dispatch (assoc on-value-changed :fx/event (into value event))})
+  {:dispatch (assoc on-value-changed :fx/event (coll/into-vector value event))})
```
matgis, line 641: `(into nil ...)` results in a list. We want a vector if `value` is `nil`.

editor/src/clj/editor/cljfx_form_view.clj
```diff
@@ -93,32 +94,57 @@
...
+      (or (field-expression/to-long v)
           (throw (RuntimeException.))))))
 
+(defn- number-string-converter [value]
+  (condp instance? value
+    Float float-converter
+    Double double-converter
+    Integer int-converter
+    Long long-converter))
+
```
matgis, line 147: As you can see, the changes to this file are mainly related to preserving data types when editing.

editor/src/clj/editor/cljfx_form_view.clj
```diff
@@ -1546,17 +1577,16 @@
                          :parent parent
                          :resource-string-converter resource-string-converter}))))))
 
+(defn- make-form-view-node [graph parent resource-node workspace project]
+  (g/make-nodes graph [view CljfxFormView]
+    (g/set-property view :renderer (create-renderer view parent workspace project))
+    (g/connect resource-node :form-data view :form-data)))
+
+(def make-form-view-node! (comp first g/tx-nodes-added g/transact make-form-view-node))
+
```
matgis, line 1586: These changes are covered by `integration.form-type-preservation-test/ensure-float-form-fields-preserve-type!`.

editor/src/clj/editor/collection.clj
```diff
@@ -179,9 +178,9 @@
 
 (g/defnk produce-go-build-targets [_node-id build-error build-resource ddf-message resource-property-build-targets source-build-targets pose]
   ;; Create a build-target for the referenced or embedded game object. Also tag
-  ;; on :instance-data with the overrides for this instance. This will later be
-  ;; extracted and compiled into the Collection - the overrides do not end up in
-  ;; the resulting game object binary.
+  ;; on :game-object-instance-data with the overrides for this instance. This
+  ;; will later be extracted and compiled into the Collection - the overrides do
+  ;; not end up in the resulting game object binary.
```
matgis, line 183: The key `:instance-data` was used for both game object instances and component instances, which was very confusing.

editor/src/clj/editor/collection.clj
```diff
@@ -374,23 +372,41 @@
...
+    :name name
+    :scale-along-z (protobuf/boolean->int scale-along-z)
+    :instances ref-inst-ddf
+    :embedded-instances embed-inst-ddf
+    :collection-instances ref-coll-ddf))
+
+(defn instance-property-desc-save-value [instance-property-desc]
+  ;; GameObject$InstancePropertyDesc in map format.
+  (-> instance-property-desc
+      (protobuf/sanitize-repeated :properties game-object/component-property-desc-save-value)))
```
matgis, line 385: * `protobuf/sanitize` works like `update`, but will exclude the key altogether if the `update-fn` returns `nil`.
* `protobuf/sanitize-repeated` works like `update (partial mapv update-fn)`, but will exclude the key altogether if the `update-fn` returns an empty sequence.

Elsewhere, you'll also see `protobuf/assign` and `protobuf/assign-repeated`, which are their `assoc` counterparts that will strip empty sequences or `nil` values in a similar fashion.

editor/src/clj/editor/defold_project.clj
```diff
@@ -936,12 +1184,12 @@
   ([endpoint]
    (case (g/endpoint-label endpoint)
      (:build-targets) (project-resource-node? (g/now) (g/endpoint-node-id endpoint))
-     (:save-data :source-value) (project-file-resource-node? (g/now) (g/endpoint-node-id endpoint))
+     (:save-data :save-value) (project-file-resource-node? (g/now) (g/endpoint-node-id endpoint))
```
matgis, line 1187: `:source-value` is now trivial to produce, since we can get it from `user-data` in the project graph. This frees us up to cache `:save-value` instead. We're not really using more memory since it is also referenced from the `save-data`, but it is commonly used in connections whereas `save-data` is only really used for dirty-checks in the save system.

editor/src/clj/editor/defold_project.clj
```diff
@@ -979,34 +1227,11 @@
...
-(defn- cache-save-data! [project]
-  ;; Save data is required for the Search in Files feature, so we pull
-  ;; it in the background here to cache it.
-  (let [evaluation-context (g/make-evaluation-context)]
-    (future
-      (error-reporting/catch-all!
-        ;; TODO: Progress reporting.
-        (g/node-value project :save-data evaluation-context)
-        (ui/run-later
-          (update-system-cache-save-data! evaluation-context))))))
```
matgis, line 1009: No need for this anymore! We have the `save-value` outputs needed for search cached and ready immediately after loading.

editor/src/clj/editor/gui.clj
```diff
@@ -2336,23 +2399,32 @@
...
-   :max-nodes max-nodes
-   :nodes node-msgs
-   :layers layer-msgs
-   :fonts font-msgs
-   :textures texture-msgs
-   :materials material-msgs
-   :layouts layout-msgs
-   :particlefxs particlefx-resource-msgs
-   :resources resource-msgs})
-
```
matgis, line 2353: Inlined just below this.

editor/src/clj/editor/gui.clj
```diff
@@ -2806,47 +2905,59 @@
...
 
 ;; SDK api
 (defn register-gui-scene-loader! [load-fn]
   (let [load-fn-sym (fn/declared-symbol load-fn)]
     (swap! custom-gui-scene-loaders assoc load-fn-sym load-fn)))
 
-;; Used by tests
-(defn clear-custom-gui-scene-loaders-for-tests! []
-  (reset! custom-gui-scene-loaders (sorted-map)))
-
```
matgis, line 2842: Moved to bottom of file. Now also clears `custom-node-type-infos`, encountered later.

editor/src/clj/editor/handler.clj
```diff
@@ -380,7 +380,7 @@
 (defn adapt [selection t]
   (if (empty? selection)
     selection
-    (let [selection (if (g/isa-node-type? t)
+    (let [selection (if (g/node-type? t)
```
matgis, line 383: Renamed.

editor/test/integration/particlefx_test.clj
```diff
@@ -59,9 +53,9 @@
           fetch-anim-fn (fn [index] (get emitter-sim-data index))
           transforms [(doto (Matrix4d.) (.setIdentity))]
           sim (plib/make-sim 16 256 prototype-msg transforms)
-          attribute-infos (into []
-                                (map graphics/attribute-key->default-attribute-info)
-                                [:position])
+          attribute-infos [(-> :position
+                               (graphics/attribute-key->default-attribute-info)
+                               (assoc :coordinate-space :coordinate-space-world))]
```
matgis, line 58: The `:coordinate-space` of the `:position` attribute is `:default` in `attribute-key->default-attribute-info`. This is replaced by either `:coordinate-space-local` or `:coordinate-space-world` in `graphics/shader-bound-attributes` dependent on the `:vertex-space` of the used material. Here, we just make it up.

editor/test/integration/save_data_test.clj
```diff
@@ -1257,95 +1305,11 @@
                     project-path)
             non-template-overridden-gui-node-field-paths)))))
 
-(defn- clear-cached-save-data! []
-  ;; Ensure any cache entries introduced by loading the project aren't covering
-  ;; up an actual dirty-check issue.
-  (g/clear-system-cache!))
```
matgis, line 1263: This stuff was moved into the `test-util` module.

vlaaad (changes requested):
Magnificent, monumental work! This is an amazing addition to the editor! Also, very impressive test suite! And all the explanations helped a lot with understanding the changes. Wow!

Before merging this, I'd like us to fix an issue with Search In Files first: when I search for a number, e.g. `1`, I get an exception:
```
clojure.lang.RT.longCast.class java.lang.NullPointerException: Cannot invoke "java.lang.Number.doubleValue()" because "x" is null
    at clojure.lang.RT.longCast(RT.java:1280)
    at editor.search_results_view$make_matched_item_row_indicator.invokeStatic(search_results_view.clj:127)
    at editor.search_results_view$make_matched_item_row_indicator.invoke(search_results_view.clj:127)
    at editor.search_results_view$make_matched_item_graphic$fn.invoke(search_results_view.clj:155)
    ...
```
I have a couple more notes about this PR. I'd prefer you to **NOT** address any of them in this PR. If you think some of the notes would be worth implementing, I'd prefer you do it **after** merging this. Anyway, here are my notes:
1. Does it make sense to create a `(g/defprotobufnode ...)` macro that ensures reasonable default handling (i.e. provides defaults, ensures no default for required field)?
2. There are changes in cljfx-form-view to preserve and ensure correct number types. Do we need something similar for editor scripts? (see https://github.com/defold/defold/blob/dev/editor/src/clj/editor/editor_extensions/graph.clj#L64-L68) — the converters only check for `number?` and `int?`. In luaj, it's impossible to have `1.0` LuaDouble, it's going to be `1` LuaInteger.
3. It's not clear that `test-util/ensure-number-type-preserving!` is actually a test assertion, because I'm trained to recognize test assertions as `(is ...)` forms. Do you think it would be better if the tests using this function were instead like `(is (test-util/number-type-preserving? ...))`?
4. Should we hide the `Upgrade File Formats...` command when we don't need to upgrade them?

engine/ddf/src/ddf/ddf_math.proto
```diff
@@ -40,6 +49,26 @@ message Vector4
...
+
+message Vector4WOne
+{
+    option (alias) = "dmVMath::Vector4";
+
+    optional float x = 1 [default = 0.0];
+    optional float y = 2 [default = 0.0];
+    optional float z = 3 [default = 0.0];
+    optional float w = 4 [default = 1.0];
+}
```
JCash, line 70: 👍

engine/gamesys/src/gamesys/resources/res_gui.cpp
```diff
@@ -195,13 +195,52 @@ namespace dmGameSystem
...
         dmDDF::Result e = dmDDF::LoadMessage<dmGuiDDF::SceneDesc>(params->m_Buffer, params->m_BufferSize, &scene_desc);
         if ( e != dmDDF::RESULT_OK )
             return dmResource::RESULT_FORMAT_ERROR;
 
+        // HACK: Gui nodes that have zero area are culled, but we can't know the
+        // size of nodes that have SIZE_MODE_AUTO at build time. Rather than
+        // having both Bob and the editor write bogus size values into the
+        // binaries, we patch it here. We can remove this hack once we address
+        // the culling issue properly.
+        EnsureAutoSizedNodesInSceneHaveNonZeroArea(scene_desc);
```
JCash, line 242: 👍

JCash (dismissed):
lgtm

matgis:
@vlaaad I've fixed the Search in Files issue! Nice find!

Regarding your other suggestions:

> 1. Does it make sense to create a `(g/defprotobufnode ...)` macro that ensures reasonable default handling (i.e. provides defaults, ensures no default for required field)?

Maybe! I considered it, but decided against it because we do a lot of renaming and transformation of the property defaults.

> 2. There are changes in cljfx-form-view to preserve and ensure correct number types. Do we need something similar for editor scripts? (see https://github.com/defold/defold/blob/dev/editor/src/clj/editor/editor_extensions/graph.clj#L64-L68) — the converters only check for `number?` and `int?`. In luaj, it's impossible to have `1.0` LuaDouble, it's going to be `1` LuaInteger.

Hmm, yes. We probably need to cover that here as well. I've created an issue for it: https://github.com/defold/defold/issues/9125

> 3. It's not clear that `test-util/ensure-number-type-preserving!` is actually a test assertion, because I'm trained to recognize test assertions as `(is ...)` forms. Do you think it would be better if the tests using this function were instead like `(is (test-util/number-type-preserving? ...))`?

It's like that to get as much information as possible from the `(is ...)` clauses. If you wrap it in an outer `(is ...)` clause, the test runner won't pick up the detailed checks within.

> 4. Should we hide the `Upgrade File Formats...` command when we don't need to upgrade them?

Since it is implemented as a forced save, it will become active immediately once you have any unsaved changes. I don't think it is worth the trouble, to be honest. It's nice to have the menu item available at all times for discoverability, and we explain thoroughly what it does when the user selects it.

vlaaad (approved):
Works great! SHIP IIIIIIIIITTTTT!!!!!11

vlaaad:
> It's nice to have the menu item available at all times for discoverability, and we explain thoroughly what it does when the user selects it.

What I mean is, after we migrated to the new format, the menu item will stay in the same way, and will show the same warnings, but it will do nothing. Do you think it still makes sense to show it in this case?

PR #9088, opened 2024-06-17
Add some missing return types to gui docs

farism:
## PR checklist

* [x] Documentation

----------

- updates gui docs to add some missing return types
- `true` and `false` uniformly display as code blocks


britzl (approved):
Thank you!

PR #9094, opened 2024-06-17
Remove callback refs from sprite components

### LINKED ISSUE #8123: `play_flipbook` callback doesn't remove itself when it's done and it doesn't remove prev. callbacks
Related: https://github.com/defold/extension-spine/issues/147

**To Reproduce (REQUIRED)**
1. Run attached example
2. Pay attention in logs how scripts create callbacks
3. when animation is done pay attention that the callback were set in `main.script`  is still exist
4. wait a few seconds when subtest will be run with `play_flipbook` in a for loop
5. pay attention that only last callback removed

I didn't test It but it is possible the same issue exist for: sounds, models, http requests (something else?)

Old callbacks should be removed when new callback created

  - 1.6.0

**Platforms (REQUIRED):**
 - All

[sprites.zip](https://github.com/defold/defold/files/12846209/sprites.zip)

Jhonnyg:
Fixed an issue for sprite components where old callback references are not removed when issuing `sprite.play_flipbook` calls. 

Fixes #8123

### DIFF
--- a/engine/gameobject/proto/gameobject/gameobject_ddf.proto
+++ b/engine/gameobject/proto/gameobject/gameobject_ddf.proto
@@ -134,16 +134,16 @@ message CollectionDesc
  *
  * Post this message to a game object instance to make that instance acquire the user input focus.
  *
- * User input is distributed by the engine to every instance that has 
+ * User input is distributed by the engine to every instance that has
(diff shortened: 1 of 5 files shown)


AGulev (approved):
Works good!

PR #9103, opened 2024-06-19
Ouput samplers to .computec resources

Jhonnyg:
* Previously, bob.jar did not output any samplers from a .compute resource, whereas the editor did.
* Compute programs have their own ComputeBuilder.java now that is used in tests as well (and tested via bob tests)

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ComputeBuilder.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ComputeBuilder.java
@@ -41,7 +41,7 @@ import java.io.OutputStream;
 @ProtoParams(srcClass = ComputeDesc.class, messageClass = ComputeDesc.class)
 @BuilderParams(name = "Compute", inExts = {".compute"}, outExt = ".computec")
 public class ComputeBuilder extends Builder<Void>  {
-	@Override
+    @Override
     public Task<Void> create(IResource input) throws IOException, CompileExceptionError {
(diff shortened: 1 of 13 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ComputeBuilder.java
```diff
@@ -0,0 +1,119 @@
...
+        TaskBuilder<Void> task = Task.<Void> newBuilder(this)
+                .setName(params.name())
+                .addInput(input)
+                .addOutput(input.changeExt(params.outExt()));
+
+        ComputeDesc.Builder computeBuilder = ComputeDesc.newBuilder();
+        ProtoUtil.merge(input, computeBuilder);
+
+        IResource computeProgramresource = this.project.getResource(computeBuilder.getComputeProgram()).changeExt(".cpc");
+        task.addInput(computeProgramresource);
```
JCash, line 55: Do we want "#include" support from the start?

JCash (approved):
Lgtm
One question about include support in the shaders.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ComputeBuilder.java
Jhonnyg: All shaderbuilders use the include preprocessor so it's already in place. But we should add tests for it for compute as well, I think we only do so for the other shader types

Issue #6444, opened 2022-02-28
Editor: add icons to properties fields to differentiate URL/hash


aglitchman:
**Is your feature request related to a problem? Please describe (REQUIRED):**
I have to append `_url` or `_id` (or `_hash`) to script properties names to differentiate their types in the editor, i.e.

```lua
go.property("script_url", msg.url(""))
-- or
go.property("script_id", hash(""))
```

Because they look like a field for a string type:
![image](https://user-images.githubusercontent.com/193258/155905525-85194f9b-09ba-4e56-acfb-6bd8196c6061.png)

**Describe the solution you'd like (REQUIRED):**
I would like not to append those suffixes:
```lua
go.property("script", msg.url(""))
```

And see icons like that:
![image](https://user-images.githubusercontent.com/193258/155905746-ee850d76-28a6-47fc-9647-cdf3779dd7e1.png)

**Additional context (OPTIONAL):**
It's just an idea :) 

Plus, you can add "123" for the number fields:
![image](https://user-images.githubusercontent.com/193258/155905885-e3b9678f-d4bf-4a47-9fc1-c8741af71259.png)

AGulev:
related https://github.com/defold/defold/issues/4064

PR #9113, opened 2024-06-21
Add icons to properties fields to differentiate URL/hash/number

AGulev:
Makes it easier to distinguish between number, hash, and URL `go.properties()` using icons.

CleanShot 2024-06-24 at 12 19 42@2x

Fix https://github.com/defold/defold/issues/6444


AGulev:
This is draft PR, I didn't add icons yet, because I'm not sure about implementation of javafx part.
That's how it looks now :
CleanShot 2024-06-21 at 13 08 34@2x

Please, see comments maybe you have better ideas.

editor/src/clj/editor/code/script_compilation.clj
```diff
@@ -77,10 +77,14 @@
...
-  (if (= resource/Resource prop-type)
-    {:type prop-type :ext (resource-kind-extensions workspace resource-kind)}
-    {:type prop-type}))
+(defn script-property-edit-type
+  ([workspace prop-type resource-kind & [script-property-type]]
+   (let [base-map (if (= resource/Resource prop-type)
+                    {:type prop-type :ext (resource-kind-extensions workspace resource-kind)}
+                    {:type prop-type})]
+     (if script-property-type
+       (assoc base-map :script-property-type script-property-type)
```
AGulev, line 86: I tried approach with new `g/MsgUrl` type, but then I can't add icons for the other fields because in such case I'm adding them all over the editor, not only the go propertiies. So, I decided to go with `edit-type` extra field

editor/src/clj/editor/properties_view.clj
```diff
@@ -97,18 +97,33 @@
     (ui/auto-commit! update-fn)
     (select-all-on-click!)))
 
+(defn set-text-field-icon!
+  [text-field image-url]
+  (let [style (str "-fx-background-image: url('" image-url "');"
+                   "-fx-background-repeat: no-repeat;"
+                   "-fx-background-position: right 6px center;")]
+    (.setStyle text-field style)))
```
AGulev, line 105: I spent a few hours fighting with javafx and trying to set everything using functions instead of styles, but it seems like we just override it with styles later. Because no matter what I tried I didn't see icons. If you have better ideas how to do it, let me know because I'm not an expert in JavaFX .

editor/src/clj/editor/properties_view.clj
```diff
@@ -97,18 +97,33 @@
...
+
+(defn script-property-type->prop-icon-png [script-property-type]
+  (case script-property-type
+    :script-property-type-number   "icons/16/Icons_34-Gamepad.png"
+    :script-property-type-hash     "icons/16/Icons_35-Inputbinding.png"
+    :script-property-type-url      "icons/16/Icons_36-Texture.png"))
```
AGulev, line 111: For now we need distinguish between 3 properties, later we will be able to add more if we want so.

AGulev:
UPD:
CleanShot 2024-06-21 at 17 49 45@2x


CleanShot 2024-06-21 at 17 49 55@2x

PR #9133, opened 2024-06-27
Add icons to properties fields to differentiate URL/hash/number

### LINKED ISSUE #6444: Editor: add icons to properties fields to differentiate URL/hash

AGulev:
Makes it easier to distinguish between number, hash, and URL `go.properties()` using icons.

image

Fix https://github.com/defold/defold/issues/6444


matgis (approved):
Ship it!

Issue #8062, opened 2023-09-29
Scriptable Bundle Dialogs


matgis:
The Bundle dialogs are a good candidate for us to explore and extend Editor Script support in the editor. We have a real need for feeding custom parameters to Bob in order to support proprietary target platforms, and it requires only a fairly basic set of UI controls. It also requires us to implement access to User Preferences.

There are a lot of open questions about how we will enable scripting of UI elements, so there is bound to be some exploration of various points. After some prototyping, we should document our findings in a design document.

* How do we declare and refresh the view?
  * Look at the approach of modern UI frameworks like React.
    * Manual render request / `set_ui_state()` of Virtual DOM from Lua?
    * Opaque state object + methods that lets us detect changes?
    * Simple polling (potentially slow due to marshalling?)
      * Perform diff on the Lua side and only marshal change instructions?
    * Use namespaces for roots (windows) owned by a particular Editor Script?
* How will we notify the Editor Script of UI events?
  * View declares `on_ui_event()` callback, event id and context data is declared as part of the Virtual DOM?
* Start with simplified pre-configured controls. We want to hide the complexity of getting the raw JavaFX controls to behave the way we want them to.
* What about styling?
  * Ideally, we'd establish a design system and the users should not have to deal with it.
  * Still a need for control over field sizes, etc.
    * Are predefined size variants enough?
* Probably need to support a few layout containers right from the start.
  * HBox, VBox, Grid with spans
* Prototype the syntax for declaring the Virtual DOM from Lua

AGulev:
related:https://github.com/defold/defold/issues/4503

vlaaad:
Related? #7047

vlaaad:
Related: https://github.com/defold/defold/issues/9422

PR #9137, opened 2024-06-27
Implement better Lua->Clj serialization

vlaaad:
Related to #8062


editor/test/integration/editor_extensions_test.clj
```diff
@@ -375,20 +377,158 @@
...
+                            (:command (first (handler/realize-menu :editor.asset-browser/context-menu-end)))
+                            (handler/eval-contexts
+                              [(handler/->context :asset-browser {} (->StaticSelection []))]
+                              false)
+                            {})]
+      @(handler/run handler+context)
+      ;; see test.editor_script: it uses editor.transact() to set a file text, then reads
+      ;; the file text from file system, then saves, then reads it again.
+      (is (= [[:out "file read: before save = 'Initial text', after save = 'New text'"]]
+             @output)))))
```
vlaaad, line 397: Oops, this test wasn't wrapped in a `deftest`

vlaaad:
@matgis Please ignore the "build failure", it's a github glitch. Initially, I pushed the commit made on top of `dev` as a `DEFEDIT-` branch, then renamed it and pushed again, and the old failure stayed.

editor/src/clj/editor/editor_extensions/vm.clj
```diff
@@ -229,3 +229,68 @@
...
+               (if (pos? max-index)
+                 (do (transduce
+                       (comp
+                         (map (fn [i]
+                                (lua-value->string vm (.rawget ^LuaTable lua-value (unchecked-inc-int (int i))))))
+                         (interpose ", "))
+                       sb-rf
+                       sb
+                       (range max-index)))
+                 (do (transduce
```
matgis, line 278: It looks like we could unwrap both these expressions from `do`?

editor/test/integration/editor_extensions_test.clj
```diff
@@ -375,20 +377,158 @@
...
+
+(deftest coercer-test
+  (let [vm (vm/make)
+        coerce #(coerce/coerce vm %1 (vm/->lua %2))]
+
+    (testing "enum"
+      (let [enum (coerce/enum 1 4 false "foo" :bar)]
+        (is (= 1 (coerce enum 1)))
+        (is (= false (coerce enum false)))
+        (is (= "foo" (coerce enum "foo")))
```
matgis, line 407: We might want to check `identical?` instead of equality for strings that are declared as enums?

It would be nice if we enforced the property that we don't want individual copies of these strings marshaled from Lua.

editor/test/integration/editor_extensions_test.clj
```diff
@@ -375,20 +377,158 @@
...
+        (is (= ["asd" 4] (coerce tuple ["asd" 4 "skipped-extra"])))
+        (is (= ["asd" nil] (coerce tuple ["asd"])))
+        (is (= ["asd" 4] (coerce tuple {1 "asd" 10 "foo" 2 4})))
+        (is (thrown? LuaError (coerce tuple [])))
+        (is (thrown? LuaError (coerce tuple "not-a-table")))
+        (is (thrown? LuaError (coerce tuple 42)))))
+
+    (testing "untouched"
+      (let [v (vm/->lua {:a 1})]
+        (is (= v (coerce coerce/untouched v)))))
```
matgis, line 436: Should we check `identical?` instead?

editor/test/integration/editor_extensions_test.clj
```diff
@@ -375,20 +377,158 @@
...
+      (is (true? (coerce coerce/to-boolean "asd"))))
+
+    (testing "any"
+      (is (= {:a 1} (coerce coerce/any {:a 1})))
+      (is (= "foo" (coerce coerce/any "foo")))
+      (is (= [1 2 3] (coerce coerce/any [1 2 3])))
+      (is (= false (coerce coerce/any false))))
+
+    (testing "userdata"
+      (is (= :key (coerce coerce/userdata (vm/wrap-userdata :key))))
```
matgis, line 462: Test `identical?` with `(Object.)`?

editor/test/integration/editor_extensions_test.clj
```diff
@@ -375,20 +377,158 @@
...
+      (is (thrown? LuaError (coerce coerce/userdata "key")))
+      (is (= inc (coerce coerce/userdata inc))))
+
+    (testing "function"
+      (let [f (vm/invoke-1 vm (vm/bind (vm/read "return function() end" "test.lua") vm))]
+        (is (= f (coerce coerce/function f)))
```
matgis, line 468: Test `identical?`

editor/test/integration/editor_extensions_test.clj
```diff
@@ -375,20 +377,158 @@
...
+      (is (thrown? LuaError (coerce (coerce/coll-of coerce/integer) [1 2 3 nil 5])))
+      (is (= [nil nil nil nil 5] (coerce (coerce/coll-of (coerce/one-of coerce/integer coerce/null)) {5 5})))
+      (is (thrown? LuaError (coerce (coerce/coll-of coerce/integer) {5 5})))
+      (testing "min-count"
+        (is (= [1 2] (coerce (coerce/coll-of coerce/any :min-count 2) [1 2])))
+        (is (= [1 2 3] (coerce (coerce/coll-of coerce/any :min-count 2) [1 2 3])))
+        (is (thrown? LuaError (coerce (coerce/coll-of coerce/any :min-count 2) [1]))))
+      (testing "distinct"
+        (is (= [1 2 3] (coerce (coerce/coll-of coerce/any :distinct true) [1 2 3])))
+        (is (thrown? LuaError (coerce (coerce/coll-of coerce/any :distinct true) [1 2 1])))))
```
matgis, line 495: Is `coll-of` the best name if we're strictly coercing to vectors?

editor/test/integration/editor_extensions_test.clj
```diff
@@ -375,20 +377,158 @@
...
+      (is (thrown? LuaError (coerce (coerce/record :opt {:a coerce/integer}) {:x 1 :y 2 :a "not-an-int"})))
+      (is (= {:a 1} (coerce (coerce/record :opt {:a coerce/integer}) {:x 1 :y 2 :a 1})))
+      (is (= {:a 1} (coerce (coerce/record
+                              :req {:a coerce/integer}
+                              :opt {:b coerce/integer})
+                            {:a 1})))
+      (is (= {:a 1 :b 2} (coerce (coerce/record
+                                   :req {:a coerce/integer}
+                                   :opt {:b coerce/integer})
+                                 {:a 1 :b 2}))))
```
matgis, line 514: Similarly, `record` seems like we might want to use for coercing into Clojure record types at some point? Maybe `coece/hashmap` would be better?

editor/test/integration/editor_extensions_test.clj
```diff
@@ -375,20 +377,158 @@
...
+      (is (= 12 (coerce (coerce/one-of coerce/string coerce/integer) 12)))
+      (is (thrown? LuaError (coerce (coerce/one-of coerce/string coerce/integer) false))))
+
+    (testing "by-key"
+      (let [by-key (coerce/by-key :type
+                                  {:rect (coerce/record :req {:width coerce/number
+                                                              :height coerce/number})
+                                   :circle (coerce/record :req {:radius coerce/number})})]
+        (is (= {:type :rect :width 15 :height 20}
+               (coerce by-key {"type" "rect" "width" 15.0 "height" 20.0})))
```
matgis, line 527: Why doesn't this fail? 🤔

`15` would not be equal to `15.0` under normal rules.

editor/src/clj/editor/editor_extensions/coerce.clj
```diff
@@ -0,0 +1,322 @@
...
+              (failure x "is not an array")))]
+    (cond-> f
+
+            min-count
+            (wrap-with-pred
+              #(<= ^long min-count (count %))
+              (str "needs at least " min-count " element" (when (< 1 ^long min-count) "s")))
+
+            distinct
+            (wrap-with-pred #(apply distinct? %) "should not have repeated elements"))))
```
matgis, line 225: This looks like a fairly involved process in order to support coercing a sparse hashmap into a non-sparse vector. I'm not sure is it worth the effort?

I think we will have some APIs that expect vectors and some that expect hashmaps. You could argue a sparse Lua table should not be able to be coerced to a vector at all, and that we should only coerce the array part of Lua tables and ignore the hashmap part when we need a vector.

I worry being flexible will significantly slow down marshaling for very little real benefit. Better to not allow it at all.

matgis (commented):
Looks very nice to use!

I'm going to commit the save-data argument fix to `disk/async-save!` directly into `dev`, so don't worry about it.

I'm not sure my finding warrant changes before merging - I'll let you be the judge. But don't feel like you have to fix things before leaving for vacation. I don't think I'm very likely to touch these files, so you likely won't run into merge conflicts if you leave it for a while.

editor/src/clj/editor/editor_extensions.clj
```diff
@@ -74,19 +74,20 @@
...
-                               :catch nil
-                               (rt/->clj rt (rt/invoke-immediate rt lua-fn lua-opts evaluation-context)))]
-                [path ret])))
+              (when-let [lua-ret (error-handling/try-with-extension-exceptions
+                                   :display-output! display-output!
+                                   :label (str label " in " path)
+                                   :catch nil
+                                   (rt/invoke-immediate rt lua-fn lua-opts evaluation-context))]
+                (when-not (rt/valid? rt coerce/null lua-ret)
+                  [path lua-ret]))))
```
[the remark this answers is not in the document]
vlaaad: Unfortunate naming, I'll rename to `coerces-to?`

editor/src/clj/editor/editor_extensions/vm.clj
vlaaad: Yes! I used to have `(.append sb "}")` on both, but then moved it after the `if`.

editor/test/integration/editor_extensions_test.clj
vlaaad: Yes!

editor/test/integration/editor_extensions_test.clj
vlaaad: Yes!

editor/test/integration/editor_extensions_test.clj
vlaaad: Yes.

editor/test/integration/editor_extensions_test.clj
vlaaad: Yep.

editor/test/integration/editor_extensions_test.clj
vlaaad: Yes, I'll rename to `vector-of`

editor/test/integration/editor_extensions_test.clj
vlaaad: We are checking deserialization from Lua, and Lua always represents whole numbers like `1.0` as integers. So, when we convert a Clojure map `{"type" "rect" "width" 15.0 "height" 20.0}` to Lua, it will be serialized as `{type = "rect", width = 15, height = 20}` Lua table.

editor/test/integration/editor_extensions_test.clj
vlaaad: I wanted to avoid the word `map` because we might want to introduce `(map-of key-coercer val-coercer)` later for homogeneous maps. I agree that `record` is not a very good term either. Clojure spec uses the term `keys`. Malli uses calls these `map`, so maybe it's fine? Yeah, I'll rename it to `hashmap` :)

editor/src/clj/editor/editor_extensions/coerce.clj
vlaaad: Yes, I'll simplify the code to use the Lua array semantics.

vlaaad:
Review changes: https://github.com/defold/defold/pull/9137/files/53e0f9c1aa57a9ac41982b05244ca31b0169713a..cf73a40bc4c6492bf8b85af14f1f470969558d84

editor/src/clj/editor/editor_extensions/coerce.clj
```diff
@@ -102,7 +102,7 @@
                     (comp
                       (map-indexed
                         (fn [i coercer]
-                          (coercer vm (.rawget x (unchecked-inc-int i)))))
+                          (coercer vm (vm/with-lock vm (.rawget x (unchecked-inc-int i))))))
```
matgis, line 105: Would it be more efficient to acquire the lock once, evaluate all the coercers, then release the lock?

matgis (dismissed):
Looks good to merge!

editor/src/clj/editor/editor_extensions/coerce.clj
vlaaad: Not sure what's more efficient, but I think semantically it will be better to use a single lock per table iteration, so other threads can't change the table. I fixed it now (+ I made `one-of` more performant in a happy-path)

editor/src/clj/editor/editor_extensions/coerce.clj
```diff
@@ -0,0 +1,322 @@
...
+  [& coercers]
+  {:pre [(not (empty? coercers))]}
+  (let [v (vec coercers)]
+    (fn coerce-one-of [vm x]
+      (let [ret (transduce
+                  (comp
+                    (map #(% vm x))
+                    (halt-when (complement failure?)))
+                  conj!
+                  v)]
```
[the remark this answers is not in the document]
vlaaad: Yes

PR #9138, opened 2024-06-27
Add `tilemap.get_tiles()` function

AGulev:
Added a new function `tilemap.get_tiles()` which returns a table of rows with all the tile indexes for the layer.

```lua
local left, bottom, columns_count, rows_count = tilemap.get_bounds("#tilemap")
local tiles = tilemap.get_tiles("#tilemap", "layer")
local tile, count = 0, 0
for row_index = bottom, bottom + rows_count - 1 do
	for column_index = left, left + columns_count - 1 do
		tile = tiles[row_index][column_index]
		count = count + 1
	end
end
```

### DIFF
--- a/engine/gamesys/src/gamesys/scripts/script_tilemap.cpp
+++ b/engine/gamesys/src/gamesys/scripts/script_tilemap.cpp
@@ -460,6 +460,74 @@ namespace dmGameSystem
         return TileMap_Get(L, true);
     }
 
+    /*# get all the tiles from a layer in a tilemap
+     * Retrieves all the tiles for the specified layer in the tilemap.
+     * It returns a table of tables where the keys are the
+     * tile positions (see [ref:tilemap.get_bounds()]).
+     *
+     * @name tilemap.get_tiles
+     * @param url [type:string|hash|url] the tilemap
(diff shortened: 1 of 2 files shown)


engine/gamesys/src/gamesys/scripts/script_tilemap.cpp
```diff
@@ -460,6 +460,74 @@ namespace dmGameSystem
...
 
+    /*# get all the tiles from a layer in a tilemap
+     * Retrieves all the tiles for the specified layer in the tilemap.
+     * It returns a table of tables where the keys are the
+     * tile positions (see [ref:tilemap.get_bounds()]).
+     *
+     * @name tilemap.get_tiles
+     * @param url [type:string|hash|url] the tilemap
+     * @param layer [type:string|hash] the name of the layer for the tiles
+     * @return tiles [type:table] a table of tables representing the layer
```
britzl, line 471: We should write that the returned table is a table of columns and not a table of rows.

tiles[y][x] vs tiles[x][y]

Personally I prefer tiles[x][y]

britzl (changes requested):
Looks good! But we need to document the order in the returned tiles table.

britzl (approved):

Issue #8630, opened 2024-03-06
Open in internal editor when using a external editor in preferences


devaex:
Even using an external editor like VS Code, it can be very useful to open scripts in the internal editor.

However, I have not been able to perform this simple task.

Is there any tip on how to open .lua/.script in the internal editor when an external editor is configured in the preferences?

I looked at the possibility of creating an editor script for this task, but it seems to me that it is not possible.

It would be interesting to have the option "Open in internal editor" in the context menu when clicking on a .lua/.script file.

Issue #4015, opened 2019-05-27
Sort out internal vs custom vs external code editor (DEFEDIT-850)


britzl:
Setting a custom code editor makes it impossible to use the internal editor, it that what we want?
Why do we even have Open As > External Editor?

AGulev:
related https://github.com/defold/defold/issues/8630

PR #9140, opened 2024-06-27
Editor: Default to internal code editor for .appmanifest files

### LINKED ISSUE #9120: Excluding '.appmanifest' file extension from external IDE list.
I'm using an external code editor for editing files. I set up the 'Preferences/Code' settings accordingly, and it works well. 

However, when I try to edit the '.appmanifest' file, it opens with the external editor. It's hard to remember every library and setting for the '.appmanifest' file and add them manually. Since there are properties for the '.appmanifest' file in the Defold Editor, it would be nice to prevent '.appmanifest' files from opening with the external editor, similar to how '.project' or '.go' files are handled.

matgis:
Default to using the internal code editor and its accompanying generator UI for `.appmanifest` files.

Fixes #9120

### Technical changes
* The `:use-custom-editor` boolean setting in`:view-opts` controls whether or not a code resource will open with the user's configured Custom Code Editor by default. Defaults to `true`.

### Note to reviewers
This changes the behavior of selecting **Open** on, or double-clicking, an `.appmanifest` resource when the user has a Custom Code Editor configured.

The user can still open an `.appmanifest` in their Custom Code Editor by selecting **Open As... > Code**.

Selecting **Open As... > External Editor** will open it using the application associated with the `.appmanifest` file extension in the host OS. This will show an error dialog unless you have an editor associated with `.appmanifest` files. I found this confusing at first, but it is consistent with other file types.

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -2012,6 +2012,8 @@ If you do not specifically require different script states, consider changing th
               :header (format "Unable to open '%s', since it contains unrecognizable data. Could the project be missing a required extension?" (resource/proj-path resource))})
            false)
        (if-let [custom-editor (and (#{:code :text} (:id view-type))
+                                   (or (:selected-view-type opts)
+                                       (get-in resource-type [:view-opts (:id view-type) :use-custom-editor] true))
                                    (let [ed-pref (some->
                                                    (prefs/get-prefs prefs "code-custom-editor" "")
                                                    string/trim)]
--- a/editor/src/clj/editor/app_manifest.clj
+++ b/editor/src/clj/editor/app_manifest.clj
@@ -482,4 +482,5 @@
     :icon "icons/32/Icons_05-Project-info.png"
     :node-type AppManifestNode
     :view-types [:code :default]
+    :view-opts {:code {:use-custom-editor false}}
     :lazy-loaded true))


editor/src/clj/editor/app_view.clj
```diff
@@ -2012,6 +2012,8 @@ If you do not specifically require different script states, consider changing th
               :header (format "Unable to open '%s', since it contains unrecognizable data. Could the project be missing a required extension?" (resource/proj-path resource))})
            false)
        (if-let [custom-editor (and (#{:code :text} (:id view-type))
+                                   (or (:selected-view-type opts)
+                                       (get-in resource-type [:view-opts (:id view-type) :use-custom-editor] true))
```
matgis, line 2016: `(:selected-view-type opts)` will have a `view-type` map if the user used **Open As...** to explicitly specify how to open the file. It will be `nil` if they used the default **Open** command, or double-clicked a file.

AGulev:
> Selecting **Open As... > External Editor** will open it using the application associated with the `.appmanifest` file extension in the host OS. This will show an error dialog unless you have an editor associated with `.appmanifest` files. I found this confusing at first, but it is consistent with other file types.

At some point we need to solve https://github.com/defold/defold/issues/4015 and https://github.com/defold/defold/issues/8630 to make it clear

AGulev (approved):
👍

PR #9141, opened 2024-06-27
Add Enter/Esc controls for properties in the editor

### LINKED ISSUE #7941: User cannot exit the editing mode of a property in the property view
Platform: macOS Ventura

Example:
Start editing the Y position of a sprite or gui node.
Notice that the user cannot press neither `Escape` nor `Enter` to exit focus.

AGulev:
Now, if a property is selected, the `Esc` button cancels uncommitted changes and unselects the property; the `Enter` button commits changes and selects the value. If the value is selected, it then unselects it.

Fix https://github.com/defold/defold/issues/7941

### DIFF
--- a/editor/src/clj/editor/properties_view.clj
+++ b/editor/src/clj/editor/properties_view.clj
@@ -160,7 +160,7 @@
                                       (properties/read-only? property))))
         cancel-fn update-prop-fn
         update-fn (fn [_]
-                    (if-let [v (field-expression/to-int (.getText text))]
+                    (when-let [v (field-expression/to-int (.getText text))]
                       (let [property (property-fn)]
                         (properties/set-values! property (repeat v))
                         (update-prop-fn nil))))]
@@ -300,9 +300,9 @@
                 cancel-fn (fn [_]
                             (let [property (property-fn)
                                   current-vals (properties/values property)]
-                              (update-ui-fn current-vals (properties/validation-message property)
-                                            (properties/read-only? property))
-                              ))
+                              (update-ui-fn current-vals
+                                            (properties/validation-message property)
+                                            (properties/read-only? property))))
                 update-fn (fn update-fn [_]
                             (let [property (property-fn)
                                   current-vals (properties/values property)
@@ -548,7 +548,7 @@
                                     (properties/validation-message property)
                                     (properties/read-only? property))))
         update-fn (fn [_]
-                      (properties/set-values! (property-fn) (repeat (.getText text))))]
+                    (properties/set-values! (property-fn) (repeat (.getText text))))]
     (customize! text update-fn cancel-fn)
     [text update-ui-fn]))
 
--- a/editor/src/clj/editor/properties_view.clj
+++ b/editor/src/clj/editor/properties_view.clj
@@ -88,32 +88,17 @@
   (doto t
     (GridPane/setHgrow Priority/ALWAYS)
     (ui/on-action! update-fn)
+    (ui/on-cancel! cancel-fn)
     (ui/auto-commit! update-fn)
-    (select-all-on-click!)
-    (ui/bind-key! "Esc" (fn []
-                          (cancel-fn t)
-                          (when-let [parent (.getParent t)]
-                            (.requestFocus parent))))
-    (ui/bind-key! "Enter" (fn []
-                            (update-fn t)
-                            (if (zero? (.getLength (.getSelection t)))
-                              (.selectAll t)
-                              (.deselect t))))))
+    (select-all-on-click!)))
 
 (defmethod customize! TextArea [^TextArea t update-fn cancel-fn]
   (doto t
     (GridPane/setHgrow Priority/ALWAYS)
+    (ui/on-action! update-fn)
+    (ui/on-cancel! cancel-fn)
     (ui/auto-commit! update-fn)
-    (select-all-on-click!)
-    (ui/bind-key! "Shortcut+Enter" (fn []
-                                     (update-fn t)
-                                     (if (zero? (.getLength (.getSelection t)))
-                                       (.selectAll t)
-                                       (.deselect t))))
-    (ui/bind-key! "Esc" (fn []
-                          (cancel-fn t)
-                          (when-let [parent (.getParent t)]
-                            (.requestFocus parent))))))
+    (select-all-on-click!)))
 
 (defn- old-num->parse-num-fn [old-num]
   {:pre [(or (number? old-num) (nil? old-num))]}
@@ -561,8 +546,9 @@
                       (update-ui-fn (properties/values property)
                                     (properties/validation-message property)
                                     (properties/read-only? property))))
-        update-fn #(properties/set-values! (property-fn) (repeat (.getText text)))]
-    (customize! text (fn [_] (update-fn)) cancel-fn)
+        update-fn (fn [_]
+                      (properties/set-values! (property-fn) (repeat (.getText text))))]
+    (customize! text update-fn cancel-fn)
     [text update-ui-fn]))
 
 (defmethod create-property-control! :default [_ _ _]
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/properties_view.clj
```diff
@@ -82,16 +82,25 @@
...
     (ui/on-action! update-fn)
     (ui/auto-commit! update-fn)
-    (select-all-on-click!)))
-
-(defmethod customize! TextArea [^TextArea t update-fn]
+    (select-all-on-click!)
+    (ui/bind-key! "Esc" (fn []
+                          (cancel-fn t)
+                          (when-let [parent (.getParent t)]
+                            (.requestFocus parent))))
```
AGulev, line 96: Cancel input and unselect the field: 

https://github.com/defold/defold/assets/2209596/59744a3f-87d9-418e-9d3a-4098ef9d7735

editor/src/clj/editor/properties_view.clj
```diff
@@ -82,16 +82,25 @@
...
+    (ui/bind-key! "Enter" (fn []
+                            (update-fn t)
+                            (if (zero? (.getLength (.getSelection t)))
+                              (.selectAll t)
+                              (.deselect t))))))
```
AGulev, line 101: Update(commit) input and select value:

https://github.com/defold/defold/assets/2209596/30e049fb-bf55-4e92-9cd9-15d818595d22

editor/src/clj/editor/properties_view.clj
```diff
@@ -82,20 +82,38 @@
...
   (doto t
     (GridPane/setHgrow Priority/ALWAYS)
     (ui/auto-commit! update-fn)
-    (select-all-on-click!)))
+    (select-all-on-click!)
+    (ui/bind-key! "Shortcut+Enter" (fn []
+                                     (update-fn t)
+                                     (if (zero? (.getLength (.getSelection t)))
+                                       (.selectAll t)
+                                       (.deselect t))))
```
AGulev, line 112: in multiline field `Enter` is a new line, so we use "Shortcut+Enter" instead

editor/src/clj/editor/properties_view.clj
```diff
@@ -129,46 +147,59 @@
...
-                           (update-ui-fn (properties/values property)
-                                         (properties/validation-message property)
-                                         (properties/read-only? property)))))]
-    (customize! text update-fn)
+        update-prop-fn (fn [_]
+                      (let [property (property-fn)]
+                        (update-ui-fn (properties/values property)
+                                      (properties/validation-message property)
+                                      (properties/read-only? property))))
+        cancel-fn update-prop-fn
```
AGulev, line 176: not sure if it's needed for readability

editor/src/clj/editor/properties_view.clj
```diff
@@ -129,46 +147,59 @@
...
+        update-fn (fn [_]
+                    (if-let [v (field-expression/to-int (.getText text))]
+                      (let [property (property-fn)]
+                        (properties/set-values! property (repeat v))
+                        (update-prop-fn nil))))]
```
AGulev, line 181: in this case we call it anyway, not sure if it's by design, but I didn't change it. But I named function `update-prop-fn` instead of cancel to make it clear

matgis (changes requested):
Nice! However, we already had code for committing the changes on Enter. If it broke for some reason, we should try to figure out why to make sure we fix it in the correct place. We had a bit of a discussion offline about this, but in summary:

We have a behavior called `auto-commit`, which would commit property changes to the graph when focus leaves a text field. However, for efficiency we only want to do it when there have been actual edits. The `auto-commit` behavior clears the `::auto-commit` flag as the field gets focus (we don't have any uncommitted edits at this point). When the user types into the text field, the `::auto-commit` flag is set to `true` (we may now have uncommitted changes). When the field loses focus, we call the function that commits the changes to the graph if the `::auto-commit` flag is `true`.

Previously the Enter key would run the `on-action!` event handler for the text field that had focus. The `on-action!` handler would set the `::auto-commit` flag to `false` before committing the changes to the graph. The field no longer has uncommitted changes, so we would not want it to commit the changes again when the user decides to move on to another field.

What I would like to see, is that we'd get the `on-action!` handler to work for text fields again. I'm not sure why it broke. Perhaps the OnAction event handler is not being called anymore after some JavaFx update made in the past, or something. If that's the case, we could re-instate the behavior by calling `ui/bind-key!` instead of `.setOnAction` in the `on-action!` implementation for `TextField`. I would then propose we add a new protocol alongside the existing `HasAction` protocol:

```clj
(defprotocol Cancellable
  (on-cancel! [this cancel-fn]))
```

And implement it for `TextField`:

```clj
(declare bind-key!) ; "forward declaration", since this is defined later in `ui.clj`.

(extend-type TextField
  HasAction
  (on-action! [this action-fn]
    ;; We could use `.setOnAction` to install the event handler here, or we
    ;; could use `bind-key!` if `.setOnAction` doesn't work anymore. Whichever
    ;; we do, it is important that we set ::auto-commit to false here like we
    ;; used to do in the existing event handler.
    ...)
  Cancellable
  (on-cancel! [this cancel-fn]
    (bind-key! "Esc" (fn []
                       (cancel-fn t)
                       (when-let [parent (.getParent t)]
                         (.requestFocus parent))))))
```

And then do the same for `TextArea`. We would then call both `on-action!` and `on-cancel!` from `property-view/customize!` to install the behaviors.


Alternatively, we could get rid of the `on-action!` implementation on `TextField` and just call `ui/bind-key!` directly from `property-view/customize!`. However, `ui/on-action!` might be used for other `TextFields` somewhere, and fixing it feels more like the "proper" fix. Regardless, we should make sure to set `::auto-commit` to `false` when the value is committed to the graph on Enter.

editor/src/clj/editor/ui.clj
```diff
@@ -750,14 +753,43 @@
...
+                           (.deselect node))
+                         (update-fn e))))
+  Cancellable
+  (on-cancel! [node cancel-fn]
+    (bind-key! node "Esc" (fn []
+                            (cancel-fn node)
+                            (when-let [parent (.getParent node)]
+                              (.requestFocus parent))))))
+
+(extend-type TextArea
```
AGulev, line 776: extend TextArea using the same protocols we use for text field, but taken into account implementation details

editor/src/clj/editor/properties_view.clj
```diff
@@ -215,17 +231,21 @@
...
+                              (update-ui-fn current-vals
+                                            (properties/validation-message property)
+                                            (properties/read-only? property))))
                 update-fn (fn update-fn [_]
                             (let [property (property-fn)
                                   current-vals (properties/values property)
                                   old-num (ffirst current-vals)
                                   num (parse-num (.getText text-field) old-num)]
-                              (if num
+                              (if (and num (not= num old-num))
```
AGulev, line 245: this small check fixes issue with not adding `0` in  the end (the same for Num properties)

https://github.com/defold/defold/assets/2209596/a3fe35b1-62f7-4869-8a94-56e0718bc7cd

AGulev:
Thank you, your explanations helped a lot!

editor/src/clj/editor/ui.clj
```diff
@@ -750,14 +753,43 @@
   (value [this] (.getValue this))
   (value! [this val] (with-on-edit-event-suppressed! this (.setValue this val))))
 
+(declare bind-key!)
+
 (extend-type TextField
   HasAction
-  (on-action! [node fn]
```
AGulev, line 755: it's better avoid using `fn` as a function argument, because it shadows `fn` keyword

matgis (approved):
LGTM!

Good work on figuring out the details around how this should work! We should port this behavior over to the Form View controls when we get the chance.

And thanks for indulging me during the lengthy review process! ❤️

AGulev:
> And thanks for indulging me during the lengthy review process! ❤️

Taking into account my experience and my knowledge in the area.... Thank you for your patience.

Issue #8550, opened 2024-02-15
Editor 'Open as Code' for .project and .settings files


dri-richard:
**Is your feature request related to a problem? Please describe:**
Right clicking on certain file types in the asset browser - project, settings and others - includes 'Open As -> Text', but that view is read-only.
Certain tasks such as adding an entirely new section or setting are not possible in the default Form view.

**Describe the solution you'd like :**
I would like to be able to open such files in an editable Code view.

**Describe alternatives you've considered :**
The alternative is to select Show in Desktop, and then open from Finder (or Windows explorer) in an external text editor.

Issue #8126, opened 2023-10-10
Make HTML files editable inside Defold's script editor


dima13230:
**Is your feature request related to a problem? Please describe (REQUIRED):**
Currently, I'm working on the SDK that requires to edit both Lua and HTML files often enough and I always have to look into Lua code to then add corresponding code into HTML file, and for Lua I prefer using Defold's script editor. However, I cannot edit HTML files inside Defold so I have to open and keep opened window of another text editor specifically for HTML, which I find quite inconvenient.

**Describe the solution you'd like (REQUIRED):**
I would like to be able to open and edit HTML files inside Defold's editor

**Describe alternatives you've considered (REQUIRED):**
The only alternative I can think of is to nevermind it and continue using external editor for HTML :smile: 

**Additional context (OPTIONAL):**
I have already started working on PR for this feature:
![screenshot](https://github.com/defold/defold/assets/12088459/f574d7a2-2479-40d8-825d-7f5485e42f60)

The main part is done, I'm currently working on syntax highlighting. Mentioning that, I have a little problem implementing it.
Here's my HTML grammar code:

```clojure
(ns editor.code.lang.html
  (:require [editor.code.data :as data])
  (:import [java.io PushbackReader]))

(set! *warn-on-reflection* true)
(set! *unchecked-math* :warn-on-boxed)

(def grammar
  {:name "HTML"
   :scope-name "source.html"
   ;; indent patterns shamelessly stolen from textmate:
   ;; https://github.com/textmate/html.tmbundle/blob/master/Preferences/Miscellaneous.plist
   :indent {
            :end #"(?x)
		^\s*
		(&lt;/(?!html)
		  [A-Za-z0-9]+\b[^&gt;]*&gt;
		|--&gt;
		|&lt;\?(php)?\s+(else(if)?|end(if|for(each)?|while))
		|\}
		)"
            }
   :patterns [
              {:match #"<[^>]*>"
               :name "tag.language.html"}
              {:match #"\b(?:true|false|null)\b"
               :name "constant.language.html"}
              {:match #"(?x)-?(?:0|[1-9]\d*)(?:\n(?:\n\.\d+)?(?:[eE][+-]?\d+)?)?"
               :name "constant.numeric.html"}
              {:begin #"\""
               :begin-captures {0 {:name "punctuation.definition.string.begin.html"}}
               :end #"\""
               :end-captures {0 {:name "punctuation.definition.string.end.html"}}
               :name "string.quoted.double.html"}
[...]

PR #9143, opened 2024-06-28
Editor: Open As Code in Defold Editor or Custom Editor

### LINKED ISSUE #4015: Sort out internal vs custom vs external code editor (DEFEDIT-850)

### LINKED ISSUE #8630: Open in internal editor when using a external editor in preferences

matgis:
* If you have a **Custom Code Editor** configured in **Preferences**, you now have the option to **Open As > Code in Defold Editor** alongside **Open As > Code in Custom Editor**.
* The **Open As > External Editor** menu option has been relabeled to **Open As > Associated Application** to differentiate it from the **Custom Code Editor** setting.

Fixes #4015
Fixes #8630

### Technical changes
* **Open As > Text** is now available for `.html` and `.animationset` files so they are consistent with similar resource types.

### DIFF
--- a/editor/src/clj/editor/animation_set.clj
+++ b/editor/src/clj/editor/animation_set.clj
@@ -191,4 +191,4 @@
     :sanitize-fn sanitize-animation-set
     :node-type AnimationSetNode
     :ddf-type Rig$AnimationSetDesc
-    :view-types [:cljfx-form-view]))
+    :view-types [:cljfx-form-view :text]))
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -1993,6 +1993,13 @@ If you do not specifically require different script states, consider changing th
             (string/replace tmpl (format "{%s}" (name key)) (str val)))
     tmpl args))
 
+(defn- custom-code-editor-executable-path-preference
+  ^String [prefs]
+  (some-> prefs
+          (prefs/get-prefs "code-custom-editor" nil)
+          (string/trim)
+          (not-empty)))
+
 (defn open-resource
   ([app-view prefs workspace project resource]
    (open-resource app-view prefs workspace project resource {}))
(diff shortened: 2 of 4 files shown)


editor/src/clj/editor/app_view.clj
```diff
@@ -2121,23 +2150,26 @@ If you do not specifically require different script states, consider changing th
...
 
 (handler/defhandler :reopen-recent-file :global
   (enabled? [prefs workspace evaluation-context app-view]
     (recent-files/exist-closed? prefs workspace app-view evaluation-context))
   (run [prefs app-view workspace project]
     (g/with-auto-evaluation-context evaluation-context
       (let [[resource view-type] (recent-files/last-closed prefs workspace app-view evaluation-context)]
-        (open-resource app-view prefs workspace project resource {:selected-view-type view-type})))))
+        (open-resource app-view prefs workspace project resource {:selected-view-type view-type
+                                                                  :use-custom-editor false})))))
```
matgis, line 2172: Opened files only get added to the Recent Files list when they are opened in the Defold editor, so we should always reopen them like that.

editor/src/clj/editor/animation_set.clj
```diff
@@ -191,4 +191,4 @@
     :sanitize-fn sanitize-animation-set
     :node-type AnimationSetNode
     :ddf-type Rig$AnimationSetDesc
-    :view-types [:cljfx-form-view]))
+    :view-types [:cljfx-form-view :text]))
```
AGulev, line 194: hmm, it isn't protobuf?

editor/src/clj/editor/html.clj
```diff
@@ -44,5 +44,5 @@
     :search-fn search-fn
     :search-value-fn search-value-fn
     :node-type HtmlNode
-    :view-types [:html]
+    :view-types [:html :text]
     :view-opts nil))
```
AGulev, line 48: Maybe we should think about it as part of this task, but it's kinda related:
https://github.com/defold/defold/issues/8550

this one is about html (but probably with much more functions we curently want) https://github.com/defold/defold/issues/8126

AGulev (approved):
It works well, should be much easier to use it now

editor/src/clj/editor/animation_set.clj
matgis: Yeah, it is Protobuf-based, but we allow you to inspect its text representation. We do this for all the other Protobuf-based types. For some reason it was missing for `.animationset`.

editor/src/clj/editor/html.clj
matgis: Yeah, it would be nice. But it is a lot more work to have it be editable, not just viewable.

PR #9144, opened 2024-06-28
Editor: Support viewing library resources in Custom Code Editor

### LINKED ISSUE #4964: Cannot open Builtins scripts or Extensions scrips with external code editors.
If I have a external code editor set in settings, when I try to open a script inside Builtin or inside a Extension the editor opens a blank document.

1. Go to Settings and set any external editor (Tested with Atom and Win7 notepad)
2. Open Builtins folder from any project.
3. Look for any script and double-click.

It should open with the external editor or at least with the internal editor.

Version [e.g. 1.2.170]

 - Platforms: tested on Windows 7.

Similar to https://github.com/defold/defold/issues/4015 ... internal editor gets completely disabled, in order to open these scripts I need to remove the external editor temporally.

matgis:
It is now possible to view library resources with the configured Custom Code Editor.

Fixes #4964

### Technical changes
* Added boolean argument to `fs/set-executable!` and `fs/set-writable!`.

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -2030,7 +2030,7 @@ If you do not specifically require different script states, consider changing th
                       (custom-code-editor-executable-path-preference prefs))))]
          (let [cursor-range (:cursor-range opts)
                arg-tmpl (string/trim (if cursor-range (prefs/get-prefs prefs "code-open-file-at-line" "{file}:{line}") (prefs/get-prefs prefs "code-open-file" "{file}")))
-               arg-sub (cond-> {:file (resource/abs-path resource)}
+               arg-sub (cond-> {:file (resource/externally-available-absolute-path resource)}
                                cursor-range (assoc :line (CursorRange->line-number cursor-range)))
                args (->> (string/split arg-tmpl #" ")
                          (map #(substitute-args % arg-sub)))]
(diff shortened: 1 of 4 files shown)


AGulev:
Now it's impossible just open a code file in a lib. it opens as text:
https://github.com/defold/defold/assets/2209596/65a6933a-0a67-41cd-98ed-b9d173a86ef6

matgis:
This was also the case before this change. It opens like that because the resource is read-only inside the `.zip` file. I want to make it open in the proper code editor in some kind of read-only mode, but that is a separate issue.

AGulev (changes requested):
Now it's impossible just open a code file in a lib. if no custom editor specified 
 it opens as text:

https://github.com/defold/defold/assets/2209596/864d470b-c2d8-4bdd-b4b3-7c21cc9be6fb

AGulev:
@matgis then it's new 1.9.1 regression. It was fine before 1.9.0:


https://github.com/defold/defold/assets/2209596/06cbbfda-f24a-49e9-8f3a-176ad030005e

matgis:
Huh. Good to know! It may have happened with the save-value changes. I'll have to look into it on Monday. I checked in `dev` a few commits prior to the recent **Open As** changes, and it does open the library files in the primitive text view before this PR.

AGulev (approved):

PR #9145, opened 2024-06-28
issue-7941 review fixes

matgis:
I made some fixes while testing, and instead of putting in review comments, I made a PR into your branch. I hope you don't mind! 😅 It seemed easier!

### DIFF
--- a/editor/src/clj/editor/properties_view.clj
+++ b/editor/src/clj/editor/properties_view.clj
@@ -160,7 +160,7 @@
                                       (properties/read-only? property))))
         cancel-fn update-prop-fn
         update-fn (fn [_]
-                    (if-let [v (field-expression/to-int (.getText text))]
+                    (when-let [v (field-expression/to-int (.getText text))]
                       (let [property (property-fn)]
                         (properties/set-values! property (repeat v))
                         (update-prop-fn nil))))]
@@ -300,9 +300,9 @@
                 cancel-fn (fn [_]
                             (let [property (property-fn)
                                   current-vals (properties/values property)]
-                              (update-ui-fn current-vals (properties/validation-message property)
-                                            (properties/read-only? property))
-                              ))
+                              (update-ui-fn current-vals
+                                            (properties/validation-message property)
+                                            (properties/read-only? property))))
                 update-fn (fn update-fn [_]
                             (let [property (property-fn)
                                   current-vals (properties/values property)
@@ -548,7 +548,7 @@
                                     (properties/validation-message property)
                                     (properties/read-only? property))))
         update-fn (fn [_]
-                      (properties/set-values! (property-fn) (repeat (.getText text))))]
+                    (properties/set-values! (property-fn) (repeat (.getText text))))]
     (customize! text update-fn cancel-fn)
     [text update-ui-fn]))
 
--- a/editor/src/clj/editor/ui.clj
+++ b/editor/src/clj/editor/ui.clj
@@ -584,6 +584,15 @@
 (defn suppress-auto-commit! [^Node node]
   (user-data! node ::suppress-auto-commit true))
 
+(defn- clear-auto-commit! [^Node node]
+  ;; Clear the auto-commit flag. You should call this whenever data has been
+  ;; synced with the graph while the field still has focus. This ensures the
+  ;; unedited value will not be committed to the graph unnecessarily after the
+  ;; user moves focus to another control. Further edits will re-apply the
+  ;; auto-commit flag, but if they leave without editing, we won't commit.
+  (when (user-data node ::auto-commit)
+    (user-data! node ::auto-commit false)))
+
 (defn increase-on-edit-event-suppress-count! [editable]
   (when-some [suppress-count (user-data editable ::on-edit-event-suppress-count)]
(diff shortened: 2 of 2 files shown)


editor/src/clj/editor/properties_view.clj
```diff
@@ -160,7 +160,7 @@
                                       (properties/read-only? property))))
         cancel-fn update-prop-fn
         update-fn (fn [_]
-                    (if-let [v (field-expression/to-int (.getText text))]
+                    (when-let [v (field-expression/to-int (.getText text))]
```
matgis, line 163: Use `when-let` to show clearly that there is no `else` branch.

editor/src/clj/editor/properties_view.clj
```diff
@@ -300,9 +300,9 @@
                 cancel-fn (fn [_]
                             (let [property (property-fn)
                                   current-vals (properties/values property)]
-                              (update-ui-fn current-vals (properties/validation-message property)
-                                            (properties/read-only? property))
-                              ))
+                              (update-ui-fn current-vals
+                                            (properties/validation-message property)
+                                            (properties/read-only? property))))
```
matgis, line 305: Fixed formatting.

editor/src/clj/editor/properties_view.clj
```diff
@@ -548,7 +548,7 @@
                                     (properties/validation-message property)
                                     (properties/read-only? property))))
         update-fn (fn [_]
-                      (properties/set-values! (property-fn) (repeat (.getText text))))]
+                    (properties/set-values! (property-fn) (repeat (.getText text))))]
```
matgis, line 551: Fixed indentation.

editor/src/clj/editor/ui.clj
```diff
@@ -584,6 +584,15 @@
...
 
+(defn- clear-auto-commit! [^Node node]
+  ;; Clear the auto-commit flag. You should call this whenever data has been
+  ;; synced with the graph while the field still has focus. This ensures the
+  ;; unedited value will not be committed to the graph unnecessarily after the
+  ;; user moves focus to another control. Further edits will re-apply the
+  ;; auto-commit flag, but if they leave without editing, we won't commit.
+  (when (user-data node ::auto-commit)
+    (user-data! node ::auto-commit false)))
+
```
matgis, line 595: Broke out this into a function and added longer explanation.

editor/src/clj/editor/ui.clj
```diff
@@ -759,27 +768,24 @@
   HasAction
   (on-action! [node update-fn]
     (.setOnAction node (event-handler e
-                         ;; Clear the auto-commit flag. Further edits will re-apply it.
-                         (when (user-data node ::auto-commit)
-                           (user-data! node ::auto-commit false))
+                         (clear-auto-commit! node)
+                         (update-fn e)
```
matgis, line 772: Use shared function to clear auto-commit flag.

Move `update-fn` call to before we adjust the selection. Not 100% sure of the correct location of the call, but now it is consistent with `TextArea`. I would think it shouldn't matter?

editor/src/clj/editor/ui.clj
```diff
@@ -759,27 +768,24 @@
...
                          (if (zero? (.getLength (.getSelection node)))
                            (.selectAll node)
-                           (.deselect node))
-                         (update-fn e))))
+                           (.deselect node)))))
   Cancellable
   (on-cancel! [node cancel-fn]
     (bind-key! node "Esc" (fn []
                             (cancel-fn node)
+                            (clear-auto-commit! node)
```
matgis, line 780: I checked with debug prints, and we should clear the auto-commit flag after calling the `cancel-fn`. If we do not, we will commit the field value when the `.requestFocus` call resolves. The property editor doesn't create an additional undo step because it also compares the property values against the old values internally, but we want to be able to use `on-cancel!` elsewhere too!

AGulev: I wanted to add it but stuck with `fn` issue and forgot =/

editor/src/clj/editor/ui.clj
AGulev: yes, sure.

AGulev (approved):

AGulev:
> I made some fixes while testing, and instead of putting in review comments, I made a PR into your branch. I hope you don't mind! 😅 It seemed easier!

Sorry for that. 
I'm fine with fixing whatever you think is far from good enough. Feel free to let me know about such cases.

editor/src/clj/editor/properties_view.clj
AGulev: I'm still in a phase with the editor and clj when I prefer to avoid touching code I didn't write and I don't need for my fix. 
But I mentioned in comments that I was confused by that, yes `when-let` is much better in such cases

editor/src/clj/editor/properties_view.clj
AGulev: I use 
CleanShot 2024-06-28 at 22 07 45@2x
everywhere I touch something. Not sure if it did something wrong or I just missed this part

editor/src/clj/editor/ui.clj
AGulev: I think so as well

PR #9151, opened 2024-07-01
Make sure that multifield values applied right 

### LINKED ISSUE #9147: Can't set y and z properties to 0.0 on Editor
Latest **1.9.1**

Position, rotation and scale values can't set to 0.0 also scale value return 0.000001

### Steps to reproduce

- Creat new GameObject
- Set position/rotation y or z to 3
- Try to set it back to 0  (Works for X )
- Returns to 3
------
- Set scale y or z to 3
- Set again to 0 -> 0.000001
- Set again to 1 -> 0.000001 (it works if you set it > 2)

https://github.com/defold/defold/assets/1324635/ff8fe608-9501-4777-8421-5b22d383f8d3

<hr/>

<table>
<tr><td>Build time</td><td>2024-06-29T07:36:54.365170</td></tr>
<tr><td>Defold channel</td><td>alpha</td></tr>
<tr><td>Defold editor sha</td><td>605867c347e5dc4b2d8ebf6871c61018133fa69e</td></tr>
<tr><td>Defold engine sha</td><td>605867c347e5dc4b2d8ebf6871c61018133fa69e</td></tr>
<tr><td>Defold version</td><td>1.9.1</td></tr>
<tr><td>GPU</td><td>Apple M2 Pro</td></tr>
<tr><td>OS version</td><td>14.5</td></tr>
<table>

AGulev:
I missed the fact that we always take the first value for the multifield input. That the reason that values which are equal to the first value can't be applied to the rest of fields.

Fix https://github.com/defold/defold/issues/9147

### DIFF
--- a/editor/src/clj/editor/properties_view.clj
+++ b/editor/src/clj/editor/properties_view.clj
@@ -240,7 +240,7 @@
                 update-fn (fn update-fn [_]
                             (let [property (property-fn)
                                   current-vals (properties/values property)
-                                  old-num (ffirst current-vals)
+                                  old-num (get (get current-vals 0) index)
                                   num (parse-num (.getText text-field) old-num)]
                               (if (and num (not= num old-num))
                                 (properties/set-values! property (mapv #(assoc % index num) current-vals))
--- a/editor/src/clj/editor/ui.clj
+++ b/editor/src/clj/editor/ui.clj
@@ -769,10 +769,11 @@
   (on-action! [node update-fn]
     (.setOnAction node (event-handler e
                          (clear-auto-commit! node)
-                         (update-fn e)
-                         (if (zero? (.getLength (.getSelection node)))
-                           (.selectAll node)
-                           (.deselect node)))))
+                         (let [length (.getLength (.getSelection node))]
+                           (update-fn e)
+                           (if (zero? length)
+                             (.selectAll node)
+                             (.deselect node))))))
   Cancellable
   (on-cancel! [node cancel-fn]
     (bind-key! node "Esc" (fn []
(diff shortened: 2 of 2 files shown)


editor/src/clj/editor/properties_view.clj
```diff
@@ -240,7 +240,7 @@
                 update-fn (fn update-fn [_]
                             (let [property (property-fn)
                                   current-vals (properties/values property)
-                                  old-num (ffirst current-vals)
+                                  old-num (get (get current-vals 0) index)
```
AGulev, line 243: The main fix.
Since we check if old-value not equal to the new one, we have to get right `old-value`

editor/src/clj/editor/ui.clj
```diff
@@ -769,10 +769,11 @@
   (on-action! [node update-fn]
     (.setOnAction node (event-handler e
                          (clear-auto-commit! node)
-                         (update-fn e)
-                         (if (zero? (.getLength (.getSelection node)))
-                           (.selectAll node)
-                           (.deselect node)))))
+                         (let [length (.getLength (.getSelection node))]
```
AGulev, line 772: `update-fn` may apply selection (when cancel input and update the field's value). So we should get length in advance to know if we need to change selection.

editor/src/clj/editor/ui.clj
```diff
@@ -785,11 +786,12 @@
   HasAction
   (on-action! [node update-fn]
     (bind-key! node "Shortcut+Enter" (fn []
-                                         (clear-auto-commit! node)
+                                       (clear-auto-commit! node)
+                                       (let [length (.getLength (.getSelection node))]
                                          (update-fn node)
-                                         (if (zero? (.getLength (.getSelection node)))
+                                         (if (zero? length)
```
AGulev, line 792: Apply the same changes for `TextArea` just for consistency

matgis (approved):
Looks good! 👍

PR #9154, opened 2024-07-02
Use Gradle as the build system for Bob.jar

AGulev:
Using Gradle for Bob.jar makes it easier to set up and develop new features for Bob.jar by speeding up iterative builds and adding the ability to run single tests when needed.

A manual for setting up IntelliJ IDEA is available [here](https://github.com/defold/defold/blob/dev/com.dynamo.cr/README.md).


com.dynamo.cr/README.md
```diff
@@ -30,28 +30,24 @@ Setup IDEA
...
+10. In  `Project Settings -> Modules -> Dependencies` click `+` then  `1 JARs or Directories...` and choose `com.dynamo.cr/com.dynamo.cr.common/ext` and repeat for `com.dynamo.cr/com.dynamo.cr.bob/lib`, `com.dynamo.cr/com.dynamo.cr.common/ext/aws`  and `com.dynamo.cr/com.dynamo.cr.bob.test/lib` ![Dependencies](https://github.com/defold/defold/assets/2209596/dd86e706-b91f-475b-b43d-aaac596ffa1f)
 
+Testing
+---------
+To be able to run single tests, change default template for JUnit tests:
+1. Open `Edit Configurations`
+2. Click `Edit Configuration Template` ![Edit configuration template](https://github.com/defold/defold/assets/2209596/1d052a0c-3e25-4dec-86f6-3d2ba42207e1)
+3. Pick JUnit and specify working derictory as `$MODULE_WORKING_DIR$/com.dynamo.cr.bob.test` ![working derictory](https://github.com/defold/defold/assets/2209596/6fb73a5f-dadb-41b0-8343-c7443d177f03)
 
 Debugging
```
AGulev, line 52: Updated manual

com.dynamo.cr/com.dynamo.cr.bob.test/build.gradle
```diff
@@ -0,0 +1,87 @@
...
+        exclude 'test/build/**/*.*'
+    }
+}
+
+task testJar(type: Test, dependsOn: createTestJar) {
+    useJUnit()
+
+    // Check if running from IntelliJ IDEA - then test bundler only for the host platform
+    doFirst {
+        if (System.getProperty('idea.active') != null) {
```
AGulev, line 65: automatically specify DM_BOB_BUNDLERTEST_ONLY_HOST if tests run from `idea`

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/bundle/test/BundlerTest.java
```diff
@@ -101,7 +101,8 @@ public static Collection<Platform[]> data() {
         String envSkipTest = System.getenv("DM_BOB_BUNDLERTEST_ONLY_HOST");
         String skipTest =  envSkipTest != null ? envSkipTest : System.getProperty("DM_BOB_BUNDLERTEST_ONLY_HOST");
         // By default property is `${DM_BOB_BUNDLERTEST_ONLY_HOST}`
-        if (skipTest != null && !skipTest.equals("0") && !skipTest.equals("${DM_BOB_BUNDLERTEST_ONLY_HOST}")) {
+        if (skipTest != null && !skipTest.isEmpty() && !skipTest.equals("0") && !skipTest.equals("${DM_BOB_BUNDLERTEST_ONLY_HOST}")) {
+            System.out.println("Bundle test only for host");
```
AGulev, line 105: Logging to see for sure when bundle test runs only for the host plaform

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ShaderProgramBuilderTest.java
```diff
@@ -334,7 +334,7 @@ public void testShaderReflection() throws Exception {
 
             debugPrintShader("Reflection Test 2", shader);
 
-            assert(shader.getTexturesCount() == 9);
+            assertEquals(8, shader.getTexturesCount());
```
AGulev, line 337: `assert` function turned off by default in jvm. After replacing to `assertEquals` I found that the value is wrong here

com.dynamo.cr/com.dynamo.cr.bob/basic.gradle
```diff
@@ -0,0 +1,86 @@
+def buildToolsDir = file('../../build_tools')
+
```
AGulev, line 2: some constants, and basic function we use in a fes places

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/AndroidBundler.java
```diff
@@ -137,8 +137,6 @@ private static void initAndroid() {
                 Bob.atomicCopy(libc_url, f, false);
             }
 
-            Bob.extract(Bob.class.getResource("/lib/android-res.zip"), rootFolder);
```
AGulev, line 140: this archive is empty for a few years, because all the resources were removed from the engine. we don't need it anymore

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/font/BMFont.java
```diff
@@ -366,7 +366,7 @@ public Boolean parseText(BufferedInputStream data) throws BMFontFormatException,
                 } break;
                 case "char":
                 {
-                    assert(this.chars <= this.charArray.size() + 1 );
+                    assert(this.chars >= this.charArray.size());
```
AGulev, line 369: The same issue with `assert`. This assert works in JUnit tests and fails build because condition always false. It is always has been like that, but `assert` turned off by default.

com.dynamo.cr/com.dynamo.cr.common/build.gradle
```diff
@@ -0,0 +1,76 @@
...
+        }
+        if (extGeneratedDir.listFiles().length == 0) {
+            throw new RuntimeException("dlib.jar is missing")
+        }
+    }
+}
+
+def protoTasks = [
+        [dir: "proto", file: "cr_ddf.proto"],
+        [dir: "proto", file: "cr_protocol_ddf.proto"],
```
AGulev, line 38: These two used by tests, and we need them to be able to run a single test. But the rest is just a duplicate of the main bob folder proto files. Having these classes in both folders end up in "duplicate symbols" error. So I removed them, and keep only unique proto files

com.dynamo.cr/.idea/com.dynamo.cr.iml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
AGulev, line 1: We don't usually commit idea files. But in this case it simplifies setup a lot. I decided to add it for now but also updated manual how to set up everything from scratch, so I can remove this.

scripts/build.py
```diff
@@ -94,7 +94,7 @@ def get_tag_suffix(self):
 def get_target_platforms():
     return BASE_PLATFORMS + build_private.get_target_platforms()
 
-PACKAGES_ALL="protobuf-3.20.1 waf-2.0.3 junit-4.6 jsign-4.2 protobuf-java-3.20.1 openal-1.1 maven-3.0.1 ant-1.9.3 vecmath vpx-1.7.0 luajit-2.1.0-6c4826f tremolo-b0cb4d1 defold-robot-0.7.0 bullet-2.77 libunwind-395b27b68c5453222378bc5fe4dab4c6db89816a jctest-0.10.2 vulkan-1.3.261.1".split()
```
AGulev, line 97: And removed, but I commit Gradle into each folder separately as we talked about it. Later we may decide to install it as well as we installed ant, but then IDE will install it's own version separately in each folder

scripts/build.py
```diff
@@ -1190,34 +1190,42 @@ def _build_engine_lib(self, args, lib, platform, skip_tests = False, dir = 'engi
...
+            return join('.', 'gradlew.bat')
+        else:  # Linux, macOS, or other Unix-like OS
+            return join('.', 'gradlew')
+
     def build_bob_light(self):
         self._log('Building bob light')
 
         bob_dir = join(self.defold_root, 'com.dynamo.cr/com.dynamo.cr.bob')
-        common_dir = join(self.defold_root, 'com.dynamo.cr/com.dynamo.cr.common')
+        # common_dir = join(self.defold_root, 'com.dynamo.cr/com.dynamo.cr.common')
```
AGulev, line 1208: It's never been used. I keep it commented in case we will need it, but I don't get what common is, and why we need it. I just found some dependencies in tests, but it throws errors only in cases I run single tests. 🤷‍♂️

JCash (approved):

Issue #4052, opened 2019-05-27
Resources from builtins and libraries should be displayed as read-only (DEFEDIT-274)


britzl:

AGulev:
Maybe it should be a message in the top of the editor window
<img width="1095" alt="image" src="https://github.com/defold/defold/assets/2209596/1b4445f3-e30f-46b3-ade3-b5d672da833e">

PR #9162, opened 2024-07-03
Use code view when opening resources as text

### LINKED ISSUE #6700: Track changes in "Open as > Text" view
The current textual view looks very outdated, while I think it can be useful to see how your changes are persisted. It does not show any changes made to the resource in the editor, even though it shows the tab with the text view as dirty. The view is not updated after saving files either.

We have the capability to keep the textual view of a resource updated when we modify it in its default view, I think we should do that. We could also use a read-only code view instead of a read-only JavaFX text field.

Delete the text view altogether?

matgis:
* The editor will now use the full-featured code editor view when you choose **Open As > Text** for a resource.
* If you have a **Custom Code Editor** configured in **Preferences**, you now have the option to **Open As > Text in Defold Editor** alongside **Open As > Text in Custom Editor**.
* Fixed an issue where it was possible to insert line breaks into the Console by pressing return.
* Fixed a CSS style regression on the **Evaluate Lua** input field in the Console.

Fixes #6700

### Technical changes
* Removed legacy `TextView`.
* Added non-editable mode to the `CodeEditorView`.
* Merged duplicate UI handlers that were shared between the code editor and the console view.
* Moved `project/connect-if-output` to `graph-util/connect-existing-outputs`.
* Moved `code.resource/Lines` `deftype` to `editor.types/Lines` so we can use it in `editor.resource-node`.
* Added read-only `lines` output to base `ResourceNode`. This will produce a vector of lines from the `save-value`. It is not `:cached`, since the `lines` input is cached in the `CodeEditorView`.
* Added a `fallback-cursor-ranges` property to the `CodeEditorView`. It can be used to store text selection state directly in the view for resource types that do not need the selection to sync up with the `lines` when editing.


editor/src/clj/editor/code/view.clj
```diff
@@ -2356,116 +2365,40 @@
...
-
-(handler/defhandler :select-beginning-of-file :code-view
-  (run [view-node] (move! view-node :selection :file-start)))
-
-(handler/defhandler :end-of-file :code-view
-  (run [view-node] (move! view-node :navigation :file-end)))
-
-(handler/defhandler :select-end-of-file :code-view
-  (run [view-node] (move! view-node :selection :file-end)))
-
```
matgis, line 2424: UI handlers for "read-only" actions, shared with the Console view, have been moved to `fundamental-read-only-handlers` later in the file.

editor/src/clj/editor/code/view.clj
```diff
@@ -3030,31 +2972,40 @@
...
+      (g/with-auto-evaluation-context evaluation-context
+        (r/ensure-loaded! resource-node evaluation-context)))
+    (let [[lines document-width]
+          (g/with-auto-evaluation-context evaluation-context
+            (let [glyph-metrics (g/node-value view-node :glyph-metrics evaluation-context)
+                  tab-spaces (g/node-value view-node :tab-spaces evaluation-context)
+                  tab-stops (data/tab-stops glyph-metrics tab-spaces)
+                  lines (g/node-value resource-node :lines evaluation-context)
+                  document-width (data/max-line-width glyph-metrics tab-stops lines)]
+              [lines document-width]))]
```
matgis, line 2994: This was slightly wrong before. Technically you should not use an `evaluation-context` after you've made any modifications to the graph, because the changes will not be seen from within the `evaluation-context`.

I've refactored the code a bit so that we make a new `evaluation-context` after we've loaded the file, but the main change here is that we only call `code.resource/ensure-loaded!` to force lazy-loaded `CodeEditorResourceNodes`. We don't want to do it for other node types.

editor/src/clj/editor/code/view.clj
```diff
@@ -3030,31 +2972,40 @@
...
+          (gu/connect-existing-outputs resource-node-type resource-node view-node
+            [[:completions :completions]
+             [:cursor-ranges :cursor-ranges]
+             [:indent-type :indent-type]
+             [:invalidated-rows :invalidated-rows]
+             [:lines :lines]
+             [:regions :regions]])))
+      (when (and is-code-resource-type
+                 (resource/file-resource? resource))
+        (lsp/open-view! lsp view-node resource lines))
```
matgis, line 3008: Don't bother with Language Server stuff unless this is a code file, and only connect outputs that exist. For non-`CodeEditorResourceNodes`, that's only really the `lines` output.

editor/src/clj/editor/code/view.clj
```diff
@@ -3102,21 +3053,20 @@
...
          canvas-repaint-info
-         cursor-repaint-info
-         completions-enabled]
+         cursor-repaint-info]
         (g/with-auto-evaluation-context evaluation-context
           [(g/node-value view-node :resource-node evaluation-context)
            (g/node-value view-node :canvas-repaint-info evaluation-context)
-           (g/node-value view-node :cursor-repaint-info evaluation-context)
-           (g/node-value view-node :completions-enabled evaluation-context)])]
+           (g/node-value view-node :cursor-repaint-info evaluation-context)])]
```
matgis, line 3060: We don't have `completions-enabled` as a property anymore. Instead, we feed `editable` as a boolean argument to this function.

editor/src/clj/editor/code/view.clj
```diff
@@ -3544,11 +3494,15 @@
 
 (defn- make-view! [graph parent resource-node opts]
   (let [{:keys [^Tab tab app-view grammar open-resource-fn project]} opts
+        basis (g/now)
+        resource-node-type (g/node-type* basis resource-node)
+        editable (and (g/has-property? resource-node-type :modified-lines)
+                      (g/connected? basis resource-node :save-data project :save-data))
```
matgis, line 3500: * The `ResourceNode` must have a `modified-lines` property to be editable using the `CodeEditorView`. This was always the case, even before this PR.
* Any `ResourceNode` that represents an editable project file will be connected to the `save-data` input, but nodes coming from library dependencies or builtins will not.

editor/src/clj/editor/code/view.clj
```diff
@@ -3544,11 +3494,15 @@
...
         grid (GridPane.)
         canvas (Canvas.)
         canvas-pane (Pane. (into-array Node [canvas]))
         undo-grouping-info (pair :navigation (gensym))
-        lsp (lsp/get-node-lsp resource-node)
+        lsp (lsp/get-node-lsp basis resource-node)
```
matgis, line 3505: Tiny cleanup: We can reuse the `basis` from above to evaluate this as well. If we do not supply it here, `get-node-lsp` will implicitly call `(g/now)` internally.

editor/src/clj/editor/code/view.clj
```diff
@@ -3583,14 +3536,15 @@
...
+        replace-bar (setup-replace-bar! (ui/load-fxml "replace.fxml") view-node editable)
         repainter (ui/->timer "repaint-code-editor-view"
                               (fn [_ elapsed-time _]
                                 (when (and (.isSelected tab) (not (ui/ui-disabled?)))
-                                  (repaint-view! view-node elapsed-time {:cursor-visible true}))))
+                                  (repaint-view! view-node elapsed-time {:cursor-visible true :editable editable}))))
         dispose-hover! (create-hover! view-node canvas tab)
         dispose-breakpoint-editor! (create-breakpoint-editor! view-node canvas tab)
         context-env {:clipboard (Clipboard/getSystemClipboard)
+                     :editable editable
```
matgis, line 3547: Putting the `editable` boolean in the UI `context` for the JavaFx `Controls` enables us to access it in the `(active? [editable] editable)` handlers.

editor/src/clj/editor/code/view.clj
```diff
@@ -3607,17 +3561,20 @@
...
       (.addEventHandler MouseEvent/MOUSE_RELEASED (ui/event-handler event (handle-mouse-released! view-node event)))
       (.addEventHandler MouseEvent/MOUSE_EXITED (ui/event-handler event (handle-mouse-exited! view-node event)))
       (.addEventHandler ScrollEvent/SCROLL (ui/event-handler event (handle-scroll! view-node event))))
 
+    (when editable
+      (doto canvas
+        (setup-input-method-requests! view-node)
+        (.addEventHandler KeyEvent/KEY_TYPED (wrap-disallow-diaeresis-after-tilde
+                                               (ui/event-handler event (handle-key-typed! view-node event))))))
+
```
matgis, line 3577: Unless we're in editable mode, we don't attach the key typing event handlers at all.

editor/src/clj/editor/code/view.clj
```diff
@@ -3688,10 +3645,134 @@
     ;; Start repaint timer.
     (ui/timer-start! repainter)
     ;; Initial draw
-    (ui/run-later (repaint-view! view-node 0 {:cursor-visible true})
+    (ui/run-later (repaint-view! view-node 0 {:cursor-visible true :editable editable})
                   (ui/run-later (slog/smoke-log "code-view-visible")))
     view-node))
 
+(def ^:private fundamental-read-only-handlers
```
matgis, line 3652: This is the "read-only" `defhandlers` shared with the Console view from earlier expressed as data. Both the code editor and the Console view now call the `register-fundamental-read-only-handlers!` function to register these handlers with the respective views.

editor/src/clj/editor/code/view.clj
```diff
@@ -3701,9 +3782,18 @@
...
+      :label "Code"
+      :make-view-fn make-view!
+      :focus-fn focus-view!
+      :text-selection-fn non-empty-single-selection-text)
+    (workspace/register-view-type workspace
+      :id :text
+      :label "Text"
+      :make-view-fn make-view!
+      :focus-fn focus-view!
+      :text-selection-fn non-empty-single-selection-text)))
```
matgis, line 3797: We now register both the `:code` and the `:text` view types to use the `CodeEditorView`.

editor/src/clj/editor/collection.clj
```diff
@@ -682,15 +682,15 @@
...
-                                  [:resource-property-build-targets :resource-property-build-targets]])
+      (gu/connect-existing-outputs node-type resource-node go-node
+        [[:_node-id :source-id]
+         [:resource :source-resource]
+         [:node-outline :source-outline]
+         [:proto-msg :proto-msg]
+         [:build-targets :source-build-targets]
+         [:scene :scene]
+         [:ddf-component-properties :ddf-component-properties]
+         [:resource-property-build-targets :resource-property-build-targets]])
```
matgis, line 693: Call moved function and updated indentation.

editor/src/clj/editor/console.clj
```diff
@@ -336,87 +337,7 @@
...
-
-(handler/defhandler :select-next-occurrence :console-view
-  (run [view-node] (view/select-next-occurrence! view-node)))
-
-(handler/defhandler :select-next-occurrence :console-tool-bar
-  (run [view-node] (view/select-next-occurrence! view-node)))
-
-(handler/defhandler :split-selection-into-lines :console-view
-  (run [view-node] (view/split-selection-into-lines! view-node)))
+(view/register-fundamental-read-only-handlers! *ns* :console-view :console-tool-bar)
```
matgis, line 340: We now call this shared function to register the "read-only" code view handlers with the Console view instead of duplicating the `defhandler` clauses across both files.

editor/src/clj/editor/defold_project.clj
```diff
@@ -447,12 +447,6 @@
                  (not (Boolean/getBoolean "defold.tests")))
         (log/info :message "Some files were migrated and will be saved in an updated format." :migrated-proj-paths migrated-proj-paths)))))
 
-(defn connect-if-output [src-type src tgt connections]
-  (let [outputs (g/output-labels src-type)]
-    (for [[src-label tgt-label] connections
-          :when (contains? outputs src-label)]
-      (g/connect src src-label tgt tgt-label))))
-
```
matgis, line 455: Moved to `editor.graph-util/connect-existing-outputs`. I didn't like the original name since it implied a single output.

editor/src/clj/editor/game_object.clj
```diff
@@ -521,14 +521,14 @@
...
-                                  [:scene :scene]
-                                  [:build-targets :source-build-targets]])
+      (gu/connect-existing-outputs node-type resource-node comp-node
+        [[:_node-id :embedded-resource-id]
+         [:resource :source-resource]
+         [:_properties :source-properties]
+         [:node-outline :source-outline]
+         [:save-value :source-save-value]
+         [:scene :scene]
+         [:build-targets :source-build-targets]])
```
matgis, line 531: Call moved function and update indentation.

editor/src/clj/editor/resource_node.clj
```diff
@@ -42,7 +44,16 @@
...
+(defn save-content
+  ^String [save-value resource-type]
+  (when (some? save-value)
+    (when-let [write-fn (:write-fn resource-type)]
+      (write-fn save-value))))
+
+(defn save-data-content
+  ^String [{:keys [resource save-value]}]
+  (let [resource-type (resource/resource-type resource)]
+    (save-content save-value resource-type)))
```
matgis, line 56: The `save-data-content` function already existed, but I broke out the `save-content` function from it and moved both up in the file, since I needed to call `save-data-content` from `produce-lines` just below this.

editor/src/clj/editor/resource_node.clj
```diff
@@ -104,6 +115,15 @@
...
 
+(g/defnk produce-lines [_node-id resource save-value]
+  (if (nil? save-value)
+    (resource-io/with-error-translation resource _node-id :lines
+      (let [content (slurp resource)]
+        (code.util/split-lines content)))
+    (let [resource-type (resource/resource-type resource)
+          content (save-content save-value resource-type)]
+      (code.util/split-lines content))))
+
```
matgis, line 126: For `ResourceNodes` that output a `save-value`, we write it to a string and split it into lines so it can be displayed using the `CodeEditorView`. This means the lines will refresh whenever the resource is edited.

For `ResourceNodes` that don't output a `save-value`, we read the lines from the resource on disk (or in a zip). This might produce an `ErrorValue` in case the `resource` refers to a non-existing or unreadable resource. If the resource is modified due to external changes to the project, our `resource` argument dependency will be invalidated, and the view will refresh as a result.

editor/src/clj/editor/resource_node.clj
```diff
@@ -224,15 +245,6 @@
...
 
-(defn save-data-content
-  ^String [save-data]
-  (when-some [save-value (:save-value save-data)]
-    (let [resource (:resource save-data)
-          resource-type (resource/resource-type resource)
-          write-fn (:write-fn resource-type)]
-      (when write-fn
-        (write-fn save-value)))))
-
```
matgis, line 235: Moved up in the file.

editor/styling/stylesheets/modules/_console.scss
```diff
@@ -233,11 +233,11 @@
...
 
   &:focused {
-    -fx-background-color: -df-background-input;
-    -fx-border-color: -df-component-dark;
+    -fx-background-color: -df-background-light;
+    -fx-border-color: -df-background-lighter;
     -fx-border-width: 1px 0px 0px 0px;
     -fx-padding: 0px;
-    -fx-text-fill: -df-text-light;
+    -fx-text-fill: -df-text;
```
matgis, line 240: Unrelated: Fixed a CSS regression visible when focusing on the **Evaluate Lua** field in the debugger.

AGulev:
Without implementing this one https://github.com/defold/defold/issues/4052 it feels a bit odd when I open something from an external lib. 

It would be nice to make as least some formats editable in the future e.g. https://github.com/defold/defold/issues/8550 or https://github.com/defold/defold/issues/5301 (I understand that it isn't part of this issue)

editor/src/clj/editor/code/view.clj
```diff
@@ -1090,17 +1093,19 @@
...
   (input indent-type r/IndentType)
   (input invalidated-rows r/InvalidatedRows)
-  (input lines r/Lines :substitute [""])
+  (input lines types/Lines :substitute [""])
   (input regions r/Regions)
   (input debugger-execution-locations g/Any)
 
   (output completion-context g/Any :cached produce-completion-context)
+  (output cursor-ranges r/CursorRanges (g/fnk [cursor-ranges fallback-cursor-ranges]
+                                         (or cursor-ranges fallback-cursor-ranges)))
```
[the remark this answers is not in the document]
AGulev: Not sure if you do so in the editor, but does it make sense to add it as a comment to know why this fallback needed

editor/src/clj/editor/code/view.clj
AGulev: how critical it is? is it only about text metrics reading, or properties as well?

editor/src/clj/editor/console.clj
```diff
@@ -683,7 +604,7 @@
                                         clear (assoc :cursor-ranges [data/document-start-cursor-range])
                                         clear (assoc :invalidated-row 0)
                                         clear (data/frame-cursor prev-layout)))))))
-  (view/repaint-view! view-node elapsed-time {:cursor-visible false}))
+  (view/repaint-view! view-node elapsed-time {:cursor-visible false :editable false}))
```
AGulev, line 607: nice, your fixes also fixed an issue when Enter worked in Console view

AGulev (dismissed):
I'm not the best editor, reviewer.
I read through your comments and tested everything that might be related, and it works good.

editor/src/clj/editor/code/view.clj
matgis: Basically, if a resource type that has been registered using the `register-code-resource-type` function specifies `:lazy-loaded true`, its `:load-fn` will not be executed at all when the project is loaded. So it won't have any content at all. You can't even show the lines of the file without loading them. Calling `ensure-loaded!` on it makes sure that it is fully loaded into memory if it hasn't yet been loaded for some other reason.

Lazy-loading is just a thing for `CodeEditorResourceNode`-derived node types, which covers all the code-editable resource types such as scripts, shaders, as well as game-specific JSON, XML, and so on.

We introduced lazy-loading because game projects often have quite a few very large JSON or XML files that is not relevant to the Defold editor, but are used in the game at runtime. For example, many projects have a few massive JSON or XML files that contain localized strings and so on.

matgis:
I think we could merge this change even if you don't feel fully comfortable reviewing the code changes, as long as it works. I can ask Vlad to have a look at the code once he's back from vacation. I don't think it's a scary change, since it is a little-used feature.

AGulev:
yes, I agree

AGulev (approved):

PR #9168, opened 2024-07-04
Better IDEa project + improved single test runner + removed old project files

AGulev:
1. Removed all old project files (eclipse, etc.) that we don't use anymore.
2. Added a better IntelliJ IDEA project in the `.idea` folder to make setup easier.
3. Added resource folders to the IDEA project, so single tests can find them now.
4. For tests, added a `runSingleTest` task where it's possible to manually specify the class you want to run in cases where the test has a very complex setup and needs a `jar` file. In all other cases:
5. Modified how tests work with the file system to make it possible to work not only with Zip (jar in our case) but with folders also to run single tests using the JUnit interface in IDEA.


com.dynamo.cr/com.dynamo.cr.bob/build.gradle
```diff
@@ -48,23 +48,25 @@ gradle.projectsEvaluated {
...
         def protoTaskName = "protoTask_${proto.file.replace('/', '_').replace('.', '_')}" as Object
-
+        tasks.named('compileJava').configure {
+            dependsOn protoTaskName
+        }
         tasks.create(name: protoTaskName, type: Exec) {
             group = 'build'
             inputs.file(protoFilePath)
-            outputs.dir('generated')
+            outputs.dir("${project.bobDir}/generated")
```
AGulev, line 57: Make sure that we use folders independently on work directory (when we run tests for example)

com.dynamo.cr/com.dynamo.cr.bob.test/.project
AGulev, line 1: A lot of old project files removed

engine/modelc/src/java/com/dynamo/bob/pipeline/ModelImporter.java
```diff
@@ -42,8 +43,14 @@ public class ModelImporter {
                 Method addToPaths = clsbob.getMethod("addToPaths", String.class);
                 File lib = (File)getSharedLib.invoke(null, LIBRARY_NAME);
                 System.load(lib.getAbsolutePath());
+            } catch (InvocationTargetException e) {
```
AGulev, line 46: Make it easier to know where exactly it tries to find file

JCash (approved):

PR #9170, opened 2024-07-04
Editor: Ensure correct numeric types in settings-based file formats

### LINKED ISSUE #9169: Exception thrown when editing integer value in `game.project`
Editing or resetting an integer field in a settings-based file format such as `game.project` throws an exception. This regression was introduced by #9081.

**To Reproduce (REQUIRED)**
1. Open `game.project` in the editor.
2. Go to the HTML 5 section.
3. Type a new value into the Heap Size field and press Enter.

The field accepts the new value.

matgis:
This PR ensures we always use 32-bit integer and floating point values for numeric settings inside settings-based file formats. Before this change, it was inconsistent, and editing an integer field would throw an exception due to a type mismatch between the default, read as a 64-bit `long` from an `.edn` file, and the parsed 32-bit `int` produced by the text field. The mismatch caused us to try to replace the `StringConverter` we configured for the default `long` value with a different `StringConverter` configured to handle `int` values for the edited field, which isn't allowed.

The change ensures we use 32-bit numbers throughout all interactions with settings files.

Fixes #9169

### DIFF
--- a/editor/src/clj/editor/settings_core.clj
+++ b/editor/src/clj/editor/settings_core.clj
@@ -132,10 +132,12 @@
   (= raw "1"))
 
 (defmethod parse-setting-value :integer [_ raw]
+  ;; Parsed as 32-bit integer in the engine runtime.
   (Integer/parseInt raw))
 
 (defmethod parse-setting-value :number [_ raw]
-  (Double/parseDouble raw))
+  ;; Parsed as 32-bit float in the engine runtime.
+  (Float/parseFloat raw))
 
 (defmethod parse-setting-value :resource [_ raw]
   raw)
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/settings_core.clj
```diff
@@ -175,39 +177,51 @@
...
+(defn- ensure-type-defaults [meta-info]
+  (update meta-info :settings
+          (fn [settings]
+            (mapv (fn [{:keys [type] :as meta-setting}]
+                    (update meta-setting :default
+                            (fn [default]
+                              (if (some? default)
+                                (sanitize-default default type)
+                                (type-defaults type)))))
+                  settings))))
```
matgis, line 198: Instead of just adding missing type defaults, we now sanitize the defaults to ensure they have the correct numeric types.

editor/src/clj/editor/settings_core.clj
```diff
@@ -175,39 +177,51 @@
...
-                    settings))))
+  (update meta-info :settings
+          (fn [settings]
+            (mapv (fn [meta-setting]
+                    (if (contains? meta-setting :options)
+                      (assoc meta-setting
+                        :from-string #(parse-setting-value meta-setting %)
+                        :to-string #(render-raw-setting-value meta-setting %))
+                      meta-setting))
+                  settings))))
```
matgis, line 211: Indentation change due to cleanup:  Use `update` and `mapv` instead of `update-in` and `map`.

editor/src/clj/editor/settings_core.clj
```diff
@@ -175,39 +177,51 @@
...
-                          (dissoc meta-setting :from-string :to-string))
-                        meta-setting))
-                    settings))))
+  (update meta-info :settings
+          (fn [settings]
+            (mapv (fn [meta-setting]
+                    (if (contains? meta-setting :options)
+                      (dissoc meta-setting :from-string :to-string)
+                      meta-setting))
+                  settings))))
```
matgis, line 220: Indentation change due to cleanup:  Use `update` and `mapv` instead of `update-in` and `map`.

britzl (approved):

PR #9175, opened 2024-07-05
Better ux for tilemap editor

AGulev:
I spited logic `op` from `cursor-mode` to be more flexible. It's still not perfect, but it is preparation work to make it great when we have Alt/Ctrl/Shirt events.

Also, I added color differentiation between mods to make sure selection mode is differed from paint mode

### DIFF
--- a/editor/src/clj/editor/tile_map.clj
+++ b/editor/src/clj/editor/tile_map.clj
@@ -179,6 +179,8 @@
 
 (def erase-brush (make-brush nil))
 
+(def ^:private empty-brush {:width 1 :height 1 :tiles [{:tile nil :h-flip false :v-flip false :rotate90 false}]})
+
 (defn flip-brush-horizontally
   [brush]
   (let [width (:width brush)
@@ -565,7 +567,6 @@
 (defn- sanitize-tile-map [{:keys [material] :as tile-grid}]
   {:pre [(map? tile-grid)]} ; Tile$TileGrid in map format.
   (cond-> tile-grid
-
           (nil? material)
           (assoc :material default-material-proj-path)))
 
@@ -758,6 +759,7 @@
 (def color-shader (shader/make-shader ::color-shader pos-color-vert pos-color-frag))
 
 (def ^:private white-color (double-array (map #(/ % 255.0) [255 255 255])))
+(def ^:private blue-color (double-array (map #(/ % 255.0) [0 191 255]  )))
 (def ^:private red-color (double-array (map #(/ % 255.0) [255 0 0])))
 (def ^:private orange-color (double-array (map #(/ % 255.0) [255 140 0])))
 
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/tile_map.clj
```diff
@@ -758,6 +759,7 @@
 (def color-shader (shader/make-shader ::color-shader pos-color-vert pos-color-frag))
 
 (def ^:private white-color (double-array (map #(/ % 255.0) [255 255 255])))
+(def ^:private blue-color (double-array (map #(/ % 255.0) [0 191 255]  )))
```
matgis, line 762: Fix whitespace.
```suggestion
(def ^:private blue-color (double-array (map #(/ % 255.0) [0 191 255])))
```

editor/src/clj/editor/tile_map.clj
```diff
@@ -179,6 +179,8 @@
 
 (def erase-brush (make-brush nil))
 
+(def ^:private empty-brush {:width 1 :height 1 :tiles [{:tile nil :h-flip false :v-flip false :rotate90 false}]})
+
```
matgis, line 183: It looks like we could also use `(make-brush nil)` here? 🤔

editor/src/clj/editor/tile_map.clj
```diff
@@ -861,7 +864,7 @@
 
 (def ^:private clamp-palette-mouse-offset (geom/clamper -0.5 0.5))
 
-(def ^:private select-ops #{:select :select-cut :select-erase})
+(def ^:private select-mods #{:select-mode :cut-mode :erase-mode})
```
matgis, line 867: Typo: `select-mods` -> `select-modes`.

matgis (changes requested):
Seems to work great! Having the cursor update on mouse move after I hold a modifier key is a great addition.

I found a typo, "select-mods", I'd like us to fix before merging, though!

editor/src/clj/editor/tile_map.clj
AGulev: I mentioned it in comment above:
>we use it in render, that's why I don't just use make brush to avoid new brush creation every frame

but maybe I don't understand right how it works. isn't this way just a single table which exist all the time?

matgis: `def` happens only once, when the file is loaded. Its similar to having
```
local empty_brush = make_brush(nil)
```
Outside of any of the functions in Lua.

AGulev: Then I misunderstood you
I thought you wanted me to use `(make-brush nil)` everywhere where I use empty-brush. I'll fix it. Thank you

matgis (approved):
LGTM! ❤️

PR #9176, opened 2024-07-05
Fix various issues related to Gui size overrides in the editor

### LINKED ISSUE #7919: Reset template size cause error
### Expected behaviour
Reset template over writes should set it to default

### Actual behaviour
error message

### Steps to reproduce

<hr/>

<table>
<tr><td>Build time</td><td>2023-08-07T11:36:43.216595</td></tr>
<tr><td>Defold editor sha</td><td>34db77fa587e44fb60bc0da0192075054def3f72</td></tr>
<tr><td>Defold engine sha</td><td>504de7800fa81847bfc2e26a21973899db9dd747</td></tr>
<tr><td>Defold version</td><td>1.4.8</td></tr>
<tr><td>Error</td><td><a href='https://sentry.io/organizations/defold/issues/?query=id%3A"890df9655a494aa6b5b8d2c72b42dac6"'>890df9655a494aa6b5b8d2c72b42dac6</a></td></tr>
<tr><td>GPU</td><td>AMD Radeon RX 7900 XT</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 Compatibility Profile Context 23.4.2.230412</td></tr>
<table>

### LINKED ISSUE #8546: Clicking reset button on changed gui template properties makes editor blow up
### Expected behaviour
Changed gui property (e.g. size) of a template node resets to the template defaults.

### Actual behaviour
Spits three errors. Requires editor restart.

### Steps to reproduce
1. New mobile project
2. Create main.gui and template.gui
3. Open template.gui and add box. Set size to manual (64x64)
4. Open main.gui, add template and set the Template to template.gui
5. Edit template/box size to something else from 64x64 and click away to make sure the change is applied
6. Click the reset button where you changed the values -> Error

<hr/>

<table>
<tr><td>Build time</td><td>2024-02-15T08:54:48.112358</td></tr>
<tr><td>Defold editor sha</td><td>7bafadda047312e68a7094c2928e5a403ebf90d8</td></tr>
<tr><td>Defold engine sha</td><td>fcec9500fd438e365d4e4f6cec24b99afb907247</td></tr>
<tr><td>Defold version</td><td>1.6.4</td></tr>
<tr><td>Error</td><td><a href='https://sentry.io/organizations/defold/issues/?query=id%3A"a850d8d521b24561b1a288d2bbf773d9"'>a850d8d521b24561b1a288d2bbf773d9</a></td></tr>
<tr><td>GPU</td><td>NVIDIA GeForce GTX 1070/PCIe/SSE2</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 NVIDIA 531.79</td></tr>
<table>

### LINKED ISSUE #8904: Gui with layouts error
### Expected behaviour
Clicking revert button should display default values
### Actual behaviour
Clicking revert button shows popup error.
### Steps to reproduce
<gui error
/tr>
<hr/>

<table>
<tr><td>Build time</td><td>2024-05-02T15:10:02.194956</td></tr>
<tr><td>Defold editor sha</td><td>acf09db558d18f7ccbbaa02c09333932f7cc8d4a</td></tr>
<tr><td>Defold engine sha</td><td>9141d9d3605e3f5d51c71293116d769da2613d39</td></tr>
<tr><td>Defold version</td><td>1.8.0</td></tr>
<tr><td>GPU</td><td>Intel(R) HD Graphics 520</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 - Build 27.20.100.8681</td></tr>
<tr><td>Java version</td><td>17.0.5+8</td>

<table>

### LINKED ISSUE #9152: Press the restart button on the size property of a template in the gui properties panel.
### Steps to reproduce

<hr/>

<table>
<tr><td>Build time</td><td>2024-06-25T16:29:23.653181</td></tr>
<tr><td>Defold editor sha</td><td>3642fa5b700b8d6e74e3f0595e75cd50eff54459</td></tr>
<tr><td>Defold engine sha</td><td>d6882f432beca85d460ec42497888157c356d058</td></tr>
<tr><td>Defold version</td><td>1.9.0</td></tr>
<tr><td>Error</td><td><a href='https://sentry.io/organizations/defold/issues/?query=id%3A"4aaf4d6b533e455caae3e50822c5e7a9"'>4aaf4d6b533e455caae3e50822c5e7a9</a></td></tr>
<tr><td>OS version</td><td>14.5</td></tr>
<table>

matgis:
* Fixed exception when clearing a manual Size override while Size Mode overrides Auto to Manual.
* Clear out hidden Gui node manual Size overrides when the Size property is controlled by the assigned Texture.

Fixes #7919
Fixes #8546
Fixes #8904
Fixes #9152

### Technical changes
Making an edit to a Gui node that puts the user-facing Size property in control of the assigned `texture` will now clear any override value from the `manual-size` property before hiding it and showing the read-only `texture-size` property instead.

Previously it was possible for an override node to have `manual-size` overridden but not visible (or clearable) if Auto Size Mode was active.

### DIFF
--- a/editor/src/clj/editor/gui.clj
+++ b/editor/src/clj/editor/gui.clj
@@ -1019,26 +1019,50 @@
                :clipping-inverted clipping-inverted))
       (fixup-shape-base-node-size-overrides)))
 
+(defn- visible-size-property-label [size-mode texture]
+  (if (and (= :size-mode-auto size-mode)
+           (coll/not-empty texture))
+    :texture-size
+    :manual-size))
+
 (g/defnode ShapeNode
   (inherits VisualNode)
 
-  (property manual-size types/Vec3 (default (protobuf/vector4->vector3 (protobuf/default Gui$NodeDesc :size))))
+  (property manual-size types/Vec3 (default (protobuf/vector4->vector3 (protobuf/default Gui$NodeDesc :size)))
+            (dynamic label (g/constantly "Size"))
+            (dynamic visible (g/fnk [size-mode texture]
+                               (= :manual-size (visible-size-property-label size-mode texture)))))
   (property texture-size types/Vec3 ; Just for presentation.
             (value (g/fnk [anim-data]
                      (if (some? anim-data)
                        [(float (:width anim-data)) (float (:height anim-data)) protobuf/float-zero]
                        protobuf/vector3-zero)))
+            (dynamic label (g/constantly "Size"))
             (dynamic read-only? (g/constantly true))
-            (dynamic visible (g/fnk [texture]
-                               (coll/not-empty texture))))
+            (dynamic visible (g/fnk [size-mode texture]
+                               (= :texture-size (visible-size-property-label size-mode texture)))))
   (property size-mode g/Keyword (default (protobuf/default Gui$NodeDesc :size-mode))
-            (set (fn [evaluation-context self old-value new-value]
-                   ;; Use the texture size for the :manual-size when the user switches
-                   ;; from :size-mode-auto to :size-mode-manual.
-                   (when (and (= :size-mode-auto old-value)
-                              (= :size-mode-manual new-value)
-                              (properties/user-edit? self :size-mode evaluation-context))
-                     (when-some [texture (not-empty (g/node-value self :texture evaluation-context))]
+            (set (fn [{:keys [basis] :as evaluation-context} self old-value new-value]
+                   (when-some [texture (coll/not-empty (g/node-value self :texture evaluation-context))]
+                     (cond
+                       ;; Clear existing :manual-size override when switching
+                       ;; from :size-mode-manual to :size-mode-auto with a
+                       ;; texture assigned.
+                       (and (= :manual-size (visible-size-property-label old-value texture))
+                            (g/property-overridden? basis self :manual-size)
+                            (let [effective-new-value
+                                  (or new-value
+                                      (let [original-node-id (g/override-original basis self)]
+                                        (g/node-value original-node-id :size-mode evaluation-context)))]
+                              (= :texture-size (visible-size-property-label effective-new-value texture))))
+                       (g/clear-property self :manual-size)
+
+                       ;; Use the size of the assigned texture as :manual-size
+                       ;; when the user switches from :size-mode-auto to
+                       ;; :size-mode-manual.
+                       (and (= :size-mode-auto old-value)
+                            (= :size-mode-manual new-value)
+                            (properties/user-edit? self :size-mode evaluation-context))
                        (when-some [texture-infos (not-empty (g/node-value self :texture-infos evaluation-context))]
                          (when-some [anim-data (:anim-data (texture-infos texture))]
                            (let [texture-size [(float (:width anim-data)) (float (:height anim-data)) protobuf/float-zero]]
@@ -1050,6 +1074,18 @@
                              (or (validate-material-resource _node-id material-infos material)
                                  (validate-material-capabilities _node-id material-infos material material-shader texture-infos texture)))))
   (property texture g/Str (default (protobuf/default Gui$NodeDesc :texture))
+            (set (fn [{:keys [basis] :as evaluation-context} self old-value new-value]
+                   ;; Clear any existing :manual-size override when assigning a
+                   ;; texture will hide the :manual-size property.
+                   (let [size-mode (g/node-value self :size-mode evaluation-context)]
+                     (when (and (= :manual-size (visible-size-property-label size-mode old-value))
+                                (g/property-overridden? basis self :manual-size)
+                                (let [effective-new-value
+                                      (or new-value
+                                          (let [original-node-id (g/override-original basis self)]
+                                            (g/node-value original-node-id :texture evaluation-context)))]
+                                  (= :texture-size (visible-size-property-label size-mode effective-new-value))))
+                       (g/clear-property self :manual-size)))))
             (dynamic edit-type (g/fnk [texture-infos] (optional-gui-resource-choicebox (keys texture-infos))))
             (dynamic error (g/fnk [_node-id texture-infos texture]
                              (validate-texture-resource _node-id texture-infos texture))))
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/gui.clj
```diff
@@ -1020,16 +1022,15 @@
...
-            (dynamic read-only? (g/fnk [size-mode] (= :size-mode-auto size-mode))))
+  (property manual-size types/Vec3 (default (protobuf/vector4->vector3 (protobuf/default Gui$NodeDesc :size))))
+  (property texture-size types/Vec3 ; Just for presentation.
+            (value (g/fnk [anim-data]
+                     (if (some? anim-data)
+                       [(float (:width anim-data)) (float (:height anim-data)) protobuf/float-zero]
+                       protobuf/vector3-zero)))
+            (dynamic read-only? (g/constantly true))
+            (dynamic visible (g/fnk [texture]
+                               (coll/not-empty texture))))
```
matgis, line 1033: Now we always show the Size property, and we label it "Manual Size". If there is a texture assigned, we also show the "Texture Size" property next to it. Even if a texture is applied, the user can edit "Manual Size" if they desire. This makes it more clear what is going on when overriding these properties.

editor/src/clj/editor/gui.clj
```diff
@@ -1040,7 +1041,7 @@
                      (when-some [texture (not-empty (g/node-value self :texture evaluation-context))]
                        (when-some [texture-infos (not-empty (g/node-value self :texture-infos evaluation-context))]
                          (when-some [anim-data (:anim-data (texture-infos texture))]
-                           (let [texture-size [(double (:width anim-data)) (double (:height anim-data)) 0.0]]
+                           (let [texture-size [(float (:width anim-data)) (float (:height anim-data)) protobuf/float-zero]]
```
matgis, line 1044: These values are floats in the Protobuf file format.

editor/src/clj/editor/gui.clj
```diff
@@ -1067,9 +1068,11 @@
...
                                          (or (texture-gpu-textures texture)
                                              (texture-gpu-textures ""))))
-  (output texture-size g/Any (g/fnk [anim-data]
-                                    (when (some? anim-data)
-                                      [(double (:width anim-data)) (double (:height anim-data)) 0.0])))
+  (output size types/Vec3 (g/fnk [manual-size size-mode texture texture-size]
+                            (if (or (= :size-mode-manual size-mode)
+                                    (coll/empty? texture))
+                              manual-size
+                              texture-size)))
```
matgis, line 1075: The previous `texture-size` output is now a read-only property so that it shows in the Property Editor.

Meanwhile, we've made the `manual-size` property that was previously hidden visible, and converted the `size` property (that used to represent both manual and texture size) into an output. This makes it more clear what is going on when the size-related properties are overridden.

AGulev:
Changing to size mode manual making manual size == texture size, but in UI manual size stay the same (even if I change node size to manual it stays the same there):

https://github.com/defold/defold/assets/2209596/06fe5d18-d718-46e8-b45d-a7e6dcf6a821

But when I change it back to manual, it resets:

https://github.com/defold/defold/assets/2209596/8dfa6676-6063-48fc-9da7-140935f0a89d

Can we reset it when I change it to auto (which will make it clear, that manual size is gone now, and it's not achievable neither editor nor runtime anymore)

AGulev:
here is prove that this doesn't exist in runtime anymore as well (at least it's can't be used there): 

https://github.com/defold/defold/assets/2209596/d14d213f-6725-4be2-a9fd-9123c10f9a4f

editor/src/clj/editor/gui.clj
AGulev: Taking into account that in runtime node has `size` this change makes it less clear.

I read code, and the whole setup looks tricky - yes =/

AGulev (dismissed):
I think it's good for now, but maybe it would be better to 
- rename manual size into size
- if `Size Mode: Auto` - make it impossible to edit `Size` (Manual size now) and reset value to texture size
- if `Size Mode: Manual`  - make it editable again

matgis:
If we set the value of Manual Size when you switch Size Mode to Auto, it will override the Manual Size property on override nodes. That's undesirable for Auto Size Mode, so that is why we set it when you switch Size Mode to Manual instead.

AGulev:
>If we set the value of Manual Size when you switch Size Mode to Auto, it will override the Manual Size property on override nodes. That's undesirable for Auto Size Mode, so that is why we set it when you switch Size Mode to Manual instead.

🤯 oh gosh. it's even trickier

AGulev:
I played with it a little bit more, now I understand why we don't wanna make it read only (to be able to use it in the override nodes). 
What if we make a bit stupid but simple thing:
CleanShot 2024-07-05 at 22 49 55@2x

and then

https://github.com/defold/defold/assets/2209596/9305204f-5931-4580-b379-dc896a50d4f1

matgis:
Poking at it some more, I think I have a pretty good solution to the UX issues you raised. We now display a single Size property to the user in all situations. When you make a change to an override node that puts the assigned Texture in control of the Size, we clear out any manual Size overrides that may be present.

There are a few situations where this can lead to somewhat surprising results with overrides, but they have a consistent internal logic to them. Overall, I think it works pretty well!

AGulev (approved):
I works great, but some strange issues appeared (maybe because you hide property)

https://github.com/defold/defold/assets/2209596/7500858f-c3cf-40b0-a9db-7c6a1fa9015f

AGulev:
It's not critical at all. Just one more click needed. We can fix it later

matgis:
Thats a strange issue. I didn't notice it at first, but yeah, the tool bar disappears, and it does indeed seem to be related to how we refresh the property view. It looks like we lose input focus somehow. Which is strange, because we do a similar thing where we hide certain properties for `.font` resources when you change between Bitmap and Distance Field font type, and I don't see the issue there.

Anyway, I agree it is minor and we should look into it separately later. I'll merge this for now. Thanks! ❤️

Issue #6519, opened 2022-04-12
Option to set GLSL ES precision for the whole shader


aglitchman:
**Is your feature request related to a problem? Please describe (REQUIRED):**
The GLSL ES spec does not define a default float precision for fragment shaders, so Defold have to specify it, and now it is `mediump` by default.

If you have a large fragment shader and want to use `highp` everywhere by default, then you have to bloat shader's source code by adding `highp` prefix everywhere.

**Describe the solution you'd like (REQUIRED):**
Defold's shader compiler should check for the line `precision <PRECISIONHERE> float;` in the shader's code to not add the line with the precision again, i.e. Defold adds `precision mediump float` into all fragment shaders.

**Describe alternatives you've considered (REQUIRED):**
We had to add "highp" at almost every line.

AGulev:
hmm, It seems like we have this logic in the ShaderUtil code. Now I'm not sure if it's outdated issue or bug in this code

AGulev:
ah, ok. it seems like it is bug

Jhonnyg:
With the new pipeline you can specify this in the shader, so I’m not sure this issue is needed any more :)

Issue #6503, opened 2022-04-01
Assigning value to material constants fails if they were optimized away from shader


izokina:
**Describe the bug (REQUIRED)**
Uniforms optimized away from shader are also optimized away from materials - which causes runtime error when assigning values to them.

**To Reproduce (REQUIRED)**
1. Create/copy material and fragment shader
2. Add constant to material and declare it as uniform in fragment shader, but never actually use
3. Use material for some sprite and try assigning value to the corresponding property from code
4. See error in logs

**Expected behavior (REQUIRED)**
Assigning to defined material constant doesn't cause error.
Instead, it's ignored if the uniform got optimised away and I'm getting a warning in code.

**Defold version (REQUIRED):**
  - Version 1.3.0

**Platforms (REQUIRED):**
 - Platforms: Windows
 - OS: Windows 11
 - Device: ASUS ROG Zephyrus ZX701ZX

**Logs (OPTIONAL):**
```
ERROR:SCRIPT: bad argument #1 to '?' (string expected, got userdata)
stack traceback:
  [C]:-1: ?
  [C]:-1: in function set
  game/root.script:7: in function spawn_line
  game/root.script:11: in function <game/root.script:10>
```

**Workaround (OPTIONAL):**
I tried using mix with zero value - didn't work out. Currently works with mixing final color with 0.001 of a uniform in question.

**Additional context (OPTIONAL):**
The behavior is somewhat reasonable - you don't want to have uniforms you don't use.
Though there are use cases when you actually need such scenarios, during development especially, so issuing a warning seems to be a much better solution.

PR #9185, opened 2024-07-08
Refactor shader pipeline to use SPIR-v as base

### LINKED ISSUE #6896: Validate shaders and show all errors and warnings in the build log.
Now it's pretty easy to make mistake in a shader which not a problem on Mac/PC but affects mobile/html5 build.
for example using `1` instead of `1.0` for floats.

That would be nice if we validate it when build and show all error and even warning related to shaders

### LINKED ISSUE #9090: Refactor shader pipeline to better utilize SPIR-v toolchain
**Task (REQUIRED):**
Currently the shader pipeline takes GLSL120 compatible shaders as input and performs different passes on them depending on platform and/or graphics API. The current setup has multiple issue:

* It's hard to understand the flow. The pipeline is used for both bob and editor, but with slightly different entry points. The code is spread out and based on a _very_ old structure that shouldn't be used any more
* There's too much magic / brittle regexp for the parsing. We should use SPIR-V for all cross compilation, even for ES2 and WebGL 1 platforms
* We cannot utilize "modern" concepts like uniform buffers. The regexp parsing puts each uniform in their own buffers
* We don't really do any shader validation.
* The engine protobuf files are unoptimized. All shaders have their own reflection structures, but OpenGL only generates reflection data in runtime, which shouldn't be necessary if we produce output shaders from spir-v.
* The NULL adapter does it's own reflection, which means we are not properly testing the pipeline in runtime. It should use the same reflection data as the other adapters. It "works", but it's a maintenance burden.

(There are more issues with the current setup, but this should be enough to warrant the refactor)

The NEW pipeline could work something like this:

* Input is GLSL version 140+ compatible shaders. Old shaders can use the old pipeline, but we should output warnings and encourage updating shaders.
* Run include preprocessor that pastes the includes into the input
* Generate SPIR-V for the shaders for all platforms
* Run an optimization pass on the spv file
* Generate reflection data
* Generate the output shaders based on language

### LINKED ISSUE #9132: Shader input ordering issue when building with bob
In the editor, the build works fine, but when building with bob, it fails.

Setup:
* Tested with bob 1.9.0
* Test project https://github.com/Dragosha/defold-things/tree/master (commit: 501f934acc773fe3f2ed2e0788a1c18bd80001b0)
* Use Vulkan backend

Cmd
```
java -jar ../../../../bob_1.9.0.jar --platform=x86_64-macos --architectures=x86_64-macos distclean build --variant=debug
```

Error message:
```
The build failed for the following reasons:
ERROR assets/materials/shadow_debug.material Location mismatch for fragment shader input 'var_texcoord0': The vertex shader specifies the input at location 3, and location 0 in the fragment shader.
```

### LINKED ISSUE #9155: Enforce safe syntax for floating point numbers in shader programs
**Is your feature request related to a problem? Please describe:**
It's possible to declare floats without periods in shader programs. It will work on PC but the game will crash on launch on html5 and Androind (and may be on other platforms also).

image

**Describe the solution you'd like:**
Defold enforces safe syntax with periods.

**Describe alternatives you've considered:**
Defold prints a warning to the logs during building/bundling.

Jhonnyg:
The shader support in Defold has gotten a major update. Defold outputs engine shaders in a pipelined manner, and currently utilises two slightly different setups: the _new_ and the _legacy_ pipelines. The legacy pipeline produces engine shaders the same way as before:

* Take ES2 compatible GLSL code as input
* Transform the shader source into ES3 / OpenGL 3 variants via mostly regexp
* Optionally take the GL3 code and cross-compile into SPIR-v for vulkan

This setup has multiple issues:
* The engine is using ancient GLSL code. Most modern sources of information has moved on, and so we should too
* Each uniform is packed into their own uniform buffer, which is very inefficient
* The ES2 -> ES3 transformation is very brittle due to the nature of regular expressions
* The GLSL code is unoptimized - there are basically no available tools that operate on old GLSL code

The new pipeline uses the same tools, but in a different order:
* Take shaders written in at least GLSL version 140 as input
* Generate SPIR-v via glslang
* Run optimisation pass on the SPIR-v binary
* Generate reflection data from the optimised shader, which will now be used by the engine for all adapters
* At this point, we can generate shaders in whatever language and version we want via spirv-cross - including GLSL for webgl 1 / ES2, HLSL, metal and whatever else we need.

The produced shaders from this new pipeline has multiple benefits:
* Produces optimised shaders, even for old GLSL variants
* The engine can now use constant buffers. This means that you can pack all your uniforms into a single buffer, which can be updated in one go instead of updating each uniform individually.
* More stable transformation of shader source into the target platform
* The engine code for the graphics adapters is more uniform than before, which helps us to write new functionality

To utilise this new setup, shaders must be written in at least version 140 (GL3/ES3+ compatible) since that is what the spirv toolchain requires. Note that it is not _required_ to rewrite your shaders in order to get the game to still work, but in order to use the new functionality you need to convert your shaders into the newer GLSL. There is a Defold specific manual entry here (<TODO: add link later!>) that you can consult to leverage the new pipeline. In the future, all builtin shaders will be upgraded into #version 140 shaders, but for now this must be done manually.

Fixes #9090 
Fixes #9155
Fixes #6896
Fixes #9132

### Technical changes

* We choose the pipeline based on #version pragma, but should we try to run the new pipeline first and use the legacy pipeline if it fails?
* Mac now supports version 330 shaders for Desktop (which is from early 2010 or so, whereas 140 is from mid 2009)
* Null adapter doesn't do reflection in runtime any more, we now utilise reflection data from the shader generation offline. This means that all test shaders that go through Bob must be valid.

Follow-up tasks:
* Migrate builtins to new GLSL version
* Improve error messages in the editor - we generate resources which means that we get errors for files that doesn't exist..


com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ModelBuilderTest.java
```diff
@@ -41,8 +41,11 @@ public void testModelDae() throws Exception {
         addFile("/test.vp", "");
         addFile("/test.fp", "");
 
-        build("/test.vp", "");
-        build("/test.fp", "");
+        StringBuilder srcShader = new StringBuilder();
+        srcShader.append("void main() {}\n");
+
+        build("/test.vp", srcShader.toString());
+        build("/test.fp", srcShader.toString());
```
Jhonnyg, line 48: We have to use valid shaders now when testing

engine/graphics/src/graphics.h
```diff
@@ -64,8 +64,6 @@ namespace dmGraphics
     //       to the users via lua: http://lua-users.org/wiki/NumbersTutorial
     typedef uint64_t HAssetHandle;
 
-    typedef uintptr_t HComputeProgram;
```
Jhonnyg, line 67: Moved to DMSDK, where vertex and fragment typedefs lives

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderCompilers.java
```diff
@@ -19,167 +19,91 @@
...
-
-    public static class NXShaderCompiler implements IShaderCompiler {
-        public ArrayList<ShaderProgramBuilder.ShaderBuildResult> compile(String shaderSource, ES2ToES3Converter.ShaderType shaderType, String resourceOutputPath, String resourceOutput, boolean isDebug, boolean outputSpirv, boolean softFail) throws IOException, CompileExceptionError {
-            ArrayList<ShaderDesc.Language> shaderLanguages = new ArrayList<ShaderDesc.Language>();
-            shaderLanguages.add(ShaderDesc.Language.LANGUAGE_SPIRV);
-            return getBaseShaderBuildResults(resourceOutputPath, shaderSource, shaderType, shaderLanguages.toArray(new ShaderDesc.Language[0]), "", isDebug, softFail);
-        }
+    public static IShaderCompiler getCommonShaderCompiler(Platform platform) {
+        return new CommonShaderCompiler(platform);
     }
```
Jhonnyg, line 108: We can potentially remove this file as well, it's a shared class now instead of separate per platform..

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderProgramBuilder.java
```diff
@@ -74,23 +75,24 @@ public Task<ShaderPreprocessor> create(IResource input) throws IOException, Comp
...
+        String platformString = this.project.getPlatformStrings()[0];
+
         // Include the spir-v flag into the cache key so we can invalidate the output results accordingly
-        String spirvCacheKey = "output_spirv=" + getOutputSpirvFlag();
+        String shaderCacheKey = String.format("output_spirv=%s;platform_key=%s", getOutputSpirvFlag(), platformString);
+
         taskBuilder.addOutput(input.changeExt(params.outExt()));
         taskBuilder.setData(shaderPreprocessor);
-        taskBuilder.addExtraCacheKey(spirvCacheKey);
+        taskBuilder.addExtraCacheKey(shaderCacheKey);
```
Jhonnyg, line 93: This doesn't work as well as I thought it would 😢 I'll leave it here for now anyway

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderProgramBuilder.java
```diff
@@ -227,9 +174,141 @@ public CommandLine GetShaderCommandLineOptions(String[] args) {
...
-    public void BuildShader(String[] args, ES2ToES3Converter.ShaderType shaderType, CommandLine cmd, IShaderCompiler shaderCompiler) throws IOException, CompileExceptionError {
+    public Platform getPlatformFromCommandLine(String[] args) throws CompileExceptionError {
+        CommandLine cmd = getShaderCommandLineOptions(args);
+        String platformName = cmd.getOptionValue("platform", "");
+        Platform platform = Platform.get(platformName);
+        if (platform == null) {
+            throw new CompileExceptionError(String.format("Invalid platform '%s'\n", platformName));
+        }
+        return platform;
+    }
```
Jhonnyg, line 185: I think we should remove this and just pass in the platform as an argument instead..

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/VertexProgramBuilder.java
```diff
@@ -22,17 +22,16 @@
...
-import com.dynamo.bob.pipeline.ShaderUtil.ES2ToES3Converter;
 import com.dynamo.bob.pipeline.IShaderCompiler;
 import com.dynamo.bob.pipeline.ShaderPreprocessor;
 
+import com.dynamo.graphics.proto.Graphics.ShaderDesc;
+
 import org.apache.commons.cli.CommandLine;
 
 @BuilderParams(name = "VertexProgram", inExts = ".vp", outExt = ".vpc")
 public class VertexProgramBuilder extends ShaderProgramBuilder {
```
Jhonnyg, line 33: I don't think we _need_ these specific builds any more, all shaders work exactly the same. The difference is the type, which we can derive from the extension..

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/shader/SPIRVReflector.java
```diff
@@ -0,0 +1,226 @@
...
+
+package com.dynamo.bob.pipeline.shader;
+
+import org.codehaus.jackson.JsonNode;
+import org.codehaus.jackson.map.ObjectMapper;
+
+import java.io.IOException;
+import java.util.*;
+
+public class SPIRVReflector {
```
Jhonnyg, line 23: Just moved this from ShaderUtils into it's own thing..

engine/gamesys/src/gamesys/resources/res_fragment_program.cpp
```diff
@@ -20,13 +20,8 @@ namespace dmGameSystem
 {
     static dmResource::Result AcquireResources(dmGraphics::HContext context, dmResource::HFactory factory, const char* filename, dmGraphics::ShaderDesc* ddf, dmGraphics::HVertexProgram* program)
     {
-        dmGraphics::ShaderDesc::Shader* shader = dmGraphics::GetShaderProgram(context, ddf);
-        if (shader == 0x0)
-        {
-            return dmResource::RESULT_FORMAT_ERROR;
-        }
```
Jhonnyg, line 27: The graphics adapters dig the shaders out from the shaderdesc by themselves now.

engine/gamesys/src/gamesys/test/material/material_constant_mixed.fp
```diff
@@ -6,5 +6,5 @@ uniform lowp mat4 constant_m4_array[2];
 
 void main()
 {
-    gl_FragColor = vec4(1.0);
+    gl_FragColor = vec4(1.0) + constant_v4 + constant_v4_array[7] + constant_m4[1] + constant_m4_array[1][1];
```
Jhonnyg, line 9: We need to use valid shaders now for the tests.

engine/graphics/src/null/graphics_null.cpp
```diff
@@ -29,7 +29,6 @@
 #include "../graphics_native.h"
 #include "../graphics_adapter.h"
 #include "graphics_null_private.h"
-#include "glsl_uniform_parser.h"
```
Jhonnyg, line 32: Not used any longer, we require actual shader reflection for the null adapter as well!

engine/graphics/src/test/test_app_graphics.cpp
```diff
@@ -253,13 +253,123 @@ struct ReadPixelsTest : ITest
...
+void DeleteShaderDesc(dmGraphics::ShaderDesc* desc)
+{
+    if (desc->m_Reflection.m_Inputs.m_Count > 0)
+        free(desc->m_Reflection.m_Inputs.m_Data);
+    if (desc->m_Reflection.m_Textures.m_Count > 0)
+        free(desc->m_Reflection.m_Textures.m_Data);
+    if (desc->m_Reflection.m_Outputs.m_Count > 0)
+        free(desc->m_Reflection.m_Outputs.m_Data);
+    if (desc->m_Shaders.m_Count > 0)
+        free(desc->m_Shaders.m_Data);
```
Jhonnyg, line 364: We should delete storage buffers here as well

engine/render/src/test/test_render_script.cpp
```diff
@@ -50,6 +50,26 @@ namespace
...
+static inline dmGraphics::ShaderDesc MakeDDFShaderDesc(dmGraphics::ShaderDesc::Shader* shader, dmGraphics::ShaderDesc::ShaderType type)
+{
+    dmGraphics::ShaderDesc ddf;
+    memset(&ddf,0,sizeof(ddf));
+    ddf.m_Shaders.m_Data = shader;
+    ddf.m_Shaders.m_Count = 1;
+    ddf.m_ShaderType = type;
+    return ddf;
+}
+
```
Jhonnyg, line 72: We should use the functions in the test_render.h file instead

share/ext/spirv-tools/build_spirv-tools.sh
```diff
@@ -60,24 +60,29 @@ EXE_SUFFIX=
 case $PLATFORM in
     win32|x86_64-win32)
         EXE_SUFFIX=.exe
-        SRC_EXE=./tools/Release/spirv-opt${EXE_SUFFIX}
+        SRC_EXE_SPIRV_OPT=./tools/Release/spirv-opt${EXE_SUFFIX}
+        SRC_EXE_SPIRV_LINK=./tools/Release/spirv-link${EXE_SUFFIX}
```
Jhonnyg, line 64: We should revert this (or actually, I think we can remove spirv-opt)

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderProgramBuilder.java
AGulev: maybe it's better to add comment into code then (just to know it doesn't work or work in unexpected way) to avoid misunderstanding when you think that some functionality is ON when it isn't

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderProgramBuilder.java
AGulev: yes, it looks a bit... hmm. strange

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderUtil.java
```diff
@@ -99,17 +102,105 @@ public static String stripComments(String source)
...
+                } else if (shaderLanguage == ShaderDesc.Language.LANGUAGE_GLSL_SM330) {
+                    version = 330;
+                } else {
+                    version = 140;
+                }
+
+                // Write our directives.
+                if (shaderLanguage == ShaderDesc.Language.LANGUAGE_GLES_SM100) {
+                    // Normally, the ES2ToES3Converter would do this
+                    writer.println("precision mediump float;");
```
AGulev, line 160: hmm, I thought it supposes to check if precision already specified if not then set default one e.g.  `mediump`, no?

AGulev (dismissed):
Looks great! 
A couple of questions, but nothing serious

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderProgramBuilder.java
Jhonnyg: It’s legacy code with some intention not necessary any longer afaik

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderUtil.java
Jhonnyg: This is legacy code, the spirv toolchain does this automatically for us

AGulev:
Could you take a look if you PR relatd of even fixes to the following issues:
https://github.com/defold/defold/issues/9132
https://github.com/defold/defold/issues/6503

engine/graphics/src/opengl/graphics_opengl_private.h
```diff
@@ -84,22 +85,35 @@ namespace dmGraphics
...
     {
         char*            m_Name;
         dmhash_t         m_NameHash;
         HUniformLocation m_Location;
         GLint            m_Count;
         GLenum           m_Type;
-        uint8_t          m_TextureUnit   : 7;
-        uint8_t          m_IsTextureType : 1;
+        uint16_t         m_TextureUnit   : 7;
+        uint16_t         m_IsTextureType : 1;
```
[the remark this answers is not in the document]
Jhonnyg: I added more fields there but then removed them and forgot to revert this change

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -1891,46 +1975,177 @@ static void LogFrameBufferError(GLenum status)
...
+            uniform.m_Location      = uniform_location;
+            uniform.m_Name          = strdup(uniform_name);
+            uniform.m_NameHash      = dmHashString64(uniform_name);
+            uniform.m_Count         = uniform_size;
+            uniform.m_Type          = uniform_type;
+            uniform.m_IsTextureType = IsTypeTextureType(GetGraphicsType(uniform_type));
+
+        #if 0
+            dmLogInfo("Uniform[%d]: %s, %llu", i, uniform.m_Name, uniform.m_Location);
+        #endif
```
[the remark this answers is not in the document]
Jhonnyg: No, I use these for debugging, when something breaks it's 9/10 uniforms :)

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -2473,6 +2485,10 @@ namespace dmGraphics
             VkDescriptorSetLayoutBinding& binding = bindings[res.m_Set][res.m_Binding];
             ProgramResourceBinding& program_resource_binding = program->m_ResourceBindings[res.m_Set][res.m_Binding];
 
+        #if 0
+            dmLogInfo("    name=%s, set=%d, binding=%d", res.m_Name, res.m_Set, res.m_Binding);
+        #endif
```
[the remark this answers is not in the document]
Jhonnyg: No, I use these for debugging!

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2125,20 +2353,26 @@ static void LogFrameBufferError(GLenum status)
         return success;
     }
 
-    static bool OpenGLReloadFragmentProgram(HFragmentProgram prog, ShaderDesc::Shader* ddf)
+    static bool OpenGLReloadFragmentProgram(HFragmentProgram prog, ShaderDesc* ddf)
     {
         assert(prog);
         assert(ddf);
 
+        ShaderDesc::Shader* ddf_shader = GetShaderProgram((HContext) g_Context, ddf);
```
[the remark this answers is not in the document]
Jhonnyg: Added new task: https://github.com/defold/defold/issues/9206
It's not related to this PR

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderUtil.java
Jhonnyg: I'm not sure, this is part of the legacy pipeline so I'm not sure we should do anything about right now (or at least in this PR).

AGulev:
@Jhonnyg btw, what about this one https://github.com/defold/defold/issues/6519 ? will new pipeline work with this?

PR #9189, opened 2024-07-08
Fix issue when the first single value input into a property field right after the property reset may be lost

### LINKED ISSUE #9182: Number and Hash script properties do not accept single character input directly after reset
### Expected behaviour
Registration of ANY valid value given to a Number or Hash `Script Property`
### Actual behaviour
Failure to register any Number or Hash value given directly after a `Script Property` reset if the value given is exactly 1 character 

### Steps to reproduce
**Video of this issue**
https://streamable.com/ia9ucf

**Video of what happens when typing more than 1 character**
https://streamable.com/f7s4vz
<hr/>

<table>
<tr><td>Build time</td><td>2024-07-06T16:52:39.732983</td></tr>
<tr><td>Defold channel</td><td>alpha</td></tr>
<tr><td>Defold editor sha</td><td>a4a62def6f1cf7bb000f331c03c38820db50b8d7</td></tr>
<tr><td>Defold engine sha</td><td>a4a62def6f1cf7bb000f331c03c38820db50b8d7</td></tr>
<tr><td>Defold version</td><td>1.9.1</td></tr>
<tr><td>GPU</td><td>NVIDIA GeForce GTX 1070 Ti/PCIe/SSE2</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 NVIDIA 555.85</td></tr>
<table>

AGulev:
Fix the issue where a property value was reset and immediately replaced with a single input value, which would not be saved after the field is unfocused.

Fix https://github.com/defold/defold/issues/9182

### DIFF
--- a/editor/src/clj/editor/properties_view.clj
+++ b/editor/src/clj/editor/properties_view.clj
@@ -663,7 +663,6 @@
                     (ui/add-styles! ["clear-button" "button-small"])
                     (ui/on-action! (fn [_]
                                      (properties/clear-override! (property-fn key))
-                                     (ui/suppress-auto-commit! control)
                                      (.requestFocus control))))
 
         label-box (let [box (GridPane.)]
--- a/editor/src/clj/editor/ui.clj
+++ b/editor/src/clj/editor/ui.clj
@@ -577,12 +577,7 @@
                                     (when (user-data node ::auto-commit)
                                       (commit-fn nil)))))
   (on-edit! node (fn [_old _new]
-                   (if (user-data node ::suppress-auto-commit)
-                     (user-data! node ::suppress-auto-commit false)
-                     (user-data! node ::auto-commit true)))))
-
-(defn suppress-auto-commit! [^Node node]
-  (user-data! node ::suppress-auto-commit true))
+                   (user-data! node ::auto-commit true))))
 
 (defn- clear-auto-commit! [^Node node]
   ;; Clear the auto-commit flag. You should call this whenever data has been


AGulev:
I found why it was added https://github.com/defold/defold/commit/fbe47b48787963a396ebb718a0b04a3e4e1c9bd8
but I can't reproduce anything like that as well as I see that commit (`update-fn`) never called after reset even without this line. 
Maybe it was changed since then.

editor/src/clj/editor/properties_view.clj
```diff
@@ -663,7 +663,6 @@
                     (ui/add-styles! ["clear-button" "button-small"])
                     (ui/on-action! (fn [_]
                                      (properties/clear-override! (property-fn key))
-                                     (ui/suppress-auto-commit! control)
```
AGulev, line 666: I found why it was added https://github.com/defold/defold/commit/fbe47b48787963a396ebb718a0b04a3e4e1c9bd8
but I can't reproduce anything like that as well as I see that commit (`update-fn`) never called after reset even without this line. 
Maybe it was changed since then.

AGulev:
This issue was introduced in 1.5.0 (1.4.9 was fine)

vlaaad (dismissed):
The change looks good to me! It would be cool if you also removed the `ui/suppress-auto-commit!` function altogether, as well as all code related to it, because Properties view was the only place using it.

AGulev:
@vlaaad good point, I missed that. Done

vlaaad (approved):
Shipit!

PR #9197, opened 2024-07-12
Fixed tilemap brush visibility

AGulev:
Fixed the issue where brush tiles were always behind the tilemap layers.

CleanShot 2024-07-12 at 19 05 00

### DIFF
--- a/editor/src/clj/editor/tile_map.clj
+++ b/editor/src/clj/editor/tile_map.clj
@@ -841,7 +841,7 @@
         vb (vtx/use-with ::brush vbuf tex-shader)
         ^Matrix4d layer-transform (:world-transform renderable)
         local-transform (doto (Matrix4d. geom/Identity4d)
-                          (.set (Vector3d. (* x w) (* y h) 0)))
+                          (.set (Vector3d. (* x w) (* y h) 0.001)))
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/tile_map.clj
```diff
@@ -841,7 +841,7 @@
         vb (vtx/use-with ::brush vbuf tex-shader)
         ^Matrix4d layer-transform (:world-transform renderable)
         local-transform (doto (Matrix4d. geom/Identity4d)
-                          (.set (Vector3d. (* x w) (* y h) 0)))
+                          (.set (Vector3d. (* x w) (* y h) 0.001)))
```
AGulev, line 844: make sure that brush has a bit bigger Z than current layer

editor/src/clj/editor/tile_map.clj
```diff
@@ -1125,7 +1125,7 @@
   [^GL2 gl render-args renderables n]
   (let [pass (:pass render-args)]
     (condp = pass
-      pass/transparent
+      pass/opaque
```
AGulev, line 1128: Use `pass` that writes into depth buffer, so brush Z will be taken into account

vlaaad (approved):
LGTM!

PR #9204, opened 2024-07-18
Frame the camera to the very first visible object added to the scene view

### LINKED ISSUE #8911: Confusing initial view of newly created tilemap
When creating a new tilemap, it has no geometry, so it makes one up.
But this is very much "zoomed in" and is quite confusing to look at and understand at a first glance, especially for newcomers.

Screenshot: The tilesource is set, and has 16x16 tiles.
Screenshot 2024-05-12 at 12 18 55

AGulev:
It may be confusing when a visible object is added to an empty scene, but nothing is visible because the camera is too zoomed in. After this fix, the very first visible object added to a scene will be framed to ensure it is visible.

Fix https://github.com/defold/defold/issues/8911

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


editor/src/clj/editor/atlas.clj
```diff
@@ -637,7 +637,8 @@
   (let [[width height] layout-size
         pages (group-by :page layout-rects)
         child-renderables (into [] (for [[page-index page-rects] pages] (produce-page-renderables aabb width height page-index page-rects gpu-texture)))]
-    {:aabb aabb
+    {:node-id _node-id
```
AGulev, line 640: I don't get why don't we add it... In all the other places we have `:node_id` but not here. I hope it's fine I added it

editor/src/clj/editor/atlas.clj
```diff
@@ -843,8 +844,8 @@
   
   (output texture-set-pb   g/Any               :cached produce-atlas-texture-set-pb)
 
-  (output aabb             AABB                (g/fnk [layout-size]
-                                                 (if (= [0 0] layout-size)
+  (output aabb             AABB                (g/fnk [layout-size layout-rects]
+                                                 (if (or (= [0 0] layout-size) (empty? layout-rects))
```
AGulev, line 848: If there are no images in atlas, we should use one of predefined AABBs (empty or null)

editor/src/clj/editor/geom.clj
```diff
@@ -387,6 +387,17 @@
          (= (.y min-p) (.y max-p))
          (= (.z min-p) (.z max-p)))))
 
+(def minimalXY-bounding-box (coords->aabb [0 0 0] [0.0011 0.0011 0]))
```
AGulev, line 390: a special "filler" AABB fo empty layers in tilemap
And a function that detects such AABB:

editor/src/clj/editor/geom.clj
```diff
@@ -387,6 +387,17 @@
...
+
+(defn minimalXY-aabb? [^AABB aabb]
+  (let [min-p ^Point3d (.min aabb)
+        max-p ^Point3d (.max aabb)]
+    (and (= (.x min-p) (- (.x max-p) 0.0011))
+         (= (.y min-p) (- (.y max-p) 0.0011)))))
+
+(defn predefined-aabb? [^AABB aabb]
+  (or (null-aabb? aabb) (empty-aabb? aabb) (minimalXY-aabb? aabb)))
```
AGulev, line 399: check if it's one of "ours" AABB but not user custom one

editor/src/clj/editor/scene.clj
```diff
@@ -600,6 +600,22 @@
...
+(declare active-scene-view)
+(declare frame-selection)
+
+(defn- reframe [scene scene-aabb active-view]
+  (when (:node-id scene)
+    (let [node-id (:node-id scene)
+          prev-aabb (g/user-data node-id :prev-scene-aabb)
+          reframe? (if prev-aabb
+                     (and (geom/predefined-aabb? prev-aabb)
+                          (not (geom/predefined-aabb? scene-aabb)))
```
AGulev, line 612: here we check if prev. saved AABB one of predefined and new AABB isn't then we reframe camera view on an added object

editor/src/clj/editor/scene.clj
```diff
@@ -1020,13 +1037,17 @@
...
-  (let [aabb (fudge-empty-aabb (g/node-value view :selected-aabb))
-        camera (view->camera view)
-        viewport (g/node-value view :viewport)
-        local-cam (g/node-value camera :local-camera)
-        end-camera (c/camera-frame-aabb local-cam viewport aabb)]
-    (set-camera! camera local-cam end-camera animate?)))
+(defn frame-selection
+  ([view animate?]
+   (let [aabb (fudge-empty-aabb (g/node-value view :selected-aabb))]
+     (frame-selection view animate? aabb)))
```
AGulev, line 1043: I already have AABB and if I try to get `selected-aabb` in my case, It might endup in an infinite loop (in case both `:scene-aabb and :selected-aabb` evicted from cache)

editor/src/clj/editor/tile_map.clj
```diff
@@ -466,8 +467,7 @@
           ;; This makes the brush for yet-to-be-painted tiles visible.
           visibility-aabb
           (if (geom/empty-aabb? aabb)
-            (geom/make-aabb (Point3d. 0.0 0.0 0.0)
-                            (Point3d. 0.001 0.001 0.0))
+            geom/minimalXY-bounding-box
```
AGulev, line 470: Use as one of predefined AABB we can later detect

editor/src/clj/editor/tile_map.clj
```diff
@@ -706,6 +706,13 @@
                 (when (and width height)
                   [width height])))))
 
+  (output scene-aabb AABB :cached
+          (g/fnk [tile-dimensions]
+            (if-not tile-dimensions
+              geom/null-aabb
+              (geom/make-aabb (Point3d. 0.0 0.0 0.0)
+                              (Point3d. (* 16 (first tile-dimensions)) (* 9 (second tile-dimensions)) 0.0)))))
```
AGulev, line 714: Calculate tilemap AABB size as 16x9 tiles field

AGulev:
That's how it works: 

https://github.com/user-attachments/assets/37039c1b-9b01-4ef0-9599-51691fa1a0ad

editor/src/clj/editor/geom.clj
vlaaad: Should we also call it `minimalXY-aabb`, like the rest of the predefined aabbs?

editor/src/clj/editor/scene.clj
```diff
@@ -600,6 +600,22 @@
             union-aabb
             (:children scene))))
 
+(declare active-scene-view)
+(declare frame-selection)
```
vlaaad, line 604: We try to avoid using `declare` if possible, since it might be a symptom of cyclic dependencies in code. Instead, we could move the definitions of the function here, or move the use of these functions lower.

editor/src/clj/editor/scene.clj
```diff
@@ -629,8 +645,9 @@
                                          (if (seq selected-aabbs)
                                            (reduce geom/aabb-union geom/null-aabb selected-aabbs)
                                            scene-aabb))))
-  (output scene-aabb AABB :cached (g/fnk [scene]
+  (output scene-aabb AABB :cached (g/fnk [scene active-view]
                                     (let [scene-aabb (calculate-scene-aabb geom/null-aabb geom/Identity4d scene)]
+                                      (reframe scene scene-aabb active-view)
```
vlaaad, line 650: This is very very bad because the only job of the outputs in the graph is producing data, not performing side effects, especially not mutating the graph. In this case, `reframe` calls `set-camera!` which, in turn, changes the graph with `g/transact`. The act of reading changes what is being read. This may cause infinite loops and stuff.

Instead, any mutation of the graph should happen outside of the graph. After a quick look, I think that possible candidates for such reframing could be `refresh-scene-view!` or `update-image-view!` functions

vlaaad (changes requested):
The improvements introduced in this PR are very useful, but they should be implemented differently, because using `g/node-value` and `g/transact` inside the `scene-aabb` function is extremely dangerous and error-prone: `g/node-value` introduces hidden dependencies, and `g/transact` is a side effect of what is supposed to be a pure computation based on explicit dependencies.

editor/src/clj/editor/scene.clj
vlaaad: Regarding subscribing to changes of a value in a graph, I think this is a very useful feature, and we will definitely need something like that in the future. Maybe with an option to subscribe to multiple outputs in the same subscription. I have a [dev-time implementation](https://github.com/defold/defold/blob/dev/editor/src/reveal/editor/reveal.clj#L167-L184) of such an observable that I sometimes use for debugging. It's useful, and we currently re-implement it ad-hoc in various places in the editor UI.

editor/src/clj/editor/scene.clj
```diff
@@ -1215,7 +1219,19 @@
               (ui/user-data! image-view ::last-renderables renderables)
               (ui/user-data! image-view ::last-frame-version frame-version)
               (scene-cache/prune-context! gl)
-              (reset! async-copy-state-atom (scene-async/finish-image! (scene-async/begin-read! @async-copy-state-atom gl) gl))))))
+              (reset! async-copy-state-atom (scene-async/finish-image! (scene-async/begin-read! @async-copy-state-atom gl) gl))
+              ;; call frame-selection if it's the very first aabb change for the scene
```
AGulev, line 1223: This code moved out from graph here after review. It's not perfect, and will be checked more often than in previous implementation, where we catch the moment of AABB change. But I took a look of the code around, and it seems like that's the way of doing it. Let me know if I'm wrong here

editor/src/clj/editor/tile_map.clj
```diff
@@ -586,11 +586,11 @@
...
-  [_node-id child-scenes]
-  {:node-id  _node-id
-   :aabb geom/null-aabb
-   :renderable {:passes [pass/selection]}
-   :children child-scenes})
+  [_node-id child-scenes scene-aabb]
+  {:node-id           _node-id
+   :aabb              scene-aabb
+   :renderable        {:passes [pass/selection]}
+   :children          child-scenes})
```
AGulev, line 593: the next `produce-node-outline` the same code formatting when create data, so it makes sense to have it here as well

editor/src/clj/editor/geom.clj
AGulev: fixed, thx

editor/src/clj/editor/scene.clj
AGulev: I've removed it, thx

editor/src/clj/editor/scene.clj
AGulev: Fixed! Thank you!

vlaaad (dismissed):
Works great and the code is very good!

editor/src/clj/editor/scene.clj
```diff
@@ -1215,7 +1219,19 @@
...
+              (let [prev-aabb (ui/user-data image-view ::prev-scene-aabb)
+                    scene-aabb (g/node-value view-id :scene-aabb evaluation-context)
+                    active-view (g/node-value view-id :active-view evaluation-context)
+                    reframe? (if prev-aabb
+                               (and (geom/predefined-aabb? prev-aabb)
+                                    (not (geom/predefined-aabb? scene-aabb)))
+                               false)]
```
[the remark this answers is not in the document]
AGulev: true. It happens when change code step by step, it's easy to miss such thing. Thanks.

editor/src/clj/editor/scene.clj
```diff
@@ -1215,7 +1219,19 @@
...
+                (ui/user-data! image-view ::prev-scene-aabb scene-aabb)
+                (when reframe?
+                  (when active-view
+                    (frame-selection active-view true scene-aabb))))))))
```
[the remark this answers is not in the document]
AGulev: fixed!

vlaaad (approved):
Ship it!

PR #9207, opened 2024-07-19
Recreate material when material resource is hot-reloaded

### LINKED ISSUE #3686: Reloading a material with changed fragment program does not work (DEF-2119)
*Description:*
Reloading a material after changing the fragment program path (pointing to a different fp-file) will not use the new fragment shader program. It will render using the old fragment program and when closing the engine an assert is triggered:
{{Assertion failed: (resource_hash), function Release, file ../src/resource.cpp, line 1011.}}.

After some investigation it seems that the {{material->m_FragmentProgram}} is not updated after a material file is reloaded.

*Steps to reproduce:*
1. Set a custom material on a sprite
2. Change the fragment program to a different .fp-file.
3. Reload the material file
4. The engine keeps rendering using the old fragment program
5. Close engine and view the assert in all its glory

*Expected result:*
The new fragment shader should be used when reloading the material.

*Actual result:*
Old fragment shader is used and engine crashes.

### LINKED ISSUE #9206: Hot-reloading shaders two times in a row crashes the engine
As the title says, hot-reloading any shader resource will crash the engine the second time its issued.

**To Reproduce (REQUIRED)**
1. Open (or create) any project
2. Run the game
3. Open any shader that is being used in the game
4. Make modifications and press hot-reload (cmd-r)
5. Do step 4 again - game crashes

  - Version 1.9.1

**Platforms (REQUIRED):**
 - Platforms: macOS (only tested)

Jhonnyg:
Fixed an issue where hot-reloading shaders have no effect. The fix also applies to swapping out shaders in the material resource as well.

Fixes #3686
Fixes #9206

### DIFF
--- a/engine/gamesys/src/gamesys/resources/res_material.cpp
+++ b/engine/gamesys/src/gamesys/resources/res_material.cpp
@@ -222,6 +222,32 @@ namespace dmGameSystem
         }
     }
 
+    static dmRender::HMaterial CreateAndInitializeRenderMaterial(dmResource::HFactory factory, dmRender::HRenderContext render_context, const MaterialResources& resources, const dmRenderDDF::MaterialDesc* ddf)
+    {
+        dmRender::HMaterial material = dmRender::NewMaterial(render_context, resources.m_VertexProgram, resources.m_FragmentProgram);
+        if (!material)
+        {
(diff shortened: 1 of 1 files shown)


AGulev (approved):
LGTM!

PR #9209, opened 2024-07-19
Use red color for the brush in tilemap if all the tiles in brush are nil

AGulev:
We use color scheme to make select/cut/draw/erase actions recognizable. Drawing with no tiles (with null tiles) is erasing. So we should use red color in such case

### DIFF
--- a/editor/src/clj/editor/tile_map.clj
+++ b/editor/src/clj/editor/tile_map.clj
@@ -130,7 +130,9 @@
 
 (defn make-brush
   [tile]
-  {:width 1 :height 1 :tiles [{:tile tile :h-flip false :v-flip false :rotate90 false}]})
+  (if-not tile
+    {:width 1 :height 1 :tiles [nil]}
+    {:width 1 :height 1 :tiles [{:tile tile :h-flip false :v-flip false :rotate90 false}]}))
 
 (defn make-brush-from-selection
   [cell-map start-cell end-cell]
@@ -1178,7 +1180,9 @@
                                            :cut-mode orange-color
                                            :erase-mode red-color
                                            :select-mode blue-color
-                                           white-color)
+                                           (if (every? nil? (:tiles brush))
+                                             red-color
+                                             white-color))
                                   :tile-dimensions tile-dimensions}}])}))
 
 (g/defnk produce-tool-renderables


editor/src/clj/editor/tile_map.clj
```diff
@@ -130,7 +130,9 @@
 
 (defn make-brush
   [tile]
-  {:width 1 :height 1 :tiles [{:tile tile :h-flip false :v-flip false :rotate90 false}]})
+  (if-not tile
+    {:width 1 :height 1 :tiles [nil]}
```
vlaaad, line 134: Are you sure that `nil` is a valid tile? Shouldn't it be `{:tile nil :h-flip false :v-flip false :rotate90 false}` instead (like `empty-brush`)?

AGulev: `empty-brush` uses the same `make-brush` function.

I'll recheck again ofc, but I think yes, it is valid. We just had two ways to represent empty tile:
- one when we `Shift+E` - this is empty tile
- another when we create brush using Shift and drag on an empty part of the field - just nil

In this PR I just unify the way we specify "no-tile" by using `nil` instad of `{:tile nil :h-flip false :v-flip false :rotate90 false}`  in one case and `nil` in another which make it harder to check if the brush is in "deleting" mode now

vlaaad: `empty-brush` uses `(make-brush nil)`, which, in the old implementation, would create a tiles array with a single `{:tile nil :h-flip false :v-flip false :rotate90 false}` hashmap, not a `nil`. I worry that using `nil` instead of a `{:tile nil :h-flip false :v-flip false :rotate90 false}` will work only until suddenly it wouldn't. `nil` and `false` are treated very similar in Clojure, but they are still different values, and they might end up in saved data in an invalid way.

It seems that previously `:tiles` could only contain hashmaps, and now they suddenly can contain nils, while the rest of the code that uses tiles continues to expect them to be maps, and it keeps working only because we use e.g. `(:h-flip tile)` instead of `(tile :h-flip)` and then e.g. `(if hflip 1 0)`. It just happens that we don't save `hflip` to a `.tilemap` file. 

If it's possible, I would just prefer if the `:tiles` data shape remained as `array of hash-maps`, not `array of hash-maps or nils`, because it's easier to reason about.

AGulev: hmmm 
But I don't introduce anything new here really.
If you'll try to create erasing brush using `Shift`+ select a few tiles outside your filed (where you don't have tiles at all) on a version without my changes, you'll get a map of nills. 
`nil` tile isn't a tile with index nil, nil tile is `there is no tile here` which is `nil`, it shouldn't have rotation, of flip. It shouldn't be saves in the `.tilemap` file but it's oposite: it should be fully removed from the tilemap

vlaaad: You are introducing a new data shape. `tiles` "type" was `{:tile int|nil :h-flip boolean :v-flip boolean :rotate90 boolean}`, where specifying the absence of the tile was done with by setting the `:tile` value to `nil`. The new "type" is `{:tile int|nil :h-flip ...} | nil`. 

It happens that Clojure treats `nil` as an empty map when looking up keys, and it happens that Clojure treats `nil` as false inside an `if`, and these are useful properties of the language that made your change work without changing the res of the code. 

My concern is that you introduce a complexity: there are now 2 ways to express the absence of a tile in a tiles array: by using a nil, i.e. `nil`, or by using a hash-map with `:tile` set to `nil`, i.e. `{:tile nil :h-flip false ...}`. I guess that's kinda cool that you can check both cases by looking up the key `:tile` — Clojure lets you do that! — but when I'm coding it I still need to think about the whole data shape.

Alternatively, you could change the data "type" into this: `{:tile int :h-flip ...} | nil` — if there is a map, the `:tile` key *has* to be set to non-null value, but then you need to change other code to actually do the right thing — check if the whole map is `nil`, not just look up the `:tile` key. I'm talking e.g. about this line: https://github.com/defold/defold/pull/9209/files#diff-6240a163144fc0ec117e9c024427e94da7078dd59f7dcdaa052c7474169d3376R111
It should check if `(first tiles)` returns nil before looking up `:tile`, `:h-flip` and other keys.

AGulev: > You are introducing a new data shape. `tiles` "type" was `{:tile int|nil :h-flip boolean :v-flip boolean :rotate90 boolean}`, where specifying the absence of the tile was done with by setting the `:tile` value to `nil`. The new "type" is `{:tile int|nil :h-flip ...} | nil`.

Am I? I'm trying to explain that I'm fixing it, not introducing.

I'm not sure if you understand what I'm saying. Let me try to show, then.
I do not introduce a new type. I just use one type for the same logical concept in two places. And instead of two ways to represent "no-tile" concept, I use only one.

Here is what I'm doing in the editor without my changes:

https://github.com/user-attachments/assets/92a4e72f-65f3-484d-8d18-ae695cbced5a

Here is how this brush looks like:
>{:width 7, :height 7, :tiles [nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil]}

Now I do `Shift + E`:
>{:width 1, :height 1, :tiles [{:tile nil, :h-flip false, :v-flip false, :rotate90 false}]}

I remind you, that this is in the editor **without my changes.** 

After my changes, I make `nil` (the first way) a unified way to represent a "no tile concept".

vlaaad (approved):
LGTM! Ship it!

PR #9218, opened 2024-07-23
Migrate dynamic GUI textures to use resource system

### LINKED ISSUE #9093: Migrate dynamic GUI textures into the resource system
**Task (REQUIRED):**
Dynamic textures (created using gui.new_texture) right now has its own system for managing them. Meaning, they cannot be shared in other parts of the engine, and we need a lot of extra code in the engine just to manage them when we already have support for dynamic texture resources in the regular .script world.

We should deprecate gui.new_texture in favor of using the regular resource system for setting dynamic textures in the GUI.

Jhonnyg:
Dynamic textures (created via `gui.new_texture`) are now managed by the resource system. There is no change in behaviour or function signature for how dynamic textures are managed, the biggest gain of this task is to remove GUI specific code in the engine and move some parts of the GUI system closer to the regular gameobject world, code wise.

Fixes #9093

### DIFF
--- a/engine/gamesys/src/gamesys/gamesys_private.h
+++ b/engine/gamesys/src/gamesys/gamesys_private.h
@@ -17,6 +17,7 @@
 
 #include <dlib/message.h>
 #include <dlib/object_pool.h>
+#include <dlib/buffer.h>
 
 #include <render/render.h>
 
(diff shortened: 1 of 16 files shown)


engine/gamesys/src/gamesys/gamesys_private.h
```diff
@@ -129,6 +130,50 @@ namespace dmGameSystem
     dmGameObject::PropertyResult ClearMaterialAttribute(DynamicAttributePool& pool, uint32_t dynamic_attribute_index, dmhash_t name_hash);
     dmGameObject::PropertyResult SetMaterialAttribute(DynamicAttributePool& pool, uint32_t* dynamic_attribute_index, dmRender::HMaterial material, dmhash_t name_hash, const dmGameObject::PropertyVar& var, CompGetMaterialAttributeCallback callback, void* callback_user_data);
     dmGameObject::PropertyResult GetMaterialAttribute(DynamicAttributePool& pool, uint32_t dynamic_attribute_index, dmRender::HMaterial material, dmhash_t name_hash, dmGameObject::PropertyDesc& out_desc, CompGetMaterialAttributeCallback callback, void* callback_user_data);
+
+    // gamesys_resource.cpp
```
Jhonnyg, line 134: All of these were moved from script_resource.cpp so we can reuse them in other parts of gamesys

AGulev (approved):
LGTM!

PR #9223, opened 2024-07-24
Remove timer callback if timer freed prematurely

### LINKED ISSUE #8700: Timer LuaRef leak
**Describe the bug**
The LuaRef count seen in the profiler does not go down when a game object that started a timer is deleted

**To Reproduce**
Steps to reproduce the behaviour:
1. Open the [Defold examples project](https://github.com/defold/examples/commits/master/)
2. Run the project
3. Start web profiler
4. Observe the LuaRefs count in the profiler Global Properties view
5. Repeatedly select and exit the repeating timer example
6. See how the LuaRef count goes up by 1 each time
7. Stop execution
8. Comment out the code in repeating_timer.gui_script that creates the timer
9. Run again and compare

**Expected behaviour**
The LuaRef count should go up when running the timer example, but return to its previous value when returning to the menu, as the proxy is unloaded and the timer is automatically cancelled/deleted

**Defold version**
Seen in 1.6.4 and 1.7.0, both on M1 Macbook Pro

**Additional context**
See #4590 which might be related.

Jhonnyg:
Fixed an issue where a timer was created with a callback and the timer was removed before it could complete. In this case, the lua reference to the timer callback was never released, causing a reference leak.

Fixes #8700

### DIFF
--- a/engine/script/src/script_timer.cpp
+++ b/engine/script/src/script_timer.cpp
@@ -166,6 +166,12 @@ namespace dmScript
         assert(timer_world != 0x0);
         assert(timer.m_IsAlive == 0);
 
+        LuaCallbackInfo* callback = (LuaCallbackInfo*) timer.m_UserData;
+        if (IsCallbackValid(callback))
+        {
+            DestroyCallback(callback);
+        }
+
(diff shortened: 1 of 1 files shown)


AGulev (approved):
👍

PR #9228, opened 2024-07-25
Upgrade GLFW to 3.4 for Win32 platforms

### LINKED ISSUE #6608: Window flashes multiple times when opening game on Windows
See our game Big Klondike built with engine version 1.3.2

https://store.steampowered.com/app/1706340/Big_Klondike__Classic_Solitaire/

It did not used to do this so it is a recent Defold change causing this issue.

Jhonnyg:
Windows is now using GLFW version 3.4 for window management. The window will always be centered now and no flickering will occur when the window is opened.

Fixes #6608


engine/engine/src/engine.cpp
```diff
@@ -1302,6 +1302,8 @@ namespace dmEngine
 
         dmGui::SetDisplayProfiles(engine->m_GuiContext, engine->m_DisplayProfiles);
 
+        dmPlatform::ShowWindow(engine->m_Window);
```
Jhonnyg, line 1305: To avoid the "flashing", we postpone showing the window until right before we draw something to it.

AGulev (dismissed):
LGTM
pay attention that editor's tests fail

PR #9249, opened 2024-07-31
Add graphics module

### LINKED ISSUE #9240: Add a 'graphics' script module
**Task (REQUIRED):**
We expose a bunch of graphics enums in _both_ the render and resource modules, but some of the enums are named differently and a lot of values are missing. We have code to convert between them, and if we add new functionality and/or enums, we need to make sure to add these enums everywhere they are used. 

Furthermore, since we are opening up the engine to be more dynamic when it comes to resources we should add functionality to query the installed graphics adapter for what it can do. Things like available device extensions, max texture size, compute shader support and so on, should be available via the graphics module. As a first step, just merging what we have into a graphics module is a good improvement.

Jhonnyg:
A new graphics script module has been added that currently only has the same graphics constants that we already use in the render and resource modules. In the future, new functionality will be added to this module for things like querying  GPU extensions, max texture size and so on.

NOTE!
`render` and `resource` constants, such as render.STATE_DEPTH_TEST still exist which means that existing projects will not need to be ported to using graphics constants. However, the constants are deprecated and removed from the autocomplete as well as the documentation in favor of using the `graphics` module!

Fixes #9240


engine/script/src/script_private.h
```diff
@@ -71,13 +71,13 @@ namespace dmScript
...
         dmConfigFile::HConfig       m_ConfigFile;
         dmResource::HFactory        m_ResourceFactory;
+        dmGraphics::HContext        m_GraphicsContext;
         dmHashTable64<Module>       m_Modules;
         dmHashTable64<Module*>      m_PathToModule;
         dmHashTable64<int>          m_HashInstances;
         dmArray<HScriptExtension>   m_ScriptExtensions;
         lua_State*                  m_LuaState;
         int                         m_ContextTableRef;
-        bool                        m_EnableExtensions;
```
Jhonnyg, line 80: This was always true, so i removed it

engine/crash/src/test/test_script_crash.cpp
```diff
@@ -49,7 +49,11 @@ class ScriptCrashTest : public jc_test_base_class
 
         dmResource::NewFactoryParams factory_params;
         m_ResourceFactory = dmResource::NewFactory(&factory_params, ".");
-        m_Context = dmScript::NewContext(m_ConfigFile, m_ResourceFactory, true);
+
+        dmScript::ContextParams script_context_params = {};
+        script_context_params.m_Factory = m_ResourceFactory;
+        script_context_params.m_ConfigFile = m_ConfigFile;
+        m_Context = dmScript::NewContext(script_context_params);
```
Jhonnyg, line 56: The signature has slightly changed, but it's not part of dmsdk so no breaking change.

engine/crash/src/test/wscript
```diff
@@ -8,7 +8,7 @@ new_copy_task('luascript', '.lua', '.luac')
 def build(bld):
     if options.skip_build_tests:
        return
-    libs = ['TESTMAIN', 'SCRIPT', 'EXTENSION', 'THREAD', 'RESOURCE', 'DDF', 'DLIB', 'PROFILE_NULL', 'LUA', 'SCRIPT']
+    libs = ['TESTMAIN', 'SCRIPT', 'EXTENSION', 'THREAD', 'RESOURCE', 'DDF', 'DLIB', 'PROFILE_NULL', 'LUA', 'GRAPHICS_NULL', 'PLATFORM_NULL']
```
Jhonnyg, line 11: The script lib has a dependency on graphics now, and graphics has a dependency on platform.

engine/engine/content/builtins/render/default.render_script
```diff
@@ -62,7 +62,7 @@ end
 
 local function update_clear_color(state, color)
     if color then
-        state.clear_buffers[render.BUFFER_COLOR_BIT] = color
```
Jhonnyg, line 65: There is no "BUFFER_COLOR_BIT" equivalent in graphics, it's BUFFER_COLORx_BIT

engine/render/src/render/render_script.cpp
```diff
@@ -611,53 +611,28 @@ namespace dmRender
         return true;
     }
 
-    /*#
-     * @name render.STATE_DEPTH_TEST
-     * @variable
-     */
```
Jhonnyg, line 617: I just moved them to the same location as all of the deprecated vars

engine/render/src/render/render_script.cpp
```diff
@@ -3203,6 +2870,407 @@ namespace dmRender
...
+    /*#
+    * Deprecated in favor of [ref:graphics.TEXTURE_FORMAT_RGB]
+    * @name render.FORMAT_RGB
+    * @variable
+    */
+
+    /*#
+    * Deprecated in favor of [ref:graphics.TEXTURE_FORMAT_RGBA]
+    * @name render.FORMAT_RGBA
+    * @variable
```
Jhonnyg, line 2918: Note: Most of the names are the same as before, but in some occasions the names are different. I've opted for using the more explicit names here that match the actual graphics enum instead of the short "FORMAT_x"

britzl (dismissed):
Very nice improvement!

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -3102,218 +3102,225 @@ static int GetTextMetrics(lua_State* L)
...
-    {"get_texture_info", GetTextureInfo},
-    {"get_render_target_info", GetRenderTargetInfo},
-    {"set_sound", SetSound},
-    {"get_buffer", GetBuffer},
-    {"set_buffer", SetBuffer},
-    {"get_text_metrics", GetTextMetrics},
-    {0, 0}
-};
-
 /*# 2D texture type
```
AGulev, line 3105: Do we want to keep it in docs? usually we remove deprecated functionality (removing # symbol)

AGulev (approved):

engine/render/src/render/render_script.cpp
```diff
@@ -873,24 +745,24 @@ namespace dmRender
...
-     * `min_filter` (optional) | `render.FILTER_LINEAR`<br/>`render.FILTER_NEAREST`
-     * `mag_filter` (optional) | `render.FILTER_LINEAR`<br/>`render.FILTER_NEAREST`
-     * `u_wrap`     (optional) | `render.WRAP_CLAMP_TO_BORDER`<br/>`render.WRAP_CLAMP_TO_EDGE`<br/>`render.WRAP_MIRRORED_REPEAT`<br/>`render.WRAP_REPEAT`<br/>
-     * `v_wrap`     (optional) | `render.WRAP_CLAMP_TO_BORDER`<br/>`render.WRAP_CLAMP_TO_EDGE`<br/>`render.WRAP_MIRRORED_REPEAT`<br/>`render.WRAP_REPEAT`
+     * `min_filter` (optional) | `graphics.TEXTURE_FILTER_LINEAR`<br/>`graphics.TEXTURE_FILTER_NEAREST`
+     * `mag_filter` (optional) | `graphics.TEXTURE_FILTER_LINEAR`<br/>`graphics.TEXTURE_FILTER_NEAREST`
+     * `u_wrap`     (optional) | `graphics.TEXTURE_WRAP_CLAMP_TO_BORDER`<br/>`graphics.TEXTURE_WRAP_CLAMP_TO_EDGE`<br/>`graphics.TEXTURE_WRAP_MIRRORED_REPEAT`<br/>`graphics.TEXTURE_WRAP_REPEAT`<br/>
+     * `v_wrap`     (optional) | `graphics.TEXTURE_WRAP_CLAMP_TO_BORDER`<br/>`graphics.TEXTURE_WRAP_CLAMP_TO_EDGE`<br/>`graphics.TEXTURE_WRAP_MIRRORED_REPEAT`<br/>`graphics.TEXTURE_WRAP_REPEAT`
      * `flags`      (optional) | `render.TEXTURE_BIT` (only applicable to depth and stencil buffers)
```
AGulev, line 755: Did you miss it? or it should be `render`?

AGulev (dismissed):
👍

engine/render/src/render/render_script.cpp
Jhonnyg: In this case it should be render, it's a render specific flag 👍

engine/gamesys/src/gamesys/scripts/script_resource.cpp
Jhonnyg: Hm, maybe we should just remove the docstrings, we shouldn't need them anymore at all

AGulev: maybe. yes. but we will have to update all the manuals after release of this version

PR #9262, opened 2024-08-05
Don't attempt to preload deprecated tile_set field

### LINKED ISSUE #9178: "ERROR:RESOURCE: Empty resource path" when async_loading a collection containing materials with no samplers.
**The bug**
When loading a collection using a collectionproxy async_load message, all materials with no samplers defined will report the following error:

> ERROR:RESOURCE: Empty resource path

**Steps to reproduce**
1. Create a material with no samplers
2. Create a collection
3. Create a GO with a Sprite component and assign the material with no samplers to it
4. Load the collection with a collection proxy with the "async_load" message
5. Observe the error in the console

**Defold version:**
  1.9.0

**Platforms**
macOS

**Repro project**
[material-error.zip](https://github.com/user-attachments/files/16116652/material-error.zip)

Jhonnyg:
Fixed an issue where loading a sprite resource with no texture samplers assigned incorrectly tries to load the tileset instead.

Fixes #9178

### DIFF
--- a/engine/gamesys/src/gamesys/resources/res_sprite.cpp
+++ b/engine/gamesys/src/gamesys/resources/res_sprite.cpp
@@ -114,10 +114,12 @@ namespace dmGameSystem
                 dmResource::PreloadHint(params->m_HintInfo, ddf->m_Textures[i].m_Texture);
             }
         }
-        else
+
+        if (ddf->m_TileSet[0] != '\0')
         {
-            dmResource::PreloadHint(params->m_HintInfo, ddf->m_TileSet);
+            dmLogInfo("Using tilesets for sprites is deprecated. '%s' will not be loaded or used.", ddf->m_TileSet);
         }
+
         dmResource::PreloadHint(params->m_HintInfo, ddf->m_Material);
 
         *params->m_PreloadData = ddf;


engine/gamesys/src/gamesys/resources/res_sprite.cpp
```diff
@@ -114,10 +114,12 @@ namespace dmGameSystem
                 dmResource::PreloadHint(params->m_HintInfo, ddf->m_Textures[i].m_Texture);
             }
         }
-        else
+
+        if (ddf->m_TileSet[0] != '\0')
         {
-            dmResource::PreloadHint(params->m_HintInfo, ddf->m_TileSet);
+            dmLogInfo("Using tilesets for sprites is deprecated. '%s' will not be loaded or used.", ddf->m_TileSet);
```
AGulev, line 120: so we can't use tilesets in sprites at all? isn't it a breaking change?

britzl: Hmm, that's not great if that is the case.

Jhonnyg: It is not used anywhere in the code. It is not a breaking change, we don’t even release it?

Jhonnyg: It is deprecated in the proto file in favor of the texture/sampler field.

AGulev: CleanShot 2024-08-05 at 14 03 33@2x

Ah, it's not related to the ability to set a tilesource as a texture on a sprite?

Jhonnyg: No, that should end up as an entry in the textures field. The tile_source field is deprecated in the source format and should not be used at all in the engine

AGulev (approved):

engine/gamesys/src/gamesys/resources/res_sprite.cpp
AGulev: Thank you for this explanation!

PR #9266, opened 2024-08-06
Only compile compute shaders for compatible shader languages

Jhonnyg:
Compute shaders cannot be built at all from the editor since the editor tries to build all languages. Compute isn't supported on webgl so spirv tools will complain and throw an error.


AGulev (approved):
👍

AGulev (approved):

PR #9269, opened 2024-08-07
Support render.enable_material with material(s) that use custom vertex attributes

### LINKED ISSUE #9246: Custom vertex attributes doesn't work together with render.enable_material
When a material is overridden by using render.enable_material, components don't generate any vertex attributes that are based on the overridden material.

**To Reproduce (REQUIRED)**
https://gitlab.com/nanaktaev/defold-vertex-attribute-issue

Jhonnyg:
Components didn't produce custom vertex attributes when a "context material" is set using render.enable_material(..) together with a material that has custom vertex attributes.

Fixes #9246


engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -983,7 +1005,7 @@ namespace dmGameSystem
         dmRender::RenderObject& ro = world->m_RenderObjects[world->m_RenderObjects.Size()-1];
 
         ro.Init();
-        ro.m_Material          = material;
+        ro.m_Material          = GetComponentMaterial(component, component->m_Resource, material_index);;
```
Jhonnyg, line 1008: We need to use the original material here to create the batches with, since the overridden material shouldn't affect what objects that should be rendered. I.e tags used for matching with the draw predicate are taken from the assigned material on the component and not from the overridden material.

AGulev (approved):
Should we changes it for extensions as well (spine, rive )?

Jhonnyg:
> Should we changes it for extensions as well (spine, rive )?

They don’t support custom vertex formats so no need!

PR #9274, opened 2024-08-08
Support turning vsync off for Vulkan

### LINKED ISSUE #9014: Vulkan doesn't take into account swap interval (vsync)
Vulkan renderer always waits for vsync, i.e. it's impossible to run the game as fast as possible.

**To Reproduce (REQUIRED)**
1. Switch on the Vulkan, i.e. `--graphics-adapter=vulkan`
2. Turn off the vsync, i.e. set swap interval to 0.
3. `dt` will be the same, i.e. 1/60 (depends on your display).

Without vsync on, the game should run as fast as possible.

  - Version: 1.8.1

**Platforms (REQUIRED):**
 - OS: Windows 10/11

[test_vsync_dt.zip](https://github.com/user-attachments/files/15519198/test_vsync_dt.zip)

**Screenshots (OPTIONAL):**
Vulkan:
image

OpenGL:
image

Jhonnyg:
Vulkan can now turn vsync off by setting the [swap interval](https://defold.com/manuals/project-settings/#swap-interval) of the game.project file to zero. Note that there is no actual concept of swap intervals built-in by Vulkan, hence this value simply turns vsync on or off. To fully support swap intervals other than these values, we need to manually wait for vblanks, which is currently not planned.

Fixes #9014


AGulev (approved):
🔥

engine/engine/src/engine.cpp
```diff
@@ -950,15 +958,6 @@ namespace dmEngine
...
         SetUpdateFrequency(engine, engine->m_UpdateFrequency);
 
-        bool setting_vsync = dmConfigFile::GetInt(engine->m_Config, "display.vsync", true); // Deprecated
-
-        uint32_t opengl_swap_interval = dmConfigFile::GetInt(engine->m_Config, "display.swap_interval", 1);
-        if (!setting_vsync)
-        {
-            opengl_swap_interval = 0;
-        }
-        SetSwapInterval(engine, opengl_swap_interval);
```
Jhonnyg, line 960: The setting is passed in to the context creation params instead of calling the function directly. We need to know this when creating the swap chain in vulkan

AGulev:
Do you want to cherry pick it into 1.9.2 ? 
CleanShot 2024-08-08 at 11 40 29@2x
if not, don't forget to change the project

Jhonnyg:
> Do you want to cherry pick it into 1.9.2 ? CleanShot 2024-08-08 at 11 40 29@2x if not, don't forget to change the project

Thanks, I changed it!

PR #9276, opened 2024-08-09
Fix issue when  debugger can't connect to the single issue after launch

AGulev:
Fixed this issue : https://forum.defold.com/t/defold-1-9-2-beta/77687/13?u=agulev

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -789,7 +789,9 @@
                                  (Thread/sleep pause-ms))        ;pause needed to make sure the launch order of instances is right
                                launched-target))
           last-launched-target (last launched-targets)]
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/app_view.clj
```diff
@@ -789,7 +789,9 @@
                                  (Thread/sleep pause-ms))        ;pause needed to make sure the launch order of instances is right
                                launched-target))
           last-launched-target (last launched-targets)]
-      (targets/select-target! prefs {:id :all-launched-targets})
+      (if (= count 1)
+        (targets/select-target! prefs last-launched-target)
+        (targets/select-target! prefs {:id :all-launched-targets}))
```
AGulev, line 794: In case if it's a single target, this target should be selected

matgis (approved):
👍

PR #9288, opened 2024-08-11
Make sure that build server url is trimmed

AGulev:
It's annoying when space in the beginning or in the end of URL prevents from building with:

>URISyntaxException: Illegal character in scheme name at index 0:  https://build.defold.com

 It happens from time to time for me and for ppl in community, I think it's easier just trim URL to make sure it's not an issue

### DIFF
--- a/editor/src/clj/editor/engine/native_extensions.clj
+++ b/editor/src/clj/editor/engine/native_extensions.clj
@@ -150,11 +150,12 @@
 (defn get-build-server-url
   (^String [prefs project]
    (g/with-auto-evaluation-context evaluation-context
-     (get-build-server-url prefs project evaluation-context)))
+     (string/trim (get-build-server-url prefs project evaluation-context))))
   (^String [prefs project evaluation-context]
-   (or (not-empty (prefs/get-prefs prefs "extensions-server" ""))
-       (not-empty (shared-editor-settings/get-setting project ["extensions" "build_server"] evaluation-context))
-       defold-build-server-url)))
+   (string/trim
+     (or (not-empty (prefs/get-prefs prefs "extensions-server" ""))
+         (not-empty (shared-editor-settings/get-setting project ["extensions" "build_server"] evaluation-context))
+         defold-build-server-url))))
 
 (defn get-build-server-headers
   ^String [prefs]
--- a/editor/src/clj/editor/engine/native_extensions.clj
+++ b/editor/src/clj/editor/engine/native_extensions.clj
@@ -150,7 +150,7 @@
 (defn get-build-server-url
   (^String [prefs project]
    (g/with-auto-evaluation-context evaluation-context
-     (string/trim (get-build-server-url prefs project evaluation-context))))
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/engine/native_extensions.clj
```diff
@@ -150,11 +150,12 @@
 (defn get-build-server-url
   (^String [prefs project]
    (g/with-auto-evaluation-context evaluation-context
-     (get-build-server-url prefs project evaluation-context)))
+     (string/trim (get-build-server-url prefs project evaluation-context))))
```
vlaaad, line 153: This code shouldn't be trimming since it uses 3-arity variant of the same `get-build-server-url` function that already trims the string.

vlaaad (changes requested):
Looks good, though we shouldn't be trimming an already-trimmed string.

editor/src/clj/editor/engine/native_extensions.clj
AGulev: true. thanks

editor/src/clj/editor/engine/native_extensions.clj
```diff
@@ -152,9 +152,10 @@
    (g/with-auto-evaluation-context evaluation-context
      (get-build-server-url prefs project evaluation-context)))
   (^String [prefs project evaluation-context]
-   (or (not-empty (prefs/get-prefs prefs "extensions-server" ""))
-       (not-empty (shared-editor-settings/get-setting project ["extensions" "build_server"] evaluation-context))
-       defold-build-server-url)))
+   (string/trim
+     (or (not-empty (prefs/get-prefs prefs "extensions-server" ""))
+         (not-empty (shared-editor-settings/get-setting project ["extensions" "build_server"] evaluation-context))
+         defold-build-server-url))))
```
vlaaad, line 158: Hmm now that got me thinking... If the user enters a build server url in preferences to be a non-empty string only with spaces, it will be trimmed and "used". We could prevent this error from happening if we trimmed the setting value before checking that it's not empty. We also don't need to trim `defold-build-server-url` since it's set by us to a string that is known to be trimmed... Then, the code would look like that:
```clj
(or (not-empty (string/trim (prefs/get-prefs prefs "extensions-server" ""))) ;; always trim because `get-prefs` does not return nil
    (not-empty (some-> (shared-editor-settings/get-setting project ["extensions" "build_server"] evaluation-context) string/trim)) ;; use `some->` because `get-setting` may return nil
    defold-build-server-url)
```

vlaaad (approved):
I think it's fine. I also think it can be improved a bit more, so non-empty whitespace-only build server preference is treated as absence of the build server, though it's optional :)

editor/src/clj/editor/engine/native_extensions.clj
AGulev: I kept your comments in code

vlaaad (approved):
🚀🚀🚀

PR #9296, opened 2024-08-13
Removed hardcoded iOS min version

britzl:
While working on some extender improvements I noticed problems with the hardcoded iOS min version of 11.0 in some of the commands. This version number should be parameterized so that it can be controller from the context or overridden when for instance building pods with a different min version.


share/extender/build_input.yml
```diff
@@ -245,8 +245,8 @@ platforms:
     zipContentPattern: 'dmengine'
     allowedLibs:    ["(\\w[\\w\\.+-]+)"]
     allowedFlags: ["-ObjC","-ObjC\\+\\+","-Wa,{{comma_separated_arg}}","-W{{warning}}","-Wno-{{warning}}","-ansi","--ansi","-std-default={{arg}}","-stdlib=(libstdc\\+\\+|libc\\+\\+)","-w","-std=(c89|c99|c\\+\\+98|c\\+\\+0x|c\\+\\+11|c\\+\\+14|c\\+\\+17|c\\+\\+20)","-Wp,{{comma_separated_arg}}","--extra-warnings","--warn-{{warning}}","--warn-={{warning}}","-f(no-)?objc-arc(-exceptions)?","-ferror-limit={{number}}","-O([0-4]?|fast|s|z)","-Wl,[-_a-zA-Z0-9,/@]+","-fno-rtti","-f\\w[\\w-=]+","-x","c"]
-    compileCmd: 'clang++ -c -arch arm64 -target arm-apple-darwin19 -m64 -O2 -g -stdlib=libc++ -miphoneos-version-min={{osMinVersion}} -isysroot {{env.SYSROOT}} -nostdinc++ {{#systemIncludes}}-isystem {{{.}}} {{/systemIncludes}} {{#defines}}-D{{{.}}} {{/defines}} {{#flags}}{{{.}}} {{/flags}} {{#ext.includes}}-I{{{.}}} {{/ext.includes}} {{#ext.frameworkPaths}}-F{{{.}}} {{/ext.frameworkPaths}} {{#includes}}-I{{{.}}} {{/includes}} {{#platformIncludes}}-I{{.}} {{/platformIncludes}} {{src}} -o{{tgt}}'
```
britzl, line 248: arm-apple-darwin19 is far from correct when building for iOS. It should be arm-apple.iosx.y.

share/extender/build_input.yml
```diff
@@ -232,7 +232,7 @@ platforms:
         dynamicLibs: []
         symbols:    ["AudioDecoderTremolo"]
         osMinVersion:   "{{env.IOS_VERSION_MIN}}"
-        swiftFlags: ["-target", "arm64-apple-ios11.0"]
```
britzl, line 235: Some pods require a higher min version than 11 to build

britzl: It's interesting because we also hardcode it in the Info.plist and users that wish to change min version need to copy the Info.plist, modify the min version and point to it from game.project. Would it not make sense to perhaps make this value configurable from game.project in the iOS section, and also send this value to the extender server?

JCash: I agree configurability is good.
Not sure about the location. Is game.project enough? Won't it need to be rebuilt? If so, then an app manifest makes more sense perhaps?

share/extender/build_input.yml
```diff
@@ -283,8 +283,8 @@ platforms:
...
+    linkCmd: 'clang++ -arch x86_64 -target x86_64-apple-ios{{osMinVersion}}-simulator -m64 -O2 -g -stdlib=libc++ -miphoneos-version-min={{osMinVersion}} -isysroot {{env.SYSROOT}} -o {{tgt}} {{#linkFlags}}{{{.}}} {{/linkFlags}} {{#ext.libPaths}}-L{{{.}}} {{/ext.libPaths}} {{#libs}}-l{{{.}}} {{/libs}} {{#ext.libs}}-l{{{.}}} {{/ext.libs}} {{#ext.dynamicLibs}}-l{{{.}}} {{/ext.dynamicLibs}} {{#dynamicLibs}}-l{{{.}}} {{/dynamicLibs}} {{#ext.frameworkPaths}}-F{{{.}}} {{/ext.frameworkPaths}} {{#ext.frameworks}}-framework {{{.}}} {{/ext.frameworks}} {{#frameworks}}-framework {{{.}}} {{/frameworks}} {{#weakFrameworks}}-weak_framework {{{.}}} {{/weakFrameworks}} {{#libPaths}}-L{{{.}}} {{/libPaths}} {{#engineLibs}}-l{{{.}}} {{/engineLibs}} {{#src}}{{{.}}} {{/src}}'
```
JCash, line 287: Should these be configurable from the input yaml as well?

AGulev:
related https://github.com/defold/defold/issues/3526

share/extender/build_input.yml
britzl: The reason I suggested game.project was because that is where we configure similar version numbers for other platforms, such as Android min and target versions. And it also means that it will be easy to parameterize Info.plist:

https://github.com/defold/defold/blob/dev/engine/engine/content/builtins/manifests/ios/Info.plist#L81

I suppose we can pull it from game.project when we send the build request to the extender?

share/extender/build_input.yml
britzl: What do you mean exactly? That it should be `x86_64-apple-ios{{env.IOS_VERSION_MIN}}-simulator` instead of `{{osMinVersion}} ?`

JCash: Yes.

britzl: I've changed to env.IOS_VERSION_MIN instead

JCash (approved):

PR #9316, opened 2024-08-19
Ensure build resource paths for embedded resources always reflect their contents

matgis:
Referencing a `.wav` or `.ogg` file directly as a component from a game object embedded in a collection no longer throws an exception when building the project.

### Technical changes
* We now have separate production functions to produce the build targets from referenced and embedded components. Some common outputs that did not make sense for both cases have been removed.
* It is no longer the responsibility of the consumer of an embedded resource build target to make sure its `:resource` field reflects its contents. Instead, `bt/with-content-hash` will now update the `:data` field of source `MemoryResources` with the calculated `content-hash`.
* All code paths that previously calculated a `:data` prior to producing the `build-target` have been updated to instead use the `:resource` of the resulting `build-target` to reference the resulting binary.

### DIFF
--- a/editor/src/clj/editor/collection.clj
+++ b/editor/src/clj/editor/collection.clj
@@ -177,25 +177,48 @@
    :icon ""
    :label ""})
 
-(g/defnk produce-go-build-targets [_node-id build-error build-resource ddf-message resource-property-build-targets source-build-targets pose]
-  ;; Create a build-target for the referenced or embedded game object. Also tag
-  ;; on :game-object-instance-data with the overrides for this instance. This
-  ;; will later be extracted and compiled into the Collection - the overrides do
-  ;; not end up in the resulting game object binary.
+(defn- path-error [node-id resource]
+  (or (validation/prop-error :fatal node-id :path validation/prop-nil? resource "Path")
+      (validation/prop-error :fatal node-id :path validation/prop-resource-not-exists? resource "Path")))
+
+(defn- component-property-error [node-id ddf-message]
+  (when-some [errors (not-empty (sequence (comp (mapcat :properties) ; Extract GameObject$PropertyDescs from GameObject$ComponentPropertyDescs.
+                                                (keep :error))
+                                          (:component-properties ddf-message)))]
+    (g/error-aggregate errors :_node-id node-id :_label :build-targets)))
+
+(g/defnk produce-referenced-go-build-targets [_node-id ddf-message resource-property-build-targets source-build-targets pose source-resource]
+  ;; Create a build-target for the referenced game object. Also tag on
+  ;; :game-object-instance-data with the overrides for this instance. This will
+  ;; later be extracted and compiled into the Collection - the overrides do not
+  ;; end up in the resulting game object binary.
   ;; Please refer to `/engine/gameobject/proto/gameobject/gameobject_ddf.proto`
   ;; when reading this. It describes how the ddf-message map is structured.
   ;; You might also want to familiarize yourself with how this process works in
   ;; `game_object.clj`, since it is similar but less complicated there.
-  (if-some [errors
-            (not-empty
-              (sequence (comp (mapcat :properties) ; Extract GameObject$PropertyDescs from GameObject$ComponentPropertyDescs.
-                              (keep :error))
-                        (:component-properties ddf-message)))]
-    (g/error-aggregate errors :_node-id _node-id :_label :build-targets)
-    (let [game-object-build-target (first source-build-targets)
-          proj-path->resource-property-build-target (bt/make-proj-path->build-target resource-property-build-targets)
-          instance-desc-with-go-props (dissoc ddf-message :data)] ; GameObject$InstanceDesc or GameObject$EmbeddedInstanceDesc in map format. We don't need the :data from GameObject$EmbeddedInstanceDesc.
-      [(collection-common/game-object-instance-build-target build-resource instance-desc-with-go-props pose game-object-build-target proj-path->resource-property-build-target)])))
+  (or (path-error _node-id source-resource)
+      (component-property-error _node-id ddf-message)
+      (let [game-object-build-target (first source-build-targets)
+            build-resource (:resource game-object-build-target)
+            proj-path->resource-property-build-target (bt/make-proj-path->build-target resource-property-build-targets)]
+        [(collection-common/game-object-instance-build-target build-resource ddf-message pose game-object-build-target proj-path->resource-property-build-target)])))
+
+(g/defnk produce-embedded-go-build-targets [_node-id ddf-message resource-property-build-targets source-build-targets pose]
+  ;; Create a build-target for the embedded game object. Also tag on
+  ;; :game-object-instance-data with the overrides for this instance. This will
+  ;; later be extracted and compiled into the Collection - the overrides do not
+  ;; end up in the resulting game object binary.
+  ;; Please refer to `/engine/gameobject/proto/gameobject/gameobject_ddf.proto`
+  ;; when reading this. It describes how the ddf-message map is structured.
+  ;; You might also want to familiarize yourself with how this process works in
+  ;; `game_object.clj`, since it is similar but less complicated there.
+  (or (component-property-error _node-id ddf-message)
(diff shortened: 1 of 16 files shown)


editor/src/clj/editor/collection.clj
```diff
@@ -177,25 +177,46 @@
    :icon ""
    :label ""})
 
-(g/defnk produce-go-build-targets [_node-id build-error build-resource ddf-message resource-property-build-targets source-build-targets pose]
-  ;; Create a build-target for the referenced or embedded game object. Also tag
-  ;; on :game-object-instance-data with the overrides for this instance. This
-  ;; will later be extracted and compiled into the Collection - the overrides do
-  ;; not end up in the resulting game object binary.
```
matgis, line 184: Split into `produce-referenced-go-build-targets` and `produce-embedded-go-build-targets`.

editor/src/clj/editor/collection.clj
```diff
@@ -216,9 +237,7 @@
   (output node-outline outline/OutlineData :cached produce-go-outline)
   (output ddf-message g/Any :abstract)
   (output node-outline-extras g/Any (g/constantly {}))
-  (output build-resource resource/Resource :abstract)
-  (output build-targets g/Any produce-go-build-targets)
-  (output build-error g/Err (g/constantly nil))
+  (output build-targets g/Any :abstract)
```
matgis, line 240: Instead of trying to share code in the base node, we now have separate `build-targets` implementations instead.

editor/src/clj/editor/collection.clj
```diff
@@ -287,15 +305,6 @@
...
 
-(defn- path-error [node-id resource]
-  (or (validation/prop-error :fatal node-id :path validation/prop-nil? resource "Path")
-      (validation/prop-error :fatal node-id :path validation/prop-resource-not-exists? resource "Path")))
-
-(defn- substitute-error [val-or-error substitute]
-  (if-not (g/error? val-or-error)
-    val-or-error
-    substitute))
-
```
matgis, line 298: `path-error` was moved up in the file, but `substitute-error` was unused.

editor/src/clj/editor/collection_common.clj
```diff
@@ -159,15 +159,14 @@
...
                                    :property-deps go-prop-dep-build-targets
                                    :instance-msg (protobuf/assign-repeated instance-desc-with-go-props
-                                                   :component-properties component-property-descs)}
-        build-target (assoc game-object-build-target
-                       :resource build-resource
-                       :game-object-instance-data game-object-instance-data)]
-    (bt/with-content-hash build-target)))
+                                                   :component-properties component-property-descs)}]
+    (assoc game-object-build-target
+      :game-object-instance-data game-object-instance-data)))
```
matgis, line 169: No need to recalculate the `content-hash` of the `GameObject$PrototypeDesc` - we're just tacking on instance data which will result in a `GameObject$InstanceDesc` binary.

editor/src/clj/editor/game_object.clj
```diff
@@ -136,34 +136,40 @@
...
-(g/defnk produce-component-build-targets [_node-id build-resource ddf-message pose resource-property-build-targets source-build-targets]
-  ;; Create a build-target for the referenced or embedded component. Also tag on
+(g/defnk produce-referenced-component-build-targets [_node-id source-resource ddf-message pose resource-property-build-targets source-build-targets]
+  ;; Create a build-target for the referenced component. Also tag on
   ;; :component-instance-data with the overrides for this instance. This will
   ;; later be extracted and compiled into the GameObject - the overrides do not
   ;; end up in the resulting component binary.
-  ;;
-  ;; TODO: We can avoid some of this processing & input dependencies for embedded components. Separate into individual production functions?
-  ;; For example, embedded components do not have resource-property-build-targets, and cannot refer to raw sounds.
```
matgis, line 151: We're doing it! Yay!

editor/src/clj/editor/game_object.clj
```diff
@@ -69,10 +69,10 @@
     (if-not (sound/supported-audio-formats ext)
       component-source-build-target ; Not a raw sound. Return unaltered.
       (let [workspace (resource/workspace source-resource)
+            dep-build-targets [component-source-build-target] ; The build-target of the referenced .wav file.
             sound-desc (protobuf/make-map-without-defaults Sound$SoundDesc
                          :sound (resource/proj-path source-resource))
-            sound-desc-resource (workspace/make-embedded-resource workspace :editable "sound" sound-desc)
-            dep-build-targets [component-source-build-target]]
+            sound-desc-resource (workspace/make-placeholder-resource workspace "sound")]
```
matgis, line 75: This is in `wrap-if-raw-sound`.

editor/src/clj/editor/game_object.clj
```diff
@@ -136,34 +136,40 @@
...
-(defn resource-path-error [_node-id source-resource]
+(defn- resource-path-error [_node-id source-resource]
     (or (validation/prop-error :fatal _node-id :path validation/prop-nil? source-resource "Path")
         (validation/prop-error :fatal _node-id :path validation/prop-resource-not-exists? source-resource "Path")
         (validation/prop-error :fatal _node-id :script validation/prop-resource-not-component? source-resource "Path")))
 
-(g/defnk produce-component-build-targets [_node-id build-resource ddf-message pose resource-property-build-targets source-build-targets]
-  ;; Create a build-target for the referenced or embedded component. Also tag on
+(g/defnk produce-referenced-component-build-targets [_node-id source-resource ddf-message pose resource-property-build-targets source-build-targets]
+  ;; Create a build-target for the referenced component. Also tag on
```
matgis, line 145: Split into `produce-referenced-component-build-targets` and `produce-embedded-component-build-targets`.

editor/src/clj/editor/game_object_common.clj
```diff
@@ -224,8 +224,8 @@
     {:resource build-resource
      :content (protobuf/map->bytes GameObject$PrototypeDesc prototype-desc)}))
 
-(defn game-object-build-target [build-resource host-resource-node-id component-instance-datas component-build-targets]
-  {:pre [(or (nil? build-resource) (workspace/build-resource? build-resource))
+(defn game-object-build-target [source-resource host-resource-node-id component-instance-datas component-build-targets]
+  {:pre [(workspace/source-resource? source-resource)
```
matgis, line 228: Changed to take a `source-resource` to encourage callers to obtain the `BuildResource` from the resulting `build-target`.

editor/src/clj/editor/game_object_non_editable.clj
```diff
@@ -54,12 +54,6 @@
       (project/load-embedded-resource-node project embedded-resource-node-id embedded-resource embedded-resource-pb-map)
       (gu/connect-existing-outputs embedded-resource-node-type embedded-resource-node-id host-node-id embedded-component-connections))))
 
-(g/defnk produce-embedded-component-build-targets [embedded-component-build-targets]
-  (into []
-        (map (comp #(assoc % :resource (bt/make-content-hash-build-resource %))
-                   util/only-or-throw))
-        embedded-component-build-targets))
-
```
matgis, line 62: We don't need to do this anymore. The `:data` of the `MemoryResources` in the `embedded-component-build-targets` will already reflect their `content-hash`.

editor/test/editor/build_target_test.clj
```diff
@@ -42,11 +42,6 @@
      (.flush writer)
      (.toString writer))))
 
-(defn- make-lua-memory-resource [workspace source]
-  (assert (string? source))
-  (let [resource-type (workspace/get-resource-type workspace "lua")]
-    (resource/make-memory-resource workspace resource-type source)))
-
```
matgis, line 49: Inlined, since we have `workspace/make-memory-resource`.

AGulev:
Does this PR fix this issue https://github.com/defold/defold/issues/4959 ?

matgis:
Not yet. That will be fixed later!

vlaaad (approved):
A-ha, so we used to calculate the data for memory resources just so that the hash is right, and now we don't have to do it, because the right hash is injected into these resources? Soods good!

PR #9339, opened 2024-08-26
Make sure that engine owns `path` in http request on web

AGulev:
I got a bunch of issues when using `path` in html5:

>"DEFOLD:ERROR:RESOURCE: Could not resolve a mount path '/dat+'\n"
"DEFOLD:ERROR:RESOURCE: Could not resolve a mount path '�,�'\n"
"DEFOLD:ERROR:RESOURCE: Could not resolve a mount path 'x7�'\n"

Take a look into code, and found that engine doesn't own `path` variable, it's owned by Lua, which may be a reason of unexpected bugs.
Now I allocate memory for `path` variable (it's an issue only for web build)

Test project where it's easy to catch: 
[DownloadFileTest.zip](https://github.com/user-attachments/files/16750566/DownloadFileTest.zip)


ekharkunov (approved):
LGTM! Thanks for the fix ❤️

britzl (approved):

PR #9347, opened 2024-08-28
Added fix for asynchronous raycast messages

JCash:
This fixes a rare crash where lingering raycast messages were handled the exact frame when the collection was freed.

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/engine/dlib/src/dlib/message.cpp
+++ b/engine/dlib/src/dlib/message.cpp
@@ -360,6 +360,21 @@ namespace dmMessage
         }
     }
 
+    dmhash_t GetSocketNameHash(HSocket socket)
+    {
+        DM_SPINLOCK_SCOPED_LOCK(g_MessageSpinlock);
+
+        MessageSocket* message_socket = g_MessageContext->m_Sockets.Get(socket);
+        if (message_socket != 0x0)
+        {
+            return message_socket->m_NameHash;
+        }
+        else
+        {
+            return 0x0;
+        }
+    }
+
     bool IsSocketValid(HSocket socket)
     {
         if (socket != 0)
--- a/engine/dlib/src/test/test_httpclient.cpp
+++ b/engine/dlib/src/test/test_httpclient.cpp
@@ -951,13 +951,19 @@ TEST_P(dmHttpClientTestExternal, PostExternal)
     m_Content = "";
     m_StatusCode = -1;
 
-    r = dmHttpClient::Post(m_Client, m_URI.m_Path);
     printf("POST'ing to %s://%s%s\n", m_URI.m_Scheme, m_URI.m_Location, m_URI.m_Path);
+    r = dmHttpClient::Post(m_Client, m_URI.m_Path);
+
+    printf("STATUS: %d\n", m_StatusCode);
+    if (503 == m_StatusCode)
+    {
+        // It's an external page, which may be down from time to time
+        return;
+    }
 
     ASSERT_EQ(dmHttpClient::RESULT_OK, r);
     ASSERT_EQ(200, m_StatusCode);
 
-    printf("STATUS: %d\n", m_StatusCode);
     printf("CONTENT:\n%s\n", m_Content.c_str());
 }
 
--- a/engine/engine/loop.sh
+++ b/engine/engine/loop.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+COUNT=$1
+shift
+ARGS=$*
+
+function usage {
+    echo "USAGE: iterate.sh <count> <cmd> [<args>]"
+    exit 1
+}
+
+if [ "" == "${COUNT}" ]; then
+    usage
+fi
+
+if [ "" == "${ARGS}" ]; then
+    usage
+fi
+
+set -e
+
+for (( i=0 ; i < ${COUNT} ; i++ )) ; do
+    echo "**************************************************************************************************************"
+    echo "Loop: ${i}: ${ARGS} "
+    ${ARGS}
+done
--- a/engine/engine/src/test/main.script
+++ b/engine/engine/src/test/main.script
@@ -59,7 +59,8 @@ local function deactivate_proxy(self, index)
     msg.post(proxy, "unload")
 end
 
-function on_message(self, message_id, message)
+ --TODO: Rearrange this mess: Make sure each state is only started after the previous is unloaded!
+function on_message(self, message_id, message, sender)
     if message_id == hash("proxy_loaded") then
         msg.post(self.loading_proxy, "init")
         msg.post(self.loading_proxy, "enable")
--- a/engine/gameobject/src/gameobject/gameobject.cpp
+++ b/engine/gameobject/src/gameobject/gameobject.cpp
@@ -328,6 +328,18 @@ namespace dmGameObject
         regist->m_Collections.SetSize(0);
     }
 
+    HCollection GetCollectionByHash(HRegister regist, dmhash_t socket_name)
+    {
+        uint32_t collection_count = regist->m_Collections.Size();
+        for (uint32_t i = 0; i < collection_count; ++i)
+        {
+            Collection* collection = regist->m_Collections[i];
+            if (collection->m_NameHash == socket_name)
+                return collection->m_HCollection;
+        }
+        return 0;
+    }
+
     void DeleteRegister(HRegister regist)
     {
         DeleteCollections(regist);
@@ -491,7 +503,7 @@ namespace dmGameObject
             return 0;
         }
 
-        collection->m_NameHash = dmHashString64(name);
+        collection->m_NameHash = dmHashString64(name); // Same as the socket name
 
         HCollection hcollection = (HCollection)new CollectionHandle;
         Result result = AttachCollection(collection, name, factory, regist, hcollection);
--- a/engine/gameobject/src/gameobject/gameobject.cpp
+++ b/engine/gameobject/src/gameobject/gameobject.cpp
@@ -2235,23 +2235,7 @@ namespace dmGameObject
         DispatchMessagesContext* context = (DispatchMessagesContext*) user_ptr;
         Collection* collection = context->m_Collection;
 
-        Instance* instance = 0x0;
-        // Start by looking for the instance in the user-data,
-        // which is the case when an instance sends to itself.
-        if (message->m_UserData1 != 0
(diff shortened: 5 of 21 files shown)


engine/dlib/src/dlib/message.cpp
```diff
@@ -360,6 +360,21 @@ namespace dmMessage
         }
     }
 
+    dmhash_t GetSocketNameHash(HSocket socket)
```
JCash, line 363: We had a way to get the name as a string, but not as a hash.

engine/dlib/src/test/test_httpclient.cpp
```diff
@@ -951,13 +951,19 @@ TEST_P(dmHttpClientTestExternal, PostExternal)
...
 
-    r = dmHttpClient::Post(m_Client, m_URI.m_Path);
     printf("POST'ing to %s://%s%s\n", m_URI.m_Scheme, m_URI.m_Location, m_URI.m_Path);
+    r = dmHttpClient::Post(m_Client, m_URI.m_Path);
+
+    printf("STATUS: %d\n", m_StatusCode);
+    if (503 == m_StatusCode)
+    {
+        // It's an external page, which may be down from time to time
+        return;
```
JCash, line 961: Another intermittent test issue: when the external server is unavailable.

engine/engine/loop.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
JCash, line 1: Perhaps useful again :)

How I triggered my ASAN crashes:
```
loop.sh 100 ./build/src/test/test_engine --test-filter RunScript
```

engine/engine/src/test/main.script
```diff
@@ -59,7 +59,8 @@ local function deactivate_proxy(self, index)
     msg.post(proxy, "unload")
 end
 
-function on_message(self, message_id, message)
+ --TODO: Rearrange this mess: Make sure each state is only started after the previous is unloaded!
+function on_message(self, message_id, message, sender)
```
JCash, line 63: The loading logic seems flawed, sending disabled messaged to non-enabled collections.

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -328,6 +328,18 @@ namespace dmGameObject
...
+        uint32_t collection_count = regist->m_Collections.Size();
+        for (uint32_t i = 0; i < collection_count; ++i)
+        {
+            Collection* collection = regist->m_Collections[i];
+            if (collection->m_NameHash == socket_name)
+                return collection->m_HCollection;
+        }
+        return 0;
+    }
+
```
JCash, line 342: I don't particularly _like_ this way, but since the solution has no context (all messages are sent to the generic `@physics:` socket, and flushed by all physics worlds), we need a way to get our context back (i.e. the calling collection).

engine/gamesys/src/gamesys/components/comp_collision_object.cpp
```diff
@@ -900,62 +900,56 @@ namespace dmGameSystem
...
+            return;
+
+        dmDDF::Descriptor* descriptor = (dmDDF::Descriptor*)message->m_Descriptor;
+        if (descriptor == dmPhysicsDDF::RequestRayCast::m_DDFDescriptor)
         {
-            dmDDF::Descriptor* descriptor = (dmDDF::Descriptor*)message->m_Descriptor;
-            if (descriptor == dmPhysicsDDF::RequestRayCast::m_DDFDescriptor)
+            dmPhysicsDDF::RequestRayCast* ddf = (dmPhysicsDDF::RequestRayCast*)message->m_Data;
+
+            dmhash_t coll_name_hash = dmMessage::GetSocketNameHash(message->m_Sender.m_Socket);
```
JCash, line 921: Instead of using the go instance pointer sent as a user data, we'll find the collection by using the sender URL.

engine/gamesys/src/gamesys/components/comp_collision_object.cpp
```diff
@@ -900,62 +900,56 @@ namespace dmGameSystem
...
+
+            // Target collection which can be different than we are updating for.
+            dmGameObject::HCollection collection = dmGameObject::GetCollectionByHash(context->m_Register, coll_name_hash);
+            if (!collection) // if the collection has been removed
+                return;
+
+            // NOTE! The collision world for the target collection is looked up using this worlds component index
+            //       which is assumed to be the same as in the target collection.
+            uint32_t component_type_index = context->m_ComponentTypeIndex;
+            CollisionWorld* world = (CollisionWorld*) dmGameObject::GetWorld(collection, component_type_index);
```
JCash, line 931: We no longer need the go instance at all, we just need the collection to find the CollisionWorld*.

engine/gamesys/src/gamesys/components/comp_collision_object.cpp
```diff
@@ -900,62 +900,56 @@ namespace dmGameSystem
...
+
+            // Give that the assumption above holds, this assert will hold too.
+            assert(world->m_ComponentTypeIndex == component_type_index);
+
+            dmMessage::URL* receiver = (dmMessage::URL*)malloc(sizeof(dmMessage::URL));
+            memcpy(receiver, &message->m_Sender, sizeof(*receiver));
```
JCash, line 937: The message comes from a script, and we want to return the messages to that script.

NOTE: A change in behavior here.
Previously, we set the receiver url to point to the instance like so (no fragment!?):
```
receiver->m_Socket = dmGameObject::GetMessageSocket(collection);
receiver->m_Path = dmGameObject::GetIdentifier(sender_instance);
```

Skipping the fragment, means that it will broadcast the physics message response to _all_ components under that game object.
That didn't seem right to me.

So, instead, I copy the sender's url (which is the script calling the raycast function).
Should be more efficient (and correct imho)

engine/gamesys/src/gamesys/components/comp_collision_object.cpp
```diff
@@ -900,62 +900,56 @@ namespace dmGameSystem
...
+
+            dmPhysics::RayCastRequest request;
+            request.m_From = ddf->m_From;
+            request.m_To = ddf->m_To;
+            request.m_Mask = ddf->m_Mask;
+            request.m_UserId = (ddf->m_RequestId & 0xff);
```
JCash, line 943: Previously, we put the component index in here as well, but I couldn't find _any_ reference to it ever being used.

engine/gamesys/src/gamesys/components/comp_collision_object.cpp
```diff
@@ -900,62 +900,56 @@ namespace dmGameSystem
...
+            request.m_UserData = (void*)receiver;
+            request.m_IgnoredUserData = 0;
```
JCash, line 945: ANother issue: Previously we passed in the `URL*`, in order to skip testing against collision objects with that userdata.
However, the user data for each collision object is the `CollisionComponent*`, so it'll never match anyways.

engine/gamesys/src/gamesys/components/comp_collision_object.cpp
```diff
@@ -1006,6 +1001,7 @@ namespace dmGameSystem
             physics_socket = dmPhysics::GetSocket2D(physics_context->m_Context2D);
 
         }
+
```
JCash, line 1004: One of the root causes of this issue, is that we only have one `@physics` socket, and each `CollisionWorld*` is flushing the message queue. It means we don't have a good way to provide better context  for handling the messages.

See the comment above this function for previous suggested solutions (e.g. one socket per physics world)

engine/gamesys/src/gamesys/scripts/script_camera.cpp
```diff
@@ -52,15 +52,14 @@ namespace dmGameSystem
     static int Camera_AcquireFocus(lua_State* L)
     {
         DM_LUA_STACK_CHECK(L, 0);
-        dmGameObject::HInstance instance = CheckGoInstance(L);
+        (void)CheckGoInstance(L); // left to check that it's not called from incorrect context.
```
JCash, line 55: Now that we no longer _need_ the instance, the question is: Should we still call the function?
Currently, it helps us checking that we're in a go script, and possibly within a life cycle function.

engine/gamesys/src/gamesys/scripts/script_camera.cpp
```diff
@@ -52,15 +52,14 @@ namespace dmGameSystem
...
 
         dmMessage::URL receiver;
         dmMessage::URL sender;
         dmScript::ResolveURL(L, 1, &receiver, &sender);
 
         dmGamesysDDF::AcquireCameraFocus msg;
-        dmMessage::Post(&sender, &receiver, dmGamesysDDF::AcquireCameraFocus::m_DDFDescriptor->m_NameHash, (uintptr_t)instance,
-            0, (uintptr_t)dmGamesysDDF::AcquireCameraFocus::m_DDFDescriptor, &msg, sizeof(msg), 0);
+        dmMessage::Post(&sender, &receiver, dmGamesysDDF::AcquireCameraFocus::m_DDFDescriptor->m_NameHash, 0, 0, (uintptr_t)dmGamesysDDF::AcquireCameraFocus::m_DDFDescriptor, &msg, sizeof(msg), 0);
```
JCash, line 62: In these cases, the "m_UserData1" was never used on the receiving end, thus safe to remove.

AGulev: As I understand for the cases you removed m_UserData1 this logic will never be touched? https://github.com/defold/defold/blob/7cf88519e88c23e0fa725b263db0c23f31a29ef9/engine/gameobject/src/gameobject/gameobject.cpp#L2227-L2238

AGulev (dismissed):

engine/gamesys/src/gamesys/scripts/script_camera.cpp
JCash: Good catch.
While it's an optimization, I think it's efficient enough to use the GetInstanceFromIdentifier() (a hash table lookup).
This keeps the code cleaner imho, and less "extra knowledge" you need to use these functions.

I'll remove it for now.

PR #9382, opened 2024-09-04
Add missing color into manufactured name keys

Jhonnyg:
"Manufactured" color attributes are never considered for models, which means that we don't pick up on vertex colors from default vertex declarations.


matgis (approved):
Ship it!

PR #9391, opened 2024-09-06
Remap gamepad input on windows for backwards compatability

### LINKED ISSUE #9348: 1.9.2 - gamepad names have changed, causing mappings to fail
Gamepad input mappings rely on the reported name of the gamepad. It seems that as of 1.9.2, gamepad names have changed. My gamepad on Windows used to be called "XBox 360 Controller" but is now called "Wireless Xbox Controller" and therefore fails to map. User Wolfe [reports](https://discord.com/channels/250018174974689280/250018634494115853/1278299962626543626) 8BitDo SN30 pro+ on Linux has changed mapping. Furthermore, I also have player reports that "Super Gamepad" and "Xbox Elite Controller 2" are not working as of an update I pushed with version 1.9.2. Notably, one user reports that "PS Dualsense" still works. I have no insight into why that is (perhaps it's a newer controller, or Steam Input interferes, or it's random chance).

This issue is perhaps related to the recent GLFW update? This update was implemented in 1.9.2: https://github.com/defold/defold/pull/9228

The problem with the gamepad names changing is that this invalidates an unknown amount of the gamepad mappings in the builtins.

Jhonnyg:
Fixed a regression issue on windows for gamepads where the names of gamepad mappings (and input) had unfortunately changed due to an update to the GLFW library that we use for platform support. Gamepads should now have the same mapping and names as before!

Fixes #9348


engine/hid/src/hid.h
```diff
@@ -35,6 +35,8 @@ namespace dmHID
 {
     /// Constant that defines invalid context handles
     const HContext INVALID_CONTEXT = 0;
+    const uint8_t MAX_GAMEPAD_NAME_LENGTH = 128;
+    const uint8_t MAX_GAMEPAD_NAME_COUNT  = 2;
```
Jhonnyg, line 39: We used a max of 128 bytes for the name elsewhere, so I figured we could make these const instead..

engine/hid/src/native/hid_gamepad_driver_glfw.cpp
```diff
@@ -112,6 +125,35 @@ namespace dmHID
...
+
+        for (int i = 0; i < num_names; ++i)
+        {
+            if (strcmp(gamepad_names[i], "Wireless Xbox Controller") == 0 ||
+                strcmp(gamepad_names[i], "Xbox Controller") == 0 ||
+                strstr(gamepad_names[i], "XInput"))
+            {
+                new_device.m_RemapStrategy = GAMEPAD_REMAP_STRATEGY_LEGACY_XINPUT;
+                break;
+            }
```
Jhonnyg, line 147: I don't know how robust this is, but it works for the gamepads I've tested.. This is how they are statically named in GLFW3 now, so the alternative would be to actually patch the GLFW3 source so we could query this directly, but I'd like to avoid that if we can..

engine/hid/src/native/hid_gamepad_driver_glfw.cpp
```diff
@@ -148,13 +183,17 @@ namespace dmHID
             {
                 return;
             }
+
             SetGamepadConnectionStatus(g_GLFWGamepadDriver->m_HidContext, gp, event == dmPlatform::GAMEPAD_EVENT_CONNECTED ? 1 : 0);
         }
     }
 
-    static inline uint8_t* RemapGamepadWin32(Gamepad* gamepad, uint8_t* buttons, uint8_t* buttons_remapped)
+    static inline uint8_t* RemapGamepadWin32(GLFWGamepadDevice* glfw_gamepad, uint8_t* buttons, uint8_t* buttons_remapped)
```
JCash, line 191: Perhaps ifdef the function as well, to make sure it's not accidentally included in another platform (even if it's called)

engine/input/src/input.cpp
```diff
@@ -142,44 +142,93 @@ namespace dmInput
...
-        } else {
-            GamepadConfig* config = GetGamepadConfigFromDeviceName(binding, dmHashString32(device_name));
-            if (config == 0x0)
+
+        GamepadConfig* best_config = 0x0;
+        int best_config_name_index = -1;
+
+        for (int i = 0; i < num_names; ++i)
+        {
+            if (strlen(device_names[i]) == 0)
```
JCash, line 157: This check seems a little redundant tbh. Both calling a function to check the first char==0, and also added code size.
I think the GetGamepadConfigFromDeviceName could handle it just as well?

engine/input/src/input.cpp
```diff
@@ -142,44 +142,93 @@ namespace dmInput
...
-             *  - The input-binding file covers several platforms and certain platforms
-             *    doesn't have support for e.g. pads. But more importantly, sometimes you might have
-             *    a device connected but sometimes not. It should be up to the user and we shouldn't
-             *    spam out warnings in such cases. In other words. It's impossible to tell whether the
-             *    warning is appropriate or not.
-             *  - We should also have support dynamic pad-connections. Whether a pad is connected
-             *    or not should be up to the game-ui.
-             */
+        char device_names[dmHID::MAX_GAMEPAD_NAME_COUNT][dmHID::MAX_GAMEPAD_NAME_LENGTH];
+        uint32_t num_names = dmHID::GetGamepadDeviceNames(binding->m_Context->m_HidContext, gamepad, device_names);
```
JCash, line 148: I wonder if it would be enough to simply add the extra name in input.cpp?

```c++
char device_name[dmHID::MAX_GAMEPAD_NAME_LENGTH];
dmHID::GetGamepadDeviceName(binding->m_Context->m_HidContext, gamepad, device_name, sizeof(device_name));

uint32_t num_names = 1;
char** device_names[2] = { device_name, 0};

#if defined(_WIN32)
    device_names[num_names++] = "XBox 360 Controller";
#endif
```

JCash (changes requested):
It feels like a lot of code for supporting legacy names.
Is the simplification I suggested a good approach?

engine/input/src/input.cpp
JCash: The "#if defined(_WIN32)" condition may ofc be different, but the overall question is the same.

engine/input/src/input.cpp
Jhonnyg: This is legacy code that I kept around

engine/input/src/input.cpp
Jhonnyg: Yeah I think that's probably a good idea, I'll try it out :)

JCash (approved):

PR #9417, opened 2024-09-11
Fix test regressions due to WebGPU adapter

Jhonnyg:
Technical changes:
* We must unfortunately expand combined samplers for vulkan and webgpu into separate texture and sampler object(s)
* The shader bindings contain an index for a sampler to which texture it originally was created from
* Sampler objects are resolved in vulkan to point to their original texture's texture unit
* The graphics WebGPU library is included in the SDK


engine/graphics/src/graphics.h
```diff
@@ -701,7 +701,6 @@ namespace dmGraphics
         return type == TYPE_SAMPLER_2D ||
                type == TYPE_SAMPLER_CUBE ||
                type == TYPE_SAMPLER_2D_ARRAY ||
-               type == TYPE_SAMPLER ||
```
Jhonnyg, line 704: Sampler isn't a valid texture type

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -1749,32 +1749,6 @@ namespace dmGraphics
         return true;
     }
 
-    static inline uint32_t GetGraphicsTypeSize(Type type)
```
Jhonnyg, line 1752: Replaced with function that does the same thing from graphics.cpp

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -2474,6 +2474,27 @@ namespace dmGraphics
         vkCreatePipelineLayout(context->m_LogicalDevice.m_Device, &vk_layout_create_info, 0, &program->m_Handle.m_PipelineLayout);
     }
 
+    static void ResolveSamplerTextureUnits(Program* program, const dmArray<ShaderResourceBinding>&  texture_resources)
```
Jhonnyg, line 2477: This is needed for the "constructed" samplers from the pipeline. Since we now have to split combined samplers into texture + sampler, we need to be able to match the sampler with the texture, so that we can assign the same texture unit to both.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderUtil.java
```diff
@@ -388,6 +388,68 @@ public static class Result {
...
+
+            if(input.isEmpty()) {
+                return result;
+            }
+
+            ArrayList<String> output = new ArrayList<>(input.length());
+            ArrayList<String[]> lineReplacements = new ArrayList<>();
+            String[] inputLines = input.split("\\r?\\n");
+
+            for(String line : inputLines) {
```
JCash, line 411: Perhaps a future improvement is to exclude commented out code, as it might cause annoying behaviors if we e.g. potentially cause warnings/errors.
iirc, we've had similar issues with our Lua parsing before.

JCash (dismissed):

JCash (approved):

PR #9419, opened 2024-09-11
Retain gui overrides if the template gui scene is modified externally

### LINKED ISSUE #9418: Gui overrides lost after template scene modified externally
Overrides on gui nodes from a template scene that is modified using an external process will be lost after the editor reloads the template gui file.

**To Reproduce (REQUIRED)**
1. Add a template to a main gui scene and override some property on a child node in it.
2. Save the project.
3. Using an external text editor, make an unrelated change to the template gui file.
4. Return to the Defold Editor and Load External Changes.

The external changes to the template gui file are seen in the editor, and the property override remains in the main gui scene.

Seen in 1.6.2, 1.9.0, and 1.9.3.

Use an external text editor to change `max_nodes: 512` to a different value in `/main/button.gui`, then return to the Defold Editor and Load External Changes. You can see the overrides being stripped out of `/main/panel.gui`.

[dev-gui-overrides.zip](https://github.com/user-attachments/files/16966463/dev-gui-overrides.zip)

matgis:
Fixed an issue where gui node property overrides would be lost from referencing `.gui` files if the referenced template `.gui` file was modified by an external process.

Fixes #9418


editor/src/clj/editor/targets.clj
```diff
@@ -326,7 +326,7 @@
 
 (defn- show-error-message [exception workspace]
   (ui/run-later
-    (let [msg (str (.getMessage exception) "\n\n"
+    (let [msg (str (ex-message exception) "\n\n"
```
matgis, line 329: Unrelated: Fix reflection warning.

vlaaad (approved):
Ship it!

PR #9420, opened 2024-09-12
Improve Lua API documentation

ghost:
Updating type names to be more consistent, helps improve auto generated annotations. Related discussion here: https://forum.defold.com/t/defold-teal-annotations/77778/5

Updates:
1. b2d.body, camera, render - vmath. prefix on types, removed the prefix as this is not part of the type
2. vmath, gui - quat vs quaternion - inconsistently named, updated quat to quaternion
3. vmath - upper case first letters on types, changed to use lower case first letters
4. tilemap.set_tile - parameter transform-bitmask is not a valid name for a variable, renamed to transform_bitmask
5. timer.delay - parameter name repeat is a reserved word in Lua, renamed to repeating

### DIFF
--- a/engine/dlib/src/dmsdk/dlib/vmath.h
+++ b/engine/dlib/src/dmsdk/dlib/vmath.h
@@ -110,30 +110,30 @@ namespace dmVMath
 
     /*# dot product between two vectors
      * @name Dot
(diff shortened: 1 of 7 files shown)


britzl (approved):
Looks good. Thank you!

PR #9421, opened 2024-09-12
Allow creating dialogs from editor scripts

vlaaad:
This changeset adds an `editor.ui` module to editor scripts with various functionality around making interactive UIs. 

Technical notes:
Docs PR: https://github.com/defold/doc/pull/476

Related to #8062


editor/src/clj/editor/sentry.clj
```diff
@@ -136,7 +136,7 @@
 
 (extend-protocol json/JSONWriter
   java.time.LocalDateTime
-  (-write [object out] (json/-write (str object) out)))
+  (-write [object out options] (json/-write (str object) out options)))
```
vlaaad, line 139: Breaking change in `clojure.data.json`.

editor/src/clj/editor/workspace.clj
```diff
@@ -50,16 +51,12 @@ ordinary paths."
...
-
-(defn load-class! [class-name]
-  (Class/forName class-name true class-loader))
-
 (def build-dir "/build/default/")
 (def plugins-dir "/build/plugins/")
 
+;; SDK api
+(def load-class! java/load-class!)
+
```
vlaaad, line 59: Turns out I can't just remove `load-class!` from the workspace ns because [other libraries use it](https://github.com/search?q=workspace%2Fload-class%21&type=code). They don't use `class-loader` though!

editor/src/java/com/defold/editor/luart/DefoldUserdata.java
```diff
@@ -29,6 +29,6 @@ public DefoldUserdata(Object obj, Object stringRepresentation) {
 
     @Override
     public String tojstring() {
-        return stringRepresentation + ": " + Integer.toHexString(Objects.hashCode(stringRepresentation));
+        return stringRepresentation + ": " + Integer.toHexString(Objects.hashCode(userdata()));
```
vlaaad, line 32: Just fixed a minor boo-boo: the hash is supposed to help differentiate various user data objects.

editor/styling/stylesheets/_dialogs.scss
```diff
@@ -6,7 +6,7 @@ $spacing-largest: 40px;
 
 @use "sass:string";
 @mixin font-size-default { -fx-font-size: 110%; }
-@mixin font-size-header { -fx-font-size: 170%; }
+@mixin font-size-header { -fx-font-size: 22px; }
```
vlaaad, line 9: 170% was almost 22px, but not quite. I made headers consistent across dialogs, markdown, and now editor scripts.

editor/styling/stylesheets/components/_markdown.scss
```diff
@@ -4,9 +4,9 @@
     -fx-font-size: 13.5px;
     & > Text {-fx-fill: -df-text;}
     & > .md {
-      &-h1 { -fx-font-size: 30px; }
-      &-h2 { -fx-font-size: 26px; }
-      &-h3 { -fx-font-size: 22px; }
+      &-h1 { -fx-font-size: 26px; -fx-font-family: "Source Sans Pro Light"; }
+      &-h2 { -fx-font-size: 24px; -fx-font-family: "Source Sans Pro Light"; }
+      &-h3 { -fx-font-size: 22px; -fx-font-family: "Source Sans Pro Light"; }
```
vlaaad, line 9: Changed for consistency with editor scripts and dialogs.

scripts/build.py
```diff
@@ -1486,7 +1486,9 @@ def build_docs(self):
         skip_tests = '--skip-tests' if self.skip_tests or self.target_platform != self.host else ''
         self._log('Building API docs')
         cwd = join(self.defold_root, 'engine/docs')
-        cmd = '%s %s/ext/bin/waf configure --prefix=%s %s distclean configure build install' % (' '.join(self.get_python()), self.dynamo_home, self.dynamo_home, skip_tests)
+        python_cmd = ' '.join(self.get_python())
+        cmd = '%s %s/ext/bin/waf configure --prefix=%s %s distclean configure build install' % (python_cmd, self.dynamo_home, self.dynamo_home, skip_tests)
+        run.env_command(self._form_env(), [python_cmd, './scripts/bundle.py', 'docs', '--docs-dir', '%s/share/doc' % self.dynamo_home], cwd = join(self.defold_root, 'editor'))
```
vlaaad, line 1491: Now there will be editor docs in the reference!

editor/src/clj/editor/editor_extensions.clj
```diff
@@ -463,45 +496,56 @@
...
   (g/with-auto-evaluation-context evaluation-context
     (let [extensions (g/node-value project :editor-extensions evaluation-context)
           old-state (ext-state project evaluation-context)
-          project-path (-> project
-                           (project/workspace evaluation-context)
-                           (workspace/project-path evaluation-context)
+          workspace (project/workspace project evaluation-context)
+          project-path (-> (workspace/project-path workspace evaluation-context)
                            .toPath
                            .normalize)
```
matgis, line 512: Should it use `.toRealPath` instead of `.normalize`? To be honest, I kind of thing this is the wrong place to normalize the project path. 🤷

editor/src/clj/editor/future.clj
```diff
@@ -29,14 +29,36 @@
...
          (get [~'_]
            (with-bindings bindings#
              ~@body))))))
 
+(def io-executor
+  (let [counter (atom 0)]
+    (Executors/newCachedThreadPool
+      (reify ThreadFactory
+        (newThread [_ r]
+          (doto (Thread. r)
```
matgis, line 47: Reflection warning?

editor/src/clj/editor/future.clj
```diff
@@ -29,14 +29,36 @@
...
+  "Asynchronously perform an IO-intensive operation (see also: compute)"
+  [& body]
+  `(let [bindings# (get-thread-bindings)]
+     (CompletableFuture/supplyAsync
+       (reify Supplier
+         (get [~'_]
+           (with-bindings bindings#
+             ~@body)))
+       io-executor)))
+
```
matgis, line 61: Should these respect `editor.disk-availability/avaliable?` and block until for instance, a save operation completes?

editor/src/clj/editor/fxui.clj
```diff
@@ -310,6 +347,7 @@
     (future
       (error-reporting/catch-all!
         (let [result @result-promise]
+          (fx/unmount-renderer state-atom renderer)
```
matgis, line 350: Does this need to be called inside the `on-fx-thread` block, since it calls the renderer with `nil`?

editor/src/clj/editor/fxui.clj
```diff
@@ -326,25 +364,30 @@
...
-    state of a dialog has `::result` key in it, dialog interaction is considered
-    complete and dialog will close
-  - `:description` (required) - fx description used for this dialog, gets merged
-    into current state map, meaning that state map contents, including
-    eventually a `::result` key, will also be present in description props. You
-    can use `editor.fxui/dialog-showing?` and pass it resulting props to check
-    if dialog stage's `:showing` property should be set to true"
-  [& {:keys [initial-state event-handler description]
-      :or {initial-state {}}}]
+  (which is checked by presence of a ::result key in state map)
```
matgis, line 367: in the state map

editor/src/clj/editor/fxui.clj
```diff
@@ -326,25 +364,30 @@
...
+    :description      required, fx description used for this dialog, gets merged
+                      into current state map, meaning that state map contents,
+                      including eventually a ::result key, will also be present
+                      in description props. Use `editor.fxui/dialog-showing?`
+                      and pass it resulting props to check if dialog stage's
+                      :showing property should be set to true
+    :initial-state    optional, defaults to {}, map containing initial state of
+                      a dialog, should not contain ::result key to be shown
+    :error-handler    optional, 1-arg Throwable handler, reports to sentry by
+                      default"
```
matgis, line 384: I think it also shows the exception dialog, in fact.

editor/src/clj/editor/lua.clj
```diff
@@ -311,67 +231,93 @@
...
+         :examples "```\neditor.delete_directory(\"/assets/gen\")\n```"}
+        {:name "editor.external_file_attributes"
+         :type :function
+         :description "Query information about file system path"
+         :parameters [{:name "path"
+                       :types ["string"]
+                       :doc "External file path, resolved against project root if relative"}]
+         :returnvalues [{:name "attributes"
+                         :types ["table"]
+                         :doc "A table with following keys:<dl>
```
matgis, line 300: following keys -> the following keys

editor/src/clj/editor/protobuf.clj
```diff
@@ -374,11 +373,11 @@ Macros currently mean no foreseeable performance gain, however."
...
                    (case (:field-rule field-info)
-                     :repeated [ [[key]] ]
-                     :required [ [key] ]
+                     :repeated [ [[key]]]
+                     :required [ [key]]
                      :optional (if-some [field-default (declared-default class key)]
-                                 [ [{key field-default}] ]
-                                 [ [key] ]))
+                                 [ [{key field-default}]]
+                                 [ [key]]))
```
matgis, line 380: Hmm, these nested vectors got a bit messed up.

editor/src/clj/editor/protobuf.clj
```diff
@@ -649,7 +648,7 @@ Macros currently mean no foreseeable performance gain, however."
                                      (->CamelCase (FilenameUtils/getBaseName (.getName file))))
                          inner-cls (desc-name desc)]
                      (str package "." outer-cls "$" inner-cls)))]
-    (workspace/load-class! cls-name)))
+    (java/load-class! cls-name)))
```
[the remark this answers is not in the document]
matgis: Love it!

editor/src/java/com/defold/control/DefoldStringConverter.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public DefoldStringConverter(IFn toString, IFn fromString) {
+        this.toString = toString;
+        this.fromString = fromString;
+    }
+
+    @Override
+    public String toString(Object o) {
+        return String.valueOf(toString.invoke(o));
```
matgis, line 21: Shouldn't the `toString` `IFn` already return a string? Seems like we're calling `.toString` twice on the object, and maybe a cast would be sufficient?

editor/styling/stylesheets/modules/_extensions.scss
```diff
@@ -0,0 +1,134 @@
...
+    }
+    &-heading {
+        &-1 { -fx-font-size: 26px; -fx-font-family: "Source Sans Pro Light"; }
+        &-2 { -fx-font-size: 24px; -fx-font-family: "Source Sans Pro Light"; }
+        &-3 { -fx-font-size: 22px; -fx-font-family: "Source Sans Pro Light"; }
+        &-4 { -fx-font-size: 20px; }
+        &-5 { -fx-font-size: 18px; }
+        &-6 { -fx-font-size: 16px; }
+    }
+    &-typography-variant {
```
matgis, line 30: There is that word again. 😬

scripts/build.py
matgis: Can we do it here, though? I thought we relied on the output of this for the `.sdoc` files in `editor/resources/docs` that end up in the editor. But perhaps we don't need a fully formed editor jar to generate the editor script docs?

matgis (changes requested):

matgis:
I found some issues, a few things I dislike with the API, and some minor niggles, but spectacular job overall! ❤️

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+
+(def ^:private dialog-props
+  [(make-prop :title
+              :coerce coerce/string
+              :required true
+              :doc "OS dialog window title")
+   (make-prop :header
+              :coerce child-coercer
+              :types ["component"]
+              :doc "top part of the dialog")
```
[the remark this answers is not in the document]
vlaaad: 👍

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+
+;; endregion
+
+;; region component definitions
+
+(defn- props-doc-html [props]
+  (lua-completion/args-doc-html
+    (map #(update % :name name) props)))
+
+(s/def :editor.editor-extensions.components.component/name string?)
```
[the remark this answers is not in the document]
vlaaad: Yeah, I forgot to fix this after renaming the namespace.

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+(def scroll-component
+  (component
+    "scroll"
+    :description "Layout container that optionally shows scroll bars if child contents overflow the assigned bounds"
+    :props scroll-props))
+
+(def label-component
+  (component
+    "label"
+    :description "Non-resizeable text label"
```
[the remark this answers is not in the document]
vlaaad: Yes, good point.

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+    :props label-props))
+
+(def paragraph-component
+  (component
+    "paragraph"
+    :description "Typography component representing a paragraph of text"
```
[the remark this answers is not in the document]
vlaaad: 👍

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+    :props paragraph-props))
+
+(def heading-component
+  (component
+    "heading"
+    :description "Typography component representing a heading"
```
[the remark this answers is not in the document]
vlaaad: 👍

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+    :props heading-props))
+
+(def icon-component
+  (component
+    "icon"
+    :description "An icon of a fixed size"
```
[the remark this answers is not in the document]
vlaaad: 👍

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+  (component
+    "select_box"
+    :props select-box-props
+    :description "Dropdown select box with an array of options"))
+
+(def text-field-component
+  (component
+    "text_field"
+    :props text-field-props
+    :description "Single-line text field, reports changes on typing"))
```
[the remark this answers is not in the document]
vlaaad: I thought we agreed to not do it 😅 
As we discussed, the components have different behavior and are intended for different purposes. You can't reproduce a `text_field` by adding an `on_text_changed` callback to the `string_field`, because it will have behaviors that you don't want. Specifically, you don't want to have "commit on Enter" and "rollback to previous value on Escape" behaviors in a `text_field`. You want these different components in different scenarios: 
- `text_field` for immediate feedback on any typing, e.g. for immediate validation of new file names.
- `value_field` and friends for feedback on explicit commit action, e.g. for number form inputs where intermediate text is often invalid (like `-`, `1.`), but immediate feedback is only distracting.

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+
+(def value-field-component
+  (component
+    "value_field"
+    :props generic-value-field-props
+    :description "Input component based on a text field, reports changes on commit (<code>Enter</code> or focus loss).\n\nSee also: <code>string_field</code>, <code>number_field</code> and <code>integer_field</code>"))
```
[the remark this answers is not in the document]
vlaaad: What do you mean? What did we change? It still reports changes on commit.

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+  (component
+    "dialog_button"
+    :props dialog-button-props
+    :description "Dialog button shown in the footer of a dialog"))
+
+(def dialog-component
+  (component
+    "dialog"
+    :props dialog-props
+    :description "Dialog component, a special component that can't be used as a child of other components"))
```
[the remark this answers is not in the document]
vlaaad: 👍

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+
+;; endregion
+
+;; region docs
+
+(defn- component->script-doc [{:keys [name props description]}]
+  (let [{req true opt false} (group-by :required props)]
```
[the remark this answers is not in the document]
vlaaad: Hmm okay. A bit inconvenient that it returns empty coll instead of `nil` if there are no items in some bucket...

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+     :description description
+     :parameters [{:name "props"
+                   :types ["table"]
+                   :doc (str (when req
+                               (str "Required props:\n"
+                                    (props-doc-html req)
+                                    "\n\n"))
+                             "Optional props:\n"
+                             (props-doc-html opt))}]
+     :returnvalues [{:name "value"
```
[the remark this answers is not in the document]
vlaaad: Yes, we did, from `sdoc` files. We also use these maps to write the editor's script doc files, so it's just more convenient to use this keyword despite the readability issues. Otherwise, we would need to convert it back and forth.

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+                     :types ["component"]
+                     :doc "UI component"}]}))
+
+(def show-dialog-doc
+  {:name "show_dialog"
+   :type :function
+   :description "Show a dialog and await for result"
```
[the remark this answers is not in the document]
vlaaad: ✅

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+   :parameters [{:name "dialog"
+                 :types ["component"]
+                 :doc "a component that resolves to <code>editor.ui.dialog(...)</code>"}]
+   :returnvalues [{:name "value"
+                   :types ["any"]
+                   :doc "dialog result"}]})
```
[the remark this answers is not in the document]
vlaaad: 👍

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+                 :types ["function"]
+                 :doc "function, will receive a single table of props when called"}]
+   :returnvalues [{:name "value"
+                   :types ["function"]
+                   :doc "decorated component function that may be invoked with a props table create component"}]})
+
+(def show-external-file-dialog-doc
+  {:name "show_external_file_dialog"
+   :type :function
+   :description "Show OS file selection dialog and await for result"
```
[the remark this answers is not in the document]
vlaaad: 👍

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+                          :types ["table[]"]
+                          :doc external-file-dialog-filters-doc}])}]
+   :returnvalues [{:name "value"
+                   :types ["string" "nil"]
+                   :doc "either absolute file path or nil if user canceled file selection"}]})
+
+(def show-external-directory-dialog-doc
+  {:name "show_external_directory_dialog"
+   :type :function
+   :description "Show OS directory selection dialog and await for result"
```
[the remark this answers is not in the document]
vlaaad: 👍

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+(def ^:private resource-dialog-extensions-doc-string
+  "if specified, restricts selectable resources in the dialog to specified file extensions; e.g. <code>{\"collection\", \"go\"}</code>")
+
+(def ^:private resource-dialog-title-doc-string
+  "dialog title, defaults to <code>\"Select Resource\"</code>")
+
+(def show-resource-dialog-doc
+  {:name "show_resource_dialog"
+   :type :function
+   :description "Show editor's resource selection dialog and await for result"
```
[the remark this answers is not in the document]
vlaaad: 👍

editor/src/clj/editor/future.clj
vlaaad: Hmm no, I don't see any reflection warnings, and there is `(set! *warn-on-reflection* true)` at the top. This is probably because `r` is type-hinted in the `reify` since `newThread` defines a `Runnable` arg.

editor/src/clj/editor/future.clj
vlaaad: No, this is in no way related to editor disc availability. `io` is used for blocking tasks, for example, for file system access or for piping a process output of a process launched by the editor script.

editor/src/clj/editor/fxui.clj
vlaaad: No, `renderer`, unlike JavaFX, is thread-safe: it already calls `on-fx-thread`.

editor/src/clj/editor/fxui.clj
vlaaad: Yes, also `the presence` 😅

editor/src/clj/editor/fxui.clj
vlaaad: 👍

editor/src/clj/editor/lua.clj
vlaaad: 👍

editor/src/clj/editor/protobuf.clj
vlaaad: It's only training space removed by parinfer... I'll remove leading space too for symmetry.

editor/src/java/com/defold/control/DefoldStringConverter.java
vlaaad: Yes!

scripts/build.py
vlaaad: Hmm yeah, we will end up with an outdated `editor_doc.sdoc` in the editor. I'll make sure to exclude it! Yes, we don't need the editor's `editor_doc.sdoc` in the editor, since the editor code is a source of truth for the editor doc.

vlaaad (commented):
I'll start fixing various things that you pointed out, meanwhile lets discuss renaming the `TYPOGRAPHY_VARIANT` enum :)

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+(def enums
+  {:alignment [:top-left :top :top-right :left :center :right :bottom-left :bottom :bottom-right]
+   :padding [:small :medium :large]
+   :spacing [:small :medium :large]
+   :text_alignment [:left :center :right :justify]
+   :icon_name [:open-resource :plus :minus :clear]
+   :orientation [:vertical :horizontal]
+   :input_variant [:default :warning :error]
+   :label_variant [:default :overridden]
+   :typography_variant [:default :hint :warning :error]})
```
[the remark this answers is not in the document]
vlaaad: I changed my mind. I'll make them kebab-cased instead. `:SCREAMING_SNAKE_CASE` keywords are hard to read in Clojure. BTW, the initial reasoning for making them snake-cased was to make `enum-prop` fn easier to use. It needs both a prop key and an enum ID, and I thought it would be nice to specify just one thing (prop name) and the enum will be automatically picked up. For example, `:text_alignment` prop has `:text_alignment` enum. But I think this is unnecessary complexity. It's no problem to specify both.

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -0,0 +1,718 @@
...
+
+(def ^:private get-enum-coercer
+  (let [m (coll/pair-map-by key #(apply coerce/enum (val %)) enums)]
+    (fn get-enum-coercer [kw]
+      {:post [(some? %)]}
+      (m kw))))
+
+(defn ->screaming-snake-case [kw]
+  (string/replace (util/upper-case* (name kw)) "-" "_"))
```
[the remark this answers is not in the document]
vlaaad: On the other hand, we probably want to memoize it, because this function is used not only on editor start but also on editor scripts reload.

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -211,8 +237,7 @@
 
 (def ^:private heading-props
   (-> typography-specific-props
-      (conj (make-prop :level :coerce (coerce/enum 1 2 3 4 5 6) :types ["integer"]
-                       :doc "Heading level, an integer from 1 to 6, defaults to 3 (intended value for headings in dialog headers)"))
+      (conj (enum-prop :style :enum :heading-style :doc "heading style, defaults to <code>editor.ui.HEADING_STYLE.DIALOG</code>"))
```
matgis, line 240: It is a bit weird that we're defaulting to `HEADING_STYLE.DIALOG`. Maybe we should change the default to something a bit more neutral like `HEADING_STYLE.H3`, seeing as how they are identical?

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -227,40 +252,54 @@
...
+                    :coerce (coerce/one-of
+                              (coerce/hash-map :req {:severity (coerce/enum :error :warning)
+                                                     :message coerce/string})
+                              absent-coercer)
+                    :types ["table"]
+                    :doc (str "issue related to the input; table with the following keys (all required):"
+                              (lua-completion/args-doc-html
+                                [{:name "severity"
+                                  :types ["string"]
+                                  :doc "either <code>\"error\"</code> or <code>\"warning\"</code>"}
```
matgis, line 274: Should these be `editor.ui.ISSUE_SEVERITY` enums rather than strings?

editor/styling/stylesheets/editor.sass
```diff
@@ -48,6 +48,7 @@
 @import 'components/_text-area'
 @import 'components/_text-field'
 @import 'components/_titled-pane'
+@import 'components/_tooltip'
```
matgis, line 51: Huh. So we had this before, but never imported it? 🤔

editor/test/resources/editor_extensions/resource_attributes_project/test.editor_script
```diff
@@ -1,10 +1,10 @@
 local M = {}
 
 local function test_resource(resource_path)
-    local success, exists = pcall(editor.resource_exists, resource_path)
+    local success, attrs = pcall(editor.resource_attributes, resource_path)
     local message
     if success then
-        message = exists and "exists" or "does not exist"
+        message = attrs.exists and ("exists (" .. (attrs.is_file and "file" or "directory") .. ")" ) or "does not exist"
```
matgis, line 7: We should explicitly check the value of `attrs.is_directory` here as well. Maybe even ensure their values are boolean?

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -58,24 +87,20 @@
   {:alignment [:top-left :top :top-right :left :center :right :bottom-left :bottom :bottom-right]
    :padding [:small :medium :large]
    :spacing [:small :medium :large]
```
matgis, line 89: I noticed the default value for `:spacing` varies between `editor.ui.horizontal` and `editor.ui.grid`. I think we should consistently use `editor.ui.SPACING.MEDIUM` for both, but we probably need to add `editor.ui.SPACING.NONE` and `editor.ui.PADDING.NONE` so you can explicitly disable spacing and padding?

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -58,24 +87,20 @@
...
-   :text_alignment [:left :center :right :justify]
-   :icon_name [:open-resource :plus :minus :clear]
+   :text-alignment [:left :center :right :justify]
+   :icon-name [:open-resource :plus :minus :clear]
```
matgis, line 91: Is `editor.ui.ICON_NAME.OPEN_RESOURCE` used for both resource and external file fields? Should its name reflect that?

Also, should it be renamed to `editor.ui.ICON`? We don't call the color enum `editor.ui.COLOR_NAME`.

matgis (changes requested):
Nice work!

I found some inconsistencies around defaults. We should make sure the default spacing and padding values are consistent among all the layout component, for example.

editor/styling/stylesheets/editor.sass
vlaaad: Hehe, yes.

editor/src/clj/editor/editor_extensions/ui_docs.clj
vlaaad: Okay!

editor/test/resources/editor_extensions/resource_attributes_project/test.editor_script
vlaaad: Okay!

editor/src/clj/editor/editor_extensions/ui_docs.clj
vlaaad: Huh, I don't see that the defaults are different.
Horizontal and Vertical default to medium here:
https://github.com/defold/defold/blob/75b5a5020abe6f288b5a9be6f1c82bad2851d47e/editor/src/clj/editor/editor_extensions/ui_components.clj#L277-L279
Grid defaults to medium here:
https://github.com/defold/defold/blob/75b5a5020abe6f288b5a9be6f1c82bad2851d47e/editor/src/clj/editor/editor_extensions/ui_components.clj#L240-L241

matgis: Hmm, not sure where I got the idea that they were different to be honest. I must have caused it myself somehow when playing around with it. 😅

matgis (approved):
Good enough for now! Let's ship it! ❤️

Issue #9397, opened 2024-09-08
Tilemap brush selection error


sergeykozlovd:
### Steps to reproduce

<hr/>

<table>
<tr><td>Build time</td><td>2024-07-30T08:34:52.703441</td></tr>
<tr><td>Defold channel</td><td>editor-alpha</td></tr>
<tr><td>Defold editor sha</td><td>075b790f01b4b3d2074a7fd9dcf04b20af309eb0</td></tr>
<tr><td>Defold engine sha</td><td>691478c02875b80e76da65d2f5756394e7a906b1</td></tr>
<tr><td>Defold version</td><td>1.9.1</td></tr>
<tr><td>Error</td><td><a href='https://sentry.io/organizations/defold/issues/?query=id%3A"e8152a1a8c8d427e8a674dc7370419ab"'>e8152a1a8c8d427e8a674dc7370419ab</a></td></tr>
<tr><td>GPU</td><td>Apple M1</td></tr>
<tr><td>GPU Driver</td><td>2.1 Metal - 88.1</td></tr>
<tr><td>Java version</td><td>17.0.5+8</td></tr>
<tr><td>OS arch</td><td>aarch64</td></tr>
<tr><td>OS name</td><td>Mac OS X</td></tr>
<tr><td>OS version</td><td>14.5</td></tr>
<table>

AGulev:
```
clojure.lang.Numbers.ops.class java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "x" is null
    at clojure.lang.Numbers.ops(Numbers.java:1068)
    at clojure.lang.Numbers.remainder(Numbers.java:200)
    at clojure.core$mod.invokeStatic(core.clj:3571)
    at editor.tile_map$palette_x.invokeStatic(tile_map.clj:154)
    at editor.tile_map$make_brush_from_selection_in_palette.invokeStatic(tile_map.clj:162)
    at editor.tile_map$handle_input_palette.invokeStatic(tile_map.clj:1328)
    at editor.tile_map$handle_input.invokeStatic(tile_map.clj:1340)
    at editor.tile_map$make_input_handler$fn.invoke(tile_map.clj:1347)
    at editor.scene$dispatch_input$fn.invoke(scene.clj:1168)
    at clojure.lang.PersistentVector.reduce(PersistentVector.java:343)
    at clojure.core$reduce.invokeStatic(core.clj:6827)
    at editor.scene$dispatch_input.invokeStatic(scene.clj:1165)
[...]

PR #9429, opened 2024-09-14
Better handling of errors in tilemaps when the tilesource is broken.

AGulev:
If a tilemap uses a tilesource without a texture, it's impossible to show the palette, which previously resulted in an error.

Fix https://github.com/defold/defold/issues/9397


matgis (approved):
Ship it! ❤️

PR #9435, opened 2024-09-16
Refactor vertex attribute generation in engine

Jhonnyg:
Technical changes:

* Refactored vertex write functions to work according to the same ruleset as editor:
  - Converting from a scalar or vector to a smaller scalar or vector will truncate values from the end of the source value.
  - Converting from a scalar or vector to a larger vector will zero-fill at the end, unless the semantic-type is position, color or tangent, in which case the W component will be one.
  - Converting from a scalar or vector to a larger matrix will zero-fill at the end.
  - Converting from a matrix to a smaller matrix will use the top-left portion of the source matrix.
  - Converting from a matrix to a larger matrix will write into the top-left portion of the destination matrix, and fill the remaining space with the identity matrix.
  - Converting from a matrix to a vector will truncate values from the end of the source matrix.
* Added multiple tests that verifies the ruleset
* All components using the "custom vertex formats" (sprite, model, particles) are now using the same function to produce vertices
  - There is now a dependency between the shared libs and graphics
* All vertex attribute code has been moved into graphics_vertices.cpp for better overview and maintenance


engine/render/src/render/material.cpp
```diff
@@ -243,6 +227,52 @@ namespace dmRender
         m->m_MaterialAttributeValues.SetCapacity(num_attribute_byte_size);
         m->m_MaterialAttributeValues.SetSize(num_attribute_byte_size);
         memset(m->m_MaterialAttributeValues.Begin(), 0, num_attribute_byte_size);
+
+        // Assign default values based on semantic type:
+        //   * position and tangent gets 0,0,0,1 if a vec4 is used
+        //   * color gets 1,1,1,1
+        //   * world and normal matrix gets a 1 in all diagonals
+        for (int i = 0; i < num_program_attributes; ++i)
```
Jhonnyg, line 235: TODO:
I think we can remove the data backing for vertex attributes that isn't customized by the editor, since we produce vertex data based on the ruleset.

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1048,16 +1058,33 @@ namespace dmGameSystem
...
+                uint32_t uv_channels_count = (uv_channels[0] ? 1 : 0) + (uv_channels[1] ? 1 : 0);
+
+                dmGraphics::WriteAttributeParams params = {};
+                dmRig::SetMeshWriteAttributeParams(&params,
                     &attribute_infos,
+                    dmGraphics::VERTEX_STEP_FUNCTION_INSTANCE,
                     &instance_render_item->m_World,
                     &normal_matrix,
-                    &params);
+                    0, // World space positions are not supported by local space materials
```
Jhonnyg, line 1078: We circumvrent this later on by taking the "world coordinate" route in the batch function. If we have to produce world space positions for local spaced materials we would essentially need to do a third "mixed" approach that can update dynamic streams with static content (similar to the instance buffer). It's a lot of work to make that work, but maybe we can solve that later? The current setup will work, but not optimally I guess.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -142,7 +146,7 @@ namespace dmGameSystem
         // Used after resolving info from all textures
         dmhash_t                        m_AnimationID;                      // The animation of the driving atlas
         uint32_t                        m_Frames[MAX_TEXTURE_COUNT];        // The resolved frame indices
-        uint32_t                        m_PageIndices[MAX_TEXTURE_COUNT];
+        float                           m_PageIndices[MAX_TEXTURE_COUNT];
```
Jhonnyg, line 149: The attribute writer function only takes floats as input (mostly for simplicity), so I changed the type here so we can pass the pointer in directly, similar to texture coordinates.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -655,7 +659,8 @@ namespace dmGameSystem
         return dmGameObject::CREATE_RESULT_OK;
     }
 
-    static void EnsureSize(dmArray<float>& array, uint32_t size)
+    template <typename T>
+    static void EnsureSize(dmArray<T>& array, uint32_t size)
     {
         if (array.Capacity() < size) {
             array.OffsetCapacity(size - array.Capacity());
```
Jhonnyg, line 666: Todo: move to dlib :)

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1002,14 +1053,15 @@ namespace dmGameSystem
         return !data->m_UsesGeometries || data->m_Geometries[0]->m_TrimMode == dmGameSystemDDF::SPRITE_TRIM_MODE_OFF;
     }
 
-    static void ResolvePositionAndUVDataFromGeometry(TexturesData* data, dmArray<float>* scratch_pos, dmArray<float>* scratch_uvs, float* scratch_uv_ptrs[MAX_TEXTURE_COUNT],
-                                                            float scale_x, float scale_y, int reverse)
+    static void ResolvePositionAndUVDataFromGeometry(TexturesData* data,
+        dmArray<Vector4>& scratch_pos, dmArray<float>* scratch_uvs,
+        float* scratch_uv_ptrs[MAX_TEXTURE_COUNT], float scale_x, float scale_y, int reverse)
```
Jhonnyg, line 1058: The scratch pos is a vec4 now since we require that all transforms are done before passing in to the vertex write function. Our Vector3 format is actually a Vector4 so it's just simpler to use a Vector4 through and through.

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1372,6 +1398,22 @@ namespace dmGameSystem
...
+        // If a "local" material space material has a world position attribute, we use world space when
+        // producing the attributes instead of trying to shoehorn this into the current local space vertex generation code.
+        // This is a bit of a shortcut because this case it not very likely to be used.
+        dmGraphics::VertexAttributeInfoMetadata material_vertex_attribute_meta;
+        dmRender::GetMaterialProgramAttributeMetadata(material, &material_vertex_attribute_meta);
+        if (material_vertex_space == dmRenderDDF::MaterialDesc::VERTEX_SPACE_LOCAL && material_vertex_attribute_meta.m_HasAttributeWorldPosition)
+        {
+            return dmRenderDDF::MaterialDesc::VERTEX_SPACE_WORLD;
+        }
+        return material_vertex_space;
```
JCash, line 1414: 👍

engine/graphics/src/graphics_vertices.cpp
```diff
@@ -0,0 +1,510 @@
...
+                return value_write_ptr + 2;
+            case dmGraphics::VertexAttribute::TYPE_INT:
+                *((int32_t*) value_write_ptr) = (int32_t) dmMath::Clamp(value, -2147483648.0f, 2147483647.0f);
+                return value_write_ptr + 4;
+            case dmGraphics::VertexAttribute::TYPE_UNSIGNED_INT:
+                *((uint32_t*) value_write_ptr) = (uint32_t) dmMath::Clamp(value, 0.0f, 4294967295.0f);
+                return value_write_ptr + 4;
+            case dmGraphics::VertexAttribute::TYPE_FLOAT:
+                *((float*) value_write_ptr) = value;
+                return value_write_ptr + 4;
```
JCash, line 70: Future thing: is there float16 support that we want to use as well?

engine/particle/src/wscript
```diff
@@ -34,7 +34,7 @@ def build(bld):
                   includes = ['.', '../proto'],
                   target = 'particle_shared',
                   protoc_includes = '../proto',
-                  use = 'DDF_NOASAN DLIB_NOASAN SOCKET PROFILE_NULL_NOASAN GRAPHICS_PROTO_NOASAN',
+                  use = 'DDF_NOASAN DLIB_NOASAN SOCKET PROFILE_NULL_NOASAN PLATFORM_NULL GRAPHICS_NOASAN',
```
JCash, line 37: Just a reminder to test with `--with-asan` to make sure you don't need a `PLATFORM_NULL_NOASAN`

engine/particle/wscript
```diff
@@ -30,10 +30,11 @@ def configure(conf):
...
 
     conf.env.append_value('INCLUDES', "default/src")
-    conf.env['STLIB_DLIB_NOASAN']           = ['dlib_noasan']
-    conf.env['STLIB_DDF_NOASAN']            = ['ddf_noasan']
-    conf.env['STLIB_GRAPHICS_NULL']         = ['graphics_null', 'graphics_proto']
-    conf.env['STLIB_GRAPHICS_PROTO_NOASAN'] = ['graphics_proto_noasan']
+    conf.env['STLIB_DLIB_NOASAN']     = ['dlib_noasan']
+    conf.env['STLIB_DDF_NOASAN']      = ['ddf_noasan']
+    conf.env['STLIB_GRAPHICS_NULL']   = ['graphics_null', 'graphics_proto']
+    conf.env['STLIB_GRAPHICS_NOASAN'] = ['graphics', 'graphics_proto', 'graphics_proto_noasan']
```
JCash, line 36: This looks a bit off. Have you verified with `--with-asan`?
1) neither graphics or graphics_proto are using the "noasan" suffix, which means that it's going to be the "ASAN" linkage used. 
2) graphics_proto and graphics_proto_noasan are the same, just with different linkage?
If so , then it's a little bit random to add two versions of the same library on the command line.

JCash (approved):
Looks good, but the NOASAN setup seems a little strange. Is that path tested?

engine/graphics/src/graphics_vertices.cpp
Jhonnyg: Yes, I think there is half_float support for attributes. Could be a future improvement!

PR #9438, opened 2024-09-17
New vmath functions

### LINKED ISSUE #5118: Add more common functions to math, vmath library
I'm always frustrated when i have to reprogram usual mathematic functions like 
- vector.magnitude 
- the nearest point on a finite or infinite line
- mathf.clamp

Implement those functions into defolds common libraries math and vmath.

You can program those yourself but its quite tiring after a while.

example implementations:
```
function module.vector3magnitude(x,y,z)
	return math.sqrt((x*x+y*y+z*z))
end

function module.clamp(x,y,z)
	return math.max(y, math.min(z,x))
end

function module.nearestPointOnVectorInfinite(start, finish, point)
	finish = vmath.normalize(finish)
	local v = point - start
	local d = vmath.dot(v, finish)
	return start + finish * d
end

function module.nearestPointOnVectorFinite(start, finish, point)
	line_direction = finish-start
	line_length = module.vector3magnitude(line_direction.x, line_direction.y, line_direction.z)
	line_direction = vmath.normalize(line_direction)
	project_length = module.clamp(vmath.dot(point-start, line_direction),0,line_length)
	return start + line_direction * project_length
end
```

ekharkunov:
Added new `vmath` functions:
* `vmath.quat_matrix4` - get quaternion from matrix4
* `vmath.matrix4_compose` - create matrix4 from translation (vector3|vector4), rotation (quaternion) and scale (vector3)
* `vmath.matrix4_scale` - create matrix4 from scale (vector3) or from number (uniform scale) or 3 number (x, y, z)
* `vmath.clamp` - clamp input value in range [min, max]. Input value can be number, vector, vector3, vector4. In case of vector type clamp applies per vector component. `min`, `max` also can be number, vector, vector3, vector4.

Fixes #5118 

## PR checklist

* [x] Code
* [x] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/engine/script/src/script_vmath.cpp
+++ b/engine/script/src/script_vmath.cpp
@@ -2481,12 +2481,63 @@ namespace dmScript
      * print(result) --> vmath.matrix4(1, 0, 0, 0, 0, 0.5, 0, 0, 0, 0, 0.5, 0, 0, 0, 0, 1)
      * ```
      */
+
+    /*# creates a new matrix4 from uniform scale
+     *
+     * creates a new matrix4 from uniform scale
+     *
+     * @name vmath.matrix4_scale
+     * @param scale [type:number] scale
+     * @return matrix [type:matrix4] new matrix4
+     * @examples
+     *
+     * ```lua
+     * local result = vmath.matrix4_scale(0.5)
+     * print(result) --> vmath.matrix4(0.5, 0, 0, 0, 0, 0.5, 0, 0, 0, 0, 0.5, 0, 0, 0, 0, 1)
+     * ```
+     */
+
+    /*# creates a new matrix4 from three scale components
+     *
+     * Creates a new matrix4 from three scale components
+     *
+     * @name vmath.matrix4_scale
+     * @param scale_x [type:number] scale along X axis
+     * @param scale_y [type:number] sclae along Y axis
+     * @param scale_z [type:number] scale along Z asis
+     * @return matrix [type:matrix4] new matrix4
+     * @examples
+     *
+     * ```lua
+     * local result = vmath.matrix4_scale(1, 0.5, 0.5)
+     * print(result) --> vmath.matrix4(1, 0, 0, 0, 0, 0.5, 0, 0, 0, 0, 0.5, 0, 0, 0, 0, 1)
+     * ```
+     */
     static int Matrix4_Scale(lua_State* L)
     {
-        Vector3* scale = CheckVector3(L, 1);
-        Matrix4 result = Matrix4::scale(*scale);
-        PushMatrix4(L, result);
-        return 1;
+        if (lua_isnumber(L, 1))
+        {
+            float x, y, z;
+            x = (float) luaL_checknumber(L, 1);
+            if (lua_isnumber(L, 2) && lua_isnumber(L, 3))
+            {
+                y = (float) luaL_checknumber(L, 2);
+                z = (float) luaL_checknumber(L, 3);
+            }
+            else
+            {
+                y = x; z = x;
+            }
+            PushMatrix4(L, Matrix4::scale(Vector3(x, y, z)));
+        }
+        else
+        {
+            Vector3* scale = CheckVector3(L, 1);
+            Matrix4 result = Matrix4::scale(*scale);
+            PushMatrix4(L, result);
+            return 1;
+        }
+        return luaL_error(L, "First argument should be number or vector3");
     }
 
     /*# clamp input value in range [min, max] and return clamped value
--- a/engine/script/src/script_vmath.cpp
+++ b/engine/script/src/script_vmath.cpp
@@ -2519,7 +2519,7 @@ namespace dmScript
         {
             float x, y, z;
             x = (float) luaL_checknumber(L, 1);
-            if (lua_isnumber(L, 2) && lua_isnumber(L, 3))
+            if (lua_gettop(L) == 3)
             {
                 y = (float) luaL_checknumber(L, 2);
                 z = (float) luaL_checknumber(L, 3);
@@ -2542,26 +2542,24 @@ namespace dmScript
 
     /*# clamp input value in range [min, max] and return clamped value
      *
-     * Clamp input value to be in range of [min, max]. In case if input value has vector|vector3|vector4 type
-     * return new vector|vector3|vector4 with clamped value at every vector's element.
-     * Min/max arguments can be vector|vector3|vector4. In that case clamp excuted per every vector's element
+     * Clamp input value to be in range of [min, max]. In case if input value has vector3|vector4 type
+     * return new vector3|vector4 with clamped value at every vector's element.
+     * Min/max arguments can be vector3|vector4. In that case clamp excuted per every vector's element
      *
      * @name vmath.clamp
-     * @param value [type:number|vector3|vector4|vector] Input value or vector of values
-     * @param min [type:number|vector3|vector4|vector] Min value(s) border
-     * @param max [type:number|vector3|vector4|vector] Max value(s) border
-     * @return clamped_value [type:number|vector3|vector4|vector] Clamped value or vector
+     * @param value [type:number|vector3|vector4] Input value or vector of values
+     * @param min [type:number|vector3|vector4] Min value(s) border
+     * @param max [type:number|vector3|vector4] Max value(s) border
+     * @return clamped_value [type:number|vector3|vector4] Clamped value or vector
      * @example
      * 
      * ```lua
      * local value1 = 56
-	 * print(vmath.clamp(value1, 89, 134)) -> 89
-	 * local v1 = vmath.vector({ -45, 34, 13, 10, 55 })
-	 * print(vmath.clamp(v1, 0, 15)) -> vmath.vector (size: 5)
-	 * local v2 = vmath.vector3(190, 190, -10)
-	 * print(vmath.clamp(v2, -50, 150)) -> vmath.vector3(150, 150, -10)
-	 * local v3 = vmath.vector4(30, -30, 45, 1)
-	 * print(vmath.clamp(v3, 0, 20)) -> vmath.vector4(20, 0, 20, 1)
+     * print(vmath.clamp(value1, 89, 134)) -> 89
+     * local v2 = vmath.vector3(190, 190, -10)
+     * print(vmath.clamp(v2, -50, 150)) -> vmath.vector3(150, 150, -10)
+     * local v3 = vmath.vector4(30, -30, 45, 1)
+     * print(vmath.clamp(v3, 0, 20)) -> vmath.vector4(20, 0, 20, 1)
      * 
      * local min_v = vmath.vector4(0, -10, -10, 1)
      * print(vmath.clamp(v3, min_v, 20)) -> vmath.vector4(20, -10, 20, 1)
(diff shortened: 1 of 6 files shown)


engine/script/src/script_vmath.cpp
```diff
@@ -2299,6 +2299,166 @@ namespace dmScript
...
+     *
+     * ```lua
+     * local translation = vmath.vector3(103, -95, 14)
+     * local quat = vmath.quat(1, 2, 3, 4)
+     * local scale = vmath.vector3(1, 0.5, 0.5)
+     * local result = vmath.matrix4_compose(translation, quat, scale)
+     * print(result) --> vmath.matrix4(-25, -10, 11, 103, 28, -9.5, 2, -95, -10, 10, -4.5, 14, 0, 0, 0, 1)
+     * ```
+     */
+    static int Matrix4_Compose(lua_State* L)
```
JCash, line 2338: Potentially, it could be part of the constructor as well.
I.e. a `vmath.matrix4(t, r, s)` (and maybe a `vmath.matrix4(m00, m01, ...)`

Perhaps mostly for discoverability, but I have no strong opinion here.

engine/script/src/script_vmath.cpp
```diff
@@ -2299,6 +2299,166 @@ namespace dmScript
...
+        if (lua_type(L, 1) == LUA_TNUMBER)
+        {
+            float value = (float) luaL_checknumber(L, 1);
+            lua_pushnumber(L, dmMath::Clamp(value, min, max));
+        }
+        else
+        {
+            const ScriptUserType argument_type = GetType(L, 1);
+            if (argument_type == SCRIPT_TYPE_VECTOR3)
+            {
```
JCash, line 2425: This is a good scenario to use ToVector3 et al, and then early out.

```
if number:
  do stuff;
  return 1
  
 Vector3* pvec3 = ToVector3(L, 1)
 if pvec3:
    ...
    return 1
 
 return luaL_error();
 ```

File: engine/script/src/script_vmath.cpp [hunk 3]
```diff
@@ -2299,6 +2299,166 @@ namespace dmScript
...
+     * ```lua
+     * local scale = vmath.vector3(1, 0.5, 0.5)
+     * local result = vmath.matrix4_scale(scale)
+     * print(result) --> vmath.matrix4(1, 0, 0, 0, 0, 0.5, 0, 0, 0, 0, 0.5, 0, 0, 0, 0, 1)
+     * ```
+     */
+    static int Matrix4_Scale(lua_State* L)
+    {
+        Vector3* scale = CheckVector3(L, 1);
+        Matrix4 result = Matrix4::scale(*scale);
```
JCash, line 2384: I think having a 1/3 number(s) is another use case, as we can then skip necessarily creating a temporary vector3.

engine/script/src/script_vmath.cpp
```diff
@@ -2299,6 +2299,166 @@ namespace dmScript
...
+	 * local v2 = vmath.vector3(190, 190, -10)
+	 * print(vmath.clamp(v2, -50, 150)) -> vmath.vector3(150, 150, -10)
+	 * local v3 = vmath.vector4(30, -30, 45, 1)
+	 * print(vmath.clamp(v3, 0, 20)) -> vmath.vector4(20, 0, 20, 1)
+     * ```
+     */
+    static int Vector_Clamp(lua_State* L)
+    {
+        float min = (float) luaL_checknumber(L, 2);
+        float max = (float) luaL_checknumber(L, 3);
```
JCash, line 2415: Potentially, we could also have a "per element" clamp.
I'm not sure what the actual user requirements are here tbh.
We can probably leave it until someone requests it.

JCash (changes requested):
Looks good.
Minor name change for consistency. `vmath.quat_from_matrix4`

Also added Björn for api review.

engine/script/src/script_vmath.cpp
```diff
@@ -2299,6 +2299,166 @@ namespace dmScript
...
+     * @name vmath.quat_matrix4
+     * @param matrix [type:matrix4] source matrix4
+     * @return q [type:quaternion] new quaternion
+     */
+    static int Quat_Matrix4(lua_State* L)
+    {
+        Matrix4* matrix = CheckMatrix4(L, 1);
+        PushQuat(L, dmVMath::Quat(matrix->getUpper3x3()));
+        return 1;
+    }
```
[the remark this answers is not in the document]
ekharkunov: In that case I would suggest to rename that function also https://github.com/defold/defold/blob/11f05ad81aa18e68e3cc59cc26a9fc2886935508/engine/script/src/script_vmath.cpp#L2484C11-L2484C30
from `vmath.matrix4_translation` to `vmath.matrix4_from_translation`. Left old one for backward compatability and mark as deprecated.
And that one also rename https://github.com/defold/defold/blob/11f05ad81aa18e68e3cc59cc26a9fc2886935508/engine/script/src/script_vmath.cpp#L2500 to `vmath.matrix4_from_scale`.
WDYT?

engine/script/src/script_vmath.cpp
ekharkunov: I thought about making it as a part of constructor. But I refused that idea because in other cases (matrix4 from scale, quaternion from matrix4, matrix4 from translation) also should be done as part of constructor.
Maybe it can be generalized like
```lua
local m = vmath.matrix4({
    translation = vmath.vector3(),
    rotation = vmath.quat(1,1,1,0)
    scale = 4
}}

local m2 = vmath.matrix4(m)
local m3 = vmath.matrix4(vmath.vector({ .... })
```
it's a matter of had several but small and clear function vs have 1 function and hide all implementation under the hood.

engine/script/src/script_vmath.cpp
ekharkunov: In that case I would suggest to have two separate functions:
* `vmath.clamp(v, min, max)` for number only
* `vmath.vclamp(vec, min_vec, max_vec)` for vectors only

WDYT?

engine/script/src/script_vmath.cpp
britzl: It's not great to have to pass a table every time. It will generate quite a bit of garbage (or the user has to keep a table around only for this purpose)

engine/script/src/script_vmath.cpp
britzl: I prefer not to complicate it. vmath.clamp() for numbers and the elements of vectors will be very useful.

engine/script/src/script_vmath.cpp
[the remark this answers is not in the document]
britzl: The odd one out is Matrix_FromQuat():

```
{"matrix4_from_quat", Matrix4_FromQuat},
{"matrix4_axis_angle", Matrix4_AxisAngle},
{"matrix4_rotation_x", Matrix4_RotationX},
{"matrix4_rotation_y", Matrix4_RotationY},
{"matrix4_rotation_z", Matrix4_RotationZ},
{"matrix4_translation", Matrix4_Translation},
```

All of the others also create a matrix4 from something but none of them have "from" in the name.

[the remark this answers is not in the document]
ekharkunov: So rename `matrix4_from_quat` to `matrix4_quat` and left `quat_matrix4` as-is?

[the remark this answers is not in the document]
britzl: Yes, but keep matri4x_from_quat but remove it from the documentation

engine/script/src/script_vmath.cpp
```diff
@@ -2299,6 +2299,166 @@ namespace dmScript
...
+     * ```
+     */
+    static int Matrix4_Compose(lua_State* L)
+    {
+        Matrix4 translation_matrix = Matrix4::identity();
+        const ScriptUserType translation_type = GetType(L, 1);
+
+        if (translation_type == SCRIPT_TYPE_VECTOR3)
+        {
+            Vector3* translation = CheckVector3(L, 1);
```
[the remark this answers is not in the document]
ekharkunov: @JCash  Suggest following function
https://github.com/defold/defold/pull/9438/files#diff-15a7f3705bd4b3e12269cedf985da8656b010506cf18223a5ea6154253e096b0R138-R192

I've used it across all vmath functions that has variation across input arguments.

engine/script/src/script_vmath.cpp
```diff
@@ -2299,6 +2401,293 @@ namespace dmScript
...
+    {
+        if (lua_isnumber(L, 1))
+        {
+            float x, y, z;
+            x = (float) luaL_checknumber(L, 1);
+            if (lua_isnumber(L, 2) && lua_isnumber(L, 3))
+            {
+                y = (float) luaL_checknumber(L, 2);
+                z = (float) luaL_checknumber(L, 3);
+            }
```
JCash, line 2526: I think that if `top > 1` then we can use "check number" regardless?
We don't want to pass in other types, do we?

engine/script/src/script_vmath.cpp
```diff
@@ -2299,6 +2401,293 @@ namespace dmScript
...
+     * @param scale [type:vector3] scale
+     * @return matrix [type:matrix4] new matrix4
+     * @examples
+     *
+     * ```lua
+     * local scale = vmath.vector3(1, 0.5, 0.5)
+     * local result = vmath.matrix4_scale(scale)
+     * print(result) --> vmath.matrix4(1, 0, 0, 0, 0, 0.5, 0, 0, 0, 0, 0.5, 0, 0, 0, 0, 1)
+     * ```
+     */
```
JCash, line 2483: This is a new thing for us I think, having explicit documentation entries for the same function.
I think it's borderline, as it might be tricky to keep the documentation in sync that way. The alternative may be visually awkward (for the arguments) thought.
However, the examples would be grouped together.

Good enough for now at least.

engine/script/src/script_vmath.cpp
```diff
@@ -111,11 +111,100 @@ namespace dmScript
...
+    static inline bool CheckMatrix4Components(Matrix4* m)
+    {
+        return
+            !isnan(m->getElem(0, 0)) && !isnan(m->getElem(1, 0)) && !isnan(m->getElem(2, 0)) && !isnan(m->getElem(3, 0)) &&
+            !isnan(m->getElem(0, 1)) && !isnan(m->getElem(1, 1)) && !isnan(m->getElem(2, 1)) && !isnan(m->getElem(3, 1)) &&
+            !isnan(m->getElem(0, 2)) && !isnan(m->getElem(1, 2)) && !isnan(m->getElem(2, 2)) && !isnan(m->getElem(3, 2)) &&
+            !isnan(m->getElem(0, 3)) && !isnan(m->getElem(1, 3)) && !isnan(m->getElem(2, 3)) && !isnan(m->getElem(3, 3));
+    }
+
+    static ScriptUserType CheckUserData(lua_State* L, int index, void** userdata)
```
JCash, line 138: 👍

engine/script/src/script_vmath.cpp
```diff
@@ -2299,6 +2401,293 @@ namespace dmScript
...
+     *
+     * @name vmath.clamp
+     * @param value [type:number|vector3|vector4|vector] Input value or vector of values
+     * @param min [type:number|vector3|vector4|vector] Min value(s) border
+     * @param max [type:number|vector3|vector4|vector] Max value(s) border
+     * @return clamped_value [type:number|vector3|vector4|vector] Clamped value or vector
+     * @example
+     * 
+     * ```lua
+     * local value1 = 56
```
JCash, line 2557: Indentation

engine/script/src/script_vmath.cpp
```diff
@@ -2299,6 +2401,293 @@ namespace dmScript
...
+                max_v = *CheckVector4(L, 3);
+            }
+            PushVector4(L, Vector4(
+                dmMath::Clamp(value->getX(), min_v.getX(), max_v.getX()),
+                dmMath::Clamp(value->getY(), min_v.getY(), max_v.getY()),
+                dmMath::Clamp(value->getZ(), min_v.getZ(), max_v.getZ()),
+                dmMath::Clamp(value->getW(), min_v.getW(), max_v.getW())
+            ));
+        }
+        else if  (argument_type == SCRIPT_TYPE_VECTOR)
```
JCash, line 2643: While I do like the implementation, I also wonder about the use case.
I rarely see anyone using the Vector type, and I think I'd personally just wait until someone needs it before adding it.

We use the mantra "every bit counts", in terms of keeping the engine size as small as possible. Sometimes the change it warranted, and sometime we want to wait until someone asks for it. I believe this case might be the latter.
I don't have an overly strong opinion here.

@britzl what do you think?

engine/script/src/script_vmath.cpp
```diff
@@ -2409,7 +2803,7 @@ namespace dmScript
     Vector4* CheckVector4(lua_State* L, int index)
     {
         Vector4* v = (Vector4*)CheckUserType(L, index, TYPE_HASHES[SCRIPT_TYPE_VECTOR4], 0);
-        if (isnan(v->getX()) || isnan(v->getY()) || isnan(v->getZ()) || isnan(v->getW()))
+        if (!CheckVector4Components(v))
```
JCash, line 2806: 👍

engine/script/src/test/test_script_vmath.cpp
```diff
@@ -292,6 +294,10 @@ TEST_F(ScriptVmathTest, TestMatrix4Fail)
     ASSERT_FALSE(RunString(L, "local m = vmath.matrix4() * true"));
     // translation
     ASSERT_FALSE(RunString(L, "local m = vmath.matrix4_translation()"));
+    // scale
+    ASSERT_FALSE(RunString(L, "local m = vmath.matrix4_scale()"));
+    // compose
+    ASSERT_FALSE(RunString(L, "local m = vmath.matrix4_compose()"));
 }
```
JCash, line 303: I'd like to see some unit test tests for the new clamp function.

JCash (changes requested):
Questions regarding:
* do we want to support matrix_scale(1) ?
* do we want to support clamp() for vector() type ?
* missing unit tests for clamp

engine/script/src/script_vmath.cpp
ekharkunov: Not new. Initially I was searching for example how to describe one function with different set of arguments and than found that https://github.com/defold/defold/blob/dev/engine/script/src/script_vmath.cpp#L1064-L1120. And did in the same way.

engine/script/src/script_vmath.cpp
britzl: I agree. The Vector type is very rarely used, and if it is used it is likely only to describe an animation curve. I think we can remove it and add a note in the documentation about which types are not supported.

engine/script/src/script_vmath.cpp
ekharkunov: > I think that if `top > 1` then we can use "check number" regardless? We don't want to pass in other types, do we?

I think only vector 3 and set of numbers. Nothing else.

ekharkunov: > Another option is to always have 3 numbers?

You can pass 1 or 3 numbers. In case if only 1 number passed - it used as value for every scale component (x, y, z).

engine/script/src/script_vmath.cpp
ekharkunov: I've removed `clamp` implementation for `vector` type but left check for `SCRIPT_TYPE_VECTOR` in `CheckUserData` to cover all types defined in `ScriptUserType`.

engine/script/src/test/test_script_vmath.lua
```diff
@@ -0,0 +1,54 @@
+-- Copyright 2020-2024 The Defold Foundation
```
JCash, line 1: 👍

JCash (approved):

PR #9471, opened 2024-09-25
Fix editor apidocs generation

vlaaad:
The problem: on CI, the API docs are generated before Bob is even built. Because of that, it's not possible to launch the full editor source code. This PR refactors the editor code so that `editor.docs` ns does not depend on bob at all. Now, instead of producing protobuf text files, it produces an `editor.apidoc` file that is then used by the build system to produce sdoc files.

As a result, `:docs` profile is sort of a separate Clojure project.

Note: we validate that it's possible to generate docs in a bundling process. This helps us ensure that docs-generating code stays independent from bob.

The latest commit succeeds on both [editor](https://github.com/defold/defold/actions/runs/11033220376/job/30643871463) and [main](https://github.com/defold/defold/actions/runs/11033210804/job/30646423024) CI runs.
You can check that Main CI run produced a ref-doc with editor docs here: http://d.defold.com/archive/4945d1878f2a6fbf2cae5136405f3576fd603c6c/engine/share/ref-doc.zip


editor/src/clj/editor/os.clj
```diff
@@ -0,0 +1,32 @@
...
+  (memoize os-raw))
+
+(defn is-mac-os? []
+  (= (os) :macos))
+
+(defn is-linux? []
+  (= (os) :linux))
+
+(defn is-win32? []
+  (= (os) :win32))
```
vlaaad, line 32: I moved bob dep from `editor.util` to `editor.os`

matgis (approved):
Looks good to me!

I did notice some excessive newlines in the resulting `.script_api` file, not sure if it is cause for concern:
```
  - desc: 'Create a directory if it does not exist, and all non-existent parent directories.


      Throws an error if the directory can''t be created.'
    name: create_directory
    notes: []
    parameters:
    - desc: Resource path (starting with <code>/</code>) of a directory to create
      name: resource_path
      type: string
    type: function
```

PR #9474, opened 2024-09-25
The physics contact_point_event now returns the correct contact point

### LINKED ISSUE #9454: Complement contact_point_event by including the contact position
When using `physics.set_listener()`, the `contact_point_event` doesn't return the contact position. The `contact_point_response` message does return it, but requires a .script file to be added to each game object with a collision object attached.

I'd prefer to use `physics.set_listener()` as it can be more performant when using many game objects, as well as keeping all collision code in one place.

The current data structure returned by `contact_point_event` looks like this:

```
{
    applied_impulse = n,
    distance = n,
    a = {},
    b = {},
}
```

It would make sense to include the contact position, too:

```
{
    position = vector3,
    applied_impulse = number,
    distance = number,
    a = {},
    b = {},
} 
```

Using the game object position, rather than the contact position.

JCash:
This is a bug fix for the `contact_point_event`, where the `position` is now the contact point itself.
This makes it similar to the `contact_point_response` event, which was the intention.

If you need the old position (the world position of the object itself), you can use either `message.a.instance_position`/`message.b.instance_position`.

Fixes https://github.com/defold/defold/issues/9454

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_collision_object.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_collision_object.cpp
@@ -689,6 +689,7 @@ namespace dmGameSystem
                 a.m_Group               = group_hash_a;
                 a.m_Id                  = instance_a_id;
                 a.m_Position            = contact_point.m_PositionA;
+                a.m_InstancePosition    = dmGameObject::GetWorldPosition(instance_a);
                 a.m_Mass                = mass_a;
                 a.m_RelativeVelocity    = -contact_point.m_RelativeVelocity;
                 a.m_Normal              = -contact_point.m_Normal;
@@ -697,6 +698,7 @@ namespace dmGameSystem
                 b.m_Group               = group_hash_b;
                 b.m_Id                  = instance_b_id;
                 b.m_Position            = contact_point.m_PositionB;
+                a.m_InstancePosition    = dmGameObject::GetWorldPosition(instance_b);
                 b.m_Mass                = mass_b;
                 b.m_RelativeVelocity    = contact_point.m_RelativeVelocity;
                 b.m_Normal              = contact_point.m_Normal;
--- a/engine/gamesys/src/gamesys/components/comp_collision_object.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_collision_object.cpp
@@ -688,7 +688,7 @@ namespace dmGameSystem
                 dmPhysicsDDF::ContactPoint& a = ddf.m_A;
                 a.m_Group               = group_hash_a;
                 a.m_Id                  = instance_a_id;
-                a.m_Position            = dmGameObject::GetWorldPosition(instance_a);
+                a.m_Position            = contact_point.m_PositionA;
                 a.m_Mass                = mass_a;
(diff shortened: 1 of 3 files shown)


engine/gamesys/src/gamesys/components/comp_collision_object.cpp
```diff
@@ -688,15 +688,15 @@ namespace dmGameSystem
...
-                a.m_Position            = dmGameObject::GetWorldPosition(instance_a);
+                a.m_Position            = contact_point.m_PositionA;
                 a.m_Mass                = mass_a;
                 a.m_RelativeVelocity    = -contact_point.m_RelativeVelocity;
                 a.m_Normal              = -contact_point.m_Normal;
 
                 dmPhysicsDDF::ContactPoint& b = ddf.m_B;
                 b.m_Group               = group_hash_b;
                 b.m_Id                  = instance_b_id;
-                b.m_Position            = dmGameObject::GetWorldPosition(instance_b);
```
AGulev, line 699: Don't we need instance positions as well as it is in `dmPhysicsDDF::ContactPointResponse` ?

JCash: I'm not sure.
The corresponding "response" message has
* position = contact point
* other_position = world point of other object

So it doesn't have the actual object position (only for the other object)
If I _did_ add the object position in the struct, what would the name be? object_position ?
It's not a big deal, but I went with "less is more" for now?

AGulev: using messages we send two of them to each object, that's why we use such terms like "other_position". But it's only one event for a callback.

so, yes, I think it should be `instance_position` or `object_possition`. 
Don't forget to change documentation either: 
https://github.com/defold/defold/blob/9a87526fe7ff093886e6575f2927300053dd0559/engine/gamesys/src/gamesys/scripts/script_physics.cpp#L1561-L1582

https://github.com/defold/defold/blob/9a87526fe7ff093886e6575f2927300053dd0559/engine/gamesys/proto/gamesys/physics_ddf.proto#L186-L197

JCash: "using messages we send two of them to each object"
I was referring to this fact, that the message itself doesn't actually have its _own_ instance position, only the _other_ position.

But, for ease of use I think it's good to add a instance position to each sub struct.

AGulev: >I was referring to this fact, that the message itself doesn't actually have its own instance position, only the other position.

yes, because each of them kinda know it's own position, and only get the other position. There are two ways to solve it: provide both because even isn't in object, so it needs both or provide none (may be requested by id as I understand?)

AGulev (approved):

britzl:
> * Make sure that manuals are updated (in github.com/defold/doc)

Reminder to update documentation too!

https://defold.com/manuals/physics-events/

PR #9475, opened 2024-09-25
Http cache improvements

### LINKED ISSUE #4116: Loading resources is slow if engine is restarted via editor
If I press Cmd+B while the game is already running, it takes forever to restart and any subsequent collection load also takes a long time. If I close the engine, then Cmd+B it’s fine.

After some digging I found that loading resources via the network is unjustifiably slow (60MB/s in Interrogation and 10MB/s for the attached repro case) and CPU-bottlenecked (dmengine uses 100% CPU).

Repro-case on forum: https://forum.defold.com/t/loading-resources-is-slow-if-engine-is-restarted-via-editor/53361

Jhonnyg:
Several improvements to the HTTP support in the engine has been made:
* Restarting a project from the editor should be significantly faster for large projects
* HTTP requests via the `http.request` function for large files should also be significantly faster
* HEAD requests in the engine should be working correctly now - you can use a HEAD request to get the size of a remote file

Fixes #4116


editor/src/clj/editor/hot_reload.clj
```diff
@@ -40,12 +41,16 @@
...
+                             0
+                             ;; Returns zero if not found.
+                             (.length file))
+            content (when (and (= method "GET") (not is-cached))
                       (try
-                        (with-open [is (io/input-stream file)]
-                          (IOUtils/toByteArray is))
-                        (catch FileNotFoundException _
+                        (Files/readAllBytes (Paths/get (.toURI file)))
+                        (catch IOException _
```
Jhonnyg, line 53: Using readAllBytes seems faster for large files due not doing copying between iostreams

editor/src/clj/editor/hot_reload.clj
```diff
@@ -40,12 +41,16 @@
...
+                             ;; that contains a content-length other than zero.
+                             0
+                             ;; Returns zero if not found.
+                             (.length file))
+            content (when (and (= method "GET") (not is-cached))
                       (try
-                        (with-open [is (io/input-stream file)]
-                          (IOUtils/toByteArray is))
-                        (catch FileNotFoundException _
+                        (Files/readAllBytes (Paths/get (.toURI file)))
```
vlaaad, line 52: You could use `(.toPath file)` instead of `(Paths/get (.toURI file))` to simplify this a bit.

vlaaad (dismissed):
LGTM!

vlaaad (approved):
Perfect!

JCash (approved):
Looks good.
Could perhaps use some tests?

PR #9480, opened 2024-09-26
Improve messages when constants and attributes are missing from the materials and compute programs

### LINKED ISSUE #6503: Assigning value to material constants fails if they were optimized away from shader

### LINKED ISSUE #9241: Unclear error when a shader properties is optimized away
When a vertex attribute is not used (even down the chain of shader programs) and is optimized away by the compiler, then the sprite property (from the material) is also gone. And it will result error when `go.set()` in Lua (even tho it's defined in the material).

It is very normal during development to define attributes and set properties before actually using them. In my case I added the vertex attribute, added it to the vertex program into a varying for the fragment shader but didn't use that yet. Then I went to set the property on the sprite using go.set and it failed. I then spend a couple of hours debugging what went wrong, not knowing the compile would optimize it away (even tho it was used in the vertex shader).

Related: if a vertex attribute or fragment constant is of the wrong type then passed by go.set, it also errors that the property does not exists (instead of complaining about the type).

**There are a couple of solutions possible:**
1. Ignore the go.set() on the property if it's defined in the material but gone in the shaders.
2. Change the error to hint that shader properties could be optimized away.
3. Add a note to the [material documentation](https://defold.com/manuals/material/).

**Steps to reproduce:**
1. Fetch the example **sprite-normalized-local-positions.zip** from [this comment in issue #7045](https://github.com/defold/defold/issues/7045#issuecomment-2082081887).
2. Open it in Defold 1.9.1 (latest version at the time of writing)
3. Verify it compiles are runs fine.
4. Remove the line `gl_FragColor.rg = var_position_local.xy;` from the fragment program.
5. Run the game again, and it should fail with `'#sprite' does not have any property called 'sprite_size'`.  
   <small>(note: sprite_size is still defined in the material, and it still defined and used in the vertex program)</small>

### LINKED ISSUE #9463: Incorrect error for sprite tint property with vec3 parameter
Providing a vec3 for sprite property "tint" displays a misleading error.

**To Reproduce (REQUIRED)**
If you provide a vec3 value for "colour_value" in this line;
```lua
go.set("#sprite", "tint", colour_value)
```
You get the error:
```
ERROR:SCRIPT: tank.script:39: '#sprite' does not have any property called 'tint'
```
This is misleading, because sprite does indeed have a property called "tint".

The error should've been something like "tint requires a vec4"

  - Version 1.9.1

**Platforms (REQUIRED):**
Probably happens on all platforms. I can produce this when building on Linux.

Jhonnyg:
When a material or compute program has specified constants, attributes and samplers in the material editor but are either missing, not used or optimized away from the shaders we now show a warning. 

Fixes #9463 
Fixes #9241
Fixes #6503

### DIFF
--- a/editor/resources/templates/template.cp
+++ b/editor/resources/templates/template.cp
@@ -4,7 +4,10 @@ layout (local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
 
 layout(rgba32f) uniform image2D texture_out;
 
-uniform vec4 color;
+uniform uniforms
+{
+    vec4 color;
+};
 
 void main()
 {
--- a/engine/render/src/render/render.h
+++ b/engine/render/src/render/render.h
@@ -51,9 +51,10 @@ namespace dmRender
     typedef struct BufferedRenderBuffer*    HBufferedRenderBuffer;
     typedef HOpaqueHandle                   HRenderCamera;
 
-    static const uint8_t RENDERLIST_INVALID_DISPATCH    = 0xff;
-    static const HRenderType INVALID_RENDER_TYPE_HANDLE = ~0ULL;
-    static const uint32_t INVALID_SAMPLER_UNIT          = 0xffffffff;
+    static const uint8_t RENDERLIST_INVALID_DISPATCH       = 0xff;
+    static const HRenderType INVALID_RENDER_TYPE_HANDLE    = ~0ULL;
+    static const uint32_t INVALID_SAMPLER_UNIT             = 0xffffffff;
+    static const uint8_t  INVALID_MATERIAL_ATTRIBUTE_INDEX = 0xff;
 
     /**
      * Display profiles handle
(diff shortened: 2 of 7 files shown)


editor/resources/templates/template.cp
```diff
@@ -4,7 +4,10 @@ layout (local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
 
 layout(rgba32f) uniform image2D texture_out;
 
-uniform vec4 color;
+uniform uniforms
+{
+    vec4 color;
+};
```
Jhonnyg, line 10: This shader is using the new pipeline, which doesn't allow non-opaque uniforms to be outside a block.

engine/render/src/render/render.h
```diff
@@ -153,6 +154,15 @@ namespace dmRender
         uint8_t          m_OrthographicProjection : 1;
     };
 
+    struct MaterialProgramAttributeInfo
```
Jhonnyg, line 157: I just moved it to the "struct" section.

JCash (approved):

PR #9482, opened 2024-09-26
Add reload extensions shortcut

vlaaad:


matgis (approved):

vlaaad:
@Insality the default hotkey for opening assets is cmd+p. The old one was for compatibility with editor 1.0 (based on ellipse). We don't really have the luxury to have multiple shortcuts for the same action.

Insality:
Oh, I already found the another hotkey, all is good!

Sadly, but I have a years defold habit to use "Shift Meta R" hotkey. As I know I can override these hotkeys on user side. Or just make another habit :)

vlaaad:
Sorry that we broke your workflow!

PR #9483, opened 2024-09-26
Fix the build report in JSON format.

AGulev:
Fix the issue where the build report in JSON format is not being generated.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
@@ -672,16 +672,28 @@ public class Bob {
         String rootDirectory = getOptionsValue(cmd, 'r', cwd);
         String sourceDirectory = getOptionsValue(cmd, 'i', ".");
 
-
-        if (cmd.hasOption("build-report") || cmd.hasOption("build-report-html")) {
-            List<File> reportFiles = new ArrayList<>();
-            String jsonReportPath = cmd.getOptionValue("build-report");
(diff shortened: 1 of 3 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1818,7 +1818,7 @@ private List<TaskResult> runTasks(IProgress monitor) throws IOException {
                 }
 
                 TimeProfiler.start(task.getName());
-                TimeProfiler.addData("output", task.getOutputsString());
+                TimeProfiler.addData("output", StringUtil.truncate(task.getOutputsString(), 1000));
```
AGulev, line 1821: this one isn't directly related to that particular issue, but it's nice to truncate strings which can be huge

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -672,16 +672,28 @@ private static void mainInternal(String[] args) throws IOException, CompileExcep
         String rootDirectory = getOptionsValue(cmd, 'r', cwd);
         String sourceDirectory = getOptionsValue(cmd, 'i', ".");
 
+        String build_report_json = null;
+        String build_report_html = null;
```
AGulev, line 676: We don't have a project here, but we want to have time profiler initialized as soon as possible, so that's why this back compatibility code has been written this way.

britzl (approved):

PR #9496, opened 2024-09-27
Expose a shared entry point usable both by the editor and Bob

vlaaad:
Related to #8062

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
@@ -26,8 +26,9 @@ import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.FileSystemException;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
-import java.util.Arrays;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -57,7 +58,7 @@ import com.dynamo.bob.cache.ResourceCacheKey;
 import com.dynamo.bob.util.FileUtil;
 
 public class Bob {
-    private static Logger logger = Logger.getLogger(Bob.class.getName());
+    private static final Logger logger = Logger.getLogger(Bob.class.getName());
 
     public static final String VARIANT_DEBUG = "debug";
     public static final String VARIANT_RELEASE = "release";
@@ -67,6 +68,24 @@ public class Bob {
 
     private static File rootFolder = null;
     private static boolean luaInitialized = false;
+
+    public static class OptionValidationException extends Exception {
+        public final int exitCode;
+
+        public OptionValidationException(int exitCode) {
+            this.exitCode = exitCode;
+        }
+    }
+
+    public static class InvocationResult {
+        public final boolean success;
+        public final List<TaskResult> taskResults;
+
+        public InvocationResult(boolean success, List<TaskResult> taskResults) {
+            this.success = success;
+            this.taskResults = taskResults;
+        }
+    }
 
     public Bob() {
     }
@@ -467,7 +486,7 @@ public class Bob {
         }
     }
 
-    private static Options getCommandLineOptions() {
+    public static Options getCommandLineOptions() {
         Options options = new Options();
         addOption(options, "r", "root", true, "Build root directory. Default is current directory", true);
         addOption(options, "o", "output", true, "Output directory. Default is \"build/default\"", false);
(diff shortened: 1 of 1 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -68,6 +69,24 @@ public class Bob {
     private static File rootFolder = null;
     private static boolean luaInitialized = false;
 
+    public static class OptionValidationException extends Exception {
+        public final int exitCode;
+
+        public OptionValidationException(int exitCode) {
+            this.exitCode = exitCode;
+        }
+    }
```
vlaaad, line 78: Exception with intended exit code instead of calling `System.exit()`.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -68,6 +69,24 @@ public class Bob {
...
+
+    public static class InvocationResult {
+        public final boolean success;
+        public final List<TaskResult> taskResults;
+
+        public InvocationResult(boolean success, List<TaskResult> taskResults) {
+            this.success = success;
+            this.taskResults = taskResults;
+        }
+    }
```
vlaaad, line 88: The `success` flag is needed for them `main` to pick the right exit code. The `taskResults` field will be used by the editor to create the build errors.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -467,7 +486,7 @@ private static void addOption(Options options, String shortOpt, String longOpt,
         }
     }
 
-    private static Options getCommandLineOptions() {
+    public static Options getCommandLineOptions() {
```
vlaaad, line 489: made public so that the editor will be able to use command line options to define editor script API to bob with docs and validation.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -549,34 +568,34 @@ private static Options getCommandLineOptions() {
...
+            throw new OptionValidationException(1);
         }
 
         return cmd;
     }
 
-    private static Project createProject(String rootDirectory, String buildDirectory, String email, String auth) {
-        Project project = new Project(new DefaultFileSystem(), rootDirectory, buildDirectory);
+    private static Project createProject(ClassLoader classLoader, String rootDirectory, String buildDirectory, String email, String auth) {
+        Project project = new Project(classLoader, new DefaultFileSystem(), rootDirectory, buildDirectory);
```
vlaaad, line 598: The editor sets its own classloader, while the bob doesn't specify any. Made it an arg.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -597,7 +616,7 @@ else if (severity == MultipleCompileException.Info.SEVERITY_WARNING)
         return String.format("%s: %s:%d: '%s'\n", strSeverity, resourceString, line, message);
     }
 
-    private static void setupProject(Project project, boolean resolveLibraries, String sourceDirectory) throws IOException, LibraryException, CompileExceptionError {
+    private static void setupProject(Project project, boolean resolveLibraries, IProgress progress, IResourceScanner resourceScanner) throws IOException, LibraryException, CompileExceptionError {
```
vlaaad, line 619: The editor provides its own `resourceScanner`.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -695,26 +721,23 @@ else if (path.endsWith(".html")) {
             if (build_report_html != null) {
                 reportFiles.add(new File(build_report_html));
             }
-            TimeProfiler.init(reportFiles, false);
+            TimeProfiler.init(reportFiles, fromEditor);
```
AGulev, line 724: we need to remove second `TimeProfiler.init()` if `InvocationResult` is the only entry point we have for both the editor and the engine.

AGulev (approved):
Great!

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
vlaaad: Yeah, I thought about that. For now, we still have another entry point in the editor. When we change the editor to only use `Bob.invoke`, we could remove that `TimeProfiler.init()` call.

PR #9497, opened 2024-09-27
Deprecate the `--input` (`-i`) option in `bob.jar`

### LINKED ISSUE #9495: Figure out the use of `--input` in bob.
**Task (REQUIRED):**
We have an `-i`/`--input` command line argument in Bob. This commit removed the only use of this argument: https://github.com/defold/defold/commit/608c26c797492781c9f4162dfd3ac76d6e523912#diff-ea0d2982212f1d66d942bc5b6fe2aa39bf5f8e04a598c033fde540ecad14d6e3L613-L616 (in `Bob.java`). So it seems, previously, we had the "input" directory separately from the "root" directory (it seems this is the dir where we load the `game.project` and its deps from now on).

We should figure out what to do with this. Deprecate/remove `--input` argument? Start using it again?

AGulev:
The `--input` (`-i`) option for `bob.jar` is deprecated and can no longer be used. If you want to specify a project folder to be built, use the `--root` option instead.

Fix https://github.com/defold/defold/issues/9495


vlaaad (dismissed):
LGTM!

vlaaad (approved):

PR #9499, opened 2024-09-27
Added Switch Pro Controller for Windows

wistpotion:
Added Switch Pro Controller for Windows.

Technical notes
Created the mapping file with gdc 1.9.3.

Gdc reports the name of the gamepad as "Wireless Gamepad", and the defold gamepad tester example project stops recognizing input if you change the name to anything else. Windows settings menu reports it as "Pro Controller".


britzl (changes requested):
Thanks! See note about a comment

britzl (approved):

PR #9510, opened 2024-09-29
Make the building artifacts independent for the `Build HTML5` and `Build` options

AGulev:
From now on, the `Build HTML5` and `Build` options produce separate build artifact folders in the `build` directory. This means alternating between these options will no longer require rebuilding the project from scratch each time, speeding up the workflow for HTML5 testing.

### DIFF
--- a/editor/src/clj/editor/pipeline/bob.clj
+++ b/editor/src/clj/editor/pipeline/bob.clj
@@ -236,7 +236,8 @@
           {:error {:causes (engine-build-errors/unsupported-platform-error-causes project evaluation-context)}}
           (let [ws (project/workspace project evaluation-context)
                 proj-path (str (workspace/project-path ws evaluation-context))
-                bob-project (Project. java/class-loader (DefaultFileSystem.) proj-path "build/default")]
+                output-path (or (get bob-args "output") "build/default")
+                bob-project (Project. java/class-loader (DefaultFileSystem.) proj-path output-path)]
             (doseq [[key val] bob-args]
               (.setOption bob-project key val))
             (when-not (string/blank? build-server-headers)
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -236,7 +236,8 @@
           {:error {:causes (engine-build-errors/unsupported-platform-error-causes project evaluation-context)}}
           (let [ws (project/workspace project evaluation-context)
                 proj-path (str (workspace/project-path ws evaluation-context))
-                bob-project (Project. java/class-loader (DefaultFileSystem.) proj-path "build/default")]
```
AGulev, line 239: Use `output` if we specify one.

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -381,12 +382,13 @@
...
-(defn- build-html5-output-path [project]
+(defn- build-html5-output-path
+  ([project]
   (let [ws (project/workspace project)
-        build-path (workspace/build-path ws)]
-    (io/file build-path "__htmlLaunchDir")))
+        build-path (workspace/build-html5-path ws)]
+    (io/file build-path "__htmlLaunchDir"))))
 
-(def build-html5-bob-commands ["distclean" "build" "bundle"])
```
AGulev, line 389: Do not use `distclean` so we don't need to rebuild everything from scratch every time.

editor/src/clj/editor/workspace.clj
```diff
@@ -52,6 +52,7 @@ ordinary paths."
 (set! *warn-on-reflection* true)
 
 (def build-dir "/build/default/")
+(def build-html5-dir "/build/default_html5/")
```
AGulev, line 55: Use a special path for html5

matgis (approved):
LGTM! ❤️

PR #9506, opened 2024-09-28
Improved output if the sdk check fails

JCash:
We improved the output in case of missing sdk paths.

You can do this check manually using:
`./scripts/build.py check_sdk --platform=arm64-android`

If it succeeds it won't print anything. You can however make it verbose:
`./scripts/build.py check_sdk --platform=arm64-android --verbose`
and it will output something like:
```
SDK info:
{'api': '21',
 'build_tools': '/Users/mawe/Library/android/sdk/build-tools/35.0.0',
 'ndk': '/Users/mawe/Library/android/sdk/ndk/25.2.9519653',
 'sdk': '/Users/mawe/Library/android/sdk',
 'version': '35.0.0'}
'check_sdk' completed in 0.00 s
```

## PR checklist

* [ ] Code
	* [ ] Add engine and/or editor unit tests.
	* [ ] New and changed code follows the overall code style of existing code
	* [ ] Add comments where needed
* [ ] Documentation
	* [ ] Make sure that API documentation is updated in code comments
	* [ ] Make sure that manuals are updated (in github.com/defold/doc)
* [ ] Prepare pull request and affected issue for automatic release notes generator
	* [ ] Pull request - Write a message that explains what this pull request does. What was the problem? How was it solved? What are the changes to APIs or the new APIs introduced? This message will be used in the generated release notes. Make sure it is well written and understandable for a user of Defold.
	* [ ] Pull request - Write a pull request title that in a sentence summarises what the pull request does. Do not include "Issue-1234 ..." in the title. This text will be used in the generated release notes.
	* [ ] Pull request - Link the pull request to the issue(s) it is closing. Use on of the [approved closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
	* [ ] Affected issue - Assign the issue to a project. Do not assign the pull request to a project if there is an issue which the pull request closes.
	* [ ] Affected issue - Assign the "breaking change" label to the issue if introducing a breaking change.
	* [ ] Affected issue - Assign the "skip release notes" is the issue should not be included in the generated release notes.

----------

Example of a well written PR description:

1. Start with the user facing changes. This will end up in the release notes.
1. Add one of the GitHub approved closing keywords
1. Optionally also add the technical changes made. This is information that might help the reviewer. It will not show up in the release notes. Technical changes are identified by a line starting with one of these:
   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```
[...]

PR #9515, opened 2024-09-30
Update waf_dynamo.py Fix Android Build on Windows OS

OscarLeif:
Fix Android Build on Windows OS

Windows must use the .cmd files for the CC and CXX variables and for AR the file must be llvm-ar.exe windows need to use .exe file.

Finally in the strip ANDROID_HOST windows case was missing.
Builds fails when CCache is available. Must disable when build on windows.


JCash:
Thanks!
I hadn't noticed the strip function yet!

I'm also doing the exact same thing, so I'll close this PR in favor of these two:
* https://github.com/defold/defold/pull/9506
* https://github.com/defold/defold/pull/9517

The last PR will contrain a few more fixes for the android.jar as well (I'm working on it now)

PR #9521, opened 2024-10-02
Make the editor use Bob as it's used on the command line

vlaaad:
Related to #8062

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
@@ -31,6 +31,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
@@ -616,7 +617,7 @@ public class Bob {
         return String.format("%s: %s:%d: '%s'\n", strSeverity, resourceString, line, message);
     }
 
-    private static void setupProject(Project project, boolean resolveLibraries, IProgress progress, IResourceScanner resourceScanner) throws IOException, LibraryException, CompileExceptionError {
+    private static void setupProject(Project project, boolean resolveLibraries, IProgress progress) throws IOException, LibraryException, CompileExceptionError {
         BobProjectProperties projectProperties = project.getProjectProperties();
         String[] dependencies = projectProperties.getStringArrayValue("project", "dependencies");
         List<URL> libUrls = new ArrayList<>();
@@ -630,7 +631,7 @@ public class Bob {
             project.resolveLibUrls(progress);
             TimeProfiler.stop();
         }
-        project.mount(resourceScanner);
+        project.mount(new ClassLoaderResourceScanner());
     }
 
     private static void validateChoices(String optionName, String value, List<String> validChoices) throws OptionValidationException {
@@ -684,7 +685,7 @@ public class Bob {
      * Common entry point for both editor and bob
      * @param classLoader may be null
      */
-    public static InvocationResult invoke(IResourceScanner resourceScanner, ClassLoader classLoader, IProgress progress, boolean fromEditor, String[] args) throws IOException, CompileExceptionError, LibraryException, OptionValidationException {
+    public static InvocationResult invoke(ClassLoader classLoader, IProgress progress, boolean fromEditor, Map<String, String> internalOptions, String[] args) throws IOException, CompileExceptionError, LibraryException, OptionValidationException {
         System.setProperty("java.awt.headless", "true");
         System.setProperty("file.encoding", "UTF-8");
 
@@ -696,7 +697,6 @@ public class Bob {
         }
         String buildDirectory = getOptionsValue(cmd, 'o', "build/default");
         String rootDirectory = getOptionsValue(cmd, 'r', cwd);
-        String sourceDirectory = getOptionsValue(cmd, 'i', ".");
 
         String build_report_json = null;
         String build_report_html = null;
@@ -798,7 +798,7 @@ public class Bob {
 
         TimeProfiler.start("setupProject");
         // resolves libraries and finds all sources
-        setupProject(project, shouldResolveLibs, progress, resourceScanner);
+        setupProject(project, shouldResolveLibs, progress);
         TimeProfiler.stop();
 
         TimeProfiler.start("setOptions");
@@ -840,6 +840,9 @@ public class Bob {
                     project.setOption(o.getLongOpt(), "true");
                 }
             }
+        }
+        if (internalOptions != null) {
+            internalOptions.forEach(project::setOption);
         }
 
         // Get and set architectures list.
@@ -987,7 +990,7 @@ public class Bob {
 
     public static void main(String[] args) throws IOException, CompileExceptionError, URISyntaxException, LibraryException {
         try {
-            boolean success = invoke(new ClassLoaderResourceScanner(), null, new ConsoleProgress(), false, args).success;
+            boolean success = invoke(null, new ConsoleProgress(), false, null, args).success;
             System.exit(success ? 0 : 1);
         } catch (OptionValidationException e) {
             System.exit(e.exitCode);
--- a/editor/scripts/bundle.py
+++ b/editor/scripts/bundle.py
@@ -283,7 +283,7 @@ def extract_build_jdk(build_jdk):
 def invoke_lein(args, jdk_path=None):
     # this weird dance with env and bash instead of supplying env kwarg to run.command is needed for the build script to work on windows
     jdk_path = jdk_path or os.environ['JAVA_HOME']
-    return run.command(['env', 'JAVA_CMD=%s/bin/java' % jdk_path, 'LEIN_HOME=target/lein', 'bash', './scripts/lein'] + args)
+    return run.command(['env', 'JAVA_CMD=%s/bin/java' % jdk_path, 'LEIN_HOME=build/lein', 'bash', './scripts/lein'] + args)
 
 def check_reflections(jdk_path):
     reflection_prefix = 'Reflection warning, ' # final space important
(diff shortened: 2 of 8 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -630,7 +631,7 @@ private static void setupProject(Project project, boolean resolveLibraries, IPro
             project.resolveLibUrls(progress);
             TimeProfiler.stop();
         }
-        project.mount(resourceScanner);
+        project.mount(new ClassLoaderResourceScanner());
```
vlaaad, line 634: Both the editor and Bob can use the same ClassLoaderResourceScanner (it's used to load resources from builtins).

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -696,7 +697,6 @@ public static InvocationResult invoke(IResourceScanner resourceScanner, ClassLoa
         }
         String buildDirectory = getOptionsValue(cmd, 'o', "build/default");
         String rootDirectory = getOptionsValue(cmd, 'r', cwd);
-        String sourceDirectory = getOptionsValue(cmd, 'i', ".");
```
vlaaad, line 699: unused code from deprecated option.

editor/scripts/bundle.py
```diff
@@ -283,7 +283,7 @@ def extract_build_jdk(build_jdk):
 def invoke_lein(args, jdk_path=None):
     # this weird dance with env and bash instead of supplying env kwarg to run.command is needed for the build script to work on windows
     jdk_path = jdk_path or os.environ['JAVA_HOME']
-    return run.command(['env', 'JAVA_CMD=%s/bin/java' % jdk_path, 'LEIN_HOME=target/lein', 'bash', './scripts/lein'] + args)
+    return run.command(['env', 'JAVA_CMD=%s/bin/java' % jdk_path, 'LEIN_HOME=build/lein', 'bash', './scripts/lein'] + args)
```
vlaaad, line 286: I moved lein install to `build` because some lein commands clear the `target` dir, which causes the CI to re-download lein multiple times per run.

editor/src/clj/editor/app_view.clj
```diff
@@ -1318,12 +1318,11 @@ If you do not specifically require different script states, consider changing th
           render-save-progress! (make-render-task-progress :save-all)
           render-build-progress! (make-render-task-progress :build)
           task-cancelled? (make-task-cancelled-query :build)
-          build-server-headers (native-extensions/get-build-server-headers prefs)
-          bob-args (bob/build-html5-bob-args project prefs)
+          bob-args (bob/build-html5-bob-options project prefs)
```
vlaaad, line 1321: build server headers are now assigned to Bob options directly, inside the `build-html5-bob-options` fn.

editor/src/clj/editor/engine/native_extensions.clj
```diff
@@ -266,15 +267,14 @@
          :extender-platform extender-platform}
         (let [extender-client (ExtenderClient. build-server-url cache-directory)
               user-info (.getUserInfo (URI. build-server-url))
-              headers (into [] (remove string/blank?) (string/split-lines build-server-headers))
```
vlaaad, line 269: Now, `build-server-headers` is already a vector of headers.

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -118,56 +114,11 @@
...
-        (if-let [r (get res-map path)]
-          (= (resource/source-type r) :file)
-          false))
-      (scan [this pattern]
-        (let [res (->> res-map
-                    (map first)
-                    (filter #(PathUtil/wildcardMatch % pattern))
-                    set)]
-          res)))))
-
```
vlaaad, line 143: There is no need for this code since it's used only for loading built-ins, and Bob already has an instrument to do it.

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -118,56 +114,11 @@
...
-(defn- run-commands! [project evaluation-context ^Project bob-project commands render-progress! task-cancelled?]
-  (try
-    (let [result (ui/with-progress [render-progress! render-progress!]
-                   (.build bob-project (->progress render-progress! task-cancelled?) (into-array String commands)))
-          failed-tasks (filter (fn [^TaskResult r] (not (.isOk r))) result)]
-      (if (empty? failed-tasks)
-        nil
-        {:error {:causes (engine-build-errors/failed-tasks-error-causes project evaluation-context failed-tasks)}}))
-    (catch Exception e
-      {:exception e})))
```
vlaaad, line 157: Inlined this in `invoke!`

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -118,56 +114,11 @@
...
-    ::build-in-progress-watch
-    (fn build-in-progress-watch-fn [_ _ _ new-state]
-      (try
-        (set-verbose-logging! new-state)
-        (catch Exception _
-          ;; We can safely ignore any errors here since we've already
-          ;; thrown an exception in case we're unable to change it to
-          ;; false at the top of this file.
-          nil)))))
+(defonce ^:private build-in-progress-atom (atom false))
```
vlaaad, line 121: I changed this to not force the verbose option on every bob invocation, but instead to do it only when the verbose option is set to true (which is `invoke!`'s default).

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -180,92 +131,148 @@
...
-            (doseq [pkg ["com.dynamo.bob" "com.dynamo.bob.pipeline"]]
-              (ClassLoaderScanner/scanClassLoader java/class-loader pkg))
-            (let [deps (workspace/dependencies ws)]
-              (when (seq deps)
-                (.setLibUrls bob-project (map #(.toURL ^URI %) deps))
-                (ui/with-progress [render-progress! render-progress!]
-                  (.resolveLibUrls bob-project (->progress render-progress! task-cancelled?)))))
-            (.mount bob-project (->graph-resource-scanner ws))
-            (ui/with-progress [render-progress! render-progress!]
-              (run-commands! project evaluation-context bob-project bob-commands render-progress! task-cancelled?))))
```
vlaaad, line 255: All this code is replaced with a `Bob/invoke` call that does the same. Caveats: 
1. we need to add `resolve` to a list of commands so that Bob does the `(.resolveLibUrls bob-project ...)` call.
2. we no longer ensure `(native-extensions/supported-platform? (get bob-args "platform"))`: this code is basically outdated because Bob also does the same, but better — the editor's hardcoded platforms don't include console plugins.

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -302,25 +310,25 @@
             ;; From BundleGenericHandler
             generate-debug-symbols? (assoc "with-symbols" "")
             generate-build-report? (assoc "build-report-html" build-report-path)
-            publish-live-update-content? (assoc "liveupdate" "true"))))
+            publish-live-update-content? (assoc "liveupdate" "yes"))))
```
vlaaad, line 313: Bob's command line expects `--liveupdate` option to be either `"yes"` or `"no"`, and then `Bob/invoke` converts it to `"true"`/`"false"` and sets it as a project option. So now we need to supply `"yes"`.

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -180,93 +131,148 @@
...
-            (doseq [pkg ["com.dynamo.bob" "com.dynamo.bob.pipeline"]]
-              (ClassLoaderScanner/scanClassLoader java/class-loader pkg))
-            (let [deps (workspace/dependencies ws)]
-              (when (seq deps)
-                (.setLibUrls bob-project (map #(.toURL ^URI %) deps))
-                (ui/with-progress [render-progress! render-progress!]
-                  (.resolveLibUrls bob-project (->progress render-progress! task-cancelled?)))))
-            (.mount bob-project (->graph-resource-scanner ws))
-            (ui/with-progress [render-progress! render-progress!]
-              (run-commands! project evaluation-context bob-project bob-commands render-progress! task-cancelled?))))
```
vlaaad, line 256: All this code is replaced with a `Bob/invoke` call that does the same. Caveats: 
1. we need to add `resolve` to a list of commands so that Bob does the `(.resolveLibUrls bob-project ...)` call.
2. we no longer ensure `(native-extensions/supported-platform? (get bob-args "platform"))`: this code is basically outdated because Bob also does the same, but better — the editor's hardcoded platforms don't include console plugins.

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -331,83 +338,83 @@
...
-             "variant" "debug"
-             "archive" "true"
-             "output" (subs workspace/build-html5-dir 1)
-             "bundle-output" (str output-path)
-             "build-server" build-server-url
-             "defoldsdk" defold-sdk-sha1
-             "local-launch" "true"
-             "email" ""
-             "auth" ""}
-            compress-archive? (assoc "compress" "true"))))
```
vlaaad, line 410: `"compress"` is a completely unused option. No such thing in Bob.

matgis (approved):
Looks good! Maybe we want to sort out the `:local-launch` hidden parameter thing in a separate PR?

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -180,93 +131,148 @@
...
   (reset! build-in-progress-atom true)
-  (let [prev-out System/out
+  (let [;; bob might notify the progress tracker AFTER it's done!
+        render-progress! (progress/until-done render-progress!)
+        provided-evaluation-context (some? evaluation-context)
+        evaluation-context (or evaluation-context (g/make-evaluation-context))
+        prev-out System/out
         prev-err System/err
+        log-output-stream (or log-output-stream
+                              (WriterOutputStream. ^PrintWriter *out* StandardCharsets/UTF_8 1024 true))
```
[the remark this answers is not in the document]
vlaaad: No, because we don't want to close `*out*`.

PR #9522, opened 2024-10-02
Consolidate writing attribute data from source "channels"

Jhonnyg:


engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1062,24 +1066,27 @@ namespace dmGameSystem
...
-                    UNPACK_ATTRIBUTE_PTR(m_Texcoord0),
-                    UNPACK_ATTRIBUTE_PTR(m_Texcoord1),
-                };
+                const float* world_matrix_channels[]  = { (float*) &instance_render_item->m_World };
+                const float* normal_matrix_channels[] = { (float*) &normal_matrix };
+                const float* uv_channels[]            = { UNPACK_ATTRIBUTE_PTR(m_Texcoord0), UNPACK_ATTRIBUTE_PTR(m_Texcoord1), };
+                const float* color_channels[]         = { UNPACK_ATTRIBUTE_PTR(m_Colors) };
+                const float* position_channels[]      = { UNPACK_ATTRIBUTE_PTR(m_Positions) };
+                const float* normal_channels[]        = { UNPACK_ATTRIBUTE_PTR(m_Normals) };
+                const float* tangent_channels[]       = { UNPACK_ATTRIBUTE_PTR(m_Tangents) };
```
Jhonnyg, line 1075: The source data has the exact same layout now for all data points, which should make it easier to introduce more channels if needed.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -690,32 +690,45 @@ namespace dmGameSystem
...
+        dmGraphics::SetWriteAttributeStreamDesc(&params->m_WorldMatrix, world_matrix, dmGraphics::VertexAttribute::VECTOR_TYPE_MAT4, 1, true);
+        dmGraphics::SetWriteAttributeStreamDesc(&params->m_PageIndices, pi_channels, dmGraphics::VertexAttribute::VECTOR_TYPE_SCALAR, pi_channels_count, true);
     }
 
-    static void CreateVertexDataSlice9(uint8_t* vertices, uint8_t* indices, bool is_indices_16_bit, bool has_local_position_attribute,
-        const Matrix4& world_matrix, Vector3 sprite_size, Vector4 slice9, uint32_t vertex_offset, uint32_t vertex_stride,
-        TexturesData* textures, dmArray<float>* scratch_uvs, float* scratch_uv_ptrs[MAX_TEXTURE_COUNT],
-        dmArray<dmVMath::Vector4>* scratch_positions_world, dmArray<dmVMath::Vector4>* scratch_positions_local,
-        bool flip_u, bool flip_v, dmGraphics::VertexAttributeInfos* sprite_infos)
+    static void CreateVertexDataSlice9(
```
Jhonnyg, line 713: This function could use a refactor..

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1224,15 +1231,36 @@ namespace dmGameSystem
...
         dmGraphics::HVertexDeclaration* vx_decl_in_out, uint32_t* vertex_stride_out,
         dmGraphics::VertexAttributeInfos* material_infos_vertex, uint32_t vertex_count, uint8_t** vb_begin)
     {
         *vb_begin = 0;
         memset(material_infos_vertex, 0, sizeof(dmGraphics::VertexAttributeInfos));
+
         dmGraphics::HVertexDeclaration vx_decl = *vx_decl_in_out;
 
-        if (!HasCustomVertexAttributes(material))
+        if (CanUseDefaultVertexDeclaration(material))
```
Jhonnyg, line 1263: This slightly differs because we only want to check if the vertex declaration contains regular "vertex" spaced attributes and not instanced. This fixes the case where someone is using a "mtx_world" attribute but no instancing is enabled (or supported), and we can use the fast route.

JCash (approved):

PR #9531, opened 2024-10-03
Add `editor.bob` to editor scripts

vlaaad:
Related to #8062

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
@@ -58,6 +58,9 @@ import com.dynamo.bob.util.HttpUtil;
 import com.dynamo.bob.cache.ResourceCacheKey;
 import com.dynamo.bob.util.FileUtil;
 
+import static com.dynamo.bob.Bob.CommandLineOption.ArgCount.*;
+import static com.dynamo.bob.Bob.CommandLineOption.ArgType.ABS_OR_CWD_REL_PATH;
+
 public class Bob {
     private static final Logger logger = Logger.getLogger(Bob.class.getName());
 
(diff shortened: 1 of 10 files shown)


editor/src/clj/editor/disk.clj
```diff
@@ -304,19 +304,24 @@
   (disk-availability/push-busy!)
   (future
     (try
-      (let [hook-opts {:output-directory (get bob-options "bundle-output")
+      (let [invoke-bundle-hooks (boolean (some #(= "bundle" %) bob-commands))
```
vlaaad, line 307: This code may be invoked by the editor script now, and it's not guaranteed we are doing a bundle. I thought about adding a flag to `editor.bob` that controls if we should invoke the bundle hooks, but coming to think of that, we can detect a bundle just by looking at the bob commands.

editor/src/clj/editor/disk.clj
```diff
@@ -304,19 +304,24 @@
...
+            hook-opts {:output-directory (or (get bob-options "bundle-output")
+                                             (get bob-options "output")
+                                             "build/default")
                        :platform (get bob-options "platform")
-                       :variant (get bob-options "variant")}]
+                       :variant (get bob-options "variant" "release")}]
```
vlaaad, line 312: Now we need to provide defaults to hook opts because they are not guaranteed to be present.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -480,97 +483,127 @@ static public void addToPaths(String dir) {
...
-        addOption(options, "r", "root", true, "Build root directory. Default is current directory", true);
-        addOption(options, "o", "output", true, "Output directory. Default is \"build/default\"", false);
-        addOption(options, "i", "input", true, "DEPRECATED! Use --root instead", true);
-        addOption(options, "v", "verbose", false, "Verbose output", false);
-        addOption(options, "h", "help", false, "This help message", false);
-        addOption(options, "a", "archive", false, "Build archive", false);
-        addOption(options, "ea", "exclude-archive", false, "Exclude resource archives from application bundle. Use this to create an empty Defold application for use as a build target", false);
-        addOption(options, "e", "email", true, "User email", false);
-        addOption(options, "u", "auth", true, "User auth token", false);
+    public static List<CommandLineOption> getCommandLineOptions() {
```
vlaaad, line 521: Some options accept relative paths. The paths are resolved against the JVM's working directory used to launch Bob. We use the same options from the editor, but we want to resolve relative paths against the opened project root instead — the working directory of the editor process is irrelevant. I changed option definitions to include the option metadata needed to support this.

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -135,48 +134,57 @@
...
+              Bob$CommandLineOption$ArgType/ABS_OR_CWD_REL_PATH (str (.normalize (.resolve project-root (str (validate-type config String value)))))))]
+    (let [{:keys [undefined cli internal]}
+          (util/group-into {} []
+                           (fn option-group [e]
+                             (let [k (key e)]
+                               (cond
+                                 (keyword? k) :internal
+                                 (long-option? k) :cli
+                                 :else :undefined)))
+                           identity
```
matgis, line 166: You can simply omit the `value-fn` argument to `group-into` if you want to use the values as-is.

matgis (approved):
Looks great! I had one concern about some Bob arguments that should maybe have `ArgType.ABS_OR_CWD_REL_PATH`, so please check that before merging!

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -480,97 +483,127 @@ static public void addToPaths(String dir) {
...
+        return List.of(
+                opt("r", "root", ONE, ABS_OR_CWD_REL_PATH, "Build root directory. Default is current directory", true),
+                opt("o", "output", ONE, "Output directory. Default is \"build/default\"", false),
```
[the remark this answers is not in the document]
vlaaad: Great question! The answer is no for both options:
- `output` should not use `ABS_OR_CWD_REL_PATH`, because it is required to be a relative path, and it's resolved against project root by bob itself, so the editor shouldn't attempt to un-relativize it at all. See: https://github.com/defold/defold/blob/6fd84fb0a19e0416fd8e81b93b3e640ee2e02f46/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/fs/AbstractFileSystem.java#L53
- supplying `input` option leads to bob aborting the build altogether, so it does not matter what's the value of the option: https://github.com/defold/defold/blob/e184ca0b3fbc7c877e5872844d7e8470646a557f/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java#L781-L784

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -135,48 +134,57 @@
...
-       (coll/pair-map-by (comp name key) (comp str val) internal))]))
+  [^Path project-root options-map]
+  (letfn [(validate-type [^Bob$CommandLineOption config ^Class expected-class value]
+            (when-not (instance? expected-class value)
+              (throw (IllegalArgumentException. (str (.-longOpt config) " option is expected to be " (.getSimpleName expected-class) ", was " (some-> (class value) .getSimpleName)))))
+            value)
+          (stringify-value [^Bob$CommandLineOption config value]
+            (condp = (.-argType config)
+              Bob$CommandLineOption$ArgType/UNSPECIFIED (str value)
+              Bob$CommandLineOption$ArgType/ABS_OR_CWD_REL_PATH (str (.normalize (.resolve project-root (str (validate-type config String value)))))))]
```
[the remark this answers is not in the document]
vlaaad: It shouldn't be necessary since Bob will do it anyway.

editor/src/clj/editor/pipeline/bob.clj
vlaaad: Good catch!

PR #9538, opened 2024-10-04
Configure default vertex attributes in the same way as the editor

Jhonnyg:
1.9.4 Beta issue with local spaced models

### DIFF
--- a/engine/gamesys/src/gamesys/gamesys_private.cpp
+++ b/engine/gamesys/src/gamesys/gamesys_private.cpp
@@ -316,7 +316,7 @@ namespace dmGameSystem
 
     // Prepares the list of material attributes by getting all the vertex attributes and values for each attribute
     // as specified in the material
-    void FillMaterialAttributeInfos(dmRender::HMaterial material, dmGraphics::HVertexDeclaration vx_decl, dmGraphics::VertexAttributeInfos* infos)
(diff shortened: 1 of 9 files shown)


engine/gamesys/src/gamesys/gamesys_private.cpp
```diff
@@ -337,6 +337,11 @@ namespace dmGameSystem
             info.m_StepFunction     = material_attributes[i].m_StepFunction;
             info.m_ValueVectorType  = material_attributes[i].m_VectorType;
 
+            if (info.m_CoordinateSpace == dmGraphics::COORDINATE_SPACE_DEFAULT)
+            {
+                info.m_CoordinateSpace = default_coordinate_space;
+            }
```
Jhonnyg, line 343: Components that use the "vertex space" option in the materials (basically only models) will produce certain streams based on that value, i.e positions normals and tangents.

engine/render/src/render/material.cpp
```diff
@@ -100,6 +52,65 @@ namespace dmRender
         return (dmGraphics::VertexAttribute::VectorType) -1;
     }
 
+    static void SetVertexAttributeDefaultSettings(dmGraphics::VertexAttribute* attribute, dmhash_t name_hash, dmGraphics::Type graphics_type, uint32_t element_count, bool instancing_supported)
```
Jhonnyg, line 55: Consolidated a bunch of functions into a single one. Basically does the same thing as "attribute-key->default-attribute-info" in graphics.clj

PR #9540, opened 2024-10-04
Editor: Added util.eduction module

matgis:
### Technical changes
Added `util.eduction` module with functions that return `eductions` for all `clojure.core` transducers.


vlaaad (approved):
Shipit!

PR #9541, opened 2024-10-04
Rename textures from spirv-cross constructed naming scheme

Jhonnyg:


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/shader/ShaderCompilePipeline.java
```diff
@@ -78,14 +78,39 @@ private static int shaderLanguageToVersion(ShaderDesc.Language shaderLanguage) {
...
+        //
+        // So these two rules together then becomes:
+        //   uniform sampler2D my_texture; ==> SPIRV_Cross_Combinedsampler_2dsampler_2d_separated
+        //
+        // Even without the separation of texture/sampler, we will still need to rename the texture in the source
+        // due to how spirv-cross works.
+        for (SPIRVReflector.Resource texture : textures) {
+            String spirvCrossSamplerName = String.format("SPIRV_Cross_Combined%s%s_separated", texture.name, texture.name);
+            source = source.replaceAll(spirvCrossSamplerName, texture.name);
+        }
```
Jhonnyg, line 112: This will fix the problem where combined samplers (sampler2D, sampler3D, samplerArray, samplerCube) is renamed for the glsl shaders, but it doesn't support the case where a "sampler object" (i.e uniform sampler my_sampler) is expanded into a combined sampler2D. But we don't really support sampler objects in any capacity in the engine, so I've left it out for now.

JCash (approved):

PR #9545, opened 2024-10-04
Fix loading of extension node types that define protocol outputs

### LINKED ISSUE #9519: Spine extension issue
### Steps to reproduce

<hr/>

<table>
<tr><td>Build time</td><td>2024-07-29T18:23:25.696563</td></tr>
<tr><td>Defold channel</td><td>dev</td></tr>
<tr><td>Defold editor sha</td><td>d4994b939edd128407b3808160dd2decd9d31377</td></tr>
<tr><td>Defold engine sha</td><td>d4994b939edd128407b3808160dd2decd9d31377</td></tr>
<tr><td>Defold version</td><td>1.6.2</td></tr>
<tr><td>Error</td><td><a href='https://sentry.io/organizations/defold/issues/?query=id%3A"c241d95be2af48009259056e795a7672"'>c241d95be2af48009259056e795a7672</a></td></tr>
<tr><td>GPU</td><td>Apple M1 Max</td></tr>
<tr><td>OS version</td><td>14.6.1</td></tr>
<table>

vlaaad:
To be honest, I'm not sure how to review the code, test the change, or reproduce the issue :D 

I had to look at macroexpansion output of the `defnode` macro. The macro-expanded form of Spine's [SpineSceneNode](https://github.com/defold/extension-spine/blob/main/defold-spine/editor/src/spineguiext.clj#L218), before it was `eval`ed, included protocol instance as an object, which included anonymous functions, which caused the emitted class to do `RT.readString("#=(anon$fn_123123.)")` during static initialization.

BTW, a customer reported that the following workaround works: altering spine extension so that `defnode` output types use `g/Any` instead of protocol refs like `resource/Resource`.

There are at least 2 problems:
1. we embed protocol objects in evaluated forms — we should embed a symbol of a protocol instead.
2. node types defined by extensions use schemas in a release build of the editor — they should not do it at all. When we compile a release version of the editor, we bind `*check-schemas*` to false, so the emitted code does not perform the schema sanity checks. But then, during the runtime, `*check-schemas*` is set to true. It does not affect pre-compiled code, but it affects dynamically loaded extensions.

Possibly, there is a 3rd problem with class loaders. It's weird that an extension can define a class with a reference to a class of an anonymous function, and then fail to load it. I tried to, but I could not come up with a repro for this scenario. So, this PR only fixes the 2 problems.

Fixes #9519


matgis (approved):
Spectacular job on finding these! I'm really happy to have both these issues fixed, and hopefully the remaining class loader strangeness is not a problem for us. LGTM! ❤️

Issue #9379, opened 2024-09-04
Incorrect displaying of trimmed sprites in atlas and GUI


Morgerion:
Relevant for all versions of the Defold Editor, including beta 1.9.3

![image](https://github.com/user-attachments/assets/81d5b482-fe34-41a2-8c23-7eba21af48ea)

This incorrect mapping does not affect rendering in any way, and does not happen for all sprites.
However, when using such sprites in GUI, any Pivot Mode other than Center stops working in node.
This happens **only in the editor**, in the build everything works correctly.

britzl:
Could you please compile a small repro project where we can try this?

AGulev:
example: ![CleanShot 2024-09-18 at 20 23 52@2x](https://github.com/user-attachments/assets/f96ea83c-72fe-4303-826c-a0fe4a602357)
[dev-texture-vertex-data.zip](https://github.com/user-attachments/files/17048653/dev-texture-vertex-data.zip)

AGulev:
![CleanShot 2024-10-20 at 15 51 10@2x](https://github.com/user-attachments/assets/17ecfb73-46c0-414f-8306-765473c36675)

PR #9552, opened 2024-10-05
Fixed an issue where trimmed images in the atlas displayed incorrect outlines

AGulev:
Fixed an issue where the editor displayed incorrect outlines for a rotated image in an atlas with trimmed mode.

Fix https://github.com/defold/defold/issues/9379


editor/src/clj/editor/atlas.clj
```diff
@@ -758,12 +758,22 @@
    page     :- types/Int32
    geometry :- s/Any])
 
+(defn rotate-vertices-90-cw [vertices]
+  (mapcat (fn [[x y]]
+            [y (- x)])
+          (partition 2 vertices)))
```
vlaaad, line 764: I'd prefer us not to use lazy sequences, and instead implement such transformations using transducers, i.e.:
```clj
(into []
      (comp (partition-all 2)
            (mapcat (fn [[x y]]
                      [y (- x)])))
      vertices)
```

editor/src/clj/editor/atlas.clj
```diff
@@ -758,12 +758,22 @@
...
   (let [[w h] layout-size
         geometries (:geometries texture-set)]
     (into {} (map (fn [{:keys [path x y width height index page]}]
-                    [path (->AtlasRect path x (- h height y) width height page (get geometries index))]))
+                    (let [geometry (get geometries index)
+                          rotated-vertices (if (:rotated geometry)
+                                             (rotate-vertices-90-cw (:vertices geometry))
+                                             (:vertices geometry))]
+                      [path (->AtlasRect path x (- h height y) width height page
+                                         (assoc geometry :vertices rotated-vertices))])))
```
vlaaad, line 776: Hmm, is this the right place for the change? I mean, is this the only place where the vertices from `texture-set` need to be rotated? Could it be that we want the vertices to be rotated everywhere once we have them? In that case, we might want to do it in `editor.pipeline.texture-set-gen/atlas->texture-set-data`, for example...

vlaaad:
Maybe @matgis could also have a look?

editor/src/clj/editor/atlas.clj
AGulev: It's a good question. I didn't find a better place  + here it's cached. But I agree, it might be not the best place and I just still have "fear" to introduce new entities in the editor code.

editor/src/clj/editor/atlas.clj
AGulev: fixed, thank you

editor/src/clj/editor/atlas.clj
matgis: Back when I was fixing various issues in this area, I tried to unify the various code paths that dealt with vertex geometry. To make sure it is picked up everywhere, I think we may want to make this change in the `texture-set/make-anim-data` code path, specifically in `->anim-frame-from-geometry` called from the `->anim-data` function. Here we deal with the flip flags, so it might make sense to also deal with rotation here?

Unless of course we need to apply this rotation to "undo" an existing rotation for this specific use case? It's a bit hard to tell, because I don't know exactly what preparations are done to the geometry by the `TextureSetGenerator` class we share with Bob. 🤔

matgis (changes requested):

editor/src/clj/editor/atlas.clj
AGulev: After investigation, I still think, that I use right place for it. 

The place @matgis recommended is the place where atlas produces its outputs, which used for all the components and the other places the atlas linked. 

But the only issue we have is representation in the atlas viewer itself. 
That's why I modified `image-path->rect` because the only place where it is used is `produce-image-scene` in `atlas`, So I fix the only place we have this issue, and it doesn't affect anything else.

matgis (approved):
OK, Sounds reasonable. Thanks for looking into it! Let's merge it like this, then! ❤️

PR #9553, opened 2024-10-05
Make sure all the changes taken into account in the bundle

### LINKED ISSUE #9548: HTML5 build reuses old cached files
HTML5 build seems to reuse old cached files, because once an HTML5 build is created, it no longer changes when you click on ‘Build HTML5’. Tested only on scripts and models.

On the left as in HTML5 and unchanged when clicking on ‘Build HTML5’, and on the right as it should be after changes in the project:
image

**To Reproduce (REQUIRED)**
1. `Build HTML5`.
2. Edit any script, save it.
3. `Build HTML5`.

  - Version 1.9.4 `4aeae3d0faad01ccb8f8eb495e2f50896e35ec62`

**Platforms (REQUIRED):**
 - Platforms: HTML5
 - OS: Windows 10

AGulev:
Fix https://github.com/defold/defold/issues/9548

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
@@ -70,7 +70,6 @@ import com.dynamo.bob.fs.FileSystemWalker;
 import com.dynamo.bob.fs.IFileSystem;
 import com.dynamo.bob.fs.IResource;
 import com.dynamo.bob.fs.ZipMountPoint;
-import com.dynamo.bob.pipeline.GameProjectBuilder;
 import com.dynamo.bob.plugin.PluginScanner;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
@@ -108,7 +107,6 @@ import com.dynamo.graphics.proto.Graphics.TextureProfiles;
 
 import com.dynamo.bob.cache.ResourceCache;
 import com.dynamo.bob.cache.ResourceCacheKey;
-import org.jagatoo.util.timing.Time;
 
 /**
  * Project abstraction. Contains input files, builder, tasks, etc
@@ -1812,7 +1810,7 @@ run:
                 // If for some reason it's not, something went wrong, and the build pipeline is broken.
                 // But some tests may run build for some particular files without building game.project at all.
                 shouldRun = shouldRun && (!allOutputExists || !allSigsEquals ||
-                            (isLastTask && buildContainsChanges && (task.getBuilder() instanceof GameProjectBuilder)));
+                            (isLastTask && buildContainsChanges && (task.getBuilder().isGameProjectBuilder())));
 
                 if (!shouldRun) {
                     if (allOutputExists && allSigsEquals)
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
@@ -1808,7 +1808,11 @@ run:
                 // game.project is always the last task
                 boolean isLastTask = completedTasks.size() + 1 == buildTasks.size();
                 boolean shouldRun = !completedTasks.contains(task);
-                shouldRun = shouldRun && (!allOutputExists || !allSigsEquals || (isLastTask && buildContainsChanges));
+                // GameProjectBuilder creates archives, and it is always the last task.
+                // If for some reason it's not, something went wrong, and the build pipeline is broken.
+                // But some tests may run build for some particular files without building game.project at all.
+                shouldRun = shouldRun && (!allOutputExists || !allSigsEquals ||
+                            (isLastTask && buildContainsChanges && (task.getBuilder() instanceof GameProjectBuilder)));
 
                 if (!shouldRun) {
                     if (allOutputExists && allSigsEquals)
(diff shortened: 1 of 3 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1800,8 +1804,10 @@ private List<TaskResult> runTasks(IProgress monitor) throws IOException {
                     }
                 }
                 TimeProfiler.stop();
-
-                boolean shouldRun = (!allOutputExists || !allSigsEquals) && !completedTasks.contains(task);
+                // game.project is always last task
+                boolean isLastTask = completedTasks.size() + 1 == buildTasks.size();
+                boolean shouldRun = !completedTasks.contains(task);
```
AGulev, line 1809: I didn't want to change it as part of my fix, but we don't process completed tasks and this is required condition, shouldn't we exit earlier?
CleanShot 2024-10-05 at 16 53 59@2x

AGulev:
@britzl are we fine with this approach? (I need your review to fix bug in beta)

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1800,8 +1804,10 @@ private List<TaskResult> runTasks(IProgress monitor) throws IOException {
                     }
                 }
                 TimeProfiler.stop();
-
-                boolean shouldRun = (!allOutputExists || !allSigsEquals) && !completedTasks.contains(task);
+                // game.project is always last task
+                boolean isLastTask = completedTasks.size() + 1 == buildTasks.size();
```
britzl, line 1808: It feels a bit unstable to rely on this. Sure, it is not likely to change, but still.... I would feel better if we did something like:

```
boolean isGameProjectTask = (task.getBuilder() instanceof GameProjectBuilder.class);
```

britzl (changes requested):
It feels unreliable to expect game.project to be the last task. Sure, it is likely going to be like that, but still..

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
AGulev: If the last task isn't game.project - the build is broken. Like totally broken.

britzl: Ok then! It might be worth adding a comment about why we do it like this. Or add a reference to this ticket.

AGulev: Ok. I'll do. 
Old logic before my refactoring had priority to use it only for game.project to make sure it builds the last one. Now when we have real build tree, it happens naturally (every task requires all the input builds to be done). 

Game.project has to be the last one, because it packs everything. If some task for some reason will be called after game.project, that means something goes really wrong and changes from such tasks will never be in the final bundle. I'll think how to add comment and somehow make sure the last task is game.project (maybe I'll add exception if it's not)

AGulev: Ok. I'll do. 
Old logic before my refactoring had priority to use it only for game.project to make sure it builds the last one. Now when we have real build tree, it happens naturally (every task requires all the input builds to be done). 

Game.project has to be the last one, because it packs everything. If some task for some reason will be called after game.project, that means something goes really wrong and changes from such tasks will never be in the final bundle. I'll think how to add comment and somehow make sure the last task is game.project (maybe I'll add exception if it's not)

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1816,6 +1823,11 @@ private List<TaskResult> runTasks(IProgress monitor) throws IOException {
                     monitor.worked(1);
                     continue;
                 }
+                if (isLastTask && !(task.getBuilder() instanceof GameProjectBuilder)) {
```
AGulev, line 1826: If it's the last task, we check if it's `GameProjectBuilder` - it it's not, the build pipeline is broken.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1800,8 +1805,14 @@ private List<TaskResult> runTasks(IProgress monitor) throws IOException {
...
-
-                boolean shouldRun = (!allOutputExists || !allSigsEquals) && !completedTasks.contains(task);
+                // game.project is always the last task
+                boolean isLastTask = completedTasks.size() + 1 == buildTasks.size();
+                boolean shouldRun = !completedTasks.contains(task);
+                // GameProjectBuilder creates archives, and it is always the last task.
+                // If for some reason it's not, something went wrong, and the build pipeline is broken.
+                // But some tests may run build for some particular files without building game.project at all.
+                shouldRun = shouldRun && (!allOutputExists || !allSigsEquals ||
+                            (isLastTask && buildContainsChanges && (task.getBuilder() instanceof GameProjectBuilder)));
```
AGulev, line 1815: Check task type, but to avoid checking every single task, use it together with `isLastTask`

AGulev: @britzl the approach with Exception didn't work, because some tests build separate files.
Instead, I did the change you initially asked but combine it with `isLastTask` to be sure it work fast on big projects

britzl: One possible tiny improvement could be to have a isGameProjectBuilder() on the Builder class, default to false and override to true for GameProjectBuilder.

britzl (dismissed):

britzl (approved):

PR #9555, opened 2024-10-06
Fix texture destroy in flight

JCash:
Fixes a crash when a texture was destroyed while it was still uploading to the gpu.

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/engine/gamesys/src/dmsdk/gamesys/resources/res_texture.h
+++ b/engine/gamesys/src/dmsdk/gamesys/resources/res_texture.h
@@ -22,6 +22,10 @@ namespace dmGameSystem
     struct TextureResource
     {
         dmGraphics::HTexture m_Texture;
+        uint8_t              m_Uploading:1;
+        uint8_t              m_DelayDelete:1;
+
+        TextureResource();
     };
 }
 
--- a/engine/gamesys/src/gamesys/resources/res_factory.cpp
+++ b/engine/gamesys/src/gamesys/resources/res_factory.cpp
@@ -89,7 +89,6 @@ namespace dmGameSystem
         else
         {
             ResFactoryDestroyResource(params->m_Factory, factory_res);
-            delete factory_res;
         }
         return res;
     }
--- a/engine/gamesys/src/gamesys/resources/res_texture.cpp
+++ b/engine/gamesys/src/gamesys/resources/res_texture.cpp
@@ -27,6 +27,11 @@ namespace dmGameSystem
 {
     static const uint32_t MAX_MIPMAP_COUNT = 15; // 2^14 => 16384 (+1 for base mipmap)
 
+    TextureResource::TextureResource()
+    {
+        memset(this, 0, sizeof(*this));
+    }
+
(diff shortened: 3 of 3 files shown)


engine/gamesys/src/gamesys/resources/res_factory.cpp
```diff
@@ -89,7 +89,6 @@ namespace dmGameSystem
         else
         {
             ResFactoryDestroyResource(params->m_Factory, factory_res);
-            delete factory_res;
```
JCash, line 92: Double delete bug

engine/gamesys/src/dmsdk/gamesys/resources/res_texture.h
```diff
@@ -22,6 +22,10 @@ namespace dmGameSystem
     struct TextureResource
     {
         dmGraphics::HTexture m_Texture;
+        uint8_t              m_Uploading:1;
+        uint8_t              m_DelayDelete:1;
```
Jhonnyg, line 26: I guess these can't both be one, right? Maybe the resource should have it's own state param instead

engine/gamesys/src/gamesys/resources/res_texture.cpp
```diff
@@ -386,8 +406,15 @@ namespace dmGameSystem
...
         TextureResource* texture_res = (TextureResource*) dmResource::GetResource(params->m_Resource);
-        dmGraphics::DeleteTexture(texture_res->m_Texture);
-        delete texture_res;
+
+        if (texture_res->m_Uploading)
+        {
+            // The job is currently in flight
+            texture_res->m_DelayDelete = 1;
+            return dmResource::RESULT_OK;
+        }
```
Jhonnyg, line 415: I wonder if we want to move more resource payloads to be "async-loadable", e.g any *other* graphics API allows you to upload data in a separate thread (OpenGL only allows a few resource types I Think), so we probably want to keep that in mind when designing render script 2.0 🤔 (just a thought)

Jhonnyg (approved):

engine/gamesys/src/dmsdk/gamesys/resources/res_texture.h
JCash: Well, we do set the delay delete because it is currently uploading.

JCash: That said, it would be nice if the resource system had some extra scaffolding around this concept.

engine/gamesys/src/gamesys/resources/res_texture.cpp
JCash: Yes, I thought the same. We probably want more resource types to be be possibly loaded at the same time.
Currently, the loader is single threaded, waiting for each resource to finish.

Issue #9557, opened 2024-10-06
Add current gui node count in the editor


JCash:
It would be nice to see the actual current node count used in the editor.

Workaround:
Set limit "max_nodes" to something small, then do divide and conquer to find the current node count.

AGulev:
Fixed https://github.com/defold/defold/pull/9562

PR #9562, opened 2024-10-07
Add a read-only property for the Current Node in the GUI.

AGulev:
Added a new read-only property `Current Nodes` for `GUI` to display the number of nodes currently being used in the GUI.

Fix https://github.com/defold/defold/issues/9557


vlaaad (approved):
LGTM!

PR #9567, opened 2024-10-07
Add vertex + index memory size to model profiling

### LINKED ISSUE #9534: Add some information about Instancing in profiler
When using instancing for models, we don't see any numbers in profiler.
It will be good to see some numbers to understand is instancing working, and find issues.

Add some number in profiler, about instancing.
I am not sure what numbers can be usefull for game developers.
Possible information
-number of groups
-instancing models(we already have total models. But i think it can be usefull to see number models that use instancing)
-if meshes or other components will be added also add them?
...

Profiler:
-----
Instancing:

    groups:10
    models:100
-----

Jhonnyg:
The profiler can now show vertex count and vertex buffer + index buffer memory for local and instanced local models.

Fixes #9534

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_model.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_model.cpp
@@ -136,7 +136,7 @@ namespace dmGameSystem
         // For profiling data:
         uint32_t                         m_StatisticsVertexCount;
         uint32_t                         m_StatisticsVertexDataSize;
-        uint16_t                         m_CurrentFrameTick;
+        uint8_t                          m_CurrentFrameTick;
     };
 
     static const uint32_t VERTEX_BUFFER_MAX_BATCHES = 16;     // Max dmRender::RenderListEntry.m_MinorOrder (4 bits)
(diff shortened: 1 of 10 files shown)


engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1120,6 +1124,14 @@ namespace dmGameSystem
         dmRender::AddToRender(render_context, &ro);
 
         world->m_InstanceBufferDataLocalSpace.SetSize(instance_write_ptr - world->m_InstanceBufferDataLocalSpace.Begin());
+
+        // Update statistics for the render item
+        if (render_item->m_Buffers->m_LastUsedFrame != world->m_CurrentFrameTick)
+        {
+            world->m_StatisticsVertexDataSize += dmGraphics::GetIndexBufferSize(ro.m_IndexBuffer) + dmGraphics::GetVertexBufferSize(ro.m_VertexBuffers[VX_DECL_BASE_BUFFER]);
+            render_item->m_Buffers->m_LastUsedFrame = world->m_CurrentFrameTick;
```
Jhonnyg, line 1132: We combine vertex + index buffer memory sizes here. An alternative would be to add a property for index buffers, but I'm not sure it's necessary..

JCash (approved):

Issue #9580, opened 2024-10-08
WebGPU detection is not necessary


aglitchman:
**Describe the bug (REQUIRED)**
WebGPU detection routine doesn't actually detect WebGPU, because it looks for the wrong class. It should be: https://developer.mozilla.org/en-US/docs/Web/API/GPUCanvasContext

And always logs the error in the browser: "An error occurred while detecting WebGPU support: ReferenceError: WebGPURenderingContext is not defined".

P.S. A new canvas and context is created to detect the availability of webgpu, and then the same is done to detect webgl. This is an unnecessary double load on the user's GPU and the system. I think it's enough to detect the availability of webgl, and whether the game will have webgl or webgpu is left up to the engine to figure out.

**To Reproduce (REQUIRED)**
1. Build HTML5 
2. Run the game
3. Check the browser logs

**Expected behavior (REQUIRED)**
dmloader.js shouldn't detect WebGPU at all.

**Defold version (REQUIRED):**
  - Version: 1.9.4 BETA (4028f3def398d6d8181ab5fe53469544e55b6893)

**Platforms (REQUIRED):**
 - Platforms: HTML5

ekharkunov:
@aglitchman Can you provide any documentation how requesting context from `canvas` element affects performance? I asked because in the future It's mandatory to check if WebGPU supported or not. Whether engine will do or not - it will be done via requesting graphical context. And WebGPU should check as primary graphics backend. It it's not supported - OpenGL will be used.
In final solution I believe that WebGL detection can be removed and handle error from engine side if opengl context can't be created

aglitchman:
> Can you provide any documentation how requesting context from `canvas` element affects performance?

I apologize 🙌 - it was just a suggestion based on the fact that getting WebGL context takes time.
So as not to fool around, I've done a small test on a $99 phone with Unisoc T606, Chrome 129 and the result is:

- WebGL: 48ms (first launch), then 33ms.
- WebGPU: 1ms, then 0ms.

It turns out WebGPU is really a step forward!! To sum up, it's ok to get the WebGPU context to check its availability. In this issue the most important thing for me was to avoid the error in the console. Plus, I didn't see the point in detecting WebGPU when Defold inside the engine itself decides what graphics backend to use if WebGPU is enabled. If I understood correctly.

<details>
  <summary>P.S. Test example</summary>
  
  ```js
(() => {
    const t1 = Date.now();
    const canvas = document.createElement('canvas');
[...]

PR #9581, opened 2024-10-08
Remove WebGPU detection from `dmloader.js`

aglitchman:
In the helper script `dmloader.js` there was added a detection of the WebGPU availability, but it worked incorrectly, because it was comparing with the wrong class and every user got an annoying error that there is no WebGPU (actually there is).

This PR removes the WebGPU detection and leaves only the WebGL detection. It is enough to detect the availability of WebGL as a minimum requirement to run the game, and whether the game will run WebGL or WebGPU will be decided by the engine itself.

Fixes [#9580](https://github.com/defold/defold/issues/9580)


AGulev:
We are merging it into beta and will plan how to fix it properly in the future releases.

AGulev (approved):

PR #9590, opened 2024-10-10
Fixed an issue where the time profiler would crash when creating the report file in large projects

### LINKED ISSUE #9588: Rewrite TimeProfiler to make it work with big projects
**Task (REQUIRED):**
Now TimeProfiler doesn't work in big projects because it generates too much data. 
It should be fixed so TimeProfiler can be used in projects of any size

TimeProfiler works

AGulev:
Fixed the method by which `TimeProfiler` in `bob.jar` creates the report file to prevent crashes when the project contains numerous files.

Fix https://github.com/defold/defold/issues/9588

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TimeProfiler.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TimeProfiler.java
@@ -166,7 +166,7 @@ public class TimeProfiler {
     }
 
     public static void createReport(Boolean fromEditor) {
-        // avoid douиle creation of the report by checking `fromEditor` flag
+        // avoid double creation of the report by checking `fromEditor` flag
         if (rootScope == null || TimeProfiler.fromEditor != fromEditor) {
             return;
         }
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TimeProfiler.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TimeProfiler.java
@@ -36,9 +36,6 @@ import org.codehaus.jackson.JsonGenerator;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.IOUtils;
 
-import com.samskivert.mustache.Mustache;
-import com.samskivert.mustache.Template;
-
 /**
  * Class helps to profile time of the Bob tool and generate report.
  */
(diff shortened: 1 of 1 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TimeProfiler.java
```diff
@@ -178,21 +158,15 @@ private static void saveHTML(String jsonReport, File reportFile) throws IOExcept
...
-        template.execute(ctx, sw);
-        sw.flush();
-
-        fileHTMMLWriter.write(sw.toString());
-        fileHTMMLWriter.close();
+        String token = "{{json-data}}";
+        int tokenIndex = templateString.indexOf(token);
+        String beforeToken = templateString.substring(0, tokenIndex - 1);
+        String afterToken = templateString.substring(tokenIndex + token.length()+1);
+        return new String[]{beforeToken, afterToken};
```
AGulev, line 165: Instead of using Mustache we manually split template into 2 parts "before" and "after token"

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TimeProfiler.java
```diff
@@ -225,14 +197,25 @@ public static void createReport(Boolean fromEditor) {
                         System.out.println("Failed to create directories: " + parentDir);
                     }
                 }
-                if (extension.equals(".json")) {
-                    saveJSON(jsonReport, finalReportFile);
-                }
-                else if (extension.equals(".html")) {
-                    saveHTML(jsonReport, finalReportFile);
+                int bufferSize = 8 * 1024 * 1024; // 8 MB in bytes
+                try (BufferedWriter writer = new BufferedWriter(new FileWriter(finalReportFile), bufferSize)) {
```
AGulev, line 201: Create a bufferWriter that writes every 8 Mb on disk directly without creation of a huge String.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TimeProfiler.java
```diff
@@ -225,14 +197,25 @@ public static void createReport(Boolean fromEditor) {
...
+                    if (extension.equals(".json")) {
+                        generateJSON(_rootScope, writer);
+                    } else if (extension.equals(".html")) {
+                        String[] htmlContent = getHtmlContent();
+                        writer.write(htmlContent[0]);
+                        writer.flush();
+                        generateJSON(_rootScope, writer);
+                        writer.flush();
+                        writer.write(htmlContent[1]);
```
AGulev, line 210: Add the first "before" html part of the time report, then write json, then add the second part of the html template.

britzl (approved):

PR #9592, opened 2024-10-11
WebGPU issues 

smagnuso:
Found a couple webgpu issues:

Make sure the context is all set to null as there lives non pod data in it that does not have constructor called now that it is malloc.

Alow interleaving buffer writes and draw calls works. In principal it should not happen as it causes a pipeline flush, but make sure the adapter handles it safely as it can be reordered in the work queue.

Make sure to account for WGPUBuffer that is current and its known size rather than relying on whole buffer in webgpu.h

### DIFF
--- a/share/extender/build_input.yml
+++ b/share/extender/build_input.yml
@@ -403,6 +403,7 @@ platforms:
         EMSCRIPTEN_BIN:         "{{env.EMSCRIPTEN_BIN_3_1_65}}"
         MANIFEST_MERGE_TOOL:    "{{env.MANIFEST_MERGE_TOOL}}"
     context:
+        optLevel: 3
         stackSize: 5242880
         minFirefoxVersion: 34
         minChromeVersion: 32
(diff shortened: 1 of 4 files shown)


share/extender/build_input.yml
```diff
@@ -438,16 +440,16 @@ platforms:
...
+    linkCmd:        '{{env.EMSCRIPTEN_BIN}}/em++ {{#src}}{{{.}}} {{/src}} -o {{tgt}} {{#linkFlags}}{{{.}}} {{/linkFlags}} {{#emscriptenLinkFlags}}-s {{{.}}} {{/emscriptenLinkFlags}} {{#ext.libPaths}}-L{{{.}}} {{/ext.libPaths}} {{#libPaths}}-L{{{.}}} {{/libPaths}} -Wl,--start-group {{#libs}}-l{{{.}}} {{/libs}} {{#ext.libs}}-l{{{.}}} {{/ext.libs}} {{#engineLibs}}-l{{{.}}} {{/engineLibs}} -Wl,--end-group {{#ext.jsLibs}}--js-library {{{.}}} {{/ext.jsLibs}} {{#engineJsLibs}}--js-library {{dynamo_home}}/lib/js-web/js/library_{{{.}}}.js {{/engineJsLibs}} {{#externalJsLibs}}--js-library {{dynamo_home}}/ext/lib/js-web/js/library_{{{.}}}.js {{/externalJsLibs}}'
 
 
   wasm-web:
     context:
         libPaths:   ["{{dynamo_home}}/lib/wasm-web", "{{dynamo_home}}/ext/lib/wasm-web"]
-        emscriptenLinkFlags: ["ALLOW_MEMORY_GROWTH=1", "WASM=1"]
+        emscriptenLinkFlags: ["ALLOW_MEMORY_GROWTH=1", "WASM=1", "ASYNCIFY=1", "ASYNCIFY_IGNORE_INDIRECT=1", "ASYNCIFY_ADD=[\"main\",\"dmEngineCreate(*)\",\"requestDeviceCallback(*)\",\"WebGPUCreateSwapchain(*)\",\"instanceRequestAdapterCallback(*)\"]"]
```
Jhonnyg, line 449: I think we need to do all the .yml changes in a separate PR, that adds these conditionally via the manifest tool.

Jhonnyg (changes requested):
As noted in the comment; We should do the manifest changes conditionally from the manifest tool similar to Vulkan. I can do that, so if you just remove the yml changes we can get the rest in and I'll prepare the PR for the manifest tool (including building tint for windows).

share/extender/build_input.yml
smagnuso: Removed.

Jhonnyg: I don't see any changes, perhaps you need to sync the PR?

Jhonnyg (approved):

PR #9615, opened 2024-10-18
Allow calling editor.bob from background threads

vlaaad:
Related to #8062

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -13,7 +13,8 @@
 ;; specific language governing permissions and limitations under the License.
 
 (ns editor.app-view
-  (:require [cljfx.fx.hyperlink :as fx.hyperlink]
+  (:require [cljfx.api :as fx]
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/app_view.clj
```diff
@@ -2697,36 +2698,37 @@ If you do not specifically require different script states, consider changing th
...
+                                                render-build-error!
+                                                commands
+                                                options
+                                                project
+                                                changes-view
+                                                (fn [successful]
+                                                  (if successful
+                                                    (future/complete! f nil)
+                                                    (future/fail! f (LuaError. "Bob invocation failed")))
+                                                  (.close out)))))
```
vlaaad, line 2731: Mostly indentation. I wrapped the whole command in `fx/on-fx-thread`. The problem was that `build-errors-view/clear-build-errors` could be invoked from the background thread, and that could cause the error `java.lang.IllegalStateException: Not on FX application thread; currentThread = ForkJoinPool.commonPool-worker-37`

matgis (approved):
Ship it!

PR #9620, opened 2024-10-21
remove auto uncompress when unpack_arc

smagnuso:
Allow unpacked archives to retain their compression on disk and have unpack_ddf do the unarchive on the fly.

### DIFF
--- a/scripts/unpack_ddf.py
+++ b/scripts/unpack_ddf.py
@@ -33,6 +33,7 @@ except:
 from google.protobuf import text_format
 import google.protobuf.message
 
+import lz4.block
 import binascii
 
 import gameobject.gameobject_ddf_pb2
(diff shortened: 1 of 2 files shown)


scripts/unpack_ddf.py
```diff
@@ -265,18 +266,27 @@ def print_shader_code(text):
...
 
 if __name__ == "__main__":
     path = sys.argv[1]
 
     with open(path, 'rb') as f:
         content = f.read()
-        _, ext = os.path.splitext(path)
-        builder = BUILDERS.get(ext, None)
+        ext = path.split(".")
+        if ext[-1] == "lz4":
```
JCash, line 280: What's the reasoning for not doing `os.splitext()` here?
When are we not having a suffix?

JCash (changes requested):
I'm not sure about going away from os.splitext() as it's a good api/contract to have.
Often when the question arises, "is the suffix with or without the .?" you don't know because there is no standard.
But here there _is_ a standard, and as an old tools/pipeline maintainer, I do appreciate the fact, and would rather not stray if not necessary.

JCash (approved):

PR #9624, opened 2024-10-22
Allow passing CFLAGS/LFLAGS from build.py to emscripten

smagnuso:
It is already possible to pass CFLAGS/LINKFLAGS into waf for running the c compiler, but I'd like to pass options specifically into emscripten as well at compile/link time.

### DIFF
--- a/build_tools/waf_dynamo.py
+++ b/build_tools/waf_dynamo.py
@@ -508,7 +508,12 @@ def default_flags(self):
         emflags_link =[j for i in emflags_link for j in i]
 
         flags = []
+        if os.environ.get('EMCFLAGS', None) is not None:
+            flags += os.environ.get("EMCFLAGS", "").split(' ')
         linkflags = []
+        if os.environ.get('EMLINKFLAGS', None) is not None:
+            linkflags += os.environ.get("EMLINKFLAGS", "").split(' ')
+
         if int(opt_level) < 2:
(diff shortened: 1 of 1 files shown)


build_tools/waf_dynamo.py
```diff
@@ -508,7 +508,12 @@ def default_flags(self):
         emflags_link =[j for i in emflags_link for j in i]
 
         flags = []
+        if os.environ.get('EMCFLAGS', None) is not None:
+            flags += os.environ.get("EMCFLAGS", "").split(' ')
         linkflags = []
+        if os.environ.get('EMLINKFLAGS', None) is not None:
+            linkflags += os.environ.get("EMLINKFLAGS", "").split(' ')
```
JCash, line 515: Would it make more sense to use `CFLAGS` and `LDFLAGS` to make it work for other platforms?

JCash (dismissed):
I think we can make it work for all platforms, by using the standard names.

JCash:
The changes would only be for custom local builds (or if ppl want to do it on CI)

JCash (approved):

smagnuso:
The reason I did EMCFLAGS instead of CFLAGS (CFLAGS already works in waf FWIW) is because I want it passed only to the link/compile lines for the wasm files. It seems there are some tools being compiled with clang to run as part of the build?

Issue #9606, opened 2024-10-16
Setting the game object max count too high causes a crash


totebo:
**Describe the bug (REQUIRED)**
When building, there is a crash if too many collection.max_instances are defined. There is no indication as to what might be the max number, but after a little digging it seems to be 32765.

```
ERROR:CRASH: CALL STACK:

# 0 pc     0x2a7764 dmengine _ZN7dmCrashL7HandlerEiP9__siginfoPv+36
# 1 pc      0x85fdd libc++abi.dylib _sigtramp+29
# 2 pc      0x40645 dmengine _Z22ResourceTypeGetContextP12ResourceType+5
# 3 pc      0x76e2c dmengine _ZN12dmGameSystemL32ResourceTypeGuiScript_UnregisterEP19ResourceTypeContextP12ResourceType+12
# 4 pc      0x4090e dmengine _ZN10dmResource15DeregisterTypesEP15ResourceFactoryP13dmHashTable64IPvE+94
# 5 pc       0x24fd dmengine _ZN8dmEngine6DeleteEPNS_6EngineE+173
# 6 pc       0x63cd dmengine _Z14dmEngineCreateiPPc+189
# 7 pc       0x70ff dmengine _ZN8dmEngine7RunLoopEPKNS_13RunLoopParamsE+95
# 8 pc       0x6fe1 dmengine _Z11engine_mainiPPc+113
# 9 pc      0x48345 libobjc.A.dylib start+1909
```

**To Reproduce (REQUIRED)**
Set collection.max_instances to 32766 or more. Run.

**Expected behavior (REQUIRED)**
Editing the number in the editor: 
Disallow entering numbers higher than 32765, with a message similar to the mass on a collisionobject:
![Screenshot 2024-10-16 at 09 55 43](https://github.com/user-attachments/assets/2300e90f-4fd1-43b4-86e4-7fc337ac57e6)

When running:
Show an error at runtime, explaining the 32765 limit.
![Screenshot 2024-10-16 at 09 57 10](https://github.com/user-attachments/assets/11317ffe-11e6-4d09-a588-48e99fc173c2)

**Defold version (REQUIRED):**
1.9.3

**Platforms (REQUIRED):**
macOS

**Minimal repro case project (OPTIONAL):**
[GameObjectLimit.zip](https://github.com/user-attachments/files/17392651/GameObjectLimit.zip)

PR #9637, opened 2024-10-23
Proper engine shutdown

AGulev:
Engine side fix for https://github.com/defold/defold/issues/9606

The engine shouldn't crash on De-init  if error happened during initialization

### DIFF
--- a/editor/src/clj/editor/console.clj
+++ b/editor/src/clj/editor/console.clj
@@ -659,6 +659,8 @@
                :name "console.reload.successful"}
               {:match #"^ERROR:.+?:"
                :name "console.error"}
+              {:match #"^FATAL:.+?:"
+               :name "console.fatal"}
               {:match #"^WARNING:.+?:"
                :name "console.warning"}
               {:match #"^INFO:.+?:"
@@ -672,6 +674,7 @@
     (view/make-color-scheme
       [["console.reload.successful" (Color/valueOf "#33CC33")]
        ["console.error" (Color/valueOf "#FF6161")]
+       ["console.fatal" (Color/valueOf "#8250df")]
        ["console.warning" (Color/valueOf "#FF9A34")]
        ["console.info" (Color/valueOf "#CCCFD3")]
        ["console.debug" (Color/valueOf "#3B8CF8")]
--- a/editor/src/clj/editor/console.clj
+++ b/editor/src/clj/editor/console.clj
@@ -674,7 +674,7 @@
     (view/make-color-scheme
       [["console.reload.successful" (Color/valueOf "#33CC33")]
        ["console.error" (Color/valueOf "#FF6161")]
-       ["console.fatal" (Color/valueOf "#8250df")]
+       ["console.fatal" (Color/valueOf "#FF6161")]
        ["console.warning" (Color/valueOf "#FF9A34")]
        ["console.info" (Color/valueOf "#CCCFD3")]
        ["console.debug" (Color/valueOf "#3B8CF8")]
--- a/engine/engine/src/engine.cpp
+++ b/engine/engine/src/engine.cpp
@@ -1052,10 +1052,11 @@ namespace dmEngine
             dmLogWarning("Failed to initialize sound system.");
         }
 
-        dmGameObject::Result go_result = dmGameObject::SetCollectionDefaultCapacity(engine->m_Register, dmConfigFile::GetInt(engine->m_Config, dmGameObject::COLLECTION_MAX_INSTANCES_KEY, dmGameObject::DEFAULT_MAX_COLLECTION_CAPACITY));
+        int max_instance_count = dmConfigFile::GetInt(engine->m_Config, dmGameObject::COLLECTION_MAX_INSTANCES_KEY, dmGameObject::DEFAULT_MAX_COLLECTION_CAPACITY);
+        dmGameObject::Result go_result = dmGameObject::SetCollectionDefaultCapacity(engine->m_Register, max_instance_count);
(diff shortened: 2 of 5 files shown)


editor/src/clj/editor/console.clj
```diff
@@ -672,6 +674,7 @@
     (view/make-color-scheme
       [["console.reload.successful" (Color/valueOf "#33CC33")]
        ["console.error" (Color/valueOf "#FF6161")]
+       ["console.fatal" (Color/valueOf "#8250df")]
```
AGulev, line 677: We didn't take into account Fatal errors, which make it unnoticeable for our users.
I added its own color: 
CleanShot 2024-10-23 at 17 15 43@2x

@vlaaad this part is for your review

engine/engine/src/engine.cpp
```diff
@@ -1052,10 +1052,11 @@ namespace dmEngine
             dmLogWarning("Failed to initialize sound system.");
         }
 
-        dmGameObject::Result go_result = dmGameObject::SetCollectionDefaultCapacity(engine->m_Register, dmConfigFile::GetInt(engine->m_Config, dmGameObject::COLLECTION_MAX_INSTANCES_KEY, dmGameObject::DEFAULT_MAX_COLLECTION_CAPACITY));
+        int max_instance_count = dmConfigFile::GetInt(engine->m_Config, dmGameObject::COLLECTION_MAX_INSTANCES_KEY, dmGameObject::DEFAULT_MAX_COLLECTION_CAPACITY);
+        dmGameObject::Result go_result = dmGameObject::SetCollectionDefaultCapacity(engine->m_Register, max_instance_count);
         if(go_result != dmGameObject::RESULT_OK)
         {
-            dmLogFatal("Failed to set max instance count for collections (%d)", go_result);
+            dmLogFatal("Failed to set max instance count %d for collections (%d)", max_instance_count, go_result);
```
AGulev, line 1059: I think it's better to show `max_instance_count` in the error message

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -249,7 +249,7 @@ namespace dmGameObject
     Result SetCollectionDefaultCapacity(HRegister regist, uint32_t capacity)
     {
         assert(regist != 0x0);
-        if(capacity >= INVALID_INSTANCE_INDEX-1)
+        if(capacity >= INVALID_INSTANCE_INDEX - 1 || capacity <= 0)
```
AGulev, line 252: `<= 0` is invalid value as well

engine/resource/src/resource_type.cpp
```diff
@@ -172,14 +172,17 @@ Result DeregisterTypes(HFactory factory, dmHashTable64<void*>* contexts)
             ctx.m_Contexts = contexts;
 
             HResourceType type = dmResource::FindResourceType(factory, desc->m_Name);
-
-            Result result = (Result)desc->m_DeregisterFn(&ctx, type);
-            if (result != RESULT_OK)
+            void* context = ResourceTypeContextGetContextByHash(&ctx, dmHashString64(desc->m_Name));
+            if (context)
```
AGulev, line 176: Unregister only in case type was registered to avoid crashes

engine/engine/src/engine.cpp
JCash: SOmetimes we also refer to the actual setting, i.e. using the dmGameObject::COLLECTION_MAX_INSTANCES_KEY

engine/gameobject/src/gameobject/gameobject.cpp
JCash: It cannot be < 0 though, as it's unsigned.
I imagine the compilation warns about this?

Change to "== 0"

JCash (changes requested):
Lgtm, only minor changes needed.

editor/src/clj/editor/console.clj
vlaaad: TBH, I'd prefer if we used the same color for errors. Throughout the editor, we use three colors for severity: blue for info, orange/yellow for warnings, and red for errors. Let us stay consistent.

vlaaad (changes requested):
I'd prefer if we kept the severity colors consistent throughout the editor.

vlaaad (approved):
LGTM!

editor/src/clj/editor/console.clj
AGulev: Fixed

engine/gameobject/src/gameobject/gameobject.cpp
AGulev: Fixed

engine/engine/src/engine.cpp
AGulev: Done

engine/engine/src/engine.cpp
```diff
@@ -1056,7 +1056,7 @@ namespace dmEngine
         dmGameObject::Result go_result = dmGameObject::SetCollectionDefaultCapacity(engine->m_Register, max_instance_count);
         if(go_result != dmGameObject::RESULT_OK)
         {
-            dmLogFatal("Failed to set max instance count %d for collections (%d)", max_instance_count, go_result);
+            dmLogFatal("Failed to set '%s' %d for collections (%d)", dmGameObject::COLLECTION_MAX_INSTANCES_KEY, max_instance_count, go_result);
```
JCash, line 1059: ❤️

JCash (approved):

PR #9640, opened 2024-10-23
Make it possible to pass an absolute path to a settings file to bob.jar

smagnuso:
I'd like to have the ability to pass in a .project that lives outside the game that will have organization wide settings, specifically manifest.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
@@ -1013,9 +1013,22 @@ public class Bob {
     }
 
     private static void logErrorAndExit(Exception e) {
-        System.err.println(e.getMessage());
-        if (e.getCause() != null) {
-            System.err.println("Cause: " + e.getCause());
+        logger.severe(e.getMessage().toString());
+        Throwable cause = e.getCause();
+        if (cause != null) {
+            for (int i = 0; cause != null; ++i) {
+                logger.severe("Cause:%d: %s", i, cause.toString());
+                StackTraceElement[] stackTrace = cause.getStackTrace();
+                for (StackTraceElement element : stackTrace) {
+                    logger.severe(element.toString());
+                }
+                cause = cause.getCause();
+            }
+        } else {
+            StackTraceElement[] stackTrace = e.getStackTrace();
+            for (StackTraceElement element : stackTrace) {
+                logger.severe(element.toString());
+            }
         }
         logger.severe(e.getMessage(), e);
         System.exit(1);
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
@@ -643,8 +643,8 @@ public class Project {
         buildServerHeaders.add(header);
     }
 
-    public void addPropertyFile(String filepath) {
-        propertyFiles.add(filepath);
+    public void addPropertyFile(String propertyFile) {
+        propertyFiles.add(propertyFile);
     }
 
     public void addEngineBuildDir(String dirpath) {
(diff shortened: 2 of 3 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -1013,9 +1013,22 @@ else if (build_report_html != null) {
...
+                for (StackTraceElement element : stackTrace) {
+                    logger.severe(element.toString());
+                }
+                cause = cause.getCause();
+            }
+        } else {
+            StackTraceElement[] stackTrace = e.getStackTrace();
+            for (StackTraceElement element : stackTrace) {
+                logger.severe(element.toString());
+            }
```
JCash, line 1031: Please add short comments to your pr that mentions the problem you're trying to solve, as it's hard to at a glance get into the correct context of the change.

Also, we like to keep separate changes apart, for easier maintenance.
So, in future, this would be in a separate PR.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -643,8 +643,8 @@ public void addBuildServerHeader(String header) {
         buildServerHeaders.add(header);
     }
 
-    public void addPropertyFile(String filepath) {
-        propertyFiles.add(filepath);
+    public void addPropertyFile(String propertyFile) throws IOException {
+        propertyFiles.add(propertyFile);
```
JCash, line 647: This function doesn't need to throw anything?
What does this change solve?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -659,10 +659,15 @@ public List<String> getPropertyFiles() {
...
-            resources.add(fileSystem.get(relativePath.toString()));
+            Path path = Paths.get(propertyFile);
+            if(path.isAbsolute()) {
+                resources.add(fileSystem.get(path.getFileName().toString()));
+            } else {
+                Path rootDir = Paths.get(getRootDirectory()).normalize().toAbsolutePath();
+                Path settingsFile = path.normalize().toAbsolutePath();
+                Path relativePath = rootDir.relativize(settingsFile);
+                resources.add(fileSystem.get(relativePath.toString()));
+            }
```
JCash, line 670: If possible, we prefer the style, where we do the final task only once.
E.g. 
```java
Path settingsPath = null;
if (path.isAbsolute())
    settingsPath = path;
else {
    settingsPath = ...
}
resources.add(fileSystem.get(settingsPath.getFileName().toString()));
```

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -719,7 +724,21 @@ public List<TaskResult> build(IProgress monitor, String... commands) throws IOEx
...
+            if(path.isAbsolute()) {
+                String normalizedRoot = path.getParent().normalize().toString();
+                if(!mounts.contains(normalizedRoot)) {
+                    DefaultFileSystem fs = new DefaultFileSystem();
+                    fs.setRootDirectory(normalizedRoot);
+                    this.fileSystem.addMountPoint(new FileSystemMountPoint(this.fileSystem, fs));
+                    mounts.add(normalizedRoot);
+                }
+            }
+        }
```
JCash, line 740: Please comment in code why this is needed, as it might not be totally obvious.

JCash (changes requested):
Overall good.
Seems like an unnecessary exception.
Also minor code style.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
smagnuso: I split it into a separate commit. The real problem here was that throw would happen and then get rethrown but when it was finally caught in the default handler it was very hard to see where the original came from. So I just added logging for cause to make it easier to get back to code that caused the initial exception.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -719,7 +722,22 @@ public List<TaskResult> build(IProgress monitor, String... commands) throws IOEx
...
+        for (String propertyFile : propertyFiles) {
+            Path path = Paths.get(propertyFile);
+            if (path.isAbsolute()) {
+                String normalizedRoot = path.getParent().normalize().toString();
+                // do not add the same mount twice if multiple settings are passed on commandline in same directory
+                if (!mounts.contains(normalizedRoot)) {
+                    DefaultFileSystem fs = new DefaultFileSystem();
+                    fs.setRootDirectory(normalizedRoot);
+                    this.fileSystem.addMountPoint(new FileSystemMountPoint(this.fileSystem, fs));
+                    mounts.add(normalizedRoot);
```
JCash, line 736: Tbh, this adds an entire folder to the project, which isn't really what your ticket title suggests.

While this feature currently is only used by you, I may also foresee issues in the future when you'll get tripped up by this.

JCash (approved):

PR #9649, opened 2024-10-24
Add the `engine-setup` script tag to the HTML5 template

### LINKED ISSUE #9645: Bring back the `engine-setup` section to the HTML5 template
A few versions of the engine ago, the html5 template was significantly improved, but as a result of this change, the `engine-setup` section was disappeared from the template. 

As it used to be:
```html
<script id='engine-loader' type='text/javascript' src="dmloader.js"></script>
<script id='engine-setup' type='text/javascript'>
// ...
</script>
<script id='engine-start' type='text/javascript'>
// ...
</script>
```

As it became:
```html
<script id='engine-loader' type='text/javascript' src="dmloader.js"></script>
<script id='engine-start' type='text/javascript'>
// ...
</script>
```

How I used it - this section is located between `dmloader.js` and the engine start code. For example, **only in between** can you put your own ProgressView to create your own variant of the loader, because the `EngineLoader.load` function directly uses ProgressView to create the view of the loader. In my games and examples for libraries (Scene3d, for example), I was simply adding a piece of code in `engine-setup` that replaced `Progress` (before it was named `ProgressView`) and got a new view of the loader. Very convenient! I did this with the `your_extension/manifests/web/engine_template.html` file, but not by completely replacing the entire template.

Bring back to HTML5 template section `engine-setup`.

1. Completely replace the HTML template with your own, but this is very inconvenient as it requires the developer to keep track of every change in the template. 
2. Replace “engine-start” with a completely your own version via `merge=“keep”`, since `merge=“merge”` adds code to the end and consequently it is not possible to replace `ProgressView` with your own.

It would be interesting to know, only I've done it that way and maybe it's just a whim of mine. But it was very convenient!

All of our games use the standard html5 template and we only include a little code through our manifest via our <script> section. For this reason, any updates to the HTML5 template don't require you to redo the template to your variant.

aglitchman:
This change brings back the `engine-setup` section to the HTML5 template. Developers can use it and native extensions can inject code here to add game startup parameters via the `CUSTOM_PARAMETERS` object, setup loading progress listeners, override the `dmloader.js` functions.

Closes https://github.com/defold/defold/issues/9645

### DIFF
--- a/engine/engine/content/builtins/manifests/web/engine_template.html
+++ b/engine/engine/content/builtins/manifests/web/engine_template.html
@@ -94,7 +94,10 @@
 	</div>
 	<!-- -->
 	<script id='engine-loader' type='text/javascript' src="dmloader.js"></script>
-
+	<script id='engine-setup' type='text/javascript'>
+		// From here you can configure game startup parameters via the CUSTOM_PARAMETERS object,
+		// override ProgressView to create your own loader. See dmloader.js for more details.
+	</script>
 	<script id='engine-start' type='text/javascript'>
 		var runningFromFileWarning = document.getElementById("running-from-file-warning");
 		if (window.location.href.startsWith("file://")) {


ekharkunov (approved):
👍

Issue #9651, opened 2024-10-25
go.property fails in certain cases


JCash:
Add a property like this (no space between the command and hash)
```
go.property("orientation",hash("landscape"))
```

Then Build&Run from the editor and you'll get a runtime error:
```
ERROR:GAMEOBJECT: Error running script: main/main.script:1: unexpected symbol near ','
```

Note: It works when building with bob.jar

PR #9652, opened 2024-10-25
Make sure `go.property()` with `hash()` works right

AGulev:
Fix https://github.com/defold/defold/issues/9651


editor/src/clj/editor/code/script_compilation.clj
```diff
@@ -175,7 +175,6 @@
                                  []
                                  nil))
                         (let [[_ end-row-skipped end-col-skipped] skipped
-                              end-col-skipped (- ^long end-col-skipped 1)
```
AGulev, line 178: I've no idea why I did this last time. there is no reason to have it

vlaaad (approved):
Ship it!

PR #9654, opened 2024-10-25
Use a global vao when enabling a vertex declaration

Jhonnyg:
In some extensions (e.g Rive) that use vertex array objects, the OpenGL render state can get mixed up because of our single global vertex array object that is only bound once during initialization.

### DIFF
--- a/engine/graphics/src/opengl/graphics_opengl.cpp
+++ b/engine/graphics/src/opengl/graphics_opengl.cpp
@@ -1429,11 +1429,7 @@ static void LogFrameBufferError(GLenum status)
         }
 
 #if !defined(GL_ES_VERSION_2_0)
-        {
-            GLuint vao;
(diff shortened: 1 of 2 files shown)


engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -1858,6 +1854,10 @@ static void LogFrameBufferError(GLenum status)
 
         OpenGLContext* context = (OpenGLContext*) _context;
 
+    #if !defined(GL_ES_VERSION_2_0)
+        glBindVertexArray(context->m_GlobalVAO);
+    #endif
```
Jhonnyg, line 1859: We must refresh the binding table at "some point" during our own rendering pipeline, and binding vertex declarations happens more rarely than binding vertex buffers.

JCash (approved):

PR #9659, opened 2024-10-28
Misc platform fixes for extensions

Jhonnyg:
A new DMSDK function has been added that can get the OpenGL id from a rendertarget: `OpenGLGetRenderTargetId`. Also, Vulkan support for iOS is working correctly again after recent GLFW changes.


engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -4540,6 +4543,14 @@ static void LogFrameBufferError(GLenum status)
         return false;
     }
 
+    // DMSDK
+    uint32_t OpenGLGetRenderTargetId(HContext _context, HRenderTarget render_target)
+    {
+        OpenGLContext* context = (OpenGLContext*) _context;
+        OpenGLRenderTarget* rt = GetAssetFromContainer<OpenGLRenderTarget>(context->m_AssetHandleContainer, render_target);
```
JCash, line 4550: Do we want to handle bad handles?

engine/graphics/src/wscript
```diff
@@ -110,6 +110,9 @@ def build(bld):
 
         elif bld.env.PLATFORM in ('arm64-ios',):
             obj.source.append('vulkan/ios/graphics_vulkan_ios_surface.mm')
+
+            obj.source.append('vulkan/osx/graphics_vulkan_osx.mm')
```
JCash, line 114: A little amazing that we haven't needed it before! ^^

JCash (approved):

engine/graphics/src/wscript
Jhonnyg: But that file is very new (two weeks old), it's just a few very specific functions for mixing vulkan and metal in an extension (you can probably guess which one!) :)

PR #9660, opened 2024-10-28
Configure opengl context creation

### LINKED ISSUE #9657: Configure OpenGL context creation from project settings
**Task (REQUIRED):**
Some projects require setting certain parameters when creating the OpenGL context, but currently the context creation process is static in the engine. 

For now, it's enough if we can set these parameters:
* Setting opengl major/minor version hints
* Use core profile on/off

Jhonnyg:
Game.project has two new options that control how the OpenGL context is created on desktop platforms:

* `opengl_version_hint`: If a specific version is selected, this will used as the minimum version required. Defaults to OpenGL 3.3, and a specific option is available that selects the highest version available.
* `opengl_core_profile_hint`: Enable/disable the 'core' profile for OpenGL. Enabling a core profile will remove all deprecated functionality from the OpenGL specification, which affects extensions as well.

Fixes #9657


engine/platform/src/platform_window_glfw.cpp
```diff
@@ -167,28 +206,29 @@ namespace dmPlatform
...
-        glfwOpenWindowHint(GLFW_OPENGL_VERSION_MINOR, 3);
-#elif defined(__MACH__)
+#elif defined(DM_PLATFORM_IOS)
         glfwOpenWindowHint(GLFW_OPENGL_VERSION_MAJOR, 3);
-    #if defined(DM_PLATFORM_IOS)
         glfwOpenWindowHint(GLFW_OPENGL_VERSION_MINOR, 0); // 3.0 on iOS
-    #else
-        glfwOpenWindowHint(GLFW_OPENGL_VERSION_MINOR, 2); // 3.2 on macOS (actually picks 4.1 anyways)
-    #endif
 #endif
```
Jhonnyg, line 212: windows and osx doesn't use glfw2 any more, these paths should have been removed before.

engine/platform/src/platform_window_glfw3.cpp
```diff
@@ -172,13 +172,74 @@ namespace dmPlatform
...
+        #ifdef _WIN32
+            // Not supported on windows when requesting the default OpenGL version (which will equate to the highest available)
+            can_set_profile_and_fc = !use_highest_version;
+        #endif
+
+            if (can_set_profile_and_fc)
+            {
+                glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
+                glfwWindowHint(GLFW_OPENGL_FORWARD_COMPAT, GL_TRUE);
+            }
```
Jhonnyg, line 240: The "forward compatability" hint is only needed when using the core profile, so I decided to not add a specific setting for it. According to the OpenGL documentation, it's only really needed for < versions 3.3, which we don't support anyway..

engine/platform/src/platform_window_glfw.cpp
```diff
@@ -154,6 +154,45 @@ namespace dmPlatform
         return 0;
     }
 
+    static void OpenGLVersionHintToMajorMinor(PlatformOpenGLVersion version, uint32_t* major, uint32_t* minor)
```
JCash, line 157: It's static but I don't see this being used in this file?
Also, it feels like the code should be in one place perhaps?

engine/engine/src/engine.cpp
```diff
@@ -411,6 +411,25 @@ namespace dmEngine
...
+
+        CHECK_AND_RETURN_OPENGL_VERSION(opengl_3_3, 3_3);
+        CHECK_AND_RETURN_OPENGL_VERSION(opengl_4_0, 4_0);
+        CHECK_AND_RETURN_OPENGL_VERSION(opengl_4_1, 4_1);
+        CHECK_AND_RETURN_OPENGL_VERSION(opengl_4_2, 4_2);
+        CHECK_AND_RETURN_OPENGL_VERSION(opengl_4_3, 4_3);
+        CHECK_AND_RETURN_OPENGL_VERSION(opengl_4_4, 4_4);
+        CHECK_AND_RETURN_OPENGL_VERSION(opengl_4_5, 4_5);
+        CHECK_AND_RETURN_OPENGL_VERSION(opengl_4_6, 4_6);
+        CHECK_AND_RETURN_OPENGL_VERSION(opengl_highest_available, HIGHEST_AVAILABLE);
```
JCash, line 427: I think an encoded int should suffice here, with e.g. 0 being the highest available.

JCash (changes requested):
Might be a code size improvement to encode version number in an int.

Also, the OpenGLVersionHintToMajorMinor isn't used in the glfw2 code flow?

engine/platform/src/platform_window_glfw3.cpp
aglitchman: If Defold stopped supporting OpenGL 3.1, it should be announced in the news 🙌. Because the minimum was OpenGL 3.1 since Defold 1.2.181. And it seems that after the updating Windows platform to GLFW 3 the engine really stopped supporting 3.1.

PR #9680, opened 2024-11-01
Support using structs in uniforms

### LINKED ISSUE #9678: Support for passing data into uniform structs in shaders
Something like this should be supported:

```glsl
// SHADER CODE

struct PbrMetallicRoughness
{
    vec4 baseColorFactor;
    float metallicFactor;
    float roughnessFactor;
};

struct PbrSpecularGlossiness
{
    vec4 diffuseFactor;
    vec3 specularFactor;
    float glossinessFactor;
};

uniform PBRMaterial
{
    PbrMetallicRoughness metallicRoughness;
    PbrSpecularGlossiness specularGlossiness;
};
```

```lua
-- render script (or via material)
self.cb = render.constant_buffer()
self.cb["PBRMaterial.metallicRoughness.baseColorFactor"] = vmath.vector4(1,2,3,4)
```

Jhonnyg:
Vertex and fragment shaders can now use struct types:

```glsl
struct CustomParameters
{
    vec4 params0;
    vec4 params1;
};

struct SceneData
{
    vec4 lightPositions[4];
    CustomParameters parameters; // some generic parameter, e.g camera position, scene bounding box or whatever.
};

uniform SceneUniforms
{
    SceneData scene;
};

uniform ParametersUniform
{
    CustomParameters customParameters;
};

void main()
{
    for (int i=0; i < 4; i++)
        color += calculateLight(scene.lightPositions[i], scene.parameters.params0);
    if (customParameters.params0.x > 0)
        // do something clever
    ...
}
```

Note that there are currently a few limitations on how to use structs in the engine - to read more head over to the `<insert-manual-entry-here>` section to read more.

Fixes #9678


engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -748,6 +748,8 @@ static void LogFrameBufferError(GLenum status)
             case CONTEXT_FEATURE_COMPUTE_SHADER:         return context->m_ComputeSupport;
             case CONTEXT_FEATURE_STORAGE_BUFFER:         return context->m_StorageBufferSupport;
             case CONTEXT_FEATURE_INSTANCING:             return context->m_InstancingSupport;
+            case CONTEXT_FEATURE_VSYNC:
+                break;
```
Jhonnyg, line 752: Fix compile warning

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -1433,7 +1435,7 @@ static void LogFrameBufferError(GLenum status)
             }
         }
 
-        if (glGenVertexArrays)
+        if (&glGenVertexArrays)
```
Jhonnyg, line 1438: Fix compile warning

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -1858,7 +1860,7 @@ static void LogFrameBufferError(GLenum status)
 
         OpenGLContext* context = (OpenGLContext*) _context;
 
-        if (glBindVertexArray)
+        if (&glBindVertexArray)
```
Jhonnyg, line 1863: Fix compile warning

engine/graphics/src/null/graphics_null.cpp
```diff
@@ -735,217 +735,293 @@ namespace dmGraphics
         return g_DrawCount;
     }
 
-    struct ShaderBinding
```
Jhonnyg, line 738: There's lots of changes in the null adapter, but mostly it is related to using the same code path and structures as the other adapters in an effort to move them closer together. This means we can test the binding code paths better, as even the null adapter can _fully_ support the type reflection. We wouldn't be able to test stuff like nested structs, or more advanced types otherwise.

engine/graphics/src/graphics.h
```diff
@@ -638,9 +649,8 @@ namespace dmGraphics
     uint8_t*         WriteAttributes(uint8_t* write_ptr, uint32_t vertex_index, const WriteAttributeParams& params);
 
     // Uniforms
-    uint32_t         GetUniformName(HProgram prog, uint32_t index, char* buffer, uint32_t buffer_size, Type* type, int32_t* size);
     uint32_t         GetUniformCount(HProgram prog);
-    HUniformLocation GetUniformLocation(HProgram prog, const char* name);
+    void             GetUniform(HProgram prog, uint32_t index, Uniform* uniform);
```
Jhonnyg, line 653: Merged these into one function.

engine/graphics/src/graphics.cpp
```diff
@@ -994,39 +995,70 @@ namespace dmGraphics
         }
     }
 
-    // TODO, comment from the PR (#4544):
-    //   "These frequent lookups could be improved by sorting on the key beforehand,
-    //   and during lookup, do a lower_bound, to find the item (or not).
-    //   E.g see: engine/render/src/render/material.cpp#L446"
-    bool GetUniformIndices(const dmArray<ShaderResourceTypeInfo> types, const dmArray<ShaderResourceBinding>& bindings, dmhash_t name_hash, uint64_t* index_out, uint64_t* index_member_out)
```
Jhonnyg, line 1001: This function isn't used anywhere (not part of DMSDK)

engine/graphics/src/null/graphics_null.cpp
```diff
@@ -735,217 +735,249 @@ namespace dmGraphics
...
-        ShaderDesc::Language   m_Language;
-        dmArray<ShaderBinding> m_Uniforms;
-        dmArray<ShaderBinding> m_Attributes;
+        uint16_t m_Binding;
+        uint8_t  m_Taken;
     };
 
-    static void PushBinding(dmArray<ShaderBinding>* binding_array, const char* name, uint32_t name_length, dmGraphics::Type type, uint32_t size)
+    // TODO: This can be merged with a certain "blue" graphics adapter
+    static void FillProgramResourceBindings(
```
Jhonnyg, line 745: We can perhaps merge this with all adapters as well (vulkan + webgpu has almost identical setup)

engine/graphics/src/null/graphics_null.cpp
```diff
@@ -1103,45 +1133,39 @@ namespace dmGraphics
...
-        dmStrlCat(buffer, uniform.m_Name, buffer_size);
-        *type = uniform.m_Type;
-        *size = uniform.m_Size;
-        return (uint32_t)strlen(buffer);
+        NullProgram* program = (NullProgram*)prog;
+        *uniform_desc = program->m_Uniforms[index];
     }
 
-    static HUniformLocation NullGetUniformLocation(HProgram prog, const char* name)
+    const Uniform* GetUniform(HProgram prog, dmhash_t name_hash)
```
Jhonnyg, line 1156: Tests only

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2149,11 +2151,17 @@ static void LogFrameBufferError(GLenum status)
         char* ptr = str;
         for (int i = len - 1; i >= 0; i--)
         {
-            if (ptr[i] == '.')
+            // For arrays, OpenGL returns the name as `name[0]`
+            if (ptr[i] == '[')
+            {
+                ptr[i] = 0;
+            }
```
Jhonnyg, line 2158: This case was handled in the renderer, but this is a better location..

engine/graphics/src/test/test_graphics.cpp
```diff
@@ -397,44 +399,6 @@ TEST_F(dmGraphicsTest, Drawing)
     dmGraphics::DeleteVertexStreamDeclaration(stream_declaration);
 }
 
-static inline dmGraphics::ShaderDesc MakeDDFShaderDesc(dmGraphics::ShaderDesc::Shader* shader,
```
Jhonnyg, line 400: Moved into a separate header since it's used in render tests as well

engine/graphics/src/test/test_graphics.cpp
```diff
@@ -570,23 +548,26 @@ TEST_F(dmGraphicsTest, TestComputeProgram)
         "void main() {\n"
         "}\n";
 
+    dmGraphics::ShaderDesc::ResourceTypeInfo types[1] = {};
+    FillShaderResourceTypeInfo(&types[0], "my_uniform", dmGraphics::ShaderDesc::SHADER_TYPE_VEC4);
+
     dmGraphics::ShaderDesc::ResourceBinding uniform = {};
-    FillResourceBinding(&uniform, "my_uniform", dmGraphics::ShaderDesc::SHADER_TYPE_VEC4);
+    FillResourceBindingTypeIndex(&uniform, "my_uniform", 0, 0);
```
Jhonnyg, line 555: For uniform buffers, we need to split the resource binding into the binding and the type. We can perhaps simplify the helper functions later

engine/graphics/src/graphics_private.h
```diff
@@ -144,7 +145,6 @@ namespace dmGraphics
         uint32_t m_StorageBufferCount;
         uint32_t m_TextureCount;
         uint32_t m_SamplerCount;
-        uint32_t m_TotalUniformCount;
```
Jhonnyg, line 147: Not used anymore

engine/graphics/src/graphics_private.h
```diff
@@ -167,6 +167,51 @@ namespace dmGraphics
...
 
+    struct Program
+    {
+        ProgramResourceBinding m_ResourceBindings[MAX_SET_COUNT][MAX_BINDINGS_PER_SET_COUNT];
+        dmArray<Uniform>       m_Uniforms;
+        uint8_t                m_MaxSet;
+        uint8_t                m_MaxBinding;
+    };
+
+    struct ProgramResourceBindingIterator
```
Jhonnyg, line 178: Just a helper to simplify looping through bindings a bit

engine/graphics/src/graphics_private.h
```diff
@@ -167,6 +167,51 @@ namespace dmGraphics
         uint8_t m_StageFlags;
     };
 
+    struct Program
+    {
+        ProgramResourceBinding m_ResourceBindings[MAX_SET_COUNT][MAX_BINDINGS_PER_SET_COUNT];
+        dmArray<Uniform>       m_Uniforms;
+        uint8_t                m_MaxSet;
+        uint8_t                m_MaxBinding;
+    };
```
Jhonnyg, line 176: This is the shared data needed for all programs, which is a first step to separating the graphics implementations into a "base" set of structs that are just data and an API-specific set of structs and functions that are API specific. A todo for later is to add this into the dmsdk:
`typedef struct Program* HProgram`

so that we don't need to cast it to a Program* in the functions that use it

engine/render/src/render/render.h
```diff
@@ -327,6 +328,9 @@ namespace dmRender
     void                            DeletePredicate(HPredicate predicate);
     Result                          AddPredicateTag(HPredicate predicate, dmhash_t tag);
 
+    HConstant                       NewConstant(dmhash_t name_hash, dmhash_t canonical_name_hash);
+    dmhash_t                        GetCanonicalConstantName(HConstant constant);
```
Jhonnyg, line 332: Avoid changing dmsdk

engine/graphics/proto/graphics/graphics_ddf.proto
```diff
@@ -353,22 +353,25 @@ message ShaderDesc
...
-        optional uint32         set           = 4;
-        optional uint32         binding       = 5;
+        required string         name               = 1;
+        required uint64         name_hash          = 2;
+        required ResourceType   type               = 3;
+        optional uint32         set                = 4;
+        optional uint32         binding            = 5;
+        optional uint32         element_count      = 6;
+        optional string         instance_name      = 7;
+        optional uint64         instance_name_hash = 8;
```
Jhonnyg, line 363: These are set to Optional, it's only exposed in this scenario:

uniform Lights {
....
} scene_lights[4];

engine/graphics/src/graphics.cpp
```diff
@@ -994,39 +1004,280 @@ namespace dmGraphics
...
+                info.m_MaxBinding = dmMath::Max(info.m_MaxBinding, (uint32_t) (res.m_Binding + 1));
+
+            #if 1
+                dmLogInfo("    name=%s, set=%d, binding=%d, data_offset=%d, texture_unit=%d", res.m_Name, res.m_Set, res.m_Binding, program_resource_binding.m_DataOffset, program_resource_binding.m_TextureUnit);
+            #endif
+            }
+        }
+    }
+
+    void FillProgramResourceBindings(
```
Jhonnyg, line 1266: This function can now we shared with vendor adapters as well!

engine/graphics/src/graphics.cpp
```diff
@@ -1161,7 +1412,22 @@ namespace dmGraphics
...
+    uint32_t GetUniformCount(HProgram prog)
+    {
+        Program* p = (Program*) prog;
+        return p->m_Uniforms.Size();
+    }
+    void GetUniform(HProgram prog, uint32_t index, Uniform* uniform_desc)
+    {
+        Program* p = (Program*) prog;
+        *uniform_desc = p->m_Uniforms[index];
+    }
```
Jhonnyg, line 1427: This is a somewhat new pattern for the graphics layer, where where we use a "base struct" that is shared for all adapters:

struct BaseThing
{
   int m_Shared;
};

struct OpenGLThing
{
  BaseThing m_Base;
  OpenGLThing m_Thing;
};

struct VulkanThing
{
  BaseThing m_Base;
  VulkanThing m_Thing;
};

engine/graphics/src/graphics.cpp
```diff
@@ -1363,18 +1629,6 @@ namespace dmGraphics
...
-        return g_functions.m_GetUniformName(prog, index, buffer, buffer_size, type, size);
-    }
-    uint32_t GetUniformCount(HProgram prog)
-    {
-        return g_functions.m_GetUniformCount(prog);
-    }
-    HUniformLocation GetUniformLocation(HProgram prog, const char* name)
-    {
-        return g_functions.m_GetUniformLocation(prog, name);
-    }
```
Jhonnyg, line 1377: Removed from adapters into the base section (and simplfied)

engine/graphics/src/null/graphics_null.cpp
```diff
@@ -1155,23 +1026,60 @@ namespace dmGraphics
...
+        uint32_t set               = UNIFORM_LOCATION_GET_VS(base_location);
+        uint32_t binding           = UNIFORM_LOCATION_GET_VS_MEMBER(base_location);
+        uint32_t buffer_offset     = UNIFORM_LOCATION_GET_FS(base_location);
+        assert(!(set == UNIFORM_LOCATION_MAX && binding == UNIFORM_LOCATION_MAX));
+
+        NullProgram* program            = (NullProgram*) context->m_Program;
+        ProgramResourceBinding& pgm_res = program->m_BaseProgram.m_ResourceBindings[set][binding];
+        uint32_t offset                 = pgm_res.m_DataOffset + buffer_offset;
+
+        WriteConstantData(program, offset, (uint8_t*) data, sizeof(dmVMath::Vector4) * count);
```
Jhonnyg, line 1064: this code could potentially now be merged with all graphics adapters except opengl, which is the odd one out.. :'(

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2320,23 +2357,222 @@ static void LogFrameBufferError(GLenum status)
...
+    // member 1: fs_uniforms.material.roughness;
+    //
+    // Case 2:
+    // We have the same, but we also use an "instance member" to rename the uniform (will also be useful later for arrayed uniforms):
+    // uniform fs_uniforms {
+    //     PBRMaterial material;
+    // } my_material;
+    // member 0: my_material.material.metallic;
+    // member 1: my_material.material.roughness;
+    static char* GetConstructedCanonicalName(BuildCanonicalPathsCallbackContext& context, char* str, uint32_t str_len, char* canonical_name_buffer, uint32_t canonical_name_buffer_len)
```
Jhonnyg, line 2487: Not sure about doing all of this work in runtime, but even if we had all the prepared paths in the data formats we still need to do this "fixup" of the paths because of how OpenGL returns the reflection data. It's only really needed for OpenGL, so I don't know if we want to do more work in the pipeline to avoid this?

Ideally I'd like to use the reflection as-is for OpenGL, but unfortunately we still need to generate the OpenGL reflection data to get the bind locations so we can bind resource to the correct shader bind points. I don't think we want OpenGL to be driving our data formats even further, since it's becoming more and more the odd one out.

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2320,23 +2357,222 @@ static void LogFrameBufferError(GLenum status)
...
+        //     PBRMaterial material;
+        // } my_material;
+        //
+        // OpenGL will yield the following binding:
+        // fs_uniforms.material
+        //
+        // But what we actually want is:
+        // my_material.material
+
+        const bool prepend_instance_name = false;
```
Jhonnyg, line 2568: For all other adapters we can add the instance name to the canonical path, but for OpenGL we can't..

engine/gamesys/src/gamesys/test/material/material_constant_array_v140.fp
```diff
@@ -0,0 +1,29 @@
+#version 140
+
+out vec4 FragColor;
+
+struct nested_nested_type
+{
+    vec4 tint[2];
+    vec4 leaf_value[2];
+    vec4 overridden_value[2];
```
Jhonnyg, line 9: Note: We can only support arrays for _leaf_ nodes in a nested type chain. This is because we don't have a way of addressing array elements earlier, e.g:

```
struct MyElement {
  vec4 data1;
  vec4 data2;
};

uniform MyUniforms {
  MyElement elements[4];  
};
```

We can't address alements[1].data1 for example, using go.set or a constant buffer.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderProgramBuilder.java
```diff
@@ -302,13 +302,20 @@ static public ShaderDesc.ResourceBinding.Builder SPIRVResourceToResourceBindingB
         ShaderDesc.ResourceType.Builder typeBuilder = getResourceTypeBuilder(res.type);
         resourceBindingBuilder.setType(typeBuilder);
         resourceBindingBuilder.setName(res.name);
-        resourceBindingBuilder.setNameHash(MurmurHash.hash64(res.name));
+        resourceBindingBuilder.setNameHash(res.nameHash);
+        resourceBindingBuilder.setId(res.id);
```
Jhonnyg, line 306: ID is used in the editor because of how the shaderc pipeline mangles resource names. If we have something like:

```
uniform my_uniform_buffer {
  vec4 thing;
  vec4 bong;
};
```

it will translate into something like:

```
struct my_uniform_buffer
{
 // ..
};
uniform my_uniform_buffer _14; // 14 is the ID here
```

The opengl reflection will now give us uniform names as:
_14.thing;
_14.bong;

So we need a way to address these uniforms, which is where the ID comes in.

editor/src/clj/editor/material.clj
```diff
@@ -145,14 +146,41 @@
...
+
 (defn- transpile-shader-source [shader-ext ^String shader-source ^long max-page-count]
   (let [shader-type (code.shader/shader-type-from-ext shader-ext)
         shader-language (code.shader/shader-language-to-java :language-glsl-sm120) ; use the old gles2 compatible shaders
         is-debug true
         result (ShaderProgramBuilderEditor/buildGLSLVariantTextureArray shader-source shader-type shader-language is-debug max-page-count)
         full-source (.source result)
-        array-sampler-names-array (.arraySamplers result)]
+        array-sampler-names-array (.arraySamplers result)
+        ^SPIRVReflector reflector (.reflector result)]
```
Jhonnyg, line 181: We have access to the reflection data now!

editor/src/clj/editor/gl/shader.clj
```diff
@@ -621,10 +642,9 @@ of GLSL strings and returns an object that satisfies GlBind and GlEnable."
                       (range (gl-shader-parameter gl program GL2/GL_ACTIVE_ATTRIBUTES)))
 
                 uniform-infos
-                (into {}
+                (into []
```
Jhonnyg, line 645: Uniform infos are just an array now since we need to do the lookup based on the :name and the :canonical-name to find a match.

matgis: This seems like it would be slower.

Can't we simply put the same `uniform-info` under both keys in the `uniform-infos` map? I.e.
```clj
uniform-infos
(into {}
      (mapcat (fn [^long uniform-index]
                (let [uniform-info (uniform-info gl program uniform-index)]
                  [(pair (:name uniform-info) uniform-info)
                   (pair (:canonical-name uniform-info) uniform-info)])))
      (range (gl-shader-parameter gl program GL2/GL_ACTIVE_UNIFORMS)))
```

editor/src/clj/editor/gl/shader.clj
```diff
@@ -634,21 +654,21 @@ of GLSL strings and returns an object that satisfies GlBind and GlEnable."
...
                 (->> sampler-name->uniform-names
                      (mapv (fn [[sampler-name uniform-names]]
                              (pair sampler-name
-                                   (uniform-infos (first uniform-names)))))
+                                   (uniform-name->uniform-infos (first uniform-names) uniform-infos))))
                      (sort-by (comp :index second))
                      (mapv first))]
             {:program program
-             :uniform-infos uniform-infos
+             :uniform-infos (strip-resource-namespaces uniform-infos resource-binding-namespaces)
```
matgis, line 671: Does this really need to happen in two phases?

It seems we could do this when we're constructing `uniform-infos` above?

editor/src/clj/editor/material.clj
```diff
@@ -145,14 +146,41 @@
...
+When crosscompiled to SM120 (which is used by the editor), we will get two uniforms:
+_<id>.material.diffuse
+_<id>.material.specular
+
+To be able to map this in a material constant, we need to strip the namespace from the
+reflected data when the shader is created (see editor.gl.shader:make-shader-program) since
+there is no way a user can know what the generated id will be for older shaders.
+"
+(defn- resource-binding-namespaces [^SPIRVReflector reflector]
+  ;; SSBOs will need the same remapping as UBOs, but since we don't support
```
matgis, line 170: As an editor person, I don't know what these acronyms mean. Maybe type them out fully?

editor/src/clj/editor/material.clj
matgis: Ohh, neat! Maybe we can move some of the logic out if the function we register with `scene-cache/request-object!` soon!

editor/src/clj/editor/material.clj
```diff
@@ -191,6 +219,12 @@
                     (:array-sampler-names augmented-vertex-shader-info)
                     (:array-sampler-names augmented-fragment-shader-info)))
 
+            resource-binding-namespaces
+            (into []
+                  (distinct (concat
+                              (:resource-binding-namespaces augmented-vertex-shader-info)
+                              (:resource-binding-namespaces augmented-fragment-shader-info))))
+
```
matgis, line 227: Why is it a vector? You're both creating it as a set and using it as a set. 😅

Although, the way we're using this, it seems we could turn it into a `Regex` here and simply use `string/replace` in `strip-resource-namespaces`.

editor/src/clj/editor/material.clj
```diff
@@ -145,14 +146,41 @@
...
+_<id>.material.specular
+
+To be able to map this in a material constant, we need to strip the namespace from the
+reflected data when the shader is created (see editor.gl.shader:make-shader-program) since
+there is no way a user can know what the generated id will be for older shaders.
+"
+(defn- resource-binding-namespaces [^SPIRVReflector reflector]
+  ;; SSBOs will need the same remapping as UBOs, but since we don't support
+  ;; them in the editor, we don't gather their namespaces here.
+  (mapv (fn [ubo] (str "_" (.id ubo))) (.getUBOs reflector)))
```
matgis, line 172: This would use reflection, since you're not type-tagging the `ubo` argument. Also, this would read better if it was split across multiple lines.

editor/src/clj/editor/material.clj
```diff
@@ -145,14 +146,41 @@
...
   (let [shader-type (code.shader/shader-type-from-ext shader-ext)
         shader-language (code.shader/shader-language-to-java :language-glsl-sm120) ; use the old gles2 compatible shaders
         is-debug true
         result (ShaderProgramBuilderEditor/buildGLSLVariantTextureArray shader-source shader-type shader-language is-debug max-page-count)
         full-source (.source result)
-        array-sampler-names-array (.arraySamplers result)]
+        array-sampler-names-array (.arraySamplers result)
+        ^SPIRVReflector reflector (.reflector result)]
     {:shader-source full-source
+     :resource-binding-namespaces (resource-binding-namespaces reflector)
```
matgis, line 183: Should it be a set here already? See my later comments.

matgis (commented):
Hmm, the editor bits could maybe be a bit more efficient?

engine/graphics/src/graphics.h
```diff
@@ -414,6 +414,17 @@ namespace dmGraphics
         uint32_t           m_ModificationVersion; // OpenGL
     };
 
+    struct Uniform
+    {
+        char*            m_Name; // Global/short name, e.g "my_member"
+        dmhash_t         m_NameHash;
+        char*            m_CanonicalName; // Full name, e.g "vsUniforms.some_struct.my_member"
```
JCash, line 421: Do we really need both?
If I have a global "tint", what will the canonical name be? 🤔

JCash (dismissed):
Lgtm.
I'm not entirely sure what distinguishes the "name" from a "canonical name" here, and raised the question if both are needed?

PR #9691, opened 2024-11-05
Detect DMFS (meant to operate like IDBFS or NODEFS) as a replacement …

smagnuso:
For my custom engine I'd like to avoid having all resources in memory, I've made it so that dmloader can detect a DMFS which will act as a replacement for in memory file access. This will have no impact on browser where we cannot implement such a thing.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
@@ -429,7 +429,9 @@ var GameArchiveLoader = {
         this.addListener(this._onFileLoadedListeners, callback);
     },
     notifyFileLoaded: function(file) {
-        this.notifyListeners(this._onFileLoadedListeners, { name: file.name, data: file.data });
+        if (file.data) {
+            this.notifyListeners(this._onFileLoadedListeners, { name: file.name, data: file.data });
+        }
     },
 
     addArchiveLoadedListener: function(callback) {
@@ -470,7 +472,11 @@ var GameArchiveLoader = {
             EngineLoader.loadAsmJsAsync(exeName);
             totalSize += EngineLoader.asmjs_size;
         }
-        this.downloadContent();
+        if (!Module['isDMFSSupported']) {
+            // we can download in parallel here because we will not rely on FS, otherwise
+            // we have to wait until after the [w]asm is loaded.
+            this.downloadContent();
+        }
         ProgressUpdater.resetCurrent();
         if (isWASMSupported) {
             EngineLoader.updateWasmInstantiateProgress(totalSize);
@@ -480,16 +486,27 @@ var GameArchiveLoader = {
 
     downloadContent: function() {
         var file = this._files[this._fileIndex];
-        // if the file consists of more than one piece we prepare an array to store the pieces in
-        if (file.pieces.length > 1) {
-            file.data = new Uint8Array(file.size);
+
+        if (Module['isDMFSSupported']) {
+            const path = `${DMSYS.GetUserPersistentDataRoot()}/${file.name}`;
+            try { // see if already and stored
+                const stat = FS.stat(path);
+                if(stat) {
+                    if (file.size == stat.size) { // should we verify hash too?
+                        this.onFileLoaded(file);
+                        return;
+                    }
+                }
+            } catch(_e) { }
+            file.stream = FS.open(path, "w");
         }
+
         // how many pieces to download at a time
         var limit = file.pieces.length;
         if (typeof this.MAX_CONCURRENT_XHR !== 'undefined') {
             limit = Math.min(limit, this.MAX_CONCURRENT_XHR);
         }
-        // download pieces
+
         for (var i=0; i<limit; ++i) {
             this.downloadPiece(file, i);
         }
@@ -532,9 +549,14 @@ var GameArchiveLoader = {
     },
 
     addPieceToFile: function(file, piece) {
-        if (1 == file.pieces.length) {
+        if (file.stream !== undefined) {
+            FS.write(file.stream, piece.data, 0, piece.data.length, piece.offset);
+        } else if (1 == file.pieces.length) {
             file.data = piece.data;
         } else {
+            if(!file.data) {
+               file.data = new Uint8Array(file.size);
+            }
             var start = piece.offset;
             var end = start + piece.data.length;
             if (0 > start) {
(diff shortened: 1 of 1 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -753,6 +777,12 @@ var Module = {
 
     setStatus: function(text) { console.log(text); },
 
+    isDMFSSupported: (function() {
```
britzl, line 780: Maybe a silly question but what does DMFS stand for?

smagnuso: I just saw the prefix DM all over, happy to call it whatever. :)

aglitchman:
Does this PR refer to WebGPU or does this PR refer to some other issue? The reason and context of why IDBFS needs to be replaced by something custom is unclear. 

**And you should always realise that IDBFS is supported and developed by Emscripten team. In contrast, supporting your own solution is extra time, work and needs to be covered with unit tests.

smagnuso:
> Does this PR refer to WebGPU or does this PR refer to some other issue? The reason and context of why IDBFS needs to be replaced by something custom is unclear.
> 

It is not webgpu specific, no. IDBFS access isn't entirely optimal if one has direct access to file system since the whole arcd is loaded into memory.

> **And you should always realise that IDBFS is supported and developed by Emscripten team. In contrast, supporting your own solution is extra time, work and needs to be covered with unit tests.

Agree that it is extra work to maintain the file system code, but that is not inside this patch, just mounting it. My platform will provide an implemtation that can be used in place of IDBFS.

aglitchman:
> It is not webgpu specific, no.

So you should first create an issue on the subject, then link to it in this PR. 

To me, as a person from the outside, i.e. from the community, this PR now looks just like your personal improvements and I don't understand the benefit of them for the community.

britzl:
> To me, as a person from the outside, i.e. from the community, this PR now looks just like your personal improvements and I don't understand the benefit of them for the community.

Not to worry @aglitchman, we are talking to @smagnuso about this in other channels.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
JCash: I think dmfs is fine since we will have to support this.
Since this is now an "api", we should also document it as well.
I.e. what functions are supported/required.
Is there like a base class we could model it on?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -553,6 +573,11 @@ var GameArchiveLoader = {
         ++file.totalLoadedPieces;
         // is all pieces of the file loaded?
         if (file.totalLoadedPieces == file.pieces.length) {
+            if (file.stream !== undefined) {
+                FS.close(file.stream);
+                file.stream = undefined;
+            }
+            this.verifyFile(file);
```
JCash, line 580: What's the rationale for moving the function call?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -470,7 +472,9 @@ var GameArchiveLoader = {
             EngineLoader.loadAsmJsAsync(exeName);
             totalSize += EngineLoader.asmjs_size;
         }
-        this.downloadContent();
+        if (!Module['isDMFSSupported']) {
```
britzl, line 475: Why is it done like this? Why can't we always call downloadContent() here?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
britzl: I'm struggling a bit with mapping in my mind how this will work and why the approach with dmfs is slightly different from the current one. Can we refactor it somehow so there are fewer branches? If I were to change anything in the functions of GameArchiveLoader I'd be worried that I would break things for one or the other solution. We also don't have any tests for any of this code which makes it even harder to reason about.

smagnuso: The reason the code had to be refactored is because we need to load the _wasm.js first as that is where the FS comes from and I need FS to figure out if we already had it on disk.

As for supporting any API at defold I am not thinking there is any new API, FS already supports mounts and this is just making use of those from emscripten. The only real difference is loading the _wasm.js first and mounting this instead of IDBFS.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
smagnuso: This moved because onFileDownloaded is called in two places now. One when the parts are downloaded and can be verified, the other were the parts have already been put back into a single file.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
smagnuso: We need to get the _wasm.js first

britzl: Ok, got it. Thanks.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
britzl: > The reason the code had to be refactored is because we need to load the _wasm.js first as that is where the FS comes from and I need FS to figure out if we already had it on disk

Thanks, that explains a lot!

> Agree testing coverage would be good, though unsure how that would be done given we cannot have file system access in general browser case. I'm happy to help out with a test, but just not sure where it would go to get the contents or how to structure the test

Yeah, I don't know where to start with this either. Let's hold off on it for now.

> but it seemed rather intention to do that in parallel.

Yes, we want to reduce loading time as much as possible so this can't be changed in the default case.

JCash (approved):

diiyw:
> For my custom engine I'd like to avoid having all resources in memory, I've made it so that dmloader can detect a DMFS which will act as a replacement for in memory file access. This will have no impact on browser where we cannot implement such a thing.

Hello, do you have a concrete DMFS implementation example? I'm having some issues with libnodefs.js after modifying Emscripten. Thank you! @smagnuso

PR #9701, opened 2024-11-06
Add euler/quat conversion functions to `vmath`

### LINKED ISSUE #8478: Add QuatToEuler and EulerToQuat in dmVMath
I spend a lot of time trying to impelemt quat->euler and euler->quat convertion in lua.
My functions worked but not always)

My cases:
1)For BlockyUniverse i use quat->euler convertion to make hp bar in gui always look in camera.(gui only worked with euler)
2) For Blocky Fighting i try to remove roll from quaterion. So i need to convert euler in quat, remove roll, convert euler to quat back.
( try to fix how items look in player hand, this not worked but i try that)

You already have that function in engine:
https://github.com/defold/defold/blob/a11256553ef4bce0f534c227769f6eacd1583ea3/engine/dlib/src/dlib/vmath.h#L82
https://github.com/defold/defold/blob/a11256553ef4bce0f534c227769f6eacd1583ea3/engine/dlib/src/dlib/vmath.h#L142

My suggestions:
1)Add QuatToEuler and EulerToQuat in native api <[dmsdk/dlib/vmath.h](https://github.com/defold/defold/blob/dev/engine/dlib/src/dmsdk/dlib/vmath.h)>
2)Add quat_to_euler and euler_to_quat in vmath(lua)

Workaround
copy QuatToEuler and EulerToQuat in native library

aglitchman:
Two new functions have been added to the `vmath` module to convert euler angles to quaternion and vice versa. Euler angles are often used to simplify handling the rotation of objects.

- `vmath.quat_to_euler` - converts a quaternion into euler angles.
- `vmath.euler_to_quat` - converts euler angles into a quaternion.

```lua
local q = vmath.quat_rotation_z(math.rad(90))
print(vmath.quat_to_euler(q)) --> 0 0 90

local q2 = vmath.euler_to_quat(0, 45, 90)
print(q2) --> vmath.quat(0.27059805393219, 0.27059805393219, 0.65328145027161, 0.65328145027161)
```

Fixes https://github.com/defold/defold/issues/8478

### DIFF
--- a/engine/script/src/script_vmath.cpp
+++ b/engine/script/src/script_vmath.cpp
@@ -2633,6 +2633,76 @@ namespace dmScript
         return 1;
     }
 
+    /*# converts a quaternion into euler angles
+     *
+     * Converts a quaternion into euler angles (r0, r1, r2), based on YZX rotation order.
+     * To handle gimbal lock (singularity at r1 ~ +/- 90 degrees), the cut off is at r0 = +/- 88.85 degrees, which snaps to +/- 90.
+     * The provided quaternion is expected to be normalized.
+     * The error is guaranteed to be less than +/- 0.02 degrees
+     *
+     * @name vmath.quat_to_euler
+     * @param q [type:quaternion] source quaternion
+     * @return x [type:number] euler angle x in degrees
+     * @return y [type:number] euler angle y in degrees
+     * @return z [type:number] euler angle z in degrees
+     * @example
+     * 
+     * ```lua
+     * local q = vmath.quat_rotation_z(math.rad(90))
+     * print(vmath.quat_to_euler(q)) --> 0 0 90
+     * 
+     * local q2 = vmath.quat_rotation_y(math.rad(45)) * vmath.quat_rotation_z(math.rad(90))
+     * local v = vmath.vector3(vmath.quat_to_euler(q2))
+     * print(v) --> vmath.vector3(0, 45, 90)
+     * ```
+     */
+    static int QuatToEuler(lua_State* L)
+    {
+        Quat* q = CheckQuat(L, 1);
+        Vector3 euler = dmVMath::QuatToEuler(q->getX(), q->getY(), q->getZ(), q->getW());
+        lua_pushnumber(L, euler.getX());
+        lua_pushnumber(L, euler.getY());
+        lua_pushnumber(L, euler.getZ());
+        return 3;
+    }
+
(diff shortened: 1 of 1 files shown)


aglitchman:
One more thing. Logically with other functions they should have been called `vmath.quat_euler` and `vmath.euler`, but then they would confuse developers with their naming. Maybe the Defold team has some ideas?

koenbollen:
Nice, this would be great to have! 👍

For the namin:
Maybe `quat_to_degrees` (and vise versa)? And also have `quat_to_radian` variants.

aglitchman:
> And also have `quat_to_radian` variants.

In the game industry (Unity/UE5/, I see that euler angles are always in degrees.

koenbollen:
> In the game industry (Unity/UE5/PlayCanvas), I see that euler angles are always in degrees.

fair, but what if you let go of euler? If I'm writing my game logics in radians, and I get a rotation-quaternion, I want to convert that rotation to radians (and the same with degrees). And if there is one method to go straight to the unit I need, it'll save a step, and a function call. (and I don't have to find out if 'to_euler' means degrees or radians).

aglitchman:
The request is mentioned in the first message - https://github.com/defold/defold/issues/8478 

The point of PR is that these two functions expose the C++ functions that exist in the engine and will always exist in the engine, since objects and gui have the "euler" property. If `dmVMath::QuatToEuler` and `dmVMath::EulerToQuat` didn't exist, I wouldn't be doing this PR.

Without these functions I usually do the conversion like that:

```lua
go.set(".", "rotation", quat)
local euler = go.get(".","euler")
```

And there are 3 other different Lua implementations (one of them is mine). But why do I need them, when this code is already in Defold and we can just let developers use it 🙌

About degrees and radians: in game object properties `euler` is in degrees, in C++ the function returns degrees, and as I pointed out above in the industry euler is in degrees. That looks ok to me.

JCash:
> The request is mentioned in the first message - https://github.com/defold/defold/issues/8478

Yeah, so sorry I didn't notice! (not sure how I missed it!)

As for Euler, it is ofc degrees, my concern was regarding the fact that in our _actual_ internal vmath headers (Sony's vector math), they don't deal with euler, only radians.
Problem is here where we've extended the vmath namespace with helper functions that don't conform fully to the previous api.
However, since rotation is problematic to work with on occasion, and we do already have the "euler" property, I think this change makes sense.

JCash (approved):

PR #9703, opened 2024-11-06
Avoid costly list removal when creating archive

britzl:
Avoid List.remove()

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ArchiveBuilder.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ArchiveBuilder.java
@@ -109,13 +109,13 @@ public class ArchiveBuilder {
     }
 
     // used in tests
-    public ArchiveEntry getArchiveEntry(int index) {
-        return this.entries.get(index);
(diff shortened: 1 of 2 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ArchiveBuilder.java
```diff
@@ -310,39 +315,41 @@ private void writeArchiveIndex(File archiveIndexFile) throws IOException {
...
-        logger.info("Creating archive data with a fixed thread pool executor using %d threads", nThreads);
+        logger.info("Creating archive entries with a fixed thread pool executor using %d threads", nThreads);
         this.executorService = Executors.newFixedThreadPool(nThreads);
         
         RandomAccessFile archiveData = new RandomAccessFile(archiveDataFile, "rw");
 
         Collections.sort(entries); // Since it has no hash, it sorts on path
 
+        excludedEntries = new ArrayList<>(entries.size());
+        includedEntries = new ArrayList<>(entries.size());
```
AGulev, line 340: Should both of them be entries.size()? 
Can we avoid creation of excludedEntries at all if build without liveupdate (no excluded entries)?

PR #9708, opened 2024-11-07
Allow separated dwarf symbols for wasm target in extender

smagnuso:
I'm debugging some wasm code and would like to get the symbols after extender relinks


AGulev:
I think @ekharkunov is working on it already

britzl (approved):

PR #9709, opened 2024-11-08
Transpiled source deletion no longer breaks the build

vlaaad:
Fixes defold/extension-teal#5

### DIFF
--- a/editor/src/clj/editor/code/transpilers.clj
+++ b/editor/src/clj/editor/code/transpilers.clj
@@ -25,6 +25,7 @@
             [editor.fs :as fs]
             [editor.graph-util :as gu]
             [editor.resource :as resource]
+            [editor.resource-io :as resource-io]
             [editor.resource-node :as resource-node]
             [editor.ui :as ui]
             [editor.workspace :as workspace]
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/code/transpilers.clj
```diff
@@ -70,54 +71,58 @@
...
+                            (let [resource (resource/make-file-resource workspace (str output-dir) file [] (constantly false))
+                                  build-targets (script-compilation/build-targets build-file-node-id resource (code.util/split-lines (slurp resource)) lua-preprocessors [] [] proj-path->node-id)]
+                              (pair (resource/proj-path resource) build-targets)))))
+                      (fs/file-walker output-dir false))))
+          (catch Exception e
+            (g/->error build-file-node-id :modified-lines :fatal (:resource build-file-save-data)
+                       (str "Compilation failed: " (ex-message e))))
+          (finally
+            (when-not use-project-dir
+              (fs/delete-directory! source-dir {:fail :silently}))))))))
```
vlaaad, line 115: The only change here is that I wrapped the whole form in `(g/precluding-errors source-code-save-datas ...)`, so that other errors still propagate.

editor/src/clj/editor/code/transpilers.clj
```diff
@@ -70,54 +71,58 @@
...
+(defn- array-substitute-remove-file-not-found-errors [coll]
+  (into [] (remove resource-io/file-not-found-error?) coll))
 
 (g/defnode TranspilerNode
   (property build-file-proj-path g/Str)
   (property instance ILuaTranspiler)
   (input root g/Str)
   (input build-file-save-data g/Any)
-  (input source-code-save-datas g/Any :array)
+  (input source-code-save-datas g/Any :array :substitute array-substitute-remove-file-not-found-errors)
```
vlaaad, line 125: We keep the nodes of deleted files around, but we didn't remove them here previously. This caused the deletion of a transpiled file to fail the build.

AGulev (approved):
👍

mikatuo:
Awesome! ⭐

PR #9714, opened 2024-11-08
Ship off batches of particles to minimize RAM overhead

smagnuso:
Introduced a CPU limit and upload in batches to the GPU up to the limit. I think we could create the buffer as it grows up to the max rather than reallocating, but at least with this a smaller number can be chosen than enough for worst case.

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_gui.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_gui.cpp
@@ -124,6 +124,7 @@ namespace dmGameSystem
         uint32_t                    m_MaxGuiComponents;
         uint32_t                    m_MaxParticleFXCount;
         uint32_t                    m_MaxParticleCount;
+        uint32_t                    m_MaxParticleBufferCount;
         uint32_t                    m_MaxAnimationCount;
     };
 
@@ -378,6 +379,7 @@ namespace dmGameSystem
 
         gui_world->m_MaxParticleFXCount = gui_context->m_MaxParticleFXCount;
         gui_world->m_MaxParticleCount = gui_context->m_MaxParticleCount;
+        gui_world->m_MaxParticleBufferCount = gui_context->m_MaxParticleBufferCount;
         gui_world->m_ParticleContext = dmParticle::CreateContext(gui_world->m_MaxParticleFXCount, gui_world->m_MaxParticleCount);
         gui_world->m_MaxAnimationCount = gui_context->m_MaxAnimationCount;
 
@@ -3095,6 +3097,7 @@ namespace dmGameSystem
         gui_context->m_MaxParticleFXCount = dmConfigFile::GetInt(ctx->m_Config, "gui.max_particlefx_count", 64);
         gui_context->m_MaxParticleCount = dmConfigFile::GetInt(ctx->m_Config, "gui.max_particle_count", 1024);
         gui_context->m_MaxAnimationCount = dmConfigFile::GetInt(ctx->m_Config, "gui.max_animation_count", 1024);
+        gui_context->m_MaxParticleBufferCount = dmConfigFile::GetInt(ctx->m_Config, "gui.max_particle_buffer_count", 1024);
 
         int32_t max_gui_count = dmConfigFile::GetInt(ctx->m_Config, "gui.max_instance_count", 128);
         gui_context->m_Worlds.SetCapacity(max_gui_count);
--- a/engine/gamesys/src/gamesys/components/comp_particlefx.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_particlefx.cpp
@@ -84,6 +84,7 @@ namespace dmGameSystem
         uint32_t                                m_DispatchCount;
         float                                   m_DT;
         uint32_t                                m_WarnOutOfROs : 1;
+        uint32_t                                m_WarnParticlesExceeded : 1;
     };
 
     dmGameObject::CreateResult CompParticleFXNewWorld(const dmGameObject::ComponentNewWorldParams& params)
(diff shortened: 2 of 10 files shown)


engine/gamesys/src/gamesys/components/comp_particlefx.cpp
```diff
@@ -396,9 +400,33 @@ namespace dmGameSystem
                 pfx_world->m_VertexBufferData.SetSize(0);
                 pfx_world->m_RenderObjects.SetSize(0);
                 break;
-            case dmRender::RENDER_LIST_OPERATION_BATCH:
-                RenderBatch(pfx_world, params.m_Context, params.m_Buf, params.m_Begin, params.m_End);
-                break;
+            case dmRender::RENDER_LIST_OPERATION_BATCH: {
```
JCash, line 403: I think you need to describe the problem and the solution in more detail, as I don't really know what we're trying to solve here.

smagnuso: We've discussed some out of band. However, to give context here: I'm finding that max_particle_count can be very large in a game to accomodate worst case emitter and there is a copy in CPU as well as GPU. This patch is meant to keep the CPU side at max_particle_count and instead grow the GPU side to accomodate worst case rather than have them match and just drop on the floor.

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
```diff
@@ -396,9 +399,43 @@ namespace dmGameSystem
...
+                    dmRender::SetBufferData(params.m_Context, pfx_world->m_VertexBuffer, vx_size, 0, dmGraphics::BUFFER_USAGE_STREAM_DRAW);
+
+                    //do the batching
+                    uint32_t vx_offset = 0;
+                    while (vx_offset < vx_size)
+                    {
+                        dmRender::SetBufferSubData(params.m_Context, pfx_world->m_VertexBuffer, vx_offset, pfx_world->m_VertexBufferData.Size(), pfx_world->m_VertexBufferData.Begin());
+                        vx_offset += pfx_world->m_VertexBufferData.Size();
+                        pfx_world->m_VertexBufferData.SetSize(0);
+                        if (!RenderBatch(pfx_world, params.m_Context, params.m_Buf, &batch_begin, params.m_End))
```
JCash, line 428: The RenderBatch() has side effects, such as clearing the render constants and also submitting one draw call each, which is not what you wanted.

So, I suggest that you put the "looping" inside the RenderBatch() function.

JCash (changes requested):
The RenderBatch should only be called once per batch.

engine/particle/src/particle.cpp
```diff
@@ -1314,12 +1316,9 @@ namespace dmParticle
 
         if (j < particle_count)
         {
-            if (emitter->m_RenderWarning == 0)
-            {
-                res = GENERATE_VERTEX_DATA_MAX_PARTICLES_EXCEEDED;
-                emitter->m_RenderWarning = 1;
-            }
+            res = GENERATE_VERTEX_DATA_MAX_PARTICLES_EXCEEDED;
```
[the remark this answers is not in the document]
smagnuso: I'm unsure this is useful since the engine can handle the case and create small cpu size and we can handle any number of particules gpu side. The solve to that is to produce fewer particles.

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
smagnuso: I agree, done.

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
```diff
@@ -330,23 +331,42 @@ namespace dmGameSystem
...
+        if (vb_offset) { // consume the last chunk of data
+            if (vb_size) {
+                dmRender::SetBufferSubData(render_context, pfx_world->m_VertexBuffer, vb_offset, vb_size, vertex_buffer.Begin());
+                vb_offset += (vb_size - vb_size_init);
+            }
+            ro_vertex_count = vb_offset / material_attribute_info.m_VertexStride;
+            vertex_buffer.SetSize(0);
+        } else {
+            ro_vertex_count = (vb_size - vb_size_init) / material_attribute_info.m_VertexStride;
+            vertex_buffer.SetSize(vb_size);
```
JCash, line 368: At this last stage, the vertex count should be possible to be calculated in the same fashion?

Also, I don't really understand why the buffer is set to two different sizes (0 vs vb_size)?

engine/particle/src/particle.cpp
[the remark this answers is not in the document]
JCash: > and we can handle any number of particules gpu side.

The point with our numbers is that the developer should have control over their budgets.

True, you've minimized CPU budget (and that should probably be a new setting in the game.project),
but, the GPU budget still stands.

With this change, you change how we work in general with memory.

> The solve to that is to produce fewer particles.

The user will not realize that they're not seeing all their particles anymore if they exceed the limit.
The particles will just silently be dropped, and that's not ok.

JCash (changes requested):

engine/particle/src/particle.cpp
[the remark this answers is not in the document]
smagnuso: Just to be clear, with this patch they are not silently dropped. However we discussed offline and agreed to add a cap on the GPU side buffer and then return to having them silently dropped (after a single warning).

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
smagnuso: In the top case the data has already been uploaded in batches so should not be done after.

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
```diff
@@ -330,23 +335,52 @@ namespace dmGameSystem
...
-                    dmLogWarning("Cannot generate vertex data for emitter (%d), particle instance handle is invalid.", (*i));
+                if (!vb_offset) { // need to set the size of the buffer
+                    uint32_t vx_size = 0;
+                    for (uint32_t *t = begin; t != end; ++t)
+                    {
+                        const dmParticle::EmitterRenderData* ed = (dmParticle::EmitterRenderData*) buf[*t].m_UserData;
+                        vx_size += GetEmitterVertexCount(pfx_world->m_ParticleContext, ed->m_Instance, ed->m_EmitterIndex);
+                    }
+                    if (vx_size >= pfx_context->m_MaxParticleCount) {
+                        if (!pfx_world->m_WarnParticlesExceeded) {
```
JCash, line 348: 👍

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -3095,6 +3097,7 @@ namespace dmGameSystem
         gui_context->m_MaxParticleFXCount = dmConfigFile::GetInt(ctx->m_Config, "gui.max_particlefx_count", 64);
         gui_context->m_MaxParticleCount = dmConfigFile::GetInt(ctx->m_Config, "gui.max_particle_count", 1024);
         gui_context->m_MaxAnimationCount = dmConfigFile::GetInt(ctx->m_Config, "gui.max_animation_count", 1024);
+        gui_context->m_MaxParticleBufferCount = dmConfigFile::GetInt(ctx->m_Config, "gui.max_particle_buffer_count", 1024);
```
JCash, line 3100: This reminds me that the same loop for particles in comp_gui.cpp hasn't been updated with the same changes.

JCash (approved):

PR #9728, opened 2024-11-12
Wait for the reload to finish before building

### LINKED ISSUE #6681: Wait for resource sync to finish before starting the build
**Task (REQUIRED):**
When we edit project files in an external editor and then switch to it to run the game, we might end up in a situation where we start the build while the resource sync is still running. The external changes will not be picked up, which is not what the user wants. We should ensure all external changes are loaded before we start the build.

Starting the build when we perform resource sync should wait until the sync completes so the build picks up all external changes.

vlaaad:
When the user focuses on the editor, it performs a resource reload. On large projects this may take a while, and if the user starts a build, the build will use outdated resources. This change fixes the issue by waiting for the reload to finish before starting the build.

Fixes #6681

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -1047,8 +1047,8 @@
           (result-fn project-build-results)
           nil)
 
-        phase-6-await-lint!
-        (fn phase-6-await-lint! [project-build-results]
+        phase-7-await-lint!
+        (fn phase-7-await-lint! [project-build-results]
           (if lint
             (do
               (render-progress! (progress/make-indeterminate "Linting code..."))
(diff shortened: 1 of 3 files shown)


editor/test/editor/analytics_test.clj
```diff
@@ -73,14 +73,15 @@
...
                                             sys/defold-version (constantly "0.1.234")]
-                                (with-config! {:cid mock-cid}
-                                  (analytics/start! validation-server-url send-interval)
-                                  (analytics/track-event! "test-category" "test-action")
-                                  (analytics/track-event! "test-category" "test-action" "test-label")
-                                  (analytics/track-exception! (NullPointerException.))
-                                  (analytics/track-screen! "test-screen-name")
-                                  (analytics/shutdown! shutdown-timeout)
-                                  @validation-messages-atom)))]
+                                (log/without-logging
```
vlaaad, line 76: Totally unrelated change, but it was annoying to see this stack trace in the tests.

editor/src/clj/editor/app_view.clj
```diff
@@ -1178,17 +1178,25 @@
               (fn process-pre-build-hook-results-on-ui-thread! [extension-error]
                 (if (some? extension-error)
                   (finish-with-result! {:error extension-error})
-                  (phase-2-start-all-build-processes!))))))]
+                  (phase-3-start-all-build-processes!))))))
```
vlaaad, line 1181: All changes above are plain renames of phases because I introduced a new phase at the start.

vlaaad:
@AGulev Build command will become unavailable. I considered showing an indicator in the bottom-left corner, but since the reload is already in progress, there is already an indicator that the editor does stuff. It will be replaced with the usual build indicators as soon as the reload is finished.

vlaaad:
@matgis This is tricky to test without a big project that has a lot of files... To reproduce it, I inserted `(Thread/sleep 5000)` in `editor.disk` on a line below `(render-progress! (progress/make-indeterminate "Loading external changes..."))`.

matgis (approved):
LGTM!

PR #9731, opened 2024-11-13
Add digest verification (sha1) for files in the archive_files.json

smagnuso:
This is an attempt to start verifying files in the archive from network.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
@@ -581,6 +581,8 @@ public class Bob {
 
                 opt("l", "liveupdate", ONE, "Yes if liveupdate content should be published", true),
 
+                opt(null, "with-sha1", ZERO, "Generate (and verify) sha1 signatures from build artifacts (when bunding for web)", false),
+
                 opt("ar", "architectures", ONE, "Comma separated list of architectures to include for the platform", true),
 
                 opt(null, "settings", MANY, ABS_OR_CWD_REL_PATH, "Path to a game project settings file. The settings files are applied left to right", false),
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
@@ -429,9 +429,7 @@ var GameArchiveLoader = {
         this.addListener(this._onFileLoadedListeners, callback);
     },
     notifyFileLoaded: function(file) {
-        if (file.data) {
-            this.notifyListeners(this._onFileLoadedListeners, { name: file.name, data: file.data });
-        }
+        this.notifyListeners(this._onFileLoadedListeners, { name: file.name, data: file.data });
     },
 
     addArchiveLoadedListener: function(callback) {
@@ -484,7 +482,7 @@ var GameArchiveLoader = {
         ProgressUpdater.setupTotal(totalSize + EngineLoader.wasm_instantiate_progress);
     },
 
-    downloadContent: function() {
+    downloadContent: async function() {
         var file = this._files[this._fileIndex];
 
         if (Module['isDMFSSupported']) {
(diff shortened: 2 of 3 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -484,21 +482,37 @@ var GameArchiveLoader = {
...
+                                    const digest = await window.crypto.subtle.digest("SHA-1", mmap);
+                                    matches = Array.from(new Uint8Array(digest)).map(b => b.toString(16).padStart(2, '0')).join('') == file.sha1;
+                                }
+                            } catch(e) { }
+                            FS.close(stream);
+                        } else {
+                            matches = false;
+                        }
+                    }
+                    if (matches) {
```
britzl, line 509: What kind of impact will this verification have on load times? We do not wish to slow down loading of content if we can avoid it. Can we hide this behind an option in game.project?

britzl (commented):
See question about impact on load time

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
smagnuso: Certainly adds some cost, I've pushed change to have it be optional

britzl: Great, but how do I disable it? We need a game.project setting and a check in HTML5Bundler too don't we?

You need an entry in com/dynamo/bob/meta.properties and in editor/resources/edn.properties. The default value should be off.

britzl (changes requested):
Question about how to disable sha1 verification

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
smagnuso: Yes, I made the default value off. I am just passing --with-sha1 into bob.jar currently. Do you want it to show up in the editor?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -581,6 +581,8 @@ public static List<CommandLineOption> getCommandLineOptions() {
 
                 opt("l", "liveupdate", ONE, "Yes if liveupdate content should be published", true),
 
+                opt(null, "with-sha1", ZERO, "Generate (and verify) sha1 signatures from build artifacts", false),
```
britzl, line 584: How will a user know this only applies to html5 builds?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
britzl: Now that I think about, it's probably fine that it is an option only available in command line builds. But the name of the option doesn't tell a user that it only applies to html5 builds. We should at least write that when printing the help.

britzl (commented):
See comment about platform availability

AGulev:
This option should be specified in game.rpoejct, not as bob argument 
How it's added (ignore cpp) https://github.com/defold/defold/commit/970042be550ad43b7d71c16c9f290d50b15dd518
and then it will be replaced in dmloader using templater: https://github.com/defold/defold/blob/1645c043da32d8b31397627532897ea974c02935/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js#L40

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
smagnuso: Added note to description

PR #9733, opened 2024-11-13
Use lua_pushliteral instead of lua_pushstring where possible

### LINKED ISSUE #9732: Use lua_pushliteral instead of lua_pushstring where possible
**Task (REQUIRED):**

`lua_pushliteral` avoids an unnecessary `strlen` call, so should be preferred over `lua_pushstring` for literals.

Change the code accordingly.

redoak:
skip release notes

Resolves #9732

### Technical changes

* No changes to behavior — simply replaces lua_pushstring with lua_pushliteral where possible.

### DIFF
--- a/engine/crash/src/script_crash.cpp
+++ b/engine/crash/src/script_crash.cpp
@@ -170,13 +170,13 @@ namespace dmCrash
             lua_pushnumber(L, i+1);
 
             lua_newtable(L);
-            lua_pushstring(L, "name");
+            lua_pushliteral(L, "name");
             lua_pushstring(L, name);
             lua_settable(L, -3);
 
             char str[64];
             sprintf(str, "%p", addr);
-            lua_pushstring(L, "address");
+            lua_pushliteral(L, "address");
             lua_pushstring(L, str);
             lua_settable(L, -3);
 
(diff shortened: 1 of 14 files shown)


engine/crash/src/script_crash.cpp
```diff
@@ -170,13 +170,13 @@ namespace dmCrash
             lua_pushnumber(L, i+1);
 
             lua_newtable(L);
-            lua_pushstring(L, "name");
+            lua_pushliteral(L, "name");
```
JCash, line 173: Thanks! I wasn't really aware of this macro before.

Since you asked questions about contributions offline, I'll elaborate some more here.
While the change is technically "idiomatic" Lua, the benefits are also quite small in terms of performance.
We sometimes leave things just because it's not a special gain, nor is it broken.
This PR lands on the plus side overall, as there is an actual official macro, and nothing else is needed.

JCash (approved):

engine/crash/src/script_crash.cpp
redoak: After your comments regarding build size for #9741, I checked size on this one too (`lua_pushliteral` is a macro that does `sizeof` like `dmHashLiteral`). It increased by ~500b.

Given this, you may want to revert this PR. The only instances that are hit often are the ones in the input code. But that's still only a few times each frame.

PR #9737, opened 2024-11-13
Use dmHashBuffer64 instead of dmHashString64 where string length is known

### LINKED ISSUE #9736: Use dmHashBuffer64 instead of dmHashString64 where string length is known
**Task (REQUIRED):**

`dmHashBuffer64` (which has a length argument) avoids an unnecessary `strlen` call, so should be preferred over `dmHashString64` (which calls `dmHashBuffer64` with `strlen`).

Change the code accordingly, where length can easily be retrieved.

redoak:
skip release notes

Resolves #9736

### Technical changes

* No changes to actual behavior — simply makes use of string length where it's easy enough to do.

### DIFF
--- a/engine/dlib/src/dlib/http_cache.cpp
+++ b/engine/dlib/src/dlib/http_cache.cpp
@@ -417,11 +417,12 @@ namespace dmHttpCache
             return RESULT_INVAL;
         }
 
-        uint64_t uri_hash = dmHashString64(uri);
+        size_t uri_length = strlen(uri);
+        uint64_t uri_hash = dmHashBuffer64(uri, uri_length);
 
         HashState64 hash_state;
         dmHashInit64(&hash_state, false);
-        dmHashUpdateBuffer64(&hash_state, uri, strlen(uri));
+        dmHashUpdateBuffer64(&hash_state, uri, uri_length);
         dmHashUpdateBuffer64(&hash_state, etag, strlen(etag));
         uint64_t identifier_hash = dmHashFinal64(&hash_state);
 
@@ -670,13 +671,14 @@ namespace dmHttpCache
     {
         dmMutex::ScopedLock lock(cache->m_Mutex);
 
+        size_t uri_length = strlen(uri);
         HashState64 hash_state;
         dmHashInit64(&hash_state, false);
-        dmHashUpdateBuffer64(&hash_state, uri, strlen(uri));
+        dmHashUpdateBuffer64(&hash_state, uri, uri_length);
         dmHashUpdateBuffer64(&hash_state, etag, strlen(etag));
         uint64_t identifier_hash = dmHashFinal64(&hash_state);
 
-        uint64_t uri_hash = dmHashString64(uri);
+        uint64_t uri_hash = dmHashBuffer64(uri, uri_length);
         Entry* entry = cache->m_CacheTable.Get(uri_hash);
         if (entry != 0 && entry->m_Info.m_IdentifierHash == identifier_hash)
         {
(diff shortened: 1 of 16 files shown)


engine/dlib/src/dlib/http_cache.cpp
```diff
@@ -417,11 +417,12 @@ namespace dmHttpCache
...
         }
 
-        uint64_t uri_hash = dmHashString64(uri);
+        size_t uri_length = strlen(uri);
+        uint64_t uri_hash = dmHashBuffer64(uri, uri_length);
 
         HashState64 hash_state;
         dmHashInit64(&hash_state, false);
-        dmHashUpdateBuffer64(&hash_state, uri, strlen(uri));
+        dmHashUpdateBuffer64(&hash_state, uri, uri_length);
```
JCash, line 425: This is a change with little real gain.
Note that you are trading runtime with code size here.

What is the size of the engine before/after your changes?
Note that every bit counts, and it's a huge selling point for our users.

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -969,8 +969,8 @@ namespace dmGameObject
     dmhash_t ConstructInstanceId(uint32_t index)
     {
         char buffer[16] = { 0 };
-        dmSnPrintf(buffer, sizeof(buffer), "%sinstance%d", ID_SEPARATOR, index);
-        return dmHashString64(buffer);
+        int length = dmSnPrintf(buffer, sizeof(buffer), "%sinstance%d", ID_SEPARATOR, index);
+        return dmHashBuffer64(buffer, length);
```
JCash, line 973: This is a good easy update.

engine/gameobject/src/gameobject/gameobject_props_lua.cpp
```diff
@@ -223,7 +223,9 @@ namespace dmGameObject
             if (lua_isstring(L, -2))
             {
                 void* userdata = 0;
-                dmhash_t id = dmHashString64(lua_tostring(L, -2));
+                size_t length;
+                const char* string = lua_tolstring(L, -2, &length);
+                dmhash_t id = dmHashBuffer64(string, length);
```
JCash, line 228: Aren't you getting warnings when bulding this, about casting from size_t to uint32_t?
Please avoid adding more warnings to the build.

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -2259,8 +2259,8 @@ namespace dmGameSystem
 
     static inline dmhash_t ResolveDynamicTexturePath(GuiComponent* component, dmhash_t path_hash, char* buffer, uint32_t buffer_size)
     {
-        dmSnPrintf(buffer, buffer_size, "%s/%llu.texturec", component->m_Resource->m_Path, (unsigned long long) path_hash);
-        return dmHashString64(buffer);
+        int length = dmSnPrintf(buffer, buffer_size, "%s/%llu.texturec", component->m_Resource->m_Path, (unsigned long long) path_hash);
+        return dmHashBuffer64(buffer, length < 0 ? buffer_size - 1 : length);
```
JCash, line 2263: Same as before. Little gain here, but losing readability and adding risk of errors. (and also adding code)

JCash (changes requested):
A few unneeded changes.
Also I'm concerned about added compile warnings about casting.

engine/dlib/src/dlib/http_cache.cpp
redoak: Good point. Size increase will not be worth it, so I will close this PR.

I think a better approach to reduce build size, while also avoiding `strlen` calls, would be to use functions like `dmScript::CheckHashOrString()`. All blocks that look something like this:

```cpp
dmhash_t property_hash;
if (dmScript::IsHash(L, 2)) {
    property_hash = dmScript::CheckHash(L, 2);
} else {
    property_hash = dmHashString64(luaL_checkstring(L, 2));
}
```

```cpp
dmhash_t property_id;
if (lua_isstring(L, 2)) {
    property_id = dmHashString64(lua_tostring(L, 2));
} else {
    property_id = dmScript::CheckHash(L, 2);
}
```

Could be replaced by a single function call. 

And if one really wants to, replace instances of `dmHashString64(lua_tostring(L, 2))` with a call to a function that does the `lua_tolstring` to `dmHashBuffer64` dance to avoid `strlen`.

engine/gameobject/src/gameobject/gameobject.cpp
redoak: I will make a separate PR for this one change and close this PR.

engine/gameobject/src/gameobject/gameobject_props_lua.cpp
redoak: I didn't notice any. But I probably lost them among all the output.

redoak:
Closing as per discussion above.

PR #9739, opened 2024-11-13
Use sizeof instead of strlen for immutable char arrays

### LINKED ISSUE #9738: Use sizeof instead of strlen for immutable char arrays
**Task (REQUIRED):**

`sizeof` avoids an unnecessary `strlen` call when getting the length of a char array/string, so should be preferred.

Change the code accordingly.

redoak:
skip release notes

Resolves #9738

### Technical changes

* No changes to behavior — simply uses `sizeof` instead of `strlen` where possible.

### DIFF
--- a/engine/dlib/src/basis/encoder/basisu_comp.cpp
+++ b/engine/dlib/src/basis/encoder/basisu_comp.cpp
@@ -1984,18 +1984,20 @@ namespace basisu
 		basist::ktx2_transcoder::key_value_vec key_values(m_params.m_ktx2_key_values);
 		key_values.enlarge(1);
 		
-		const char* pKTXwriter = "KTXwriter";
-		key_values.back().m_key.resize(strlen(pKTXwriter) + 1);
-		memcpy(key_values.back().m_key.data(), pKTXwriter, strlen(pKTXwriter) + 1);
+		const char pKTXwriter[] = "KTXwriter";
+		key_values.back().m_key.resize(sizeof(pKTXwriter));
+		memcpy(key_values.back().m_key.data(), pKTXwriter, sizeof(pKTXwriter));
 
 		char writer_id[128];
+		int writer_id_length =
 #ifdef _MSC_VER
-		sprintf_s(writer_id, sizeof(writer_id), "Basis Universal %s", BASISU_LIB_VERSION_STRING);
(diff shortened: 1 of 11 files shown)


engine/dlib/src/basis/encoder/basisu_comp.cpp
```diff
@@ -1984,18 +1984,20 @@ namespace basisu
 		basist::ktx2_transcoder::key_value_vec key_values(m_params.m_ktx2_key_values);
 		key_values.enlarge(1);
 		
-		const char* pKTXwriter = "KTXwriter";
-		key_values.back().m_key.resize(strlen(pKTXwriter) + 1);
-		memcpy(key_values.back().m_key.data(), pKTXwriter, strlen(pKTXwriter) + 1);
+		const char pKTXwriter[] = "KTXwriter";
```
JCash, line 1987: No. This is a 3rd party library.
Please go to their site to add changes.

engine/dlib/src/dlib/log.cpp
```diff
@@ -45,7 +45,7 @@
 namespace dmLog
 {
 
-const char* LOG_OUTPUT_TRUNCATED_MESSAGE = "...\n[Output truncated]\n";
+const char LOG_OUTPUT_TRUNCATED_MESSAGE[] = "...\n[Output truncated]\n";
```
JCash, line 48: Same as before. No need for this change.

engine/dlib/src/dlib/log.cpp
```diff
@@ -237,15 +237,15 @@ static void dmLogUpdateNetwork()
                 if (connections_full)
                 {
                     dmLogError("Too many log connections opened");
-                    const char* resp = "1 Too many log connections opened\n";
-                    SendAll(client_socket, resp, strlen(resp));
+                    const char resp[] = "1 Too many log connections opened\n";
```
JCash, line 240: No need to change to array.

JCash (changes requested):
A lot of changes that are not improving things significantly.
Also, the behavioral change in the for loop is not something we want to encourage, so please revert that.

engine/dlib/src/dlib/log.cpp
redoak: Changing to array makes `sizeof` give the size of the array and not the pointer.

redoak:
I'm closing this PR. I agree with the issues you point out, and after removing/fixing those there wouldn't be enough left to warrant a code change, IMO.

PR #9741, opened 2024-11-13
Add string hash function dmHashLiteral to avoid strlen call, like lua_pushliteral

### LINKED ISSUE #9740: Add string hash function dmHashLiteral to avoid strlen call, like lua_pushliteral
**Task (REQUIRED):**

The length of literals and char arrays can be retrieved with `sizeof`. Passing length to `dmHashBuffer64` avoids the `strlen` call of `dmHashString64`. A `dmHashLiteral64` macro or function, like `lua_pushliteral`, that calls `sizeof` for the user would be convenient.

Add the code accordingly, and use it where applicable.

redoak:
skip release notes

Resolves #9740

### Technical changes

* Added macros `dmHashLiteral32` and `dmHashLiteral64` in hash.h.
* No changes to behavior — simply uses the new macro where possible.

### DIFF
--- a/engine/dlib/src/dmsdk/dlib/hash.h
+++ b/engine/dlib/src/dmsdk/dlib/hash.h
@@ -84,6 +84,7 @@ DM_DLLEXPORT uint64_t dmHashBuffer64(const void* buffer, uint32_t buffer_len);
  * @return hash [type:uint32_t] hash value
  */
 DM_DLLEXPORT uint32_t dmHashString32(const char* string);
+#define dmHashLiteral32(string) dmHashBuffer32((string), sizeof((string)) - 1)
 
 
 /*# calculate 64-bit hash value from string
@@ -93,6 +94,7 @@ DM_DLLEXPORT uint32_t dmHashString32(const char* string);
  * @return hash [type:uint64_t] hash value
  */
 DM_DLLEXPORT uint64_t dmHashString64(const char* string);
+#define dmHashLiteral64(string) dmHashBuffer64((string), sizeof((string)) - 1)
 
 
 /*# get string value from hash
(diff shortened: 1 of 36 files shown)


redoak:
I went with macros for `dmHashLiteral` rather than template functions or some other construct. Let me know what you prefer.

I named them like functions instead of all uppercase. Let me know what you prefer.

I haven't added documentation for the functions yet, but will do once there's agreement on if and how to implement these.

engine/dlib/src/dmsdk/dlib/hash.h
```diff
@@ -84,6 +84,7 @@ DM_DLLEXPORT uint64_t dmHashBuffer64(const void* buffer, uint32_t buffer_len);
  * @return hash [type:uint32_t] hash value
  */
 DM_DLLEXPORT uint32_t dmHashString32(const char* string);
+#define dmHashLiteral32(string) dmHashBuffer32((string), sizeof((string)) - 1)
```
JCash, line 87: The PR title is wrong, as it isn't a function, but a macro.

JCash:
Macro is likely preferred here.

What are the results?
Code size before/after?
Runtime performance during startup difference?
Which platforms did you measure on?

engine/dlib/src/dmsdk/dlib/hash.h
JCash: Also, our code style says you should use CAPS for macros: `DM_HASH_LITERAL_32` (which arguably isn't as good looking, but we should try to stay consistent.

redoak:
> Macro is likely preferred here.
> 
> What are the results? Code size before/after? Runtime performance during startup difference? Which platforms did you measure on?

I checked the size now on Windows, and it increased by ~1.2kb. Didn't think of it, but of course for each invocation of the macro it will need to store the result. 8b per invocation gives ~1.5kb with the number of lines I changed.

I'm closing this PR; it's not worth the cost. If any instance is in a hot spot where `strlen` has impact, it can always be changed when discovered.

Issue #9743, opened 2024-11-13
Use sizeof instead of strlen for elements of arrays of string literals


redoak:
**Task (REQUIRED):**

`sizeof` avoids an unnecessary `strlen` call when getting the length of a string literal (change proposed in #9738).

We define const arrays of literals in several places, and at runtime pass one of them to `dmHashString64`:

```cpp
const char* properties_common[] = { "type", "custom_type", "id", "pivot" };
//...
pit->m_Property.m_NameHash = dmHashString64(properties_common[index]);
```

Since the length is known, a call to `strlen` can be avoided by using `dmHashBuffer64` (see #9736).

Since the length values need to be stored in an intermediate location until use, new macros, functions and/or structs are needed.

See #9744 for rough draft of a possible implementation.

**Expected outcome (REQUIRED):**

New constructs that make it easy to work with arrays of literals and `dmHashBuffer64`.

redoak:
Closed after discussion in the linked PR.

PR #9744, opened 2024-11-13
Use sizeof instead of strlen for elements of arrays of string literals

redoak:
This is only to illustrate what I mean by #9743, and a rough draft for a possible implementation.

There are several other locations where this would be applied — I only implemented it in this one function to get feedback.

Having some kind of string view struct that holds the char* and string length would be useful. If one already exist, let me know.

As in #9741 I went with a macro for constructing the string view. The macro `dmHashStrView64` could just as well be a function, though.

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_gui.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_gui.cpp
@@ -2895,6 +2895,10 @@ namespace dmGameSystem
 
     static bool CompGuiIterPropertiesGetNext(dmGameObject::SceneNodePropertyIterator* pit)
     {
+        struct StrView { const char* str; size_t len; };
+#define STR_VIEW_LITERAL(literal) {(literal), sizeof((literal)) - 1}
+#define dmHashStrView64(str_view) dmHashBuffer64((str_view).str, (str_view).len)
+
         if (pit->m_Node->m_Type == dmGameObject::SCENE_NODE_TYPE_COMPONENT)
         {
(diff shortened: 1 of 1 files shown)


engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -2906,21 +2910,34 @@ namespace dmGameSystem
 
         uint64_t index = pit->m_Next++;
 
-        const char* properties_common[] = { "type", "custom_type", "id", "pivot" };
+        const StrView properties_common[] = {
+            STR_VIEW_LITERAL("type"),
+            STR_VIEW_LITERAL("custom_type"),
+            STR_VIEW_LITERAL("id"),
+            STR_VIEW_LITERAL("pivot"),
+        };
```
JCash, line 2918: I don't think the code is improved by this change.
Also, I think the readability suffers a lot, for this code that is also only used when debugging.

JCash (changes requested):
Thanks, but I think we won't add this.

engine/gamesys/src/gamesys/components/comp_gui.cpp
redoak: Understood. I take the closing of this PR as a rejection of the proposal altogether. What's the process for the associated task #9743? Should I as the author close it, will you close it on your Monday meeting, or would you normally do it when closing the PR but simply forgot this time?

PR #9748, opened 2024-11-13
Remove platform specific handling of snprintf on Windows since it's not needed with Universal CRT

### LINKED ISSUE #9747: Remove platform specific handling of snprintf on Windows since it's not needed with Universal CRT
**Task (REQUIRED):**

Platform specific handling of `snprintf` was needed historically on Windows as they didn't comply with the C Standard:

```cpp
#if defined(_WIN32)
int result = _vsnprintf_s(buffer, size, _TRUNCATE, format, argp);
#else
int result = vsnprintf(buffer, size, format, argp);
#endif
```

With the Universal CRT in Visual Studio 2015 and Windows 10, these functions now comply with C99. As a result, these platform specifics should be removed.

Remove the Windows specifics accordingly.

redoak:
skip release notes

Resolves #9747

### Technical changes

* Replaced use of `_vsnprintf_s` with `vsnprintf` on `_WIN32`. With the Universal CRT in Visual Studio 2015 and Windows 10, the `snprintf` family of functions now comply with C99, and as a result, there's no need to have platform specific code. Since no platform besides Windows use the `_s` function variant, I'm assuming that a decision has been made to not use it, so neither should Windows now with this update.
* No changes to actual behavior.

### DIFF
--- a/engine/dlib/src/dlib/dstrings.cpp
+++ b/engine/dlib/src/dlib/dstrings.cpp
@@ -25,18 +25,14 @@
 
 int dmSnPrintf(char *buffer, size_t count, const char *format, ...)
 {
-    // MS-compliance
+    // mimics ms pre-ucrt vsnprintf_s behavior
     if (buffer == 0x0 || count == 0 || format == 0x0)
         return -1;
     va_list argp;
     va_start(argp, format);
-#if defined(_WIN32)
-    int result = _vsnprintf_s(buffer, count, _TRUNCATE, format, argp);
-#else
     int result = vsnprintf(buffer, count, format, argp);
-#endif
     va_end(argp);
-    // MS-compliance
+    // mimics ms pre-ucrt vsnprintf_s behavior
     if (count == 0 || (count > 0 && result >= (int)count))
         return -1;
     return result;
--- a/engine/dlib/src/dlib/dstrings.cpp
+++ b/engine/dlib/src/dlib/dstrings.cpp
@@ -33,7 +33,7 @@ int dmSnPrintf(char *buffer, size_t count, const char *format, ...)
     int result = vsnprintf(buffer, count, format, argp);
     va_end(argp);
     // mimics ms pre-ucrt vsnprintf_s behavior
-    if (count == 0 || (count > 0 && result >= (int)count))
+    if (result >= (int)count)
         return -1;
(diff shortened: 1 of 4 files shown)


engine/dlib/src/dlib/dstrings.cpp
```diff
@@ -25,19 +25,15 @@
...
-#if defined(_WIN32)
-    int result = _vsnprintf_s(buffer, count, _TRUNCATE, format, argp);
-#else
     int result = vsnprintf(buffer, count, format, argp);
-#endif
     va_end(argp);
-    // MS-compliance
-    if (count == 0 || (count > 0 && result >= (int)count))
+    // mimics ms pre-ucrt vsnprintf_s behavior
+    if (result >= (int)count)
```
JCash, line 36: This is a `DMSDK`function you are changing, which affects plenty of extensions our there.
Since you did the chagne for a reason, I expect to see the tests for this?

And, you didn't comment _why_ the `count == 0` isn't good anymore?

engine/dlib/src/dlib/pprint.cpp
```diff
@@ -48,11 +48,7 @@ namespace dmPPrint
         }
 
         int c = m_BufferSize - m_Cursor;
-    #if defined(_WIN32)
-        _vsnprintf_s(m_Buffer + m_Cursor, c, _TRUNCATE, format, argp);
-    #else
```
JCash, line 53: Uncommented change?

JCash (changes requested):
Uncommented changes to DMSDK.
Missing tests.

engine/dlib/src/dlib/pprint.cpp
redoak: I've updated the PR description. I originally wrote the reasoning behind the change and linked to references in the task, as I was thinking the reviewer would first read it and then the PR. But I see how it's better to have the info here in the PR.

I've rebased to latest dev and added a change: replaced an instance of `_vsnprintf_s` in script_table.cpp that I missed.

JCash: I see no links in your description though.
Where is the support of the `vsnprintf` mentioned?

JCash (approved):

PR #9750, opened 2024-11-13
Add handling of -1 return value from dmSnPrintf when the result was truncated

redoak:
skip release notes

Resolves #9749

### Technical changes

* Added handling of -1 return value from `dmSnPrintf`, which is returned when the resulting string was truncated.
* Fixed a couple of erroneous return value and bound checks.

### DIFF
--- a/engine/dlib/src/dlib/http_client.cpp
+++ b/engine/dlib/src/dlib/http_client.cpp
@@ -522,7 +522,7 @@ namespace dmHttpClient
         // DEF-2889 most webservers have a header length limit of 8096 bytes
         char buf[8096];
         const int bufsize = sizeof(buf);
-        if(dmSnPrintf(buf, bufsize, "%s: %s\r\n", name, value) > bufsize) {
+        if(dmSnPrintf(buf, bufsize, "%s: %s\r\n", name, value) == -1) {
             dmLogWarning("Truncated HTTP request header %s since it was larger than %d", name, bufsize);
         }
 
--- a/engine/dlib/src/dlib/log.cpp
+++ b/engine/dlib/src/dlib/log.cpp
@@ -671,7 +671,8 @@ void LogInternal(LogSeverity severity, const char* domain, const char* format, .
 
     if (n < dmLog::MAX_STRING_SIZE)
     {
-        n += dmSnPrintf(str_buf + n, dmLog::MAX_STRING_SIZE - n, "\n");
+        dmSnPrintf(str_buf + n, dmLog::MAX_STRING_SIZE - n, "\n");
+        ++n; // Since dmSnPrintf returns -1 on truncation, don't add the return value to n, and instead increment n separately
     }
 
     if (n >= dmLog::MAX_STRING_SIZE)
--- a/engine/dlib/src/dmsdk/dlib/dstrings.h
+++ b/engine/dlib/src/dmsdk/dlib/dstrings.h
@@ -29,7 +29,8 @@
 
 /*# Size-bounded string formating.
  *
- * Size-bounded string formating. Resulting string is guaranteed to be 0-terminated.
+ * Size-bounded string formating. Resulting string is guaranteed to be 0-terminated. Unlike snprintf, which
+ * always returns the untruncated string length, this function returns -1 if the string was truncated.
  *
  * @name dmSnPrintf
  * @param buffer Buffer to write to
--- a/engine/gamesys/src/gamesys/gamesys_private.cpp
+++ b/engine/gamesys/src/gamesys/gamesys_private.cpp
@@ -46,13 +46,21 @@ namespace dmGameSystem
             const char* path_name_receiver = dmHashReverseSafe64(receiver->m_Path);
             const char* fragment_name_receiver = dmHashReverseSafe64(receiver->m_Fragment);
 
-            n+= dmSnPrintf(buf + n, sizeof(buf) - n, " Message '%s' sent from %s:%s#%s to %s:%s#%s.",
+            int length = dmSnPrintf(buf + n, sizeof(buf) - n, " Message '%s' sent from %s:%s#%s to %s:%s#%s.",
                             id_str,
                             socket_name_sender, path_name_sender, fragment_name_sender,
                             socket_name_receiver, path_name_receiver, fragment_name_receiver);
+            if (length == -1)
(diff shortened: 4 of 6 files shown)


engine/dlib/src/dlib/dstrings.cpp
```diff
@@ -193,7 +193,7 @@ size_t dmStrlCpy(char *dst, const char *src, size_t siz)
  * Appends src to string dst of size siz (unlike strncat, siz is the
  * full size of dst, not space left).  At most siz-1 characters
  * will be copied.  Always NUL terminates (unless siz == 0).
- * Returns strlen(src); if retval >= siz, truncation occurred.
+ * Returns strlen(dst) + strlen(src); if retval >= siz, truncation occurred.
```
JCash, line 196: WHy isn't this change in the other PR, that actually alters the bevhavior in dstrings.cpp?

engine/dlib/src/dlib/http_client.cpp
```diff
@@ -522,11 +522,13 @@ namespace dmHttpClient
         // DEF-2889 most webservers have a header length limit of 8096 bytes
         char buf[8096];
         const int bufsize = sizeof(buf);
-        if(dmSnPrintf(buf, bufsize, "%s: %s\r\n", name, value) > bufsize) {
+        int buf_length = snprintf(buf, bufsize, "%s: %s\r\n", name, value);
```
JCash, line 525: No.
We use dmSnPrintf() because it works on all platforms.
Please revert to dmSnPrintf()

JCash (changes requested):
Unrelated changes in the same PR.
Removed usage of dmSnPrintf() in favor of platform dependent snprintf.

engine/dlib/src/dlib/http_client.cpp
redoak: I've updated the changes to only handle the return value properly.

engine/dlib/src/dlib/log.cpp
```diff
@@ -665,7 +665,7 @@ void LogInternal(LogSeverity severity, const char* domain, const char* format, .
         n += vsnprintf(str_buf + n, dmLog::MAX_STRING_SIZE - n, format, lst);
     }
 
-    if (n < dmLog::MAX_STRING_SIZE)
+    if (n + 1 < dmLog::MAX_STRING_SIZE)
```
JCash, line 668: No comment here? What was the issue?
Did you catch this in the tests?

engine/gamesys/src/gamesys/gamesys_private.cpp
```diff
@@ -46,13 +46,21 @@ namespace dmGameSystem
...
             const char* fragment_name_receiver = dmHashReverseSafe64(receiver->m_Fragment);
 
-            n+= dmSnPrintf(buf + n, sizeof(buf) - n, " Message '%s' sent from %s:%s#%s to %s:%s#%s.",
+            int length = dmSnPrintf(buf + n, sizeof(buf) - n, " Message '%s' sent from %s:%s#%s to %s:%s#%s.",
                             id_str,
                             socket_name_sender, path_name_sender, fragment_name_sender,
                             socket_name_receiver, path_name_receiver, fragment_name_receiver);
+            if (length == -1)
+            {
+                n = sizeof(buf);
```
JCash, line 55: Is this correct?
The string length is `n - 1` here?

JCash (changes requested):

engine/dlib/src/dlib/log.cpp
redoak: The +1 is to guarantee that there's space in the string for the "\n" appended by `dmSnPrintf` below. If there isn't space and `dmSnPrintf` is called, it would return -1 to signal truncation, which would then be added to `n`, causing the last char to be overwritten if another string were to be printed to `str_buf + n`.

No test — I spotted this as I was looking over all uses of `dmSnPrintf` (since I was fixing one misuse of `dmSnPrintf` return value, I thought I might as well fix all instances of the same misuse).

For this, would you expect a comment in the PR or a comment in the code?

engine/gamesys/src/gamesys/gamesys_private.cpp
redoak: The length is indeed `n - 1` as a result of truncation.

However, in order for the underflow error below to fire, `n` can't be set to the truncated length, as then we won't know if it was truncated or if the string's size happens to be exactly that of the buffer.

```cpp
if (n > (int) sizeof(buf) - 1) {
    dmLogError("Buffer underflow when formatting message-error (LogMessageError)");
}
```

The original code expected `dmSnPrintf` to have returned the untruncated length, which would have made it all work. Here, I get the same result by manually setting `n` to a length greater than what the buffer can hold. Alternatively, I could instead add a `bool is_truncated`, set it here, and check it in the condition below; it would mean more changed lines though.

engine/dlib/src/dlib/log.cpp
JCash: Well, yes, since it's old untouched code, it's curious that it's suddenly in need of a +1.
And, it also tells us that we didn't have a test for it.

redoak: I've added two tests to verify:

1. Logging a message with the maximum allowed length doesn't truncate
2. Logging a message with the maximum allowed length + 1 does truncate

To make the second test pass, I had to change this code again. No more `n + 1 < dmLog::MAX_STRING_SIZE`, since if `n` is `MAX_STRING_SIZE - 1` we still need to increment `n`, even though no character was appended, in order for the truncation to trigger with the next if-statement.

I added a code comment to clarify reasoning.

engine/dlib/src/dmsdk/dlib/dstrings.h
```diff
@@ -29,7 +29,8 @@
 
 /*# Size-bounded string formating.
  *
- * Size-bounded string formating. Resulting string is guaranteed to be 0-terminated.
+ * Size-bounded string formating. Resulting string is guaranteed to be 0-terminated. Unlike snprintf, which
```
JCash, line 32: Just a note that (as seen in the implementation), consideration for cross platform, and the MSVC vsnprintf_s is why is implemented this way.

JCash (approved):

Issue #9626, opened 2024-10-22
Remove library resolving step when using "Build HTML" 


AGulev:
**Task (REQUIRED):**  
When we build HTML5 from the editor, it always resolves dependencies. However, this option should serve as a replacement for `Cmd+B` but for the HTML5 platform. `Cmd+B` never resolves dependencies, leaving it up to the developer to resolve them manually when changes are made. On large projects with many dependencies, resolving them every time slows down the process significantly.

**Expected Outcome (REQUIRED):**  
Ensure that Bob uses the same dependencies as the editor and remove the automatic dependency resolution from the editor when using the "Build HTML" option to speed up the build process. This change will streamline HTML5 builds, aligning the process with how `Cmd+B` works. 

This could be implemented after one of the following tasks:
https://github.com/defold/defold/issues/9549
or
https://github.com/defold/defold/issues/6557

PR #9766, opened 2024-11-15
Better "Build HTML5" options

AGulev:
Build HTML5 no longer fetches dependencies to speed up the build process.
If you need to clean the build folder and fetch dependencies, use Rebuild HTML5.

Both options now have shortcuts:
CleanShot 2024-11-15 at 16 00 46@2x

Fix https://github.com/defold/defold/issues/9626


editor/src/clj/editor/app_view.clj
```diff
@@ -1320,23 +1320,32 @@ If you do not specifically require different script states, consider changing th
...
+        render-build-error! (make-render-build-error main-scene tool-tab-pane build-errors-view)
+        render-reload-progress! (make-render-task-progress :resource-sync)
+        render-save-progress! (make-render-task-progress :save-all)
+        render-build-progress! (make-render-task-progress :build)
+        task-cancelled? (make-task-cancelled-query :build)
+        bob-args (bob/build-html5-bob-options project prefs)
+        out (start-new-log-pipe!)]
+    (build-errors-view/clear-build-errors build-errors-view)
+    (disk/async-bob-build! render-reload-progress! render-save-progress! render-build-progress! out task-cancelled?
+                           render-build-error! (vec (concat pre-bob-cmds bob/build-html5-bob-commands)) bob-args project changes-view
```
AGulev, line 1334: The only addition in this line:
> (vec (concat pre-bob-cmds bob/build-html5-bob-commands))

editor/src/clj/editor/app_view.clj
```diff
@@ -1320,23 +1320,32 @@ If you do not specifically require different script states, consider changing th
...
+                           (fn [successful?]
+                             (when successful?
+                               (ui/open-url (format "http://localhost:%d%s/index.html" (http-server/port web-server) bob/html5-url-prefix)))
+                             (.close out)))))
+
+(handler/defhandler :rebuild-html5 :global
```
AGulev, line 1340: new option that passes extra build commands

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -403,7 +403,7 @@
          build-path (workspace/build-html5-path ws)]
      (io/file build-path "__htmlLaunchDir"))))
 
-(def build-html5-bob-commands ["resolve" "build" "bundle"])
+(def build-html5-bob-commands ["build" "bundle"])
```
AGulev, line 406: No need to resolve dependencies in a regular step anymore

editor/src/clj/editor/app_view.clj
```diff
@@ -1320,23 +1320,32 @@ If you do not specifically require different script states, consider changing th
...
+  (enabled? [] (not (build-in-progress?)))
+  (run [project prefs web-server build-errors-view changes-view main-stage tool-tab-pane]
+       (build-html5! project prefs web-server build-errors-view changes-view main-stage tool-tab-pane
+                     ["clean" "clear" "resolve"])))
```
vlaaad, line 1344: This is the first time I see `clear` command...

editor/src/clj/editor/app_view.clj
vlaaad: Hmm, I think it would be clearer if instead of `pre-bob-commands` this function accepted `bob-commands`. Then we would have `bob/build-html5-bob-commands` and `bob/rebuild-html5-bob-commands`. I think it would be easier to follow.

vlaaad (changes requested):
LGTM, but I think we don't have a `clear` bob command.

editor/src/clj/editor/app_view.clj
AGulev: 😳 how it's end up in my head and then here? O_o
I'll replace these two with "distclean"

editor/src/clj/editor/app_view.clj
AGulev: yes, this is much better! Thanks! Fixed!

AGulev:
I’ve checked, and in both cases, we use the same path for dependencies, so there’s no need to resolve them every time if the user has already fetched them.

editor/src/clj/editor/app_view.clj
```diff
@@ -1320,23 +1320,33 @@ If you do not specifically require different script states, consider changing th
...
+    (build-errors-view/clear-build-errors build-errors-view)
+    (disk/async-bob-build! render-reload-progress! render-save-progress! render-build-progress! out task-cancelled?
+                           render-build-error! bob-commands bob-args project changes-view
+                           (fn [successful?]
+                             (when successful?
+                               (ui/open-url (format "http://localhost:%d%s/index.html" (http-server/port web-server) bob/html5-url-prefix)))
+                             (.close out)))))
+
+(handler/defhandler :rebuild-html5 :global
+  (enabled? [] (not (build-in-progress?)))
```
vlaaad, line 1341: Hmm, did you copy this from `:rebuild` command? The `build-in-progess?` check is unnecessary here because is is used for the editor build pipeline, while this html5 build is made using bob pipeline.

vlaaad (changes requested):
Please remove the unnecessary `enabled?` check, then I think it's fine to merge 👍

editor/src/clj/editor/app_view.clj
AGulev: Yes, I did.
ok, I'll remove it

vlaaad (approved):
Ship it on this Friday evening 👍

PR #9777, opened 2024-11-17
Avoid strlen when constructing a hash of an game object instance id

redoak:
skip release notes

This is a separate PR for the one worthwhile change in #9737.

### Technical changes

* No changes to actual behavior — simply makes use of string length in `dmGameObject::ConstructInstanceId()`.

### DIFF
--- a/engine/gameobject/src/gameobject/gameobject.cpp
+++ b/engine/gameobject/src/gameobject/gameobject.cpp
@@ -969,8 +969,8 @@ namespace dmGameObject
     dmhash_t ConstructInstanceId(uint32_t index)
     {
         char buffer[16] = { 0 };
-        dmSnPrintf(buffer, sizeof(buffer), "%sinstance%d", ID_SEPARATOR, index);
-        return dmHashString64(buffer);
+        int length = dmSnPrintf(buffer, sizeof(buffer), "%sinstance%d", ID_SEPARATOR, index);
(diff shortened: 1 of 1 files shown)


engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -969,8 +969,8 @@ namespace dmGameObject
     dmhash_t ConstructInstanceId(uint32_t index)
     {
         char buffer[16] = { 0 };
-        dmSnPrintf(buffer, sizeof(buffer), "%sinstance%d", ID_SEPARATOR, index);
-        return dmHashString64(buffer);
+        int length = dmSnPrintf(buffer, sizeof(buffer), "%sinstance%d", ID_SEPARATOR, index);
+        return dmHashBuffer64(buffer, length);
```
JCash, line 973: No compiler warnings here?

JCash (changes requested):
Passing an int to a uint32_t argument should probably have produced a warning on your end?

engine/gameobject/src/gameobject/gameobject.cpp
redoak: I couldn't find any warning. But that might be a Windows thing. In any case, I added a cast to `uint32_t`.

JCash (approved):

PR #9782, opened 2024-11-18
Moved Lua obfuscator cretion to create() in LuaBuilder

britzl:
Create Lua obfuscators in create() to avoid concurrency problems when building.


vlaaad (approved):
I haven't tested it, but the change looks fine 👍

PR #9792, opened 2024-11-19
Use actual dpi value in glfw get display scale factor

Jhonnyg:
Fixes an issue where the `glfwGetDisplayScaleFactor` function always returns 1 on web. The function will now take `window.devicePixelRatio` into account, since this is what the glfw implementation is using as well.


AGulev (approved):
LGTM!

PR #9794, opened 2024-11-19
Fixed warning "not all control paths return a value" in socket.cpp

redoak:
skip release notes

On Windows, the following warnings are generated:

```
socket_posix.cpp(176) : warning C4715: 'dmSocket::TypeToNative': not all control paths return a value
socket_posix.cpp(184) : warning C4715: 'dmSocket::ProtocolToNative': not all control paths return a value
socket_posix.cpp(194) : warning C4715: 'dmSocket::DomainToNative': not all control paths return a value
socket_posix.cpp(436) : warning C4715: 'dmSocket::ShutdownTypeToNative': not all control paths return a value
```

No associated issue.

### Technical changes

* Add a return statement at the end of these switch-statement functions to resolve the warning.

### DIFF
--- a/engine/dlib/src/dlib/socket_posix.cpp
+++ b/engine/dlib/src/dlib/socket_posix.cpp
@@ -173,6 +173,7 @@ namespace dmSocket
             case TYPE_STREAM:  return SOCK_STREAM;
             case TYPE_DGRAM:   return SOCK_DGRAM;
         }
+        return 0;
     }
     static int ProtocolToNative(Protocol protocol)
     {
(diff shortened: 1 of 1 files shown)


engine/dlib/src/dlib/socket_posix.cpp
```diff
@@ -433,6 +436,7 @@ namespace dmSocket
             case SHUTDOWNTYPE_READWRITE:    return SHUT_RDWR;
 #endif
         }
+        return 0;
```
JCash, line 439: While good, I'd expected a build system flag?

JCash (approved):

PR #9799, opened 2024-11-19
Removed Windows-specific direct includes of malloc.h to get access to alloca outside of dalloca.h

redoak:
skip release notes

#9798 describes and fixes the following warning that is generated on Windows:

```
dalloca.h:33:13: warning: 'alloca' macro redefined [-Wmacro-redefined]
```

However, in several other locations malloc.h is directly included and `alloca` potentially redefined. To avoid this, these includes are removed and replaced with dalloca.h which centralizes the include and `alloca` definition.

No associated issue.

### Technical changes

* Replaced Windows-specific includes of malloc.h with includes of dalloca.h

### DIFF
--- a/engine/liveupdate/src/liveupdate.cpp
+++ b/engine/liveupdate/src/liveupdate.cpp
@@ -41,11 +41,6 @@
 #include <resource/resource_util.h>     // BytesToHexString for debug printing
 #include <resource/providers/provider.h>
 
-#if defined(_WIN32)
-#include <malloc.h>
-#define alloca(_SIZE) _alloca(_SIZE)
-#endif
-
 namespace dmLiveUpdate
 {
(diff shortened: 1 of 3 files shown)


engine/liveupdate/src/liveupdate.cpp
```diff
@@ -41,11 +41,6 @@
 #include <resource/resource_util.h>     // BytesToHexString for debug printing
 #include <resource/providers/provider.h>
 
-#if defined(_WIN32)
-#include <malloc.h>
-#define alloca(_SIZE) _alloca(_SIZE)
-#endif
-
```
redoak, line 48: There's no use of `alloca` in this file, so no need to include dalloca.h

JCash (approved):

Issue #6739, opened 2022-06-22
Add the geometry bounds visibility filter


astrochili:
Sometimes it's hard to figure out and navigate when the whole scene is filled with gray boxes.

![](https://user-images.githubusercontent.com/4752473/174963048-8ff7ff5d-1013-4bd9-99d9-a73b6779ff63.png)

TheKing0x9:
`View->Toggle Component Guides` or `Ctrl+H` should disable the grey boxes.

astrochili:
@TheKing0x9 It really works, I didn't know about this option! Thanks!

Perhaps it would be convenient to move it to the same list as the rest of the filters 🤔

AGulev:
related https://github.com/defold/defold/issues/6691

Issue #6691, opened 2022-06-09
Allow turning off the visibility of all debug lines in the scene


vlaaad:
**Is your feature request related to a problem? Please describe (REQUIRED):**
When developing a GUI scene, it's useful to see how it will look in the game without launching the game. Currently, it's not possible — we can't turn off grey node bounds.

**Describe the solution you'd like (REQUIRED):**
Allow turning off the visibility of node bounds.

**Describe alternatives you've considered (REQUIRED):**
I haven't considered any alternatives

matgis:
Doesn't **View > Toggle Component Guides** do just that? Maybe I'm not fully understanding what you want.

vlaaad:
Hmm. @Insality does **View > Toggle Component Guides** work for you?

matgis:
Let's add these options (Grid, Component Guides) to the Scene View visibility popup!

PR #9810, opened 2024-11-21
Add new toggles "Grid" and "Component Guides" into the Scene View visibility popup

AGulev:
"Grid" and "Component Guides" added into the Scene View visibility popup for better discoverability

Fix https://github.com/defold/defold/issues/6691
Fix https://github.com/defold/defold/issues/6739

### DIFF
--- a/editor/src/clj/editor/scene_visibility.clj
+++ b/editor/src/clj/editor/scene_visibility.clj
@@ -48,8 +48,8 @@
            {:label "Text" :tag :text}
            {:label "Tile Maps" :tag :tilemap}
            {:label :separator}
-           {:label "Component Guides" :tag :outline :command :toggle-component-guides :always-active true}
-           {:label "Grid" :tag :grid :always-active true}]
+           {:label "Component Guides" :tag :outline :command :toggle-component-guides :always-enabled true}
+           {:label "Grid" :tag :grid :always-enabled true}]
 
           (system/defold-dev?)
           (into [{:label :separator}
@@ -315,9 +315,12 @@
   (let [keymap (g/node-value app-view :keymap)
         command->shortcut (keymap/command->shortcut keymap)
         command->display-text (fn [command]
-                                (keymap/key-combo->display-text (get command->shortcut command)))
+                                (let [shortcut (get command->shortcut command)]
+                                  (if shortcut
+                                    (keymap/key-combo->display-text shortcut)
+                                    "")))
         make-control
-        (fn [{:keys [label tag command always-active]}]
+        (fn [{:keys [label tag command always-enabled]}]
           (if (= :separator label)
             [(Separator.) nil]
             (let [[control update-fn]
@@ -328,7 +331,7 @@
                   update-from-hidden-tags
                   (fn [hidden-tags enabled]
                     (let [checked (not (contains? hidden-tags tag))]
-                      (update-fn checked (or always-active enabled))))]
+                      (update-fn checked (or always-enabled enabled))))]
               [control update-from-hidden-tags])))
 
         tag-toggles (mapv make-control renderable-tag-toggles-info)
--- a/editor/src/clj/editor/scene_visibility.clj
+++ b/editor/src/clj/editor/scene_visibility.clj
@@ -33,10 +33,8 @@
 
 (def ^:private renderable-tag-toggles-info
   (cond-> [{:label "Collision Shapes" :tag :collision-shape}
-           {:label "Component Guides" :tag :outline :command :toggle-component-guides}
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/scene_visibility.clj
```diff
@@ -46,7 +46,10 @@
            {:label "Spine Scenes" :tag :spine}
            {:label "Sprites" :tag :sprite}
            {:label "Text" :tag :text}
-           {:label "Tile Maps" :tag :tilemap}]
+           {:label "Tile Maps" :tag :tilemap}
+           {:label :separator}
+           {:label "Component Guides" :tag :outline :command :toggle-component-guides :always-active true}
+           {:label "Grid" :tag :grid :always-active true}]
```
AGulev, line 52: We keep them always active, so user can on/off them independently of the rest of options.

editor/src/clj/editor/scene_visibility.clj
```diff
@@ -61,9 +64,7 @@
         renderable-tag-toggles-info))
 
 (defn filters-appear-active?
-  "Returns true if some parts of the scene are hidden due to visibility filters.
-  Does not consider scene elements that you'd not typically expect to be there,
-  such as debug rendering of bounding volumes, etc."
+  "Returns true if some parts of the scene are hidden due to visibility filters."
```
AGulev, line 67: it's not true anymore

editor/src/clj/editor/scene_visibility.clj
```diff
@@ -336,7 +341,7 @@
 
         [filters-enabled-control filters-enabled-update-fn]
         (make-toggle {:label "Visibility Filters"
-                      :acc (keymap/key-combo->display-text "Shift+Shortcut+I")
```
AGulev, line 339: I think it's not good that we have hardcoded shortcuts here. I changed it to make sure that we use the actual keymap and command name for getting a shortcut from the keymap

editor/src/clj/editor/scene_visibility.clj
```diff
@@ -310,20 +311,24 @@
...
                   (make-toggle {:label label
-                                :acc ""
+                                :acc (if command (command->display-text command) "")
                                 :on-change (fn [checked]
                                              (set-tag-visibility! scene-visibility tag checked))})
                   update-from-hidden-tags
                   (fn [hidden-tags enabled]
                     (let [checked (not (contains? hidden-tags tag))]
-                      (update-fn checked enabled)))]
+                      (update-fn checked (or always-active enabled))))]
```
AGulev, line 331: an option to keep new toggles always active

editor/src/clj/editor/scene_visibility.clj
```diff
@@ -46,7 +46,10 @@
            {:label "Spine Scenes" :tag :spine}
            {:label "Sprites" :tag :sprite}
            {:label "Text" :tag :text}
-           {:label "Tile Maps" :tag :tilemap}]
+           {:label "Tile Maps" :tag :tilemap}
+           {:label :separator}
+           {:label "Component Guides" :tag :outline :command :toggle-component-guides :always-active true}
```
AGulev, line 51: `command` added for cases when there is a shortcut for an option

editor/src/clj/editor/scene_visibility.clj
```diff
@@ -310,20 +311,24 @@
...
 
 (defn- make-visibility-toggles-list
-  ^Region [scene-visibility]
-  (let [make-control
-        (fn [{:keys [label tag]}]
+  ^Region [app-view scene-visibility]
+  (let [keymap (g/node-value app-view :keymap)
+        command->shortcut (keymap/command->shortcut keymap)
+        command->display-text (fn [command]
+                                (keymap/key-combo->display-text (get command->shortcut command)))
```
vlaaad, line 318: Bug: `(get command->shortcut command)` may return `nil` if a custom keymap reassigned the shortcut to another command. This will cause `keymap/key-combo->display-text` to throw an exception.

vlaaad (changes requested):
LGTM! I found a couple of issues, they should be easy to address.

editor/src/clj/editor/scene_visibility.clj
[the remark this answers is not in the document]
AGulev: done!

editor/src/clj/editor/scene_visibility.clj
AGulev: thank you, fixed!

vlaaad (approved):
Ship it!

PR #9830, opened 2024-11-25
Shaderc library based on spirv cross

Jhonnyg:
The spirv-cross command-line binary has been replaced with a library version of the tool that does the same thing, which means that we don't need to write intermediate files when crosscompiling, and we can now reconfigure resource bindings however we want. This means that we can merge resources between fragment and vertex program, change location decorations for input/output bindings and much more!


com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ShaderProgramBuilderTest.java
```diff
@@ -323,30 +324,30 @@ public void testShaderReflection() throws Exception {
...
-                    "   color_out += texture(sampler_2d_array, vec3(0.0));\n" +
-                    "   color_out += texture(sampler_cube, vec3(0.0));\n" +
-                    // TODO:
-                    //"   color_out += texelFetch(sampler_buffer, 0);\n" +
-                    "   color_out += texture(sampler2D(texture_2d, sampler_name), vec2(0.0));\n" +
-                    "   color_out += texture(usampler2D(utexture_2d, sampler_name), vec2(0));\n" +
-                    "   color_out += imageLoad(uimage_2d, ivec2(0));\n" +
-                    "   color_out += imageLoad(image_2d, ivec2(0));\n" +
-                    "}\n";
+            "uniform sampler2D sampler_2d;\n" +
```
Jhonnyg, line 327: Indentation (it looked weird in code)

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/shader/ShaderCompilePipeline.java
```diff
@@ -50,20 +52,22 @@ public ShaderModule(String source, ShaderDesc.ShaderType type) {
...
-        if (this.tintExe == null) this.tintExe = Bob.getExe(Platform.getHostPlatform(), "tint");
-        if (this.glslangExe == null) this.glslangExe = Bob.getExe(Platform.getHostPlatform(), "glslang");
-        if (this.spirvOptExe == null) this.spirvOptExe = Bob.getExe(Platform.getHostPlatform(), "spirv-opt");
-        if (this.spirvCrossExe == null) this.spirvCrossExe = Bob.getExe(Platform.getHostPlatform(), "spirv-cross");
+        if (tintExe == null)
+            tintExe = Bob.getExe(Platform.getHostPlatform(), "tint");
+        if (glslangExe == null)
+            glslangExe = Bob.getExe(Platform.getHostPlatform(), "glslang");
+        if (spirvOptExe == null)
+            spirvOptExe = Bob.getExe(Platform.getHostPlatform(), "spirv-opt");
```
Jhonnyg, line 70: Readability

engine/shaderc/src/test/data/bindings.vert
```diff
@@ -0,0 +1,31 @@
...
+    mat4 mtx_world;
+    mat4 mtx_view;
+    mat4 mtx_projection;
+    mat4 mtx_normal;
+};
+
+uniform extra
+{
+    vec4 offset;
+} extra_variable;
```
Jhonnyg, line 18: Part of the reason we need this library, there is no way to get the instance variable name here (extra_variable) from the client tool! But now we can :)

engine/shaderc/src/test/test_shaderc.cpp
```diff
@@ -0,0 +1,418 @@
...
+
+    ASSERT_NE((void*) 0, tex_coord);
+    ASSERT_EQ(2, tex_coord->m_Location);
+
+    dmShaderc::SetResourceLocation(shader_ctx, compiler, dmHashString64("position"), 3);
+    dmShaderc::SetResourceLocation(shader_ctx, compiler, dmHashString64("normal"), 4);
+    dmShaderc::SetResourceLocation(shader_ctx, compiler, dmHashString64("tex_coord"), 5);
+
+    dmShaderc::SetResourceBinding(shader_ctx, compiler, dmHashString64("matrices"), 3);
+    dmShaderc::SetResourceBinding(shader_ctx, compiler, dmHashString64("extra"), 4);
```
Jhonnyg, line 140: This will be useful for coming tasks. Vertex and fragment shaders are compiled to SPIRv separately, which means that locations and bindings for the same resources can have different values, which we can post-fix with these functions. We might need to take an extra compilation step to fixup the .spv file from the .fp and .vp shaders before crosscompiling, but it will fix all the issues with bindings we have.

JCash (approved):
lgtm!

PR #9834, opened 2024-11-26
Use more precise method to calculate `dt`

### LINKED ISSUE #3450: Revisit dmTime::GetTime (DEF-2013)
On linux (CI), it seems to suffer from poor resolution (~100 hz). The API should probably be split in a small timing function as well as a calendar based time. One goal with this task should be to lower the TOL in test_profile.cpp:35. Links:
(LINK REMOVED)(LINK REMOVED)CI error when used in the profiler tests:

```
[ RUN      ] dmProfile.Profile
../src/test/test_profile.cpp:96: Failure
The difference between (100000 + 50000 + 40000 + 50000 + 40000 + 60000) / 1000000.0 and samples[0].m_Elapsed / ticks_per_sec is 0.01633399999999996, which exceeds 0.001, where
(100000 + 50000 + 40000 + 50000 + 40000 + 60000) / 1000000.0 evaluates to 0.34000000000000002,
samples[0].m_Elapsed / ticks_per_sec evaluates to 0.35633399999999998, and
0.001 evaluates to 0.001.
[  FAILED  ] dmProfile.Profile (673 ms)
[ RUN      ] dmProfile.Nested
WARNING:DLIB: Unable to start profile http-server (-1)
../src/test/test_profile.cpp:160: Failure
The difference between (50000 + 50000) / 1000000.0 and samples[0].m_Elapsed / ticks_per_sec is 0.0026419999999999916, which exceeds 0.001, where
(50000 + 50000) / 1000000.0 evaluates to 0.10000000000000001,
samples[0].m_Elapsed / ticks_per_sec evaluates to 0.102642, and
0.001 evaluates to 0.001.
[  FAILED  ] dmProfile.Nested (146 ms)
```

2018-05-15 09:58:18 (M.Westerdahl)
Same problem with Win32 and QueryPerformanceTimer. It should be used on one thread, for very small scopes.

### LINKED ISSUE #8948: Occasional object teleportation (game speed up) on Macbook Pro M1 Max
Bug description:

When playing some Defold sample projects objects are being teleported. It looks like occasionally 1 in-game second passes during 1 frame. Better to see attached videos to understand

1. Create new project in Defold from one of the projects below
1.1. From template "Platformer game"
1.2. From sample "Lightning VFX"
1.3. From sample "Pixel line platformer"
2. Run game using Project -> Build
3. Play game for some time and see bug

Expected behavior:
Playing game without occasional object teleportation (game speed-up)

Defold version:
  - Version 1.8.0
  - Channel: editor-alpha
  - Build time: 2024-05-1515:14:00.109381
  - Defold CPU arch: Apple Silicon

Platforms:
 - Platforms: MacOS
 - OS: MacOS Sonoma 14.4.1
 - Device: Macbook Pro M1 Max

Minimal repro case project:
https://github.com/defold/template-platformer/archive/main.zip
https://github.com/defold/sample-pixel-line-platformer/archive/refs/heads/main.zip
https://github.com/FlexYourBrain/sample_LightningVFX/archive/master.zip

Videos of bug:
https://github.com/defold/defold/assets/47689344/0d74ef64-6c11-4cb0-9dc9-456ac9b3d01b
https://github.com/defold/defold/assets/47689344/9cbb5b44-e217-418d-ac5d-4aa90209e3eb
https://github.com/defold/defold/assets/47689344/64bce953-d2fa-4b3b-904d-9341292af801

Additional Info:
On some sample projects there is no such problem

### LINKED ISSUE #9812: Engine uses incorrect value for `dt` (affects physics; update; timers; animation)

When running on Apple silicon (my tests on M1 Pro) the engine sends incorrect values for `dt`.
Depending on combinations of `update_frequency` and `swap_interval`:

1.
When `update_frequency` is `60` then `dt` is often incorrectly set to `0.5` - which causes physics objects, timers, animations, and anything using `dt` passed into `update`, to leap ahead in time of where it should actually be. This can look like a choppy or laggy frame-rate, when the real frame-rate is just fine.

2.
When `update_frequency` is `0` and `swap_interval` is `1` the value of `dt` alternates between a low value (3ms) and a higher value (13ms), giving approximately 120 updates per second when the display refresh rate is actually 60 (`swap_interval` not respected?). This looks ok most of the time because on average everything is updated at the right speed, but periodically the actual screen-update event drifts over the boundary of these alternating frame-rates, and for a short time (0.5s maybe) objects appear to move fast/slow/fast/slow, jittering around from their expected position based on current movement speed. This happens every few minutes.

Note: This is almost certainly caused by or related to #8948 - but I have reported it separately because I think it may be possible to understand and solve this issue of incorrect `dt` values, even if the other issue turns out to be a problem with GLFW that cannot be fixed by Defold team.

**To Reproduce (REQUIRED)**

Use the repro project attached and run it on Apple silicon.
It may also manifest on other hardware but I haven't tested it there.

The value of `dt` in all the game systems (physics, timers, etc.) is consistent with the actual frame-rate being achieved.

1.9.5

**Platforms (REQUIRED):**

 - Platforms: macOS
 - OS: Sonoma 14.7.1
 - Device: Macbook Pro M1 Pro chip (arm64)

[dt_tests.zip](https://github.com/user-attachments/files/17851793/dt_tests.zip)

I've written a small project (not minimal) that runs a sequence of tests and shows the results in the GUI (also prints to the debug log). Results from a few runs are shown below.

The program as-is runs 4 tests, with the combinations of `update_frequency` (Freq. in the table) and `swap_interval` (Swap in the table).

|Test|Freq.| Swap|        Duration|Frames|   Deviation min/max| Reported dt min/max|   Actual dt min/max|
|----|-----|-----|----------------|------|--------------------|--------------------|--------------------|
|   1|    0|    0|  15 ( 15.00039)| 47759|  -0.005276/0.005269|   0.000215/0.005752|   0.000216/0.005572|
|   2|    0|    1| 120 (120.00126)| 13925|  -0.002990/0.003138|   0.000670/0.020534|   0.000646/0.020579|
|   3|   60|    0| 300 (297.47633)| 17858|  -0.001310/0.499518|   0.016667/0.500000|   0.000482/0.017977|
|   4|   60|    1| 120 (116.45937)|  6993|  -0.000882/0.499210|   0.016667/0.500000|   0.000790/0.022428|

The test runs gathering data until a `timer.delay` triggers a message (to ensure test is stopped at the end of a frame). For example in test 3 above, the Duration field in the table shows the expected/desired duration of 300s, but the actual duration (according to `socket.gettime()`) is only 297.4s - this is because `timer` is incorrectly leaping ahead in time. Looking at Reported dt max (0.5) it appears that on this run this event happened about 5 times in 300s. In test 4 it looks like it happened about 7 times in 120s.

|Test|Freq.| Swap|        Duration|Frames|   Deviation min/max| Reported dt min/max|   Actual dt min/max|
|----|-----|-----|----------------|------|--------------------|--------------------|--------------------|
|   1|    0|    0|  15 ( 15.00008)| 47610|  -0.004757/0.004751|   0.000211/0.005302|   0.000208/0.005048|
|   2|    0|    1| 120 (120.01098)| 14167|  -0.003471/0.003492|   0.001222/0.020334|   0.000622/0.020402|
|   3|   60|    0| 180 (176.96882)| 10627|  -0.000277/0.499305|   0.016667/0.500000|   0.000695/0.016943|
|   4|   60|    1| 180 (178.49587)| 10707|  -0.013460/0.499106|   0.016667/0.500000|   0.000894/0.030127|

Again in this run tests 3 and 4 show that the 0.5s time-leap happened 6 times and 3 times respectively. Note also that in all of these tables looking at test 2 shows the 120fps alternating short/long frame times described at the top.

|Test|Freq.| Swap|        Duration|Frames|   Deviation min/max| Reported dt min/max|   Actual dt min/max|eee|
|----|-----|-----|----------------|------|--------------------|--------------------|--------------------|---|
|   1|    0|    0|  15 ( 14.99996)| 48159|  -0.000349/0.000354|   0.000237/0.002664|   0.000235/0.002709|  0|
|   2|    0|    1|  60 ( 60.00170)|  6938|  -0.004556/0.004729|   0.001140/0.023031|   0.000482/0.025210|  0|
|   3|   60|    0|  60 ( 58.97552)|  3543|  -0.000780/0.499458|   0.016667/0.500000|   0.000542/0.017447|  2|
|   4|   60|    1|  60 ( 58.97144)|  3541|  -0.000376/0.496189|   0.016667/0.500000|   0.003811/0.020894|  2|

I added a counter (eee column, meaning estimated error events) which counts how many frames the reported `dt` is greater than `0.05` - since the actual dt max is always lower than this, and comparing to the actual duration value, it validates the count.

AGulev:
Fixes an issue where `dt` occasionally becomes 0.5s, causing jitters in the game.
Also, a new function has been introduced in `dmsdk`: `dmTime::GetMonotonicTime()`. This is the recommended way to measure time intervals (i.e., the time elapsed between events).

Fix https://github.com/defold/defold/issues/3450
Fix https://github.com/defold/defold/issues/8948
Fix https://github.com/defold/defold/issues/9812

### DIFF
--- a/engine/dlib/src/dlib/connection_pool.cpp
+++ b/engine/dlib/src/dlib/connection_pool.cpp
@@ -304,7 +304,7 @@ namespace dmConnectionPool
     static Result Connect(HPool pool, const char* host, dmSocket::Address address, uint16_t port, bool ssl, int timeout,
                                     dmSocket::Socket* socket, dmSSLSocket::Socket* sslsocket, dmSocket::Result* sr)
     {
-        uint64_t connectstart = dmTime::GetTime();
+        uint64_t connectstart = dmTime::GetMonotonicTime();
 
         Result r = ConnectSocket(pool, address, port, timeout, socket, sr);
         if( r != RESULT_OK )
@@ -313,7 +313,7 @@ namespace dmConnectionPool
             return r;
         }
 
-        uint64_t handshakestart = dmTime::GetTime();
+        uint64_t handshakestart = dmTime::GetMonotonicTime();
         if( timeout > 0 && (handshakestart - connectstart) > (uint64_t)timeout )
         {
             dmSocket::Delete(*socket);
@@ -350,11 +350,11 @@ namespace dmConnectionPool
         // that the caller can try to connect first to ipv4 and then to ipv6 if ipv4 failed.
         dmSocket::Address address;
 
-        uint64_t dial_started = dmTime::GetTime();
+        uint64_t dial_started = dmTime::GetMonotonicTime();
         bool gethost_did_succeed = dmSocket::GetHostByNameT(host, &address, timeout, cancelflag, ipv4, ipv6) == dmSocket::RESULT_OK;
         if (timeout > 0)
         {
-            timeout = timeout - (int)(dmTime::GetTime() - dial_started);
+            timeout = timeout - (int)(dmTime::GetMonotonicTime() - dial_started);
             if (timeout <= 0)
             {
                 return RESULT_SOCKET_ERROR;
@@ -415,7 +415,7 @@ namespace dmConnectionPool
     Result Dial(HPool pool, const char* host, uint16_t port, bool ssl, int timeout, int* cancelflag, HConnection* connection, dmSocket::Result* sock_res)
     {
         // try connecting to the host using ipv4 first
-        uint64_t dial_started = dmTime::GetTime();
+        uint64_t dial_started = dmTime::GetMonotonicTime();
         Result r = DoDial(pool, host, port, ssl, timeout, cancelflag, connection, sock_res, 1, 0);
         // Only if handshake failed NOT because of timeout
         if (r == RESULT_OK || r == RESULT_SHUT_DOWN || r == RESULT_OUT_OF_RESOURCES ||
@@ -426,7 +426,7 @@ namespace dmConnectionPool
         // ipv4 connection failed - reduce timeout (if needed) and try using ipv6 instead
         if (timeout > 0)
         {
-            timeout = timeout - (int)(dmTime::GetTime() - dial_started);
+            timeout = timeout - (int)(dmTime::GetMonotonicTime() - dial_started);
             if (timeout <= 0)
             {
                 return RESULT_SOCKET_ERROR;
--- a/engine/dlib/src/dlib/connection_pool.cpp
+++ b/engine/dlib/src/dlib/connection_pool.cpp
@@ -227,7 +227,7 @@ namespace dmConnectionPool
     static void PurgeExpired(HPool pool) {
         uint32_t n = pool->m_Connections.Size();
 
-        uint64_t now = dmTime::GetTime();
+        uint64_t now = dmTime::GetMonotonicTime();
         for (uint32_t i = 0; i < n; ++i) {
             Connection* c = &pool->m_Connections[i];
             if (c->m_State == STATE_CONNECTED && now >= c->m_Expires) {
@@ -399,7 +399,7 @@ namespace dmConnectionPool
                 c->m_ID = conn_id;
                 c->m_ReuseCount = 0;
                 c->m_State = STATE_INUSE;
-                c->m_Expires = pool->m_MaxKeepAlive * 1000000U + dmTime::GetTime();
+                c->m_Expires = pool->m_MaxKeepAlive * 1000000U + dmTime::GetMonotonicTime();
                 c->m_Address = address;
                 c->m_Port = port;
                 c->m_WasShutdown = 0;
--- a/engine/dlib/src/dlib/darwin/time_darwin.cpp
+++ b/engine/dlib/src/dlib/darwin/time_darwin.cpp
@@ -31,7 +31,7 @@ namespace dmTime
         return ((uint64_t) tv.tv_sec) * 1000000U + tv.tv_usec;
     }
 
-    uint64_t GetMonotomicTime()
+    uint64_t GetMonotonicTime()
     {
         uint64_t nanoseconds = clock_gettime_nsec_np(CLOCK_UPTIME_RAW);
         return nanoseconds / 1000U;
--- a/engine/dlib/src/dlib/darwin/time_darwin.cpp
+++ b/engine/dlib/src/dlib/darwin/time_darwin.cpp
@@ -33,8 +33,7 @@ namespace dmTime
 
     uint64_t GetMonotomicTime()
     {
-        struct timespec ts;
-        clock_gettime(CLOCK_MONOTONIC, &ts);
-        return (uint64_t) ts.tv_sec * 1000000U + ts.tv_nsec / 1000U;
+        uint64_t nanoseconds = clock_gettime_nsec_np(CLOCK_UPTIME_RAW);
+        return nanoseconds / 1000U;
     }
 }
--- a/engine/dlib/src/dlib/time.h
+++ b/engine/dlib/src/dlib/time.h
@@ -21,8 +21,8 @@
 namespace dmTime
 {
     inline void BusyWait(uint32_t useconds) {
-        uint64_t end = dmTime::GetMonotomicTime() + (uint64_t)useconds;
-        while (dmTime::GetMonotomicTime() < end);
+        uint64_t end = dmTime::GetMonotonicTime() + (uint64_t)useconds;
+        while (dmTime::GetMonotonicTime() < end);
     }
 }
 
--- a/engine/dlib/src/dlib/time.h
+++ b/engine/dlib/src/dlib/time.h
@@ -20,14 +20,9 @@
 
 namespace dmTime
 {
-    /**
-     * Busy wait thread with high precision (~10 microseconds).
-     * NOTE This function currently has low precision, see DEF-2013
-     * @param useconds Time to wait in microseconds
-     */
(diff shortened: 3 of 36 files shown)


AGulev:
Tested on:

- [ ] ios 11.4
- [ ] ios 18.1.1
- [ ] android 12
- [ ] android 5.0.1
- [ ] macos Sonoma 14.7.1 (23H222)
- [ ] windows 11
- [ ] ubuntu

engine/dlib/src/dlib/connection_pool.cpp
```diff
@@ -227,7 +227,7 @@ namespace dmConnectionPool
     static void PurgeExpired(HPool pool) {
         uint32_t n = pool->m_Connections.Size();
 
-        uint64_t now = dmTime::GetTime();
+        uint64_t now = dmTime::GetMonotonicTime();
```
AGulev, line 230: In all places where we need to measure time intervals, I replaced `dmTime::GetTime()` with  `dmTime::GetMonotonicTime()` to avoid unexpected issues

engine/dlib/src/dlib/darwin/time_darwin.cpp
```diff
@@ -0,0 +1,39 @@
...
+    {
+        timeval tv;
+        gettimeofday(&tv, 0);
+        return ((uint64_t) tv.tv_sec) * 1000000U + tv.tv_usec;
+    }
+
+    uint64_t GetMonotonicTime()
+    {
+        uint64_t nanoseconds = clock_gettime_nsec_np(CLOCK_UPTIME_RAW);
+        return nanoseconds / 1000U;
```
AGulev, line 37: When I read[ info about the topic](https://wadetregaskis.com/swifts-native-clocks-are-very-inefficient/), I found that the recommended function is [mach_absolute_time](https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time) , but apple added this API into [their list of "dirty" APIs](https://developer.apple.com/documentation/bundleresources/describing-use-of-required-reason-api?language=objc)
 and instead recommend using `clock_gettime_nsec_np(CLOCK_UPTIME_RAW)`

engine/dlib/src/dlib/time.h
```diff
@@ -20,14 +20,9 @@
 
 namespace dmTime
 {
-    /**
-     * Busy wait thread with high precision (~10 microseconds).
-     * NOTE This function currently has low precision, see DEF-2013
-     * @param useconds Time to wait in microseconds
-     */
```
AGulev, line 27: Should work now

engine/dlib/src/dlib/time_posix.cpp
```diff
@@ -26,8 +26,15 @@ namespace dmTime
...
-        return ((uint64_t) tv.tv_sec) * 1000000U + tv.tv_usec;
+        struct timespec ts;
+        clock_gettime(CLOCK_REALTIME, &ts);
+        return (uint64_t)ts.tv_sec * 1000000U + ts.tv_nsec / 1000U;
+    }
+
+    uint64_t GetMonotonicTime()
+    {
+        struct timespec ts;
+        clock_gettime(CLOCK_MONOTONIC, &ts);
```
AGulev, line 37: Replaced with new function which does the same as old, but we don't need 2 includes

engine/dlib/src/dlib/win32/time_win32.cpp
```diff
@@ -44,4 +44,18 @@ namespace dmTime
...
+    {
+        if (frequency == 0) {
+            LARGE_INTEGER freq;
+            QueryPerformanceFrequency(&freq);
+            frequency = freq.QuadPart;
+        }
+        LARGE_INTEGER counter;
+        QueryPerformanceCounter(&counter);
+        return (uint64_t)counter.QuadPart * 1000000ULL / frequency;
+    }
```
AGulev, line 60: windows implementation

engine/dlib/src/test/test_profile.cpp
```diff
@@ -168,15 +166,6 @@ TEST(dmProfile, SmallTest)
...
 
-#if 0
-// This test is currently disabled until we can figure out a good way to test it.
-// It is currentlyu very tricky to do due to its threadeded mechanic.
-
-// TODO
-// 100 msec, which is in fact much higher than the expected time of the profiler
-// On OSX, the time is usually a few microseconds, but once in a while the time spikes to ~0.5 ms
-// On Linux CI, the time can be as high as 16 msec
-// The timings (dmTime::BusyWait) is based around dmTime::GetTime, this issue is a revisit to improve the expected granularity: DEF-2013
```
AGulev, line 179: It should work now: https://github.com/defold/defold/issues/3450 (DEF-2013)

engine/dlib/src/dlib/darwin/time_darwin.cpp
AGulev: But will remotery ever end up in app stroe? it is Debug tool, no?

britzl: That is true though... hmm, so maybe we could remove that entry.

JCash (approved):
Looks good. Make sure to create a ticket for the consoles as well.

engine/dlib/src/dlib/darwin/time_darwin.cpp
AGulev: hmm, maybe we should remove it, not sure

PR #9838, opened 2024-11-26
Use PushbackReader when listing iOS devices

### LINKED ISSUE #9837: Use PushbackReader when listing devices with `ios-deploy`
**Task (REQUIRED):**
The updated JSON library requires a pushback reader to read multiple JSON documents from the same source (which is the case when we list devices using `ios-deploy`). Currently, it fails with the following message:
```
Resolving ios-deploy location...
Resolved to '/opt/homebrew/bin/ios-deploy'
Listing devices...
JSON error (unexpected character): K
```

`ios-deploy` works correctly.

vlaaad:
Fixes #9837

### DIFF
--- a/editor/src/clj/editor/ios_deploy.clj
+++ b/editor/src/clj/editor/ios_deploy.clj
@@ -19,7 +19,8 @@
             [editor.os :as os]
             [editor.prefs :as prefs]
             [editor.process :as process])
-  (:import [clojure.lang IReduceInit]))
+  (:import [clojure.lang IReduceInit]
+           [java.io PushbackReader]))
 
 (defn get-ios-deploy-path
   "Find ios-deploy path or throw if it's not found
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/ios_deploy.clj
```diff
@@ -40,7 +41,7 @@
 (defn- json-stream [input & {:as json-opts}]
   (reify IReduceInit
     (reduce [_ f init]
-      (with-open [reader (io/reader input)]
+      (with-open [reader (PushbackReader. (io/reader input) 64)]
```
vlaaad, line 44: The pushback buffer size has to be at least 64 chars. The new `data.json` lib version has some performance optimizations for faster string reading. This is only needed for repeated JSON reads from a stream, which seems to be the only place where we do it in the editor.

matgis (approved):
Ship it!

PR #9847, opened 2024-11-27
Camera components should have precedence in the default render script

### LINKED ISSUE #9842: Camera component(s) should have precedence in the render script by default
**Task (REQUIRED):**
To use a camera component in Defold, you need to send a message to the renderer:
`msg.post("@render:", "use_camera_projection")`

This will set the projection and view matrices from the _last_ camera in the camera stack. It's a really hidden workflow (especially for 3D) and really confusing for beginners coming from other engines where a camera component would work immediately. I see no reason at all why we have this UX.

We currently have a few ways of working with "cameras":
* actual camera components
* set view/proj from messages
* set view/proj function from message
* set view/proj manually in render script

My suggestion to improve this situation is that if there is a camera component, it should have precedence over the other projection modes. To use the custom "fixed", "fixed_fit" and "stretch" modes, the users should disable the camera component or modify the render script. In the long(er) run, we should consider adding add these projection modes to the camera component itself. 

If I add a camera to the scene, it should be used without any further setup.

Jhonnyg:
The default render script has been updated to take camera components into account - if there are camera components in the scene, they will be used in favor of the various "projection modes" specified in the render script. For a camera component to be used in rendering, a need function has also been added to the camera API:

`camera.get_enabled(url)`

This function can be used to check wether or not a camera is enabled or not. If there exists multiple camera components in the scene, the last enabled camera will be used for rendering. This is usually the camera component created last, but enabling and disabling camera components can change this order in the engine as well.

Fixes #9842

### DIFF
--- a/engine/engine/content/builtins/render/default.render_script
+++ b/engine/engine/content/builtins/render/default.render_script
@@ -3,10 +3,10 @@
 -- Copyright 2009-2014 Ragnar Svensson, Christian Murray
 -- Licensed under the Defold License version 1.0 (the "License"); you may not use
 -- this file except in compliance with the License.
--- 
+--
 -- You may obtain a copy of the License, together with FAQs at
 -- https://www.defold.com/license
--- 
+--
 -- Unless required by applicable law or agreed to in writing, software distributed
 -- under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
 -- CONDITIONS OF ANY KIND, either express or implied. See the License for the
@@ -67,8 +67,10 @@ local function update_clear_color(state, color)
 end
 
 local function update_camera(camera, state)
-    camera.proj = camera.projection_fn(camera, state)
-    camera.frustum.frustum = camera.proj * camera.view
+    if camera.projection_fn then
+        camera.proj = camera.projection_fn(camera, state)
+        camera.options.frustum = camera.proj * camera.view
+    end
 end
 
 local function update_state(state)
@@ -111,7 +113,7 @@ end
 
 local function create_camera(state, name, is_main_camera)
     local camera = {}
-    camera.frustum = {}
+    camera.options = {}
     state.cameras[name] = camera
     if is_main_camera then
         state.main_camera = camera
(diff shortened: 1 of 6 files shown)


engine/engine/content/builtins/render/default.render_script
```diff
@@ -67,8 +67,10 @@ local function update_clear_color(state, color)
 end
 
 local function update_camera(camera, state)
-    camera.proj = camera.projection_fn(camera, state)
-    camera.frustum.frustum = camera.proj * camera.view
```
Jhonnyg, line 71: renamed the inner table from frustum to options, since that is how it's used later in the rendering (i.e as the second parameter to draw, which is a table of options). This was a bit confusing for users, so this should make it more clear.

engine/engine/content/builtins/render/default.render_script
```diff
@@ -135,6 +137,38 @@ local function create_state()
     return state
 end
 
+local function set_camera_world(state)
+    local camera_components = camera.get_cameras()
+
+    -- This will set the last enabled camera from the stack of camera components
+    if #camera_components > 0 then
+        for i = #camera_components, 1, -1 do
+            if camera.get_enabled(camera_components[i]) then
```
Jhonnyg, line 146: New function - we shouldn't render with a  disabled camera.

Otherwise we check each camera from the top of the stack and backwards until we find one, and if we don't we use the "other" method of setting view/projection matrices.

engine/render/src/render/render_script_camera.cpp
```diff
@@ -246,6 +260,7 @@ namespace dmRender
         // READ-ONLY
         {"get_projection",          RenderScriptCamera_GetProjection},
         {"get_view",                RenderScriptCamera_GetView},
+        {"get_enabled",             RenderScriptCamera_GetEnabled},
```
Jhonnyg, line 263: This is added because we don't have a way to check this from render scripts otherwise. We could consider adding a "set_enabled" as well, but it's not needed for now

engine/engine/content/builtins/render/default.render_script
```diff
@@ -212,6 +250,7 @@ end
 function on_message(self, message_id, message)
     local state = self.state
     local camera = state.main_camera
+
     if message_id == MSG_CLEAR_COLOR then
         update_clear_color(state, message.color)
     elseif message_id == MSG_WINDOW_RESIZED then
```
AGulev, line 256: so we can remove `MSG_SET_VIEW_PROJ` and `MSG_SET_CAMERA_PROJ` in `on_message` and in our docs as well.

AGulev (commented):
👍❤️

engine/engine/content/builtins/render/default.render_script
britzl: This will break camera libraries that do not have a camera component but send these messages pretending to be a camera.

AGulev: Ah, ok

Issue #6694, opened 2022-06-10
Allow defining custom templates for file types


vlaaad:
**Is your feature request related to a problem? Please describe (REQUIRED):**
It's common for guis in the project to share some base configuration: textures, fonts, and layers. We should make it easier to create new guis by pre-filling the default configuration.

**Describe the solution you'd like (REQUIRED):**
We can allow selecting files to be used as templates when creating a new file with a particular extension. This should probably be a per-project configuration.

**Describe alternatives you've considered (REQUIRED):**
Instead of creating new files, developers can copy and rename the template files. I think it's somewhat clunky and slow.

PR #9854, opened 2024-11-28
Added the ability to specify custom templates for the project.

AGulev:
It is now possible to specify custom templates for each project. To do so, create a new folder named templates in the project’s root directory, and add new files named `default.*` with the desired extensions, such as `/templates/default.gui` or `/templates/default.script`. Additionally, if the `{{NAME}}` token is used in these files, it will be replaced with the filename specified in the file creation window.

Fix https://github.com/defold/defold/issues/6694

### DIFF
--- a/editor/src/clj/editor/game_object_common.clj
+++ b/editor/src/clj/editor/game_object_common.clj
@@ -38,13 +38,11 @@
 
 (def component-transform-property-keys (set (keys scene/identity-transform-properties)))
 
-(defn- template-pb-map-raw [workspace resource-type]
+(defn template-pb-map [workspace resource-type]
   (let [template (workspace/template workspace resource-type)
         read-fn (:read-fn resource-type)]
     (with-open [reader (StringReader. template)]
       (read-fn reader))))
-
-(def template-pb-map (fn/memoize template-pb-map-raw))
 
 (defn strip-default-scale-from-component-desc [component-desc]
   ;; GameObject$ComponentDesc or GameObject$EmbeddedComponentDesc in map format.
--- a/editor/src/clj/editor/workspace.clj
+++ b/editor/src/clj/editor/workspace.clj
@@ -467,17 +467,20 @@ ordinary paths."
   (or (:template resource-type)
       (some->> resource-type :ext (str "templates/template."))))
 
+(def ^:private user-resource-path "/templates/template.")
+
 (defn- get-template-resource [workspace resource-type]
   (let [path (template-path resource-type)
         java-resource (when path (io/resource path))
+        user-resource (find-resource workspace (some->> resource-type :ext (str user-resource-path)))
         editor-resource (when path (find-resource workspace path))]
-    (or java-resource editor-resource)))
-  
+    (or user-resource java-resource editor-resource)))
+
 (defn has-template? [workspace resource-type]
   (let [resource (get-template-resource workspace resource-type)]
     (not= resource nil)))
 
-(defn- template-raw [workspace resource-type]
+(defn template [workspace resource-type]
   (when-let [resource (get-template-resource workspace resource-type)]
     (let [{:keys [read-fn write-fn]} resource-type]
       (if (and read-fn write-fn)
@@ -489,8 +492,6 @@ ordinary paths."
         ;; Just read the file as-is.
         (with-open [reader (io/reader resource)]
           (slurp reader))))))
-
-(def template (fn/memoize template-raw))
 
 (defn- update-dependency-notifications! [workspace lib-states]
   (let [{:keys [error missing]} (->> lib-states
--- a/editor/src/clj/editor/workspace.clj
+++ b/editor/src/clj/editor/workspace.clj
@@ -467,15 +467,16 @@ ordinary paths."
 (def ^:private java-resource-path "templates/template.")
 
 (defn- get-template-resource [workspace resource-type]
-  (let [resource-path (:template resource-type)
-        ext (:ext resource-type)]
-    (or
-      ;; default user resource
-      (when ext (find-resource workspace (str default-user-resource-path ext)))
-      ;; editor resource provided from extensions
-      (when resource-path (find-resource workspace resource-path))
-      ;; java resource
-      (when ext (io/resource (str java-resource-path ext))))))
+  (when resource-type
+    (let [resource-path (:template resource-type)
+          ext (:ext resource-type)]
+      (or
+        ;; default user resource
+        (find-resource workspace (str default-user-resource-path ext))
+        ;; editor resource provided from extensions
+        (when resource-path (find-resource workspace resource-path))
+        ;; java resource
+        (io/resource (str java-resource-path ext))))))
 
 (defn has-template? [workspace resource-type]
   (let [resource (get-template-resource workspace resource-type)]
--- a/editor/src/clj/editor/workspace.clj
+++ b/editor/src/clj/editor/workspace.clj
@@ -463,18 +463,18 @@ ordinary paths."
                                                                         path))))]
       (resolve-workspace-resource workspace path))))
 
-(defn- template-path [resource-type]
-  (or (:template resource-type)
-      (some->> resource-type :ext (str "templates/template."))))
-
-(def ^:private user-resource-path "/templates/template.")
+(def ^:private default-user-resource-path "/templates/default.")
+(def ^:private java-resource-path "templates/template.")
 
 (defn- get-template-resource [workspace resource-type]
-  (let [path (template-path resource-type)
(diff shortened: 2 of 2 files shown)


editor/src/clj/editor/game_object_common.clj
```diff
@@ -38,14 +38,12 @@
...
 (def component-transform-property-keys (set (keys scene/identity-transform-properties)))
 
-(defn- template-pb-map-raw [workspace resource-type]
+(defn template-pb-map [workspace resource-type]
   (let [template (workspace/template workspace resource-type)
         read-fn (:read-fn resource-type)]
     (with-open [reader (StringReader. template)]
       (read-fn reader))))
 
-(def template-pb-map (fn/memoize template-pb-map-raw))
```
AGulev, line 47: I removed `fn/memoize` as we agreed (because it's a lightweight operation rarely used)

editor/src/clj/editor/workspace.clj
```diff
@@ -490,8 +493,6 @@ ordinary paths."
         (with-open [reader (io/reader resource)]
           (slurp reader))))))
 
-(def template (fn/memoize template-raw))
```
AGulev, line 493: no more memorize here as well

editor/src/clj/editor/workspace.clj
```diff
@@ -467,17 +467,20 @@ ordinary paths."
   (or (:template resource-type)
       (some->> resource-type :ext (str "templates/template."))))
 
+(def ^:private user-resource-path "/templates/template.")
```
AGulev, line 470: It's not really the same path as path to a resource in the jar file because of `/` in the beginning. 
I think it's easier to use two separate variables to avoid mess here.

editor/src/clj/editor/workspace.clj
```diff
@@ -467,17 +467,20 @@ ordinary paths."
...
+
 (defn- get-template-resource [workspace resource-type]
   (let [path (template-path resource-type)
         java-resource (when path (io/resource path))
+        user-resource (find-resource workspace (some->> resource-type :ext (str user-resource-path)))
         editor-resource (when path (find-resource workspace path))]
-    (or java-resource editor-resource)))
-  
+    (or user-resource java-resource editor-resource)))
```
AGulev, line 477: This is actual change.  The order priority for different template PATHs:
if user path in the project exist - use this one, if not use one from java, if not check if we have a special path specified (for extensions)

vlaaad: I think it would be nice if we could avoid unnecessary lookups. For example, we do `editor-resource` lookup even if `user-resource` exists. I'm suggesting something like:
```clj
(or 
  ;; user resource
  (find-resource workspace (some->> resource-type :ext (str user-resource-path)))
  ;; java resource
  (when path (io/resource path))
  ;; editor resource
  (when path (find-resource workspace path)))
```

vlaaad (approved):
LGTM!

editor/src/clj/editor/workspace.clj
AGulev: it makes sense

editor/src/clj/editor/workspace.clj
AGulev: true

editor/src/clj/editor/workspace.clj
AGulev: Fixed!

editor/src/clj/editor/workspace.clj
AGulev: I go further and modified it a little bit more (tested with spine extension to make sure it forks)

editor/src/clj/editor/workspace.clj
```diff
@@ -463,18 +463,18 @@ ordinary paths."
...
-        editor-resource (when path (find-resource workspace path))]
-    (or user-resource java-resource editor-resource)))
+  (let [resource-path (:template resource-type)]
+    (or
+      ;; default user resource
+      (find-resource workspace (some->> resource-type :ext (str default-user-resource-path)))
+      ;; editor resource provided from extensions
+      (when resource-path (find-resource workspace resource-path))
+      ;; java resource
+      (io/resource (some->> resource-type :ext (str java-resource-path))))))
```
vlaaad, line 477: Hmm, this is getting strange. `(some->> resource-type :ext (str java-resource-path))` implies it may return `nil`. `(io/resource nil)` will throw an exception.

AGulev: IDK why it was like that? Because `:ext` can be nil? if so, I can change it to just check it

AGulev:
```
(defn- get-template-resource [workspace resource-type]
  (let [resource-path (:template resource-type)
        ext (:ext resource-type)]
    (or
      ;; default user resource
      (when ext (find-resource workspace (str default-user-resource-path ext)))
      ;; editor resource provided from extensions
      (when resource-path (find-resource workspace resource-path))
      ;; java resource
      (when ext (io/resource (str java-resource-path ext))))))
      ```

File: editor/src/clj/editor/workspace.clj [hunk 6]
```diff
@@ -463,21 +463,24 @@ ordinary paths."
...
-  
+  (let [resource-path (:template resource-type)]
+    (or
+      ;; default user resource
+      (find-resource workspace (some->> resource-type :ext (str default-user-resource-path)))
+      ;; editor resource provided from extensions
+      (when resource-path (find-resource workspace resource-path))
+      ;; java resource
+      (io/resource (some->> resource-type :ext (str java-resource-path))))))
+
```
matgis, line 478: The priority order changed here.

Previously the template from the editor jar would take precedence over any templates matching the built-in editor file types from extension, but now a `template.sprite` from an extension would win over the one in the editor jar.

Was this intentional?

I think it might be better this way, actually, because it allows an extension to tailor the defaults for newly-created resources to its requirements. For example, if an extension's render script has specific expectations of materials and so on. 🤔

matgis (approved):

editor/src/clj/editor/workspace.clj
```diff
@@ -467,14 +467,15 @@ ordinary paths."
...
+        ext (:ext resource-type)]
     (or
       ;; default user resource
-      (find-resource workspace (some->> resource-type :ext (str default-user-resource-path)))
+      (when ext (find-resource workspace (str default-user-resource-path ext)))
       ;; editor resource provided from extensions
       (when resource-path (find-resource workspace resource-path))
       ;; java resource
-      (io/resource (some->> resource-type :ext (str java-resource-path))))))
+      (when ext (io/resource (str java-resource-path ext))))))
```
vlaaad, line 478: `:ext` is definitely not nil. We should either remove nil checks here (i.e. replace `some->>` with `->>`) because they were added by mistake or figure out why there were nullability checks (I don't know if `resource-type` could be `nil` here).

vlaaad (changes requested):

editor/src/clj/editor/workspace.clj
vlaaad: I think I understand now. Yes, the resource type may be `nil` if we create a file using `New File` and type an unregistered extension, e.g. `new file -> foo.bleh`. See `handler/defhandler :new-file :global`'s `run` function that invokes `workspace/template`. It would be nice for the editor to still use a template `/templates/default.bleh` in this case, but that would require more refactoring — we would need to pass an extension, and the function would look up resource type by this extension to see if there is an explicit template; and if there isn't, then it would look up fallbacks.

AGulev: It feels like a small fraction out of a % of use cases. Do we want to solve it now as part of this issue?
Or maybe it's enough to make sure it's not nil:
```clj
(defn- get-template-resource [workspace resource-type]
  (when resource-type
    (let [resource-path (:template resource-type)
          ext (:ext resource-type)]
      (or
        ;; default user resource
        (find-resource workspace (str default-user-resource-path ext))
        ;; editor resource provided from extensions
        (when resource-path (find-resource workspace resource-path))
        ;; java resource
        (io/resource (str java-resource-path ext))))))

vlaaad: I think so, it's a fairly small change to accept an extension instead of a resource type.

AGulev: it can't be extension
it should be a full path
because native extensions provide full path
maybe I'll have time to take a look into it next week

vlaaad: when you have a file extension and a workspace, you can look up the resource type using `get-resource-type`

vlaaad (approved):
LGTM!

PR #9871, opened 2024-12-02
Consider successors from implicit connections inherited by override nodes

matgis:
Fixed an issue where stale values would be retained in the editor system cache in certain scenarios involving overrides.

### Technical changes
* We now correctly consider successors from implicit connections inherited by override nodes when altering the topology of the editor graph.

### DIFF
--- a/editor/src/clj/internal/transaction.clj
+++ b/editor/src/clj/internal/transaction.clj
@@ -21,6 +21,7 @@
             [internal.system :as is]
             [internal.util :as util]
             [schema.core :as s]
+            [util.array :as array]
             [util.coll :refer [pair]]
             [util.debug-util :as du]
             [util.eduction :as e])
@@ -744,9 +745,10 @@
             ;; nodes of the source node, since these will inherit an implicit
             ;; connection between them and the corresponding override nodes of
             ;; the target node.
-            (flag-successors-changed (cons (pair source-id source-label)
-                                           (e/map #(pair % source-label)
-                                                  (ig/get-overrides basis source-id))))
+            (flag-successors-changed (e/concat
+                                       (array/of (pair source-id source-label))
+                                       (e/map #(pair % source-label)
+                                              (ig/get-overrides basis source-id))))
             (flag-override-nodes-affected target target-label)))
       ctx)
     ctx))
@@ -785,9 +787,10 @@
       ;; When updating the successors, we must also consider any override nodes
       ;; of the source node, since these will inherit an implicit connection
       ;; between them and the corresponding override nodes of the target node.
-      (flag-successors-changed (cons (pair source-id source-label)
-                                     (e/map #(pair % source-label)
-                                            (ig/get-overrides (:basis ctx) source-id))))
+      (flag-successors-changed (e/concat
+                                 (array/of (pair source-id source-label))
+                                 (e/map #(pair % source-label)
+                                        (ig/get-overrides (:basis ctx) source-id))))
       (ctx-remove-overrides source-id source-label target-id target-label)))
 
 (defmethod perform :disconnect
--- a/editor/src/clj/internal/transaction.clj
+++ b/editor/src/clj/internal/transaction.clj
@@ -22,7 +22,8 @@
             [internal.util :as util]
             [schema.core :as s]
             [util.coll :refer [pair]]
-            [util.debug-util :as du])
+            [util.debug-util :as du]
+            [util.eduction :as e])
   (:import [internal.graph.types Arc]))
 
 (set! *warn-on-reflection* true)
@@ -341,8 +342,14 @@
         all-originals (ig/override-originals basis original-node-id)]
     (-> ctx
         (assoc :basis (gt/override-node basis original-node-id override-node-id))
+
+        ;; Any property, input or output on any original nodes must now take the
+        ;; new override node into account.
         (flag-all-successors-changed all-originals)
-        (flag-successors-changed (mapcat #(gt/sources basis %) all-originals)))))
+
+        ;; Similarly, so must the source outputs of any arcs that target any of
+        ;; the original nodes.
+        (flag-successors-changed (e/mapcat #(gt/sources basis %) all-originals)))))
 
 (defmethod perform :override-node
   [ctx {:keys [original-node-id override-node-id]}]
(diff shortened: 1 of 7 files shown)


editor/src/clj/internal/transaction.clj
```diff
@@ -723,17 +730,23 @@
...
             (mark-input-activated target-id target-label)
             (update :basis gt/connect source-id source-label target-id target-label)
-            (flag-successors-changed [[source-id source-label]])
+            ;; When updating the successors, we must also consider any override
+            ;; nodes of the source node, since these will inherit an implicit
+            ;; connection between them and the corresponding override nodes of
+            ;; the target node.
+            (flag-successors-changed (cons (pair source-id source-label)
+                                           (e/map #(pair % source-label)
+                                                  (ig/get-overrides basis source-id))))
```
matgis, line 749: Note: We use the "before" `basis` here since we want to consider the state before the change. Besides, override nodes are not created or destroyed immediately here, we just flag them for consideration at the end of the transaction.

editor/test/internal/node_test.clj
```diff
@@ -175,7 +170,7 @@
                    (fn [node-id] (g/set-property    node-id :foo "two")))
   (expect-modified SimpleTestNode
                    #{:_declared-properties :_properties :foo}
-                   (fn [node-id] (g/update-property node-id :foo str/reverse)))
+                   (fn [node-id] (g/update-property node-id :foo string/reverse)))
```
matgis, line 173: Unrelated pet peeve.

editor/test/internal/node_test.clj
```diff
@@ -611,30 +606,551 @@
...
+  ;; Dependent on array-input.
+  (output uncached-output<=array-input g/Any (g/fnk [array-input] array-input))
+  (output cached-output<=array-input g/Any :cached (g/fnk [array-input] array-input))
+  (output uncached-secondary-output<=uncached-output<=array-input g/Any (g/fnk [uncached-output<=array-input] uncached-output<=array-input))
+  (output cached-secondary-output<=uncached-output<=array-input g/Any :cached (g/fnk [uncached-output<=array-input] uncached-output<=array-input))
+  (output uncached-secondary-output<=cached-output<=array-input g/Any (g/fnk [cached-output<=array-input] cached-output<=array-input))
+  (output cached-secondary-output<=cached-output<=array-input g/Any :cached (g/fnk [cached-output<=array-input] cached-output<=array-input)))
 
 (deftest dependency-caching
+  (letfn [(cacheable-property-dependent-endpoints [node-id]
```
matgis, line 638: Fleshed out this existing test to cover more editing scenarios. This was done during the discovery process. The new checks don't actually fail if I remove the fix in `transaction.clj`, but I like the added coverage.

editor/test/internal/node_test.clj
```diff
@@ -611,30 +606,551 @@
...
+  ([basis node-id label]
+   (into (sorted-set)
+         (map #(gt/endpoint % label))
+         (g/overrides basis node-id))))
+
+(deftest override-successors
+  ;; This test recreates a scenario discovered when working on the gui system.
+  ;; We discovered an issue where the successors stored in the graph would not
+  ;; correct after adding a gui node to a scene imported as a template from
+  ;; another scene.
```
matgis, line 999: This added test focuses on the successors in particular, and will fail if I remove the fix in `transaction.clj`.

editor/src/clj/internal/transaction.clj
vlaaad: It seems that using `cons` erases all the benefits of using eduction here. Should we use `concat` with a singleton pair here instead? Or perhaps we could implement `cons` as a transducer...

editor/src/clj/internal/transaction.clj
```diff
@@ -769,7 +782,12 @@
   (-> ctx
       (mark-input-activated target-id target-label)
       (update :basis gt/disconnect source-id source-label target-id target-label)
-      (flag-successors-changed [[source-id source-label]])
+      ;; When updating the successors, we must also consider any override nodes
+      ;; of the source node, since these will inherit an implicit connection
+      ;; between them and the corresponding override nodes of the target node.
+      (flag-successors-changed (cons (pair source-id source-label)
```
vlaaad, line 788: same issue with `cons` here.

editor/test/internal/node_test.clj
vlaaad: This is great!

vlaaad (commented):
LGTM!

I have a worry that `cons` forcing eductions into `seq` will become a performance problem for us. I'm not sure if that place is going to be a hot path though, and I'm not sure what would be a better solution there...

matgis:
Your hunch was right! Using `e/concat` with a one-element object array is around three times faster. 👍

editor/src/clj/util/eduction.clj
```diff
@@ -13,115 +13,130 @@
...
+  (:require [util.array :as array]))
 
 (set! *warn-on-reflection* true)
+(set! *unchecked-math* :warn-on-boxed)
 
 (defonce empty-eduction (eduction))
 
 (definline cat [coll]
-  `(eduction
+  `(->Eduction
```
matgis, line 25: `eduction` is a function and creates intermediate `seqs` from the arguments. We can avoid that by constructing the `Eduction` from arguments ourselves when we have exactly two arguments.

editor/src/clj/util/array.clj
```diff
@@ -0,0 +1,94 @@
...
+
+(ns util.array
+  (:import [java.lang.reflect Array]))
+
+(set! *warn-on-reflection* true)
+(set! *unchecked-math* :warn-on-boxed)
+
+(def ^"[Ljava.lang.Object;" empty-object-array (Array/newInstance Object 0))
+
+(defn of
```
matgis, line 23: Intended use: `(array/of "item one" :item-two ItemThree)`.

Perhaps we'll add primitive flavors like `of-longs`, `of-doubles` and so on in the future.

I decided to make this its own module so I could use it from `util.eduction` as well.

editor/test/util/array_test.clj
```diff
@@ -0,0 +1,33 @@
...
+
+(set! *warn-on-reflection* true)
+(set! *unchecked-math* :warn-on-boxed)
+
+(defn- object-arrays-equal? [^"[Ljava.lang.Object;" a ^"[Ljava.lang.Object;" b]
+  (Arrays/equals a b))
+
+(deftest of-test
+  (let [items [:a :b :c :d :e :f :g :h :i :j :k :l :m :n :o :p]]
+    (doseq [^long length (range 4)]
```
vlaaad, line 28: shouldn't it be `(range (count items))`? and maybe `dotimes` instead of `doseq`...

vlaaad (approved):
LGTM, though maybe we don't test `array/of` enough? Since we test the array creation only up to 3 elements, we don't test the vararg arity, which seems to be the trickiest part.

vlaaad:
Ship it (if the tests succeed 😁)!

PR #9876, opened 2024-12-03
Add `editor.browse` editor script function

### LINKED ISSUE #9875: Add `editor.open_url` editor scripts API
**Task (REQUIRED):**
Add `editor.open_url` editor scripts API.

Editor scripts should be able to open URLs.

vlaaad:
Fixes #9875

### DIFF
--- a/editor/src/clj/editor/editor_extensions.clj
+++ b/editor/src/clj/editor/editor_extensions.clj
@@ -359,8 +359,8 @@
       (when (and resource (resource/exists? resource) (resource/openable? resource))
         (open-resource! resource)))))
 
-(def ext-open-url-fn
-  (rt/suspendable-lua-fn open-url [{:keys [rt]} lua-string]
+(def ext-browse-fn
+  (rt/suspendable-lua-fn browse [{:keys [rt]} lua-string]
     (let [s (rt/->clj rt coerce/string lua-string)]
       (future/io
         (try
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/editor_extensions.clj
```diff
@@ -357,6 +359,15 @@
       (when (and resource (resource/exists? resource) (resource/openable? resource))
         (open-resource! resource)))))
 
+(def ext-open-url-fn
+  (rt/suspendable-lua-fn open-url [{:keys [rt]} lua-string]
+    (let [s (rt/->clj rt coerce/string lua-string)]
+      (future/io
+        (try
+          (.browse ui/desktop (URI. s))
```
vlaaad, line 367: I didn't use `ui/open-url` because I need to raise an error on failure to open. I checked this changeset on Linux and it does not block the editor (`java.awt.Desktop` and JavaFX...)

matgis (changes requested):
I feel `editor.open_url` might be too generic of a name. Defold uses its own concept of urls to reference objects inside scenes, and `open_url` feels like it could be used to follow such a url and open the referenced object in the editor. Maybe something like `browse_url` or `open_in_browser` would be better?

Also, should it be in the `editor.ui` module? We already have `editor.ui.open_resource` there. 🤔

vlaaad:
I think the function should not be in the `editor.ui` module because it does not open the URL in the editor UI, unlike `open_resource`, which opens a resource inside the editor. Also `open_in_browser` implies it only opens URLs in the browser, while file URLs may be opened in a text editor. I think `browse` fits better here.

matgis (approved):
Ship it!

PR #9880, opened 2024-12-03
Add optional SHA1 validation for wasm, wasmjs, and asmjs files

smagnuso:
Second step to adding SHA1 authentication. This is lumped in `--with-sha1` and will add hash verification for .wasm, _wasm.js _asm.js and the split file .json in the dmloader.


britzl (approved):
Looks good!

PR #9883, opened 2024-12-04
Add ASTC texture compressor

### LINKED ISSUE #9885: Add ASTC texture compression in the texture generator
**Task (REQUIRED):**
Similar to Basis, we should be able to produce ASTC data via our TexC library

Jhonnyg:
A new texture encoder has been added that can encode textures into ASTC binaries. While ASTC isn't supported on all platforms, if can still be useful if you are building for devices and platforms that are guaranteed to have it. This is also a step in the direction towards supporting more texture compressors as plugins in the engine (e.g Webp compression). 

Note that this is a technical preview of this feature, there is no editor support available yet!

Fixes #9885


editor/src/clj/editor/protobuf_forms.clj
```diff
@@ -230,12 +230,26 @@
...
+    :texture-format-rgba-astc-6x6
+    :texture-format-rgba-astc-8x5
+    :texture-format-rgba-astc-8x6
+    :texture-format-rgba-astc-8x8
+    :texture-format-rgba-astc-10x5
+    :texture-format-rgba-astc-10x6
+    :texture-format-rgba-astc-10x8
+    :texture-format-rgba-astc-10x10
+    :texture-format-rgba-astc-12x10
+    :texture-format-rgba-astc-12x12})
```
Jhonnyg, line 246: This means that the formats will NOT be shown in the editor, just so people don't start using them prematurely.

engine/graphics/src/dmsdk/graphics/graphics.h
```diff
@@ -264,6 +264,21 @@ namespace dmGraphics
...
+        TEXTURE_FORMAT_RGBA_ASTC_8x5        = 37,
+        TEXTURE_FORMAT_RGBA_ASTC_8x6        = 38,
+        TEXTURE_FORMAT_RGBA_ASTC_8x8        = 39,
+        TEXTURE_FORMAT_RGBA_ASTC_10x5       = 40,
+        TEXTURE_FORMAT_RGBA_ASTC_10x6       = 41,
+        TEXTURE_FORMAT_RGBA_ASTC_10x8       = 42,
+        TEXTURE_FORMAT_RGBA_ASTC_10x10      = 43,
+        TEXTURE_FORMAT_RGBA_ASTC_12x10      = 44,
+        TEXTURE_FORMAT_RGBA_ASTC_12x12      = 45,
+
```
Jhonnyg, line 281: UGH, we should really merge these formats with the proto file.. :')

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -543,7 +543,6 @@ static void LogFrameBufferError(GLenum status)
         memset(this, 0, sizeof(*this));
         m_ModificationVersion     = 1;
         m_VerifyGraphicsCalls     = params.m_VerifyGraphicsCalls;
-        m_RenderDocSupport        = params.m_RenderDocSupport;
```
Jhonnyg, line 546: Not used in opengl any longer

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -1270,7 +1269,7 @@ static void LogFrameBufferError(GLenum status)
             OpenGLIsExtensionSupported(context, "OES_texture_compression_astc") ||
             OpenGLIsExtensionSupported(context, "WEBGL_compressed_texture_astc"))
         {
-            context->m_TextureFormatSupport |= 1 << TEXTURE_FORMAT_RGBA_ASTC_4x4;
+            context->m_ASTCSupport = 1;
```
Jhonnyg, line 1272: No need to set specific formats here, they should all be supported if we have any of the extensions.

engine/texc/src/test/test_texc.cpp
```diff
@@ -425,6 +425,27 @@ TEST_P(TexcCompileTest, PreMultiplyAlpha)
...
+    settings.m_Width       = m_Width;
+    settings.m_Height      = m_Height;
+    settings.m_PixelFormat = dmTexc::PF_RGBA_ASTC_4x4;
+    settings.m_ColorSpace  = info.m_ColorSpace;
+    settings.m_Data        = m_Image->m_Data;
+    settings.m_DataCount   = m_Image->m_DataCount;
+
+    uint8_t* out = 0;
+    uint32_t out_size = 0;
+    ASSERT_TRUE(dmTexc::ASTCEncode(&settings, &out, &out_size));
```
JCash, line 446: Perhaps also check that the out size is <some_limit ?

JCash (changes requested):
Looks good.
However, a small bpp issue left I believe

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -3867,16 +3893,15 @@ namespace dmGraphics
...
-            uint8_t bpp_new           = 32;
-            uint8_t* data_new         = new uint8_t[data_pixel_count * bpp_new];
+            uint8_t* data_new         = new uint8_t[data_pixel_count * 4]; // RGBA => 4 bytes per pixel
 
             RepackRGBToRGBA(data_pixel_count, (uint8_t*) tex_data_ptr, data_new);
             vk_format     = VK_FORMAT_R8G8B8A8_UNORM;
             tex_data_ptr  = data_new;
-            tex_bpp       = bpp_new;
+            tex_bpp       = 32;
+            tex_data_size = tex_bpp * params.m_Width * params.m_Height * tex_layer_count;
```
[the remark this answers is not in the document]
Jhonnyg: Hmm, we might want to add a function that takes a width, height, format and slice count and gives a byte size back or something. Some formats are still 2 bits wide, so we need to account for that.

Issue #9787, opened 2024-11-19
Inconsistent Behavior of GUI Templates and Layouts Between Linux and HTML5 Builds


0xbeefbeef:
Describe the bug (REQUIRED)
GUI templates and layouts do not work consistently across different platforms (e.g., Linux vs HTML5). Using GUI templates is intended to avoid redundant code and design, but when supporting different layouts, this functionality breaks:

On Linux builds, GUI templates always use the Default layout, completely ignoring other defined layouts.
On HTML5 builds, the layout functionality works better, but overriding text node content in a GUI template is ignored. On Linux, text overrides work in the Default layout.

To Reproduce (REQUIRED)

Create a GUI template with multiple layouts (e.g., Default and an additional layout).
Use the GUI template in a scene.
On Linux: Attempt to use a layout other than Default.
On HTML5: Override a text node's content in the GUI template.
Build and observe the behavior differences.

Expected behavior (REQUIRED)
On Linux, the GUI template should respect the selected layout and not default to the Default layout.
On HTML5, overriding text node content in a GUI template should function as expected.

Defold version (REQUIRED):
Version 1.7.0 (unable to use newer versions due to FMOD asset compatibility issues).

Platforms: Linux, HTML5
OS: Linux Mint, HTML5 browser (tested in Chrome)

Logs (OPTIONAL):
No specific logs show errors; the issue is with inconsistent runtime behavior.

Workaround (OPTIONAL):
Currently, no viable workaround exists to ensure consistent behavior across platforms.

Screenshots (OPTIONAL):
Add screenshots of your setup, showcasing the layout behavior differences if helpful.

Additional context (OPTIONAL):
This bug limits the practicality of using GUI templates for cross-platform projects and undermines the concept of reusable and adaptable GUI components.

AGulev:
Please provide an example project where the issue you reported is reproducible.

0xbeefbeef:
[template_layout_problem.zip](https://github.com/user-attachments/files/17815570/template_layout_problem.zip)

Here is a minimal example:
- Overriding nodes doesn't work in HTML5 builds, but the layouts are applied to the parent as well as the template
- Overriding nodes seems to work in Linux build (at least in the Default layout), but layouts are only applied to the parent gui, not the template.

AGulev:
I think this issue fixed in 1.9.6 

here screenshot from 1.9.5 
[...]

Issue #7213, opened 2022-12-08
The GUI template location is not applied when changing the layout if there has been no change in the position of the parent node.


megalanthus:
**Describe the bug (REQUIRED)**
The location of the GUI template is not applied when changing the layout unless there has been a change in the position of the parent node (when running in the editor). In the build it works as expected.

**To Reproduce (REQUIRED)**
1. Create gui template.
2. Move the template to the desired positions for different layouts, and do not change the position of the parent node in the template.
3. Run project (ctrl-b)
4. When the window is resized after switching the layout, the template stays in place in the default position.

**Expected behavior (REQUIRED)**
The template is expected to move to the specified position.

**Defold version (REQUIRED):**
  - Version 1.4.0.

**Platforms (REQUIRED):**
 - Platforms: Windows
 - OS: Windows 10

**Minimal repro case project (OPTIONAL):**
[layout_with_template_bug.zip](https://github.com/defold/defold/files/10183906/layout_with_template_bug.zip)

**Screenshots (OPTIONAL):**
![image](https://user-images.githubusercontent.com/66716124/206407817-a13ec530-0c15-44c4-946c-7e19eee232c9.png)
![image](https://user-images.githubusercontent.com/66716124/206407966-a7e76dbc-e81d-4e89-b015-b3d18eba1af0.png)

AGulev:
If I understand everything right, it is fixed in 1.9.6 

1.9.5
![CleanShot 2024-12-07 at 17 56 57@2x](https://github.com/user-attachments/assets/74d93afe-ae48-467d-a381-6208bd24d092)

1.9.6
![CleanShot 2024-12-07 at 17 57 42@2x](https://github.com/user-attachments/assets/218fa815-efc7-4e3a-aa5c-aef87f726f80)


FYI @matgis

megalanthus:
> If I understand everything right, it is fixed in 1.9.6

Yes, in version 1.9.6 everything works as expected! Thank you!

Issue #4061, opened 2019-05-27
Template chain with missing layouts (DEFEDIT-322)


britzl:
When:
* there is a chain of gui templates
* the sub-most scene and super-most scene chair a layout
* the in-between scenes do not have this layout
then, the sub-most layout will not be used in the super-most. This can probably be fixed, but there is a bigger issue here. If it was used, and the layout was introduced in the in-between scene, the overrides in the super-most would be lost. This requires a lot of thinking, right now I can't see a good fix for this problem.

AGulev:
@matgis maybe this one is also related to your latest fix

matgis:
I believe this should be fixed in 1.9.6, yes!

PR #9884, opened 2024-12-04
Rewrote Gui layout editing to work efficiently with templates

### LINKED ISSUE #3987: Update references to gui resources in templates from external scenes (DEFEDIT-1034)
When a gui resource such as a Texture or Font is renamed, we should update all references to it from gui scenes that use it as a template. Currently we do this for all references inside the Template scene, but we could do this recursively for external scenes that use the template as well.
2017-06-16 09:37:38 (Mats.Gisselson)
A work-in-progress branch with failing tests for this feature exists here:
(LINK REMOVED)

matgis:
* Fixed various fundamental issues around Gui templates and layouts in the editor. You should now be able to freely combine Gui templates and layouts without worry.
* Projects that use a lot of Gui templates will load faster and use less memory.
* When renaming a resource (such as a font) inside a Gui scene, we will update references to the original name in layouts and scenes that import it as a template.
* Clearing an overridden property will no longer focus on and select all text in the cleared field in the Property Editor.

Fixes #3987, #4061, #7213, #9787

### Technical changes
* You can now tag property `fnk` arguments with `^:raw` to bypass the `value` `fnk` and get the raw property value.
* Added `g/invalidate-output` transaction step function.
* You can now specify a `:init-props-fn` in `g/override` to initialize properties on override nodes created automatically from establishing connections and so on.
* You can now inject the `_evaluation-context` as a `fnk` argument, but you must tag it with `^:unsafe` to signal that you know what you are doing.

This PR should be merged after the corresponding PR in `extension-spine`: https://github.com/defold/extension-spine/pull/192


editor/src/clj/editor/gui.clj
```diff
@@ -567,32 +598,86 @@
    (or (prop-resource-error node-id prop-kw prop-value prop-name)
        (validation/prop-error :fatal node-id prop-kw validation/prop-resource-ext? prop-value resource-ext prop-name))))
 
-;; SDK api
-(defn references-gui-resource? [evaluation-context node-id prop-kw gui-resource-name]
-  (and (g/property-value-origin? (:basis evaluation-context) node-id prop-kw)
-       (= gui-resource-name (g/node-value node-id prop-kw evaluation-context))))
-
```
matgis, line 574: I've changed how this works. The new system supports updating references to renamed gui resources inside template overrides and layouts, and even across scenes as long as the gui resource name is not shadowed by another resource in a referencing gui.

I added a bunch of tests for this in `integration.gui-test/template-layout-resource-rename-test`.

editor/src/clj/editor/gui.clj
```diff
@@ -661,11 +746,190 @@
         :type type ; Explicitly include the type (pb-field is optional, so :type-box would be stripped otherwise).
         :child-index child-index))) ; Used to order sibling nodes in the SceneDesc.
 
+;; SDK api
+(defmacro layout-property-getter [prop-sym]
+  {:pre [(symbol? prop-sym)]}
+  (let [prop-kw (keyword prop-sym)]
+    `(g/fnk [~'prop->value ~prop-sym]
+       (get ~'prop->value ~prop-kw ~prop-sym))))
```
matgis, line 754: Layout-specific overrides are now stored in a `:layout->prop->override` map in the `GuiNodes`. We produce the `prop->value` output from it and the `current-layout`, merging in the overrides from override originals recursively to produce the final map. However, if we're using the default layout, we can immediately return `nil` from the `prop->value` output, and the raw property value of the `prop-sym` will be used.

editor/src/clj/editor/gui.clj
```diff
@@ -661,11 +746,190 @@
...
+
+(defn layout-property-set-fn [_evaluation-context node-id _old-value _new-value]
+  (concat
+    (g/invalidate-output node-id :prop->value)
+    (g/invalidate-output node-id :layout->prop->value)))
```
matgis, line 759: We explicitly invalidate these outputs whenever a layout property is set. They both evaluate properties in a bit of a naughty way using `g/node-value` and `_this` in their `fnks`.

* `prop->value` is used in the visual presentation, and is lighter to produce since it focuses on the `current-layout` only.
* `layout->prop->value` is used to produce `build-targets` and `save-value` only.

editor/src/clj/editor/gui.clj
```diff
@@ -661,11 +746,190 @@
...
+
+;; SDK api
+(defmacro layout-property-setter [prop-sym]
+  {:pre [(symbol? prop-sym)]}
+  layout-property-set-fn)
```
matgis, line 764: We don't actually need the `prop-sym` since they all use the same setter, but I added it for symmetry with `layout-property-getter`, and it means we can decide to use it if we need to in the future without updating `extension-spine` yet again.

editor/src/clj/editor/gui.clj
```diff
@@ -661,11 +746,190 @@
...
+(defn- update-layout-property [evaluation-context node-id prop-kw update-fn & args]
+  (let [old-value (g/node-value node-id prop-kw evaluation-context)
+        new-value (apply update-fn old-value args)
+        current-layout (g/node-value node-id :current-layout evaluation-context)]
+    (if (str/blank? current-layout)
+      (g/set-property node-id prop-kw new-value)
+      (g/update-property
+        node-id :layout->prop->override
+        update current-layout
+        assoc prop-kw new-value))))
```
matgis, line 823: This is the mechanism for updating the value of a layout property. We use this function in particular in our custom implementations of the transform manipulator `defmethods`. Moving a node in the scene view should update its pose in the `current-layout` only, after all.

editor/src/clj/editor/gui.clj
```diff
@@ -661,11 +746,190 @@
...
+      (if (str/blank? current-layout)
+        (coll/mapcat
+          #(g/clear-property node-id %)
+          cleared-prop-kws)
+        (g/update-property
+          node-id :layout->prop->override
+          update current-layout
+          (fn [prop->override]
+            (coll/not-empty
+              (apply dissoc prop->override cleared-prop-kws))))))))
```
matgis, line 869: We use the `:edit-type` mechanism to ensure the property editor updates values in the correct place. Either directly on the edited property, or inside the `:layout->prop->override` map.

editor/src/clj/editor/gui.clj
```diff
@@ -661,11 +746,190 @@
...
+            (value (layout-property-getter position))
+            (set (layout-property-setter position)))
+  (property rotation types/Vec4 (default scene/default-rotation)
+            (dynamic edit-type (layout-property-edit-type rotation properties/quat-rotation-edit-type))
+            (value (layout-property-getter rotation))
+            (set (layout-property-setter rotation)))
+  (property scale types/Vec3 (default scene/default-scale)
+            (dynamic edit-type (layout-property-edit-type scale {:type types/Vec3}
+                                 (fn [evaluation-context self prop-kw old-value new-value]
+                                   {:scale (some-> new-value scene/non-zeroify-scale)})))
```
matgis, line 929: We sometimes need to do something special when setting values. Instead of the usual `(set (fn []))` declaration on the property, we have the ability to specify a `:changes-fn` in the `edit-type` that returns a map of `{:prop-kw prop-value}` that should be set, and the `:set-fn` in the `edit-type` will make sure it ends up in the correct place (I.e.  directly on the edited property, or inside the `:layout->prop->override` map).

editor/src/clj/editor/gui.clj
```diff
@@ -764,19 +1039,25 @@
...
-                            (into [node-msg]
-                                  (map #(dissoc % :child-index))
-                                  (flatten
-                                    (sort-by #(get-in % [0 :child-index])
-                                             node-msgs)))))
+  (output node-msgs g/Any (g/fnk [layout->prop->override layout->prop->value node-msg node-msgs]
+                            (let [decorated-node-msg
+                                  (assoc node-msg
+                                    :layout->prop->override layout->prop->override
+                                    :layout->prop->value layout->prop->value)]
```
matgis, line 1055: We decorate the `node-msgs` with these. They are used in the `build-targets` and `save-value` outputs of `GuiSceneNode`.

editor/src/clj/editor/gui.clj
```diff
@@ -810,8 +1091,123 @@
   (output node-overrides g/Any :cached (g/fnk [node-overrides id _overridden-properties]
                                          (into {id _overridden-properties}
                                                node-overrides)))
+  (input layout-names g/Any)
+  (output layout-names g/Any (gu/passthrough layout-names))
```
matgis, line 1095: Layout names are passed from scene to all nodes so we can figure out where a layout was introduced along the override chain.

editor/src/clj/editor/gui.clj
```diff
@@ -810,8 +1091,123 @@
...
+                                   layout-names-for-original
+                                   (merge (get layout->prop->override-for-original current-layout)
+                                          prop->value))))))
+
+                    ;; We've reached the override root. Anything not overridden
+                    ;; for the current-layout by this point will use values from
+                    ;; our default layout.
+                    (let [node (g/node-by-id basis node-id)]
+                      (merge (make-prop->value-for-default-layout node)
+                             prop->value))))))))
```
matgis, line 1177: The `layout->prop->value` and `prop->value` outputs do similar work, but `prop->value` can be more efficient by focusing on the `current-layout`. It's worth it, because it is used in the `value` `fnk` of every layout property. However, it could also be implemented as simply looking up the `current-layout` in `layout->prop->value`.

editor/src/clj/internal/node.clj
```diff
@@ -1727,11 +1748,23 @@
...
+                                       false
+
+                                       (contains? properties prop-kw)
+                                       (:assoc-original-value? prop true)
+
+                                       (not (declared? prop-kw))
+                                       (:assoc-original-value? prop false)
+
+                                       :else
+                                       false)]
```
matgis, line 1765: We do some trickery in the `_properties` output of `GuiNode` to make properties appear overridden based on what `current-layout` is selected for editing. For our purposes, we need to stop this code from `associng` in the `:original-value` from the original node. You can now set `:assoc-original-value?` to `false` in the produced property to disable this. Previously you could only set it to `true` to force it to happen.

editor/src/clj/internal/transaction.clj
```diff
@@ -330,9 +338,11 @@
...
   (let [successors (get ctx :successors-changed)]
     (assoc ctx :successors-changed (reduce (fn [successors [node-id label]]
-                                             (if-let [node-succ (get successors node-id #{})]
-                                               (assoc successors node-id (conj node-succ label))
-                                               successors)) ; Found nil - all successors already flagged as changed.
+                                             (let [node-succ (get successors node-id ::not-found)]
+                                               (case node-succ
+                                                 nil successors ; Found nil - all successors already flagged as changed.
+                                                 ::not-found (assoc successors node-id #{label})
+                                                 (assoc successors node-id (conj node-succ label)))))
```
matgis, line 345: Bug fix: Hmm, this looked pretty bad actually. 🤔 Looks like we would throw away the effect of calling `flag-all-successors-changed` if `flag-successors-changed` was called afterwards. 😬 Am I reading it wrong?

editor/test/integration/gui_test.clj
```diff
@@ -365,9 +385,9 @@
                   (gui-node gui-node-name)
                   (select-labels labels)))]
 
-      (let [button-gui-resource (workspace/find-resource workspace "/main/button.gui")
-            panel-gui-resource (workspace/find-resource workspace "/main/panel.gui")
-            window-gui-resource (workspace/find-resource workspace "/main/window.gui")
+      (let [button-gui-resource (workspace/find-resource workspace "/gui/template/button.gui")
+            panel-gui-resource (workspace/find-resource workspace "/gui/template/panel.gui")
+            window-gui-resource (workspace/find-resource workspace "/gui/template/window.gui")
```
matgis, line 390: I moved these files in the gui test project because I added a bunch of new scenes to it and felt it needed reorganising a bit.

editor/test/integration/gui_test.clj
```diff
@@ -786,10 +822,7 @@
   (#'gui/move-child-node! node-id offset))
 
 (defn- scene-gui-node-map [scene]
-  (into {}
-        (map (juxt :label :node-id)
-             (tree-seq :children :children (g/node-value scene :node-outline)))))
-
+  (g/node-value scene :node-ids))
```
matgis, line 825: The `node-ids` output on the `GuiSceneNode` is exactly this - a map of `{:name :node-id}`. Although `:name` is called `:id` on the `GuiNode`.

editor/test/integration/save_data_test.clj
```diff
@@ -1304,7 +1304,7 @@
   ;; being overridden from a template node in one of the root-level files in the
   ;; save data test project. If you add a field to the NodeDesc protobuf
   ;; message, you'll either need to add a template override for it (we suggest
-  ;; you add it to `checked02.gui`, which hosts the majority of the template
+  ;; you add it to `checked01.gui`, which hosts the majority of the template
```
matgis, line 1307: Previously we separated all template override checks from the layout override checks into distinct `.gui` resources, because they didn't play well together when combined. Now we want to check more extensively that both layout and template overrides are saved when they are used in combination. We now have `checked01.gui`, that references `checked02.gui` as a template. `checked01.gui` overrides all properties on all node types, both in its own nodes and on nodes imported from `checked02.gui`. In addition, a Landscape layout overrides all these properties across both scenes to different values.

editor/test/integration/save_data_test.clj
```diff
@@ -1332,7 +1332,7 @@
   (test-util/with-loaded-project project-path
     (test-util/clear-cached-save-data! project)
     (doseq [save-data (project->save-datas project)]
-      (test-util/check-save-data-disk-equivalence! save-data))))
+      (test-util/check-save-data-disk-equivalence! save-data project-path))))
```
matgis, line 1335: Fix: The message on test failure would report the wrong path to the resources on disk.

vlaaad (dismissed):
Lookin' good!

editor/src/clj/internal/util.clj
```diff
@@ -147,9 +147,9 @@
   [pred m]
   (into {} (filter pred m)))
 
-(defn key-set
-  [m]
-  (set (keys m)))
+(defmacro key-set [m]
+  ;; TODO: Replace all calls so we can get rid of this macro.
```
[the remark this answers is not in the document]
matgis: I wanted to move the `key-set` function to the `coll` module, because the `util` module depends on a lot of other modules. However, `util/key-set` is used in a bunch of places and I didn't want to complicate this PR further. I'll do a separate PR and change all calls to `util/key-set` to use `coll/key-set` instead later.

editor/src/clj/editor/gui.clj
```diff
@@ -661,11 +746,190 @@
...
+
+(defn- node-type-deref->stripped-prop-kws-raw [node-type-deref]
+  {:pre [(in/node-type-deref? node-type-deref)]}
+  (->> (:property node-type-deref)
+       (e/keep
+         (fn [[prop-kw prop-def]]
+           (when-not (layout-property-prop-def? prop-def)
+             prop-kw)))
+       (sort)
+       (vec)))
```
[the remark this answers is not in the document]
matgis: I think `core.sort` copies the input into a Java array and then does an in-place mutating quick sort. I don't think we can do much better than that efficiency-wise. But sometimes you just want the results as a vector, you know. 😅

editor/src/clj/editor/gui.clj
```diff
@@ -2454,32 +2829,71 @@
...
+                          (reduce
+                            (fn [node-desc [pb-field pb-value]]
+                              (let [default-pb-value (default-node-desc-pb-field-values pb-field)]
+                                (if (= default-pb-value pb-value)
+                                  (dissoc node-desc pb-field)
+                                  (assoc node-desc pb-field pb-value))))
+                            (-> node-msg
+                                (dissoc :layout->prop->override :layout->prop->value)
+                                (protobuf/assign-repeated :overridden-fields overridden-fields)))
+                          (strip-unused-overridden-fields-from-node-desc)))))))))
```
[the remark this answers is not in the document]
matgis: It does, but you need to indent the whole thing using the `->>` macro if you want to have the input up front. I think it just reads better this way. 🤷

editor/src/clj/editor/gui.clj
```diff
@@ -661,11 +746,190 @@
...
+
+;; SDK api
+(defmacro layout-property-setter [prop-sym]
+  {:pre [(symbol? prop-sym)]}
+  layout-property-set-fn)
+
+(defn- layout-property-prop-def? [prop-def]
+  (let [value-dependencies (-> prop-def :value :dependencies)]
+    (and (= 2 (count value-dependencies))
+         (contains? value-dependencies :prop->value))))
```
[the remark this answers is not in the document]
matgis: It depends on only `prop->value` and the raw property argument.

editor/src/clj/editor/gui.clj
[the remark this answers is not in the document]
vlaaad: I meant e.g. this:
```clj
(let [src (vec (shuffle (range 64)))]
  (println "sort->seq->vec")
  (criterium.core/quick-bench
    (vec (sort src)))
  (println "sort->vec")
  (criterium.core/quick-bench
    (let [a (to-array src)]
      (java.util.Arrays/sort a)
      (with-meta (clojure.lang.LazilyPersistentVector/createOwning a) (meta src)))))
=>
sort->seq->vec
             Execution time mean : 5,729477 µs
sort->vec
             Execution time mean : 2,157710 µs
```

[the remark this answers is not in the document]
vlaaad: Vector, twice as fast :)

editor/src/clj/editor/gui.clj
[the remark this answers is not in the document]
vlaaad: Got it!

editor/src/clj/editor/gui.clj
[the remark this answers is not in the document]
vlaaad: Hmm, I now realized that all speedup here is only caused because my implementation didn't use `compare` comparator.

vlaaad (approved):
Shipit!

PR #9890, opened 2024-12-05
Show default build server as a prompt text in prefs dialog

vlaaad:
Technical notes:
When the text field is empty, we can now display greyed-out prompt text in the prefs dialog. Now we use it for the build server text field.


AGulev (approved):
👍

matgis (approved):
Ship it!

PR #9931, opened 2024-12-16
Align astc to block sizes

Jhonnyg:
ASTC textures needs to be aligned (and padded) to nearest block size.


com.dynamo.cr/com.dynamo.cr.bob/src/com/defold/extension/pipeline/texture/ITextureCompressor.java
```diff
@@ -23,4 +24,8 @@ public interface ITextureCompressor {
     public String getName();
 
     public byte[] compress(TextureCompressorPreset preset, TextureCompressorParams params, byte[] input);
+
+    public default int getAlignedWidth(TextureImage.TextureFormat format, int width) { return width; }
+
+    public default int getAlignedHeight(TextureImage.TextureFormat format, int height) { return height; }
```
Jhonnyg, line 30: An alternative to changing the API would be to check specifically in the texture compressor if we are using an ASTC format, but at least this lets the plugins decide themselves how to calculate the alignment.

engine/graphics/proto/graphics/graphics_ddf.proto
```diff
@@ -214,6 +214,7 @@ message TextureImage
         optional CompressionType compression_type = 9 [default = COMPRESSION_TYPE_DEFAULT];
         optional uint64 compression_flags         = 10;
         repeated uint32 mip_map_size_compressed   = 11;
+        repeated uint32 mip_map_dimensions        = 12; // w0, h0, w1, h1, ...
```
Jhonnyg, line 217: I think we can do a cleanup pass on this data format later when we migrate mipmaps into a separate image.

engine/gamesys/src/gamesys/gamesys_resource.cpp
```diff
@@ -134,6 +142,7 @@ namespace dmGameSystem
             GRAPHCIS_TO_TEXTURE_IMAGE_ENUM_CASE(TEXTURE_FORMAT_RG16F);
             GRAPHCIS_TO_TEXTURE_IMAGE_ENUM_CASE(TEXTURE_FORMAT_R32F);
             GRAPHCIS_TO_TEXTURE_IMAGE_ENUM_CASE(TEXTURE_FORMAT_RG32F);
+            default:break;
```
Jhonnyg, line 145: Fix some warnings.

engine/gamesys/src/gamesys/resources/res_texture.cpp
```diff
@@ -272,11 +272,10 @@ namespace dmGameSystem
...
 
-                    params.m_MipMap   = i;
-                    dmGraphics::SetTextureAsync(texture, params, 0, 0);
+                    params.m_MipMap = i;
+                    params.m_Width  = image->m_MipMapDimensions[i * 2];
+                    params.m_Height = image->m_MipMapDimensions[i * 2 + 1];
 
-                    params.m_Width >>= 1;
-                    params.m_Height >>= 1;
                     if (params.m_Width == 0)
```
Jhonnyg, line 279: Can't calculate it like this for compressed formats due to alignment.

JCash (approved):

PR #9943, opened 2024-12-17
Fix frustrum culling for sprites with pivot point

### LINKED ISSUE #9917: Frustum culling does not take into account sprite pivot points
**Describe the bug**
Modifying a sprite's pivot point means frustum culling behaves incorrectly, e.g. objects being culled even though they should be visible.

In this example I have created two versions of a sprite, one with a default y pivot (0.5) and one of a y pivot of 1.

All of these sprites should be visible, but the leftmost two should be mostly off screen:
image

In practice, the second sprite is incorrectly culled:
image

**Defold version:**
1.9.5

[Frustum_pivot.zip](https://github.com/user-attachments/files/18125875/Frustum_pivot.zip)

ekharkunov:
Fixed culling for sprites with pivot points. Also implemented some optimization on how geometry data calculated and stored between batches. So no it should be a bit faster in terms of calculation but consume some memory to store cached information.

Fixes #9917 

### Technical changes
`TexturesData` completely removed because all data can be retrieved from `SpriteComponent`.
In some inner function replace 3 or 4 function arguments with 1 SpriteComponent and access data inside the function.

For animation data cache used the same approach as it was implemented in `ResourceChunkCache`. The only difference is cache eviction rule. I counted how every sprite world update. Every time when cache entry is retrieved from the cache - updated special field in cache entry (like timestamp) which stores what engine tick entry was read at.
2 structures to work with cache are used:
* hastable to fast access to cache data by key
* double linked list to implement Least Recently Used cache

When cache entry is accessed - record in LRU cache moved to head. Then when cache eviction rule is checked - I iterate from the end of the LRU until I found the first cache entry which should not be evicted.

P.S. Checked rotated image in atlas with pivot point - calculation is already correct.

Measurement:
Project contains 2 collection. 5 times collection will be changed (one collection unloaded, second - load). Every collection contains a set of game objects. First collection contains 4 types of game object. Every game object has their own texture with one animation. Second collection contains 5 types of game object. Every object contains their own texture with one sprite.
Every collection loads, spawn 6k of random objects, then remove 5k, then spawn 5k and load next collection.

Memory
|Build|Peak (Mb)| Average (Mb)|
|-----|----------|--------------|
|base |110,38     |~92                |
|cache|100,27    |~83,5             |

Time
base:  CompSpriteRender -2,68s, CompSpriteUpdate - 820 ms
cache: CompSpriteRender - 2,92s, CompSpriteUpdate - 901ms

Base:

Screenshot 2025-05-26 at 00 57 59
Screenshot 2025-05-26 at 00 57 40

Cache:
Screenshot 2025-05-26 at 01 02 51
Screenshot 2025-05-26 at 01 03 38

Another set of optimizations was done in https://github.com/defold/defold/pull/10736

## PR checklist

* [x] Code
* [x] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -65,8 +65,27 @@ namespace dmGameSystem
...
+        float                           m_PageIndices[MAX_TEXTURE_COUNT];
+
+        const dmGameSystemDDF::TextureSetAnimation* m_Animations[MAX_TEXTURE_COUNT];
+        const dmGameSystemDDF::SpriteGeometry*      m_Geometries[MAX_TEXTURE_COUNT];
+        bool                                        m_UsesGeometries;
+    };
+
     struct SpriteComponent
     {
+        TexturesData                m_TexturesData; // store precalculated gemetry data that used in different places during the frame
```
ekharkunov, line 88: Stores textures data to avoid double calculation in `UpdateTextureDataAndCounts` (ex `UpdateVertexAndIndexCount`) and in `CreateVertexData`. So calculate once in `UpdateTextureDataAndCounts` and use that result in `CreateVertexData`.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1218,14 +1213,12 @@ namespace dmGameSystem
...
             uint32_t component_index         = (uint32_t)buf[*i].m_UserData;
             const SpriteComponent* component = (const SpriteComponent*) &components[component_index];
+            const TexturesData& textures     = component->m_TexturesData;
             const Matrix4& world_matrix      = component->m_World;
 
             float sp_width  = component->m_Size.getX();
             float sp_height = component->m_Size.getY();
 
-            // Get the correct animation frames, and other meta data
-            ResolveAnimationData(&textures, component->m_CurrentAnimation, component->m_CurrentAnimationFrame);
```
ekharkunov, line 1227: No need to do calculation here any more.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1737,11 +1736,12 @@ namespace dmGameSystem
         }
     }
 
-    static void UpdateVertexAndIndexCount(SpriteWorld* sprite_world, dmRender::HRenderContext render_context)
+    static void UpdateTextureDataAndCounts(SpriteWorld* sprite_world, dmRender::HRenderContext render_context)
```
ekharkunov, line 1739: Update function name to be more meaningful to new functionality.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1962,7 +1962,25 @@ namespace dmGameSystem
...
+                    float pivot_x = 0.f;
+                    float pivot_y = 0.f;
+                    const dmGameSystemDDF::SpriteGeometry* geometry = component.m_TexturesData.m_Geometries[0];
+                    if (geometry)
+                    {
+                        pivot_x = geometry->m_PivotX;
+                        pivot_y = geometry->m_PivotY;
+                    }
+
+                    trans = world_transform * Point3(-pivot_x, -pivot_y, 0.0f);
```
ekharkunov, line 1978: Main fix. I looked that `m_WorldsPosition` in `RenderListEntry` used only for sorting and for visibility check so it's ok to set `m_WorldPosition` to value with pivot points.
Let me know if it's not ok.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
ekharkunov, line 88: Probably we can cache only geometries and left double calculation 'as-is' to avoid grows of memory footprint. Let's discuss.

JCash: I was a bit surprised to see this in a PR regarding the pivot point?

Also, I'd expect some numbers on the trade offs for this.
E.g. now the sprite component structure becomes much larger.
I.e. what was the size of 32k sprites before/after?
What is the _measured_ performance benefit.
I feel we need those things before making a judgement call, especially as we are currently looking to minimize runtime footprint where available.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
JCash: Structurally, it feels a bit off, since the other application of the pivot is for vertex calculations, but here a special case is added on the object's world position.

So now we apply pivot in "two" places.

Also, the pivot isn't about changing the position of the component, but shifting the visual representation.
So this change should go into the corresponding place for the vertex calculation.

JCash (changes requested):
Please split change into two PR's.
Also use the pivot when generating the vertices, to leave the object position as-is.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
ekharkunov: Adding those data to `SpriteComponent` is more about to solve current issue.
Pivots we calculate only inside `CreateVertexData`. Call chain: `RenderListDispatch`->`RenderBatch`->`CreateVertexData`. So `RenderListFrustumCulling` is calling before pivot points will be calculated.
So I guess more correct solution is to precalculate pivot points and stored it in `SpriteWorld` like as done for `m_BoundingVolumes`.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -65,6 +66,31 @@ namespace dmGameSystem
         SpriteResourceOverrides() : m_Material(0) {}
     };
 
+    const uint32_t MAX_TEXTURE_COUNT = dmRender::RenderObject::MAX_TEXTURE_COUNT;
+    const uint8_t CACHE_EVICTION_FRAMES = 10; // how many frames cache entry is stored before eviction
+
+    struct AnimationData
```
ekharkunov, line 72: Use the same approach as in ResourceChunkCache.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -65,6 +66,31 @@ namespace dmGameSystem
...
+    {
+        dmDoubleLinkedList::ListNode    node;
+        // Used after resolving info from all textures
+        dmhash_t                        m_AnimationID;                      // The animation of the driving atlas
+        uint32_t                        m_Frames[MAX_TEXTURE_COUNT];        // The resolved frame indices
+        uint32_t                        m_LastAccessTick;
+        uint32_t                        m_CacheKey;
```
ekharkunov, line 79: Added cache key as a field because in case when cache entry will be removed because of cache eviction we should know the key to correctly update hashtable.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -65,6 +66,31 @@ namespace dmGameSystem
...
+    const uint32_t MAX_TEXTURE_COUNT = dmRender::RenderObject::MAX_TEXTURE_COUNT;
+    const uint8_t CACHE_EVICTION_FRAMES = 10; // how many frames cache entry is stored before eviction
+
+    struct AnimationData
+    {
+        dmDoubleLinkedList::ListNode    node;
+        // Used after resolving info from all textures
+        dmhash_t                        m_AnimationID;                      // The animation of the driving atlas
+        uint32_t                        m_Frames[MAX_TEXTURE_COUNT];        // The resolved frame indices
+        uint32_t                        m_LastAccessTick;
```
ekharkunov, line 78: Use engine tick as a marker for cache eviction.
I can use for ticks counting `uint16_t` instead of `uint32_t` because overflow case correctly handles.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -65,6 +66,31 @@ namespace dmGameSystem
...
+        const dmGameSystemDDF::TextureSetAnimation* m_Animations[MAX_TEXTURE_COUNT];
+        const dmGameSystemDDF::SpriteGeometry*      m_Geometries[MAX_TEXTURE_COUNT];
+        bool                                        m_UsesGeometries;
+    };
+
+    struct AnimationDataCache
+    {
+        dmHashTable32<AnimationData*> m_Cache;
+        dmDoubleLinkedList::List      m_LRU;
+        uint32_t                      m_CurrentEngineTick;
```
ekharkunov, line 91: That filed updates every call of `CompSpriteUpdate`

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -137,22 +162,6 @@ namespace dmGameSystem
         uint8_t                             m_ReallocBuffers : 1;
     };
 
-    struct TexturesData
```
ekharkunov, line 140: Completely removed that structure because all information contains inside `SpriteComponent`

engine/resource/src/resource_chunk_cache.cpp
```diff
@@ -16,28 +16,17 @@
...
 #include <dlib/array.h>
+#include <dlib/double_linked_list.h>
 #include <dlib/hash.h>
 #include <dlib/log.h>
 
 #include <stdio.h>   // printf
 #include <stddef.h>  // offsetof
 #include <algorithm> // lower_bound
 
-struct DLListNode
```
ekharkunov, line 26: Moved to dlib to use in comp_sprite.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1844,6 +1942,8 @@ namespace dmGameSystem
          */
 
         SpriteWorld* world = (SpriteWorld*)params.m_World;
+        UpdateCache(world);
```
JCash, line 1945: Perhaps rename to UpdateAnimationDataCache() ?

engine/dlib/src/dlib/double_linked_list.h
```diff
@@ -0,0 +1,42 @@
+// Copyright 2020-2025 The Defold Foundation
```
JCash, line 1: The resource chunk cache has some unit tests, but now that the api is formalized, we need to add tests for this container too.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1182,7 +1230,46 @@ namespace dmGameSystem
...
-    static void CreateVertexData(SpriteWorld* sprite_world, dmGraphics::VertexAttributeInfos* material_attribute_info, bool has_local_position_attribute, uint8_t** vb_where, uint8_t** ib_where, dmRender::RenderListEntry* buf, uint32_t* begin, uint32_t* end)
+    static AnimationData* GetOrCreateAnimationData(SpriteWorld* sprite_world, const SpriteComponent* component)
+    {
+        // component, component->m_MixedHash, component->m_CurrentAnimation, component->m_CurrentAnimationFrame
+        HashState32 state;
+        dmHashInit32(&state, false);
+        dmHashUpdateBuffer32(&state, &component->m_MixedHash, sizeof(component->m_MixedHash));
+        dmHashUpdateBuffer32(&state, &component->m_CurrentAnimation, sizeof(component->m_CurrentAnimation));
+        dmHashUpdateBuffer32(&state, &component->m_CurrentAnimationFrame, sizeof(component->m_CurrentAnimationFrame));
+        uint32_t hash = dmHashFinal32(&state);
```
JCash, line 1241: As the animation doesn't generally change each update, it might be appropriate to update this when the m_CurrentAnimation/m_CurrentAnimationFrame changes, and then use a m_AnimationHash to do this lookup.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -65,6 +66,31 @@ namespace dmGameSystem
...
+        uint32_t                        m_LastAccessTick;
+        uint32_t                        m_CacheKey;
+        float                           m_PageIndices[MAX_TEXTURE_COUNT];
+
+        const dmGameSystemDDF::TextureSetAnimation* m_Animations[MAX_TEXTURE_COUNT];
+        const dmGameSystemDDF::SpriteGeometry*      m_Geometries[MAX_TEXTURE_COUNT];
+        bool                                        m_UsesGeometries;
+    };
+
+    struct AnimationDataCache
```
JCash, line 87: It would be good with a note on _why_ we want to use a cache at all, as opposed to using a ref count for the structure.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
JCash, line 72: I'd also like a code comment about how the animation data usage is intended to work and why.
This component is getting quite complex, and I think we'll need more documentaion moving on.

JCash (changes requested):
Overall good. Only smaller things needed
* Can pre calculate the animation state hash
* Needs some code comments
* Need a unit test for the new container
* What's the performance comparison before/after this change?

engine/gamesys/src/gamesys/components/comp_sprite.cpp
ekharkunov: Renamed

engine/dlib/src/dlib/double_linked_list.h
ekharkunov: Added test_double_linked_list.cpp

JCash (approved):

PR #9947, opened 2024-12-17
Restore WebGPU check using proper class GPUCanvasContext #9581

smagnuso:
I need to check for WebGPU as I do not have WebGL available.


britzl (approved):
Looks good

PR #9950, opened 2024-12-18
Append raw texture data after protobuf header

### LINKED ISSUE #9605: Look into using a regular binary format for .texturec
The benefit would be to avoid creating an extra copy of the payload as a Protobuf message.

Note: we can still use the protobuf as the header of the file, and append the payload afterwards. 

Related: https://github.com/defold/defold/issues/8738

Jhonnyg:
The engine format for textures has been modified so that the data is stored after the protobuf message. This means that no texture unpacking is needed, and no extra buffer is needed to copy the data from the resource file.

Fixes #9605


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/TextureGenerator.java
```diff
@@ -401,11 +405,7 @@ private static TextureImage.Image.Builder generateFromColorAndFormat(String name
                 byte[] uncompressed = TexcLibraryJni.GetData(mipImage);
                 int mipWidth        = TexcLibraryJni.GetWidth(mipImage);
                 int mipHeight       = TexcLibraryJni.GetHeight(mipImage);
-
-                String paramsName = name;
-                if (paramsName == null) {
-                    paramsName = "MipMap_" + mipMapLevel;
-                }
```
Jhonnyg, line 408: Name was always passed as null to the function

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TextureUtil.java
```diff
@@ -214,8 +217,19 @@ public static TextureProfile getTextureProfileByPath(TextureProfiles textureProf
         return null;
     }
 
-    public static TextureImage createCombinedTextureImage(TextureImage[] textures, TextureImage.Type type) throws TextureGeneratorException {
-        int numTextures = textures.length;
+    private static int getTextureGenerateResultByteIndex(TextureImage textureImage, int alternative, int mipMap) throws TextureGeneratorException {
```
Jhonnyg, line 220: This can probably be written differently, but I didn't understand the old code so I opted for the simplest solution.

editor/project.clj
```diff
@@ -281,7 +281,7 @@
                                                     "-Ddefold.extension.simpledata.url=https://github.com/defold/extension-simpledata/archive/refs/tags/v1.1.0.zip"
                                                     "-Ddefold.extension.spine.url=https://github.com/defold/extension-spine/archive/refs/tags/3.6.2.zip"
                                                     "-Ddefold.extension.teal.url=https://github.com/defold/extension-teal/archive/main.zip"
-                                                    "-Ddefold.extension.texturepacker.url=https://github.com/defold/extension-texturepacker/archive/refs/tags/1.2.zip"
+                                                    "-Ddefold.extension.texturepacker.url=https://github.com/defold/extension-texturepacker/archive/refs/heads/update-to-new-texture-format.zip"
```
Jhonnyg, line 284: When this has been approved, I will merge TP extension PR and use the new release.

editor/test/resources/save_data_project/checked.tpinfo
```diff
@@ -41,6 +41,10 @@ pages {
       x: 47.0
       y: 25.0
     }
+    pivot {
+      x: 32
+      y: 32
+    }
```
Jhonnyg, line 47: We were using an old TP extension, the new one has pivots so it's included now

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TextureUtil.java
JCash: Yeah, I want the "alternatives" to be removed.
It's not the way we want to provide alternative texture packs imo.

engine/graphics/proto/graphics/graphics_ddf.proto
```diff
@@ -209,20 +209,26 @@ message TextureImage
...
-        required bytes data                       = 8;
-        optional CompressionType compression_type = 9 [default = COMPRESSION_TYPE_DEFAULT];
-        optional uint64 compression_flags         = 10;
-        repeated uint32 mip_map_size_compressed   = 11;
-        repeated uint32 mip_map_dimensions        = 12; // w0, h0, w1, h1, ...
+        repeated uint32 mip_map_size              = 7; // always uncompressed (native) size
+        optional CompressionType compression_type = 8 [default = COMPRESSION_TYPE_DEFAULT];
+        repeated uint32 mip_map_size_compressed   = 9;
+        repeated uint32 mip_map_dimensions        = 10; // w0, h0, w1, h1, ...
+        optional uint32 data_size                 = 11;
```
JCash, line 216: What does this represent? isn't it either `mip_map_size` or `mip_map_size_compressed`, depending on the `compression_type`?

Perhaps add a comment in the file?

JCash (dismissed):
lgtm

AGulev (dismissed):
👍

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TextureUtil.java
matgis, line 220: The method name is a bit confusing. `ByteIndex` sounds like it refers to an offset in a continuous array of bytes, but it is really the index of one of the byte arrays inside the `GenerateResult.imageDatas`.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TextureUtil.java
```diff
@@ -225,36 +239,35 @@ public static TextureImage createCombinedTextureImage(TextureImage[] textures, T
...
 
-            } catch (IOException e) {
-                throw new TextureGeneratorException(e.getMessage());
+            for (int j = 0; j < mipMapCount; j++) {
+
+                for (int k = 0; k < numTextures; k++) {
+                    int mipSize = generateResults[k].textureImage.getAlternatives(i).getMipMapSize(j);
+                    alternativeImageBuilder.addMipMapSizeCompressed(mipSize);
+
+                    ArrayList<byte[]> textureBytes = generateResults[k].imageDatas;
```
matgis, line 260: Consider renaming to `imageDatas` or `mipImageByteArrays` for clarity?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TextureUtil.java
```diff
@@ -225,36 +239,35 @@ public static TextureImage createCombinedTextureImage(TextureImage[] textures, T
...
+                    int byteIndex = getTextureGenerateResultByteIndex(generateResults[k].textureImage, i, j);
```
matgis, line 261: `mipImageIndex`?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TextureUtil.java
```diff
@@ -274,19 +289,65 @@ public static BufferedImage Resize(BufferedImage img, int newW, int newH, int ty
...
+        int currentIndex = 0;
+        for (byte[] array : arrays) {
+            System.arraycopy(array, 0, result, currentIndex, array.length);
+            currentIndex += array.length;
+        }
+
+        return result;
+    }
+
+    public static byte[] generateResultToByteArray(TextureGenerator.GenerateResult generateResult) {
```
matgis, line 312: The method name is a bit confusing to me.

What exactly is in this `ByteArray` that this method returns? It seems to return the bytes we should write to a `.texturec` resource when building?

Maybe we could name it `generateResultToTextureResourceBytes` or something like that?

Although to be honest, I would prefer a `writeTextureResource(OutputStream)` method on `GenerateResult` that writes it to an `OutputStream` instead of returning a concatenated byte array like this one does. Then the editor could be changed to write the `GenerateResult` directly to an output `.texturec` file without having to allocate the concatenated byte array. And you could implement this helper method by calling that with a `ByteArrayOutputStream`. If you know the output byte size, you can specify the capacity of the `ByteArrayOutputStream` when constructing it and avoid intermediate allocations during the write, so it might be helpful to have a `calculateTextureResourceSize()` method on `GenerateResult` as well for this purpose.

editor/src/clj/editor/pipeline/tex_gen.clj
```diff
@@ -90,12 +91,11 @@
...
 (defn assemble-texture-images
-  ^Graphics$TextureImage [texture-images max-page-count]
-  (let [texture-images (into-array Graphics$TextureImage texture-images)
-        texture-type (if (pos? max-page-count)
+  ^TextureGenerator$GenerateResult [texture-generator-results max-page-count]
+  (let [texture-type (if (pos? max-page-count)
                        Graphics$TextureImage$Type/TYPE_2D_ARRAY
                        Graphics$TextureImage$Type/TYPE_2D)]
-    (TextureUtil/createCombinedTextureImage texture-images texture-type)))
+    (TextureUtil/createCombinedTextureImage (into-array texture-generator-results) texture-type)))
```
matgis, line 98: `TextureUtil/createCombinedTextureImage` is called in the `assemble-cubemap-texture-images` function below as well, so you need to update that to return a `TextureGenerator$GenerateResult` and all callers of that too.

editor/src/clj/editor/image.clj
```diff
@@ -37,9 +39,10 @@
         image ((:f content-generator) (:args content-generator))]
     (g/precluding-errors
       [image]
-      (let [texture-image (tex-gen/make-texture-image image texture-profile compress?)]
+      (let [texture-generator-result (tex-gen/make-texture-image image texture-profile compress?)
+            texture-protobuf-bytes (TextureUtil/generateResultToByteArray texture-generator-result)]
```
matgis, line 43: It's not really a `texture-protobuf` anymore, right? Maybe `texture-resource-bytes` or `texturec-bytes`?

editor/test/integration/extension_texturepacker_test.clj
```diff
@@ -377,7 +378,7 @@
     (let [node-id (test-util/resource-node project "/examples/basic/basic.tpatlas")]
       (with-open [_ (test-util/build! node-id)]
         (let [built-textureset (->> node-id test-util/node-build-output (protobuf/bytes->map-without-defaults TextureSetProto$TextureSet))
-              built-texture (->> built-textureset :texture (workspace/build-path workspace) fs/read-bytes (protobuf/bytes->map-with-defaults Graphics$TextureImage))]
+              built-texture (->> built-textureset :texture (workspace/build-path workspace) fs/read-bytes TextureUtil/byteArrayToTextureImage (protobuf/bytes->map-with-defaults Graphics$TextureImage))]
```
matgis, line 381: We can use `protobuf/pb->map-with-defaults` here if we change `TextureUtil/byteArrayToTextureImage` to return a `TextureImage` instead of a byte array.

matgis (changes requested):
Looks like cube map code paths need to be updated in the editor. I'm also concerned about peak memory usage during project builds increasing after this change.

editor/test/editor/pipeline/tex_gen_test.clj
```diff
@@ -61,9 +49,31 @@
...
+                                                    false)
+        tex-img (.-textureImage generate-result)]
+    (is (= 3 (.getAlternativesCount tex-img)))
+    (is (= (* 3 32 64) (.. tex-img (getAlternatives 0) (getDataSize))))
+    (is (= (* 4 32 64) (.. tex-img (getAlternatives 1) (getDataSize))))
+    (is (= (* 1 32 64) (.. tex-img (getAlternatives 2) (getDataSize))))))
+
+(deftest make-preview-texture-image-test
+  (let [img     (ImageIO/read (io/resource "test_project/graphics/ball.png"))
+        ^TextureGenerator$GenerateResult
```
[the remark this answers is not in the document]
Jhonnyg: It really isn't clear to me when and why I should add type hints, it _feels_ completely arbitrary sometimes when it makes sense.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TextureUtil.java
```diff
@@ -274,19 +289,65 @@ public static BufferedImage Resize(BufferedImage img, int newW, int newH, int ty
...
+
+        // Create a new array to hold all the data
+        byte[] result = new byte[totalLength];
+
+        // Copy each array into the result array
+        int currentIndex = 0;
+        for (byte[] array : arrays) {
+            System.arraycopy(array, 0, result, currentIndex, array.length);
+            currentIndex += array.length;
+        }
```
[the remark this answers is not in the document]
Jhonnyg: I don't know for certain, but if I can make the change as you suggested it is better than before since we used to copy mipmap data into sequential byte strings for each textureimage. But if we can stream out the byte data it should be better than before.

editor/test/editor/pipeline/tex_gen_test.clj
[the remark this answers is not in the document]
matgis: Yeah, it is a bit of a blemish on the language. There are basically three reasons why you might want to add a type hint.

The most important reason is to get rid of Reflection Warnings. When a particular named Java method has multiple overloads that vary only by argument type, applying a type hint can help it pick the correct one at compile time. If the argument types cannot be inferred, the compiler will output a Reflection Warning when you reload the file in the REPL, or print it to stdout when you run the program. These warnings should be taken seriously, since it means the call will use reflection to find the correct method implementation based on the types at runtime instead of compile time, which can make the call around 50 times slower.

The second reason to type hint is it to avoid boxed math. The JVM has primitive types like `long` and `double`, but all of its collection classes are designed to accept any value that is a `java.lang.Object`. So your `ArrayList<int>` is actually an `ArrayList<Object>` that stores a bunch of boxed `ints` as `java.lang.Integer` objects. In Java, you would get a value from such a collection and assign it to an `int num = arrayList.get(0))`, which would put an `int` on the stack and allow you to do efficient math with it and other primitive numbers. However, in Clojure vars aren't typed by default - everything is of type `java.lang.Object`. That is, unless you add a primitive type hint in the let binding like `(let [^int num (nth list 0)] ...)`. You can also type hint function arguments and return values as `^long` or `^double` to enable something similar to a `fastcall` (but only when a function has four or fewer arguments). You can turn on a compiler warning for boxed math by setting `(set! *unchecked-math* :warn-on-boxed)` at the top of your file. We do this in a bunch of places, but not everywhere since it can be a bit fiddly to solve completely. But it is probably worth addressing in places where we're doing a lot of math.

Finally, type hints are sometimes used for documentation purposes on function arguments and return values that are Java types. It might be nice to see that an argument is expected to be a `^String name`, for example.

The second question is where to put them. I prefer to use type hints on the function arguments, and always type hint the return value of any function that returns a Java type. If I add them in `let` bindings it is either because I want to put a primitive on the stack, or because a function or method I call does not have the correct return type declared. If it is our function, I add a return type hint to the function if possible (because that will benefit every call), but sometimes I need to put the hint at the call site instead. An example of that is stuff like the `nth` function. It has no way of knowing the type of the object it obtains from a collection, but maybe I know that I have a homogenous collection of objects of a particular type.

editor/test/editor/pipeline/tex_gen_test.clj
```diff
@@ -17,42 +17,30 @@
...
-                                                                                             :compression-level :fast}]
-                                                                                  :mipmaps false}]}))
+        ^TextureGenerator$GenerateResult
+        generate-result (tex-gen/make-texture-image img {:name      "test-profile"
+                                                         :platforms [{:os      :os-id-generic
+                                                                      :formats [{:format            :texture-format-rgba
+                                                                                 :compression-level :fast}]
+                                                                      :mipmaps false}]})
+        protobuf-bytes (TextureUtil/generateResultToByteArray generate-result)
+        bytes   (TextureUtil/byteArrayToTextureImage protobuf-bytes)
```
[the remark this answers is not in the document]
Jhonnyg: right, yes that's true!

editor/src/clj/editor/pipeline/tex_gen.clj
```diff
@@ -114,12 +116,21 @@
...
     (make-cubemap-texture-images images preview-profile false)))
+
+(defn build-texture-resource-fn [resource user-data]
+  (let [digest-output-stream
+        (-> resource
+            (io/output-stream)
+            (digest/make-digest-output-stream "SHA-1"))]
+    ;; writeGenerateResultToOutputStream flushes the stream
+    (TextureUtil/writeGenerateResultToOutputStream (:texture-generator-result user-data) digest-output-stream)
+    (digest/completed-stream->hex digest-output-stream)))
```
matgis, line 136: Do we need to flush the stream before calling `completed-stream->hex` here? 🤔

If `writeGenerateResultToOutputStream` flushes it internally, we should probably add a comment here.

editor/src/clj/editor/pipeline.clj
```diff
@@ -151,21 +151,25 @@
       (and (.exists f) (= mtime (.lastModified f)) (= size (.length f))))))
 
 (defn- to-disk! [artifact content-hash]
-  (assert (some? (:content artifact)))
+  (assert (or (some? (:build-fn artifact)) (some? (:content artifact))))
```
matgis, line 154: I love this change!

Although I'm not entirely happy about calling this `:build-fn`, since we also have `:build-fn` in the `build-targets` that means something else entirely.

Maybe something like `:write-content-fn` would be better?

editor/src/clj/editor/pipeline/tex_gen.clj
```diff
@@ -114,12 +116,21 @@
...
         type Graphics$TextureImage$Type/TYPE_CUBEMAP]
     (TextureUtil/createCombinedTextureImage texture-images type)))
 
 (defn make-preview-cubemap-texture-images
-  ^Graphics$TextureImage [images texture-profile]
+  ^TextureGenerator$GenerateResult [images texture-profile]
   (let [preview-profile (make-preview-profile texture-profile)]
     (make-cubemap-texture-images images preview-profile false)))
+
+(defn build-texture-resource-fn [resource user-data]
```
matgis, line 129: Rename to `write-texturec-content-fn`?

matgis (changes requested):
Awesome work! ❤️

I would like to change the name of the `:build-fn` entry in the `artifact` to `:write-content-fn`, so it doesn't get confused with the existing `:build-fn` in the `build-targets`. Also, we might need to flush the `digest-output-stream` before getting the `sha1` from it. Not entirely sure.

matgis (dismissed):
Editor changes look good!

PR #9951, opened 2024-12-18
Add initial dmsdk for WebGPU

Jhonnyg:
Needed for WebGPU support in certain extensions (e.g rive).


JCash (approved):

davebalmer:
Appreciate this a lot. Rive on web is very interesting to me right now. Small file size from its vector animations is very Poki-friendl. It opens up different art styles while keeping the bundle download small.

PR #9977, opened 2024-12-25
Add audio frame count calculation for JS

aglitchman:
The change provides a dynamic frame count calculation for the HTML5 platform audio that scales with the device sample rate. Also it fixes sound issues when running the HTML5 build in the Samsung Internet browser.

Fixes #9445

### Technical changes
* I don't know the formula for calculating frame count and why it defaults to 768 (set by [this commit](https://github.com/defold/defold/commit/f59eb246a40e16d3353c30bcc38b5bb5a28c228e) looong time ago). But I checked the frame count in the new Wasapi (Windows) at 48KHz and the value was 1056. It means that sample rate can be multiplied by 0.022 to get the right value.
* Testing this formula didn't fix the sound on Samsung Internet Browser, at least on my S22 Ultra device.
* I empirically and subjectively set the multiplier to 0.025 to eliminate problems on Samsung Internet. At least on my phone.


JCash (changes requested):
Replace with a user configurable setting. E.g. `sound.frame_count`

aglitchman:
Summary:
- deleted the unused `sound_common.cpp` file, because it duplicates the code that is in `sound.cpp`.
- added a new option `sound.frame_count`, default value 0.
- if the `sound.frame_count` option is not equal to 0, it is used for frame count. If it is 0, then the value from the device is used, i.e. on `wasapi` it is from the system, on HTML5 it is from my “formula”. If the value from the device is 0, then the value 768 is used.

engine/sound/src/devices/device_js.cpp
```diff
@@ -87,6 +87,7 @@ namespace dmDeviceJS
         assert(info);
         JSDevice *dev = (JSDevice*) device;
         info->m_MixRate = dmGetDeviceSampleRate(dev->devId);
+        info->m_FrameCount = info->m_MixRate * 0.025;
```
JCash, line 90: Hmm, well, I wanted to remove this hardcoded constant, which was why I wanted the frame count configuration.
(the property was for developer to use.

And, any runtime manipulation/calculation should more likely happen further "up" in the callstack, not at this level.

So, if you're not going to use the setting, I suggest coming up with a better (documented) calculation in the code.
Sure, it works, but we want as stable solutions as possible (which means we need to calculate a better frame count).

JCash (changes requested):
Now we're stuck with two solutions.

I do prefer a runtime solution, I just don't want a hardcoded value there.

aglitchman:
My initial PR on fixing the sound on Samsung Internet turned in a slightly different direction.

Plus, on HTML5 with the occasional FPS drop, now we still hear issues in the audio. Ideally, it would be great to solve everything at once, i.e. this fps drop problem and the Samsung Internet problem. And it seems to me that this is only possible by dynamically calculating the buffer size for sound frames.

That's why I filled in the issue - https://github.com/defold/defold/issues/10041

So, I think this PR can be closed?

PR #9983, opened 2024-12-29
GPU skinning support

### LINKED ISSUE #3691: GPU Skinning (DEF-1974)
GPU Skinning related code may or may not end up in librig, but we label it there initially

Jhonnyg:
Model components can now support GPU skinning when using materials with a "local" vertex space. There are now two new builtin model materials that can be used (`model_skinned` and `model_skinned_instanced`) that can support either instanced or non-instanced models. 

ScreenRecording2025-03-03at09 54 141-ezgif com-cut (1)

The GPU skinning uses a `floating point texture` to store all the bones of each animation. This means that in order for this to work, `the client must have support for floating point textures`. This can potentially be an issue on WebGL1, but in generally floating point textures are extremely well supported so in most cases this will not be an issue. 

Fixes #3691


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderProgramBuilderEditor.java
```diff
@@ -35,6 +35,8 @@ static public ShaderUtil.Common.GLSLCompileResult buildGLSLVariantTextureArray(S
         shaderDescs.add(module);
 
         ShaderCompilePipeline.Options options = new ShaderCompilePipeline.Options();
+        options.defines.add("EDITOR");
```
Jhonnyg, line 38: Preview shaders are build with the "EDITOR" define, but not the build targets.

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -413,7 +486,14 @@ namespace dmGameSystem
...
 
             ro->m_Textures[i] = texture_res ? texture_res->m_Texture : 0;
+
+            if (ro->m_Textures[i] == 0 && !first_free_index_set)
+            {
+                first_free_index_set = true;
+                first_free_index = i;
+            }
         }
+        return first_free_index;
```
Jhonnyg, line 496: We bind the bone cache texture to the first free index we can find. It's an internal resource (like the font cache), and as such we shouldn't expose it to the materials to the users.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/Exec.java
```diff
@@ -161,6 +174,7 @@ public static Result execResultWithEnvironmentWorkDir(Map<String, String> env, F
     /**
      * Exec command
      * @param env environment variables to use
+     * @param args arguments
```
Jhonnyg, line 177: Missing dosctring

editor/src/clj/editor/material.clj
```diff
@@ -250,11 +250,15 @@ there is no way a user can know what the generated id will be for older shaders.
     :vector-type-mat3 :mat4
     :vector-type-mat4 :mat4))
 
+(def unsupported-semantic-types
+  #{:semantic-type-bone-weights
+    :semantic-type-bone-indices})
```
Jhonnyg, line 255: At some point we might want to have these be configurable, but as an MVP1 it's not necessary.

engine/engine/content/builtins/materials/model_skinned_instanced.vp
```diff
@@ -0,0 +1,45 @@
...
+out mediump vec4 var_light;
+
+uniform vs_uniforms
+{
+    uniform mediump mat4 mtx_view;
+    uniform mediump mat4 mtx_proj;
+    uniform mediump vec4 light;
+};
+
+#include "/builtins/materials/skinning.glsl"
```
Jhonnyg, line 34: I decided to use includes for the shared parts here, both to reduce the code but also to make sure we test incudes more in "production code" apart from the unit tests.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/Exec.java
```diff
@@ -130,6 +130,19 @@ public static Result execResult(String... args) throws IOException {
         return startAndWaitForProcessBuilderResult(pb);
     }
 
+    /**
+     * Exec command
+     * @param args arguments
+     * @return Result instance with return code and stdout/stderr combined
+     * @throws IOException
+     */
+    public static Result execResult(List<String> args) throws IOException {
```
Jhonnyg, line 139: TODO: Remove, we can just convert to array instead (I didn't know this)

engine/rig/src/rig.cpp
```diff
@@ -1118,13 +1170,54 @@ namespace dmRig
...
+
+    void ResetPoseMatrixCache(HRigContext context)
+    {
+        ResetPoseMatrixCache(&context->m_PoseMatrixCache);
+
+        const dmArray<RigInstance*>& instances = context->m_Instances.GetRawObjects();
+        uint32_t n = context->m_Instances.Size();
+        for (uint32_t i = 0; i < n; ++i)
+        {
+            instances[i]->m_PoseMatrixCacheIndex = INVALID_POSE_MATRIX_CACHE_INDEX;
```
Jhonnyg, line 1186: We can probably use memset here

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -114,15 +120,33 @@ namespace dmGameSystem
...
+    };
+
+    struct ModelSkinnedAnimationData
+    {
+        HComponentRenderConstants        m_AnimationRenderConstants;
+        dmGraphics::HTexture             m_BindPoseCacheTexture;
+        // We use an intermediate buffer here to copy the bind poses to before uploading to texture.
+        // This means that we have both a buffer for the matrices in dmRig as well as here.
+        // A potential improvement here would be to just allocate this buffer in comp_model and pass
+        // a pointer to dmRig instead.
```
JCash, line 134: 👍

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1006,16 +1153,18 @@ namespace dmGameSystem
         ro.m_IndexBuffer                                  = render_item->m_Buffers->m_IndexBuffer;              // May be 0
         ro.m_IndexType                                    = render_item->m_Buffers->m_IndexBufferElementType;
         ro.m_InstanceCount                                = instance_count;
-        ro.m_VertexDeclarations[VX_DECL_BASE_BUFFER]      = world->m_VertexDeclaration;
+        ro.m_VertexDeclarations[VX_DECL_BASE_BUFFER]      = vx_decl_base;
```
JCash, line 1156: Haven't noticed this before, but shouldn't the declaration come from the assigned material? 🤔 
Or perhaps I'm missing something?

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1196,12 +1386,44 @@ namespace dmGameSystem
...
+
+                // *4 = 4 vectors per matrix (RGBA)
+                uint32_t cache_offset = 4 * dmRig::GetPoseMatrixCacheDataOffset(world->m_RigContext, component->m_RigInstance);
 
-            if (component->m_RenderConstants)
+                dmVMath::Vector4 animation_data;
+                animation_data.setX((float) cache_offset);
+                animation_data.setY((float) GetBoneCount(component->m_RigInstance));
+                animation_data.setZ(dmGraphics::GetTextureWidth(world->m_SkinnedAnimationData.m_BindPoseCacheTexture));
+                animation_data.setW(dmGraphics::GetTextureHeight(world->m_SkinnedAnimationData.m_BindPoseCacheTexture));
```
JCash, line 1417: No (float) cast needed here?

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1196,12 +1386,44 @@ namespace dmGameSystem
...
+                else
+                {
+                    dmLogOnceError("Unable to bind bone matrix cache texture for component '%s', no free texture slot available.",
+                        dmHashReverseSafe64(dmGameObject::GetIdentifier(component->m_Instance)));
+                }
+
+                constants = constants ? constants : world->m_SkinnedAnimationData.m_AnimationRenderConstants;
+
+                // *4 = 4 vectors per matrix (RGBA)
+                uint32_t cache_offset = 4 * dmRig::GetPoseMatrixCacheDataOffset(world->m_RigContext, component->m_RigInstance);
```
JCash, line 1411: I'm used to only storing 3*4 = 12 floats (as the last column is (0,0,0,1)), to save bandwidth. (doing a transpose to make it into 3 vectors)
Is that something we want to do or perhaps it's not valid these days?
(may still be a mvp2 thing, but not sure when we'll get to that again)

JCash (dismissed):
lgtm

Will you add the example project to a public defold repo and documentation page?

engine/gamesys/src/gamesys/components/comp_model.cpp
Jhonnyg: Well, this is a bit complicated :)
We need a declaration that matches the exact data we pass in, and in this case the data is a fixed format that is pre-baked in the resource. So if we want to use the vertex declaration that the material has "as-is", we need to look at all the streams to see which data streams they want that we have available. This is what we do for custom vertex formats, but for models we try to avoid that route if possible since it requires more vertex buffers to be created.

And for instancing, we need a separate vertex declaration so that we can setup the "step function" for each vertex stream that should be repeated per instance and not per vertex.

In total, the vertex layout for models can potentially be split between three declarations:
* a "default" vertex declaration for all the streams that is part of the fixed data layout (either skinned or non-skinned data)
* a "custom" vertex declaration that holds all the custom attributes
* an "instance" vertex declaration that holds all the instanced streams

engine/gamesys/src/gamesys/components/comp_model.cpp
Jhonnyg: I couldn't get it to work so I'm pausing it for now in favor of merging this PR. I added a note about it in the code for future reference.

PR #9979, opened 2024-12-27
Add `astcenc` for Linux ARM64

aglitchman:
This PR adds `astcenc` package build for `arm64-linux`. Also modifies the build script, which allows you to successfully build a package for a specific commit from astcenc Git repository.

### Technical changes
* Related to https://github.com/defold/defold/issues/9885
* Do not include the PR in the public changelog since the point is to help the team, i.e. no new features.

PR #9984, opened 2024-12-29
GL ES for the Linux ARM64 target to support running on Raspberry Pi and Linux handhelds

### LINKED ISSUE #9826: OpenGL ES as backend for Linux ARM64
The new version 1.9.6 will include experimental support for Linux ARM64. Thank you very much!

I've already started trying to make builds and test on devices and it turns out that Defold expects desktop OpenGL on Linux ARM64. The thing is that ARM as a platform usually includes GPUs designed for mobile platforms. That's why they are limited to OpenGL ES support and newer ones support Vulkan.

For example:
- Raspberry Pi 5: VideoCore VII GPU, supporting OpenGL ES 3.1, Vulkan 1.2 (https://datasheets.raspberrypi.com/rpi5/raspberry-pi-5-product-brief.pdf)
- [RK3588](https://www.rock-chips.com/uploads/pdf/2022.8.26/192/RK3588%20Brief%20Datasheet.pdf) (Mali-G610 MC3, OpenGL ES 1.1, 2.0, 3.2, Vulkan 1.2), [RK3399](https://opensource.rock-chips.com/wiki_RK3399) (Mali-T864 GPU, OpenGL ES1.1/2.0/3.0/3.1, OpenCL, DX11) - popular chips for a single-board computers and handhelnds.
- [Amlogic S905X3](https://www.cnx-software.com/2019/04/12/amlogic-s905x3-specifications-block-diagram/): Arm Mali G31MP2 GPU with support for OpenGL ES 3.2, Vulkan 1.0 and OpenCL 2.0 support.
- [Allwinner A133P](https://www.allwinnertech.com/uploads/pdf/20210803142431e6.pdf): OpenGL ES3.2, Vulkan1.1, OpenCL1.2
- etc

On devices like the M1 and others enthusiasts have added [desktop OpenGL, GL ES, Vulkan support](https://rosenzweig.io/blog/aaa-gaming-on-m1.html), but this is a special case as I understand it.

I suggest a compromise to support the whole range of devices:
- To be able to run the engine on simple SBCs, as stated above, use OpenGL ES, i.e. build the engine with GL ES, not desktop GL.
- Those Defold developers who cannot use GL ES for some reason, let them use Vulkan as the most advanced backend.

Most of the GPUs support Vulkan, but, on the above devices, except RPi, Vulkan drivers have problems and are usually disabled (only judging by community posts), so it's better to use GL ES for them.

And unfortunately, it's impossible to build Defold with simultaneous GL ES and desktop OpenGL support, so we have to make a choice.

* https://github.com/defold/defold/issues/8311

aglitchman:
This PR changes the graphics API from desktop OpenGL to OpenGL ES for the Linux ARM64 target. Devices on this platform usually do not have support for desktop OpenGL because they are built on mobile chips. Also added gamepad mapping for Linux handhelds to the gamepads database.

Fixes #9826

### Technical changes
* I rebuilt the GLFW package for arm64 linux for the reason that the previous version required libGL.so, which is not present on the target system. It was built using clang 16 and the build script was improved accordingly.
* Related (and required to be merged before) https://github.com/defold/defold/pull/9979?new_mergebox=true


engine/platform/src/platform_window_glfw3.cpp
```diff
@@ -142,26 +142,35 @@ namespace dmPlatform
 
     HWindow NewWindow()
     {
+        glfwSetErrorCallback(OnError);
```
aglitchman, line 145: I moved the error callback setup here, since it can be done before `glfwInit`, according to the documentation.

build_tools/waf_dynamo.py
```diff
@@ -1881,8 +1881,10 @@ def detect(conf):
         conf.env['LIB_OPENGL'] = ['EGL', 'GLESv1_CM', 'GLESv2']
     elif platform in ('win32', 'x86_64-win32'):
         conf.env['LINKFLAGS_OPENGL'] = ['opengl32.lib', 'glu32.lib']
-    elif platform in ('x86_64-linux','arm64-linux'):
+    elif platform in ('x86_64-linux'):
         conf.env['LIB_OPENGL'] = ['GL', 'GLU']
+    elif platform in ('arm64-linux'):
+        conf.env['LIB_OPENGL'] = ['EGL', 'GLESv1_CM', 'GLESv2']
```
JCash, line 1887: My main concern is if there will be an actual "desktop" target using this.
For better support in the future, I think the these options should be chosen using an app manifest approach (i.e. choosing the graphics backend: opengl, gles, webgpu etc).

Question is, can we go a head with this, and not make it breaking change if we want to support arm64-linux desktop targets?
I wonder if it's not better to default to OpenGL, and leave the custom backend selection for the app manifest.
(we've also wanted to move to Vulkan as default which just highlights the point)

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderCompilers.java
```diff
@@ -45,14 +45,19 @@ private ArrayList<ShaderDesc.Language> getPlatformShaderLanguages(ShaderDesc.Sha
...
                     } else {
                         shaderLanguages.add(ShaderDesc.Language.LANGUAGE_GLSL_SM140);
                     }
                 } break;
+                case Arm64Linux: {
+                    if (shaderType != ShaderDesc.ShaderType.SHADER_TYPE_COMPUTE) {
+                        shaderLanguages.add(ShaderDesc.Language.LANGUAGE_GLES_SM300);
+                        shaderLanguages.add(ShaderDesc.Language.LANGUAGE_GLES_SM100);
+                    }
+                } break;
```
JCash, line 60: I think this is a bit too limiting.
Using this approach we cannot switch from gles to glsl.

I think we need another way of selecting the shader language. @Jhonnyg do you have any thoughts around this?

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -72,7 +72,7 @@
 #endif
 
 #if defined(__linux__) && !defined(ANDROID)
-    #include <GL/glext.h>
+    // NOP
```
JCash, line 75: If it's not needed, please remove the dead code.

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -1073,7 +1073,7 @@ static void LogFrameBufferError(GLenum status)
         context->m_IsGles3Version = 1; // 0 == gles 2, 1 == gles 3
         context->m_PipelineState  = GetDefaultPipelineState();
 
-#if defined(__EMSCRIPTEN__) || defined(__ANDROID__)
+#if defined(__EMSCRIPTEN__) || defined(__ANDROID__) || (defined(__linux__) && defined(__aarch64__))
```
JCash, line 1076: These switches start to get complicated.
I propose moving this switch to the `wscript`, and adding a `DM_GRAPHICS_USE_OPENGL`/`DM_GRAPHICS_USE_OPENGLES` respectively.

engine/gamesys/wscript
```diff
@@ -40,8 +40,10 @@ def configure(conf):
         pass
     elif re.match('arm.*?android', platform):
         conf.env.append_value('LINKFLAGS', ['-lEGL', '-lGLESv1_CM', '-lGLESv2', '-landroid'])
-    elif platform == "linux" or platform == "x86_64-linux":
+    elif platform == "x86_64-linux":
         conf.env['LIB_X'] = ['Xext', 'X11', 'Xi', 'GL', 'GLU']
+    elif platform == "arm64-linux":
+        conf.env['LIB_X'] = ['Xext', 'X11', 'Xi', 'EGL', 'GLESv1_CM', 'GLESv2']
```
JCash, line 46: Ok for tests, and I know it's old code, but the build artefact should probably use `use = "X OPENGL"` and set those libraries into the LIB_OPENGL variable instead (as you did in the wscript above)

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -72,7 +72,7 @@
 #endif
```
JCash, line 72: Since we are talking about a target that may want to see both OpenGL and OpenGLES, I think there needs to be some design decisions:
* Should we only support OpenGLES?
* How to select between the targets? Compile time or runtime?

We could for instance easily compile two versions of this file (graphics_opengl.cpp` into two libraries `libgraphics_opengl.a` and `libgraphics_opengles.a`, only difference is using different defines (USE_OPENGLES).

Or, do we want to be able to check at runtime?
(I'm not even sure it's even possible, given that they use different headers+libraries)
And it also requires the built project to include two versions of the same shaders, which might not be what one wants.

My gut feeling is that the two-library-approach is the easier route to take.

engine/platform/src/platform_window_glfw3.cpp
```diff
@@ -142,26 +142,35 @@ namespace dmPlatform
...
+        if (glfwInit() == GL_FALSE)
+        {
+            glfwInitHint(GLFW_PLATFORM, GLFW_PLATFORM_WAYLAND);
+            glfwInitHint(GLFW_WAYLAND_LIBDECOR, GLFW_WAYLAND_DISABLE_LIBDECOR);
+            if (glfwInit() == GL_FALSE)
+            {
+                dmLogError("Could not initialize glfw.");
+                return 0;
+            }
+        }
```
JCash, line 162: Is this a fix for something else?
It doesn't seem like it's arm64 related?

engine/platform/src/platform_window_glfw3.cpp
```diff
@@ -181,6 +190,12 @@ namespace dmPlatform
 
     static PlatformResult OpenWindowOpenGL(dmWindow* wnd, const WindowParams& params)
     {
+    #if defined(__linux__) && defined(__aarch64__) && !defined(ANDROID)
+        glfwWindowHint(GLFW_CLIENT_API, GLFW_OPENGL_ES_API);
+        glfwWindowHint(GLFW_CONTEXT_CREATION_API, GLFW_EGL_CONTEXT_API);
+        glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 2);
```
JCash, line 196: Not knowing what the different targets support, but I would think it's not unreasonable to think one of them might support opengles 3?

E.g. the android code does this to detect it:
```
        const char* version = (char *) glGetString(GL_VERSION);
        if (strstr(version, "OpenGL ES 2.") != 0) {
            context->m_IsGles3Version = 0;
        } else {
            context->m_IsGles3Version = 1;
        }
```

Also, I'd expect the gles variables to be set accordingly:
```
context->m_IsGles3Version = ...;
context->m_IsShaderLanguageGles = ...;
```

JCash (changes requested):
While it might work as it is right now, I do wonder if we want to lock down into opengles/handhelds only.

I think my suggested changes gets us 90% of the way to a more modular approach (like we usually do when selecting backends)

Only piece missing is how to change the shader language+version properly.

While I mark this for "change requested", it's also not set in stone, I wanted to mark it as "not done with decisions yet" so we can discuss it further :)

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderCompilers.java
Jhonnyg: Hmm hm, yes I think we need a different approach to this. I kinda feel like you should be able to configure your project to build with both ES and desktop OpenGL support + configure which shader versions to include. If for example you don't care about ES2/WebGL1 you shouldn't be forced to include them. As it stands right now, I don't have a solution for this, we need to design that setup a bit.

engine/graphics/src/opengl/graphics_opengl.cpp
Jhonnyg: Agreed.

engine/graphics/src/opengl/graphics_opengl.cpp
Jhonnyg: I think you can technically link both desktop and ES into the same binary and then load the correct symbols depending on which of the libraries you use. I think there's loaders like GLAD or GLEW that we could use instead of our current setup, but I'd like to take the shortest path for this as possible. And that likely means building two libraries that use different headers and libraries.

aglitchman:
> While it might work as it is right now, I do wonder if we want to lock down into opengles/handhelds only.

> Question is, can we go a head with this, and not make it breaking change if we want to support arm64-linux desktop targets?
> I wonder if it's not better to default to OpenGL, and leave the custom backend selection for the app manifest.
(we've also wanted to move to Vulkan as default which just highlights the point)

I want to discuss this, because I think the decision will make the work more complicated 🙌

I think the Arm64 Linux platform should be kept simple in support, i.e. only GL ES, and for more advanced users let it be Vulkan (which is not there yet, but it's easy to add in the next phase).

Arguments for:
- Arm64 were originally mobile chips and they didn't have desktop OpenGL ([mentioned in the original issue](https://github.com/defold/defold/issues/9826)). At most they got Vulkan and DirectX because of Windows Phone and Windows ARM. And Apple added it for backward compatibility (100% through some layer on Metal).
- Google search tells me that desktop OpenGL on current modern devices is only on: NVIDIA's Jetson and Asahi Linux for the M1/M2. Which have GL ES and Vulkan, too. Why then do these platforms need desktop OpenGL from Defold?
- By supporting Desktop OpenGL and GL ES we will x2 complicate support of this platform. For example, I don't have an arm64 linux device with desktop GL and I won't be able to test it in real life at all.
- OpenGL desktop has been dead for 7 years.

To sum up, I believe that Arm64 Linux platform will have an audience of few Defold users, including me. And I'm sure that hardly anyone here will need desktop OpenGL. And if it is left on by default instead of GL ES, I'm sure that the first question will be: “I built an arm64 linux version, tried to run it on my RPi, but the game doesn't run because there is no libGL.so library. what should I do?”.

JCash:
> I think the Arm64 Linux platform should be kept simple in support

Agreed.
I would however wish to see a `libgraphics_opengles.a` library, with a `GraphicsAdapterOpenGLES`.
Granted, Android doesn't have it, but I think that's a bug imho, as we know GL/GLES work differently.
A lot of our architecture revolve around switching out libraries, and as such, I'd like to make sure the libraries do what they promise. (e.g. "libgraphics_opengl.a" is _not_ an opengles library)

We can still use the new "libgraphics_opengles.a" as the default library, and have a "libgraphics_opengl.a" later on (if someone wants it).

So, to facilitate this, we should build the "graphics_opengl.cpp" into two different versions:
* `libgraphics_opengl.a` - using `#DM_GRAPHICS_USE_OPENGL`
* `libgraphics_opengles.a` - using `#DM_GRAPHICS_USE_OPENGLES`

The waf_dynamo.py should also define the library using `LIB_OPENGLES` just like we do with `LIB_OPENGL`
And the vanilla engine would then use the `GraphicsAdapterOpenGLES` and `OPENGLES` when building arm64-linux.

And, in the shader compiler, we'll still have to hardcode the gles shader, as we can currently only detect `GraphicsAdapterXXX` symbols from the app manifest.

aglitchman:
> I would however wish to see a `libgraphics_opengles.a` library, with a `GraphicsAdapterOpenGLES`.
> Granted, Android uses `libgraphics_opengl.a), but I consider that a bug imho.

Now it's clear! 🙌 Many thanks. I'll update the PR as advised.

yaing11:
@aglitchman Many thanks for your work. It's a great job. I've been waiting for a long time this PR. 

I have two questions.
1. Please tell me, will it work on the **wayland** linux-devices with arm64 ?
2. Do you have success in the building defold game on the RPI ?

aglitchman:
> @aglitchman Many thanks for your work. It's a great job. I've been waiting for a long time this PR.
> 
> I have two questions.
> 
> 1. Please tell me, will it work on the **wayland** linux-devices with arm64 ?
> 2. Do you have success in the building defold game on the RPI ?

🙌🏻

1. Yes
2. I didn't test, but it should run well.

I'll finish the work on the changes this week.

aglitchman:
Not ready for a review yet - I think I've not finished something yet (I'll look tomorrow).

aglitchman (commented):
Summary of changes:
1) Added a split into `GraphicsAdapterOpenGL` and `GraphicsAdapterOpenGLES` adapters for desktop GL and GL ES, using a single source code based on the definition `DM_GRAPHICS_USE_OPENGLES`. For arm64 linux only, so as not to break all other platforms. And for the reason that a full refactoring of GL/GLES into two adapters for all platforms has to be done in a separate PR and by someone from the team, since you need experience and access to CI (i.e. I as a 3rd party contributor without your CI can't even check the correctness of what I'll do).
2) Bob for arm64 linux builds shaders for GL ES, without desktop GL, as right now there will only be a GL ES adapter. I think that's ok.
3) Platform code GLFW3 chooses the proper initialisation path by graphics adapter type.

Only one thing: I didn't find where I should add new definitions for `GraphicsAdapterOpenGLES` for extender and for generating appmanifest from editor. _Update: found it, in app_manifest.clj. Oh, this is the hardest nut to crack so far - I didn't think I'd have to set up an environment for IDE development as well._

engine/graphics/src/opengl/graphics_opengl_defines.h
```diff
@@ -17,10 +17,17 @@
 
 #if defined(__linux__) && !defined(ANDROID)
 
+#if defined(__aarch64__)
```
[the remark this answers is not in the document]
JCash: Hmm, I don't think this header is included from dmSDK?
Do you have a link to that?

engine/tools/wscript
```diff
@@ -45,8 +45,10 @@ def configure(conf):
         conf.env.append_value('LINKFLAGS', ['-framework', 'Cocoa', '-framework', 'OpenGL', '-framework', 'AGL', '-framework', 'IOKit', '-framework', 'Carbon', '-framework', 'CoreVideo', '-framework', 'QuartzCore'])
     elif platform == 'x86_64-ios':
         pass
-    elif platform in ["x86_64-linux", "arm64-linux"]:
+    elif platform == 'x86_64-linux':
         conf.env['LIB_X'] = ['Xext', 'X11', 'Xi', 'GL', 'GLU']
+    elif platform == 'arm64-linux':
+        conf.env['LIB_X'] = ['Xext', 'X11', 'Xi', 'EGL', 'GLESv1_CM', 'GLESv2']
```
JCash, line 51: (old code, I know)
The "X" library shouldn't need these extra GL libraries.
Instead, they should go into "LIB_OPENGLES" (although that is also already setup in waf_dynamo.py)
Our wscripts have tons of small issues like that.

JCash (dismissed):
Lgtm!
Well done!

Jhonnyg (dismissed):
LGTM! Well done 👏

yaing11:
@aglitchman Hello sir. Great job.  Please tell me, when I can to try to build defold for my linux/arm64 device.
Many thanks in advice.

engine/graphics/src/opengl/graphics_opengl_defines.h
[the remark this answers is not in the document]
aglitchman: My fault! Changed to `DM_GRAPHICS_USE_OPENGLES`. Thanks!

engine/tools/wscript
aglitchman: I apologise - I overlooked this change. As with other wscripts, `LIB_X` is no longer needed for Linux, so I removed it for both platforms. And as I understand it, if the engine builds with all the tests after these deletions, then it's ok.

aglitchman:
Finalised the changes taking into account the last two comments from @JCash.

( I'm completely brain-dead from the Github PR interface and I think I'm hitting the wrong buttons - I'll go study their documentation).

Jhonnyg:
@aglitchman if you could fix the conflict we are ready to merge this pr!

aglitchman:
> @aglitchman if you could fix the conflict we are ready to merge this pr!

Give me a couple of hours - I'll merge and check on the device if the gamepad mapping needs to be changed that was written before your fix.

aglitchman:
Done! It can be merged now.

Jhonnyg (approved):

PR #10000, opened 2025-01-07
Remove argument to --tc

Jhonnyg:
The `--texture-compression` argument in bob no longer requires a secondary flag to indicate wether or not texture compression should be applied according to the projects texture profile settings.

Before:
```
java -jar bob.jar --texture-compression=true # texture compression enabled
java -jar bob.jar --texture-compression=false # texture compression disabled
```

After:
```
java -jar bob.jar --texture-compression # texture compression enabled
java -jar bob.jar # texture compression disabled (argument omitted)
```

NOTE: bob is backwards compatible, meaning --texture-compression=false will work for the time being, but will likely be removed at a later time. A warning will be raised about the argument, but the feature will still work.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
@@ -613,7 +613,7 @@ public class Bob {
     private static String[] migrateInvalidArguments(String[] args) {
         ArrayList<String> migratedArguments = new ArrayList<>();
         for (String arg : args) {
-            // Migrate texture compression argument
+            // Migrate texture compression argument. Deprecated in 1.9.7 (https://github.com/defold/defold/pull/10000)
             boolean argIsTextureCompression = arg.startsWith("--tc") || arg.startsWith("--texture-compression");
             if (argIsTextureCompression && arg.contains("=false")) {
                 System.out.println("Warning: --texture-compression and --tc no longer accepts an argument. To disable texture compression, remove the argument.");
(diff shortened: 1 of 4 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -560,7 +560,7 @@ public static List<CommandLineOption> getCommandLineOptions() {
                 opt(null, "with-symbols", ZERO, "Generate the symbol file (if applicable)", false),
 
                 opt("tp", "texture-profiles", ONE, "DEPRECATED! Use --texture-compression instead", true),
-                opt("tc", "texture-compression", ONE, "Use texture compression as specified in texture profiles", true),
+                opt("tc", "texture-compression", ZERO, "Use texture compression as specified in texture profiles", true),
```
JCash, line 563: What happens with user scripts passing an argument to this property? Breaking change?
User facing documentation?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -608,8 +608,20 @@ public static List<CommandLineOption> getCommandLineOptions() {
...
+            // Migrate texture compression argument
+            boolean argIsTextureCompression = arg.startsWith("--tc") || arg.startsWith("--texture-compression");
+            if (argIsTextureCompression && arg.contains("=false")) {
+                System.out.println("Warning: --texture-compression and --tc no longer accepts an argument. To disable texture compression, remove the argument.");
+            } else {
+                migratedArguments.add(arg);
+            }
+        }
+        return migratedArguments.toArray(new String[0]);
     }
```
JCash, line 625: Perhaps add a "DEPRECATED" comment in the code as well, to be able to remove it later when we do a cleanup?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1725,6 +1725,9 @@ private List<TaskResult> doBuild(IProgress monitor, String... commands) throws T
         }
 
         boolean texture_compress = this.option("texture-compression", "false").equals("true");
+
+        System.out.println("Has texture compression? " + texture_compress);
```
JCash: I don't think we should add more "printf" style logging, as this is not required very often.
I suggest looking into the logger that is available in most systems (including Project.java) (e.g. using Level.FINE)

JCash (dismissed):

PR #10003, opened 2025-01-07
Fixing looping issues with Ogg decoders on some streamed data

### LINKED ISSUE #9968: Fix looping .ogg files when streaming.
When looping an .ogg sound while streaming, it "skips" a few frames when it starts over.

thengel01:
EOS detectionn was "off" for Ogg decoders, resulting in in-precise looping of sounds under certain conditions.


JCash (approved):
verified locally
thanks!

PR #10009, opened 2025-01-08
Cache only big files such as textures and fonts

### LINKED ISSUE #9719: Make most resources uncacheable
**Task (REQUIRED):**
Currently, we cache most resources in an external cache system, disabling it only for a few builders:

CleanShot 2024-11-11 at 14 12 30@2x

However, this approach doesn’t make sense because building most resources takes less time than checking their signature and then copying them from the cache.
I propose we disable caching for all builders by default, with some exceptions:
	•	Textures (e.g., atlas, tilemap, cubemap, etc.)
	•	Fonts (especially GlyphBank)
	•	Possibly other exceptions as needed

Caching should enhance speed by focusing on avoiding the building of large, slow tasks while minimizing unnecessary checks and the overhead of copying and pasting smaller tasks.

AGulev:
It's faster to build most small resources than to read them, calculate their hash, and copy/paste them back and forth. However, this rule does not apply to large resources, such as textures and fonts, which take longer to build.  

Fix https://github.com/defold/defold/issues/9719


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/CopyCustomResourcesBuilder.java
```diff
@@ -30,8 +30,7 @@ public Task create(IResource input) throws CompileExceptionError {
         String[] resources = properties.getStringArrayValue("project", "custom_resources", new String[0]);
 
         TaskBuilder b = Task.newBuilder(this)
-                .setName("Copy Custom Resources")
-                .disableCache();
```
AGulev, line 34: Instead of disabling cache for a few resource types we enable it just for a few of them

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/AnimationSetBuilder.java
```diff
@@ -39,7 +39,7 @@
 
 import javax.xml.stream.XMLStreamException;
 
-@BuilderParams(name="AnimationSet", inExts=".animationset", outExt=".animationsetc")
+@BuilderParams(name="AnimationSet", inExts=".animationset", outExt=".animationsetc", isCacheble = true)
```
AGulev, line 42: it is nice to have it in `BuilderParams`

britzl (approved):
Nice improvement!

PR #10018, opened 2025-01-10
Ensure that Bob stops building immediately if any of the build steps fail

### LINKED ISSUE #9695: bob build process fast fail
if an error occurs during engine assembly, the build process continues to pack resources, wasting time on detecting the error and the resources of the assembly machine. 

It would be nice to add a flag to change the behavior and the assembly process ends immediately after the first error.

AGulev:
Ensure that if the engine build fails, Bob stops building resources and displays the error.

Fix https://github.com/defold/defold/issues/9695

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/ConsoleProgress.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/ConsoleProgress.java
@@ -23,6 +23,7 @@ public class ConsoleProgress implements IProgress {
     private float ticks;
     private float scale = 1;
     private Boolean isATTY = false;
+    private boolean canceled = false;
 
     public ConsoleProgress() {
         reportTo = this;
@@ -96,7 +97,14 @@ public class ConsoleProgress implements IProgress {
 
     @Override
     public boolean isCanceled() {
-        return false;
+        if (reportTo != null) {
+            return reportTo.canceled || canceled;
+        }
+        return canceled;
     }
 
+    @Override
+    public void setCanceled(boolean canceled) {
+        this.canceled = canceled;
+    }
 }
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
@@ -1361,7 +1361,16 @@ public class Project {
 
                 long tstart = System.currentTimeMillis();
 
-                buildEngine(monitor, architectures, appmanifestOptions);
+                try {
+                    buildEngine(monitor, architectures, appmanifestOptions);
+                }
+                catch (Exception e) {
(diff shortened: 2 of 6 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1361,7 +1361,16 @@ public Void call() throws Exception {
...
                 long tstart = System.currentTimeMillis();
 
-                buildEngine(monitor, architectures, appmanifestOptions);
+                try {
+                    buildEngine(monitor, architectures, appmanifestOptions);
+                }
+                catch (Exception e) {
+                    if ((e instanceof MultipleCompileException) ||
+                        (e instanceof CompileExceptionError)) {
+                        monitor.setCanceled(true);
```
AGulev, line 1370: If compilation error was caught - flag "cancel build" rised

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1773,32 +1782,40 @@ private List<TaskResult> doBuild(IProgress monitor, String... commands) throws T
...
                     }
-
-                    if (remoteBuildFuture != null) {
-                        // get the result from the remote build and catch
-                        // if an exception was thrown in buildRemoteEngine() the
-                        // original exception is included in the ExecutionException
-                        try {
-                            remoteBuildFuture.get();
+                    catch(Exception e) {
+                        if (!monitor.isCanceled()) {
```
AGulev, line 1792: Do not throw exception if build canceled.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1773,32 +1782,40 @@ private List<TaskResult> doBuild(IProgress monitor, String... commands) throws T
...
+                            resourceBuildingFailed = true;
+                            throw e;
                         }
-                        catch (ExecutionException|InterruptedException e) {
-                            Throwable cause = e.getCause();
-                            if ((cause instanceof MultipleCompileException) ||
-                                (cause instanceof CompileExceptionError)) {
-                                throw cause;
+                    }
+                    finally {
```
AGulev, line 1797: If build canceled and canceled because of engine build fails, we need to throw this exception here

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/ConsoleProgress.java
```diff
@@ -96,7 +97,14 @@ public void done() {
 
     @Override
     public boolean isCanceled() {
```
AGulev, line 99: We already have this checks in many places, I think it makes sense to reuse it to avoid new entities which will do almost the same thing

britzl (approved):
Looks good

PR #9963, opened 2024-12-19
Editor: Consolidate connections from host gui scene to template nodes

matgis:
* Simplified the editor graph connections between host gui scenes and gui nodes imported from referenced templates. This saves a bit of memory in projects with a large number of gui nodes.
* Nodes imported from template scenes that are assigned to a layer will now correctly use the layer configuration from the host scene when viewed in the editor.
* Renaming a gui resource in a referenced scene will no longer erroneously update references that have been redeclared as a different resource in referencing scenes. I.e, if `host.gui` references `template.gui`, and they each declare a font named "header", renaming it inside `template.gui` will not change `host.gui`, because there "header" refers to its the "header" font inside `host.gui`.
* Improved sorting of choice box entries with numbers when referencing gui resources in the Properties View.

Fixes #6356

### Technical changes
* Instead of individual connections for various data originating from the host gui scene, the data is consolidated into two connections, `basic-gui-scene-info` and `costly-gui-scene-info`. The latter contains data whose dependencies are more costly to produce, and is typically only needed when the scene is shown in the Scene View.

This change requires users to update `extension-spine`. Corresponding PR: defold/extension-spine#196.

PR #10024, opened 2025-01-10
Fix excessive memory usage regression introduced by #9963

matgis:
This PR fixes excessive memory usage when building the project in the editor, a regression introduced by #9963. In that PR, we consolidated a bunch of connections into fewer connections of maps of the same data. However, one of the outputs used to produce Gui scene `build-targets` was using the `costly-gui-scene-info` map in error. This caused `gpu-textures` to be produced for each atlas referenced from a Gui. The `gpu-texture` is unnecessary unless you view the Gui scene using the Scene View, so we shouldn't produce them when building the project.


vlaaad (approved):
LGTM!

PR #10032, opened 2025-01-13
Introduce pivot point controller in atlas view

### LINKED ISSUE #9700: Visual marker of sprite pivot point in atlas
Without visualisation, it is not very clear where the sprite anchor point will be, especially when the point is not on the edge of the sprite, or  the sprite in the atlas was rotated when packing. I think visual marker will solve this situation.

#3666 
[forum discuss](https://forum.defold.com/t/teaser-fridays-and-roadmap-talks/2803/232?u=dragosha)

sprocketc:
Introduced a pivot point controller in atlas view. Snapping while moving the pivot point is enabled on `Shift` down.

pivot-point

Fixes #9700

### Technical changes
Implemented a custom tool controller to get around a performance issue on pivot drag. When we update the pivot, atlas image outputs are regenerated. I avoided the transaction on `mouse-move`, and used the tool delta to correct the position of the indicator. The move transaction happens  on `mouse-released` only.


editor/src/clj/editor/atlas.clj
```diff
@@ -1109,3 +1098,159 @@
...
+  (let [scene (g/node-value node-id :scene)
+        rect (-> scene :renderable :user-data :rect)
+        {:keys [geometry width height]} rect
+        rotated (:rotated geometry)
+        [pivot-x pivot-y] (g/node-value node-id :pivot)
+        delta-x (/ (cond-> (.x delta) rotated -) width)
+        delta-y (/ (cond-> (.y delta) (not rotated) -) height)]
+    (if rotated
+      [(- pivot-x delta-y) (+ pivot-y delta-x)]
+      [(+ pivot-x delta-x) (+ pivot-y delta-y)])))
```
sprocketc, line 1120: I decided not to round here eventually. Since this is a normalized value, I am not sure about the maximum required precision.

vlaaad:
I found a strange behavior where the point of view of the camera changes the size of the pivot handle:
https://github.com/user-attachments/assets/203aaa56-9b97-4b82-900a-a97858c0f8e6

editor/src/clj/editor/atlas.clj
```diff
@@ -1109,3 +1098,156 @@
...
+                       (when (< (Math/abs (- % snap-value)) threshold)
+                         snap-value)) snap-values) %) pivot)))
+
+(defn- updated-pivot
+  [node-id ^Vector3d delta]
+  (let [scene (g/node-value node-id :scene)
+        rect (-> scene :renderable :user-data :rect)
+        {:keys [geometry width height]} rect
+        rotated (:rotated geometry)
+        [pivot-x pivot-y] (g/node-value node-id :pivot)
```
vlaaad, line 1115: This function calls `g/node-value`. Because of that, we are not allowed to use it in `produce-` functions — these functions must be pure functions of their arguments. Using `g/node-value` inside them is dangerous because it creates an implicit dependency in code: `produce-renderables` output might change when `:pivot` of this `node-id` changes, but changing the pivot will not invalidate the output of the `produce-renderables` function because there is no explicit connection. 

Instead, we have to connect nodes in the graph so that the necessary data flows through the graph so that you have a `pivot` here as an input.

editor/src/clj/editor/atlas.clj
```diff
@@ -1109,3 +1098,156 @@
...
+  (when (and start-action action)
+    (let [{:keys [world-rotation world-transform]} (peek original-values)
+          manip-origin (math/translation world-transform)
+          manip-rotation (scene-tools/get-manip-rotation manip-space world-rotation)
+          lead-transform (doto (Matrix4d.) (.set manip-origin))
+          [start-pos pos] (map #(let [manip-dir (math/rotate manip-rotation (Vector3d. 0.0 0.0 1.0))
+                                      _ (.transform lead-transform manip-dir)
+                                      _ (.normalize manip-dir)
+                                      manip-pos (math/translation lead-transform)]
+                                  (math/line-plane-intersection (:world-pos %) (:world-dir %) (Point3d. manip-pos) manip-dir)) [start-action action])]
```
vlaaad, line 1148: I also find this `map` hard to follow. Maybe add a linebreak between an anonymous function and a coll? Also, it should probably be `mapv`, we try to avoid laziness...

editor/src/clj/editor/atlas.clj
```diff
@@ -1109,3 +1098,156 @@
...
+          (not= (-> selected-renderables last :node-id g/node-by-id g/node-type) AtlasImage))
+    {}
+    (let [reference-renderable (last selected-renderables)
+          {:keys [node-id world-translation world-rotation]} reference-renderable
+          scale (scene-tools/scale-factor camera viewport world-translation)
+          world-transform (scene-tools/manip-world-transform reference-renderable manip-space scale)
+          pivot-pos (if manip-delta
+                      (cond-> (updated-pivot node-id manip-delta)
+                        snap-enabled (snap-pivot snap-threshold))
+                      (g/node-value node-id :pivot))
```
vlaaad, line 1163: In addition to `updated-pivot`, this line also uses `g/node-value` which is not allowed inside `produce-` functions.

editor/src/clj/editor/atlas.clj
```diff
@@ -1109,3 +1098,156 @@
...
+          [start-pos pos] (map #(let [manip-dir (math/rotate manip-rotation (Vector3d. 0.0 0.0 1.0))
+                                      _ (.transform lead-transform manip-dir)
+                                      _ (.normalize manip-dir)
+                                      manip-pos (math/translation lead-transform)]
+                                  (math/line-plane-intersection (:world-pos %) (:world-dir %) (Point3d. manip-pos) manip-dir)) [start-action action])]
+      (doto (Vector3d.) (.sub pos start-pos)))))
+
+(g/defnk produce-renderables [_node-id manip-space camera viewport selected-renderables manip-delta snap-enabled snap-threshold]
+  (if (or (empty? selected-renderables)
+          (second selected-renderables)
```
vlaaad, line 1153: We could use `(= 1 (coll/bounded-count 2 selected-renderables))` to express that we want only a single element. Or, if `selected-renderables` is guaranteed to be a vector or nil, we could just use `count` instead of bounded count.

editor/src/clj/editor/atlas.clj
```diff
@@ -1109,3 +1098,156 @@
...
+          (not= (-> selected-renderables last :node-id g/node-by-id g/node-type) AtlasImage))
```
vlaaad, line 1154: I worry that `g/node-by-id` is the same as `g/node-value`, meaning that using it should be forbidden from the `produce-` functions. On the other hand, it's guaranteed that the node type of a node does not change. If there is a way to avoid it, I'd prefer us to avoid it...

editor/src/clj/editor/atlas.clj
```diff
@@ -1109,3 +1098,156 @@
...
+  (when-not (empty? original-values)
+    (let [{:keys [world-transform]} (peek original-values)
+          manip-origin (math/translation world-transform)
+          lead-transform (doto (Matrix4d.) (.set manip-origin))
+          manip-pos (math/translation lead-transform)]
+      (scene-tools/scale-factor camera viewport manip-pos))))
+
+(defn handle-input [self action selection-data]
+  (case (:type action)
+    :mouse-pressed (if (first (get selection-data self))
```
vlaaad, line 1180: Maybe here, on mouse pressed, we could inspect the selection, filter the atlas images, and then connect the pivots to the controller?

vlaaad (changes requested):
Looks great and even has snapping, wow!

I found some strange behavior (see the video in the PR conversation) when it comes to the pivot handle presentation. Also, it's very important to not use `g/node-value` inside output functions used in the graph, we should instead connect/disconnect values in the graph as necessary.

matgis:
Please note that we have an automated script that generates release notes from the PR description. I fear the formatting in the current PR description may need to be adjusted before merging! ❤️

matgis:
Regarding `extension-texturepacker`, it would be nice to visualize it’s pivots, but as you concluded may need a different approach.

These pivots should also not be adjustable from within our editor, since they are exported as part of the `.tpinfo` files from the Texture Packer editor.

editor/src/clj/editor/atlas.clj
```diff
@@ -1109,3 +1098,156 @@
...
+                     (let [selected-renderables (g/node-value self :selected-renderables)
+                           original-values (mapv #(select-keys % [:node-id :world-rotation :world-transform]) 
+                                                 selected-renderables)]
+                       (when-not (empty? original-values)
+                         (g/transact
+                            (concat
+                              (g/set-property self :start-action action)
+                              (g/set-property self :action action)
+                              (g/set-property self :original-values original-values)
+                              (g/set-property self :op-seq (gensym)))))
```
matgis, line 1190: Style: Indentation looks messed up?

editor/src/clj/editor/atlas.clj
matgis: We also have `util/only` for this. 🙂

(It could be more efficient internally, but it doesn't matter here.)

editor/src/clj/editor/atlas.clj
```diff
@@ -1109,3 +1098,156 @@
...
+        [pivot-x pivot-y] pivot-pos
+        absolute-pivot-x (* pivot-x (if rotated height width))
+        absolute-pivot-y (* pivot-y (if rotated width height))
+        x (+ x (if rotated (- width absolute-pivot-y) absolute-pivot-x))
+        y (+ y (- height (if rotated absolute-pivot-x absolute-pivot-y))) 
+        position (-> (mapv #(/ % scale) [x y])
+                     (conj 0.0))]
+    (concat
+     (scene-tools/vtx-add position (scene-tools/vtx-scale [7.0 7.0 1.0] (scene-tools/gen-circle 64)))
+     (scene-tools/vtx-add position (scene-tools/gen-point)))))
```
matgis, line 1135: Style: We indent with two spaces.

editor/src/clj/editor/atlas.clj
matgis: You should use either `properties/round-scalar` or `properties/round-scalar-float`, depending on if `pivot-x` is a `double` or a `float`. It's important to maintain the same numeric type as the value was before you transform it, otherwise it can mess with structural equality and hashing of before / after states. We have similar rounding functions for `coarse` values.

Have a look at `editor.scene/apply-move-delta`: https://github.com/defold/defold/blob/dev/editor/src/clj/editor/scene.clj#L1636-L1638

editor/src/clj/editor/atlas.clj
```diff
@@ -1109,3 +1098,156 @@
...
+                                      _ (.normalize manip-dir)
+                                      manip-pos (math/translation lead-transform)]
+                                  (math/line-plane-intersection (:world-pos %) (:world-dir %) (Point3d. manip-pos) manip-dir)) [start-action action])]
+      (doto (Vector3d.) (.sub pos start-pos)))))
+
+(g/defnk produce-renderables [_node-id manip-space camera viewport selected-renderables manip-delta snap-enabled snap-threshold]
+  (if (or (empty? selected-renderables)
+          (second selected-renderables)
+          (not= (-> selected-renderables last :node-id g/node-by-id g/node-type) AtlasImage))
+    {}
```
matgis, line 1155: I'm pretty sure you could return `nil` in this case if you'd rather use `when-some` with `only`.

editor/src/clj/editor/atlas.clj
```diff
@@ -1109,3 +1098,156 @@
...
+  [original-values manip-space start-action action]
+  (when (and start-action action)
+    (let [{:keys [world-rotation world-transform]} (peek original-values)
+          manip-origin (math/translation world-transform)
+          manip-rotation (scene-tools/get-manip-rotation manip-space world-rotation)
+          lead-transform (doto (Matrix4d.) (.set manip-origin))
+          [start-pos pos] (map #(let [manip-dir (math/rotate manip-rotation (Vector3d. 0.0 0.0 1.0))
+                                      _ (.transform lead-transform manip-dir)
+                                      _ (.normalize manip-dir)
+                                      manip-pos (math/translation lead-transform)]
```
matgis, line 1147: These calculation could be moved out of the loop, unless I'm missing something? 🤔

You could also refactor this so you get whatever need from the `start-action` when the drag is initiated, since it will have the same result across the entire drag operation.

editor/src/clj/editor/scene_tools.clj
```diff
@@ -108,7 +108,8 @@
     :scale-xy [0.0 0.0 1.0]
     :scale-xz [0.0 1.0 0.0]
     :scale-yz [1.0 0.0 0.0]
-    :scale-uniform [0.0 0.0 1.0]))
+    :scale-uniform [0.0 0.0 1.0]
+    [0.0 0.0 1.0]))
```
matgis, line 112: Do we really want this default case? It's kind of nice that the `case` expression is exhaustive, and I'm not sure there is a sensible default behavior for this without context? 🤔

editor/src/clj/editor/atlas.clj
```diff
@@ -1109,3 +1098,156 @@
                               node-child-index (.indexOf children node-id)]
                           (< node-child-index (dec (.size children)))))
   (run [selection] (move-node! (selection->image selection) 1)))
+
+(defn- snap-pivot
+  [pivot threshold]
+  (let [snap-values [0.0 0.5 1.0]]
+    (mapv #(or (some (fn [^double snap-value]
+                       (when (< (Math/abs (- % snap-value)) threshold)
```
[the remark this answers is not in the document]
matgis: I would suggest doing this change in a separate PR, since you'll likely have to fix a bunch of places where others neglected this. 😬

matgis (changes requested):
Very nice! Snapping feels great! ❤️

I think we need to round the values. After all, they're `float` values in the file formats, so showing them with all the digits of double-precision is misleading. But the main reason is to avoid changing the value from float to double after the drag, because it messes with the before / after equality test we use for save tracking.

I'm also uncertain about adding the default case to `manip->direction`.

editor/src/clj/editor/atlas.clj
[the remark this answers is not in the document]
sprocketc: I added the check to fix the issues introduced here and then removed it.

editor/src/clj/editor/atlas.clj
sprocketc: The pivot point positions also exist on the `:geometry` (although there is a 0.5 offset), so I didn't have to create any connections. I had to some additional changes to always use the info provided by the selected renderable.

editor/src/clj/editor/scene_tools.clj
sprocketc: I wanted to avoid adding a key from a different ns here, but since we already pass `:move-pivot` to `gen-manip-renderable`, we might as well add an explicit case.

sprocketc:
> I found a strange behavior where the point of view of the camera changes the size of the pivot handle: https://github.com/user-attachments/assets/203aaa56-9b97-4b82-900a-a97858c0f8e6

The scale now depends on the position on the pivot point. That should fix the issue.

matgis:
Don't forget to update the PR description so will format well with our release notes script! ❤️

matgis (approved):
Looks good to me! Nice work! ❤️

editor/src/clj/editor/atlas.clj
```diff
@@ -1109,3 +1098,159 @@
...
+
+  (input active-tool g/Keyword)
+  (input manip-space g/Keyword)
+  (input camera g/Any)
+  (input viewport g/Any)
+  (input selected-renderables g/Any)
+
+  (output scale g/Any :cached produce-scale)
+  (output snap-threshold g/Any :cached (g/fnk [scale] (cond-> 0.1 scale (* ^double scale))))
+  (output snap-enabled g/Bool :cached (g/fnk [start-action action] (and start-action (:shift action))))
```
vlaaad, line 1239: Nitpick: There is no point in caching outputs that are so trivial to produce. There is not need to change it now because `AtlasToolController` is created once per open view, but such details are important for resource nodes that are created per file (meaning, there might be thousands of such nodes).

vlaaad (approved):
Great job! Ship it!

blendjam:
Amazing Work! 👏

PR #10038, opened 2025-01-14
Re-implement bundle dialogs as editor extensions

### LINKED ISSUE #8062: Scriptable Bundle Dialogs

vlaaad:
The bundle dialogs now use newer UI, and are implemented as editor extensions.

In addition to that, we added some new features to editor scripts:
- commands can now use a new location: `"Bundle"` (menu items are added to **Project -> Bundle menu**)
- commands can now define another query parameter in addition to `selection`: `argument`. Currently, command arguments are only passed to the commands in the `"Bundle"` location. There, the argument is `true` if the user asked to configure the bundle options, and `false` otherwise (i.e. when re-bundling).
- commands can now define an `id`: an identifier that may be used to refer to the command by the editor. Currently, it's only used to remember the last used bundle command for re-bundling.
- we added `editor.open_external_file` function
- we added `json` module, with the usual `encode` and `decode` functions

Technical notes:

There are some minor changes to how the new bundling process invokes bob:
- we no longer include `--auth` and `--email` args that were always set by the editor to empty strings. The args are no longer required
- we no longer provide `--defoldsdk` arg, as it's set by bob to the same value that the editor would have provided

Fixes #8062


editor/src/clj/editor/adb.clj
```diff
@@ -1,118 +0,0 @@
...
-;; 
-;; You may obtain a copy of the License, together with FAQs at
-;; https://www.defold.com/license
-;; 
-;; Unless required by applicable law or agreed to in writing, software distributed
-;; under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
-;; CONDITIONS OF ANY KIND, either express or implied. See the License for the
-;; specific language governing permissions and limitations under the License.
-
-(ns editor.adb
```
vlaaad, line 15: Deleted, now the functionality is implemented directly in the bundle editor script.

editor/src/clj/editor/app_view.clj
```diff
@@ -2536,116 +2527,33 @@ If you do not specifically require different script states, consider changing th
...
+                     {:command :bundle
+                      :label (handler/label handler+context)
+                      :user-data {:command command
+                                  :handler+context handler+context}}))))
+             (into [])))))
+  (run [prefs user-data]
+    (let [{:keys [command handler+context]} user-data]
+      (prefs/set! prefs [:bundle :last-bundle-command] (if (handler/synthetic-command? command) nil command))
+      (when (handler/enabled? handler+context)
+        (handler/run handler+context)))))
```
vlaaad, line 2548: The `:bundle` command is now a "meta-command" — its work is presenting and running other commands that live in `:editor.bundle/menu` location.

editor/src/clj/editor/bundle.clj
```diff
@@ -1,38 +0,0 @@
...
-;; 
-;; You may obtain a copy of the License, together with FAQs at
-;; https://www.defold.com/license
-;; 
-;; Unless required by applicable law or agreed to in writing, software distributed
-;; under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
-;; CONDITIONS OF ANY KIND, either express or implied. See the License for the
-;; specific language governing permissions and limitations under the License.
-
-(ns editor.bundle
```
vlaaad, line 15: Deleted: the functionality here moved to the ios bundling section of `bundle.editor_script`.

editor/src/clj/editor/bundle_dialog.clj
```diff
@@ -1,1003 +0,0 @@
...
-;; 
-;; You may obtain a copy of the License, together with FAQs at
-;; https://www.defold.com/license
-;; 
-;; Unless required by applicable law or agreed to in writing, software distributed
-;; under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
-;; CONDITIONS OF ANY KIND, either express or implied. See the License for the
-;; specific language governing permissions and limitations under the License.
-
-(ns editor.bundle-dialog
```
vlaaad, line 15: Deleted: all functionality is re-implemented in `bundle.editor_script`

editor/src/clj/editor/code/view.clj
```diff
@@ -2970,7 +2970,7 @@
    {:command :toggle-breakpoint :label "Toggle Breakpoint"}
    {:command :edit-breakpoint :label "Edit Breakpoint"}])
 
-(handler/register-menu! ::menubar :editor.app-view/view-end
+(handler/register-menu! ::menubar-view :editor.app-view/view-end
```
vlaaad, line 2973: I had to refactor `editor.handler` ns to make built-in and extension commands more similar (e.g. to support command ids). Because of that, id handling was changed, and now you need separate registration ids (first arg to `handler/register-menu!`) for separate things. Previously, the id was scoped to location (`:editor.app-view/view-end` or `:editor.app-view/edit-end`). Now it is scoped to the handler system. Because of that, this ns has to use different registration ids for different locations.

editor/src/clj/editor/editor_extensions.clj
```diff
@@ -368,6 +372,58 @@
...
+       coerce/boolean
+       coerce/number
+       coerce/string
+       (coerce/vector-of % :min-count 1)
+       (coerce/map-of coerce/string %))))
+
+(def ext-json-encode
+  (rt/lua-fn ext-json-decode
+    ([{:keys [rt]} lua-value]
+     (json/write-str (rt/->clj rt json-value-coercer lua-value)))))
```
vlaaad, line 419: For `ios-deploy`, we need to `decode` json. I added `encode` for completeness, even though the editor scripts do not use it.

editor/src/clj/editor/editor_extensions.clj
```diff
@@ -601,7 +665,8 @@
                                 (future/then (reload-resources!) rt/and-refresh-context))
                :out (line-writer #(display-output! :out %))
                :err (line-writer #(display-output! :err %))
-               :env {"editor" {"get" (make-ext-get-fn project)
+               :env {"editor" {"bundle" {"project_binary_name" ext-project-binary-name} ;; undocumented, hidden API!
```
vlaaad, line 668: `editor.bundle` module is internal. It exists in the runtime, but the docs/completions are displayed only when running the editor in dev. This will allow us to change the bundling in the future if we decide to change the bundling UI/UX.

editor/src/clj/editor/editor_extensions/graph.clj
```diff
@@ -121,14 +122,17 @@
...
-      (when-let [to (-> outline-property
-                        properties/edit-type-id
-                        edit-type-id->value-converter
-                        :to)]
-        #(some-> (properties/value outline-property) to)))))
+  (or (when (fn/multi-responds? ext-get node-id property evaluation-context)
+        #(ext-get node-id property evaluation-context))
+      (when (and (= :editor.game-project/GameProjectNode (node-id->type-keyword node-id evaluation-context))
+                 (re-matches #"^[a-zA-Z_][a-zA-Z0-9_]*\.[a-zA-Z_][a-zA-Z0-9_]*$" property))
+        #(game-project/get-setting node-id (string/split property #"\." 2) evaluation-context))
```
vlaaad, line 129: Now we can do e.g. `editor.get("/game.project", "project.title")` — we do this in `bundle.editor_script`

editor/src/clj/editor/editor_extensions/ui_components.clj
```diff
@@ -494,6 +494,7 @@
     {:fx/type ext-with-extra-check-box-props
      :desc (-> {:fx/type fx.check-box/lifecycle
                 :style-class ["check-box" "ext-check-box"]
+                :mnemonic-parsing false
```
vlaaad, line 497: Somewhere it bit me... maybe in a check-box that had a label like `x86_64-linux`

editor/src/clj/editor/handler.clj
```diff
@@ -19,197 +19,264 @@
...
+                  handlers))]
+    (-> state
+        (unregister registration)
+        (update :menus add-menus menus registration)
+        (update :handlers add-handlers handlers registration)
+        (update :registrations assoc registration {:locations (vec (keys menus))
+                                                   :command+contexts (vec (e/mapcat
+                                                                            (fn [{:keys [command contexts]}]
+                                                                              (e/map #(coll/pair command %) contexts))
+                                                                            handlers))}))))
```
vlaaad, line 96: I refactored handler registration so it has more approachable semantics w.r.t. registration identifier, that were previously scoped differently in different places. Now there is a single `registration` id that is used to register both menus and handlers.

editor/src/clj/editor/handler.clj
```diff
@@ -274,14 +341,6 @@
               handler))
           (vals (get-in state [:handlers [command ctx-name]])))))
 
-(defn- get-active [command command-contexts user-data evaluation-context]
-  (let [state @state-atom]
-    (some (fn [ctx]
-            (let [full-ctx (assoc-in ctx [:env :user-data] user-data)]
-              (when-let [handler (get-active-handler state command full-ctx evaluation-context)]
-                [handler full-ctx])))
-          command-contexts)))
```
vlaaad, line 283: inlined in `active`

editor/src/clj/editor/ios_deploy.clj
```diff
@@ -1,123 +0,0 @@
...
-;; 
-;; You may obtain a copy of the License, together with FAQs at
-;; https://www.defold.com/license
-;; 
-;; Unless required by applicable law or agreed to in writing, software distributed
-;; under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
-;; CONDITIONS OF ANY KIND, either express or implied. See the License for the
-;; specific language governing permissions and limitations under the License.
-
-(ns editor.ios-deploy
```
vlaaad, line 15: Deleted: the functionality is now implemented in the ios section of `bundle.editor_script`

editor/resources/prelude.lua
```diff
@@ -99,4 +101,282 @@ editor.ui.resource_field = editor.ui.component(function(props)
...
+        config = editor.ui.show_dialog(dialog_component({config = config, errors = errors}))
+        if not config then
+            error(editor.bundle.abort_message)
+        end
+        editor.prefs.set(prefs_key, config)
+    end
+    return config
+end
+
+function editor.bundle.output_directory(requested_dialog, output_subdir)
```
vlaaad, line 271: The "get output directory" step. Translated from now-deleted `editor.bundle-dialog` ns.

editor/resources/prelude.lua
```diff
@@ -99,4 +101,282 @@ editor.ui.resource_field = editor.ui.component(function(props)
...
+            }))
+            if not overwrite then
+                error(editor.bundle.abort_message)
+            end
+        end
+    end
+    return output_directory
+end
+
+function editor.bundle.create(config, output_directory, extra_bob_opts)
```
vlaaad, line 320: The "invoke bob" step. Mostly corresponds to generic bundle options that are now removed from `editor.pipeline.bob` ns.

editor/src/clj/editor/app_view.clj
```diff
@@ -2536,116 +2527,33 @@ If you do not specifically require different script states, consider changing th
...
+        (->> (handler/realize-menu :editor.bundle/menu)
+             (e/keep
+               (fn [{:keys [command]}]
+                 (when command
+                   (when-let [handler+context (handler/active command contexts true)]
+                     {:command :bundle
+                      :label (handler/label handler+context)
+                      :user-data {:command command
+                                  :handler+context handler+context}}))))
+             (into [])))))
```
matgis, line 2543: Use transducer with `into` instead of creating an eduction?

Or you could make use of the new `coll/transfer` function if you prefer the added clarity.

editor/src/clj/editor/editor_extensions.clj
```diff
@@ -368,6 +372,58 @@
...
+            (throw (LuaError. ^String (ex-message e)))))))))
+
+;; region json
+
+(def json-decode-options-coercer
+  (coerce/hash-map :opt {:all coerce/boolean}))
+
+(def ext-json-decode
+  (letfn [(decode [^LuaString lua-string opts]
+            (with-open [r (-> lua-string .toInputStream io/reader (PushbackReader. 64))]
```
matgis, line 391: Is 64 bytes enough of a buffer?

editor/src/clj/editor/editor_extensions/commands.clj
```diff
@@ -82,53 +82,58 @@
...
-                         (lua-fn->env-fn
-                           (fn [env opts]
-                             (error-handling/try-with-extension-exceptions
-                               :display-output! display-output!
-                               :label (str label "'s \"active\" in " path)
-                               :catch false
-                               (rt/->clj rt coerce/to-boolean (rt/invoke-immediate-1 (:rt state) active (rt/->lua opts) (:evaluation-context env)))))))
+    (cond-> {:contexts contexts
+             :label label
+             :locations locations}
```
matgis, line 111: Style: Missing newline?

editor/test/editor/prefs_test.clj
```diff
@@ -231,6 +231,14 @@
...
              (prefs/get p []))))))
 
+(deftest set?-test
+  (with-schemas {::test {:type :string}}
+    (let [p (prefs/make :scopes {:global (fs/create-temp-file! "is-set-test" "test.editor_settings")}
+                        :schemas [::test])]
+      (is (false? (prefs/set? p [])))
+      (prefs/set! p [] "new-string")
+      (is (true? (prefs/set? p []))))))
+
```
matgis, line 241: Maybe add tests for entries below some path as well?

matgis (changes requested):
Looks good! I think I may have found a place where a preference is read from the old unmigrated location.

I also have some suggestions for moving some Lua helper functions into utility modules, but that can wait.

editor/test/editor/prefs_test.clj
vlaaad: Good idea, I'll add!

editor/resources/bundle.editor_script
```diff
@@ -0,0 +1,587 @@
...
+local linux_bundle_dialog = editor.ui.component(function(props)
+    local config, set_config = editor.ui.use_state(props.config)
+    return editor.bundle.dialog("Bundle Linux Application", config, nil, nil, {
+        editor.bundle.desktop_variant_grid_row(config, set_config),
+        editor.bundle.texture_compression_grid_row(config, set_config),
+        editor.bundle.check_boxes_grid_row(config, set_config)
+    })
+end)
+
+local function bundle_linux(requested_dialog)
```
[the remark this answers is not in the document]
vlaaad: Good idea, I'll rename the parameter

editor/resources/bundle.editor_script
```diff
@@ -0,0 +1,587 @@
...
+            resolved_path = success and result
+        end
+        if resolved_path and editor.external_file_attributes(resolved_path).exists then
+            return resolved_path
+        end
+    end
+    return nil
+end
+
+---@return string cmd ios-deploy command path
```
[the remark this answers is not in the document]
vlaaad: Yeah, I was using them for myself when writing the code, by they are not really needed. I'll remove them.

editor/resources/bundle.editor_script
```diff
@@ -0,0 +1,587 @@
...
+local function get_ios_deploy_command()
+    local prefs_path = editor.prefs.get("tools.ios-deploy-path")
+    if #prefs_path > 0 then
+        if editor.external_file_attributes(prefs_path).exists then
+            return prefs_path
+        else
+            print(("ios-deploy path defined in preferences does not exist: '%s'"):format(prefs_path))
```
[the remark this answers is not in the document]
vlaaad: These end up in the console. Maybe we want to introduce a `log` function that will print to editor logs (in addition to console? or instead of?), but for now I think writing to console with print is good enough.

editor/resources/bundle.editor_script
```diff
@@ -0,0 +1,587 @@
...
+        if config.sign and #identities == 0 then
+            ret.sign = "No code signing identities found on this computer."
+        end
+        if config.sign and #config["code-signing-identity"] == 0 then
+            ret.identity = "Code signing identity must be set to sign the application."
+        end
+        if config.sign then
+            if #config["provisioning-profile"] == 0 then
+                ret.mobileprovisioning = "Provisioning profile must be set to sign the application."
+            elseif config["provisioning-profile"]:sub(-#".mobileprovision") ~= ".mobileprovision" then
```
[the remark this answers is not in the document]
vlaaad: Maybe, though I think that would require a design phase to also add some more goodies to it (like basename, all extensions etc.)

editor/resources/bundle.editor_script
```diff
@@ -0,0 +1,587 @@
...
+            print("Install and launch done.")
+        else
+            print("Install done.")
+        end
+    end
+end
+
+--- ANDROID
+
+
```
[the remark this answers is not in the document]
vlaaad: Good catch!

editor/resources/bundle.editor_script
```diff
@@ -0,0 +1,587 @@
...
+        end
+    end
+end
+
+--- EDITOR EXTENSIONS
+
+function M.get_commands()
+    return {
+        editor.bundle.command("iOS Application...", "bundle-ios", bundle_ios, {
+            active = function() return editor.platform:sub(-#"macos") == "macos" end
```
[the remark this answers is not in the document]
vlaaad: Great idea, I'll do it!

vlaaad:
@matgis I didn't quite understand your concerns about reading prefs from the old unmigrated location. If this is about `build.texture-compression`, we migrated it from the old prefs and it's still in `prefs.clj` built-in schema definition. I removed `bundle` (not `build`) prefs definitions from `prefs.clj` and moved them to `bundle.editor_script` since this is where they are used:
https://github.com/defold/defold/blob/71c54eeebe228f868de55d16a791278bffacc9f0/editor/resources/bundle.editor_script#L533-L540

matgis (approved):
Ship it!

Issue #10001, opened 2025-01-07
Shortcut CTRL + PLUS for increase Font Size just don't work in code editor [Linux]


Dolbaeb777K:
**Describe the bug (REQUIRED)**
Clicking on CTRL+PLUS (CTRL is left or right) for increase Font Size in the Code Editor just don't work

**To Reproduce (REQUIRED)**
1. Go to Code Editor
2. Click on CTRL+PLUS on keyboard
4.  just nothing

**Expected behavior (REQUIRED)**
Increase font size

**Defold version (REQUIRED):**
  - Version 1.9.6

**Platforms (REQUIRED):**
 - Platforms: Linux, KDE Plasma 6.2.5 **Wayland**
 - OS: EndeavourOS
 - Device: Lenovo B590 (on the my main PC it don't work too)

**Video:**

https://github.com/user-attachments/assets/c93e5624-d8ff-433c-9cf2-61bd7f96568d

Dolbaeb777K:
also i have tried to change keyboard layout. Nothing too

AGulev:
related https://github.com/defold/defold/issues/8127

sprocketc:
I reproduced this on my machine (Arch/Wayland). The issue is not reproducible on windows using the same keyboard+layout.

@Dolbaeb777K For now, you might be able to fix this by [setting a custom shortcut](https://defold.com/manuals/editor-keyboard-shortcuts/#customizing-shortcuts). The config depends on the layout and key that you are using. If you want to use your numpad `+` you might have to add `["Ctrl+Add" :zoom-in]`. If you are using a combination like `"Shift+Ctrl+=`, you need to add `["Shift+Ctrl+'='" :zoom-in]`. A custom `keymap.edn` configuration with both should look like this
```
[
 ["Ctrl+Add" :zoom-in]
 ["Shift+Ctrl+'='" :zoom-in]
]
```
You will have to restart the editor after setting a custom keymap.

Dolbaeb777K:
https://github.com/user-attachments/assets/6899afc0-1e46-481c-bf08-a2e88974ea09

thank you! 🙏 )

vlaaad:
In the upcoming 1.10.1 version (beta starts next week), the Defold editor will allow configuring shortcuts in the preferences dialog by pressing a desired configuration (see #10512). Because of that, I think this issue may be closed.

PR #10043, opened 2025-01-15
Enable wheel zoom and add related keyboard shortcuts in code view

### LINKED ISSUE #4842: Add ctrl+mouse wheel to adjust text editor text size
This is a common feature which should be supported.

ctrl+mouse wheel up = zoom in
ctrl+mouse wheel down = zoom out

Ctrl+ and Ctrl- work but Ctrl + mouse wheel should still be supported since it's faster and standard.

sprocketc:
This aims to make code zoom in/out/reset easier by providing additional defaults. It won't solve keyboard-layout specific issues, but having alternative defaults should help. The added shortcuts are also defaults on popular editors like VSCode. We can replace `Ctr+0` with `Ctr+Equals` and `Ctr+'='`, if you think that's a better option.
Related to #10001

Also enabled zoom on Ctrl+Wheel Y events, that resolves #4842.


editor/src/clj/editor/code/view.clj
```diff
@@ -2325,15 +2325,20 @@
...
-                                      (get-property view-node :scroll-y)
-                                      (get-property view-node :layout)
-                                      (get-property view-node :gesture-start)
-                                      (.getDeltaX event)
-                                      (.getDeltaY event)))
-    (hide-suggestions! view-node)))
+  (when (if (or (.isControlDown event) 
+                (.isMetaDown event))
+          (-> (ui/contexts (ui/main-scene))
+              (ui/invoke-handler (cond (pos? (.getDeltaY event)) :zoom-in
```
sprocketc, line 2331: Is there a better way to dispatch the event?

vlaaad: If your action is executed in the context of some JavaFX Node (in this case, the node should be accessible from the event, or using `(g/node-value view-node :canvas)`), you should use this node instead of `(ui/main-scene)` for invoking a command. You probably should use `ui/run-command` here instead.

PR #10046, opened 2025-01-16
Throw a Lua error if `nil` passed into `go.exists()` function

### LINKED ISSUE #9998: go.exists(nil) returns true
## Describe the bug

When calling `go.exists(nil)` I expect the return value to be **false**, currently it's **true**.

## To Reproduce
```lua
print("go.exists(nil) = ", go.exists(nil))
```

## Expected behavior
I expect go.exists(nil) to be false.

## Defold version
Only tested on: 1.9.6

## Platforms
 - Platforms: Win, Web
 - OS: Windows 10
(I presume this is the same on all platforms)

## Workaround
I now always have to write:
```lua
if my_url and go.exists(my_url) then
  ...
```

AGulev:
⚠️ Passing `nil` to `go.exists()` results in a Lua error, as this aligns with expected behavior and helps catch logical issues in the code.

Fixes https://github.com/defold/defold/issues/9998

### DIFF
--- a/engine/gameobject/src/gameobject/gameobject_script.cpp
+++ b/engine/gameobject/src/gameobject/gameobject_script.cpp
@@ -2084,6 +2084,11 @@ namespace dmGameObject
     int Script_Exists(lua_State* L)
     {
         DM_LUA_STACK_CHECK(L, 1);
+        int top = lua_gettop(L);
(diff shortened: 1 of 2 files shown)


koenbollen (approved):
Looks good to me. 👍

britzl (approved):

PR #10048, opened 2025-01-16
Add remapping of SPIR-v resources to dmShaderc

Jhonnyg:
### Technical changes

Pipeline functionality only. Needed so that we can take a produced spirv binary from our shader compilers and modify the bindings, set and locations from the resources. 

For example, lets say we have a vertex and a fragment shader from a project that should be used in a runtime pipeline. These two are being compiled into spirv modules separately with auto-generated locations and bindings, which means that there might be a mismatch between input and output locations for the same variables. Since users don't write these values themselves, we need to be able to remap the resources in the build pipeline to produce correct shaders (this will come in a later PR).


engine/shaderc/src/shaderc_private.h
```diff
@@ -0,0 +1,98 @@
...
+                uint8_t m_Set;
+                uint8_t m_Value;
+            };
+
+            enum Type
+            {
+                TYPE_LOCATION,
+                TYPE_BINDING,
+                TYPE_SET,
+            };
```
JCash, line 68: Are we missing a TYPE_VALUE?
Or is it confusing that they're not 1:1?

engine/shaderc/src/shaderc_spirv.cpp
```diff
@@ -0,0 +1,177 @@
...
+            }
+        }
+        return 0;
+    }
+
+    ShaderCompileResult* CompileSPIRV(HShaderContext context, ShaderCompilerSPIRV* compiler, const ShaderCompilerOptions& options)
+    {
+        uint32_t* source = (uint32_t*) compiler->m_SPIRVCode;
+        uint32_t num_words_total = compiler->m_SPIRVCodeSize / sizeof(uint32_t);
+        uint32_t spirv_header_size_in_words = 5; // 20 bytes
```
JCash, line 136: How is "word" defined here (or rather by spir-v)?

I'm used to the Win32 WORD (16 bits), and DWORD (32 bits).
But I'm not sure there is a C standard for it.

Is there a another name for it?

JCash (dismissed):

engine/shaderc/src/shaderc_private.h
Jhonnyg: Right, yeah I can probably remove the m_Value. I just wanted to set "any of them" using a single value but it doesn't matter.

engine/shaderc/src/shaderc_spirv.cpp
Jhonnyg: Good call, it's taken from here:
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Word

I'll add to the code

PR #10049, opened 2025-01-16
Fix "search already submitted feature requests" link in CONTRIBUTING.md

amirbegan:
The existing URL for the mentioned link contains an underscore instead of a space character resulting in opening a web page that shows zero issues labeled as "feature request" (it is searching for open issues with label "feature_request" instead of "feature request").


britzl (approved):
Thanks!

PR #10052, opened 2025-01-16
Bundling from the editor no longer invalidates build information

AGulev:
Bundling from the editor no longer overrides the build folder, which means that the building process (Ctrl+B) after the bundling process will not take extra time.


vlaaad (approved):
Not sure if we need a slash at the end. Looks good though!

AGulev:
I did it the same way it is in the editor: 
https://github.com/defold/defold/blob/5b82c054f139872b0552dd10f7dd8fb057532a7a/editor/src/clj/editor/workspace.clj#L54-L56

PR #10057, opened 2025-01-17
Fix dataflow issue that caused long sprite save times

### LINKED ISSUE #10050: Saving script in this project takes a lot of time
If you make changes in a file and save it it takes a lot of time to save for some reason

**To Reproduce (REQUIRED)**
1. Open `main/1.script`
2. Change any line
3. Save it
Pay attention how long it takes

it should be saves immidiatly

  - Version 1.9.6

**Platforms (REQUIRED):**
 - Platforms: mac
 - OS: macOS
 - Device: m1 max

[BigAtlases.zip](https://github.com/user-attachments/files/18441175/BigAtlases.zip)

**Logs (OPTIONAL):**

**Workaround (OPTIONAL):**

**Screenshots (OPTIONAL):**

vlaaad:
Previously, saving a sprite caused its atlas to be laid out. We don't do that anymore.

Technical notes:

Initially, I tried to refactor the `texture-binding-infos` output to not require the atlas layout by replacing `anim-data` with `anim-ids`. This was not enough because there is still a dependency on `texture-page-count` that is needed for validation (to check that the number of pages used in the atlas — known only after layout — corresponds to the right page count in the material). So I just used another connection.

Fixes #10050

### DIFF
--- a/editor/src/clj/editor/sprite.clj
+++ b/editor/src/clj/editor/sprite.clj
@@ -356,6 +356,7 @@
                                          (g/error-value? anim-data) (dissoc :anim-data)
                                          (g/error-value? anim-ids) (dissoc :anim-ids)
                                          (g/error-value? texture-page-count) (dissoc :texture-page-count))))
+  (output texture-binding-save-value (g/fnk [sampler texture :as info] info))
   (output scene-info g/Any (g/fnk [sampler gpu-texture anim-data :as info] info))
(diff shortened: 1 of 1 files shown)


vlaaad:
I found an issue with my solution...

matgis (approved):
Works like a charm! Ship it!

I think we may want to cherry-pick this into the `beta` branch as well. 🤔

PR #10061, opened 2025-01-17
Editor profiling helpers

matgis:
### Technical changes
* Restructured the `load-project` dev module to facilitate profiling the various stages of loading the project.
* Updated the `yourkit` dev module to use the new `v2` protocol.
* Changed byte size logging to use human-readable strings (`"1.32 MB"`).
* Added some dev functions to help with profiling.

### DIFF
--- a/editor/src/clj/util/debug_util.clj
+++ b/editor/src/clj/util/debug_util.clj
@@ -25,6 +25,11 @@
 (def ^:private ^:const nanos-per-second 1000000000.0)
 (def ^:private ^:const nanos-per-minute (* nanos-per-second 60.0))
 
+(def ^:private ^:const bytes-per-kilobyte 1024.0)
+(def ^:private ^:const bytes-per-megabyte (* 1024.0 bytes-per-kilobyte))
+(def ^:private ^:const bytes-per-gigabyte (* 1024.0 bytes-per-megabyte))
+(def ^:private ^:const bytes-per-terabyte (* 1024.0 bytes-per-gigabyte))
+
 (defn- decimal-string
   ^String [num ^String unit]
   (String/format Locale/ROOT "%.2f %s" (to-array [num unit])))
@@ -57,6 +62,17 @@
             (if (> 60.0 minutes)
               (DurationFormatUtils/formatDuration ms "m:ss")
               (DurationFormatUtils/formatDuration ms "H:mm:ss"))))))))
+
+(defn bytes->string
+  "Converts a byte size into a human-readable size string."
+  ^String [^long bytes]
+  (let [double-bytes (double bytes)]
+    (condp > double-bytes
+      bytes-per-kilobyte (str bytes " B")
+      bytes-per-megabyte (decimal-string (/ double-bytes bytes-per-kilobyte) "KB")
+      bytes-per-gigabyte (decimal-string (/ double-bytes bytes-per-megabyte) "MB")
+      bytes-per-terabyte (decimal-string (/ double-bytes bytes-per-gigabyte) "GB")
+      (decimal-string (/ double-bytes bytes-per-terabyte) "TB"))))
 
 (defn stack-trace
   "Returns a human-readable stack trace as a vector of strings. Elements are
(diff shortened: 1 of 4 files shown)


editor/src/dev/yourkit.clj
```diff
@@ -71,67 +85,134 @@
...
+  ([^Controller controller]
+   `(.stopTelemetry ~(with-meta controller {:tag `Controller}))))
+
+(defmacro clear-telemetry!
+  "Clears telemetry. Note: when telemetry starts, collected telemetry is cleared
+  automatically, so you do not have to explicitly call this function."
+  ([] `(clear-telemetry! (controller)))
+  ([^Controller controller]
+   `(.resetTelemetry ~(with-meta controller {:tag `Controller}))))
+
```
matgis, line 131: We use macros a lot in this file to minimize function call overhead when profiling.

editor/src/dev/yourkit.clj
```diff
@@ -71,67 +85,134 @@
...
 ;; -----------------------------------------------------------------------------
 ;; CPU profiling
 ;; -----------------------------------------------------------------------------
 
-(def default-cpu-profiling-settings (CpuProfilingSettings.))
+(def ^CpuProfilingSettings default-cpu-profiling-settings (CpuProfilingSettings.))
 
-(definline start-cpu-sampling!
+(defmacro start-cpu-sampling!
```
matgis, line 138: I wanted multiple arities for convenience.

editor/src/dev/yourkit.clj
```diff
@@ -141,42 +222,69 @@
...
+(defmacro capture-garbage-collected-allocation-snapshot!
+  "Forces a garbage collection, then writes the profiling results to a snapshot
+  file on disk. Includes both any recorded CPU profiling results, and recorded
+  memory allocations. Returns the path to the snapshot file."
+  (^String [] (capture-garbage-collected-allocation-snapshot! (controller)))
+  (^String [^Controller controller]
+   `(let [^Controller controller# ~controller]
+      (force-garbage-collection! controller#)
+      (force-garbage-collection! controller#)
+      (capture-allocation-snapshot! controller#))))
```
matgis, line 259: It's intended to call `force-garbage-collection!` twice here. Seems like the first call will sometimes not take care of it all.

editor/src/clj/util/debug_util.clj
```diff
@@ -135,19 +151,49 @@
...
+            end-ns# (System/nanoTime)
+            end-bytes# (allocated-bytes runtime#)
+            allocated-bytes# (- end-bytes# start-bytes#)
+            elapsed-ns# (- end-ns# start-ns#)]
+        (if (pos? allocated-bytes#)
+          (log/info :message ~label
+                    :elapsed (nanos->string elapsed-ns#)
+                    :allocated (bytes->string allocated-bytes#))
+          (log/info :message ~label
+                    :elapsed (nanos->string elapsed-ns#)))
```
matgis, line 188: `log/info` is a variadic macro, so I have to do it like this.

editor/src/clj/util/debug_util.clj
```diff
@@ -58,6 +63,17 @@
               (DurationFormatUtils/formatDuration ms "m:ss")
               (DurationFormatUtils/formatDuration ms "H:mm:ss"))))))))
 
+(defn bytes->string
+  "Converts a byte size into a human-readable size string."
+  ^String [^long bytes]
+  (let [double-bytes (double bytes)]
+    (condp > double-bytes
+      bytes-per-kilobyte (str bytes " B")
+      bytes-per-megabyte (decimal-string (/ double-bytes bytes-per-kilobyte) "KB")
```
vlaaad, line 72: It is very good that you used "KB" instead of "kB" to avoid confusion. Because [people use](https://en.wikipedia.org/wiki/Byte#Multiple-byte_units) e.g. "MB" to mean both `1000*1000 bytes` and `1024*1024 bytes` depending on the context. Which is completely normal and not confusing at all, because there is "KB" vs "kB" to differentiate different things...

vlaaad (approved):
LGTM!

PR #10068, opened 2025-01-19
API doc fix on lua example of timer.trigger

### LINKED ISSUE #10067: API doc: `Timer.trigger` example is broken
The current [`timer.trigger` example](https://defold.com/ref/stable/timer/#timer.trigger:handle) is broken: it's all on one line and without syntax-highlighting.

**To Reproduce (REQUIRED)**
Open `timer.trigger` API doc and check the example code.

The example should be formatted as the other examples

I checked the `alpha`, `beta` and `stable` versions of the documentation.

Andre-LA:
Fix the formatting problem of `timer.trigger` example in API docs.

The reason is that there's one backtick lacking on `timer.trigger` documentation code.

Closes #10067

### DIFF
--- a/engine/script/src/script_timer.cpp
+++ b/engine/script/src/script_timer.cpp
@@ -689,7 +689,7 @@ namespace dmScript
      * if not result then
      *    print("the timer is already cancelled or complete")
      * end
-     * ``
+     * ```
      */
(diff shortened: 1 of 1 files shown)


britzl (approved):
Thanks!

PR #10073, opened 2025-01-20
Allow editing markdown files

### LINKED ISSUE #5301: Create and edit markdown and text files within the Defold editor.

sprocketc:
Allow editing markdown files on text view. The web view should be updated accordingly.

Fixes #5301

### Technical changes
* Clicking on external and internal anchors should work as intended.
* We should avoid unneeded re-rendering.

### DIFF
--- a/editor/src/clj/editor/html_view.clj
+++ b/editor/src/clj/editor/html_view.clj
@@ -243,21 +243,20 @@
   If we somehow manage to browse away from the project to an external
   page, this will somewhat annoyingly bring us back."
   (when resource-node
-    (let [web-engine  (.getEngine web-view)
-          resource    (g/node-value resource-node :resource)
-          project     (project/get-project resource-node)
-          url         (resource-url (get-http-server! project) resource)
-          current-url (.getLocation web-engine)]
-      (if (not= url current-url)
-        (.load web-engine url)
-        (when html (.loadContent web-engine html))))))
+    (let [web-engine (.getEngine web-view)
+          resource   (g/node-value resource-node :resource)
+          project    (project/get-project resource-node)
+          url        (resource-url (get-http-server! project) resource)]
+      (.load web-engine url))))
 
 (defn- handle-location-change! [project view-id new-location]
   (if-let [new-resource-node
-           (if (string/starts-with? new-location (http-server/url (get-http-server! project)))
-             (let [resource-path (subs new-location (count (http-server/url (get-http-server! project))))]
-               (project/get-resource-node project resource-path))
-             nil)]
+           (when (string/starts-with? new-location (http-server/url (get-http-server! project)))
+             (let [resource-path (subs new-location (count (http-server/url (get-http-server! project))))
+                   resource-node (project/get-resource-node project resource-path)
+                   resource (g/node-value resource-node :resource)]
+               (when (resource/openable-in-view-type? resource :html)
+                 resource-node)))]
     (g/transact [(g/connect new-resource-node :html view-id :html)
                  (view/connect-resource-node view-id new-resource-node)])
     (log/warn :message (format "Moving to non-local url or missing resource: %s" new-location))))
@@ -274,8 +273,7 @@
         web-view      (make-web-view project)
         web-engine    (.getEngine web-view)
         view-id       (g/make-node! graph WebViewNode :web-view web-view)
-        repainter     (ui/->timer 1 "update-web-view!" (fn [_ _ _]
-                                                         (g/node-value view-id :load-resource)))]
+        repainter     (ui/->timer 30 "update-web-view!" (fn [_ _ _] (g/node-value view-id :load-resource)))]
 
     (.addListener (.locationProperty web-engine)
                   (ui/change-listener _ _ new-location (handle-location-change! project view-id new-location)))
--- a/editor/src/clj/editor/html_view.clj
+++ b/editor/src/clj/editor/html_view.clj
@@ -235,12 +235,12 @@
     (str (http-server/url http-server)
          (resource/proj-path resource))))
 
-(g/defnk load-resource! [^WebView web-view resource-node _node-id html]
+(g/defnk produce-update-web-view! [^WebView web-view resource-node _node-id html]
   (when resource-node
     (let [web-engine (.getEngine web-view)
-          resource   (g/node-value resource-node :resource)
-          project    (project/get-project resource-node)
-          url        (resource-url (get-http-server! project) resource)]
+          resource (g/node-value resource-node :resource)
+          project (project/get-project resource-node)
+          url (resource-url (get-http-server! project) resource)]
       (.load web-engine url))))
 
 (defn- handle-location-change! [project view-id new-location]
@@ -249,7 +249,7 @@
              (let [resource-path (subs new-location (count (http-server/url (get-http-server! project))))
                    resource-node (project/get-resource-node project resource-path)
                    resource (g/node-value resource-node :resource)]
-               (when (resource/openable-in-view-type? resource :html)
+               (when (resource/has-view-type? resource :html)
                  resource-node)))]
     (g/transact [(g/connect new-resource-node :html view-id :html)
                  (view/connect-resource-node view-id new-resource-node)])
@@ -259,17 +259,17 @@
   (inherits view/WorkbenchView)
   (property web-view WebView)
   (input html g/Str)
-  (output load-resource g/Any :cached load-resource!)) 
+  (output update-web-view g/Any :cached produce-update-web-view!)) 
 
 (defn make-view
   [graph ^Parent parent html-node opts]
-  (let [project       (or (:project opts) (project/get-project html-node))
-        web-view      (make-web-view project)
-        web-engine    (.getEngine web-view)
-        view-id       (g/make-node! graph WebViewNode :web-view web-view)
-        repainter     (ui/->timer 30 "update-web-view!" (fn [_ _ _] 
-                                                          (when (.isSelected (:tab opts))
-                                                            (g/node-value view-id :load-resource))))]
+  (let [project (or (:project opts) (project/get-project html-node))
+        web-view (make-web-view project)
+        web-engine (.getEngine web-view)
+        view-id (g/make-node! graph WebViewNode :web-view web-view)
+        repainter (ui/->timer 30 "update-web-view!" (fn [_ _ _]
+                                                      (when (.isSelected (:tab opts))
+                                                        (g/node-value view-id :update-web-view))))]
 
     (.addListener (.locationProperty web-engine)
                   (ui/change-listener _ _ new-location (handle-location-change! project view-id new-location)))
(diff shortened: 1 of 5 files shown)


sprocketc:
For future reference, removing parts of the code related to http-server and using `loadContent` to inject the html instead of `load` that accepts a url worked for the most part, but I couldn't get the `on-load-succeeded` run properly after re-rendering the content. I reverted the changes because I wanted to preserve this.

editor/src/clj/editor/html_view.clj
```diff
@@ -239,41 +235,45 @@
...
-           (if (string/starts-with? new-location (http-server/url (get-http-server! project)))
-             (let [resource-path (subs new-location (count (http-server/url (get-http-server! project))))]
-               (project/get-resource-node project resource-path))
-             nil)]
-    (g/transact (view/connect-resource-node view-id new-resource-node))
+           (when (string/starts-with? new-location (http-server/url (get-http-server! project)))
+             (let [resource-path (subs new-location (count (http-server/url (get-http-server! project))))
+                   resource-node (project/get-resource-node project resource-path)
+                   resource (g/node-value resource-node :resource)]
+               (when (resource/openable-in-view-type? resource :html)
```
sprocketc, line 258: This fixes and already existing bug. If we try to link to a non html/md asset (like `<a href="game.project">game.project</a>`) we would try to open it on the web viewer.

editor/src/clj/editor/html_view.clj
```diff
@@ -239,41 +235,47 @@
...
 (defn make-view
   [graph ^Parent parent html-node opts]
   (let [project       (or (:project opts) (project/get-project html-node))
         web-view      (make-web-view project)
         web-engine    (.getEngine web-view)
         view-id       (g/make-node! graph WebViewNode :web-view web-view)
-        repainter     (ui/->timer 1 "update-web-view!" (fn [_ _ _] (update-web-view! view-id project)))]
+        repainter     (ui/->timer 30 "update-web-view!" (fn [_ _ _] 
+                                                          (when (.isSelected (:tab opts))
+                                                            (g/node-value view-id :load-resource))))]
```
sprocketc, line 278: Increasing the frames here makes the update more responsive. Since we also check if the tab is selected, that shouldn.t be a problem. Maybe there is a better way to handle this?

editor/test/resources/save_data_project/checked.md
```diff
@@ -0,0 +1,16 @@
...
+### Sub-heading
+Sub-heading paragraph text.
+
+* Unordered
+* List
+
+---
+
+1. Ordered
+2. List
```
sprocketc, line 16: I added this because of  `save_data_test/all-resource-types-covered-test` and fixed the tests under `save_test.clj`. Is there any additional action needed?

editor/src/clj/editor/html_view.clj
```diff
@@ -239,41 +235,47 @@
...
-  (let [web-view ^WebView (g/node-value view-id :web-view)
-        web-engine (.getEngine web-view)]
-    (when-let [resource-node (g/node-value view-id :resource-node)]
-      (load-resource! web-engine project (g/node-value resource-node :resource)))))
+  (when resource-node
+    (let [web-engine (.getEngine web-view)
+          resource   (g/node-value resource-node :resource)
+          project    (project/get-project resource-node)
+          url        (resource-url (get-http-server! project) resource)]
+      (.load web-engine url))))
```
sprocketc, line 250: We always use `load` here, even if it's just the html input that has changed, to ensure that `on-load-succeeded` is going to be triggered.

editor/src/clj/editor/markdown.clj
```diff
@@ -432,34 +431,20 @@
...
 (defn register-resource-types [workspace]
-  (workspace/register-resource-type workspace
+  (r/register-code-resource-type workspace
     :ext "md"
     :label "Markdown"
-    :textual? true
-    :search-fn search-fn
-    :search-value-fn search-value-fn
     :node-type MarkdownNode
     :view-types [:html :text]
```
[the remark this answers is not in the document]
vlaaad: I think we should use `:code` here. The intention is that `:code` is for editable text, and `:text` is for read-only text previews derived from data.

editor/src/clj/editor/resource.clj
```diff
@@ -81,6 +81,10 @@
   (and (openable-resource? value)
        (true? (:editor-openable (resource-type value)))))
 
+(defn openable-in-view-type?
+  [value view-type]
+  (contains? (into #{} (map :id (:view-types (resource-type value)))) view-type))
```
vlaaad, line 86: I think it would be better to use `(some #(= view-type %) (:view-types ...))`: it is still O(n), but without creating an intermediate set that will be immediately thrown away. We also have `coll/some` to reduce GC pressure even more.

editor/test/resources/save_data_project/checked.md
vlaaad: I think that's enough for tests :)

editor/src/clj/editor/html_view.clj
```diff
@@ -239,41 +235,47 @@
...
   This will open the resource at the new location if we've moved it.
 
   If we somehow manage to browse away from the project to an external
   page, this will somewhat annoyingly bring us back."
-  [view-id project]
-  (let [web-view ^WebView (g/node-value view-id :web-view)
-        web-engine (.getEngine web-view)]
-    (when-let [resource-node (g/node-value view-id :resource-node)]
-      (load-resource! web-engine project (g/node-value resource-node :resource)))))
+  (when resource-node
```
vlaaad, line 245: It seems to me that what was previously a docstring (`"Loads the url..."`) is now part of a function body. Maybe remove the string altogether?

vlaaad (changes requested):
LGTM, though I'd prefer if we used `:code` instead of `:text` since there are different intentions behind them.

editor/test/integration/save_test.clj
```diff
@@ -237,25 +237,26 @@
...
-                                            external-lua-lines (code.util/split-lines external-lua-text)]
+                                            external-lua-lines (code.util/split-lines external-lua-text)
+                                            external-markdown-lines (code.util/split-lines external-markdown-text)]
                                         (is (check-eager-loaded-save-data! project "/game_object/test.go" external-game-object-text external-game-object-pb-map))
                                         (is (check-lazy-loaded-save-data! project "/json/test.json" external-json-text))
                                         (is (check-eager-loaded-save-data! project "/script/test_module.lua" external-lua-text external-lua-lines))
-                                        (is (check-lazy-loaded-save-data! project "/markdown/test.md" external-markdown-text))
+                                        (is (check-eager-loaded-save-data! project "/markdown/test.md" external-markdown-text external-markdown-lines))
```
matgis, line 254: Hmm, I think we should still be able to use lazy-loading for the markdown files?

If you add `:lazy-loaded true` to the `register-code-resource-type` call, we won't load the contents of the `.md` file until the user opens it.

Currently only code resources support lazy loading. You can use as long as your node type does not need to establish connections to other nodes as a result of the file contents changing.

(We can't use it for Lua code files, for example, because they connect the referenced resources of any `require` statements as dependent build targets of themselves.)

editor/test/resources/save_data_project/checked.md
matgis: That should do it! ❤️

matgis (changes requested):
Looks good overall! ❤️

In addition to the things Vlad mentioned, I would like us to try to use lazy loading for `.md` files. I believe it should be possible!

editor/src/clj/editor/html_view.clj
```diff
@@ -239,41 +235,47 @@
...
+    (let [web-engine (.getEngine web-view)
+          resource   (g/node-value resource-node :resource)
+          project    (project/get-project resource-node)
+          url        (resource-url (get-http-server! project) resource)]
```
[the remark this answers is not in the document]
sprocketc: Good to hear that! I also dislike column alignment, but didn't want to change the current state.

editor/test/integration/save_test.clj
sprocketc: I tried lazy loading in my initial attempt, but couldn't update the web view. Is this because of the html output? I can give it another shot if you think should work.

editor/src/clj/editor/html_view.clj
```diff
@@ -239,41 +235,47 @@
...
          (resource/proj-path resource))))
 
-(defn- load-resource! [^WebEngine web-engine project resource]
-  (let [url         (resource-url (get-http-server! project) resource)
-        current-url (.getLocation web-engine)]
-    (when (not= url current-url)
-      (.load web-engine url))))
-
-(defn- update-web-view!
+(g/defnk load-resource! [^WebView web-view resource-node _node-id html]
```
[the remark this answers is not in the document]
sprocketc: I updated the name but kept the bang on the produce fn, to have an indication of the side-effect. Does this break your current naming convention?

editor/src/clj/editor/html_view.clj
sprocketc: Oh, I missed that :flushed:. Thanks!

editor/src/clj/editor/markdown.clj
```diff
@@ -433,18 +433,18 @@
...
   (inherits r/CodeEditorResourceNode)
 
-  (output html g/Str :cached (g/fnk [save-value]
-                                    (str "<!DOCTYPE html>"
-                                         "<html><head></head><body>"
-                                         (-> save-value
-                                             data/lines->string
-                                             markdown->html)
-                                         "</body></html>"))))
+  (output html g/Str :cached (g/fnk [arg-list] [save-value]
```
vlaaad, line 436: `[arg-list] [save-value]`? What's going on here?

sprocketc: Me, trying to fix auto-indent in CSCode before committing the changes.. Sorry about that.

vlaaad (dismissed):
LGTM!

editor/test/integration/save_test.clj
matgis: Oh, OK. It's probably fine to eager-load. Maybe we'll look into it later!

matgis (dismissed):
Looks good to me! 👍

matgis (approved):

PR #10084, opened 2025-01-21
Fix issues related to async texture creation

### LINKED ISSUE #10078: Crash on using resource.create_texture_async in Defold 1.9.4
crash on use - `resource.create_texture_async`
NO crash on use - `resource.create_texture` with the same data

**To Reproduce (REQUIRED)**
Project attached

[defold create_texture_async issue.zip](https://github.com/user-attachments/files/18477300/defold.create_texture_async.issue.zip)

  - Version 1.9.4 - 3788157b95959f2d8a6bc763bd8e6eb93f503d2a

**Platforms (REQUIRED):**
 - Platforms: macOS

Run project.

**Logs (OPTIONAL):**

```
ERROR:CRASH: 0   dmengine                            0x0000000103092cc0 _ZN7dmCrash7OnCrashEi + 104
ERROR:CRASH: 1   libsystem_platform.dylib            0x000000018aaaf584 _sigtramp + 56
ERROR:CRASH: 2   libsystem_pthread.dylib             0x000000018aa7ec20 pthread_kill + 288
ERROR:CRASH: 3   libsystem_c.dylib                   0x000000018a98ba30 abort + 180
ERROR:CRASH: 4   libsystem_c.dylib                   0x000000018a98ad20 err + 0
ERROR:CRASH: 5   dmengine                            0x0000000103116f90 _ZN12dmGameSystemL14RenderBoxNodesEPN5dmGui5SceneEPKNS0_11RenderEntryEPKN10Vectormath3Aos7Matrix4EPKfPPKNS0_12StencilScopeEPNS_19CompRenderConstantsEjPNS_16RenderGuiContextE.cold.4 + 0
ERROR:CRASH: 6   dmengine                            0x0000000102e91b90 _ZN12dmGameSystemL14RenderBoxNodesEPN5dmGui5SceneEPKNS0_11RenderEntryEPKN10Vectormath3Aos7Matrix4EPKfPPKNS0_12StencilScopeEPNS_19CompRenderConstantsEjPNS_16RenderGuiContextE + 4484
ERROR:CRASH: 7   dmengine                            0x0000000102e90424 _ZN12dmGameSystemL11RenderNodesEPN5dmGui5SceneEPKNS0_11RenderEntryEPKN10Vectormath3Aos7Matrix4EPKfPPKNS0_12StencilScopeEjPv + 1964
ERROR:CRASH: 8   dmengine                            0x000000010307bf90 _ZN5dmGui11RenderSceneEPNS_5SceneERKNS_17RenderSceneParamsEPv + 2164
ERROR:CRASH: 9   dmengine                            0x0000000102e8d6b0 _ZN12dmGameSystemL13CompGuiRenderERKN12dmGameObject22ComponentsRenderParamsE + 484
ERROR:CRASH: 10  dmengine                            0x0000000102e619c8 _ZN12dmGameObject6RenderEPNS_16CollectionHandleE + 192
ERROR:CRASH: 11  dmengine                            0x0000000102e0eaac _ZN8dmEngine4StepEPNS_6EngineE + 988
ERROR:CRASH: 12  dmengine                            0x0000000102e0f090 _Z14dmEngineUpdatePN8dmEngine6EngineE + 32
ERROR:CRASH: 13  dmengine                            0x0000000102e0fc70 _ZN8dmEngine7RunLoopEPKNS_13RunLoopParamsE + 104
ERROR:CRASH: 14  dmengine                            0x0000000102e0fb2c _Z11engine_mainiPPc + 104
ERROR:CRASH: 15  dyld                                0x000000018a6f60e0 start + 2360
ERROR:CRASH: 
```

Jhonnyg:
Fixed an issue where `resource.create_texture_async` only updated the width and height when setting the texture data, but it should have a valid width and height when _creating_ the texture as well. This would cause GUIs that use dynamic textures to crash, because the "original texture" dimensions were never set. The function can now properly use transcoded texture data as well.

Fixes #10078

### DIFF
--- a/engine/gamesys/src/gamesys/scripts/script_resource.cpp
+++ b/engine/gamesys/src/gamesys/scripts/script_resource.cpp
@@ -259,7 +259,6 @@ struct SetTextureAsyncRequest
     dmBuffer::HBuffer          m_Buffer;
     int32_t                    m_BufferRef;
     HOpaqueHandle              m_Handle;
-    bool                       m_UseRawData;
 };
 
 struct ResourceModule
@@ -701,7 +700,7 @@ static void HandleRequestCompleted(dmGraphics::HTexture texture, void* user_data
         dmScript::DestroyCallback(request->m_CallbackInfo);
     }
 
-    if (request->m_UseRawData)
+    if (request->m_RawData)
     {
         delete request->m_RawData;
     }
@@ -1047,9 +1046,8 @@ static int CreateTextureAsync(lua_State* L)
     CheckCreateTextureResourceParams(L, &create_params);
 
     // We need to create an empty upload buffer if no explicit buffer is passed
-    bool is_transcoded              = create_params.m_CompressionType != dmGraphics::TextureImage::COMPRESSION_TYPE_DEFAULT && dmGraphics::IsFormatTranscoded(create_params.m_CompressionType);
-    bool use_raw_data               = create_params.m_Buffer == 0 || is_transcoded;
-    uint8_t* raw_data               = 0;
+    bool is_transcoded = dmGraphics::IsFormatTranscoded(create_params.m_CompressionType);
+    uint8_t* raw_data  = 0;
 
     if (create_params.m_Buffer == 0)
     {
@@ -1109,7 +1107,6 @@ static int CreateTextureAsync(lua_State* L)
     request->m_Buffer            = create_params.m_Buffer;
     request->m_PathHash          = create_params.m_PathHash;
     request->m_RawData           = raw_data;
-    request->m_UseRawData        = use_raw_data;
 
     dmGraphics::TextureParams texture_params;
     texture_params.m_Width  = create_params.m_Width;
@@ -1152,8 +1149,7 @@ static int CreateTextureAsync(lua_State* L)
         texture_params.m_Data     = decompressed_data;
         texture_params.m_DataSize = decompressed_data_size;
 
-        request->m_RawData    = decompressed_data;
-        request->m_UseRawData = true;
+        request->m_RawData = decompressed_data;
     }
 
     // Execute the upload, the upload buffer should now be locked by this request
--- a/engine/gamesys/src/gamesys/scripts/script_resource.cpp
+++ b/engine/gamesys/src/gamesys/scripts/script_resource.cpp
@@ -255,10 +255,11 @@ struct SetTextureAsyncRequest
     lua_State*                 m_LuaState;
     dmScript::LuaCallbackInfo* m_CallbackInfo;
     TextureResource*           m_TextureResource;
+    uint8_t*                   m_RawData;
     dmBuffer::HBuffer          m_Buffer;
     int32_t                    m_BufferRef;
     HOpaqueHandle              m_Handle;
-    bool                       m_UseUploadBuffer;
+    bool                       m_UseRawData;
 };
 
 struct ResourceModule
@@ -700,12 +701,11 @@ static void HandleRequestCompleted(dmGraphics::HTexture texture, void* user_data
         dmScript::DestroyCallback(request->m_CallbackInfo);
     }
 
-    // If we used a temporary upload buffer, we need to delete it.
-    if (request->m_UseUploadBuffer)
+    if (request->m_UseRawData)
     {
-        dmBuffer::Destroy(request->m_Buffer);
+        delete request->m_RawData;
     }
-    else
+    if (request->m_Buffer)
     {
         dmScript::Unref(request->m_LuaState, LUA_REGISTRYINDEX, request->m_BufferRef);
     }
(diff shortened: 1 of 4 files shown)


engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1046,17 +1046,14 @@ static int CreateTextureAsync(lua_State* L)
...
+    uint8_t* raw_data               = 0;
+
+    if (create_params.m_Buffer == 0)
     {
-        const dmBuffer::StreamDeclaration streams_decl = {dmHashString64("data"), dmBuffer::VALUE_TYPE_UINT8, 1};
-        dmBuffer::Result r = dmBuffer::Create(create_params.m_Width * create_params.m_Height * create_params.m_TextureBpp, &streams_decl, 1, &upload_buffer);
-        if (r != dmBuffer::RESULT_OK)
-        {
-            return DM_LUA_ERROR("Unable to create an empty upload buffer: %s (%d)", dmBuffer::GetResultString(r), r);
-        }
```
Jhonnyg, line 1059: Not sure why we used an upload buffer and not just a raw data pointer 🤔 
I guess in some sense we could use the same code path, but it _feels_ like it was just doing a malloc with extra steps. We don't really need the alignment, streams etc.

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1095,6 +1095,8 @@ static int CreateTextureAsync(lua_State* L)
 
     dmGraphics::TextureCreationParams texture_create_params;
     texture_create_params.m_Type = create_params.m_Type;
+    texture_create_params.m_Width = create_params.m_Width;
+    texture_create_params.m_Height = create_params.m_Height;
```
Jhonnyg, line 1099: Crash fix: If the texture is created without width and height, the "original" width and height was never set, causing GIUs to crash when using it for some things like slice-9. It's not completely clear to me why we need these original width and heights, but it fixes the crash at least.

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1104,25 +1106,56 @@ static int CreateTextureAsync(lua_State* L)
...
+    if (create_params.m_Buffer)
     {
         lua_pushvalue(L, 3);
         request->m_BufferRef = dmScript::Ref(L, LUA_REGISTRYINDEX);
     }
 
+    dmBuffer::GetBytes(request->m_Buffer, (void**) &texture_params.m_Data, &texture_params.m_DataSize);
+
+    // If the data is transcoded, we need an extra pass here to unpack the data before uploading it
+    if (is_transcoded)
```
Jhonnyg, line 1130: This block of code is essentially what the res_texture.cpp does. It would be nice if we could merge the code paths somehow since we eventually want to support installing custom decoders (like webp?), but it can perhaps wait when we do those refactors. Plus, this function is getting complex now :/

engine/gamesys/src/gamesys/test/resource/set_texture.script
```diff
@@ -13,110 +13,110 @@
 -- specific language governing permissions and limitations under the License.
```
Jhonnyg, line 13: We had mixed indentation here, so actually no new changes.

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -255,10 +255,11 @@ struct SetTextureAsyncRequest
     lua_State*                 m_LuaState;
     dmScript::LuaCallbackInfo* m_CallbackInfo;
     TextureResource*           m_TextureResource;
+    uint8_t*                   m_RawData;
     dmBuffer::HBuffer          m_Buffer;
     int32_t                    m_BufferRef;
     HOpaqueHandle              m_Handle;
-    bool                       m_UseUploadBuffer;
+    bool                       m_UseRawData;
```
JCash, line 262: Isn't it enough if the raw data is set?

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1046,17 +1046,14 @@ static int CreateTextureAsync(lua_State* L)
     CreateTextureResourceParams create_params = {};
     CheckCreateTextureResourceParams(L, &create_params);
 
-    // Create an empty upload buffer
-    dmBuffer::HBuffer upload_buffer = create_params.m_Buffer;
-    bool use_upload_buffer = create_params.m_Buffer == 0;
-    if (use_upload_buffer)
+    // We need to create an empty upload buffer if no explicit buffer is passed
+    bool is_transcoded              = create_params.m_CompressionType != dmGraphics::TextureImage::COMPRESSION_TYPE_DEFAULT && dmGraphics::IsFormatTranscoded(create_params.m_CompressionType);
```
JCash, line 1050: Is the "!= dmGraphics::TextureImage::COMPRESSION_TYPE_DEFAULT" needed?
Seems the "IsFormatTranscoded(type)" already should handle this question?

JCash (dismissed):
lgtm.
some comments about redundancy

engine/gamesys/src/gamesys/scripts/script_resource.cpp
AGulev: Isn't it about ownership? Lua vs Resource system or something like this. I don't remember for sure, I just remember we had conversation around it

Jhonnyg: But this buffer will never be placed into lua, it's just an intermediate storage for the texture data during async upload.. The other buffer that can come from lua is handled according to this, we increase the ref while the data is uploaded and then its decref'd when upload is done. But in this case I Don't really see why we use an intermediate HBuffer 🤔

PR #10096, opened 2025-01-22
Allow editing html files

### LINKED ISSUE #8126: Make HTML files editable inside Defold's script editor

sprocketc:
Allow editing html files. The web view should be updated accordingly.

Fixes #8126


editor/src/clj/editor/html_view.clj
```diff
@@ -247,7 +247,7 @@
   (if-let [new-resource-node
            (when (string/starts-with? new-location (http-server/url (get-http-server! project)))
              (let [resource-path (subs new-location (count (http-server/url (get-http-server! project))))
-                   resource-node (project/get-resource-node project resource-path)
+                   resource-node (project/get-resource-node project (URLDecoder/decode resource-path))
```
sprocketc, line 250: Fixes a bug that throws an error when the file name of the asset contains special characters like spaces (we get the url by listening to the location of the web-engine).

editor/src/clj/editor/html_view.clj
```diff
@@ -244,16 +244,17 @@
...
-                   resource (g/node-value resource-node :resource)]
-               (when (resource/has-view-type? resource :html)
-                 resource-node)))]
-    (g/transact [(g/connect new-resource-node :html view-id :html)
-                 (view/connect-resource-node view-id new-resource-node)])
-    (log/warn :message (format "Moving to non-local url or missing resource: %s" new-location))))
+  (let [url (http-server/url (get-http-server! project))]
+    (if-let [new-resource-node
+             (when (string/starts-with? new-location url)
+               (let [resource-path (URLDecoder/decode (subs new-location (count url)))
```
sprocketc, line 250: Decode path to fix opening files that contain spaces.

sprocketc:
It looks like a contributor was working on syntax highlighting on the resolved issue, but was stuck on the editing part. This PR might help pushing this forward.

matgis (approved):
Looks good to me! 👍

vlaaad:
Works great! LGTM!

vlaaad (approved):

PR #10098, opened 2025-01-23
Default to `$XDG_STATE_HOME/Defold` instead of `$HOME/.Defold` for Linux logs

### LINKED ISSUE #10102: Respect `$XDG_STATE_HOME` on Linux
**Task (REQUIRED):**
We should respect the `$XDG_STATE_HOME` env var on Linux and unpack/put logs there

We don't unpack/put logs into the home directory.

menaechmi:
On Linux, the `.Defold` folder was hardcoded to the `$HOME` folder. It now follows the XDG spec and uses `$XDG_STATE_HOME` or `$HOME/.local/state/Defold`.

Fixes #10102

### Technical changes
* It would be better to search for the $XDG_STATE_HOME env first, but that would force current users who also have the env set to do an unpack and split their logs (which I would consider a breaking change)
* Please note line 97 and 99 `Paths.get(userHome, "." + applicationName)` might match if applicationName = "" because `.` refers to the folder itself (i.e. `$HOME/.` can return `$HOME`). 
This shouldn't be an issue, it is only called through `getSupportPath()` which calls with applicationName = "Defold", and it's been present since the file was created, but it is worth noting.
* I haven't had the ability to build Defold yet

### DIFF
--- a/editor/src/java/com/defold/util/SupportPath.java
+++ b/editor/src/java/com/defold/util/SupportPath.java
@@ -92,26 +92,27 @@ public class SupportPath {
         if (userHome == null) {
             return null;
         }
+        final Path home = Paths.get(userHome);
 
         // 1. $HOME/.Defold
-        final Path homeDefold = Paths.get(userHome, "." + applicationName);
-        if (Files.isDirectory(homeDefold)) {
-            return userHome.resolve("." + applicationName);
+        final Path homeDefoldDir = home.resolve("." + applicationName);
+        if (Files.isDirectory(homeDefoldDir)) {
+            return homeDefoldDir;
         }
 
         // 2. $XDG_STATE_HOME
-        final String localStateEnv = System.getenv("XDG_STATE_HOME");
-        if (localStateEnv != null) {
-            Path localState = Paths.get(localStateEnv);
-            if (Files.isDirectory(localState)) {
-                return localState.resolve(applicationName);
+        final String xdgStateEnv = System.getenv("XDG_STATE_HOME");
+        if (xdgStateEnv != null) {
+            Path xdgStateHome = Paths.get(xdgStateEnv);
+            if (Files.isDirectory(xdgStateHome)) {
+                return xdgStateHome.resolve(applicationName);
             }
         }
 
         // 3. $HOME/.local/state
-        final Path home = Paths.get(userHome);
         if (Files.isDirectory(home)) {
-            return home.resolve(".local", "state", applicationName);
+            Path homeLocalState = Paths.get(userHome, ".local", "state");
+            return homeLocalState.resolve(applicationName);
         }
 
         return null;
--- a/editor/src/java/com/defold/util/SupportPath.java
+++ b/editor/src/java/com/defold/util/SupportPath.java
@@ -94,13 +94,7 @@ public class SupportPath {
         }
         final Path home = Paths.get(userHome);
 
-        // 1. $HOME/.Defold
-        final Path homeDefoldDir = home.resolve("." + applicationName);
-        if (Files.isDirectory(homeDefoldDir)) {
-            return homeDefoldDir;
-        }
-
-        // 2. $XDG_STATE_HOME
+        // 1. $XDG_STATE_HOME
         final String xdgStateEnv = System.getenv("XDG_STATE_HOME");
         if (xdgStateEnv != null) {
             Path xdgStateHome = Paths.get(xdgStateEnv);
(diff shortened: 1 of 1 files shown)


vlaaad:
I've read the [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/latest/), and the `XDG_STATE_HOME` environment variable seems like the right place for logs and unpacked files. The editor unpacks files to the support directory on launch if the needed files aren't there, so I would expect things to keep working without any breakage. We should test it though. I'd prefer we use `$XDG_STATE_HOME/Defold`, with a fallback to `~/.local/state/Defold`.

editor/src/java/com/defold/util/SupportPath.java
```diff
@@ -87,13 +87,31 @@ public static Path getWindowsSupportPath(String applicationName) {
...
+        // Ensure we have a user.home set
         final String userHome = System.getProperty("user.home");
         if (userHome == null) {
             return null;
         }
+
+        // 1. $HOME/.Defold
+        final Path homeDefold = Paths.get(userHome, "." + applicationName);
+        if (Files.isDirectory(homeDefold)) {
+            return userHome.resolve("." + applicationName);
```
sprocketc, line 99: `userHome` is a string here, so I think you wanted to use `homeDefold`. If that's the case shouldn't you also remove `, "." + applicationName` from `homeDefold`? homeDefold wouldn't make sense as a name though.

editor/src/java/com/defold/util/SupportPath.java
```diff
@@ -87,13 +87,31 @@ public static Path getWindowsSupportPath(String applicationName) {
...
+            if (Files.isDirectory(localState)) {
+                return localState.resolve(applicationName);
+            }
+        }
+
+        // 3. $HOME/.local/state
         final Path home = Paths.get(userHome);
         if (Files.isDirectory(home)) {
-            return home.resolve("." + applicationName);
+            return home.resolve(".local", "state", applicationName);
```
sprocketc, line 114: Similarly to the previews comment, shouldn't `".local", "state",` be used on `Paths.get` and then resolve the applicationName?

sprocketc (changes requested):

editor/src/java/com/defold/util/SupportPath.java
menaechmi: I was trying to stay consistent by returning a resolve for each, but just returning an already resolved name makes more sense (and actually works). Thank you!

I think `homeDefoldDir` is a better, clearer variable name to show that the directory is `$HOME/.defold`?

menaechmi:
@vlaaad I separated a version that skips checking for `$HOME/.Defold` into its own commit (eaa9b0d) so that if further testing reveals a problem it doesn't need to be merged. 

I tested both in 1.9.5 stable, 1.9.6 stable, and 1.9.7 -beta and -alpha and they both correctly identify the support folder in the correct order. The only thing that seems to cause any real issue is deleting the folder when Defold is already open.

editor/src/java/com/defold/util/SupportPath.java
```diff
@@ -87,16 +87,14 @@ public static Path getWindowsSupportPath(String applicationName) {
     // linux
 
     private static Path getLinuxSupportPath(String applicationName) {
-        final String userHome = System.getProperty("user.home");
-        if (userHome == null) {
-            return null;
```
vlaaad, line 92: This code was unnecessarily defensive: the only use of the function (`Editor.getSupportPath()`) expects the returned value to be non-null. Also, `"user.home"` is mostly guaranteed to be non-null, and we expect it to be set to some valid value throughout the editor.

vlaaad:
@sprocketc please review it on Linux when you have time. (for example, the `Help -> Show Logs` command of the built editor bundle should open the new folder).

editor/src/java/com/defold/util/SupportPath.java
```diff
@@ -87,16 +87,14 @@ public static Path getWindowsSupportPath(String applicationName) {
...
+        final String xdgStateHome = System.getenv("XDG_STATE_HOME");
+        if (xdgStateHome != null) {
+            // 1. $XDG_STATE_HOME
+            return Path.of(xdgStateHome, applicationName);
+        } else {
+            // 2. ~/.local/state
+            return Path.of(System.getProperty("user.home"), ".local", "state", applicationName);
         }
```
sprocketc, line 97: Nit, but I guess the else is not needed now. Feel free to leave this as it is.

```suggestion
        }

        // 2. ~/.local/state
        return Path.of(System.getProperty("user.home"), ".local", "state", applicationName);
```

sprocketc (dismissed):
Seems to work fine. If `XDG_STATE_HOME` is set, it respects the location. If not, it creates the folder under `.local/state`.

It doesn't check if a folder already exists, so it breaks the contract of the original PR. Since this is about logs only, I don't think this is an issue, but the description should be updated before we merge this.

PR #10099, opened 2025-01-23
Disable change tracking for initial project load transactions

matgis:
Improved project load time and peak memory usage when loading large projects in the editor.

### Technical changes
You can now supply `{:track-changes false}` in `opts` to `g/transact` to disable change tracking inside the transaction. Change tracking is used to figure out the set of node outputs that were affected by the changes and need to be evicted from the cache as a result. This is also stored in the undo step so we can evict any affected entries from the cache when we move through the undo history.

Tracking the changes and figuring out the set of dependent outputs is a memory-intensive and lengthy process, so we save quite a bit of peak memory usage and time spent loading the project if we skip it.

When we load the project initially, we will clear the undo history once we're done, so we don't really need to track changes as long as we're careful to manage the system cache manually during the loading process.

### DIFF
--- a/editor/src/clj/dynamo/graph.clj
+++ b/editor/src/clj/dynamo/graph.clj
@@ -200,7 +200,8 @@
          override-id-generator (is/override-id-generator system)
          tx-data-context-map (or (:tx-data-context-map opts) {})
          metrics-collector (:metrics opts)
-         transaction-context (it/new-transaction-context basis id-generators override-id-generator tx-data-context-map metrics-collector)
+         track-changes (:track-changes opts true)
+         transaction-context (it/new-transaction-context basis id-generators override-id-generator tx-data-context-map metrics-collector track-changes)
          tx-result (it/transact* transaction-context txs)]
      (when (and (not (:dry-run opts))
                 (= :ok (:status tx-result)))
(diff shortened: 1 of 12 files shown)


editor/src/clj/dynamo/graph.clj
```diff
@@ -227,16 +228,6 @@
...
-(defn is-added?
-  "Returns a boolean if a node was added as a result of a transaction given a tx-result and node."
-  [tx-result node-id]
-  (contains? (:nodes-added tx-result) node-id))
-
-(defn is-deleted?
-  "Returns a boolean if a node was delete as a result of a transaction given a tx-result and node."
-  [tx-result node-id]
-  (contains? (:nodes-deleted tx-result) node-id))
-
```
matgis, line 239: These will not work anymore. Fortunately, they are both unused.

editor/src/clj/editor/lsp/async.clj
```diff
@@ -105,5 +105,5 @@
   [ec & body]
   `(let [~ec (g/make-evaluation-context)
          ret# (do ~@body)]
-     (ui/run-later (g/update-cache-from-evaluation-context! ~ec))
+     (fx/on-fx-thread (g/update-cache-from-evaluation-context! ~ec))
```
matgis, line 108: Fix tests failing when transferring new cache entries to the main thread from the LSP worker threads.

editor/src/clj/internal/graph.clj
```diff
@@ -824,20 +823,21 @@
...
+              (fn [acc graph-id graph]
+                (assoc! acc graph-id (:successors graph)))
+              (transient {})
+              (:graphs basis)))
+          cache-key (into [endpoints]
+                          (map #(System/identityHashCode (val %)))
+                          graph-id->node-successor-map)]
+      (.get basis-dependencies-cache
+            cache-key
+            (fn [_]
```
matgis, line 840: Indentation change: Return empty set when no endpoints.

vlaaad (approved):
LGTM! Ship it! 

Also, one idea worth investigating is to see if we can improve `mark-input-activated` by only adding the input endpoint in the `:nodes-affected` instead of its direct internal dependencies. This looks like unnecessary work given that internal input dependencies will be covered by walking the successors afterwards anyway.

PR #10108, opened 2025-01-24
Activate tabs on right click

### LINKED ISSUE #9071: [Bug] Editor - Context menu for inactive tab has wrong context
Снимок экрана 2024-06-13 в 16 06 44

Context menu in editor work only for active tab. If click on inactive tab and try to click on "move to other tab pane" or "copy requre path" and etc it call for active tab not for clicked inactive tab

**To Reproduce (REQUIRED)**
1) Open two tabs in editor
2) Open context menu for inactive tab
3) Copy full path for tab (or Close)
4) Path will be for active tab

Context menu should work properly for tab where click was made, not only for active tab. 

  - Version 1.8.1

**Platforms (REQUIRED):**
 - Mac OS Editor

sprocketc:
Activate tabs on right click, to fix the context of the displayed menu.

Fixes #9071

### Technical changes
Tabs are not JavaFX Nodes, so we can't add event handlers to them (mouse-clicks/context-menu-requests). We have to find a way to select the tab through the parent `TabPane`. Using `tab.setContextMenu` could work, but we would have to change the way we register menus, to be able to pass the correct view context.

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -95,7 +95,7 @@
            [javafx.collections ListChangeListener ObservableList]
            [javafx.event Event]
            [javafx.geometry HPos Orientation Pos]
-           [javafx.scene Parent Scene Node]
+           [javafx.scene Parent Scene]
            [javafx.scene.control Label MenuBar SplitPane Tab TabPane TabPane$TabClosingPolicy TabPane$TabDragPolicy Tooltip]
            [javafx.scene.image Image ImageView]
            [javafx.scene.input Clipboard ClipboardContent MouseEvent MouseButton]
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -1876,10 +1876,9 @@ If you do not specifically require different script states, consider changing th
   (defn- handle-tab-pane-mouse-pressed! 
     [^TabPane tab-pane ^MouseEvent event]
     (when (= MouseButton/SECONDARY (.getButton event))
-      (->> (.invoke getTab
-                    (ui/closest-node-where #(instance? TabHeaderSkin %) (.getTarget event))
-                    (into-array Object []))
-           (.select (.getSelectionModel tab-pane))))))
+      (when-let [node (ui/closest-node-where #(instance? TabHeaderSkin %) (.getTarget event))]
+        (->> (.invoke getTab node (into-array Object []))
+             (.select (.getSelectionModel tab-pane)))))))
 
 (defn- configure-editor-tab-pane! [^TabPane tab-pane ^Scene app-scene app-view]
   (.setTabClosingPolicy tab-pane TabPane$TabClosingPolicy/ALL_TABS)
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -1870,19 +1870,21 @@ If you do not specifically require different script states, consider changing th
 (defn- tab->view-type [^Tab tab]
   (some-> tab (ui/user-data ::view-type) :id))
 
+(defn- get-target-tab
+  [^MouseEvent event]
+  (let [TabHeaderSkin (Class/forName "javafx.scene.control.skin.TabPaneSkin$TabHeaderSkin")
+        getTab (.getDeclaredMethod TabHeaderSkin "getTab" (into-array Class []))]
+    (.setAccessible getTab true)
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/app_view.clj
```diff
@@ -1870,6 +1870,20 @@ If you do not specifically require different script states, consider changing th
 (defn- tab->view-type [^Tab tab]
   (some-> tab (ui/user-data ::view-type) :id))
 
+(defn- handle-tab-pane-mouse-pressed!
+  "Selects the tab on right-click."
+  [^TabPane tab-pane ^MouseEvent event]
+  (when (= MouseButton/SECONDARY (.getButton event))
+    ;; Hackish way to get the tab index under the mouse cursor.
+    (when-let [target-tab (->> (.lookupAll tab-pane ".tab")
```
sprocketc, line 1878: Using `(.getTabs tab-pane)` instead of a lookup didn't work, because once again, tabs are not nodes.

sprocketc:
I am not thrilled about this, but I don't think it's terrible, If we just want to activate the tab. Maybe there is a more involved way to do this, without using the event position (like extending the tab class to support those events).

editor/src/clj/editor/app_view.clj
```diff
@@ -1870,6 +1870,20 @@ If you do not specifically require different script states, consider changing th
...
+                               (keep-indexed
+                                (fn [i ^Node node]
+                                  (when (.contains (.getBoundsInParent node)
+                                                   (.getX event) (.getY event))
+                                    i)))
+                               first)]
+      (.select (.getSelectionModel tab-pane) (int target-tab)))))
+
```
vlaaad, line 1886: Hmm, I think there might be a more reliable way to get the pressed tab — by walking the parent chain of the event target, i.e.:
```clj
(let [TabHeaderSkin (Class/forName "javafx.scene.control.skin.TabPaneSkin$TabHeaderSkin")
      getTab (.getDeclaredMethod TabHeaderSkin "getTab" (into-array Class []))]
  (.setAccessible getTab true)
  (.invoke getTab
           (ui/closest-node-where #(instance? TabHeaderSkin %) (.getTarget event))
           (into-array Object [])))
```
Unfortunately, it requires reflection. Still, I think it's preferred over walking the child tree of the tab pane. Realistically, the JavaFX dev team is not going to rewrite TreeView, so it's very unlikely this class changes. Performance-wise, I think it does not matter if we walk a parent chain of a target node and make a single reflective call vs. walking all children of a tab pane and making bounds checks on some of them. I just think that since we know the pressed node, it's more reliable to go from there.

vlaaad (commented):
Maybe we should do it a bit differently, what do you think?

editor/src/clj/editor/app_view.clj
sprocketc: Avoiding the lookup and the bounds check is clearly a better option, so I applied your suggestion.

editor/src/clj/editor/app_view.clj
```diff
@@ -1870,6 +1870,22 @@ If you do not specifically require different script states, consider changing th
 (defn- tab->view-type [^Tab tab]
   (some-> tab (ui/user-data ::view-type) :id))
 
+(defn- get-target-tab
+  [^MouseEvent event]
+  (let [TabHeaderSkin (Class/forName "javafx.scene.control.skin.TabPaneSkin$TabHeaderSkin")
+        getTab (.getDeclaredMethod TabHeaderSkin "getTab" (into-array Class []))]
+    (.setAccessible getTab true)
```
vlaaad, line 1877: I think we should not do class/method reflection on every call to `get-target-tab`. Instead, we should make it a top-level statement, like here: https://github.com/defold/defold/blob/4e8483c3143df92f66ae6086d4d2cfb0e843c8a7/editor/src/clj/editor/lsp.clj#L599-L604
This way, it's both more performant in the runtime, and if a dependency update breaks our use of reflection, we will know about it immediately when the app is started.

vlaaad (changes requested):

vlaaad (approved):
Works great, ship it!

vlaaad (changes requested):
Actually, I just found a bug in the code that I suggested :D 

You need to check that `(ui/closest-node-where #(instance? TabHeaderSkin %) (.getTarget event))` is not nil. If you click on empty space near tabs, you get NPE.

vlaaad (approved):
LGTM!

PR #10109, opened 2025-01-24
Migrate builtin shaders to the new shader pipeline

### LINKED ISSUE #10107: Migrate builtin shaders to use the new shader pipeline
Notes:

* All shaders will be converted to `#version 140`
* Precision for ES will change slightly:
  Vertex shaders have default highp precision. All lowp will be mediump (spirv-cross ignores lowp)
  Fragment shaders have default float precision set to mediump and int to highp. Setting precision highp float; at the top will be ignored. We will have to add a flag for this. All lowp declarations will be converted into mediump
* Uniforms will be turned into a single uniform buffer that will be named based on the stage they are in, i.e vs_uniforms and fs_uniforms
* gl_FragColor doesn't exist anymore, it will be called out_FragColor to converge with what we call varyings

Jhonnyg:
The builtin shaders are now written in `#version 140` glsl, which means that they will now be compiled to the target shader language with the new shader compile pipeline. To read more about the new pipeline and how to migrate existing shaders, head over to https://defold.com/manuals/shader/#writing-modern-glsl-shaders

Fixes #10107

### DIFF
--- a/engine/engine/content/builtins/materials/sprite_paged_atlas.fp
+++ b/engine/engine/content/builtins/materials/sprite_paged_atlas.fp
@@ -3,7 +3,7 @@
 in mediump vec2 var_texcoord0;
 in mediump float var_page_index;
 
-out vec4 out_FragColor;
+out vec4 out_fragColor;
 
 uniform mediump sampler2DArray texture_sampler;
 
@@ -16,5 +16,5 @@ void main()
 {
     // Pre-multiply alpha since all runtime textures already are
     mediump vec4 tint_pm = vec4(tint.xyz * tint.w, tint.w);
-    out_fragColor      = texture2DArray(texture_sampler, vec3(var_texcoord0.xy, var_page_index)) * tint_pm;
+    out_fragColor = texture(texture_sampler, vec3(var_texcoord0.xy, var_page_index)) * tint_pm;
 }
--- a/engine/engine/content/builtins/materials/sprite_paged_atlas.fp
+++ b/engine/engine/content/builtins/materials/sprite_paged_atlas.fp
@@ -1,12 +1,20 @@
-varying mediump vec2 var_texcoord0;
-varying lowp float   var_page_index;
+#version 140
 
-uniform lowp sampler2DArray texture_sampler;
-uniform lowp vec4           tint;
+in mediump vec2 var_texcoord0;
+in mediump float var_page_index;
+
+out vec4 out_FragColor;
+
+uniform mediump sampler2DArray texture_sampler;
+
+uniform fs_uniforms
+{
+    mediump vec4 tint;
+};
 
 void main()
 {
     // Pre-multiply alpha since all runtime textures already are
-    lowp vec4 tint_pm = vec4(tint.xyz * tint.w, tint.w);
-    gl_FragColor      = texture2DArray(texture_sampler, vec3(var_texcoord0.xy, var_page_index)) * tint_pm;
+    mediump vec4 tint_pm = vec4(tint.xyz * tint.w, tint.w);
+    out_fragColor      = texture2DArray(texture_sampler, vec3(var_texcoord0.xy, var_page_index)) * tint_pm;
 }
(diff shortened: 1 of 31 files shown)


aglitchman:
Great!

I have a request - in the version of the engine that will do this, extend the beta test time by +1 week. Because quite a few libraries and games (and ours including) made materials like that: vertex shader was `builtins`, and fragment was new. Accordingly, the materials will break and it will take a long time to update. Oh...

Jhonnyg:
@aglitchman hm, I'm not so sure they will break. They should be 1:1 compatible with your existing materials 🤔

aglitchman:
> @aglitchman hm, I'm not so sure they will break. They should be 1:1 compatible with your existing materials 🤔

You mean we can use the vertex shader for the new GLSL 140 version and the fragment shader for the old GLSL version in the same material? I was sure it was mandatory to upgrade both shaders to the new one.

Jhonnyg:
@aglitchman no that should be fine, you can mix and match versions (although it's recommended to upgrade)

engine/shaderc/src/shaderc_spvc.cpp
```diff
@@ -440,6 +440,11 @@ namespace dmShaderc
             spvc_compiler_options_set_uint(spv_options, SPVC_COMPILER_OPTION_GLSL_VERSION,                  options.m_Version);
             spvc_compiler_options_set_bool(spv_options, SPVC_COMPILER_OPTION_GLSL_ES,                       options.m_GlslEs);
             spvc_compiler_options_set_bool(spv_options, SPVC_COMPILER_OPTION_GLSL_ENABLE_420PACK_EXTENSION, !options.m_No420PackExtension);
+
+            // TODO:
+            // SPVC_COMPILER_OPTION_GLSL_ES_DEFAULT_FLOAT_PRECISION_HIGHP
+            // SPVC_COMPILER_OPTION_GLSL_ES_DEFAULT_INT_PRECISION_HIGHP
```
Jhonnyg, line 446: This should be configurable!

engine/engine/content/builtins/materials/sprite_paged_atlas.fp
```diff
@@ -1,12 +1,20 @@
...
+};
 
 void main()
 {
     // Pre-multiply alpha since all runtime textures already are
-    lowp vec4 tint_pm = vec4(tint.xyz * tint.w, tint.w);
-    gl_FragColor      = texture2DArray(texture_sampler, vec3(var_texcoord0.xy, var_page_index)) * tint_pm;
+    mediump vec4 tint_pm = vec4(tint.xyz * tint.w, tint.w);
+    out_fragColor = texture(texture_sampler, vec3(var_texcoord0.xy, var_page_index)) * tint_pm;
 }
```
AGulev, line 20: Artem asked it, I want to ask again to make sure:
It's pretty frequent to have a material where `vp` shader from `builtins`, but `fp` one is custom.
After this change, such materials will have `vp` shader version 140 and `fp` stay old. 
Will it work? If we expect it to work, maybe it make sense to have test for it?

AGulev (approved):
Looks good, just a small question

engine/shaderc/src/shaderc_spvc.cpp
Jhonnyg: Yep, that would be a good place to put it. Not sure if we should do it in this PR or not 🤔

engine/engine/content/builtins/materials/sprite_paged_atlas.fp
Jhonnyg: My other PR has such tests, but both pipelines we will produce valid shaders so I know for a fact this will work :) But in any case since we don't run an actual graphics API for tests I can't really write a unit test for it. Perhaps if we can run metal or something on dedicated hardware? 🤔

AGulev: For me, it's enough if you are sure it works :)
I agree, it's not that easy to test.

JCash (approved):

PR #10112, opened 2025-01-27
Fix arrow key movement of GUI elements on non default layouts

### LINKED ISSUE #10077: When using gui layouts, the arrow keys move the default node instead of the one in the active layout
When trying to move gui nodes with arrow keys and the layouts are in use (unsure if this happens with templates too), the selected node does not move if it is not 1:1 with the default node (blue node name). If moving a gui node while non-default layout selected and the node is not blue, then the arrow keys move the default node instead of the selected layout one.

Everything seems to work as intended when using mouse drag to move the nodes.

**To Reproduce (REQUIRED)**
1. Open the repro project and open main.gui.
2. Select Landscape layout and try to arrow key move "box". Nothing seems to move.
3. Now try to move "box1" with arrow keys (still on the Landscape layout) and that seems to move the node in the default layout.
4. Change active layout to Portrait and try to arrow key move either of the box nodes. They move, but so does the default layout nodes.

The selected gui node move in their selected layout.

  - Version 1.9.6 83536c65275c04366b704b2fd3f3d681c88b8267

**Platforms (REQUIRED):**
 - Platforms: Windows
 - OS: Windows 10
 - Device: PC

(Tried to add a zip file but it just failed 🤷 )
https://github.com/NaakkaDev/layoutarrowissue

sprocketc:
Moving gui elements using arrow keys on non default layouts now works as intended.

Fixes #10077

### DIFF
--- a/editor/src/clj/editor/scene.clj
+++ b/editor/src/clj/editor/scene.clj
@@ -1237,17 +1237,10 @@
           (.setImage image-view new-image))))))
 
 (defn- nudge! [scene-node-ids ^double dx ^double dy ^double dz]
-  (g/transact
-   (for [node-id scene-node-ids
-         :let [[^double x ^double y ^double z] (g/node-value node-id :position)
-               current-layout (g/maybe-node-value node-id :current-layout)
-               new-position [(+ x dx) (+ y dy) (+ z dz)]]]
-     (if (str/blank? current-layout)
-       (g/set-property node-id :position new-position)
-       (g/update-property
-        node-id :layout->prop->override
-        update current-layout
-        assoc :position new-position)))))
+  (g/with-auto-evaluation-context evaluation-context
+    (g/transact
+      (for [node-id scene-node-ids]
+        (scene-tools/manip-move evaluation-context node-id (Vector3d. dx dy dz))))))
 
 (declare selection->movable)
 
--- a/editor/src/clj/editor/scene.clj
+++ b/editor/src/clj/editor/scene.clj
@@ -14,6 +14,7 @@
 
 (ns editor.scene
   (:require [clojure.set :as set]
+            [clojure.string :as str]
             [dynamo.graph :as g]
             [editor.background :as background]
             [editor.camera :as c]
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/scene.clj
```diff
@@ -1237,9 +1238,16 @@
...
+   (for [node-id scene-node-ids
+         :let [[^double x ^double y ^double z] (g/node-value node-id :position)
+               current-layout (g/maybe-node-value node-id :current-layout)
+               new-position [(+ x dx) (+ y dy) (+ z dz)]]]
+     (if (str/blank? current-layout)
+       (g/set-property node-id :position new-position)
+       (g/update-property
+        node-id :layout->prop->override
+        update current-layout
+        assoc :position new-position)))))
```
matgis, line 1250: We should probably make `gui/update-layout-property` public and call it here instead of re-implementing it?

Style node: We use two spaces for indentation.

matgis (changes requested):

editor/src/clj/editor/scene.clj
matgis: We could use `g/maybe-node-value` in `gui/update-layout-property` without breaking things. However, thinking about it some more, I'm starting to think maybe `nudge!` should simply call `scene-tools/manip-move`, since it already has a specialization for `GuiNode` that respects the layout property. 🤔

sprocketc: That seems to work. The evaluation-context is not required in this case (as far as I understand), but I don't think that is a problem.

editor/src/clj/editor/scene.clj
```diff
@@ -1236,10 +1236,10 @@
           (.setImage image-view new-image))))))
 
 (defn- nudge! [scene-node-ids ^double dx ^double dy ^double dz]
-  (g/transact
-    (for [node-id scene-node-ids
-          :let [[^double x ^double y ^double z] (g/node-value node-id :position)]]
-      (g/set-property node-id :position [(+ x dx) (+ y dy) (+ z dz)]))))
+  (g/with-auto-evaluation-context evaluation-context
+    (g/transact
```
matgis, line 1240:
```suggestion
  (g/transact
    (g/with-auto-evaluation-context evaluation-context
```
I would swap the call order so that `g/transact` surrounds the `g/with-auto-evaluation-context` block. 

This is because `g/with-auto-evaluation-context` will transfer any changes that have been made to the local cache inside the `evaluation-context` to the system cache so they're visible to everyone at the end of its scope. But `g/transact` may also cause entries to be evicted from the cache due to its changes.

We have code to deal with conflicts like this in `internal.system/update-cache-from-evaluation-context`, but for efficiency, we want to avoid situations where an `g/with-auto-evaluation-context` block surrounds code that modifies the graph state.

matgis (changes requested):

matgis (approved):
LGTM! ❤️

Issue #10113, opened 2025-01-27
Investigate and improve build process memory usage in large projects


vlaaad:
**Task (REQUIRED):**
When building large projects, the editor uses a lot of memory. It should use less memory.

**Expected outcome (REQUIRED):**
The editor uses quantifiably less memory when building large projects.

PR #10120, opened 2025-01-28
Improve memory usage when building large projects

vlaaad:
This change improves:
- memory usage when building large projects (memory reduced by around 15%)
- performance when building large projects (time reduced by around 25%)

Technical notes:

When looking into memory usage during the build, there are several concerns:
- initial memory usage before starting the build. This is not related to the build itself, but it shows the relative importance of the build-time memory usage in relation to normal editor memory usage.
- final memory usage after finishing the build. The difference between final and initial memory usage is the size of the additional in-memory cache data used by the editor.
- peak temporary memory usage during the build. If some phase of the build requires an unexpectedly large amount of memory, it can become a problem when building large projects on machines with limited memory.

In this changeset I mostly focused on allocation hotspots, i.e. places in code that allocate a lot of temporary objects. This changeset mostly fixes low-handing fruits in the build process. This helps a bit with peak temporary memory usage, I assume this is because there is less GC pressure and fewer throw-away objects overwhelming the eden space. As a nice side-effect, reducing temporary allocations makes the build process noticeably faster.

I measured 3 "modes" of the build process:
- "full": when the project is built, there is no on-disc or in-memory cache.
- "cold": when the project is built, there is an on-disc cache but the in-memory cache is empty (i.e. build after starting the editor).
- "hot": when the project is built, there is both an on-disc and in-memory cache (i.e. build after a previous build).

All measurements were done after running GC. Typical initial memory usage is between **8 and 10 GB** before starting the build. After the build, the editor memory usage increases by around **4 to 5** GB.

Here is the before/after comparison for this changeset:

1. Peak temporary memory usage (measured as peak memory usage during the build minus final memory usage after the build):
   mode | before    | after    | improvement
   -----|----------:|---------:|-----------:
   full | `11.2 GB` | `7.7 GB` | `30.8%`
   cold | `8.2 GB`  | `5.5 GB` | `33.2%`
   hot  | `11.3 GB` | `6.6 GB` | `41.7%`

   How is this possible that hot build actually uses more throw-away memory than the cold one, if it does less by definition? Great question. We do dynamic build target resolution, which involves creating a lot of throw-away data structures. Probably GC is invoked less frequently when the build only does that... The numbers should probably be taken with a grain of salt. Still, the improvement is observable.
2. Build time:
   mode | before   | after   | improvement
   -----|---------:|--------:|-----------:
   full | `12m30s` | `8m20s` | `33.3%`
   cold | `10m15s` | `6m11s` | `39.7%`
   hot  |    `28s` |   `25s` | `8.6%`

   I observed some variation between measurements, e.g. +- 30s time per cold build is typical. Even with the variation, the improvement is noticeable.
3. Average temporary memory usage (measured as average memory usage throughout the build process minus final memory usage after the build). Average memory was calculated by collecting used memory every second during the build. The idea of this metric is to measure GC pressure: if we create fewer temporary objects, there saw pattern of memory usage is less spikey, and that should make the average memory lower.
   mode | before    | after    | improvement
   -----|----------:|---------:|-----------:
   full | `1.9 GB`  | `1.4 GB` | `26.8%`
   cold | `1.2 GB`  | `0.4 GB` | `65.7%`
   hot  | `3.1 GB`  | `2.6 GB` | `16.1%`

   Yet again, somehow the hot build uses more temporary memory than the cold one. I think this metric is not the most useful, because it smoothes out the peaks, and if there are particular points in the build process that are problematic, they will not be visible in this metric. Still, reducing the amount of allocations in hot loops is visible here.

Future work:
When investigating, I found some places that generate very big peaks in memory usage (up to 10 GBs!) during the build. Fixing those is more involved, and I will address them in a separate PR. The problematic places are:
- clj-antlr (used in `editor.lua-parser/lua-info` and `lua-parser/parse-requires`) creates a lot of lazy sequences
- `texture-page-counts` output of `GuiSceneNode` creates a lot of red/black tree nodes

Related to #10113


editor/src/clj/editor/pipeline/fontc.clj
```diff
@@ -86,21 +86,48 @@
...
+      (.write (str descent))
+      (.write " ")
+      (.write (str (:width glyph-cell-wh)))
+      (.write " ")
+      (.write (str (:height glyph-cell-wh)))
+      (.write " ")
+      (.write (str glyph-data-offset))
+      (.write " ")
+      (.write (str glyph-data-size))
+      (.write "]"))))
```
vlaaad, line 113: Font compilation created a lot of throw-away objects when computing the content hash of the font build target because there were so many of these glyph maps created. Using a record here both helps with memory and allows us to provide a more efficient digesting.

editor/src/clj/editor/pipeline/fontc.clj
```diff
@@ -285,12 +311,12 @@
...
-          height (min (.getHeight src) (.getHeight dst-in) (.getHeight dst-out))
-          int0 (int 0)
-          int1 (int 1)]
-      (doseq [^int i (range width)
+(defn- do-blend-rasters [^Raster src ^Raster dst-in ^WritableRaster dst-out]
+  (let [width (min (.getWidth src) (.getWidth dst-in) (.getWidth dst-out))
+        height (min (.getHeight src) (.getHeight dst-in) (.getHeight dst-out))
+        int0 (int 0)
+        int1 (int 1)]
+    (doseq [^int i (range width)
```
vlaaad, line 319: Just an indentation change.

editor/src/clj/editor/resource.clj
```diff
@@ -52,7 +52,17 @@
...
+  (let [cache (atom {})
+        not-found (Object.)]
+    (fn resource? [x]
+      (let [c (class x)
+            ret (@cache c not-found)]
+        (if (identical? not-found ret)
+          (let [ret (satisfies? Resource x)]
+            (swap! cache assoc c ret)
+            ret)
+          ret)))))
```
vlaaad, line 65: Making resource check much faster by using a `class->result` cache

editor/src/clj/editor/resource.clj
```diff
@@ -61,18 +71,18 @@
 
 (defn- get-resource-type [workspace resource]
   (let [output-label (if (editable? resource) :resource-types :resource-types-non-editable)
-        resource-types (g/node-value workspace output-label)
+        resource-types (output-label (g/node-by-id workspace))
```
vlaaad, line 74: We frequently call `resource/resource-type` during compilation, and every call creates an evaluation context. I observed a lot of throw-away atoms (caches in the evaluation context) originating from here. Using basis to lookup the property is much more efficient.

editor/src/clj/editor/sprite.clj
```diff
@@ -271,7 +271,7 @@
   (or (validation/prop-error :fatal _node-id :material validation/prop-nil? material "Material")
       (validation/prop-error :fatal _node-id :material validation/prop-resource-not-exists? material "Material")))
 
-(g/defnk produce-build-targets [_node-id resource textures texture-binding-infos default-animation material material-attribute-infos material-max-page-count material-samplers material-shader blend-mode size-mode manual-size slice9 dep-build-targets offset playback-rate vertex-attribute-bytes vertex-attribute-overrides]
+(g/defnk produce-build-targets [_node-id resource textures texture-binding-infos default-animation material material-attribute-infos material-max-page-count material-samplers material-shader blend-mode size-mode manual-size slice9 offset playback-rate vertex-attribute-bytes vertex-attribute-overrides]
```
vlaaad, line 274: In the large project I tested, most of the build-target -> dep-build-target connections in the graph were from sprites to their materials and from factories to their prototypes. Making these dependencies dynamic did not affect the memory usage, but it shaved off a minute from the total build time.

editor/src/clj/editor/sprite.clj
```diff
@@ -293,7 +293,7 @@
               (if unassigned-texture-error
                 [unassigned-texture-error]
                 [(validation/prop-error :fatal _node-id :textures shader/page-count-mismatch-error-message is-paged-material texture-page-count material-max-page-count label)
-                 (when (nil? unassigned-default-animation-error)
+                 (when (and anim-data (nil? unassigned-default-animation-error))
```
vlaaad, line 296: If anim-data is nil, that means that the referenced atlas failed to compile, and we are going to get a compilation error there when building. Making an animation missing error here is unnecessary at this point.

editor/src/clj/internal/graph.clj
```diff
@@ -563,10 +563,15 @@
...
-        (overrides graph node-id)))
+  (let [^Indexed os (overrides graph node-id)
+        n (count os)]
+    (loop [i 0]
+      (if (= i n)
+        nil
+        (let [override-node-id (.nth os i)]
+          (if (= override-id (gt/override-id (node-id->node graph override-node-id)))
+            override-node-id
+            (recur (inc i))))))))
```
vlaaad, line 574: This was an allocation hotspot because `some` created lazy sequences. I refactored it as a `loop`.

editor/src/clj/util/digestable.clj
```diff
@@ -205,7 +205,7 @@
 
   java.util.Map
   (digest! [value writer opts]
-    (if (satisfies? resource/Resource value)
+    (if (resource/resource? value)
```
vlaaad, line 208: This was an allocation hotspot with hashing build targets.

editor/src/clj/util/digestable.clj
```diff
@@ -232,7 +232,7 @@
    (sha1-hash object nil))
   (^String [object opts]
    (with-open [digest-output-stream (digest/make-digest-output-stream "SHA-1")
-               writer (OutputStreamWriter. digest-output-stream)]
+               writer (BufferedWriter. (OutputStreamWriter. digest-output-stream))]
```
vlaaad, line 235: There was an allocation hotspot in `digest-tagged-bytes!` that writes bytes one by one. It created a lot of `java.nio.HeapCharBuffer` objects. Buffering helped.

editor/test/integration/build_test.clj
```diff
@@ -854,19 +854,12 @@
...
-          uppercase-image-resource (workspace/resolve-workspace-resource workspace uppercase-image-path)]
-      (g/set-property! atlas-image-node :image uppercase-image-resource)
-      (let [build-error (:error (project-build project game-project-node (g/make-evaluation-context)))
-            error-message (some :message (tree-seq :causes :causes build-error))]
-        (is (g/error? build-error))
-        (is (= (str "The file '" uppercase-image-path "' could not be found.") error-message))))))
+          build-error (:error (project-build project game-project-node (g/make-evaluation-context)))
+          error-message (some :message (tree-seq :causes :causes build-error))]
+      (is (g/error? build-error))
+      (is (= "The file '/MAIN/BUTTON_CLOUDY.png' could not be found." error-message)))))
```
vlaaad, line 862: Refactored to a separate project. Other build errors in the project started triggering first because the expected one instead became a dynamic dependency, and that caused it to appear later.

editor/test/integration/sprite_test.clj
```diff
@@ -21,17 +21,25 @@
 
 (deftest replacing-sprite-image-replaces-dep-build-targets
   (test-util/with-loaded-project
-    (let [node-id (project/get-resource-node project "/logic/session/pow.sprite")
-          old-texture-binding-info (first (g/node-value node-id :texture-binding-infos))
+    (let [sprite-id (project/get-resource-node project "/logic/session/pow.sprite")
+          old-texture-binding-info (first (g/node-value sprite-id :texture-binding-infos))
           old-texture-binding-node-id (:_node-id old-texture-binding-info)
-          old-image (:texture old-texture-binding-info)]
-      (let [old-sources (g/sources-of old-texture-binding-node-id :build-targets)]
```
vlaaad, line 28: The old test impl relied on explicit build target connections, but now we use dynamic deps there.

editor/src/clj/editor/resource.clj
matgis: Ohh, this should be good!

The program feels more healthy already! ❤️

editor/src/clj/internal/graph.clj
matgis: Maybe we should have our own`coll/some`?

Is using a loop more efficient than `reduce`? 🤔

editor/src/clj/editor/pipeline.clj
```diff
@@ -30,86 +30,69 @@
...
+          (throw (ex-info "value for resource field in pb-map is not a Resource or proj-path."
+                          {:field-value field-value :field-path field-path})))
+        (if-let [fused-path (field-value->fused-path field-value)]
+          (coll/assoc-in-ex pb-map field-path fused-path)
+          (throw (ex-info "deps-resources is missing a referenced source-resource. Ensure it is among the dep-build-targets."
+                          {:field-value field-value
+                           :field-path field-path
+                           :deps-by-source field-value->fused-path}))))
+      pb-map
+      (protobuf/get-field-value-paths pb-map pb-class))))
```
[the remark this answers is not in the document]
matgis: Since there are a few details that weren't fully nailed down around `make-protobuf-build-target`, I think we should take this opportunity to flesh out `make-protobuf-build-target-test` somewhat.

matgis (changes requested):
I'd like to see the existing `make-protobuf-build-target-test` extended a bit to nail down some of the details that were unclear now that we're rewriting it. Looks good otherwise! ❤️

editor/src/clj/editor/resource.clj
matgis: But `extends?` is supposed to check if the class implements a particular protocol? 
https://clojuredocs.org/clojure.core/extends_q

Are you perhaps thinking of some other function? 🤔

editor/src/clj/internal/graph.clj
matgis: Ah, fair enough!

editor/src/clj/editor/resource.clj
vlaaad: It looks like `extends?` has a more narrow scope and does not walk the inheritance chain, which can lead to false negatives. Ah, it would be so much simpler if `Resource` was an interface...

editor/test/editor/pipeline_test.clj
```diff
@@ -201,3 +201,73 @@
...
+  (abs-path [_] proj-path)
+  (proj-path [_] proj-path)
+  (resource-name [_] (FilenameUtils/getName proj-path))
+  (workspace [_])
+  (resource-hash [_] (hash proj-path))
+  (openable? [_] false)
+  (editable? [_] false))
+
+(defn- build-resource [proj-path]
+  (workspace/make-build-resource (->TestResource proj-path)))
```
matgis, line 223: Would `test-util/make-fake-file-resource` work for your use case?

editor/src/clj/editor/pipeline.clj
```diff
@@ -56,12 +57,13 @@
...
                           {:field-value field-value :field-path field-path})))
-        (if-let [fused-path (field-value->fused-path field-value)]
-          (coll/assoc-in-ex pb-map field-path fused-path)
-          (throw (ex-info "deps-resources is missing a referenced source-resource. Ensure it is among the dep-build-targets."
-                          {:field-value field-value
-                           :field-path field-path
-                           :deps-by-source field-value->fused-path}))))
+        (let [fused-path (field-value->fused-path field-value ::not-found)]
+          (if (identical? ::not-found fused-path)
+            (throw (ex-info "deps-resources is missing a referenced source-resource. Ensure it is among the dep-build-targets."
```
matgis, line 62: Typo: `deps-resources` -> `dep-resources`? It was like that before, though.

matgis (dismissed):
LGTM! ❤️

editor/test/editor/pipeline_test.clj
vlaaad: I kinda like this one more, just a project path basically.

PR #10124, opened 2025-01-28
Make it possible to hook the imports

smagnuso:


ekharkunov (dismissed):

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -43,6 +43,8 @@ var CUSTOM_PARAMETERS = {
         var e = document.getElementById("webgl-not-supported");
         e.style.display = "block";
     },
+    update_imports: function() {
```
AGulev, line 46: Above `var CUSTOM_PARAMETERS` we have a short explanation for each parameter which used as documentation on manual page, could you please add the same for `update_imports` ?

AGulev (changes requested):
Looks good, but documentation missed

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
smagnuso: Sure

AGulev (approved):
Great! Thank you!!!

PR #10129, opened 2025-01-30
Fix mouse scroll regression from GLFW update

### LINKED ISSUE #9494: Mouse wheel up and down events are triggered much less than before
After the Defold update, the mouse wheel events starts triggered much less and requires do a "bigger" scroll gesture to trigger this event.
I'm using MacOS and their trackpad. From Defold 1.9.0 or 1.9.1 probably (I've tried to update to 1.9.3)

Before update I can do a "precission" scroll for elements, but after update I can't, only a fast gesture is registered

**To Reproduce (REQUIRED)**
1. Add mouse wheel events in bindings
2. Do a print message in on_input events
3. See a logs to see when event triggered. In Defold 1.8.0 It will be much often and more precise

No changed behaviour from 1.8.0

Tested on Defold 1.9.3

**Platforms (REQUIRED):**
- Tested on MacOS, trackpad

Jhonnyg:
Fixed an issue where the mouse wheel events from GLFW had changed, causing the mouse wheel to be less precise than before.

Fixes #9494


Insality:
Thanks for this!

Jhonnyg:
@Insality would you mind testing this to make sure we fixed the issue for you?

Insality:
Yeah
I need a sha and channel

Insality:
Seems the channel is "dev" btw not "alpha" :)

@Jhonnyg the scrolling works! I tested on MacOS desktop, feels good now

engine/input/src/input.cpp
```diff
@@ -701,10 +702,6 @@ namespace dmInput
         }
         if (binding->m_MouseBinding != 0x0)
         {
-            bool is_wheel_offsets = false; // glfw 2.x
-            #if defined(DM_INPUT_USE_GLFW3)
-                is_wheel_offsets = true; // glfw 3, the wheel scroll is offsets
-            #endif
```
Jhonnyg, line 707: Moved this logic into the platform layer instead.

engine/platform/src/platform_window_glfw3.cpp
```diff
@@ -481,7 +480,7 @@ namespace dmPlatform
 
     int32_t GetMouseWheel(HWindow window)
     {
-        return (int32_t) window->m_MouseScrollY;
+        return (int32_t) round(window->m_MouseScrollY);
```
Jhonnyg, line 483: This is what the old GLFW implementation did with the double values. Seems to have most impact on mac, on windows its works the same regardless of the rounding

JCash (approved):

Issue #9818, opened 2024-11-22
Show/hide selected objects (Ctrl+E) does not work with models component.


Dragosha:
Show/hide selected objects stopped working with the model component. This seems to have happened in version 1.9.5. Other components such as the GUI are hidden from the screen.

https://github.com/user-attachments/assets/833efabd-100a-4f6b-bd8a-80893bb14de5

<hr/>

<table>
<tr><td>Build time</td><td>2024-11-21T14:48:58.772099</td></tr>
<tr><td>Defold channel</td><td>editor-alpha</td></tr>
<tr><td>Defold editor sha</td><td>addf94a3946b973a491a20b276314cb672835951</td></tr>
<tr><td>Defold engine sha</td><td>87b6907759f7b8dff830d54b2250b8d721bde291</td></tr>
<tr><td>Defold version</td><td>1.9.5</td></tr>
<tr><td>GPU</td><td>NVIDIA GeForce GTX 1060 6GB/PCIe/SSE2</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 NVIDIA 555.85</td></tr>
<tr><td>Java version</td><td>17.0.5+8</td></tr>
<tr><td>OS arch</td><td>amd64</td></tr>
<tr><td>OS name</td><td>Windows 10</td></tr>
<tr><td>OS version</td><td>10.0</td></tr>
<table>

selimanac:
Now we have a great visibility toggle on the outline, but unfortunately, it is not working with models.:(

https://github.com/user-attachments/assets/0624af51-54d0-48e3-b346-f264adee78a5

PR #10133, opened 2025-01-30
Introduce visibility toggle on the outline panel

### LINKED ISSUE #3916: Toggle object visibility from the outline panel (DEFEDIT-1416)
The outline-interactions described in section 4a from this document:
(LINK REMOVED)Show the visibility status of individual object in the outline, and allow the user to toggle visibility by clicking on an eye icon in the outline.

Note that this selection state will need to be layered in such a way that it will work alongside DEFEDIT-1386 once implemented.

### LINKED ISSUE #6026: Add Show/Hide visibility switch button for components in the editor
As mentioned in the forum post below, an option for switching visibility of components would be a very comfortable solution!
There is already an option to Hide/Show objects, so the feature could probably use it and add a button near components to utilize it for switching between visibility states. 

https://forum.defold.com/t/idea-for-selection-and-visualization-filter/69190

A button near components name to switch (show/hide) visibility state of the component in the editor.

Alternative is the usage of current Hide/Show buttons from the menu.

In huge projects with extensive collections this could really speed up development process.

sprocketc:
Introduced a visibility toggle to the outline tree. The toggle will be visible on mouse over, or when the node is hidden.

visibility-toggle

Fixes #6026

### Technical changes
Rendering an additional button on the tree cell that sticks to the right side of the panel, without blocking horizontal scrolling.


editor/src/clj/editor/outline_view.clj
```diff
@@ -112,23 +115,24 @@
...
-         data (mapv #(decorate hidden-node-outline-key-paths node-id-path node-outline-key-path % (or parent-reference? (:outline-reference? item))) (:children item))
+         data (mapv #(decorate hidden-node-outline-key-paths node-id-path node-outline-key-path % (or parent-reference? (:outline-reference? item)) outline-name-paths) (:children item))
          item (assoc item
                 :node-id-path node-id-path
                 :node-outline-key-path node-outline-key-path
                 :parent-reference? parent-reference?
                 :children (mapv :item data)
                 :child-error? (boolean (some :child-error? data))
                 :child-overridden? (boolean (some :child-overridden? data))
+                :hideable? (contains? outline-name-paths (rest node-outline-key-path))
```
sprocketc, line 135: Is there an easier way to get to this info? Connecting to `outline-name-paths` seemed like the logical solution here, given the current implementation.

editor/src/clj/editor/outline_view.clj
```diff
@@ -458,9 +463,29 @@
...
+  (.consume event))
+
+(def eye-open-path (ui/load-svg-path "scene/images/eye_open.svg"))
+(def eye-closed-path (ui/load-svg-path "scene/images/eye_closed.svg"))
+
 (defn make-tree-cell
   [^TreeView tree-view drag-entered-handler drag-exited-handler]
-  (let [cell (proxy [TreeCell] []
+  (let [eye-icon-open (app-view/make-svg-icon-graphic eye-open-path)
+        eye-icon-closed (app-view/make-svg-icon-graphic eye-closed-path)
```
sprocketc, line 476: Part of the initialization was moved here for performance purposes. That solved almost all blinking issues on visibility toggling, but there are still some rare cases where a tree cell might appear to be re-rendered, although it shouldn't. Is there anything else we can do?

editor/src/clj/editor/scene_visibility.clj
```diff
@@ -254,6 +255,16 @@
...
+(handler/defhandler :hide-toggle :workbench
+  (active? [scene-visibility evaluation-context user-data]
+           (g/node-value scene-visibility :active-scene-resource-node evaluation-context))
+  (run [scene-visibility user-data]
+       (let [{:keys [node-outline-key-path]} user-data
+             outline-name-paths #{(vec (rest node-outline-key-path))}]
+         (if (contains? (g/node-value scene-visibility :hidden-node-outline-key-paths) node-outline-key-path)
+           (show-outline-name-paths! scene-visibility outline-name-paths)
+           (hide-outline-name-paths! scene-visibility outline-name-paths)))))
+
```
sprocketc, line 267: A new handler was introduced to toggle the visibility state of a single nodes. The existing context menu action is applied to all selected elements and its children. That might result in unexpected behavior, so I decided to deviate. For instance, if we try to unhide a node that has at least one visible children using the context menu, the child will turn invisible, instead of changing the visibility of the selected parent.

editor/src/clj/editor/outline_view.clj
```diff
@@ -471,15 +496,20 @@
...
-                           label (if show-link? (format "%s - %s" label (resource/resource->proj-path link)) label)]
-                       (proxy-super setText label)
-                       (proxy-super setGraphic (icons/get-image-view icon 16))
+                           text (if show-link? (format "%s - %s" label (resource/resource->proj-path link)) label)
+                           hidden? (= :hidden scene-visibility)
+                           visibility-icon (if hidden? eye-icon-closed eye-icon-open)]
+                       (.setImage image-view-icon (icons/get-image icon))
+                       (.setText text-label text)
+                       (.setGraphic visibility-button visibility-icon)
+                       (proxy-super setGraphic pane)
```
sprocketc, line 509: I initially tried switching to a `TreeTableView` component that seemed like a better fit, but that would require a lot of changes that might produce more problems down the line. I worked around the issue here by using `setGraphic` to render the icon+label+toggle.

editor/src/clj/editor/outline_view.clj
```diff
@@ -458,9 +463,29 @@
...
+  (ui/run-command (.getSource event) :hide-toggle {:node-outline-key-path node-outline-key-path})
+  (.consume event))
+
+(def eye-open-path (ui/load-svg-path "scene/images/eye_open.svg"))
+(def eye-closed-path (ui/load-svg-path "scene/images/eye_closed.svg"))
+
 (defn make-tree-cell
   [^TreeView tree-view drag-entered-handler drag-exited-handler]
-  (let [cell (proxy [TreeCell] []
+  (let [eye-icon-open (app-view/make-svg-icon-graphic eye-open-path)
```
sprocketc, line 475: Not that important, but should we move `make-svg-icon-graphic` under `editor.ui.clj`, or maybe `editor.icons.clj`?

editor/styling/stylesheets/modules/_outline.scss
```diff
@@ -76,6 +68,39 @@
       @include colorize-disclosure-arrow(-df-error-severity-fatal-dim);
     }
 
+    .label {
+      -fx-min-width: 40px;
+    }
```
sprocketc, line 73: This PR enables ellipsis on text-overrun, to allow displaying the toggle at the right side of the cell. That might not be ideal on deeply nested outlines. Although horizontal scrolling is not ideal either, this is a middle ground solution that always shows part of the label, and enables horizontal scrolling sooner. For what is worth, I tried to make the visibility toggle stick to the right, without blocking scroll, but I didn't manage to produce a good result. The main problem was that it required watching the scrolling position to rerender the button on a different anchor point, and I couldn't make this work properly.

editor/src/clj/editor/outline_view.clj
matgis: I think both `hideable` and `hidable` are alternative but valid ways to spell it?

Regarding using `outline-name-paths`, it works for now, but may not be exactly what we want. For the existing "hide this and everything below it" action, we want to be able to select a "group" element like for example a Game Object, and hide everything related to it. But the Game Object itself doesn't have a visual component - it is the various components below it that may or may not be hideable. So technically I think we may want to not show the eye icon next to Game Objects, just like we do not show it for things like Script components.

We might be able to check if the `child-scene` contains a `:renderable` when we produce the `outline-name-paths` in `scene-visibility/scene-outline-name-paths`, but I think we want the existing behavior for the "hide this and everything below it" action, so maybe we need to restructure the return value to include a bool signifying whether or not the `outline-name-path` is associated with a `:renderable`.

I think we can do these changes in a separate PR, though.

editor/src/clj/editor/outline_view.clj
matgis: I suspect there might be a bug in the JavaFX redraw code. I've certainly seen issues around hover highlighting and the disclosure triangle earlier. I don't think we should worry about it for this PR.

editor/src/clj/editor/tile_map.clj
```diff
@@ -486,7 +486,7 @@
                     :passes [pass/transparent pass/selection]}})))
 
 (g/defnk produce-layer-outline
-  [_node-id id z]
+  [_node-id id z visible]
```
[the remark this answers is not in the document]
matgis: The same is true for Gui scene nodes, BTW. They have a Visible checkbox.

With the previous "hide this and everything below it" functionality, we explicitly didn't want to tie it to the setting that affects visibility in-game - it was considered more of a "scene management" feature for users of the editor, so they could get rid of clutter and focus on particular parts of the scene.

With the eye-icon visibility toggle, I personally think it should work the same way (and it nicely does interact with the existing show / hide stack 👍). However, I'm sure some users might expect it to be tied to the in-game state as well.

For now, I think we should leave it as implemented, and if enough people yell about it, we might reconsider. 😅

editor/styling/stylesheets/modules/_outline.scss
matgis: In many ways it would be nicer if the eye icons were in their own column to the right of the tree view item, but I understand that would necessitate more drastic changes to the view.

For now, I think this is fine. Maybe we'll revisit this later if we need to!

editor/resources/scene/images/eye_closed.svg
matgis, line 1: Beautiful icons! ❤️ Did you make these?

matgis (dismissed):
Works great! ❤️

vlaaad:
Before I got sick, I started reviewing it and wanted to discuss aspects of this PR:
- are we sure that using ellipsis instead of horizontal scroll bar is not a problem? In gui files with lots of nested templates, a lot of nodes in the outline view share the same (very long) text prefix. It might be a problem
- should the on/off switch be recursive? I.e. should turning off go visibility make its sprites invisible? Currently it does not work like that, and I found it puzzling in my tests 

Either way, the improvements are very cool!

sprocketc:
> * are we sure that using ellipsis instead of horizontal scroll bar is not a problem? In gui files with lots of nested templates, a lot of nodes in the outline view share the same (very long) text prefix. It might be a problem

I have the same fear. Horizontal scrolling will be enabled btw, when the rightmost element reaches 40px, so it won't block them. The icon will be hidden in this case until you scroll. Ideally, I would like to have a TreeTable component,  with the visibility column at the left side of the tree, and free scrolling. I think it can be done, but requires a lot of changes

> * should the on/off switch be recursive? I.e. should turning off go visibility make its sprites invisible? Currently it does not work like that, and I found it puzzling in my tests

Yeah, I think I need to do this. The problem in my opinion though, is that when a node is hidden, it should also automatically block child rendering. Until now, we toggle the visibility of all children. I made a little research to see how similar editors work, and it looks like there is no standard. I guess being able to see a node, without having to make its parent visible has its pros, I just think it's not intuitive. Toggling all children will also be consistent with the existing behavior.

In any case, I won't merge this until we make a decision,

editor/resources/scene/images/eye_closed.svg
sprocketc: They are based on the existing `eye_icon_eye_arrow.svg` for consistency.

editor/src/clj/editor/outline_view.clj
sprocketc: I will probably change the behavior to toggle all children, so this case would make more sense.

editor/src/clj/editor/tile_map.clj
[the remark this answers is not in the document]
sprocketc: I don't think we should tie those options. What I found weird was that if you make the layer invisible, you are blocking the in-editor visibility toggling. For example, if I turn the in-game visibility of a tilemap layer off, and then hide the parent, the change won't be propagated to the layer. Now that I think about it, it will be weird if the state on the view is not synced with the icon, so I guess it's fine.

sprocketc:
Recent changes
- We now also toggle the visibility of children. This feels fine, because it depends on the state of a single node, so the expected action will always be triggered.
- Increased the minimum label width to `100px`
- Moved make-svg-icon-graphic to icons ns
- Fixed some color issues introduced here, and also fixed an existing issue with the default/hover color of the icon that was not consistent with the asset tree. I don't know if the specificity changes introduced here #7904 are still relevant, but I tried to maintain them. I would assume `-fx-effect` also has a negative impact, but that requires further investigation.

I can give `TreeTableView` another try , since there is `treeColumnProperty` that can move the visibility column to the left, but the setup feels a little fragile to make additional changes.

sprocketc:
I managed to make the toggle button stick to the right by observing multiple values

https://github.com/user-attachments/assets/16d3d320-7ec8-450c-89ce-0f93b0898dd8

I also fixed 
 > - Sometimes, toggling the visibility moves the focus to the first input of the properties panel.
 > - In some cases, the title-bar menu is removed until you click somewhere on the app to trigger a re-render.
 
by adding an additional event filter to stop the propagation. I cannot find a way to get rid of the random flickering that already existed.

matgis (dismissed):
And now it is even better! Love it. ❤️

vlaaad (dismissed):
This is seriously amazing! Great job!

My only tiny nitpick is that double-clicking the eye icon of a selected tree item opens the referenced file instead of toggling the visibility:

https://github.com/user-attachments/assets/3ef56855-a07e-4ccf-a4fb-4ede2bd585e8

sprocketc:
> My only tiny nitpick is that double-clicking the eye icon of a selected tree item opens the referenced file instead of toggling the visibility:

@vlaaad I wouldn't call that a nitpick :smile: I am aware of this, but I cannot block the action for some reason, even if I add another `MOUSE_CLICKED` filter here https://github.com/defold/defold/pull/10133/files#diff-4c381bb725e8f7d75b6700638fe4d057dbaa9372f3d8cdf7f5bb6c94f104b518R487-R489. This only happens if the element is currently selected. Anyway I pushed a minor fix to add a missing typehint, so I'll need another pass. I appreciate the ultra careful testing btw.

vlaaad (approved):
LGTM!

selimanac:
Great addition! So useful—thank you so much!
But unfortunately, the visibility toggle doesn't do anything on model components. Is this a known issue? Should I open an issue or submit a feature request?

https://github.com/user-attachments/assets/a122973a-2f03-43b1-860a-728978b4d068

selimanac:
> @selimanac Hey, thanks for testing this! It looks like there is an open issue #9818.

Ah, thanks! Sorry, I didn't see that.

PR #10139, opened 2025-01-31
Improve build time and memory for large projects

vlaaad:
This changeset improves the build time by 15% (e.g. 8m03s -> 6m54s) and reduces final memory usage after the build by 10% (e.g. 12.6Gb -> 11.2Gb) for large projects.

Related to #10113

### DIFF
--- a/editor/project.clj
+++ b/editor/project.clj
@@ -225,6 +225,7 @@
                       :no-schemas {:jvm-opts ["-Ddefold.schema.check.disable=true"]}
                       :no-spec-asserts {:jvm-opts ["-Dclojure.spec.compile-asserts=false"]}
                       :performance [:no-decorated-exceptions :no-schemas :no-spec-asserts]
+                      :16gb {:jvm-opts ["-Xmx16g"]}
                       :x86_64-linux {:dependencies [[org.openjfx/javafx-base "23.0.1" :classifier "linux" :exclusions [org.openjfx/javafx-base]]
                                                     [org.openjfx/javafx-controls "23.0.1" :classifier "linux" :exclusions [org.openjfx/javafx-controls org.openjfx/javafx-graphics]]
                                                     [org.openjfx/javafx-graphics "23.0.1" :classifier "linux" :exclusions [org.openjfx/javafx-graphics org.openjfx/javafx-base]]
--- a/editor/src/clj/editor/gui.clj
+++ b/editor/src/clj/editor/gui.clj
@@ -2143,7 +2143,7 @@
   (property name g/Str)
   (property gpu-texture TextureLifecycle)
   (output texture-infos GuiResourceTextureInfos (g/fnk [name] (sorted-map name {})))
-  (output texture-page-counts GuiResourcePageCounts (g/fnk [name] (sorted-map name nil))) ; Use a nil texture-page-count to disable validation against the material page count.
+  (output texture-page-counts GuiResourcePageCounts (g/fnk [name] {name nil})) ; Use a nil texture-page-count to disable validation against the material page count.
   (output texture-gpu-textures GuiResourceTextures (g/fnk [name gpu-texture] {name gpu-texture})))
 
 (g/defnk produce-texture-gpu-textures [_node-id texture-names gpu-texture default-tex-params samplers]
@@ -2192,7 +2192,7 @@
   ;; The texture-page-count can be nil, which will disable validation against
   ;; the material page count. We still want an entry in the map since the keys
   ;; are used to validate the texture names.
-  (into (sorted-map)
+  (into {}
         (map (fn [texture-name]
                (pair texture-name texture-page-count)))
         texture-names))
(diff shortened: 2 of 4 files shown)


editor/project.clj
```diff
@@ -225,6 +225,7 @@
                       :no-schemas {:jvm-opts ["-Ddefold.schema.check.disable=true"]}
                       :no-spec-asserts {:jvm-opts ["-Dclojure.spec.compile-asserts=false"]}
                       :performance [:no-decorated-exceptions :no-schemas :no-spec-asserts]
+                      :16gb {:jvm-opts ["-Xmx16g"]}
```
vlaaad, line 228: I did a test on a smaller heap, the large project can still be built even if the memory usage peaks at 20Gb on my machine!

editor/src/clj/editor/gui.clj
```diff
@@ -2143,7 +2143,7 @@
   (property name g/Str)
   (property gpu-texture TextureLifecycle)
   (output texture-infos GuiResourceTextureInfos (g/fnk [name] (sorted-map name {})))
-  (output texture-page-counts GuiResourcePageCounts (g/fnk [name] (sorted-map name nil))) ; Use a nil texture-page-count to disable validation against the material page count.
+  (output texture-page-counts GuiResourcePageCounts (g/fnk [name] {name nil})) ; Use a nil texture-page-count to disable validation against the material page count.
```
vlaaad, line 2146: We don't need to sort the `texture-page-counts`, but using sorted maps instead of hash maps significantly contributes to GC pressure.

editor/project.clj
AGulev, line 228: usually when we are talking about 16GB, it's about the laptop memory, and as I remember we allow jvm to use only 75% out of it, so it should be something like 12gb

editor/src/clj/editor/gui.clj
```diff
@@ -2574,6 +2574,9 @@
   (output texture-resource-names GuiResourceNames :cached (g/fnk [names] (into (sorted-set) names)))
   (input build-errors g/Any :array)
   (output build-errors g/Any (gu/passthrough build-errors))
+  (input texture-page-counts g/Any :array)
+  (output texture-page-counts g/Any :cached (g/fnk [texture-page-counts]
+                                              (into {} cat texture-page-counts)))
```
vlaaad, line 2579: Now we aggregate `texture-page-counts` from all textures on a `TexturesNode`, and have a single non-array connection from here to `GuiSceneNode`.

editor/src/clj/editor/pipeline/fontc.clj
```diff
@@ -592,13 +592,14 @@
...
-        ^double shadow-alpha shadow-alpha
-        ^double sdf-outline sdf-outline
+  (let [glyph-left-bearing (double glyph-left-bearing)
+        ^int padding padding
+        outline-width (double outline-width)
+        sdf-spread (double sdf-spread)
+        sdf-shadow-spread (double sdf-shadow-spread)
+        edge (double edge)
+        shadow-alpha (double shadow-alpha)
+        sdf-outline (double sdf-outline)
```
vlaaad, line 602: Hinting locals as `^double` is not enough apparently: locals keep being objects that get cast to double on use. This place generated a lot of `Double`s. How I figured this out: I reloaded the whole ns, and then did `(compile 'editor.pipeline.fontc)`. This dumped the classes on disc. Then, I opened `target/classes/editor/pipeline/fontc$draw_ttf_distance_field.class` and saw a lot of `RT.uncheckedDoubleCast`s and `(Number)a_local`s.

editor/src/clj/editor/pipeline/fontc.clj
```diff
@@ -720,26 +721,31 @@
...
+                                        (* y row-size))]
+                        (aset glyph-data-bank y-offset (unchecked-byte 0))
+                        (System/arraycopy field (+ (* y image-width channel-count) 0) glyph-data-bank (+ y-offset channel-count) (* image-width channel-count))
+                        (aset glyph-data-bank (+ (+ y-offset channel-count) (* image-width channel-count)) (unchecked-byte 0))))
+                    (let [last-y-offset (- (+ glyph-data-offset glyph-data-size) row-size)]
+                      (Arrays/fill glyph-data-bank last-y-offset (+ last-y-offset row-size) (unchecked-byte 0)))))
+                batch))
+        (sequence (comp
+                    positive-glyph-extent-pairs-xf
+                    (partition-all 100))
```
vlaaad, line 746: Batching improved font compilation speed from 3s to 2s. Also, I did replaced `doall` with `dorun` so we don't keep the seq in memory while performing the drawing. Not that it matters now with batching...

editor/src/clj/util/coll.clj
```diff
@@ -221,6 +221,7 @@
    (cond
      (empty? a) b
      (empty? b) a
+     (and (map? a) (supports-transient? a) (map? b)) (-> (reduce-kv assoc! (transient a) b) persistent! (with-meta (meta a)))
```
vlaaad, line 224: A lot of map entries were created when compiling guis (`GuiSceneNode`'s `basic-gui-scene-info` output).

editor/project.clj
vlaaad: Good point, I'll try to use even lower heaps.

editor/src/clj/editor/pipeline/fontc.clj
matgis: Whaaat? 😨

matgis (approved):
Nice findings! LGTM! ❤️

PR #10144, opened 2025-02-01
Fixed an issue where some node properties became hidden when multiple nodes were selected

### LINKED ISSUE #9992: Selecting multiple gui nodes in the outline window hides some properties
When selecting two or more text nodes, the property field for Font and Material disappear, making it impossible to change the current font for multiple nodes at the same time. When selecting two or mode box nodes then the Texture and Material property fields disappear. I am fairly sure this didn't happen 4-6 versions ago.

**To Reproduce (REQUIRED)**
1. Open any gui with two text nodes.
2. CTRL+Click two or more text nodes in the outline window.
3. Try to change the font property, you cannot.

Property fields not disappearing when selecting multipe nodes of the same type.

  - Version 1.9.6

**Platforms (REQUIRED):**
 - Platforms: Windows
 - OS: Windows 10
 - Device: PC

**Screenshots (OPTIONAL):**
One box node selected.
image

Two box nodes selected.
image

AGulev:
Fixed an issue where some node properties were not editable when more than one node was selected.  

Fix https://github.com/defold/defold/issues/9992

### DIFF
--- a/editor/src/clj/editor/properties.clj
+++ b/editor/src/clj/editor/properties.clj
@@ -457,6 +457,13 @@
 (defn visible? [property]
   (:visible property true))
 
+;; Return property, replacing function instances with function names for comparison
+(defn normalize-edit-type [property]
+  (reduce-kv (fn [m k v]
+               (assoc m k (if (fn? v) (class v) v)))
+             {}
+             property))
+
 (defn coalesce [properties]
   (let [node-ids (mapv :node-id properties)
         properties (mapv flatten-properties properties)
@@ -474,9 +481,11 @@
                           ;; Heuristic is to compare count and also type
                           (filter
                             (fn [e]
-                              (let [v (val e)]
+                              (let [v (val e)
+                                    edit-type (map property-edit-type v)
+                                    normalized-edit-type (map normalize-edit-type edit-type)]
                                 (and (= node-count (count v))
-                                     (apply = (map property-edit-type v))))))
+                                     (apply = normalized-edit-type)))))
                           ;; Coalesce into properties consumable by e.g. the properties view
                           (map (fn [[k v]]
                                  (let [prop {:key k
@@ -508,7 +517,8 @@
                                        prop (cond-> prop
                                                     (not-every? nil? original-values)
                                                     (assoc :original-values original-values))]
-                                   (pair k prop)))))
+                                   (pair k prop))))
+                          )
                         visible-prop-colls)]
     {:properties coalesced
      :display-order (prune-display-order (first display-orders) (set (keys coalesced)))
--- a/editor/src/clj/editor/properties.clj
+++ b/editor/src/clj/editor/properties.clj
@@ -517,8 +517,7 @@
                                        prop (cond-> prop
                                                     (not-every? nil? original-values)
                                                     (assoc :original-values original-values))]
-                                   (pair k prop))))
-                          )
+                                   (pair k prop)))))
                         visible-prop-colls)]
     {:properties coalesced
      :display-order (prune-display-order (first display-orders) (set (keys coalesced)))


editor/src/clj/editor/properties.clj
```diff
@@ -457,6 +457,13 @@
 (defn visible? [property]
   (:visible property true))
 
+;; Return property, replacing function instances with function names for comparison
+(defn normalize-edit-type [property]
+  (reduce-kv (fn [m k v]
+               (assoc m k (if (fn? v) (class v) v)))
```
vlaaad, line 463: I'm not sure this change is safe. We need to check for equality because we use a single edit type on multiple values afterward. This is not safe because the actual functions in the edit type may close over different values, and ignoring this may result in data corruption.

Instead, I think we should use `fn/partial` in the edit type function because it provides equality semantics for functions. For example, gui's `layout-property-edit-type-set-impl` should reorder its arguments to receive `prop-kw` and `changes-fn` as first 2 args, and then `wrap-layout-property-edit-type` should create `(fn/partial layout-property-edit-type-set-impl prop-kw changes-fn)` instead of an anonymous function to be used as `set-fn`.

vlaaad (changes requested):
I think we should do it differently, because this change is not safe: we really do need equality. It's unfortunate that anonymous functions in Clojure don't define equality semantics based on the values they close over, but we have `fn/partial` to help with that. With that, we can show the shared properties when they are truly equal, and hide them when they are not.

vlaaad:
@matgis you might be interested in this.

editor/src/clj/editor/properties.clj
AGulev: hmm. you know better ofc, but the result of this operation is function but without instance. If we can set different values using the same function, we can edit it together. Isn't it?  For us it's important to know, that if we call this function to set value - it works, and if values works with the same functions it will work.
Maybe there are some situations I don't know about, but it seems logical to me

AGulev: That's how this object looks like:
```
#object[editor.gui$fn__108378$position_edit_type_set_fn__108379 0x3e6c66fc editor.gui$fn__108378$position_edit_type_set_fn__108379@3e6c66fc]
```
So, we get only function (class) part:
```
editor.gui$fn__108378$position_edit_type_set_fn__108379
```

vlaaad: That's my point though: the fields that the instance of a function class has are important and should not be ignored.
Here is an example:
```clj
;; function for creating set-fn for edit-type
(defn- make-set-fn [prop-kw]
  (fn [node-id value]
    (dynamo.graph/set-property node-id prop-kw value)))

;; node A has this
(make-set-fn :foo)

;; node B has this
(make-set-fn :foo)

;; when A and B are selected, the shared property will be visible
(= (class (make-set-fn :foo))
   (class (make-set-fn :bar)))
=> true
```
The problem is here:
https://github.com/defold/defold/blob/99da8ae275c1c64da08f2d0cdb0f8f5d212e23f0/editor/src/clj/editor/properties.clj#L504

If we decide to show the property for multi-selection, we will show one of them — the first one. This will cause `(make-set-fn :foo)` to be used both for nodes A (expected) and nodes B (unexpected, will cause bugs since it expects the property `:bar` to be set).

AGulev: ok, that means all the `clean-fn` and `set-fn` should be rewritten?

vlaaad: Not all, but those that use anonymous functions.

vlaaad: There is e.g. :set-fn in `editor.sprite` that does the right thing, and `curve-set-fn` with `curve-spread-set-fn` are fine because they refer to a top-level def of a function, not a closure.

AGulev: ok, I'll take a look, thank you

AGulev:
this solution is wrong, 
I can't provide right solution, it requires changes in `defmacro wrap-layout-property-edit-type`

PR #10150, opened 2025-02-03
Minor toolbar enhancements

sprocketc:
Add tooltips to toolbar and separate tool activation buttons from the rest.

Screenshot From 2025-02-03 17-11-24


sprocketc:
This is one of the first things I noticed, so I decided to open a PR to make a simple suggestion. The reasoning here is that the tool activation buttons are part of a separate group. Also, it might be hard to interpret certain icons depending on the view (e.g. the perspective icon on a 2d view).

matgis (approved):
Ship it! ❤️

PR #10153, opened 2025-02-03
Split the two physics engine integrations into separate implementations

### LINKED ISSUE #10132: Split physics implementation from a shared structure into physics engine counterparts (box2d vs bullet3d)
**Task (REQUIRED):**
The current physics implementation uses the same resource types as well as component in the engine. There is a lot of logic in place in these source files where we take decisions based on what "physics type" the project has been configured with. But this also means that it is currently impossible to use both 2d and 3d physics in the same project (e.g making mini-games might want to use different physics). 

Better codebase that is prepared for implementing physics as extensions.

Jhonnyg:
The internal implementation of the two physics engine integrations (box2d and bullet3d) have been refactored into separate component implementations. This is the first step into updating the box2d library to the latest version, as well as pave the way for supporting physics engines as extensions (e.g Jolt physics as an extension).

Fixes #10132


engine/gamesys/src/dmsdk/gamesys/resources/res_convex_shape.h
```diff
@@ -23,14 +23,14 @@ namespace dmPhysics
 
 namespace dmGameSystem
 {
-    struct ConvexShapeResource
+    struct ConvexShapeResourceBox2D
```
Jhonnyg, line 26: Should we keep the shape resource as before? I don't know if this is a breaking change or not, but I'd like to avoid the "m_3D" and having separate objects.. Maybe it should be just in the resource type?

`dmPhysics::HCollisionShape m_Shape;`

engine/gamesys/src/gamesys/components/comp_collision_object.h
```diff
@@ -104,15 +64,32 @@ namespace dmGameSystem
...
-    bool GetShapeIndex(void* _component, dmhash_t shape_name_hash, uint32_t* index_out);
-    bool GetShape(void* _world, void* _component, uint32_t shape_ix, ShapeInfo* shape_info);
-    bool SetShape(void* _world, void* _component, uint32_t shape_ix, ShapeInfo* shape_info);
-
-
-    // For script_box2d.cpp
-    b2World* CompCollisionObjectGetBox2DWorld(dmGameObject::HComponentWorld world);
-    b2Body* CompCollisionObjectGetBox2DBody(dmGameObject::HComponent component);
-    dmGameObject::HInstance CompCollisionObjectGetInstance(void* _user_data);
+    // Adapter API
```
Jhonnyg, line 67: I think this is just for scripts, so maybe it should be in comp_collision_object_script or something?

engine/gamesys/src/dmsdk/gamesys/resources/res_convex_shape.h
JCash: I think it makes sense to keep it as a resource. I don't think we want to bake it into a particular shape as those are more lightweight and less reusable. The custom shape feels more relevant to have as a standalone resource.

engine/engine/src/engine.cpp
```diff
@@ -311,10 +311,10 @@ namespace dmEngine
...
         m_InputBuffer.SetCapacity(64);
         m_ResourceTypeContexts.SetCapacity(31, 64);
-
-        m_PhysicsContext.m_Context3D = 0x0; // it'a union
-        m_PhysicsContext.m_Debug = false;
-        m_PhysicsContext.m_3D = false;
+        m_PhysicsContextBox2D.m_Context = 0x0;
+        m_PhysicsContextBox2D.m_BaseContext.m_PhysicsType = dmGameSystem::PHYSICS_ENGINE_BOX2D;
+        m_PhysicsContextBullet3D.m_Context = 0x0;
+        m_PhysicsContextBullet3D.m_BaseContext.m_PhysicsType = dmGameSystem::PHYSICS_ENGINE_BULLET3D;
```
JCash, line 317: These enums are a little bit surprising though. I wouldn't have thought we needed such info after the split?

engine/gamesys/src/gamesys/components/box2d/comp_collision_object_box2d.cpp
```diff
@@ -0,0 +1,1349 @@
...
+        g_PhysicsAdapter->m_CreateJoint            = CreateJointBox2D;
+        g_PhysicsAdapter->m_DestroyJoint           = DestroyJointBox2D;
+        g_PhysicsAdapter->m_GetJointParams         = GetJointParamsBox2D;
+        g_PhysicsAdapter->m_GetJointType           = GetJointTypeBox2D;
+        g_PhysicsAdapter->m_SetJointParams         = SetJointParamsBox2D;
+        g_PhysicsAdapter->m_GetJointReactionForce  = GetJointReactionForceBox2D;
+        g_PhysicsAdapter->m_GetJointReactionTorque = GetJointReactionTorqueBox2D;
+        g_PhysicsAdapter->m_GetPhysicsEngineType   = GetPhysicsEngineTypeBox2D;
+    }
+}
```
JCash, line 1349: Next part in the saga may be to fully move the collision objects to an extension as well, so it's not initialized by gamesys.cpp but with a static constructor here.
But that's for an upcoming PR.

engine/gamesys/src/gamesys/components/comp_collision_object.cpp
```diff
@@ -57,219 +36,197 @@ namespace dmGameSystem
...
-    };
+        PhysicsContext* physics_context = (PhysicsContext*) params.m_Context;
+        if (physics_context->m_PhysicsType == PHYSICS_ENGINE_BOX2D)
+        {
+            return CompCollisionObjectBox2DDeleteWorld(params);
+        }
+        else if (physics_context->m_PhysicsType == PHYSICS_ENGINE_BULLET3D)
+        {
+            return CompCollisionObjectBullet3DDeleteWorld(params);
+        }
```
JCash, line 63: we're almost there! :)

JCash (approved):
Lgtm

engine/engine/src/engine.cpp
JCash: Nvm, I saw the code in collision object cpp 👍

PR #10155, opened 2025-02-04
Fix single separator issue on toolbar

sprocketc:
I did not expect #10150 to break something, but a single separator is visible when we focus on the asset tree. We need to also remove separators when they are the first element of a toolbar. I hadn't noticed that we clear the toolbar when we focus on the asset tree until now.


matgis (approved):
LGTM!

PR #10160, opened 2025-02-04
Fix some warnings during editor tests

matgis:
* Fix some warnings that appear in the output from the editor tests.
* Update extension URLs for extensions that have an editor component to the latest released version (except `extension-rive`, which in its latest release causes editor test failures).


editor/project.clj
```diff
@@ -278,12 +278,12 @@
...
                                                     "-Ddefold.extension.rive.url=https://github.com/defold/extension-rive/archive/refs/tags/2.4.0.zip"
                                                     "-Ddefold.extension.simpledata.url=https://github.com/defold/extension-simpledata/archive/refs/tags/v1.1.0.zip"
                                                     "-Ddefold.extension.spine.url=https://github.com/defold/extension-spine/archive/refs/tags/3.6.5.zip"
-                                                    "-Ddefold.extension.teal.url=https://github.com/defold/extension-teal/archive/main.zip"
-                                                    "-Ddefold.extension.texturepacker.url=https://github.com/defold/extension-texturepacker/archive/refs/tags/2.1.zip"
+                                                    "-Ddefold.extension.teal.url=https://github.com/defold/extension-teal/archive/refs/tags/v1.2.zip"
+                                                    "-Ddefold.extension.texturepacker.url=https://github.com/defold/extension-texturepacker/archive/refs/tags/2.1.2.zip"
```
matgis, line 286: The editor tests will use these revisions of our extensions in the integration tests.

I've updated all of them to use a particular version number, so that none of them are on a pre-release version. We should try to keep these URLs updated whenever we release new versions of these extensions.

sprocketc (approved):
Looks good. I assume removing `squirrel.spinejson` was intentional since there is no reference to it on the tests, except in `checked.spinescene`.

matgis:
It’s not actually removed, it’s just that the new content is all on a single line. But thanks for bringing my attention to it! I had to double-check. 😅

PR #10162, opened 2025-02-05
Insert shader sources if no crosscompiling required

Jhonnyg:
Fixes regression on linux due to spirv-cross giving incompatible shaders (for linux!)

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/shader/ShaderCompilePipeline.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/shader/ShaderCompilePipeline.java
@@ -294,7 +294,7 @@ public class ShaderCompilePipeline {
 
         boolean isLanguageGLSL = shaderLanguageIsGLSL(shaderLanguage);
 
-        if (isLanguageGLSL && module.shaderInfo.version == version) {
+        if (isLanguageGLSL && module.shaderInfo != null && module.shaderInfo.version == version) {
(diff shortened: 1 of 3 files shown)


engine/engine/content/builtins/materials/model.fp
```diff
@@ -1,3 +1,5 @@
+#version 140
```
Jhonnyg, line 1: Sneaky fix!

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/shader/ShaderCompilePipeline.java
```diff
@@ -292,9 +292,17 @@ public byte[] crossCompile(ShaderDesc.ShaderType shaderType, ShaderDesc.Language
...
         assert module != null;
 
-        if (shaderLanguage == ShaderDesc.Language.LANGUAGE_SPIRV) {
+        boolean isLanguageGLSL = shaderLanguageIsGLSL(shaderLanguage);
+
+        if (isLanguageGLSL && module.shaderInfo != null && module.shaderInfo.version == version) {
+            // The input shader is already valid GLSL code in the output version we are trying to produce,
+            // no need to crosscompile!
+            return module.desc.source.getBytes();
+        } else if (shaderLanguage == ShaderDesc.Language.LANGUAGE_SPIRV) {
```
Jhonnyg, line 301: Now that I think about it, we won't benefit from shader optimizations from the pipeline now since we are sidestepping the SPIR-v pipeline.. Hm, something to think about

JCash (approved):

PR #10165, opened 2025-02-06
Updated gamepad input mapping for 8BitDo Pro on Linux

Domarius:
Updated inputs on Linux, looks like something changed in Linux's gamepad support:
- Both stick axiis have been inverted
- the dpad buttons are now a higher index


britzl:
It is likely that the glFW update changed something on Linux. @Jhonnyg did a remapping on Windows and we probably need the same on Linux.

Domarius:
> It is likely that the glFW update changed something on Linux. @Jhonnyg did a remapping on Windows and we probably need the same on Linux.

Ok I can do the 3 types of gamepads I have here, I wonder if the changes should be carried across to the other gamepads under Linux...

britzl:
> Ok I can do the 3 types of gamepads I have here

Did all of them change their mappings from what you had previously?

Domarius:
> > Ok I can do the 3 types of gamepads I have here
> 
> Did all of them change their mappings from what you had previously?

I think I did it correctly - I just now pushed the remaining gamepad mappings into the pull request.
And yes, they all changed in the same way. Makes me think we can "cheat" and update all the other gamepads for Linux manually by directly editing the file.

PR #10170, opened 2025-02-06
Add context menu to scene view

sprocketc:
Introduce a context menu to scene view.

context-menu

### Technical changes
Simply registering a context menu wouldn't work properly in this case. We need to explicitly show it on certain conditions, and make it play nice with element selection.

### DIFF
--- a/editor/src/clj/editor/scene_selection.clj
+++ b/editor/src/clj/editor/scene_selection.clj
@@ -144,7 +144,7 @@
                          (let [node ^Node (:target action)
                                scene ^Scene (.getScene node)
                                context-menu (ui/init-context-menu! ::scene-selection-menu scene)]
-                           (.show context-menu node ^double (:screen-x action) ^double  (:screen-y action))))
+                           (.show context-menu node ^double (:screen-x action) ^double (:screen-y action))))
                        nil)
       :mouse-released (do
                         (g/transact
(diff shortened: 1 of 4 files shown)


editor/src/clj/editor/scene_selection.clj
```diff
@@ -90,15 +82,26 @@
...
                                    prev-selection (g/node-value controller :prev-selection)
                                    prev-selection-set (set prev-selection)]
                                (into [] (concat (filter (complement selection-set) prev-selection) (filter (complement prev-selection-set) selection)))))
+
+                           contextual?
+                           (fn [selection]
+                             (let [prev-selection (g/node-value controller :prev-selection)]
+                               (if (some #(= (first selection) %) prev-selection)
+                                 prev-selection
+                                 selection)))
```
sprocketc, line 102: If the element under the mouse is already selected, we should maintain the previous selection to avoid deselecting other elements.

editor/src/clj/editor/scene_selection.clj
```diff
@@ -133,8 +136,12 @@
                            (g/set-property self :current cursor-pos)
                            (g/set-property self :mode mode)
                            (g/set-property self :toggle? toggle?)
+                           (g/set-property self :contextual? contextual?)
                            (g/set-property self :prev-selection (g/node-value self :selection))))
                        (select self op-seq mode toggle?)
+                       (when contextual?
+                         (-> (ui/init-context-menu! ::scene-selection-menu (.getScene (:parent action)))
+                             (.show (:parent action) (:screen-x action) (:screen-y action))))
```
sprocketc, line 144: Simply registering a context menu wouldn't work in this case properly. We need to explicitly show it on certain conditions, and make it work properly along with element selection.

vlaaad (approved):
LGTM!

matgis (approved):
Nice! LGTM!

PR #10176, opened 2025-02-07
Fix initial position of visibility toggle

sprocketc:
To reproduce
- Open a scene and make a node invisible
- Resize the right sidebar to make the outline tree scrollable (horizontally)
- Close and reopen the scene

The toggle button is not properly positioned until you scroll.


vlaaad (approved):
LGTM!

PR #10183, opened 2025-02-09
Visibility fixes and enhancements

### LINKED ISSUE #9818: Show/hide selected objects (Ctrl+E) does not work with models component.

sprocketc:
Hide children of invisible nodes without toggling their visibility state, fix hiding models and double clicking on the visibility toggle.

Fixes #9818

### DIFF
--- a/editor/src/clj/editor/outline_view.clj
+++ b/editor/src/clj/editor/outline_view.clj
@@ -134,6 +134,7 @@
                 :child-error? (boolean (some :child-error? data))
                 :child-overridden? (boolean (some :child-overridden? data))
                 :hideable? (contains? outline-name-paths (rest node-outline-key-path))
+                :hidden-parent? (scene-visibility/hidden-outline-key-path? hidden-node-outline-key-paths node-outline-key-path)
                 :scene-visibility (if (contains? hidden-node-outline-key-paths node-outline-key-path)
                                     :hidden
                                     :visible))]
@@ -489,7 +490,8 @@
                             (ui/add-style! "visibility-toggle")
                             (.addEventFilter MouseEvent/MOUSE_PRESSED ui/ignore-event-filter)
                             (AnchorPane/setRightAnchor 0.0))
-        text-label (Label.)
+        text-label (doto (Label.)
+                     (ui/bind-double-click! :open))
         h-box (doto (HBox. 5 (ui/node-array [image-view-icon text-label]))
                 (ui/add-style! "h-box")
                 (AnchorPane/setRightAnchor 0.0)
@@ -508,7 +510,7 @@
                        (proxy-super setGraphic nil)
                        (proxy-super setContextMenu nil)
                        (proxy-super setStyle nil))
-                     (let [{:keys [label icon link color outline-error? outline-overridden? outline-reference? outline-show-link? parent-reference? child-error? child-overridden? scene-visibility hideable? node-outline-key-path]} item
+                     (let [{:keys [label icon link color outline-error? outline-overridden? outline-reference? outline-show-link? parent-reference? child-error? child-overridden? scene-visibility hideable? node-outline-key-path hidden-parent?]} item
                            icon (if outline-error? "icons/32/Icons_E_02_error.png" icon)
                            show-link? (and (some? link)
                                            (or outline-reference? outline-show-link?))
@@ -544,6 +546,9 @@
                        (if hideable?
                          (ui/add-style! this "hideable")
                          (ui/remove-style! this "hideable"))
+                       (if hidden-parent?
+                         (ui/add-style! this "hidden-parent")
+                         (ui/remove-style! this "hidden-parent"))
                        (if hidden?
                          (ui/add-style! this "scene-visibility-hidden")
                          (ui/remove-style! this "scene-visibility-hidden")))))))]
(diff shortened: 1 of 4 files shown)


editor/src/clj/editor/outline_view.clj
```diff
@@ -489,7 +490,8 @@
                             (ui/add-style! "visibility-toggle")
                             (.addEventFilter MouseEvent/MOUSE_PRESSED ui/ignore-event-filter)
                             (AnchorPane/setRightAnchor 0.0))
-        text-label (Label.)
+        text-label (doto (Label.)
+                     (ui/bind-double-click! :open))
```
sprocketc, line 494: Moved the double click handler on the label to avoid opening the resource when we double click on the visibility toggle. I think this move makes sense regardless of the bug.

sprocketc:
This PR changes the behavior of `editor.scene-visibility` to hide children of invisible parents, without adding their `node-outline-key-path` to `hidden-outline-name-paths`. The outline tree was also modified to change the opacity of those children without displaying the toggle. I think this is more intuitive, but it might affect the current workflow of some users, so I would appreciate some feedback on this. Not persisting the visibility state when we close the editor, reduces the chance of frustration. It's probably also less prone to bugs similar to #9818, that is indirectly fixed by this PR (probably introduced [here](https://github.com/defold/defold/pull/8769/files#diff-6046ca4ceae4f8cffb0e51b10ea675540542b33302ca3f504b334e1140cd25ffR324-R342)).

editor/src/clj/editor/outline_view.clj
```diff
@@ -134,6 +134,7 @@
                 :child-error? (boolean (some :child-error? data))
                 :child-overridden? (boolean (some :child-overridden? data))
                 :hideable? (contains? outline-name-paths (rest node-outline-key-path))
+                :hidden-parent? (scene-visibility/hidden-outline-key-path? hidden-node-outline-key-paths node-outline-key-path)
```
sprocketc, line 137: Do we have a preference on boolean naming? I am leaning towards using a question mark on predicate methods only, but I don't mind either way. Removing `?` also feels inadequate in some cases, so having a general rule will make the decision easier.

matgis: Yes, these days we tend to reserve the `?` suffix for predicates, and often prefixing booleans with words like `is-`, `has-`, `should-`, and `will-` when appropriate.

I don't really mind us following the convention of the surrounding code like this, though.

editor/src/clj/editor/scene_visibility.clj
```diff
@@ -154,20 +154,12 @@
...
-                                                         (outline-name-paths-by-selection-state true)))
-
-  (output unselected-outline-name-paths OutlineNamePaths (g/fnk [outline-name-paths-by-selection-state]
-                                                           (outline-name-paths-by-selection-state false)))
+  (output selected-outline-name-paths OutlineNamePaths (g/fnk [outline-selection]
+                                                         (set (into [] (keep outline-selection-entry->outline-name-path) outline-selection))))
+
+  (output unselected-outline-name-paths OutlineNamePaths (g/fnk [selected-outline-name-paths outline-name-paths]
+                                                           (set/difference outline-name-paths selected-outline-name-paths)))
+
```
matgis, line 162: I think we should `:cache` these outputs now that they are no longer trivial to evaluate?

editor/src/clj/editor/scene_visibility.clj
```diff
@@ -260,9 +252,7 @@
            (g/node-value scene-visibility :active-scene-resource-node evaluation-context))
   (run [scene-visibility user-data]
        (let [{:keys [node-outline-key-path]} user-data
-             outline-name-paths (g/node-value scene-visibility :outline-name-paths)
-             name-path-child? #(iutil/seq-starts-with? % (rest node-outline-key-path))
-             name-paths-to-toggle (set (filter name-path-child? outline-name-paths))]
+             name-paths-to-toggle #{(vec (rest node-outline-key-path))}]
```
matgis, line 255: Would it be better to use `subvec` here?

matgis (changes requested):
I think we should perhaps `:cache` the `OutlineNamePaths` outputs of `SceneVisibilityNode` now that they are not trivial to produce.

matgis (approved):
Looks good to me! 👍

PR #10188, opened 2025-02-10
Improve build performance

vlaaad:
Technical notes:

There is no noticeable improvement in build time or memory usage, though... But the places appeared in the memory allocations profiler...

Related to #10113


vlaaad:
Actually, one noticeable change is that the average memory use during the build decreased from 12.5Gb to 11.3Gb.

matgis (approved):
Looks good to me! Some nice cleanup in there, even if it doesn't have a direct effect on performance, it gives us cleaner measurements. 👍

PR #10190, opened 2025-02-10
Add support for Android devices with 16 KB memory page sizes

### LINKED ISSUE #10058: Support 16 KB page sizes on Android
**Task (REQUIRED):**
https://developer.android.com/guide/practices/page-sizes#cmake

Android build should work on new devices with 16 KB page size

AGulev:
Support for 16 KB memory pages will soon be a requirement for the Google Play Store.  

The engine now includes an extra flag to ensure compatibility with 16 KB page sizes while remaining fully compatible with 4 KB page size Android devices.  

More information is available here:  
[Support 16 KB page sizes](https://developer.android.com/guide/practices/page-sizes)  

Fix https://github.com/defold/defold/issues/10058


engine/profiler/src/profiler_proc_utils.cpp
```diff
@@ -184,7 +184,7 @@ uint64_t dmProfilerExt::GetProcMemoryUsage()
     }
     fclose(fp);
 
-    long page_size = sysconf( _SC_PAGESIZE);
+    long page_size = sysconf(_SC_PAGESIZE);
```
AGulev, line 187: I just removed an extra space (it is easier to search for this function if it's without space)

build_tools/waf_dynamo.py
```diff
@@ -276,10 +276,11 @@ def getAndroidCompileFlags(target_arch):
         return ['-D__ARM_ARCH_5__', '-D__ARM_ARCH_5T__', '-D__ARM_ARCH_5E__', '-D__ARM_ARCH_5TE__', '-DGOOGLE_PROTOBUF_NO_RTTI', '-march=armv7-a', '-mfloat-abi=softfp', '-mfpu=vfp', '-fvisibility=hidden']
 
 def getAndroidLinkFlags(target_arch):
-    if 'arm64' == target_arch:
-        return ['-Wl,--no-undefined', '-Wl,-z,noexecstack', '-landroid', '-fpic', '-z', 'text']
+    common_flags = ['-Wl,--no-undefined', '-Wl,-z,noexecstack', '-landroid', '-fpic', '-z', 'text']
```
AGulev, line 279: The most of the flags are common, it's a bit hard to read and modify it when we have these in two places

AGulev (commented):
I checked all the places where we use `PAGE_SIZE` or functions that may require alignment:  

- `mmap()` and `mprotect()`  
- `posix_memalign()`  
- `memalign()`  
- `aligned_alloc()`  
- `mlock()` and `mlockall()`  
- `madvise()` and `mincore()`  
- `shmat()`  
- `dlopen()`  

We don't use most of these, but for the functions we do use, it seems like we don’t depend on the page size—only on our own required alignment.  

Of course, it's possible that I misunderstood or missed something, especially since I can't run tests on an actual Android device. However, I tested the application on a simulator with a 16 KB page size, and it works fine.

JCash (approved):

PR #10198, opened 2025-02-11
Fixed an issue where the `sys.open_url()` function did not work

### LINKED ISSUE #10164: sys.open_url() do not work in Defold version 1.9.7 (iOS build)
sys.open_url() do not work in Defold version 1.9.7

**To Reproduce (REQUIRED)**
Build any project for iOS with exec of
sys.open_url("https://www.googlr.com")

Nothing to happens

  - Version 1.9.7

**Platforms (REQUIRED):**
 iOS

Please attach a minimal project where the bug is reproduced. This will greatly help the person trying to investigate and fix the bug.

**Logs (OPTIONAL):**

**Workaround (OPTIONAL):**

**Screenshots (OPTIONAL):**

AGulev:
Fixed an issue on iOS where a deprecated function was used for URL opening, causing it to not work on iOS and iPadOS.

Fix https://github.com/defold/defold/issues/10164

### DIFF
--- a/engine/dlib/src/dlib/sys_cocoa.mm
+++ b/engine/dlib/src/dlib/sys_cocoa.mm
@@ -163,24 +163,11 @@ namespace dmSys
 
     Result OpenURL(const char* url, const char* target)
     {
-        NSString* ns_url = [NSString stringWithUTF8String:url];
-        NSURL* nsURL = [NSURL URLWithString:ns_url];
-
-        __block Result result = RESULT_UNKNOWN;
-        dispatch_semaphore_t sema = dispatch_semaphore_create(0);
-
-        [[UIApplication sharedApplication] openURL:nsURL
-                                           options:@{}
-                                 completionHandler:^(BOOL success) {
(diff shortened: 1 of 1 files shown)


engine/dlib/src/dlib/sys_cocoa.mm
```diff
@@ -163,16 +163,11 @@ Result GetLogPath(char* path, uint32_t path_len)
...
-        }
-        else
-        {
-            return RESULT_UNKNOWN;
-        }
+        NSString* ns_str = [NSString stringWithUTF8String:url];
+        NSURL* ns_url = [NSURL URLWithString:ns_str];
+        UIApplication *app = [UIApplication sharedApplication];
+        [app openURL:ns_url options:@{} completionHandler:nil];
+        return [app canOpenURL:ns_url] ? RESULT_OK : RESULT_UNKNOWN;
```
AGulev, line 170: The new `openURL` function is asynchronous, so we can't get a result the same way we did before, but we can use another function to check if URL can be opened or not, and return this info instead.

britzl (approved):

PR #10200, opened 2025-02-11
Zoom at pointer position on mouse wheel

sprocketc:
When 2d-mode is enabled, zooming will now take the pointer position into account, so that the focus point remains the same on scroll. We can use the mouse position to target the point that we want to zoom into, and avoid panning as a separate navigation step. Holding `ALT` will activate the previous behavior. On 3d-mode the opposite is applied. We maintain a fixed camera target on mouse wheel (as we used to), and pan to the pointed position on `ALT` down.

zoom-at-pointer-position

### DIFF
--- a/editor/src/clj/editor/camera.clj
+++ b/editor/src/clj/editor/camera.clj
@@ -388,9 +388,9 @@
   [^Camera camera ^Camera prev-camera ^Region viewport [^double x ^double y]]
   (let [focus ^Vector4d (:focus-point camera)
         point (camera-project camera viewport (Point3d. (.x focus) (.y focus) (.z focus)))
-        screen-z (.z point)
-        world (camera-unproject camera viewport x y screen-z)
-        delta (camera-unproject prev-camera viewport x y screen-z)]
+        prev-point (camera-project prev-camera viewport (Point3d. (.x focus) (.y focus) (.z focus)))
+        world (camera-unproject camera viewport x y (.z point))
+        delta (camera-unproject prev-camera viewport x y (.z prev-point))]
     (.sub delta world)
     (assoc (camera-move camera (.x delta) (.y delta) (.z delta))
            :focus-point (doto focus (.add delta)))))
--- a/editor/src/clj/editor/camera.clj
+++ b/editor/src/clj/editor/camera.clj
@@ -384,6 +384,17 @@
     (assoc (camera-move camera (.x delta) (.y delta) (.z delta))
       :focus-point (doto focus (.add delta)))))
 
+(defn pan-at-pointer-position
+  [^Camera camera ^Camera prev-camera ^Region viewport [^double x ^double y]]
+  (let [focus ^Vector4d (:focus-point camera)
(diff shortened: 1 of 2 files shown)


sprocketc:
An easy way to test if zooming at pointer position works as intended, is to target something on the scene with your mouse, and then zoom in and out. Your target should always remain under the pointer.

editor/src/clj/editor/camera.clj
```diff
@@ -384,6 +384,17 @@
...
+  [^Camera camera ^Camera prev-camera ^Region viewport [^double x ^double y]]
+  (let [focus ^Vector4d (:focus-point camera)
+        point (camera-project camera viewport (Point3d. (.x focus) (.y focus) (.z focus)))
+        screen-z (.z point)
+        world (camera-unproject camera viewport x y screen-z)
+        delta (camera-unproject prev-camera viewport x y screen-z)]
+    (.sub delta world)
+    (assoc (camera-move camera (.x delta) (.y delta) (.z delta))
+           :focus-point (doto focus (.add delta)))))
+
```
sprocketc, line 397: Panning depends on the diff of the current camera, and the camera before calling `dolly`.

editor/src/clj/editor/camera.clj
```diff
@@ -611,34 +622,44 @@
...
-                                       (= :track movement)
-                                       (track viewport last-x last-y x y)
-                                       (= :tumble movement)
-                                       (tumble last-x last-y x y))
-
-                                     true
-                                     filter-fn)]
+(defn mode-2d? [camera]
+  (and (= 1.0 (some-> camera camera-view-matrix (.getElement 2 2)))
+       (= :orthographic (:type camera))))
```
sprocketc, line 627: Rulers will now be hidden when we activate perspective.

editor/src/clj/editor/rulers.clj
```diff
@@ -189,80 +188,77 @@
...
-                           (doseq [[y-w y-s] ys
-                                   :let [text (String/format nil nf (to-array [y-w]))]]
-                             (label! vertex-buffer (- x 2) (- y-s 1) text false)))
-                         ;; This is unfortunate but should not have any severe consequence
-                         (catch BufferOverflowException _))
-                       ;; end of triangles; start of lines
-                       (let [tri-count (count vertex-buffer)]
-                         (try
-                           (let [[r g b a] marker-color]
-                             (let [x0 width
```
sprocketc, line 235: Removing the nested let was the only change.

matgis (dismissed):
Works great! Good job getting it to co-exist alongside the focus-point based 3D navigation too! ❤️

matgis (approved):

vlaaad (approved):
LGTM! Panning became so much nicer!

vlaaad:
Do we have an issue for this? It would be a shame if such a welcome change would not be presented in the release notes...

AGulev:
it's enough to assign PR itse;f to the project

PR #10207, opened 2025-02-12
Add 2d mode toolbar button

sprocketc:
Introduced a 2d-mode button that toggles the sate of the scene camera. `realign-camera` now also toggles this state, instead of just switching to 2d. The 3d camera is cached, so we can switch to the previous 3d view.

2d-mode


editor/src/clj/editor/scene.clj
```diff
@@ -989,21 +989,28 @@
...
-    (g/transact
-      (g/set-property camera-node :local-camera end-camera)))
-  nil)
+(defn set-camera!
+  ([camera-node start-camera end-camera animate?]
+   (set-camera! camera-node start-camera end-camera animate? nil))
+  ([camera-node start-camera end-camera animate? on-animation-end]
+   (if animate?
+     (let [duration 0.5]
+       (g/transact (g/set-property camera-node :animating true))
```
sprocketc, line 998: Maintaining this state was necessary to fix issues related to spamming camera-realign, and also some flickering issues of the toolbar buttons.

editor/src/clj/editor/scene.clj
```diff
@@ -989,21 +989,28 @@
...
+                 (fn [^double t]
+                   (let [t (- (* t t 3) (* t t t 2))
+                         cam (c/interpolate start-camera end-camera t)]
+                     (g/transact
+                       (g/set-property camera-node :local-camera cam))))
+                 (fn []
+                   (g/transact
+                     [(g/set-property camera-node :local-camera end-camera)
+                      (g/set-property camera-node :animating false)])
+                   (ui/user-data! (ui/main-scene) ::ui/refresh-requested? true)
```
sprocketc, line 1009: Dirty, but we need to update the toolbar when the animation ends.

editor/src/clj/editor/scene.clj
```diff
@@ -1051,21 +1049,74 @@
...
+  [view animate?]
+  (let [camera-node (view->camera view)
+        local-cam (g/node-value camera-node :local-camera)
+        is-perspective (= (:type local-cam) :perspective)]
+    (g/transact (g/set-property camera-node :cached-3d-camera local-cam))
+    (let [end-camera (cond-> local-cam
+                       is-perspective c/camera-perspective->orthographic
+                       :always c/camera-orthographic-realign
+                       is-perspective (c/camera-orthographic->perspective c/fov-y-35mm-full-frame))]
+      (set-camera! camera-node local-cam end-camera animate? #(set-camera-type! view :orthographic)))))
```
sprocketc, line 1096: This looks weird, but improves the animation from perspective 3d -> 2d, by animating to a perspective camera and setting its type to orthographic on animation end.

editor/src/clj/editor/scene.clj
```diff
@@ -1051,21 +1049,74 @@
...
         (set-camera! camera-controller old-camera new-camera false)))))
 
+(defn- sync-camera-position
+  [^Camera camera-a ^Camera camera-b viewport]
+  (let [focus ^Vector4d (:focus-point camera-b)
+        point (c/camera-project camera-b viewport (Point3d. (.x focus) (.y focus) (.z focus)))
+        world (c/camera-unproject camera-a viewport (.x point) (.y point) (.z point))
+        delta (c/camera-unproject camera-b viewport (.x point) (.y point) (.z point))]
+    (.sub delta world)
+    (c/camera-move camera-a (.x delta) (.y delta) (.z delta))))
```
sprocketc, line 1059: When we navigate on 2d and then switch to 3d, we attempt to move the camera to the updated location, but keep the rotation of the cached 3d camera. Not sure if that's the best way to handle this.

editor/src/clj/editor/scene.clj
```diff
@@ -1051,21 +1049,74 @@
...
+          (g/node-value :animating)))
 
 (handler/defhandler :realign-camera :global
   (active? [app-view evaluation-context]
            (active-scene-view app-view evaluation-context))
-  (run [app-view] (when-let [view (active-scene-view app-view)]
-                    (realign-camera view true))))
+  (enabled? [app-view] (not (camera-animating? app-view)))
+  (run [app-view] (some-> (active-scene-view app-view) 
+                          (realign-camera true))))
```
sprocketc, line 1119: `:realign-camera` is probably not a good name anymore, but I didn't want to add a new shortcut. Using `period` to toggle the camera is fine in my opinion, and migrating to a new key seems like an overkill.

matgis (approved):
Works great! Ship it! ❤️

vlaaad (approved):
It works great! Ship it!

PR #10209, opened 2025-02-12
Fixed an issue where the LiveUpdate archive path could become invalid after an app update

### LINKED ISSUE #9664: Mounted liveupdate archives are being unmounted after app update on iOS
On iOS, after app update, absolute path to app home directory is changing, like
`/var/mobile/Containers/Data/Application/0E51B748-1CAE-4B41-9D42-6AA70BBCFAD9`
to
`/var/mobile/Containers/Data/Application/20AAFA4F-BEF8-4CF8-9326-7551C938D0B3`
But looks like liveupdate doesn't take it into account and store absolute path, and we got error like this during first run on new version
` ERROR:RESOURCE: Could not resolve a mount path '/var/mobile/Containers/Data/Application/0E51B748-1CAE-4B41-9D42-6AA70BBCFAD9/Documents/http-storage/archive_name'`

**To Reproduce (REQUIRED)**
1. Create two iOS builds of some app with possibility to mount liveupdate archive
2. Run lowest version and mount liveupdate archive
3. Install highest version above lowest and run
4. Check mounted archives / errors in logs

Archives should be mounted after install update (if it was mounted before and we don't have mount_on_start = false)

  - version 1.6.2

**Platforms (REQUIRED):**
 - Platforms: iOS
 - OS: wide range of iOS versions (as example iPhone OS 17.6.1)
 - Device: wide range of iPhone/iPad devices (as example iPhone 14)

AGulev:
Use a dynamic path for the `HOME` part of the LiveUpdate mount path to prevent iOS from breaking it when the app `UUID` changes after an update. 

Fix https://github.com/defold/defold/issues/9664

### DIFF
--- a/engine/dlib/src/dlib/sys.h
+++ b/engine/dlib/src/dlib/sys.h
@@ -237,6 +237,19 @@ namespace dmSys
      */
     Result GetLogPath(char* path, uint32_t path_len);
 
+    /**
+     * Get path to home directory.
+     * Platform notes:
+     * <ul>
+     * <li>iOS: In iOS, the home directory is the application’s sandbox directory
+     * <li>MacOs: In macOS, it’s the application’s sandbox directory, or the current user’s home directory if the application isn’t in a sandbox
+     * </ul>
+     * @param path
+     * @param path_len
+     * @return
+     */
+    Result GetHomePath(char* path, uint32_t path_len);
+
 
     /**
      * Get system information
--- a/engine/dlib/src/dlib/sys_cocoa.mm
+++ b/engine/dlib/src/dlib/sys_cocoa.mm
@@ -93,6 +93,17 @@ namespace dmSys
         }
     }
 
+    Result GetHomePath(char* path, uint32_t path_len)
+    {
+        NSString* p = NSHomeDirectory();
+        const char* s = [p UTF8String];
+
+        if (dmStrlCpy(path, s, path_len) >= path_len)
+            return RESULT_INVAL;
+
+        return RESULT_OK;
+    }
+
 #if defined(DM_PLATFORM_IOS)
 
     static NetworkConnectivity g_NetworkConnectivity = NETWORK_DISCONNECTED;
--- a/engine/dlib/src/dlib/sys_posix.cpp
+++ b/engine/dlib/src/dlib/sys_posix.cpp
@@ -599,7 +599,7 @@ namespace dmSys
 
 
 // NOTE: iOS/Mac implementation in sys_cocoa.mm
-#if !defined(__MACH__) && !defined(DM_PLATFORM_IOS)
+#if !defined(__MACH__) && !defined(DM_PLATFORM_IOS) && !defined(__ANDROID__)
     Result GetHomePath(char* path, uint32_t path_len)
     {
         const char* home = NULL;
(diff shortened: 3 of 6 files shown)


engine/dlib/src/dlib/sys.h
```diff
@@ -237,6 +237,19 @@ namespace dmSys
...
+     * Platform notes:
+     * <ul>
+     * <li>iOS: In iOS, the home directory is the application’s sandbox directory
+     * <li>MacOs: In macOS, it’s the application’s sandbox directory, or the current user’s home directory if the application isn’t in a sandbox
+     * </ul>
+     * @param path
+     * @param path_len
+     * @return
+     */
+    Result GetHomePath(char* path, uint32_t path_len);
```
AGulev, line 251: this path needed so we can replace it with a token

engine/dlib/src/dlib/sys_posix.cpp
```diff
@@ -597,6 +598,103 @@ namespace dmSys
...
+        // First, try getDataDir() (API 24+)
+        jmethodID get_data_dir_method = env->GetMethodID(activity_class, "getDataDir", "()Ljava/io/File;");
+        if (get_data_dir_method) {
+            jobject data_dir_obj = env->CallObjectMethod(activity_obj, get_data_dir_method);
+            if (data_dir_obj) {
+                jclass file_class = env->FindClass("java/io/File");
+                jmethodID getPathMethod = env->GetMethodID(file_class, "getPath", "()Ljava/lang/String;");
+                path_obj = (jstring) env->CallObjectMethod(data_dir_obj, getPathMethod);
+            }
+        }
```
AGulev, line 630: This is a preferable way to do it, once we update minimal android version to 24 we will keep only this one

engine/resource/src/resource_mounts.cpp
```diff
@@ -316,6 +344,24 @@ dmResource::Result SaveMounts(HContext ctx, const char* app_support_path)
...
+                char before[1024];
+                dmStrlCpy(before, uri_str, before_len);
+                before[before_len] = '\0';
+
+                char uricopy[1024];
+                const char* rest = save_path + strlen(app_home_path);
+                dmSnPrintf(uricopy, sizeof(uricopy), "%s${HOME_PATH}%s", before, rest);
+                memcpy(uri_str, uricopy, sizeof(uri_str));
+            }
+        }
```
AGulev, line 363: replace HOME part of the "path" with token before save

engine/resource/src/test/wscript
```diff
@@ -201,6 +201,7 @@ def build(bld):
     bld.program(features     = 'cxx test',
                 includes     = '.. ../../proto',
                 use          = 'TESTMAIN DDF DLIB PROFILE_NULL SOCKET THREAD LUA resource',
+                web_libs     = ['library_sys.js'],
```
AGulev, line 204: fix web build tests

engine/dlib/src/dlib/sys_cocoa.mm
```diff
@@ -93,6 +93,17 @@ Result GetApplicationSupportPath(const char* application_name, char* path, uint3
         }
     }
 
+    Result GetHomePath(char* path, uint32_t path_len)
+    {
+        NSString* p = NSHomeDirectory();
+        const char* s = [p UTF8String];
+
+        if (dmStrlCpy(path, s, path_len) >= path_len)
+            return RESULT_INVAL;
```
JCash, line 102: 👍

engine/resource/src/resource_mounts.cpp
```diff
@@ -271,13 +282,27 @@ dmResource::Result LoadMounts(HContext ctx, const char* app_support_path)
         return result;
     }
 
+    char app_home_path[1024];
```
JCash, line 285: Use `DMPATH_MAX_PATH` from dlib/path.h

engine/resource/src/resource_mounts.cpp
```diff
@@ -248,6 +249,16 @@ static dmResource::Result LoadMount(HContext ctx, int priority, const char* name
     return dmResource::RESULT_OK;
 }
 
+const char* UriTeplateReplacer(void* user_data, const char* key)
```
JCash, line 252: typo. I believe it should be "UriTeplateReplacer" -> "UriTe**m**plateReplacer"

engine/resource/src/resource_mounts.cpp
```diff
@@ -271,13 +282,27 @@ dmResource::Result LoadMounts(HContext ctx, const char* app_support_path)
         return result;
     }
 
+    char app_home_path[1024];
+    dmSys::Result app_path_result = dmSys::GetHomePath(app_home_path, sizeof(app_home_path));
```
JCash, line 286: To be honest, I were thinking we were going to use GetApplicationSupportPath() for this?
What's the benefit of the new function?

JCash (dismissed):
lgtm.
Needs fixing some path size constants.

engine/resource/src/resource_mounts.cpp
AGulev: Unfortunately, we can't use `GetApplicationSupportPath` because this is just one of  the subfolders:
`/var/mobile/Containers/Data/Application/{app-uid?}/*subfolder here*`
and there are many subfolders may be used. We need to replace the main path before `{app-uid?}`

AGulev:
closed as not needed
see https://github.com/defold/defold/issues/9664

PR #10213, opened 2025-02-13
Make sure each sampler is present only once when binding textures to samplers with hashes in render scripts

Jhonnyg:
Fixed an issue where multiple samplers could be present in the render contexts "texture bind table" that gets applied during rendering. In some cases this could cause the engine to not be able to bind textures from render scripts because all slots are taken.


engine/render/src/render/render.cpp
```diff
@@ -721,8 +723,19 @@ namespace dmRender
...
+        }
+
+        // If we are unassigning the sampler, but it wasn't found we can exit here.
+        if (texture == 0)
+        {
+            return;
+        }
+
+        // Second pass
+        // We try to find the first empty slot in the list
```
JCash, line 735: Can't we store an empty slot index from the first loop and use it here?

JCash (dismissed):
lgtm
question about avoiding the second loop

PR #10218, opened 2025-02-13
Add a new preference to prevent the browser from opening after `Build HTML5`

### LINKED ISSUE #10181: Make silent html5 build process (without opening new browser tab)
Image

it would be cool if there was an option to silently build the project. and then i could just manually reload the tab in the browser

I think I found a [place in the code 1](https://github.com/defold/defold/blob/dev/editor/src/clj/editor/app_view.clj#L1346) where the project opens in the browser, but I don’t understand anything in the Closure

the problem is that every time you build and rebuild the html5 project, a new browser tab opens, although the host and port remain the same.
this is quite inconvenient, especially if the developer tools are open in the current build tab

perhaps it is worth creating an additional command to build the project, but not open a tab in the browser. and then you could call this method through [http requests](https://github.com/defold/defold/blob/dev/editor/doc/http-api.md#the-command-endpoint) (for example, from neovim)

as a result:
- create a setting in the project build config that would allow building the html5 project without opening a new browser tab
- make it possible to build silently via a request [http request](https://github.com/defold/defold/blob/dev/editor/doc/http-api.md#the-command-endpoint) (like this `http://localhost:[port]/command/build-html5?silent=1`)

AGulev:
Some browsers always create a new tab when opening the same URL.  
In such cases, it may be useful to rebuild the HTML5 build without reopening the browser and instead manually update the page.  
This is now possible with a new checkbox in **Preferences**.  

Fixes https://github.com/defold/defold/issues/10181

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -1348,7 +1348,7 @@ If you do not specifically require different script states, consider changing th
                                (let [url (format "http://localhost:%d%s/index.html" (http-server/port web-server) bob/html5-url-prefix)]
                                  (if (prefs/get prefs [:build :open-html5-build])
                                    (ui/open-url url)
-                                   (console/append-console-entry! nil (format "INFO:The game is available at %s" url))))
+                                   (console/append-console-entry! nil (format "INFO: The game is available at %s" url))))
                                (.close out))))))
 
 (handler/defhandler :rebuild-html5 :global
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -1345,8 +1345,11 @@ If you do not specifically require different script states, consider changing th
                            render-build-error! bob-commands bob-args project changes-view
                            (fn [successful?]
                              (when successful?
-                               (ui/open-url (format "http://localhost:%d%s/index.html" (http-server/port web-server) bob/html5-url-prefix)))
-                             (.close out)))))
+                               (let [url (format "http://localhost:%d%s/index.html" (http-server/port web-server) bob/html5-url-prefix)]
+                                 (if (prefs/get prefs [:build :open-html5-build])
+                                   (ui/open-url url)
+                                   (console/append-console-entry! nil (format "INFO:The game is available at %s" url))))
+                               (.close out))))))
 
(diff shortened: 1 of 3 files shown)


editor/src/clj/editor/app_view.clj
```diff
@@ -1345,8 +1345,11 @@ If you do not specifically require different script states, consider changing th
                            render-build-error! bob-commands bob-args project changes-view
                            (fn [successful?]
                              (when successful?
-                               (ui/open-url (format "http://localhost:%d%s/index.html" (http-server/port web-server) bob/html5-url-prefix)))
-                             (.close out)))))
+                               (let [url (format "http://localhost:%d%s/index.html" (http-server/port web-server) bob/html5-url-prefix)]
+                                 (if (prefs/get prefs [:build :open-html5-build])
+                                   (ui/open-url url)
+                                   (console/append-console-entry! nil (format "INFO:The game is available at %s" url))))
```
vlaaad, line 1351: It pains me to see there is no space between `INFO:` and `The`

editor/src/clj/editor/prefs_dialog.clj
```diff
@@ -100,6 +100,7 @@
   (cond-> [{:name  "General"
             :prefs [{:label "Load External Changes on App Focus" :type :boolean :key [:workflow :load-external-changes-on-app-focus]}
                     {:label "Open Bundle Target Folder" :type :boolean :key [:bundle :open-output-directory]}
+                    {:label "Open browser after `Build HTML5`" :type :boolean :key [:build :open-html5-build]}
```
vlaaad, line 103: The rest of the preferences here use Title Cased Labels.

vlaaad (changes requested):
The code is good, but the messaging could be improved...

editor/src/clj/editor/app_view.clj
AGulev: 😳  I'm sorry

editor/src/clj/editor/prefs_dialog.clj
AGulev: 👍 fixed

vlaaad (approved):
LGTM!

PR #10222, opened 2025-02-14
The default render script was not using an option table when drawing the gui

### LINKED ISSUE #10221: The default render script is not providing the correct options table when drawing gui and text
The default render script draws the gui as the last step of the `update()` function:

```
    -- render GUI
    --
    local camera_gui = state.cameras.camera_gui
    render.set_view(camera_gui.view)
    render.set_projection(camera_gui.proj)

    render.enable_state(graphics.STATE_STENCIL_TEST)
    render.draw(predicates.gui, camera_gui.frustum)
    render.draw(predicates.debug_text, camera_gui.frustum)
    render.disable_state(graphics.STATE_STENCIL_TEST)
    render.disable_state(graphics.STATE_BLEND)
```

The problem is that `render.draw()` is expecting an options table as the second argument and not a frustum. Also, looking at `camera_gui` there is no `frustum` key. There is however an `options` table which is what should be passed to `render.draw()`:

Image

britzl:
The default render script was now passing an options table when drawing the gui and text in the last step of the `update()` function.

Fixes #10221

### DIFF
--- a/engine/engine/content/builtins/render/default.render_script
+++ b/engine/engine/content/builtins/render/default.render_script
@@ -241,8 +241,8 @@ function update(self)
     render.set_projection(camera_gui.proj)
 
     render.enable_state(graphics.STATE_STENCIL_TEST)
-    render.draw(predicates.gui, camera_gui.frustum)
-    render.draw(predicates.debug_text, camera_gui.frustum)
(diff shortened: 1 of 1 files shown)


AGulev (approved):
👍

Issue #8941, opened 2024-05-17
Viewport controls for perspective and view filters only get highlighted when the mouse is over the bright pixels


f2l2pe:
**Describe the bug (REQUIRED)**
Viewport controls for perspective and filters only get highlighted when the mouse is over the bright pixels

**To Reproduce (REQUIRED)**
1. Go to the viewport
2. Put the mouse close to the perspective toggle or view filters

**Expected behavior (REQUIRED)**
The icons should be highlighted like the move, rotate, and scale icons, when the mouse is near the icon and not only when it's over the bright pixels

**Defold version (REQUIRED):**
  - 1.8.1

**Platforms (REQUIRED):**
 - Platforms: Any

**Screenshots (OPTIONAL):**

https://github.com/defold/defold/assets/6012227/8b59a747-6522-4118-8724-6f7e98a97dde

matgis:
Regression from #8855. Before, these icons did not highlight on hover.

britzl:
Fixed by https://github.com/defold/defold/commit/1a074c21277d3355b68c3fe12538a3b32fc86eac

PR #10228, opened 2025-02-17
Fix mouse over state of toolbar buttons

sprocketc:
The fix [here](https://github.com/defold/defold/commit/1a074c21277d3355b68c3fe12538a3b32fc86eac#diff-5b2c365ea12152c04181092a182be946f1197a57a290c1d82dd4c5d302f44759R32-R35) disabled the mouse over state. Changing the color of svg paths would maintain that, and also fix the related #8941.


matgis (approved):
Ship it!

PR #10231, opened 2025-02-17
Update properties on label drag

### LINKED ISSUE #6417: Editor: incremental adjustment of properties values.
It's time-consuming to enter the values of properties manually.

I would like to have an option to adjust values incrementally.

PlayCanvas Editor shows a button when the mouse cursor is over the property value:

https://user-images.githubusercontent.com/193258/154329277-1965c320-4862-4c93-a0cd-51e30cde9f59.mp4

Houdini has a little bit more advanced widget (the video is recorded from [Houdini isn't Scary](https://www.sidefx.com/tutorials/houdini-isnt-scary-part-1-basics/?collection=80)):

https://user-images.githubusercontent.com/193258/154329225-7a185afb-0919-4b53-ae35-f89ebb4610d5.mp4

sprocketc:
Dragging the labels of numeric properties now updates their value depending on the drag direction. When this action is supported, the label will be highlighted on mouse over, and the mouse cursor will switch to horizontal-resize. Left-right / up-down gestures both work. Some property types may override the default step (1), to a more sensible default. Holding `SHIFT` while dragging multiplies this step by 10, and holding `CTRL` divides it by 10, to allow different precision levels on a case by case basis.

drag-label

Fixes #6417


editor/src/clj/editor/atlas.clj
```diff
@@ -283,7 +283,8 @@
                    (concat
                      (g/set-property self :pivot-x (new-value 0))
                      (g/set-property self :pivot-y (new-value 1)))))
-            (dynamic edit-type (g/constantly {:type types/Vec2 :labels ["X" "Y"]})))
+            (dynamic edit-type (g/constantly {:type types/Vec2 :labels ["X" "Y"]
+                                              :precision 0.1})))
```
sprocketc, line 287: I thought about using `:step` at first, but we already use `:precision` for our slider component.

editor/src/clj/editor/properties_view.clj
```diff
@@ -212,6 +214,49 @@
...
+        [prev-x prev-y] @label-drag-position
+        delta-x (- x prev-x)
+        delta-y (- prev-y y)
+        max-delta (if (> (abs delta-x) (abs delta-y)) delta-x delta-y)
+        update-val (cond-> (or precision 1.0)
+                     (.isShiftDown event) (* 10.0)
+                     (.isControlDown event) (* 0.1)
+                     (neg? max-delta) -)]
+    (reset! label-drag-position [x y])
+    (when (> (abs max-delta) 1)
```
sprocketc, line 235: This check avoids some quirky updates, mainly because we support gestures on both axises.

editor/src/clj/editor/gui.clj
```diff
@@ -1665,7 +1665,9 @@
             (set (layout-property-setter inner-radius)))
   (property perimeter-vertices g/Int (default (protobuf/default Gui$NodeDesc :perimeter-vertices))
             (dynamic error (g/fnk [_node-id perimeter-vertices] (validate-perimeter-vertices _node-id perimeter-vertices)))
-            (dynamic edit-type (layout-property-edit-type perimeter-vertices {:type g/Int}))
+            (dynamic edit-type (layout-property-edit-type perimeter-vertices {:type g/Int
+                                                                              :min perimeter-vertices-min
+                                                                              :max perimeter-vertices-max}))
```
sprocketc, line 1670: I think we always use a slider when we have a min & max, except in this case. I didn't change that, because I wanted to support value clamping on mouse drag anyway. If the gap is huge, or if we only have a min or max, it doesn't make sense to use a slider.

editor/src/clj/editor/properties_view.clj
```diff
@@ -223,10 +271,15 @@
...
-                            text-field]
+          (let [drag-update-fn (fn [v update-val]
+                                 (update v index #(properties/round-scalar (+ % update-val))))
+                children (if (seq label-text)
+                           (let [label (doto (Label. label-text)
+                                         (.setMinWidth Region/USE_PREF_SIZE))]
+                             (ui/do-run-later
+                               #(when-not (properties/read-only? (property-fn))
+                                  (make-label-draggable! label (partial handle-label-drag-event! property-fn drag-update-fn update-ui-fn))))
+                             [label text-field])
```
sprocketc, line 282: I couldn't make this work properly without using `do-run-later`.

vlaaad:
Hmm, I find it strange that the slow movement of the mouse (which I do if I want to move things precisely) does not trigger the change of the value:
https://github.com/user-attachments/assets/aeffe4f1-1366-4670-a4d8-bf46c818f696

sprocketc:
> Hmm, I find it strange that the slow movement of the mouse (which I do if I want to move things precisely) does not trigger the change of the value: https://github.com/user-attachments/assets/aeffe4f1-1366-4670-a4d8-bf46c818f696

@vlaaad Good catch. I was trying to test something and forget to revert that. It should be fixed now.

editor/src/clj/editor/properties_view.clj
```diff
@@ -212,6 +214,52 @@
                         ctrls))
     box))
 
+(def label-drag-position (atom nil))
+(def label-drag-op-seq (atom nil))
```
sprocketc, line 218: Should I use `user-data` instead?

vlaaad: Hmm.. I understand that using a global state here is not really a problem because cursor position / mouse events are a global resource as well. Still, it irks me the wrong way. If it's not hard, I'd prefer us to move it to user-data, but I think it's okay this way too.

editor/src/clj/editor/properties_view.clj
```diff
@@ -223,10 +270,15 @@
...
-                              (.setMinWidth Region/USE_PREF_SIZE))
-                            text-field]
+          (let [drag-update-fn (fn [v update-val]
+                                 (update v index #(properties/round-scalar (+ % update-val))))
+                children (if (seq label-text)
+                           (let [label (doto (Label. label-text)
+                                         (.setMinWidth Region/USE_PREF_SIZE))]
+                             (ui/do-run-later
+                               #(when-not (properties/read-only? (property-fn))
+                                  (make-label-draggable! label (partial handle-label-drag-event! property-fn drag-update-fn update-ui-fn))))
```
vlaaad, line 280: We can use `run-later` here since we use an anonymous function. No need to change it here.

vlaaad (approved):
Works great! Ship it!

aglitchman:
I just love you already!!

Because it makes the job so much easier.

PR #10234, opened 2025-02-18
Add TextureProfilesBuilder instead of building it in GameProjectBuilder

### LINKED ISSUE #10217: Make sure that changes in bob arguments which may affect output are taken into account for cachekey calculation
If I build with the `--texture-compression` flag and then build again without this flag, Bob does not rebuild the textures, and the bundle still contains compressed textures.  

**To Reproduce (REQUIRED)**  
1. Build with `--texture-compression` using heavy compression to notice the difference.  
2. Build again without `--texture-compression` and without the `--clean` option.  
3. Observe that the resulting build still uses texture compression.  

Files affected by the `--texture-compression` flag should be rebuilt when the flag is removed.  

- Version 1.9.7

AGulev:
We already have a dependency graph that checks signatures, protocol file formats, output data, etc., as part of the signature, making our cache invalidation more robust.  

I also want `texture_profiles` to be part of this system to ensure that if `texture_profiles` change (as an input for textures), the textures are rebuilt accordingly.  

This step is necessary to fix [issue #10217](https://github.com/defold/defold/issues/10217).

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/TextureProfilesBuilder.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/TextureProfilesBuilder.java
@@ -1,0 +1,60 @@
+package com.dynamo.bob.pipeline;
+
+import com.dynamo.bob.BuilderParams;
+import com.dynamo.bob.CompileExceptionError;
+import com.dynamo.bob.Platform;
+import com.dynamo.bob.ProtoBuilder;
+import com.dynamo.bob.ProtoParams;
+import com.dynamo.bob.Task;
+import com.dynamo.bob.fs.IResource;
+import com.dynamo.graphics.proto.Graphics;
+
+import java.io.IOException;
+import java.util.LinkedList;
+import java.util.List;
+
+@ProtoParams(srcClass = Graphics.TextureProfiles.class, messageClass = Graphics.TextureProfiles.class)
+@BuilderParams(name = "TextureProfile", inExts = {".texture_profiles"}, outExt = ".texture_profilesc")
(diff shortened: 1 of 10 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/TextureProfilesBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+        // Now it should only contain profiles with platform entries
+        // relevant for the target platform...
+        texProfilesBuilder.clearProfiles();
+        texProfilesBuilder.addAllProfiles(newProfiles);
+
+        // Add the current texture profiles to the project, since this
+        // needs to be reachedable by the TextureGenerator.
+        Graphics.TextureProfiles textureProfiles = texProfilesBuilder.build();
+
+        task.output(0).setContent(textureProfiles.toByteArray());
```
AGulev, line 58: Write as output instead of setting it into `Project`

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TextureUtil.java
```diff
@@ -216,6 +221,24 @@ public static TextureProfile getTextureProfileByPath(TextureProfiles textureProf
 
         return null;
     }
+    private static String textureProfilesExt = TextureProfilesBuilder.class.getAnnotation(BuilderParams.class).outExt();
```
AGulev, line 224: not sure if it's a good idea, on the other hand it's not a runtime value, so I think it's ok to cache it like this

britzl (approved):

PR #10252, opened 2025-02-21
Added new window API functions

### LINKED ISSUE #3223: More options to specify executable name and window title separately (DEF-3034)
*Description:*
 From ed2-issues: [(LINK REMOVED)Suggestion/request from original issue;
{quote}Use window title for the default window title - in one project I want a blank default window title, if native extensions are used then bake this window title into the exe and do a rebake if the users changes this in the game.project file... right now when an exe open it first says "GLFW Window" then it says "my_project_name" and then it says "My Project Name" as I change it with DefOS. I'd rather it not need to change like this.

Use project name for the editor to know what project it has open, right now it uses project title for this, but uses the folder name of where the game.project file is in the launcher

Use executable name for the final exe/js/etc. filename for the project. It's annoying to have a window title with spaces and capitalization and a JS file with spaces and capitalization because of it.
{quote}

### LINKED ISSUE #6115: Expose content scale from the window api (DPI)

The `window.get_size()` api will return the window size after it has been dpi scaled for display.
For instance on a retina display set to default scaling the window will be twice the size to the window size configured in the project settings. If your monitor is scaled on Linux the `window.get_size` function reports the scaled window size as well. I have not tested this on windows but would presume it is the same.

Additionally, if you move a window between a retina or scaled display and another display with a different scaling value (or unscaled) the window size / content scale can change, and there is no API event on the `window.listener` to react to this change.

- Add a `window.get_scale()` function that returns a floating point scale factor for the x and y axis  that is the difference between the default DPI of the OS and the currentDPI on each axis., modeled after glfw, or simply a single scaling factor if it is determined that no device scales unevenly across a given axis.
- Add a new event to the `window.listener` window.WINDOW_EVENT_SCALE_CHANGED to signal that the window scaling has changed

A user on the discord server suggested that I read the configuration file to determine what window size was initial requested. This works so long as the projection is default (scaling) when a window is resized, If the projection differs, you could work around this by calculating the scale based on the requested window size and the created window size, though this method breaks down if you then move the window to a new monitor with a different scale, as no event is fired to inform you that you may need to recalculate the scale.

GLFW provides a [`glfwGetWindowContentScale`](https://www.glfw.org/docs/3.3/group__window.html#gaf5d31de9c19c4f994facea64d2b3106c) method that works reasonably well to determine this information on each supported platform. Additionally you would want to hook the [`glfwSetWindowContentScaleCallback`](https://www.glfw.org/docs/3.3/group__window.html#gaf2832ebb5aa6c252a2d261de002c92d6) and provide a user facing window callback to react to this event as well.
On platforms that do not use glfw and that do not apply a content scale, this could be defaulted to 1.0 - e.g. consoles.

britzl:
Added new window.* API functions:

```
window.set_size(width, height)
window.set_position(x, y)
window.set_title("Defold rocks!")
local scale = window.get_display_scale()
```

Fixes #6115 
Fixes #3223

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


engine/gamesys/src/gamesys/scripts/script_window.cpp
```diff
@@ -291,22 +291,99 @@ static int GetSize(lua_State* L)
...
-    assert(top + 1 == lua_gettop(L));
     return 1;
 }
 
+/*# get the display scale
+ *
+ * This returns the content scale of the current display.
+ *
+ * @name window.get_display_scale
+ * @return scale [type:number] The display scale
```
britzl, line 308: I'm a bit torn about this tbh. We do a max(scale_x, scale_y) in out code, but glfw actually returns both the horizontal and vertical scale. I think it is ok for now. If we ever have a need for it in the future we can return scale, scale_x, scale_y.

engine/platform/src/platform_window_glfw.cpp
```diff
@@ -361,6 +366,11 @@ namespace dmPlatform
         }
     }
 
+    void SetWindowPosition(HWindow window, int32_t x, int32_t y)
+    {
+        glfwSetWindowPos(x, y);
```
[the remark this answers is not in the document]
britzl: My expectation is that it will do nothing on Android and iOS. A lot of the window functionality has no meaning on mobile. Like pointer lock or window title.

engine/gamesys/src/gamesys/scripts/script_window.cpp
```diff
@@ -291,22 +291,99 @@ static int GetSize(lua_State* L)
...
 {
     {"set_listener",   SetListener},
     {"set_dim_mode",   SetDimMode},
     {"set_mouse_lock", SetMouseLock},
+    {"set_title",      SetTitle},
     {"get_dim_mode",   GetDimMode},
     {"get_size",       GetSize},
+    {"set_size",       SetSize},
+    {"set_position",   SetPosition},
+    {"get_scale",      GetScale},
```
JCash, line 386: Mismatch between name "get_scale" vs "get_display_scale" in names/comments (also in PR description)

engine/platform/src/platform_window_glfw.cpp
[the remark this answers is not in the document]
JCash: Not sure about the tables (android), as they might have some other layouts. E.g. in landscape etc.
But I guess this should work as a no-op for now.

But I guess documentation of the Lua function should cover what platforms are supported and which are not?

engine/platform/src/platform_window_glfw3.cpp
```diff
@@ -413,6 +413,11 @@ namespace dmPlatform
         }
     }
 
+    void SetWindowTitle(HWindow window, const char* title)
+    {
+        glfwSetWindowTitle(window->m_Window, title);
```
JCash, line 418: Same here, is it supported on all platforms?
What should the documentation say?

engine/platform/src/platform_window_null.cpp
```diff
@@ -106,6 +109,9 @@ namespace dmPlatform
         }
     }
 
+    void SetWindowPosition(HWindow window, int32_t x, int32_t y)
+    {}
+
```
JCash, line 114: Note that this is a change for consoles as well.
Should be a small fix though, but worth remembering.

JCash (changes requested):

engine/platform/src/platform_window_null.cpp
britzl: Yes, we need null implementations on consoles too I suppose?

engine/platform/src/platform_window_glfw3.cpp
britzl: Updated docs to say "desktop only". I suppose we could add some JS code too and do `document.title = "new title";` on HTML5?

engine/platform/src/platform_window_glfw.cpp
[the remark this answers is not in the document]
britzl: Added note that this is for desktop only.

engine/platform/src/platform_window_glfw3.cpp
JCash: However, I think it makes less sense on html5, as the game canvas is deeper in the hierarchy of html5 elements.

JCash (approved):
lgtm

robotigerhub:
Thank you for this feature! Could you tell me in which Defold version this API will be available? I'm not so familiar with how releasing (and GitHub in general) works 😓

PR #10254, opened 2025-02-21
Take Bob’s parameters into consideration to decide which resources should be rebuilt

### LINKED ISSUE #10217: Make sure that changes in bob arguments which may affect output are taken into account for cachekey calculation

AGulev:
Now Bob knows which resources may be affected by its parameters and rebuilds only those resources. For example, if the first bundle is done with `--texture-compression` and the second bundle is done without it, all textures, atlases, and other related resources will be rebuilt.
This should improve the experience when bundling without the `clean` parameter for Bob.

Fix https://github.com/defold/defold/issues/10217
Fix https://github.com/defold/defold/issues/3306

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/cache/test/ResourceCacheKeyTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/cache/test/ResourceCacheKeyTest.java
@@ -14,15 +14,28 @@
 
 package com.dynamo.bob.cache.test;
 
+import static java.util.Map.entry;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
 
 import java.io.IOException;
+import java.lang.reflect.Field;
+import java.security.NoSuchAlgorithmException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
-import java.util.HashMap;
+import java.util.Set;
+import java.util.stream.Collectors;
 
+import com.dynamo.bob.Bob;
+import com.dynamo.bob.BuilderParams;
+import com.dynamo.bob.Project;
+import com.dynamo.bob.fs.DefaultFileSystem;
 import org.junit.After;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
@@ -37,6 +50,7 @@ import com.dynamo.bob.test.util.MockResource;
 
 public class ResourceCacheKeyTest {
 
+	@BuilderParams(name = "DummyBuilder", outExt = "", inExts = {}, paramsForSignature = {"important_option"})
 	private class DummyBuilder extends Builder {
 		private TaskBuilder builder;
 
@@ -75,20 +89,10 @@ public class ResourceCacheKeyTest {
 		return fs.addFile(name, content.getBytes());
 	}
 
-	private Map<String, String> createImportantOptions() {
-		Map<String, String> options = new HashMap<String, String>();
-		options.put("important_option", "yep");
-		return options;
-	}
-	private Map<String, String> createEmptyOptions() {
-		return new HashMap<String, String>();
-	}
-
 	@Before
 	public void setUp() throws Exception {
 		fs = new MockFileSystem();
 		fs.setBuildDirectory("build");
-		ResourceCacheKey.includeOption("important_option");
 	}
 
 	@After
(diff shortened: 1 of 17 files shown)


com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/cache/test/ResourceCacheKeyTest.java
```diff
@@ -97,103 +101,233 @@ public void tearDown() throws IOException {
...
 		Task task2 = builder2.addInput(input).addOutput(output).create(null);
-		String key2 = ResourceCacheKey.calculate(task2, ignoredOptions2, output);
+		String key2 = ResourceCacheKey.calculate(task2.calculateSignature(), output);
 
 		assertEquals(key1, key2);
 	}
+
+	// Ensure that all parameters specified in all builders exist in Bob
+	@Test
+	public void testAllParametersExist() throws IOException, CompileExceptionError, NoSuchFieldException, IllegalAccessException {
```
AGulev, line 205: We use strings for parameters, it's useful to test if we didn't typo and all the parameters really exist

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/cache/test/ResourceCacheKeyTest.java
```diff
@@ -97,103 +101,233 @@ public void tearDown() throws IOException {
...
+			}
+		}
+	}
+
+	// Fails if a new Bob parameter is added.
+	// Ensure that new Bob parameters do not affect build signatures.
+	// If a parameter does affect builders, add it to @BuilderParams -> paramsForSignature,
+	// then update the list in this test.
+	@Test
+	public void testNoNewParametersAdded() throws IOException, CompileExceptionError, NoSuchFieldException, IllegalAccessException {
```
AGulev, line 248: I need to make sure that every time new parameter added, the developer who added it take into account this new logic. So, I added this test to make sure it will fail and the developer will check why and take it into considiration

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -512,102 +511,98 @@ public enum ArgCount {ZERO, ONE, MANY}
         public enum ArgType {UNSPECIFIED, ABS_OR_CWD_REL_PATH}
     }
 
-    private static CommandLineOption opt(String shortOpt, String longOpt, CommandLineOption.ArgCount argCount, CommandLineOption.ArgType argType, String description,
-                                         boolean usedByResourceCacheKey) {
-        if (usedByResourceCacheKey) {
-            ResourceCacheKey.includeOption(longOpt);
```
AGulev, line 518: Now it's up to builder to know what exactly bob parameter may affect it

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Builder.java
```diff
@@ -26,6 +30,29 @@
  */
 public abstract class Builder {
 
+    private static Map<Class<?>,  byte[]> classToParamsDigest = new HashMap<Class<?>,  byte[]>();
+
+    public static void addParamsDigest(Class<?> klass, Map<String, String> options, BuilderParams builderParams) throws NoSuchAlgorithmException {
```
AGulev, line 35: I don't want to do it too often, so I do it once per Builder class when we scan for classes.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/cache/ResourceCacheKey.java
```diff
@@ -15,65 +15,26 @@
...
 	 * @param resource The resource to calculate cache key for
 	 * @return The cache key as a hex string
 	 */
-	public static String calculate(Task task, Map<String, String> projectOptions, IResource resource) throws RuntimeException, IOException {
-		MessageDigest digest = task.calculateSignatureDigest();
-
+	public static String calculate(byte[] taskSignature, IResource resource) throws RuntimeException, IOException, NoSuchAlgorithmException {
+		MessageDigest digest = MessageDigest.getInstance("SHA1");
+        digest.update(taskSignature);
 		digest.update(resource.getPath().getBytes());
```
AGulev, line 37: I didn't get why it's needed, but I decided to keep it as is.
This function does not that much now.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/AtlasBuilder.java
```diff
@@ -48,7 +48,7 @@
 import com.google.protobuf.TextFormat;
 
 @ProtoParams(srcClass = Atlas.class, messageClass = TextureSet.class)
-@BuilderParams(name = "Atlas", inExts = {".atlas"}, outExt = ".a.texturesetc", isCacheble = true)
+@BuilderParams(name = "Atlas", inExts = {".atlas"}, outExt = ".a.texturesetc", isCacheble = true, paramsForSignature = {"texture-compression"})
```
AGulev, line 51: now, we specify for each builder what exactly may affect its signature

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Builder.java
```diff
@@ -26,6 +32,28 @@
...
+        String[] params = builderParams.paramsForSignature();
+        Set<String> paramSet = new HashSet<>(Arrays.asList(params));
+        if (paramSet.add("defoldsdk")) {
```
AGulev, line 40: `defoldsdk` sha1 should always be part of the signature for each task. If it's specified as parameter it will be used if not, default one will be used instead see https://github.com/defold/defold/blob/20133ef8a564ae4724ea2ce0fd307be2b0da50a0/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java#L866-L868


Also it's better to do it here to calculate it just once per builder instead of doing it for each task

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1324,7 +1321,6 @@ public void scanJavaClasses() throws IOException, CompileExceptionError {
         createClassLoaderScanner();
         registerPipelinePlugins();
         scan(scanner, "com.dynamo.bob");
-        scan(scanner, "com.dynamo.bob.pipeline");
```
AGulev, line 1327: this is just a sub-namespace of `com.dynamo.bob` - we don't want to scan it twice

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/cache/ResourceCacheKey.java
britzl: Hmm, yeah, it should be the protobuf file formats that are important, not the engine version?

britzl (approved):

PR #10261, opened 2025-02-24
Fix property-type-preservation-test

vlaaad:
Now that we use `run-later` in the properties view, we need to rebind it for tests that use the views.

### DIFF
--- a/editor/src/clj/editor/app_manifest.clj
+++ b/editor/src/clj/editor/app_manifest.clj
@@ -382,7 +382,8 @@
     :vulkan (concat
               vulkan-toggles
               (exclude-libs-toggles vulkan ["graphics"])
-              (generic-contains-toggles vulkan :excludeSymbols ["GraphicsAdapterOpenGL"]))
+              (generic-contains-toggles (disj vulkan :arm64-linux) :excludeSymbols ["GraphicsAdapterOpenGL"])
+              [(contains-toggle :arm64-linux :excludeSymbols "GraphicsAdapterOpenGLES")])
     :both vulkan-toggles
     :open-gl))
 
--- a/editor/test/integration/property_type_preservation_test.clj
+++ b/editor/test/integration/property_type_preservation_test.clj
@@ -106,7 +106,6 @@
 
 (defn- make-property-widget
   ^Parent [edit-type node-id prop-kw]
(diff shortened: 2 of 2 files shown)


editor/test/integration/property_type_preservation_test.clj
```diff
@@ -207,28 +207,29 @@
...
-                       (let [decorated-value (if (vector? prop-value)
-                                               (with-meta prop-value original-meta)
-                                               prop-value)]
-                         [prop-kw decorated-value])))
-                original-property-values)
-
-          graph-id (g/make-graph! :history true)
-          node-id (apply g/make-node! graph-id NumericPropertiesNode (mapcat identity property-values))]
-      (let [edit-type-by-prop-kw
+    (test-util/with-ui-run-later-rebound
```
vlaaad, line 210: The only real change is wrapping the test in `with-ui-run-later-rebound`, everything else is indentation.

editor/src/clj/editor/app_manifest.clj
```diff
@@ -382,7 +382,8 @@
     :vulkan (concat
               vulkan-toggles
               (exclude-libs-toggles vulkan ["graphics"])
-              (generic-contains-toggles vulkan :excludeSymbols ["GraphicsAdapterOpenGL"]))
+              (generic-contains-toggles (disj vulkan :arm64-linux) :excludeSymbols ["GraphicsAdapterOpenGL"])
+              [(contains-toggle :arm64-linux :excludeSymbols "GraphicsAdapterOpenGLES")])
```
vlaaad, line 386: Fix for test failure introduced in a commit by an external contributor (no tests were run for the contribution PR before merging it) — on arm64-linux, we now exclude `GraphicsAdapterOpenGLES` instead of  `GraphicsAdapterOpenGL`  (see https://github.com/defold/defold/commit/001803ecd5d9bf695ea0c9e1c7383acc790c70ac#diff-1fc20001e513238f7f1d25c1a628b6287e34cc7b64a287e6a13535547dac6511L34)

editor/test/integration/property_type_preservation_test.clj
vlaaad: I'm not sure why the original PR had no issues with this test...

matgis (approved):
Nice find! LGTM!

PR #10262, opened 2025-02-24
Fix a memory leak when using tables in render constants

### LINKED ISSUE #10259: Memory leak when storing table in render constants buffer
When you create a `render.constants_buffer` and repeatedly store a table in it (even if it's the same static table being stored into the same field), a small amount of memory will be leaked each time.

**To Reproduce (REQUIRED)**
Please consult this repo to reproduce the issue: https://github.com/akhleung/defold-issues
When you run the project, it will show the profiler and you can see that LuaMem keeps increasing over time. In `my.render_script`, you can see that a static table is being assigned to the same field of a render constants buffer in each frame. If you comment out the assignment or replace the table with a `vmath.vector4`, LuaMem will no longer increase. Note that `collect_garbage` is being called each frame as well.

Storing a table into a render constants buffer should not leak memory (especially reassigning the same table into the same field).

  - Version 1.9.6
  - Version 1.9.8 beta

**Platforms (REQUIRED):**
 - Windows 10, Samsung Galaxy Book Pro

https://github.com/akhleung/defold-issues

**Workaround (OPTIONAL):**

UPDATE 1: I thought this could be worked around by storing the table into the buffer only once, and then modifying the table in-place afterward, but the updated table isn't passed to the shader :-/

UPDATE 2: Okay, what _does_ work is to fetch the table back out of the render constants buffer and modify _that_ in place. E.g., something like `my_render_constants.light_matrices[i] = whatever`

~~Since the contents of a table can be modified in-place, it's not really necessary to reassign the entire table into the render constants buffer each frame. (However, if large updates to the table are necessary, then assigning a new table might be more convenient than updating the existing one. It's also conceivable that one might want to rotate between different tables each frame. In these cases, the leak would still occur.)~~

Jhonnyg:
Setting a table value with a previously allocated table + key caused memory leaks in the engine because the old element user data reference was never released.

Fixes #10259

### DIFF
--- a/engine/render/src/render/render_script.cpp
+++ b/engine/render/src/render/render_script.cpp
@@ -207,6 +207,8 @@ namespace dmRender
         if (lua_istable(L, 3))
         {
             ConstantBufferTableEntry* table_entry = cb_table->m_ConstantArrayEntries.Get(name_hash);
+            // If we are re-assigning an existing entry that was previously ref'd by this function
+            // we have to unref the old value, otherwise we will leak memory
             if (table_entry)
             {
                 dmScript::Unref(L, LUA_REGISTRYINDEX, table_entry->m_LuaRef);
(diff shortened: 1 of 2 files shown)


JCash (approved):
lgtm

PR #10263, opened 2025-02-24
Add scene context menu to editor scripts

### LINKED ISSUE #10235: Enable Scene Context Menu for Editor Scripts
Currently we can add new commands with Editor Scripts in context of Menu bar, Assets and Outline Elements.
In 1.9.8 a Scene Context Menu was added to Editor. It would be great if such context could be used with Editor Scripts enabling intuitive operations in Scenes!

Expose context "Scene Menu" or "Scene" or similar to Editor Scripts for commands https://defold.com/manuals/editor-scripts/#commands

-

https://forum.defold.com/t/defold-1-9-8-beta/79816/4

sprocketc:
Fixes #10235


vlaaad:
Regarding the first issue, it's probably because in this place we expect the selection to be an OutlineNode, but instead we have something else: https://github.com/defold/defold/blob/4e8483c3143df92f66ae6086d4d2cfb0e843c8a7/editor/src/clj/editor/editor_extensions/commands.clj#L75-L77

sprocketc:
Marking this as draft, since it requires further investigation. Besides fixing the issue, I am also skeptical about the fact that we are going to use the outline menu to register commands for the scene view. I think the underlying problem is that we have two different UI contexts (scene and outline), but the selection context remains the same. The right click menu related to the selection in this case.

editor/src/clj/editor/code/view.clj
```diff
@@ -2505,7 +2506,7 @@
    {:command :edit.copy :label "Copy"}
    {:command :edit.paste :label "Paste"}
    {:command :code.select-all :label "Select All"}
-   {:label :separator :id :editor.app-view/edit-end}])
+   (menu-items/separator-with-id :editor.app-view/edit-end)])
```
sprocketc, line 2509: Unrelated, but it looks like we are making an effort to replace those.

editor/src/clj/editor/editor_extensions/commands.clj
```diff
@@ -135,7 +135,7 @@
                              "Bundle" :global
                              "Debug" :global
                              "Edit" :global
-                             "Outline" :outline
+                             "Outline" :global
```
sprocketc: I had to change the context to get the selected nodes from the `env`, but I am not sure why we need to filter by context here (meaning that `:global` would work for all cases as far as I understand, because we filter nodes by type on `gen-selection-query`).

vlaaad: Hmm, I think the `:outline` context might be required here so that we get selection from a selected tree item in the outline, and not from, e.g., scene.

vlaaad: An outline is not the same as a scene. For example, the outline might have a game object with a sound component, and we can select the sound component, but there is no corresponding visual element in the scene. Perhaps it's wrong to use outline context menu in the scene altogether...

sprocketc: isn't the selection provider the same for both? So if you select something on the outline and then right click on the scene, the selection will be altered, because they share the same selection provider. So if you cannot click on it on the scene, it will be deselected, unless you deliberately hold shift?

sprocketc: If the selection is always synced, they should probably share the same context menu (I think).

vlaaad: But it's not always synced. Try creating a collection with a game object with a sound component. When the sound component is selected in the Outline, there is nothing selected in the scene.

vlaaad:
We probably want to approach the scene context menu a bit differently. It should not be exactly the same as the Outline context menu. At least, we should not try to show Outline context menu commands from editor scripts in the Scene view. Instead, we probably want to create a separate editor script command location like "Scene". Sorry for misleading you about editor script commands....

vlaaad:
And we don't need to do the Scene editor script command location as a part of this task.

vlaaad (approved):
Wonderful! Ship it! And don't forget to update [the docs](https://github.com/defold/doc/blob/master/docs/en/manuals/editor-scripts.md) 😉

sprocketc:
@vlaaad I used the initial custom context menu and added a `Scene` editor script location (that was what the original github issue was about). Reusing entries seemed unneeded, since we loose the ability to use different labels (especially on a localized context that might be added in the future), and also icons per location. I'll also update the docs when we merge this.

PR #10267, opened 2025-02-25
Async texture update support for vulkan

Jhonnyg:
The vulkan backend can now upload texture data in parallell.


engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -3982,14 +4043,181 @@ namespace dmGraphics
...
+        VulkanContext* context     = (VulkanContext*) _context;
+        uint16_t param_array_index = (uint16_t) (size_t) data;
+        SetTextureAsyncParams ap   = GetSetTextureAsyncParams(context->m_SetTextureAsyncState, param_array_index);
+
+        if (dmAtomicGet32(&context->m_DeleteContextRequested))
+        {
+            return 0;
+        }
+
+        VulkanTexture* tex = GetAssetFromContainer<VulkanTexture>(context->m_AssetHandleContainer, ap.m_Texture);
```
Jhonnyg, line 4077: This might suffer from the same thread safe issues we have with the async texture uploading in opengl as well, I'm not sure. We should probably guard the asset container

engine/graphics/src/vulkan/graphics_vulkan_device.cpp
```diff
@@ -1667,6 +1673,16 @@ namespace dmGraphics
             vk_create_pool_info.queueFamilyIndex = (uint32_t) queueFamily.m_GraphicsQueueIx;
             vk_create_pool_info.flags            = VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT;
             res = vkCreateCommandPool(logicalDeviceOut->m_Device, &vk_create_pool_info, 0, &logicalDeviceOut->m_CommandPool);
+
+            if (res == VK_SUCCESS)
+            {
+                memset(&vk_create_pool_info, 0, sizeof(vk_create_pool_info));
+                vk_create_pool_info.sType            = VK_STRUCTURE_TYPE_COMMAND_POOL_CREATE_INFO;
+                vk_create_pool_info.queueFamilyIndex = (uint32_t) queueFamily.m_GraphicsQueueIx; // Use the same queue for now (use a transfer queue at some point)
```
Jhonnyg, line 1681: Vulkan supports a dedicated transfer queue just for data uploads, that can run in parallell with the graphics queue. It's a bit more complicated to set that up, but might be worth it at some point.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -3634,6 +3661,47 @@ namespace dmGraphics
         return offset;
     }
 
+    static void CopyToTextureLayerWithtTageBuffer(VkCommandBuffer cmd_buffer, VkBufferImageCopy* copy_regions, DeviceBuffer* stage_buffer, VulkanTexture* texture, const TextureParams& params, uint32_t layer_count, uint32_t slice_size)
```
JCash, line 3664: Typo?  "Tage" -> "Stage"

JCash (dismissed):

PR #10279, opened 2025-03-01
Do not clean the bundle folder to speed up the bundling process

### LINKED ISSUE #3369: Bundling from editor starts Bob with 'clean' option (DEF-3578)
When spawning Bob for a bundle the editor adds the 'clean' option which leads to unnecessary long bundle times.

AGulev:
From now on, the bundle option in the editor does not clean the bundle folder (equivalent to the `distclean` option in Bob) to speed up bundling.

Fix https://github.com/defold/defold/issues/3369

### Technical notes

Now, when [this PR](https://github.com/defold/defold/pull/10254) has been merged, we should have much more robust cache invalidation when bundling. And now it might be useful to reuse already built assets between builds.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ScriptBuilders.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ScriptBuilders.java
@@ -21,19 +21,19 @@ import com.dynamo.lua.proto.Lua.LuaModule;
 public class ScriptBuilders {
 
     @ProtoParams(srcClass = LuaModule.class, messageClass = LuaModule.class)
-    @BuilderParams(name = "Lua", inExts = ".lua", outExt = ".luac", paramsForSignature = {"use-uncompressed-lua-source", "use-lua-bytecode-delta", "platform", "architectures"})
+    @BuilderParams(name = "Lua", inExts = ".lua", outExt = ".luac", paramsForSignature = {"use-uncompressed-lua-source", "use-lua-bytecode-delta", "platform", "architectures", "variant"})
     public static class LuaScriptBuilder extends LuaBuilder {}
 
     @ProtoParams(srcClass = LuaModule.class, messageClass = LuaModule.class)
-    @BuilderParams(name = "Script", inExts = ".script", outExt = ".scriptc", paramsForSignature = {"use-uncompressed-lua-source", "use-lua-bytecode-delta", "platform", "architectures"})
+    @BuilderParams(name = "Script", inExts = ".script", outExt = ".scriptc", paramsForSignature = {"use-uncompressed-lua-source", "use-lua-bytecode-delta", "platform", "architectures", "variant"})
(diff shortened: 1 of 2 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ScriptBuilders.java
```diff
@@ -21,19 +21,19 @@
 public class ScriptBuilders {
 
     @ProtoParams(srcClass = LuaModule.class, messageClass = LuaModule.class)
-    @BuilderParams(name = "Lua", inExts = ".lua", outExt = ".luac", paramsForSignature = {"use-uncompressed-lua-source", "use-lua-bytecode-delta", "platform", "architectures"})
+    @BuilderParams(name = "Lua", inExts = ".lua", outExt = ".luac", paramsForSignature = {"use-uncompressed-lua-source", "use-lua-bytecode-delta", "platform", "architectures", "variant"})
```
AGulev, line 24: I found that it is useful to take into account variant as well, because of Lua Preprocessor

vlaaad (commented):
The code is fine. I approve the code. I don't like the UI changes, I think we are adding too much internal knobs that bloat the dialogs. Could we instead just not do `distclean`, or is the bob not ready for this yet? Perhaps, if the bob is not ready, we could make an issue to ensure we don't need distclean?

AGulev:
>Could we instead just not do distclean, or is the bob not ready for this yet? Perhaps, if the bob is not ready, we could make an issue to ensure we don't need distclean?

It's a good question, I think it is ready, but we have to check to know it for sure.
Let me ping @britzl for his opinion

vlaaad:
If it is most probably ready, maybe a better solution would be adding a separate command to clear the build directory in case some users find a bob issue that we missed? This will allow us to get rid of the "rebundle" and "rebuild html5" commands.

AGulev:
We don't have "rebundle" only rebuild (which is clean build in the editor) and "rebuild html5" which is clean html5 bundle folder and build 
We can have one "clean bundle folders" to clean bundle folder and html5 bundle folder

vlaaad:
Sorry, I meant "rebuild"

AGulev:
"Rebuild" isn't related to bob, so we have to keep it anyways

vlaaad:
Aren't they all the same? I thought all of these just clean `build`, but maybe they are more specific, I didn't check

AGulev:
We have separate folders for build assets:
- build folder for editor build
- bundle folder for html5 build
- bundle folder for bundles

vlaaad:
Gotcha, thanks for clearing it up!

britzl:
> > Could we instead just not do distclean, or is the bob not ready for this yet? Perhaps, if the bob is not ready, we could make an issue to ensure we don't need distclean?
> 
> It's a good question, I think it is ready, but we have to check to know it for sure. Let me ping @britzl for his opinion

I believe bob should be ready for this now 👍

vlaaad (approved):
I hope it works 😁

PR #10283, opened 2025-03-03
Improve bundle UI and reliability

### LINKED ISSUE #10282: Improve bundling UI and process
**Task (REQUIRED):**
See https://forum.defold.com/t/defold-1-9-8-beta/79816/26

vlaaad:
Fixes #10282

### DIFF
--- a/editor/resources/bundle.editor_script
+++ b/editor/resources/bundle.editor_script
@@ -122,7 +122,8 @@ local function find_existing_path(paths)
             for j = 1, #path do args[#args+1] = path[j] end
             args[#args+1] = {reload_resources = false, out = "capture"}
             local success, result = pcall(editor.execute, table.unpack(args))
-            resolved_path = success and result
+            -- take the first line in case e.g. `where` returns multiple matches
+            resolved_path = success and result and (result:match("^(.-)\r?\n") or result)
         else -- string
             local success, result = pcall(string.gsub, path, "%$(%w+)", function(var)
                 return assert(os.getenv(var))
--- a/editor/resources/prelude.lua
+++ b/editor/resources/prelude.lua
@@ -237,10 +237,12 @@ function editor.bundle.dialog(heading, config, hint, error, content)
                 color = error and editor.ui.COLOR.ERROR or editor.ui.COLOR.HINT
             })
         }}),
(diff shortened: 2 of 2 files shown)


editor/resources/bundle.editor_script
```diff
@@ -122,7 +122,8 @@ local function find_existing_path(paths)
             for j = 1, #path do args[#args+1] = path[j] end
             args[#args+1] = {reload_resources = false, out = "capture"}
             local success, result = pcall(editor.execute, table.unpack(args))
-            resolved_path = success and result
+            -- take the first line in case e.g. `where` returns multiple matches
+            resolved_path = success and result and (result:match("^(.-)\r?\n") or result)
```
vlaaad, line 126: This fixes the following error from [the post](https://forum.defold.com/t/defold-1-9-8-beta/79816/26?u=vlaaad):
```
Resolving ADB location...
ERROR:EXT: bundle.editor_script:132 vm error: java.nio.file.InvalidPathException: Illegal char <
> at index 59: C:\Users\U\AppData\Local\Android\sdk\platform-tools\adb.exe
ERROR:EXT: C:\Users\U\adb\adb.exe
```
It's not immediately apparent from the error message, but it complains that there is an unexpected newline character in path, so the resolved adb path was something like `"C:\\...\\platform-tools\\adb.exe\r\nC:\\...\\adb\\adb.exe"`. This is because Windows' `where` command can actually return multiple results, each on a separate line, which I didn't know when I coded this initially.

editor/resources/prelude.lua
```diff
@@ -237,10 +237,12 @@ function editor.bundle.dialog(heading, config, hint, error, content)
...
-        content = editor.ui.grid({
-            padding = editor.ui.PADDING.LARGE,
-            columns = {{}, {grow = true}},
-            children = content
+        content = editor.ui.scroll({
+            content = editor.ui.grid({
+                padding = editor.ui.PADDING.LARGE,
+                columns = {{}, {grow = true}},
+                children = content
+            })
```
vlaaad, line 245: Adding a scroll pane ensures the dialog fits into the window.

britzl (approved):

PR #10284, opened 2025-03-03
Add callbacks for progress and completed load

smagnuso:
Added two extra parameters to the [`CUSTOM_PARAMTERS` for HTML5 builds](https://defold.com/manuals/html5/#extra-parameters):

* `start_success` - Function that is called just before main is called upon successful load
* `update_progress` - Function that is called as progress is updated. Parameter progress is updated 0-100

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
@@ -29,6 +29,12 @@
 *     'resize_window_callback':
 *         Function that is called when resize/orientationchanges/focus events happened
 *
+*     'start_success':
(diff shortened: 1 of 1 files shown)


britzl (approved):

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
```diff
@@ -48,6 +48,10 @@ var CUSTOM_PARAMETERS = {
         var e = document.getElementById("webgl-not-supported");
         e.style.display = "block";
     },
+    start_success: function() {
```
britzl, line 51: Please add both of these in the comment above.

britzl (changes requested):
Add comment about both of the new parameters

britzl (approved):

PR #10288, opened 2025-03-04
Refactor editor http server implementation

### LINKED ISSUE #10272: Expose editor HTTP server to editor scripts
**Task (REQUIRED):**
Editor scripts should be able to respond to HTTP requests for the editor server.

vlaaad:
Now editor scripts may add routes to the built-in HTTP server, e.g.:
```lua
function M.get_server_routes()
  return {
    http.server.route("/my-extension/files/{*file}", function(request)
      local attrs = editor.external_file_attributes(request.file)
      if attrs.is_file then
        return http.server.external_file_response(request.file)
      else
        return http.server.response(404, "Not found")
      end
    end)
  }
end
```

Fixes #10272


editor/project.clj
```diff
@@ -63,6 +63,8 @@
                      [com.defold.lib/bob                          "1.0"]
                      [com.defold.lib/openmali                     "1.0"]
 
+                     [metosin/reitit-core "0.8.0-alpha1"]
```
vlaaad, line 66: Router library. This dependency increases the editor size by 33kb.

editor/test/integration/web_server_test.clj
```diff
@@ -0,0 +1,66 @@
...
+            (let [{:keys [status headers body]} @(http/request (str url "/command/build-html5") :as :string)]
+              (is (= 405 status))
+              (is (= "OPTIONS, POST" (get headers "allow")))
+              (is (= "405 Method Not Allowed\n" body)))
+            (with-redefs [ui/open-url (promise)]
+              (let [{:keys [status body]} @(http/request (str url "/command/report-issue") :method "POST" :as :string)]
+                (is (= 202 status))
+                (is (= "202 Accepted\n" body)))
+              (when (is (realized? ui/open-url))
+                (is (string/includes? @ui/open-url "github.com/defold/defold/issues"))))))))))
```
vlaaad, line 66: I added a test for the other endpoints of the HTTP server besides the hot-reload ones tested in `hot-reload-test`

editor/src/clj/util/http_server.clj
```diff
@@ -13,98 +13,415 @@
...
+                            ;; Response length argument:
+                            ;; -1 means empty response body (0 bytes)
+                            ;; 0 is unknown size (chunked)
+                            ;; positive is exact size
+                            (if (nil? connection)
+                              -1
+                              (if-let [explicit-length (get headers "content-length")]
+                                (let [n (Long/parseUnsignedLong explicit-length)]
+                                  (if (zero? n) -1 n))
+                                0)))
```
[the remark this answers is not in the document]
matgis: Really annoying for the `responseLength` argument to behave like this. 😬 I would have expected the zero and -1 cases to be swapped, if anything.

editor/src/clj/util/http_server.clj
```diff
@@ -13,98 +13,415 @@
...
+                          (when connection
+                            (try
+                              (with-open [is (io/input-stream connection)]
+                                (try
+                                  (io/copy is (.getResponseBody exchange))
+                                  ;; A browser or remote http client can close
+                                  ;; the connection before we finished sending
+                                  ;; the response: ignore such exceptions since
+                                  ;; it's not an issue.
+                                  (catch IOException _)))
```
matgis, line 342: According to the docs for `getResponseBody`:

> If too many bytes are written, then the write method of `OutputStream` will throw an `IOException`. If too few bytes are written then the stream `OutputStream.close()` will throw an `IOException`.

I think the case you're catching here is when _too many_ bytes are written to the output stream returned by `.getResponseBody`? `io/copy` doesn't close the stream, so presumably that won't happen until `.close` is called on the `exchange` in the `finally` clause below?

editor/src/clj/util/http_server.clj
```diff
@@ -13,98 +13,415 @@
...
+                when responding to HEAD requests)
+  See also:
+    https://cljdoc.org/d/metosin/reitit-core/0.8.0-alpha1/doc/introduction"
+  [routes]
+  (let [allowed-methods (fn allowed-methods [method->handler]
+                          (let [method-set (-> method->handler keys set (conj "OPTIONS"))
+                                method-set (cond-> method-set (contains? method-set "GET") (conj "HEAD"))]
+                            (string/join ", " (sort method-set))))
+        invoke-handler (fn invoke-handler [handler request match]
+                         (handler (assoc request :path-params (:path-params match))))
```
matgis, line 405: These functions don't appear to use anything from the `routes`, so they could be global?

editor/src/clj/util/http_server.clj
```diff
@@ -13,98 +13,415 @@
...
+        router (reitit/router routes {:syntax :bracket})]
+    (fn route-request [request]
+      (let [path (:path request)]
+        (if-let [match (or (reitit/match-by-path router path)
+                           (if (string/ends-with? path "/")
+                             (reitit/match-by-path router (subs path 0 (dec (count path))))
+                             (reitit/match-by-path router (str path "/"))))]
```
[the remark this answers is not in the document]
matgis: In their own examples, they do a redirect when this happens, possibly to ensure the client ends up at the corrected url. Not sure if we want to do the same? 🤔

See https://cljdoc.org/d/metosin/reitit/0.5.18/doc/ring/slash-handler

editor/test/integration/editor_extensions_test.clj
```diff
@@ -855,16 +854,14 @@ POST http://localhost:23456/echo hello world! as string => 200
...
-                       "/json" (response/response "{\"a\": 1, \"b\": [true]}")
-                       "/echo" (response/response (slurp (:body request)))
-                       (response/not-found "404")))
-                   {:port 23456 :join? false})
+    (let [server (http-server/start!
+                   (http-server/router-handler
+                     {"/redirect/foo" {"GET" (constantly (http-server/redirect "/foo"))}
+                      "/foo" {"GET" (constantly (http-server/response 200 "successfully redirected"))}
+                      "/" {"GET" (constantly (http-server/response 200 ""))}
+                      "/json" {"GET" (constantly (http-server/response 200 "{\"a\": 1, \"b\": [true]}"))}
```
matgis, line 862: Should we use `http-server/json-response` here?

editor/test/integration/test_util.clj
```diff
@@ -793,33 +793,34 @@
...
-                                                    (.closeEntry out))))
-                                              (.finish out)
-                                              (let [bytes (.toByteArray byte-stream)]
-                                                {:headers {"ETag" "tag"}
-                                                 :body bytes}))))})
-    (http-server/start!)))
-
-(defn kill-lib-server [server]
-  (http-server/stop! server))
+(defmacro with-server [handler & body]
```
matgis, line 796: Rename to `with-http-server`? This is a bit generic for `test-util`. There can be other servers eventually, I guess?

editor/test/integration/test_util.clj
```diff
@@ -793,33 +793,34 @@
...
+  `(let [~'server (http-server/start! ~handler)]
+     (try
+       (do ~@body)
+       (finally (http-server/stop! ~'server 0)))))
+
+(def lib-server-handler
+  (http-server/router-handler
+    {"/lib/{*lib}" {"GET" (fn [request]
```
matgis, line 804: Break up into multiple lines to control indentation?

editor/test/integration/web_server_test.clj
```diff
@@ -0,0 +1,66 @@
...
+            [editor.progress :as progress]
+            [editor.ui :as ui]
+            [editor.web-server :as web-server]
+            [integration.test-util :as test-util]
+            [util.http-client :as http]
+            [util.http-server :as http-server])
+  (:import [javafx.scene Scene]
+           [javafx.scene.layout Region]
+           [javafx.stage Stage]))
+
```
matgis, line 19: Add missing copyright header and `(set! *warn-on-reflection* true)`.

matgis (changes requested):
Stellar effort! It's a lot nicer after this. ❤️

I've not done any manual testing of this, because it seems like a lot of surface area to cover. I'm gonna rely on the automated tests.

I'm also not very knowledgeable when it comes to this stuff, but I think I may have found some concerns. I'll let you have a look and respond! 😅

editor/src/clj/editor/hot_reload.clj
```diff
@@ -91,16 +44,24 @@
...
+(defn routes [workspace]
+  (let [build-path (.toPath (workspace/build-path workspace))]
+    {"/build/{*path}" {"GET" (bound-fn [{:keys [path-params headers]}]
+                               (let [^String path (:path path-params)
+                                     full-path (.normalize (.resolve build-path path))]
+                                 (if (and (.startsWith full-path build-path) ;; Avoid going outside the build path with '..'
+                                          (fs/path-exists? full-path)
+                                          (not (fs/path-is-directory? full-path)))
+                                   (let [etag (workspace/etag workspace (str "/" path))
+                                         remote-etag (get headers "if-none-match")]
```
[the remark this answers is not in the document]
vlaaad: Yes, exactly. I ensure lower-case headers (both in the server and client!)

editor/src/clj/editor/html_view.clj
```diff
@@ -25,66 +25,35 @@
...
+                                (let [body (g/node-value node :html)]
+                                  (if-let [error (g/error? body)]
+                                    (http-server/response 500 (format "Couldn't render %s: %s\n"
+                                                                      (resource/resource->proj-path resource)
+                                                                      (or (:message error) "Unknown reason")))
+                                    (http-server/response 200 {"content-type" "text/html; charset=utf-8"} body)))
+                                (http-server/response 200 resource)))
+                            (do (log/warn :message (format "Cannot find resource for %s" path))
+                                http-server/not-found)))}})
+    :host "0.0.0.0"))
```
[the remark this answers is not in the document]
vlaaad: Oops, I didn't notice it was `127.0.0.1` and not `0.0.0.0`. I guess there is no difference between these in this case, but it is important to provide a host here; otherwise, clicking on links that point to other in-project HTML resources won't work: it's used to make `(string/starts-with? new-location url)` work in `handle-location-change!`.

editor/src/clj/editor/resource.clj
```diff
@@ -326,7 +333,26 @@
...
 
   io/Coercions
-  (as-file [this] (io/as-file zip-uri)))
+  (as-file [this] (io/as-file zip-uri))
+
+  http-server/ContentType
+  (content-type [this] (http-server/ext->content-type (type-ext this)))
+  http-server/DataConnection
+  (data-connection [_]
+    (let [zip-file (ZipFile. ^File (io/as-file zip-uri))
```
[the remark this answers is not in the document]
vlaaad: Good point, I'll change it.

editor/src/clj/editor/resource.clj
```diff
@@ -326,7 +333,26 @@
...
+        (data-length [_]
+          (let [size (.getSize entry)]
+            (when-not (= -1 size) size)))
+        io/IOFactory
+        (make-input-stream [_ opts] (io/make-input-stream (.getInputStream zip-file entry) opts))
+        (make-reader [this opts] (io/make-reader (io/make-input-stream this opts) opts))
+        (make-output-stream [_ _] (throw (Exception. "Zip resources are read-only")))
+        (make-writer [_ _] (throw (Exception. "Zip resources are read-only")))
+        AutoCloseable
+        (close [_] (.close zip-file))))))
```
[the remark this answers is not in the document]
vlaaad: Yes, because the entry input stream is the only input stream we return when opening the file here. FYI, this code mostly repeats the `make-zip-resource-input-stream` fn that we use for reading zip resources but does so using a reify instead of a proxy.

editor/src/clj/util/http_server.clj
```diff
@@ -13,98 +13,415 @@
...
+
+(extend-protocol ContentData
+  String (content-data [s] (.getBytes s StandardCharsets/UTF_8))
+  File (content-data [file] (.toPath file))
+  Object (content-data [x] x)
+  nil (content-data [x] x))
+
+(extend-protocol DataLength
+  byte/1 (data-length [bytes] (count bytes))
+  Object (data-length [_])
```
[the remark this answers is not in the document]
vlaaad: No, it would not. These protocols define an open extension point for inferring content length and type for inputs, but there are still other valid bodies (i.e., anything implementing `io/IOFactory`) that can't report their content length. For example, you can make a response with an `InputStream` instance as a body — the HTTP server will successfully write it to the output using chunked transfer encoding. We don't want any errors in this case.

editor/src/clj/util/http_server.clj
```diff
@@ -13,98 +13,415 @@
...
+(extend-protocol DataConnection
+  Path (data-connection [path]
+         (reify
+           DataLength
+           (data-length [_] (Files/size path))
+           io/IOFactory
+           (make-input-stream [_ opts] (io/make-input-stream path opts))
+           (make-output-stream [_ opts] (io/make-output-stream path opts))
+           (make-reader [_ opts] (io/make-reader path opts))
+           (make-writer [_ opts] (io/make-writer path opts))))
```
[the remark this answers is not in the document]
vlaaad: Yes, I agree.

editor/src/clj/util/http_server.clj
```diff
@@ -13,98 +13,415 @@
...
+            io/IOFactory
+            (make-input-stream [_ opts] (io/make-input-stream (.getInputStream connection) opts))
+            (make-output-stream [_ opts] (io/make-output-stream (.getOutputStream connection) opts))
+            (make-reader [_ opts] (io/make-reader (.getInputStream connection) opts))
+            (make-writer [_ opts] (io/make-writer (.getOutputStream connection) opts))
+            AutoCloseable
+            (close [_]
+              (.close (.getInputStream connection))))))
+  URI (data-connection [uri] (data-connection (.toURL uri)))
+  Object (data-connection [x] x)
```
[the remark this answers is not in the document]
vlaaad: No, this enables automatic support for all the other `io/IOFactory` implementations like `InputStream` instances.

editor/src/clj/util/http_server.clj
vlaaad: Hmm, that's a good point. The actual exceptions I was getting were `Broken pipe` and `Connection reset by peer` when I was holding `F5` in the browser, which caused it to constantly reload the engine profiler page. I think I'll log the exceptions instead of ignoring them completely (in another PR)

editor/src/clj/util/http_server.clj
vlaaad: Yes, I'll make them global (in another PR)

editor/test/integration/test_util.clj
vlaaad: Heh, I thought about it but didn't come up with other servers we might have. Maybe we could do it when we actually have another non-http server? 😄

matgis: We could, but I prefer we do it now for clarity. We're a desktop application, so "server" could mean other things (a build server or native extension server maybe?).

I'd especially like to address this now since it has an entry in `codeStyles/Project.xml`, so a refactoring rename will mess up the indentation if we're not careful.

matgis (changes requested):

PR #10298, opened 2025-03-05
Create atlas images on drag & drop from the asset tree

### LINKED ISSUE #7055: Allow drag&drop of image files from the Assets pane into Atlas.
Adding new images to an atlas when there are many assets in a project is a pain.

Atlas should support accepting new image files and folders by drag&drop from the assets pane.

Using the regular search box.

sprocketc:
Enable dragging & dropping images from the asset tree to the atlas scene view.

dnd-to-atlas

Fixes #7055

### Technical changes
The drag event handlers are registered on the `gl-pane` and can be handled using the augmented `action-from-jfx`. This would allow us to handle drag events on any `input-handler` of scene controllers. Abstracting the original event away has some cons, because it reduces the level of control that the original event provides.


editor/src/clj/editor/scene.clj
```diff
@@ -1574,6 +1576,7 @@
                   (g/connect view-id         :viewport                      tool-controller :viewport)
                   (g/connect camera          :camera                        tool-controller :camera)
                   (g/connect view-id         :selected-renderables          tool-controller :selected-renderables)
+                  (g/connect app-view-id     :_node-id                      tool-controller :app-view)
```
sprocketc, line 1579: We need the app-view to get and also update the selection.

editor/src/clj/editor/input.clj
```diff
@@ -28,7 +28,9 @@
                  MouseEvent/MOUSE_RELEASED :mouse-released
                  MouseEvent/MOUSE_CLICKED :mouse-clicked
                  MouseEvent/MOUSE_MOVED :mouse-moved
-                 MouseEvent/MOUSE_DRAGGED :mouse-moved})
+                 MouseEvent/MOUSE_DRAGGED :mouse-moved
+                 DragEvent/DRAG_OVER :drag-over
+                 DragEvent/DRAG_DROPPED :drag-dropped})
```
sprocketc, line 33: The list of drag events is not exhaustive, but those are enough for this case.

editor/src/clj/editor/input.clj
```diff
@@ -56,6 +58,20 @@
                        :shift (.isShiftDown scroll-event)
                        :meta (.isMetaDown scroll-event)
                        :control (.isControlDown scroll-event)))
+      :drag-over (let [drag-event ^DragEvent jfx-event]
+                   (.acceptTransferModes drag-event TransferMode/ANY)
+                   (assoc action
+                          :x (.getX drag-event)
+                          :y (.getY drag-event)))
+      :drag-dropped (let [drag-event ^DragEvent jfx-event]
+                      (.setDropCompleted drag-event true)
```
sprocketc, line 67: Quoting javafx doc
> the gesture needs to be completed by calling setDropCompleted(Boolean) on the event. The Boolean argument indicates if the data has been transferred successfully or not. If it is not called, the gesture is considered unsuccessful.

Since this transforms the original jfx event to an action, we won't have access to the event from the input handler. The data has been transferred successfully is a way, but I think that's not exactly what the docs suggest.

vlaaad: Hmm, maybe we could add the event to the action map so that we can use it if needed. As I understand, the effect of it is that if the drop is unsuccessful, there is an animation that moves the unsuccessfully-dragged icons back. Maybe we want that?

vlaaad (dismissed):
LGTM, just a minor nitpick: we should use `keep` over `map + remove nil?`. I'm not sure if detecting the success of the drag operation is important here for playing the "rewind" of the dragged icons...

sprocketc:
@vlaaad I applied your suggestions along with some minor naming fixes.
>As I understand, the effect of it is that if the drop is unsuccessful, there is an animation that moves the unsuccessfully-dragged icons back.

About that, it doesn't look like it, but maybe I am doing something wrong. Anyway, adding the event to handle this later is a better option regardless.

editor/src/clj/editor/atlas.clj
```diff
@@ -1186,8 +1191,67 @@
...
+                            workspace (project/workspace project)
+                            image-resources (get-image-resources-from-dragboard dragboard workspace)
+                            op-seq (gensym)
+                            image-nodes (create-dropped-images! parent image-resources op-seq)
+                            drag-event ^DragEvent (:event action)]
+                        (when (seq image-nodes)
+                          (select! app-view image-nodes op-seq)
+                          (ui/user-data! (ui/main-scene) ::ui/refresh-requested? true)
+                          (.setDropCompleted drag-event true))
+                        nil)))
```
matgis, line 1254: I think you should be able to get what you need here from the ui context of the `:target` of the `action`, to avoid connecting the `app-view` to each `ToolController`.
```clj
(let [image-view (:target action) ; JavaFX ImageView that was the drop target.
      ui-context (first (ui/node-contexts image-view false))
      {:keys [app-view project selection workspace]} (:env ui-context)]
  ...)
```

matgis (changes requested):
Works great overall, but it seems to throw an exception if I have a different resource being the active tab in split view ("Move to Another Tab Pane" from the Tab context menu), and drag images onto the Atlas scene view in the inactive Tab Pane.

editor/src/clj/editor/input.clj
```diff
@@ -56,6 +59,19 @@
                        :shift (.isShiftDown scroll-event)
                        :meta (.isMetaDown scroll-event)
                        :control (.isControlDown scroll-event)))
+      :drag-over (let [drag-event ^DragEvent jfx-event]
+                   (.acceptTransferModes drag-event TransferMode/ANY)
+                   (assoc action
+                          :x (.getX drag-event)
+                          :y (.getY drag-event)))
```
matgis, line 66: Style: `assoc` indents key-value pairs with two spaces.

editor/src/clj/editor/atlas.clj
```diff
@@ -44,23 +44,27 @@
...
+            [editor.ui :as ui]
             [editor.validation :as validation]
             [editor.workspace :as workspace]
             [internal.util :as util]
             [schema.core :as s]
             [util.digestable :as digestable]
             [util.fn :as fn]
-            [util.murmur :as murmur])
+            [util.murmur :as murmur]
+            [clojure.string :as str])
```
matgis, line 55: Style: Keep requires sorted.

editor/src/clj/editor/atlas.clj
```diff
@@ -44,23 +44,27 @@
...
            [com.jogamp.opengl GL GL2]
            [editor.gl.vertex2 VertexBuffer]
            [editor.types AABB Animation Image]
            [java.awt.image BufferedImage]
+           [java.io File]
            [java.nio ByteBuffer]
            [java.util List]
-           [javax.vecmath AxisAngle4d Matrix4d Point3d Vector3d]))
+           [javax.vecmath AxisAngle4d Matrix4d Point3d Vector3d]
+           [javafx.scene.input Dragboard DragEvent]))
```
matgis, line 67: Style: Keep imports sorted.

matgis (dismissed):
Works great! LGTM! ❤️

editor/src/clj/editor/input.clj
sprocketc: Not trying to argue, but what's the reason behind this rule? I would assume that it falls under this category https://guide.clojure.style/#vertically-align-fn-args.

matgis: Not sure. 🤔 I think it might just be the default behavior in Cursive. There are a bunch of exceptions to that rule specified in `editor/.idea/codeStyles/Project.xml`.

matgis (dismissed):

vlaaad (dismissed):
LGTM!

matgis (approved):
Oh, neat! That's even better! Seems to work great! LGTM! ❤️

PR #10308, opened 2025-03-07
Persist saved colors of color picker

### LINKED ISSUE #6684: Persist saved colors from color picker in project-wide palette
When using the color picker, you can "save" a color. It's not saved anywhere, but the functionality is very useful because it's common to have a defined color palette in a game. 

We should actually save the "saved" colors in a project-level palette.

We could save in a system-level palette (i.e. prefs), but it's useful to have a defined color palette that is shared between all team members.

### LINKED ISSUE #6693: Make a better color picker
The current color picker is hard to use:
1. Entering hex color code requires too many clicks in the color picker.
2. Developers often would prefer to pick a color by pointing at a pixel on the screen, not possible with the current color picker
3. Add support for project-wide `.palette` files as a way to pick colors.

We should provide a more useful color picker

We could either create it ourselves or maybe try to find 3rd-party JavaFX color pickers

sprocketc:
Persist custom saved colors on project prefs.

Screenshot From 2025-03-12 13-04-05

Fixes #6684
Fixes #6693

### Technical changes
I tried persisting the colors to `game.project` or `project.shared_editor_settings`, but I decided to use prefs eventually. The reason was that custom colors are not intended to be used as a color palette manager. Saving a color is more of a quick way to store a value for later use. Sharing a single project palette in a collaborating context will introduce more problems. The closest option that we have is probably `[:workflow :recent-files]`, so I used `[:workflow :saved-colors]` for the color path. I also decided not to expose the list to our prefs dialog, since there is a way to remove a saved color, by right clicking on the color rect.


editor/src/clj/editor/properties_view.clj
```diff
@@ -594,6 +612,7 @@
                                   (let [c (.getValue color-picker)]
                                     (set-color-value! property-fn ignore-alpha c)
                                     (ui/user-data! (ui/main-scene) ::ui/refresh-requested? true))))
+    (.setOnHidden color-picker (fn [_] (save-colors! color-picker prefs)))
```
sprocketc, line 615: We also need to save the list when we remove a color, so calling `save-colors!` this on `ui/on-action!` won't suffice.

vlaaad: Since ColorPicker's `getCustomColors` returns an observable list, I think it might be better to add a `ChangeListener` to it to save them.

editor/src/clj/editor/properties_view.clj
```diff
@@ -553,10 +557,22 @@
...
+(defn- get-saved-colors
+  [prefs]
+  (->> (prefs/get prefs saved-colors-prefs-path)
+       (mapv #(Color/valueOf ^String %))))
+
+(defmethod create-property-control! types/Color [edit-type {:keys [color-dropper-view prefs]} property-fn]
   (let [wrapper (doto (HBox.)
                   (.setPrefWidth Double/MAX_VALUE))
         pick-fn (fn [c] (set-color-value! property-fn (:ignore-alpha? edit-type) c))
+        saved-colors (when prefs ^Collection (get-saved-colors prefs))
```
vlaaad, line 575: Hmm, I think `prefs` must not be `nil` here, is that right? Saving them does not perform a `nil` check in `save-colors!`, so either both places should check for `nil`, or neither.

vlaaad (approved):
LGTM! Nice!

PR #10319, opened 2025-03-11
Optimized the gui node size

AGulev:
The size is now 512 bytes instead of 720 bytes for dynamically created nodes and 704 bytes for nodes loaded from resources. 

### Technical changes
Smaller enums helped save a few bytes, and together with alignment improvements, reduced the structure size to 688 bytes.  

However, I then discovered that each structure contains a copy of all the properties for reset, which is rarely need. Yet, we "pay" with a much larger node size for each node just to have this rarely used functionality.

### DIFF
--- a/engine/gui/src/dmsdk/gui/gui.h
+++ b/engine/gui/src/dmsdk/gui/gui.h
@@ -117,7 +117,7 @@ namespace dmGui
      * @member NODE_TEXTURE_TYPE_TEXTURE
      * @member NODE_TEXTURE_TYPE_TEXTURE_SET
      */
-    enum NodeTextureType
+    enum NodeTextureType : uint8_t
     {
         NODE_TEXTURE_TYPE_NONE,
         NODE_TEXTURE_TYPE_TEXTURE,
--- a/engine/gui/src/gui.cpp
+++ b/engine/gui/src/gui.cpp
@@ -2519,9 +2519,7 @@ namespace dmGui
         }
         if (n->m_Node.m_Text)
             free((void*)n->m_Node.m_Text);
-        if (n->m_Node.m_HasResetPoint) {
-            free(n->m_Node.m_ResetPointProperties);
-        }
+        free(n->m_Node.m_ResetPointProperties);
         memset(n, 0, sizeof(InternalNode));
         n->m_Index = INVALID_INDEX;
     }
@@ -4233,7 +4231,7 @@ namespace dmGui
 
         InternalNode* n = GetNode(scene, node);
         out_n->m_Node = n->m_Node;
-        out_n->m_Node.m_HasResetPoint = false;
+        out_n->m_Node.m_HasResetPoint = 0;
         out_n->m_Node.m_ResetPointProperties = 0;
         if (n->m_Node.m_Text != 0x0)
             out_n->m_Node.m_Text = strdup(n->m_Node.m_Text);
--- a/engine/gui/src/gui.cpp
+++ b/engine/gui/src/gui.cpp
@@ -2519,6 +2519,9 @@ namespace dmGui
         }
         if (n->m_Node.m_Text)
             free((void*)n->m_Node.m_Text);
+        if (n->m_Node.m_HasResetPoint) {
+            free(n->m_Node.m_ResetPointProperties);
+        }
         memset(n, 0, sizeof(InternalNode));
         n->m_Index = INVALID_INDEX;
     }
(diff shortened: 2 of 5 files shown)


engine/gui/src/dmsdk/gui/gui.h
```diff
@@ -117,7 +117,7 @@ namespace dmGui
      * @member NODE_TEXTURE_TYPE_TEXTURE
      * @member NODE_TEXTURE_TYPE_TEXTURE_SET
      */
-    enum NodeTextureType
```
AGulev, line 120: this is cpp 11 feature, I think it's ok to use it?

engine/gui/src/gui.cpp
```diff
@@ -2923,6 +2926,10 @@ namespace dmGui
     void SetNodeResetPoint(HScene scene, HNode node)
     {
         InternalNode* n = GetNode(scene, node);
+        if (!n->m_Node.m_ResetPointProperties)
+        {
+            n->m_Node.m_ResetPointProperties = (dmVMath::Vector4*)malloc(sizeof(dmVMath::Vector4) * PROPERTY_COUNT);
+        }
```
AGulev, line 2932: `m_Node.m_ResetPointProperties` is rarely needed when we reset values, so we don't need these values all the time as part of the node structure. It helps to reduce node structure size to 512 bytes instead of 720 which should improve cache locality

engine/gui/src/gui.cpp
```diff
@@ -4226,6 +4233,8 @@ namespace dmGui
 
         InternalNode* n = GetNode(scene, node);
         out_n->m_Node = n->m_Node;
+        out_n->m_Node.m_HasResetPoint = false;
+        out_n->m_Node.m_ResetPointProperties = 0;
```
AGulev, line 4237: According to documentation `guidelines.reset_nodes()`shouldn't affect dynamic nodes:
CleanShot 2025-03-11 at 08 03 00@2x

That means we don't need this data here

engine/gui/src/gui.cpp
```diff
@@ -2519,6 +2519,9 @@ namespace dmGui
         }
         if (n->m_Node.m_Text)
             free((void*)n->m_Node.m_Text);
+        if (n->m_Node.m_HasResetPoint) {
+            free(n->m_Node.m_ResetPointProperties);
+        }
```
JCash, line 2524: No need for the if-statement, free() can handle a null pointer.

engine/gui/src/dmsdk/gui/gui.h
JCash: I think so too, yes.

engine/gui/src/gui.cpp
```diff
@@ -4226,6 +4233,8 @@ namespace dmGui
 
         InternalNode* n = GetNode(scene, node);
         out_n->m_Node = n->m_Node;
+        out_n->m_Node.m_HasResetPoint = false;
```
JCash, line 4236: It's technically a utin32, so it should be 0 here

engine/gui/src/gui_private.h
```diff
@@ -105,72 +105,65 @@ namespace dmGui
...
     struct Node
     {
-        dmVMath::Vector4    m_Properties[PROPERTY_COUNT];
-        dmVMath::Vector4    m_ResetPointProperties[PROPERTY_COUNT];
-        dmVMath::Matrix4    m_LocalTransform;
-        dmVMath::Vector4    m_LocalAdjustScale;
-        uint32_t            m_ResetPointState;
+        dmVMath::Matrix4        m_LocalTransform;
+        dmVMath::Vector4        m_Properties[PROPERTY_COUNT];
+        dmVMath::Vector4        m_LocalAdjustScale;
```
JCash, line 110: Fyi: While biggest-first is a common approach, they all have the same alignment (vector4), so it wasn't strictly necessary to switch these around.

engine/gui/src/gui_private.h
```diff
@@ -105,72 +105,65 @@ namespace dmGui
...
 
-        uint32_t    m_HasResetPoint         : 1; // If true, we have stored a copy of the m_State, into m_ResetPointState
-        uint32_t    m_PerimeterVertices     :31;
-        PieBounds   m_OuterBounds;
+        uint64_t                m_TextureHash;
+        uint64_t                m_FlipbookAnimHash;
+        uint64_t                m_FontHash;
+        uint64_t                m_ParticlefxHash;
+        dmhash_t                m_LayerHash;
+        dmhash_t                m_MaterialNameHash;
```
JCash, line 117: I wonder if we should change all the hash to use `dmhash_t` here, or if it causes problems?
I don't want the PR to grow because of that. 🤔

JCash (dismissed):

engine/gui/src/gui.cpp
AGulev: fixed, thank you

engine/gui/src/gui_private.h
AGulev: if you don't mind, let's keep it for now as is, and fix it in the future iterations

PR #10341, opened 2025-03-14
Send all events through a callback to be handled in caller if desired

smagnuso:
Much like getting start_success I'd like notification on the outside for errors in startup.


britzl (approved):
Looks good!

PR #10345, opened 2025-03-17
Make scene grid configurable

### LINKED ISSUE #6420: Editor: render axis Z and horizonal grid in perspective camera mode
There is a vertical grid only and no axis Z.

image

Render the same grid but horizontally. Plus Z-axis.

### LINKED ISSUE #6421: Use grid main lines every 8/16 units instead of the default 10 units.
**Related problem :**

Using grid main lines every 8/16 units will be helpful while aligning sprites instead of the 10 units, we usually are using power of 2 sizes for all our textures 16/32/64/128/256/512, this way the main lines will fit all those sizes by default.

Screenshot from 2022-02-18 11-20-02

**Solution :**
Because currently we can't control the grid (size / main-lines ) so it will be better to make the default main lines draw every 8/16 units instead of the default value "10".

**Alternatives :**
An alternative is to be able to control the [ grid -> size / main-lines ]

### LINKED ISSUE #8407: 2D Grid for The Editor
While designing levels, it always frustrated me that there is no sense of measurement besides side rulers and numerical data which does not help in terms of placing objects efficiently.

A 2 dimensional grid system which can have configurable cell sizing and line color. If a specific shortcut key is triggered the object or component could snap to the grid. This system I described is standard with many engines editors.

Using another external editor but this defeats the purpose of Defold engine.

## How other Editors Provide Grid and Snapping

In other editors, snapping usually involves a grid, that’s why I combined them. In some cases snapping mode is triggered by a keybinding (like shift) or a UI button. After it is enabled, when the object’s position is modified with UI handles, it changes the position of the object according to its dragged position but approximating it to the configured cell size of the grid to fit it to the grid. I saw similar implementations with adjusting size/scale as well.

sprocketc:
customizable-grid

The grid toggle was moved from the visibility dropdown, at the top level of the scene toolbar. A dedicated grid options dropdown was added next to the toggle, to allow users to configure the size per axis, the visibility plane, the color, and the opacity, on a project by project basis. A default keyboard shortcut (`Ctrl+G`) that toggles the grid was also introduced.

Fixes #6421
Fixes #8407
Fixes #6420

### Technical changes
`curve_grid.clj` and `tile_map_grid.clj` removed. `TileMapGrid` and `CurveGrid` now inherit our main `Grid` node, and provide an `options` map as an output. The Grid is using the provided options to override the grid prefs. Overridden preferences are not available on the popover settings, since we cannot change their value.


selimanac:
This is probably my first Defold Editor [issue](https://github.com/defold/editor2-issues/issues/1477) from 2017. I’m going to cry :D Thank you!

editor/src/clj/editor/curve_view.clj
```diff
@@ -588,20 +588,37 @@
...
+                                 {:active-plane :z
+                                  :color [1.0 1.0 1.0 1.0]
+                                  :axes-colors {:x [1.0 1.0 1.0 1.0]
+                                                :y [1.0 1.0 1.0 1.0]
+                                                :z [1.0 1.0 1.0 1.0]}
+                                  :opacity 0.1
+                                  :auto-scale false
+                                  :size {:x 0.2
+                                         :y y-size
+                                         :z 1}}))))
```
sprocketc, line 626: I tried to get as close as possible to the existing curve grid. This provides a dynamic y grid size that is based on the camera zoom level and the viewport. The scaling is not equally smooth, but it has the upside that it also adjusts based on the viewport height, to avoid excessive amount of horizontal grid lines. The white vertical line at x 1.0 is now a simple grid line, but I think the colors are close enough. I can probably find a way to render an additional primary axis, if this is important. Reverting `curve_grid.clj` is also an option.

editor/src/clj/editor/gl/pass.clj
```diff
@@ -41,9 +41,9 @@
 (make-passes
   ; name selection model-transform depth-clipping
   background            false false false
-  infinity-grid         false true  false
   opaque                false true  true
   transparent           false true  true
+  infinity-grid         false true  false
```
sprocketc, line 46: I moved the grid pass above the scene items. I think it makes sense to render the grid on top of your scene. A substantial amount of objects may occlude the grid, if we render it on the background. Now that the toggle is a top level toolbar control, it is easy to deactivate it, and we also added a default shortcut. We can make this configurable if we receive complains, but I don't think we should do that yet.

editor/src/clj/editor/prefs.clj
```diff
@@ -211,7 +211,20 @@
...
+                                        :scope :project
+                                        :properties
+                                        {:x {:type :integer :default 10}
+                                         :y {:type :integer :default 10}
+                                         :z {:type :integer :default 10}}}
+                                 :active-plane {:type :keyword :default :z}
+                                 :opacity {:type :number :default 0.25}
+                                 :color {:type :tuple
+                                         :items [{:type :number} {:type :number} {:type :number} {:type :number}]
+                                         :default [0.5 0.5 0.5 1.0]}}}}}
```
sprocketc, line 227: The `active-plane` could be just `plane`. Not sure if it is clear enough.

I also changed my mind a couple of times about the type of the color. The alpha channel can be omitted, or the whole thing could be a string. In the end, I decided to use the type that the grid used to expect. We also seem to use this type internally for colors in general.

The default color is gray and the opacity is `0.25`. It looks similar to what we already had, and it is distinguishable on various types of scenes (dark/light).

editor/src/clj/editor/scene_visibility.clj
```diff
@@ -368,15 +365,15 @@
...
+    (Point2D. (- (.getMaxX container-screen-bounds) width)
               (+ (.getMaxY container-screen-bounds) y-gap))))
 
 (defn show-visibility-settings! [app-view ^Parent owner scene-visibility]
   (if-let [popup ^PopupControl (ui/user-data owner ::popup)]
     (.hide popup)
     (let [[^Region toggles update-fn] (make-visibility-toggles-list app-view scene-visibility)
           popup (popup/make-popup owner toggles)
-          anchor (pref-popup-position owner (.getMinWidth toggles) -5)
+          anchor (pref-popup-position owner (.getMinWidth toggles) 10)
```
sprocketc, line 376: I moved the `visibility-settings` popup down and to the right, to make it consistent with the grid settings. I am not sure why it used to be like that, but I couldn't think of a reason to maintain the overlap (see before/after below)

Screenshot From 2025-06-17 17-52-59
Screenshot From 2025-06-17 12-14-17

editor/src/clj/editor/scene_visibility.clj
```diff
@@ -399,9 +396,13 @@
   (run [scene-visibility] (toggle-tag-visibility! scene-visibility :outline)))
 
 (handler/defhandler :scene.visibility.toggle-grid :workbench
-  (active? [scene-visibility evaluation-context]
-           (g/node-value scene-visibility :active-scene-resource-node evaluation-context))
-  (run [scene-visibility] (toggle-tag-visibility! scene-visibility :grid)))
+  (active? [app-view scene-visibility evaluation-context]
+           (and (g/node-value scene-visibility :active-scene-resource-node evaluation-context)
+                (when-let [active-view (g/node-value app-view :active-view evaluation-context)]
+                  (some? (g/node-value active-view :grid evaluation-context)))))
```
sprocketc, line 402: This filters out the toggle on scene that don't have a grid attached.

editor/src/clj/editor/tile_map.clj
```diff
@@ -1547,6 +1547,24 @@
...
 
+(g/defnode TileMapGrid
+  (inherits grid/Grid)
+  (input grid-size g/Any :substitute nil)
+  (output options g/Any (g/fnk [grid-size]
+                          {:active-plane :z
+                           :auto-scale false
+                           :size {:x (first grid-size)
+                                  :y (second grid-size)
+                                  :z 1}})))
```
sprocketc, line 1558: We feed the tile size to the grid. Zooming out after a certain point, used to make the editor hang. The grid will now be hidden to avoid this.

editor/src/clj/editor/ui.clj
```diff
@@ -1803,6 +1803,69 @@
...
+                              (on-action! (fn [_event]
+                                            (execute-command (contexts scene) command (:user-data menu-item)))))]
+            (.add (.getChildren group) button)
+            (.add (.getChildren group) more-button)
+            (when id (.setId more-button (name id)))
+            (observe (.selectedProperty button)
+                     (fn [_observable _old-val new-val]
+                       (if new-val
+                         (add-style! group "active")
+                         (remove-style! group "active"))))
```
sprocketc, line 1886: That's the actual change. We pass a more key if we want to display an additional dropdown button within the toolbar control.

editor/src/clj/editor/app_view.clj
```diff
@@ -431,6 +431,21 @@
         (ui/remove-style! btn "filters-active"))
       (scene-visibility/settings-visible? btn))))
 
+(defn get-grid-settings-button
+  [app-view]
+  (some-> ^TabPane (g/node-value app-view :active-tab-pane)
+          ui/selected-tab
+          .getContent
+          (.lookup "#show-grid-settings")))
```
sprocketc, line 445: Not elegant. I am following the example above. Both dropdowns seem to misbehave from time to time. The click event sometimes does not trigger the popup. If you try to toggle the visibility settings multiple times in a row, you will probably reproduce this. The state of the toggle might also be invalid in this case. The problem is less prominent on the grid toggle, because its state depends on the grid. Created a dedicated issue here #10853

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+        grid-id (g/node-value scene-view-id :grid)]
+    (g/transact [(g/invalidate-output grid-id :grids)])))
+
+(defmulti settings-row (fn [_app-view _prefs option] option))
+
+(defmethod settings-row :opacity
+  [app-view prefs option]
+  (let [prefs-path (conj grid-prefs-path option)
+        value (prefs/get prefs prefs-path)
+        slider (Slider. 0.0 1.0 value)
```
sprocketc, line 270: There is a javafx related problem with this slider, and I can't find a proper way to fix it. If we start dragging the slider, and then release the mouse within the bounds of the setting popup, the second time we will try to use it, it won't work when the pointer is outside of the slider handle. The issue seems to be related to the `autoHide` feature of the popup. I tried to disable it and implement a custom autohide, bit I wasn't satisfied with the result. I also tried to filter the mouse events to prevent this, but I couldn't make it work.

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+    (->> [:size :active-plane :color :opacity]
+         (e/remove (partial contains? options))
+         (reduce (fn [rows option]
+                   (conj rows (doto (HBox. 5 (ui/node-array (settings-row app-view prefs option)))
+                                (.setAlignment Pos/CENTER))))
+                 [reset-btn]))))
+
+(defn- pref-popup-position
+  ^Point2D [^Parent container]
+  (Utils/pointRelativeTo container 0 0 HPos/RIGHT VPos/BOTTOM -190.0 10.0 true))
```
sprocketc, line 383: Not sure if it's worth trying to avoid the hardcoded `190` here, since I already had to set a fixed `pref-width` for the popup.

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+                    (do (let [active-plane (-> (.getText new-value) string/lower-case keyword)]
+                          (prefs/set! prefs prefs-path active-plane))
+                        (invalidate-grids! app-view))
+                    (.setSelected old-value true))))
+    (concat [label] buttons)))
+
+(defmethod settings-row :color
+  [app-view prefs option]
+  (let [prefs-path (conj grid-prefs-path option)
+        text-field (TextField.)
```
sprocketc: I wanted to avoid showing a popover within a popover, so I decided to display a simple text field fro the color. I am not thrilled about this.

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
-    (.invert m)
-    (.transform m v)
-    v))
+        v (Point3d. 0.0 (- (.w plane1)) (- (.w plane2)))
+        det (.determinant m)]
+    (if (< (Math/abs det) 1e-10)
+      (Point3d. 0.0 0.0 0.0)
+      (do (.invert m)
+          (.transform m v)
+          v))))
```
sprocketc, line 172: While rotating the camera, `invert` would throw when the grid was perpendicular to the viewport. The reason seems to be that the matrix cannot be inverted, If the determinant is zero. Also, the editor would sometimes hang while getting close to this point. Providing a default when the determinant approaches zero seems to fix both cases.

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+    (->> [:size :active-plane :color :opacity]
+         (e/remove (partial contains? options))
+         (reduce (fn [rows option]
+                   (conj rows (doto (HBox. 5 (ui/node-array (settings-row app-view prefs option)))
+                                (.setAlignment Pos/CENTER))))
+                 [reset-btn]))))
+
+(defn- pref-popup-position
+  ^Point2D [^Parent container]
+  (Utils/pointRelativeTo container 0 0 HPos/RIGHT VPos/BOTTOM -200.0 10.0 true))
```
sprocketc: I don't know if it's worth trying to avoid the hardcoded `200.0`, since the `pref-width` of the popup is also set to fixed `200px`.

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+     :aabbs (cond-> [(snap-out-to-grid aabb grid-size-small)]
+              grid-size-large
+              (conj (snap-out-to-grid aabb grid-size-large)))
+     :plane plane}))
+
+(g/defnk produce-merged-options
+  [prefs camera options]
+  (cond-> (if prefs (prefs/get prefs grid-prefs-path) {})
+    :always
+    (assoc :auto-scale true)
```
sprocketc: The auto scale is not part of the grid prefs (it doesn't need to be), so this provides the default.

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+      (geom/aabb-incorporate (frustum-plane-projection (nth planes 1) (nth planes 3) plane))))
 
 (defn grid-ratio [extent]
   (let [exp (Math/log10 extent)]
     (- 1.0 (- exp (Math/floor exp)))))
 
-(defn small-grid-size [extent]  (Math/pow 10.0 (dec (Math/floor (Math/log10 extent)))))
-(defn large-grid-size [extent]  (Math/pow 10.0      (Math/floor (Math/log10 extent))))
+(defn small-grid-axis-size [^double extent ^double size] (* size (Math/pow 10 (- (Math/floor (Math/log10 extent)) 2))))
+(defn large-grid-axis-size [^double extent ^double size] (* size (Math/pow 10 (- (Math/floor (Math/log10 extent)) 1))))
```
sprocketc: This is the core of the auto-scale calculation, in case anyone wants to suggest or try something different.

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+  [prefs camera options]
+  (cond-> (if prefs (prefs/get prefs grid-prefs-path) {})
+    :always
+    (assoc :auto-scale true)
+
+    options
+    (merge options)
+
+    (c/mode-2d? camera)
+    (assoc :active-plane :z)))
```
sprocketc: When the 2D mode is enabled, we always display a z plane grid. The plane option on the popup won't be in sync with this. It will be available, and show the currently selected pane as active.

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+(defn grid-snap-up   [^double a ^double sz] (* sz (Math/ceil  (/ a sz))))
 
-(defn grid-snap-down [a sz] (* sz (Math/floor (/ a sz))))
-(defn grid-snap-up   [a sz] (* sz (Math/ceil  (/ a sz))))
+(defn grid-size
+  [size auto-scale extent grid-axis-size-fn]
+  (into {}
+        (map (fn [[k ^double v]]
+               [k (max v (cond->> v auto-scale ^double (grid-axis-size-fn extent)))])
+             size)))
```
sprocketc, line 201: Grid sizes smaller that the size provided by the user will not be displayed. I think this makes sense, and it will also be consistent with the snapping behavior that will be added.

editor/src/clj/editor/scene_visibility.clj
```diff
@@ -48,8 +48,7 @@
            {:label "Text" :tag :text}
            {:label "Tile Maps" :tag :tilemap}
            {:label :separator}
-           {:label "Component Guides" :tag :outline :command :scene.visibility.toggle-component-guides :always-enabled true}
-           {:label "Grid" :tag :grid :command :scene.visibility.toggle-grid :always-enabled true}]
```
sprocketc, line 52: The grid toggle was remove from the visibility settings.

editor/src/clj/editor/ui.clj
```diff
@@ -1878,18 +1908,27 @@
...
+                      item (.getSelectedItem selection-model)]
+                  (when (not= item state)
+                    (binding [*programmatic-selection* true]
+                      (.select selection-model state))))))
+
+            :else
+            (let [toggle-button (.get (.getChildren box) 0)]
+              (if (handler/state handler-ctx)
+                (.setSelected ^Toggle toggle-button true)
+                (.setSelected ^Toggle toggle-button false)))))))))
```
sprocketc, line 1952: Handles the state of the grid toggle on refresh.

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+    {:ratios [first-grid-ratio (- 1.0 ^double first-grid-ratio)]
+     :sizes [grid-size-small grid-size-large]
+     :aabbs (cond-> [(snap-out-to-grid aabb grid-size-small)]
+              grid-size-large
+              (conj (snap-out-to-grid aabb grid-size-large)))
+     :plane plane}))
+
+(g/defnk produce-merged-options
+  [prefs camera options]
+  (cond-> (if prefs (prefs/get prefs grid-prefs-path) {})
```
vlaaad, line 233: When can prefs be nil here?

editor/src/clj/editor/keymap.clj
```diff
@@ -193,6 +194,7 @@
            ["Ctrl+G" :code.find-next]
            ["Ctrl+H" :code.replace-text]
            ["Ctrl+H" :scene.visibility.toggle-component-guides]
+           ["Ctrl+G" :scene.visibility.toggle-grid]
```
vlaaad: Maybe move this line 2 lines up so it's near another `Ctrl+G` shortcut?

editor/src/clj/editor/keymap.clj
```diff
@@ -308,6 +310,7 @@
...
```
vlaaad: same here

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
-                      alpha (Math/abs (* (aget dir axis) ratio))]]
-     (do
-       (gl/gl-color gl (colors/alpha grid-color alpha))
-       (render-grid gl axis
-                    (nth (:sizes grids) grid-index)
-                    (nth (:aabbs grids) grid-index))))))
+  [^GL2 gl ^doubles dir grids options]
+  (let [{:keys [^double opacity color auto-scale]} options]
+    (doall
+     (for [grid-index (range (if auto-scale 2 1))
```
vlaaad: I understand the code was like that before, but still, it would be nice if we used `(doseq ...)` instead of `(doall (for ...))` when we don't need the actual sequence... We don't need it, right?

editor/src/clj/editor/prefs.clj
vlaaad: The fourth number is opacity, right? But we also have a separate opacity pref... maybe we should only keep 3 numbers here, given that the 4th number is supposedly unused?

vlaaad (dismissed):
This is simply awesome! I love it!

One note: using <kbd>Tab</kbd> to switch between size text fields (X/Y/Z) does not work for me. I'd expect to set the grid size by typing a number in the X field, then pressing Tab, then typing a number in the Y field, then pressing Tab again, and finally typing a number in the Z field and pressing Enter. It would be nice if this got fixed, but it could be done in a separate PR — the grid is so great!

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+           :let [^double fixed-axis (:plane grids)
+                 ^double ratio (nth (:ratios grids) grid-index)
+                 alpha (cond-> opacity
+                         auto-scale
+                         (->> (* ^double (aget dir fixed-axis) ratio)
+                              (Math/abs)))
```
matgis: Style: We indent macros like `cond->` and `->>` the same way.  Everything aligned to the first argument, like it is for `->>` here.

editor/src/clj/editor/grid.clj
sprocketc: Yeah, I was thinking about adding a comment here (I probably should). `Prefs` are a property of the `Grid`. We pass them to the grid on `editor/scene.clj` (see https://github.com/defold/defold/pull/10345/files#diff-1661e0896d2a5cabc1cc1222e277c2fb745b21efddef8a47300c89b7ad0423ffR1651), but they are not guaranteed to exists, because they are passed through the scene `opts`.

editor/src/clj/editor/grid.clj
matgis: I have long wanted to have our own compact color wheel control for stuff like this. I think it may be neat to have it directly inlined in the Properties panel as well. Or have it pop out on demand instead of the system color picker dialog that we use today. One day!

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+  [app-view prefs option]
+  (let [prefs-path (conj grid-prefs-path option)]
+    (into []
+          (comp (map (partial axis-group app-view prefs prefs-path))
+                (mapcat identity))
+          axes)))
+
+(defn- reset-button
+  [app-view prefs ^PopupControl popup]
+  (let [button (doto (javafx.scene.control.Button. "Reset to defaults")
```
matgis: We use AP-style Title Case for labels: "Reset to Defaults"

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+                   (doseq [path [[:size :x]
+                                 [:size :y]
+                                 [:size :z]
+                                 [:active-plane]
+                                 [:opacity]
+                                 [:color]]]
+                     (let [path (into grid-prefs-path path)]
+                       (prefs/set! prefs path (:default (prefs/schema prefs path)))))
+                   (invalidate-grids! app-view)
+                   (.hide popup))]
```
matgis: I got a bit annoyed that clicking the Reset button would also hide the popup. Could we make it only reset the settings instead?

editor/src/clj/editor/grid.clj
matgis: I think maybe the `anchorLocation` property of the `PopupWindow` is intended to allow this sort of positioning? Not sure if it is settable from CSS, but there's a setter method at least.

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
-    {:ratios [first-grid-ratio                        (- 1.0 first-grid-ratio)]
-     :sizes  [grid-size-small                         grid-size-large]
-     :aabbs  [(snap-out-to-grid aabb grid-size-small) (snap-out-to-grid aabb grid-size-large)]}))
+        grid-size-small (grid-size size auto-scale smallest-extent small-grid-axis-size)
+        grid-size-large (when auto-scale (grid-size size auto-scale smallest-extent large-grid-axis-size))]
+    {:ratios [first-grid-ratio (- 1.0 ^double first-grid-ratio)]
+     :sizes [grid-size-small grid-size-large]
+     :aabbs (cond-> [(snap-out-to-grid aabb grid-size-small)]
+              grid-size-large
+              (conj (snap-out-to-grid aabb grid-size-large)))
```
matgis: Style: `cond->` indentation.

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+     :plane plane}))
+
+(g/defnk produce-merged-options
+  [prefs camera options]
+  (cond-> (if prefs (prefs/get prefs grid-prefs-path) {})
+    :always
+    (assoc :auto-scale true)
+
+    options
+    (merge options)
```
matgis: Style: `cond->` indentation.

editor/src/clj/editor/scene.clj
```diff
@@ -1616,8 +1617,9 @@
 (defmethod attach-grid :editor.grid/Grid
   [_ grid-node-id view-id resource-node camera]
   (concat
-    (g/connect grid-node-id :renderable view-id      :aux-renderables)
-    (g/connect camera       :camera     grid-node-id :camera)))
+    (g/connect grid-node-id :_node-id        view-id      :grid)
+    (g/connect grid-node-id :renderable      view-id      :aux-renderables)
+    (g/connect camera       :camera          grid-node-id :camera)))
```
matgis: Style: The fact that you have to "fix" three lines instead of just adding a line is why we prefer to not align column values like this anymore. Let's take this opportunity to remove the column alignment here!

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+        value (prefs/get prefs prefs-path)
+        slider (Slider. 0.0 1.0 value)
+        label (doto (Label. "Opacity")
+                (.setPrefWidth 70))]
+    (ui/observe
+     (.valueProperty slider)
+     (fn [_observable _old-val new-val]
+       (let [val (math/round-with-precision new-val 0.01)]
+         (prefs/set! prefs prefs-path val)
+         (invalidate-grids! app-view))))
```
matgis: Style: We indent function arguments with two spaces when they are on their own line.

editor/src/clj/editor/scene_visibility.clj
matgis: Ohh, that's very nice! Looks a lot better!

matgis (dismissed):
Very nice! Love the added functionality and the fact that the various grid implementations have been unified. I found a few minor code style issues, but nothing that needs immediate attention before merging. I do have some suggestions for further improvements, though!

I think the automatic grid size calculation is a bit strange. If I use tiny values for the grid steps, it draws a tremendous number of lines when I zoom out. It seems to me that the density of grid lines we draw when zoomed out shouldn't change with smaller values - it really should only change the density at the zoomed-in level, right?

I was also thinking that maybe we should have separate grid settings for 2D and 3D mode, given how 2D scenes typically use 1 unit per pixel, whereas 3D scenes use 1 unit per meter? I would expect users to often want to use something like a 10 x 10 grid for 2D scenes but a 0.1 x 0.1 grid for 3D scenes.

Some UX feedback:
* Like Vlad already mentioned, I'd love to be able to Tab between the text fields in the grid popup!
* When I focus on a text field, I expect the text to be selected, so I can start typing into it immediately (In `properties_view.clj` we have a `customize!` method that does this to the `TextFields`).
* The styling of the controls in the popup deviates from the style we use elsewhere for controls hosted on a brighter background. For example, see the **Open from Disk...** button and the **Title** input field in the Welcome dialog.
* The Opacity slider control looks bad to me. The drag handle color is too similar to the popup background, and the black line below the track looks almost like a graphical glitch. I feel like we need to style it differently for use on a lighter background like this.
* The height of the row that hosts the Opacity setting needs to be increased, so it is the same as the other setting rows. Right now the distances between the labels appear inconsistent.
* There appears to be some slight margin or padding differences between the Color field and the other controls that cause them to not align properly.
* The color of the selected Plane is a small amount brighter than the other controls, not sure if intentional? Should we use the blue button style for the selected one instead, like the **Create Project** button in the Welcome dialog?
* Visually, the space between the grid icon and the dropdown arrow in the toolbar button feels a bit wide to me.

editor/src/clj/editor/grid.clj
```diff
@@ -13,192 +13,385 @@
...
+
+(defn- axis-group
+  [app-view prefs prefs-path axis]
+  (let [text-field (TextField.)
+        label (Label. (string/upper-case (name axis)))
+        size-val (str (get (prefs/get prefs prefs-path) axis))
+        cancel-fn (fn [_] (ui/text! text-field size-val))
+        update-fn (fn [_] (try
+                            (let [value (some-> (.getText text-field) Float/parseFloat)]
+                              (if (pos? value)
```
matgis, line 343: I get a boxed math warning from this line. I don't think the `some->` clause above is warranted, since `.getText` should always return a string, and `Float/parseFloat` will in fact throw an exception if the text can't be interpreted as a `float`. 🤔

editor/src/clj/editor/grid.clj
sprocketc: As it turns out, this was a Linux specific issue. I pushed a hacky fix to handle that.

sprocketc:
@matgis 

> I think the automatic grid size calculation is a bit strange. If I use tiny values for the grid steps, it draws a tremendous number of lines when I zoom out. It seems to me that the density of grid lines we draw when zoomed out shouldn't change with smaller values - it really should only change the density at the zoomed-in level, right?

I added an additional normalization step to mitigate this. The reason I didn't do this previously, is that using the zoom level is not enough to avoid this, although that kind of depends on what you expect to see. For example, setting `x` to `0.1` and `y` to `10` would now produce a 1/1 grid, until you reach a zoom level that allows you to see the actual grid. I don't think there is a way to avoid high density lines (or no lines) for both axes, without altering their ratio, but I might be wrong. Using a common 1-10 range would look fine though.

> I was also thinking that maybe we should have separate grid settings for 2D and 3D mode, given how 2D scenes typically use 1 unit per pixel, whereas 3D scenes use 1 unit per meter? I would expect users to often want to use something like a 10 x 10 grid for 2D scenes but a 0.1 x 0.1 grid for 3D scenes.

I think that having dedicated grid prefs per scene type (or even per scene) might make more sense, but I am not sure if we want to go that far. I wouldn't expect the grid to change while working on the same scene. I believe that the per pixel / per meter analogy emerges from the fact that we use `pow10` and `log10` to generate the sub-grids, and that resembles the metric system. Using power of 2 might make more sense in some cases (GUI scenes?), in order to get 2->4->8->16...128 instead of 2->20->200, but that would require an additional pref (per axis?), and the zoom experience wouldn't be smooth, because we would only split by half.

> Should we use the blue button style for the selected one instead, like the Create Project button in the Welcome dialog?

That was the only suggestion that looked a little distracting. The rest were handled.

matgis:
Regarding grid preferences by scene type, I don't think it will make sense. While we can assume most Gui resources will likely be authored in pixels, we cannot make assumptions about the content inside collections and game objects. They might even contain a mix of 2D and 3D elements. I don't think this needs to be addressed in this PR, though.

matgis (dismissed):
Lovely! Let's merge it! ❤️

It seems the intensity of the grid can vary quite a bit between different plane alignments when using a perspective camera. I think we may want to perhaps fade out distant lines instead of the entire grid in this mode, but that is definitely a later enhancement!

vlaaad (approved):
Great work! Ship it!

PR #10351, opened 2025-03-18
Merge shader resources

Jhonnyg:
Shader resources can now be merged between different stages of the same shader program:

my_shader.vp:
```glsl
uniform shared_uniforms {
    vec4 tint;
};
uniform sampler2D shared_texture;
void main()
{
    vec4 shared_color = texture(shared_texture, vec2(texcoord0));
    gl_Position = view_proj * vec4(position.xyz * tint.x + shared_color.r * 0.001, 1.0);
    var_texcoord0 = texcoord0;
}
```
my_shader.fp:
```glsl
uniform shared_uniforms {
    vec4 tint;
};
uniform sampler2D shared_texture;
void main()
{
    vec4 shared_color = texture(shared_texture, var_texcoord0.st);
    color = tint * shared_color;
    color.a = 1.0;
}
```

Previously, shared_uniforms and shared_texture would be used as two separate resources in the engine, which means that the same resource would be bound twice even though they are supposed to be used as a single resource.

Furthermore, the pipeline will now also make sure that outputs from a shader stage matches the inputs of the next shader stage (e.g vertex shader + fragment shader). Previously, this example would cause non-opengl adapters to behave incorrect:

my_shader.vp:
```glsl
out vec4 out_position;
out vec3 out_normal;
out vec2 out_uv;
```

my_shader.fp
```glsl
in vec3 out_normal;
in vec2 out_uv;
in vec4 out_position;
```

With the new remapping system, the locations of the outputs from the vertex shader will correctly match the inputs of the fragment shader.

Note: Remapping of resources will currently only work with the "new shader pipeline". To migrate your old shaders to the new pipeline, head over to https://defold.com/manuals/shader/#writing-modern-glsl-shaders to read more.


com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ShaderProgramBuilderTest.java
```diff
@@ -783,7 +783,7 @@ public void testNativeReflectionAPI() throws Exception {
         byte[] spvSimple = getFile("simple.spv");
         assertNotNull(spvSimple);
 
-        long ctx = ShadercJni.NewShaderContext(spvSimple);
+        long ctx = ShadercJni.NewShaderContext(Shaderc.ShaderStage.SHADER_STAGE_FRAGMENT.getValue(), spvSimple);
```
Jhonnyg, line 786: The "stage" is moved from the crosscompile function into creating the context constructor. This makes more sense as we already know upfront which stage it should be compatible with.

engine/shaderc/src/shaderc.h
```diff
@@ -39,7 +39,7 @@ namespace dmShaderc
     {
         SHADER_STAGE_VERTEX   = 1,
         SHADER_STAGE_FRAGMENT = 2,
-        SHADER_STAGE_COMPUTE  = 3,
+        SHADER_STAGE_COMPUTE  = 4,
```
Jhonnyg, line 42: Need to be able to use the stage as a mask

engine/shaderc/src/shaderc.h
```diff
@@ -135,7 +135,6 @@ namespace dmShaderc
         ShaderCompilerOptions()
         : m_Version(330)
         , m_EntryPoint("main")
-        , m_Stage(SHADER_STAGE_VERTEX)
```
Jhonnyg, line 138: This is removed from the options. This will probably require a new release of PS4/PS5 extensions for next release!

JCash (approved):

PR #10355, opened 2025-03-19
Use vulkan as default renderer on osx

### LINKED ISSUE #10157: Make Vulkan the default renderer on MacOS

Jhonnyg:
Vulkan is now the default renderer on OSX (x86_64-macos and ram64-macos).

Q: What are the benefits of using Vulkan?
A: Many! These are some typical points:

* OpenGL is deprecated on osx. This means that OpenGL can get fully removed from the mac ecosystem at any point in time.
* More new potential features (compute shaders are supported by default, among many other modern features)
* Improved rendering performance (mainly for 3D)
* You can profile the rendering in Xcode to get accurate timings for a frame
* In general reduced runtime memory footprint
* OpenGL is complete and will likely not get any more updates in the future. Using modern tech means we can add and improve rendering in the future.

Q: I don't want to use Vulkan, can I use OpenGL?
A: Yes, you can use an app manifest to switch back to OpenGL. Read more about app manifests here: https://defold.com/manuals/app-manifest/

Fixes #10157


vlaaad (dismissed):
Editor changes look good!

engine/graphics/src/vulkan/graphics_vulkan_context.cpp
```diff
@@ -61,7 +61,7 @@ namespace dmGraphics
     {
         if (!IsSupported(supported_extensions, extension))
         {
-            dmLogWarning("%s was not supported", extension);
+            dmLogDebug("%s was not supported", extension);
```
Jhonnyg, line 64: I think people will be annoyed by this, and we only add extensions if they are supported. Meaning, they are not required extensions to do rendering.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderCompilers.java
```diff
@@ -199,7 +208,24 @@ public ShaderProgramBuilder.ShaderCompileResult compile(ArrayList<ShaderCompileP
         }
     }
 
-    public static IShaderCompiler getCommonShaderCompiler(Platform platform) {
+    public static IShaderCompiler GetCommonShaderCompiler(Platform platform) {
         return new CommonShaderCompiler(platform);
     }
+
+    public static ArrayList<ShaderDesc.Language> GetSupportedOpenGLVersionsForPlatform(Platform platform) {
```
Jhonnyg, line 215: Maybe we should use something like a static feature table (e.g like Platform.java has) for these things instead of multiple functions.. For now it's not a priority, but this could be data driven I think.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderProgramBuilder.java
```diff
@@ -125,28 +131,34 @@ public void build(Task task) throws IOException, CompileExceptionError {
...
         }
         if (getOutputSpirvFlag()) {
-            compileOptions.forceIncludeShaderLanguages.add(ShaderDesc.Language.LANGUAGE_SPIRV);
+            addUniqueShaderLanguage(ShaderDesc.Language.LANGUAGE_SPIRV);
         }
         if (getOutputWGSLFlag()) {
-            compileOptions.forceIncludeShaderLanguages.add(ShaderDesc.Language.LANGUAGE_WGSL);
+            addUniqueShaderLanguage(ShaderDesc.Language.LANGUAGE_WGSL);
+        }
+        if (getOutputGLSLFlag()) {
```
Jhonnyg, line 142: This flag, compared to the specific language flags, adds the GLSL languages that is supported by the platform we are building for, and not just specific languages (which is more for tests).

editor/src/clj/editor/app_manifest.clj
```diff
@@ -532,13 +547,21 @@
             (value (setting-property-getter use-android-support-lib-setting))
             (set (setting-property-setter use-android-support-lib-setting)))
   (property graphics g/Any
-            (dynamic tooltip (g/constantly "Vulkan support is in BETA (desktop and mobile platforms)"))
```
Jhonnyg, line 535: It's not beta any more!

engine/graphics/src/null/graphics_null.cpp
```diff
@@ -852,10 +852,12 @@ namespace dmGraphics
         {
             case ShaderDesc::SHADER_TYPE_VERTEX:
                 return language == ShaderDesc::LANGUAGE_GLSL_SM330 ||
-                       language == ShaderDesc::LANGUAGE_GLES_SM300;
+                       language == ShaderDesc::LANGUAGE_GLES_SM300 ||
+                       language == ShaderDesc::LANGUAGE_SPIRV;
```
Jhonnyg, line 856: TODO (future): The null adapter should support all languages and a mechanism to turn on/off specific languages, so that we can test more things. The actual shader code is never run anyway.

Jhonnyg:
IRL reviewed by @JCash

PR #10363, opened 2025-03-21
Fix menu bar visibility on toolbar refresh

### LINKED ISSUE #10361: The Main App Menu disappears.
### Expected behaviour

The Main App Menu does not disappear.

### Actual behaviour

https://youtu.be/kBpgU1hF7g4  (non-public link)

When I add a Node / Box in a gui file (say, game.gui) and I want to make a button out of it, e.g. I add an icon that I want to scale then:
- selecting the Size Mode drop-down list and changing it to Manual or to Auto makes the main program menu disappear.
Clicking anywhere on the program window restores the menu.
I haven't noticed the problem with other drop-down lists.

If the application is active (focus on the application), interaction with the keyboard (ex. esc key) or mouse (ex. LPM) causes the menu to appear.

### Steps to reproduce

https://youtu.be/kBpgU1hF7g4

1. add Gui file
2. add Nodes/Box
3. add a texture, let's say 50x50
4. try to change Size Mode to Manual and v-versa
5. Main App Menu disappears.

Linux Mint, Defold 1.9.8

<hr/>

<table>
<tr><td>Build time</td><td>2025-03-11T20:28:12.023660</td></tr>
<tr><td>Defold editor sha</td><td>f68db9583283029bd2c2425839119f34c253143b</td></tr>
<tr><td>Defold engine sha</td><td>67542769598a1b794877c96f740f3f527f63f491</td></tr>
<tr><td>Defold version</td><td>1.9.8</td></tr>
<tr><td>GPU</td><td>Mesa Intel(R) HD Graphics 630 (KBL GT2)</td></tr>
<tr><td>GPU Driver</td><td>4.6 (Compatibility Profile) Mesa 23.2.1-1ubuntu3.1~22.04.3</td></tr>
<tr><td>OS name</td><td>Linux</td></tr>
<tr><td>OS version</td><td>5.15.0-134-generic</td></tr>
<table>

sprocketc:
This fixes an occasional menu bar visibility issue, triggered by a toolbar refresh. 

Fixes #10361

### Technical changes
Refreshing the toolbars was triggering a command execution, when the selection of the ChoiceBox was programmatically set. That was affecting the focus state and the context of the menu-bar. Although this looks linux specific, it is also reproducible on macOS (and probably windows). The difference is that on mac, the custom title-bar is not visible, so the problem is less prominent. Hiding the native menu does not look as bad as the case on video of the reported issue, but it's part of the same problem. Note that the initial trigger of the reported problem is calling `(ui/user-data! (ui/main-scene) ::ui/refresh-requested? true)`, so whenever we attempt to automatically refresh the UI, there is a chance to face this issue.

### DIFF
--- a/editor/src/clj/editor/outline_view.clj
+++ b/editor/src/clj/editor/outline_view.clj
@@ -488,7 +488,6 @@
         image-view-icon (ImageView.)
         visibility-button (doto (ToggleButton.)
                             (ui/add-style! "visibility-toggle")
-                            (.addEventFilter MouseEvent/MOUSE_PRESSED ui/ignore-event-filter)
                             (AnchorPane/setRightAnchor 0.0))
         text-label (doto (Label.)
                      (ui/bind-double-click! :open))
--- a/editor/src/clj/editor/properties_view.clj
+++ b/editor/src/clj/editor/properties_view.clj
@@ -625,7 +625,7 @@
         listen-fn (fn [_old-val new-val]
                     (when-not *programmatic-setting*
                       (properties/set-values! (property-fn) (repeat new-val))
-                      (ui/user-data! (ui/main-scene) ::ui/refresh-requested? true)))]
+                      (ui/refresh (ui/main-scene))))]
     (fuzzy-combo-box/observe! combo-box listen-fn)
     [combo-box update-ui-fn]))
 
--- a/editor/src/clj/editor/properties_view.clj
+++ b/editor/src/clj/editor/properties_view.clj
@@ -625,7 +625,7 @@
         listen-fn (fn [_old-val new-val]
                     (when-not *programmatic-setting*
                       (properties/set-values! (property-fn) (repeat new-val))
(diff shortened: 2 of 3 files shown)


editor/src/clj/editor/outline_view.clj
```diff
@@ -488,7 +488,6 @@
         image-view-icon (ImageView.)
         visibility-button (doto (ToggleButton.)
                             (ui/add-style! "visibility-toggle")
-                            (.addEventFilter MouseEvent/MOUSE_PRESSED ui/ignore-event-filter)
```
sprocketc, line 491: Clicking on the visibility toggle was somehow triggering this issue, and I was also trying to avoid opening the resource on double click. The latter was fixed [here](https://github.com/defold/defold/pull/10183/files#r1948029163), so this is not required.

sprocketc:
I think we could also fix this by using `onAction`, but I am not sure if we are observing the value on purpose.

matgis (approved):
LGTM!

PR #10367, opened 2025-03-22
Fix multi-editing of gui node properties

### LINKED ISSUE #9992: Selecting multiple gui nodes in the outline window hides some properties

matgis:
Fixed a regression where most Gui node properties would be hidden in the property editor when multiple Gui nodes were selected.

Fixes #9992

### DIFF
--- a/editor/src/clj/editor/gui.clj
+++ b/editor/src/clj/editor/gui.clj
@@ -856,7 +856,7 @@
         update current-layout
         assoc prop-kw new-value))))
 
-(defn layout-property-edit-type-set-impl [evaluation-context node-id prop-kw old-value new-value changes-fn]
+(defn layout-property-edit-type-set-impl [changes-fn prop-kw evaluation-context node-id old-value new-value]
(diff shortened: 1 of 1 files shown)


vlaaad (approved):
LGTM!

PR #10369, opened 2025-03-22
Debug source paths

ekharkunov:
* Set `-fdebug-compilation-dir` and `-fdebug-prefix-map` as additional compilation flags to control paths to debug sources. Expected folder structure is following:

Overall
Screenshot 2025-03-24 at 17 22 24

Inside engine
Screenshot 2025-03-24 at 17 23 06

Paths to debug sources set as relative path (without leading /) because On Extender's side I didn't find simple way to replace path to source files in the debug information inside already compiled binaries. So use relative paths helps to make path portable. Expected that debugger process will start from the root directory (see in the example).

Related to https://github.com/defold/extender/issues/259
Related to #10379 

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


engine/crash/wscript
```diff
@@ -53,6 +53,9 @@ def configure(conf):
     if operating_sys == "win":
         conf.env.append_value('LINKFLAGS', ['dbghelp.lib'])
 
+    conf.env.append_value('CFLAGS', ["-fdebug-compilation-dir=engine/crash", "-fdebug-prefix-map=../src=src", "-fdebug-prefix-map=../../../tmp/dynamo_home=../../defoldsdk"])
+    conf.env.append_value('CXXFLAGS', ["-fdebug-compilation-dir=engine/crash", "-fdebug-prefix-map=../src=src", "-fdebug-prefix-map=../../../tmp/dynamo_home=../../defoldsdk"])
+
```
JCash, line 58: These should be set in waf_dynamo.py, where we set common build settings.

Also, how will this work for windows, where we currently use cl.exe?

JCash (changes requested):

share/extender/build_input.yml
```diff
@@ -158,7 +158,7 @@ platforms:
     writeShLibPattern: 'lib%s.dylib'
     writeExePattern: 'dmengine'
     symbolsPattern: '.*dSYM'
-    zipContentPattern: 'dmengine'
+    zipContentPattern: 'dmengine|main.cpp'
```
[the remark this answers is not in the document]
ekharkunov: Zip pattern is used to determine what content should be added to build result.
IMO now isn't clear implementation:
1. Zip pattern defined in build_input.yaml
2. In Extender's code zip pattern defined as zipContentPattern + symbolsPattern + patternForSharedLibs.

My suggestion is to move over time all regex to build_input.yaml and avoid create a combination at Extender's side. That's why new stuff I added to build_input.yaml.

engine/crash/wscript
ekharkunov: `"-fdebug-compilation-dir="` has different values for different modules. `-fdebug-prefix-map=` I will move to common flags.

> Also, how will this work for windows, where we currently use cl.exe?
I understood in the way that on Windows we also use clang to build. Will recheck

share/extender/build_input.yml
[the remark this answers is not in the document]
JCash: I know what the pattern does, but I was asking "why do you need to include main.cpp to be sent back to the user?".
It's good to comment such things in a pull request.

But now that I read my question again, I realize that to be able to fully debug an application locally, it's good to have the generated sources available. Is that the reason?

Note that this also moves us towards sending back other generated files (e.g. protobuf .cpp files)

engine/crash/wscript
JCash: Ah, I missed the initial commonalities.

Still, I assume it will be a lot of errors or warnings for windows and xbox that uses cl.exe

share/extender/build_input.yml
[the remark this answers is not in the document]
ekharkunov: > I know what the pattern does, but I was asking "why do you need to include main.cpp to be sent back to the user?".

Sorry, I misunderstood initial question.

> But now that I read my question again, I realize that to be able to fully debug an application locally, it's good to have the generated sources available. Is that the reason?
Correct. Sending back generated file is one of the piece to collect all needed source in one place.

> Note that this also moves us towards sending back other generated files (e.g. protobuf .cpp files)
Thanks for highlight that! I missed that part.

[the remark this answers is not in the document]
ekharkunov: I decided to add main.cpp and other generated files on Extender's side (add generated files to list of output files, which added to build result).

engine/crash/wscript
ekharkunov: I wrapped flags with check for target OS because for Windows there are no direct analog for `-fdebug-*` flags.

build_tools/build_constants.py
```diff
@@ -0,0 +1,14 @@
+from typing import NamedTuple
+
+class TargetOSConts(NamedTuple):
```
ekharkunov, line 3: Added constants for target OS names

build_tools/waf_dynamo.py
```diff
@@ -367,29 +368,34 @@ def apply_framework(self):
 @before('process_source')
 def default_flags(self):
     build_util = create_build_utility(self.env)
+    target_os = build_util.get_target_os()
```
ekharkunov, line 371: Use precalculated variables instead of function's calls every time

build_tools/waf_dynamo.py
```diff
@@ -367,29 +368,34 @@ def apply_framework(self):
...
     if (Options.options.with_asan or Options.options.with_ubsan or Options.options.with_tsan) and opt_level != '0':
         opt_level = 1
 
-    FLAG_ST = '/%s' if 'win' == build_util.get_target_os() else '-%s'
+    FLAG_ST = '/%s' if TargetOS.WINDOWS == target_os else '-%s'
 
     # Common for all platforms
     flags = []
+    if target_os != TargetOS.WINDOWS:
+        flags += ["-fdebug-prefix-map=../src=src", "-fdebug-prefix-map=../../../tmp/dynamo_home=../../defoldsdk"]
```
ekharkunov, line 389: Common `-fdebug-prefix-map` flags for all modules

build_tools/waf_dynamo.py
```diff
@@ -1665,14 +1666,17 @@ def detect(conf):
     except BuildUtilityException as ex:
         conf.fatal(ex.msg)
 
+    target_os = build_util.get_target_os()
+    conf.env['TARGET_OS'] = target_os
```
ekharkunov, line 1670: Store target OS into environment to easier retrieve it later in every module's wscript

build_tools/waf_dynamo.py
```diff
@@ -1927,27 +1931,28 @@ def detect(conf):
...
         conf.env['LIB_PLATFORM_SOCKET'] = ''
         conf.env['LIB_DL'] = 'dl'
-    elif platform in ('x86_64-macos', 'arm64-macos', 'arm64-ios', 'x86_64-ios'):
+    elif target_os in (TargetOS.MACOS, TargetOS.IOS):
         conf.env['LIB_PLATFORM_SOCKET'] = ''
-    elif 'android' in platform:
+    elif TargetOS.ANDROID == target_os:
         conf.env['LIB_PLATFORM_SOCKET'] = ''
+    # TODO: only 32bit windows or 64 also?????
     elif platform == 'win32':
```
ekharkunov, line 1949: Q: Is it bug here or it shoule be only for 32bit windows?

engine/crash/wscript
```diff
@@ -28,14 +28,8 @@ def configure(conf):
     conf.load('waf_ddf')
     waf_ddf.configure(conf)
 
-    build_util = None
```
ekharkunov, line 31: Use stored variable from `env` instead of create and use `build_util`

engine/gamesys/wscript
```diff
@@ -26,23 +27,26 @@ def options(opt):
...
 
     waf_ddf.configure(conf)
     waf_graphics.configure(conf)
     waf_physics.configure(conf)
     waf_render.configure(conf)
 
-    platform = conf.env['PLATFORM']
-
-    if platform in ("x86_64-macos",):
+    if target_os == TargetOS.MACOS:
```
ekharkunov, line 37: I guess it was bug here to add flags only for x86_64 macos variant.

engine/modelc/wscript
```diff
@@ -45,10 +48,12 @@ def configure(conf):
         #conf.env['LIB_JDK'] = ['native']
         conf.env['DEFINES_JDK'] = ['DM_HAS_JDK']
 
-    platform = conf.env['PLATFORM']
```
ekharkunov, line 48: Unused

engine/particle/wscript
```diff
@@ -29,17 +30,20 @@ def configure(conf):
...
 
+    target_os = conf.env['TARGET_OS']
     conf.env.append_value('INCLUDES', "default/src")
     conf.env['STLIB_DLIB_NOASAN']          = ['dlib_noasan']
     conf.env['STLIB_DDF_NOASAN']           = ['ddf_noasan']
     conf.env['STLIB_GRAPHICS_NULL']        = ['graphics_null', 'graphics_proto']
     conf.env['STLIB_GRAPHICS_NULL_NOASAN'] = ['graphics_null_noasan', 'graphics_proto_noasan'] # For the shared library used in editor
     conf.env['STLIB_PLATFORM_NULL']        = ['platform_null']
 
-    platform = conf.env['PLATFORM']
```
ekharkunov, line 39: Unused

engine/render/wscript
```diff
@@ -30,15 +31,16 @@ def configure(conf):
 
     conf.recurse('src')
 
+    target_os = conf.env['TARGET_OS']
     platform = conf.env['PLATFORM']
 
-    if platform in "linux":
+    # TODO: check if here was mistake - if platform in "linux":
+    if target_os == TargetOS.LINUX:
```
ekharkunov, line 38: Q: There was no platform called `linux`. I guess it was bug here and fix it with `target_os ==`. Is it correct?

ekharkunov:
@JCash I decided to split task into 2 parts: for non-Windows and Windows platforms because `cl.exe` doesn't have direct analog for used flags.
Also I've added a named constants for target OS in build scripts to avoid mistyped of os name.

Please, pay attention to a couple of questions (marked as Q:) regarding possible bugs in build scripts.

build_tools/build_constants.py
JCash: typo "Conts" -> "Constants"

build_tools/build_constants.py
```diff
@@ -0,0 +1,14 @@
...
+    WINDOWS: str = 'win'
+    LINUX: str = 'linux'
+    ANDROID: str = 'android'
+    IOS: str = 'ios'
+    MACOS: str = 'macos'
+    WEB: str = 'web'
+    NX64: str = 'nx64'
+    PS4: str = 'ps4'
+    PS5: str = 'ps5'
```
JCash, line 12: MIssing XBONE

engine/ddf/wscript
```diff
@@ -53,6 +55,10 @@ def configure(conf):
 
     conf.env.append_unique('DEFINES', 'DLIB_LOG_DOMAIN="DDF"')
 
+    if target_os != TargetOS.WINDOWS:
```
JCash, line 58: XBONE also uses cl.exe

engine/dlib/wscript
```diff
@@ -40,6 +42,10 @@ def configure(conf):
     conf.env.append_value('LIBPATH', build_util.get_dynamo_ext('lib', build_util.get_target_platform()))
     conf.env.append_unique('DEFINES', 'DLIB_LOG_DOMAIN="DLIB"')
 
+    if target_os != TargetOS.WINDOWS:
+        conf.env.append_value('CFLAGS', ["-fdebug-compilation-dir=engine/resource"])
+        conf.env.append_value('CXXFLAGS', ["-fdebug-compilation-dir=engine/resource"])
```
JCash, line 47: An alternative would be to add this to waf_dynamo, as part of one of the "@feature` hooks instead
E.g. checking the output folder of a TaskGen and creating the path dynamically.

In its current state, it feels easy to forget or miss the nuance if it gets left out (e.g. when creating a new library)

engine/gamesys/wscript
JCash: I consider setting all these configure steps in libraries' wscript deprecated, as it just duplicates things all over the scripts. I tend to prefer putting them in waf_dynamo.py instead.
Occasionally, it's beneficial to put config in the wscript, in order to make a test link the way one wants it to.

JCash (changes requested):
XBONE is missing

engine/dlib/wscript
ekharkunov: Changes https://github.com/defold/defold/pull/10369#discussion_r2024630854

build_tools/waf_dynamo.py
JCash: looks like a bug

engine/render/wscript
JCash: Looks like an old bug yes. "linux" (32 bit) was removed some years ago.
I guess this code path wasn't used for that period of time, which suggests it isn't needed now either?

engine/render/wscript
```diff
@@ -30,15 +31,16 @@ def configure(conf):
...
         conf.env.append_value('CXXFLAGS', '-DGL_GLEXT_PROTOTYPES')
 
-    if "win32" in platform:
+    if target_os == TargetOS.WINDOWS:
         conf.env.append_value('LINKFLAGS', ['opengl32.lib', 'user32.lib', 'shell32.lib'])
```
JCash, line 42: I'm not sure this is needed, as it's been working well for x86_64 without these flags?
Try removing the code?

I believe all the necessary link flags are setup with the `SOCKET APP` in render/src/test/wscript

JCash (dismissed):
Some minor potential cleanup possibilities.

JCash (approved):

PR #10373, opened 2025-03-24
Introduce code zoom on scroll preference

### LINKED ISSUE #10371: Add a Prefrence option for the Editor to disable [Ctrl+Mouse wheel] to adjust text size
I often will hold control and mouse wheel while moving around code when copy/pasting variables. So this change has made it jarring when the text gets suddenly larger or smaller. 

This is mostly a muscle memory issue for something I do quiet often.

Add an option to the preferences to disable the ability to change the font size when holding ctrl+mouse wheel. 

Attempt to get use to it.

sprocketc:
Screenshot From 2025-03-24 14-23-03

Fixes #10371

### DIFF
--- a/editor/src/clj/editor/code/view.clj
+++ b/editor/src/clj/editor/code/view.clj
@@ -2471,12 +2471,13 @@
         (when-not (get-property view-node :hover-mouse-over-popup evaluation-context)
           (assoc (schedule-hover-refresh! view-node) :hover-cursor nil))))))
 
-(defn handle-scroll! [view-node ^ScrollEvent event]
+(defn handle-scroll! [view-node zoom-on-mouse-wheel ^ScrollEvent event]
   (.consume event)
   (when (if (.isShortcutDown event)
-          (do (-> (g/node-value view-node :canvas)
-                  (ui/run-command (cond (pos? (.getDeltaY event)) :zoom-in
-                                        (neg? (.getDeltaY event)) :zoom-out)))
+          (do (when zoom-on-mouse-wheel
+                (-> (g/node-value view-node :canvas)
+                    (ui/run-command (cond (pos? (.getDeltaY event)) :zoom-in
+                                          (neg? (.getDeltaY event)) :zoom-out))))
               true)
           (set-properties! view-node :navigation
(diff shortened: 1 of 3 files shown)


editor/src/clj/editor/code/view.clj
```diff
@@ -3810,6 +3811,7 @@
     (let [find-case-sensitive-setter (make-property-change-setter view-node :find-case-sensitive?)
           find-whole-word-setter (make-property-change-setter view-node :find-whole-word?)
           font-size-setter (make-property-change-setter view-node :font-size)
+          zoom-on-mouse-wheel-setter (make-property-change-setter view-node :zoom-on-mouse-wheel?)
```
vlaaad, line 3814: Do we need this?

editor/src/clj/editor/prefs_dialog.clj
```diff
@@ -113,7 +113,8 @@
             :prefs [{:label "Custom Editor" :type :string :key [:code :custom-editor]}
                     {:label "Open File" :type :string :key [:code :open-file]}
                     {:label "Open File at Line" :type :string :key [:code :open-file-at-line]}
-                    {:label "Code Editor Font (Requires Restart)" :type :string :key [:code :font :name]}]}
+                    {:label "Code Editor Font (Requires Restart)" :type :string :key [:code :font :name]}
+                    {:label "Zoom on mouse wheel" :type :boolean :key [:code :zoom-on-mouse-wheel]}]}
```
vlaaad, line 117: We use "Title Case" for the settings labels.

vlaaad (changes requested):
The code looks good, but it seems there were some accidental changes. We should also use Title Case for the settings label. Also, I'm not sure if "mouse wheel" is the right term to use here, because the font size change also works when using trackpad. Maybe "Zoom on Scroll"?

vlaaad (approved):
Lessgoooo!

editor/src/clj/editor/code/view.clj
sprocketc: I initially thought we did and forgot to remove it. If we the pref within the wheel handler, we won't have to reopen the view to apply the change.

PR #10378, opened 2025-03-25
Auto-insert closing parens in the code editor

### LINKED ISSUE #8988: Automatically insert closing brackets
**Task (REQUIRED):**
To reduce the "sea of red" effect of the linter, we should finally implement automatic insertion of the closing brackets.

Code editor automatically inserts closing brackets on typing open brackets. On subsequent typing of the closing bracket, it should simply move the cursor forward instead of inserting another one. The editor should not auto-insert the closing brackets inside the strings.

### LINKED ISSUE #9126: Editor support for automatically wrapping selected string with tokens

A common pattern when editing code is to select a string and hit the `"`, `'`, `[`, `(`, `{`, or `` ` `` keys to automatically wrap that string in the corresponding token.

In the Defold editor this doesn't happen, the text is simply replaced.

It would be nice if the text was instead wrapped, which would match the common behavior of other editors.

e.g. 

If I have the string `hello` selected and I hit the `"` key, the resulting text becomes `"hello"`.

Crucially, the current selection shouldnt change and `hello` stays the selected part.

Using an external editor, but I really want to use the Defold editor.

vlaaad:
record

Fixes #8988
Fixes #9126

### DIFF
--- a/editor/src/clj/editor/code/data.clj
+++ b/editor/src/clj/editor/code/data.clj
@@ -1265,6 +1265,17 @@
                    contexts))
           (persistent! syntax-info'))))))
 
+(defn syntax-scope-before-cursor [syntax-info grammar ^Cursor cursor]
+  {:pre [(< (.-row cursor) (count syntax-info))]}
+  (or
+    (let [row (.-row cursor)
+          col-before-cursor (dec (.-col cursor))
+          runs (second (get syntax-info row))]
+      (let [result-index (dec ^long (util/find-insert-index runs [col-before-cursor] #(compare (%1 0) (%2 0))))]
+        (when-let [run (get runs result-index)]
+          (second run))))
+    (:scope-name grammar "source")))
+
 (defn last-visible-row
   "Returns 1-indexed row number of the last visible row"
   [^LayoutInfo layout]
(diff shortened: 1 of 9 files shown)


matgis (approved):
Works very nicely! ❤️

I did find a small issue: If you type `\"` when inside a string, it will not insert a closing `"` character as expected. However, if you do this when you have multiple cursors inside several strings, it will insert closing `"` characters in all the places except the first one.

If this is tricky to fix, I think we can merge it like this, but if it is trivial, it would be nice to address it before merging.

vlaaad:
Could it be because you initially typed `\` before the closing `"` of a string? This then caused the syntax to change. 
I can't reproduce it if I type `\"` when `\` is not at the end of a string, e.g.:
```lua
--(pipe=cursor)
"a|sd" "a|sd" "a|sd" 
-- typed `\`
"a\|sd" "a\|sd" "a\|sd"
-- typed `"`
"a\"sd" "a\"sd" "a\"sd"
```
But I observe some weirdness when typing `\` at the end of a string:
```lua
-- initial state
"asd|""asd|""asd|"
-- typed `\`
"asd\""asd\""asd\"
-- as you can see, the first `\` escaped the previous end of a string, so the second and third `asd`s are no longer inside strings. This causes auto-inserting of closing quotes for subsequent cursors when typing `"`:
"asd\"""asd\""""asd\"""
```
I'm afraid I don't know how to fix this. The syntax changed; what can we do?

PR #10380, opened 2025-03-25
Add build server username/password prefs fields

### LINKED ISSUE #9785: Project preferences, extensions, build server, dedicated fields for username and password
In our company we are using passwords vault, so if build server url exposed in any public chat, IT have to update password.

It's really inconvenient to share urls, you have to remove password manually

It would be nice to have dedicate fields for username and password in Preferences->Extensions tab

vlaaad:
This changeset also adds a new prefs schema type in editor scripts: password, e.g.
```lua
function M.get_prefs_schema()
    return {["my-extension.third-party-api-key"] = editor.prefs.schema.password()}
end
```
Password schema behaves as a string schema, but the value is encrypted in the prefs file.

Fixes #9785

### DIFF
--- a/editor/src/clj/editor/prefs.clj
+++ b/editor/src/clj/editor/prefs.clj
@@ -22,6 +22,7 @@
     :any           anything goes, no validation is performed
     :boolean       a boolean value
     :string        a string
+    :password      a string, stored securely
     :keyword       a keyword
     :integer       an integer
     :number        floating point number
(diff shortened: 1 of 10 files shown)


vlaaad:
Mats, don't jump on this PR yet, I found an issue with prefs code...

editor/src/clj/editor/prefs.clj
```diff
@@ -22,6 +22,7 @@
     :any           anything goes, no validation is performed
     :boolean       a boolean value
     :string        a string
+    :password      a string, stored securely
```
matgis, line 25: Calling this "securely" is a bit much, seeing as how you could easily figure out how to decrypt it from the source code. Maybe something like "not stored in clear text" or "stored encrypted" instead? 🤔

matgis (dismissed):
Looks good! Maybe change the wording so we're not claiming to store the password "securely"?

PR #10384, opened 2025-03-26
Make zoom on scroll pref disabled by default

sprocketc:
### Technical changes
I think we may have to mark this as breaking, since a default functionality is going be disabled. We could also cherry pick this and the introduced pref for 1.9.9 to reduce the exposure, although I am not sure if that's going to help much.


vlaaad (approved):
LGTM! I think this is a pretty minor workflow change, we typically assign breaking labels only to changes that require users to modify their projects.

PR #10389, opened 2025-03-26
An attempt to fix null hexDigest

### LINKED ISSUE #10356: Bug in game.graph.json
It is not clear whether the bug is platform-specific, but it occurred on the iOS platform for us

This bug occurs extremely rarely; we encountered it only once in one build, so unfortunately, there are no clear steps to reproduce it. The essence of the bug is that after the build, in game.graph.json, the hexDigest field for one of the pruned resources ended up with a null value. A new build from the same commit where the bug was caught fixed the situation, which means the issue is not with the resources themselves.

Defold version:
1.9.7
639c8e2666cab8d7474dc5adccc56683dbe377b1

Platforms:
iOS

Problem file:
{
  "path" : "/main/generated/_split_group_specific/25110_patrick_p2/spine/A_decor_spring_fireplace_res_0-02.a.texturesetc",
  "hexDigest" : null,
  "isInMainBundle" : false,
  "children" : [ "/main/generated/_split_group_specific/25110_patrick_p2/spine/A_decor_spring_fireplace_res_0-02.texturec" ]
}

[game.graph.zip](https://github.com/user-attachments/files/19365585/game.graph.zip)

AGulev:
This issue is hard to debug, so here is an attempt to fix it + Exception when the issue happens to make sure wrong data will not end up in the production build.

Fix https://github.com/defold/defold/issues/10356
Fix https://github.com/defold/defold/issues/10196

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ArchiveBuilder.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ArchiveBuilder.java
@@ -30,7 +30,7 @@ import java.util.List;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.Map;
-import java.util.HashMap;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


britzl (approved):

selimanac:
If it helps, I can share my project. This happens every time, and I can’t use Live Update at all:
https://forum.defold.com/t/live-update-bundle-error-has-no-hex-digest/80262

AGulev:
@selimanac yes, please. And open a separate issue so it will not lose

PR #10396, opened 2025-03-27
Throw an exception if the quaternion in GO or Collection is invalid

AGulev:
This isn't an issue we have, since the editor itself can't produce invalid values.  
But if the user adds an invalid value manually, the editor and runtime handle it fine, while Bob and the runtime do not.  
So I decided to validate the value on Bob's side to avoid inconsistencies.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ModelUtilTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ModelUtilTest.java
@@ -124,7 +124,7 @@ public class ModelUtilTest {
      */
     private void assertBone(Rig.Bone bone, Vector3d expectedPosition, Quat4d expectedRotation) {
         assertV(expectedPosition, MathUtil.ddfToVecmath(bone.getLocal().getTranslation()));
-        assertV(expectedRotation, MathUtil.ddfToVecmath(bone.getLocal().getRotation(), "bone " + bone.getName()));
+        assertV(expectedRotation, MathUtil.ddfToVecmath(bone.getLocal().getRotation(), "bone %s".formatted(bone.getName())));
     }
 
     /*
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ModelUtilTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ModelUtilTest.java
@@ -124,7 +124,7 @@ public class ModelUtilTest {
      */
     private void assertBone(Rig.Bone bone, Vector3d expectedPosition, Quat4d expectedRotation) {
         assertV(expectedPosition, MathUtil.ddfToVecmath(bone.getLocal().getTranslation()));
-        assertV(expectedRotation, MathUtil.ddfToVecmath(bone.getLocal().getRotation()));
+        assertV(expectedRotation, MathUtil.ddfToVecmath(bone.getLocal().getRotation(), "bone " + bone.getName()));
     }
 
     /*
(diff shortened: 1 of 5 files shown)


britzl:
> But if the user adds an invalid value manually, the editor and runtime handle it fine

What does the editor do to handle it?

AGulev:
> What does the editor do to handle it?

The editor works with Euler. So it converts this invalid quaternion into a valid euler and works with it

britzl:
> But if the user adds an invalid value manually, the editor and runtime handle it fine, while Bob and the runtime do not.
> So I decided to validate the value on Bob's side to avoid inconsistencies

I am a bit concerned that we set a dangerous precedence of sanity checking values. I'm sure there are a lot of other places where a user could enter weird/invalid values.

> The editor works with Euler. So it converts this invalid quaternion into a valid euler and works with it

Could we do the same and simply treat it as "no rotation"?

AGulev:
> Q: What happens at runtime? Does the engine crash?

Undefined behavior.
The editor keep w:0 and the engine works just fine with it.

But bob as result of merging makes
x:nan
y:nan
z:nan
w:nan
which is undefined behavior.

AGulev:
> Could we do the same and simply treat it as "no rotation"?

if one of components "nan" use 
 x:0
y:0
z:0
w:1

instead?

AGulev:
@britzl do you think we should just silently fix if and use default quaternion if one or elements is nan?

JCash:
Perhaps we can do this in `MathUtil.vecmathToDDF()` and return (0,0,01) it was invalid, and log an error.
This would help discoery for _all_ broken quats.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/MathUtil.java
```diff
@@ -46,7 +48,28 @@ public static Vector3d ddfToVecmath(Vector3One v) {
...
+        if (Math.abs(lenSq - 1.0f) > 1e-4f) {
+            return false;
+        }
+        return true;
+    }
+
+    public static Quat4d ddfToVecmath(Quat q, String owner) {
+        if (!isValid(q)) {
+            String err = new StringBuilder().append("Invalid quaternion: ").append(q).append(" in ").append(owner).toString();
+            throw new RuntimeException(new CompileExceptionError(err));
```
AGulev, line 71: `CompileExceptionError` wrapped in `RuntimeException` works good in both the editor and in bob

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ModelUtilTest.java
```diff
@@ -124,7 +124,7 @@ private void assertVertBone(Tuple4i expected, List<Integer> actual) {
      */
     private void assertBone(Rig.Bone bone, Vector3d expectedPosition, Quat4d expectedRotation) {
         assertV(expectedPosition, MathUtil.ddfToVecmath(bone.getLocal().getTranslation()));
-        assertV(expectedRotation, MathUtil.ddfToVecmath(bone.getLocal().getRotation()));
+        assertV(expectedRotation, MathUtil.ddfToVecmath(bone.getLocal().getRotation(), "bone " + bone.getName()));
```
AGulev, line 127: this is a test, so I don't use string builder here, it is not necessary

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/CollectionBuilder.java
```diff
@@ -209,7 +209,7 @@ private void mergeSubCollections(IResource owner, CollectionDesc.Builder collect
                 properties.put(pathPrefix + instProps.getId(), instProps.getPropertiesList());
             }
             Point3d p = MathUtil.ddfToVecmath(collInst.getPosition());
-            Quat4d r = MathUtil.ddfToVecmath(collInst.getRotation());
+            Quat4d r = MathUtil.ddfToVecmath(collInst.getRotation(), new StringBuilder().append(owner).append(" collection: ").append(collInst.getId()).toString());
```
AGulev, line 212: the string builder is a bit more efficient here

JCash (changes requested):
a small redundant check in the validate function

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/MathUtil.java
```diff
@@ -46,7 +48,28 @@ public static Vector3d ddfToVecmath(Vector3One v) {
         return new Vector3d(v.getX(), v.getY(), v.getZ());
     }
 
-    public static Quat4d ddfToVecmath(Quat q) {
+    public static boolean isValid(Quat q) {
+        if (q.getX() == 0  && q.getY() == 0 && q.getZ() == 0 && q.getW() == 0) {
+            return false;
+        }
```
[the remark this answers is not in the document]
AGulev: Good one!

PR #10398, opened 2025-03-27
Hotfix: syntax info

vlaaad:
This change fixes an issue where typing in a file without grammar would throw exceptions.


matgis (approved):
Ship it!

PR #10405, opened 2025-03-29
Editor: Replace all calls to workspace/project-path with the new project-directory function

matgis:
### Technical changes
* Removed `workspace/project-path`, which took an `evaluation-context`.
* Replaced all calls to `workspace/project-path` with `workspace/project-directory`, which takes a `basis`.
* Replaced some instances where we were accessing the `:root` property of the `workspace` with calls to `workspace/project-directory`.


vlaaad (approved):
Good stuff!

PR #10414, opened 2025-03-31
Remove duplicate setting

vlaaad:


editor/resources/meta.edn
```diff
@@ -508,10 +508,6 @@
    :help "max number of particle fx emitters, 64 by default",
    :default 64,
    :path ["particle_fx" "max_emitter_count"]}
-  {:type :integer,
-   :help "max total number of living particles, 1024 by default",
-   :default 1024,
-   :path ["particle_fx" "max_particle_count"]}
```
vlaaad, line 514: Remove this one since the other has better help string (`"max total number of living particles. limits the GPU vertex buffer size. 1024 by default"`)

JCash (approved):

PR #10416, opened 2025-03-31
Fix script property label drag

sprocketc:
Fixes dragging labels of numeric script properties.

Fixes #10411


editor/src/clj/editor/properties_view.clj
```diff
@@ -223,12 +224,15 @@
...
+        {:keys [key node-ids edit-type]} property
         {:keys [precision from-type to-type]} edit-type
-        [x y] [(.getX event) (.getY event)]
+        x (.getX event)
+        y (.getY event)
         [prev-x prev-y] (ui/user-data target ::position)
         delta-x (- x prev-x)
         delta-y (- prev-y y)
+        script-property? (partial g/node-instance? script/ScriptPropertyNode)
+        key-fn (fn [node-id] (if (script-property? node-id) :value key))
```
sprocketc, line 235: Script properties are represented by dedicated graph nodes as far as I understand. This should fix the issue, but it looks like `properties/set-values!` abstracts this away. Not sure if its worth it changing the implementation to use that instead.

vlaaad: Hmm... I think we should use `properties/set-values!` or implement some function there specifically for drag operations. There is a set of rules for changing properties, and the rules should preferably be located in some shared place (like `properties` ns) to separate concerns. I don't quite remember this set of rules, but it seems that these property maps may include `:prop-kw` keyword to denote the right key to use for setting the value. The logic should probably check for that instead of checking for the node instance: script properties define the `prop-kw` here: https://github.com/defold/defold/blob/0272aeb58dd014dbe30a0adcc0dfecc52e00bfb7/editor/src/clj/editor/code/script.clj#L187

vlaaad (commented):
I'm not 100% sure, but I think we might want to use `properties/set-values!` or implement a separate function in the `properties` ns so that we don't duplicate the logic in the view ns.

editor/src/clj/editor/properties_view.clj
```diff
@@ -235,21 +236,25 @@
...
+                :let [current-value (cond-> (g/node-value node-id prop-kw) to-type to-type)
+                      new-value (cond-> (drag-update-fn current-value update-val)
+                                  from-type from-type
+                                  (:min edit-type) (max (:min edit-type))
+                                  (:max edit-type) (min (:max edit-type)))]]
+            (concat (g/operation-sequence op-seq)
+                    (g/set-property node-id prop-kw new-value))))
+        (ui/user-data! target ::position [x y])
+        (when (apply = values)
+          (let [[node-id prop-kw] (first set-operations)]
```
vlaaad, line 253: Hmm, now that I think of it, this code might be unsafe since prop-kw is not guaranteed to exist. Here is an example of such a property definition without `:prop-kw`: https://github.com/defold/defold/blob/690c5d630d7dd3e5b0da935a6f5c1bc941fef17d/editor/src/clj/editor/sprite.clj#L430-L437

There is another place above that uses `g/node-value` on `prop-kw`: https://github.com/defold/defold/blob/8855500c663f4c891fedbe122d0ecf40dbf78ac1/editor/src/clj/editor/properties_view.clj#L244

vlaaad (changes requested):
Hmm, I think I found some more issues with the property handling. We might define `set-fn` on the `edit-type` of a property, and in this case, the `prop-kw` will be absent. This would lead to errors in the current implementation.

I found a place where we create numeric attributes without `prop-kw` in the code, so I think it's possible to trigger the exception here still by creating a special material: https://github.com/defold/defold/blob/4e8483c3143df92f66ae6086d4d2cfb0e843c8a7/editor/src/clj/editor/graphics.clj#L908-L913

sprocketc:
I ended up calling `properties/set-values!` by collecting the vector using `set-operations`, so we also avoid using `:from-type`.
> Hmm, I think I found some more issues with the property handling. We might define set-fn on the edit-type of a property, and in this case, the prop-kw will be absent. This would lead to errors in the current implementation.

I couldn't find a way to test this, but I assume we need to use something like `(or prop-kw :value)`?

vlaaad:
We use `set-fn` without `prop-kw` for cases where we want to present something as a property in a properties view when the graph node does not have a corresponding property to get/set (so we can't use `g/node-value`). It's probably a value inside some map somewhere on another node property.

I made a repro case where this issue can be triggered:
[properties-drag-repro.zip](https://github.com/user-attachments/files/19562534/properties-drag-repro.zip)
To reproduce the error, open `main.collection`, select `sprite` component of a root `go` game object, and try dragging the X/Y/Z/W label of a property named `"New Attribute"` (at the bottom of the properties list):

https://github.com/user-attachments/assets/494faffc-49d7-4b7b-b261-4d5cdfe41187

sprocketc:
@vlaaad Thanks for the repro case! I should have seen it sooner, but we can avoid all of this by simply caching the values while dragging.

editor/src/clj/editor/properties_view.clj
```diff
@@ -223,31 +223,33 @@
...
+            values (or (ui/user-data target ::values)
+                       (properties/values property))
+            new-values (mapv (fn [value]
+                               (cond-> (drag-update-fn value update-val)
+                                 min-val (max min-val)
+                                 max-val (min max-val))) values)]
+        (properties/set-values! property values op-seq)
+        (ui/user-data! target ::position [x y])
+        (ui/user-data! target ::values new-values)
+        (update-ui-fn (mapv (comp to-fn from-fn) new-values)
```
sprocketc, line 252: This is not required. We can use new-values directly, but the value will be changed on drag end in some cases (e.g. when a rotation value exceeds 180). I am tempted to do that, to get rid of `to-fn` and `from-fn` and only rely on methods that abstract all of this.

vlaaad (approved):
Works great, and the code is great! Ship it!

PR #10418, opened 2025-04-01
Sparse node property assignments

matgis:
Instead of assigning values for all properties in the node maps, we now look up the default values in a defaults map associated with the node type.

### Technical changes
* Renamed the `own-properties` method of the `Node` protocol to `assigned-properties` to reflect this change.
* Added `g/own-property-values` function that returns the equivalent result.
* Updated the list of opened Java modules in the `:jamm` Lein profile to make `mem/node-size-report` work again after recent changes.
* Improved `gu/node-debug-label` to distinguish between resource nodes and their overrides.
* You can no longer treat a `NodeImpl` instance like a Clojure map, because it won't contain values for defaults. This was arguably not the case before either, because it has the additional `:node-type` key, and it was never the case for `OverrideNode`. However, some older tests were using `keys` to test that the `NodeImpl` had all values assigned after construction.

### DIFF
--- a/editor/project.clj
+++ b/editor/project.clj
@@ -219,9 +219,12 @@
                                            (.replace (System/getProperty "user.home") \\ \/)
                                            "/.m2/repository/com/github/jbellis/jamm/0.4.0/jamm-0.4.0.jar")
                                         "-Ddefold.jamm=true"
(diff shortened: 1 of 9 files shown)


vlaaad (approved):
Ship it!

PR #10422, opened 2025-04-02
Consume the mouse event when clicking outside of the color popup

### LINKED ISSUE #9284: Click outside a color picker widget also triggers the editor view node selection
When color picker palette is active and I click on editor view to close it, this click is also used as a node selection on editor view

Don't register editor view click if any of popup from properties panel is selected

Always click on properties panel or use "esc" to close the properties popup widget

https://github.com/user-attachments/assets/836d50cf-6294-4f2e-ae08-a47c33321aeb

sprocketc:
Clicking outside of a color picker popup should not propagate.

Fixes #9284

### Technical changes
I couldn't find a better way to filter the mouse event and consume it. The introduced overlay is not scoped to the color-picker property multimethod (although it could), because this doesn't look like a color picker specific issue. `ChoiceBox` seems to have the same problem (e.g. the layout select on the toolbar). If we are OK with this approach, we can use the same overlay to block the event propagation in similar cases.


vlaaad (approved):
Cool solution! 

I think there might be a way to implement event consumption using ColorPicker only, but it's much more hacky than this: we would need to wait until the color picker skin is created (or force its creation by assigning a stylesheet to it... hacky), then access the skin, then reflectively (hacky!) access the skin-private popup, and finally set it's `consumeAutoHidingEvents` property to true... I think your solution is better!

sprocketc:
@Vlaad I tried that at first (it took me a lot longer to figure out what to do..), but it it didn't work. I guess I might have done something wrong after multiple failed attempts.

PR #10425, opened 2025-04-02
sound mixer fix for negative offsets on 64-bit platforms

thengel01:
Fixes an issue that calculate pointer offset incorrectly for 64-bit pointers depending on how precisely the compiler generates code

### DIFF
--- a/engine/sound/src/sound.cpp
+++ b/engine/sound/src/sound.cpp
@@ -1080,7 +1080,7 @@ namespace dmSound
         instance->m_FrameCount = SOUND_MAX_HISTORY + (avail_framecount - used_framecount) + SOUND_MAX_FUTURE;
         assert(instance->m_FrameCount <= SOUND_INSTANCE_STATEFRAMECOUNT);
         uint32_t state_bytes = instance->m_FrameCount * sizeof(float);
-        uint32_t state_offset = used_framecount - SOUND_MAX_HISTORY;
+        int32_t state_offset = (int32_t)(used_framecount - SOUND_MAX_HISTORY);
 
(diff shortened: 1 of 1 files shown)


engine/sound/src/sound.cpp
```diff
@@ -1080,7 +1080,7 @@ namespace dmSound
         instance->m_FrameCount = SOUND_MAX_HISTORY + (avail_framecount - used_framecount) + SOUND_MAX_FUTURE;
         assert(instance->m_FrameCount <= SOUND_INSTANCE_STATEFRAMECOUNT);
         uint32_t state_bytes = instance->m_FrameCount * sizeof(float);
-        uint32_t state_offset = used_framecount - SOUND_MAX_HISTORY;
+        int32_t state_offset = (int32_t)(used_framecount - SOUND_MAX_HISTORY);
```
JCash, line 1083: What do you mean 64bit platforms?
I don't see anything related to bitness here?

Also, allowing "state_offset" to be negative surely has an effect on the calculations below?

I think you need to explain the scenarios here. what if the `state_offset == -4` now?

thengel01: - 64 bit: anything using 64-bit pointers (there is an either signed or unsigned extension from 32 to 64 bit in the code below. Depends on the type here. It was incorrectly unsigned.
- state offset can "legally" be negative if very few / no samples have been mixed for some reason. This is ok, as the bsae pointer operated here is actually sufficiently moved post the beginning of the allocated memory. For ease of processing all mixing code refers to the offset base, while "historic" data is available imn front of it. Hence negatie values are "ok" - and should better be extended in a signed fashion

Issue #6901, opened 2022-08-22
Add an ability to have shortcuts for the editor scripts actions


AGulev:
**Is your feature request related to a problem? Please describe (REQUIRED):**
Sometimes it's useful to have a fast way to run an action from editor_scripts

**Describe the solution you'd like (REQUIRED):**
A way to specify a shortcut for an action

**Describe alternatives you've considered (REQUIRED):**
Clicking to actions with mouse which is annoying for actions used frequent

PR #10427, opened 2025-04-03
Actualize prefs dialog

vlaaad:
Now, it uses our new look and feel:
Screenshot 2025-04-03 at 10 13 08

Related to #6901


editor/src/clj/editor/analytics.clj
```diff
@@ -15,18 +15,15 @@
...
-           (java.nio.charset StandardCharsets)
-           (java.util UUID)
-           (java.util.concurrent CancellationException)))
+  (:import [clojure.lang PersistentQueue]
+           [com.defold.editor Editor]
+           [java.io File]
+           [java.net HttpURLConnection MalformedURLException URL]
+           [java.nio.charset StandardCharsets]
+           [java.util UUID]
+           [java.util.concurrent CancellationException]))
```
vlaaad, line 26: Unrelated imports cleanup

editor/src/clj/editor/app_view.clj
```diff
@@ -869,13 +869,13 @@
           {:title "Launch Failed"
            :icon :icon/triangle-error
            :header {:fx/type fx.v-box/lifecycle
-                    :children [{:fx/type fxui/label
+                    :children [{:fx/type fxui/legacy-label
```
vlaaad, line 872: There was a label component, but now there is another, which is also used in editor scripts. I had to rename the old one...

editor/src/clj/editor/code/view.clj
```diff
@@ -3649,7 +3649,7 @@
                [{:fx/type fx.label/lifecycle
                  :style-class ["label" "breakpoint-editor-label"]
                  :text "Condition"}
-                {:fx/type fxui/text-field
+                {:fx/type fxui/legacy-text-field
```
vlaaad, line 3652: Same with text field

editor/src/clj/editor/editor_extensions/ui_components.clj
```diff
@@ -218,50 +215,40 @@
...
+                                      (let [{:keys [row_span column_span]
+                                             :or {row_span 1 column_span 1}} (:props (meta child))]
+                                        (assoc child :grid-pane/row row
+                                                     :grid-pane/column column
+                                                     :grid-pane/row-span row_span
+                                                     :grid-pane/column-span column_span)))))
+                                row-children)))
+                          children))
+    rows (apply-constraints :row-constraints fx.row-constraints/lifecycle :vgrow rows)
+    columns (apply-constraints :column-constraints fx.column-constraints/lifecycle :hgrow columns)
```
vlaaad, line 249: There was actually a bug where using the `rows` prop would instead use the `:column-contraints` key, but with `row-constraint` values, which would cause the runtime exceptions.

editor/src/clj/editor/editor_extensions/ui_components.clj
```diff
@@ -218,50 +215,40 @@
...
-                                                         :grid-pane/column-span column_span)))))
-                                    row-children)))
-                              children))
-        rows (apply-constraints fx.row-constraints/lifecycle :vgrow rows)
-        columns (apply-constraints fx.column-constraints/lifecycle :hgrow columns)
-        padding (apply-padding padding)
-        alignment (apply-alignment alignment))))
+  (cond->
+    {:fx/type fxui/grid
+     :spacing spacing}
```
vlaaad, line 231: I started working on unifying the views between editor scripts and clojure editor code, one bit at a time...

editor/src/clj/editor/fxui.clj
```diff
@@ -636,3 +663,274 @@
...
+  "Text field with value commit/reset semantics
+
+  Supports all :text-field props, plus:
+    :value               the edited value
+    :on-value-changed    value change callback
+    :to-string           value->string converter, default str
+    :to-value            string->value converter, default identity, returning
+                         nil implies value could not be converted"
+  [props]
+  (make-value-field text-field true props))
```
vlaaad, line 911: `value-field` is a copy of the editor script implementation. I thought it would be possible to avoid code duplication here and reuse this value-view in the editor scripts, but unfortunately, there are some differences in the implementations that would take more time to resolve. The editor scripts' implementation uses different handling of focus change because it also needs to access the dialog window when the focus changes. It passes it to its value change handler so that in case the lua value change listener decides to open a new dialog, the current one will be set as its owner, transparently for the user. We probably want this here as well... eventually... But I thought this would be too out of scope for the keyboard shortcuts feature.

editor/src/clj/editor/fxui.clj
```diff
@@ -636,3 +663,274 @@
...
+      (add-style-classes "text-input" "text-area" "ext-text-area")
+      resolve-input-color
+      resolve-tooltip))
+
+(defn- handle-value-field-key-pressed [edit text swap-state on-value-changed to-value commit-on-enter ^KeyEvent e]
+  (condp = (.getCode e)
+    KeyCode/ENTER
+    (when (and (or commit-on-enter (.isShortcutDown e))
+               (not= edit text))
+      (.consume e)
```
matgis, line 828: Should we really consume the event here? We also consume it in the invalid-value branch below.

editor/src/clj/editor/prefs.clj
```diff
@@ -75,19 +78,20 @@
    :properties
    {:asset-browser {:type :object
                     :properties
-                    {:track-active-tab {:type :boolean :label "Track Active Tab in Asset Browser"}}}
+                    {:track-active-tab {:type :boolean
+                                        :ui {:label "Track Active Tab in Asset Browser"}}}}
```
matgis, line 82: I kind of feel like the `:ui` clause could be warranted to keep on its own line like this everywhere.

matgis (dismissed):
Nice! Love to see us starting to use the new components in the fxui views!

editor/src/clj/editor/fxui.clj
```diff
@@ -392,6 +401,24 @@
                    :middleware (fx/wrap-map-desc merge description))]
     (mount-renderer-and-await-result! state-atom renderer)))
 
+(defn show-stateless-dialog-and-await-result!
```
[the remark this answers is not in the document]
vlaaad: It is named stateless in contrast to `show-dialog-and-await-result!`, which explicitly creates a state atom for the dialog and watches it to determine when it is done (checks if a key was set). This stateless approach is simpler: it passes back a completion function that the view invokes with a result to notify that it's done.

editor/src/clj/editor/fxui.clj
vlaaad: Nice catch! I forgot to remove the consume call from the invalid value branch! It should be in both branches.

PR #10430, opened 2025-04-03
Faster LU archive creation + Added option for compression level.  

### LINKED ISSUE #10012: Find faster way to create ZIP archive in Bob for LU
**Task (REQUIRED):**  
Bob has two separate builders:  
- Our own archive format builder (`arcd`, `arci`, etc.)  
- A ZIP archive builder used for Live Update.  

Testing the same project with a large number of resources, we found the following:  
- Our archive builder takes just **25-32 seconds** to create an archive with approximately 330,000 files.  
- The ZIP archive builder for Live Update, however, takes about **850 seconds**, which is **26 times slower**.  

We need to identify a way to speed up the ZIP archive creation process in Java, making it as close as possible in performance to our internal format builder.  

**Expected Outcome (REQUIRED):**  
The time to create a Live Update archive should be significantly reduced. Ideally, it should not take much more time than creating an `arcd` archive.  
- If the ZIP archive creation process is **2-3 times slower** than the `arcd` archive creation process, this would still be a significant improvement over the current performance.

AGulev:
A new option was added to the LiveUpdate settings: ZIP archive compression level, with levels ranging from 0 to 9.  
The default level is 1 (the same as before the changes).
This option is useful in cases where compression is not needed at all—e.g., when Poki repacks the archive and applies their own compression.  In such cases, using level 0 will significantly speed up archive creation and the bundling process.  
Additionally, the archive writer itself was improved, making archive creation at least twice as fast (regardless of whether compression is used or not).

On a test suite with 172,296 files, which were used for benchmarking on a Mac M1 Pro Max:  

- Archive creation before changes (level 1 compression): **56 seconds**  
- With performance improvements and default level 1 compression: **28 seconds**  
- With performance improvements and level 0 compression: **1.8 seconds**  

Fix https://github.com/defold/defold/issues/10012

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/archive/test/ZipPublisherTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/archive/test/ZipPublisherTest.java
@@ -1,5 +1,6 @@
 package com.dynamo.bob.archive.test;
 
+import com.dynamo.bob.CompileExceptionError;
 import com.dynamo.bob.archive.ArchiveEntry;
 import com.dynamo.bob.archive.publisher.PublisherSettings;
 import com.dynamo.bob.archive.publisher.ZipPublisher;
@@ -57,9 +58,9 @@ public class ZipPublisherTest {
             // Step 2: Publish all ArchiveEntry objects
             long startPublishTime = System.currentTimeMillis();
             for (Map.Entry<String, AbstractMap.SimpleEntry<ArchiveEntry, byte[]>> entry : archiveMap.entrySet()) {
-                try (ByteArrayInputStream bais = new ByteArrayInputStream(entry.getValue().getValue())) {
-                    zipPublisher.publish(entry.getValue().getKey(), bais);
-                } catch (IOException e) {
+                try {
+                    zipPublisher.publish(entry.getValue().getKey(), entry.getValue().getValue());
+                } catch (CompileExceptionError e) {
                     fail("Error publishing entry: " + entry.getKey());
                 }
             }
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/archive/test/ZipPublisherTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/archive/test/ZipPublisherTest.java
@@ -1,0 +1,63 @@
+package com.dynamo.bob.archive.test;
+
+import com.dynamo.bob.archive.ArchiveEntry;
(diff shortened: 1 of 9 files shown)


com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/archive/test/ZipPublisherTest.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ForkJoinPool;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class ZipPublisherTest {
+// This isn't a test really, this is setup for development to be able to iterate faster on the archiving
+// see https://github.com/defold/defold/issues/10012
+//    @Test
```
AGulev, line 29: I’m keeping it for the future in case we need to improve it even more.  
It’s much easier to have it as a test and use it from IntelliJ IDEA (no need to rebuild Bob all the time).

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/publisher/ZipPublisher.java
```diff
@@ -82,10 +78,11 @@ public void start() throws CompileExceptionError {
                 this.destZipFile = new File(cwd, this.destZipFile.getPath());
             }
 
-            zipEntries.clear();
+            entries.clear();
```
AGulev, line 81: we use only entries now instead of having both `entries` and `zipEntries`

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/publisher/ZipPublisher.java
```diff
@@ -118,19 +115,15 @@ public void publish(ArchiveEntry entry, InputStream data) throws CompileExceptio
         final String archiveEntryHexdigest = entry.getHexDigest();
         final String archiveEntryName = entry.getName();
         final String zipEntryName = (archiveEntryHexdigest != null) ? archiveEntryHexdigest : archiveEntryName;
-        synchronized (zipEntries) {
-            if (zipEntries.contains(zipEntryName)) {
-                return;
-            }
-            zipEntries.add(zipEntryName);
+        if (entries.put(archiveEntryName, entry) != null) {
```
AGulev, line 118: Simpler logic when using `ConcurrentHashMap`
also, no need to check contains and add it separatly

britzl (approved):

PR #10433, opened 2025-04-03
Introduce global console filtering toggle 

### LINKED ISSUE #10400: Editor Console All Filters Toggler
When I'm using a several filters in Defold Editor, sometimes I want to remove all filters to see full output. But in this case I should turn off manually each filter and then return back each of them.

Can we have a global filter checkbox? Seems are very convenient to me

sprocketc:
A global filter toggle button was introduced to the console tab.

Screenshot From 2025-04-04 18-10-20

Fixes #10400


editor/src/clj/editor/console.clj
```diff
@@ -311,18 +315,33 @@
...
 (defonce ^SimpleStringProperty find-term-property (doto (SimpleStringProperty.) (.setValue "")))
 
+(def ^:private filter-svg-path
+  (ui/load-svg-path "scene/images/filter.svg"))
+
+(defn- toggle-filtering
+  [^ToggleButton toggle-button prefs]
+  (let [is-selected (.isSelected toggle-button)]
+    (prefs/set! prefs console-filtering-key is-selected)
+    (set-filters! (if is-selected (prefs/get prefs console-filters-prefs-key) []))))
```
sprocketc, line 325: I initially used a handler, but decided to use a simple toggle action for now.

editor/resources/console-view.fxml
```diff
@@ -13,7 +13,9 @@
     <RowConstraints vgrow="NEVER" />
   </rowConstraints>
   <HBox id="console-tool-bar" GridPane.rowIndex="0">
-    <Button id="filter-console" text="Filter" />
+    <Button id="filter-console" text="Filters" />
```
sprocketc, line 16: Now that we have a toggle button next to the dropdown, I think it makes sense to use a plural noun for the multiselect filter list.

editor/styling/stylesheets/modules/_console.scss
```diff
@@ -163,6 +160,30 @@
...
+    }
+  }
+
+  &:selected {
+    -fx-background-color: -df-component-light;
+
+    SVGPath {
+      -fx-fill: -df-text-selected;
+    }
+  }
```
sprocketc, line 184: Styling this similarly to the toolbar toggles looks weird in this context.

editor/src/clj/editor/console.clj
```diff
@@ -311,18 +315,33 @@
 
 (defonce ^SimpleStringProperty find-term-property (doto (SimpleStringProperty.) (.setValue "")))
 
+(def ^:private filter-svg-path
+  (ui/load-svg-path "scene/images/filter.svg"))
```
sprocketc, line 319: `scene/images/`is the only dir that hosts SVG icons for now. It' s probably not a great fit for this case, but I don't know if there is a better option, that does not require broader restructuring.

sprocketc:
A select / deselect all button could also be added to the multiselect dropdown, but that wouldn't exactly solve the feature request. I can add this here, or in a future PR, if we think this is useful.

matgis:
Would it be better to put the global filter toggle inside the Filters popup next to the individual toggles?

@vlaaad You did the initial implementation of this, what do you think?

vlaaad:
Hmm.. In my opinion, the filtering toggle should be inside the Filter button popup, and when it's off, it should not show a number of active filters counter. Otherwise it seems strange:
Screenshot 2025-04-04 at 13 38 34
I'm used to thinking that the counter is an indicator for a number of active filters. But here it's off, so the indicator of whether we are filtering or not is both the funnel icon and the counter. I also think that the funnel icon in its active state draws too much attention. Let's move it into the popup!

sprocketc:
@vlaaad @matgis I forgot to reply here, but this should be ready for review now.

editor/src/clj/editor/console.clj
```diff
@@ -265,7 +273,18 @@
...
                             :children
-                            [{:fx/type fx.list-view/lifecycle
+                            [{:fx/type fx.check-box/lifecycle
+                              :focus-traversable false
+                              :max-width ##Inf
+                              :v-box/margin 4
+                              :id "global-console-filtering"
+                              :selected (prefs/get prefs console-filtering-key)
+                              :on-action {:event-type :toggle-global-filtering}
+                              :on-selected-changed (partial set-global-filtering prefs)
```
vlaaad, line 283: I would use `:on-selected-changed {:event-type :toggle-global-filtering}` instead, and I'd updated the prefs in the `handle-filter-event!` function like we do for filter changes. It would be more consistent that way.

editor/src/clj/editor/console.clj
```diff
@@ -298,16 +318,17 @@
...
+    (set-filters! (if filtering filters []))
     (fx/mount-renderer
       state
       (fx/create-renderer
         :error-handler error-reporting/report-exception!
         :opts {:fx.opt/map-event-handler #(handle-filter-event! state prefs %)}
         :middleware (comp
                       fxui/wrap-dedupe-desc
-                      (fx/wrap-map-desc #(filter-console-view filter-console-button %)))))))
+                      (fx/wrap-map-desc #(filter-console-view filter-console-button prefs %)))))))
```
vlaaad, line 331: If we update the `filtering` preference in `handle-filter-event!` like we currently do for `filters`, then passing the prefs to view will be unnecessary.

editor/src/clj/editor/console.clj
```diff
@@ -319,7 +340,7 @@
                               ^Button clear-console
                               filter-console]
     (init-console-filter! filter-console prefs)
-    (ui/context! tool-bar :console-tool-bar {:term-field search-console :view-node view-node} nil)
+    (ui/context! tool-bar :console-tool-bar {:term-field search-console :view-node view-node :prefs prefs} nil)
```
vlaaad, line 343: Is it needed?

vlaaad (dismissed):
Works great! 

Nitpick: we could improve the consistency of updating the prefs in the popup. Currently, `filters` are updated in the event handler, and `filtering` is updated in the view itself.

vlaaad (approved):
Shipit!

PR #10434, opened 2025-04-03
Remove some unrealized sequences from the graph and elsewhere

matgis:
We generally want to avoid storing references to lazy sequences and eductions, since these may hold on to various large objects and prevent them from being garbage collected.

### Technical changes
* Converted various eductions and lazy sequences into vectors.
* Fixed `PaperTape` equivalence check throwing when compared to a non-`PaperTape` object.
* Added `eduction/eduction?` function.
* Added `eduction/mapcat-indexed` function.


vlaaad (approved):
LGTM!

PR #10439, opened 2025-04-03
Added `--keep-bob-uncompressed` for bob to speedup development

AGulev:
New flag for bob development that speed up bob building twice from the terminal:
> ./scripts/build.py build_bob --skip-tests --keep-bob-uncompressed

and 10 times using gradle daemon (e.g. IDEA):
CleanShot 2025-04-09 at 15 37 36@2x

### DIFF
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -410,6 +410,7 @@ class Configuration(object):
                  dynamo_home = None,
                  target_platform = None,
                  skip_tests = False,
+                 uncompressed = False,
                  skip_codesign = False,
                  skip_docs = False,
                  incremental = False,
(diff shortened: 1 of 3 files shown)


scripts/build.py
```diff
@@ -410,6 +410,7 @@ def __init__(self,
                  dynamo_home = None,
                  target_platform = None,
                  skip_tests = False,
+                 uncompressed = False,
```
JCash, line 413: Given the "global" context here, I think it should be bob_use_compression (or something with bob at least)

JCash (dismissed):
I think it should be named something regarding bob.

scripts/build.py
AGulev: Renamed to `keep-bob-uncompressed`:
> ./scripts/build.py build_bob --skip-tests --keep-bob-uncompressed

JCash (approved):

PR #10454, opened 2025-04-07
Throw a Lua error if `nil` passed into `go.delete()` function

### LINKED ISSUE #10428: Add script and line to "go.delete() invoked with nil and self will be deleted"
If I get the warning `WARNING:GAMEOBJECT: go.delete() invoked with nil and self will be deleted` it would be really helpful to know _where_ this occurred. I use a manager script which creates and deletes lots of objects, so if this happens unexpectedly it is my manager script which gets deleted with catastrophic consequences. Not knowing where this happened makes it difficult to debug.

AGulev:
⚠️ Passing `nil` to `go.delete()` results in a Lua error, as this aligns with expected behavior and helps catch logical issues in the code.

Fix https://github.com/defold/defold/issues/10428

### DIFF
--- a/engine/gameobject/src/gameobject/gameobject_script.cpp
+++ b/engine/gameobject/src/gameobject/gameobject_script.cpp
@@ -1935,7 +1935,7 @@ namespace dmGameObject
             }
             else if(lua_isnil(L, 1))
             {
-                return luaL_error(L, "The url shouldn't be `nil`");
+                return luaL_error(L, "go.delete() invoked with first argument 'id' set to 'nil'");
             }
         }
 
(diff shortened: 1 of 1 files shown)


britzl (changes requested):
I think we should be more explicit in the error message and still mention `go.delete()`. Some suggestions:

"go.delete() invoked with id nil"
"go.delete() invoked with id set to nil"
"go.delete() invoked with first argument 'id' set to nil"
"go.delete() invoked with first argument 'id' set to nil by accident"

britzl (approved):

PR #10461, opened 2025-04-08
Create GUI dependencies on drop 

sprocketc:
Allow dropping assets to GUI scenes to create dependencies.

gui-drop

### Technical changes
* Refactor some of the code added on #10298.
* Use string paths instead of files on dnd.
* Pass a `drop-fn` function to the `SelectionController` and call the existing `select-fn` when new node ids are returned. I think it makes more sense to handle dropping on the selection controller.
* Use the introduced `drop-fn` property to allow adding GUI dependencies on dnd from the asset tree.

### DIFF
--- a/editor/src/clj/editor/asset_browser.clj
+++ b/editor/src/clj/editor/asset_browser.clj
@@ -691,6 +691,9 @@
 (defn- drag-detected [^MouseEvent e selection]
   (let [resources (roots selection)
         files (fileify-resources! resources)
+        paths (->> resources
+                   (mapv resource/proj-path)
+                   (string/join "\n"))
         ;; Note: It would seem we should use the TransferMode/COPY_OR_MOVE mode
         ;; here in order to support making copies of non-readonly files, but
         ;; that results in every drag operation becoming a copy on macOS due to
@@ -707,6 +710,7 @@
       (.setDragView db (icons/get-image (workspace/resource-icon (first resources)) 16)
                     0 16))
     (.putFiles content files)
+    (.putString content paths)
     (.setContent db content)
     (.consume e)))
 
--- a/editor/src/clj/editor/gui.clj
+++ b/editor/src/clj/editor/gui.clj
@@ -3266,11 +3266,11 @@
   (input script-resource resource/Resource)
 
   (input node-tree g/NodeID)
-  (input layers-node g/NodeID) ; for tests
-  (input layouts-node g/NodeID) ; for tests
-  (input fonts-node g/NodeID) ; for tests
-  (input textures-node g/NodeID) ; for tests
-  (input particlefx-resources-node g/NodeID) ; for tests
+  (input layers-node g/NodeID)
+  (input layouts-node g/NodeID)
+  (input fonts-node g/NodeID)
+  (input textures-node g/NodeID)
+  (input particlefx-resources-node g/NodeID)
   (input materials-node g/NodeID)
   (input handler-infos g/Any :array)
   (output handler-infos g/Any (g/fnk [handler-infos]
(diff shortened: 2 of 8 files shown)


editor/src/clj/editor/asset_browser.clj
```diff
@@ -691,6 +691,9 @@
 (defn- drag-detected [^MouseEvent e selection]
   (let [resources (roots selection)
         files (fileify-resources! resources)
+        paths (->> resources
+                   (mapv resource/proj-path)
+                   (string/join "\n"))
```
sprocketc, line 696: By adding plain paths to content we avoid unnecessary back and forth (see removed `get-image-resources-from-files`). We also allow dropping files from builtin assets that wasn't previously working because `fileify-resources!` creates a temporary path. What we really want on drop is a string that will be resolved to an existing workspace resource.

editor/src/clj/editor/image.clj
```diff
@@ -25,13 +26,18 @@
...
 (set! *warn-on-reflection* true)
 
 (def exts ["jpg" "png"])
 
+(defn image-path?
+  [path]
+  (->> exts
+       (some (partial str/ends-with? (str/lower-case path)))
+       boolean))
+
```
sprocketc, line 40: Moved to `image` ns and also added `str/lower-case` to handle some additional missed cases.

editor/src/clj/editor/workspace.clj
```diff
@@ -32,7 +32,6 @@ ordinary paths."
             [editor.ui :as ui]
             [editor.url :as url]
             [editor.util :as util]
-            [internal.cache :as c]
```
sprocketc, line 35: Unrelated unused require.

editor/src/clj/editor/scene.clj
```diff
@@ -929,7 +929,6 @@
   (let [x          (:x action)
         y          (:y action)
         screen-pos (Vector3d. x y 0)
-        view-graph (g/node-id->graph-id view)
```
sprocketc, line 932: Unrelated but seems to be redundant.

editor/src/clj/editor/gui.clj
```diff
@@ -3266,11 +3266,12 @@
...
-  (input layouts-node g/NodeID) ; for tests
-  (input fonts-node g/NodeID) ; for tests
-  (input textures-node g/NodeID) ; for tests
-  (input particlefx-resources-node g/NodeID) ; for tests
+  (input layers-node g/NodeID)
+  (input layouts-node g/NodeID)
+  (input fonts-node g/NodeID)
+  (input textures-node g/NodeID)
+  (input particlefx-resources-node g/NodeID)
+  (input materials-node g/NodeID)
```
sprocketc, line 3274: Added `materials-node` and removed the comments since I used some of the inputs to get to the appropriate node. I don't know if those comments were meant to discourage usage outside of test.

editor/src/clj/editor/gui.clj
```diff
@@ -3934,6 +3936,42 @@
...
+                       (outline/resolve-id base-name))]
+    (cond
+      (= ext "particlefx")
+      (add-particlefx-resource scene (g/node-value scene :particlefx-resources-node) resource (gen-name :particlefx-resources-node))
+
+      (= ext "font")
+      (add-font scene (g/node-value scene :fonts-node) resource (gen-name :fonts-node))
+
+      (some #{ext} (workspace/resource-kind-extensions workspace :atlas))
+      (add-texture scene (g/node-value scene :textures-node) resource (gen-name :textures-node))
```
sprocketc, line 3954: This case should include `["atlas" "tilesource"]` and custom formats if I understand this correctly.

vlaaad (dismissed):
Works great! LGTM!

matgis (changes requested):
Looks great! I think we may want to prevent adding duplicates of already-existing Gui resources, though. If a Gui already references `player.atlas`, there's no good reason to create a new reference to it under a different name. You do want that when dragging images onto an AtlasAnimation, though. But similarly, it makes no sense to be able to add the same image to the non-animation section of an Atlas multiple times either (not sure if this was possible).

sprocketc:
> Looks great! I think we may want to prevent adding duplicates of already-existing Gui resources, though. If a Gui already references `player.atlas`, there's no good reason to create a new reference to it under a different name. You do want that when dragging images onto an AtlasAnimation, though. But similarly, it makes no sense to be able to add the same image to the non-animation section of an Atlas multiple times either (not sure if this was possible).

@matgis That's what I thought, but then I realized that we don't block this if you use the modal to add a dependency, so I tried to mimic the behavior. Should I change both cases? I wasn't sure if there is a valid case, so I decided to leave it as it is.

matgis:
Oh, I see! Yeah, we should fix that in both cases. I'll approve this PR for now, and let you decide if you want to address it here or in a separate PR. 👍

matgis (dismissed):

sprocketc:
@matgis We now filter out existing resources in both cases. Note that in atlases we already do that, but we don't seem to use `outline/resolve-id` to generate a unique name for different resources that share the same name. I'll handle this on  different PR.

sprocketc:
I overlooked the SDK comment, so I had to fix the arguments of `query-and-add-resources`. If we want to avoid using optional arguments with varargs, I can introduce a new method.

sprocketc:
I added a `:link` to the material-resource to fix the filtering of materials, and enabled `:outline-show-link?`. I also decided to convert `query-and-add-resources!` to multi-arity and pass the `accept-fn` directly, to make it easier for extensions to use this.

vlaaad:
> If a Gui already references player.atlas, there's no good reason to create a new reference to it under a different name

I think there might be a use case for using duplicate atlases with different names. First occurrence might contain default UI elements, and second might contain overridable UI elements, where overriding is achieved by using a different atlas with the name of the second texture (this doesn't seem to work in the editor, but it works in runtime)

sprocketc:
Reverted to the initial submission.

matgis (approved):
I say let't merge it!

PR #10462, opened 2025-04-08
Use unique identifiers for factory-spawned game objects

### LINKED ISSUE #9851: GO ids are reused / not unique
This can lead to bugs that are very difficult to debug. For example:

* I create an object in a manager script, and set up a timer to delete this object.
* I unload the current collection proxy and load in a new one. The object is deleted, but I forget to remove the timer.
* The timer fires, and deletes an object with the original id - which is now potentially a completely different object.

Note that this setup was poorly constructed by me in the first place. Because I neglected to reset the timer on collection switch, I _was_ often getting an error about game objects not existing when a deletion request came up. So in an attempt to fix this, I wrapped go.delete() in a go.exists() check first. Inadvertently, this means I created a situation that worked fine most of the time, but when I bumped into reused game object ids, I experienced this very tricky debug situation.

AGulev:
All objects spawned with factories now have unique identifiers, which helps to avoid hidden issues

Fix https://github.com/defold/defold/issues/9851

### DIFF
--- a/engine/gameobject/src/dmsdk/gameobject/gameobject.h
+++ b/engine/gameobject/src/dmsdk/gameobject/gameobject.h
@@ -547,11 +547,11 @@ namespace dmGameObject
     void Delete(HCollection collection, HInstance instance, bool recursive);
 
     /*#
-     * Construct a hash of an instance id.
-     * @name ConstructInstanceId
-     * @return id [type: dmhash_t] hash of the instance id constructed.
+     * Creates a new unique instance ID and returns its hash.
+     * @name CreateIntanceId
+     * @return id [type: dmhash_t] hash of the new unique instance id
      */
-    dmhash_t ConstructInstanceId();
+    dmhash_t CreateIntanceId();
 
     /*#
      * Retrieve an instance index from the index pool for the collection.
@@ -594,15 +594,6 @@ namespace dmGameObject
      * @return instance [type: dmGameObject::HInstance] Instance. NULL if instance isn't found.
      */
     HInstance GetInstanceFromIdentifier(HCollection collection, dmhash_t identifier);
-
-    /*#
-     * Construct and set instance identifier.
-     * @name SetConstructedIdentifier
-     * @param collection [type: dmGameObject::HCollection] Collection
-     * @param instance [type: dmGameObject::HInstance] Instance
-     * @return result [type: dmGameObject::Result]  RESULT_OK on success
-     */
-    Result SetConstructedIdentifier(HCollection collection, HInstance instance);
 
     /*#
      * Get component id from component index.
--- a/engine/gameobject/src/dmsdk/gameobject/gameobject.h
+++ b/engine/gameobject/src/dmsdk/gameobject/gameobject.h
@@ -548,10 +548,10 @@ namespace dmGameObject
 
     /*#
      * Creates a new unique instance ID and returns its hash.
-     * @name CreateIntanceId
+     * @name CreateInstanceId
      * @return id [type: dmhash_t] hash of the new unique instance id
      */
-    dmhash_t CreateIntanceId();
+    dmhash_t CreateInstanceId();
 
     /*#
      * Retrieve an instance index from the index pool for the collection.
--- a/engine/gameobject/src/dmsdk/gameobject/gameobject.h
+++ b/engine/gameobject/src/dmsdk/gameobject/gameobject.h
@@ -547,7 +547,7 @@ namespace dmGameObject
     void Delete(HCollection collection, HInstance instance, bool recursive);
 
     /*#
-     * Construct a hash of an instance id based on the index provided.
+     * Construct a hash of an instance id.
      * @name ConstructInstanceId
      * @return id [type: dmhash_t] hash of the instance id constructed.
      */
@@ -594,6 +594,15 @@ namespace dmGameObject
      * @return instance [type: dmGameObject::HInstance] Instance. NULL if instance isn't found.
      */
     HInstance GetInstanceFromIdentifier(HCollection collection, dmhash_t identifier);
+
+    /*#
+     * Construct and set instance identifier.
+     * @name SetConstructedIdentifier
+     * @param collection [type: dmGameObject::HCollection] Collection
+     * @param instance [type: dmGameObject::HInstance] Instance
+     * @return result [type: dmGameObject::Result]  RESULT_OK on success
+     */
+    Result SetConstructedIdentifier(HCollection collection, HInstance instance);
 
     /*#
      * Get component id from component index.
--- a/engine/gameobject/src/dmsdk/gameobject/gameobject.h
+++ b/engine/gameobject/src/dmsdk/gameobject/gameobject.h
@@ -549,10 +549,9 @@ namespace dmGameObject
     /*#
      * Construct a hash of an instance id based on the index provided.
      * @name ConstructInstanceId
-     * @param index [type: uint32_t] The index to base the id off of.
      * @return id [type: dmhash_t] hash of the instance id constructed.
      */
-    dmhash_t ConstructInstanceId(uint32_t index);
+    dmhash_t ConstructInstanceId();
 
     /*#
      * Retrieve an instance index from the index pool for the collection.
--- a/engine/gameobject/src/gameobject/gameobject.cpp
+++ b/engine/gameobject/src/gameobject/gameobject.cpp
@@ -970,7 +970,7 @@ namespace dmGameObject
     {
         static uint32_t index = 0;
         index += 1;
-        char buffer[16] = { 0 };
+        char buffer[32] = { 0 };
         int length = dmSnPrintf(buffer, sizeof(buffer), "%sinstance%d", ID_SEPARATOR, index);
         return dmHashBuffer64(buffer, (uint32_t)length);
     }
@@ -1035,9 +1035,15 @@ namespace dmGameObject
         return RESULT_OK;
     }
 
-    Result SetIdentifier(HCollection hcollection, HInstance instance, dmhash_t id)
+    Result SetIdentifier(HCollection collection, HInstance instance, dmhash_t id)
     {
-        return SetIdentifier(hcollection->m_Collection, instance, id);
+        return SetIdentifier(collection->m_Collection, instance, id);
+    }
+
+    Result SetConstructedIdentifier(HCollection collection, HInstance instance)
+    {
+        instance->m_Generated = true;
+        return SetIdentifier(collection->m_Collection, instance, ConstructInstanceId());
     }
 
     void ReleaseIdentifier(Collection* collection, HInstance instance)
@@ -1657,6 +1663,7 @@ namespace dmGameObject
         }
 
         HInstance instance = SpawnInternal(hcollection->m_Collection, proto, prototype_name, id, property_container, position, rotation, scale);
+        instance->m_Generated = 1;
 
         if (instance == 0) {
             dmLogError("Could not spawn an instance of prototype %s.", prototype_name);
--- a/engine/gameobject/src/gameobject/gameobject.cpp
+++ b/engine/gameobject/src/gameobject/gameobject.cpp
@@ -91,7 +91,7 @@ namespace dmGameObject
     static bool InitCollection(Collection* collection);
     static bool FinalCollection(Collection* collection);
 
-    static uint32_t g_instance_index = 0;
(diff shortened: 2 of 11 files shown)


engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -966,9 +966,11 @@ namespace dmGameObject
         return instance;
     }
 
-    dmhash_t ConstructInstanceId(uint32_t index)
+    dmhash_t ConstructInstanceId()
     {
-        char buffer[16] = { 0 };
+        static uint32_t index = 0;
+        index += 1;
+        char buffer[32] = { 0 };
```
AGulev, line 973: We need at least 20 bytes `/` + `instance` + `uint32` + = 1 + 8 + 10 + 1

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -1033,9 +1035,15 @@ namespace dmGameObject
...
+    Result SetIdentifier(HCollection collection, HInstance instance, dmhash_t id)
     {
-        return SetIdentifier(hcollection->m_Collection, instance, id);
+        return SetIdentifier(collection->m_Collection, instance, id);
+    }
+
+    Result SetConstructedIdentifier(HCollection collection, HInstance instance)
+    {
+        instance->m_Generated = true;
+        return SetIdentifier(collection->m_Collection, instance, ConstructInstanceId());
```
AGulev, line 1046: If we generate ID, we want to mark the instance as `m_Generated`, so it will release hash when object will be removed (we already have this functionality which used only for collection factories)

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -1655,6 +1663,7 @@ namespace dmGameObject
         }
 
         HInstance instance = SpawnInternal(hcollection->m_Collection, proto, prototype_name, id, property_container, position, rotation, scale);
+        instance->m_Generated = 1;
```
AGulev, line 1666: all factory spawned objects are `m_Generated`

AGulev:
I also carefully tested it to make sure the memory doesn't grow constantly 
CleanShot 2025-04-08 at 18 54 38@2x

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -1217,6 +1225,7 @@ namespace dmGameObject
 
         if (success) {
             AddToUpdate(collection, instance);
+            instance->m_Generated = 1;
```
AGulev, line 1228: all factory spawned objects are m_Generated

engine/gameobject/src/dmsdk/gameobject/gameobject.h
```diff
@@ -547,12 +547,11 @@ namespace dmGameObject
...
 
     /*#
-     * Construct a hash of an instance id based on the index provided.
+     * Construct a hash of an instance id.
      * @name ConstructInstanceId
-     * @param index [type: uint32_t] The index to base the id off of.
      * @return id [type: dmhash_t] hash of the instance id constructed.
      */
-    dmhash_t ConstructInstanceId(uint32_t index);
+    dmhash_t ConstructInstanceId();
```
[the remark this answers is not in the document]
britzl: It would be nice to not break stuff, but on the other hand we do have control over both the Spine and Rive extensions. But perhaps best to keep the index and respect it if > 0 and skip if == 0?

engine/gameobject/src/gameobject/test/factory/test_gameobject_factory.cpp
```diff
@@ -160,7 +160,7 @@ TEST_F(FactoryTest, Factory)
     for (int i = 0; i < count; ++i)
     {
         uint32_t index = dmGameObject::AcquireInstanceIndex(m_Collection);
```
britzl, line 162: Is this line needed if the index is no longer used?

engine/gameobject/src/dmsdk/gameobject/gameobject.h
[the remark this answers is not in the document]
JCash: I think it's best to move forward here and make it a breaking change.
As mentioned, it's only Rive/Spine that uses this so it's well worth doing it right now, as opposed to waiting for more extensions to use the api.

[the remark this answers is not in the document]
AGulev: It maybe a reason of collisions, only if some other unique prefix will be used for such cases, not default `instance`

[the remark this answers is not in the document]
JCash: Btw, shall we rename it to CreateInstanceId()?
"Construct" feels like an unusual name for an api imho

engine/gameobject/src/gameobject/gameobject.cpp
JCash: Perhaps a code comment?

engine/gameobject/src/gameobject/test/factory/test_gameobject_factory.cpp
AGulev: In this particular case it is not necessary, but in general we use index not just for name, but also to check if we can spawn an instance like
```
(index != dmGameObject::INVALID_INSTANCE_POOL_INDEX)
```
and then we assign index to the instance:
```
dmGameObject::AssignInstanceIndex(index, spawned_instance);
```

I can remove it in cases we don't call `AssignInstanceIndex`

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -1038,6 +1046,12 @@ namespace dmGameObject
         return SetIdentifier(hcollection->m_Collection, instance, id);
     }
 
+    Result SetConstructedIdentifier(HCollection hcollection, HInstance instance)
+    {
+        instance->m_Generated = 1;
+        return SetIdentifier(hcollection->m_Collection, instance, ConstructInstanceId());
+    }
```
JCash, line 1053: I don't think we need this function?
The reason is I don't think we need the `m_Generated`?
The only place it's used is for the debug functionality of checking to call the `dmHashReverseErase64(instance->m_Identifier);` and `dmScript::ReleaseHash(L, instance->m_Identifier);`

Both of those functions can (should at least!) handle non existent entries.
So, we can reduce complexity altogether by removing the m_Generated flag.

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -1202,6 +1216,7 @@ namespace dmGameObject
 
         bool success = CreateComponents(collection, instance);
         if (!success) {
+            dmLogError("Could not create components.");
```
JCash, line 1219: If we don't have the reason here, then the reason should be printed in the CreateComponents() perhaps?
I don't feel this error message is super helpful tbh (but still adds to engine size)

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -966,10 +968,16 @@ namespace dmGameObject
         return instance;
     }
 
-    dmhash_t ConstructInstanceId(uint32_t index)
+    void ResetInstanceIndex()
+    {
+        g_instance_index = 0;
```
JCash, line 973: style: `g_InstanceIndex`

engine/gameobject/src/gameobject/test/factory/test_gameobject_factory.cpp
britzl: Ah, got it. The AcquireInstanceIndex() name is a bit confusing then if we no longer care about the index it returns? Instead of Pushing and Popping indices it seems like we should have an uint32_t counter or something? Or am I misunderstanding something?

JCash (changes requested):
overall good, but feel we can do some adjustments

engine/gameobject/src/gameobject/test/factory/test_gameobject_factory.cpp
AGulev: You can return them in random order. So it can't be a counter.

engine/gameobject/src/gameobject/gameobject.cpp
AGulev: I think the biggest issue here not non-existent entries, but removing some hashes which are in use.
Let me explain.
We don't have reference counters for hashes.
User may have just a "/go" in two collections, and it's not factory generated one, just a collection. As a result, when such `/go` removed in one place, it will be removed from the ReverseHashTable, and in all the places where `/go` used for the second `/go` it will be hard to debug.

That's why I mentioned that having a unique name to be able to freely release it from reverse table is important. Which is possible only for the cases we create them ourselves

JCash: Sure I get it.
The problem though is that the id's you mention are thus "leaked", and will always stay in the reverse hash table 
I think that's not a "clean" solution either. Granted, we've had it for a long time, but I just don't like the fact that we can't treat all game objects or id's equal.

Also, I _really_ don't like that the api's gets so dictated by the debug functionality either.
E.g. this new function, giving us 2 functions to call, which is quite ambiguous.

I think a better way to solve this, is if the gameobject.cpp has it's own debug store of hashes, with ref counts.
And when they get deleted from there, we remove it from the actual reverse hash as well.

AGulev: I think I can remove this `SetConstructedIdentifier` function. Let me think about it a little bit

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -2175,6 +2185,7 @@ namespace dmGameObject
     void SetBone(HInstance instance, bool bone)
     {
         instance->m_Bone = bone;
+        instance->m_Generated = 1;
```
AGulev, line 2188: as well as bones (Bones by design created in runtime)

engine/gameobject/src/gameobject/gameobject.h
```diff
@@ -391,6 +391,9 @@ namespace dmGameObject
      * tracked by the collection, e.g resource.release(id) has been called
      */
     void RemoveDynamicResourceHash(HCollection collection, dmhash_t resource_hash);
+
+    // For tests
+    void ResetInstanceIndex();
```
AGulev, line 396: I don't like it. Should I rewrite these tests to make sure we don't expect exact names?

engine/gameobject/src/dmsdk/gameobject/gameobject.h
```diff
@@ -547,12 +547,11 @@ namespace dmGameObject
...
-     * Construct a hash of an instance id based on the index provided.
-     * @name ConstructInstanceId
-     * @param index [type: uint32_t] The index to base the id off of.
-     * @return id [type: dmhash_t] hash of the instance id constructed.
+     * Creates a new unique instance ID and returns its hash.
+     * @name CreateInstanceId
+     * @return id [type: dmhash_t] hash of the new unique instance id
      */
-    dmhash_t ConstructInstanceId(uint32_t index);
+    dmhash_t CreateInstanceId();
```
AGulev, line 554: New function name. I'll update Spine and Rive when merge this PR

engine/gameobject/src/gameobject/gameobject.h
JCash: Sure, it may be beneficial to not rely on the names?
And also one function less to take into consideration when testing.

JCash (dismissed):

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -966,10 +968,16 @@ namespace dmGameObject
         return instance;
     }
 
-    dmhash_t ConstructInstanceId(uint32_t index)
+    void ResetInstanceIndex()
+    {
+        g_instance_index = 0;
+    }
```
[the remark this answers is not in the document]
AGulev: Fixed

engine/gameobject/src/gameobject/gameobject.h
AGulev: Fixed!

engine/gameobject/src/gameobject/test/factory/test_gameobject_factory.cpp
britzl: Keep them!

britzl (approved):

PR #10469, opened 2025-04-10
Hide toolbars of inactive tabs

### LINKED ISSUE #10468: Shared toolbar on split pane configuration
When we open multiple split tab panes, the toolbar content is populated based on the context of the active tab, but is displayed on all tabs. Interacting with a toolbar of an inactive tab, updates the active tab instead.

https://github.com/user-attachments/assets/c34b5cbd-dbde-4494-a78b-462c5ad97c78

sprocketc:
This should hide all toolbars of inactive tabs.

Fixes #10468

### DIFF
--- a/editor/src/clj/editor/ui.clj
+++ b/editor/src/clj/editor/ui.clj
@@ -1914,10 +1914,17 @@
   [^Scene scene evaluation-context]
   (let [visible-command-contexts (visible-command-contexts scene)
         current-command-contexts (current-command-contexts scene)
-        root (.getRoot scene)]
+        root (.getRoot scene)
+        app-view (-> current-command-contexts first :env :app-view)
+        active-tab (g/maybe-node-value app-view :active-tab evaluation-context)]
     (doseq [td (vals (user-data root ::toolbars))]
-      (refresh-toolbar td visible-command-contexts evaluation-context)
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/ui.clj
```diff
@@ -1914,10 +1914,17 @@
...
-        root (.getRoot scene)]
+        root (.getRoot scene)
+        app-view (-> current-command-contexts first :env :app-view)
+        active-tab (g/maybe-node-value app-view :active-tab evaluation-context)]
     (doseq [td (vals (user-data root ::toolbars))]
-      (refresh-toolbar td visible-command-contexts evaluation-context)
-      (refresh-toolbar-state (:control td) current-command-contexts evaluation-context))))
+      (let [control (:control td)
+            is-active (boolean (when active-tab
+                                 (nodes-along-path? control (.getContent ^Tab active-tab) root)))]
```
sprocketc, line 1923: This would also hide all toolbars when we focus out of tabs. I don't think that's bad, since we already do that when we focus on the asset tree.

vlaaad (approved):
LGTM!

matgis (approved):
Nice catch! I think this may have been a regression introduced at some point. I could swear it used to only show the tool bar in the focused view before. 🤔

Anyway, nice to get it working as intended! ❤️

PR #10475, opened 2025-04-10
Material constants now contain no duplicates

### LINKED ISSUE #10456: Undefined uniform value when using the same name in vertex and fragment shaders
I haven't updated all shaders in old games to version 140 yet, and when testing our games on Vulkan macOS, I discovered that most of our materials don't look as they should.

I checked all the shaders and realized that the problem is that I use the same names for uniforms in vertex and fragment shaders, as I use the value of one uniform in both shaders.

1) This works perfectly if the shader is an old version and the renderer is OpenGL (or WebGL).
2) When specifying shader version 140 and using OpenGL renderer, only the vertex shader receives the uniform value, while the fragment shader receives a value of `vec4(0.0, 0.0, 0.0, 0.0)`.
3) With Vulkan renderer, regardless of shader version, only the vertex shader receives the uniform value, while the fragment shader receives a value of `vec4(0.0, 0.0, 0.0, 0.0)`.
4) <!> In all variants above, if you set a new uniform value from a script using `go.set()`, then both shaders receive the new value.

Question: which behavior is correct? I believe that both shaders should receive the uniform value specified in the material, as it was for older shader versions and OpenGL backend - this doesn't violate standards, as far as I can tell.

  - Version: 1.9.8, 1.10.0

**Platforms (REQUIRED):**
 - Platforms: Windows, macOS, HTML5

The project: 

[test_model_tint_new_vs_old.zip](https://github.com/user-attachments/files/19639351/test_model_tint_new_vs_old.zip)

From my point of view, the demo should look like that:
Image

In reality, it looks like that on OpenGL (Windows):
Image

On Vulkan (Windows):
Image

Screenshot from RenderDoc (sorry, not from the repro project, but from my game - but the point is to show that the value of uniform is `vec4(0.0)`):
Image

JCash:
Fixes https://github.com/defold/defold/issues/10456

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/engine/render/src/render/program_utils.cpp
+++ b/engine/render/src/render/program_utils.cpp
@@ -251,7 +251,7 @@ namespace dmRender
 
                 dmRender::SetConstantLocation(render_constant, uniform_desc.m_Location);
 
-                // We are be about to add a duplicate. Make sure to reuse the data
+                // We are about to add a duplicate. Make sure to reuse the data
                 HConstant prev_constant = FindConstant(constants, uniform_desc.m_NameHash);
                 if (prev_constant != 0)
                 {
--- a/engine/render/src/render/program_utils.cpp
+++ b/engine/render/src/render/program_utils.cpp
@@ -197,6 +197,18 @@ namespace dmRender
         *samplers_count_out = samplers_count;
     }
 
+    static Constant* FindConstant(dmArray<RenderConstant>& constants, dmhash_t name_hash)
+    {
+        uint32_t n = constants.Size();
+        for (uint32_t i = 0; i < n; ++i)
+        {
+            RenderConstant* constant = &constants[i];
+            if (constant->m_Constant->m_NameHash == name_hash)
+                return constant->m_Constant;
+        }
+        return 0;
+    }
+
     void SetProgramConstantValues(dmGraphics::HContext graphics_context, dmGraphics::HProgram program, uint32_t total_constants_count, dmHashTable64<dmGraphics::HUniformLocation>& name_hash_to_location, dmArray<RenderConstant>& constants, dmArray<Sampler>& samplers)
     {
         const uint32_t buffer_size = 128;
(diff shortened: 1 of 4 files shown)


engine/render/src/render/program_utils.cpp
```diff
@@ -236,23 +248,38 @@ namespace dmRender
             if (uniform_desc.m_Type == dmGraphics::TYPE_FLOAT_VEC4 || uniform_desc.m_Type == dmGraphics::TYPE_FLOAT_MAT4)
             {
                 HConstant render_constant = dmRender::NewConstant(uniform_desc.m_NameHash);
+
                 dmRender::SetConstantLocation(render_constant, uniform_desc.m_Location);
 
-                if (uniform_desc.m_Type == dmGraphics::TYPE_FLOAT_MAT4)
+                // We are be about to add a duplicate. Make sure to reuse the data
+                HConstant prev_constant = FindConstant(constants, uniform_desc.m_NameHash);
+                if (prev_constant != 0)
```
JCash, line 256: When creating the initial material constants (when loading the material), we check if there are duplicates.
If so, we make sure it uses the same memory as the first constant (with the same name).

This makes sure we can have multiple uniforms (vs + fs) with different locations, but still use get/set the same values.

engine/render/src/render/program_utils.cpp
```diff
@@ -236,23 +248,38 @@ namespace dmRender
...
-                    num_values *= 4;
-                    dmRender::SetConstantType(render_constant, dmRenderDDF::MaterialDesc::CONSTANT_TYPE_USER_MATRIX4);
-                }
+                    uint32_t prev_num_values;
+                    dmVMath::Vector4* prev_values = GetConstantValues(prev_constant, &prev_num_values);
 
-                // Set correct size of the constant (Until the shader builder provides all the default values)
-                if (num_values > default_values_capacity)
+                    dmRender::SetConstantValuesRef(render_constant, prev_values, prev_num_values);
+                    dmRender::SetConstantType(render_constant, GetConstantType(prev_constant));
```
JCash, line 262: I do believe we need to check mor thoroughly in editor/bob for mismatching types etc.
For now, we use the type of the first uniform (as we are using an existing pointer to the data)

engine/render/src/render/program_utils.cpp
```diff
@@ -236,23 +248,38 @@ namespace dmRender
...
+                }
+                else
                 {
-                    default_values_capacity = num_values;
-                    delete[] default_values;
-                    default_values = new dmVMath::Vector4[default_values_capacity];
-                    memset(default_values, 0, default_values_capacity * sizeof(dmVMath::Vector4));
+                    if (uniform_desc.m_Type == dmGraphics::TYPE_FLOAT_MAT4)
```
JCash, line 266: Not a duplicate, then do the things we did before...

engine/render/src/render/program_utils.cpp
AGulev: Could you please create a task for bob?

AGulev (dismissed):
lgtm

PR #10476, opened 2025-04-10
Fix state when closing the inactive tab on split configurations

### LINKED ISSUE #10473: Closing the inactive tab on a split configuration makes the remaining tab unresponsive
If we close the inactive tab on a split configuration, the app will be left in an invalid state. The scene view will be unresponsive, and the outline will be empty. We will have to close and reopen the scene to fix this.

https://github.com/user-attachments/assets/0b75f83b-12e4-4bf9-a05b-3697f1b4fd2b

sprocketc:
Fixes #10473


editor/src/clj/editor/app_view.clj
```diff
@@ -1924,14 +1924,14 @@ If you do not specifically require different script states, consider changing th
...
+        (ui/add-style! old-editor-tab-pane "inactive")
+        (ui/remove-style! new-editor-tab-pane "inactive")
+        (g/set-property! app-view :active-tab-pane new-editor-tab-pane))
       (let [selected-tab (ui/selected-tab new-editor-tab-pane)
             resource-node (tab->resource-node selected-tab)
             view-type (tab->view-type selected-tab)]
-        (ui/add-style! old-editor-tab-pane "inactive")
-        (ui/remove-style! new-editor-tab-pane "inactive")
-        (g/set-property! app-view :active-tab-pane new-editor-tab-pane)
         (on-selected-tab-changed! app-view app-scene selected-tab resource-node view-type)))))
```
sprocketc, line 1935: When we remove the inactive tab-pane [here](https://github.com/defold/defold/pull/10476/files#diff-4094da286709d7f8895e995fa84b2aa81727ed5610ff8fcbfb8ceee4f831e699R1919), the selected tab seems to be left in an invalid state. Calling `on-selected-tab-changed!` seems to fix the issue. We could add this after the pane removal, but since we also call `.requestFocus`, we are going to end up here. The `new-editor-tab-pane` will be the same with the `old-editor-tab-pane`, so this change ensures that `on-selected-tab-changed!` will always get called.

matgis (changes requested):
I'm not sure about this fix. It seems we now end up calling `on-selected-tab-changed!` every time the input focus changes between fields when editing things like `game.project`, for example.

It is curious to me that `old-editor-tab-pane` and `new-editor-tab-pane` are identical when we receive the focus after the `requestFocus` call in the `ListChangeListener`. I would expect the `:active-tab-pane` property to still refer to the `TabPane` that was removed immediately prior to calling `requestFocus`, and `new-editor-tab-pane` to be the `TabPane` that just received focus.

Maybe the proper way to fix this is to have the `editor-tab-pane` function return the `node` we supply it if it is already a `TabPane` directly under the `"editor-tabs-split"` `SplitPane`? The implementation of the `editor-tab-pane` function also seems to not handle the situation where non-"editor-tab" `TabPanes` are used inside an "editor-tab", although I don't think we do that anywhere yet. 🤔

sprocketc:
> I'm not sure about this fix. It seems we now end up calling on-selected-tab-changed! every time the input focus changes between fields when editing things like game.project, for example.

I moved the call to avoid this.

> It is curious to me that old-editor-tab-pane and new-editor-tab-pane are identical when we receive the focus after the requestFocus call in the ListChangeListener. I would expect the :active-tab-pane property to still refer to the TabPane that was removed immediately prior to calling requestFocus, and new-editor-tab-pane to be the TabPane that just received focus.

I believe that when we click to close the tab, `on-selected-tab-changed!` is called. Then in this specific case, we remove the pane because there are no tabs under it. The `:active-tab-pane` is not updated because the element is removed. Then the `old-editor-tab-pane` and the `new-editor-tab-pane` will always be same until we create a new pane. I think that if we delay the pane removal, we will be able to focus on the pane afterwards, but that doesn't look like a solid solution.

> The implementation of the editor-tab-pane function also seems to not handle the situation where non-"editor-tab" TabPanes are used inside an "editor-tab", although I don't think we do that anywhere yet.

Adding a `recur` should do, if we want to fix this here. Is that what you had in mind?

matgis (approved):
LGTM!

PR #10483, opened 2025-04-11
⚠️ Removed `scale_along_z` from the engine

### LINKED ISSUE #5631: 3D position of models are not correct in engine vs editor
Sample project, go to example 003
https://github.com/subsoap/defold-shader-examples/archive/4ba05858380d0611823b4075a9b2f1d84d0b7761.zip

In editor
editor

In engine
2021-03-09 22_41_54-Defold Shader Examples

I've seen similar issues in past projects like with https://github.com/subsoap/fps where seemingly objects placed in the editor too closely bug out position wise

The models use a custom material but AFAIK it has nothing to do with it, I changed material to standard model material without changing the positions and they still bug out.

### LINKED ISSUE #5679: Scaling the Z axis scales the rendered mesh, but not the child game object positions
I have the following hierarchy:
* map: scale vec3(70, 70, 70)
  * object1: position vec3(0, 0, 0)
  * object2: position vec3(0, 0, 10)
  
The objects end up 10 units apart in world space, instead of 700 units. Each object is about 0.2 units in height and it correctly gets scaled when rendered (renders as 14 units high).

It seems like Defold, by default ignores Z scaling for game object transforms, which makes it really annoying for working with 3D. I discovered this is caused by `scale_along_z: 0` in the collection file and there is no way to change it from the editor.

IMHO, that's a bad default. It's irrelevant for 2D and very confusing once you start working in 3D. I know there was discussion about removing `scale_along_z` and I agree that it's not very useful, but I think that the default should be `scale_along_z: 1`, like in any other 3D engine.

**To Reproduce (REQUIRED)**
Do the above

The objects should end up 700 units apart in world space and should be rendered as 0.2 * 70 = 14 units high 

  - Current (1.2.180)

**Platforms (REQUIRED):**
 - Platforms: macOS
 - OS: Big Sur

**Workaround (OPTIONAL):**
`scale_along_z: 1` in the collection file

**Screenshots (OPTIONAL):**
Screenshot 2021-03-23 at 17 24 40

### LINKED ISSUE #6165: Scale of parent object broke local position of children objects in realtime 3d view.
how it looks in editor:

how it looks in runtime:

As you can see all planes are shifted from original position.
All ok if parent object scale is 1.0

Tested in Defold version 187, 188 and 189 beta.

A simple project based on 3d template reproducing this issue:
[Basic 3D project.zip](https://github.com/defold/defold/files/7473835/Basic.3D.project.zip)
Editor (189 beta):

Runtime:

### Steps to reproduce

<hr/>

<table>
<tr><td>Build time</td><td>2021-09-20T09:08:57.266447</td></tr>
<tr><td>Defold editor sha</td><td>03b14982258f69eb8a06d28810ae879e16bd63d7</td></tr>
<tr><td>Defold engine sha</td><td>581c6439ae93755a8a6bcf58732c39c724fa193c</td></tr>
<tr><td>Defold version</td><td>1.2.187</td></tr>
<tr><td>GPU</td><td>NVIDIA GeForce GTX 1060 6GB/PCIe/SSE2</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 NVIDIA 471.68</td></tr>
<tr><td>Java version</td><td>11.0.1+13</td></tr>
<table>

### LINKED ISSUE #7441: gameobject with multiple meshes not scaling correctly in-game
I made an interior cube out of six quads all under the same gameobject. When I scale the gameobject in the editor, it looks as it should. However, when running the game, two of the quads are not where they need to be.

To reproduce should be as such:
1. create project with 3d template
2. create empty gameobject, attach six quad mesh components arranged to form a cube
3. scale the gameobject in editor by any uniform amount greater than 1
4. build and run

You should see a scaled up cube like in the editor, but it looks like the positions of two of the sides haven't been scaled

I'm using Defold version 1.4.3 on Windows 10

Here's how it looks in the inspector (cube and transform)

inspector

inspector_transform

Here's how it looks in-game

bad_scale

Here I printed out the scale of the object just to make sure it's scaled correctly, and it says it is

debug

I first made a post about this on the forum: https://forum.defold.com/t/object-not-scaling-properly/72782/4

### LINKED ISSUE #9529: Remove Scale Along Z property in collections
**Task Overview:**

The goal is to remove the old hidden `scale_along_z` field from collections in the engine, which forces the engine to ignore the Z-component of scale for all entities (Game Objects and components). Since removing it outright would break existing projects that rely on this behavior, the objective is to gradually phase out this feature while providing an alternative mechanism for developers to control the scaling behavior.

**Proposed Steps:**

1. **Major Engine Release with Breaking Change**  
   The engine's next major release will introduce a breaking change by removing the `scale_along_z` behavior. This will ensure that future versions of the engine no longer rely on hidden fields that manage scaling, improving maintainability.

2. **Introduce a New `scale.xy` Property in the Lua API**  
   To replace the old behavior, a new property (`scale.xy` or a similar name) will be introduced for Game Objects and components via the Lua API. This new property will specifically ignore the Z-component of the scale. 

   - **For Old Projects:** Projects relying on the old behavior will need to replace all instances of `scale` properties (used in `go.set()` and `go.animate()`) with this new `scale.xy` property. 
   - This migration will be similar to the earlier transition from `gui.PROP_ROTATION` to `gui.PROP_EULER`, making it easier for developers to update their projects without thoroughly inspecting all scale values.

3. **Editor Migration Tool**  
   To assist developers with this transition, an editor migration tool will be introduced. This tool will:
   - Automatically replace all instances of `scale.z` in Game Objects and components within collections with `scale_along_z:0` to 1.0 or a similar mechanism to remove reliance on the deprecated field.
   - This automated tool will simplify the migration process, helping developers update their projects without manual intervention.

**Reasoning for the New `scale.xy` Property:**  
The new property allows developers to easily replace constants in their code, streamlining the process. Replacing `scale` with `scale.xy` across the project is far simpler than manually investigating and adjusting individual values for every use of `scale` property, making the transition smoother and less error-prone.

**Expected Outcome (REQUIRED):**
The engine will no longer contain any references or code related to `scale_along_z`. Instead, the correct scaling behavior (whether 2D or 3D) will be explicitly managed by the user through the Lua API (`scale.xy`), ensuring that scaling logic is clear and intentional in all projects. This will improve maintainability, remove legacy dependencies, and allow users greater control over their scaling behavior.

AGulev:
`scale_along_z` has been fully removed from the engine. Read the following information to understand how to migrate your projects.

#### What is `scale_along_z`?
`scale_along_z` was a hidden option in `*.collection` files, set to `0` by default.  
When set to `0`, scaling on objects and collections did **not** affect their z-position.  

For example, if your sprite has a position `x: 0.0, y: 0.0, z: 0.1`, and its GameObject has a scale `x: 1.0, y: 1.0, z: 2.0`:
- With `scale_along_z = 0`, the transform of the sprite remains `x: 0.0, y: 0.0, z: 0.1`
- With `scale_along_z = 1`, the transform becomes `x: 0.0, y: 0.0, z: 0.2`

#### Why Was It Removed?
This was a legacy feature that introduced special-case logic and added complexity to the engine. While it had utility in some 2D projects, it was often a source of confusion in 3D projects. To unify 2D and 3D workflows, we’ve made the difficult decision to remove it, even though it’s a breaking change.

#### Who Needs to Take Action?
If you are unfamiliar with `scale_along_z`, or you never explicitly set it to `1`, **you need to update your project**. If your project consistently used `scale_along_z: 1` across all collections, **no action is required**.

#### What Should Be Done in Lua Code?
To avoid unexpected behavior now that scaling affects Z-position:

- Replace all usages of `go.set_scale(number|vector3)` with `go.set_scale_xy(number|vector3)`
- When accessing the `scale` property via `go.get()`, `go.set()`, or `go.animate()`, use the `scale.xy` property instead

#### What Should Be Done in Project Files?
Review your GameObjects and Collections and check for use of the Z component in the `scale` property.  
If you find a non 1.0 Z scale in your project, investigate why it's there. In most of 2D cases, the Z scale should always be 1.0 because it doesn't affect anything. However, in some cases—such as pseudo-3D effects, perspective cameras, particles, 3D models, etc. Z-scaling may be intentional. These situations require manual review.

If all children of such an object have `position.z == 0`, then behavior will not change.  
But if `position.z` is non-zero, then the scale will now affect it, and you'll need to decide case by case how to proceed based on your game’s requirements.  
For example:
- You might need to adjust object positions to compensate for `scale.z` now affecting the children's positions.
- Or, set the parent object's `scale.z` to 1.0 and instead scale all children individually (adjusting positions accordingly).

Automatic migration is not possible due to the variety of use cases.  
Please review and adjust your project accordingly.  

We apologize for the inconvenience and hope that this change leads to a more consistent and understandable workflow for all users.

Fix https://github.com/defold/defold/issues/9529
Fix https://github.com/defold/defold/issues/7441
Fix https://github.com/defold/defold/issues/6165
Fix https://github.com/defold/defold/issues/5679
Fix https://github.com/defold/defold/issues/5631

### DIFF
--- a/engine/dlib/src/dlib/transform.h
+++ b/engine/dlib/src/dlib/transform.h
@@ -91,17 +91,6 @@ namespace dmTransform
     }
 
     /**
-     * Apply the transform on a point, but without scaling the Z-component of the point (includes the transform translation).
-     * @param t Transform
-     * @param p Point
-     * @return Transformed point
-     */
-    inline dmVMath::Point3 ApplyNoScaleZ(const TransformS1& t, const dmVMath::Point3 p)
-    {
-        return dmVMath::Point3(rotate(t.GetRotation(), mulPerElem(dmVMath::Vector3(t.GetScale(), t.GetScale(), 1.0f), dmVMath::Vector3(p))) + t.GetTranslation());
-    }
-
-    /**
      * Apply the transform on a vector (excludes the transform translation).
      * @param t Transform
      * @param v Vector
@@ -110,17 +99,6 @@ namespace dmTransform
     inline dmVMath::Vector3 Apply(const TransformS1& t, const dmVMath::Vector3 v)
     {
         return dmVMath::Vector3(rotate(t.GetRotation(), v * t.GetScale()));
-    }
-
-    /**
-     * Apply the transform on a vector, but without scaling the Z-component of the vector (excludes the transform translation).
-     * @param t Transform
-     * @param v Vector
-     * @return Transformed vector
-     */
-    inline dmVMath::Vector3 ApplyNoScaleZ(const TransformS1& t, const dmVMath::Vector3 v)
-    {
-        return dmVMath::Vector3(rotate(t.GetRotation(), mulPerElem(dmVMath::Vector3(t.GetScale(), t.GetScale(), 1.0f), v)));
     }
 
     /**
@@ -134,21 +112,6 @@ namespace dmTransform
         TransformS1 res;
         res.SetRotation(lhs.GetRotation() * rhs.GetRotation());
         res.SetTranslation(dmVMath::Vector3(Apply(lhs, dmVMath::Point3(rhs.GetTranslation()))));
-        res.SetScale(lhs.GetScale() * rhs.GetScale());
-        return res;
-    }
-
-    /**
-     * Transforms the right-hand transform by the left-hand transform, without scaling the Z-component of the transition of the transformed transform
-     * @param lhs Transforming transform
-     * @param rhs Transformed transform
-     * @return Transformed transform
-     */
-    inline TransformS1 MulNoScaleZ(const TransformS1& lhs, const TransformS1& rhs)
-    {
-        TransformS1 res;
-        res.SetRotation(lhs.GetRotation() * rhs.GetRotation());
-        res.SetTranslation(dmVMath::Vector3(ApplyNoScaleZ(lhs, dmVMath::Point3(rhs.GetTranslation()))));
         res.SetScale(lhs.GetScale() * rhs.GetScale());
         return res;
     }
--- a/engine/dlib/src/dmsdk/dlib/transform.h
+++ b/engine/dlib/src/dmsdk/dlib/transform.h
@@ -161,8 +161,9 @@ namespace dmTransform
         }
 
         /*# set scale for x and y
-         * @name SetScale
-         * @return scale [type: dmVMath::Vector3]
+         * @name SetScaleXY
+         * @param scale_x [type: float]
+         * @param scale_y [type: float]
          */
         inline void SetScaleXY(float scale_x, float scale_y)
         {
--- a/engine/dlib/src/dmsdk/dlib/transform.h
+++ b/engine/dlib/src/dmsdk/dlib/transform.h
@@ -160,6 +160,16 @@ namespace dmTransform
             m_Scale = dmVMath::Vector3(scale);
         }
 
+        /*# set scale for x and y
+         * @name SetScale
+         * @return scale [type: dmVMath::Vector3]
+         */
+        inline void SetScaleXY(float scale_x, float scale_y)
+        {
+            m_Scale.setX(scale_x);
+            m_Scale.setY(scale_y);
+        }
+
         /*# get rotatiom
          * @name GetRotation
          * @return rotation [type: dmVMath::Quat]
--- a/engine/dlib/src/dmsdk/dlib/transform.h
+++ b/engine/dlib/src/dmsdk/dlib/transform.h
@@ -197,20 +197,6 @@ namespace dmTransform
     }
 
     /*#
-     * Apply the transform on a point, but without scaling the Z-component of the point (includes the transform translation).
-     * @name ApplyNoScaleZ
-     * @param t [type: dmTransform::Transform&] Transform
-     * @param p [type: dmVMath::Point3&] Point
-     * @return point [type: dmVMath::Point3] Transformed point
-     */
-    inline dmVMath::Point3 ApplyNoScaleZ(const Transform& t, const dmVMath::Point3 p)
-    {
-        dmVMath::Vector3 s = t.GetScale();
-        s.setZ(1.0f);
-        return dmVMath::Point3(rotate(t.GetRotation(), mulPerElem(dmVMath::Vector3(p), s)) + t.GetTranslation());
-    }
-
-    /*#
      * Apply the transform on a vector (excludes the transform translation).
      * @name Apply
      * @param t [type: dmTransform::Transform&] Transform
@@ -220,20 +206,6 @@ namespace dmTransform
     inline dmVMath::Vector3 Apply(const Transform& t, const dmVMath::Vector3 v)
     {
         return dmVMath::Vector3(rotate(t.GetRotation(), mulPerElem(v, t.GetScale())));
-    }
-
-    /*#
-     * Apply the transform on a vector, but without scaling the Z-component of the vector (excludes the transform translation).
-     * @name ApplyNoScaleZ
-     * @param t [type: dmTransform::Transform&] Transform
-     * @param v [type: dmVMath::Vector3&] Vector
-     * @return point [type: dmVMath::Vector3] Transformed vector
-     */
-    inline dmVMath::Vector3 ApplyNoScaleZ(const Transform& t, const dmVMath::Vector3 v)
-    {
-        dmVMath::Vector3 s = t.GetScale();
-        s.setZ(1.0f);
-        return dmVMath::Vector3(rotate(t.GetRotation(), mulPerElem(v, s)));
     }
 
     /*#
@@ -248,22 +220,6 @@ namespace dmTransform
         Transform res;
         res.SetRotation(lhs.GetRotation() * rhs.GetRotation());
         res.SetTranslation(dmVMath::Vector3(Apply(lhs, dmVMath::Point3(rhs.GetTranslation()))));
-        res.SetScale(mulPerElem(lhs.GetScale(), rhs.GetScale()));
-        return res;
-    }
-
-    /*#
-     * Transforms the right-hand transform by the left-hand transform, without scaling the Z-component of the transition of the transformed transform
-     * @name MulNoScaleZ
-     * @param lhs [type: const dmTransform::Transform&]
-     * @param rhs [type: const dmTransform::Transform&]
-     * @return result [type: dmTransform::Transform] Transformed transform
-     */
-    inline Transform MulNoScaleZ(const Transform& lhs, const Transform& rhs)
-    {
-        Transform res;
-        res.SetRotation(lhs.GetRotation() * rhs.GetRotation());
-        res.SetTranslation(dmVMath::Vector3(ApplyNoScaleZ(lhs, dmVMath::Point3(rhs.GetTranslation()))));
         res.SetScale(mulPerElem(lhs.GetScale(), rhs.GetScale()));
         return res;
     }
@@ -370,23 +326,6 @@ namespace dmTransform
     {
         *target = source;
         NormalizeZScale(target);
-    }
-
-    /*#
-     * Multiply two matrices without z-scaling the translation in m2
-     * @name MulNoScaleZ
-     * @param m1 [type: const dmVMath::Matrix&] First matrix
-     * @param m2 [type: const dmVMath::Matrix&] Second matrix
-     * @return result [type: dmVMath::Matrix] The resulting transform
-     */
-    inline dmVMath::Matrix4 MulNoScaleZ(dmVMath::Matrix4 const &m1, dmVMath::Matrix4 const &m2)
-    {
-        // tmp is the non-z scaling version of m1
-        dmVMath::Matrix4 tmp, res;
-        NormalizeZScale(m1, &tmp);
-        res = m1 * m2;
-        res.setCol(3, tmp * m2.getCol(3));
-        return res;
     }
 }
 
--- a/engine/engine/src/test/label/label.script
+++ b/engine/engine/src/test/label/label.script
@@ -97,6 +97,20 @@ local function test_set_vector4(name, startvalue)
 	test_vector4(name, expected)
 end
 
+local function test_set_scale_xy(val)
+	local expected = vmath.vector3(val, val+1, 3)
+	local set_value = vmath.vector3(val, val+1, val+2)
+	go.set("#label", "scale.xy", set_value)
+    local v = go.get("#label", "scale")
+    assert(expected == v)
+
+	expected = vmath.vector3(val+3, val+3, 3)
+	set_value = val + 3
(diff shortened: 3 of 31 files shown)


engine/dlib/src/dlib/transform.h
```diff
@@ -90,17 +90,6 @@ namespace dmTransform
...
-     * Apply the transform on a point, but without scaling the Z-component of the point (includes the transform translation).
-     * @param t Transform
-     * @param p Point
-     * @return Transformed point
-     */
-    inline dmVMath::Point3 ApplyNoScaleZ(const TransformS1& t, const dmVMath::Point3 p)
-    {
-        return dmVMath::Point3(rotate(t.GetRotation(), mulPerElem(dmVMath::Vector3(t.GetScale(), t.GetScale(), 1.0f), dmVMath::Vector3(p))) + t.GetTranslation());
-    }
-
```
AGulev, line 103: We don't need these "noZ" mat functions anymore

engine/gameobject/src/gameobject/comp_anim.cpp
```diff
@@ -514,14 +514,18 @@ namespace dmGameObject
...
     {
         switch (type)
         {
         case PROPERTY_TYPE_NUMBER:
             return 1;
         case PROPERTY_TYPE_VECTOR3:
+        {
+            if (element_ids[2] == 0)
+                return 2;
             return 3;
```
AGulev, line 527: we don't need new type, we use vector3, but if only 2 element's specified, we use only 2 elements

engine/gameobject/src/gameobject/gameobject_script.cpp
```diff
@@ -1102,11 +1102,11 @@ namespace dmGameObject
      * go.set_scale(s)
      * ```
      *
-     * Set the scale of another game object instance with id "x":
+     * Set the scale of another game object instance with id "obj_id":
      *
      * ```lua
      * local s = 1.2
-     * go.set_scale(s, "x")
+     * go.set_scale(s, "obj_id")
```
AGulev, line 1109: it's a bit confusing to use `x` as an object name in this example

engine/gamesys/src/gamesys/components/comp_private.cpp
```diff
@@ -57,6 +57,13 @@ dmGameObject::PropertyResult GetProperty(dmGameObject::PropertyDesc& out_value,
     {
         out_value.m_Variant = dmGameObject::PropertyVar(ref_value.getZ());
     }
+    else if (get_property == property.m_XY)
+    {
+        out_value.m_ElementIds[0] = property.m_X;
+        out_value.m_ElementIds[1] = property.m_Y;
+        out_value.m_ElementIds[2] = 1.0f;
```
AGulev, line 64: always return 1.0 as z if only XY requested

engine/gameobject/src/gameobject/gameobject_script.cpp
```diff
@@ -1135,6 +1135,57 @@ namespace dmGameObject
...
+        Instance* instance = ResolveInstance(L, 2);
+
+        // Supports both vector and number
+        Vector3* v = dmScript::ToVector3(L, 1);
+        if (v != 0)
+        {
+            Vector3 scale = *v;
+            if (scale.getX() <= 0.0f || scale.getY() <= 0.0f)
+            {
+                return luaL_error(L, "Vector passed to go.set_scale contains components that are below or equal to zero");
```
britzl, line 1173: go.set_scale -> go.set_scale_xy

engine/gameobject/src/gameobject/gameobject_script.cpp
```diff
@@ -1135,6 +1135,57 @@ namespace dmGameObject
...
+            }
+            dmGameObject::SetScaleXY(instance, scale.getX(), scale.getY());
+            return 0;
+        }
+
+        lua_Number n = luaL_checknumber(L, 1);
+        if (n <= 0.0)
+        {
+            return luaL_error(L, "The scale supplied to go.set_scale must be greater than 0.");
```
britzl, line 1182: go.set_scale -> go.set_scale_xy

engine/gameobject/src/gameobject/gameobject_script.cpp
```diff
@@ -1135,6 +1135,57 @@ namespace dmGameObject
...
+     * ```lua
+     * local s = vmath.vector3(2.0, 1.0, 5.0)
+     * go.set_scale_xy(s) -- z will not be set here, only x and y
+     * ```
+     *
+     * Set the scale of another game object instance with id "obj_id":
+     *
+     * ```lua
+     * local s = 1.2
+     * go.set_scale_xy(s, "obj_id")
```
britzl, line 1159: Perhaps add the same comment here that "z will not be set here ... "

engine/dlib/src/dmsdk/dlib/transform.h
```diff
@@ -160,6 +160,16 @@ namespace dmTransform
             m_Scale = dmVMath::Vector3(scale);
         }
 
+        /*# set scale for x and y
+         * @name SetScale
+         * @return scale [type: dmVMath::Vector3]
+         */
```
britzl, line 166: Missing @param for scale_x and scale_y

britzl: And there is not return value

engine/engine/src/test/label/label.script
```diff
@@ -97,6 +103,20 @@ local function test_set_vector4(name, startvalue)
 	test_vector4(name, expected)
 end
 
+local function test_set_scale_xy(val)
+	local expected = vmath.vector3(val, val+1, 3)
+	local set_value = vmath.vector3(val, val+1, val+2)
+	go.set("#label", "scale.xy", set_value)
+    local v = go.get("#label", "scale")
+    assert(expected == v)
```
britzl, line 111: Indentation

engine/gameobject/src/dmsdk/gameobject/gameobject.h
```diff
@@ -664,6 +664,15 @@ namespace dmGameObject
      */
     void SetScale(HInstance instance, dmVMath::Vector3 scale);
 
+    /*# set scale only for width and height (x and y)
```
britzl, line 667: Are we talking about scale x and y as width and height elsewhere? I don't remember that we do. Perhaps better to just keep referring to it as "x and y scale" or "x and y axis" or perhaps only as "x and y"?

engine/gameobject/src/gameobject/comp_anim.cpp
britzl: How can we know this? Can't I do a go.animate() of "scale" with the z-component set to 0?

britzl (changes requested):
See comments

engine/gameobject/src/gameobject/comp_anim.cpp
AGulev: `element_id`is an id for a field I set here: https://github.com/defold/defold/blob/1fcad5cf368af8bd6f7234a64e438aa172ec9b64/engine/gameobject/src/gameobject/gameobject.cpp#L3242-L3245

for cases when we need only two components like `xy` I set Z to 0

engine/gameobject/src/dmsdk/gameobject/gameobject.h
AGulev: Thank you. Fixed!

engine/engine/src/test/label/label.script
AGulev: Thank you. Fixed!

engine/gameobject/src/gameobject/gameobject_script.cpp
AGulev: Thank you. Fixed!

engine/gameobject/src/gameobject/gameobject_script.cpp
AGulev: Thank you. Fixed!

engine/gameobject/src/gameobject/gameobject_script.cpp
AGulev: Thank you. Fixed!

engine/gameobject/src/gameobject/comp_anim.cpp
britzl: Oh, yes, now I see. 👍

britzl (dismissed):

matgis (dismissed):
The change itself might be disruptive, but the editor changes look good to me.

engine/dlib/src/dmsdk/dlib/transform.h
```diff
@@ -196,20 +207,6 @@ namespace dmTransform
...
-     * @param t [type: dmTransform::Transform&] Transform
-     * @param p [type: dmVMath::Point3&] Point
-     * @return point [type: dmVMath::Point3] Transformed point
-     */
-    inline dmVMath::Point3 ApplyNoScaleZ(const Transform& t, const dmVMath::Point3 p)
-    {
-        dmVMath::Vector3 s = t.GetScale();
-        s.setZ(1.0f);
-        return dmVMath::Point3(rotate(t.GetRotation(), mulPerElem(dmVMath::Vector3(p), s)) + t.GetTranslation());
-    }
```
JCash, line 211: I think this is acceptable, as we already will designate the entire release as a breaking change.

We ofc need to update Spine + Rive externsions.

engine/engine/src/test/label/label.script
```diff
@@ -13,38 +13,44 @@
...
-    local vy = go.get("#label", name .. ".y")
-    local vz = go.get("#label", name .. ".z")
-    local vw = go.get("#label", name .. ".w")
-
-    assert(expected == v)
-    assert(expected.x == vx)
-    assert(expected.y == vy)
-    assert(expected.z == vz)
-    assert(expected.w == vw)
+	local v = go.get("#label", name)
```
JCash, line 30: Replaced spaces with tabs?
Makes it harder to detect/read the actual changes.

engine/gameobject/src/gameobject/gameobject_script.cpp
```diff
@@ -1135,6 +1135,57 @@ namespace dmGameObject
...
+     * ```
+     *
+     * Set the scale of another game object instance with id "obj_id":
+     *
+     * ```lua
+     * local s = 1.2
+     * go.set_scale_xy(s, "obj_id") -- z will not be set here, only x and y
+     * ```
+     */
+    int Script_SetScaleXY(lua_State* L)
```
JCash, line 1162: Missing "static"

JCash (dismissed):
lgtm
some minor comments about "static" in the script module

engine/engine/src/test/label/label.script
AGulev: yes, I just made it consistent because most of the file uses tabs, and only a bit of code spaces.

JCash: well, in that case it should be the other way around (we use spaces in our code style) :)

JCash (approved):

PR #10490, opened 2025-04-13
Do not autocomplete Lua code in comments

### LINKED ISSUE #7840: Don't autocomplete when writing comments
image

In the picture above you can see that it's trying to autocomplete my text when I'm inserting a comment.

It's annoying if I want to type the word then press enter because it will with change the text if it matches a key word.

Can this be removed for comments or at least an option to stop it?

AGulev:
The code editor will no longer try to auto-complete Lua code in comments.

Fix https://github.com/defold/defold/issues/7840


vlaaad (approved):
LGTM!

PR #10491, opened 2025-04-13
Fixed the issue where horizontal scrolling in the text editor doesn't account for changes in code font size

### LINKED ISSUE #10002: Sometimes horizontal Scrolling doesn't work AND horizontal Scroll Bar just doesn't appear too
Sometimes, I can't see end of my code line in the Code Editor because Horizontal scrolling just don't work and also horizontal Scroll Bar don't appear too while the Code Overview prevent to see it.

**To Reproduce (REQUIRED)**
1. Go to Code Editor
2. Scroll to the end of the code line by clicking on right arrow on keyboard
3. Code Overview prevent to see the end of code line

See the end of my code.

  - Version 1.9.6

**Platforms (REQUIRED):**
 - Platforms: Linux, KDE Plasma 6.2.5 Wayland session
 - OS: EndeavourOS
 - Device: Lenovo B590
 - Display: 1366x768

**Video:**

https://github.com/user-attachments/assets/adbd7e50-8a6b-4534-a670-203eca5c6994

AGulev:
Fix https://github.com/defold/defold/issues/10002

### DIFF
--- a/editor/src/clj/editor/code/view.clj
+++ b/editor/src/clj/editor/code/view.clj
@@ -3820,13 +3820,6 @@
     (let [find-case-sensitive-setter (make-property-change-setter view-node :find-case-sensitive?)
           find-whole-word-setter (make-property-change-setter view-node :find-whole-word?)
           font-size-setter (make-property-change-setter view-node :font-size)
-          document-width-setter (make-property-change-setter view-node :document-width
-                                                             (fn [_]
-                                                               (let [glyph-metrics (g/node-value view-node :glyph-metrics)
-                                                                     tab-spaces (g/node-value view-node :tab-spaces)
-                                                                     tab-stops (data/tab-stops glyph-metrics tab-spaces)
-                                                                     lines (g/node-value view-node :lines)]
-                                                                 (data/max-line-width glyph-metrics tab-stops lines))))
           highlighted-find-term-setter (make-property-change-setter view-node :highlighted-find-term)
           visible-indentation-guides-setter (make-property-change-setter view-node :visible-indentation-guides?)
           visible-minimap-setter (make-property-change-setter view-node :visible-minimap?)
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/code/view.clj
```diff
@@ -676,6 +676,10 @@
 (g/defnk produce-indent-level-pattern [tab-spaces]
   (data/indent-level-pattern tab-spaces))
 
+(g/defnk produce-document-width [glyph-metrics tab-spaces lines]
+  (let [tab-stops (data/tab-stops glyph-metrics tab-spaces)]
+    (data/max-line-width glyph-metrics tab-stops lines)))
```
AGulev, line 681: `document-width` shuold be updated if text metrics changes, so we show horizontal scrollbar in time. We can do it using a listener for change text size event (I did it initially), or using graph and dependencies

matgis (changes requested):
You've correctly identified the issue, in that we must account for the updated `glyph-metrics` to calculate the document width. However, the new implementation is rather inefficient since it will measure all the lines of the whole document whenever the user edits the file.

Previously (and still, presumably broken?), the `document-width` property was adjusted to grow by considering the previous width and the new widths of just the lines affected by each edit in the `code.data/update-document-width-after-splice` function.

Your approach of calculating the `document-width` in an output is more elegant and in-line with how we typically want to solve this sort of thing in the graph. However, to remain efficient when typing into large files, it would be better to retain the `document-width` property and also update it in `set` clauses added to the `font-name` and `font-size` properties.

editor/src/clj/editor/code/view.clj
```diff
@@ -1136,8 +1146,13 @@
...
 
-  (property font-name g/Str (default "Dejavu Sans Mono"))
-  (property font-size g/Num (default (g/constantly default-font-size)))
+  (property font-name g/Str (default "Dejavu Sans Mono")
+            (set (fn [evaluation-context self _old-value _new-value]
+                   (update-document-width evaluation-context self))))
+  (property font-size g/Num
+            (default (g/constantly default-font-size))
+            (set (fn [evaluation-context self _old-value _new-value]
+                   (update-document-width evaluation-context self))))
```
AGulev, line 1155: I'm not happy with the result, because `tab-spaces` should be taken into account as well.  
Also, I see that we have `produce-layout`, which is called with all the same values—maybe it's a bit more lightweight.

The current implementation fixes the main issue, though.

matgis (approved):
LGTM! ❤️

PR #10492, opened 2025-04-14
Make it clear when a collision object shape is unnamed

### LINKED ISSUE #10429: Display unnamed physics shapes in the editor in a clearer way
By default, a newly created collision Box shape is displayed as "Box", although its id is "".
By looking at the outline, it's misleading and may cause extra work in finding out why the code doesn't work.

AGulev:
Make sure it's clear from the outline when physics shapes are unnamed (names are only needed for runtime operations):
CleanShot 2025-04-14 at 08 05 16@2x

Fix https://github.com/defold/defold/issues/10429


matgis (approved):
LGTM! ❤️

PR #10495, opened 2025-04-14
Increased text contrast in diff view

### LINKED ISSUE #10480: [SUGGESTION] improve color contrast in 'View Diff' window.
When displaying the difference between two files via 'View Diff' in the editor, the green highlighted text is difficult to read with the grey text.

A higher contrast color, such as black, would be better for the text.

Image

britzl:
Increased text contrast on green background when showing change diffs.

Old:

CleanShot 2025-04-14 at 11 31 45@2x

New:

CleanShot 2025-04-14 at 11 29 26@2x

Fixes #10480

### DIFF
--- a/editor/styling/stylesheets/modules/_diff-view.scss
+++ b/editor/styling/stylesheets/modules/_diff-view.scss
@@ -12,7 +12,15 @@
   }
 
   Text {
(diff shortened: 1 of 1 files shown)


vlaaad (approved):
Ship it!

PR #10507, opened 2025-04-15
Adding support to decode IMA ADPCM to WAV sound decoder

### LINKED ISSUE #9768: Support for ADPCM based sample compression in basic sound system
For lower end platform with both limited memory resources as well as low CPU power, it would be useful to support a version of this simple compression method. For many sounds (effects) the quality achievable will be more then adequate and decompression costs are low, while still overing a decent compression ration if compared to plain PCM16.

thengel01:
WAV decoder now recognizes format 0x11 (IMA ADPCM) and decodes it to PCM16.

Fixes https://github.com/defold/defold/issues/9768

### Technical changes

Suitable unit tests have been added.
Code should be cross platform. Tested on HTML5/WASM and x86_64 Linux.


engine/sound/src/decoders/decoder_wav.cpp
```diff
@@ -110,9 +111,22 @@ namespace dmSoundCodec
 
         struct DecodeStreamInfo {
             Info m_Info;
+            int m_IsADPCM : 1;
+            int : 31;
```
JCash, line 115: it is ok to use a `bool` as it's the first flag.
A bool is also smaller than an integer, which we like. (every bit counts).
For flags, we choose the smallest type first, in this case `uint8_t`.
Then ofc, you need to sort the structure members to respect the layout, to not introduce superfluous padding.

engine/sound/src/decoders/decoder_wav.cpp
```diff
@@ -110,9 +111,22 @@ namespace dmSoundCodec
...
+            struct ADPCM {
+                uint16_t m_BlockAlign;
+                uint16_t m_BlockFrames;
+                uint16_t m_OutFramesOffset;
+                dmArray<int16_t> m_OutBuffer;
+                dmArray<int8_t> m_InBuffer;
+                uint32_t m_InBufferOffset;
+                int32_t m_Pred[2];
+                int32_t m_StepIndex[2];
+                int32_t m_Step[2];
```
JCash, line 128: Sort on structure member size

engine/sound/src/decoders/decoder_wav.cpp
```diff
@@ -110,9 +111,22 @@ namespace dmSoundCodec
...
+            } m_ADPCM;
```
JCash, line 129: How often is this case expected to happen?
DO we have any numbers on % of sounds using ADPCM?
Why not allocate this memory when creating the decoder, as opposed to incurring this cost on every decoder?

engine/sound/src/decoders/decoder_wav.cpp
```diff
@@ -147,6 +158,9 @@ namespace dmSoundCodec
         if (header.m_ChunkID == FOUR_CC('R', 'I', 'F', 'F') &&
             header.m_Format == FOUR_CC('W', 'A', 'V', 'E')) {
 
+            DecodeStreamInfo *streamOut = new DecodeStreamInfo;
+            assert(streamOut);
```
JCash, line 162: Do not assert here (Have you actually gotten an OOM here?)
We avoid overly defensive coding (code size: every bit counts)
We either return an error (as seen above), but most often for "general use", we don't handle OOM.

engine/sound/src/decoders/decoder_wav.cpp
JCash: Also, the flag isn't needed if the m_ADPCM is a pointer.

engine/sound/src/test/test_sound.cpp
```diff
@@ -1401,6 +1412,205 @@ const TestParams params_verify_opus_test[] = {TestParams("loopback",
...
 
+#if !defined(GITHUB_CI) || (defined(GITHUB_CI) && !(defined(WIN32) || defined(__MACH__)))
+TEST_P(dmSoundVerifyAdpcmTest, Mix)
+{
+    TestParams params = GetParam();
+    dmSound::Result r;
+    dmSound::HSoundData sd = 0;
+    dmSound::NewSoundData(params.m_Sound, params.m_SoundSize, params.m_Type, &sd, 1234);
+
+    printf("verifying adpcm mix: frames: %d\n", params.m_FrameCount);
```
JCash, line 1423: While I don't know the length of this test, I remind to keep it to a minimum, as it's easy for them to add up into a long waiting time for them to finish.

JCash (changes requested):

engine/sound/src/decoders/decoder_wav.cpp
thengel01: Making it a pointer adds another allocation from the heap and will introduce more overhead.  Bool is not necessarily smaller than an int: rather compiler depended. Anyhow: adjusting layout.

engine/sound/src/decoders/decoder_wav.cpp
thengel01: Will adjust for alignment, Hardly makes sense to sort by size.

engine/sound/src/decoders/decoder_wav.cpp
thengel01: It's hardly "every deccoder". Every WAV decoder: yes. I'd expect a large number of SFX rather being ADPCM than plain vanilla PCM(16) -- just because of size. And if it's PCM16... frankly the data size will drawf anything allocatd here. You are looking at 3*2*4 bytes. Anything else involved in the process of loading and mixing this data is vastly larger. I frankly do not see the need to add the complexity of moving this into something separately allocated -- and introducing all the overhead caused by that.

engine/sound/src/decoders/decoder_wav.cpp
thengel01: So... you just want it to crash mysteriously "sometime later" once an OOM occurrs? Well, so be it: removed.
(aren't asserts anyways out when doing a release build?)

engine/sound/src/test/test_sound.cpp
thengel01: You are suggesting shorter test material? Or what is your concern here?

JCash: Yes, shorter wall clock time. We're running these tests a lot, so adding overly long test sounds adds to our turnaround times. E.g. in general, I'd like to restrict most sounds to <0.5 s. Not a hard limit ofc, but you get the idea.

engine/sound/src/decoders/decoder_wav.cpp
JCash: > Bool is not necessarily smaller than an int:

Yes, but we're also looking at our current set of compilers.
Point still stands, why not use an uint8_t then?

engine/sound/src/decoders/decoder_wav.cpp
JCash: Sure, but it's also about setting the code style for other people that arrive here.

engine/sound/src/decoders/decoder_wav.cpp
JCash: > So... you just want it to crash mysteriously "sometime later" once an OOM occurrs? 

it's part of keeping the build small. If you're running out of memory, you're having larger problems anyways.
Also, OOM happens quite rarely imo, and due to other factors like programmer miscalculations (e.g. integer overflows etc).

And, as I mentioned, you _do_ have the option to return. However, it's likely, it will still crash elsewhere.
And there is little _we_ can do about OOM, nor can the user at that point.

> aren't asserts anyways out when doing a release build?

There is no debug vs release. It's the same libraries. Only difference is logging and debugging services are disabled in a release build. Debug builds are built with -O2, but we still want asserts to catch programmer errors, so we still have them enabled in the libraries.

engine/sound/src/decoders/decoder_wav.cpp
JCash: Another option would be to create an DecodeStreamInfo_ADMPCM:
```c++
struct DecodeStreamInfo {
    ...
    uint8_t m_IsADPCM : 1;
 };
 
struct DecodeStreamInfoAdpcm {
    DecodeStreamInfo m_Info;
    <adpcm memmbers>
};
```

And then you'd new it with `new DecodeStreamInfoAdpcm` instead.
That way you can avoid any extra pointer lookups.

JCash (changes requested):
See latest suggestion about a new struct DecodeStreamInfoAdpcm

engine/sound/src/decoders/decoder_wav.cpp
```diff
@@ -109,12 +110,32 @@ namespace dmSoundCodec
...
+            uint8_t : 7;
             uint32_t m_Cursor;
             const void* m_Buffer;
             uint32_t m_BufferOffset;
             dmSound::HSoundData m_SoundData;
         };
+
+        struct DecodeStreamInfoADPCM : public DecodeStreamInfo {
+            DecodeStreamInfoADPCM() { m_IsADPCM = true; }
+            virtual ~DecodeStreamInfoADPCM() = default;
```
JCash, line 127: Note that we're more of a C engine than C++. Also known as"c-like-c++". As such we prefer a `has-a` strategy as opposed to an `is-a`.

We only have class overrides in two places:
* where the 3rd party c++ libraries force us to
* type specializations for `dmHashTable` and `dmIndexPool`

I think this is the last change needed for this PR, and we're good to go.

JCash (changes requested):
One tiny thing left.
DOn't forget to re-request a re-review, so it'll show up faster for me to review.

JCash (approved):

PR #10513, opened 2025-04-16
Optimized output path for HTML5 audio output

thengel01:
- Adds capability for sound devices to report back information about the output format they expect
- Configured JS output device code to work with non-interleaved float data (avoids conversion from mixer float format to int16_t and back & allows to use faster (native) copy code to feed data into audio context)
- Moved setup of audio output queue length to be a bit more configurable (two values: threaded vs non-threaded; both tweaked down a bit from original (very) conservative value)
- Changed time tracking in JS output code to use frames rather than seconds to avoid numerical errors

Tested for Linux and wasm-web targets. HTML5 target used Chrome as browser.


engine/sound/src/js/library_sound.js
```diff
@@ -67,32 +70,50 @@ var LibrarySoundDevice =
...
-                        c0[i] = getValue(samples+4*i, 'i16') / 32768;
-                        c1[i] = getValue(samples+4*i+2, 'i16') / 32768;
+
+                    // Setup buffer for data delivery...
+
+//Optimize buffer allocation (assumes immediate copy-out - does not work on many browsers (it should per spec, though))
+//                    var buf = this.bufferCache[frame_count];
+//                    if (!buf) {
+//                        buf = this.bufferCache[frame_count] = shared.audioCtx.createBuffer(2, frame_count, this.sampleRate);
+//                    }
```
JCash, line 80: Dead code

engine/sound/src/js/library_sound.js
```diff
@@ -67,32 +70,50 @@ var LibrarySoundDevice =
...
+                    var buf = shared.audioCtx.createBuffer(2, frame_count, this.sampleRate);
+
+                    for(var c=0;c<2;c++) {
+//Optimize sub array creation by caching -> will fail as soon as the internal HEAPF32 representation changes (often after many seconds of use). Disabled.
+//                        var input = this.arrayCache[samples];
+//                        if (!input) {
+//                            input = this.arrayCache[samples] = HEAPF32.subarray(samples / 4, samples / 4 + frame_count);
+//                        }
```
JCash, line 88: dead code

engine/sound/src/sound.cpp
```diff
@@ -334,8 +339,11 @@ namespace dmSound
         HDevice device = 0;
         OpenDeviceParams device_params;
 
-        // TODO: m_BufferCount configurable?
-        device_params.m_BufferCount = SOUND_OUTBUFFER_COUNT;
+// TODO: m_BufferCount configurable?
```
JCash, line 342: Indentation
TODO? Should if be a new task?

engine/sound/src/sound_dsp.h
```diff
@@ -350,6 +350,38 @@ static inline void ApplyGainAndInterleaveToS16(int16_t* out, float* in[], uint32
     }
 }
 
+static inline void ApplyGain(float* out[], float* in[], uint32_t num, float scale, float scale_delta)
```
JCash, line 353: I think it's time to split the file into `sound_dsp_wasm.h`, `sound_dsp_sse2.h` and `sound_dsp_cpu.h` as this file is quickly becoming large and hard to read.

engine/sound/src/sound.cpp
```diff
@@ -334,8 +339,11 @@ namespace dmSound
...
+        const uint16_t num_outbuffers = params->m_UseThread ? SOUND_OUTBUFFER_COUNT : SOUND_OUTBUFFER_COUNT_NO_THREADS;
```
JCash, line 343: What's the relevant difference between 3 and 4 here?
1) as mentioned, it should likely be configurable, or 2) using the same number to keep complexity down?

engine/sound/src/devices/device_wasapi.cpp
```diff
@@ -299,7 +299,7 @@ namespace dmDeviceWasapi
     }
 
     // We get this call after we've returned a non zero value from the DeviceWasapiGetAvailableFrames()
-    static dmSound::Result DeviceWasapiQueue(dmSound::HDevice _device, const int16_t* samples, uint32_t sample_count)
+    static dmSound::Result DeviceWasapiQueue(dmSound::HDevice _device, const void* samples, uint32_t sample_count)
```
JCash, line 302: Our "standard" is to rename this to `void* _samples`and then do:
```c++
const uint16_t* samples = (const uint16_t*)_samples;
```

This will result in less code changes, which makes PR more readable, and doesn't add to the distory of the code.

JCash (changes requested):

engine/sound/src/js/library_sound.js
thengel01: Yes. And? It is left in there to inform future updates.

engine/sound/src/js/library_sound.js
thengel01: Yes. And? It is left in there to inform future updates.

engine/sound/src/sound.cpp
thengel01: It's NOT my "TODO". This is as I found it. My changes are just "around it".

engine/sound/src/sound_dsp.h
thengel01: I'd like to keep things together for now.

engine/sound/src/sound.cpp
thengel01: 20ms more tollerance for late updates when no threading is used.

engine/sound/src/devices/device_wasapi.cpp
thengel01: Changed.

engine/sound/src/js/library_sound.js
JCash: > Yes. And?

There's a big different between writing a comment on whqt we want to do, and just leaving code as-is.
We don't generally keep dead code around as it clutters the reading experience.
This isn't even a TODO, and there's no official ticket to change it.
The task in particular was previously committed but quickly reverted as it didn't work as expected.
Thus, it is dead code.

engine/sound/src/sound.cpp
JCash: Sorry, I misread here.
But why did you change indentation?

engine/sound/src/sound_dsp.h
JCash: Perhaps a little longer. It is getting hard to read and follow.

engine/sound/src/devices/device_wasapi.cpp
```diff
@@ -343,7 +344,7 @@ namespace dmDeviceWasapi
                 for (int i = 0; i < frames_available; ++i)
                 {
                     float left = samples[i*channels_in+0]/32768.0f;
-                    float right = samples[i*channels_in+1]/32768.0f;
+                    float right = samplesi*channels_in+1]/32768.0f;
```
JCash, line 347: This doesn't compile, does it?

JCash (changes requested):
doesn't compile

thengel01:
K. Respectfully agree to disagree. I'll remove the questionable code ASAP (but might be a few days, as I am travelling)

engine/sound/src/devices/device_wasapi.cpp
thengel01: Oops! Could have sworn I tried Windows...

engine/sound/src/sound.cpp
thengel01: Restored

engine/sound/src/js/library_sound.js
thengel01: And now I removed it.

engine/sound/src/js/library_sound.js
thengel01: It's removed.

JCash (dismissed):

JCash (approved):

PR #10514, opened 2025-04-16
Removing redundant dmSound::Update() from sound component update

thengel01:
- sound components called directly into dmSound::Update() on each of their own update calls
- this was a noop for platforms with threading enabled, but did cause extra update calls on non-threaded setups


JCash (changes requested):
The unit tests are failing.

(For the Linux tests, we had an issue on defold/dev, but it's fixed now, so I suggest you update this branch with latest defold/dev.)

thengel01:
Updated from 'dev'. The only(!) change is now the single line being deleted. Hardly a trigger for fails - so the update should fix things.

JCash:
> Hardly a trigger for fails - so the update should fix things.

While that's our desire, it seems some tests were written in a way that relies on this issue (updating sound twice for each engine update).

JCash (changes requested):
unit tests are failing

thengel01:
Umm... those fails are something about collision data... not quite something these changes would trigger :-P

thengel01:
Updated from "dev"... in the hope that fixes things.

thengel01:
K. Now I saw the fail regarding to sound. Not sure why the Lua "final() call finds the sound_done still not set. Any pointers from your side what might trigger something like this?

PR #10529, opened 2025-04-17
Updating volume and panning handling for sound system

thengel01:
* Sound panning now configured to correctly pan stereo sound instances (basically treats them as 2 mono sources for panning).

  * *NOTE:* This fixes a (very old) bug where the mechanism removed ~3dB of the sounds. The fix will make the sound strength back to normal again. If this bug fix affects your game, you can lower the main `sound.gain` (default 1.0) in the `game.project` settings.

* Added logic that can support exponential gain (as in: 0.5 perceptually is "half volume"). This is useful for a nicer fade ins/outs.
  * The functionality is enabled by setting the `sound.use_linear_gain = 0` in `game.project`


engine/sound/src/sound.cpp
```diff
@@ -281,6 +282,24 @@ namespace dmSound
...
+    static float GainToScale(float gain)
+    {
+    #if defined(DM_SOUND_USE_LEGACY_GAIN) && (DM_SOUND_USE_LEGACY_GAIN != 0)
+        return gain;
+    #else
+        gain = dmMath::Clamp(gain, 0.0f, 1.0f);
+        // Convert "gain" to scale so progression over the range 'feels' linear
+        // (roughly 60dB(A) range assumed; rough approximation would be simply X^4 -- if this ever is too costly)
+        const float l = 0.1f;   // linear taper-off range
+        const float a = 1e-3f;
```
JCash, line 294: Feel free to use `0.001f` here for readability (especially as the other ones are in decimal form)

engine/sound/src/sound.cpp
```diff
@@ -281,6 +282,24 @@ namespace dmSound
         return RESULT_DEVICE_NOT_FOUND;
     }
 
+    static float GainToScale(float gain)
+    {
+    #if defined(DM_SOUND_USE_LEGACY_GAIN) && (DM_SOUND_USE_LEGACY_GAIN != 0)
+        return gain;
+    #else
```
JCash, line 289: We don't want to carry an extra compile time code path for this.
We either consider it a bug fix, or add a new script function.
Remember, we want the code, but also have to consider the old projects.
An extra compile time code path would then mean for us to ship _an extra build target_ for each platform+architecture.

engine/sound/src/test/test_sound.cpp
```diff
@@ -523,6 +527,23 @@ const TestParams params_looping_test[] = {
...
+    gain = dmMath::Clamp(gain, 0.0f, 1.0f);
+    const float l = 0.1f;   // linear taper-off range
+    const float a = 1e-3f;
+    const float b = 6.908f;
+    float scale = a * expf(gain * b);
+    if (gain < l)
+        scale *= gain * (1.0f / l);
+    return dmMath::Min(scale, 1.0f);
+#endif
+}
```
JCash, line 545: Not sure why this is duplicated here?
Isn't the point to test the _actual_ function from sound.cpp?
Testing duplicated code is error prone.

engine/sound/src/test/test_sound.cpp
```diff
@@ -881,7 +922,7 @@ TEST_P(dmSoundTestSpeedTest, Speed)
     ASSERT_EQ(dmSound::RESULT_OK, r);
     ASSERT_NE((dmSound::HSoundInstance) 0, instance);
 
-    r = dmSound::SetParameter(instance, dmSound::PARAMETER_GAIN, dmVMath::Vector4(0.5f,0,0,0));
+    r = dmSound::SetParameter(instance, dmSound::PARAMETER_GAIN, dmVMath::Vector4(0.8,0,0,0));
```
JCash, line 925: .5 to .8 does sound like a huge step?

If this is required, I wonder how silent the current games are when e.g. animating the gain from 1 to 0.
Will they become quiet at 0.7?

I'm ok with 1.0 regaining the sound level we used to have. The question is if they all have to tweak their gain levels now 🤔

JCash (changes requested):

JCash:
Also, why is the PR title suggesting _two_ changes?
I read the code as one change (panning fix), but perhaps I missed something.
If there are indeed two changes, you should split them into two separate PR's.

engine/sound/src/sound.cpp
thengel01: Well, so i take it then you decided on how to deal with possible incompatibility by now. Last time we chatted, I seemed to be the cautious one and you seemed very fine with changing things as it's "just a bug".

Any I can make this a runtime switch if so desired. Cost overhead will be minimal.

engine/sound/src/test/test_sound.cpp
thengel01: Yes,but it seemed not possible to easily access the code from within the test -- unless I park it in a header. Which I found  very inappropriate.

engine/sound/src/test/test_sound.cpp
thengel01: Yes, the volume change for existing mixing in existing games would be huge at times. This is why I didn't activate any of the volume related code paths, yet. I wanted to put the code in with the panning as it is - in a sense - the same "theme". Still the volume adjustments seem to me almost like a possible 2nd mode.

As we didn't get to any decision on how / if / when (me being the cautious one in the past ;-) to enable this, I wanted to at least not make it get lost, but have it on the ready once such a decision would be made.

The change cannot quite be made to be compatible. If we want to make this a runtime switchable thing - which it could be - it would likely be something games projects opt in on earlish in their development.

engine/sound/src/test/test_sound.cpp
thengel01: I did in the end add a new API to dmSound to enable access.

engine/sound/src/sound.cpp
thengel01: All is now configured dynamically at runtime. Defaults are still to "new panning == yes" and "new gain == no".
But I'd have no issue with switching both off by default if so desired.

PR #10544, opened 2025-04-21
Make sure the engine picks the best GPU

### LINKED ISSUE #5954: OpenGL graphics card selection issue
In a lot of cases, the OpenGL driver seems to select the wrong graphics card.

### LINKED ISSUE #10437: Engine doesn't pick the best GPU with GLFW 3
Picking dedicated GPU was implemented some time ago - https://github.com/defold/defold/issues/5185

But it looks like this no longer works after upgrading to GLFW 3, because the symbols to help select the dedicated GPU are no longer in the executable.

`dumpbin.exe /exports dmengine.exe`

1.9.1:
```
         15    E 001189F0 AmdPowerXpressRequestHighPerformance
         16    F 001189EC NvOptimusEnablement
```

1.9.8: no symbols like that

**To Reproduce (REQUIRED)**
1. Run the game on iGPU / dGPU PC
2. The engine picks iGPU

The engine should pick dGPU

  - Version: 1.9.2 and newer

**Platforms (REQUIRED):**
 - Platforms: Windows and other

AGulev:
GLFW for Windows is now built with `GLFW_USE_HYBRID_HPG` enabled to export the NvOptimusEnablement and AmdPowerXpressRequestHighPerformance symbols, which force the use of the high-performance GPU on Nvidia Optimus and AMD PowerXpress systems.

Fix https://github.com/defold/defold/issues/10437
Fix https://github.com/defold/defold/issues/5954


share/ext/glfw/build_glfw.sh
```diff
@@ -44,6 +44,7 @@ CMAKE_FLAGS="-DCMAKE_BUILD_TYPE=Release ${CMAKE_FLAGS}"
 CMAKE_FLAGS="-DGLFW_BUILD_EXAMPLES=OFF ${CMAKE_FLAGS}"
 CMAKE_FLAGS="-DGLFW_BUILD_TESTS=OFF ${CMAKE_FLAGS}"
 CMAKE_FLAGS="-DGLFW_BUILD_DOCS=OFF ${CMAKE_FLAGS}"
+CMAKE_FLAGS="-DGLFW_USE_HYBRID_HPG=ON ${CMAKE_FLAGS}"
```
AGulev, line 47: Actual fix

share/ext/glfw/build_glfw.sh
```diff
@@ -82,14 +83,23 @@ function cmi_unpack() {
...
 }
 
+function convert_line_endings() {
+    if [[ "${PLATFORM}" == *win* ]]; then
+        if [ -f "../patch_${VERSION}" ]; then
+            echo "Converting patch file ../patch_${VERSION} to Unix line endings..."
+            dos2unix ../patch_${VERSION}
+        fi
+    fi
+}
```
AGulev, line 93: Can't apply patch on windows without it

.github/workflows/build-glfw.yml
```diff
@@ -0,0 +1,149 @@
+name: Build GLFW
+
+on:
+  workflow_dispatch:
```
AGulev, line 4: a manual workflow that rebuilds glfw for all the platforms on CI
similiar to what we have for LuaJit

.github/workflows/build-glfw.yml
```diff
@@ -0,0 +1,149 @@
...
+          echo "Artifact not found. Listing contents of share/ext/glfw:"
+          ls -lh share/ext/glfw || echo "Directory missing"
+        fi
+    - name: Archive results
+      uses: actions/upload-artifact@v4
+      with:
+        name: glfw-3.4-${{ matrix.platform }}.tar.gz
+        path: share/ext/glfw/glfw-3.4-${{ matrix.platform }}.tar.gz
+        if-no-files-found: error
+        retention-days: 1
```
AGulev, line 112: CleanShot 2025-04-21 at 11 44 37@2x
all the results available in Summary for one day

share/ext/glfw/build_glfw.sh
JCash: I think we need this on Linux platforms as well, as they're likely to run on same/similar hardware?

JCash (changes requested):
Are we missing the Linux packages?

AGulev:
@JCash hmm, yes, it makes sense. For mac as well (exp for x86, but I'll update both)

AGulev (commented):
It's finally green!

JCash (approved):

PR #10550, opened 2025-04-22
Fix toolbar visibility

### LINKED ISSUE #10546: Scene View tool bar is hidden when the Outline View has input focus
The Scene View tool bar should remain visible when selecting objects using the Outline View, as well as editing selected object properties using the Properties View. Both of these views are considered subviews of the active editor tab.

**To Reproduce (REQUIRED)**
1. Open a scene view.
2. Click on an item in the Outline View.

The tool bar becomes hidden as soon as input focus is transferred to the Outline View.

The tool bar remains visible while the input focus is in the Outline View or the Properties View.

sprocketc:
Introduced here #10469

Fixes #10546

### DIFF
--- a/editor/src/clj/editor/ui.clj
+++ b/editor/src/clj/editor/ui.clj
@@ -1918,13 +1918,11 @@
         app-view (-> current-command-contexts first :env :app-view)
         active-tab (g/maybe-node-value app-view :active-tab evaluation-context)]
     (doseq [td (vals (user-data root ::toolbars))]
-      (let [control (:control td)
-            is-active (boolean (when active-tab
-                                 (nodes-along-path? control (.getContent ^Tab active-tab) root)))]
-        (visible! control is-active)
-        (when is-active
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/ui.clj
```diff
@@ -1918,13 +1918,11 @@
...
-      (let [control (:control td)
-            is-active (boolean (when active-tab
-                                 (nodes-along-path? control (.getContent ^Tab active-tab) root)))]
-        (visible! control is-active)
-        (when is-active
-          (refresh-toolbar td visible-command-contexts evaluation-context)
-          (refresh-toolbar-state (:control td) current-command-contexts evaluation-context))))))
+      (let [control (:control td)]
+        (when active-tab
+          (visible! control (nodes-along-path? control (.getContent ^Tab active-tab) root)))
```
sprocketc, line 1923: If there is no active tab, we simply skip the visibility toggling in order to maintain the current state.

vlaaad (approved):
LGTM!

PR #10556, opened 2025-04-23
Fix label drag issues

### LINKED ISSUE #10523: Change position results in a Java error
Change the Y position

Results in a JAVA error on Mac M1

Does not happen every time, but this time I copied an object, then changed the Y position on the new object

<hr/>

<table>
<tr><td>Build time</td><td>2025-04-14T17:27:31.739719</td></tr>
<tr><td>Defold editor sha</td><td>591eb496d52f4140bc2c7de547131f1b9408b9b4</td></tr>
<tr><td>Defold engine sha</td><td>591eb496d52f4140bc2c7de547131f1b9408b9b4</td></tr>
<tr><td>Defold version</td><td>1.10.0</td></tr>
<tr><td>Error</td><td><a href='https://sentry.io/organizations/defold/issues/?query=id%3A"299fbe1f64cc4894b23b7766de4eeabc"'>299fbe1f64cc4894b23b7766de4eeabc</a></td></tr>
<tr><td>GPU Driver</td><td>2.1 Metal - 89.4</td></tr>
<tr><td>OS version</td><td>15.4.1</td></tr>
<table>

sprocketc:
Handle various label drag issues.

Fixes #10523

### DIFF
--- a/editor/src/clj/editor/properties_view.clj
+++ b/editor/src/clj/editor/properties_view.clj
@@ -222,49 +222,59 @@
 (defn- handle-label-drag-event! [property-fn drag-update-fn update-ui-fn ^MouseDragEvent event]
   (.consume event)
   (let [target (.getTarget event)
-        property (property-fn)
-        edit-type (:edit-type property)
-        to-fn (:to-type edit-type identity)
-        from-fn (:from-type edit-type identity)
-        min-val (:min edit-type)
-        max-val (:max edit-type)
-        x (.getX event)
-        y (.getY event)
-        [prev-x prev-y] (ui/user-data target ::position)
-        delta-x (- x prev-x)
-        delta-y (- prev-y y)
-        max-delta (if (> (abs delta-x) (abs delta-y)) delta-x delta-y)
-        update-val (cond-> (or (:precision edit-type) 1.0)
-                     (.isShiftDown event) (* 10.0)
-                     (.isControlDown event) (* 0.1)
-                     (neg? max-delta) -)]
-    (when (> (abs max-delta) 1)
-      (let [op-seq (ui/user-data target ::op-seq)
-            values (or (ui/user-data target ::values)
-                       (properties/values property))
-            new-values (mapv (fn [value]
-                               (cond-> (drag-update-fn value update-val)
-                                 min-val (max min-val)
-                                 max-val (min max-val))) values)]
-        (properties/set-values! property values op-seq)
-        (ui/user-data! target ::position [x y])
-        (ui/user-data! target ::values new-values)
-        (update-ui-fn (mapv (comp to-fn from-fn) new-values)
-                      (properties/validation-message property)
-                      (properties/read-only? property))))))
+        position (ui/user-data target ::position)
+        op-seq (ui/user-data target ::op-seq)]
+    (when (and position op-seq)
+      (let [property (property-fn)
+            edit-type (:edit-type property)
+            to-fn (:to-type edit-type identity)
+            from-fn (:from-type edit-type identity)
+            min-val (:min edit-type)
+            max-val (:max edit-type)
+            x (.getX event)
+            y (.getY event)
+            [prev-x prev-y] (ui/user-data target ::position)
+            delta-x (- x prev-x)
+            delta-y (- prev-y y)
+            max-delta (if (> (abs delta-x) (abs delta-y)) delta-x delta-y)
+            update-val (cond-> (or (:precision edit-type) 1.0)
+                         (.isShiftDown event) (* 10.0)
+                         (.isControlDown event) (* 0.1)
+                         (neg? max-delta) -)]
+        (when (> (abs max-delta) 1)
+          (let [values (or (ui/user-data target ::values)
+                           (properties/values property))
+                new-values (mapv (fn [value]
+                                   (cond-> (drag-update-fn value update-val)
+                                     min-val (max min-val)
+                                     max-val (min max-val))) values)]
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/properties_view.clj
```diff
@@ -222,49 +222,59 @@
...
-                                 max-val (min max-val))) values)]
-        (properties/set-values! property values op-seq)
-        (ui/user-data! target ::position [x y])
-        (ui/user-data! target ::values new-values)
-        (update-ui-fn (mapv (comp to-fn from-fn) new-values)
-                      (properties/validation-message property)
-                      (properties/read-only? property))))))
+        position (ui/user-data target ::position)
+        op-seq (ui/user-data target ::op-seq)]
+    (when (and position op-seq)
```
sprocketc, line 227: This is the fix that is related to the mentioned issue. I couldn't reproduce it, so I don't know how the drag event is triggered without going through a `MOUSE_PRESSED` first, but I have experienced cases where the drag event is moved to a different target while dragging, so this is not surprising.

editor/src/clj/editor/properties_view.clj
```diff
@@ -222,49 +222,59 @@
...
     (ui/user-data! ::op-seq (gensym))
-    (ui/user-data! ::position [(.getX event) (.getY event)])))
+    (ui/user-data! ::position [(.getX event) (.getY event)])
+    (ui/user-data! ::values nil)))
+
+(defn handle-label-release-event!
+  [^MouseEvent event]
+  (doto (.getTarget event)
+    (ui/user-data! ::op-seq nil)
+    (ui/user-data! ::position nil)))
```
sprocketc, line 269: Clearing the data on mouse release ensures that the drag handler will be blocked, if those values where previously set.

editor/src/clj/editor/properties_view.clj
```diff
@@ -282,9 +292,9 @@
                 children (if (seq label-text)
                            (let [label (doto (Label. label-text)
                                          (.setMinWidth Region/USE_PREF_SIZE))]
-                             (ui/do-run-later
-                               #(when-not (properties/read-only? (property-fn))
-                                  (make-label-draggable! label (partial handle-label-drag-event! property-fn drag-update-fn update-ui-fn))))
+                             (ui/run-later
+                               (when-not (properties/read-only? (property-fn))
+                                 (make-label-draggable! label (partial handle-label-drag-event! property-fn drag-update-fn update-ui-fn))))
```
sprocketc, line 297: Unrelated to the issue. This has the same effect, but I think I should have used `ui/run-later` in the first place.

editor/src/clj/editor/properties_view.clj
```diff
@@ -222,49 +222,59 @@
...
+                          (properties/validation-message property)
+                          (properties/read-only? property))))))))
 
 (defn handle-label-press-event!
   [^MouseEvent event]
   (doto (.getTarget event)
     (ui/user-data! ::op-seq (gensym))
-    (ui/user-data! ::position [(.getX event) (.getY event)])))
+    (ui/user-data! ::position [(.getX event) (.getY event)])
+    (ui/user-data! ::values nil)))
```
sprocketc, line 263: Added this to always use the initial property value when we start dragging. You might experience a glitchy behavior on subsequent drag events of the same label without this.

editor/src/clj/editor/properties_view.clj
```diff
@@ -221,50 +221,61 @@
...
-        (ui/user-data! target ::values new-values)
-        (update-ui-fn (mapv (comp to-fn from-fn) new-values)
-                      (properties/validation-message property)
-                      (properties/read-only? property))))))
+  (let [property (property-fn)
+        target (.getTarget event)
+        position (ui/user-data target ::position)
+        op-seq (ui/user-data target ::op-seq)
+        is-read-only (properties/read-only? (property-fn))]
+    (when (and position op-seq (not is-read-only))
```
sprocketc, line 229: This is the fix that is related to the mentioned issue. We somehow end up here without a position. I couldn't reproduce this, so I don't know how the drag event is triggered without going through a `MOUSE_PRESSED` event first, but I have experienced cases where the drag event is moved to a different target while dragging, so this is not surprising.

The read-only check was also moved here, because this state is dynamic, and we used to check for it upon the initialization of the property control. Changing a property that alters the read-only state of other props would result in an inconsistent state.

editor/src/clj/editor/properties_view.clj
```diff
@@ -221,50 +221,61 @@
...
-        (ui/user-data! target ::values new-values)
-        (update-ui-fn (mapv (comp to-fn from-fn) new-values)
-                      (properties/validation-message property)
-                      (properties/read-only? property))))))
+  (let [property (property-fn)
+        target (.getTarget event)
+        position (ui/user-data target ::position)
+        op-seq (ui/user-data target ::op-seq)
+        is-read-only (properties/read-only? property)]
+    (when (and position op-seq (not is-read-only))
```
sprocketc, line 229: This is the fix that is related to the mentioned issue. We somehow end up here without a position. I couldn't reproduce this, so I don't know how the drag event is triggered without going through a `MOUSE_PRESSED` event first, but I have experienced cases where the drag event is moved to a different target while dragging, so this is not surprising.

The read-only check was also moved here, because its state is dynamic, and we used to check for it upon the initialization of the property control. Changing a property that alters the read-only state of other props, would result in an inconsistent state.

vlaaad (approved):
LGTM, though I also couldn't reproduce the issue 🤷‍♂️

PR #10561, opened 2025-04-23
Fix initial document-width calculation for tab-indented code files

matgis:
Fixed an issue where you could not scroll the code editor far enough horizontally to view the whole document in code files indented using tabs.

### DIFF
--- a/editor/src/clj/editor/code/view.clj
+++ b/editor/src/clj/editor/code/view.clj
@@ -994,12 +994,13 @@
       :text)
     display-string))
 
-(defn- update-document-width
-  [evaluation-context view-node]
-  (when-let [resource-node (g/maybe-node-value view-node :resource-node)]
+(defn- update-document-width [evaluation-context view-node]
+  ;; This is called in response to changing layout-related properties on the
+  ;; view-node. We only want to do something when we're connected to a resource
+  ;; node. If we're not, the lines output will return nil.
+  (when-some [lines (g/node-value view-node :lines evaluation-context)]
     (let [glyph-metrics (g/node-value view-node :glyph-metrics evaluation-context)
           tab-spaces (g/node-value view-node :tab-spaces evaluation-context)
-          lines (g/node-value resource-node :lines evaluation-context)
           tab-stops (data/tab-stops glyph-metrics tab-spaces)
           document-width (data/max-line-width glyph-metrics tab-stops lines)]
       (g/set-property view-node :document-width document-width))))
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/code/view.clj
```diff
@@ -3354,29 +3355,24 @@
...
+           [:lines :lines]
+           [:regions :regions]])))
+    (g/transact
+      (g/with-auto-evaluation-context evaluation-context
+        (update-document-width evaluation-context view-node)))
+    (when (and is-code-resource-type
+               (resource/file-resource? resource))
+      (let [lines (g/node-value view-node :lines)]
+        (lsp/open-view! lsp view-node resource lines)))
+    view-node))
```
matgis, line 3375: Mostly indentation, but the change is to update the `document-width` property of the `view-node` in a separate transaction after it has been connected to the `resource-node`. We need this for the `tab-spaces` output of the `view-node` to be valid. Previously we would evaluate the `tab-spaces` output on the `view-node` before the connection to the `resource-node` had been established, which would cause us to fall back to the default indentation type instead of basing it on the content of the file.

editor/src/clj/editor/code/view.clj
```diff
@@ -994,12 +994,13 @@
...
     display-string))
 
-(defn- update-document-width
-  [evaluation-context view-node]
-  (when-let [resource-node (g/maybe-node-value view-node :resource-node)]
+(defn- update-document-width [evaluation-context view-node]
+  ;; This is called in response to changing layout-related properties on the
+  ;; view-node. We only want to do something when we're connected to a resource
+  ;; node. If we're not, the lines output will return nil.
+  (when-some [lines (g/node-value view-node :lines evaluation-context)]
```
matgis, line 1001: The `resource-node` output will not be valid until the `app-view` has connected the `view-node` that we're constructing, so instead we rely on the fact that the `lines` input will not have anything connected until we have established the connection to the `resource-node`. This allows us to call this function from inside the `setup-view!` function.

editor/src/clj/editor/code/view.clj
```diff
@@ -3354,29 +3355,24 @@
...
+           [:indent-type :indent-type]
+           [:invalidated-rows :invalidated-rows]
+           [:lines :lines]
+           [:regions :regions]])))
+    (g/transact
+      (g/with-auto-evaluation-context evaluation-context
+        (update-document-width evaluation-context view-node)))
+    (when (and is-code-resource-type
+               (resource/file-resource? resource))
+      (let [lines (g/node-value view-node :lines)]
```
matgis, line 3373: The `lines` output is `:cached` and also evaluated inside `update-document-width`, so we're not doing more work after this change.

vlaaad (approved):
LGTM! Ship it!

PR #10570, opened 2025-04-25
Transform top selected nodes only

### LINKED ISSUE #9286: Don't move child nodes inside parents if both are selected
@astrochili suggest a good thing here: https://github.com/Insality/panthera/issues/5

Text:
> When selecting all nodes in the tree - it would be great to move everything only globally as the solid structure, otherwise moving child nodes (if they are also selected with parents) is done twice or more, depending on the hierarchy level.

https://github.com/user-attachments/assets/fbaef843-eb14-4629-b390-032acf9c0bc4

sprocketc:
Selected children of already selected parents should not be transformed.

Fixes #9286

### Technical changes
We now filter the `selected-renderables` to keep the top selected parents only. The related issue is about moving, but the problem also occurs when we scale or rotate elements. The `SelectionController` under `editor.scene-selection` seems to mitigate this when we select multiple elements on the scene, by selecting the top elements only, but that does not happen when we select elements on the outline. It also doesn't seem to work on gui scenes. I didn't investigate this, since we would have to find a way to handle this on a lower level because of the outline case.

### DIFF
--- a/editor/src/clj/editor/scene_tools.clj
+++ b/editor/src/clj/editor/scene_tools.clj
@@ -24,7 +24,9 @@
             [editor.gl.vertex :as vtx]
             [editor.math :as math]
             [editor.prefs :as prefs]
-            [editor.scene-picking :as scene-picking])
+            [editor.scene-picking :as scene-picking]
+            [util.coll :as coll]
+            [util.eduction :as e])
   (:import [com.jogamp.opengl GL GL2]
            [java.lang Math Runnable]
            [javax.vecmath AxisAngle4d Matrix3d Matrix4d Point3d Quat4d Tuple3d Vector3d]))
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/scene_tools.clj
```diff
@@ -500,16 +500,27 @@
       (apply-fn start-pos pos))))
 
 (def ^:private original-values #(select-keys % [:node-id :world-rotation :world-transform :parent-world-transform]))
+ 
+(defn- top-nodes
+  [nodes]
+  (let [node-ids (->> nodes (map :node-id) set)
```
vlaaad, line 506: Please use transducers (i.e. `(into #{} (map :node-id) nodes)`) or our `util.eduction` ns (i.e. `(->> nodes (e/map :node-id) set)`)

editor/src/clj/editor/scene_tools.clj
```diff
@@ -500,16 +500,27 @@
...
+        child? (fn [node]
+                 (->> (:node-id-path node)
+                      (some #(and (not= % (:node-id node))
```
vlaaad, line 509: Please use `coll/some` instead of `some` so there is less GC pressure.

editor/src/clj/editor/scene_tools.clj
```diff
@@ -500,16 +500,27 @@
...
+                                  (contains? node-ids %)))))]
+    (filter (complement child?) nodes)))
```
vlaaad, line 511: Please use `e/filter` here... Or maybe `(e/remove child? nodes)`

vlaaad (dismissed):
Works great! My only worry is the potential performance impact of using a lot of lazy sequences on click when there are many selected nodes — we could use transducers/eductions here instead.

vlaaad (approved):
Ship it!

PR #10581, opened 2025-04-28
Hide reset button if some of the props are not overridable

### LINKED ISSUE #10551: Using the reset button of a property when multiple nodes are selected may throw an error
### Steps to reproduce
- Select an outline element with an overridable property
- Modify the value to enable the reset button
- Add another element to the current selection that has the same property, but it is not overridable
- Click on the reset button that is still visible

<hr/>

<table>
<tr><td>Build time</td><td></td></tr>
<tr><td>Defold channel</td><td></td></tr>
<tr><td>Defold editor sha</td><td>fa9b09bd744ea40653b5ba98a6dab1f8e93f0cb8</td></tr>
<tr><td>Defold engine sha</td><td>f0ae29af45e8fab92e4c45b409041174cfb68235</td></tr>
<tr><td>Defold version</td><td>dev</td></tr>
<tr><td>Error</td><td><a href='https://sentry.io/organizations/defold/issues/?query=id%3A"243c26eb00404228a9f6aaf2ba482806"'>243c26eb00404228a9f6aaf2ba482806</a></td></tr>
<tr><td>GPU</td><td>AMD Radeon RX 6700 (radeonsi, navi22, LLVM 19.1.7, DRM 3.61, 6.14.2-arch1-1)</td></tr>
<tr><td>GPU Driver</td><td>4.6 (Compatibility Profile) Mesa 25.0.3-arch1.1</td></tr>
<tr><td>Java version</td><td>21.0.7+6</td></tr>
<tr><td>OS name</td><td>Linux</td></tr>
<tr><td>OS version</td><td>6.14.2-arch1-1</td></tr>
<table>

sprocketc:
Check if all props of on a multi-selection are overridable in order to display the reset button.

Fixes #10551

### DIFF
--- a/editor/src/clj/editor/properties.clj
+++ b/editor/src/clj/editor/properties.clj
@@ -626,7 +626,9 @@
       v0)))
 
 (defn overridden? [property]
-  (and (contains? property :original-values) (not-every? nil? (:values property))))
+  (and (contains? property :original-values)
+       (every? some? (:original-values property))
+       (not-every? nil? (:values property))))
 
 (defn error-aggregate [vals]
   (when-let [errors (seq (remove nil? (distinct (filter g/error? vals))))]


editor/src/clj/editor/properties.clj
```diff
@@ -626,7 +626,9 @@
       v0)))
 
 (defn overridden? [property]
-  (and (contains? property :original-values) (not-every? nil? (:values property))))
+  (and (contains? property :original-values)
+       (every? some? (:original-values property))
+       (not-every? nil? (:values property))))
```
sprocketc, line 631: This would now return false if some of the properties are not overridable. That shouldn't break anything as far as I understand, since the only place we use this pred is on `preperties_view`, to decide if the reset button is going to be displayed.

We could also modify `clear-override!` to always show the button if there is an overridden prop, and filter out the rest to avoid the exception.

vlaaad: Makes sense!

vlaaad (approved):
LGTM!

PR #10587, opened 2025-04-29
Allow getting/setting more properties using editor scripts

### LINKED ISSUE #5371: Access to all Outline properties in Editor Script
The "New editor script feature just got released: Outline Properties" is not working as intended.

I tried to get a Box Node's pivot but got
`ERROR:EXT: /editor-script-align/editor-script-align.editor_script:106 BoxNode has no "pivot" property`
This got me to try all fo them and these are not available:
`size_mode, texture, slice_9, color, alpha, layer, blend_mode, pivot, x_anchor, y_anchor, adjust_mode, clipping_mode`

These are the ones you can get on a Box Node:
`position, rotation, scale, size, inherit_alpha, clipping_visible, clipping_inverted`

vlaaad:
Previously, some properties couldn't be get/set by editor scripts. Now, it's possible to do so! We added support for getting and setting:
- Color properties
- Multiline text properties
- Choice box properties 
- Slider properties

Remaining unsettable properties are curve and curve spread (in particle fx).

Fixes #5371


editor/src/clj/editor/properties.clj
```diff
@@ -362,7 +362,8 @@
       {:type (:type property)}))
 
 (defn edit-type-id [property]
-  (let [t (:type (property-edit-type property))]
+  (let [t (:type (property-edit-type property))
+        t (or (g/value-type-dispatch-value t) t)]
```
vlaaad, line 366: Now we use the same logic as in `properties-view` for dispatching on edit type.

matgis (approved):
Looks great! Bonus points for adding a test! ❤️

vlaaad:
😊

PR #10591, opened 2025-04-30
Fix delete shortcut on code view

sprocketc:
Reported here https://forum.defold.com/t/defold-1-10-1-beta/80375/7
> The “Del” key stopped deleting text when editing code

Can be reproduced on Windows and Linux.

### Technical changes
The conflict with `:edit.remove` seems to be the issue here. You might have to restart the editor to verify this after removing the `delete` shortcut from `Edit -> Remove`. I couldn't find a better way to handle this promptly, so I reverted the name change that was introduced by #10512. Also related to the discussion here https://github.com/defold/defold/pull/10512#discussion_r2054172726


sprocketc:
Closing this in favor of #10593

PR #10593, opened 2025-04-30
Fix shortcuts not triggering when conflicting shortcuts are disabled in the menu bar

vlaaad:


editor/src/clj/editor/keymap.clj
```diff
@@ -700,14 +701,20 @@
 (defn install!
   "Install a keymap on a scene, replacing any predefined accelerators"
   [keymap ^Scene scene execute-fn]
-  (doto (.getAccelerators scene)
```
vlaaad, line 703: The problem was that scene accelerators are mutated both by us here and by JavaFX on Windows and Linux when using a menubar with accelerators. Using `KEY_PRESSED` handler instead helps.

sprocketc (approved):
Works great now!

editor/src/clj/editor/keymap.clj
```diff
@@ -700,14 +701,20 @@
...
+  (let [{:keys [shortcut->commands]} keymap
+        new-handler (reify EventHandler
+                      (handle [_ e]
+                        (reduce-kv
+                          (fn [_ ^KeyCombination shortcut commands]
+                            (when (.match shortcut e)
+                              (.consume e)
+                              (reduced (execute-fn commands))))
+                          nil
+                          shortcut->commands)))]
```
vlaaad, line 715: Fun fact: this is semantically the same code that JavaFX uses internally when invoking a scene accelerator. Yes, accelerators are a map, but JavaFX also iterates over every entry in an attempt to `.match` a key combination with a handler:
https://github.com/openjdk/jfx/blob/3fdd21386d6db96294fcecd80afc25d09732c067/modules/javafx.graphics/src/main/java/com/sun/javafx/scene/KeyboardShortcutsHandler.java#L372-L387

PR #10619, opened 2025-05-07
Make it possible to set GUI resources using `gui.set()`

### LINKED ISSUE #10471: Add a gui function to reference a new atlas
_Is it a private platform specific feature? (i.e. would it break your NDA?)_
NO

Defold proposes a complicated way цhen you want to use dynamic atlases (created by [resource.create_atlas](https://defold.com/ref/resource/#resource.create_atlas) in a gui under different collection proxies / factories :

1. load and create a texture, load and create an atlas with that texture
2. you need to create a script for attaching this atlas to gui. This script attaches that atlas to gui component by `go.set()` (Example: `go.set("#gui", "textures", my_atlas, {key = "my_atlas"})`). You have to set the script to the same go which contains gui component (or exact to any go but inside of the same collection. Because of proxies)
3. Call the script to attach atlases before showing your gui and wait for an answer. It produces a synchronization between different parts of code.

The best solution is to have an API in `gui` to add an atlas into `textures` property. It can work with already loaded atlases.

See the problem section

AGulev:
It is now possible to set Font or Texture resources for a GUI using the `gui.set()` function from the GUI component itself, provided that `msg.url()` is used as the first argument instead of a node.

```lua
function on_message(self, message_id, message, sender)
   if message_id == hash("set_font") then
       gui.set(msg.url(), "fonts", message.font, {key = "my_font_name"})
       gui.set_font(gui.get_node("text"), "my_font_name")
   elseif message_id == hash("set_texture") then
       gui.set(msg.url(), "textures", message.texture, {key = "my_texture"})
       gui.set_texture(gui.get_node("box"), "my_texture")
       gui.play_flipbook(gui.get_node("box"), "logo_256")
   end
end
```

Fix https://github.com/defold/defold/issues/10471

### DIFF
--- a/engine/gameobject/src/gameobject/gameobject_script.cpp
+++ b/engine/gameobject/src/gameobject/gameobject_script.cpp
@@ -168,17 +168,6 @@ namespace dmGameObject
         "on_reload"
     };
 
-    static const char* TYPE_NAMES[PROPERTY_TYPE_COUNT] = {
-        "number",        // PROPERTY_TYPE_NUMBER
-        "hash",          // PROPERTY_TYPE_HASH
-        "msg.url",       // PROPERTY_TYPE_URL
-        "vmath.vector3", // PROPERTY_TYPE_VECTOR3
-        "vmath.vector4", // PROPERTY_TYPE_VECTOR4
-        "vmath.quat",    // PROPERTY_TYPE_QUAT
-        "boolean",       // PROPERTY_TYPE_BOOLEAN
-        "vmath.matrix4", // PROPERTY_TYPE_MATRIX4
-    };
-
     HRegister g_Register = 0;
 
     CompScriptWorld::CompScriptWorld(uint32_t max_instance_count)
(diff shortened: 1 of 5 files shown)


engine/gameobject/src/gameobject/gameobject_script.cpp
```diff
@@ -168,17 +168,6 @@ namespace dmGameObject
...
-        "number",        // PROPERTY_TYPE_NUMBER
-        "hash",          // PROPERTY_TYPE_HASH
-        "msg.url",       // PROPERTY_TYPE_URL
-        "vmath.vector3", // PROPERTY_TYPE_VECTOR3
-        "vmath.vector4", // PROPERTY_TYPE_VECTOR4
-        "vmath.quat",    // PROPERTY_TYPE_QUAT
-        "boolean",       // PROPERTY_TYPE_BOOLEAN
-        "vmath.matrix4", // PROPERTY_TYPE_MATRIX4
-    };
-
```
AGulev, line 181: moved to `gameobject_script_util.h`

engine/gameobject/src/gameobject/gameobject_script.cpp
```diff
@@ -677,70 +666,6 @@ namespace dmGameObject
         return CheckGetPropertyResult(L, "go", result, property_desc, property_id, target, property_options, index_requested);
     }
 
-    static int HandleGoSetResult(lua_State* L, dmGameObject::PropertyResult result, dmhash_t property_id, dmGameObject::HInstance target_instance, const dmMessage::URL& target, const dmGameObject::PropertyOptions& property_options)
```
AGulev, line 680: moved to `gameobject_script_util.cpp` to be able to use in gui

engine/gui/src/gui_script.cpp
```diff
@@ -855,19 +855,73 @@ namespace dmGui
...
+            dmMessage::URL target;
+            dmScript::ResolveURL(L, 1, &target, &sender);
+            bool is_self = (sender.m_Socket == target.m_Socket) &&
+                           (sender.m_Path == target.m_Path) &&
+                           (sender.m_Fragment == target.m_Fragment);
+            if (!is_self)
+            {
+                return DM_LUA_ERROR("'gui.set()' can only be used to change a property of the GUI component itself, use 'msg.url()'");
+            }
+            dmGameObject::HInstance instance = (dmGameObject::HInstance)scene->m_Context->m_GetUserDataCallback(scene);
```
AGulev, line 914: Callback used to get instance, so no more dependency on gamesys

engine/gui/src/gui_script.cpp
```diff
@@ -855,19 +855,73 @@ namespace dmGui
...
+
+        dmGameObject::PropertyOptions property_options = {};
+        if (lua_gettop(L) > 3)
+        {
+            int options_result = LuaToPropertyOptions(L, 4, &property_options, property_hash, 0);
+            if (options_result != 0)
+            {
+                return options_result;
+            }
+        }
```
AGulev, line 899: the lines above just moved from the function bottom to be able to reuse this code

engine/gui/src/gui_script.cpp
```diff
@@ -855,19 +855,73 @@ namespace dmGui
...
+
+        if (dmScript::IsURL(L, 1))
+        {
+            dmMessage::URL sender;
+            dmScript::GetURL(L, &sender);
+            dmMessage::URL target;
+            dmScript::ResolveURL(L, 1, &target, &sender);
+            bool is_self = (sender.m_Socket == target.m_Socket) &&
```
AGulev, line 907: Make it possible to set properties only from within the GUI component itself.

JCash (approved):
lgtm

PR #10626, opened 2025-05-08
Faster atlas creation

AGulev:
Atlas creation is now twice as fast in `bob.jar` and about three times faster in the editor.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/MaxRectsLayoutStrategy.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/MaxRectsLayoutStrategy.java
@@ -584,12 +584,12 @@ public class MaxRectsLayoutStrategy implements TextureSetLayoutStrategy {
         }
 
         // / Returns 0 if the two intervals i1 and i2 are disjoint, or the length of their overlap otherwise.
-        private int commonIntervalLength (int i1start, int i1end, int i2start, int i2end) {
+        private int commonIntervalLength(int i1start, int i1end, int i2start, int i2end) {
             if (i1end < i2start || i2end < i1start) return 0;
             return Math.min(i1end, i2end) - Math.max(i1start, i2start);
         }
 
-        private int contactPointScoreNode (int x, int y, int width, int height) {
+        private int contactPointScoreNode(int x, int y, int width, int height) {
             int score = 0;
 
             if (x == 0 || x + width == binWidth) score += height;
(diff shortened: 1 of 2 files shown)


AGulev:
Left side is 1.10.0, right side is this branch:
https://github.com/user-attachments/assets/a1066b46-57bc-46a9-847e-8d03dc3f4e36

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/MaxRectsLayoutStrategy.java
```diff
@@ -607,34 +607,57 @@ private int contactPointScoreNode (int x, int y, int width, int height) {
             return score;
         }
 
-        private RectNode findPositionForNewNodeContactPoint (int width, int height, int rotatedWidth, int rotatedHeight, boolean rotate) {
+        private RectNode evaluateContactPoint(int width, int height, int rotatedWidth, int rotatedHeight, boolean rotate, Rect currRect) {
```
AGulev, line 610: This function does the same thing, but only for one item; the loop is moved outside to allow parallelization.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/MaxRectsLayoutStrategy.java
```diff
@@ -607,34 +607,57 @@ private int contactPointScoreNode (int x, int y, int width, int height) {
...
+
             return bestNode;
         }
 
+        private RectNode findPositionForNewNodeContactPoint(int width, int height, int rotatedWidth, int rotatedHeight, boolean rotate) {
+            RectNode globalBestNode = new RectNode();
+            globalBestNode.rect = new Rect(null, 0,0,0,0,0);
+            globalBestNode.score1 = Integer.MIN_VALUE; // maximize contact points
+
+            if (freeRectangles.size() < 100) {
```
AGulev, line 642: Not sure if it's a good idea, But I think there is no sense to run it in parallel for a small set of images. On the other hand, it is extra complexity in code... What do you think?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/MaxRectsLayoutStrategy.java
```diff
@@ -607,34 +607,57 @@ private int contactPointScoreNode (int x, int y, int width, int height) {
...
+                    if (candidate.score1 > globalBestNode.score1) {
+                        globalBestNode = candidate;
+                    }
+                }
+            } else {
+                // Big set: parallel scan
+                globalBestNode = freeRectangles.parallelStream()
+                    .map(currentNode -> evaluateContactPoint(width, height, rotatedWidth, rotatedHeight, rotate, currentNode.rect))
+                    .reduce((best1, best2) -> (best1.score1 > best2.score1) ? best1 : best2)
+                    .orElse(globalBestNode);
```
AGulev, line 655: looks very functional, but this is the simplest way to make it in parallel

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TextureUtil.java
```diff
@@ -400,16 +401,28 @@ public static TextureGenerator.GenerateResult createMultiPageTexture(List<Buffer
...
     // Public api
     public static List<BufferedImage> loadImages(List<IResource> resources) throws IOException, CompileExceptionError {
-        List<BufferedImage> images = new ArrayList<BufferedImage>(resources.size());
-
-        for (IResource resource : resources) {
-            BufferedImage image = ImageIO.read(new ByteArrayInputStream(resource.getContent()));
-            if (image == null) {
-                throw new CompileExceptionError(resource, -1, "Unable to load image " + resource.getPath());
+        try {
+            return resources.parallelStream()
```
AGulev, line 405: There is no sense to read files one by one, parallel reading should always be faster or at least not slower in case of IO bottleneck

It's only for bob

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/MaxRectsLayoutStrategy.java
britzl: Is there a drawback to running it in parallel also for a small set of images? If there isn't then I suggest doing it also for a small set.

britzl (approved):
Looks good, but question if we shouldn't always run in parallel regardless of set size?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/MaxRectsLayoutStrategy.java
britzl: Sure, but surely not that much time? Packing an atlas with 1-100 images should be instant even with the overhead. But lets go ahead and merge. It's not a big deal!

PR #10627, opened 2025-05-08
Added core engine support for runtime generation of SDF fonts

JCash:
In this release we've added support for runtime generation of signed distance field fonts.
The main benefits are smaller bundle size, and runtime memory footprint. The drawback is very slightly longer load times.

It is currently an opt-in, and you can enable it by setting the project variable `font.runtime_generation`.
Our goal is to make this our default way of generating fonts, so we'd like our users to please test the this feature out and report any issues.
Note: This feature is currently only available when using Bob.jar. So either use bob.jar directly or bundle your game.

For the visual before/after comparisons, both the old and new distance field generators have artefacts, and we don't see any real difference in improvement or degradation. They're just different, which is expected for a single channel SDF implementation.

The font glyph can prewarm the glyph cache if you've deselected "All Characters" on your font, and specified a small subset of characters.

At runtime, you can also add/remove glyphs to the font dynamically by calling `font.add_glyphs()`/`font.remove_glyphs()`

You can read more about usage of runtime fonts in the [font documentation](https://defold.com/manuals/font/).

### Technical details

[Design document](https://docs.google.com/document/d/1hvAmz7JW-9GQZMXhQIpseLf2DURa0rtdS15Y5E1r23A/edit?tab=t.0#heading=h.s14yhi5eu2yf)

Results:
image

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


apply_license.py
```diff
@@ -69,6 +69,7 @@ def license(comment):
     "engine/ddf/src/ddfc.py",
     "engine/engine/contents/builtins/edn.lua",
     "engine/engine/contents/builtins/mobdebug.lua",
+    "engine/font/src/stb_truetype.h",
```
JCash, line 72: don't add defold license to this file

build_tools/waf_dynamo.py
```diff
@@ -626,7 +626,7 @@ def default_flags(self):
         self.env.append_value('DEFINES', ['DM_TEST_DLIB_HTTPCLIENT_NO_HOST_SERVER'])
 
         for f in ['CFLAGS', 'CXXFLAGS']:
-            self.env.append_value(f, ['-Wall', '-fPIC', '-fno-exceptions', '-fno-rtti',
+            self.env.append_value(f, ['-Wall', '-fPIC', '-fno-exceptions', '-fno-rtti', '-Wno-nontrivial-memcall',
```
JCash, line 629: we've started getting annoying compile warnings from Emscripten about `memset(this, ...)`

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/meta.properties
```diff
@@ -771,3 +771,12 @@ fixed_update_frequency.default = 60
...
 max_time_step.default = 0.5
+
+[fontgen]
+sdf_base_padding.type = integer
+sdf_base_padding.help = The base padding when generating sdf glyphs. [0-255]
+sdf_base_padding.default = 3
+
+sdf_edge_value.type = integer
+sdf_edge_value.help = The on edge when generating sdf glyphs. [0-255]
+sdf_edge_value.default = 190
```
JCash, line 782: 1) Should it be renamed `[font]` ?
2) If we want to support other generators in the long run, should these variables be prefixed `stbtt_` to signify that they belong to the STB TrueType settings?

engine/dlib/src/dlib/job_thread.cpp
```diff
@@ -115,16 +114,22 @@ static void JobThread(void* _ctx)
...
+        uint64_t tstart = dmTime::GetTime();
+        JobItem item = ctx->m_Work.Pop();
+
+        item.m_Result = item.m_Process(item.m_Context, item.m_Data);
+        PutDone(ctx, &item);
 
-    item.m_Result = item.m_Process(item.m_Context, item.m_Data);
-    PutDone(ctx, &item);
+        uint64_t tend = dmTime::GetTime();
+        if ((tend-tstart) > max_time)
```
JCash, line 128: If max_time == 0, then it still behaves as before: Process one job per update.

engine/engine/src/engine.cpp
```diff
@@ -158,6 +158,9 @@ namespace dmEngine
             ExtensionParamsSetContext(&m_Params, "render", engine->m_RenderContext);
             if (engine->m_HttpCache)
                 ExtensionParamsSetContext(&m_Params, "http_cache", engine->m_HttpCache);
+
+            if (engine->m_JobThreadContext)
+                ExtensionParamsSetContext(&m_Params, "job_thread", engine->m_JobThreadContext);
```
JCash, line 163: Allow getting the same job thread context from within an extension.

engine/engine/src/engine.cpp
```diff
@@ -1798,8 +1801,14 @@ namespace dmEngine
...
                     }
                 }
-
-                dmJobThread::Update(engine->m_JobThreadContext);
+#if defined(DM_HAS_THREADS)
+                uint64_t jobthread_max_time_us = 0;
+#else
+                // Dictates max time the job thread may take.
+                // Note that it will always process at least one item if available
+                uint64_t jobthread_max_time_us = 3 * 1000;
```
JCash, line 1809: May well be configurable later. Also, should it be shared between the different job thread contexts?

Side note:
We currently create multiple job thread contexts, but each context may have 0, 1 or more threads.
So, in terms of keeping the threads contained, it probably makes sense to use 1 job thread context for all our internal systems.

engine/engine/src/wscript
```diff
@@ -94,6 +95,12 @@ def build(bld):
 
     sound_lib = sound_lib.split()
 
+    if 'font' in waflib.Options.options.enable_features:
+        Logs.info("font library enabled")
+        additional_libs.append('FONT')
+        exported_symbols.append('ResourceTypeTTFFont')
+        exported_symbols.append('ScriptFontExt')
```
JCash, line 102: debug locally with `--enable-feature=font`

engine/font/src/fontgen.cpp
```diff
@@ -0,0 +1,564 @@
+// Copyright 2020-2025 The Defold Foundation
```
JCash, line 1: Copied mostly as-is from extension-fontgen

engine/resource/src/resource.cpp
```diff
@@ -460,7 +460,7 @@ void UpdateFactory(HFactory factory)
 {
     DM_PROFILE(__FUNCTION__);
     dmMessage::Dispatch(factory->m_Socket, &Dispatch, factory);
-    dmJobThread::Update(factory->m_JobThreadContext);
+    dmJobThread::Update(factory->m_JobThreadContext, 0);
```
JCash, line 463: Should probably test loading times on HTML5 as well, as 0 means "1 job per frame", but if we have budget for it, we can process many more items per frame.

engine/dlib/src/dlib/job_thread.cpp
ekharkunov: I think it's worth to accumulate all time spent on job process and compare accumulated time to `max_time`. Because it can be lots of small jobs which doesn't exceed the `max_time` but in sum it takes much more time than `max_time`. For example,
* 10 jobs. Each takes 0.05s to process
* max_time == 0.1s
* total time is 0.05 * 10 = 0.5s which is much more than 0.1s

engine/engine/src/engine.cpp
ekharkunov: My opinion is every job context should have their own `max_time` setting. And also I think that it's good to have multiple job_contexts because different subsystems can have different type of jobs. I mean one subsystem can have lots of small jobs, other - small amount of long-running jobs. I think if we have multiple job context it helps to tweak each job context according to job types requirements.

engine/font/src/wscript
```diff
@@ -0,0 +1,23 @@
+#! /usr/bin/env python
+import os
+from waf_dynamo import dmsdk_add_files
```
ekharkunov, line 3: Three unused imports

engine/gamesys/src/gamesys/resources/res_font.cpp
```diff
@@ -184,6 +184,9 @@ namespace dmGameSystem
 
     static void GetDynamicGlyphMetric(dmRender::FontMetrics* metrics, const uint32_t* key, DynamicGlyph** pglyph)
     {
+        if ((*key) == 0x0A) // new line doesn't have a size
```
ekharkunov, line 187: Maybe `return caret` should be added here?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/meta.properties
ekharkunov: I suggest to have naming like `[stbtt_font]`. It helps avoid collisions with future implementations and helps use the same keys for parameters across different implementations (sections).

build_tools/waf_dynamo.py
ekharkunov: Maybe it's worth to create a separate ticket and accumulate all kinds of warning which compiler reports and then fix it?

ekharkunov (changes requested):
Overall good. Small comments about job context max_time and a couple of other suggestions

build_tools/waf_dynamo.py
JCash: In this case, I've already checked in this flag, as it was overly spammy.
Fixing it makes little sense as it's merely convertiing `memset(ptr, ..)` to `memset((void*)ptr, ...)`. It doesn't really fix anything.

engine/dlib/src/dlib/job_thread.cpp
JCash: Hmm, yeah that was the intent, but I wonder how I got it turned around in my head, as that's not how it's implemented?!

Thanks for the heads up!

engine/engine/src/engine.cpp
JCash: Yeah, I think overall, it generally needs more work if we are to mix different work types on the same thread.
E.g. having different priorities on jobs, max time per type etc.
Scheduling is a tricky thing.
So far, we've been reacting to the tasks, but I feel we soon need to model this system after some exsiting job system out in the wild. TBD.

engine/dlib/src/dlib/job_thread.cpp
JCash: Yeah, I didn't notice it because the workload I tested with was way faster than the limit I set :)

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/font/Fontc.java
```diff
@@ -1038,24 +1059,49 @@ public BufferedImage generatePreviewImage() throws IOException {
...
         try {
             System.setProperty("java.awt.headless", "true");
-            if (args.length != 2 && args.length != 3 && args.length != 4)    {
-                System.err.println("Usage: fontc fontfile [basedir] outfile");
-                System.exit(1);
+
+            File    fontInput   = null;
+            String  outfile     = null;
+            String  basedir     = ".";
+            boolean dynamic     = false;
```
JCash, line 1072: This main() function is used for building test content (and also builtins)

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/font/Fontc.java
```diff
@@ -483,24 +519,9 @@ public BufferedImage generateGlyphData(boolean preview, final FontResourceResolv
...
-            {
-                shadow_edge = 1.0f;
-            }
-
-            glyphBankBuilder.setSdfSpread(sdf_spread);
-            glyphBankBuilder.setSdfOutline(outline_edge);
-            glyphBankBuilder.setSdfShadow(shadow_edge);
+            glyphBankBuilder.setSdfSpread(GetFontMapSdfSpread(fontDesc));
+            glyphBankBuilder.setSdfOutline(GetFontMapSdfOutline(fontDesc));
+            glyphBankBuilder.setSdfShadow(GetFontMapSdfShadow(fontDesc));
```
JCash, line 521: Strictly not necessary, but it helped make separate functions here, and also recreating them at runtime.

scripts/unpack_ddf.py
```diff
@@ -52,32 +52,34 @@
 import gamesys.sprite_ddf_pb2
 import gamesys.physics_ddf_pb2
 import gamesys.gui_ddf_pb2
+import gamesys.label_ddf_pb2
 
 BUILDERS = {}
-BUILDERS['.gamepadsc']      = input.input_ddf_pb2.GamepadMaps
```
JCash, line 57: Sorted alphabetically, for readability

scripts/unpack_ddf.py
```diff
@@ -124,7 +126,7 @@ def print_bytes(descriptor, data):
     print(descriptor.name, ":", str.format('[%s]' % binascii.hexlify(data).decode('utf8')), "len:", len(data) )
 
 def print_string(descriptor, data):
-    print(descriptor.name, ":", str.format('"%s"' % data), "len:", len(data))
+    print(descriptor.name, ":", '"' + data.replace('"', '\\"') + '"', "len:", len(data))
```
JCash, line 129: Fix for quotes in the actual payload string

engine/render/proto/render/font_ddf.proto
```diff
@@ -57,60 +57,74 @@ message GlyphBank
...
+    optional float             max_ascent            = 5 [default = 0];
+    optional float             max_descent           = 6 [default = 0];
+    optional float             max_advance           = 7 [default = 0];
+    optional float             max_width             = 8 [default = 0];
+    optional float             max_height            = 9 [default = 0];
     optional FontTextureFormat image_format          = 10 [default = TYPE_BITMAP];
 
+    // TODO: Move to fontDesc
     optional float             sdf_spread            = 11 [default = 1];
-    optional float             sdf_offset            = 12 [default = 0];
```
JCash, line 81: Removed unused `sdf_offset`

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/FontBuilder.java
```diff
@@ -26,44 +26,88 @@
...
+            // input(2)
+            subTask = createSubTask(fontResource, CopyBuilders.TTFBuilder.class, taskBuilder);
+        }
+        else
+        {
+            // input(2)
+            taskBuilder.addInput(fontResource);
+            // input(3)
+            subTask = createSubTask(input, GlyphBankBuilder.class, taskBuilder);
+        }
```
JCash, line 72: Only output .glyphbankc if it's an offline font.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/FontBuilder.java
```diff
@@ -75,6 +119,18 @@ public void build(Task task) throws CompileExceptionError, IOException {
         fontMapBuilder.setOutlineAlpha(fontDesc.getOutlineAlpha());
         fontMapBuilder.setOutlineWidth(fontDesc.getOutlineWidth());
         fontMapBuilder.setLayerMask(Fontc.GetFontMapLayerMask(fontDesc));
+        fontMapBuilder.setCacheWidth(fontDesc.getCacheWidth());
```
JCash, line 122: For runtime fonts, these are calculated at runtime.

engine/engine/src/engine.cpp
```diff
@@ -1417,6 +1420,7 @@ namespace dmEngine
             engine->m_ResourceTypeContexts.Put(dmHashString64("gui_scriptc"), engine->m_GuiScriptContext);
             engine->m_ResourceTypeContexts.Put(dmHashString64("guic"), engine->m_GuiContext);
         }
+        engine->m_ResourceTypeContexts.Put(dmHashString64("fontc"), engine->m_RenderContext);
```
JCash, line 1423: We've made .fontc into a true extension

engine/render/src/render/font_renderer_default.cpp
```diff
@@ -231,10 +231,9 @@ namespace dmRender
         //     back to front layer in the order of shadow -> outline -> face, where the offset of each
         //     layer depends on how many glyphs we actually can place in the buffer. To get a valid count, we
         //     do a dry run first over the input string and place glyphs in the cache if they are renderable.
-        if (HAS_LAYER(layer_mask,OUTLINE) || HAS_LAYER(layer_mask,SHADOW))
+        layer_count += HAS_LAYER(layer_mask,OUTLINE) + HAS_LAYER(layer_mask,SHADOW);
+        if (layer_count > 1)
```
JCash, line 235: Small optimization, skipping if we don't need it.

engine/resource/src/resource.h
```diff
@@ -436,6 +437,9 @@ namespace dmResource
     typedef int (*FPreloadDataCallback)(HFactory factory, void* cbk_ctx, HResourceDescriptor resource, uint32_t offset, uint32_t nread, uint8_t* buffer);
 
     Result PreloadData(HFactory factory, const char* path, uint32_t offset, uint32_t size, FPreloadDataCallback cbk, void* cbk_ctx);
+
+    // Get the assigned Job thread
+    dmJobThread::HContext GetJobThread(const dmResource::HFactory factory);
```
JCash, line 442: For adding resource related jobs

engine/gamesys/src/gamesys/scripts/script_font.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+dmResource::HFactory g_ResourceFactory = 0;
+
+struct CallbackContext
+{
+    dmScript::LuaCallbackInfo* m_Callback;
+    int                        m_Request;
+};
+
+// TODO: Determine more about our actual use case!
```
JCash, line 40: I've postponed this api a little bit, as I believe it's not fully matching our use case yet.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/font/Fontc.java
```diff
@@ -219,6 +221,51 @@ public static int GetFontMapLayerMask(FontDesc fontDesc)
...
+            // always gets padded by the sqrt of a pixel diagonal
+            return fontDesc.getShadowBlur() + (int)(fontDesc.getOutlineWidth()) + 1;
+        } else {
+            return Math.min(4, fontDesc.getShadowBlur()) + (int)(fontDesc.getOutlineWidth());
+        }
+    }
+
+    public static float GetFontMapSdfSpread(FontDesc fontDesc)
+    {
+        float sdf_spread = getPaddedSdfSpread(fontDesc.getOutlineWidth());
```
ekharkunov: [Code style] Here and below: variable names should be in camelCase

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/FontBuilder.java
```diff
@@ -26,44 +26,88 @@
...
+            int buildDirLen        = this.project.getBuildDirectory().length();
+            String glyphBankPath   = task.input(3).getPath().substring(buildDirLen);
+            fontMapBuilder.setGlyphBank(glyphBankPath);
+        }
+
+        fontMapBuilder.setFont(fontDesc.getFont()); // Keep the suffix as-is (i.e. ".ttf")
         fontMapBuilder.setMaterial(BuilderUtil.replaceExt(fontDesc.getMaterial(), ".material", ".materialc"));
 
-        fontMapBuilder.setGlyphBank(genResourcePath);
+        boolean all_chars = fontDesc.getAllChars();
```
ekharkunov: all_chars -> allChars

engine/dlib/src/dlib/job_thread.cpp
```diff
@@ -115,16 +115,22 @@ static void JobThread(void* _ctx)
...
 #else
-static void UpdateSingleThread(JobThreadContext* ctx)
+static void UpdateSingleThread(JobThreadContext* ctx, uint64_t max_time)
 {
-    DM_PROFILE("SingleThread");
-    if (ctx->m_Work.Empty())
-        return;
-    // TODO: Perhaps time scope a number of items!
-    JobItem item = ctx->m_Work.Pop();
+    uint64_t tstart = dmTime::GetTime();
```
ekharkunov: Should we use `dmTime::GetMonotonicTime()` here? If I remember correctly we had some troubles if we use `GetTime` to calculate dt (https://github.com/defold/defold/pull/9834)

engine/engine/src/wscript
```diff
@@ -2,6 +2,7 @@
 import os, re
 from waf_dynamo import dmsdk_add_files, copy_file_task, apidoc_extract_task, platform_supports_feature, platform_graphics_libs_and_symbols
 import waflib.Options
+from waflib import Logs
```
ekharkunov: Unused import

share/extender/build_input.yml
```diff
@@ -4,19 +4,23 @@ context:
...
                   "opus", "libopus.lib",
                   "decoder_wav", "libdecoder_wav.lib",
                   "decoder_ogg", "libdecoder_ogg.lib",
-                  "decoder_opus", "libdecoder_opus.lib"]
+                  "decoder_opus", "libdecoder_opus.lib",
+                  "fontgen","libfontgen.lib"]
     allowedSymbols: ["DefaultSoundDevice", "NullSoundDevice", "AudioDecoderWav", "CrashExt", "AudioDecoderStbVorbis", "AudioDecoderOpus", "GraphicsAdapterWebGPU",
                     "ProfilerExt", "GraphicsAdapterOpenGL", "GraphicsAdapterOpenGLES", "GraphicsAdapterVulkan", "GraphicsAdapterNull",
                     "LiveUpdateExt", "ScriptBox2DExt", "ScriptImageExt", "ScriptModelExt", "ScriptTypesExt",
+                    "ScriptFont", "ResourceTypeFont", "ResourceTypeTTF",
```
ekharkunov, line 58: Is `"ScriptFontgenExt"` missed here?

engine/render/src/render/font.h
```diff
@@ -91,7 +91,8 @@ namespace dmRender
         uint8_t m_LayerMask;
 
         uint8_t m_IsMonospaced:1;
-        uint8_t m_Padding:7;
+        uint8_t m_IsDynamic:1;
+        uint8_t m_Padding:6;        // Note: Not C struct padding, but actual glyph padding.
```
ekharkunov, line 95: Is it worth to rename field to `m_StructPadding` or just leave unnamed like `uint8_t : 6;`

engine/render/src/render/font.h
```diff
@@ -165,7 +166,7 @@ namespace dmRender
      * @param font_map Font map handle
      * @param params Parameters to update
      */
-    void SetFontMap(HFontMap font_map, dmRender::HRenderContext render_context, dmGraphics::HContext graphics_context, FontMapParams& params);
+    bool SetFontMap(HFontMap font_map, dmRender::HRenderContext render_context, dmGraphics::HContext graphics_context, FontMapParams& params);
```
ekharkunov: Missed `@return` statement in comment

engine/render/proto/render/font_ddf.proto
```diff
@@ -57,60 +57,74 @@ message GlyphBank
...
+    optional FontRenderMode render_mode      = 15 [default = MODE_SINGLE_LAYER];
+
+    optional bool all_chars       = 16 [default = false]; // 0x000000 - 0x10FFFF
+    optional string characters    = 17 [default = ""];
+
+    // If zero, it'll grow dynamically to our max size (2k x 4k)
+    optional uint32 cache_width   = 18 [default = 0];
+    optional uint32 cache_height  = 19 [default = 0];
+
+    optional float sdf_spread     = 20 [default = 1];
```
ekharkunov: Here and below: default value for float should be 1.0/0.0

engine/render/proto/render/font_ddf.proto
```diff
@@ -57,60 +57,74 @@ message GlyphBank
...
+    optional uint64            glyph_padding         = 2 [default = 0];
+    optional uint32            glyph_channels        = 3 [default = 0];
     optional bytes             glyph_data            = 4; // glyph data may be compressed
 
-    optional float             max_ascent            = 5;
-    optional float             max_descent           = 6;
-    optional float             max_advance           = 7;
-    optional float             max_width             = 8;
-    optional float             max_height            = 9;
+    optional float             max_ascent            = 5 [default = 0];
```
ekharkunov: Default value for float should be with .0.

engine/gamesys/src/gamesys/test/fontview/fontview.cpp
```diff
@@ -132,6 +132,7 @@ namespace dmFontView
             window_params.m_Height           = 540;
             window_params.m_Title            = "FontView";
             window_params.m_ContextAlphabits = 8;
+            window_params.m_GraphicsApi      = dmPlatform::PLATFORM_GRAPHICS_API_VULKAN;
```
ekharkunov, line 135: Is it correct? If I understand correctly Vulakn isn't supported on every platform.

engine/gamesys/src/gamesys/scripts/script_font.cpp
```diff
@@ -0,0 +1,289 @@
...
+    lua_pushinteger(L, request_id);
+    return 1;
+}
+
+/*# removes glyphs from the font
+ * Removes glyphs from the font
+ *
+ * @name font.remove_glyphs
+ * @param path [type:string|hash] The path to the .fontc resource
+ * @param text [type:string] A unique list of unicode characters to be loaded
```
ekharkunov: I guess `to be loaded` -> `to be removed`

engine/gamesys/src/gamesys/scripts/script_font.cpp
```diff
@@ -0,0 +1,289 @@
...
+static int AddGlyphs(lua_State* L)
+{
+    DM_LUA_STACK_CHECK(L, 1);
+    int top = lua_gettop(L);
+
+    dmhash_t fontc_path_hash = dmScript::CheckHashOrString(L, 1);
+    const char* text = luaL_checkstring(L, 2);
+
+    dmScript::LuaCallbackInfo* luacbk = 0;
+    if (top > 2 && !lua_isnil(L, 3)) {
```
ekharkunov: [code style] Open bracket should be on separate line

engine/gamesys/src/gamesys/scripts/script_font.cpp
```diff
@@ -0,0 +1,289 @@
...
+    g_ResourceFactory = 0;
+    return dmGameSystem::FontGenFinalize(params);
+}
+
+static dmExtension::Result ScriptFontUpdate(dmExtension::Params* params)
+{
+    return dmGameSystem::FontGenUpdate(params);
+}
+
+DM_DECLARE_EXTENSION(ScriptFont, "ScriptFont", 0, 0, ScriptFontInitialize, ScriptFontUpdate, 0, ScriptFontFinalize)
```
ekharkunov, line 282: Should it be `ScriptFontgenExt` instead of `ScriptFont` because `ScriptFontgenExt` is used in app_manifest.clj?

engine/gamesys/src/gamesys/scripts/script_font.cpp
ekharkunov: Maybe it's better tp phrase like `A string with unique unicode characters to be removed` because `list` refers more to Lua table but type defined as "string"?

engine/gamesys/src/gamesys/scripts/script_font.cpp
ekharkunov: I think it's worth to check id 3rd argument is function or not? `CreateCallback` doesn't check type correctness.

engine/gamesys/src/gamesys/scripts/script_font.cpp
```diff
@@ -0,0 +1,289 @@
...
+ * ```
+ *
+ * ```lua
+ * -- Remove glyphs
+ * local requestid = font.remove_glyphs("/path/to/my.fontc", "abcABC123")
+ * ```
+ *
+ * ```lua
+ * -- Remove all ranges and glyphs associated with a .ttf resource
+ * local requestid = font.remove_subfont("/path/to/my.fontc", "/path/to/my.ttf")
```
ekharkunov: I don't find `font.remove_subfont` API.

engine/gamesys/src/gamesys/scripts/script_font.cpp
```diff
@@ -0,0 +1,289 @@
...
+ * You can call `font.remove_glyphs()` to remove them from memory.
+ *
+ * @note A glyph residing in the glyph cache will stay there until it's evicted.
+ * It will not be removed from the glyph cache by removing the loaded glyphs from the .fontc resource.
+ *
+ * @name font.add_glyphs
+ *
+ * @param path [type:string|hash] The path to the .fontc resource
+ * @param text [type:string] A unique list of unicode characters to be loaded
+ * @param [callback] [type:function(self, request_id, result, errstring)] A callback function that is called after the request is finished
```
ekharkunov: I think `[callback]` should be without square brackets

ekharkunov: Type should be `[type:function(self, request_id, result, errstring)|nil]` because callback is optional.

engine/gamesys/src/gamesys/scripts/script_font.cpp
```diff
@@ -0,0 +1,289 @@
...
+//     if (dmResource::RESULT_OK != r)
+//     {
+//         return DM_LUA_ERROR("Failed to remove glyph source '%s' from font '%s'", dmHashReverseSafe64(ttf_path_hash), dmHashReverseSafe64(fontc_path_hash));
+//     }
+//     return 0;
+// }
+
+static void AddGlyphsCallback(void* _ctx, int result, const char* errmsg)
+{
+    CallbackContext* ctx = (CallbackContext*)_ctx;
```
ekharkunov, line 103: How we handle case when `_ctx` is NULL? If I understand correctly - callback context created only of user pass function as 3rfd argument into `font.add_glyphs`.

engine/font/wscript
```diff
@@ -0,0 +1,38 @@
...
+    conf.env.append_value('INCLUDES', "default/src")
+    conf.env.append_unique('DEFINES', 'DLIB_LOG_DOMAIN="FONT"')
+
+def build(bld):
+    global test_context
+    bld.recurse('src')
+    TestContext.initialize_test_context(test_context, bld)
+
+def shutdown(self):
+    waf_dynamo.run_tests(test_context, valgrind = True)
```
ekharkunov, line 36: Q: Is it correct to run tests in `shutdown` hook?

engine/font/src/util.h
```diff
@@ -0,0 +1,27 @@
...
+//
+// You may obtain a copy of the License, together with FAQs at
+// https://www.defold.com/license
+//
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#if !defined(DM_UTIL_H)
```
ekharkunov: Suggest to use `DM_FONT_UTIL_H` define because it can be more util headers in different subsystems.

engine/font/src/font_ttf.cpp
```diff
@@ -0,0 +1,208 @@
...
+        memcpy((void*)font->m_Data, buffer, buffer_size);
+        font->m_Allocated = 1;
+    }
+    else
+    {
+        font->m_Data    = buffer;
+        font->m_DataSize= buffer_size;
+    }
+
+    int index = stbtt_GetFontOffsetForIndex((const unsigned char*)font->m_Data,0);
```
ekharkunov: [code style] Space before 0.

engine/font/src/font_ttf.cpp
```diff
@@ -0,0 +1,208 @@
...
+    int result = stbtt_InitFont(&font->m_Font, (const unsigned char*)font->m_Data, index);
+    if (!result)
+    {
+        dmLogError("Failed to load font from '%s'", path);
+        DestroyFontTTF((HFont)font);
+        delete font;
```
ekharkunov: Here will be double deallocation. First happened in `DestroyFontTTF`, second - `delete font`.

engine/font/src/font_ttf.cpp
```diff
@@ -0,0 +1,208 @@
...
+}
+
+void DestroyFontTTF(HFont hfont)
+{
+    TTFFont* font = ToFont(hfont);
+    if (font->m_Allocated)
+    {
+        free((void*)font->m_Data);
+    }
+    free((void*)font);
```
ekharkunov, line 57: Also need free font->m_Path, because it's intialized as `strdup(....)`

engine/font/src/font_ttf.cpp
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    int advx, lsb;
+    stbtt_GetGlyphHMetrics(info, glyph_index, &advx, &lsb);
+
+    int x0, y0, x1, y1;
+    stbtt_GetGlyphBox(info, glyph_index, &x0, &y0, &x1, &y1);
+
+    float scale = options->m_Scale;
+    float padding = options->m_StbttSDFPadding;
```
ekharkunov: Q: Can `padding` be 0? If yes - at line 167 we will get division by zero.

engine/font/src/font_ttf.cpp
```diff
@@ -0,0 +1,208 @@
...
+        x0 -= padding;
+        y0 -= padding;
+        x1 += padding;
+        y1 += padding;
+    }
+
+
+    glyph->m_Width = (x1 - x0) * scale;
+    glyph->m_Height = (y1 - y0) * scale;
+    glyph->m_Advance = advx*scale;;
```
ekharkunov: [code style] Double semi-colon in the end of the line.

engine/font/src/dmsdk/font/font.h
```diff
@@ -0,0 +1,272 @@
...
+        RESULT_OK,
+        RESULT_NOT_SUPPORTED,
+        RESULT_ERROR,
+    };
+
+    /*#
+     * FontType
+     * @enum
+     * @name FontType
+     * @member FONT_TYPE_UNKNOWN = -1
```
ekharkunov: Q: Is doc generator supports assignment? I mean constructions like `FONT_TYPE_UNKNOWN = -1`

engine/font/src/dmsdk/font/font.h
```diff
@@ -0,0 +1,272 @@
...
+    };
+
+    /*#
+     * Holds the bitmap data of a glyph.
+     * If there's an associated image, it is of size width * height * channels.
+     * @struct
+     * @name GlyphBitmap
+     * @member m_Width [type: uint16_t] The glyph image width
+     * @member m_Height [type: uint16_t] The glyph image height
+     * @member m_Channels [type: uint16_t] The glyph image height
```
ekharkunov: Incorrect description of the field

engine/font/src/font_ttf.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+static HFont LoadTTFInternal(const char* path, const void* buffer, uint32_t buffer_size, bool allocate)
+{
+    TTFFont* font = new TTFFont;
+    memset(font, 0, sizeof(*font));
+
+    if (allocate)
+    {
+        font->m_Data = (const void*)malloc(buffer_size);
+        memcpy((void*)font->m_Data, buffer, buffer_size);
```
ekharkunov, line 69: Missed `font->m_DataSize` assignment here. As result for allocated font reported data size will be zero.

engine/font/src/dmsdk/font/font.h
```diff
@@ -0,0 +1,272 @@
...
+    // Glyphs
+
+    /*#
+     * Holds the bitmap data of a glyph.
+     * If there's an associated image, it is of size width * height * channels.
+     * @struct
+     * @name GlyphOptions
+     * @member m_Scale [type: float] The font scale
+     * @member m_GenerateImage [type: bool] If true, generates an SDF image, and fills out the glyph.m_Bitmap structure.
+     * @member m_StbttSDFPadding [type: int] The sdk padding value (valid for FONT_TYPE_STBTTF fonts)
```
ekharkunov: Incorrect type. Should be `float`

engine/font/src/dmsdk/font/font.h
```diff
@@ -0,0 +1,272 @@
...
+        bool  m_GenerateImage;
+
+        // stbtt options (see stbtt_GetGlyphSDF)
+        float m_StbttSDFPadding;
+        int   m_StbttSDFOnEdgeValue;
+
+        GlyphOptions()
+        : m_Scale(1.0f)
+        , m_GenerateImage(false)
+        , m_StbttSDFPadding(3)
```
ekharkunov: [minor] IMO it's better to initialize float field like `m_StbttSDFPadding(3.f)`

engine/font/src/dmsdk/font/font.h
```diff
@@ -0,0 +1,272 @@
...
+        , m_StbttSDFOnEdgeValue(190)
+        {
+        }
+    };
+
+    /*#
+     * Get the metrics of a glyph
```
ekharkunov: Incorrect description

engine/font/src/dmsdk/font/font.h
```diff
@@ -0,0 +1,272 @@
...
+     * @name GetResourceSize
```
ekharkunov: Incorrect function name

engine/font/src/dmsdk/font/font.h
```diff
@@ -0,0 +1,272 @@
...
+     * @param font [type: dmFont::HFont] The font
+     * @param glyph [type: dmFont::Glyph*] The glyph
+     * @return result [type: dmFont::FontResult] The result
+     */
+    // In case a bit map was allocated, and you wish to free the memory
+    FontResult FreeGlyph(HFont font, Glyph* glyph);
+
+} // namespace
+
+#endif // DM_FONT_H
```
ekharkunov: [minor] Correct define in the comment should be `DMSDK_FONT_H`

engine/gamesys/src/dmsdk/gamesys/resources/res_font.h
```diff
@@ -179,7 +187,30 @@ namespace dmGameSystem
      */
     dmResource::Result ResFontRemoveGlyph(FontResource* font, uint32_t codepoint);
 
-    void   ResFontDebugPrint(FontResource* font);
+    /*# add a new glyph range
+     * Add a new glyph range
+     * @note Does not check if
```
ekharkunov: Incomplete sentence

engine/gamesys/src/dmsdk/gamesys/resources/res_font.h
```diff
@@ -179,7 +187,30 @@ namespace dmGameSystem
...
+     * @name ResFontAddGlyphSource
+     * @param factory [type: dmResource::HFactory] The factory
+     * @param font_hash [type: dmhash_t] The font path hash (.fontc)
```
ekharkunov: Correct  variable name is `fontc_hash`

engine/gamesys/src/dmsdk/gamesys/resources/res_font.h
```diff
@@ -179,7 +187,30 @@ namespace dmGameSystem
...
+     */
+    dmResource::Result ResFontAddGlyphSource(dmResource::HFactory factory, dmhash_t fontc_hash, dmhash_t ttf_hash, uint32_t codepoint_min, uint32_t codepoint_max);
+
+
+    /*# removes all glyph ranges associated with a ttfresource
+     * Removes all glyph ranges associated with a ttfresource
+     *
+     * @name ResFontRemoveGlyphSource
+     * @param factory [type: dmResource::HFactory] The factory
+     * @param font_hash [type: dmhash_t] The font path hash (.fontc)
```
ekharkunov: Correct variable name is `fontc_hash`

engine/gamesys/src/gamesys/resources/res_font_private.h
```diff
@@ -0,0 +1,81 @@
...
+        dmRenderDDF::FontMap*   m_DDF;
+        dmRender::HFont         m_FontMap;
+        HResourceDescriptor     m_Resource;             // For updating the resource size dynamically
+        MaterialResource*       m_MaterialResource;
+        GlyphBankResource*      m_GlyphBankResource;
+        TTFResource*            m_TTFResource;          // the first ttf resource
+        dmJobThread::HContext   m_Jobs;
+        uint32_t                m_CacheCellPadding;
+        uint32_t                m_ResourceSize;         // For correct resource usage reporting
+        bool                    m_IsDynamic;            // Are the glyphs populated at runtime?
```
ekharkunov: `bool m_IsDynamic;` -> `uint8_t m_IsDynamic : 1;`?

engine/gamesys/src/gamesys/fontgen/fontgen.h
```diff
@@ -0,0 +1,38 @@
...
+//
+// You may obtain a copy of the License, together with FAQs at
+// https://www.defold.com/license
+//
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#pragma once
```
ekharkunov: Q: Why somewhere we use pattern `#pragma once`, somewhere - `#if !defined(SOME_MACRO)`?

engine/gamesys/src/gamesys/fontgen/fontgen.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+    // As we've incref'd these resources, they should not have gone out of scope
+    if (!item->m_FontResource)
+        return 0;
+    if (!item->m_TTFResource)
+        return 0;
+
+    uint32_t codepoint = item->m_Codepoint;
+
+    uint64_t tstart = dmTime::GetTime();
```
ekharkunov: Better to use `dmTime::GetMonotonicTime()` to measure intervals.

engine/gamesys/src/gamesys/fontgen/fontgen.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+static void RemoveGlyphs(Context* ctx, FontResource* resource, const char* text)
+{
+    DM_MUTEX_SCOPED_LOCK(ctx->m_Mutex);
+
+    const char* cursor = text;
+    uint32_t c = 0;
+    while ((c = dmUtf8::NextChar(&cursor)))
+    {
+        dmGameSystem::ResFontRemoveGlyph(resource, c);
```
ekharkunov, line 420: Suggestion: Introduce function which removes set of glyphs in one call instead of several calls.

engine/gamesys/src/gamesys/fontgen/fontgen.cpp
```diff
@@ -0,0 +1,484 @@
...
+    delete item;
+}
+
+// Called on the main thread
+static void JobPostProcessGlyph(void* context, void* data, int result)
+{
+    Context* ctx = (Context*)context;
+    JobItem* item = (JobItem*)data;
+
+    DM_MUTEX_OPTIONAL_SCOPED_LOCK(item->m_Mutex);
```
ekharkunov, line 281: I'm a bit worried about invoking callback inside critical section because it can be situation when 
1. Mutex is locked
2. Callback is invoked
3. User do some other stuff with resource system inside callback which requires to lock resource mutex (which is the same as `JobItem` uses)
4. Deadlock appears.

WDYT about that case?

engine/gamesys/src/gamesys/resources/res_font.cpp
```diff
@@ -226,98 +309,311 @@ namespace dmGameSystem
...
+        params->m_ImageFormat        = ddf->m_OutputFormat;
+
+        params->m_SdfSpread          = ddf->m_SdfSpread;
+        params->m_SdfOutline         = ddf->m_SdfOutline;
+        params->m_SdfShadow          = ddf->m_SdfShadow;
+    }
+
+    static void GetMaxCellSize(dmFont::HFont hfont, float scale, const char* text, float* cell_width, float* cell_height)
+    {
+        *cell_width = 0;
```
ekharkunov: [minor] `*cell_width = 0.f;`

engine/gamesys/src/gamesys/resources/res_font.cpp
```diff
@@ -226,98 +309,311 @@ namespace dmGameSystem
...
+            {
+                *cell_width = dmMath::Max(*cell_width, glyph.m_Width);
+                *cell_height = dmMath::Max(*cell_height, glyph.m_Height);
+            }
+        }
+    }
+
+    static void SetupParamsForDynamicFont(dmRenderDDF::FontMap* ddf, const char* filename, dmFont::HFont hfont, dmRender::FontMapParams* params)
+    {
+        if (ddf->m_ShadowBlur > 0.0f && ddf->m_ShadowAlpha > 0.0f) {
```
ekharkunov: [code style] Brackets should be at separate lines

engine/gamesys/src/gamesys/resources/res_font.cpp
```diff
@@ -226,98 +309,311 @@ namespace dmGameSystem
...
+        {
+            sdf_shadow = CalcSdfValue(padding, shadow_padding);
+        }
+        params->m_SdfShadow = sdf_shadow;
+
+        float scale = dmFont::GetPixelScaleFromSize(hfont, ddf->m_Size);
+        params->m_MaxAscent = dmFont::GetAscent(hfont, scale);
+        params->m_MaxDescent = -dmFont::GetDescent(hfont, scale);
+
+        bool dynamic_cache_size = ddf->m_CacheWidth == 0 && ddf->m_CacheHeight == 0;
```
ekharkunov: Q: Should both size be zero to set "dynamic" flag to true? I mean how it should work if only width is 0?

engine/gamesys/src/gamesys/resources/res_font.cpp
```diff
@@ -226,98 +309,311 @@ namespace dmGameSystem
...
+
+        params->m_CacheCellWidth     = 0;
+        params->m_CacheCellHeight    = 0;
+        params->m_CacheCellMaxAscent = 0;
+
+        bool all_chars = ddf->m_AllChars;
+        bool has_chars = ddf->m_Characters != 0 && ddf->m_Characters[0] != 0;
+        if (!all_chars && has_chars)
+        {
+            bool dynamic_cache_size = ddf->m_CacheWidth == 0 && ddf->m_CacheHeight == 0;
```
ekharkunov: The same variable already calculated at line 460.

engine/gamesys/src/gamesys/wscript
```diff
@@ -23,7 +23,7 @@ def remove_from_node_list(lst, name):
 def build(bld):
     build_util = create_build_utility(bld.env)
 
-    sources = bld.path.ant_glob(['*.cpp', 'resources/**/*.cpp', 'components/**/*.cpp', 'scripts/*.cpp']) + bld.path.parent.parent.ant_glob('proto/gamesys/*')
+    sources = bld.path.ant_glob(['*.cpp', 'resources/**/*.cpp', 'components/**/*.cpp', 'scripts/*.cpp', 'fontgen/*.cpp']) + bld.path.parent.parent.ant_glob('proto/gamesys/*')
```
ekharkunov, line 26: Q: If I understand correctly we include `fontgen`source every time and doesn't create separate library to include/exclude via appmanifest. Is it correct understanding? If yes - should it be like this?

engine/render/src/render/font_renderer_private.h
```diff
@@ -120,20 +91,24 @@ namespace dmRender
...
 
-        uint32_t                m_CacheWidth;           // In texels
-        uint32_t                m_CacheHeight;          // In texels
-        uint32_t                m_CacheCellWidth;       // In texels
-        uint32_t                m_CacheCellHeight;      // In texels
-        uint32_t                m_CacheCellMaxAscent;   // In texels
-        uint32_t                m_CacheColumns;         // Number of cells in horizontal direction
-        uint32_t                m_CacheRows;            // Number of cells in horizontal direction
-        uint32_t                m_CacheCellCount;       // Number of cells in total
+        uint16_t                m_CacheMaxWidth;        // In texels
```
ekharkunov, line 94: Suggest to use here and below `uint16_t : 15;` because we can handle texture of maximum size 32k. So we don't need 65k maximum value. And we can use free bits for some of uint8_t value. Small but still optimization)

engine/render/src/render/font_renderer_private.h
```diff
@@ -120,20 +91,24 @@ namespace dmRender
...
+        uint16_t                m_CacheMaxHeight;       // In texels
+        uint16_t                m_CacheWidth;           // In texels
+        uint16_t                m_CacheHeight;          // In texels
+        uint16_t                m_CacheCellWidth;       // In texels
+        uint16_t                m_CacheCellHeight;      // In texels
+        uint16_t                m_CacheCellMaxAscent;   // In texels
+        uint16_t                m_CacheColumns;         // Number of cells in horizontal direction
+        uint16_t                m_CacheRows;            // Number of cells in horizontal direction
+        uint16_t                m_CacheCellCount;       // Number of cells in total
         uint8_t                 m_CacheChannels;        // Number of channels
```
ekharkunov: Here we can use `uint8_t : 2;` because possible values is 1 or 3. 6 free bits more)

ekharkunov (changes requested):
See code comments.
Also additional questions:
1. Padding has different types in the different structures (int vs float). Should be the same type to be consistent?
2. New option in Bob Project.java should affects how asset cache calculates?
3. Should font asset be cached after the first build via Bob?

share/extender/build_input.yml
JCash: No, it's called `ScriptFont` now.

engine/gamesys/src/gamesys/resources/res_font.cpp
JCash: makes sense

engine/gamesys/src/gamesys/resources/res_font.cpp
JCash: thanks, forgot to remove it

engine/gamesys/src/gamesys/wscript
JCash: Yes, the idea is that we want to move towards an easier-to-use font system.
At least that is the current goal.
As shown in the numbers table in the OP, using runtime fonts is a net gain, with this file included in those numbers.

engine/render/src/render/font_renderer_private.h
JCash: While true, fiy it won't carry over to uint8, so I'd need to switch from uint8 to uint16 to make it work.
And, I'd have to interleave every uint16 with a "unit8 : 1" which  honestly makes it all quite unreadable.
Even with uint64, it'd be messy.
It's one of the big shortcomings of bitpacking unfortunately.
So, for this case, we'll avoid that opt for now, as I favor readability in this case.
Also, we only have a few fonts in memory at the same time.

JCash:
Thanks for the feedback!
For the questions:

> Padding has different types in the different structures (int vs float). Should be the same type to be consistent?

When it's an int, it comes from a pixel bounddaries.
But at runtime, we need to do calculations with floats.

> New option in Bob Project.java should affects how asset cache calculates?

I believe it already is by this `paramsForSignature = {"font-runtime-generation"}` in FontBuilder.java ?

> Should font asset be cached after the first build via Bob?

Good call, I was sure it was the default but it seems we've opted for not caching as the default.
However, as recommended by Alexey, simple copy steps are not to be cached, as it's more effective to simply copy the file.

engine/font/src/dmsdk/font/font.h
JCash: probably not.

engine/font/src/font_ttf.cpp
JCash: Theoretically yes

engine/font/wscript
JCash: That's how it's been setup for as long as I can remember.
"Correct", I'm not sure, but I'd rather move to another build system before changing this.

engine/gamesys/src/gamesys/fontgen/fontgen.cpp
JCash: It shouldn't be a problem, as the mutex is (should be) reentrant.
That is, if it's already locked on the same thread, it can be locked again on the same thread.
I'm pretty sure we have more than one place where we use this fact, although it's a bit tricky to pinpoint an example.

engine/gamesys/src/gamesys/fontgen/fontgen.cpp
JCash: Probably good idea. I'll add it in my next PR

engine/gamesys/src/gamesys/fontgen/fontgen.h
JCash: Yeah, my mistake here.
We shouldn't use pragma once as it has some caveats (https://en.wikipedia.org/wiki/Pragma_once#:~:text=Additionally%2C%20%23pragma%20once%20can%20do,in%20a%20compiler%2Ddependent%20way.)

engine/gamesys/src/gamesys/scripts/script_font.cpp
JCash: If user doesn't pass that argument, then `callback = 0;` and there should be no callback to begin with, thus no need to check the ctx for null.

engine/gamesys/src/gamesys/scripts/script_font.cpp
JCash: As Björn just fixed all docs, and we use brackets [here](https://github.com/defold/defold/blob/dev/engine/script/src/script_sys.cpp#L454), I'm going to leave it in for now at least.

JCash: I don't believe so, as `nil` is an explicit argument.
Having it optional means you don't have to specify anything. See [example](https://github.com/defold/defold/blob/dev/engine/script/src/script_sys.cpp#L454)

engine/gamesys/src/gamesys/scripts/script_font.cpp
JCash: Nah, I opted for `ScriptFont` to make slightly nicer.
I'll fix the app_manifest.clj instead.

engine/gamesys/src/gamesys/test/fontview/fontview.cpp
JCash: This tool isn't really used either, so it likely needs much more love.
It's also a tool only built for our desktops for debugging.
I don't remember why I had to do this change, but iirc it had to do with some compilation issue.

engine/render/src/render/font.h
JCash: I thought the comment would explain that it's _not_ structure padding, but an actual value called padding? :)

engine/render/src/render/font.h
JCash: thanks. I'll also move it back to the cpp file (not used outside of the font.cpp)

engine/font/src/dmsdk/font/font.h
```diff
@@ -0,0 +1,272 @@
...
+        uint32_t    m_Codepoint; // Unicode code point
+        float       m_Width;
+        float       m_Height;
+        float       m_Advance;
+        float       m_LeftBearing;
+        float       m_Ascent;
+        float       m_Descent;
+    };
+
+    typedef struct Font* HFont;
```
ekharkunov, line 132: Should also be documented

engine/gamesys/src/gamesys/fontgen/fontgen.cpp
```diff
@@ -0,0 +1,484 @@
...
+    }
+}
+
+static void DeleteItem(Context* ctx, JobItem* item)
+{
+    ReleaseResources(ctx, item);
+
+    if (item->m_Callback) // It's the last item
+    {
+        delete item->m_Status;
```
ekharkunov, line 270: [memort leak] `item->m_Status->m_Error` doesn't deallocate memory here

engine/gamesys/src/gamesys/fontgen/fontgen.cpp
```diff
@@ -0,0 +1,484 @@
...
+        // Incref the resources so that they're not accidentally removed during this threaded work
+        if (!loading) // If we're in the process of creating this resource, then we cannot incref it!
+            dmResource::IncRef(ctx->m_ResourceFactory, fontresource);
+        dmResource::IncRef(ctx->m_ResourceFactory, ttfresource);
+
+        dmFont::HFont font = dmGameSystem::GetFont(ttfresource);
+        float scale = dmFont::GetPixelScaleFromSize(font, font_info.m_Size);
+
+        // TODO: Pass dmFont::HFont instead of ttfresource
+        GenerateGlyph(ctx, fontresource, ttfresource, c, scale, stbtt_padding, stbtt_edge, is_sdf,
```
ekharkunov, line 404: Is it worth to pack all arguments into struct to avoid passing 13 arguments? If I remember correctly it's recommended to create intermediate structs for function arguments if overall arguments count is more than 6.

engine/gamesys/src/gamesys/fontgen/fontgen.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+dmExtension::Result FontGenInitialize(dmExtension::Params* params)
+{
+    g_FontExtContext = new Context;
+    g_FontExtContext->m_ResourceFactory = params->m_ResourceFactory;
+    // As we're interacting with the resources, it's good to use the resource mutex
+    g_FontExtContext->m_Mutex = dmResource::GetLoadMutex(params->m_ResourceFactory);
+
+    // 3 is arbitrary but resembles the output from our old generator
+    g_FontExtContext->m_StbttDefaultSdfPadding = dmConfigFile::GetInt(params->m_ConfigFile, "fontgen.stbtt_sdf_base_padding", 3);
```
ekharkunov, line 432: Here and below: properties' definitions are missed in meta.properties, meta.edn and test/resources/save_data_project/game.project

ekharkunov (changes requested):
One more place with possible memory leak and one more entry need to be documented. Everything else if good to me. After fix you can merge without additional review from my side

engine/gamesys/src/gamesys/fontgen/fontgen.cpp
JCash: Yeah, I made them private again. We can always add them later if we need to.

engine/gamesys/src/gamesys/fontgen/fontgen.cpp
JCash: Yeah, we can probably clean it up later.
A structure is good for api-purposes, but this is internal to this file, so not super urgent.
And the argument count is for making it easy for a human to remember, but it's also not a function one writes more than once (i.e. here), so also not super urgent.

PR #10631, opened 2025-05-11
Make sure that GUI components are sorted per world

### LINKED ISSUE #9123: Stencils overwriting each other from different collections
Feels there is an issues with stencils overwriting each other from different collections, so I created a test project to demo this.

### Expected behaviour

image

### Actual behaviour

image

### Steps to reproduce
I have two gui-s.
In every gui is a stencil-node, and a "stick" node.

image
image

Every gui is placed in own collection, and downloaded dynamicly with proxy 
image

### Can just build my test project:
https://github.com/ivolff/defold_stencil_example

<hr/>

<table>
<tr><td>Build time</td><td>2024-05-28T14:20:53.141474</td></tr>
<tr><td>Defold editor sha</td><td>7247f038fba6f40cda23c5ea2e38e9fde69a2ea6</td></tr>
<tr><td>Defold engine sha</td><td>fd1ad4c17bfdcd890ea7176f2672c35102384419</td></tr>
<tr><td>Defold version</td><td>1.8.1</td></tr>
<tr><td>GPU</td><td>NVIDIA GeForce RTX 3070 Laptop GPU/PCIe/SSE2</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 NVIDIA 551.61</td></tr>
<table>

AGulev:
Make sure that GUI components spawned in different worlds (in different collection proxies) are sorted separately, and that the later-spawned world is always rendered on top unless another behavior is specified using `gui.set_render_order()`.

Fix https://github.com/defold/defold/issues/9123

### Technical changes
In our sorting order, we have 12 unused bits, so I’m using them for world sorting to ensure that each new world is sorted separately.  
At the same time, I’ve kept `scene_order` as the highest priority so that users can still explicitly define the order from code.
This change helps avoid unpredictable behavior caused by GUI sorting clashes between collections, while still keeping the system flexible for users who specify render order using `gui.set_render_order()`.

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_gui.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_gui.cpp
@@ -1031,8 +1031,10 @@ namespace dmGameSystem
 
     inline uint32_t MakeFinalRenderOrder(uint32_t world_order, uint32_t scene_order, uint32_t sub_order)
     {
-        // scene_order limited to 0-15, see gui_script.gui LuaSetRenderOrder() funtion
-        return (world_order << 20) | (scene_order << 4) | sub_order;
+        // scene_order: 4 bits (max 15)
+        // world_order: 12 bits (max 4095)
+        // sub_order:   16 bits (max 65535)
+        return (scene_order << 28) | (world_order << 16) | sub_order;
     }
 
     static void SetBlendMode(dmRender::RenderObject& ro, dmGui::BlendMode blend_mode)
@@ -3153,7 +3155,7 @@ namespace dmGameSystem
         gui_context->m_MaxAnimationCount = dmConfigFile::GetInt(ctx->m_Config, "gui.max_animation_count", 1024);
         gui_context->m_MaxParticleBufferCount = dmConfigFile::GetInt(ctx->m_Config, "gui.max_particle_buffer_count", 1024);
 
-        // TODO: it seems like hidden game.project property which should be gui.world_count like it is in physics
+        // TODO: it seems like this is a hidden game.project property which should be gui.world_count like it is in physics
         int32_t max_world_count = dmConfigFile::GetInt(ctx->m_Config, "gui.max_instance_count", 128);
         if (max_world_count > 4096)
         {
--- a/engine/gamesys/src/gamesys/components/comp_gui.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_gui.cpp
@@ -425,6 +425,7 @@ namespace dmGameSystem
 
         for (uint32_t i = 0; i < gui_world->m_RenderConstants.Size(); ++i)
         {
+            gui_world->m_RenderOrder = i;
             if (gui_world->m_RenderConstants[i])
             {
                 dmGameSystem::DestroyRenderConstants(gui_world->m_RenderConstants[i]);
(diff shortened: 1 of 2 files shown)


engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -1028,9 +1029,12 @@ namespace dmGameSystem
...
     }
 
-    inline uint32_t MakeFinalRenderOrder(uint32_t scene_order, uint32_t sub_order)
+    inline uint32_t MakeFinalRenderOrder(uint32_t world_order, uint32_t scene_order, uint32_t sub_order)
     {
-        return (scene_order << 16) + sub_order;
+        // scene_order: 4 bits (max 15)
+        // world_order: 12 bits (max 4095)
+        // sub_order:   16 bits (max 65535)
+        return (scene_order << 28) | (world_order << 16) | sub_order;
```
AGulev, line 1037: `scene_order` - the order user specified using `gui.set_render_order()` still have the highest priority.

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -3149,8 +3155,14 @@ namespace dmGameSystem
         gui_context->m_MaxAnimationCount = dmConfigFile::GetInt(ctx->m_Config, "gui.max_animation_count", 1024);
         gui_context->m_MaxParticleBufferCount = dmConfigFile::GetInt(ctx->m_Config, "gui.max_particle_buffer_count", 1024);
 
-        int32_t max_gui_count = dmConfigFile::GetInt(ctx->m_Config, "gui.max_instance_count", 128);
-        gui_context->m_Worlds.SetCapacity(max_gui_count);
+        // TODO: it seems like this is a hidden game.project property which should be gui.world_count like it is in physics
+        int32_t max_world_count = dmConfigFile::GetInt(ctx->m_Config, "gui.max_instance_count", 128);
+        if (max_world_count > 4096)
```
AGulev, line 3160: it seems like this is a hidden option. I added a comment here, not sure if I should fix it (make it visible in the game.project + rename it) as part of this PR

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -302,6 +302,7 @@ namespace dmGameSystem
         GuiWorld* gui_world = new GuiWorld();
         if (!gui_context->m_Worlds.Full())
         {
+            gui_world->m_RenderOrder = gui_context->m_Worlds.Size();
```
JCash, line 305: I don't think this is good enough to keep the number unique.
Scenario:
```
Spawn Coll A -> order = 0
Spawn Coll B -> order = 1
Despawn Coll A
Spawn Coll C -> order = 1
```

It would work, if the order is updated in the `CompGuiDeleteWorld()` function, and in fact we're already looping over all the worlds there.

engine/gamesys/src/gamesys/components/comp_gui.cpp
JCash: I'm not sure why it's hidden. Was it ever present and then removed? (i.e. deprecated?)

JCash (changes requested):
overall good. small bug in the render order calculation

JCash (approved):

PR #10635, opened 2025-05-12
Improve Scene View error experience

matgis:
* Render errors from Scene View objects are now presented in a more readable view and can be copied into the clipboard.
* Most shader syntax errors will now cause render errors in the Scene View.

### Technical changes
* You must now explicitly specify the byte order when creating buffers using the `editor.buffers` module.
* Moved the code related to shader transpilation and reflection to new `editor.pipeline.shader-gen` module.
* Refactored shader transpilation and reflection so it no longer needs an OpenGL context, which lets us determine the attribute requirements earlier and cache attribute buffers in the graph once we port over all the editor-specific shaders to the transpiled pipeline. 
* Made some preparations to support compute shaders in addition to vertex and fragment shaders in the shader compilation pipeline.
* You can now specify `:overlay-anchor-pane-props`in the map produced by the `scene` of your resources `SceneNode` to configure an `AnchorPane` overlaid on top of the Scene View using `cljfx.fx.anchor-pane` properties.
* Added `util.defonce` module to make it possible to avoid redefining JVM classes when reloading Clojure files that define types.
* Made it possible to clear the `scene-cache` from the REPL.
* `VolatileCache` now supports metadata.
* Added `dev/clear-caches!` as a one-stop function to clear all the caches during development.
* Added `combine-comparisons` and `combine-hashes` helper macros to the `java` module.
* Added `of-type`, `from-type` and `from` functions to the `array` module for creating typed JVM arrays.
* Added `conj-set` and `conj-vector` functions to the `coll` module.
* Renamed `nm` field in `RenderPass` record to `name` (only used when debugging).
* Renamed `_toolbar.scss` to `_scene-view.scss`.
* Removed unused `glslang` and `tint` executables from the editor bundle, since we were using the ones from Bob already.
* Fixed various inefficiencies in the rendering-related code.

### Notes
This PR is mostly intended to get some of the preparations I've done in order to improve scene view rendering performance and workflows merged into `dev` so my own branch doesn't deviate too much during this long-running task. It is titled "Improve Scene View error experience" for the benefit of the release notes, but as you can see, there are lots of changes unrelated to that.


editor/src/clj/editor/gl/shader.clj
```diff
@@ -340,13 +350,25 @@ This must be submitted to the driver for compilation before you can use it. See
...
+  (case value
+    (:shader-type-vertex :shader-type-fragment #_ :shader-type-compute) true
+    false))
+
+(defn- shader-type->gl-shader-type
+  ^long [shader-type]
+  (case shader-type
+    :shader-type-vertex GL2/GL_VERTEX_SHADER
+    :shader-type-fragment GL2/GL_FRAGMENT_SHADER
+    #_#_ :shader-type-compute GL3/GL_COMPUTE_SHADER))
```
vlaaad, line 370: What's the significance of this comment?

editor/src/clj/editor/gl/shader.clj
```diff
@@ -499,38 +525,98 @@ This must be submitted to the driver for compilation before you can use it. See
...
+         (every? string? (keys uniform-values-by-name))]}
+  ;; TODO(instancing): We should be able to derive the attributes-infos from the
+  ;; attribute-reflection-infos we get after transpiling the shader. This would
+  ;; enable us to get the attribute-infos from a shader without an active OpenGL
+  ;; context. However, we first need to update all the editor-specific shaders
+  ;; created using the defshader macro to pass through the same transpilation
+  ;; pipeline as the material shaders. Unfortunately this also involves porting
+  ;; them to #version 140 first, since the transpiler does not support versions
+  ;; earlier than that. Such shaders are also created from editor plugins, so we
+  ;; need to update those.
```
vlaaad, line 562: 👍

editor/src/clj/editor/gl/texture.clj
```diff
@@ -28,34 +30,37 @@
...
-   :swizzle-g    GL3/GL_TEXTURE_SWIZZLE_G
-   :swizzle-b    GL3/GL_TEXTURE_SWIZZLE_B
-   :swizzle-a    GL3/GL_TEXTURE_SWIZZLE_A
-   :wrap-s       GL2/GL_TEXTURE_WRAP_S
-   :wrap-t       GL2/GL_TEXTURE_WRAP_T
-   :wrap-r       GL2/GL_TEXTURE_WRAP_R})
+(defn texture-param->gl-texture-param
+  ^long [texture-param]
+  "This function translates Clojure keywords into OpenGL constants. You can use
+  the keywords in texture parameter maps."
```
vlaaad, line 44: This string is supposed to be a docstring, but instead it is inside the function body. Move it above the arglist?

editor/src/clj/editor/scene.clj
```diff
@@ -738,6 +749,30 @@
...
+                 :anchor-pane/left 0
+                 :anchor-pane/right 0
+                 :anchor-pane/top 0
+                 :style-class "info-text-area"
+                 :editable false
+                 :wrap-text true
+                 :text (string/join \newline (error-message-lines [error]))}]}
+    (if-let [overlay-anchor-pane-props (:overlay-anchor-pane-props scene)]
+      overlay-anchor-pane-props
+      (let [info-text (or tool-info-text (:info-text scene))]
```
vlaaad, line 765: Since you use `^:try`, `tool-info-text` here could be an `ErrorValue`. But it seems that the code expects it to be a string. Is it because `tool-info-text` is an error value only when `scene` has `:error`?

editor/src/clj/util/array.clj
```diff
@@ -13,82 +13,144 @@
...
-     (reduce (fn [^long index item]
-               (Array/set array index item)
-               (inc index))
-             (- length more-count)
-             more)
+         array (of-type-impl Object length a b c d e f g h)]
+     (set-items-impl! array more 8)
      array)))
+
+(defn of-type
```
vlaaad, line 86: Do you have any performance numbers for this compared to, e.g., `into-array`?

editor/src/clj/util/defonce.clj
```diff
@@ -0,0 +1,39 @@
...
+  [name & args]
+  (if (not (class? (resolve name)))
+    (list* 'defrecord name args)))
+
+(defmacro type
+  "Like core.deftype, but will never be redefined."
+  {:arglists '([name [& fields] & opts+specs])}
+  [name & args]
+  (if (not (class? (resolve name)))
+    (list* 'deftype name args)))
```
vlaaad, line 39: Nice! Typical Clojure suggestion is to define all types and protocols in another ns that is not reloaded, but I like this approach more.

editor/src/dev/dev.clj
```diff
@@ -1030,6 +1030,30 @@
       (pair context-id entry-count))
     (scene-cache/cache-stats)))
 
+(defn clear-caches!
```
vlaaad, line 1033: 🔥

vlaaad (changes requested):
Looks good! I only want us to fix a function definition with a misplaced docstring that causes a string to be created and immediately discarded on every function invocation.

editor/src/clj/editor/scene.clj
matgis: Well, the code will only use it as a string if it is a string. If it is an `ErrorValue`, we won't render anything here.

However, I think we should fall back on the `:info-text` from the `scene` if the `tool-info-text` is an `ErrorValue` rather than showing nothing. I'll fix!

editor/src/clj/editor/gl/texture.clj
matgis: Oops. Nice catch! I'll fix it!

editor/src/clj/editor/gl/shader.clj
matgis: I've included the compute shader constants from the Protobuf enum, but compute shaders are not really supported by the OpenGL profile we currently use in the editor. I'll add a comment to clarify!

vlaaad (approved):
Ship it!

PR #10651, opened 2025-05-14
Directories are now treated as resources in editor scripts

### LINKED ISSUE #5372: Directories should be resources in editor scripts
Currently I do not get the path to the selected directory.
If I want to create something from an directory. For example creating an atlas from a folder of images or a level collection from a folder of json files or similar.

vlaaad:
Now, directories may be accessed as resources in editor scripts. The directories have 2 properties:
- `path` - the path to the directory, same as with file resources
- `children` - list of children resources

For example, here is how you can ensure that an atlas has all images from a directory:
```lua
local images = editor.get("/assets/images", "children")
local atlas = "/assets/main.atlas"
local txs = {editor.tx.clear(atlas, "images")}
for i = 1, #images do
    local resource_path = editor.get(images[i], "path")
    if resource_path:match(".png$") ~= nil then
        txs[#txs + 1] = editor.tx.add(atlas, "images", {image = resource_path})
    end
end
editor.transact(txs)
```

Fixes #5372


matgis (dismissed):
Looks good to me!

PR #10661, opened 2025-05-15
Make sure that changes in shaders are always taken into account during rebuild

AGulev:
Fix https://github.com/defold/defold/issues/10657


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderProgramBuilder.java
```diff
@@ -82,7 +82,7 @@ public Task create(IResource input) throws IOException, CompileExceptionError {
         ShaderProgramBuilderBundle.ModuleBundle modules = ShaderProgramBuilderBundle.ModuleBundle.load(input);
         for (String path : modules.getModules()) {
             IResource moduleInput = this.project.getResource(path);
-
+            taskBuilder.addInput(moduleInput);
```
AGulev, line 85: `inputs` and `outputs` are our dependency graph. All the inputs uses to calculate a hash key which used to know if cache entry should be invalidated. In this cases shader code wasn't added as input, so changes weren't taken inti account by the build system

britzl (approved):

PR #10679, opened 2025-05-16
Added support for spawning from collectionfactory components in dmsdk.

Gert-Jan:
Replicated the already existing style and dmsdk functionality of comp_factory.h but for collectionfactories instead of normal factories.

### Technical changes
* A test can be found here: https://github.com/Gert-Jan/tutorial-side-scroller/tree/dmsdk-version-collectionfactory-test
* Replaced `CompFactorySpawn()` based on the conversation down here. Changes:
  * No longer takes an int index as param.
  * The result can be retrieved using the `HInstance* out_instance` param.
  * Returns a `dmGameObject::Result` so it can return errors such as `RESULT_IDENTIFIER_IN_USE`.
* Added `CompCollectionFactorySpawn()`. Notes:
  * Takes a `const char* id_prefix` param. If nullptr is passed a default id will be generated, e.g. `/collection1/*`, `/collection2/*` etc.
  * Returns a `dmGameObject::Result`. In particular so it can return errors such as `RESULT_IDENTIFIER_IN_USE`.
  * The result can be retrieved using a `dmGameObject::InstanceIdMap* out_instances` param.
* Added a number of new `dmGameObject::Result` entries to cover what can go wrong during game object spawning.
* `dmGameObject::SpawnInternal()`, `dmGameObject::CollectionSpawnFromDescInternal()`, `dmGameObject::SpawnFromCollection()` now also return `dmGameObject::Result` instead of bool.
* Renamed `dmGameObject::InstancePropertyBuffers` to `dmGameObject::InstancePropertyContainers` and moved it to dmsdk.
* Moved `dmGameObject::InstanceIdMap` to dmsdk.
* CollectionFactoryComp_Create is now also using CompCollectionFactorySpawn for some unit test coverage.
* Modified unit test to also test for collection factory spawning with custom id's.
* Added unit test to test if RESULT_IDENTIFIER_INVALID and RESULT_IDENTIFIER_IN_USE are returned when collection spawning.
* Added documentation for `CompFactorySpawn()` and `CompCollectionFactorySpawn()`.


engine/gamesys/src/dmsdk/gamesys/components/comp_collection_factory.h
```diff
@@ -0,0 +1,56 @@
...
+#include <dmsdk/dlib/hash.h>
+#include <dmsdk/dlib/vmath.h>
+#include <dmsdk/dlib/hashtable.h>
+
+namespace dmGameObject
+{
+    typedef struct CollectionHandle* HCollection;
+    typedef dmHashTable<dmhash_t, HPropertyContainer> InstancePropertyBuffers;
+    typedef dmHashTable<dmhash_t, dmhash_t> InstanceIdMap;
+}
```
JCash, line 28: While I'm all for forward declarations, it feels strange to not have them public as well.
Please add these to `engine/gameobject/src/dmsdk/gameobject/gameobject.h`.

engine/gamesys/src/dmsdk/gamesys/components/comp_collection_factory.h
```diff
@@ -0,0 +1,56 @@
...
+        COMP_COLLECTION_FACTORY_STATUS_LOADED = 2,  //!< COMP_COLLECTION_FACTORY_STATUS_LOADED
+    };
+
+    CompCollectionFactoryStatus CompCollectionFactoryGetStatus(CollectionFactoryComponent* component);
+    bool                        CompCollectionFactoryIsLoading(CollectionFactoryComponent* component);
+    bool                        CompCollectionFactoryIsDynamicPrototype(CollectionFactoryComponent* component);
+
+    bool CompCollectionFactorySpawn(HCollectionFactoryComponent component, dmGameObject::HCollection collection,
+                                    const dmVMath::Point3& position, const dmVMath::Quat& rotation, const dmVMath::Vector3& scale,
+                                    dmGameObject::InstancePropertyBuffers* properties, dmGameObject::InstanceIdMap* out_instances);
```
JCash, line 53: * I believe a `bool` is insufficient in the long run.
Please replace with `dmGameObject::Result`

*  Please add `world` to the signatures. This is for code consistency, and longevity of these api's.
E.g. follow the style of preexisting apis, in this case `comp_factory.h`

* Another point of our refactor is to make sure the script api's are simplified. Please update the `CollectionFactoryComp_Create` function in `script_collection_factory.cpp` accordingly to use this new function.

* By using it in the script api, you will also get testing done on the functionality, which is missing from the PR.

JCash (changes requested):
Thanks!
Needs a few minor changes

engine/gamesys/src/dmsdk/gamesys/components/comp_collection_factory.h
Gert-Jan: Done. Note:

> * I believe a `bool` is insufficient in the long run.
Please replace with `dmGameObject::Result`

For now it just returns RESULT_UNKNOWN_ERROR in case of fail.

engine/gamesys/src/dmsdk/gamesys/components/comp_collection_factory.h
```diff
@@ -0,0 +1,56 @@
...
+        COMP_COLLECTION_FACTORY_STATUS_UNLOADED = 0,//!< COMP_COLLECTION_FACTORY_STATUS_UNLOADED
+        COMP_COLLECTION_FACTORY_STATUS_LOADING = 1, //!< COMP_COLLECTION_FACTORY_STATUS_LOADING
+        COMP_COLLECTION_FACTORY_STATUS_LOADED = 2,  //!< COMP_COLLECTION_FACTORY_STATUS_LOADED
+    };
+
+    CompCollectionFactoryStatus CompCollectionFactoryGetStatus(CollectionFactoryComponent* component);
+    bool                        CompCollectionFactoryIsLoading(CollectionFactoryComponent* component);
+    bool                        CompCollectionFactoryIsDynamicPrototype(CollectionFactoryComponent* component);
+
+    dmGameObject::Result CompCollectionFactorySpawn(HCollectionFactoryWorld world, HCollectionFactoryComponent component, dmGameObject::HCollection collection,
```
JCash, line 51: Question is, do we need the forward declaration now that we need to include gameobject.h for this "Result" enum?
And, how does this build if it's not including gameobject.h?

JCash (changes requested):

engine/gamesys/src/dmsdk/gamesys/components/comp_collection_factory.h
Gert-Jan: Interesting... I haven't looked into this very deeply, but I did experiment commenting out the whole dmGameObject forward declaration and it did build just fine.
The reason it builds is because in the [compilation unit (`comp_collection_factory.cpp`)](https://github.com/defold/defold/blob/af70a8015d186030645f5f711d70354d5034804b/engine/gamesys/src/gamesys/components/comp_collection_factory.cpp#L26) it includes `gameobject/gameobject.h` which in turn [includes `dmsdk/gameobject/gameobject.h`](https://github.com/defold/defold/blob/af70a8015d186030645f5f711d70354d5034804b/engine/gameobject/src/gameobject/gameobject.h#L20).

So I didn't touch the existing forward decleration because:
1. This is how it's done in `comp_factory.h`
2. My guess of the reason why it's done like that in `comp_factory.h` is because then you can only include `comp_factory.h` (and not `gameobject.h`) in a native extension and it works? Maybe?

If that's the case than I'll also need to forward declare `dmGameObject::Result`.

Gert-Jan:
There is something else that need to be mentioned here about dmsdk factory and collectionfactory feature parity.

Factory:
* [ `CompFactorySpawn()` takes a `dmhash_t id` ](https://github.com/defold/defold/blob/2c5040e7729013b395a9e85307d8d34c7dbab1b4/engine/gamesys/src/gamesys/components/comp_factory.cpp#L468)
  * This is pretty cool, because this allows for overriding the defailt `/instance1` with `/my_own_unique_instance_id`.
* [`CompFactorySpawn()` takes `uint32_t index`](https://github.com/defold/defold/blob/2c5040e7729013b395a9e85307d8d34c7dbab1b4/engine/gamesys/src/gamesys/components/comp_factory.cpp#L468)
  * I'm not sure why this is there. As far as I can tell it is only used to specify a location in the instance pool.
  * You could potentially put anything in there but I think it would not play nice if you're not using [`dmGameObject::AcquireInstanceIndex()`](https://github.com/defold/defold/blob/2c5040e7729013b395a9e85307d8d34c7dbab1b4/engine/gameobject/src/gameobject/gameobject.cpp#L980) to generate this index.
  * I guess my question would be: should this be part of the interface of `CompFactorySpawn()`?
  * It is also used in the [cppgame example to figure out beforehand if there is room in the pool for another instance.](https://github.com/defold/tutorial-side-scroller/blob/719fbdf3eead5ecafc7e3d268b077ca7013be24e/cppgame/src/cppgame.cpp#L123) This functionality could be covered with a dmGameObject::Result just like is done here for `CompCollectionFactorySpawn()`

Collectionfactory:
* `CompCollectionFactorySpawn()` as proposed here does not take a `dmhash_t id`
  * Passing a `dmhash_t` would not make much sense because collection factory produces a number of hashes formatted like `/collection1/foo`, `/collection1/bar`.
  * It would only make sense if you could pass the `/collection1` part, but you can't combine `dmhash_t prefix = hash("/collection1")` with another `dmhash_t`.
  * To get similar functionality you would either need to pass a prefix `const char*` or as is done in [`CollectionSpawnFromDescInternal()`](https://github.com/defold/defold/blob/2c5040e7729013b395a9e85307d8d34c7dbab1b4/engine/gameobject/src/gameobject/gameobject.cpp#L1339), use/pass a `HashState64` to combine hashes (pretty cool haven't seen anything like this before).
  * This could then generate `/my_own_unique_instance_id/foo` and `/my_own_unique_instance_id/bar`
* `CompCollectionFactorySpawn()` as proposed here does not take a `uint32_t index`.
  * Should it?

As far as a concrete use-case for id overriding:
* I'm working on something that tries to link spawned objects over a network.
* Currently I'm not overriding ids, but instead use translation tables (e.g. `/instance1` at Alice == `/instance7` at Bob).
* With id overrides I could use the same id on all clients (e.g. `/i_[client_id]_1`).

In conclusion. Questions:
* Should id overriding be added to `CompCollectionFactorySpawn()`?
  * If so in what form, `const char*` or `HashState64` or something else?
  * Note: This would require the top bits of `CollectionSpawnFromDescInternal()` to be moved around.
* Should index be added to `CompCollectionFactorySpawn()`?
  * If no, what about `CompFactorySpawn()`?

JCash:
Hi!

Yes, you are fully correct that the functions should be as similar as possible, and that's indeed a major part of the process of moving them from internal to public headers.


CompFactorySpawn:

* I agree that the `index` is quite messy, and should be internalized (as you suggest). I think it's a bit due to the scripting api, which allocates the index upfront, and then sends it via message.
But that doesn't have to be the case for the synchronous call to CompFactorySpawn().

CompCollectionFactorySpawn: 

* I agree that it shouldn't take any indices into it.

> Should id overriding be added to CompCollectionFactorySpawn()?

I think that generating the id hashes is not something we want to mess with at this stage.
You have the id mappings, in which the keys will be the same on all clients.
Isn't it enough for you to do the mapping on your end?

> Should index be added to CompCollectionFactorySpawn()?

Let's remove the index from the CompFactorySpawn()!

> Add an overloaded CompFactorySpawn() that:
returns a dmGameObject::Result instead of a bool.

Yes, that sounds good (it didn't return a bool though?)

> takes a const char* id instead of the uint32_t index, dmhash_t id.

The ID has to be unique, which is why the generation of the ID is done with the helper functions. I don't think we should introduce an "id" to the arguments.

> Change CollectionSpawnFromDescInternal() so it returns a dmGameObject::Result

Yes.

> Change CollectionSpawnFromDescInternal() so it takes a HashState64 prefixHashState.

Same here, I don't wish to overcomplicate things.
You should be able to use the id mappings as you mentioned earlier.

Also, given the requirement of uniqueness, the counter in the ID, still isn't guarantueed to be exact on two different machines.
1) due to lag, and 2) you're likely to spawn effects etc that aren't replicated.

So `/i_[client_id]_1` on Alice may still be `/i_[client_id]_2` on Bob.

In short, I'm in favour of most suggested changes here, but I'm not convinced by the "custom" prefix here.
The dmSDK C++ api is still experimental though, so we may revisit your ideas later, when we've had more time to process the problems/ideas around it.

Makes sense?

Gert-Jan:
Thanks for the reply, all clear, but for the following:

> > Should id overriding be added to CompCollectionFactorySpawn()?
> 
> I think that generating the id hashes is not something we want to mess with at this stage. You have the id mappings, in which the keys will be the same on all clients. Isn't it enough for you to do the mapping on your end?

I'm already doing the mapping on my end, it's just a step I could completely skip with custom id's. I'll advocate one last time for this and then I'll let it go ok? :D
- Custom id's is currently **already possible** but only for `CompFactorySpawn()` as it takes a `dmhash_t id` which is used as id. So you can do id overriding for factory but currently not for *collection*factory.
- Currently id uniqueness for `CompFactorySpawn()` [is already checked](https://github.com/defold/defold/blob/dc7f0b578cc02fc679a61d6ac28c9cdf446007b6/engine/gameobject/src/gameobject/gameobject.cpp#L1027) which returns a nice error, but the error will not (yet) return to the dmsdk api.
- The suggestion for dealing with non-unique id's is to return a dmGameObject::Result::RESULT_IDENTIFIER_IN_USE from `CompFactorySpawn()` instead of hoping dmsdk users will always use the result of `dmGameObject::CreateInstanceId()` to pass as `dmhash_t id` in `CompFactorySpawn()`. The same could be done for `CompCollectionFactorySpawn()`
- Alternatively, maybe `dmhash_t id` also should not be there at `CompFactorySpawn()`?
- Another option is having overloaded functions with and without `dmhash_t id` param.
  - Without id: uses `dmGameObject::CreateInstanceId()` internally. 
  - With id: now you as the api user are responsible for id uniqueness, but we'll tell you specifically when the id isn't unique.
  - Or use `dmGameObject::CreateInstanceId()` internally if 0 is passed as `dmhash_t id`.

> Also, given the requirement of uniqueness, the counter in the ID, still isn't guarantueed to be exact on two different machines.
> 
> 1. due to lag, and 2) you're likely to spawn effects etc that aren't replicated.
> 
> So `/i_[client_id]_1` on Alice may still be `/i_[client_id]_2` on Bob.

No it's guaranteed to be the same:
1. Alice will spawn object `foo` as `/i_[Alice_id]_5`.
2. Alice will send a message to Bob: With `foo_factory` spawn at position `x,y,z` using index `5`.
3. Bob knows Alice's unique client id (in fact, any other client will know Alice by the same unique client id).
4. Bob will spawn `foo` as `/i_[Alice_id]_5`.
We may even use not the client id but a session id to guard for disconnect/reconnects of Alice.

In any case, I'm ok with anything, I'll just work with what will be there in the end.

JCash:
Hi!
(Sorry for the late answer!)

First of all, let me thank you for your effort and insights!
It's very valuable to us!

I think the custom ID topic was brought up recently by the team as well, so I believe I'm warming to the idea now.

> Custom id's is currently already possible but only for CompFactorySpawn() as it takes a dmhash_t id which is used as id. So you can do id overriding for factory but currently not for collectionfactory.

I think this is an accidental feature from the fact that this C++ api was moved from internal to public.
I don't consider the api complete yet (as this discussion highlights.)

I think one big difference with your ID proposal, from what we've discussed internally and from how it's currently implemented, is that your solution allows for a completely unique id. As opposed to our internal suggestions to just alter the "prefix" (i.e. the "instance" part), while still using a running number.

As long as we can detect/handle unique id's and return a good return value, I think that your suggestion is good.

> No it's guaranteed to be the same:

Just to make sure we're on the same page here.

Regarding this statement, it will only be the same, if you provide the full ID yourself, like `id = hash("i_[Alice_id]_5")`
Because, as our current implementation works, it's using an internal running number, which you cannot guarantuee will be in sync between clients.

So, the suggested solution above, will be needed to assert the statement "No it's guaranteed to be the same".
Am I understanding you correctly here?

> dmGameObject::Result::RESULT_IDENTIFIER_IN_USE

I think this is a good solution.

engine/gamesys/src/dmsdk/gamesys/components/comp_collection_factory.h
JCash: > This is how it's done in comp_factory.h

Forward declarations is a constant work imho, as things shift over time.
I'm not sure the comp_factory.h is super "correct".
Sometimes I wonder if we want things like "gameobject_forward.h" with forward declarations (including the enums, as enums cannot be forward declared)

No need to put to much effort into it at this moment.

engine/gameobject/src/dmsdk/gameobject/gameobject.h
```diff
@@ -101,6 +102,20 @@ namespace dmGameObject
...
+     * @name InstanceIdMap
+     */
+    typedef dmHashTable<dmhash_t, dmhash_t> InstanceIdMap;
+
+    /*#
+     * Contains property buffers for game objects to be spawned
+     * @typedef
+     * @name InstancePropertyBuffers
+     */
+    typedef dmHashTable<dmhash_t, HPropertyContainer> InstancePropertyBuffers;
```
JCash: I think we should rename it to `InstancePropertyContainers`, to make it clearer that it holds `HPropertyContainer`
Also, the doc mentions "buffers" which could be anything, so I think "property containers" is more specific.

JCash (changes requested):
Looks overall good.
Small typedef name change I think is good.

engine/gameobject/src/dmsdk/gameobject/gameobject.h
Gert-Jan: About the forward declarations:
Locally I already deleted all dmGameObject forward declarations in `comp_factory.h` and `comp_collection_factory.h` and instead just added `#include <dmsdk/gameobject/gameobject.h>` precisely because of the `dmGameObject::Result` enum which would otherwise need to be kept in sync in multiple places.

The only disadvantage of doing it like this, that I can think of, is that compilation may be every so slightly slower when a user only includes `#include <dmsdk/gamesys/components/comp_factory.h>` (and not gameobject.h). Not sure if it's worth it to worry too much about that, because, as a dmsdk user I think it's rare to only include `comp_factory.h` and not also `#include <dmsdk/gameobject/gameobject.h>` given how intertwined they are.

Maybe there are other reasons I'm missing to not simply replace the forward declarations with pulling in gameobject.h entirely?

Also, with a `gameobject_forward.h` you may overtime just end up with almost all that is in `gameobject.h` minus the function definitions. Alternatively it's also possible to do a `gameobject_enums.h` to solve the enum problem specifically.

Gert-Jan: - Renamed `InstancePropertyBuffers` to `InstancePropertyContainers`
- Removed the `dmGameObject` forward declarations from `comp_factory.h` and `comp_collection_factory.h` to prevent having to copy in enums.
- Both factory and collectionfactory support custom id's now and will return `RESULT_IDENTIFIER_IN_USE` if an identical id is used. They can also take nullptr as id, which means it will revert to the default id's (`/instance1`, 1/collection1/*` etc.)
- Returning `dmGameObject::Result` was quite impactful, code changes-wise. I used 2 slightly different approaches for factory and collectionfactory.
  -  `CompFactorySpawn()` does not use `dmGameObject::Spawn()` but `dmGameObject::SpawnInternal()` instead. If `dmGameObject::Spawn() were to be used it would also need to return `dmGameObject::Result` and return the result through a `HInstance *` parameter. Which then also requires updates everywhere else `Spawn()` is used in the engine.
  - `CompCollectionFactorySpawn()` uses `dmGameObject::SpawnFromCollection()` instead of `dmGameObject::CollectionSpawnFromDescInternal()`. Mainly because `SpawnFromCollection()` was not widely used in the engine, which made the function signature changes less impactful.
  - Maybe this should be one approach? Maybe this is ok?

engine/gameobject/src/gameobject/gameobject.h
```diff
@@ -121,6 +121,22 @@ namespace dmGameObject
...
+     * @param id Identifier for the new instance, must be unique within the collection
+     * @param property_container Properties that should be set on the new instance
+     * @param position Position
+     * @param rotation Rotation
+     * @param scale Scale
+     * @param out_instance Output parameter for the new instance
+     * @return RESULT_OK on success
+     */
+    Result SpawnInternal(HCollection collection, HPrototype proto, const char *prototype_name, dmhash_t id, 
+                        HPropertyContainer property_container, const Point3& position, const Quat& rotation, const Vector3& scale, HInstance* out_instance);
```
JCash: If it's not internal (it's used outside of gameobject.cpp), we shouldn't name it "Internal"
Perhaps just "Spawn" ?

engine/gamesys/src/dmsdk/gamesys/components/comp_factory.h
```diff
@@ -51,6 +56,24 @@ namespace dmGameSystem
...
+     * @param position [type: dmVMath::Point3] Position of the spawned object
+     * @param rotation [type: dmVMath::Quat] Rotation of the spawned object
+     * @param scale [type: dmVMath::Vector3] Scale of the spawned object
+     * @param properties [type: dmGameObject::HPropertyContainer] Property container with override properties
+     * @param out_instance [type: dmGameObject::HInstance] Output parameter for the new instance
+     * @return result [type: dmGameObject::Result] Result of the operation
+     */
+    dmGameObject::Result CompFactorySpawn(HFactoryWorld world, HFactoryComponent component, dmGameObject::HCollection collection, dmhash_t id,
+                                                const dmVMath::Point3& position, const dmVMath::Quat& rotation, const dmVMath::Vector3& scale,
+                                                dmGameObject::HPropertyContainer properties, dmGameObject::HInstance* out_instance);
```
JCash: We don't want two public interfaces.
If we want to return a Result (which makes sense), then we should use the new function.

engine/gamesys/src/gamesys/components/comp_factory.cpp
```diff
@@ -484,5 +484,44 @@ namespace dmGameSystem
         return instance;
     }
 
+    dmGameObject::Result CompFactorySpawn(HFactoryWorld world, HFactoryComponent component, dmGameObject::HCollection collection, dmhash_t id,
```
JCash, line 466: Same here, we don't want two functions for the same thing.

JCash (changes requested):
We're getting real close.
Some changes needed with the last update.

Also, one major thing missing is the unit tests for the custom id prefix.

engine/gameobject/src/gameobject/gameobject.h
Gert-Jan: Done. Note that there already is another "Spawn" that returns an HInstance (instead of a Result), which is the one that is used everywhere else, and why I used SpawnInternal for dmsdk in the first place.

Gert-Jan:
> Also, one major thing missing is the unit tests for the custom id prefix.

Added some unit tests for custom id prefixes and for the duplicate id error result.
Updated the PR description.

engine/gameobject/src/gameobject/test/collection/test_gameobject_collection.cpp
```diff
@@ -262,6 +263,33 @@ TEST_F(CollectionTest, CollectionSpawning)
...
     }
 
+    // spawn 5 collections with a custom id
+    const char* prefix = "/custom";
+    char buffer[32];
+    for (int i=0;i!=5;i++)
+    {
+        dmGameObject::InstanceIdMap output;
+        dmGameObject::InstancePropertyContainers props;
+        sprintf(buffer, "%s%d\0", prefix, i);
```
JCash: FYI, we have dmSnPrintf().
Not required here, but good for visibility (educational) to help users into our code style.

engine/gameobject/src/gameobject/test/collection/test_gameobject_collection.cpp
```diff
@@ -306,6 +334,46 @@ TEST_F(CollectionTest, CollectionSpawningToFail)
...
+    ret = dmGameObject::Update(coll, &m_UpdateContext);
+    ASSERT_TRUE(ret);
+
+    dmGameObject::InstanceIdMap output;
+    dmGameObject::InstancePropertyContainers props;
+
+    // RESULT_IDENTIFIER_INVALID: custom prefix without leading slash
+    dmGameObject::Result result = Spawn(m_Factory, coll, "/root1.collectionc", "custom", &props, pos, rot, scale, &output);
+    ASSERT_TRUE(result == dmGameObject::RESULT_IDENTIFIER_INVALID);
+    ASSERT_EQ(output.Size(), 0u);
```
JCash, line 363: 👍

engine/gameobject/src/gameobject/test/collection/test_gameobject_collection.cpp
```diff
@@ -306,6 +334,46 @@ TEST_F(CollectionTest, CollectionSpawningToFail)
...
+
+    // RESULT_IDENTIFIER_IN_USE: duplicate id (prefix)
+    result = Spawn(m_Factory, coll, "/root1.collectionc", "/custom", &props, pos, rot, scale, &output);
+    ASSERT_TRUE(result == dmGameObject::RESULT_OK);
+    ASSERT_NE(output.Size(), 0u);
+    output.Clear();
+    result = Spawn(m_Factory, coll, "/root1.collectionc", "/custom", &props, pos, rot, scale, &output);
+    ASSERT_TRUE(result == dmGameObject::RESULT_IDENTIFIER_IN_USE);
+    ASSERT_EQ(output.Size(), 0u);
```
JCash, line 372: 👍

JCash (approved):
Looks good!
Thanks for your patience!

Gert-Jan:
> Looks good! Thanks for your patience!

Thanks for all the reviews!

britzl:
Good job @Gert-Jan ! Thank you!

JCash (changes requested):
I tried to build this locally and it failed.
Which platfom(s) have you run the tests on?

```
../src/gameobject/test/collection/test_gameobject_collection.cpp:282:41: error: format string contains '\0' within the string body [-Werror,-Wformat]
  282 |         sprintf(buffer, "%s%d/sub1/child\0", prefix, i);
  ```

Gert-Jan:
> I tried to build this locally and it failed. Which platfom(s) have you run the tests on?
> 
> ```
> ../src/gameobject/test/collection/test_gameobject_collection.cpp:282:41: error: format string contains '\0' within the string body [-Werror,-Wformat]
>   282 |         sprintf(buffer, "%s%d/sub1/child\0", prefix, i);
> ```

Windows, worked fine there. Was under the impression that Defold's build system was using Clang on windows too, but it looks like it's using MSVC, so that's where things went wrong here.

Anyway I checked the docs, sprintf() (and variants) automatically adds a null terminator so the \0 can go... The more you know... While at it, also replaced sprintf() with dmSnPrintf() as suggested earlier.

JCash (approved):

AGulev:
@JCash we added this PR into 1.10.3, but beta was already released. 
should we change it to 1.10.4 or maybe cherry-pick it into beta then?

britzl:
> @JCash we added this PR into 1.10.3, but beta was already released. should we change it to 1.10.4 or maybe cherry-pick it into beta then?

FYI I'm cherry picking this into the beta.

PR #10681, opened 2025-05-16
Support texture arrays in dx12

Jhonnyg:


engine/graphics/src/dx12/graphics_dx12.cpp
```diff
@@ -502,6 +526,45 @@ namespace dmGraphics
...
+            depth_clear.DepthStencil.Depth   = 1.0f;
+            depth_clear.DepthStencil.Stencil = 0;
+
+            hr = context->m_Device->CreateCommittedResource(
+                &CD3DX12_HEAP_PROPERTIES(D3D12_HEAP_TYPE_DEFAULT),
+                D3D12_HEAP_FLAG_NONE,
+                &depth_desc,
+                D3D12_RESOURCE_STATE_DEPTH_WRITE,
+                &depth_clear,
+                IID_PPV_ARGS(&context->m_FrameResources[i].m_DepthStencil)
```
JCash, line 553: Try to use our `DM_IID_PPV_ARGS` we use on the xbox branch, as they're different on the platforms.
Makes it easier to merge.
(No, we haven't moved that from xbox branch to public branch yet afaik)

JCash (dismissed):

PR #10683, opened 2025-05-16
Add warnings to open-url and open-file

### LINKED ISSUE #7819: Debug -> Open Web Profiler silently fails in some Linux environments
Selecting Debug -> Open Web Profiler does nothing. Tried while running from Project -> Build and Debug -> Start/Attach both with same results.

**To Reproduce (REQUIRED)**
1. Start project with Project -> Build.
2. Click Debug -> Open Web Profiler.
3. Wait for nothing to happen?

Expecting the web profiler to open in the default web browser.

  - Version 1.4.8

**Platforms (REQUIRED):**
 - Platforms: Linux
 - OS: Linux (Kubuntu 23.04)
 - Device: Desktop (Default browser set to Firefox)

**Workaround (OPTIONAL):**
The in-app overlay profiler works fine. May also be able to manually enter the URL into the web browser but not real sure what that URL would be.

This is likely specific to local environment conditions. It also generates no related logs. Opening web links from applications on Linux distros in general are hit or miss. After discussing with @britzl on Discord, it was suggested to make this issue: "We should try to work around it and at the least let the user know about it."

A potentially easy, non-fix, but informative option may be to print to console or pop up something that has the URL the editor just tried to open with the 'Open Web Profiler' click. This way even if nothing happens the user could copy/paste that into their browser (assuming it works that way.)

menaechmi:
On Linux, users reported not being able to open the web profiler. The problem seems to stem from missing `gvfs` required to open their browser. Users experiencing this will now see a warning in their log telling them that installing `gvfs` might fix the issue. The editor will now also log a warning if opening files isn't supported.

Fixes #7819

### Technical Changes
* This change is against `editor-dev` due to its low impact. (but actually this time)
* I doubt anyone has encountered the false clause of open-file, but we might as well log it if it happens.
* Because the GVFS issue is specific to Linux, but any future failure of Desktop$Action/BROWSE may not be, we just print the warning, and then only on Linux print another line about GVFS. I will be so embarrassed if a different issue also causes this error.
* It's in the linked issue, but here is my comment about why this fixes the issue: 
* Opening the web profiler uses `java.awt.Desktop.Action.BROWSE` which calls `gtk_show_uri()` [The GTK docs](https://docs.gtk.org/gtk3/func.show_uri_on_window.html) say:
>The uri must be of a form understood by GIO (i.e. you need to install gvfs to get support for uri schemes such as http:// or ftp://, as only local files are handled by GIO itself).

* I can reproduce this error successfully by uninstalling and re-installing `gvfs`.
* And here's what it looks like in a terminal (in case you can't uninstall GVFS or have no access to a Linux host)
defold-gvfs

## PR checklist

* [x] Code
* [x] Documentation

----------

### DIFF
--- a/editor/src/clj/editor/ui.clj
+++ b/editor/src/clj/editor/ui.clj
@@ -2234,7 +2234,10 @@
     (do
       (.start (Thread. #(.browse desktop (as-url url))))
       true)
-    false))
+    (do
+      (log/warn :message (format "Cannot open browser"))
+      (if (os/is-linux?) (log/warn :message (format "Installing gvfs may fix this error")))
+    false)))
 
 (defn open-file
   ([^File file]
@@ -2250,7 +2253,9 @@
                                 (on-error-fn (.getMessage e))
                                 (throw e)))))))
        true)
-     false)))
+     (do
+       (log/warn :message (format "Cannot open file handler"))
+     false))))
 
 (defn- make-path-data
   ^String [^double col ^double row outlines]
--- a/editor/src/clj/editor/ui.clj
+++ b/editor/src/clj/editor/ui.clj
@@ -2235,9 +2235,8 @@
       (.start (Thread. #(.browse desktop (as-url url))))
       true)
     (do
-      (log/warn :message (format "Cannot open browser"))
-      (if (os/is-linux?) (log/warn :message (format "Installing gvfs may fix this error")))
-    false)))
+      (log/warn :message (str "Cannot open browser." (when (os/is-linux?) " Installing gvfs may fix this.")))
+      false)))
 
 (defn open-file
   ([^File file]
@@ -2254,8 +2253,8 @@
                                 (throw e)))))))
        true)
      (do
-       (log/warn :message (format "Cannot open file handler"))
-     false))))
+       (log/warn :message "Cannot open file handler")
+       false))))
 
 (defn- make-path-data
   ^String [^double col ^double row outlines]


editor/src/clj/editor/ui.clj
```diff
@@ -2234,7 +2234,10 @@
     (do
       (.start (Thread. #(.browse desktop (as-url url))))
       true)
-    false))
+    (do
+      (log/warn :message (format "Cannot open browser"))
+      (if (os/is-linux?) (log/warn :message (format "Installing gvfs may fix this error")))
+    false)))
```
sprocketc, line 2240: We don't have to use `format` in this case, since we are not interpolating. We can also use `when` instead of `if`, since there is not `else` branch. We could also concatenate the strings to use a single `log/warn`. Additionally, we should fix the indentation of the returned bools. The end result would look something like this
```clojure
(do
  (log/warn :message (str "Cannot open browser." (when (os/is-linux?) "Installing gvfs may fix this.")))
  false)
```

sprocketc (changes requested):
Hey @menaechmi, thanks for your contribution! 

I tested this on GNOME and KDE, and confirmed that the implementation relies on `gvfs`, so having something like `kio-fuse` installed on KDE won't fix this. I also left a couple of comments about some minor issues.

sprocketc:
Also note that we now work directly on the `dev` branch, so you should change the base from `editor-dev` to `dev`.

britzl:
> you should change the base from `editor-dev` to `dev`

Note: I updated the contribution guidelines where @menaechmi probably read about using editor-dev

https://github.com/defold/defold/commit/d91ec4edb2dd5e025828ce896daecc69cd81fd3f

editor/src/clj/editor/ui.clj
```diff
@@ -2250,7 +2253,9 @@
                                 (on-error-fn (.getMessage e))
                                 (throw e)))))))
        true)
-     false)))
+     (do
+       (log/warn :message (format "Cannot open file handler"))
+     false))))
```
sprocketc, line 2258:
```suggestion
     (do
       (log/warn :message "Cannot open file handler")
       false))))
```

sprocketc (approved):
Thanks!

PR #10685, opened 2025-05-18
Fixed an issue with editor logs

### LINKED ISSUE #10226: No output to editor console.
### Expected behaviour
When launched via F5, information is output to the console
### Actual behaviour
When launched via F5, information is not output to the console
### Steps to reproduce
A periodic bug that occurs when launching via F5. No information is output to the console.

If the launch occurs without delays, then everything works fine. If there is a delay during launch (due to what??) then no information is sent to the console. In this case, you can select a device via the menu (project -> target) and then information starts to be sent to the console.
Does not occur on an empty or small project.

https://github.com/user-attachments/assets/a2d0a29c-6c68-4398-aacd-89009ebd62de

Also, if the connection was normal, then Local dmengine appears in the list, otherwise not.

Image
Image

It feels like there is some kind of timeout when trying to connect.

The bug appears on versions 1.9.4 and higher, I suspect that after this https://github.com/defold/defold/pull/9393
<hr/>

<table>
<tr><td>Build time</td><td>2025-02-12T18:00:16.249806</td></tr>
<tr><td>Defold editor sha</td><td>868769ba7a3458db12d149188bf3be80a339a85c</td></tr>
<tr><td>Defold engine sha</td><td>868769ba7a3458db12d149188bf3be80a339a85c</td></tr>
<tr><td>Defold version</td><td>1.9.7</td></tr>
<tr><td>GPU</td><td>NVIDIA GeForce GTX 1060 3GB/PCIe/SSE2</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 NVIDIA 566.36</td></tr>
<table>

AGulev:
Fixes the issue where logs in the Editor's console don't work when Simulated Resolution is used (Windows only).

Fix https://github.com/defold/defold/issues/10226

### DIFF
--- a/editor/src/clj/editor/targets.clj
+++ b/editor/src/clj/editor/targets.clj
@@ -117,14 +117,14 @@
 (defn update-launched-target! [target target-info on-service-url-found]
   (let [old @launched-targets]
     (reset! launched-targets
-            (map (fn [launched-target]
-                   (if (= (:id launched-target) (:id target))
-                     (let [result-target (merge launched-target target-info)]
-                       ;(when (and (:url target-info) (not (:url launched-target)))
-                       ;  (on-service-url-found result-target))
-                       result-target)
-                     launched-target))
-                 old))
+            (mapv (fn [launched-target]
+                    (if (= (:id launched-target) (:id target))
+                      (let [result-target (merge launched-target target-info)]
+                        (when (and (:url target-info) (not (:url launched-target)))
+                          (on-service-url-found result-target))
+                        result-target)
+                      launched-target))
+                  old))
     (when (not= old @launched-targets)
       (clear-selected-target-hint!)
       (invalidate-target-menu!))))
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/targets.clj
```diff
@@ -121,7 +121,14 @@
...
                        (when (and (:url target-info) (not (:url launched-target)))
-                         (on-service-url-found result-target))
+                         (future
+                           ;; Wait for the console stream to be marked as active
+                           (loop []
+                             (if (console/current-stream? (:log-stream result-target))
+                               (on-service-url-found result-target)
+                               (do
+                                 (Thread/sleep 100)
+                                 (recur))))))
```
AGulev, line 131: I'm not sure if it's a good solution tbh, more like a workaround
But it works

matgis (approved):
LGTM!

PR #8203, opened 2023-11-01
Split bob atlas layout calculation into two, removing dependency on animations

JCash:
## PR checklist

* [ ] Code
	* [ ] Add engine and/or editor unit tests.
	* [ ] New and changed code follows the overall code style of existing code
	* [ ] Add comments where needed
* [ ] Documentation
	* [ ] Make sure that API documentation is updated in code comments
	* [ ] Make sure that manuals are updated (in github.com/defold/doc)
* [ ] Prepare pull request and affected issue for automatic release notes generator
	* [ ] Pull request - Write a message that explains what this pull request does. What was the problem? How was it solved? What are the changes to APIs or the new APIs introduced? This message will be used in the generated release notes. Make sure it is well written and understandable for a user of Defold.
	* [ ] Pull request - Write a pull request title that in a sentence summarises what the pull request does. Do not include "Issue-1234 ..." in the title. This text will be used in the generated release notes.
	* [ ] Pull request - Link the pull request to the issue(s) it is closing. Use on of the [approved closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
	* [ ] Affected issue - Assign the issue to a project. Do not assign the pull request to a project if there is an issue which the pull request closes.
	* [ ] Affected issue - Assign the "breaking change" label to the issue if introducing a breaking change.
	* [ ] Affected issue - Assign the "skip release notes" is the issue should not be included in the generated release notes.

----------

Example of a well written PR description:

1. Start with the user facing changes. This will end up in the release notes.
1. Add one of the GitHub approved closing keywords
1. Optionally also add the technical changes made. This is information that might help the reviewer. It will not show up in the release notes. Technical changes are identified by a line starting with one of these:
   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```
There was a anomaly in the carbon chroniton propeller, introduced in version 8.10.2. This fix will make sure to reset the phaser collector on application startup.

Fixes #1234

### Technical changes
* Pay special attention to line 23 of phaser_collector.clj as it contains some interesting optimizations
* The propeller code was not taking into account a negative phase.
```

PR #10697, opened 2025-05-19
Improve performance when editing a large tilesource

### LINKED ISSUE #5919: Editing a tilesource is slow
Two reports of tilesource editing being very slow:

* https://forum.defold.com/t/defold-editor-slow/68356/15
* https://forum.defold.com/t/tilesource-collision-groups-unusually-slow/68092?u=britzl

**To Reproduce (REQUIRED)**
See forum posts above

It should not be slow

  - Version 1.2.182

**Platforms (REQUIRED):**
 - Platforms: ?

Related to (or similar fix as) https://github.com/defold/defold/pull/5898 ?

sprocketc:
Editing collision groups or other properties of a tilesource which do not affect the tilesource layout will no longer regenerate the texture.

Fixes #5919

### Technical changes
- Decoupled layout calculation from texture generation as per https://github.com/defold/defold/pull/8203#pullrequestreview-1727088174


matgis (dismissed):
LGTM! ❤️

matgis:
@sprocketc Typically the `digest-ignored` prefix should be used in cases where the value does not contribute meaningfully to the files we build for the engine to consume. You can use it to ignore things like node-ids, which are typically not deterministic between runs of the editor, and may only be used to report errors in specific cases. The layout-result does seem to affect the resulting texture image binaries, so we cannot ignore it. We either need to implement the `Digestable` protocol for it, or convert it to a Clojure representation.

matgis:
Regarding the `IllegalFormatConversionException`, we should either case the pivot components to `int`, or use `%g` or `%f` instead of `%d` for the floats.

matgis (approved):
Looks good to me!

PR #10701, opened 2025-05-20
Introduce property drag handle

sprocketc:
drag-handle
Draggable properties now display a dedicated drag controller on mouse over input, instead of using the label as a handle.


matgis (approved):
LGTM! ❤️

PR #10702, opened 2025-05-20
Fix issue where profiler doesn't work after reboot

### LINKED ISSUE #9427: Profiler is not working if build was relaunched by Defold Editor
The profiler is not working if the game build was relaunched from Editor. I mean If the game is running, we returning to the Defold Editor and run Build command. The new instance of build can be attached with profiler in this case

**To Reproduce (REQUIRED)**
1. Run game
2. Click Profiler - it's working
3. Relaunch game without close it from Defold Editor
4. Click Profiler - it's not working
5. Close game before start - it's working

Profiler is working

Tested on 1.8.0 and 1.9.2, both affected

**Platforms (REQUIRED):**
Tested on MacOS

[issue_profiler.zip](https://github.com/user-attachments/files/16993884/issue_profiler.zip)

**Screenshots (OPTIONAL):**

https://github.com/user-attachments/assets/4de0f419-cb8b-49b9-adc1-95f7217cf102

### LINKED ISSUE #10671: Rebuild game from Editor and sys.reboot breaks Profiler
If the game was build from editor while previous instance is running (not manually closed)
or used `sys.reboot()` before, profiler will not work

**To Reproduce (REQUIRED)**
1. Run game, press "R"
2. Profiler working
3. Rebuild game or click to `sys.reboot()`
4. Profiler breaks

Profiler works

1.10.1

**Platforms (REQUIRED):**
Tested on MacOS

https://github.com/user-attachments/assets/42f9f434-9e5a-4d97-9054-958ed69c5f8c

AGulev:
Make sure that the profiler shows the correct scope and property names after the engine reboot.

Fix https://github.com/defold/defold/issues/10671
Fix https://github.com/defold/defold/issues/9427


engine/dlib/src/dlib/profile/profile_remotery.cpp
```diff
@@ -35,6 +35,7 @@ namespace dmProfile
     static void*                    g_PropertyTreeCallbackCtx = 0;
     static int32_atomic_t           g_ProfilerInitialized = 0;
     static dmSpinlock::Spinlock     g_ProfilerLock;
+    uint8_t g_ProfilerGeneration = 0;
```
AGulev, line 38: this variable used to know when static properties and scopes should be recreated

JCash (changes requested):
Some minor change (u32 -> u8)

However, the bigger question is: why are the properties/scopes invalid after a reboot?
I feel it isn't discussed, nor is the choice of this solution.

Overall, I would have seen an attempt at not resetting the current property/scope data.

PR #10715, opened 2025-05-22
Added support for hash userdata type keys in lua table (de)serialization.

Gert-Jan:
It is now possible to use `sys.save()` and `sys.load()` to save and load tables with hashes as keys. This allows for example to (de)serialize tables that are produced and used by collectionfactory. This change also means that `sys.serialize()` and `sys.deserialize()` can serialize and deserialize tables with hashes as keys.

⚠️ Using a hash as a key in a serialized table makes the binary format incompatible with older versions of the engine.

Fixes #3208 

### Technical changes
* Added a special LUA_THASH type much like LUA_TNEGATIVENUMBER, instead of using LUA_TUSERDATA  and adding a userdata sub type, as is done for serializing hashes in values. This will make hash keys more compact. The alternative would be to add an extra SubType byte before each hash key, which would also often misalign the data.
* Incremented the version to 5 because version 4 can't read hash type keys.
* Changed the if structure for some version checks to not explicitly specify each version but use version ranges instead. This should perform slightly better at the cost of making it slightly harder to read. Additionally it's already done this way in other parts in script_table.cpp.
* Added a new test for size checking different key types.
* Updated MixedKeys test, added hash keys.

### DIFF
--- a/engine/script/src/script_table.cpp
+++ b/engine/script/src/script_table.cpp
@@ -27,32 +27,33 @@ extern "C"
 #include <lua/lualib.h>
 }
 
-// custom type when writing negative numbers as keys
+// custom type when writing negative numbers or hashes as keys
 // the rest of the types used when serializing a table come from lua.h
-// make sure this type has a value quite a bit higher than the types in lua.h
+// make sure these types have a value quite a bit higher than the types in lua.h
 #define LUA_TNEGATIVENUMBER 64
+#define LUA_THASH 65
 
 namespace dmScript
 {
     const int TABLE_MAGIC = 0x42544448;
-    const uint32_t TABLE_VERSION_CURRENT = 4;
+    const uint32_t TABLE_VERSION_CURRENT = 5;
 
     /*
      * Original table serialization format:
      *
      * uint16_t   count
      *
-     * char   key_type (LUA_TSTRING or LUA_TNUMBER)
+     * char   key_type (LUA_TSTRING, LUA_TNUMBER or LUA_THASH)
      * char   value_type (LUA_TXXX)
      * T      key (null terminated string or uint16_t)
      * T      value
      *
-     * char   key_type (LUA_TSTRING or LUA_TNUMBER)
+     * char   key_type (LUA_TSTRING, LUA_TNUMBER or LUA_THASH)
      * char   value_type (LUA_TXXX)
      * T      key (null terminated string or uint16_t)
      * T      value
      * ...
-     * if value is of type Vector3, Vector4, Quat, Matrix4 or Hash ie LUA_TUSERDATA, the first byte in value is the SubType
+     * if value is of type Vector3, Vector4, Quat, Matrix4 or Hash i.e. LUA_TUSERDATA, the first byte in value is the SubType
      *
      *    Version 1 table serialization format:
      *
@@ -85,6 +86,9 @@ namespace dmScript
      *
      *    Version 4:
      *    Adds support for more than 65535 keys in a table.
+     * 
+     *    Version 5:
+     *    Adds support for hash userdata type as keys.
      */
 
     struct TableHeader
@@ -183,6 +187,7 @@ namespace dmScript
         case 2:
         case 3:
         case 4:
+        case 5:
             supported = true;
             break;
         default:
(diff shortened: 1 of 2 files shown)


AGulev:
Related issue: https://github.com/defold/defold/issues/3208

engine/script/src/script_table.cpp
```diff
@@ -487,9 +496,15 @@ namespace dmScript
 
             int key_type = lua_type(L, -2);
             int value_type = lua_type(L, -1);
-            if (key_type != LUA_TSTRING && key_type != LUA_TNUMBER)
+
+            if (IsHash(L, -2))
```
JCash, line 500: use `dmScript::ToHash()` instead, to avoid getting the value twice.

engine/script/src/script_table.cpp
```diff
@@ -510,6 +525,22 @@ namespace dmScript
...
+            {
+                (*buffer++) = (char) LUA_THASH;
+                (*buffer++) = (char) value_type;
+            
+                dmhash_t *hash = (dmhash_t*)lua_touserdata(L, -2);
+                const uint32_t hash_size = sizeof(dmhash_t);
+
+                if (buffer_end - buffer < int32_t(hash_size))
+                {
+                    luaL_error(L, "buffer (%d bytes) too small for table, exceeded at key (%s) for element #%d", buffer_size, lua_typename(L, key_type), count);
```
JCash, line 538: Will `lua_typename()` work on our type? Or will it produce `null` or an error?

engine/script/src/script_table.cpp
```diff
@@ -510,6 +525,22 @@ namespace dmScript
                 (*buffer++) = (char) value_type;
                 buffer = WriteEncodedIndex(L, key, header, buffer, buffer_end);
             }
+            else if (key_type == LUA_THASH)
+            {
+                (*buffer++) = (char) LUA_THASH;
+                (*buffer++) = (char) value_type;
+            
+                dmhash_t *hash = (dmhash_t*)lua_touserdata(L, -2);
+                const uint32_t hash_size = sizeof(dmhash_t);
```
JCash, line 534: See comment above about dmScript::ToHash()

JCash (changes requested):
Lgtm.
Suggested small change to ToHash()

engine/script/src/script_table.cpp
britzl: Good point. I don't think lua_typename() will work for non standard types.

britzl (changes requested):
See lua_typename() concerns

engine/script/src/script_table.cpp
Gert-Jan: Makes sense, done!

engine/script/src/script_table.cpp
Gert-Jan: Right, this would likely result in an access violation.  Fixed by not using lua_typename() anymore, we already know it's a hash here.

Gert-Jan:
Also found a (self inflicted) bug in CheckTableSize(). Added a test for that.

engine/script/src/script_table.cpp
```diff
@@ -536,7 +536,7 @@ namespace dmScript
 
                 if (buffer_end - buffer < int32_t(hash_size))
                 {
-                    luaL_error(L, "buffer (%d bytes) too small for table, exceeded at key (%s) for element #%d", buffer_size, lua_typename(L, key_type), count);
+                    luaL_error(L, "buffer (%d bytes) too small for table, exceeded at key (hash) for element #%d", buffer_size, lua_typename(L, key_type), count);
```
JCash: Now it's another bug: Parsing the wrong parameter type for the string format.

JCash (changes requested):
still issue related to the lua_typename() string format

engine/script/src/script_table.cpp
```diff
@@ -339,7 +344,13 @@ namespace dmScript
...
-            if (key_type != LUA_TSTRING && key_type != LUA_TNUMBER)
+
+            if (IsHash(L, -2))
+            {
+                key_type = LUA_THASH;
+            }
+
+            if (key_type != LUA_TSTRING && key_type != LUA_TNUMBER && key_type != LUA_THASH)
             {
                 luaL_error(L, "keys in table must be of type number or string (found %s)", lua_typename(L, key_type));
```
britzl, line 355: The error message should now include hash as well.

```cpp
luaL_error(L, "keys in table must be of type number, string or hash (found %s)", lua_typename(L, key_type));
```

britzl (changes requested):
See note about error message

JCash (approved):

britzl (approved):

Issue #8504, opened 2024-02-05
Editor's scripting proposal


ekharkunov:
_Is it a private platform specific feature? (i.e. would it break your NDA?)<br>_
No

**Is your feature request related to a problem? Please describe (REQUIRED):**
Some of project/extensions should make assets generation/modification. Introduce new API which add possibility to generate/modify project's assets.

**Use cases:**
1. Editor scripts which update atlas with all images from folder
You add/remove image - with editor's script you list all images from folder and recreate atlas asset with new API.

2. Generate "spawner" objects
For example, you have some king of strategy game. In your game you can create new type of units manually and after that you should update game object that used to spawn new unit type in runtime. I mean you should add new factory/collectionfactory component to "spawner" object.

3. Extend previous case - generate game objects/collections from balance table
For example, you have some Excel/Google sheets tables that can be exported to csv and stored inside repository. During build phase you need to generate game objects/collections/other assets related to balance table. It helps to make assets to be consistent with balance tables.

4. Generate collection for level
For example, you used Photoshop to combine different sprites and create a game level. Use tool like PsdSplit to create a bunch of sprites and xml description with sprite position. After that use editor's script to create set of game objects, atlas with all sprites and level collection where you place game objects in the right place according to xml description.

**Describe the solution you'd like (REQUIRED):**
Expose more API in editor's script for:
1. Creating all type of assets: atlas, game object, collection, etc.
2. Modify assets: 
2a. add/remove components
2b. modify atlas content/settings
2c. modify game objects hierarchy inside collection
2d. change component properties
3. Iterating over nested data
4. Getting component/game object/piece of data via id (see example below)

**Describe alternatives you've considered (REQUIRED):**
You can use any external scripted or compiled language to create external tool to generate/modifying assets. For now workflow can be looked like:
1. Download defoldsdk.zip according to pinned version
2. Unzip archive
3. Generate code for certain language from proto definitions (if it scripting language like Python) (optional)
4. Run script or compiled program to make asset generation/modification

**Additional context (OPTIONAL):**
Some examples how it can be looks like:
1. Group game object as children of newly created empty game object
```
local work_collection = assets.get_collection("/main/test/main.collection")
local new_parent = assets.collection.create_game_object("new_parent_id")
for id, game_object in pairs(work_collection.get_game_objects()) do
    if string.find(id, "tile_") then
        assets.collection.set_parent(game_object, new_parent, {}) --- some additional options
    end
end
[...]

PR #10723, opened 2025-05-26
Edit particlefx emitters and modifiers using editor scripts

vlaaad:
With this change, it's now possible to edit particlefx emitters and modifiers using editor scripts. We added:
- `"emitters"` and `"modifiers"` properties on particlefx component
- `"modifiers"` property on particlefx emitter component
- ability to read and write particlefx curve properties in editor scripts

New `emitters` and `modifiers` properties are node lists, where each node may be accessed using `editor.get()` and modified using `editor.tx.set()`. For example, adding a new circle emitter works like this:
```lua
editor.transact({
    editor.tx.add("/fire.particlefx", "emitters", {
        type = "emitter-type-circle"
    })
})
```

Curves are represented as a table with a non-empty list of `points`, where each point is a table with the following properties:
- `x` - the x coordinate of the point, should start at 0 and end at 1
- `y` - the value of the point
- `tx` (0 to 1) and `ty` (-1 to 1) - tangents of the point. E.g., for an 80-degree angle, `tx` should be `math.cos(math.rad(80))` and `ty` should be `math.sin(math.rad(80))`.
Curve spreads additionally have a `spread` number property.

For example, setting a particle lifetime alpha curve might look like this:
```lua
editor.transact({
    editor.tx.set(emitter, "particle_key_alpha", { points = {
        {x = 0,   y = 0, tx = 0.1, ty = 1}, -- start at 0, go up quickly
        {x = 0.2, y = 1, tx = 1,   ty = 0}, -- reach 1 at 20% of a lifetime
        {x = 1,   y = 0, tx = 1,   ty = 0}  -- slowly go down to 0
    }})
})
```

Related to #8504

### DIFF
--- a/editor/src/clj/editor/particlefx.clj
+++ b/editor/src/clj/editor/particlefx.clj
@@ -16,6 +16,7 @@
   (:require [clojure.java.io :as io]
             [dynamo.graph :as g]
             [editor.app-view :as app-view]
+            [editor.attachment :as attachment]
             [editor.build-target :as bt]
             [editor.camera :as camera]
             [editor.colors :as colors]
@@ -277,9 +278,9 @@
 ; Modifier geometry
 
 (def drag-geom-data (let [top (->> geom/origin-geom
-                                (geom/transl [0 10 0])
-                                (geom/chain 4 (partial geom/transl [15 10 0])))
-                         bottom (geom/scale [1 -1 1] top)]
+                                   (geom/transl [0 10 0])
+                                   (geom/chain 4 (partial geom/transl [15 10 0])))
+                          bottom (geom/scale [1 -1 1] top)]
                       (interleave top bottom)))
 
 (def vortex-geom-data (let [ps (->> geom/origin-geom
@@ -388,7 +389,7 @@
   (property magnitude CurveSpread (default props/default-curve-spread))
   (property max-distance Curve (default props/default-curve)
             (dynamic visible (g/fnk [type] (contains? #{:modifier-type-radial :modifier-type-vortex} type))))
-  (property node-outline-key g/Str (dynamic visible (g/constantly false))) ; Always assigned in load-fn.
+  (property node-outline-key g/Str (dynamic visible (g/constantly false))) ; Either assigned in load-fn or generated when attached
 
   (output transform-properties g/Any scene/produce-unscalable-transform-properties)
   (output pb-msg g/Any produce-modifier-pb)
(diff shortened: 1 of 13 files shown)


engine/particle/src/particle.cpp
```diff
@@ -200,7 +200,18 @@ namespace dmParticle
...
+        for (uint32_t i = 0; i < emitter_ddf->m_Properties.m_Count; ++i)
+        {
+            const dmParticleDDF::Emitter::Property& property = emitter_ddf->m_Properties[i];
+            if (property.m_Key == EMITTER_KEY_SPAWN_RATE)
+            {
+                spawn_rate_spread = dmMath::Rand11(&seed) * property.m_Spread;
+                break;
+            }
+        }
+        emitter->m_SpawnRateSpread = spawn_rate_spread;
```
vlaaad, line 214: The old code was wrong: it accessed the `emitter_ddf->m_Properties` array at index 0 (`EMITTER_KEY_SPAWN_RATE` value) to get the spawn rate during initialization. The `m_Properties` array was not guaranteed to have the spawn rate property as the first value, though in practice it always had. But the editor excludes entries with 0 values, so it could be omitted! In that case, the engine would use some other unrelated value as a spawn rate. I discovered this issue when I started implementing particlefx editing using editor scripts, and it filled all properties with zeroes — this caused the `m_Properties` to be empty, and the engine (and editor) hard crashed because array access has an assert for index access. 

I fixed the code, and that's the first C++ code I ever committed, so I can't vouch for it to be correct, but the idea is to find the spawn rate entry in the array instead of doing array access.

editor/test/integration/editor_extensions_test.clj
```diff
@@ -1119,15 +1119,12 @@ GET /test/resources/test.json as json => 200
                             :properties
                             vals
                             (map #(assoc % :outline outline)))))
-             (keep (fn [{:keys [outline key edit-type] :as p}]
+             (run! (fn [{:keys [outline key] :as p}]
                      (let [{:keys [node-id]} outline
                            ext-key (string/replace (name key) \- \_)]
-                       (when-not (contains? #{:editor.properties.CurveSpread :editor.properties.Curve} (:k (:type edit-type)))
```
vlaaad, line 1125: no we support all properties!

editor/src/clj/editor/particlefx.clj
```diff
@@ -277,9 +278,9 @@
 ; Modifier geometry
 
 (def drag-geom-data (let [top (->> geom/origin-geom
-                                (geom/transl [0 10 0])
-                                (geom/chain 4 (partial geom/transl [15 10 0])))
-                         bottom (geom/scale [1 -1 1] top)]
+                                   (geom/transl [0 10 0])
+                                   (geom/chain 4 (partial geom/transl [15 10 0])))
+                          bottom (geom/scale [1 -1 1] top)]
```
vlaaad, line 283: Just an indentation change.

editor/src/clj/editor/particlefx.clj
```diff
@@ -437,36 +438,36 @@
...
-            (let [gpu-texture (:gpu-texture emitter-sim-data)
-                  vtx-binding (vtx/use-with context vbuf shader)
-                  blend-mode (convert-blend-mode (:blend-mode render-data))]
-              (gl/with-gl-bindings gl render-args [shader vtx-binding gpu-texture]
-                (shader/set-samplers-by-index shader gl 0 (:texture-units gpu-texture))
-                (gl/set-blend-mode gl blend-mode)
-                (gl/gl-draw-arrays gl GL/GL_TRIANGLES (:v-index render-data) (:v-count render-data))
-                (.glBlendFunc gl GL/GL_SRC_ALPHA GL/GL_ONE_MINUS_SRC_ALPHA)))))))))
+    (let [{:keys [color emitter-index emitter-sim-data material-attribute-infos max-particle-count vertex-attribute-bytes]} (:user-data renderable)]
+      (when-let [shader (:shader emitter-sim-data)]
```
vlaaad, line 442: Mostly indentation, but I check that the shader exists, so it doesn't spam errors in logs when there is no material.

editor/src/clj/editor/particlefx.clj
```diff
@@ -782,6 +792,7 @@
 (def ^:private default-start-delay (pair pb-default-start-delay pb-default-start-delay-spread))
 
 (g/defnode EmitterNode
+  (inherits core/Scope)
```
vlaaad, line 795: I refactored the graph a bit to simplify wiring. Previously, all modifiers were connected directly to the ParticleFX node, even when they were children of emitter nodes. I made EmitterNode a scope node and started connecting emitter modifiers here instead. This change simplified the code around connecting modifiers.

editor/src/clj/editor/particlefx.clj
```diff
@@ -1028,36 +1050,55 @@
...
-                                                                                      (attach-emitter self-id child-id true))}
-                                                                     {:node-type ModifierNode
-                                                                      :tx-attach-fn (fn [self-id child-id]
-                                                                                      (attach-modifier self-id self-id child-id true))}]})))
+  (output node-outline outline/OutlineData :cached
+          (g/fnk [^:unsafe _evaluation-context _node-id child-outlines]
+            ;; We use unsafe evaluation context to get child node types: these
+            ;; should never change
+            (let [basis (:basis _evaluation-context)
+                  outlines (group-by #(g/node-instance? basis ModifierNode (:node-id %)) child-outlines)
```
vlaaad, line 1073: mostly indentation, but using evaluation context here is important since it allows evaluating the `node-outline` more reliably (e.g., it works with evaluation context created in transaction, when children were also created in the same transaction).

JCash (approved):

matgis (approved):
LGTM! ❤️

PR #10730, opened 2025-05-26
Improve `hash` management when creating objects using collectionfactory

### LINKED ISSUE #10704: Invalid object id when created from collectionfactory inside proxy collection
I have `unknown` or some `integer` value as object id when use `collectionfactory.create` inside proxy collection. It happens if proxy was loaded-unloaded and loaded again, as if player go to level - main menu - level.

**To Reproduce (REQUIRED)**
1. Create prefab collection (e.g. item).
2. Create collection (e.g. level) and load it using proxy. Inside this collection spawn several objects using `collectionfactory.create` and prefab from step 1. Print id's of the created objects, or try to move them (go.set_position).
3. Unload proxy from step 2.
4. Load proxy collection again.
5. When `collectionfactory.create` spawn new objects their IDs could be not valid.
For example:
Valid ID: `hash: [/item] = hash: [/collection0/item],`
Invalid ID: `hash: [/item] = 17,`
Invalid ID: `hash: [/item] = hash [<unknown:5348857206967936022>],`

`collectionfactory.create` always returns valid object ids.

  - Version 1.10.1 - has issue
  - Version 1.10.2-alpha (d91ec4edb2dd5e025828ce896daecc69cd81fd3f) - has issue
  - Veersion 1.10.0 - no issue

**Platforms (REQUIRED):**
 - Platforms: macOS
 - OS: 15.4.1 (24E263)
 - Device: MacBook Air M2

Steps:
1. Project load proxy with "menu" collection.
2. Press button "Level" to load proxy collection.
3. Notice that there is no errors, all collections are spawned ok.
4. Click on screen to return to "menu".
5. Click "Level" button again.
6. Notice, that not all objects spawned correctly. In console you'll see that some objects has invalid IDs. If I try to use this id in `go.set_position` there is an error:
`ERROR:SCRIPT: main/level/level.script:9: Instance 17 not found`

[collection-create.zip](https://github.com/user-attachments/files/20346737/collection-create.zip)

**Logs (OPTIONAL):**

**Workaround (OPTIONAL):**

**Screenshots (OPTIONAL):**

AGulev:
Fix issue where `hash` can be removed while it's in use

Fix https://github.com/defold/defold/issues/10704

### DIFF
--- a/engine/script/src/script.cpp
+++ b/engine/script/src/script.cpp
@@ -222,10 +222,6 @@ namespace dmScript
         // [-2] context_table
         // [-1] weak_table
 
-        // context_table["__weak"] = weak_table
-        lua_pushvalue(L, -1);        // duplicate weak_table for storing in context
-        lua_setfield(L, -3, "__weak");  // context_table["__weak"] = weak_table
-
         // Now store weak_table ref separately
         context->m_ContextWeakTableRef = Ref(L, LUA_REGISTRYINDEX);
 
--- a/engine/script/src/script_ddf.cpp
+++ b/engine/script/src/script_ddf.cpp
@@ -547,16 +547,16 @@ namespace dmScript
                     else if (d->m_NameHash == DDF_TYPE_NAME_HASH_LUAREF)
                     {
                         dmScriptDDF::LuaRef* lua_ref = (dmScriptDDF::LuaRef*) ptr;
-                        lua_rawgeti(L, LUA_REGISTRYINDEX, lua_ref->m_ContextTableRef);
-                        lua_rawgeti(L, -1, lua_ref->m_Ref);
-                        if (lua_isnil(L, -1))
+                        if (lua_ref->m_Ref)
                         {
-                            lua_pop(L, 1); // remove nil
-                            lua_getfield(L, -1, "__weak"); // context_table["__weak"]
+                            lua_rawgeti(L, LUA_REGISTRYINDEX, lua_ref->m_ContextTableRef);
                             lua_rawgeti(L, -1, lua_ref->m_Ref);
-                            lua_remove(L, -2); // remove weak table
+                            lua_remove(L, -2);
                         }
-                        lua_remove(L, -2); // remove context table
+                        else
+                        {
+                            lua_pushnil(L);
+                        }
                     }
                     else
                     {
--- a/engine/script/src/script_hash.cpp
+++ b/engine/script/src/script_hash.cpp
@@ -393,9 +393,18 @@ namespace dmScript
         if (context)
         {
             int* refp = context->m_HashInstances.Get(hash);
-            if (refp)
+            if (refp && context->m_ContextWeakTableRef != LUA_NOREF)
             {
-                context->m_HashInstances.Erase(hash);
+                lua_rawgeti(L, LUA_REGISTRYINDEX, context->m_ContextWeakTableRef);
+                // [-1] weak_table
+                lua_rawgeti(L, -1, (lua_Integer)hash);
+                // [-2] weak_table
+                // [-1] hash
+                if (lua_isnil(L, -1))
+                {
+                    context->m_HashInstances.Erase(hash);
+                }
+                lua_pop(L, 2);
             }
         }
         return 0;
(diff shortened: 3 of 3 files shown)


engine/script/src/script.cpp
```diff
@@ -222,10 +222,6 @@ namespace dmScript
         // [-2] context_table
         // [-1] weak_table
 
-        // context_table["__weak"] = weak_table
-        lua_pushvalue(L, -1);        // duplicate weak_table for storing in context
-        lua_setfield(L, -3, "__weak");  // context_table["__weak"] = weak_table
```
AGulev, line 227: We don't need it really

engine/script/src/script_ddf.cpp
```diff
@@ -547,16 +547,16 @@ namespace dmScript
                     else if (d->m_NameHash == DDF_TYPE_NAME_HASH_LUAREF)
                     {
                         dmScriptDDF::LuaRef* lua_ref = (dmScriptDDF::LuaRef*) ptr;
-                        lua_rawgeti(L, LUA_REGISTRYINDEX, lua_ref->m_ContextTableRef);
-                        lua_rawgeti(L, -1, lua_ref->m_Ref);
-                        if (lua_isnil(L, -1))
+                        if (lua_ref->m_Ref)
```
AGulev, line 550: revert to old logic. It shouldn't be touched at all

engine/script/src/script_hash.cpp
```diff
@@ -142,6 +142,57 @@ namespace dmScript
         return 1;
     }
 
+    static int CreateLuaHashUserdata(lua_State* L, dmhash_t hash)
```
AGulev, line 145: 1. Create hash
2. Put it into registry
3. Save `ref` in instances
4. Put it into weak table with `(lua_Integer)hash` as a key

engine/script/src/script_hash.cpp
```diff
@@ -150,51 +201,35 @@ namespace dmScript
...
         int* refp = instances->Get(hash);
+
         if (refp)
         {
-            lua_rawgeti(L, LUA_REGISTRYINDEX, context->m_ContextTableRef);
-            // [-1] Context table
-            lua_rawgeti(L, -1, *refp);
-            // [-2] Context table
-            // [-1] hash
+            lua_rawgeti(L, LUA_REGISTRYINDEX, context->m_ContextWeakTableRef);
```
AGulev, line 207: Always, check weak table. Registry needed only to increase ref counter. `instances` needed to know if the object still alive.

engine/script/src/script_hash.cpp
```diff
@@ -150,51 +201,35 @@ namespace dmScript
...
-                lua_remove(L, -2);
-            }
+                lua_pop(L, 1); // remove nil
+                //[-1] weak_table
 
-            lua_remove(L, -2);
-            // [-1] hash
+                // Create new userdata and re-register it
+                lua_pop(L, 1); // remove weak_table
+                CreateLuaHashUserdata(L, hash);
```
AGulev, line 222: if weak table is empty - create a new hash

engine/script/src/script_hash.cpp
```diff
@@ -214,15 +249,14 @@ namespace dmScript
...
-
         // Retrieve the value from the strong table
         lua_rawgeti(L, LUA_REGISTRYINDEX, context->m_ContextTableRef);
         // [-1] context_table
         lua_rawgeti(L, -1, ref);
         // [-2] context_table
         // [-1] value
 
-        if (lua_isnil(L, -1))
+        if (!lua_isuserdata(L, -1))
```
AGulev, line 259: it may be nil or a number, because that's how registry work (something like a linked list of empty slots)

engine/script/src/script_hash.cpp
```diff
@@ -376,14 +393,18 @@ namespace dmScript
...
                 lua_rawgeti(L, LUA_REGISTRYINDEX, context->m_ContextWeakTableRef);
-                lua_pushinteger(L, *refp);
-                lua_pushnil(L);
-                lua_settable(L, -3);
-                lua_pop(L, 1);
+                // [-1] weak_table
+                lua_rawgeti(L, -1, (lua_Integer)hash);
+                // [-2] weak_table
+                // [-1] hash
+                if (lua_isnil(L, -1))
```
AGulev, line 403: Remove only if it's not recreated

JCash (approved):

PR #10731, opened 2025-05-27
Fix issue with gui order render order

AGulev:


engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -1030,12 +1030,13 @@ namespace dmGameSystem
         return (dmGameObject::HComponent)params.m_UserData;
     }
 
+    // we can use only 24 bits, take a look render_private.h -> RenderListSortValue
```
AGulev, line 1033: This is a bit hidden knowlege, So I thought I can use all 32 bits what I did here: https://github.com/defold/defold/pull/10631/files

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -1030,12 +1030,13 @@ namespace dmGameSystem
...
     inline uint32_t MakeFinalRenderOrder(uint32_t world_order, uint32_t scene_order, uint32_t sub_order)
     {
         // scene_order: 4 bits (max 15)
-        // world_order: 12 bits (max 4095)
-        // sub_order:   16 bits (max 65535)
-        return (scene_order << 28) | (world_order << 16) | sub_order;
+        // world_order: 7 bits (max 127)
+        // sub_order:   13 bits (max 8191)
```
AGulev, line 1038: It is basically draw calls count for all GUIs, I think it is enough to have 8192

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -3158,9 +3159,9 @@ namespace dmGameSystem
 
         // TODO: it seems like this is a hidden game.project property which should be gui.world_count like it is in physics
         int32_t max_world_count = dmConfigFile::GetInt(ctx->m_Config, "gui.max_instance_count", 128);
-        if (max_world_count > 4096)
+        if (max_world_count > 128)
```
AGulev, line 3162: Since this is a hidden option nobody knows about, we can make the default the maximum, because nobody uses more. This is the count of simultaneously existing GUI worlds—a world is similar to one in physics; it’s a separate batch of components in a collection proxy.

britzl (approved):

PR #10733, opened 2025-05-27
Edit collision shapes using editor scripts

vlaaad:
It is now possible to edit collision shapes using editor scripts! To support this, we added new `"shapes"` node list property to collision objects. For example, adding a new shape is done like this:
```lua
editor.transact({
    editor.tx.add("/hero.collisionobject", "shapes", {
        type = "shape-type-box" -- or "shape-type-sphere", "shape-type-capsule"
    })
})
```

To create a new collision shape, it is necessary to specify the type of the shape. The available types are:
- `shape-type-box` - box shape with `dimensions` property
- `shape-type-sphere` - sphere shape with `diameter` property
- `shape-type-capsule` - capsule shape with `diameter` and `height` properties

Related to #8504

### DIFF
--- a/editor/src/clj/editor/attachment.clj
+++ b/editor/src/clj/editor/attachment.clj
@@ -20,7 +20,7 @@
             [util.coll :as coll]))
 
 (defonce ^:private state-atom
-  ;; :add -> type -> list-kw -> type -> {:node-type ... :tx-attach-fn ...}
+  ;; :add -> type -> list-kw -> type -> tx-attach-fn
   ;; :get -> type -> list-kw -> get-fn
   ;;
   ;; tx-attach-fn: fn of parent-node, child-node -> txs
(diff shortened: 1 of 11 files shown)


editor/src/clj/editor/attachment.clj
```diff
@@ -79,10 +81,13 @@
...
                        is a list-kw keyword, and second is a map with the
                        following keys:
-                         :init    required, data structure that will be supplied
-                                  to the init-fn to create element
-                                  initialization transaction steps
-                         :add     optional, attachment tree of the created item
+                         :init         required, data structure that will be
+                                       supplied to the init-fn to create element
+                                       initialization transaction steps
+                         :node-type    required, node type of child node
```
vlaaad, line 87: Now we require a node type here. Editor scripts, though, only require a node type when there is more than 1 alternative type registered.

editor/src/clj/editor/collision_object.clj
```diff
@@ -652,14 +667,24 @@
...
 
+(attachment/register!
+  CollisionObjectNode :shapes
+  :add {SphereShape attach-shape-node
+        BoxShape attach-shape-node
+        CapsuleShape attach-shape-node}
+  :get attachment/nodes-getter)
+(node-types/register-node-type-name! SphereShape "shape-type-sphere")
+(node-types/register-node-type-name! BoxShape "shape-type-box")
+(node-types/register-node-type-name! CapsuleShape "shape-type-capsule")
```
vlaaad, line 686: Also, we define explicit extension type names for editor node types.

editor/src/clj/editor/editor_extensions/graph.clj
```diff
@@ -388,68 +410,103 @@
...
+      (attachment->set-tx-steps child-node-id rt project evaluation-context)))
+
+(def ^:private default-sphere-diameter (rt/->lua 20.0))
+(defmethod init-attachment :editor.collision-object/SphereShape [evaluation-context rt project _ _ child-node-id attachment]
+  (concat
+    (g/set-property child-node-id :shape-type :type-sphere)
+    (-> attachment
+        (util/provide-defaults
+          "diameter" default-sphere-diameter)
+        (attachment->set-tx-steps child-node-id rt project evaluation-context))))
```
vlaaad, line 457: Refactoring of `init-attachment` was needed to support initializing the collision shapes: we set the `shape-type` property without exposing it to editor scripts, since it shouldn't be set after node construction.

matgis (dismissed):
Looks good to me! I had some questions about using node types as keys, though.

PR #10734, opened 2025-05-27
Remove all dynamically created hashes (from factories) when they aren't used anymore

AGulev:


engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -1976,7 +1979,7 @@ namespace dmGameObject
         dmHashRelease64(&instance->m_CollectionPathHashState);
         if(instance->m_Generated)
         {
-            dmHashReverseErase64(instance->m_Identifier);
+            dmScript::ReleaseHash(dmScript::GetLuaState(g_ScriptContext), instance->m_Identifier);
```
AGulev, line 1982: Delete hash on deletion of all dynamically generated game objects (spawned with a factory)

engine/gameobject/src/gameobject/gameobject_script.cpp
```diff
@@ -1872,10 +1868,6 @@ namespace dmGameObject
             DM_HASH_REVERSE_MEM(hash_ctx, 256);
             return luaL_error(L, "Can not delete subinstances of spine or model components. '%s'", dmHashReverseSafe64Alloc(&hash_ctx, dmGameObject::GetIdentifier(instance)));
         }
-        if (instance->m_Generated)
-        {
-            dmScript::ReleaseHash(L, instance->m_Identifier);
-        }
```
AGulev, line 1878: one more sub ccase.

JCash (approved):

PR #10736, opened 2025-05-27
Various sprite component optimizations

ekharkunov:
The goal was:
* restructure data to enhance data locality
* compress result structure of SpriteComponent to decrease memory footprint
* decrease time spent on `CompSpriteUpdate` and `CompSpriteRender`

### Caching and data locality
* Added fields with vertex count and indices count in AnimationData. Helps avoid chain AnimationData->SpriteResource->TextureSetResource->....
* Cache flag `m_CanUseQuads`. `m_UsesGeometry` is used only in one place (`CanUseQuads` function). So from memory point of view struct size isn't changed but we avoid chain AnimationData->m_Geometries->m_TrimMode
* Store number of textures inside SpriteComponent because it's often used during the frame
* Store animation playback mode inside SpriteComponent. It's also often used during the frame. Removed 2 bits flags which can be deduced from playback mode. Playback mode defined as 7 bits field because maximum value in the Playback enum is 7

### Structure compressing
* Narrow `m_AnimationID` and `m_DynamicVertexAttributeIndex` from uint32_t to uint16_t. I believe it should be enough to store possible values. In some places dynamic vertex attribute index already passed as uint16_t
* Remove 2 flags which can be deduced from playback mode

### Execution time optimization
* Instead of executes 3 cycles over all sprite component and execute logic I merged all in one cycle which contains all necessary logic for every component. Helps avoid fetch data into CPU cache line several time during the frame.
* Move some logic from `CompSpriteRender` to `CompSpriteUpdate` to include it into one big cycle over all component

### Profiling results
I used Time profiler from Xcode instruments. It doesn't provide exact execution time. Just amount of samples where function was appeared. I calculated percentage of samples where CompSpriteUpdate and CompSpriteRender were appeared.

Screenshot 2025-09-21 at 21 32 13

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1665,64 +1688,19 @@ namespace dmGameSystem
...
-                float radius_sq = dmVMath::LengthSqr((c->m_World.getCol(0).getXYZ() + c->m_World.getCol(1).getXYZ()) * 0.5f);
-                sprite_world->m_BoundingVolumes[i] = radius_sq;
-            }
-        }
-
-        // The "sub_pixels" is set by default
+        Matrix4 local = dmTransform::ToMatrix4(dmTransform::Transform(component->m_Position, component->m_Rotation, 1.0f));
+        Matrix4 world = dmGameObject::GetWorldMatrix(component->m_Instance);
+        Vector3 size( component->m_Size.getX() * component->m_Scale.getX(), component->m_Size.getY() * component->m_Scale.getY(), 1);
+        Matrix4 w = scale_along_z ? world * local : dmTransform::MulNoScaleZ(world, local);
```
ekharkunov, line 1696: Collapse the same code into smaller section and use here a ternary operator. I believe that CPU should make branch prediction optimization here because `scale_along_z` and `sun_pixels` doesn't change during the frame.

ekharkunov: In any case it will be merge conflict with your changes (for example, I moved bounding box assignment out of the function). I suggest to discuss the order of merging. You will be the first then I will merge my changes (if changes will pass the review)

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -2180,12 +2120,17 @@ namespace dmGameSystem
     {
         float cursor = component->m_AnimTimer;
 
-        if (component->m_AnimBackwards)
+        if (component->m_AnimationPlayback == dmGameSystemDDF::PLAYBACK_ONCE_BACKWARD 
+            || component->m_AnimationPlayback == dmGameSystemDDF::PLAYBACK_LOOP_BACKWARD)
```
JCash: Optional improvement:`Playback playback = component->m_AnimationPlayback`

This tells the compiler that the value is fixed during the entire function, as opposed to making the pointer dereference for each case.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -145,19 +145,22 @@ namespace dmGameSystem
...
         uint16_t                    m_FlipVertical : 1;
         uint16_t                    m_AddedToUpdate : 1;
         uint16_t                    m_ReHash : 1;
         uint16_t                    m_UseSlice9 : 1;
         uint16_t                    m_AnimationReHash : 1;
-        uint16_t                    : 5;
+        uint16_t                    m_IsPlaying : 1;
+        // currently we don't support multiple animation cursors that's why we can use Playback from the first
+        // texture set
+        dmGameSystemDDF::Playback   m_AnimationPlayback : 7;
```
JCash: I'm not convinced this becomes a "uint16_t", as it's an enum.
I think it's more likely it beomces an int, in which case your bitcounts are off.
Better to make sure to use a uint16_t:7 instead.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -532,21 +530,20 @@ namespace dmGameSystem
...
         TextureSetResource* texture_set = GetFirstTextureSet(component);
         dmGameSystemDDF::TextureSet* texture_set_ddf = texture_set->m_TextureSet;
-        dmGameSystemDDF::TextureSetAnimation* animation_ddf = &texture_set_ddf->m_Animations[component->m_AnimationID];
 
         // Set frame from cursor (tileindex or animframe)
         float t = component->m_AnimTimer;
-        float backwards = (animation_ddf->m_Playback == dmGameSystemDDF::PLAYBACK_ONCE_BACKWARD
-                        || animation_ddf->m_Playback == dmGameSystemDDF::PLAYBACK_LOOP_BACKWARD) ? 1.0f : 0;
+        float backwards = (component->m_AnimationPlayback == dmGameSystemDDF::PLAYBACK_ONCE_BACKWARD
+                        || component->m_AnimationPlayback == dmGameSystemDDF::PLAYBACK_LOOP_BACKWARD) ? 1.0f : 0;
```
JCash: Same as comment below (regarding getting a variable to avoid dereferncing).
Probably less important here, as we are storing values in component as well.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -134,9 +136,7 @@ namespace dmGameSystem
         uint32_t                    m_MixedHash;
         uint32_t                    m_AnimationDataHash;
 
-        uint32_t                    m_AnimationID; // index into array
-        uint32_t                    m_DynamicVertexAttributeIndex;
-
+        int32_t                     m_AnimationInterval; // cached value when animation is playing
```
JCash: I'd prefer "m_AnimationFrameCount" instead.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -671,6 +669,8 @@ namespace dmGameSystem
         component->m_Scale = params.m_Scale;
         SpriteResource* resource = (SpriteResource*)params.m_Resource;
         component->m_Resource = resource;
+        // narrow to 8 bits because I don't believe that sprite can has more than 255 textures
```
JCash: 👍 

typo "can has" -> "can have"

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1085,7 +1087,36 @@ namespace dmGameSystem
             uses_geometries |= data->m_Geometries[i]->m_TrimMode != dmGameSystemDDF::SPRITE_TRIM_MODE_OFF;
         }
 
-        data->m_UsesGeometries = uses_geometries;
+        data->m_CanUseQuads = !uses_geometries || data->m_Geometries[0]->m_TrimMode == dmGameSystemDDF::SPRITE_TRIM_MODE_OFF;
```
JCash, line 1093: 👍

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1665,64 +1690,19 @@ namespace dmGameSystem
         dmRender::AddToRender(render_context, &ro);
     }
 
-    static void UpdateTransforms(SpriteWorld* sprite_world, bool sub_pixels)
+    static void UpdateTransforms(SpriteComponent* component, bool scale_along_z, bool sub_pixels)
```
JCash: Now that it only deals with a single component: "UpdateTransforms()" -> "UpdateTransform()"

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1665,64 +1690,19 @@ namespace dmGameSystem
...
-                SpriteComponent* c = &components[i];
-                Vector4 position = c->m_World.getCol3();
-                position.setX((int) position.getX());
-                position.setY((int) position.getY());
-                c->m_World.setCol3(position);
-            }
+            Vector4 position = component->m_World.getCol3();
+            position.setX((int) position.getX());
+            position.setY((int) position.getY());
+            component->m_World.setCol3(position);
```
JCash: Btw, have you profiled with subpixels on?

This is also part of the hot path, so I wouldn't store the matrix until it's done.
E.g. something like:
```
        w = dmVMath::AppendScale(w, size);
        if (!sub_pixels)
        {
            Vector4 position = w.getCol3();
            position.setX((int) position.getX());
            position.setY((int) position.getY());
            w.setCol3(position);
        }
        component->m_World = w;
```

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1665,64 +1690,19 @@ namespace dmGameSystem
         dmRender::AddToRender(render_context, &ro);
     }
 
-    static void UpdateTransforms(SpriteWorld* sprite_world, bool sub_pixels)
+    static void UpdateTransforms(SpriteComponent* component, bool scale_along_z, bool sub_pixels)
     {
-        DM_PROFILE("UpdateTransforms");
```
JCash, line 1670: Why the removal of profiling scopes?
I don't see them replaced anywhere?

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1983,18 +1883,69 @@ namespace dmGameSystem
...
+        bool scale_along_z = dmGameObject::ScaleAlongZ(dmGameObject::GetCollection(c->m_Instance));
+        bool sub_pixels = sprite_context->m_Subpixels;
 
-        SpriteContext* sprite_context = (SpriteContext*)params.m_Context;
-        dmRender::TrimBuffer(sprite_context->m_RenderContext, world->m_VertexBuffer);
-        dmRender::RewindBuffer(sprite_context->m_RenderContext, world->m_VertexBuffer);
+        for (uint32_t i = 0; i < n; ++i)
+        {
+            SpriteComponent* component = &components[i];
+            // Note: We update all sprites, even though they might be disabled, or not added to update
```
JCash: Note that this comment refers to the previous implementation, where we tried to do a cache friendly loop.
Now, that case isn't the same, as we call `continue` within this loop, so we can skip this work now.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1983,18 +1883,69 @@ namespace dmGameSystem
...
+            UpdateTransforms(component, scale_along_z, sub_pixels);
+            // we need to consider the full scale here
+            // I.e. we want the length of the diagonal C, where C = X + Y
+            float radius_sq = dmVMath::LengthSqr((component->m_World.getCol(0).getXYZ() + component->m_World.getCol(1).getXYZ()) * 0.5f);
+            world->m_BoundingVolumes[i] = radius_sq;
+
+            if (!component->m_Enabled)
+                continue;
```
JCash: Add to top of loop.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1983,18 +1883,69 @@ namespace dmGameSystem
...
+            Animate(component, params.m_UpdateContext->m_DT);
+            if (!component->m_AddedToUpdate)
+                continue;
```
JCash: Add to top of loop

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -2204,6 +2149,9 @@ namespace dmGameSystem
 
     static inline float GetAnimationFrameCount(SpriteComponent* component)
     {
+        // don't use cached m_AnimationInterval here because GetAnimationFrameCount
+        // can be called without start playing animation as a result m_AnimationInterval is not
+        // updated to actual length of animation
```
JCash: Good comment!

Can we detect this?
E.g. if `m_AnimationInterval == 0` (uninitialized)

Or, can we update `m_AnimationInterval` when the component is created?

JCash (changes requested):
Overall very nnice!
Some room for improvement I believe.

Also, could you add some of the profile data in the PR comments?

engine/gamesys/src/gamesys/components/comp_sprite.cpp
ekharkunov: I've removed profiler scope because adding and removing scope inside the loop spend pretty much time. Yesterday I reveal a `DM_PROFILE_DYN` macro which I believe can be an option here. But I haven't tried it yet.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
ekharkunov: > Or, can we update m_AnimationInterval when the component is created?

No, because frame count depends on animation id. `m_IsPlaying` isn't reliable also because for `*ONCE` playback it can be stopped already when we ask for frame count.

> Can we detect this?
E.g. if m_AnimationInterval == 0 (uninitialized)

Let me look and try something like this.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
ekharkunov: > Btw, have you profiled with subpixels on?

No... Will do it as part of the next iteration

engine/gamesys/src/gamesys/components/comp_sprite.cpp
JCash: Dyn is for dynamic names. Shouldn't be needed here though.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
JCash: >No, because frame count depends on animation id

We already call "`PlayAnimation()` (which calls UpdateCurrentAnimationFrame()) in `CompSpriteCreate()`, so I see no reason why we shouldn't be able to update the m_AnimationFrameCount at that time?

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1697,7 +1697,8 @@ namespace dmGameSystem
     {
         Matrix4 local = dmTransform::ToMatrix4(dmTransform::Transform(component->m_Position, component->m_Rotation, 1.0f));
         Matrix4 world = dmGameObject::GetWorldMatrix(component->m_Instance);
-        Vector3 size( component->m_Size.getX() * component->m_Scale.getX(), component->m_Size.getY() * component->m_Scale.getY(), 1);
+        Vector3 size = dmVMath::MulPerElem(component->m_Size, component->m_Scale);
+        size.setZ(1.f);//(component->m_Size.getX() * component->m_Scale.getX(), component->m_Size.getY() * component->m_Scale.getY(), 1);
```
JCash: dead code

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -2386,14 +2382,38 @@ namespace dmGameSystem
             if (res == dmGameObject::PROPERTY_RESULT_OK)
             {
                 TextureSetResource* texture_set = GetFirstTextureSet(component);
-                uint32_t* anim_id = texture_set ? texture_set->m_AnimationIds.Get(component->m_CurrentAnimation) : 0;
+                dmhash_t current_animation = component->m_CurrentAnimation;
+                uint32_t* anim_id = texture_set ? texture_set->m_AnimationIds.Get(current_animation) : 0;
+                if (!anim_id)
+                {
+                    // it means that new atlas doesn't contain animation with the same name as it played before
+                    const void* data = dmHashReverse64(current_animation, 0);
```
JCash: This won't work in release builds, as the reverse hash table is disabled.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -2386,14 +2382,38 @@ namespace dmGameSystem
...
+                        const char* error_message = (const char*)data;
+                        dmHashTable64<uint32_t>::Iterator animation_iterator = texture_set->m_AnimationIds.GetIterator();
+                        if (animation_iterator.Next())
+                        {
+                            current_animation = animation_iterator.GetKey();
+                            anim_id = const_cast<uint32_t*>(&animation_iterator.GetValue());
+
+                            const void* new_animation_name = dmHashReverse64(current_animation, 0);
+                            dmLogError("Atlas doesn't contains animation '%s'. Animation '%s' will be used", error_message, new_animation_name != 0 ? (const char*)new_animation_name : "<unknown>");
+                        }
```
JCash: I think the fallback behavior should be the same in debug and release builds.

JCash (changes requested):
I think the fallback logic should work the same in release builds as well.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -2388,24 +2388,21 @@ namespace dmGameSystem
...
-                        }
-                        else
-                        {
-                            dmLogError("Atlas doesn't contains animation '%s'. No animation will be used", error_message);
-                        }
-                        // else anim_id still == 0
+                        current_animation = animation_iterator.GetKey();
+                        anim_id = const_cast<uint32_t*>(&animation_iterator.GetValue());
+
+                        const void* new_animation_name = dmHashReverse64(current_animation, 0);
```
JCash: Prefer dmHashReverseSafe64() which always returns a valid string.

JCash (approved):

PR #10743, opened 2025-05-28
Set drop position to z-plane intersection

sprocketc:
When we drop a collection to a scene, we should probably use the intersection with the z-plane to position the element, in order to have the Z value set to zero by default.

Reported [here](https://forum.defold.com/t/defold-1-10-2-beta/80587/5?)


matgis (approved):
Ship it! ❤️

PR #10752, opened 2025-05-29
Fix issue where profiler doesn't work after reboot

### LINKED ISSUE #9427: Profiler is not working if build was relaunched by Defold Editor

### LINKED ISSUE #10671: Rebuild game from Editor and sys.reboot breaks Profiler

AGulev:
Make sure that the profiler shows the correct scope and property names after the engine reboot.

Fix https://github.com/defold/defold/issues/10671
Fix https://github.com/defold/defold/issues/9427


engine/engine/src/engine.cpp
```diff
@@ -1857,7 +1874,7 @@ namespace dmEngine
 
                 if (esc_pressed || !dmGraphics::GetWindowStateParam(engine->m_GraphicsContext, dmPlatform::WINDOW_STATE_OPENED))
                 {
-                    engine->m_Alive = false;
+                    Exit(engine, 0);
```
AGulev, line 1877: we need to use `Exit` here so `engine->m_RunResult.m_Action` updated as well

engine/engine/src/engine.h
```diff
@@ -23,12 +23,12 @@ namespace dmEngine
...
-	    RESULT_REBOOT   =  1,
-	    RESULT_EXIT     = -1,
-	};
+    enum UpdateResult
+    {
+        RESULT_OK       =  0,
+        RESULT_REBOOT   =  1,
+        RESULT_EXIT     = -1,
+    };
 };
```
AGulev, line 32: Tabs -> spaces

engine/extension/src/dmsdk/extension/extension.h
```diff
@@ -98,11 +98,28 @@ typedef enum ExtensionCallbackType
...
+    EXTENSION_APP_REBOOT   =  1,
+    EXTENSION_APP_EXIT     = -1,
+} AppExitStatus;
 
 typedef struct ExtensionAppParams
 {
-    HConfigFile m_ConfigFile; // Deprecated. Here for backwards compatibility
-
+    HConfigFile     m_ConfigFile; // Deprecated. Here for backwards compatibility
+    AppExitStatus   m_ExitStatus;
```
AGulev, line 122: If/when I merge this PR I'll run `./scripts/build.py gen_sdk_source` separately in another PR.

engine/profiler/src/profiler.cpp
```diff
@@ -885,7 +888,11 @@ static dmExtension::Result AppFinalizeProfiler(dmExtension::AppParams* params)
     }
 
     dmProfiler::SetEnabled(false);
-    dmProfile::Finalize();
+
+    if (params->m_ExitStatus != EXTENSION_APP_REBOOT)
+    {
+        dmProfile::Finalize();
+    }
```
AGulev, line 895: The actual fix. Just avoid finalization on restart

engine/extension/src/dmsdk/extension/extension.h
JCash: Theoretically, it's nice to also have the actual "exit code", but let's wait with that until someone asks for it.

engine/profiler/src/profiler.cpp
```diff
@@ -885,7 +888,11 @@ static dmExtension::Result AppFinalizeProfiler(dmExtension::AppParams* params)
     }
 
     dmProfiler::SetEnabled(false);
-    dmProfile::Finalize();
+
+    if (params->m_ExitStatus != EXTENSION_APP_REBOOT)
```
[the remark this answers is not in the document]
JCash: I would like to avoid using the structure directly though.
Please add a function like the other sdk functions for accessing this member.

JCash (changes requested):
overall good.
Needs a new sdk function to retrieve the exit status

engine/extension/src/dmsdk/extension/extension.hpp
```diff
@@ -48,6 +48,13 @@ namespace dmExtension {
       return (T)ExtensionParamsGetContext((ExtensionParams*)params, name_hash);
    }
 
+   /*# get the app exit status
+   * @name GetAppExitStatus
+   * @param app_params [type:dmExtension::AppParams*] The app params sent to the extension dmExtension::AppInitialize / dmExtension::AppInitialize
+   * @return status [type:AppExitStatus] engine exit status
+   */
+   dmExtension::AppExitStatus GetAppExitStatus(dmExtension::AppParams* app_params);
```
JCash: The `ExtensionAppParamsGetAppExitCode()` is missing from `extension.h` (the C header file).
For these generated apis, we want to treat our C api as the "truth", and only need the ".hpp" files when the generator can't or doesn't do a good job.
(We want minimal maintenance here).

So, move this function to extension.h (and rename accordingly) and regenerate the bindings.
You'll have to implement both:
E.g. in source (here C api calls C++ api):
```
int ExtensionAppParamsSetContext(ExtensionAppParams* params, const char* name, void* context)
{
    return SetContext(&params->m_Impl->m_Contexts, name, context);
}
```

Or vice versa, let C++ api call C api.

JCash (changes requested):
Almost there.
Some rearrangement required for the C api needed.

engine/extension/src/dmsdk/extension/extension.hpp
AGulev: I hope now I got it right—thank you for your patience.

JCash (approved):
lgtm

PR #10754, opened 2025-05-29
Render convex-hull point cloud when physics type is 3d

sprocketc:
Reported [here](https://forum.defold.com/t/defold-1-10-2-beta/80587/9). The post contains a repro case that can be used to test this.

When the physics type is "3d", we should render a simple point cloud for convex shapes, to avoid calculating the hull. Trying to use the same algorithm with the physics engine might be possible in the future, but we would have to make sure that what we render is the same with the interpretation of the engine. Rendering a simple point cloud also has the benefit of recognizing potential issues with the provided points. For instance, some of them might be ignored by the engine, if they are contained by the hull. Rendering the collision hull on runtime, when users have physics debugging enabled, might be a better option.

Before / After


editor/src/clj/editor/collision_object.clj
```diff
@@ -439,32 +441,43 @@
...
+                                  :tags #{:collision-shape :outline}
+                                  :passes [pass/outline]
+                                  :user-data {:color color
+                                              :geometry {:primitive-type GL2/GL_LINE_LOOP
+                                                         :vbuf vbuf}}}}]}
+        {:node-id _node-id
+         :node-outline-key "3D Convex Hull"
+         :aabb aabb
+         :renderable {:render-fn render-points-uniform-scale
+                      :tags #{:collision-shape :outline}
```
sprocketc, line 467: Not exactly an outline, but kind of?

vlaaad (approved):
LGTM! Works great!

PR #10759, opened 2025-05-30
Fix issue where rebuilding the game from the Editor doesn't keep the simulated resolution

### LINKED ISSUE #10670: Rebuild game from Editor and sys.reboot don't keep simulated resolution
If I select Debug -> Select Resolution, it will be saved and set each game instance run

If I rebuild the game, while previous build is running, the resolution will be default
If I call sys.reboot, the resolution will be default

**To Reproduce (REQUIRED)**
1. Load example
2. Run App
3. Debug -> Set Resolution -> Any
4. Window size changed
5. If game restart - size keeped
6. If sys.reboot or rebuild game - size not changed

Size always changed

1.10.1

**Platforms (REQUIRED):**
Tested on MacOS

https://github.com/user-attachments/assets/f6cd573f-c459-44f7-b858-cc742874a42a

AGulev:
If the window resolution has been set in the Editor, it will be applied again when the engine is restarted from the Editor using Build To Target.

Fix https://github.com/defold/defold/issues/10670

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -820,7 +820,10 @@
               ;; running to keep engine process
               ;; from halting because stdout/err is
               ;; not consumed.
-              (reboot-engine! selected-target web-server debug?))
+              (reboot-engine! selected-target web-server debug?)
+              (console/start-one-line-log-pump!
+                (:log-stream selected-target)
+                (fn [_] (on-service-url-found prefs selected-target))))
             (launch-new-engine!))))
       (catch Exception e
         (log/warn :exception e)
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -690,8 +690,6 @@
 (defn- build-in-progress? []
   @build-in-progress-atom)
 
-(def ^:private engine-just-rebooted(atom false))
-
 (defn- async-reload-on-app-focus? [prefs]
   (prefs/get prefs [:workflow :load-external-changes-on-app-focus]))
 
@@ -745,16 +743,20 @@
       (throw e))))
 
 (defn- make-launched-log-sink [launched-target on-service-url-found]
-  (let [initial-output (atom "")]
+  (let [initial-output (atom "")
+        version-line (atom nil)
+        updated-target (atom nil)]
     (fn [line]
       (when (< (count @initial-output) 5000)
         (swap! initial-output str line "\n")
         (when-let [target-info (engine/parse-launched-target-info @initial-output)]
-          (targets/update-launched-target! launched-target target-info on-service-url-found)))
-      (when @engine-just-rebooted
-        (reset! engine-just-rebooted false)
-        (doseq [l-target (targets/all-launched-targets)]
-          (on-service-url-found l-target)))
+          (let [result-target (targets/update-launched-target! launched-target target-info)]
+            (reset! updated-target result-target)))
+        (when (not @version-line)
+          (when-let [engine-version-line (engine/parse-engine-version-line @initial-output)]
+            (reset! version-line engine-version-line))))
+      (when (and @updated-target line (= @version-line line))
+        (on-service-url-found @updated-target))
       (when (console/current-stream? (:log-stream launched-target))
         (console/append-console-line! line)))))
 
@@ -763,7 +765,6 @@
     (report-build-launch-progress! (format "Rebooting %s..." (targets/target-message-label target)))
     (engine/reboot! target (local-url target web-server) debug?)
     (report-build-launch-progress! (format "Rebooted %s" (targets/target-message-label target)))
-    (reset! engine-just-rebooted true)
     target
     (catch Exception e
       (report-build-launch-progress! "Reboot failed")
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -753,9 +753,9 @@
           (let [result-target (targets/update-launched-target! launched-target target-info)]
             (reset! updated-target result-target)))
         (when (not @version-line)
-          (when-let [engine-version-line (engine/parse-engine-version-line @initial-output)]
+          (when-let [engine-version-line (engine/parse-engine-version-line line)]
             (reset! version-line engine-version-line))))
-      (when (and @updated-target line (= @version-line line))
+      (when (and @updated-target (= @version-line line))
         (on-service-url-found @updated-target))
       (when (console/current-stream? (:log-stream launched-target))
         (console/append-console-line! line)))))
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -690,6 +690,8 @@
 (defn- build-in-progress? []
   @build-in-progress-atom)
 
+(def ^:private engine-just-rebooted(atom false))
+
 (defn- async-reload-on-app-focus? [prefs]
   (prefs/get prefs [:workflow :load-external-changes-on-app-focus]))
 
@@ -749,6 +751,10 @@
         (swap! initial-output str line "\n")
         (when-let [target-info (engine/parse-launched-target-info @initial-output)]
           (targets/update-launched-target! launched-target target-info on-service-url-found)))
+      (when @engine-just-rebooted
+        (reset! engine-just-rebooted false)
+        (doseq [l-target (targets/all-launched-targets)]
+          (on-service-url-found l-target)))
       (when (console/current-stream? (:log-stream launched-target))
         (console/append-console-line! line)))))
 
@@ -757,6 +763,7 @@
     (report-build-launch-progress! (format "Rebooting %s..." (targets/target-message-label target)))
     (engine/reboot! target (local-url target web-server) debug?)
     (report-build-launch-progress! (format "Rebooted %s" (targets/target-message-label target)))
+    (reset! engine-just-rebooted true)
     target
     (catch Exception e
       (report-build-launch-progress! "Reboot failed")
@@ -820,10 +827,7 @@
               ;; running to keep engine process
               ;; from halting because stdout/err is
               ;; not consumed.
-              (reboot-engine! selected-target web-server debug?)
-              (console/start-one-line-log-pump!
-                (:log-stream selected-target)
-                (fn [_] (on-service-url-found prefs selected-target))))
+              (reboot-engine! selected-target web-server debug?))
             (launch-new-engine!))))
       (catch Exception e
         (log/warn :exception e)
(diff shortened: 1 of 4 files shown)


editor/src/clj/editor/app_view.clj
```diff
@@ -820,7 +820,10 @@
               ;; running to keep engine process
               ;; from halting because stdout/err is
               ;; not consumed.
-              (reboot-engine! selected-target web-server debug?))
+              (reboot-engine! selected-target web-server debug?)
+              (console/start-one-line-log-pump!
+                (:log-stream selected-target)
+                (fn [_] (on-service-url-found prefs selected-target))))
```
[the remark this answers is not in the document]
vlaaad: I'm not quite sure this will work reliably. As I understand from the comments on lines 819-822, there is already a thread that consumes all output from the target and displays it in the console. If that's the case, then launching another thread that will also read input from the same log stream will work only in cases where this one-line thread gets lucky to read a line first.

editor/src/clj/editor/console.clj
```diff
@@ -128,6 +129,22 @@
...
+;; Start a background thread that waits for a single line from the log-stream,
+;; calls sink-fn with it, and then stops.
+(defn start-one-line-log-pump!
+  "Reads a single line from log-stream in a background thread and calls sink-fn with it.
+   Does NOT close the stream."
+  [log-stream sink-fn]
+  (doto (Thread.
+          (fn []
+            (try
+              (let [reader (BufferedReader. (io/reader log-stream :encoding "UTF-8"))]
```
vlaaad: Also, using a buffered reader might read too much, which will lead to log loss. The log line already seems lost in this case, since it is not submitted to the console.

vlaaad (commented):
I'm not sure this solution works, can you explain it?

editor/src/clj/editor/app_view.clj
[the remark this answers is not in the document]
AGulev: what the best way to know the reboot is done?
Ofc I can just thread sleep for 100 ms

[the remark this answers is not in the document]
vlaaad: Great question, I'm not sure. In `editor.engine/reboot!`, we send a reboot request and then wait until all output is processed. Not sure how the engine does it, maybe it closes this connection after the reboot is done? In that case, the editor doesn't have to do anything special and could just apply the resolution after the call to reboot...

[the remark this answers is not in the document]
AGulev: No, it doesn't work this way, I tried it. 
Maybe then just sleep (we do it when we start the engine)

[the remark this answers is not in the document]
AGulev: I have an idea, let me try

AGulev:
I tried new way to implement it, please take a look

editor/src/clj/editor/app_view.clj
```diff
@@ -749,6 +751,10 @@
         (swap! initial-output str line "\n")
         (when-let [target-info (engine/parse-launched-target-info @initial-output)]
           (targets/update-launched-target! launched-target target-info on-service-url-found)))
+      (when @engine-just-rebooted
+        (reset! engine-just-rebooted false)
+        (doseq [l-target (targets/all-launched-targets)]
+          (on-service-url-found l-target)))
```
vlaaad: Shouldn't the code call `on-service-url-found` only for the `launched-target` that is an argument to this function?

vlaaad: Does it work if the old engine is constantly outputting something into logs? Then this branch might be executed before the engine is actually restarted. There is code above (`engine/parse-launched-target-info`) that parses log output, perhaps we should do something like that instead? maybe even without the global state `engine-just-rebooted` atom...

vlaaad (commented):
I'm not a big fan of communicating via shared global state, but perhaps we don't have other options here. Or maybe we do? Maybe it's worth exploring if we could parse the log similarly to what `engine/parse-launched-target-info` does, and call `on-service-url-found` only when we get a particular message indicating that the engine has rebooted? 

Also, as I understand, when we have multiple instances running, we have a separate log sink for each instance (see `launch-new-engine!` fn in `launch-built-project!`). In that case, we should only call `on-service-url-found` for the target that we have as an argument in `make-launched-log-sink`, since this is the engine that got rebooted.

editor/src/clj/editor/app_view.clj
AGulev: I don't like the idea with parsing, because then we have to parse it every single line which is far way from perfect. For now we parse only 5000 messages (which is too many I think).

> Shouldn't the code call on-service-url-found only for the launched-target that is an argument to this function?

no, because it will be changed in `targets/update-launched-target!` 
but I'll recheck logic with multiple instance again (when I have time)

vlaaad: Initially, we could wait for a message in the shape `INFO:ENGINE: Defold Engine $version ($sha)`. Then we could store it inside the function, and then we would only need to compare new lines for equality with the stored string. Then, subsequent "parsing" of every line is going to be fast. I think this approach of gradually gathering information from every log line is more reliable than relying on the next log output line after notifying the engine about a reboot request to indicate that the engine has already rebooted.

Btw, the info parser looks for the first 5000 characters, not 5000 messages.

editor/src/clj/editor/app_view.clj
```diff
@@ -743,12 +743,20 @@
       (throw e))))
 
 (defn- make-launched-log-sink [launched-target on-service-url-found]
-  (let [initial-output (atom "")]
+  (let [initial-output (atom "")
+        version-line (atom nil)
+        updated-target (atom nil)]
```
AGulev, line 748: instead of global state we have these two atoms

editor/src/clj/editor/targets.clj
```diff
@@ -114,27 +114,21 @@
...
-                           ;; Wait for the console stream to be marked as active
-                           (loop []
-                             (if (console/current-stream? (:log-stream result-target))
-                               (on-service-url-found result-target)
-                               (do
-                                 (Thread/sleep 100)
-                                 (recur))))))
-                       result-target)
+                     (let [rt (merge launched-target target-info)]
+                       (vreset! result-target rt)
```
AGulev, line 124: we need it to be able to return updated target as result of update-launched-target!. Update target will contain `:url` we need to be able to send resolution change message

editor/src/clj/editor/engine.clj
```diff
@@ -174,6 +174,13 @@
              {:log-port log-port
               :address loopback-address}))))
 
+;; Parse a line from engine output to extract engine version info.
+(defn parse-engine-version-line [output]
+  (some (fn [line]
+          (when-let [[_ version sha] (re-find #"INFO:ENGINE: Defold Engine ([^\s]+) \(([^)]+)\)" line)]
+            line))
+        (split-lines output)))
```
AGulev: if line has form of `INFO:ENGINE: Defold Engine $version ($sha)` - return it

editor/src/clj/editor/app_view.clj
```diff
@@ -743,12 +743,20 @@
...
         (swap! initial-output str line "\n")
         (when-let [target-info (engine/parse-launched-target-info @initial-output)]
-          (targets/update-launched-target! launched-target target-info on-service-url-found)))
+          (let [result-target (targets/update-launched-target! launched-target target-info)]
+            (reset! updated-target result-target)))
+        (when (not @version-line)
+          (when-let [engine-version-line (engine/parse-engine-version-line @initial-output)]
+            (reset! version-line engine-version-line))))
+      (when (and @updated-target line (= @version-line line))
+        (on-service-url-found @updated-target))
```
AGulev, line 759: (see comments below first)
now we have result of `update-launched-target!` and `version-line` we use to compare every line with it.
if we get such line - then update window size

editor/src/clj/editor/targets.clj
```diff
@@ -114,27 +114,21 @@
...
-                       (when (and (:url target-info) (not (:url launched-target)))
-                         (future
-                           ;; Wait for the console stream to be marked as active
-                           (loop []
-                             (if (console/current-stream? (:log-stream result-target))
-                               (on-service-url-found result-target)
-                               (do
-                                 (Thread/sleep 100)
-                                 (recur))))))
-                       result-target)
```
AGulev, line 132: I think we don't need it anymore. New even that waits `INFO:ENGINE: Defold Engine $version ($sha)` should work just fine

AGulev (commented):
One more attempt...

editor/src/clj/editor/app_view.clj
```diff
@@ -743,12 +743,20 @@
...
+        updated-target (atom nil)]
     (fn [line]
       (when (< (count @initial-output) 5000)
         (swap! initial-output str line "\n")
         (when-let [target-info (engine/parse-launched-target-info @initial-output)]
-          (targets/update-launched-target! launched-target target-info on-service-url-found)))
+          (let [result-target (targets/update-launched-target! launched-target target-info)]
+            (reset! updated-target result-target)))
+        (when (not @version-line)
+          (when-let [engine-version-line (engine/parse-engine-version-line @initial-output)]
```
vlaaad: This code could parse a `line` instead of `@initial-input`, so it checks every new line instead of re-checking a set of initial lines multiple times. This would improve performance a bit.

editor/src/clj/editor/app_view.clj
```diff
@@ -743,12 +743,20 @@
...
+            (reset! version-line engine-version-line))))
+      (when (and @updated-target line (= @version-line line))
```
vlaaad: As I understand, `line` is never `nil`, it's always a string, so there is no need to check if it exists.

vlaaad (dismissed):
LGTM!

editor/src/clj/editor/app_view.clj
AGulev: true!

AGulev:
Thank you @vlaaad ! 
I think this approach is more robust, even for the initial resolution setup (instead of using sleep).  
I like it much more now!

editor/src/clj/editor/engine.clj
```diff
@@ -174,6 +174,13 @@
              {:log-port log-port
               :address loopback-address}))))
 
+;; Parse a line from engine output to extract engine version info.
+(defn parse-engine-version-line [output]
+  (some (fn [line]
+          (when-let [[_ version sha] (re-find #"INFO:ENGINE: Defold Engine ([^\s]+) \(([^)]+)\)" line)]
+            line))
```
vlaaad: Since you always pass a single line here, there is no need to split it into lines. The whole function body can be simplified to:
```clj
(when (re-find #"INFO:ENGINE: Defold Engine ([^\s]+) \(([^)]+)\)" line)
  line)
```

vlaaad (approved):
shipit!

PR #10767, opened 2025-06-02
Sanitize float values and escape strings for valid JSON output in scene graph dump

### LINKED ISSUE #10690: Invalid scene graph dump with multiline text
No

**To Reproduce (REQUIRED)**
1. Open attached project
2. Run game
3. In the browser open `localhost:8002/scene_graph`

Engine produces valid scen graph dump

  - Version 1.10.1

**Platforms (REQUIRED):**

[scene-graph.zip](https://github.com/user-attachments/files/20272600/scene-graph.zip)

**Logs (OPTIONAL):**

**Workaround (OPTIONAL):**

**Screenshots (OPTIONAL):**

Result scene graph dump
```
{
    "id": "main",
    "type": "collectionc",
    "resource": "/main/main.collectionc",
    "children": [
    {
        "id": "/go",
        "type": "goc",
        "resource": "/_generated_43b90d8c.goc",
        "position": [0.000000, 0.000000, 0.000000],
        "rotation": [0.000000, 0.000000, 0.000000, 1.000000],
        "scale": [1.000000, 1.000000, 1.000000],
        "world_position": [0.000000, 0.000000, 0.000000],
        "world_rotation": [0.000000, 0.000000, 0.000000, 1.000000],
        "world_scale": [1.000000, 1.000000, 1.000000],
        "children": [
        {
            "id": "test",
            "type": "guic",
            "resource": "/main/test.guic",
            "children": [
            {
                "type": "gui_node_text",
                "custom_type": "",
                "id": "text",
                "pivot": "PIVOT_CENTER",
                "position": [438.000000, 291.000000, 0.000000, 1.000000],
                "rotation": [0.000000, 0.000000, 0.000000, 1.000000],
                "euler": [0.000000, 0.000000, 0.000000, 0.000000],
                "scale": [1.000000, 1.000000, 1.000000, 1.000000],
                "color": [1.000000, 1.000000, 1.000000, 1.000000],
                "size": [200.000000, 100.000000, 0.000000, 0.000000],
                "outline": [0.000000, 0.000000, 0.000000, 1.000000],
                "shadow": [0.000000, 0.000000, 0.000000, 1.000000],
                "slice9": [0.000000, 0.000000, 0.000000, 0.000000],
                "pie_params": [0.000000, 360.000000, 0.000000, 0.000000],
                "text_params": [1.000000, 0.000000, 0.000000, 0.000000],
                "world_position": [438.000000, 291.000000, 0.000000],
                "world_rotation": [0.000000, 0.000000, 0.000000, 1.000000],
                "world_scale": [1.000000, 1.000000, 1.000000],
                "world_size": [200.000000, 100.000000, 0.000000],
                "enabled": 1,
                "text": "Some
multiline
text",
                "children": []
            }]
        }]
    }]
}
```

### LINKED ISSUE #10691: Invalid scene graph dump with sprite size == 0

I scene has sprite with size.x or size.y == 0 then scene graph contains `inf` in `world_scale` field which leads to invalid JSON syntax.
Use case of such sprite: bar that shows current shield of the character and made as 9-slices.

**To Reproduce (REQUIRED)**
1. Open attached project
2. Run game
3. Open in browser `localhost:8002/scene_graph`

Engine produces correct scene graph dump

  - Version 1.10.1

**Platforms (REQUIRED):**

[scene-graph.zip](https://github.com/user-attachments/files/20272610/scene-graph.zip)

**Logs (OPTIONAL):**

**Workaround (OPTIONAL):**

**Screenshots (OPTIONAL):**

Scene graph dump
```
{
    "id": "main",
    "type": "collectionc",
    "resource": "/main/main.collectionc",
    "children": [
    {
        "id": "/go",
        "type": "goc",
        "resource": "/_generated_41def0fd.goc",
        "position": [0.000000, 0.000000, 0.000000],
        "rotation": [0.000000, 0.000000, 0.000000, 1.000000],
        "scale": [1.000000, 1.000000, 1.000000],
        "world_position": [0.000000, 0.000000, 0.000000],
        "world_rotation": [0.000000, 0.000000, 0.000000, 1.000000],
        "world_scale": [1.000000, 1.000000, 1.000000],
        "children": [
        {
            "id": "sprite",
            "type": "spritec",
            "resource": "/_generated_b5976e66.spritec",
            "position": [357.000000, 288.000000, 0.000000],
            "rotation": [0.000000, 0.000000, 0.000000, 1.000000],
            "scale": [1.000000, 1.000000, 1.000000],
            "size": [0.000000, 256.000000, 0.000000],
            "world_position": [357.000000, 288.000000, 0.000000],
            "world_rotation": [0.000000, 0.000000, 0.000000, 8.031189],
            "world_scale": [inf, 0.003906, 1.000000],
            "world_size": [1.000000, 1.000000, 1.000000],
            "enabled": 1,
            "children": []
        }]
    }]
}
```

deurell:
Fix for non-finite floats and unescaped strings generating invalid JSON output in scene graph dumps.

Fixes #10690 and #10691

Technical changes
- Added a utility to sanitize float values to ensure only valid JSON numbers are output.
- Updated string output to properly escape special characters for JSON compliance.
- Fix sprite world_scale calculation to not be dependent on sprite size.

### DIFF
--- a/engine/engine/src/engine_service.cpp
+++ b/engine/engine/src/engine_service.cpp
@@ -816,7 +816,12 @@ namespace dmEngineService
                 case '\n': SendText(request, "\\n"); break;
                 case '\r': SendText(request, "\\r"); break;
                 case '\t': SendText(request, "\\t"); break;
-                default: SendText(request, (char[2]){*p, 0});
+                default:
+                {
+                    char buf[2] = {*p, 0};
+                    SendText(request, buf);
+                    break;
+                }
             }
         }
         SendText(request, "\"");
--- a/engine/engine/src/engine_service.cpp
+++ b/engine/engine/src/engine_service.cpp
@@ -829,7 +829,7 @@ namespace dmEngineService
 
     static float JsonSafeFloat(float f)
     {
-        return isfinite(f) ? f : 0.0f;
+        return isfinite(f) ? f : std::numeric_limits<float>::max();
     }
 
     static void OutputJsonProperty(dmGameObject::SceneNodeProperty* property, dmWebServer::Request* request, int indent)
--- a/engine/engine/src/engine_service.cpp
+++ b/engine/engine/src/engine_service.cpp
@@ -14,7 +14,7 @@
 
 #include <stdint.h>
 #include <string.h>
-#include <limits>
+#include <float.h>
 #include <dlib/webserver.h>
 #include <dlib/message.h>
 #include <dlib/dstrings.h>
@@ -830,7 +830,7 @@ namespace dmEngineService
 
     static float JsonSafeFloat(float f)
     {
-        return isfinite(f) ? f : std::numeric_limits<float>::max();
+        return isfinite(f) ? f : FLT_MAX;
     }
 
     static void OutputJsonProperty(dmGameObject::SceneNodeProperty* property, dmWebServer::Request* request, int indent)
--- a/engine/engine/src/engine_service.cpp
+++ b/engine/engine/src/engine_service.cpp
@@ -804,6 +804,29 @@ namespace dmEngineService
             dmWebServer::Send(request, buf, sizeof(buf));
     }
 
+    static void SendJsonEscapedText(dmWebServer::Request* request, const char* text)
+    {
+        SendText(request, "\"");
+        for (const char* p = text; *p; ++p)
+        {
+            switch (*p)
+            {
+                case '\"': SendText(request, "\\\""); break;
+                case '\\': SendText(request, "\\\\"); break;
+                case '\n': SendText(request, "\\n"); break;
+                case '\r': SendText(request, "\\r"); break;
+                case '\t': SendText(request, "\\t"); break;
+                default: SendText(request, (char[2]){*p, 0});
+            }
+        }
+        SendText(request, "\"");
+    }
+
+    static float JsonSafeFloat(float f)
+    {
+        return isfinite(f) ? f : 0.0f;
+    }
+
     static void OutputJsonProperty(dmGameObject::SceneNodeProperty* property, dmWebServer::Request* request, int indent)
     {
         SendIndent(request, indent);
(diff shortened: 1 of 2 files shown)


engine/engine/src/engine_service.cpp
```diff
@@ -825,13 +853,22 @@ namespace dmEngineService
                     dmSnPrintf(buffer, sizeof(buffer), "\"0x%016llX\"", (unsigned long long)property->m_Value.m_Hash);
             }
             break;
-        case dmGameObject::SCENE_NODE_PROPERTY_TYPE_NUMBER: dmSnPrintf(buffer, sizeof(buffer), "%f", property->m_Value.m_Number); break;
+        case dmGameObject::SCENE_NODE_PROPERTY_TYPE_NUMBER: dmSnPrintf(buffer, sizeof(buffer), "%f", JsonSafeFloat(property->m_Value.m_Number)); break;
```
JCash, line 857: The more I think about it, what should the default be? 0 seems like it may be misleading here?
Should it be an overly large number, to signify the issue? E.g. max float or something?

Also, for https://github.com/defold/defold/issues/10691, I would have expected something sprite related? (i..e in comp_sprite.cpp when returning `world_scale`)

JCash (changes requested):
The catch-all as a concept is good.
However, is 0 a good "invalid number"?

Also want a fix for the sprite component though.

engine/engine/src/engine_service.cpp
deurell: I wanted to make sure we never send non valid json so that's the catch all scenario. I went for 0 as a catch all but maybe maxfloat is better. I'll have a look at comp_sprite.cpp to see if I can come up with a nice solution to not get here in the first place.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -2384,6 +2384,10 @@ namespace dmGameSystem
                     {
                         // Since the size is baked into the matrix, we divide by it here
                         Vector3 size( component->m_Size.getX() * component->m_Scale.getX(), component->m_Size.getY() * component->m_Scale.getY(), 1);
+                        const float epsilon = 1e-6f;
+                        if (fabsf(size.getX()) < epsilon) size.setX(1.0f);
+                        if (fabsf(size.getY()) < epsilon) size.setY(1.0f);
+                        if (fabsf(size.getZ()) < epsilon) size.setZ(1.0f);
```
deurell: ~~Would a fallback value of 1.0 be ok for size in the case when the sprite size is near 0 and making it return the world scale for the axis directly from the transform?~~ Changing this to do a parent world scale * local scale instead of using baked scale

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -2381,12 +2381,13 @@ namespace dmGameSystem
...
-                    {
-                        // Since the size is baked into the matrix, we divide by it here
-                        Vector3 size( component->m_Size.getX() * component->m_Scale.getX(), component->m_Size.getY() * component->m_Scale.getY(), 1);
-                        value = Vector4(dmVMath::DivPerElem(transform.GetScale(), size));
-                    }
+                {
+                    Matrix4 parent_world = dmGameObject::GetWorldMatrix(component->m_Instance);
+                    Vector3 parent_scale = dmTransform::ToTransform(parent_world).GetScale();
+                    Vector3 world_scale = dmVMath::MulPerElem(parent_scale, component->m_Scale);
+                    value = Vector4(world_scale);
```
deurell, line 2388: calculate sprite world scale without using baked transform and divide with size

engine/engine/src/engine_service.cpp
deurell: I changed it to max float and also change the way we calculate sprite world scale so we don't do the division with size.

engine/engine/src/engine_service.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include <stdint.h>
 #include <string.h>
+#include <limits>
```
JCash: Please prefer the C api (i.e. `float.h`)
We try to avoid the C++ standard libraries (templates etc).
It's also about communication to others reading the code, we don't want any C++ standard library code in our engine.
You can read about our code style and some of our reasons here:
https://defold.com/2020/05/31/The-Defold-engine-code-style/

engine/engine/src/engine_service.cpp
```diff
@@ -804,6 +805,34 @@ namespace dmEngineService
...
+                    break;
+                }
+            }
+        }
+        SendText(request, "\"");
+    }
+
+    static float JsonSafeFloat(float f)
+    {
+        return isfinite(f) ? f : std::numeric_limits<float>::max();
```
JCash: As above, we don't want to add `std::` usage.

JCash (changes requested):
lgtm
A code style change needed.

engine/engine/src/engine_service.cpp
deurell: fixing it asap

JCash (approved):

britzl:
I have one comment about the PR from a release note perspective:

We use the title and initial PR comment in the release notes. We generate the release notes using GitHub's GraphQL language and a Python script (https://github.com/defold/defold/blob/dev/scripts/releasenotes_github_projectv2.py).

This means that the title of the PR should be something which an end user can understand when they read the release notes on the forum. Example: https://forum.defold.com/t/defold-1-10-2-beta/80587

The PR comment should start with the text we want to show in the release note details.

GitHub closing keywords such as `Fix`, `Fixes`, `Closes` etc are automatically removed.

Any text following after a line starting with "Technical changes" (and [permutations](https://github.com/defold/defold/blob/dev/scripts/releasenotes_github_projectv2.py#L285-L291)) will be removed. This is helpful since you often want to write something about the technical solution and additional information which only the reviewer or us developers care about.

britzl:
What this means in this specific case is that I'd put the "Background" text at the top as the release notes description. And the "Solution" heading should be renamed to "Technical changes" so that it isn't included in the generated release notes.

britzl (approved):
See comment regarding release notes

deurell:
> What this means in this specific case is that I'd put the "Background" text at the top as the release notes description. And the "Solution" heading should be renamed to "Technical changes" so that it isn't included in the generated release notes.

Thanks! Changed the description. Is the new one in a correct format?

PR #10774, opened 2025-06-03
Edit GUI layers using editor scripts

vlaaad:
Now it's possible to edit GUI layers using editor scripts, e.g.:
```lua
editor.transact({
    editor.tx.add("/main.gui", "layers", {name = "foreground"}),
    editor.tx.add("/main.gui", "layers", {name = "background"})
})
```
Additionally, it's possible to reorder layers:
```lua
local fg, bg = table.unpack(editor.get("/main.gui", "layers"))
editor.transact({
    editor.tx.reorder("/main.gui", "layers", {bg, fg})
})
```

Related to #8504

### DIFF
--- a/editor/src/clj/editor/attachment.clj
+++ b/editor/src/clj/editor/attachment.clj
@@ -119,7 +119,7 @@
   [node-type list-kw]
   (-> @state-atom :add (get node-type) (contains? list-kw)))
 
-(defn reorder?
+(defn reorderable?
   "Checks if a node type allows reordering of a list-kw list"
   [node-type list-kw]
   (-> @state-atom :reorder (get node-type) (contains? list-kw)))
@@ -189,7 +189,7 @@
 
   The implementation will assert that the supplied child node ids are the same
   node ids as defined by [[getter]]. It will also assert that the container
-  node-id defines reorder of a list identified by list-kw (see [[reorder?]])"
+  node-id defines reorder of a list identified by list-kw (see [[reorderable?]])"
   [node-id list-kw reordered-child-node-ids]
   (g/expand-ec reorder-tx node-id list-kw reordered-child-node-ids))
 
--- a/editor/src/clj/editor/attachment.clj
+++ b/editor/src/clj/editor/attachment.clj
@@ -22,10 +22,12 @@
 (defonce ^:private state-atom
   ;; :add -> type -> list-kw -> type -> tx-attach-fn
   ;; :get -> type -> list-kw -> get-fn
+  ;; :reorder -> type -> list-kw -> reorder-fn
   ;;
   ;; tx-attach-fn: fn of parent-node, child-node -> txs
   ;; get-fn: fn of node, evaluation-context -> vector of nodes
-  (atom {:add {} :get {}}))
+  ;; reorder-fn: fn of reordered-nodes -> txs
+  (atom {:add {} :get {} :reorder {}}))
 
 (defn register!
   "Register a semantical list of child components
@@ -35,20 +37,25 @@
     list-kw      name of the node component list
 
   Kv-args (at least one is required):
-    :add    a map that defines how new items are added to the list;
-            where keys are child node types used for constructing new nodes, and
-            values are attach functions, i.e. functions of parent-node-id
-            (container) and child-node-id (item) that should return transaction
-            steps that connect the child node into the parent
-    :get    a function that defines how to read a list of children, will receive
-            2 args: parent-node-id (container) and evaluation-context; should
-            return a vector of node ids"
-  [node-type list-kw & {:keys [add get]}]
-  {:pre [(or (ifn? get) (map? add))]}
+    :add        a map that defines how new items are added to the list; where
+                keys are child node types used for constructing new nodes, and
+                values are attach functions, i.e. functions of parent-node-id
+                (container) and child-node-id (item) that should return
+                transaction steps that connect the child node into the parent
+    :get        a function that defines how to read a list of children, will
+                receive 2 args: parent-node-id (container) and
+                evaluation-context; should return a vector of node ids
+    :reorder    a function that defines how to reorder a list of children, will
+                receive 1 arg: reordered-node-ids (vector of item node ids,
+                validated to be the same node ids as those returned by :get);
+                should return transaction steps that set the new order"
+  [node-type list-kw & {:keys [add get reorder]}]
+  {:pre [(or (ifn? get) (map? add) (ifn? reorder))]}
   (swap! state-atom (fn [s]
                       (cond-> s
                               add (update :add update node-type update list-kw merge add)
-                              get (update :get update node-type assoc list-kw get))))
+                              get (update :get update node-type assoc list-kw get)
+                              reorder (update :reorder update node-type assoc list-kw reorder))))
   nil)
 
 (defn add-impl [current-state parent-node-type parent-node-id attachment-tree init-fn]
@@ -112,6 +119,11 @@
   [node-type list-kw]
   (-> @state-atom :add (get node-type) (contains? list-kw)))
 
+(defn reorder?
+  "Checks if a node type allows reordering of a list-kw list"
+  [node-type list-kw]
+  (-> @state-atom :reorder (get node-type) (contains? list-kw)))
+
 (defn child-node-types
   "Returns defined child node-types for a parent node-type's list-kw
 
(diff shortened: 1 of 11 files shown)


editor/src/clj/editor/gui.clj
```diff
@@ -460,19 +462,16 @@
...
-          child-indices (g/node-value target :child-indices)
-          next-index (next-child-index child-indices)]
-      (concat
-        (g/update-property source :id id/resolve taken-ids)
-        (g/set-property source :child-index next-index)
-        (attach-gui-node node-tree target source type)))))
+    (g/with-auto-evaluation-context evaluation-context
+      (let [node-tree (node->node-tree (:basis evaluation-context) target)
+            taken-ids (g/node-value node-tree :id-counts evaluation-context)
+            next-index (gui-attachment/next-child-index target evaluation-context)]
```
vlaaad, line 470: I moved `next-child-index` to a separate namespace. Here, I additionally improved the code so we don't create too many short-lived evaluation contexts.

editor/src/clj/editor/gui.clj
```diff
@@ -1322,8 +1321,6 @@
                                (validate-material-resource _node-id material-infos material)))))
 
   (output visual-base-node-msg g/Any produce-visual-base-node-msg)
-  (output gui-scene g/Any (g/fnk [_node-id]
-                                 (node->gui-scene _node-id)))
```
vlaaad, line 1326: This was very wrong because `node->gui-scene` walks the graph instead of relying only on the `fnk` inputs. Thankfully, this output was unused.

editor/src/clj/editor/gui.clj
```diff
@@ -2691,36 +2688,31 @@
...
 (defn- attach-layer
-  ([self layers-node layer]
-   (attach-layer self layers-node layer false))
   ;; Self is not used here but added to conform all attach-*** functions
-  ([_self layers-node layer internal?]
-   (concat
+  [_self layers-node layer]
+  (concat
     (g/connect layer :_node-id layers-node :nodes)
-    (when (not internal?)
```
vlaaad, line 2700: Removed `internal?` flag — there was no point in creating an unused layer node in every GUI.

editor/src/clj/editor/gui.clj
```diff
@@ -2691,36 +2688,31 @@
...
+    (g/connect layer :node-outline layers-node :child-outlines)
+    (g/connect layer :node-id+child-index layers-node :child-indices)
+    (g/connect layer :name+child-index layers-node :indexed-layer-names)
+    (g/connect layers-node :name-counts layer :name-counts)))
+
+(defn add-layer [scene parent name child-index select-fn]
   (g/make-nodes (g/node-id->graph-id scene) [node [LayerNode :name name :child-index child-index]]
-    ;; Self is not used when attaching layers
-    (attach-layer nil parent node)
+    (attach-layer scene parent node)
```
vlaaad, line 2704: The fact that `scene` is unused is only a concern of the `attach-layer` function.

editor/src/clj/editor/gui.clj
```diff
@@ -3476,27 +3468,27 @@
...
+                                :type node-type)]
+          (concat
+            (g/operation-label "Add Gui Node")
+            (g/make-nodes (g/node-id->graph-id scene) [gui-node [def-node-type node-properties]]
+              (attach-gui-node node-tree parent gui-node node-type)
+              (when select-fn
+                (select-fn [gui-node]))))))
+      g/transact
+      g/tx-nodes-added
+      first))
```
vlaaad, line 3491: Mostly indentation and minor improvement with evaluation context, prompted by using the `gui-attachment/next-child-index` function that I moved to a separate ns.

editor/src/clj/editor/gui.clj
```diff
@@ -3762,18 +3754,15 @@
                                                                      :particlefx (resolve-resource (:particlefx particlefx-desc))]]
                                       (attach-particlefx-resource self particlefx-resources-node particlefx-resource)))))
 
-      (g/make-nodes graph-id [layers-node LayersNode
-                              no-layer [LayerNode
-                                        :name ""]]
```
vlaaad, line 3767: we no longer create useless layer nodes

editor/src/clj/editor/gui.clj
```diff
@@ -3762,18 +3754,15 @@
...
-                    (g/connect layers-node :_node-id self :layers-node) ; for the tests :/
+      (g/make-nodes graph-id [layers-node LayersNode]
+                    (g/connect layers-node :_node-id self :layers-node)
```
vlaaad, line 3758: Removed the comment `; for the tests :/` since it's also now used in editor scripts

editor/src/clj/editor/attachment.clj
```diff
@@ -112,6 +119,11 @@
   [node-type list-kw]
   (-> @state-atom :add (get node-type) (contains? list-kw)))
 
+(defn reorder?
+  "Checks if a node type allows reordering of a list-kw list"
+  [node-type list-kw]
+  (-> @state-atom :reorder (get node-type) (contains? list-kw)))
+
```
matgis, line 126: Maybe clearer to name it `can-reorder?` or `reorderable?`?

editor/src/clj/editor/gui.clj
```diff
@@ -460,19 +462,16 @@
...
-        (g/set-property source :child-index next-index)
-        (attach-gui-node node-tree target source type)))))
+    (g/with-auto-evaluation-context evaluation-context
+      (let [node-tree (node->node-tree (:basis evaluation-context) target)
+            taken-ids (g/node-value node-tree :id-counts evaluation-context)
+            next-index (gui-attachment/next-child-index target evaluation-context)]
+        (concat
+          (g/update-property source :id id/resolve taken-ids)
+          (g/set-property source :child-index next-index)
+          (attach-gui-node node-tree target source type))))))
```
matgis, line 474: Are we sure we're not holding a reference to the `evaluation-context` here due to laziness?

editor/src/clj/editor/gui.clj
```diff
@@ -2691,36 +2688,31 @@
...
-      (g/operation-label "Add Layer")
-      (add-layer project scene parent name next-index select-fn)))))
+(defn- add-layer-handler [_project {:keys [scene parent]} select-fn]
+  (g/transact
+    (g/with-auto-evaluation-context evaluation-context
+      (let [name (id/gen "layer" (g/node-value parent :name-counts evaluation-context))
+            next-index (gui-attachment/next-child-index parent evaluation-context)]
+        (concat
+          (g/operation-label "Add Layer")
+          (add-layer scene parent name next-index select-fn))))))
```
matgis, line 2715: I think we may want to wrap the internals of the `g/with-auto-evaluation-context` call in a `doall` here? The lazy sequences will be realized by `g/transact`, but any cache entries produced will be committed when we exit the `g/with-auto-evaluation-context` scope, so anything lazily produced will be placed in the `evaluation-context` cache instead of the system cache and never be committed.

editor/src/clj/editor/gui.clj
```diff
@@ -3476,27 +3468,27 @@
...
+                                :id id
+                                :child-index next-index
+                                :custom-type custom-type
+                                :type node-type)]
+          (concat
+            (g/operation-label "Add Gui Node")
+            (g/make-nodes (g/node-id->graph-id scene) [gui-node [def-node-type node-properties]]
+              (attach-gui-node node-tree parent gui-node node-type)
+              (when select-fn
+                (select-fn [gui-node]))))))
```
matgis, line 3488: Same thing here - wrap internals in `doall`?

matgis (changes requested):
I think we may need to force evaluation of the lazy sequences produced within `g/with-auto-evaluation-context` blocks?

editor/src/clj/editor/gui.clj
vlaaad: Yes, because we use the evaluation context only in the let binding and not within the lazy sequence.

editor/src/clj/editor/gui.clj
vlaaad: I think this is unnecessary. It would only be an issue if the lazy sequence could use the evaluation context; then it would create cache entries after exiting the `with-auto-evaluation-context` block. But here, we instead use the evaluation context only in the let binding.

editor/src/clj/editor/gui.clj
vlaaad: Same here: we only use the evaluation context in the let binding, and don't use it within the lazy sequence.

matgis (approved):
LGTM!

PR #10784, opened 2025-06-04
Added IsURL(), ToURL(), PushURL() and 2 variants of CheckURL() to dmsdk.

### LINKED ISSUE #7220: Add CheckURL function to sdk
I need to work with url in native extension.
In src you have function CheckURL. But looks like it is not added to sdk?
https://github.com/defold/defold/blob/37245683c052ffc814c1b07010221722e1a5f0c0/engine/script/src/script_msg.cpp#L580

Add dmScript::CheckURL to sdk

Add definition in my project.
```c++
namespace dmScript {
    dmMessage::URL* CheckURL(lua_State* L, int index);
}
```

Gert-Jan:
Allows native extensions to have better control over URLs in the lua state.
Closes https://github.com/defold/defold/issues/7220
### Technical changes
* Moved `IsURL()` in `script_msg.cpp` to the other URL related functions. Changed the `IsURL()` implementation to do the same as other [`IsXxx()`](https://github.com/defold/defold/blob/027491c8d5dbe0b592a7d41f348056f902c78113/engine/script/src/script_vmath.cpp#L188) implentations.
* As it stands the URL related implementations are split in [script_msg.cpp](https://github.com/defold/defold/blob/a562970c7fea4e059faf1f57ca704625fa1090ab/engine/script/src/script_msg.cpp#L582) and [script.cpp](https://github.com/defold/defold/blob/a562970c7fea4e059faf1f57ca704625fa1090ab/engine/script/src/script.cpp#L848). For now I put the implementation for CheckURL() in script.cpp because it's so closely related to it. Maybe a better place would be script_msg.cpp because there already is [different variant](https://github.com/defold/defold/blob/a562970c7fea4e059faf1f57ca704625fa1090ab/engine/script/src/script_msg.cpp#L590) of CheckURL().
* The specific use case for this change was getting an absolute URL from a relative URL string, when calling from Lua. The rest (IsURL(), ToURL(), PushURL() and other CheckURL()) of this PR escalated from there. Example of getting an absolute URL:
```
// called from lua
dmMessage::URL sender;
dmMessage::URL absoluteUrl;
if (dmScript::CheckURL(L, &sender)) {
  // lua stack 1: "object#component"
  dmScript::ResolveURL(L, 1, &absoluteUrl, &sender);
}
```

### DIFF
--- a/engine/script/src/dmsdk/script/script.h
+++ b/engine/script/src/dmsdk/script/script.h
@@ -460,6 +460,50 @@ namespace dmScript
     const char* GetStringFromHashOrString(lua_State* L, int index, char* buffer, uint32_t bufferlength);
 
     /*#
+     * Check if the value at #index is a URL
+     * @name IsURL
+     * @param L [type:lua_State*] Lua state
+     * @param index [type:int] Index of the value
+     * @return result [type:bool] true if the value at #index is a URL
+     */
+    bool IsURL(lua_State *L, int index);
+
+    /*#
+     * get the value at index as a dmMessage::URL*
+     * @name ToURL
+     * @param L [type:lua_State*] Lua state
+     * @param index [type:int] Index of the value
+     * @return hash [type:dmhash_t*] pointer to URL or 0 if it's not a URL
+     */
+    dmMessage::URL* ToURL(lua_State *L, int index);
+
+    /*#
+     * Push a URL value onto the supplied lua state, will increase the stack by 1.
+     * @name PushURL
+     * @param L [type:lua_State*] Lua state
+     * @param url [type:dmMessage::URL&] URL reference to push
+     */
+    void PushURL(lua_State* L, const dmMessage::URL& url);
+
+    /*#
+     * Check if the value in the supplied index on the lua stack is a dmMessage::URL and returns it if so.
+     * @name CheckURL
+     * @param L [type:lua_State*] Lua state
+     * @param index [type:int] Index of the value
+     * @return url [type:dmMessage::URL*] The pointer to the value
+     */
+    dmMessage::URL* CheckURL(lua_State* L, int index);
+
+    /*#
+     * Get the current game object URL
+     * @name CheckURL
+     * @param L [type:lua_State*] Lua state
+     * @param out_url [type:dmMessage::URL*] where to store the result
+     * @return result [type:bool] true if successful
+     */
+    bool CheckURL(lua_State* L, dmMessage::URL* out_url);
+
+    /*#
      * Push DDF message to Lua stack
      * @name PushDDF
      * @param L [type: lua_State*] the Lua state
@@ -639,15 +683,6 @@ namespace dmScript
      * Lua stack on exit
      */
     void UnrefInInstance(lua_State* L, int ref);
-
-    /*#
-     * Get the current game object URL
-     * @name CheckURL
-     * @param L [type:lua_State*] Lua state
-     * @param out_url [type:dmMessage::URL*] where to store the result
-     * @return result [type:bool] true if successful
-     */
-    bool CheckURL(lua_State* L, dmMessage::URL* out_url);
 
     /*#
      * Resolves the value in the supplied index on the lua stack to a URL. It long jumps (calls luaL_error) on failure.
--- a/engine/script/src/dmsdk/script/script.h
+++ b/engine/script/src/dmsdk/script/script.h
@@ -641,6 +641,15 @@ namespace dmScript
     void UnrefInInstance(lua_State* L, int ref);
 
     /*#
+     * Get the current game object URL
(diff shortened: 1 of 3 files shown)


engine/script/src/dmsdk/script/script.h
```diff
@@ -640,6 +640,15 @@ namespace dmScript
...
     void UnrefInInstance(lua_State* L, int ref);
 
+    /*#
+     * Get the current game object URL
+     * @name CheckURL
+     * @param L [type:lua_State*] Lua state
+     * @param out_url [type:dmMessage::URL*] where to store the result
+     * @return result [type:bool] true if successful
+     */
+    bool CheckURL(lua_State* L, dmMessage::URL* out_url);
```
JCash: Some api inconsistencies:
* The `CheckXxx` api should return the value type, which is (dmMessage::URL*) in this case.
* It should accept the `index` of the argument on the Lua stack
* Each type `Xxx`should have `IsXxx()`, `CheckXxx()`, `ToXxx()` and `PushXxx()`,

Note that we already have these functions, and they should only require to be moved, and update the documentation.

JCash (changes requested):

engine/script/src/dmsdk/script/script.h
Gert-Jan: This CheckURL is more like [CheckGOInstance(lua_State *L)](https://github.com/defold/defold/blob/2f0c5de4c8dd4ca3afd3dcf26fac5ef788c1866a/engine/gamesys/src/gamesys/gamesys_script.cpp#L63) without an index. where the URL is retrieved through a lua metamethod.

I didn't return a URL value here because as far as I can tell there is no clear null check for a URL value. So then the question becomes: how does the user know when CheckURL failed?

I'll add the other URL getter type functions.

Gert-Jan: Ok looked into it a bit more, I'll just copy what is done for To/Check Vector4 where for both a pointer is returned (as opposed to for example To/Check Hash where the former returns a pointer and the latter a value).

Gert-Jan: > Ok looked into it a bit more, I'll just copy what is done for To/Check Vector4 where for both a pointer is returned (as opposed to for example To/Check Hash where the former returns a pointer and the latter a value).

This is not going to work because the URL is retrieved [from a lua metamethod](https://github.com/defold/defold/blob/63b978edf497b248b50215e44f715511d3ffb156/engine/script/src/script.cpp#L852), which pushes the url on the lua stack. So in order to return a pointer to the URL you would need to keep it around on the lua stack but ideally you would not want to change the lua state based on a CheckXxx() function.

So I'm back at the original bool on success with an out_url arg.

Gert-Jan:
Updated the title and description of this PR to cover the other URL related functions added to dmsdk.
@JCash Check the [resolved conversation up here](https://github.com/defold/defold/pull/10784#discussion_r2126492860) to see why one of the CheckURL() function doesn't have an index and returns it's result through a pointer.

engine/script/src/script.cpp
```diff
@@ -872,6 +872,14 @@ namespace dmScript
         return false;
     }
 
+    bool CheckURL(lua_State* L, dmMessage::URL* out_url) {
```
JCash: style: the curly brace should be on a new line.

engine/script/src/script.cpp
```diff
@@ -872,6 +872,14 @@ namespace dmScript
...
+        bool result = GetURL(L, out_url);
+        if (!result) {
+            luaL_error(L, "no URL could be found in the current script environment");
```
JCash: The recommendation from Lua is to always do a return: "return luaL_error()"

Also type: "no" -> "No"

engine/script/src/dmsdk/script/script.h
JCash: Right, yeah, it is a good use case for getting the url of the current script.

> how does the user know when CheckURL failed?

The "CheckXxx" should always throw an error if they fail.
If we want to implement something more graceful, that's what we use the IsXxx/ToXxx for.

JCash (changes requested):
lgtm:
Minor style changes needed

AGulev:
It also fixes this issue https://github.com/defold/defold/issues/7220

Please, add it into your PR description (see explanation here: https://github.com/defold/defold/pull/10767#issuecomment-2939849009)

Gert-Jan:
> It also fixes this issue #7220
> 
> Please, add it into your PR description (see explanation here: [#10767 (comment)](https://github.com/defold/defold/pull/10767#issuecomment-2939849009))

Good to see I'm not the only one who would want this. Added!

JCash (approved):

PR #10788, opened 2025-06-04
Moved GetCollectionByHash() to dmsdk.

Gert-Jan:
Moved GetCollectionByHash() to dmsdk to allow for URLs to be turned into HCollection/HInstance/HComponent.

### Technical changes
* Specific use case: going from absolute URL to collection/instance/component without a lua context.
* Example:
```
// not called from lua
dmGameObject::HCollection collection = dmGameObject::GetCollectionByHash(
    register, absoluteUrl.m_Socket);
dmGameObject::HInstance instance = dmGameObject::GetInstanceFromIdentifier(collection, factoryUrl.m_Path);
dmGameObject::HComponentWorld world;
dmGameObject::HComponent component;
uint32_t componentTypeIndex = 0;
dmGameObject::Result result = dmGameObject::GetComponent(instance, absoluteUrl.m_Fragment, &componentTypeIndex, &component, &world);
```

### DIFF
--- a/engine/gameobject/src/dmsdk/gameobject/gameobject.h
+++ b/engine/gameobject/src/dmsdk/gameobject/gameobject.h
@@ -536,7 +536,7 @@ namespace dmGameObject
      * @param socket_name [type: dmhash_t] Socket name
      * @return collection [type: dmGameObject::HCollection] The collection the specified instance belongs to
      */
-    HCollection GetCollectionByHash(HRegister register, dmhash_t socket_name);
+    HCollection GetCollectionByHash(HRegister regist, dmhash_t socket_name);
 
     /*#
      * Create a new gameobject instance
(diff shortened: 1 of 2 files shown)


engine/gameobject/src/dmsdk/gameobject/gameobject.h
```diff
@@ -528,6 +528,16 @@ namespace dmGameObject
...
 
+    /*#
+     * Retrieve a collection by socket name hash
+     * Note: in native extensions, the register can be retrieved during init using dmEngine::GetGameObjectRegister(dmExtension::AppParams *params)
+     * @name GetCollectionByHash
+     * @param regist [type: dmGameObject::HRegister] Register
+     * @param socket_name [type: dmhash_t] Socket name
+     * @return collection [type: dmGameObject::HCollection] The collection the specified instance belongs to
+     */
+    HCollection GetCollectionByHash(HRegister register, dmhash_t socket_name);
```
JCash: Two issues:
* `register` is a keyword. We use `regist` as a workaround. See prior code.
* For a function that is "moved",  I'd expect a removal somewhere?

JCash (changes requested):

engine/gameobject/src/dmsdk/gameobject/gameobject.h
Gert-Jan: > * `register` is a keyword. We use `regist` as a workaround. See prior code.

Interesting, I think it compiled just fine because a header doesn't care about the var name, only the type.

Gert-Jan:
Changed as requested.

JCash (approved):
lgtm

PR #10789, opened 2025-06-04
Fixed issue with font cache invalidation

AGulev:
Fixed issue where the cache stayed invalidated after updating to a newer version of Defold where the font or glyph bank output format was changed.  
Usually, it appeared as a `Unable to create resource: /_generated_font_f8188ecab1e82fcc.glyph_bankc: FORMAT_ERROR` and could only be fixed by removing the `build` folder, which is no longer necessary.

Fix https://github.com/defold/defold/issues/10786

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/FontBuilder.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/FontBuilder.java
@@ -52,9 +52,8 @@ public class FontBuilder extends ProtoBuilder<FontDesc.Builder> {
 
     @Override
     public void build(Task task) throws CompileExceptionError, IOException {
-        FontDesc.Builder fontDescbuilder = FontDesc.newBuilder();
-        ProtoUtil.merge(task.firstInput(), fontDescbuilder);
-        FontDesc fontDesc = fontDescbuilder.build();
+        FontDesc.Builder builder = getSrcBuilder(task.firstInput());
+        FontDesc fontDesc = builder.build();
 
         BuilderUtil.checkResource(this.project, task.firstInput(), "material", fontDesc.getMaterial());
 
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/FontBuilder.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/FontBuilder.java
@@ -16,9 +16,10 @@ package com.dynamo.bob.pipeline;
 
 import java.io.IOException;
 
-import com.dynamo.bob.Builder;
 import com.dynamo.bob.BuilderParams;
 import com.dynamo.bob.CompileExceptionError;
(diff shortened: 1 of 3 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/FontBuilder.java
```diff
@@ -16,24 +16,25 @@
...
 import com.dynamo.bob.font.Fontc;
 import com.dynamo.bob.fs.IResource;
 
 import com.dynamo.render.proto.Font.FontDesc;
 import com.dynamo.render.proto.Font.FontMap;
 
+@ProtoParams(srcClass = FontDesc.class, messageClass = FontDesc.class)
 @BuilderParams(name = "Font", inExts = ".font", outExt = ".fontc")
-public class FontBuilder extends Builder  {
+public class FontBuilder extends ProtoBuilder<FontDesc.Builder> {
```
AGulev, line 32: When our class scanner finds a `ProtoBuilder`, it gets `messageClass` and uses its fields as part of the cache key.  
This is one of the reasons why it's necessary to use `ProtoBuilder` for all the resources that produce protobuf messages.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/GlyphBankBuilder.java
```diff
@@ -22,25 +22,27 @@
...
+import com.dynamo.bob.ProtoParams;
 import com.dynamo.bob.Task;
 import com.dynamo.bob.fs.IResource;
 
 import com.dynamo.bob.font.Fontc;
 import com.dynamo.bob.font.Fontc.FontResourceResolver;
+import com.dynamo.render.proto.Font.GlyphBank;
 import com.dynamo.render.proto.Font.FontDesc;
 
+@ProtoParams(srcClass = FontDesc.class, messageClass = GlyphBank.class)
```
AGulev, line 37: It's fine to have different classes for input and output—we support it.  
This could be useful in case we decide to add extra editor-only info, for example.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/FontBuilder.java
britzl: Are there more such builders that we need to change or is this the only one?

britzl (approved):

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/FontBuilder.java
AGulev: I've checked the rest of them, and they look fine

AGulev: After One more check I found AnimationSetDesc
But We rarely touch it

PR #10797, opened 2025-06-05
Collision physics messages should be still sent when collision events are disabled in only one of the collision objects

### LINKED ISSUE #10796: Disabling `Generate Collision Events` leads to undefined behaviour in 3D physics
Disabling `Generate Collision Events` or `Generate Contact Events` options leads to undefined behaviour in 3D physics: if one of the objects has one of the options disabled, the other objects _**may**_ stop generating `collision_point_response`/`collision_response` events.

**To Reproduce (REQUIRED)**
1) Disable this option on one of the objects in the scene
2) Add several such objects
3) Collide them
4) The remaining objects _**may**_ stop generating `collision_point_response`/`collision_response`  events.

Physics shouldn't drop collision callbacks handling.

  - Version: 1.10.1, 1.10.2 BETA

**Platforms (REQUIRED):**
 - Platforms: Windows, HTML5

[issue_3d_physics_events.zip](https://github.com/user-attachments/files/20604507/issue_3d_physics_events.zip)

In the repo case, the debug print should output a number greater than zero.

aglitchman:
Fixed a bug where disabling `Generate Collision Events` or `Generate Contact Events` in a collision object could cause collision physics messages to stop being sent in 3D physics for objects where this was not disabled.

Fixes #10796

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_collision_object.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_collision_object.cpp
@@ -552,8 +552,8 @@ namespace dmGameSystem
 
         bool event_supported_a = SupportsEvent(component_a, EVENT_MASK_TRIGGER);
         bool event_supported_b = SupportsEvent(component_b, EVENT_MASK_TRIGGER);
-        if (!event_supported_a && event_supported_a == event_supported_b)
-            return; // Neither supported this event
+        if (!event_supported_a && !event_supported_b)
+            return; // We early out because neither supported this event
 
         dmGameObject::HInstance instance_a = component_a->m_Instance;
         dmGameObject::HInstance instance_b = component_b->m_Instance;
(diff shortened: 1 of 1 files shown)


engine/gamesys/src/gamesys/components/comp_collision_object.cpp
```diff
@@ -616,7 +616,7 @@ namespace dmGameSystem
             bool event_supported_a = SupportsEvent(component_a, EVENT_MASK_COLLISION);
             bool event_supported_b = SupportsEvent(component_b, EVENT_MASK_COLLISION);
             if (!event_supported_a && event_supported_a == event_supported_b)
-                return false; // Neither supported this event
+                return true; // Neither supported this event
```
AGulev: But it checks is two events don't support it:
if event_supported_a == false // not supported

event_supported_a == event_supported_b // both not supported.

so condition the condition is
if event_a doesn't support the event, and A is equal to B

JCash: Thanks for the fix!

After some discussion, we suggest making it even clearer:
```
            if (!event_supported_a && !event_supported_b)
                return true; // We early out because neither supported this event
```

JCash (changes requested):
good fix
however we'd like a small style/comment change as well.

aglitchman:
> however we'd like a small style/comment change as well.

Many thanks! Your suggested change makes the code much clearer. I updated all three places - trigger/collision/contact.

(The bug does not apply to the trigger, but the trigger has the same condition — let it be the same everywhere.)

JCash (approved):

PR #10800, opened 2025-06-05
Moved PropertyContainerCreateFromLua() to dmsdk

Gert-Jan:
Added `dmScript:: PropertyContainerCreateFromLua()` to `dmsdk/gameobject/script.h`

### Technical changes
* Placed `PropertyContainerCreateFromLua()` in `dmsdk/gameobject/script.h` [per request](https://github.com/defold/defold/pull/10800#discussion_r2137476565), left the implementation in the dmGameObject namespace, because it contains a great number of references within the dmGameObject namespace.


engine/gameobject/src/dmsdk/gameobject/gameobject_props_lua.h
```diff
@@ -0,0 +1,33 @@
...
+namespace dmGameObject
+{
+    /*#
+     * Creates a property container from a lua table
+     * @name PropertyContainerCreateFromLua
+     * @param L [type: lua_State*] The lua state
+     * @param index [type: int] The lua stack index of the lua table
+     * @return container [type: HPropertyContainer] The property container
+     */
+    HPropertyContainer PropertyContainerCreateFromLua(lua_State* L, int index);
```
JCash: As this is a new file, it would need a @document paragraph (see e.g. gameobject.h)

However, I think we should move this function to gameobject/src/dmsdk/gameobject/script.h as it's part of the scripting api.
That would of course make it `dmScript:: PropertyContainerCreateFromLua()`, but I think that's a good change.

JCash (changes requested):
overall good, but I'd like the public function to be placed in the script.h file instead.

engine/gameobject/src/dmsdk/gameobject/gameobject_props_lua.h
Gert-Jan: Done and updated the PR description.

JCash (approved):
lgtm

PR #10804, opened 2025-06-05
Make sure that if the user doesn't serialize a table with a hash key, the version of the data remains backward compatible

AGulev:
It fixes the issue where a file can't be loaded anymore in Defold 1.10.1 if it was saved in Defold 1.10.2, even if no hash keys were used.

### DIFF
--- a/engine/script/src/script_private.h
+++ b/engine/script/src/script_private.h
@@ -81,6 +81,18 @@ namespace dmScript
         int                         m_ContextWeakTableRef;
     };
 
+    struct TableHeader
+    {
+        uint32_t m_Magic;
+        uint32_t m_Version;
+
+        TableHeader() : m_Magic(0), m_Version(0)
+        {
+        }
+    };
+
+    const char* ReadHeader(const char* buffer, TableHeader& header);
+
     HContext GetScriptContext(lua_State* L);
 
     bool ResolvePath(lua_State* L, const char* path, uint32_t path_size, dmhash_t& out_hash);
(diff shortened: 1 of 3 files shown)


engine/script/src/script_private.h
```diff
@@ -81,6 +81,18 @@ namespace dmScript
...
+    {
+        uint32_t m_Magic;
+        uint32_t m_Version;
+
+        TableHeader() : m_Magic(0), m_Version(0)
+        {
+        }
+    };
+
+    const char* ReadHeader(const char* buffer, TableHeader& header);
```
AGulev, line 94: Moved it to the private script header to be able to use it in tests

engine/script/src/script_table.cpp
```diff
@@ -37,6 +37,8 @@ namespace dmScript
 {
     const int TABLE_MAGIC = 0x42544448;
     const uint32_t TABLE_VERSION_CURRENT = 5;
+    const uint32_t TABLE_VERSION_BASE = 4;
+    const uint32_t TABLE_VERSION_HASH_KEYS_ADDED = TABLE_VERSION_CURRENT;
```
AGulev, line 41: Use per-feature version

engine/script/src/script_table.cpp
```diff
@@ -751,7 +744,7 @@ namespace dmScript
 
             TableHeader* header = (TableHeader*)buffer;
             header->m_Magic = TABLE_MAGIC;
-            header->m_Version = TABLE_VERSION_CURRENT;
+            header->m_Version = TABLE_VERSION_BASE;
```
AGulev, line 747: Use base version by default

engine/script/src/script_table.cpp
```diff
@@ -528,6 +520,7 @@ namespace dmScript
             }
             else if (key_hash)
             {
+                header.m_Version = TABLE_VERSION_HASH_KEYS_ADDED;
```
AGulev, line 523: Upgrade version in case hash-key used

britzl (approved):

PR #10808, opened 2025-06-08
Generate and inject root signature into HLSL files when running on windows

Jhonnyg:
This PR adds:
* Generate and inject root signatures into hlsl shaders
* Compute shader support and image texture (i.e storage textures) support for DX12

This means that we can now generate raw hlsl shaders with injected root signatures (via reflection) that can now be compiled into binary hlsl shaders for xbox via xdc!


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/shader/ShaderCompilePipeline.java
```diff
@@ -92,7 +93,7 @@ private static Integer ShaderLanguageToVersion(ShaderDesc.Language shaderLanguag
             case LANGUAGE_GLES_SM300 -> 300;
             case LANGUAGE_GLSL_SM330 -> 330;
             case LANGUAGE_GLSL_SM430 -> 430;
-            case LANGUAGE_HLSL       -> 50;
+            case LANGUAGE_HLSL       -> 51;
```
Jhonnyg, line 96: Maybe we need HLSL_51 and so forth here. XBox requires 6.0+ i think. It would be _even_ better if we could just pass the version into the shader blob instead of having these enums with specific versions?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/shader/ShaderCompilePipeline.java
```diff
@@ -439,23 +468,30 @@ public byte[] crossCompile(ShaderDesc.ShaderType shaderType, ShaderDesc.Language
...
             if (ShaderLanguageIsGLSL(shaderLanguage)) {
-                bytes = RemapTextureSamplers(module.spirvReflector.getTextures(), new String(bytes));
+                result.data = RemapTextureSamplers(module.spirvReflector.getTextures(), new String(result.data));
+            }
+
+            // If we are cross-compiling a compute shader to HLSL, we need to inject an extra resource into the reflection,
+            // because there is no gl_NumWorkGroups equivalent in HLSL
+            if (shaderType == ShaderDesc.ShaderType.SHADER_TYPE_COMPUTE && result.hLSLNumWorkGroupsId >= 0) {
+                injectHLSLNumWorkGroupsIdIntoReflection(module, result.hLSLNumWorkGroupsId);
             }
```
Jhonnyg, line 493: Hm, one problem with this is if we need to store multilple hlsl files in the same shader file (e.g building for windows + xbox), they could potentially be stored at different bind points but only exist once in the reflection 🤔

engine/shaderc/src/shaderc_private.h
```diff
@@ -92,6 +92,26 @@ namespace dmShaderc
...
+        uint8_t     m_CombinedId;
+        const char* m_ImageName;
+        uint8_t     m_ImageId;
+        const char* m_SamplerName;
+        uint8_t     m_SamplerId;
+    };
+
+    void GetCombinedSamplerMapSPIRV(HShaderContext context, ShaderCompilerSPVC* compiler, dmArray<CombinedSampler>& samplers);
+
+#ifdef DM_BINARY_HLSL_SUPPORTED
```
Jhonnyg, line 114: Hm, maybe some other name here.

engine/graphics/src/dx12/graphics_dx12.cpp
```diff
@@ -77,13 +80,6 @@ namespace dmGraphics
...
         m_UseValidationLayers     = params.m_UseValidationLayers;
 
-        m_TextureFormatSupport |= 1 << TEXTURE_FORMAT_LUMINANCE;
-        m_TextureFormatSupport |= 1 << TEXTURE_FORMAT_LUMINANCE_ALPHA;
-        m_TextureFormatSupport |= 1 << TEXTURE_FORMAT_RGB;
-        m_TextureFormatSupport |= 1 << TEXTURE_FORMAT_RGBA;
-        m_TextureFormatSupport |= 1 << TEXTURE_FORMAT_RGB_16BPP;
-        m_TextureFormatSupport |= 1 << TEXTURE_FORMAT_RGBA_16BPP;
-
         assert(dmPlatform::GetWindowStateParam(m_Window, dmPlatform::WINDOW_STATE_OPENED));
```
Jhonnyg, line 83: These are queried in the same way as we do for vulkan (later down in this file).

engine/graphics/src/dx12/graphics_dx12.cpp
```diff
@@ -177,17 +173,6 @@ namespace dmGraphics
         }
     }
 
-    static void CreateRootSignature(DX12Context* context, CD3DX12_ROOT_SIGNATURE_DESC* desc, DX12ShaderProgram* program)
```
Jhonnyg, line 180: We will have to re-add this function later on to support building shaders on non-windows platforms.

engine/graphics/src/dx12/graphics_dx12.cpp
```diff
@@ -2150,6 +2282,13 @@ namespace dmGraphics
             return hr;
         }
 
+
+        hr = D3DGetBlobPart(shader->m_ShaderBlob->GetBufferPointer(), shader->m_ShaderBlob->GetBufferSize(), D3D_BLOB_ROOT_SIGNATURE, 0, &shader->m_RootSignatureBlob);
+        if (FAILED(hr))
+        {
+            // TODO: Extract reflection and generate root signature that way. There is no embedded root signature (which can happen when content has been built on non-windows platforms)
```
Jhonnyg, line 2289: we can do this later.

engine/shaderc/src/shaderc_spvc.cpp
```diff
@@ -464,13 +542,33 @@ namespace dmShaderc
         }
         else if (compiler->m_BaseCompiler.m_Language == SHADER_LANGUAGE_HLSL)
         {
+            assert(options.m_Version > 50);
```
Jhonnyg, line 521: We can't support anything less than 51 right now.

engine/shaderc/src/shaderc_hlsl.cpp
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    ShaderCompileResult* CompileRawHLSLToBinary(HShaderContext context, HShaderCompiler compiler, ShaderCompileResult* raw_hlsl)
+    {
+        ID3DBlob* shader_blob = NULL;
+        ID3DBlob* error_blob = NULL;
+
+        const char* profile = "";
+
+        // TODO: correct versions here
```
Jhonnyg, line 305: We need to fix this later for when we want this to work for xbox as well.

JCash (dismissed):
Lgtm!

PR #10810, opened 2025-06-09
Edit GUI layouts using editor scripts

vlaaad:
Now you can edit a list of GUI layouts using editor scripts, e.g.:
```lua
editor.transact({
    editor.tx.add("/main.gui", layouts, {name = "Landscape"}),
    editor.tx.add("/main.gui", layouts, {name = "Portrait"})
})
```

Related to #8504

### DIFF
--- a/editor/src/clj/editor/code/script.clj
+++ b/editor/src/clj/editor/code/script.clj
@@ -75,7 +75,7 @@
               (cond
                 escaped (recur (inc i) token in-quote false skip-rest last-non-space tokens)
                 (= ch \\) (recur (inc i) token in-quote true skip-rest last-non-space tokens)
-                (= ch in-quote) (recur (inc i) token nil false skip-rest last-non-space tokens)
+                (= ch (.charValue ^Character in-quote)) (recur (inc i) token nil false skip-rest last-non-space tokens)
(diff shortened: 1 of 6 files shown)


editor/src/clj/editor/code/script.clj
```diff
@@ -75,7 +75,7 @@
               (cond
                 escaped (recur (inc i) token in-quote false skip-rest last-non-space tokens)
                 (= ch \\) (recur (inc i) token in-quote true skip-rest last-non-space tokens)
-                (= ch in-quote) (recur (inc i) token nil false skip-rest last-non-space tokens)
+                (= ch (.charValue ^Character in-quote)) (recur (inc i) token nil false skip-rest last-non-space tokens)
```
vlaaad, line 78: Unrelated: fixed boxed math warning.

matgis (changes requested):
I think we may need to restrict the allowed layout names to the ones listed as `unused-display-profiles` in `LayoutsNode`, like we do when adding layouts through the UI?

We also want to prevent adding duplicate layout names within the same transaction to avoid some potentially painful states.

matgis (approved):
LGTM!

PR #10812, opened 2025-06-09
Edit GUI spine scenes list using editor scripts

vlaaad:
Now, when using extension spine, it's possible to edit the list of spine scenes in the GUI:
```lua
editor.transact({
    editor.tx.add("/main.gui", "spine_scenes", {
        spine_scene = "/bg.spinescene"
    }),
})
```

Related to #8504

### DIFF
--- a/editor/project.clj
+++ b/editor/project.clj
@@ -163,9 +163,9 @@
                       ;; hide warnings about illegal reflective access by clojure
                       "--add-opens=java.xml/com.sun.org.apache.xerces.internal.jaxp=ALL-UNNAMED"
                       ;; used in editor.scene$read_to_buffered_image
-                      "--add-opens=java.desktop/sun.awt.image=ALL-UNNAMED"
+                      "--add-opens=java.desktop/sun.awt.image=ALL-UNNAMED"]
                       ;; "-XX:MaxJavaStackTraceDepth=1073741823"
-                      ]
+
   :main ^:skip-aot   com.defold.editor.Main
 
   :uberjar-exclusions [#"^natives/"]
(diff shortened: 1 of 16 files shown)


editor/project.clj
```diff
@@ -287,7 +287,7 @@
                                 :jvm-opts          ["-Ddefold.extension.lua-preprocessor.url=https://github.com/defold/extension-lua-preprocessor/archive/refs/tags/1.1.3.zip"
                                                     "-Ddefold.extension.rive.url=https://github.com/defold/extension-rive/archive/refs/tags/3.9.0.zip"
                                                     "-Ddefold.extension.simpledata.url=https://github.com/defold/extension-simpledata/archive/refs/tags/v1.1.0.zip"
-                                                    "-Ddefold.extension.spine.url=https://github.com/defold/extension-spine/archive/refs/tags/3.7.0.zip"
+                                                    "-Ddefold.extension.spine.url=https://github.com/defold/extension-spine/archive/refs/heads/issue-8054.zip"
```
vlaaad, line 290: The idea is to merge this PR, then do a spine release, then update the ref here.

vlaaad:
See https://github.com/defold/extension-spine/pull/207

editor/src/clj/editor/attachment.clj
```diff
@@ -29,6 +29,7 @@
   ;; reorder-fn: fn of reordered-nodes -> txs
   (atom {:add {} :get {} :reorder {}}))
 
+;; SDK api
```
matgis, line 23: Hmm. Depending on how this is used, having a `state-atom` like this will pollute the tests following a test that brings in the Spine plugin.

We have a similar issue around defmultis and protocols we expose through the SDK api, but most of those will not have any effect unless a matching node enters the graph. How is this `register!` function different?

Would it make sense to move the state to the workspace node and have the plugin supply a workspace argument before exposing this as SDK api?

editor/src/clj/editor/editor_extensions/graph.clj
```diff
@@ -368,6 +369,7 @@
   (fn init-attachment-dispatch-fn [_evaluation-context _rt _project _parent-node-id child-node-type _child-node-id _attachment]
     (:k child-node-type)))
 
+;; SDK api
```
matgis: Presumably it is the `defmulti` that is exposed as SDK api, not this default implementation?

matgis (changes requested):
Loading the spine plugin in an editor test seems to potentially leak state to subsequent tests. Perhaps we want to address this before exposing this functionality to editor plugins?

vlaaad:
@matgis I moved the state to the graph (workspace property).

matgis (approved):
LGTM!

PR #10817, opened 2025-06-09
Fix sys.get_sys_info() returning junk `system_version` on web platform

### LINKED ISSUE #10815: System version from `sys.get_sys_info()` may contain garbage on HTML5
A user of [Sentinel](https://github.com/indiesoftby/defold-sentinel), a Sentry SDK for Defold, reported that his bug reports were not reaching the Sentry servers. Upon investigation, it turned out that the system version info included in the report being sent may have random binary values, so it was not being accepted by Sentry.

**To Reproduce (REQUIRED)**
Steps to reproduce the behavior: `json.encode(sys.get_sys_info())` - it **may** have invalid data in the `system_version` field.

`sys.get_sys_info()` should always return valid data.

  - Version: 1.10.1, 1.10.2.

**Platforms (REQUIRED):**
 - Platforms: HTML5

aglitchman:
The `sys.get_sys_info()` function may have returned junk in the `system_version` field for the web platform because it used an uninitialised structure as a source of data.

Fixes #10815

### Technical changes
* For Emscripten, data from the uninitialised variable ‘uts’ was used.
* I split the code for Linux and Emscripten platforms for readability.

### DIFF
--- a/engine/dlib/src/dlib/sys_posix.cpp
+++ b/engine/dlib/src/dlib/sys_posix.cpp
@@ -750,40 +750,45 @@ namespace dmSys
 #endif
     }
 
-#if (defined(__linux__) && !defined(__ANDROID__)) || defined(__EMSCRIPTEN__)
+#if defined(__EMSCRIPTEN__)
+    void GetSystemInfo(SystemInfo* info)
+    {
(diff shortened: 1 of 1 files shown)


AGulev (approved):
👍

PR #10823, opened 2025-06-11
The textures that were set using `go.set()/gui.set()` for the GUI component are internally used as dynamic textures.

### LINKED ISSUE #10472: Editable limit for number of atlases for gui scene
_Is it a private platform specific feature? (i.e. would it break your NDA?)_
NO

It's impossible to use more than 128 atlases in gui scene
https://github.com/defold/defold/blob/master/engine/gamesys/src/gamesys/components/comp_gui.cpp#L931

Add an editable property for gui scene to control the limit of atlases

No alternatives except an optimization of the content of atlases

AGulev:
Now, all the textures that were set to a `GUI` component from outside are internally considered dynamic GUI textures in the engine.

⚠️ Pay attention: if you set textures into the GUI this way, you may need to increase `Max Dynamic Textures` (default is 128) in the GUI component.

Fix https://github.com/defold/defold/issues/10472

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_gui.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_gui.cpp
@@ -2355,7 +2355,12 @@ namespace dmGameSystem
         GuiComponent* component = (GuiComponent*)dmGui::GetSceneUserData(scene);
         char resource_path[dmResource::RESOURCE_PATH_MAX];
         dmhash_t resolved_path_hash = ResolveDynamicTexturePath(component, path_hash, resource_path, sizeof(resource_path));
-        ReleaseDynamicResource(dmGameObject::GetFactory(component->m_Instance), dmGameObject::GetCollection(component->m_Instance), resolved_path_hash);
+        dmResource::HFactory factory = dmGameObject::GetFactory(component->m_Instance);
+        ResourceDescriptor* rd = dmResource::FindByHash(factory, resolved_path_hash);
+        if (rd)
+        {
+            dmResource::Release(factory, dmResource::GetResource(rd));
+        }
     }
 
     static void SetTextureResourceCallback(dmGui::HScene scene, const dmhash_t path_hash, uint32_t width, uint32_t height, dmImage::Type type, const void* buffer)
--- a/engine/gamesys/src/gamesys/components/comp_gui.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_gui.cpp
@@ -931,7 +931,7 @@ namespace dmGameSystem
         scene_params.m_UserData = gui_component;
         scene_params.m_MaxFonts = 64;
         scene_params.m_MaxDynamicTextures = scene_desc->m_MaxDynamicTextures;
-        int texture_count = scene_resource->m_GuiTextureSets.Size();
+        uint32_t texture_count = scene_resource->m_GuiTextureSets.Size();
         scene_params.m_MaxTextures = texture_count > 0 ? texture_count : 1;
         scene_params.m_MaxMaterials = 16;
         scene_params.m_MaxAnimations = gui_world->m_MaxAnimationCount;
--- a/engine/gamesys/src/gamesys/components/comp_gui.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_gui.cpp
@@ -931,7 +931,8 @@ namespace dmGameSystem
         scene_params.m_UserData = gui_component;
         scene_params.m_MaxFonts = 64;
         scene_params.m_MaxDynamicTextures = scene_desc->m_MaxDynamicTextures;
-        scene_params.m_MaxTextures = 128;
+        int texture_count = scene_resource->m_GuiTextureSets.Size();
+        scene_params.m_MaxTextures = texture_count > 0 ? texture_count : 1;
         scene_params.m_MaxMaterials = 16;
         scene_params.m_MaxAnimations = gui_world->m_MaxAnimationCount;
         scene_params.m_MaxParticlefx = gui_world->m_MaxParticleFXCount;
@@ -2856,7 +2857,7 @@ namespace dmGameSystem
             if (res == dmGameObject::PROPERTY_RESULT_OK)
             {
                 dmGraphics::HTexture texture = texture_source->m_Texture->m_Texture;
-                dmGui::Result r = dmGui::AddTexture(gui_component->m_Scene, params.m_Options.m_Key, (dmGui::HTextureSource) texture_source, dmGui::NODE_TEXTURE_TYPE_TEXTURE_SET, dmGraphics::GetOriginalTextureWidth(texture), dmGraphics::GetOriginalTextureHeight(texture));
+                dmGui::Result r = dmGui::AddDynamicTexture(gui_component->m_Scene, params.m_Options.m_Key, (dmGui::HTextureSource) texture_source, dmGui::NODE_TEXTURE_TYPE_TEXTURE_SET, dmGraphics::GetOriginalTextureWidth(texture), dmGraphics::GetOriginalTextureHeight(texture));
                 if (r != dmGui::RESULT_OK)
                 {
                     dmLogError("Unable to add texture '%s' to scene (%d)", dmHashReverseSafe64(params.m_Options.m_Key),  r);
--- a/engine/gamesys/src/gamesys/gamesys_resource.cpp
+++ b/engine/gamesys/src/gamesys/gamesys_resource.cpp
@@ -209,7 +209,6 @@ namespace dmGameSystem
         }
 
         assert(create_params.m_Collection);
-        dmGameObject::AddDynamicResourceHash(create_params.m_Collection, create_params.m_PathHash);
         *resource_out = resource;
         return dmResource::RESULT_OK;
     }
@@ -265,20 +264,5 @@ namespace dmGameSystem
         upload_params.m_UploadSpecificMipmap  = 1;
 
         return dmResource::SetResource(factory, params.m_PathHash, (void*) &recreate_params);
-    }
-
-    dmResource::Result ReleaseDynamicResource(dmResource::HFactory factory, dmGameObject::HCollection collection, dmhash_t path_hash)
-    {
-        HResourceDescriptor rd = dmResource::FindByHash(factory, path_hash);
-        if (!rd)
-        {
-            return dmResource::RESULT_RESOURCE_NOT_FOUND;
-        }
-
-        // This will remove the entry in the collections list of dynamically allocated resource (if it exists),
-        // but we do the actual release here since we allow releasing arbitrary resources now
-        dmGameObject::RemoveDynamicResourceHash(collection, path_hash);
-        dmResource::Release(factory, dmResource::GetResource(rd));
-        return dmResource::RESULT_OK;
     }
 }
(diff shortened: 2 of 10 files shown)


engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -931,7 +931,8 @@ namespace dmGameSystem
         scene_params.m_UserData = gui_component;
         scene_params.m_MaxFonts = 64;
         scene_params.m_MaxDynamicTextures = scene_desc->m_MaxDynamicTextures;
-        scene_params.m_MaxTextures = 128;
+        int texture_count = scene_resource->m_GuiTextureSets.Size();
+        scene_params.m_MaxTextures = texture_count > 0 ? texture_count : 1;
```
AGulev, line 935: since these textures now used only for the textures we set in the editor, we can allocate the minimal size for this hashtable

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -2355,7 +2356,12 @@ namespace dmGameSystem
         GuiComponent* component = (GuiComponent*)dmGui::GetSceneUserData(scene);
         char resource_path[dmResource::RESOURCE_PATH_MAX];
         dmhash_t resolved_path_hash = ResolveDynamicTexturePath(component, path_hash, resource_path, sizeof(resource_path));
-        ReleaseDynamicResource(dmGameObject::GetFactory(component->m_Instance), dmGameObject::GetCollection(component->m_Instance), resolved_path_hash);
```
AGulev, line 2358: I removed `ReleaseDynamicResource` because there are only two places where we use it, and we don't need it in gui anymore, so only one place

engine/gamesys/src/gamesys/gamesys_resource.cpp
```diff
@@ -209,7 +209,6 @@ namespace dmGameSystem
         }
 
         assert(create_params.m_Collection);
-        dmGameObject::AddDynamicResourceHash(create_params.m_Collection, create_params.m_PathHash);
```
AGulev, line 212: moved out from `gamesys_resource.cpp` to use it the same way as we do it in the other functions

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -930,6 +930,7 @@ static int CreateTexture(lua_State* L)
         return ReportPathError(L, res, create_params.m_PathHash);
     }
 
+    dmGameObject::AddDynamicResourceHash(create_params.m_Collection, create_params.m_PathHash);
```
AGulev, line 933: we always do `AddDynamicResourceHash` outside or resource functions but for some reason in case of `CreateTextureResource` we do it there

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1251,13 +1252,15 @@ static int ReleaseResource(lua_State* L)
...
+    }
     dmGameObject::HInstance sender_instance = dmScript::CheckGOInstance(L);
     dmGameObject::HCollection collection    = dmGameObject::GetCollection(sender_instance);
-
-    if (ReleaseDynamicResource(g_ResourceModule.m_Factory, collection, path_hash) != dmResource::RESULT_OK)
-    {
-        return DM_LUA_ERROR("Could not release resource: %s", dmHashReverseSafe64(path_hash));
-    }
+    dmGameObject::RemoveDynamicResourceHash(collection, path_hash);
+    dmResource::Release(g_ResourceModule.m_Factory, dmResource::GetResource(rd));
```
AGulev, line 1262: logic that replace `ReleaseDynamicResource`

engine/gamesys/src/gamesys/test/gui/gui_max_dynamic_textures.go
```diff
@@ -1,4 +1,4 @@
 components {
   id: "gui"
-  component: "/gui/gui_max_dynamic_textures_256.gui"
+  component: "/gui/gui_max_dynamic_textures_32.gui"
```
AGulev, line 3: wehave limit for resources 64, because of that test wasn't right, so I limited it to a number which is actually less that resource count.

engine/gui/src/gui.cpp
```diff
@@ -496,6 +496,21 @@ namespace dmGui
         return AddTexture(scene, scene->m_Textures, texture_name_hash, texture_source, texture_type, original_width, original_height, (dmImage::Type) -1);
     }
 
+    Result AddDynamicTexture(HScene scene, dmhash_t texture_name_hash, HTextureSource texture_source, NodeTextureType texture_type, uint32_t original_width, uint32_t original_height)
+    {
+        TextureInfo* t = scene->m_DynamicTextures.Get(texture_name_hash);
+        if (t)
```
AGulev, line 507: if we replace texture, we need to make sure prev. one removed (it wasn't necessary before, because we don't have manually managed textures in `m_Textures`)

engine/gui/src/gui.cpp
```diff
@@ -539,9 +554,15 @@ namespace dmGui
         }
     }
 
+    static inline TextureInfo* GetTextureInfo(HScene scene, dmhash_t texture_id)
+    {
+        TextureInfo* texture_info = scene->m_DynamicTextures.Get(texture_id);
```
AGulev, line 564: now we check `m_DynamicTextures` before `m_Textures`, because we don't override textures in `m_Textures` anymore, and to keep functionality (replacing existing texture with a new one) we heed `m_DynamicTextures` to have higher priority that  `m_Textures `

engine/gui/src/gui.cpp
```diff
@@ -616,6 +637,11 @@ namespace dmGui
         HTextureSource res = scene->m_NewTextureResourceCallback(scene, path, width, height, type, data);
         free(data);
 
+        if (!res)
+        {
+            return RESULT_OUT_OF_RESOURCES;
+        }
```
AGulev, line 648: this case was just missed after ne of refactorings. If we didn't ceate a resource, we are out of resource slots

engine/gui/src/gui.h
```diff
@@ -543,6 +543,19 @@ namespace dmGui
...
+     * @note Any nodes connected to the same texture_name will also be connected to the new texture/textureset. This makes this function O(n), where n is #nodes.
+     * @param scene Scene to add the texture/textureset to
+     * @param texture_name_hash Hash of the texture name that will be used in the gui scripts
+     * @param texture The texture to add
+     * @param textureset The textureset to add if animation is used, otherwise zero. If set, texture parameter is expected to be equal to textureset texture.
+     * @param original_width Original With of the texture
+     * @param original_height Original Height of the texture
+     * @return Outcome of the operation
+     */
+    Result AddDynamicTexture(HScene scene, dmhash_t texture_name_hash, HTextureSource texture_source, NodeTextureType texture_type, uint32_t original_width, uint32_t original_height);
```
AGulev, line 557: to be able to set new textures for gui, we need this new function

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -2355,7 +2356,12 @@ namespace dmGameSystem
...
+        dmResource::HFactory factory = dmGameObject::GetFactory(component->m_Instance);
+        ResourceDescriptor* rd = dmResource::FindByHash(factory, resolved_path_hash);
+        if (rd)
+        {
+            dmResource::Release(factory, dmResource::GetResource(rd));
+        }
```
AGulev, line 2364: This is reimplementation of the `ReleaseDynamicResource` but without `RemoveDynamicResourceHash`, we don't really need to manage these textures on collection level when we create them in gui. If users reuse them somewhere, they will be refcounted there, otherwise removed together with gui component remove. I don't see why it's needed to add them to the collection level resource management.

engine/gamesys/src/gamesys/gamesys_resource.cpp
```diff
@@ -266,19 +265,4 @@ namespace dmGameSystem
...
-        HResourceDescriptor rd = dmResource::FindByHash(factory, path_hash);
-        if (!rd)
-        {
-            return dmResource::RESULT_RESOURCE_NOT_FOUND;
-        }
-
-        // This will remove the entry in the collections list of dynamically allocated resource (if it exists),
-        // but we do the actual release here since we allow releasing arbitrary resources now
-        dmGameObject::RemoveDynamicResourceHash(collection, path_hash);
-        dmResource::Release(factory, dmResource::GetResource(rd));
```
AGulev, line 281: I don't use `RemoveDynamicResourceHash` in gui anymore. 
Also, in all the other places we use `RemoveDynamicResourceHash` and `AddDynamicResourceHash` together **with** `gamesys_resource.cpp` APIs, not inside it

engine/gui/src/gui.cpp
```diff
@@ -635,7 +666,15 @@ namespace dmGui
 
         scene->m_DeleteTextureResourceCallback(scene, texture_hash, t->m_TextureSource);
         scene->m_DynamicTextures.Erase(texture_hash);
-        UnassignTexture(scene, texture_hash);
+        t = scene->m_Textures.Get(texture_hash);
+        if (t)
+        {
+            UpdateTexture(scene, texture_hash, t->m_TextureSource, t->m_TextureSourceType);
```
AGulev, line 672: if we "shadowed" the default texture with a new one, we need to update nodes to "unshadow" old texture, because we have just removed the old one.

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -931,7 +931,8 @@ namespace dmGameSystem
         scene_params.m_UserData = gui_component;
         scene_params.m_MaxFonts = 64;
         scene_params.m_MaxDynamicTextures = scene_desc->m_MaxDynamicTextures;
-        scene_params.m_MaxTextures = 128;
+        int texture_count = scene_resource->m_GuiTextureSets.Size();
```
JCash: It's a uint32_t (this produces a new compiler warning)

engine/gamesys/src/gamesys/scripts/script_resource.cpp
```diff
@@ -1251,13 +1252,15 @@ static int ReleaseResource(lua_State* L)
 {
     DM_LUA_STACK_CHECK(L, 0);
     dmhash_t path_hash                      = dmScript::CheckHashOrString(L, 1);
+    HResourceDescriptor rd = dmResource::FindByHash(g_ResourceModule.m_Factory, path_hash);
+    if (!rd) {
+        luaL_error(L, "Could not release resource: %s", dmHashReverseSafe64(path_hash));
+        return 0;
```
JCash: it should be "return luaL_error()" (it's the recommended way from Lua)

JCash (dismissed):
lgtm
some minor issues

engine/gamesys/src/gamesys/components/comp_gui.cpp
AGulev: my bad, thank you!

engine/gamesys/src/gamesys/scripts/script_resource.cpp
AGulev: 👍

JCash (approved):
lgtm

PR #10824, opened 2025-06-11
Add collections and game objects to root on drop

sprocketc:
Fixes #10822

### Technical changes
I initially tried to fix the position recursively (see dc62eb6e54f65eca246ba88a66a30e00442ef875). This approach kind of worked, but I faced a considerable amount of issues that had to be be resolved
- Nodes and InstanceNodes should be handled differently.
- If the `z` position of the parent is not zero, there is an ambiguity regarding the expected `z` position of the dropped object.
- There is a preexisting issue when we add game objects or collections to nested collections. The created nodes are not going to be selected on the outline view, and that makes the visual feedback weird.
- If we look at the linked issue, users seem to prefer adding the dropped nodes at root.

Overall, I think maintaining the parent selection can be fixed, but it introduces a lot of problems to add questionable value. I decided to directly use the root node in this case. After multiple attempts to consistently get to the root node id from the current selection, I realized that the safest way to do that, is to pass it down to the drop fn from the selection controller. This change also fixed some edge cases where there is no selection (e.g. you can hold `Ctrl` and click on the selected root on the outline to remove it from the selection). We also got rid of some error prone calculations by directly using the root id.

### DIFF
--- a/editor/src/clj/editor/atlas.clj
+++ b/editor/src/clj/editor/atlas.clj
@@ -1231,16 +1231,11 @@
     AtlasAnimation (->> (image-resources->image-msgs image-resources)
                         (make-image-nodes-in-animation parent))))
 
-(defn- parent-animation-or-atlas
-  [selection]
-  (or (first (handler/adapt-every selection AtlasAnimation))
-      (some #(core/scope-of-type % AtlasAnimation) selection)
-      (some #(core/scope-of-type % AtlasNode) selection)
-      (first (handler/adapt-every selection AtlasNode))))
-
 (defn- handle-drop
   [root-id selection _workspace _world-pos resources]
-  (let [parent (or (parent-animation-or-atlas selection) root-id)]
+  (let [parent (or (handler/adapt-single selection AtlasAnimation)
+                   (some #(core/scope-of-type % AtlasAnimation) selection)
+                   root-id)]
     (->> resources
          (e/filter image/image-resource?)
          (create-dropped-images parent))))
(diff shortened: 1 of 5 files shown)


Wolfe2x7:
I am not certain that dropping nodes into root is the best way to go; if game objects are instanced as children of whatever is selected, that is probably fine too. The worst bit is the way GOs appeared in an unexpected place.

sprocketc:
> I am not certain that dropping nodes into root is the best way to go; if game objects are instanced as children of whatever is selected, that is probably fine too.

@Wolfe2x7 I agree with this. The problem is that when you create an object (in this case by dropping), it is going to be selected by default. I believe this is useful, and I don't think we should change that. So when you consecutively drop items to the scene, you might get a nested tree, like the one you described. Using the root node is the only alternative I could think of, but I am open to suggestions. 

I am also worried that being aware of the current selection while dropping, is an additional mental overhead. There are also various cases that are not straightforward. What if we have multiple selected elements, and more that one is eligible to be a parent of the dropped object? The parent object might also be rotated or have a non zero Z position value. Should we try to use its intersection with the mouse position in this case? I think supporting dropping assets to outline nodes to explicitly select a parent, might be more useful, but I am not sure either.

Wolfe2x7:
I just compared the behavior to Godot Engine and I stand corrected; in Godot nodes (game objects) always drop into the owner of the scene (the collection file being edited), so that is a precedent users may expect.

matgis (approved):
Makes sense to me!

PR #10830, opened 2025-06-12
Scene view tools reflect behavior changes from held modifier keys without moving the mouse

### LINKED ISSUE #9366: Scene view tools should reflect behavior changes from held modifier keys without moving the mouse
Currently, some tools like the Tile Editor Draw Tool behave differently when certain modifier keys are held. We should visualize this behavior difference immediately when the user holds down the modifier key.

AGulev:
Now holding modifier keys such as Alt, Ctrl, Shift, and Cmd sends its own event, which helps reflect cursor changes immediately without needing to move the cursor.  
It's important, for example, for the tilemap editor, where using modifier keys changes the behavior of the cursor (e.g. Shift+Alt for tile eraser).

Fix https://github.com/defold/defold/issues/9366

### DIFF
--- a/editor/src/clj/editor/scene.clj
+++ b/editor/src/clj/editor/scene.clj
@@ -1467,7 +1467,7 @@
                             (catch Throwable error
                               (reset! process-events? false)
                               (error-reporting/report-exception! error)))))
-        simulate-mouse-on-modifier-keys! (fn [e]
+        simulate-mouse-on-modifier-keys! (fn [^KeyEvent e]
                                            (when (and @process-events? (.isEmpty (.getText e)))
                                              (when-let [last-action (ui/user-data parent ::last-mouse-action)]
                                                (let [updated-action (assoc last-action
--- a/editor/src/clj/editor/scene.clj
+++ b/editor/src/clj/editor/scene.clj
@@ -1444,27 +1444,38 @@
 
 (defn register-event-handler! [^Parent parent view-id]
   (let [process-events? (atom true)
-        event-handler   (ui/event-handler e
-                          (when @process-events?
-                            (try
-                              (profiler/profile "input-event" -1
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/scene.clj
```diff
@@ -1444,27 +1444,38 @@
...
+                                    x (:x action)
+                                    y (:y action)
+                                    pos [x y 0.0]
+                                    picking-rect (selection/calc-picking-rect pos pos)]
+                                (when (= :mouse-pressed (:type action))
+                                  ;; Request focus and consume event to prevent someone else from stealing focus
+                                  (.requestFocus parent)
+                                  (.consume e))
+                                (when (= :mouse-moved (:type action))
+                                  (ui/user-data! parent ::last-mouse-action action))
```
AGulev, line 1461: two lines I added: save last action

editor/src/clj/editor/scene.clj
```diff
@@ -1444,27 +1444,38 @@
...
+                                (g/transact
+                                  (concat
+                                    (g/set-property view-id :cursor-pos [x y])
+                                    (g/set-property view-id :tool-picking-rect picking-rect)
+                                    (g/update-property view-id :input-action-queue conj action)))))
+                            (catch Throwable error
+                              (reset! process-events? false)
+                              (error-reporting/report-exception! error)))))
+        simulate-mouse-on-modifier-keys! (fn [e]
```
AGulev: new function which re-send the last event if non-text key is pressed

matgis: I'm getting reflection warnings in the console. We need to type-hint the argument here to avoid reflection.
```suggestion
        simulate-mouse-on-modifier-keys! (fn [^KeyEvent e]
```

matgis (dismissed):
Seems to work! But please fix the reflection warnings before merging.

PR #10837, opened 2025-06-13
Editor: Added functions for modifying specific GuiNode layouts

matgis:
This change is a minor refactoring that makes it simpler to modify specific layouts associated with a `GuiNode`, rather than relying on the `visible-layout` setting of the owning `GuiSceneNode`.

We need this for property override transfer and editor scripts.

### DIFF
--- a/editor/src/clj/dynamo/graph.clj
+++ b/editor/src/clj/dynamo/graph.clj
@@ -1581,6 +1581,21 @@
      (property-overridden? basis node-id prop-kw)
      true)))
 
+(defn node-property-dynamic
+  "Returns the value of a dynamic associated with a specific node property."
+  [node property-label dynamic-label evaluation-context]
+  (let [node-type (node-type node)
+        property-def (get (in/all-properties node-type) property-label)
+        _ (assert property-def (format "Property %s not found in node-type %s." property-label (:k node-type)))
+        dynamic-fnk (-> property-def (get :dynamics) (get dynamic-label))
+        _ (assert dynamic-fnk (format "Dynamic %s not found on property %s in node-type %s." dynamic-label property-label (:k node-type)))
+        dynamic-fn (:fn dynamic-fnk)
+        dynamic-args (coll/transfer (:arguments dynamic-fnk) {}
+                       (map (fn [arg-label]
+                              (let [arg-value (in/node-value node arg-label evaluation-context)]
+                                (pair arg-label arg-value)))))]
+    (dynamic-fn dynamic-args)))
+
 (defmulti node-key
   "Used to identify a node uniquely within a scope. This has various uses,
   among them is that we will restore overridden properties during resource sync
--- a/editor/src/clj/editor/gui.clj
+++ b/editor/src/clj/editor/gui.clj
@@ -843,35 +843,23 @@
         (persistent!
           (reduce dissoc! node-properties stripped-prop-kws))))))
 
-(defn- update-layout-property [evaluation-context node-id prop-kw update-fn & args]
-  (let [old-value (g/node-value node-id prop-kw evaluation-context)
-        new-value (apply update-fn old-value args)
-        trivial-gui-scene-info (g/valid-node-value node-id :trivial-gui-scene-info evaluation-context)
-        current-layout (:current-layout trivial-gui-scene-info)]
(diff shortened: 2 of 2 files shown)


editor/src/clj/dynamo/graph.clj
```diff
@@ -1581,6 +1581,21 @@
...
+        _ (assert property-def (format "Property %s not found in node-type %s." property-label (:k node-type)))
+        dynamic-fnk (-> property-def (get :dynamics) (get dynamic-label))
+        _ (assert dynamic-fnk (format "Dynamic %s not found on property %s in node-type %s." dynamic-label property-label (:k node-type)))
+        dynamic-fn (:fn dynamic-fnk)
+        dynamic-args (coll/transfer (:arguments dynamic-fnk) {}
+                       (map (fn [arg-label]
+                              (let [arg-value (in/node-value node arg-label evaluation-context)]
+                                (pair arg-label arg-value)))))]
+    (dynamic-fn dynamic-args)))
+
```
matgis, line 1598: Used to get the `edit-type` of a property. We need the `changes-fn` from it.

editor/src/clj/editor/gui.clj
```diff
@@ -843,83 +843,150 @@
...
-  (let [old-value (g/node-value node-id prop-kw evaluation-context)
-        new-value (apply update-fn old-value args)
-        trivial-gui-scene-info (g/valid-node-value node-id :trivial-gui-scene-info evaluation-context)
-        current-layout (:current-layout trivial-gui-scene-info)]
-    (if (str/blank? current-layout)
-      (g/set-property node-id prop-kw new-value)
-      (g/update-property
-        node-id :layout->prop->override
-        update current-layout
-        assoc prop-kw new-value))))
```
matgis, line 856: This function was moved down in the file and renamed to `basic-layout-property-update-in-current-layout` to match the naming pattern of similar functions that were introduced in this PR.

editor/src/clj/editor/gui.clj
```diff
@@ -843,83 +843,150 @@
...
+      (fn [prop->override]
+        (coll/not-empty
           (reduce-kv (fn [prop->override prop-kw new-value]
                        (if (nil? new-value)
                          (dissoc prop->override prop-kw)
                          (assoc prop->override prop-kw new-value)))
                      prop->override
                      changes))))))
 
-(defn basic-layout-property-edit-type-clear-fn [node-id prop-kw]
```
matgis, line 882: Moved down and renamed to `basic-layout-property-clear-in-current-layout`.

editor/src/clj/editor/gui.clj
```diff
@@ -843,83 +843,150 @@
...
+(defn- prop->value-for-specific-layout
+  [node-id layout-name evaluation-context]
+  (let [layout->prop->value (g/node-value node-id :layout->prop->value evaluation-context)
+        prop->value (get layout->prop->value layout-name ::not-found)]
+    (when (= ::not-found prop->value)
+      (throw (ex-info (format "Layout '%s' does not exist in the scene." layout-name)
+                      {:node-id node-id
+                       :layout-name layout-name
+                       :layout-name-candidates (vec (sort (keys layout->prop->value)))})))
+    prop->value))
```
matgis, line 896: The map returned here should contain all values, not only overrides, so it could be used to query the value of a property in a specific layout.

editor/src/clj/editor/gui.clj
```diff
@@ -843,83 +843,150 @@
...
+            (coll/transfer prop-kws-and-values {}
+              (partition-all 2)
+              (mapcat (fn [[prop-kw new-value]]
+                        (if-let [changes-fn (:changes-fn (g/node-property-dynamic node prop-kw :edit-type evaluation-context))]
+                          (let [old-value (get @prop->value-delay prop-kw)]
+                            (changes-fn evaluation-context node-id prop-kw old-value new-value))
+                          {prop-kw new-value}))))]
+
+        (coll/not-empty
+          (layout-property-changes-tx-data layout-name node-id changes))))))
```
matgis, line 971: This function could be used to set property values in a specific layout, similar to how an edit from the Properties View would work.

vlaaad (approved):
Nice!

PR #10839, opened 2025-06-13
Fixed issue where `drag handle` interrupted the Tab key from moving to the next field

### LINKED ISSUE #10829: The property drag handle interrupt Tab key to go next field
If in Defold Editor the new drag handle is enabled (while mouse hover it), I can't with single tab key go to other field, it requires double tab instead

No focus on Tab for drag handle

https://github.com/user-attachments/assets/79f59976-7d31-483b-88ee-d5bc82f4f0ff

AGulev:
Fix https://github.com/defold/defold/issues/10829


sprocketc (approved):
Looks good, thanks for handling this!

PR #10845, opened 2025-06-16
Fix drag handle usability issues

sprocketc:
Reported [here]( https://forum.defold.com/t/defold-1-10-2-has-been-released/80687/8?u=sprocketc)

### Technical changes
- Made the handle smaller, used the scaleupdown icon, added a background
- Users can click trough the handle to edit the value
- The handle is now hidden when the input is focused
- Added  a min-width rule to text fields

### DIFF
--- a/editor/src/clj/editor/properties_view.clj
+++ b/editor/src/clj/editor/properties_view.clj
@@ -266,29 +266,37 @@
     (ui/user-data! ::values nil)))
 
 (defn handle-label-release-event!
-  [^MouseEvent event]
+  [^MouseEvent event ^Node control]
   (.consume event)
-  (doto (.getTarget event)
-    (ui/remove-style! "active")
-    (ui/user-data! ::op-seq nil)
-    (ui/user-data! ::position nil)))
+  (let [target (.getTarget event)]
+    (when-not (ui/user-data target ::values)
+      (.requestFocus control))
+    (doto target
+      (ui/remove-style! "active")
+      (ui/user-data! ::op-seq nil)
+      (ui/user-data! ::position nil))))
 
 (defn- make-control-draggable
   ([^Node control drag-event-handler]
    (make-control-draggable control drag-event-handler true))
   ([^Node control drag-event-handler is-left-aligned]
-   (let [drag-icon (doto (Button. "" (jfx/get-image-view "icons/32/Icons_X_10_scalesides.png" 16))
-                     (ui/add-style! "action-button")
+   (let [drag-icon (doto (Button. "" (jfx/get-image-view "icons/32/Icons_X_11_scaleupdown.png" 14))
+                     (ui/add-styles! ["action-button" "drag-handle"])
                      (.setFocusTraversable false)
                      (.addEventHandler MouseEvent/MOUSE_DRAGGED (ui/event-handler event (drag-event-handler event)))
                      (.addEventHandler MouseEvent/MOUSE_PRESSED (ui/event-handler event (handle-label-press-event! event)))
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/properties_view.clj
```diff
@@ -266,29 +266,37 @@
...
-  [^MouseEvent event]
+  [^MouseEvent event ^Node control]
   (.consume event)
-  (doto (.getTarget event)
-    (ui/remove-style! "active")
-    (ui/user-data! ::op-seq nil)
-    (ui/user-data! ::position nil)))
+  (let [target (.getTarget event)]
+    (when-not (ui/user-data target ::values)
+      (.requestFocus control))
```
sprocketc, line 273: If we release the mouse without dragging, we focus on the control to allow users to click through the handle. I was tempted to make the handle fit the input size, but I was afraid that if this doesn't work as expected, users won't be able to edit the value at all.

editor/styling/stylesheets/modules/_properties.scss
```diff
@@ -3,6 +3,10 @@
         -fx-spacing: 0;
     }
 
+    .text-field {
+        -fx-min-width: 32px;
+    }
+
```
sprocketc, line 9: I believe setting a min-width makes sense regardless of the handle. I think we can increase this value a little bit, but I wasn't sure if we want that.

editor/styling/stylesheets/modules/_properties.scss
```diff
@@ -20,12 +24,21 @@
...
             visibility: hidden;
 
+            &.drag-handle {
+                -fx-max-width: 16px;
+                -fx-min-width: 16px;
+            }
+
+            &.color-dropper {
+                -fx-background-color: transparent;
+            }
```
sprocketc, line 37: I removed this rule from the handle. Now that its width is a lot smaller, having a background helps us grab the handle, because its bounds are visible. It also makes the handle distinguishable if there is a value behind it.

editor/src/clj/editor/properties_view.clj
```diff
@@ -266,29 +266,37 @@
...
+       (AnchorPane/setRightAnchor drag-icon 0.0)
+       (AnchorPane/setLeftAnchor drag-icon 0.0))
      (doto control
        (AnchorPane/setRightAnchor 0.0)
        (AnchorPane/setLeftAnchor 0.0))
+     (ui/observe (.focusedProperty control)
+                 (fn [_ _ is-focused]
+                   (if is-focused
+                     (ui/add-style! drag-icon "hidden")
+                     (ui/remove-style! drag-icon "hidden"))))
```
sprocketc, line 299: When the input is focused, the handle will now be hidden. This gives more room to the input while editing, and prevents typing a value while dragging. It is also consistent with [this change](https://github.com/defold/defold/pull/10839).

matgis (changes requested):
Looks neat! I found a couple of issues:
* Pressing shift while dragging awkwardly cancels the drag. I would expect it to accelerate subsequent movements just as if I had pressed shift before starting the drag.
* The mouse pointer while hovering the handle still signals a horizontal drag. I think we should rotate it 90 degrees now that the drag handle shows a different axis.

matgis (approved):
Pressing shift while dragging awkwardly cancels the drag for certain node types. I could get it to happen 100% while dragging the position of `GuiTemplate` nodes in my project, but when do the same to nodes imported by the `GuiTemplate` nodes, it instead appears to perform some sort of refresh of the property so it shows up as overridden. I can still keep dragging, though.

Anyway, I'm approving this for now and we could investigate further later.

PR #10852, opened 2025-06-18
Bump Box2D to 3.1.0, add thread pool backend for Box2D 3.x task system + GitHub actions to build Box2D

### LINKED ISSUE #10617: Update to Box2D 3.1
**Task (REQUIRED):**
Box2D 3.1 has several performance improvements and bugfixes that we want:

https://box2d.org/posts/2025/04/box2d-3.1/

Defold should work with Box2D 3.1

deurell:
Add optional support for Box2D 3.1.0 and provide a GitHub action to build Box2D for all platforms.
Fixes #10617 

Technical changes
Added a workflow to build Box2D for all platforms.
`gh workflow run build-box2d.yml --ref <branch-name>`

- Added Box2D 3.1.0 files to external
- Added GitHub action to build Box2D for all platforms
- Added install_waf command to build.py
- Added fix to handle -ext suffix in Android SDK folders
- Fixed breaking changes in Box2D integration
- Added thread pool backend for Box2D 3.x in order to do multithread performance tests

Defold is still set to use Box2D 2.2 as default. Thread pool is set to run on main as default. On emscripten builds we need to add a Asyncify dependency to enable the thread pool as blocking waits aren't allowed on the main browser thread. We solve this with an active wait, yielding with emscripten_sleep().

A branch containing the thread pool implementation used for performance testing is available at [here](https://github.com/defold/defold/tree/box2d-threadpool-prototype). Instructions on how to run it under wasm_pthread-web is available [here](https://github.com/defold/defold/blob/box2d-threadpool-prototype/README_BOX2D_THREADPOOL.md).


.github/workflows/build-box2d.yml
```diff
@@ -0,0 +1,329 @@
...
+
+    - name: Build Box2D
+      run: |
+        ./scripts/build.py --save-env-path=ci/env.sh save_env
+        source ci/env.sh
+        echo "DYNAMO_HOME=$DYNAMO_HOME"
+        source "$EMSDK/emsdk_env.sh"
+        export CC=emcc
+        export CXX=em++
+        export CFLAGS="-DBOX2D_DISABLE_SIMD"
```
deurell: Should we enable SIMD on selected web platform builds?

.github/workflows/build-box2d.yml
```diff
@@ -0,0 +1,329 @@
...
+        ./scripts/build.py --save-env-path=ci/env.sh save_env
+        source ci/env.sh
+        echo "DYNAMO_HOME=$DYNAMO_HOME"
+        export PATH="$ANDROID_NDK_ROOT/toolchains/llvm/prebuilt/linux-x86_64/bin:$PATH"
+        ./scripts/build.py install_waf
+        cd external/box2d
+        if [[ "${{ matrix.platform }}" == "armv7-android" ]]; then
+          export CC="armv7a-linux-androideabi21-clang"
+          export CXX="armv7a-linux-androideabi21-clang++"
+          export CFLAGS="-march=armv7-a -mfloat-abi=softfp -DBOX2D_DISABLE_SIMD"
```
deurell: Disable simd on armv7

build_tools/sdk.py
```diff
@@ -136,6 +137,9 @@ def _get_latest_version_from_folders(path, replace_patterns=[]):
     def _replace_pattern(s, patterns):
         for pattern, replace in patterns:
             s = s.replace(pattern, replace)
+            # handle -ext to fix Android versions like android-34-ext12
+        if '-ext' in s:
+            s = re.sub(r'-ext\d+$', '', s)
```
JCash: Indentation looks strange here?
SHould the comment or the code move? (presumably the code?)

scripts/build.py
```diff
@@ -629,14 +628,23 @@ def _check_package_path(self):
...
+        self._extract_tgz(waf_path, self.ext)
 
     def install_ext(self):
         def make_package_path(root, platform, package):
             return join(root, 'packages', package) + '-%s.tar.gz' % platform
 
         def make_package_paths(root, platform, packages):
             return [make_package_path(root, platform, package) for package in packages]
+        
+        self.install_waf()
```
JCash, line 650: I would have expected some code to also be removed? 🤔

JCash:
Perhaps rephase "Bump Box2D to 3.1.0" to "Added optional support for".
This is to beeter communicate what the user till actually get when updating the editor version.

engine/engine/src/test/test_engine.cpp
```diff
@@ -424,9 +424,11 @@ TEST_P(DrawCountTest, DrawCount)
 
 DrawCountParams draw_count_params[] =
 {
-    // Box2d v2:
+#ifdef DM_PHYSICS_BOX2D_V3
+    {"/game.projectc", 3, 2},    // 1 draw call for sprite, 1 for debug physics
+#else
```
JCash, line 429: 👍

engine/gamesys/src/gamesys/test/collision_object/joint_test.script
```diff
@@ -96,11 +96,13 @@ function init(self)
     physics.create_joint(physics.JOINT_TYPE_HINGE, self.a_coll, self.joint_id, zp, self.b_coll, zp, { lower_angle = 0, upper_angle = 10, enable_limit = true })
     properties = physics.get_joint_properties(self.a_coll, self.joint_id)
 
+    if PHYSICS=="box2dv3" then
```
JCash, line 103: 👍

engine/gamesys/src/gamesys/test/test_gamesys.h
```diff
@@ -528,6 +528,17 @@ void GamesysTest<T>::SetUp()
...
     dmScript::Initialize(m_ScriptContext);
 
+    lua_State* L = dmScript::GetLuaState(m_ScriptContext);
+
+    #ifdef DM_PHYSICS_BOX2D_V3
+        lua_pushstring(L, "box2dv3");
+    #elif defined(DM_PHYSICS_BOX2D_V2)
+       lua_pushstring(L, "box2dv2");
+    #else
+        lua_pushstring(L, "box2dv2");
```
JCash, line 537: This catch-all, was it supposed to say "bullet3d" ?
Feels strange to have the same value as above imho, if it's already the default.

engine/physics/src/physics/box2d/box2d_debug_draw.cpp
```diff
@@ -203,22 +203,22 @@ namespace dmPhysics
...
+        m_DebugDraw.DrawPointFcn            = DrawPoint;
+        m_DebugDraw.DrawStringFcn           = DrawString;
 
         // Debug!
         const bool force_flag = false;
         m_DebugDraw.drawShapes           = force_flag;
         m_DebugDraw.drawJoints           = force_flag;
         m_DebugDraw.drawJointExtras      = force_flag;
-        m_DebugDraw.drawAABBs            = force_flag;
+        //m_DebugDraw.drawAABBs            = force_flag;
```
JCash: dead code

engine/physics/src/physics/box2d/box2d_physics.cpp
```diff
@@ -25,8 +25,169 @@
...
 #include "box2d_physics.h"
 
+#include <thread>
+#include <mutex>
+#include <condition_variable>
+#include <queue>
+#include <functional>
+#include <atomic>
+#include <memory>
+#include <unordered_map>
```
JCash: Let's continue using our own implementations, in order to stay on top of engine size and performance.
Also, prefer C headers over C++.
And, we more or less have a strict ban on STL, so keep that in mind.

* dlib/thread.h
* dlib/mutex.h
* dlib/conditionals.h
* dlib/atomic.h
* dlib/hashtable.h

engine/physics/src/physics/box2d/box2d_physics.cpp
```diff
@@ -25,8 +25,169 @@
...
+    s_QueueCV.notify_all();
+    for (auto& thread : s_PhysicsThreads)
+        if (thread.joinable())
+            thread.join();
+    s_PhysicsThreads.clear();
+}
+
+using b2TaskHandle = void*;
+
+static b2TaskHandle MyEnqueueTask(
```
JCash: "MyEnqueueTask" -> EnqueueTask?

engine/physics/src/physics/box2d/box2d_physics.cpp
```diff
@@ -25,8 +25,169 @@
...
+        s_TaskHandles[handle.get()] = handle;
+    }
+
+    THREAD_LOG("EnqueueTask: %d items, %d minRange, %d chunks", itemCount, minRange, chunks);
+
+    int offset = 0;
+    for (int i = 0; i < chunks; ++i)
+    {
+        int b = offset;
+        int e = std::min(itemCount, offset + minRange);
```
JCash: dmMath::Min()

engine/physics/src/physics/box2d/box2d_physics.cpp
```diff
@@ -25,8 +25,169 @@
...
+
+//#define DEBUG_THREAD_POOL
+#ifdef DEBUG_THREAD_POOL
+#include <dmsdk/dlib/log.h>
+#define THREAD_LOG(fmt, ...) dmLogInfo(fmt, ##__VA_ARGS__)
+#else
+#define THREAD_LOG(fmt, ...) (void)0
+#endif
+
+static std::vector<std::thread>             s_PhysicsThreads;
```
JCash: * dlib/array.h dmArray<>

engine/physics/src/physics/box2d/box2d_physics.cpp
```diff
@@ -25,8 +25,169 @@
...
+static std::mutex                           s_QueueMutex;
+static std::condition_variable              s_QueueCV;
+static std::queue<std::function<void(int)>> s_JobQueue;
```
JCash: We don't have a queue, but we do have "ringbuffer.h" which can fill that purpose.
Or, if the jobs are unordered, you can use `dmArray::EraseSwap()` that moves tha last item into the index you deleted.

engine/physics/src/physics/box2d/box2d_physics.cpp
```diff
@@ -25,8 +25,169 @@
...
+{
+    s_Running.store(false);
+    s_QueueCV.notify_all();
+    for (auto& thread : s_PhysicsThreads)
+        if (thread.joinable())
+            thread.join();
+    s_PhysicsThreads.clear();
+}
+
+using b2TaskHandle = void*;
```
JCash: In the engine we use typedefs: `typedef void* b2TaskHandle;`

engine/physics/src/physics/box2d/box2d_physics.cpp
```diff
@@ -25,8 +25,169 @@
...
+    b2TaskCallback* task,
+    int             itemCount,
+    int             minRange,
+    void*           taskContext,
+    void* /*userContext*/)
+{
+    EnsurePhysicsThreads();
+
+    int chunks = (itemCount + minRange - 1) / minRange;
+    auto handle = std::make_shared<TaskHandle>();
```
JCash: No `auto`. This is for readability/maintainability/searchability.

engine/physics/src/physics/box2d/box2d_physics.cpp
```diff
@@ -64,6 +225,16 @@ namespace dmPhysics
         worldDef.gravity              = context->m_Gravity;
         worldDef.restitutionThreshold = context->m_VelocityThreshold * context->m_Scale;
 
+        // optimizations from box2d samples
+        worldDef.contactHertz = 30.0f;
+        worldDef.contactDampingRatio = 10.0f;
+        worldDef.enableContinuous = true;
```
JCash: In the future, I can totally see more game.project settings for these.

engine/physics/src/physics/box2d/box2d_physics.cpp
```diff
@@ -25,8 +25,169 @@
...
+
+static void ShutdownPhysicsThreads()
+{
+    s_Running.store(false);
+    s_QueueCV.notify_all();
+    for (auto& thread : s_PhysicsThreads)
+        if (thread.joinable())
+            thread.join();
+    s_PhysicsThreads.clear();
+}
```
JCash: What is the life span of these threads?
I don't see a call to `ShutdownPhysicsThreads()`?

engine/physics/src/physics/box2d/box2d_physics.cpp
```diff
@@ -25,8 +25,169 @@
...
+    bool expected = false;
+    if (s_Running.compare_exchange_weak(expected, true))
+    {
+        THREAD_LOG("Starting %d worker threads", WORKER_COUNT);
+        for (int tid = 0; tid < WORKER_COUNT; ++tid)
+        {
+            s_PhysicsThreads.emplace_back([tid]() { PhysicsWorker(tid); });
+        }
+    }
+}
```
JCash: What is the strategy/separation here?
If I have two separate physics worlds running, will that work with this global pool of workers?

Also, I was under the impression that one couldn't have this code in a non pthread environment? 🤔

JCash (changes requested):
There are some questions in the PR, and we need to port the usage of C++ headers and STL to our code style.

engine/physics/src/physics/box2d/box2d_physics.cpp
deurell: So the reason for the thread pool prototype was to quickly verify the effect of threading impact and have perf data available quickly. That's why I wrote it in the quickest/most compact way i could think of in order to support the performance test. It wasn't supposed to be used in prod. If we find a thread pool worth it I can port it to our standard. As you pointed out in only uses a global pool for for all worlds as I only had this is the perf test session.

engine/physics/src/physics/box2d/box2d_physics.cpp
deurell: There are more tweaks but wanted to follow the ones used in official Box2d perf tests.

engine/physics/src/physics/box2d/box2d_physics.cpp
deurell: It will work fine if you set thread# to 0. That will run on all platforms and continue to run on main. Running on wasm/pthread it will perform a emscripten.sleep() poll loop and all other platforms will do a normal wait.

scripts/build.py
deurell: It was installed with PACKAGES_ALL before. Removed  it (waf-2.0.3) from there, made a function for it and called it from install_ext instead. This is just to expose install_waf so I can run it from the github actions when setting everything up.

build_tools/sdk.py
deurell: Comment indentation is off. Fixed it.

engine/gamesys/src/gamesys/test/test_gamesys.h
deurell: simplified it to remove redundant code

engine/physics/src/physics/box2d/box2d_debug_draw.cpp
deurell: removed

engine/physics/src/physics/box2d/box2d_physics.cpp
```diff
@@ -32,38 +31,42 @@ namespace dmPhysics
...
+        : m_Worlds()
+        , m_DebugCallbacks()
+        , m_Socket(0)
+        , m_Scale(1.0f)
+        , m_InvScale(1.0f)
+        , m_ContactImpulseLimit(0.0f)
+        , m_TriggerEnterLimit(0.0f)
+        , m_RayCastLimit(0)
+        , m_TriggerOverlapCapacity(0)
+        , m_AllowDynamicTransforms(0)
```
JCash: Please avoid large scale indentation+formatting in the same PR.
This makes it much harder to see what's actually changed (both during a PR, but also later when looking at a specific commit when e.g. debugging)

engine/physics/src/physics/box2d/box2d_physics.cpp
```diff
@@ -587,26 +590,26 @@ namespace dmPhysics
...
-                                b2Shape_GetUserData(contact.shapeIdA),
-                                b2Shape_GetFilter(contact.shapeIdA).categoryBits,
-                                b2Shape_GetUserData(contact.shapeIdB),
-                                b2Shape_GetFilter(contact.shapeIdB).categoryBits,
-                                step_context.m_CollisionUserData);
+                            b2Shape_GetUserData(contact.shapeIdA),
+                            b2Shape_GetFilter(contact.shapeIdA).categoryBits,
+                            b2Shape_GetUserData(contact.shapeIdB),
+                            b2Shape_GetFilter(contact.shapeIdB).categoryBits,
+                            step_context.m_CollisionUserData);
```
JCash: Is this our clang format that aligns all the parameters to the same line as the call site? (then I guess we need to fix that later as well)

JCash (changes requested):
Overall good, but please revert the white space changes in the box2d_physics.cpp as it distracts from the actual changes. (use a separate PR if desired)

engine/physics/src/physics/box2d/box2d_physics.cpp
deurell: I changed editor to clion and it autoformatted this for some reason. My bad.

engine/physics/src/physics/box2d/box2d_physics.cpp
deurell: I had some fighting between our clang format and the current file code style. I tried to improve my workflow/editor setup a bit but it landed in this. I'll stay true to the original file code style.

deurell: restored indentation to original file format

engine/physics/src/physics/box2d/box2d_physics.cpp
deurell: restored indentation to original file format

JCash (approved):

PR #10855, opened 2025-06-18
Avoid unnecessary properties pane remake

### LINKED ISSUE #10847: Property field value drag adjustment cancels on shift press
Pressing shift while dragging the property field values for certain objects seem to have unwanted effects. For Gui template nodes, pressing shift will cancel the drag. In I instead drag a property on one of the Gui nodes imported by the template, the property refreshes to appear overridden in the Properties View, but I can still drag the value.

sprocketc:
Fixes #10847

### DIFF
--- a/editor/src/clj/editor/properties_view.clj
+++ b/editor/src/clj/editor/properties_view.clj
@@ -1014,7 +1014,7 @@
       (when-let [update-ui-fn (get update-fns key)]
         (update-ui-fn property)))))
 
-(def ^:private ephemeral-edit-type-fields [:from-type :to-type :set-fn :clear-fn])
+(def ^:private ephemeral-edit-type-fields [:from-type :to-type :set-fn :clear-fn :dialog-accept-fn])
 
 (defn- edit-type->template [edit-type]
   (apply dissoc edit-type ephemeral-edit-type-fields))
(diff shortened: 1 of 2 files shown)


editor/styling/stylesheets/modules/_properties.scss
```diff
@@ -30,6 +30,10 @@
             &.drag-handle {
                 -fx-max-width: 16px;
                 -fx-min-width: 16px;
+
+                &:hover {
+                    -fx-cursor: v-resize;
+                }
```
sprocketc, line 36: Unrelated, but the cursor rule should be applied to the drag-handle only.

editor/src/clj/editor/properties_view.clj
```diff
@@ -1014,7 +1014,7 @@
       (when-let [update-ui-fn (get update-fns key)]
         (update-ui-fn property)))))
 
-(def ^:private ephemeral-edit-type-fields [:from-type :to-type :set-fn :clear-fn])
+(def ^:private ephemeral-edit-type-fields [:from-type :to-type :set-fn :clear-fn :dialog-accept-fn])
```
sprocketc, line 1017: `:dialog-accept-fn` affects the equality check `(not= template prev-template)` below, and `make-pane!` is called instead of `refresh-pane!`.

matgis (approved):
Nice find! LGTM! ❤️

Issue #5370, opened 2019-11-05
editor.get(id, "parent")


Jerakin:
Love that we can no gain properties from objects. This really enables some neat stuff.

I would like to get the absolute position of a node but to do that I need to iterate over its parent nodes and add the positions together. This would be possible if I where able to get the ID of the nodes parent.

Jerakin:
I have seen some movement on the editor-script font and wanted to take the opportunity to do a sneaky double ping @matgis @vlaaad to bring attention to this one!

PR #10858, opened 2025-06-19
Edit GUI nodes using editor scripts

vlaaad:
Now you can edit GUI nodes using editor scripts, e.g.:
```lua
editor.transact({
    editor.tx.add("/main.gui", "nodes", {
        type = "gui-node-type-box",
        position = {20, 20, 20}
    }),
    editor.tx.add("/main.gui", "nodes", {
        type = "gui-node-type-template",
        template = "/button.gui"
    }),
})
```

If the GUI file defines layouts, you can get and set the values from layouts using `layout:property` syntax, e.g.:
```lua
local node = editor.get("/main.gui", "nodes")[1]

-- GET:
local position = editor.get(node, "position")
pprint(position) -- {20, 20, 20}
local landscape_position = editor.get(node, "Landscape:position")
pprint(landscape_position) -- {20, 20, 20}

-- SET:
editor.transact({
    editor.tx.set(node, "Landscape:position", {30, 30, 30})
})
pprint(editor.get(node, "Landscape:position")) -- {30, 30, 30}
```

Layout properties that were set can be reset to their default values using `editor.tx.reset`:
```lua
print(editor.can_reset(node, "Landscape:position")) -- true
editor.transact({
    editor.tx.reset(node, "Landscape:position")
})
```
Template nodes can be read, but not edited: you can only set the properties of the template node tree:
```lua
local template = editor.get("/main.gui", "nodes")[2]
print(editor.can_add(template, "nodes")) -- false
local node_in_template = editor.get(template, "nodes")[1]
editor.transact({
    editor.tx.set(node_in_template, "text", "Button text")
})
print(editor.can_reset(node_in_template, "text")) -- true (overrides a value in the template)
```

Related to #8504

### DIFF
--- a/editor/src/clj/dynamo/graph.clj
+++ b/editor/src/clj/dynamo/graph.clj
@@ -1582,10 +1582,14 @@
      true)))
 
 (defn node-property-dynamic
-  "Returns the value of a dynamic associated with a specific node property, or
-  not-found value if the dynamic does not exist"
+  "Returns the value of a dynamic associated with a specific node property
+  If the dynamic does not exist, returns the provided not-found value, or throws
+  an assertion error if there was no not-found value provided"
   ([node property-label dynamic-label evaluation-context]
-   (node-property-dynamic node property-label dynamic-label nil evaluation-context))
+   (let [ret (node-property-dynamic node property-label dynamic-label ::not-found evaluation-context)]
+     (assert (not (identical? ret ::not-found))
+             (format "Dynamic %s not found on property %s in node-type %s." dynamic-label property-label (:k (node-type node))))
+     ret))
   ([node property-label dynamic-label not-found evaluation-context]
    (let [node-type (node-type node)
          property-def (get (in/all-properties node-type) property-label)
--- a/editor/src/clj/dynamo/graph.clj
+++ b/editor/src/clj/dynamo/graph.clj
@@ -98,7 +98,7 @@
      (gt/node-type n))))
 
 (defn node-type
-  "Return the node-type given a node. Uses the current basis if not provided."
+  "Return the node-type given a node."
   [node]
   (when node
     (gt/node-type node)))
@@ -1553,7 +1553,7 @@
   ([node-id]
    (override? (now) node-id))
   ([basis node-id]
-   (not (nil? (override-original basis node-id)))))
+   (some? (override-original basis node-id))))
 
 (defn override-id
   ([node-id]
@@ -1582,19 +1582,25 @@
      true)))
 
 (defn node-property-dynamic
-  "Returns the value of a dynamic associated with a specific node property."
-  [node property-label dynamic-label evaluation-context]
-  (let [node-type (node-type node)
-        property-def (get (in/all-properties node-type) property-label)
-        _ (assert property-def (format "Property %s not found in node-type %s." property-label (:k node-type)))
-        dynamic-fnk (-> property-def (get :dynamics) (get dynamic-label))
-        _ (assert dynamic-fnk (format "Dynamic %s not found on property %s in node-type %s." dynamic-label property-label (:k node-type)))
-        dynamic-fn (:fn dynamic-fnk)
-        dynamic-args (coll/transfer (:arguments dynamic-fnk) {}
-                       (map (fn [arg-label]
-                              (let [arg-value (in/node-value node arg-label evaluation-context)]
-                                (pair arg-label arg-value)))))]
-    (dynamic-fn dynamic-args)))
+  "Returns the value of a dynamic associated with a specific node property, or
+  not-found value if the dynamic does not exist"
+  ([node property-label dynamic-label evaluation-context]
+   (node-property-dynamic node property-label dynamic-label nil evaluation-context))
+  ([node property-label dynamic-label not-found evaluation-context]
+   (let [node-type (node-type node)
+         property-def (get (in/all-properties node-type) property-label)
+         _ (assert property-def (format "Property %s not found in node-type %s." property-label (:k node-type)))
+         dynamic-fnk (-> property-def (get :dynamics) (get dynamic-label))]
+     (if dynamic-fnk
+       (let [dynamic-fn (:fn dynamic-fnk)
+             dynamic-args (coll/transfer (:arguments dynamic-fnk) {}
+                            (map (fn [arg-label]
+                                     (let [arg-value (case arg-label
+                                                       :_this node
+                                                       (in/node-value node arg-label evaluation-context))]
+                                       (pair arg-label arg-value)))))]
+         (dynamic-fn dynamic-args))
+       not-found))))
 
 (defmulti node-key
   "Used to identify a node uniquely within a scope. This has various uses,
--- a/editor/src/clj/editor/attachment.clj
+++ b/editor/src/clj/editor/attachment.clj
@@ -263,6 +263,7 @@
         get-fn (getter basis workspace node-type list-kw)
         children-set (set (get-fn node-id evaluation-context))
         reorder-fn (-> basis (workspace/node-attachments workspace) (get node-type) list-kw :reorder)]
+    (assert (editable? basis workspace node-type list-kw))
     (assert (not (read-only? workspace node-id list-kw evaluation-context)))
     (assert (every? children-set reordered-child-node-ids))
     (assert (= (count children-set) (count reordered-child-node-ids))) ;; no duplicates
@@ -274,8 +275,8 @@
 
   The implementation will assert that the supplied child node ids are the same
   node ids as defined by [[getter]]. It will also assert that the container
-  node-id defines reorder of a list identified by list-kw (see [[reorderable?]])
-  and is not [[read-only?]]"
+  node-id defines an editable and reorderable list (see [[reorderable?]],
+  [[editable?]]) and is not [[read-only?]]"
   [workspace node-id list-kw reordered-child-node-ids]
   (g/expand-ec reorder-tx workspace node-id list-kw reordered-child-node-ids))
 
--- a/editor/src/clj/editor/attachment.clj
+++ b/editor/src/clj/editor/attachment.clj
@@ -14,11 +14,20 @@
 
 (ns editor.attachment
   "Extensible definitions for semantical node attachments"
-  (:refer-clojure :exclude [remove])
+  (:refer-clojure :exclude [remove alias])
   (:require [dynamo.graph :as g]
             [editor.workspace :as workspace]
             [internal.graph.types :as gt]
             [util.coll :as coll]))
+
+(defn- assoc-list-definition [state node-type list-kw {:keys [aliases] :as definition}]
+  (let [state (update state node-type assoc list-kw definition)]
+    (if aliases
+      (let [aliased-definition (-> definition
+                                   (dissoc :aliases)
+                                   (assoc :alias node-type))]
+        (reduce #(update %1 %2 assoc list-kw aliased-definition) state aliases))
+      state)))
 
 ;; SDK api
 (defn register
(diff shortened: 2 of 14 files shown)


vlaaad:
See also: https://github.com/defold/extension-spine/pull/209

editor/src/clj/dynamo/graph.clj
```diff
@@ -98,7 +98,7 @@
      (gt/node-type n))))
 
 (defn node-type
-  "Return the node-type given a node. Uses the current basis if not provided."
+  "Return the node-type given a node."
```
vlaaad, line 101: No basis: it works on nodes, not node ids.

editor/src/clj/dynamo/graph.clj
```diff
@@ -1582,19 +1582,25 @@
...
-        dynamic-fnk (-> property-def (get :dynamics) (get dynamic-label))
-        _ (assert dynamic-fnk (format "Dynamic %s not found on property %s in node-type %s." dynamic-label property-label (:k node-type)))
-        dynamic-fn (:fn dynamic-fnk)
-        dynamic-args (coll/transfer (:arguments dynamic-fnk) {}
-                       (map (fn [arg-label]
-                              (let [arg-value (in/node-value node arg-label evaluation-context)]
-                                (pair arg-label arg-value)))))]
-    (dynamic-fn dynamic-args)))
+  "Returns the value of a dynamic associated with a specific node property, or
+  not-found value if the dynamic does not exist"
```
vlaaad: I changed the semantics of this function. I think it makes sense since dynamics are generally optional. I needed it for looking up `read-only?` dynamic that very few properties define.

editor/src/clj/editor/attachment.clj
```diff
@@ -30,50 +39,99 @@
...
-                        {:parent-node-type parent-node-type
-                         :list-kw list-kw}))))
+      (let [definition (-> current-state (get parent-node-type) list-kw)]
+        (if-let [node-type->tx-attach-fn (:add definition)]
+          (let [tx-attach-fn (or (node-type->tx-attach-fn node-type)
+                                 (throw (ex-info (str (name (:k parent-node-type)) " does not support " (name list-kw) " attachments of type " (name (:k node-type)))
+                                                 {:parent-node-type parent-node-id :list-kw list-kw :node-type node-type})))
+                child-node-id (first (g/take-node-ids (g/node-id->graph-id parent-node-id) 1))]
+            (with-read-only-check
+              parent-node-type parent-node-id list-kw (:read-only? definition)
```
vlaaad, line 126: Main change: added `read-only?` check.

editor/src/clj/editor/editor_extensions/graph.clj
```diff
@@ -201,38 +206,172 @@
...
+  [node-type-kw f]
+  {:pre [(qualified-keyword? node-type-kw) (ifn? f)]}
+  (ext-property-getter node-type-kw f))
+
+(register-property-getter!
+  :editor.code.resource/CodeEditorResourceNode
+  (fn CodeEditorResourceNode-getter [node-id property evaluation-context]
+    (case property
+      "text" #(coll/join-to-string \newline (g/node-value node-id :lines evaluation-context))
+      nil)))
```
vlaaad, line 298: Previously, the equivalent code looked like this:
```clj
(defmethod ext-get [:editor.code.resource/CodeEditorResourceNode "text"] [node-id _ evaluation-context]
  (string/join \newline (g/node-value node-id :lines evaluation-context)))
```

The main reason I couldn't use a multi-method dispatching on 2 args (node type + property name) is that I needed to define GUI property getters/setters that have layout-sensitive properties. Such properties are not known beforehand...

editor/src/clj/editor/editor_extensions/graph.clj
```diff
@@ -250,109 +389,196 @@
...
-                node-type (g/node-type* basis node-id)
-                list-kw (property->prop-kw property)]
-            (when (attachment/defines? basis workspace node-type list-kw)
-              (let [f (attachment/getter basis workspace node-type list-kw)]
-                #(mapv rt/wrap-userdata (f node-id evaluation-context)))))))))
+            (let [workspace (project/workspace project evaluation-context)
+                  list-kw (property->prop-kw property)]
+              (when (attachment/defines? basis workspace node-type list-kw)
+                (let [f (attachment/getter basis workspace node-type list-kw)]
+                  #(mapv rt/wrap-userdata (f node-id evaluation-context))))))))))
```
vlaaad, line 374: Also, the inheritance chain approach simplified previously hard-coded type-based dispatch for outline and game project properties.

editor/src/clj/editor/gui.clj
```diff
@@ -462,16 +467,15 @@
...
-          (attach-gui-node node-tree target source type))))))
+(defn- gui-node-attach-fn [target source]
+  (g/with-auto-evaluation-context evaluation-context
+    (let [node-tree (node->node-tree (:basis evaluation-context) target)
+          taken-ids (g/node-value node-tree :id-counts evaluation-context)
+          next-index (gui-attachment/next-child-index target evaluation-context)]
+      (concat
+        (g/update-property source :id id/resolve taken-ids)
+        (g/set-property source :child-index next-index)
+        (attach-gui-node node-tree target source)))))
```
vlaaad, line 478: Simplified: `type` was unused, so there was no point in using a higher-order function.

editor/src/clj/editor/gui.clj
```diff
@@ -4215,27 +4326,145 @@
...
     :defaults visual-base-node-defaults}])
 
-(defonce ^:private custom-node-type-infos (atom (sorted-map)))
+(def ^:private empty-node-type-registry
+  {;; graph-node-type -> non-deprecated info
+   :node-cls->type-info {}
+   ;; node-type-kw -> custom-type -> info
+   :node-type->custom-type->type-info {}
+   ;; flat list
+   :type-infos []})
```
vlaaad: I refactored the custom node type registry because I needed access from the node type to initialization defaults. The previous implementation had linear complexity, not it's constant time.

editor/src/clj/editor/gui.clj
```diff
@@ -4215,27 +4326,145 @@
...
+          node (g/node-by-id basis node-id)
+          node-type (g/node-type node)]
+      (when-let [last-colon-index (str/last-index-of property \:)]
+        (let [property-name (subs property (inc last-colon-index))]
+          (when-let [prop-kw ((node-type->layout-property-names node-type) property-name)]
+            (let [layout-name (subs property 0 last-colon-index)
+                  layout->prop->override (g/node-value node-id :layout->prop->override evaluation-context)]
+              (when-let [prop-kw->override (layout->prop->override layout-name)]
+                (when (contains? prop-kw->override prop-kw)
+                  #(layout-property-clear-in-specific-layout evaluation-context layout-name node-id prop-kw))))))))))
```
vlaaad, line 4433: This is what makes the `layout:property` syntax to work.

editor/src/clj/editor/gui.clj
```diff
@@ -4215,27 +4326,145 @@
...
+
+;; TODO: template nodes????
```
vlaaad: Whoops, leftover todo. Template nodes are supported :)

editor/src/clj/editor/workspace.clj
```diff
@@ -896,14 +896,24 @@ ordinary paths."
...
-  ;; :add -> type -> list-kw -> type -> tx-attach-fn
-  ;; :get -> type -> list-kw -> get-fn
-  ;; :reorder -> type -> list-kw -> reorder-fn
+  ;; {node-type {list-kw {:add {node-type tx-attach-fn}
+  ;;                      :get get-fn
+  ;;                      :reorder reorder-fn
+  ;;                      :read-only? read-only-pred
+  ;;                      ;; one of:
+  ;;                      :alias node-type
+  ;;                      :aliases #{node-types}}}}
```
vlaaad, line 905: I also changed the node attachment state to something that's easier to follow.

editor/src/clj/editor/gui.clj
```diff
@@ -4215,27 +4326,145 @@
     :icon particlefx/particle-fx-icon
     :defaults visual-base-node-defaults}])
 
-(defonce ^:private custom-node-type-infos (atom (sorted-map)))
+(def ^:private empty-node-type-registry
+  {;; graph-node-type -> non-deprecated info
+   :node-cls->type-info {}
+   ;; node-type-kw -> custom-type -> info
```
matgis: This comment is a bit misleading. `node-type-kw` is a term used in many places to mean the fully-qualified keyword name of a `NodeType` defined using `defnode`. But this seems to be the keyword representation of a `Gui$NodeDesc$Type` like `:type-box` or `:type-pie`?

Using `node-type` as the term for a `Gui$NodeDesc$Type` keyword was an unfortunate choice prior to your changes, and I think we should fix it at some point.

editor/src/clj/editor/attachment.clj
```diff
@@ -177,7 +262,8 @@
         node-type (g/node-type* basis node-id)
         get-fn (getter basis workspace node-type list-kw)
         children-set (set (get-fn node-id evaluation-context))
-        reorder-fn (-> basis (workspace/node-attachments workspace) :reorder (get node-type) (get list-kw))]
+        reorder-fn (-> basis (workspace/node-attachments workspace) (get node-type) list-kw :reorder)]
+    (assert (not (read-only? workspace node-id list-kw evaluation-context)))
```
matgis, line 267: Should we `assert` `editable?` here too, like we do in other places?

editor/src/clj/editor/editor_extensions/graph.clj
```diff
@@ -201,38 +206,172 @@
...
+            ([a] (coll/some #(% a) fs))
+            ([a b] (coll/some #(% a b) fs))
+            ([a b c] (coll/some #(% a b c) fs))
+            ([a b c d] (coll/some #(% a b c d) fs))
+            ([a b c d e] (coll/some #(% a b c d e) fs))))))))
+
+(defonce/type InheritanceChain
+  [^:unsynchronized-mutable node-type-kw->f
+   ^:unsynchronized-mutable hierarchy
+   ^:unsynchronized-mutable builder
```
matgis: These should all be `:volatile-mutable`, right? This whole thing makes me nervous. Couldn't we implement it using an `atom` instead of these low-level constructs? These are soo hard to get right, and the fact that we don't even have soak tests for this scares me.

editor/src/clj/editor/editor_extensions/docs.clj
```diff
@@ -290,7 +295,7 @@ editor.command({
           :type :function
           :parameters [node-param property-param]
           :returnvalues [transaction-step-param]
-          :description "Create a transaction step that will deletes all items from node's list property when transacted with `editor.transact()`."}
+          :description "Create a transaction step that will delete all items from node's list property when transacted with `editor.transact()`."}
```
matgis: Use the term "remove" instead of "delete" to correspond to `editor.tx.remove` here?

editor/src/clj/dynamo/graph.clj
matgis: Good call. I think the `not-found` argument shouldn't be optional, though. You should always have to supply it, otherwise it could lead to subtle bugs.

matgis (changes requested):

editor/src/clj/editor/editor_extensions/graph.clj
vlaaad: As I understand, the fields don't have to be volatile, since the lock is the entity that establishes how data is synced between threads. I do agree though, it's spooky. I'll use a simple atom instead.

editor/src/clj/editor/gui.clj
matgis: Don't forget to remove the leftover TODO! 🙂

editor/src/clj/editor/editor_extensions/graph.clj
matgis: As I understand it, `:volatile-mutable` will mark the field as `volatile` in the resulting JVM class. The state of fields marked `volatile` will always be fetched from memory, whereas the default is to keep their values in CPU cache or registers for maximum efficiency. Thus, one thread may not see changes to the field from another thread because it already has the value cached and hasn't been made aware that it needs to re-fetch the value from RAM.

vlaaad: Yes, volatile is necessary for multi-threaded access without locks. But there are locks.  See [this SO discussion](https://stackoverflow.com/questions/31467826/do-i-need-to-use-volatile-if-update-to-other-threadcache-is-not-time-constraint), for example: the field access is guarded by a lock that enforces synchronization between threads; therefore, the field itself does not have to be volatile.

editor/src/clj/dynamo/graph.clj
```diff
@@ -1582,19 +1582,29 @@
...
-                                (pair arg-label arg-value)))))]
-    (dynamic-fn dynamic-args)))
+  "Returns the value of a dynamic associated with a specific node property
+  If the dynamic does not exist, returns the provided not-found value, or throws
+  an assertion error if there was no not-found value provided"
+  ([node property-label dynamic-label evaluation-context]
+   (let [ret (node-property-dynamic node property-label dynamic-label ::not-found evaluation-context)]
+     (assert (not (identical? ret ::not-found))
+             (format "Dynamic %s not found on property %s in node-type %s." dynamic-label property-label (:k (node-type node))))
+     ret))
```
vlaaad, line 1592: When I started providing defaults in the places that didn't do that explicitly, I realized they actually expected the dynamic to be defined. So instead made this function throw if `not-found` was not explicitly provided when there is no dynamic.

editor/src/clj/editor/editor_extensions/graph.clj
```diff
@@ -301,7 +257,22 @@
...
+              {::keys [builder hierarchy] :as current-state} @state-atom
+              builder (if (and builder (identical? hierarchy h))
+                        builder
+                        (let [builder (make-property-fn-builder h (::node-type-kw->f current-state))]
+                          (swap! state-atom assoc ::hierarchy h ::builder builder)
+                          builder))]
+          (builder node-type-kw)))
+       ([node-type-kw f]
+        (swap! state-atom #(-> % (update ::node-type-kw->f assoc node-type-kw f) (assoc ::builder nil)))
+        nil)))))
```
vlaaad, line 275: So much smaller without locks...

matgis (approved):
Ship it! ❤️

I guess you forgot to remove the leftover TODO, though? 😄

editor/src/clj/editor/gui.clj
```diff
@@ -4432,8 +4432,6 @@
                 (when (contains? prop-kw->override prop-kw)
                   #(layout-property-clear-in-specific-layout evaluation-context layout-name node-id prop-kw))))))))))
 
-;; TODO: template nodes????
-
```
vlaaad: 😏

matgis: Oh, never mind! 😄

Jerakin:
This doesn't add any attributes we can get right?

Just double checking if it's possible to get the parent with this ( like requested in #5370 )

vlaaad:
No, we haven't done that yet.

PR #10871, opened 2025-06-21
Fix example for resource.create_atlas()

### LINKED ISSUE #10870: Incorrect example for resource.create_atlas()
The example for `resource.create_atlas()` has a mistake regarding the pivot. It says:
```
geometries = {
  {
    id = 'idle0',
    width = 128,
    height = 128,
    pivot_x = 64,
    pivot_y = 64,
```
So the texture of size 128x128 is supposed to have a pivot in the center, at (64, 64). But pivot_x is not defined in pixels, but in unit coords. So the center is not (64, 64), but (0.5, 0.5), like the reference says:

> pivot_x: The pivot x value of the image in unit coords. (0,0) is upper left corner, (1,1) is bottom right. Default is 0.5.

While `pivot_x = 64` would be accepted, that would set the pivot at `64*128`, so render the image at (8192, 8192), and is certainly not what the example is supposed to do.

So the better example would be:
```
geometries = {
  {
    id = 'idle0',
    width = 128,
    height = 128,
    pivot_x = 0.5,
    pivot_y = 0.5,
```

HalfstarDev:
Fixes #10870

### DIFF
--- a/engine/gamesys/src/gamesys/scripts/script_resource.cpp
+++ b/engine/gamesys/src/gamesys/scripts/script_resource.cpp
@@ -2461,8 +2461,8 @@ static void MakeTextureSetFromLua(lua_State* L, dmhash_t texture_path_hash, dmGr
  *                 id = 'idle0',
  *                 width = 128,
  *                 height = 128,
- *                 pivot_x = 64,
- *                 pivot_y = 64,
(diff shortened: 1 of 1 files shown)


britzl (approved):
Thanks

PR #10874, opened 2025-06-23
Enhance curve colors

### LINKED ISSUE #10857: Different colors for individual curves in Curve Editor
While fine-tuning particle effects I often stack several curves in the Curve Editor (e.g. Life Scale, Life Rotation, Life Alpha). Because every curve is currently drawn in the same yellow, it’s hard to see at a glance which keyframes belong to which property—especially once there are a lot of control points.

Image

sprocketc:
Fixes #10857

### DIFF
--- a/editor/src/clj/editor/colors.clj
+++ b/editor/src/clj/editor/colors.clj
@@ -103,3 +103,6 @@
 
 (defn hsl->rgba [h s l]
   (alpha (hsl->rgb h s l) 1.0))
+
+(defn hsla->rgba [h s l a]
+  (alpha (hsl->rgb h s l) a))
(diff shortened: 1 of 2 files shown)


sprocketc:
@matgis Sorry about the copilot misclick..

matgis (commented):
This doesn't seem like the correct fix to the issue to me. I agree yellow might not be the best choice for the alpha channel (maybe gray?), but the issue seems to be that the "hardcoded" colors we use for the color channels do not "claim" that color so the evenly-distributed color selections will avoid it?

sprocketc:
Added a saturation value  to support gray and used the alpha channel to differentiate the colors on the UI. We now also try to avoid overlapping hues. This is far from sophisticated, but I feel like this is already overengineered.

matgis (approved):
Works great! Ship it!

PR #10878, opened 2025-06-23
Pull up or push down property overrides in the editor

### LINKED ISSUE #6685: Allow moving overridden properties back to original
It's a common pattern when authoring guis to create some template and then configure it when it's referenced in another gui. After completing this configuration, we want to move the changes back to the original because this configuration will be repeated in other uses of the template.

We should provide a way to lift the overridden property back to the original.

We could make this "lifting" work on a per property basis or per whole GUI basis.

matgis:
* The Clear Override button next to an overridden property in the Properties View has been moved into a dropdown menu available by clicking on the property label.
* You can now pull up property overrides to ancestors, or push down overrides to descendants in the editor. These commands are available from the property label dropdown menu, or from the context menu when right-clicking on an overridden object in the Outline View.
* Pull Up and Push Down Overrides buttons have been added to the Search Results view that appears when Show Overrides... is selected for an object or property.
* Fixed various small positioning and size issues around controls used in the Properties View.

Fixes #6685

### Technical changes
* `g/set-property` has been updated to only set individual properties to allow for an optional `opts` map argument. To set a bunch or properties at once, use `g/set-properties` instead.
* Added `menu_items.clj` and moved a few common menu item definitions into it.


editor/src/clj/dynamo/graph.clj
```diff
@@ -666,32 +666,52 @@
...
+    (partition-all 2)
+    (mapcat (fn [[property-label new-value]]
+              (it/set-property node-id property-label new-value nil)))))
+
+(defn set-properties!
+  "Creates transaction steps to assign multiple values to a node's properties,
+  then executes them in a transaction. Returns the result of the transaction,
+  (tx-result)"
+  [node-id & kvs]
+  (transact (apply set-properties node-id kvs)))
```
matgis, line 686: `g/set-property` has been updated to only set individual properties to allow for an optional `opts` map argument. To set a bunch or properties at once, use `g/set-properties` instead.

editor/src/clj/dynamo/graph.clj
```diff
@@ -1624,13 +1642,14 @@
...
+  (let [node (node-by-id basis node-id)
+        node-type (gt/node-type node)]
     (into {}
           (map (fn [[prop-kw raw-prop-value]]
                  [prop-kw (if (has-property? node-type prop-kw)
-                            (node-value node-id prop-kw evaluation-context)
+                            (in/node-value node prop-kw evaluation-context)
                             raw-prop-value)]))
-          (node-value node-id :_overridden-properties evaluation-context))))
+          (in/node-value node :_overridden-properties evaluation-context))))
```
matgis, line 1652: Avoid multiple node lookups.

editor/src/clj/dynamo/graph.clj
```diff
@@ -1660,7 +1679,11 @@
                                    node-key
                                    node-type-kw)
                            {:node-key node-key
-                            :node-type node-type-kw})))
+                            :node-type node-type-kw
+                            :source-node-id source-node-id
+                            :override-node-id override-node-id
+                            :override-node-key override-node-key
+                            :properties-by-override-node-key (persistent! properties-by-override-node-key)})))
```
matgis, line 1686: Additional `ex-data` for exception thrown when two nodes produce the same node key for storing overrides. This is used when a file is reloaded from external changes and we want to re-apply the overrides to the replacement node hierarchy.

editor/src/clj/dynamo/graph.clj
```diff
@@ -666,32 +666,52 @@
...
-   (partition-all 2 kvs)))
+  `(transact (set-property node-id :opacity 0.5 {:force true}))`"
+  ([node-id property-label value]
+   (it/set-property node-id property-label value nil))
+  ([node-id property-label value opts]
+   (it/set-property node-id property-label value opts))
+  ([node-id property-label value another-property-label another-value & more]
+   ;; Deprecated. To set multiple properties at once, use set-properties.
+   ;; Kept for compatibility with editor plugins until we can update them.
+   (apply set-properties node-id property-label value another-property-label another-value more)))
```
matgis, line 702: Unfortunately we have some calls to set-property with multiple kv-pairs in extensions. I didn't want this to PR to be a disruptive change for our users. We can remove this arity once we've updated the plugins.

editor/src/clj/editor/app_view.clj
```diff
@@ -362,11 +363,15 @@
...
+                           (g/node-value app-view :active-resource))]
+     (when (resource/openable-resource? resource)
+       resource)))
   ([app-view selection evaluation-context]
-   (or (selection->single-openable-resource selection)
-       (g/node-value app-view :active-resource evaluation-context))))
+   (when-let [resource (or (selection->single-resource selection)
+                           (g/node-value app-view :active-resource evaluation-context))]
+     (when (resource/openable-resource? resource)
+       resource))))
```
matgis, line 374: Corrected the logic here - previously it could return a non-openable resource in some cases. This is part of making the Open As... command available from the editor tab context menu.

editor/src/clj/editor/app_view.clj
```diff
@@ -1626,28 +1631,31 @@ If you do not specifically require different script states, consider changing th
...
                 :command :file.show-references}
                {:label "Dependencies..."
                 :command :file.show-dependencies}
-               {:label "Show Overrides"
-                :command :edit.show-overrides}
+               menu-items/separator
+               menu-items/show-overrides
+               menu-items/pull-up-overrides
+               menu-items/push-down-overrides
+               menu-items/separator
```
matgis, line 1655: I've created a shared module for the menu item definitions that appear in several menus to ensure they use the same label, icon, and so on.

editor/src/clj/editor/app_view.clj
```diff
@@ -2133,14 +2143,21 @@ If you do not specifically require different script states, consider changing th
...
            (when-not user-data
-             (let [resource (selection->single-openable-resource selection)
+             (let [[resource active-view-type-id]
+                   (g/with-auto-evaluation-context evaluation-context
+                     (if-let [selected-resource (selection->single-resource selection)]
+                       (pair selected-resource nil)
+                       (let [active-resource (g/node-value app-view :active-resource evaluation-context)
+                             active-view-type-id (:id (:view-type (g/node-value app-view :active-view-info evaluation-context)))]
+                         (pair active-resource active-view-type-id))))
+
```
matgis, line 2160: When invoked from an editor tab, we want to exclude the view-type that the tab is already showing from the options.

editor/src/clj/editor/collection.clj
```diff
@@ -262,12 +262,6 @@
 (defn- component-source-resource [basis component-node-id]
   (g/node-value component-node-id :source-resource (g/make-evaluation-context {:basis basis :cache c/null-cache})))
 
-(defn- overridable-resource-type? [resource-type]
-  (some-> resource-type :tags (contains? :overridable-properties)))
-
-(defn- overridable-resource? [resource]
-  (some-> resource resource/resource-type overridable-resource-type?))
-
```
matgis, line 270: Moved to `editor.resource`.

editor/src/clj/editor/fxui.clj
```diff
@@ -479,8 +483,37 @@
...
+        :show-duration [30 :s])))
+
+(defn- resolve-tooltip [props]
+  (let [tooltip-text (:tooltip props)]
+    (cond-> props
+            (string? tooltip-text)
+            (assoc :tooltip
+                   {:fx/type tooltip
+                    :text tooltip-text}))))
+
```
matgis, line 514: Moved up from below and broke out `tooltip` into separate component.

editor/src/clj/editor/fxui.clj
```diff
@@ -495,7 +528,27 @@
                                     :primary "button-primary"
                                     :secondary "button-secondary"
                                     :danger "button-danger"
-                                    :icon "button-icon"))))
+                                    :icon "button-icon"))
+      resolve-alignment
+      resolve-tooltip))
```
matgis, line 533: Buttons now support alignment and tooltip.

editor/src/clj/editor/fxui.clj
```diff
@@ -495,7 +528,27 @@
...
+    :or {variant :secondary}
+    :as props}]
+  (-> props
+      (assoc :fx/type fx.menu-button/lifecycle)
+      (dissoc :variant)
+      (add-style-classes "menu-button" (case variant
+                                         :secondary "button-secondary"
+                                         :icon "button-icon"))
+      resolve-alignment
+      resolve-tooltip))
```
matgis, line 551: Added `menu-button` component. We use this in the `search-results-view` when showing overrides.

editor/src/clj/editor/fxui.clj
```diff
@@ -593,85 +646,128 @@
                             :desc (dissoc popup :showing :anchor-x :anchor-y :on-window)}})
           :desc {:fx/type fx/ext-get-ref :ref ::on}}})
 
+(defn- icon-svg-path-data
```
matgis, line 649: Broke out the SVG data from the `icon` function below so we can also calculate their bounds in the new `icon-svg-path-bounds` function. Added `:icon/pull-up-override`, `:icon/push-down-override`, and `:icon/refresh`.

editor/src/clj/editor/fxui.clj
```diff
@@ -593,85 +646,128 @@
...
+        (assoc :fx/type fx.region/lifecycle
+               :min-width width
+               :max-width width
+               :pref-width width
+               :min-height height
+               :max-height height
+               :pref-height height
+               :shape {:fx/type icon
+                       :type type}))))
+
```
matgis, line 770: This allows us to use the SVG icons as a graphic for buttons, menu items and so on.

editor/src/clj/editor/fxui.clj
```diff
@@ -714,17 +810,6 @@
...
-  (let [alignment (get props :alignment ::not-found)]
-    (if (identical? alignment ::not-found)
-      props
-      (case alignment
-        (:top-left :top-center :top-right :center-left :center :center-right :bottom-left :bottom-center :bottom-right) props
-        :top (assoc props :alignment :top-center)
-        :left (assoc props :alignment :center-left)
-        :right (assoc props :alignment :center-right)
-        :bottom (assoc props :alignment :bottom-center)))))
-
```
matgis, line 727: Moved up in the file.

editor/src/clj/editor/fxui.clj
```diff
@@ -775,17 +860,6 @@
...
-  (if (string? tooltip)
-    (assoc props :tooltip {:fx/type fx.tooltip/lifecycle
-                           :wrap-text true
-                           :max-width 600
-                           :text tooltip
-                           :hide-delay [200 :ms]
-                           :show-delay [200 :ms]
-                           :show-duration [30 :s]})
-    props))
-
```
matgis, line 788: Moved up in the file.

editor/src/clj/editor/graph_util.clj
```diff
@@ -50,6 +51,32 @@
...
+                 ;;   "#book_script" -> "book_script"
+                 ;;   "/referenced_book#book_script" -> "referenced_book/book_script"
+                 (-> value
+                     (subs 1)
+                     (string/replace "#" "/")))))
+           (when (in/behavior node-type :id)
+             (let [value (in/node-value node :id evaluation-context)]
+               (when (string? value)
+                 (coll/not-empty value)))))))))
+
```
matgis, line 79: Similar to `node-debug-label` below, but presents a more user-friendly label suitable for use in UIs. We use it to identify the target node when transferring overrides.

editor/src/clj/editor/gui.clj
```diff
@@ -384,7 +385,7 @@
                (transient pb-field-index->pb-field)
                pb-field-index->pb-field)))
 
-(def ^:private prop-key->pb-field-index
+(def prop-key->pb-field-index
```
matgis, line 388: Made these public since they are used from tests.

editor/src/clj/editor/handler.clj
```diff
@@ -315,34 +315,39 @@
...
+              (catch Exception e
+                (when (not= :run fsym)
+                  (swap! throwing-handlers assoc throwing-id fnk))
+                (error-reporting/report-exception!
+                  (ex-info (format "handler '%s' in context '%s' failed at '%s' with message '%s'"
+                                   (:command handler) (:name command-context) fsym (.getMessage e))
+                           {:handler handler
+                            :command-context (update command-context :env dissoc :evaluation-context)}
+                           e))
+                nil))))))))
```
matgis, line 350: Smoother development process: We now re-enable handlers that had previously been disabled due to throwing exceptions when they are redefined from the REPL.

editor/src/clj/editor/handler.clj
```diff
@@ -402,10 +407,17 @@
...
-    (update :env merge (into {} (map (fn [[k [node v]]] [k (g/node-value (get-in context [:env node]) v)]) (:dynamics context))))))
+          (contains? context :dynamics)
+          (update :env
+                  (fn [env]
+                    (into env
+                          (map (fn [[dynamic-key [node-key label]]]
+                                 (let [node-id (env node-key)
+                                       value (g/node-value node-id label evaluation-context)]
+                                   (pair dynamic-key value))))
+                          (:dynamics context))))))
```
matgis, line 420: Use `evaluation-context` and fix readability.

editor/src/clj/editor/handler.clj
```diff
@@ -433,14 +445,19 @@
...
 
-(defn- eval-selection-context [context]
+(defn- eval-selection-context [{:keys [name selection-provider] :as context}]
   (let [selections (context-selections context)]
     (mapv (fn [selection]
-            (update context :env assoc :selection selection :selection-context (:name context) :selection-provider (:selection-provider context)))
+            (update context :env assoc
+                    :selection selection
+                    :selection-context name
+                    :selection-provider selection-provider))
```
matgis, line 454: Readability and look up `(:name context)` only once.

editor/src/clj/editor/handler.clj
```diff
@@ -433,14 +445,19 @@
...
           selections)))
 
 (defn eval-contexts [contexts all-selections?]
-  (let [contexts (mapv eval-dynamics contexts)]
+  (let [contexts (g/with-auto-evaluation-context evaluation-context
+                   (mapv #(eval-dynamics % evaluation-context)
+                         contexts))]
```
matgis, line 460: Use shared `evaluation-context` when evaluating dynamics.

editor/src/clj/editor/properties.clj
```diff
@@ -968,3 +980,385 @@
...
+      (coll/transfer source-prop-infos-by-prop-kw {}
+        (filter (fn [[_prop-kw prop-info]]
+                  ;; Note that properties that have validation errors may be
+                  ;; included in the transfer.
+                  (and (contains? prop-info :original-value)
+                       (:visible prop-info true)
+                       (not (:read-only? prop-info)))))))))
+
+(def ^:private property-transfer-target-status?
+  #{:ok :owner-resource-not-editable :property-not-found})
```
matgis, line 1007: We might want to present these different failure states in an error message in the future, but right now we will simply disable the menu items.

editor/src/clj/editor/properties.clj
```diff
@@ -968,3 +980,385 @@
...
+            (if target-aspect-qualifier
+              (str target-aspect-qualifier " Across " (text-util/amount-text target-proj-path-count "Resource"))
+              (str "Across " (text-util/amount-text target-proj-path-count "Resource")))))]
+
+    (-> (format "%s %s to %s %s"
+                action-text
+                overrides-qualifier
+                target-nodes-qualifier
+                target-resources-qualifier)
+        (string/replace "_" "__"))))
```
matgis, line 1272: This will be fun to localize one day, I'm sure. 😬

editor/src/clj/editor/properties_view.clj
```diff
@@ -137,10 +139,10 @@
     :script-property-type-url "script-property-text-field-icon-url"
     nil))
 
-(defmulti create-property-control! (fn [edit-type _ _]
-                                     (edit-type->type edit-type)))
+(defmulti make-property-control (fn [edit-type _context _property-fn]
+                                  (edit-type->type edit-type)))
```
matgis, line 143: For some reason I decided to rename this multimethod. I think I was very annoyed at the `!` prefix (maybe it was mutating before, but it's not anymore), so with the damage done I also went on and changed `create` to `make`, which we tend to use everywhere else for this sort of thing. Forgive me.

editor/src/clj/editor/properties_view.clj
```diff
@@ -329,13 +332,13 @@
                                 (properties/set-values! property (mapv #(assoc % index num) current-vals))
                                 (cancel-fn nil))))
                 _ (customize! text-field update-fn cancel-fn)
-                text-field (make-control-draggable text-field (partial handle-control-drag-event! property-fn drag-update-fn update-ui-fn) false) 
+                control (make-control-draggable text-field (partial handle-control-drag-event! property-fn drag-update-fn update-ui-fn) false)
```
matgis, line 335: This wasn't returning a `text-field`.

editor/src/clj/editor/properties_view.clj
```diff
@@ -666,14 +669,16 @@
...
   (let [box           (GridPane.)
         browse-button (doto (Button. "\u2026") ; "..." (HORIZONTAL ELLIPSIS)
                         (.setPrefWidth 26)
+                        (.setPrefHeight 27)
                         (.setFocusTraversable false)
                         (ui/add-style! "button-small"))
         open-button   (doto (Button. "" (jfx/get-image-view "icons/32/Icons_S_14_linkarrow.png" 16))
-                        (.setMaxWidth 26)
+                        (.setPrefWidth 26)
+                        (.setPrefHeight 27)
```
matgis, line 681: These where off by a few pixels.

editor/src/clj/editor/properties_view.clj
```diff
@@ -696,16 +701,31 @@
...
+    (ui/on-action! browse-button (fn [_]
+                                   (let [property (property-fn)
+                                         resource (g/with-auto-evaluation-context evaluation-context
+                                                    (let [basis (:basis evaluation-context)
+                                                          node-id (first (:node-ids property))
+                                                          project (project/get-project basis node-id)
+                                                          workspace (project/workspace project evaluation-context)]
+                                                      (first (resource-dialog/make workspace project dialog-opts))))]
+                                     (when (some? resource)
+                                       (properties/set-values! property (repeat resource))))))
```
matgis: We no longer rely on the `context` containing the `project` and the `workspace`.

editor/src/clj/editor/properties_view.clj
```diff
@@ -858,161 +878,237 @@
...
-                   (.setText (cond->> (format "Available as `%s` in editor scripts"
-                                              (string/replace (name key) \- \_))
-                                      tooltip
-                                      (str tooltip "\n\n")))
-                   (.setHideDelay Duration/ZERO)
-                   (.setShowDuration (Duration/seconds 30))))
-    (ui/add-style! "property-label")
-    (.setMinWidth Label/USE_PREF_SIZE)
-    (.setMinHeight 28.0)))
-
```
matgis, line 873: This was inlined into `make-property-grid-row` below. We need more surrounding context to create the dropdown menu for the property label.

editor/src/clj/editor/properties_view.clj
```diff
@@ -858,161 +878,237 @@
...
+    (properties/clear-override! property)
+    (ui/clear-auto-commit! property-control)))
+
+(handler/register-menu! ::property-menu
+  [menu-items/show-overrides
+   menu-items/pull-up-overrides
+   menu-items/push-down-overrides
+   {:label "Clear Override"
+    :icon "icons/32/Icons_S_02_Reset.png"
+    :command :private/clear-override}])
```
matgis, line 949: The Clear button that used to appear next to an overridden property is now a menu item available from the property label dropdown.

editor/src/clj/editor/protobuf.clj
```diff
@@ -45,14 +46,19 @@ Macros currently mean no foreseeable performance gain, however."
        (system/defold-dev?)
        (not (Boolean/getBoolean "defold.exception.decorate.disable"))))
 
-(defprotocol GenericDescriptor
+(def ^:private enforce-strict-pb-map-keys
+  (and *assert*
+       (system/defold-dev?)
+       (Boolean/getBoolean "defold.protobuf.strict.enable")))
```
matgis, line 52: Disabled by default, but can be useful during development. When enabled, any unrecognized keys in pb-maps will throw exceptions.

editor/src/clj/editor/protobuf.clj
```diff
@@ -752,14 +758,32 @@ Macros currently mean no foreseeable performance gain, however."
...
+                                  :clj-value clj-value}))))
+                         (.build builder)))
+                     (fn builder-fn [pb-map]
+                       (let [builder (new-builder class)]
+                         (doseq [[field-key clj-value] pb-map
+                                 :when (some? clj-value)
+                                 :let [setter! (get setters field-key)]
+                                 :when setter!]
+                           (setter! builder clj-value))
+                         (.build builder))))]
```
matgis, line 786: Updated for `enforce-strict-pb-map-keys`.

editor/src/clj/editor/search_results_view.clj
```diff
@@ -382,15 +394,21 @@
...
-  {:graphic {:fx/type fx.h-box/lifecycle
-             :spacing 6
-             :children (cond-> [{:fx/type fxui/legacy-label
-                                 :text (resource/resource->proj-path resource)}]
-                               qualifier
-                               (conj {:fx/type fxui/legacy-label
-                                      :style {:-fx-text-fill :-df-text-dark}
-                                      :text qualifier}))}})
+  (let [proj-path (resource/proj-path resource)]
+    {:graphic (resource-cell-view proj-path qualifier)}))
```
matgis, line 411: We also use the `resource-cell-view` for the added title that displays the original query (referred to as the "query label" below).

editor/src/clj/editor/search_results_view.clj
```diff
@@ -477,133 +495,266 @@
...
-(defn open-hyperlink-resource! [open-resource-fn resource _]
-  (open-resource-fn resource {}))
-
-(defmethod override-value-cell-view resource/Resource [{:keys [value open-resource-fn] :as property}]
+(defmethod override-value-cell-view resource/Resource [{:keys [value] :as property}]
   {:fx/type fx.hyperlink/lifecycle
    :style-class (into ["override-inspector-hyperlink"] (overridden-style-classes property))
-   :on-action (fn/partial #'open-hyperlink-resource! open-resource-fn value)
+   :on-action {:event-type :on-click-resource
+               :resource value}
```
matgis, line 502: Changed to use map-based event handler.

editor/src/clj/editor/search_results_view.clj
```diff
@@ -477,133 +495,266 @@
...
+  {:graphic (override-value-cell-view (assoc property :value (properties/value property)))})
+
+(defn- property-value [property-keyword state]
+  (-> state :properties property-keyword))
+
+(defmulti handle-override-inspector-event :event-type)
+
+(defmethod handle-override-inspector-event :on-refresh-view [{:keys [^Event fx/event]}]
+  (.consume event)
+  [[:refresh-view nil]])
```
matgis, line 515: The Refresh command will re-run the previous query against the new state of the graph.

editor/src/clj/editor/ui.clj
```diff
@@ -1490,6 +1491,26 @@
...
+
+            menu-items
+            (g/with-auto-or-fake-evaluation-context evaluation-context
+              (make-menu-items scene menu command-contexts keymap evaluation-context))]
+
+        (refresh-separator-visibility menu-items)
+        (refresh-menu-item-styles menu-items)
+        (.setAll (.getItems menu-button) (to-array menu-items))
+        nil))))
+
```
matgis, line 1513: Needed to be able to use handlers in the property label dropdown menus.

editor/styling/stylesheets/components/_fuzzy-combo-box.scss
```diff
@@ -25,8 +25,8 @@
   }
 
   .arrow-button {
-    -fx-max-height: 26px;
-    -fx-min-height: 26px;
+    -fx-max-height: 25px;
+    -fx-min-height: 25px;
```
matgis, line 29: The combo box controls were one pixel higher than all the other controls.

editor/test/editor/handler_test.clj
```diff
@@ -269,14 +269,15 @@
...
-      (is (= [[1] [1]] (eval-selections [local global] false))))))
+  (with-clean-system
+    (let [global (handler/->context :global {:selection [0]} nil {} {})]
+      (is (= [0] (eval-selection [global] true))))
+    (let [global (handler/->context :global {} (StaticSelection. [0]) {} {})]
+      (is (= [0] (eval-selection [global] true)))
+      (let [local (handler/->context :local {} (StaticSelection. [1]) {} {})]
+        (is (= [[1] [1] [0]] (eval-selections [local global] true))))
+      (let [local (handler/->context :local {} (StaticSelection. [1]) {} {})]
+        (is (= [[1] [1]] (eval-selections [local global] false)))))))
```
matgis, line 280: Needed to have a system because we create an `evaluation-context` when evaluating the dynamics now.

editor/test/integration/test_util.clj
```diff
@@ -971,7 +1065,6 @@
      game-object/ReferencedComponent
      (validate-component-resource-node-id basis (g/node-feeding-into basis node-id :source-resource))
 
-     :else
```
matgis, line 974: Whoops, that's not how `condp` works!

editor/test/internal/transaction_test.clj
```diff
@@ -77,7 +77,7 @@
   (testing "simple update"
     (ts/with-clean-system
       (let [[resource] (ts/tx-nodes (g/make-node world Resource :marker (int 0)))
-            tx-result  (g/transact (it/update-property resource :marker safe+ [42]))]
+            tx-result  (g/transact (it/update-property resource :marker safe+ [42] nil))]
```
matgis, line 80: `it/update-property` now takes an `opts` argument.

editor/test/support/test_support.clj
```diff
@@ -82,3 +82,13 @@
...
+
+(defmacro with-post-ec
+  "Given a symbol that resolves to a function, returns a fn that executes that
+  function in an implicit evaluation-context supplied as the final argument to
+  the function."
+  [fn-sym]
+  {:pre [(symbol? fn-sym)]}
+  `(fn ~(symbol (name fn-sym)) [& ~'args]
+     (g/with-auto-evaluation-context ~'evaluation-context
+       (apply ~fn-sym (conj (vec ~'args) ~'evaluation-context)))))
```
matgis, line 94: I've started to think having an arity that omits the `evaluation-context` for convenient use in tests is somewhat of an anti-pattern. When I removed the arity without the `evaluation-context` argument for my functions, I immediately noticed several places in the production code where we were creating temporary `evaluation-contexts` inside loop constructs. So I've stopped having those arities at all, and instead using this macro to create a wrapped function for use in tests.

editor/src/clj/util/coll.clj
```diff
@@ -608,14 +631,41 @@
...
               (reduced ret)))
           nil
           coll))
 
+(defn any?
+  "Returns whether any item in the supplied collection matches the predicate. Do
+  not confuse with clojure.core/any?, which takes a single argument and always
+  returns true."
+  [pred coll]
+  (boolean (some pred coll)))
```
matgis, line 647: Side note: The behavior of `clojure.core/any?` really annoys me, especially in light of `not-any?`, `every?`, and `not-every?` existing and leaving you with a reasonable expectation of what `any?` should do.

editor/src/clj/dynamo/graph.clj
vlaaad: Should we then add `;; SDK api` comment above?

editor/src/clj/editor/app_view.clj
vlaaad: That's nice!

editor/src/clj/editor/fxui.clj
```diff
@@ -593,85 +646,128 @@
...
+    (:icon/android) "M28.5,11c-0.6,0-1.1,0.2-1.5,0.5v0c0-3.4-1.8-6.3-4.4-8l1.1-2.2c0.1-0.2,0-0.5-0.2-0.7c-0.2-0.1-0.5,0-0.7,0.2L21.7,3c-1.3-0.6-2.7-1-4.2-1s-2.9,0.4-4.2,1l-1.1-2.1c-0.1-0.2-0.4-0.3-0.7-0.2c-0.2,0.1-0.3,0.4-0.2,0.7l1.1,2.2C9.8,5.2,8,8.1,8,11.5v0C7.6,11.2,7.1,11,6.5,11C5.1,11,4,12.1,4,13.5v8C4,22.9,5.1,24,6.5,24c0.6,0,1.1-0.2,1.5-0.5v1c0,1.4,1.1,2.5,2.5,2.5H11v4.5c0,1.4,1.1,2.5,2.5,2.5s2.5-1.1,2.5-2.5V27h3v4.6c0,1.3,1.1,2.4,2.4,2.4h0.3c1.3,0,2.4-1.1,2.4-2.4V27h0.5c1.4,0,2.5-1.1,2.5-2.5v-1c0.4,0.3,0.9,0.5,1.5,0.5c1.4,0,2.5-1.1,2.5-2.5v-8C31,12.1,29.9,11,28.5,11zM17.5,3c4.5,0,8.2,3.5,8.5,8H9C9.3,6.5,13,3,17.5,3zM6.5,23C5.7,23,5,22.3,5,21.5v-8C5,12.7,5.7,12,6.5,12S8,12.7,8,13.5v8C8,22.3,7.3,23,6.5,23zM26,24.5c0,0.8-0.7,1.5-1.5,1.5H23v5.6c0,0.8-0.6,1.4-1.4,1.4h-0.3c-0.8,0-1.4-0.6-1.4-1.4V26h-5v5.5c0,0.8-0.7,1.5-1.5,1.5S12,32.3,12,31.5V26h-1.5C9.7,26,9,25.3,9,24.5v-3v-8V12h17v1.5v8V24.5zM30,21.5c0,0.8-0.7,1.5-1.5,1.5S27,22.3,27,21.5v-8c0-0.8,0.7-1.5,1.5-1.5s1.5,0.7,1.5,1.5V21.5zM12.5,7.3c0-0.5,0.4-0.8,0.8-0.8c0.5,0,0.9,0.4,0.9,0.8s-0.4,0.9-0.9,0.9C12.9,8.2,12.5,7.8,12.5,7.3zM20.8,7.3c0-0.5,0.4-0.8,0.9-0.8c0.5,0,0.8,0.4,0.8,0.8s-0.4,0.9-0.8,0.9C21.2,8.2,20.8,7.8,20.8,7.3z"
```
vlaaad, line 652: Why wrap `:icon/android` etc. in a list?

editor/src/clj/editor/graph_util.clj
vlaaad: Nice!

editor/src/clj/editor/menu_items.clj
```diff
@@ -0,0 +1,64 @@
...
+
+(ns editor.menu-items
+  (:require [util.fn :as fn]))
+
+(set! *warn-on-reflection* true)
+(set! *unchecked-math* :warn-on-boxed)
+
+;; -----------------------------------------------------------------------------
+;; Common menu definitions shared across files.
+;; Keep sorted alphabetically!
```
vlaaad, line 23: > Keep sorted alphabetically!

:D

editor/src/clj/editor/properties.clj
```diff
@@ -569,12 +579,14 @@
...
 
-(defn keyword->name [kw]
-  (-> kw
-    name
-    camel/->Camel_Snake_Case_String
-    (clojure.string/replace "_" " ")
-    clojure.string/trim))
+(defn- keyword->name-raw [property-keyword]
+  (-> property-keyword
+      (name)
```
vlaaad, line 584: Why wrap with parens?

editor/src/clj/editor/properties.clj
vlaaad: Yeah, this kind of stuff is what worries me about l10n. Also pluralization... I did it once for a game, and while developing it, I even learned something about the Russian language! Turns out, while English has 2 forms for counted words (singular and plural), the Russian language actually has 3: singular, plural for numbers ending on 2/3/4, and plural for numbers ending on 5/6/7/8/9/0. It's not obvious, given that only a very small percentage of words have different forms for different plural types...

editor/src/clj/editor/properties_view.clj
vlaaad: Hmm, is it possible for us to show a property for a node that is not in the project graph? It could be in the view graph instead...

editor/src/clj/util/coll.clj
```diff
@@ -189,6 +190,28 @@
...
+  collection are themselves lazily-evaluated."
+  [value]
+  `(condp identical? (class ~value)
+     LazySeq true
+     Eduction true
+     Repeat true
+     Cycle true
+     false))
+
+(defn eager-seq?
```
vlaaad: Maybe call it `eager-seqable?` instead? `seq` is a well-defined Clojure term, and this function does does not satisfy the meaning of the term

editor/src/clj/util/text_util.clj
```diff
@@ -322,3 +322,70 @@
...
+     1 "One"
+     2 "Two"
+     3 "Three"
+     4 "Four"
+     5 "Five"
+     6 "Six"
+     7 "Seven"
+     8 "Eight"
+     9 "Nine"
+     (str count))))
```
vlaaad, line 368: I would prefer if we just used numbers, and 0 instead of `no`. In a technical context, I think numerals are much clearer, i.e.:
- "Push Down Color Override to 2 Descendants Across 2 Resources" — good action description
- "Push Down Color Override to Two Descendants Across Two Resources" — good blog post title

editor/src/clj/util/text_util.clj
```diff
@@ -322,3 +322,70 @@
...
+  represent the amount."
+  (^String [^long count ^String singular]
+   (amount-text count singular nil))
+  (^String [^long count ^String singular ^String plural]
+   (amount-text count singular plural
+                (if (and (not (.isEmpty singular))
+                         (Character/isUpperCase (.codePointAt singular 0)))
+                  count->upper-case-string
+                  count->lower-case-string)))
+  (^String [^long count ^String singular ^String plural count->string]
```
vlaaad: Uh oh, only one plural form... It's not going to be easy to add localization...

editor/styling/stylesheets/components/_fuzzy-combo-box.scss
vlaaad: Nice catch!

editor/test/support/test_support.clj
vlaaad: Yeah, I agree, the code would be clearer if more functions always required `evaluation-context`. Similarly, if there were fewer functions like `g/set-property!` that call `g/transact` for convenience. Also, it should be called `g/transact!`

editor/test/util/coll_test.clj
```diff
@@ -275,6 +275,49 @@
...
+  (is (true? (coll/lazy? (repeat 1))))
+  (is (true? (coll/lazy? (repeatedly 1 rand))))
+  (is (true? (coll/lazy? (cycle [1]))))
+  (is (true? (coll/lazy? (map identity [1]))))
+  (is (true? (coll/lazy? (eduction (map identity) [1])))))
+
+(deftest eager-seq?-test
+  (is (true? (coll/eager-seq? nil)))
+  (is (true? (coll/eager-seq? "")))
+  (is (true? (coll/eager-seq? [])))
```
vlaaad: vector is not a seq

editor/test/util/coll_test.clj
```diff
@@ -275,6 +275,49 @@
...
+  (is (true? (coll/eager-seq? (vector-of :long))))
+  (is (true? (coll/eager-seq? '())))
+  (is (true? (coll/eager-seq? {})))
+  (is (true? (coll/eager-seq? #{})))
+  (is (true? (coll/eager-seq? (sorted-map))))
+  (is (true? (coll/eager-seq? (sorted-set))))
+  (is (true? (coll/eager-seq? (double-array 0))))
+  (is (true? (coll/eager-seq? (object-array 0))))
```
vlaaad: Maps, sets, and arrays are not seqs.

vlaaad (dismissed):
LGTM! I only have a few nitpicks about naming and labels for your consideration, but don't let them stop you from merging :)

vlaaad:
I forgot to add, the new property label UI is very nice, I like it!

vlaaad:
Styling nitpick:

Screenshot 2025-06-25 at 14 38 21

I found the "Color overrides of" text to be unexpectedly bright.

editor/src/clj/editor/fxui.clj
matgis: Simply because I think mixing bare keywords and lists looks messy. We match several keywords for `(:icon/triangle-error :icon/triangle-warning)`.

editor/src/clj/editor/properties.clj
matgis: It just looks nicer when there is a mix of calls that need additional arguments. I wouldn't do it for a one-liner, but for a multi-line expression like this I think it looks neater.

editor/src/clj/util/coll.clj
matgis: Good point! I'll rename it to `eager-seqable?`.

editor/src/clj/util/text_util.clj
matgis: Yep, proper localization will take some effort, but I don't want that to dissuade us from having nice things while we're still focusing on English. I suspect we might end up using a library that will have its own take on this sort of thing when we get around to it.

editor/src/clj/util/text_util.clj
matgis: Makes sense! I think I'll update the `amount-text` function so you'll always have to supply a `count->string` function to make it explicit how you want to format the numbers.

editor/test/support/test_support.clj
matgis: God yes.

editor/src/clj/dynamo/graph.clj
matgis: Not sure, to be honest. I kind of feel like most of the stuff in here is SDK api. Maybe we only add it for somewhat unexpected stuff?

vlaaad: Yeah perhaps...

editor/src/clj/editor/properties_view.clj
matgis: Hmm, good point. I think I got rid of the `context` argument at some earlier stage, but then restored it because it was needed for the new color picker functionality or something. I think I'll restore this code to use the `project` and `workspace` from the `context` again.

matgis:
Hmm, yes. I think I'll change it to a single label that reads like the transfer menu options.

vlaaad (approved):
Ship it!

PR #10883, opened 2025-06-24
Increase initial delay before showing code hover

vlaaad:
Now the hover popup in the code editor is less annoying.

### DIFF
--- a/editor/src/clj/editor/code/view.clj
+++ b/editor/src/clj/editor/code/view.clj
@@ -1676,9 +1676,12 @@
     (and (get-property view-node :completions-showing evaluation-context)
          (pos? (count (get-property view-node :completions-combined evaluation-context))))))
 
-(defn- hover-visible? [view-node]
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/code/view.clj
```diff
@@ -2570,7 +2573,8 @@
      (schedule-hover-refresh! view-node evaluation-context)))
   ([view-node evaluation-context]
    (some-> (g/node-value view-node :hover-request evaluation-context) ui/cancel)
-   {:hover-request (ui/->future 0.2 #(refresh-hover-state! view-node))}))
+   (let [delay (if (hover-visible? view-node evaluation-context) 0.2 1.0)]
```
vlaaad, line 2576: This behavior is closer to vscode: initially, there is a longer delay before showing the popup. Once the popup is shown, its state is refreshed more often so it can, e.g., close faster.

AGulev (approved):
❤️

PR #10885, opened 2025-06-25
Added an option to not create game objects for the bones of a model

### LINKED ISSUE #10884: Add option to not create game objects for the bones of a model
For each bone of a model's skeleton we also create an associated game object. With a large number of models it's possible to quickly reach the max limit of game objects in a collection. 

An option to not create a game object per bone.

britzl:
The model component now has an option to not create one game object per bone of a skeleton. When bone creation is disabled it is possible to have a high number of animated and skinned models in a scene (example with 15k models):

image

image

In addition to the above change the engine will now also warn if the bone pose matrix texture cache is too small so that the developer can take action and increase the max size in game.project (`model.max_bone_matrix_texture_width` and `model.max_bone_matrix_texture_height`).

Fixes #10884 

## Technical notes
This PR also contains a fix for the rig where the pose matrix cache (backed by a dmArray) capacity was reset each frame, causing new allocations every update. Instead of resetting the capacity the cache size is set to 0 to clear it but maintain the capacity. Prior to this change almost all time was spent resizing the array, which quickly becomes noticeable when having a large number of models:

image

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


editor/src/clj/editor/model.clj
```diff
@@ -443,6 +445,8 @@
           (g/fnk [material-scene-infos]
             (model-scene/make-material-name->material-scene-info material-scene-infos)))
 
+  (property create-go-bones g/Bool (default false))
```
vlaaad: We use defaults from protobuf (and the default there is true). Is this intentional? Looks like a breaking change...

vlaaad: Also, for defaults, we don't typically hard-code them, but instead look them up, e.g. `(protobuf/default ModelProto$ModelDesc :create-go-bones)`

britzl: You are completely right. The default should be true.

britzl: Side-note: extension-spine should be changed in a similar fashion:

https://github.com/defold/extension-spine/blob/main/defold-spine/editor/src/spineext.clj#L938

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -364,6 +364,10 @@ namespace dmGameSystem
     static void CompModelPoseCallback(void* user_data1, void* user_data2)
     {
         ModelComponent* component = (ModelComponent*)user_data1;
+        if (!component->m_Resource->m_Model->m_CreateGoBones)
+        {
+            return;
+        }
```
JCash: If we don't want this callback, I think it's probably better to not register it?
This will be called each update, and we do 3 pointer dereferences here.

engine/rig/src/rig.cpp
```diff
@@ -52,8 +52,8 @@ namespace dmRig
 
     static void ResetPoseMatrixCache(PoseMatrixCache* cache)
     {
-        cache->m_PoseMatrices.SetCapacity(0);
-        cache->m_CacheEntryOffsets.SetCapacity(0);
+        cache->m_PoseMatrices.SetSize(0);
+        cache->m_CacheEntryOffsets.SetSize(0);
```
JCash, line 56: Nice find!

JCash (changes requested):
lgtm
could avoid the callback altogether

JCash (dismissed):

Issue #10781, opened 2025-06-04
Editor setting needed to disable auto-inserting of closing parens in the code editor.


Sublustris:
Is it possible to add a setting to the editor to disable automatic completion of brackets?
It does not always work conveniently, and it is very annoying.

Here is one example - I need to turn a block of code into a multi-page comment, I put the first part of the quotes at the beginning of the block, and I have to delete the auto-completed part, because I need the closing quote in a completely different place.

![Image](https://github.com/user-attachments/assets/8029c58c-c05b-4395-8df0-417a05cd5dd0)

![Image](https://github.com/user-attachments/assets/e26bd3f2-a561-4a9f-a868-951ee1c93239)

Sublustris:
another case - during rapid input, when i type something like **print()** there is one unexpected bracket appearing and i have to delete it manually.

This doesn't happen when typing slowly, but I've been coding for decades, and some typing patterns are automatic. I'm unlikely to be able to force myself to type slower.

![Image](https://github.com/user-attachments/assets/f82a00f6-6714-47d1-8571-50287d9322ef)

Qwarr:
Optionally would like an option to disable auto completing `" "` too, I just have muscle memory of manually typing it out so having it auto complete and duplicating just throws me off.

Wolfe2x7:
I can't stand these things or the mandatory warm orange theme, which left me to pursue an external code editor out of need, rather than for the sake of familiarity.

I really like Lite-XL, though, so overall it's a win. 🙂

Klaleus:
Adding options to disable this kind of behavior is probably the wrong approach. There are several other small but frequent annoyances like this. Instead of disabling them, they need to be prioritized as editor bugs and outright fixed.

AGulev:
1. Case 1
```
sprite.play_flipbook("#sprite", go.property("value"))
```
add this line to youe code, put cursor afetr "value" and try to print: `msg.url()`

https://github.com/user-attachments/assets/5deec2c5-16d3-4bab-b10d-fbb1a79bcc74

2. Case 2: `print()` - autocompletion happens and second bracket appears

https://github.com/user-attachments/assets/372ff8a0-78ab-454e-8b30-f3f5c3ed13a3

AGulev:
The issues I reported in my comment were fixed in beta (should be available. in 20-30 mins)

Sublustris:
1.10.4 alpha

[...]

PR #10888, opened 2025-06-25
Don't use `.` and `(` as completion auto-commit characters

vlaaad:
Previously, the text editor would accept displayed completion if the user typed `.` or `(`. It results in a surprising behavior if the user was simply typing and didn't look at the completion list. Other editors like vscode don't accept selected completions when typing these characters, so we now follow the same behavior.

Related to #10781


vlaaad:
This PR solves both issues reported in [this comment](https://github.com/defold/defold/issues/10781#issuecomment-2972859283).

AGulev (approved):
It SOOOOOO much better now, thank you!

PR #10892, opened 2025-06-25
Improve callback invalidation

### LINKED ISSUE #8759: Add dmLogError if can't create callback
I am trying to create websocket server in  lua module. So server was created when i require that lua module.
So websocket try to create callback but script instance not exist.

I spend 2+h trying to understand why server not call callback on events.

https://github.com/defold/defold/blob/427457b596d6d7f8a6c1da1c489c403d471b4333/engine/script/src/script.cpp#L1556

Please add dmLogErrors if callback was not created

https://github.com/defold/extension-websocket/pull/49/files#r1551530184

AGulev:
Ensure that callbacks properly track whether they are still valid and avoid situations where the wrong callback could be accidentally destroyed under certain conditions.

Fix https://github.com/defold/defold/issues/10868
Fix https://github.com/defold/defold/issues/8759

### DIFF
--- a/engine/gameobject/src/gameobject/gameobject_script.cpp
+++ b/engine/gameobject/src/gameobject/gameobject_script.cpp
@@ -363,6 +363,13 @@ namespace dmGameObject
         return 1;
     }
 
+    static int ScriptGetUniqueScriptId(lua_State* L)
+    {
+        ScriptInstance* inst = (ScriptInstance*)lua_touserdata(L, 1);
+        lua_pushinteger(L, (lua_Integer)inst->m_UniqueScriptId);
+        return 1;
+    }
+
     static const luaL_reg ScriptInstance_methods[] =
     {
         {0,0}
@@ -379,6 +386,7 @@ namespace dmGameObject
         {dmScript::META_TABLE_IS_VALID,                 ScriptInstanceIsValid},
         {dmScript::META_GET_INSTANCE_CONTEXT_TABLE_REF, ScriptGetInstanceContextTableRef},
         {dmScript::META_GET_INSTANCE_DATA_TABLE_REF,    ScriptGetInstanceDataTableRef},
+        {dmScript::META_GET_UNIQUE_SCRIPT_ID,           ScriptGetUniqueScriptId},
         {0, 0}
     };
 
@@ -2451,6 +2459,7 @@ bail:
         i->m_Instance = instance;
         i->m_ScriptWorld = script_world->m_ScriptWorld;
         i->m_ComponentIndex = component_index;
+        i->m_UniqueScriptId = dmScript::GenerateUniqueScriptId();
         NewPropertiesParams params;
         params.m_ResolvePathCallback = ScriptInstanceResolvePathCB;
         params.m_ResolvePathUserData = (uintptr_t)L;
--- a/engine/script/src/script.cpp
+++ b/engine/script/src/script.cpp
@@ -65,12 +65,14 @@ namespace dmScript
     const char META_TABLE_IS_VALID[]                 = "__is_valid";
     const char META_GET_INSTANCE_CONTEXT_TABLE_REF[] = "__get_instance_context_table_ref";
     const char META_GET_INSTANCE_DATA_TABLE_REF[]    = "__get_instance_data_table_ref";
+    const char META_GET_UNIQUE_SCRIPT_ID[]           = "__get_unique_script_id";
 
     const char SCRIPT_METATABLE_TYPE_HASH_KEY_NAME[] = "__dmengine_type";
     static const uint32_t SCRIPT_METATABLE_TYPE_HASH_KEY = dmHashBufferNoReverse32(SCRIPT_METATABLE_TYPE_HASH_KEY_NAME, sizeof(SCRIPT_METATABLE_TYPE_HASH_KEY_NAME) - 1);
 
     // A debug value for profiling lua references
     int g_LuaReferenceCount = 0;
+    const uint32_t INVALID_SCRIPT_ID = 0xFFFFFFFF;
 
     HContext NewContext(const ContextParams& params)
     {
@@ -1127,6 +1129,18 @@ namespace dmScript
 
         lua_pop(L, 1);
         // [-1] value
+    }
+
+    uint32_t GenerateUniqueScriptId()
+    {
+        static uint32_t counter = 0;
+
+        if (counter == INVALID_SCRIPT_ID - 1)
+            counter = 0;
+        else
+            ++counter;
+
+        return counter;
     }
 
     HScriptWorld NewScriptWorld(HContext context)
--- a/engine/script/src/script.cpp
+++ b/engine/script/src/script.cpp
@@ -1136,9 +1136,13 @@ namespace dmScript
         static uint32_t counter = 0;
 
         if (counter == INVALID_SCRIPT_ID - 1)
+        {
             counter = 0;
+        }
         else
+        {
             ++counter;
+        }
 
         return counter;
     }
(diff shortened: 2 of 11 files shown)


engine/gameobject/src/gameobject/gameobject_script.cpp
```diff
@@ -379,6 +386,7 @@ namespace dmGameObject
         {dmScript::META_TABLE_IS_VALID,                 ScriptInstanceIsValid},
         {dmScript::META_GET_INSTANCE_CONTEXT_TABLE_REF, ScriptGetInstanceContextTableRef},
         {dmScript::META_GET_INSTANCE_DATA_TABLE_REF,    ScriptGetInstanceDataTableRef},
+        {dmScript::META_GET_UNIQUE_SCRIPT_ID,           ScriptGetUniqueScriptId},
```
AGulev, line 389: All script types require new `META_GET_UNIQUE_SCRIPT_ID` now, so we can check it in `script.cpp`

engine/script/src/script.cpp
```diff
@@ -1587,7 +1606,9 @@ namespace dmScript
         GetInstance(L);
         // [-1] instance
 
-        if (!GetMetaFunction(L, -1, META_GET_INSTANCE_CONTEXT_TABLE_REF, sizeof(META_GET_INSTANCE_CONTEXT_TABLE_REF) - 1)) {
+        if (!GetMetaFunction(L, -1, META_GET_INSTANCE_CONTEXT_TABLE_REF, sizeof(META_GET_INSTANCE_CONTEXT_TABLE_REF) - 1))
+        {
+            dmLogError("CreateCallback failed: missing meta function for instance context table");
```
AGulev: Added more info if callback can't be created. In some cases it happens silently  which makes debugging hard. I also found this one https://github.com/defold/defold/issues/8759

engine/script/src/script.cpp
```diff
@@ -1628,6 +1671,7 @@ namespace dmScript
         // [-2] callback
         // [-1] LuaCallbackInfo
 
+        cbk->m_UniqueScriptId = unique_script_id;
```
AGulev, line 1677: save `unique_script_id` in callback, so we can use it for comparasing.

engine/script/src/script.cpp
```diff
@@ -1655,14 +1699,96 @@ namespace dmScript
         return cbk;
      }
 
+    static bool IsCallbackInstanceValid(LuaCallbackInfo* cbk)
```
AGulev, line 1705: This function checks if the instance the callback refers to is the same instance it expects by comparing the IDs.

engine/script/src/script.cpp
```diff
@@ -1655,14 +1699,96 @@ namespace dmScript
...
     bool IsCallbackValid(LuaCallbackInfo* cbk)
     {
         if (cbk == NULL ||
             cbk->m_L == NULL ||
             cbk->m_ContextTableRef == LUA_NOREF ||
             cbk->m_CallbackInfoRef == LUA_NOREF ||
             cbk->m_Callback == LUA_NOREF ||
-            cbk->m_Self == LUA_NOREF) {
+            cbk->m_Self == LUA_NOREF ||
+            !IsCallbackInstanceValid(cbk))
```
AGulev: no 100%, do we need it here?

engine/script/src/script.cpp
```diff
@@ -1655,14 +1699,96 @@ namespace dmScript
...
+
+        const int context_table_stack_index = lua_gettop(L);
+        lua_rawgeti(L, context_table_stack_index, cbk->m_Self);
+        // [-3] old instance
+        // [-2] context table
+        // [-1] instance
+        if (lua_isnil(L, -1))
+        {
+            lua_pop(L, 3);
+            cbk->m_UniqueScriptId = INVALID_SCRIPT_ID;
```
AGulev, line 1734: if we found that the callback is invalid once - cache the result.

engine/script/src/script.cpp
```diff
@@ -1681,23 +1807,27 @@ namespace dmScript
...
-
-                // For the callback (that can actually outlive the script instance)
-                // we want to add to the lua debug count
+            }
+            
+            // For the callback (that can actually outlive the script instance)
+            // we want to add to the lua debug count
+            if (cbk->m_CallbackInfoRef != LUA_NOREF)
+            {
                 dmScript::Unref(L, LUA_REGISTRYINDEX, cbk->m_CallbackInfoRef);
```
AGulev, line 1836: we call `dmScript::Ref` for  `cbk->m_CallbackInfoRef` separatly from `cbk->m_ContextTableRef` when we create callback. So, even if callback isn't valid, this Ref still exist and we have unref it.

engine/script/src/test/test_script_lua.cpp
```diff
@@ -1066,6 +1076,64 @@ TEST_F(ScriptTestLua, LuaBooleanFunctions)
...
+    dmScript::InvokeCallback(cbk3, 0, 0);
+    dmScript::Unref(L, LUA_REGISTRYINDEX, instanceref3);
+    dmScript::DestroyCallback(cbk3);
+
+    // At the end, assert that _G.cb_call_count == 2
+    ASSERT_TRUE(RunString(L, "assert(_G.cb_call_count == 2)"));
+    // If DestroyCallback(cbk1) works fine and doesn't remove what it doesn't own,
+    // this assert should pass
+    ASSERT_TRUE(RunString(L, "assert(_G.cb2_call_count == 1)"));
+}
```
AGulev, line 1135: this test cases validates issue from the https://github.com/defold/defold/issues/10868

engine/script/src/script.cpp
```diff
@@ -1605,7 +1626,28 @@ namespace dmScript
...
 
         int context_table_ref = lua_tonumber(L, -1);
-        lua_pop(L, 2);
+
+        lua_pop(L, 1);
+        // [-1] instance
+
+        uint32_t unique_script_id = INVALID_SCRIPT_ID;
+        if (GetMetaFunction(L, -1, META_GET_UNIQUE_SCRIPT_ID, sizeof(META_GET_UNIQUE_SCRIPT_ID) - 1))
+        {
```
JCash: In which cases will this fail?
Should we allow creating a callback in that case?
(i..e I don't see a return upon error?)

engine/script/src/script.cpp
```diff
@@ -1655,14 +1699,96 @@ namespace dmScript
...
+
+        GetInstance(L);
+        // [-1] old instance
+        lua_rawgeti(L, LUA_REGISTRYINDEX, cbk->m_ContextTableRef);
+        // [-2] old instance
+        // [-1] context table
+        if (lua_type(L, -1) != LUA_TTABLE)
+        {
+            lua_pop(L, 2);
+            cbk->m_UniqueScriptId  = INVALID_SCRIPT_ID;
```
JCash, line 1722: Hmm, I'm not a fan of modifying data in a check function (i.e. it's name "Is...")
I.e. I'd rather update the callback after the `IsCallbackInstanceValid()` returned false.

engine/script/src/test/test_script_lua.cpp
```diff
@@ -219,6 +219,7 @@ struct TestDummy {
     dmMessage::URL  m_URL;
     int             m_InstanceReference;
     int             m_ContextTableReference;
+    int             m_UniqueScriptId;
```
JCash: uint32_t

JCash (dismissed):
lgtm

tiny change request though, so move the modification outside of the "Is.." function

engine/script/src/script.cpp
```diff
@@ -1631,21 +1631,24 @@ namespace dmScript
...
-            // [-1] META_GET_UNIQUE_SCRIPT_ID()
-            lua_pushvalue(L, -2); // push instance
-            // [-3] instance
-            // [-2] META_GET_UNIQUE_SCRIPT_ID()
-            // [-1] instance
-            lua_call(L, 1, 1);    // call __get_unique_script_id(self)
-            // [-2] instance
-            // [-1] unique script id
-            unique_script_id = (uint32_t)lua_tointeger(L, -1);
+            dmLogError("CreateCallback failed: missing %s meta function for instance context table", "UNIQUE_SCRIPT");
```
JCash, line 1636: 👍

JCash (approved):

PR #10897, opened 2025-06-26
Fix glitches on the first frame of the GPU-skinned instanced animation right after spawn.

AGulev:
It also increases performance for GPU-skinned instanced animation by about 30–50%, depending on the platform and the specific case.

### Technical changes
This fix moves writing texture matrices into the component update to ensure that the animation's cached texture is ready in advance, before the first render call happens.

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_model.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_model.cpp
@@ -1906,6 +1906,8 @@ namespace dmGameSystem
 
         dmRig::Result rig_res = dmRig::Update(world->m_RigContext, params.m_UpdateContext->m_DT);
 
+        WritePoseMatricesToTexture(world);
+
         assert(world->m_MaxBatchIndex < VERTEX_BUFFER_MAX_BATCHES);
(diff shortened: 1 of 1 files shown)


engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1906,6 +1906,8 @@ namespace dmGameSystem
 
         dmRig::Result rig_res = dmRig::Update(world->m_RigContext, params.m_UpdateContext->m_DT);
 
+        WritePoseMatricesToTexture(world);
```
JCash, line 1909: I think this is a good fix.

For correctness, I think we should (later on) make sure we can verify that the async upload of the texture is actually done. Theoretically, we could still end up rendering and the texture hasn't finished uploading.

JCash (approved):

engine/gamesys/src/gamesys/components/comp_model.cpp
AGulev: But `WritePoseMatricesToTexture ` uses `SetTexture` not `SetTextureAsync`

PR #10900, opened 2025-06-27
Fix layout restore on Linux

sprocketc:
Fixes #10865

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -582,10 +582,12 @@
 (defn restore-split-positions! [^Scene scene prefs]
   (let [split-positions (stored-split-positions prefs)
         split-panes (existing-split-panes scene)]
-    (doseq [[id positions] split-positions]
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/app_view.clj
```diff
@@ -582,10 +582,12 @@
...
-    (doseq [[id positions] split-positions]
-      (when-some [^SplitPane split-pane (get split-panes id)]
-        (.setDividerPositions split-pane (double-array positions))
-        (.layout split-pane)))))
+    ;; The nested run-later fixes restore on Linux, by forcing an initial rendering pass. 
+    (ui/run-later
+      (doseq [[id positions] split-positions]
+        (when-some [^SplitPane split-pane (get split-panes id)]
+          (.setDividerPositions split-pane (double-array positions))
+          (.layout split-pane))))))
```
sprocketc, line 590: Everything works as expected up until this point. Although `restore-split-positions!` is already wrapped in a `run-later`, the layout pass won't work on Linux.

matgis (approved):

PR #10915, opened 2025-06-30
Set size of pose matrix cache only once per frame if needed

britzl:
The pose matrix cache in rig.cpp was resized once per rig instance. This caused a big slowdown when a large number of instances were added in a single frame. The pose matrix size is now calculated once per frame and the cache itself is resized only once per frame if needed.

### DIFF
--- a/engine/rig/src/rig.cpp
+++ b/engine/rig/src/rig.cpp
@@ -54,6 +54,7 @@ namespace dmRig
     {
         cache->m_PoseMatrices.SetSize(0);
         cache->m_CacheEntryOffsets.SetSize(0);
+        cache->m_TotalBoneCount = 0;
         cache->m_MaxBoneCount = 0;
     }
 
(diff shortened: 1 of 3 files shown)


engine/rig/src/rig.cpp
```diff
@@ -519,14 +533,10 @@ namespace dmRig
             return;
         }
 
-        dmArray<Matrix4>& cache_pose_matrices = context->m_PoseMatrixCache.m_PoseMatrices;
-
-        const uint32_t num_pose_matrices = cache_pose_matrices.Size();
-        EnsureSize(cache_pose_matrices, num_pose_matrices + bone_count);
-
-        context->m_PoseMatrixCache.m_CacheEntryOffsets[instance->m_PoseMatrixCacheIndex] = num_pose_matrices;
+        PoseMatrixCache* cache = &context->m_PoseMatrixCache;
```
britzl, line 536: We no longer calculate the offset in this function. The offset is instead calculated when a rig instance acquires a pose matrix slot.

engine/rig/src/rig.cpp
```diff
@@ -1205,16 +1215,20 @@ namespace dmRig
...
-            context->m_PoseMatrixCache.m_CacheEntryOffsets.OffsetCapacity(32);
+            cache->m_CacheEntryOffsets.OffsetCapacity(32);
         }
 
-        context->m_PoseMatrixCache.m_CacheEntryOffsets.SetSize(next_index + 1);
+        cache->m_CacheEntryOffsets.SetSize(next_index + 1);
 
         instance->m_PoseMatrixCacheIndex = next_index;
 
+        cache->m_CacheEntryOffsets[next_index] = cache->m_TotalBoneCount;
```
britzl: New: calculate offset into cache here. Also sum up the bone count here so that we can resize the cache only once at the beginning of the Update

engine/rig/src/rig.cpp
```diff
@@ -519,14 +533,10 @@ namespace dmRig
             return;
         }
 
-        dmArray<Matrix4>& cache_pose_matrices = context->m_PoseMatrixCache.m_PoseMatrices;
-
-        const uint32_t num_pose_matrices = cache_pose_matrices.Size();
-        EnsureSize(cache_pose_matrices, num_pose_matrices + bone_count);
```
britzl, line 525: Here's the problem. If we add 100 new rig instances in a frame we resize the array 100 times in the first frame. Better to resize the array once outside of the loop over the rig instances (which starts in the Update function)

JCash (approved):

PR #10918, opened 2025-07-01
Edit go files using editor scripts

vlaaad:
It's now possible to edit components of a game object file using editor scripts. The components come in 2 flavors: referenced and embedded. Referenced components use type `reference` and act as references to other resources, only allowing overrides of go properties defined in scripts. Embedded components use types like `sprite`, `label`, etc., and allow editing of all properties defined in the component type, as well as adding sub-components like shapes of collision objects. For example, you can use the following code to set up a game object:
```lua
editor.transact({
    editor.tx.add("/npc.go", "components", {
        type = "sprite",
        id = "view"
    }),
    editor.tx.add("/npc.go", "components", {
        type = "collisionobject",
        id = "collision",
        shapes = {
            {
                type = "shape-type-box",
                dimensions = {32, 32, 32}
            }
        }
    }),
    editor.tx.add("/npc.go", "components", {
        type = "reference",
        path = "/npc.script"
        id = "controller",
        __hp = 100 -- set a go property defined in the script
    })
})
```

### Technical notes

I had to refactor the whole attachment implementation to support what outline calls "alt-outline", meaning looking up an alternative node for attaching the attachment. Previously, attachment mostly worked on types: node types had lists, e.g., `AtlasNode` had `images`. With a dynamic alternative node lookup, now individual nodes have lists. This is necessary because the node type `EmbeddedComponent` has no way of telling if it has any lists. Instead, an instance of an `EmbeddedComponent` might refer to a resource node of type `CollisionObjectNode`, and only when we find the embedded node can we say that it has `shapes`.

Related to #8504

### DIFF
--- a/editor/src/clj/editor/attachment.clj
+++ b/editor/src/clj/editor/attachment.clj
@@ -261,7 +261,7 @@
 (defn- clear-tx [{:keys [basis] :as evaluation-context} workspace node-id list-kw]
   (let [[node-id list-definition] (require-list-definition workspace node-id list-kw evaluation-context)]
     (assert (list-definition-editable? list-definition node-id evaluation-context))
-    (mapcat
+    (coll/mapcat
       #(g/delete-node (g/override-root basis %))
       (list-definition-get list-definition node-id evaluation-context))))
 
--- a/editor/src/clj/editor/attachment.clj
+++ b/editor/src/clj/editor/attachment.clj
@@ -13,25 +13,61 @@
 ;; specific language governing permissions and limitations under the License.
 
 (ns editor.attachment
-  "Extensible definitions for semantical node attachments"
-  (:refer-clojure :exclude [remove alias])
+  "Extensible definitions for semantical node attachment lists
+
+  Use register, alias, or define-alternative to configure the lists.
+  Use defined?, editable?, and reorderable? to query the status of node lists on
+  a given node.
+  Use get to list attached nodes.
+  Use add, remove, reorder and clear to edit the nodes.
+
+  The node attachment state has the following data shape:
+
+  {node-type {:lists {list-kw {:add {node-type tx-attach-fn}
+                               :get get-fn
+                               :reorder reorder-fn
+                               :read-only? read-only-pred
+                               :alias node-type
+                               :aliases #{node-type}}}
+              :alternative alternative-fn}}
+
+  In the list definition map, only :get is required. Fns:
+    tx-attach-fn      fn of parent-node, child-node -> txs
+    get-fn            fn of node, evaluation-context -> vector of nodes
+    reorder-fn        fn of reordered-nodes -> txs
+    read-only-pred    fn of parent-node, evaluation-context -> boolean
+
+  Aliasing allows treating a list as the same as another list. Any list
+  definition may have at most one of :alias, :aliases keys, where:
+    :alias      refers to a node type that this list definition follows, i.e., a
+                link to the \"original\"
+    :aliases    a set of all node types that follow this definition, i.e., a
+                link to \"copies\"
+
+  Similarly to outline's :alt-outline feature, it's possible to define an
+  alternative node to look up lists by defining an alternative-fn (fn of node,
+  evaluation-context -> alt-node | nil)"
+  (:refer-clojure :exclude [alias remove get])
   (:require [dynamo.graph :as g]
             [editor.workspace :as workspace]
             [internal.graph.types :as gt]
             [util.coll :as coll]))
 
 (defn- assoc-list-definition [state node-type list-kw {:keys [aliases] :as definition}]
-  (let [state (update state node-type assoc list-kw definition)]
+  (let [state (assoc-in state [node-type :lists list-kw] definition)]
     (if aliases
       (let [aliased-definition (-> definition
                                    (dissoc :aliases)
                                    (assoc :alias node-type))]
-        (reduce #(update %1 %2 assoc list-kw aliased-definition) state aliases))
+        (reduce #(assoc-in %1 [%2 :lists list-kw] aliased-definition) state aliases))
       state)))
 
 ;; SDK api
 (defn register
   "Create transaction steps that register a semantical list of child components
+
+  The very first registration has to specify :get. Further registrations may
+  extend the definition, for example, by adding additional :add entries
 
   Args:
     workspace    the workspace node id
@@ -71,6 +107,7 @@
     (fn [s]
       (let [definition (-> s
                            (clojure.core/get node-type)
+                           :lists
                            list-kw
                            (cond->
                              add (update :add coll/merge add)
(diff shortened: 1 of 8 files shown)


editor/src/clj/editor/attachment.clj
```diff
@@ -13,26 +13,62 @@
...
+
+  The node attachment state has the following data shape:
+
+  {node-type {:lists {list-kw {:add {node-type tx-attach-fn}
+                               :get get-fn
+                               :reorder reorder-fn
+                               :read-only? read-only-pred
+                               :alias node-type
+                               :aliases #{node-type}}}
+              :alternative alternative-fn}}
```
vlaaad, line 32: Moved this data shape description from `editor.workspace` ns because it belongs here.

editor/src/clj/editor/attachment.clj
```diff
@@ -84,206 +121,199 @@
...
+          (g/add-node (g/construct child-node-type :_node-id child-node-id))
+          (init-fn parent-node-id child-node-id)
+          (tx-attach-fn parent-node-id child-node-id)
+          (config-fn child-node-id))))))
 
-  Asserts that it exists. See [[editable?]]"
-  [basis workspace node-type list-kw]
-  {:post [(some? %)]}
-  (-> basis (workspace/node-attachments workspace) (get node-type) list-kw :add))
+(defn add
```
vlaaad, line 227: Changed the `add` function. Previously, it accepted a tree. Now, it can only add one item, but adding the tree is now supported on a use site: in addition to `init-fn` that initializes the node before attaching it, `add` now supports `config-fn` that is invoked after the node is attached. Editor scripts attach more nodes to the attached node using `config-fn`.

editor/src/clj/editor/attachment.clj
```diff
@@ -84,206 +121,199 @@
...
+(defn- list-definition-editable? [list-definition node-id evaluation-context]
+  (and (contains? list-definition :add)
+       (if-let [pred (:read-only? list-definition)]
+         (not (pred node-id evaluation-context))
+         true)))
 
-(defn read-only?
-  "Checks if a node is read-only, i.e., can't be edited even when editable
+(defn editable?
+  "Checks if a node list is editable
```
vlaaad, line 187: Removed `read-only?` — merged it into `editable?`. Both functions were essentially about the same thing, but `editable?` used to work on node types, while `read-only?` worked on node instances. Since now all attachment code works on node instances, there is no need to separate them.

editor/src/clj/editor/editor_extensions/graph.clj
```diff
@@ -420,12 +416,11 @@
...
           (g/expand-ec
             (fn [evaluation-context]
-              (let [basis (:basis evaluation-context)
-                    workspace (project/workspace project evaluation-context)
+              (let [workspace (project/workspace project evaluation-context)
                     collision-id->node-id
                     (coll/pair-map-by
                       #(g/node-value % :id evaluation-context)
-                      ((attachment/getter basis workspace (g/node-type* basis node-id) :collision-groups) node-id evaluation-context))
+                      (attachment/get workspace node-id :collision-groups evaluation-context))
```
vlaaad, line 423: Also, I renamed `getter` to `get` since the only need to look up a `getter` was to invoke it immediately.

editor/src/clj/editor/game_object.clj
```diff
@@ -388,14 +395,10 @@
...
 
-(defn- attach-component [self-id comp-id ddf-input resolve-id?]
+(defn- attach-component [self-id comp-id ddf-input]
   ;; self-id: GameObjectNode
   ;; comp-id: EmbeddedComponent, ReferencedComponent
   (concat
-    (when resolve-id?
-      (->> (g/node-value self-id :component-ids)
-           keys
-           (g/update-property comp-id :id id/resolve)))
```
vlaaad, line 398: Simplified the code: I moved the id resolution code where it was requested.

editor/test/integration/editor_extensions_test.clj
```diff
@@ -1506,6 +1506,69 @@ After transaction (clear):
...
+  - type: embedded-component-type-collectionproxy
+    id: collectionproxy
+  - type: embedded-component-type-collectionproxy
+    id: collectionproxy1
+  - type: embedded-component-type-collisionobject
+    id: collisionobject-embedded
+    shapes: 1
+    - id: box
+      type: shape-type-box
+      dimensions: 2.5 2.5 2.5
```
vlaaad, line 1527: Yay! We support adding embedded components with nested data.

editor/src/clj/editor/attachment.clj
```diff
@@ -84,206 +121,199 @@
...
-(defn defines?
+    workspace         the workspace to transact to
+    node-type         the node type that defines an alternative
+    alternative-fn    a function that looks up an alternative node id if no
+                      matching list was found on the input node; will receive 2
+                      args: node id (instance of provided node-type) and
+                      evaluation-context, should return an alternative node-id
+                      or nil"
+  [workspace node-type alternative-fn]
+  (g/update-property workspace :node-attachments #(update % node-type assoc :alternative alternative-fn)))
```
matgis, line 154: Should it use the `node-type-kw` as the key instead of the `node-type` here? I'm thinking of what might happen if the `node-type` is redefined when loading an editor plugin or through the REPL.

editor/src/clj/editor/attachment.clj
```diff
@@ -84,206 +121,199 @@
...
+  (let [[parent-node-id list-definition] (require-list-definition workspace parent-node-id list-kw evaluation-context)]
+    (assert (list-definition-editable? list-definition parent-node-id evaluation-context))
+    (let [tx-attach-fn (-> list-definition :add (clojure.core/get child-node-type))]
+      (assert tx-attach-fn)
+      (let [child-node-id (first (g/take-node-ids (g/node-id->graph-id parent-node-id) 1))]
+        (concat
+          (g/add-node (g/construct child-node-type :_node-id child-node-id))
+          (init-fn parent-node-id child-node-id)
+          (tx-attach-fn parent-node-id child-node-id)
+          (config-fn child-node-id))))))
```
matgis, line 225: Should we supply the `evaluation-context` to these functions?

editor/src/clj/editor/attachment.clj
```diff
@@ -84,206 +121,199 @@
...
-    (assert (not (read-only? workspace node-id list-kw evaluation-context)))
+  Asserts that list is defined. See [[defined?]]"
+  [workspace node-id list-kw evaluation-context]
+  (let [[node-id list-definition] (require-list-definition workspace node-id list-kw evaluation-context)]
+    (list-definition-get list-definition node-id evaluation-context)))
+
+(defn- clear-tx [{:keys [basis] :as evaluation-context} workspace node-id list-kw]
+  (let [[node-id list-definition] (require-list-definition workspace node-id list-kw evaluation-context)]
+    (assert (list-definition-editable? list-definition node-id evaluation-context))
     (mapcat
```
matgis: `core.mapcat` is really slow. Should we take this opportunity to use `coll/mapcat` instead?

editor/src/clj/editor/game_object.clj
```diff
@@ -620,19 +641,111 @@
...
+    (-> attachment
+        (dissoc "type")
+        (eutil/provide-defaults
+          "id" (rt/->lua (id/gen component-ext (g/node-value parent-node-id :component-ids evaluation-context))))
+        (ext-graph/attachment->set-tx-steps child-node-id rt project evaluation-context))))
+
+(defn- set-referenced-attachment-properties [evaluation-context rt project parent-node-id child-node-id attachment]
+  (-> attachment
+      ;; Ignore the path since it's applied first in init-attachment
+      (dissoc "path")
```
matgis, line 704: Maybe this is not the correct place for this question, but is is possible to set the path to refer to a different component resource using the script API?

I think we may want to disallow that if we intend to collapse the `ReferencedComponent` node with the component `ResourceNode` it is referencing in the future.

We could still make it work by replacing the entire node in response to setting the property, of course, but we may not want to expose ourselves to that potential hassle?

editor/src/clj/editor/game_object.clj
```diff
@@ -620,19 +641,111 @@
...
+  (fn ReferencedComponent-getter [_node-id property _evaluation-context]
+    (case property
+      "type" constantly-ext-referenced-component-type
+      nil)))
+
+(ext-graph/register-property-getter!
+  ::EmbeddedComponent
+  (fn EmbeddedComponent-getter [node-id property evaluation-context]
+    (case property
+      "type" #(str ext-embedded-component-type-prefix (resource/ext (g/node-value node-id :source-resource evaluation-context)))
```
matgis: Maybe we should intern these strings, or memoize them?

editor/test/resources/editor_extensions/transact_attachment_project/test.editor_script
```diff
@@ -497,6 +521,116 @@ function M.get_commands()
...
+                        id = "collisionobject-referenced",
+                        type = "referenced-component",
+                        path = "/test.collisionobject"
+                    }),
+                    editor.tx.add(go, "components", {
+                        id = "referenced-tilemap",
+                        type = "referenced-component",
+                        path = "/test.tilemap"
+                    })
+                })
```
matgis, line 601: I'm a bit torn about the type names. Would it make sense to have the type of embedded components be just the `type-ext`? The string prefix and `starts-with?` checks kind of irk me, and it sort of feels like simply adding a `type = "sprite"` is nicer. And maybe `"referenced-component"` could just be `"reference"` or something? Maybe I'm missing something important, though...

editor/src/clj/editor/attachment.clj
vlaaad: Hmm, didn't we discuss it before? It does not matter since `node-type` is a ref, i.e., a `NodeTypeRef` instance, so redefining the node does not affect it.

editor/src/clj/editor/attachment.clj
vlaaad: No, it's actually pretty important that we don't do that here. The evaluation context here is pre-transaction, but `init-fn` and `config-fn` are invoked at different points within the transaction, and they often need to use an evaluation context at these points within the transaction. For example, `init-fn` is used when a node is constructed but before it's attached. This means it can resolve the `id` property without its id polluting the id counters. Similarly, `config-fn` is invoked after the node is attached, so that if it recursively adds more nodes, their parent nodes will be properly connected and will report correct id counters.

editor/src/clj/editor/attachment.clj
vlaaad: Yes!

editor/src/clj/editor/game_object.clj
vlaaad: Yes, it is possible — `path` is an outline property, and all outline properties that are editable by the user in the UI are also editable by editor scripts in the same way. I'm not aware of the decision to get rid of ReferencedComponent nodes... Feels wrong to make it read-only while keeping it editable in the Outline.

editor/src/clj/editor/attachment.clj
matgis: Yeah, I brought it up before. I think I was confused about what a `NodeTypeRef` is. Its not like an `atom`, where the internal state (and as a result, the `hashCode`) might change - its just a keyword that resolves to a value in a global map, which should be fine.

editor/src/clj/editor/attachment.clj
matgis: Makes sense!

editor/src/clj/editor/game_object.clj
matgis: Yeah, perhaps the main thing is to not have an `EmbeddedComponentNode` that "wraps" each component `ResourceNode`, but it would be nice to also do this for referenced components (like we do in `NonEditableGameObjectNode`, which is vastly more efficient than the regular `GameObjectNode`).

When the it was only in the UI, we could get rid of the Path field and instead have a **Replace With...** action available from the Outline Context Menu for this sort of refactoring, but if we expose it to scripting we're kind of stuck with it. 🤔

matgis: But maybe having a `path` property from the script point of view would be preferable anyway? We could still have it perform the transaction steps for replacing the node entirely if we decide to get rid of the `ReferencedComponentNode` wrapper. So maybe it's not really an issue after all. 👍

matgis (approved):

editor/src/clj/editor/game_object.clj
vlaaad: We don't expose `ReferencedComponent` or `EmbeddedComponent` to editor scripts at all, so at most I foresee a small breaking change where an option becomes read-only. Given that it's an editor script, the damage is extremely localized and inconsequential.

vlaaad: And it will be localized only to editor scripts that change resources of existing referenced components. Given that this minor breaking change will be released in conjunction with UI changes that remove the said property, this is entirely justified and expected. I don't see the benefit of adding restrictions here for the sake of us anticipating some future changes that may or may not be breaking...

vlaaad: Is this removal of ReferencedComponent something that has already been decided? I don't remember us discussing it previously.

matgis: We talked about various ways to decrease the number of nodes in the graph during one of our weekly meetups, and removing the wrapper nodes around components was discussed as a prime target given that they serve no function apart from providing transform properties.

I think it was right after we introduced the `NonEditable*` nodes? 🤔

But we haven't scheduled a formal task for it, I think. So it's more of a thing we may look into the next time we run into memory issues, or as a mitigating general efficiency change.

PR #10925, opened 2025-07-04
Edit collections using editor scripts

vlaaad:
Now it's possible to edit collections using editor scripts. You can add game objects (embedded or referenced) and collections (referenced). For example:
```lua
local coll = "/char.collection"
editor.transact({
    editor.tx.add(coll, "children", {
        -- embbedded game object
        type = "go",
        id = "root",
        children = {
            {
                -- referenced game object
                type = "go-reference",
                path = "/char-view.go"
                id = "view"
            },
            {
                -- referenced collection
                type = "collection-reference",
                path = "/body-attachments.collection"
                id = "attachments"
            }
        },
        -- embedded gos can also have components
        components = {
            {
                type = "collisionobject",
                id = "collision",
                shapes = {
                    {type = "shape-type-box", dimensions = {2.5, 2.5, 2.5}}
                }
            },
            {
                type = "component-reference",
                id = "controller",
                path = "/char.script",
                __hp = 100 -- set a go property defined in the script
            }
        }
    })
})
```

Related to #8504

### DIFF
--- a/editor/src/clj/editor/collection.clj
+++ b/editor/src/clj/editor/collection.clj
@@ -15,12 +15,17 @@
 (ns editor.collection
   (:require [dynamo.graph :as g]
             [editor.app-view :as app-view]
+            [editor.attachment :as attachment]
             [editor.build-target :as bt]
             [editor.code.script :as script]
             [editor.collection-common :as collection-common]
             [editor.collection-string-data :as collection-string-data]
             [editor.core :as core]
             [editor.defold-project :as project]
+            [editor.editor-extensions.coerce :as coerce]
+            [editor.editor-extensions.graph :as ext-graph]
+            [editor.editor-extensions.node-types :as node-types]
+            [editor.editor-extensions.runtime :as rt]
             [editor.game-object :as game-object]
             [editor.game-object-common :as game-object-common]
             [editor.graph-util :as gu]
@@ -34,13 +39,15 @@
             [editor.resource-node :as resource-node]
             [editor.scene :as scene]
             [editor.types :as types]
+            [editor.util :as eutil]
             [editor.validation :as validation]
             [editor.workspace :as workspace]
             [internal.cache :as c]
             [internal.util :as util]
             [util.eduction :as e])
   (:import [com.dynamo.gameobject.proto GameObject$CollectionDesc GameObject$CollectionInstanceDesc GameObject$EmbeddedInstanceDesc GameObject$InstanceDesc]
-           [internal.graph.types Arc]))
+           [internal.graph.types Arc]
+           [org.apache.commons.io FilenameUtils]))
 
 (set! *warn-on-reflection* true)
 
@@ -70,7 +77,12 @@
   (inherits core/Scope)
   (property id g/Str ; Required pb field.
             (dynamic error (g/fnk [_node-id id id-counts]
-                             (or (validation/prop-error :fatal _node-id :id (partial validation/prop-id-duplicate? id-counts) id)
+                             (or (when id-counts
+                                   ;; We evaluate the _properties output before
+                                   ;; the node is connected into parent when
+                                   ;; creating nodes from editor scripts. At
+                                   ;; that point, id-counts may be nil
+                                   (validation/prop-error :fatal _node-id :id (partial validation/prop-id-duplicate? id-counts) id))
                                  (validation/prop-error :warning _node-id :id validation/prop-contains-prohibited-characters? id "Id"))))
             (dynamic read-only? (g/fnk [_node-id]
                                   (g/override? _node-id))))
@@ -130,8 +142,7 @@
   ;; coll-id: CollectionNode
   ;; child-id: CollectionInstanceNode
   (concat
-    (for [[from to] [[:_node-id :nodes]
-                     [:ddf-message :ref-coll-ddf]
+    (for [[from to] [[:ddf-message :ref-coll-ddf]
                      [:id :ids]
                      [:source-resource :referenced-collection-resources]
                      [:build-targets :sub-build-targets]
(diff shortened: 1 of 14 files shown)


editor/src/clj/editor/collection.clj
```diff
@@ -130,8 +142,7 @@
   ;; coll-id: CollectionNode
   ;; child-id: CollectionInstanceNode
   (concat
-    (for [[from to] [[:_node-id :nodes]
```
vlaaad, line 133: It was a mistake to connect the attached collection's `_node-id` to `nodes` of its parent here. The function is called `attach-coll-coll`, but this is a misnomer. The actual code that performs an attachment is spread across 2 functions: `attach-coll-coll` and `child-coll-any`. This split exists to support various attachment function flavors that differ depending on a parent (embedded go, ref go, coll node) and a child (embedded go, ref go, ref coll). It is always a responsibility of `child-*` functions to connect the `_node-id` output to the `nodes` input. Previously, the attached referenced collection was connected twice: first in `attach-coll-coll`, and then in `child-coll-any`. Now it's only connected once, as expected.

editor/src/clj/editor/game_object_non_editable.clj
```diff
@@ -374,18 +375,20 @@
...
-    :sanitize-fn (partial sanitize-non-editable-game-object workspace)
-    :string-encode-fn (partial string-encode-non-editable-game-object workspace)
-    :load-fn load-non-editable-game-object
-    :allow-unloaded-use true
-    :icon game-object-common/game-object-icon
-    :icon-class :design
-    :view-types [:scene :text]
-    :view-opts {:scene {:grid true}}))
+  (concat
+    (attachment/register workspace NonEditableGameObjectNode :components :get (constantly []))
```
vlaaad, line 379: The only addition.

editor/src/clj/editor/collection_non_editable.clj
```diff
@@ -426,18 +427,20 @@
...
-    :sanitize-fn (partial sanitize-non-editable-collection workspace)
-    :string-encode-fn (partial string-encode-non-editable-collection workspace)
-    :load-fn load-non-editable-collection
-    :allow-unloaded-use true
-    :icon collection-common/collection-icon
-    :icon-class :design
-    :view-types [:scene :text]
-    :view-opts {:scene {:grid true}}))
+  (concat
+    (attachment/register workspace NonEditableCollectionNode :children :get (constantly []))
```
vlaaad, line 431: The only addition.

editor/src/clj/editor/particlefx.clj
```diff
@@ -916,8 +916,7 @@
                                                       :icon emitter-icon
                                                       :children (outline/natural-sort child-outlines)
                                                       :child-reqs [{:node-type ModifierNode
-                                                                    :tx-attach-fn (fn [self-id child-id]
-                                                                                    (attach-modifier self-id child-id true))}]}))
+                                                                    :tx-attach-fn attach-modifier}]}))
```
vlaaad, line 919: The default last arg is true.

editor/src/clj/editor/collection.clj
```diff
@@ -145,9 +156,31 @@
 
 (declare EmbeddedGOInstanceNode ReferencedGOInstanceNode CollectionNode)
 
-(defn- go-id->node-ids [go-id]
-  (let [collection (core/scope-of-type go-id CollectionNode)]
-    (g/node-value collection :ids)))
+(defn- tx-attach-go-referenced-go
```
vlaaad, line 159: The real attach fn...

editor/src/clj/editor/collection.clj
matgis: I'm not really a fan of connecting the child nodes to the `:nodes` input like we do in some places. It's an extra connection that doesn't serve any purpose outside establishing a `:cascade-delete` relationship. In many cases we should probably designate a different (necessary) input as the `:cascade-delete` input.

Not really something for this PR, but it is something we may want to address at some point.I suppose the editor script tests will start failing if we decide to refactor things later?

editor/test/resources/editor_extensions/transact_attachment_project/test.editor_script
```diff
@@ -631,6 +661,93 @@ function M.get_commands()
...
+                                                        type = "shape-type-box",
+                                                        dimensions = {2.5, 2.5, 2.5},
+                                                    }
+                                                }
+                                            },
+                                            {
+                                                type = "reference",
+                                                path = "/test.script",
+                                                __num = 37
+                                            }
```
matgis, line 708: Hmm. I don't like that we have `type = "reference"` for component references, but `type = "reference-go"` and `type = "reference-collection"` for game object and collection references, respectively. Since both game object and collection references can be added to the `"children"` list, I suppose we can't use `type = "reference"` for both of these as well?

If they absolutely need to be three distinct types, I think it would feel better to not use `type = "reference"` for component references either. In that case I think I would prefer something like `"component-reference"`, `"go-reference"` and `"collection-reference"`. What do you think?

editor/test/resources/editor_extensions/transact_attachment_project/test.editor_script
```diff
@@ -631,6 +661,93 @@ function M.get_commands()
...
+                })
+
+                print("After transaction (add gos and collections):")
+                print_coll(coll)
+
+                print("Transaction: clear collection")
+                editor.transact({ editor.tx.clear(coll, "children") })
+
+                print("After transaction (clear collection)")
+                print_coll(coll)
```
matgis, line 804: I'd like to also see some tests that add referenced and embedded game objects below an already-existing game object in a collection.

Similarly, we're missing tests for adding them to an existing game object located below an existing collection reference. Not really sure if the editor lets you do that, and where it ends up in that case, but I think we need to have a well-defined behavior for it. 

Finally, we should test that you get an error if you try to add or remove stuff from game objects or collections below a read-only directory. Although maybe that's already covered in a resource-agnostic way? There may be special circumstances due to them being different node types, though.

matgis (changes requested):

editor/test/resources/editor_extensions/transact_attachment_project/test.editor_script
vlaaad: Yes, we need distinct types. I think your suggestion makes sense, I'll rename types.

editor/test/resources/editor_extensions/transact_attachment_project/test.editor_script
vlaaad: Good call, I'll add more tests for editing already existing structures.

> Similarly, we're missing tests for adding them to an existing game object located below an existing collection reference. Not really sure if the editor lets you do that, and where it ends up in that case, but I think we need to have a well-defined behavior for it.

The editor lets you do that, but the edit is applied to the referenced collection. I disallow such edits in editor scripts, since they give you a false impression that you are editing this collection. Basically, the only allowed edits are those that make the edited file dirty.

editor/test/resources/editor_extensions/transact_attachment_project/test.editor_script
```diff
@@ -743,6 +743,60 @@ function M.get_commands()
...
+                assert(referenced_coll, "referenced collection not found")
+                assert(referenced_coll_go, "referenced collection's go not found")
+                assert(readonly_referenced_coll, "readonly referenced collection not found")
+                assert(readonly_referenced_go, "readonly referenced go not found")
+
+                print("Transaction: edit already existing collection elements")
+                editor.transact({
+                    editor.tx.add(embedded_go, "children", {type = "go", id = "new-embedded-go-child"}),
+                    editor.tx.add(referenced_go, "children", {type = "go", id = "new-referenced-go-child"}),
+                    editor.tx.add(readonly_referenced_go, "children", {type = "go", id = "new-readonly-referenced-go-child"}),
```
vlaaad, line 787: We actually can add children to gos from read-only dirs, because new gos are added in the edited collection.

matgis (approved):
LGTM! ❤️

PR #10928, opened 2025-07-04
Improve editor tests performance

vlaaad:
I decided to profile the test run and noticed one obvious performance issue that was trivial to fix. ZipOutputStream is horribly slow. I replaced the uses of it with ZipFile. For comparison, `load-zip` performance before on builtins.zip:
```
Evaluation count : 54 in 6 samples of 9 calls.
             Execution time mean : 12,272537 ms
    Execution time std-deviation : 43,141670 µs
   Execution time lower quantile : 12,229410 ms ( 2,5%)
   Execution time upper quantile : 12,332157 ms (97,5%)
                   Overhead used : 1,961034 ns
```
And after:
```
Evaluation count : 11154 in 6 samples of 1859 calls.
             Execution time mean : 53,891997 µs
    Execution time std-deviation : 405,917057 ns
   Execution time lower quantile : 53,447161 µs ( 2,5%)
   Execution time upper quantile : 54,386894 µs (97,5%)
                   Overhead used : 1,961034 ns
```
Yep, 53 µs vs 12 ms. That's a 200x improvement.

Here is a comparison of the test run time before and after the change:
```
before:
lein test  322.99s user 70.70s system 110% cpu 5:54.92 total
after:
lein test  279.82s user 66.18s system 115% cpu 4:59.84 total
```

This is a 43 second (or 13%) improvement.

### DIFF
--- a/editor/src/clj/editor/library.clj
+++ b/editor/src/clj/editor/library.clj
@@ -27,7 +27,7 @@
            [java.net HttpURLConnection URI]
            [java.nio.file.attribute FileTime]
            [java.util Base64]
-           [java.util.zip ZipEntry ZipInputStream ZipOutputStream]
+           [java.util.zip ZipEntry ZipFile ZipInputStream ZipOutputStream]
            [org.apache.commons.codec.digest DigestUtils]
            [org.apache.commons.io FilenameUtils]))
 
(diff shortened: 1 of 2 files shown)


vlaaad:
On CI, it reduced the time to build the editor from `19m52s` to `18m26s`, a 7% improvement.

editor/src/clj/editor/library.clj
```diff
@@ -230,15 +230,16 @@
...
+      (fn [_ ^ZipEntry entry]
         (let [parts (str/split (FilenameUtils/separatorsToUnix (.getName entry)) #"/")
               name (last parts)]
-          (if (= file-name name)
-            {:name name :path (str/join "/" (butlast parts))}
-            (recur (.getNextEntry zip))))))))
+          (when (= file-name name)
+            (reduced {:name name :path (str/join "/" (butlast parts))}))))
+      nil
+      (.stream zip))))
```
matgis, line 242: This change looks fine, but if this is indeed a hot spot, we could probably do even better if we use `string/ends-with?` on the `ZipEntry` name (and check that the character right before the match is one of the separator characters). We don't need to split and convert the separators in the `ZipEntry` name until after we've found a match.

matgis (approved):

editor/src/clj/editor/library.clj
vlaaad: The hotspot was in .getNextEntry, i.e. IO.

PR #10930, opened 2025-07-04
Improve create-go-bones checkbox handling

vlaaad:


editor/src/clj/editor/model.clj
```diff
@@ -136,9 +136,9 @@
             rig-scene-pb-msg {}
             rig-scene-build-target (rig/make-rig-scene-build-target workspace _node-id rig-scene-pb-msg dep-build-targets rig-scene-dep-build-targets)
             rt-pb-msg (-> {:rig-scene (:resource rig-scene-build-target)
-                           :default-animation (:default-animation pb-msg)
+                           :default-animation default-animation
                            :materials (:materials pb-msg)
-                           :create-go-bones (:create-go-bones pb-msg)}
+                           :create-go-bones create-go-bones}
```
vlaaad, line 141: `pb-msg` omits defaults, so there could be nils.

editor/src/clj/editor/model.clj
```diff
@@ -529,15 +529,16 @@
...
-      :animations (workspace/resolve-resource resource animations)
-      :create-go-bones create-go-bones)
+    (let [resolve-resource #(workspace/resolve-resource resource %)]
+      (gu/set-properties-from-pb-map self ModelProto$ModelDesc model-desc
+        name :name
+        default-animation :default-animation
+        mesh (resolve-resource :mesh)
+        skeleton (resolve-resource :skeleton)
+        animations (resolve-resource :animations)
+        create-go-bones :create-go-bones))
```
vlaaad, line 541: `create-go-bones` could be nil, and that would cause the value to be set to nil, and then it was displayed as an indeterminate check-box in the editor.

matgis (approved):
LGTM!

PR #10933, opened 2025-07-06
Make sure the Windows bundle works when run via Proton

### LINKED ISSUE #8767: Defold games now crash when run via Proton (Could not find gl function 'glGenPrograms')
Running a bundle of a Windows project on Linux via Proton makes the game open as a black window for a moment before closing. Native Linux bundles work as normal.

**To Reproduce (REQUIRED)**
1. Make a Windows bundle of a project
2. Add it to Steam on Linux
3. Set it to use Proton (any version) in the properties window
4. Run it

Tested with 3 different projects built with 1.7.0 (a project of mine, the desktop template, and the [Chromacell demo](https://store.steampowered.com/app/2293310/Chromacell/)).

**Platforms (REQUIRED):**
- Fedora Linux 39
- Radeon RX 6700 XT GPU

**Logs (OPTIONAL):**
Logs taken from the desktop template.
- Defold log:  [log.txt](https://github.com/defold/defold/files/14898229/log.txt)

```
ERROR:GRAPHICS: Could not find gl function 'glGenPrograms'.
FATAL:ENGINE: Unable to create the graphics context.
```

- Proton log (via PROTON_LOG=1): [steam-11239744614107709440.log](https://github.com/defold/defold/files/14898232/steam-11239744614107709440.log)
- [_crash](https://github.com/defold/defold/files/14898284/_crash.txt)
- [_crash.dmp](https://github.com/defold/defold/files/14898286/_crash.dmp.txt)

A very old Windows bundle I have (Defold 1.2.171) still runs just fine on Proton.

This shouldn't be a big issue with how easy it is to bundle for other platforms with Defold, but 7 out of the 25 [Defold games on Steam](https://steamdb.info/instantsearch/?refinementList%5Btechnologies%5D%5B0%5D=Engine.Defold) are still Windows-only.

AGulev:
Fixed an issue where running a Windows project bundle on Linux via Proton would close with `Could not find gl function 'glGenPrograms'`.

Fix https://github.com/defold/defold/issues/8767#event-18490035618

# Technical changes
Defold requests an OpenGL 3.3 core profile (or 3.1 if 3.3 is not available), but then tries to load legacy `ARB_vertex_program` / `ARB_fragment_program` entry points like `glGenProgramsARB`, which were removed from the core spec.  
Strict drivers, such as Mesa (used by Proton/Wine), return `NULL` for those functions.

It seems that native Windows drivers still expose these outdated symbols for compatibility, so the bug is only reproducible when the game runs using Proton.  
Since we don’t use these functions internally, they can be safely removed.


engine/modelc/scripts/modelviewer.py
```diff
@@ -1,172 +0,0 @@
-# Copyright 2020-2025 The Defold Foundation
```
AGulev, line 1: We don't really use it. It looks like very outdated script.

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -996,11 +988,6 @@ static void LogFrameBufferError(GLenum status)
             return false;\
         }
 
-        GET_PROC_ADDRESS(glGenProgramsARB, "glGenPrograms", PFNGLGENPROGRAMARBPROC);
-        GET_PROC_ADDRESS(glBindProgramARB, "glBindProgram", PFNGLBINDPROGRAMARBPROC);
-        GET_PROC_ADDRESS(glDeleteProgramsARB, "glDeletePrograms", PFNGLDELETEPROGRAMSARBPROC);
-        GET_PROC_ADDRESS(glProgramStringARB, "glProgramString", PFNGLPROGRAMSTRINGARBPROC);
-        GET_PROC_ADDRESS(glProgramLocalParameter4fARB, "glProgramLocalParameter4f", PFNGLVERTEXPARAMFLOAT4ARBPROC);
```
AGulev, line 1003: These functions never used and outdated

JCash (approved):

PR #10941, opened 2025-07-07
Edit node id in outline

### LINKED ISSUE #9306: Edit Id property with F2
Sort of. When ever I try to edit the **Id** property of a GUI node or a GO (anything really that has an **Id** property) I instinctively press F2 and start typing the new name just to realize that doesn't work.

Pressing F2 when an item in the **Outline** window is selected would make the **Id** field focused.

None

I just now noticed that pressing TAB does exactly what I want. I would still want to have an alternative hotkey for that, which is F2. Rationale being that brain thinks of "renaming" the **Id** field and F2 is used to rename everything.

### LINKED ISSUE #10740: Change node ID from outline
It is time consuming when we need to always go to the properties of the nodes in order to change their ID. Sometimes we need to quickly change just it and nothing else from the properties.

We would like to have the option to change the nodes ID from the outline and not only via clicking on it and changing from the property panel

We could have it as a double click or with a hotkey or something

sprocketc:
rename-id

Node ids can be edited in place on the outline tree, by pressing `F2` or with a delayed double click. The keyboard shortcut is customizable. 

Fixes #10740
Fixes #9306


editor/src/clj/editor/outline_view.clj
```diff
@@ -569,6 +659,24 @@
...
+    (condp = (.getCode event)
+      KeyCode/ENTER
+      (when-not editing-id
+        (.consume event)
+        (ui/run-command (.getSource event) :file.open-selected))
+
+      KeyCode/F2
+      (when (some #(= "F2" (.toString ^KeyCodeCombination %)) rename-shortcuts)
+        (.consume event)
+        (ui/run-command (.getSource event) :node.rename))
```
sprocketc: This replaces `(ui/bind-key-commands! {"Enter" :file.open-selected})`, because we need to block `:file.open-selected` when we commit the editing changes on `Enter`. Also, the key-down `F2` event was previously consumed by javafx, even though the built-in editing that uses it is set to false, so `:node.rename :outline` handler wasn't working for `F2`. Since the shortcut is customizable, we need to check if it exists first.

editor/src/clj/editor/outline_view.clj
```diff
@@ -476,6 +478,75 @@
...
+                                   (.getDesktopProperty "awt.multiClickInterval"))
+                               400)]
+    (.consume event)
+    (ui/user-data! text-label ::last-click-time current-click-time)
+    (cond
+      (< time-diff db-click-threshold) 
+      (ui/run-command (.getSource event) :file.open-selected)
+      
+      (< time-diff (* db-click-threshold 2)) 
+      (ui/run-command (.getSource event) :node.rename)
```
sprocketc: Attempt to honor the system db-click-threshold. 400 to 500 should be a reasonable default range. Doubling the threshold for the delayed db-click also seems reasonable to me.

editor/src/clj/editor/outline_view.clj
```diff
@@ -476,6 +478,75 @@
...
                                                                   (AnchorPane/setRightAnchor visibility-button 0.0))))))
 
+(defn- set-editing-id!
+  [^TreeView tree-view node-id]
+  (let [editing-id (ui/user-data tree-view ::editing-id)]
+    (when (not= editing-id node-id)
+      (doto tree-view
+        (ui/user-data! ::editing-id node-id)
+        (.refresh)
+        (.requestFocus)))))
```
sprocketc, line 491: A refresh is required to update the editing state.

editor/src/clj/editor/outline_view.clj
matgis: I think it would be good to add a comment to the code here with this information, because it sort of looks like the rename action is hard-coded to F2 at first glance.

vlaaad (changes requested):
I found 2 issues:
1. Exception if there is no id property — open code file, focus Outline, press F2: `java.lang.AssertionError: Assert failed: Property :id not found in node-type :editor.code.script/ScriptNode.
property-def`
2. It allows renaming things that should not be renamable. For example, open GUI file, select `Nodes`, press F2 — it can be renamed. We should exclude properties that have `visible` dynamic return `false`.

editor/src/clj/editor/keymap.clj
```diff
@@ -71,6 +71,7 @@
            ["F11" :debugger.step-into]
            ["F2" :file.rename]
            ["F2" :code.rename]
+           ["F2" :node.rename]
```
vlaaad: Hmm, perhaps we should have a single `:edit.rename` command that renames either a file, or a code symbol, or an id, depending on the context? I don't remember what was the reasoning when I added a separate `code.rename` command, but now when I see 3 of them together they looks wrong. Perhaps create an issue for that?

matgis (changes requested):
The renaming feels really nice! Great work on getting it to feel like a natural part of the Tree View! Some feedback:
* I get an exception if I press F2 on the root node in the Outline.
* I think it would be nice for discoverability if there was also a **Rename...** menu item displaying the F2 shortcut in the context menu, like we have in the Assets panel.
* I don't think the delayed double-click interaction is quite how entering the rename state typically works. I find I can't get into the rename state consistently, and sometimes enter it unexpectedly when selecting nodes in quick succession. In Finder, it feels more like it goes into the rename state if:
  1. A single file is selected.
  2. I click on the filename label itself.
  3. Enough time elapses to make sure it is not a double-click.

vlaaad (approved):
LGTM! Works wonders!

sprocketc:
@vlaaad I was trying to also migrate `:code.rename` and `:file.rename` to `:edit.rename` on the same PR. Should we skip it?

vlaaad:
Nah, it's good!

matgis (approved):
Works great! LGTM! ❤️

I think we should also make it possible to rename certain nodes that have a `:name` property instead of an `:id` (fonts, materials and textures inside guis come to mind), but let's merge this for now! 👍

PR #10944, opened 2025-07-08
Create resources in editor scripts

vlaaad:
It's now possible to create resources (from templates or with a custom file content) using editor scripts. To do it, use the new `editor.create_resources` function. For example:
```lua
-- create a single resource:
editor.create_resources({ "/npc.go" })

-- create multiple resources:
editor.create_resources({"/npc.go", "/level.collection"})

-- create a resource with custom content:
editor.create_resources({
  {"/npc.script", "go.property('hp', 100)"}
})
```
Creating multiple resources in a single call is more efficient than creating them one by one, as it requires only a single editor resource sync.


vlaaad:
Requesting @britzl so he can provide feedback on the API.

editor/test/resources/editor_extensions/resources_io_project/test.editor_script
```diff
@@ -0,0 +1,81 @@
...
+
+function M.get_commands()
+    return {
+        editor.command({
+            label = "Test",
+            locations = {"Edit"},
+            id = "defold.test",
+            run = function()
+                create_resources({
+                    {"/test/config.json", content = "{\"test\": true}"}
```
matgis: Would it be better to have a key for the resource path as well? This looks cute, but might be more awkward to work with programmatically? It feels like it leverages a quirk of Lua to achieve a nicer syntax, but is a bit odd. Can I supply multiple paths? Maybe it's a feature to create a bunch of files with the same content? 🤔

editor/src/clj/editor/editor_extensions.clj
```diff
@@ -154,6 +155,56 @@
       (and (not (resource/resource? node-id-or-resource))
            (some? (graph/ext-lua-value-setter node-id-or-resource property rt project evaluation-context))))))
 
+(def ^:private created-resources-coercer
+  (coerce/vector-of
+    (coerce/one-of
+      graph/resource-path-coercer
+      (coerce/hash-map :req {1 graph/resource-path-coercer}
+                       :opt {:content coerce/string}
```
matgis: Is creating resources from string content really the preferred way to do it? I suppose it's useful, but it's really annoying to create files with resources embedded as strings like this. Maybe we could rename the key to `text`, and reserve `content` for something like a Lua table representation like the Outline attachment syntax we have for adding stuff to an existing resource?

editor/src/clj/editor/editor_extensions.clj
```diff
@@ -154,6 +155,56 @@
...
+                                          (when (fs/path-exists? file-path)
+                                            (if (fs/path-is-directory? file-path)
+                                              (throw (LuaError. (str "Directory already exists in place of file: " proj-path)))
+                                              (throw (LuaError. (str "Resource already exists: " proj-path)))))
+                                          (let [content (or content
+                                                            (let [file-name (str (.getFileName file-path))
+                                                                  base-name (FilenameUtils/removeExtension file-name)
+                                                                  ext (string/lower-case (FilenameUtils/getExtension file-name))
+                                                                  resource-type (get (resource/resource-types-by-type-ext basis workspace :editable) ext)
+                                                                  template (or (workspace/template workspace resource-type evaluation-context) "")]
```
matgis: I would move these two out of the `future/io` block so we don't hold on to the `basis`. We could get the `resource-types-by-type-ext` map and a `template-by-type-ext` map before entering the `future/io` block.

matgis (changes requested):

editor/src/clj/editor/editor_extensions.clj
vlaaad: We could add `:properties` later if needed, which would be a more descriptive name.

editor/test/resources/editor_extensions/resources_io_project/test.editor_script
vlaaad: Like `{resource = "/foo.json", content = "{}"}`? Seems too verbose, @britzl didn't like this approach when I designed the `zip.pack` API, so we decided to use similar approach there: https://defold.com/ref/stable/editor-lua/#zip.pack:output_path-opts-entries

vlaaad: Perhaps we could make the create_resources arg to be an array or a map of path to string content, e.g.:
```
editor.create_resources({
  "/foo.go",
  "/foo.collection",
  ["/foo.json"] = "[true, {}]"
})
```
That way, later, we could support properties by accepting table value for kv pairs. @britzl ?

editor/src/clj/editor/editor_extensions.clj
matgis: Sounds good!

editor/test/resources/editor_extensions/resources_io_project/test.editor_script
matgis: I think I prefer the syntax you implemented now that I realize what specifying table keys from string literals looks like.

Maybe it's not weird, and it is just me being unfamiliar with Lua idioms.

matgis: One drawback of the content-by-path-map approach might be that the order we create the files is undefined. Might not matter, though.

Another approach could be to have individual calls to create resources, but being able to run them in a continuation that defers resource-sync until all the changes have been made, similar to how we run transactions. Could be nice if you wanted to mix creations with other changes like deletions and renames. 🤔

editor/src/clj/editor/editor_extensions.clj
vlaaad: Regarding usefulness, it's very convenient, e.g., for creating resources of custom type, e.g. see the trenchfold script:
https://github.com/astrochili/defold-trenchfold/blob/master/trenchfold/trenchfold.lua

I initially wanted to support property tables, but then I realized it's not only too hard to make it work with the editor sync model, but even the way to do it can already be replicated if we don't support setting properties in this function.

Creating files requires resource sync: it's not undoable. So, an API that accepts properties would have to create files, then make a transaction, and then do a save. This could be done anyway, since the script can transact and save.

editor/test/resources/editor_extensions/resources_io_project/test.editor_script
vlaaad: Hm, not sure what could be the use case where the order matters when you want to create multiple files. Like, ensuring that in case of an IO error, some files are created before others? Like a component before a collection that refers to the component? Not sure how important this is...

editor/src/clj/editor/editor_extensions.clj
vlaaad: We still need to access the `evaluation-context` that references the basis to get templates here. Besides, the lua runtime manages this evaluation context and holds a reference to it until the returned future completes — only then does it update the caches. See https://github.com/defold/defold/blob/4b67cd730e998b90a31e2dc7e0241a6781133acc/editor/src/clj/editor/editor_extensions/runtime.clj#L401

To verify that, I added taps in `ext-create-resources`, then inside `io` block, and then finally in a place that updates the caches:
Screenshot 2025-07-09 at 11 32 38
Screenshot 2025-07-09 at 11 33 17
When using the function, it always caches last:
```
tap> [:ext-create-resources 900925723]
tap> [:access-in-io 900925723]
tap> [:update-cache 900925723]
tap> [:ext-create-resources 199188633]
tap> [:access-in-io 199188633]
tap> [:update-cache 199188633]
tap> [:ext-create-resources 2099725489]
tap> [:access-in-io 2099725489]
tap> [:update-cache 2099725489]
tap> [:ext-create-resources 964960844]
tap> [:ext-create-resources 964960844]
tap> [:ext-create-resources 964960844]
tap> [:access-in-io 964960844]
tap> [:update-cache 964960844]
```

vlaaad:
I changed the syntax to creating a resource with custom content to be a tuple of path, content:
```lua
editor.create_resources({
  "/npc.go",
  "/npc.sprite",
  {"/npc.script", "go.property('hp', 100)"}
})
```

editor/src/clj/editor/editor_extensions.clj
matgis: Just like you did for `resource-types`, what I was suggesting is that we add a `workspace/templates-by-type-ext` function that could return a similar map we could use outside the `future/io` block. If possible, we shouldn't hold on to an `evaluation-context` longer than we need to, because others will not see the values produced until they are committed to the cache. I'm sure it works the way you've done it, but it doesn't seem optimal.

matgis: Also, because the template call is in the loop, it will read and sanitize the template every time a resource type is repeated. The `workspace/template` function seems ill-conceived for this sort of thing. Maybe `workspace` should have an `templates-by-type-ext` output? 🤔

vlaaad: Hm, I think this might be too excessive to create a template string for every resource type when we might need only a very small subset of templates, especially given that they are typically going through read + write fns before being returned. I think holding onto an evaluation context and letting the lua runtime handle it once it finishes is the lesser evil here. Especially since this evaluation context is managed by the lua runtime, and there are cases where we want to hold on to the evaluation context that runs long-running, suspendable computations (e.g., lua scripts that look up stuff in the graph, then call `editor.execute(..., {reload_resources=false})`, and then continue looking up stuff — they expect consistent editor state, which requires holding onto evaluation context).

matgis (changes requested):
I still think it's worth it to address the use of `workspace/template` inside the loop.

vlaaad:
The evaluation context won't be released until returned future completes anyway, what's here to address?

vlaaad:
Ah, I see your comment about a cached output, that might work better!

matgis:
Alternatively, if having the templates in an output doesn’t work, you could have a memoized local function of `type-ext->template-string` to avoid the issue of repeat resource types in the loop.

matgis (approved):
LGTM! ❤️

PR #10968, opened 2025-07-15
Allow users to cancel an ongoing build in the editor

### LINKED ISSUE #10725: Add ability to cancel editor build before completion (same as html5)
Sometimes we accidentally press build (cmd+b), but we don't need it. It takes our game around 10 mins to compile a clean build, we wish to have the possibility to abort the build operation without waiting for the build to finish and then closing the window.

Add the possibility to cancel build before build window opens, like we have with HTML5.

Don't really have alternatives

matgis:
You can now cancel the build process in the editor by clicking the Stop button next to the progress indicator in the status bar.

Fixes #10725

### Technical changes
* Simplified the optional arguments to `app-view/build-project!` and `build/build-project!`. Both now take an `opts` map instead.
* Made sure the progress cancellation flag set by `app-view/cancel-task!` is not overwritten by subsequent calls to `app-view/render-task-progress!`.
* Check the cancellation flag at frequent points inside `build/build-project!`.
* Use a record for Progress state instead of a map and fix boxed math.
* Removed unused `progress/mapv` function.
* Fixed boxed math in `editor.ui` module.

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -83,6 +83,7 @@
             [service.smoke-log :as slog]
             [util.coll :as coll :refer [pair]]
             [util.eduction :as e]
+            [util.fn :as fn]
             [util.http-server :as http-server]
             [util.profiler :as profiler]
             [util.thread-util :as thread-util])
@@ -637,11 +638,13 @@
    :fetch-libraries (ref progress/done)
    :download-update (ref progress/done)})
 
+(declare ^:private render-task-progress!)
+
 (defn- cancel-task!
   [task-key]
   (dosync
     (let [progress-ref (task-key app-task-progress)]
-      (ref-set progress-ref (progress/cancel! @progress-ref)))))
+      (render-task-progress! task-key (progress/cancel @progress-ref)))))
 
 (def ^:private app-task-ui-priority
   "Task priority in descending order (from highest to lowest)"
@@ -705,8 +708,27 @@
   (progress/throttle-render-progress
     (fn [progress] (render-task-progress! key progress))))
 
-(defn make-task-cancelled-query [keyword]
-  (fn [] (progress/cancelled? @(keyword app-task-progress))))
+(defn begin-task-progress! [key]
+  (let [progress-ref (get app-task-progress key)
+        prev-progress-atom (atom nil)]
+    (assert (some? progress-ref))
+    (pair
+      (fn render-progress! [progress]
+        ;; Combined throttling and inheritance of cancel state.
+        ;; The first call to render-progress! overwrites the cancel state of the
+        ;; progress-ref, and then subsequent calls will inherit the cancel state
+        ;; from the progress-ref. This is to ensure we see changes to the
+        ;; progress-refs cancel state from the cancel-task! function.
+        (let [prev-progress @prev-progress-atom
+              progress (cond-> progress
+                               prev-progress
+                               (progress/with-inherited-cancel-state @progress-ref))]
+          (when (progress/relevant-change? prev-progress progress)
+            (reset! prev-progress-atom progress)
+            (render-task-progress! key progress))))
+
+      (fn task-cancelled? []
+        (progress/cancelled? @progress-ref)))))
 
 (defn render-main-task-progress! [progress]
   (render-task-progress! :main progress))
@@ -900,19 +922,25 @@
     exception))
 
 (defn- build-project!
-  [project evaluation-context extra-build-targets old-artifact-map render-progress!]
+  [project old-artifact-map opts evaluation-context]
   (let [game-project (project/get-resource-node project "/game.project" evaluation-context)
-        render-progress! (progress/throttle-render-progress render-progress!)]
+        render-progress! (or (:render-progress! opts)
+                             progress/null-render-progress!)]
     (try
       (ui/with-progress [render-progress! render-progress!]
-        (build/build-project! project game-project evaluation-context extra-build-targets old-artifact-map render-progress!))
+        (build/build-project! project game-project old-artifact-map opts evaluation-context))
       (catch Throwable error
         (let [error (if (instance? ExecutionException error)
                       (ex-cause error)
                       error)
-              cause-ex-data (-> error ex-root-cause ex-data)
-              is-cyclic-resource-dependency-error (= :cycle-detected (:ex-type cause-ex-data))]
-          (if is-cyclic-resource-dependency-error
+              cause (ex-root-cause error)
+              cause-ex-data (ex-data cause)
+              ex-type (:ex-type cause-ex-data)]
+          (case ex-type
+            :task-cancelled
+            nil ; We'll just produce an error signaling that the build was cancelled below.
+
+            :cycle-detected
             (ui/run-later
               (dialogs/make-info-dialog
                 {:title "Build Error"
@@ -921,9 +949,15 @@
                  :content {:fx/type fxui/legacy-label
                            :style-class "dialog-content-padding"
                            :text (get-cycle-detected-help-message (-> cause-ex-data :endpoint gt/endpoint-node-id))}}))
+
+            ;; Default case.
             (error-reporting/report-exception! error))
           {:error (cond
-                    is-cyclic-resource-dependency-error
+                    (= :task-cancelled ex-type)
+                    {:severity :fatal
+                     :message (ex-message cause)}
+
+                    (= :cycle-detected ex-type)
                     {:severity :fatal
                      :message (str "Cyclic resource dependency detected. " (get-cycle-detected-help-message (-> cause-ex-data :endpoint gt/endpoint-node-id)))}
 
@@ -969,17 +1003,20 @@
     :run-build-hooks     optional flag that indicates whether to run pre- and
                          post-build hooks
     :render-progress!    optional progress reporter fn
+    :task-cancelled?     optional fn that will be called periodically to check
+                         if the user has cancelled the build process.
     :old-artifact-map    optional old artifact map with previous build results
                          to speed up the build process"
   [project & {:keys [;; required
                      result-fn
                      prefs
                      ;; optional
-                     debug build-engine run-build-hooks render-progress! old-artifact-map lint]
+                     debug build-engine run-build-hooks render-progress! task-cancelled? old-artifact-map lint]
               :or {debug false
                    build-engine true
                    run-build-hooks true
                    render-progress! progress/null-render-progress!
+                   task-cancelled? fn/constantly-false
                    old-artifact-map {}}}]
   {:pre [(ifn? result-fn)
          (or (not build-engine) (some? prefs))]}
@@ -1145,8 +1182,13 @@
               (fn run-project-build-on-background-thread! []
                 (let [extra-build-targets
                       (when debug
-                        (debug-view/build-targets project evaluation-context))]
-                  (build-project! project evaluation-context extra-build-targets old-artifact-map render-progress!)))
+                        (debug-view/build-targets project evaluation-context))
+
+                      opts
+                      {:extra-build-targets extra-build-targets
+                       :render-progress! render-progress!
+                       :task-cancelled? task-cancelled?}]
+                  (build-project! project old-artifact-map opts evaluation-context)))
               (fn process-project-build-results-on-ui-thread! [project-build-results]
                 (project/update-system-cache-build-targets! evaluation-context)
                 (project/log-cache-info! (g/cache) "Cached compiled build targets in system cache.")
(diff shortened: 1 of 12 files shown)


editor/src/clj/editor/app_view.clj
```diff
@@ -637,11 +638,13 @@
...
    :download-update (ref progress/done)})
 
+(declare ^:private render-task-progress!)
+
 (defn- cancel-task!
   [task-key]
   (dosync
     (let [progress-ref (task-key app-task-progress)]
-      (ref-set progress-ref (progress/cancel! @progress-ref)))))
+      (render-task-progress! task-key (progress/cancel @progress-ref)))))
```
matgis, line 647: Calling `render-task-progress!` ensures the UI is refreshed immediately after cancelling.

editor/src/clj/editor/app_view.clj
```diff
@@ -705,8 +708,27 @@
...
+        (let [prev-progress @prev-progress-atom
+              progress (cond-> progress
+                               prev-progress
+                               (progress/with-inherited-cancel-state @progress-ref))]
+          (when (progress/relevant-change? prev-progress progress)
+            (reset! prev-progress-atom progress)
+            (render-task-progress! key progress))))
+
+      (fn task-cancelled? []
+        (progress/cancelled? @progress-ref)))))
```
matgis, line 731: The progress reporting is quite difficult to work with and feels like it gets in the way sometimes. 😞 Here, I needed to make sure the cancel flag is not overwritten by subsequent calls to `render-progress!`, while also not inheriting the cancelled state from a prior cancelled task. This feels stateful and messy, and we might want to take another stab at it at some point.

editor/src/clj/editor/app_view.clj
```diff
@@ -900,19 +922,25 @@
     exception))
 
 (defn- build-project!
-  [project evaluation-context extra-build-targets old-artifact-map render-progress!]
+  [project old-artifact-map opts evaluation-context]
```
matgis, line 925: Instead of adding yet another optional `task-cancelled?` argument, I decided to put the optional stuff in an `opts` map here and in `build/build-project!`. While doing this, I also moved the `evaluation-context` argument to the end, as is our convention.

editor/src/clj/editor/app_view.clj
```diff
@@ -1292,19 +1338,21 @@ If you do not specifically require different script states, consider changing th
...
+                  :lint false
+                  :render-progress! render-progress!
+                  :task-cancelled? task-cancelled?
+                  :old-artifact-map (workspace/artifact-map workspace)
+                  :prefs prefs
+                  :result-fn (fn [build-results]
+                               (when (handle-build-results! workspace render-build-error! build-results)
+                                 (let [target (targets/selected-target prefs)]
+                                   (when (targets/controllable-target? target)
+                                     (debug-view/attach! debug-view project target (:artifacts build-results)))))))))
```
matgis, line 1355: Mostly indentation, other than adding `:task-cancelled?`.

editor/src/clj/editor/build.clj
```diff
@@ -42,11 +43,6 @@
     (when-some [resource-path (node-id->resource-path node-id)]
       (str "Compiling " resource-path))))
 
-(defn make-collect-progress-steps-tracer [watched-label steps-atom]
-  (fn [state node output-type label]
-    (when (and (= label watched-label) (= state :begin) (= output-type :output))
-      (swap! steps-atom conj node))))
-
```
matgis, line 49: This was inlined at the only call site so we can check the cancellation flag in-between steps.

editor/src/clj/editor/build.clj
```diff
@@ -165,15 +161,43 @@
...
+        supplied-render-progress! (or (:render-progress! opts)
+                                      progress/null-render-progress!)
+
+        render-progress!
+        (if (= fn/constantly-false task-cancelled?)
+          supplied-render-progress!
+          (fn render-progress! [progress]
+            (if (task-cancelled?)
+              (throw-build-cancelled-exception!)
+              (supplied-render-progress! progress))))
```
matgis, line 181: I wrap the supplied `render-progress!` here so it will throw a `:task-cancelled` exception if the cancellation flag has been set. It's a convenient way to intersperse cancellation flag checks in most parts of the build process.

editor/src/clj/editor/build.clj
```diff
@@ -165,15 +161,43 @@
...
+          (cond
+            (task-cancelled?)
+            (throw-build-cancelled-exception!)
+
+            (and (= :build-targets label)
+                 (= :begin state)
+                 (= :output output-type))
+            (swap! steps conj node)))
+
+        _ (g/node-value node-id :build-targets (assoc evaluation-context :dry-run true :tracer collect-tracer-fn))
```
matgis, line 196: This is a preflight step that figures out all the graph outputs we'll need to evaluate to build the project. This in itself can potentially take a little while in large projects, so I intersperse cancellation flag checks here too.

editor/src/clj/editor/build.clj
```diff
@@ -165,15 +161,43 @@
...
         node-id->resource-path (set/map-invert (g/node-value project :nodes-by-resource-path evaluation-context))
         progress-message-fn (partial compiling-progress-message node-id->resource-path)
         step-count (count @steps)
-        progress-tracer (project/make-progress-tracer :build-targets step-count progress-message-fn (progress/nest-render-progress render-progress! (progress/make "" 10) 5))
+        progress-tracer (project/make-progress-tracer :build-targets step-count progress-message-fn (progress/nest-render-progress render-progress! (progress/make "" 10 0 true) 5))
```
matgis, line 200: Marked parent task as cancellable.

editor/src/clj/editor/library.clj
```diff
@@ -287,7 +287,7 @@
   * :new-file to downloaded file if any
   * :tag with etag from resolver"
   [resolver render-progress! lib-states]
-  (progress/mapv
+  (progress/progress-mapv
```
matgis, line 290: Unrelated bug fix: This arity of `progress/mapv` was actually calling progress/progress-mapv by mistake. We didn't have any calls to the "correct" arity of `progress/mapv`, so I removed it.

editor/src/clj/editor/progress.clj
```diff
@@ -13,89 +13,138 @@
...
+            [util.defonce :as defonce]))
+
+(set! *warn-on-reflection* true)
+(set! *unchecked-math* :warn-on-boxed)
+
+(defonce/record Progress
+  [^String message
+   ^long size
+   ^long pos
+   cancel-state])
```
matgis, line 26: I decided to change progress to use a record so I could fix all the boxed math issues. We're surprisingly spammy with progress reporting - only a few of them actually make it into UI updates. I also merged the `:cancellable` and `:cancelled` booleans into a three-state `:cancel-state`.

editor/src/clj/editor/progress.clj
```diff
@@ -13,89 +13,138 @@
...
 
-(defn- relevant-change? [last-progress progress]
+(defn relevant-change? [^Progress last-progress ^Progress progress]
   (or (nil? last-progress)
-      (not= (message last-progress)
-            (message progress))
+      (not= (.-cancel-state last-progress)
+            (.-cancel-state progress))
+      (not= (.-message last-progress)
+            (.-message progress))
```
matgis, line 131: The `relevant-change?` function is used to check if a UI refresh is warranted. Since we're so spammy with progress updates, it can take a lot of progress to increase the completion percentage one tick.

editor/src/clj/editor/progress.clj
```diff
@@ -13,89 +13,138 @@
...
     (println (message progress) percentage "%")
     (println (message progress))))
 
 (defn throttle-render-progress [render-progress!]
   (let [last-progress (atom nil)]
-    (fn [progress]
+    (fn throttled-render-progress! [^Progress progress]
       (when (relevant-change? @last-progress progress)
-        (swap! last-progress progress)
+        (reset! last-progress progress)
```
matgis, line 146: Whoops! Converting Progress to a record found a bug! Previous code would look up the `progress` as a key in the `last-progress` map, presumably resulting in `nil` all the time. 🤔

editor/src/clj/editor/progress.clj
```diff
@@ -125,48 +174,37 @@
...
+       (assert (<= (+ parent-pos span) parent-size))
+       (fn nested-render-progress! [^Progress progress]
+         (let [size (.-size progress)
+               pos (.-pos progress)]
+           (render-progress!
+             (->progress
+               (message progress)
+               (* size parent-size)
+               (+ (* size parent-pos) (* pos span))
+               (inherited-cancel-state progress parent-cancellable parent-cancelled)))))))))
```
matgis, line 194: Updated `nest-render-progress!` to inherit the cancellation state of the `parent-progress`.

Render progress nesting is also one of those things I feel could benefit from a redesign at some point. It's a bit awkward and confusing to use. 😞

editor/src/clj/editor/progress.clj
```diff
@@ -125,48 +174,37 @@
...
-  [render-progress! message-fn]
-  (fn [f coll] (mapv f coll render-progress! message-fn)))
+       (reduce (fn [[result progress] item]
+                 (let [progress (with-message progress (or (message-fn item) ""))]
+                   (render-progress! progress)
+                   (pair (conj! result (f item progress))
+                         (advance progress))))
+               (pair (transient [])
+                     (->progress "" (count coll) 0 :not-cancellable))
+               coll)))))
```
matgis, line 209: The diff here is a bit weird since both these functions had very similar implementations. To be clear, `progress/mapv` was removed. I also updated `progress/progress-mapv` to use a transient vector.

editor/src/clj/editor/ui.clj
```diff
@@ -2227,10 +2239,10 @@
...
   (let [prev-exception-handler (Thread/getDefaultUncaughtExceptionHandler)
         thrown-exception (volatile! nil)]
-    (Thread/setDefaultUncaughtExceptionHandler (reify Thread$UncaughtExceptionHandler
-                                                 (uncaughtException [_ _ exception]
-                                                   (vreset! thrown-exception exception)
-                                                   (close! stage))))
+    (Thread/setDefaultUncaughtExceptionHandler
+      (fn [_ exception]
+        (vreset! thrown-exception exception)
+        (close! stage)))
```
matgis, line 2245: Modern Clojure lets you use functions for this sort of thing.

editor/styling/stylesheets/_editor.scss
```diff
@@ -150,11 +150,11 @@ Links:
...
     -fx-content-display: center;
     &:hover {
-        -fx-background-color: -df-component-lighter;
-        -cross-color: -df-text-selected;
+        -fx-background-color: -df-background-lighter;
+        -cross-color: -df-text;
     }
     &:armed {
-        -fx-background-color: -df-background-light;
+        -fx-background-color: -df-background-lighter;
```
matgis, line 157: Updated the styling of the Stop button next to the progress indicator in the status bar so it is less intense. It felt a bit out of place at the previous intensity.

editor/src/clj/editor/progress.clj
```diff
@@ -13,89 +13,137 @@
...
+   ^long pos
+   cancel-state])
+
+(defn ->progress
+  ^Progress [^String message ^long size ^long pos cancel-state]
+  {:pre [(string? message)
+         (nat-int? size) ; size 0 means indeterminate.
+         (nat-int? pos)
+         (<= pos size)
+         (case cancel-state (:not-cancellable :cancellable :cancelled) true false)]}
```
sprocketc, line 34: I think I have seen this pattern before on our codebase. Is there a reason to avoid `contains?` in this case, or is it just a matter of preference?

sprocketc (approved):
Works great!

editor/src/clj/editor/progress.clj
matgis: When measured, it looks like it is around 40% faster than using `contains?`, probably because it can be compiled into a jump table rather than doing hash lookups. Although I do agree it doesn't read that well.

Curiously, the common Clojure idiom `(#{:not-cancellable :cancellable :cancelled} keyword)` does not appear to suffer from as bad of a performance hit as using `contains?`. Here, `case` is less than 10% faster. Maybe the compiler is clever enough to compile it to a `case` statement under the hood and we're only seeing the performance hit of converting the matching value into a boolean? 🤔 It should be noted that this optimization only works when the set is inlined. And using `contains?` with an inlined set doesn't trigger the optimization either.

Anyway, I think using `case` might be worth it in hot paths, but an inlined set lookup is probably sufficient and clearer in most cases.

PR #10972, opened 2025-07-16
Bob performance optimizations for sequential builds

### LINKED ISSUE #10854: Optimize Bob for faster iterative builds
**Task (REQUIRED):**
Currently, iterative Bob builds (without `distclean`) on a small project may take 5-6 seconds.
This time can be reduced at least twice by optimizing Java classes scanner and engine hash (how we calculate if engine should be rebuild on extender or not):

Image

https://github.com/defold/defold/issues/8490#issuecomment-2981003910

Faster iterative builds using bob

AGulev:
This fix speeds up Bob.jar initialization, which is especially useful for sequential builds without many changes (e.g. Cmd+M to Build HTML5 from the editor, or bundling).  
* About 4 times faster initialization. In the test setup, it's 400 ms vs 1600 ms (for a project with a large file count, it's 3.5 s vs 20 s)  
* Speed-up of the task creation stage for a small project in the test setup: 900 ms vs 2000 ms (for a large project, about 5%)

Fix https://github.com/defold/defold/issues/10854

# Technical details

## `findResourcePaths`
Reading through the code, `findResourcePaths` appears suspicious.  
A dedicated scope in the build time profiler, highlighted in green, shows the usage of this function and the time it takes.

CleanShot 2025-07-15 at 14 34 52@2x
Profiler shows that `findResourcePaths` is called multiple times, each time scanning the file system.  
The `findResourcePaths()` method internally calls `findResourcePathsByExtension()`, where a file extension can be specified.  
In this test project, that function is called 7+ times, resulting in at least 20 full scan operations during a build (depending on project setup).

Given that `bob.jar` doesn’t modify the project folder during the build, it makes sense to scan the disk once and reuse the results.  
A cache was implemented at the `findResourcePathsByExtension()` level to avoid redundant scans.

**Result**:  
This change significantly speeds up the build process — about 5x faster for the test project.

CleanShot 2025-07-15 at 17 09 06@2x

## doScan
Further investigation shows that our Java class scanner calls all the static initializers, which isn't necessary at this stage.  
Excluding such classes from the scan stage helped reduce `doScan` time by 2x — from about 1s to 300–500 ms.  
Excluding the `com.dynamo.bob.pipeline.antlr` namespace further reduced scan time to about 160–230 ms.

## Ogg Validation
The Defold build pipeline has two main stages: task creation and task building.  
Profiling showed anomalously long creation times for `.ogg` files. Investigation revealed that `.ogg` validation occurred during the task creation stage, which is unnecessary.  
Moving validation to the task building stage saved about 500 ms during creation — important for sequential builds.

## Various Optimizations
- By default, we have a fairly large gamepads file that doesn't reference other resources.  
  Calling `createSubTasks` on it was a waste of time. Overriding the default `create` method saved about 120 ms on each iterative build.
- Lua parser is one of the heaviest operations during task creation.  
  Small improvements in the parser reduced parsing time by about 10%.

**Result**:  
All these improvements helped reduce sequential build time on a test project from **3.6s to 1.4s** (on average).

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
@@ -159,6 +159,7 @@ public class Bob {
     }
 
     public static void initLua() {
+        init();
         PackedResources.unpackAllLibsAsync(Platform.getHostPlatform());
         PackedResources.waitForuUpackAllLibsAsync();
     }
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
@@ -643,6 +643,7 @@ public class Bob {
     }
 
     private static void setupProject(Project project, boolean resolveLibraries, IProgress progress) throws IOException, LibraryException, CompileExceptionError {
+        project.loadProjectFile(false);
         BobProjectProperties projectProperties = project.getProjectProperties();
         String[] dependencies = projectProperties.getStringArrayValue("project", "dependencies");
         List<URL> libUrls = new ArrayList<>();
@@ -820,10 +821,6 @@ public class Bob {
                     project.addBuildServerHeader(header);
                 }
             }
-            TimeProfiler.stop();
-
-            TimeProfiler.start("loadProjectFile");
-            project.loadProjectFile();
             TimeProfiler.stop();
 
             TimeProfiler.start("setupProject");
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
@@ -2262,7 +2262,7 @@ public class Project {
         TimeProfiler.start("findResourcePathsByExtension");
         TimeProfiler.addData("path", _path);
         TimeProfiler.addData("ext", ext);
-
+        _path = ".".equals(_path) ? "" : _path;
         final String path = Project.stripLeadingSlash(_path);
 
         // Initialize and cache all paths only once
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
@@ -2267,7 +2267,7 @@ public class Project {
         TimeProfiler.start("findResourcePathsByExtension");
         TimeProfiler.addData("path", _path);
         TimeProfiler.addData("ext", ext);
-        _path = ".".equals(_path) ? "" : _path;
+        _path = FilenameUtils.getPath(_path);
         final String path = Project.stripLeadingSlash(_path);
 
         // Initialize and cache all paths only once
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
@@ -357,7 +357,6 @@ public class Project {
     private void doScan(IClassScanner scanner, Set<String> classNames) {
         TimeProfiler.start("doScan");
         boolean is_bob_light = getManifestInfo("is-bob-light") != null;
-        TimeProfiler.start("filter out classes");
         List<String> filteredClassNames = classNames.stream()
                 .filter(className ->
                         // classes with static initializers we don't want to initialize on this stage
@@ -369,10 +368,7 @@ public class Project {
                         !(is_bob_light && className.startsWith("com.dynamo.bob.pipeline.ExtenderUtil")) &&
                         !(is_bob_light && className.startsWith("com.dynamo.bob.bundle.BundleHelper")))
                 .collect(Collectors.toList());
-        TimeProfiler.stop();
         for (String className : filteredClassNames) {
-            TimeProfiler.start("doScan.builder_registration");
-            TimeProfiler.addData("className", className);
             try {
                 Class<?> klass = Class.forName(className, true, scanner.getClassLoader());
                 BuilderParams builderParams = klass.getAnnotation(BuilderParams.class);
@@ -395,38 +391,29 @@ public class Project {
                     }
                 }
 
-                TimeProfiler.stop();
-                TimeProfiler.start("doScan.bundler_registration");
                 if (IBundler.class.isAssignableFrom(klass)) {
                     if (!klass.equals(IBundler.class)) {
                         bundlerClasses.add((Class<? extends IBundler>) klass);
                     }
                 }
 
-                TimeProfiler.stop();
-                TimeProfiler.start("doScan.shader_registration");
                 if (IShaderCompiler.class.isAssignableFrom(klass)) {
                     if (!klass.equals(IShaderCompiler.class)) {
                         shaderCompilerClasses.add((Class<? extends IShaderCompiler>) klass);
                     }
                 }
 
-                TimeProfiler.stop();
-                TimeProfiler.start("doScan.texture_registration");
                 if (ITextureCompressor.class.isAssignableFrom(klass)) {
                     if (!klass.equals(ITextureCompressor.class)) {
                         textureCompressorClasses.add((Class<? extends ITextureCompressor>) klass);
                     }
                 }
 
-                TimeProfiler.stop();
-                TimeProfiler.start("doScan.plugin_registration");
                 if (IPlugin.class.isAssignableFrom(klass)) {
                     if (!klass.equals(IPlugin.class)) {
                         pluginClasses.add((Class<? extends IPlugin>) klass);
                     }
                 }
-                TimeProfiler.stop();
             } catch (Exception e) {
                 throw new RuntimeException(e);
             }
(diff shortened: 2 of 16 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -159,6 +159,7 @@ public static void init() {
     }
 
     public static void initLua() {
+        init();
```
AGulev, line 162: Class scanner doesn't unpack anything anymore, so we should initialize it here otherwise files will be unpacked into a wrong folder

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -643,6 +644,7 @@ else if (severity == MultipleCompileException.Info.SEVERITY_WARNING)
     }
 
     private static void setupProject(Project project, boolean resolveLibraries, IProgress progress) throws IOException, LibraryException, CompileExceptionError {
+        project.loadProjectFile(false);
```
AGulev, line 647: We do it twice.
And the first time we need it for dependencies to be able to download them.
So, this extra parameter helps to reduce scope and avoid unnecessary operations and avoid paths caching until we have dependencies downloaded.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -350,71 +357,72 @@ private static String getManifestInfo(String attribute) {
...
+                        !className.startsWith("com.dynamo.bob.pipeline.TexcLibrary") &&
+                        !className.startsWith("com.dynamo.bob.pipeline.Shaderc") &&
+                        !className.startsWith("com.dynamo.bob.pipeline.ModelImporter") &&
+                        // namespaces we don't need to scan
+                        !className.startsWith("com.dynamo.bob.pipeline.antlr") &&
+                        // classes we don't need to bob light
+                        !(is_bob_light && className.startsWith("com.dynamo.bob.archive.publisher.AWSPublisher")) &&
+                        !(is_bob_light && className.startsWith("com.dynamo.bob.pipeline.ExtenderUtil")) &&
+                        !(is_bob_light && className.startsWith("com.dynamo.bob.bundle.BundleHelper")))
+                .collect(Collectors.toList());
```
AGulev, line 372: Filter out all the classes we are not going to use and iterate only needed. It makes it easier to read code.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -350,71 +357,72 @@ private static String getManifestInfo(String attribute) {
...
-                                Class<?> inputClass = protoParams.srcClass();
-                                if (inputClass != null) {
-                                    ProtoBuilder.addMessageClass(ext, protoParams.srcClass());
-                                }
+        List<String> filteredClassNames = classNames.stream()
+                .filter(className ->
+                        // classes with static initializers we don't want to initialize on this stage
+                        !className.startsWith("com.dynamo.bob.pipeline.TexcLibrary") &&
+                        !className.startsWith("com.dynamo.bob.pipeline.Shaderc") &&
+                        !className.startsWith("com.dynamo.bob.pipeline.ModelImporter") &&
```
AGulev, line 365: all these classes used to be filter out because they unpack libs in static initializer (see `com.dynamo.bob.TexcLibrary` in old code), but then we moved `TexcLibrary` and added a couple more such classes with libs, so I filter out them all.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -350,71 +357,72 @@ private static String getManifestInfo(String attribute) {
...
+                        // namespaces we don't need to scan
+                        !className.startsWith("com.dynamo.bob.pipeline.antlr") &&
```
AGulev, line 367: just a "heavy" namespace where we don't expect anything we need

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -2255,29 +2269,40 @@ else if (availableProcessors > 1) {
...
-                    shouldAdd = path.endsWith(ext);
+        // Initialize and cache all paths only once
+        if (allResourcePathsCache == null) {
+            List<String> allPaths = new ArrayList<>();
+            fileSystem.walk("", new FileSystemWalker() {
+                public void handleFile(String filePath, Collection<String> results) {
+                    results.add(FilenameUtils.normalize(filePath, true));
                 }
+            }, allPaths);
+            allResourcePathsCache = allPaths;
```
AGulev, line 2289: Always read all files and cache them . All the other operations works only with cached strings

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -2255,29 +2269,40 @@ else if (availableProcessors > 1) {
...
+            result.addAll(allResourcePathsCache);
+            TimeProfiler.stop();
+            return;
+        }
+
+        // Filter the cached list
+        allResourcePathsCache.parallelStream()
+            .filter(p -> (ext == null || p.endsWith(ext)) &&
+                         (path == null || path.isEmpty() || p.startsWith(path)))
+            .forEach(result::add);
```
AGulev, line 2303: multithreaded filter by path or/and extension

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LuaScanner.java
```diff
@@ -113,7 +114,8 @@ public int getLineNumber() {
     private CommonTokenStream tokenStream = null;
     private TokenStreamRewriter rewriter;
```
AGulev, line 115: All these changes are micro-optimisations mostly, but taking into account that this is the heaviest operation during creation phase it gives boost

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/OggBuilder.java
```diff
@@ -37,34 +37,33 @@ public class OggBuilder extends CopyBuilder{
 
     @Override
     public Task create(IResource input) throws IOException, CompileExceptionError {
+        oggzValidateExePath = Bob.getHostExeOnce("oggz-validate", oggzValidateExePath);
+        return super.create(input);
```
AGulev, line 41: Validation moved to Build stage vut it requires `CompileExceptionError` for all CopyBuilders

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/OpusBuilder.java
```diff
@@ -31,7 +31,7 @@
 import com.dynamo.bob.Task;
 import com.dynamo.bob.fs.IResource;
 
-@BuilderParams(name = "Opus", inExts = ".opus", outExt = ".opusc")
+@BuilderParams(name = "Opus", inExts = ".opus", outExt = ".opusc", paramsForSignature = {"sound-stream-enabled"})
```
AGulev, line 34: `sound-stream-enabled` should invalidate cache

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ProtoBuilders.java
```diff
@@ -172,7 +172,16 @@ public static class InputBindingBuilder extends ProtoBuilder<InputBinding.Builde
...
-    public static class GamepadMapsBuilder extends ProtoBuilder<GamepadMaps.Builder> {}
+    public static class GamepadMapsBuilder extends ProtoBuilder<GamepadMaps.Builder> {
+        @Override
+        public Task create(IResource input) throws IOException, CompileExceptionError {
+            Task.TaskBuilder taskBuilder = Task.newBuilder(this)
+                    .setName(params.name())
+                    .addInput(input)
+                    .addOutput(input.changeExt(params.outExt()));
+            return taskBuilder.build();
+        }
```
AGulev, line 183: it's ne file, but it has a lot of fields in it. Calling `createSubTasks` iterate over all the fields checking if it's resounce or not. 
We know that there are no resources in this file, so we can skip it and save this time (on my machine about 120ms)

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/BobProjectProperties.java
```diff
@@ -537,7 +537,7 @@ private Map<String, Map<String, ProjectProperty>> doLoad(InputStream in, boolean
         int cursor = 0;
         String line = reader.readLine();
         while (line != null) {
-            line.trim();
+            line = line.trim();
```
AGulev, line 540: just missed bug

AGulev:
One more comparison from a real project on another machine (sequential build):
5 seconds before (you may see that it takes a lot of time to scan disk to detect that native extensions don't need to be rebuilt)
CleanShot 2025-07-16 at 22 41 21@2x
1 second one
CleanShot 2025-07-16 at 22 42 13@2x

britzl (approved):

PR #10983, opened 2025-07-18
Improve responsiveness when typing a filter term in the Asset Picker

### LINKED ISSUE #10739: Faster search using search dialog (cmd+p)
In our project, which is very big and very heavy, it takes a long time to search for resources with cmd+p option. Sometimes it takes several seconds and it can accumulate to minutes.

We'd like to hasten the search function and make it respond faster

The only alternative is to search the project files and not use the dialog, and sometimes it's faster

https://github.com/user-attachments/assets/f085dc19-250a-4d34-ae9f-e2656e7c8568

matgis:
Filterable item list dialogs like the Open Asset dialog will now remain responsive while the list is filtered. A progress indicator in the bottom-left of the dialog indicates that we're still awaiting results from the search.

Fixes #10739

### DIFF
--- a/editor/src/clj/editor/dialogs.clj
+++ b/editor/src/clj/editor/dialogs.clj
@@ -23,6 +23,7 @@
             [cljfx.fx.list-cell :as fx.list-cell]
             [cljfx.fx.list-view :as fx.list-view]
             [cljfx.fx.progress-bar :as fx.progress-bar]
+            [cljfx.fx.progress-indicator :as fx.progress-indicator]
             [cljfx.fx.region :as fx.region]
             [cljfx.fx.scene :as fx.scene]
             [cljfx.fx.v-box :as fx.v-box]
@@ -38,7 +39,9 @@
             [editor.progress :as progress]
             [editor.ui :as ui]
             [editor.util :as util]
-            [service.log :as log])
+            [service.log :as log]
+            [util.coll :as coll]
+            [util.thread-util :as thread-util])
   (:import [clojure.lang Named]
            [java.io File]
            [java.nio.file Path Paths]
@@ -505,10 +508,13 @@
       (.showDialog owner-window)))
 
 (defn- default-filter-fn [filter-on text items]
-  (let [text (string/lower-case text)]
-    (filterv (fn [item]
-               (string/starts-with? (string/lower-case (filter-on item)) text))
-             items)))
+  (if (coll/empty? text)
+    items
+    (let [text (string/lower-case text)]
+      (filterv (fn [item]
+                 (thread-util/throw-if-interrupted!)
+                 (string/starts-with? (string/lower-case (filter-on item)) text))
+               items))))
 
 (def ext-with-identity-items-props
   (fx/make-ext-with-props
@@ -531,7 +537,7 @@
                               (.put properties "isDefaultAnchor" true)))))))
 
 (defn- select-list-dialog
-  [{:keys [filter-term filtered-items title ok-label prompt cell-fn selection owner]
+  [{:keys [filter-term filtered-items filter-in-progress title ok-label prompt cell-fn selection owner]
     :as props}]
   {:fx/type dialog-stage
    :title title
@@ -555,20 +561,28 @@
                                   :fixed-cell-size 27
                                   :cell-factory {:fx/cell-type fx.list-cell/lifecycle
                                                  :describe cell-fn}}}}}
-   :footer {:fx/type dialog-buttons
-            :children [{:fx/type fxui/button
-                        :text ok-label
-                        :variant :primary
-                        :disable (zero? (count filtered-items))
-                        :on-action {:event-type :confirm}
-                        :default-button true}]}})
+   :footer {:fx/type fx.h-box/lifecycle
+            :alignment :center-left
+            :children [{:fx/type fx.progress-indicator/lifecycle
+                        :h-box/hgrow :never
+                        :pref-width 32.0
+                        :pref-height 32.0
+                        :visible filter-in-progress}
+                       {:fx/type dialog-buttons
+                        :h-box/hgrow :always
+                        :children [{:fx/type fxui/button
+                                    :text ok-label
+                                    :variant :primary
+                                    :disable (zero? (count filtered-items))
+                                    :on-action {:event-type :confirm}
+                                    :default-button true}]}]}})
 
 (defn- wrap-cell-fn [f]
   (fn [item]
     (when (some? item)
       (assoc (f item) :on-mouse-clicked {:event-type :select-item-on-double-click}))))
 
-(defn- select-list-dialog-event-handler [filter-fn items]
+(defn- select-list-dialog-event-handler [set-filter-term-fn]
   (fn [state event]
     (case (:event-type event)
       :cancel (assoc state ::fxui/result nil)
@@ -620,12 +634,8 @@
                                   KeyCode/ESCAPE (recur state (assoc event :event-type :cancel))
                                   state))
                               state))
-      :set-filter-term (let [term (:fx/event event)
-                             filtered-items (vec (filter-fn term items))]
-                         (assoc state
-                           :filter-term term
-                           :filtered-items filtered-items
-                           :selected-indices (if (seq filtered-items) [0] []))))))
+      :set-filter-term (let [filter-term (:fx/event event)]
+                         (set-filter-term-fn state filter-term)))))
 
 (defn make-select-list-dialog
   "Show dialog that allows the user to select one or many of the suggested items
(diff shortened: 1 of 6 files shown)


editor/src/clj/editor/dialogs.clj
```diff
@@ -505,10 +508,13 @@
...
-    (filterv (fn [item]
-               (string/starts-with? (string/lower-case (filter-on item)) text))
-             items)))
+  (if (coll/empty? text)
+    items
+    (let [text (string/lower-case text)]
+      (filterv (fn [item]
+                 (thread-util/throw-if-interrupted!)
+                 (string/starts-with? (string/lower-case (filter-on item)) text))
+               items))))
```
matgis, line 517: This now runs in a `future` in the thread pool. If the user updates the filter term by typing into the filter field, we cancel the future. This flags the running thread as interrupted. At convenient points, we check this flag and will throw an `InterruptedException` to terminate the future. The `InterruptedException` is caught at root-level of the future performing the filtering.

editor/src/clj/editor/dialogs.clj
```diff
@@ -531,7 +537,7 @@
                               (.put properties "isDefaultAnchor" true)))))))
 
 (defn- select-list-dialog
-  [{:keys [filter-term filtered-items title ok-label prompt cell-fn selection owner]
+  [{:keys [filter-term filtered-items filter-in-progress title ok-label prompt cell-fn selection owner]
```
matgis, line 540: Added `filter-in-progress` boolean to control if the progress indicator is visible or not.

editor/src/clj/editor/dialogs.clj
```diff
@@ -555,20 +561,28 @@
...
+                        :pref-height 32.0
+                        :visible filter-in-progress}
+                       {:fx/type dialog-buttons
+                        :h-box/hgrow :always
+                        :children [{:fx/type fxui/button
+                                    :text ok-label
+                                    :variant :primary
+                                    :disable (zero? (count filtered-items))
+                                    :on-action {:event-type :confirm}
+                                    :default-button true}]}]}})
```
matgis, line 578: Buttons are unchanged. They are wrapped in another h-box so we can put the progress indicator to the left.

editor/src/clj/editor/dialogs.clj
```diff
@@ -620,12 +634,8 @@
...
                                   state))
                               state))
-      :set-filter-term (let [term (:fx/event event)
-                             filtered-items (vec (filter-fn term items))]
-                         (assoc state
-                           :filter-term term
-                           :filtered-items filtered-items
-                           :selected-indices (if (seq filtered-items) [0] []))))))
+      :set-filter-term (let [filter-term (:fx/event event)]
+                         (set-filter-term-fn state filter-term)))))
```
matgis, line 638: Changed this to a callback fn so we can do more stuff outside of this.

editor/src/clj/editor/dialogs.clj
```diff
@@ -660,19 +670,52 @@
...
+                                           :filter-in-progress false
+                                           :filtered-items filtered-items
+                                           :selected-indices selected-indices))))
+                              (catch InterruptedException _
+                                nil)
+                              (catch Throwable error
+                                (error-reporting/report-exception! error))))))
+                 (assoc state
+                   :filter-term filter-term
+                   :filter-in-progress true))))
```
matgis, line 704: Typing into the filter-term field will call this function for each keypress. The filtering is performed in a `future` running in the thread pool. We keep the `future` in an `atom` so we can cancel and replace it when the term changes. Once it completes, it will update the `:filtered-items` and so on in the `state-atom` observed by the view. It is also here that we catch the `InterruptedException` thrown elsewhere if we detect that the thread has been interrupted, which will exit the `future` cleanly without doing anything.

editor/src/clj/editor/dialogs.clj
```diff
@@ -660,19 +670,52 @@
...
+
+         _ (swap! state-atom set-filter-term filter-term)
+         event-handler (select-list-dialog-event-handler set-filter-term)
          result (fxui/show-dialog-and-await-result!
-                  :initial-state {:filter-term filter-term
-                                  :filtered-items initial-filtered-items
-                                  :selected-indices (if (seq initial-filtered-items) [0] [])}
-                  :event-handler (select-list-dialog-event-handler filter-fn items)
+                  :state-atom state-atom
+                  :event-handler event-handler
```
matgis, line 710: The `:initial-state` has been moved into the `state-atom`, which we need access to from the outside so that the completed filter `future` can set the `:filtered-items` and so on when it completes.

editor/src/clj/editor/dialogs.clj
```diff
@@ -660,19 +670,52 @@
...
                   :description {:fx/type select-list-dialog
                                 :title (:title options "Select Item")
                                 :ok-label (:ok-label options "OK")
                                 :prompt (:prompt options "Type to filter")
                                 :cell-fn cell-fn
                                 :owner (or (:owner options) (ui/main-stage))
                                 :selection (:selection options :single)})]
+     (swap! filter-future-atom thread-util/cancel-future!)
```
matgis, line 718: The call to `fxui/show-dialog-and-await-result!` immediately before this blocks until the dialog is closed. At that point, we want to cancel any pending filtering.

editor/src/clj/editor/fxui.clj
```diff
@@ -385,13 +385,17 @@
...
     :error-handler    optional, 1-arg Throwable handler, by default it shows an
                       error dialog and reports the exception to sentry"
-  [& {:keys [initial-state event-handler description error-handler]
-      :or {initial-state {}
-           error-handler error-reporting/report-exception!}}]
-  (let [state-atom (atom initial-state)
+  [& {:keys [state-atom initial-state event-handler description error-handler]
+      :or {error-handler error-reporting/report-exception!}}]
+  (let [state-atom (or state-atom
+                       (atom (or initial-state {})))
```
matgis, line 398: Added the ability to supply a `:state-atom` from the outside instead of this function creating an internal one from the `:initial-state`.

editor/src/clj/editor/resource_dialog.clj
```diff
@@ -25,19 +25,22 @@
...
+            [util.fn :as fn]
+            [util.thread-util :as thread-util]))
 
 (def ^:private fuzzy-resource-filter-fn (partial fuzzy-choices/filter-options resource/proj-path resource/proj-path))
 
 (defn- resource-node-ids->project-resources [basis resource-node-ids]
   (let [project-resources
         (into []
               (comp
+                (map thread-util/abortable-identity!)
```
matgis, line 37: This throws an `InterruptedException` if the thread we're running on has been interrupted by someone cancelling the `future`, otherwise returns the value unaltered.

editor/src/clj/editor/resource_dialog.clj
```diff
@@ -25,19 +25,22 @@
...
                 (map #(resource-node/resource basis %))
                 (filter (fn [resource]
                           (and (some? (resource/proj-path resource))
                                (resource/exists? resource)))))
               resource-node-ids)]
+    (thread-util/throw-if-interrupted!)
```
matgis, line 43: One final check before we embark on sorting the results.

editor/src/clj/editor/resource_dialog.clj
```diff
@@ -115,8 +120,10 @@
                                          (accept-fn %)))
                            (g/node-value workspace :resource-list))
         tooltip-gen (:tooltip-gen options)
+        special-filter-fns {"refs" (partial refs-filter-fn project)
+                            "deps" (partial deps-filter-fn project)}
```
matgis, line 124: Moved outside the `:filter-fn` below. No need to do it more than once.

editor/src/clj/editor/ui/fuzzy_choices.clj
```diff
@@ -53,10 +54,16 @@
...
+    (let [fuzzy-matched-options
+          (->> options
+               (r/take-while (thread-util/thread-uninterrupted-predicate))
+               (r/map #(option->fuzzy-matched-option option->matched-text filter-text %))
+               (r/filter some?)
+               (r/foldcat))]
+
+      (thread-util/throw-if-interrupted!)
+      (sort (partial option-order option->label-text)
+            (seq fuzzy-matched-options)))))
```
matgis, line 66: Tidy up formatting and inject interruption checks. Since reducers run on other threads in the thread pool, we use `(thread-util/thread-uninterrupted-predicate)` to check the interrupted flag of the the instigating thread, not the worker thread.

sprocketc (approved):
This nicely unblocks the UI when we type to filter the items.

The only noticeable and maybe questionable difference, was that the initial filtered items calculation was removed, so the list is empty when we open the dialog, until we start typing (https://github.com/defold/defold/pull/10983/files#diff-1c9a536e685c01a76e5860248531f14522b70744c2c82750607a9780d15164a7L663)

I am not sure if this was intentional. I'll approve this for now to review #10986. Reintroducing the initial list seems to be trivial.

matgis:
Oh, I realize now I fixed the empty list issue in the subsequent PR, but not this one. Sorry about that.

PR #10986, opened 2025-07-20
Asset Picker optimizations

### LINKED ISSUE #10739: Faster search using search dialog (cmd+p)

matgis:
Shortened the time it takes before you see the results when filtering in the Asset Picker.

Fixes #10739

### Technical changes
* Added `util.bit-set` module.
* The fuzzy text matching algorithm now uses bit-sets instead of vectors of indices.
* Moved some prep-work outside the fuzzy text matching loop.

-----

⚠️ This PR was made on top of #10983, which should be reviewed and merged before tackling this one.


editor/src/clj/util/coll.clj
```diff
@@ -28,6 +28,11 @@
   collection specified as the second argument, using a transducer composed of
   the remaining arguments. Returns the resulting collection. Supplying :eduction
   as the destination returns an eduction instead."
+  ([from to]
+   (case to
+     :eduction `(->Eduction identity ~from)
+     `(into ~to
+            ~from)))
```
matgis, line 35: Unrelated: Added a `coll/transfer` arity without any transducer.

editor/src/clj/editor/fuzzy_text.clj
```diff
@@ -226,69 +242,95 @@
...
+  contain the character indices in the string that matched the pattern in
+  sequential order. A lower score represents a better match."
+  [prepared-pattern ^String string]
+  (let [[upper-pattern lower-pattern] prepared-pattern]
+    (match-impl upper-pattern lower-pattern string 0)))
 
 (defn- apply-filename-bonus
   "Applies a bonus for a match on the filename part of a path."
-  [^String path ^long basename-start [^long score matched-indices]]
+  [^String path ^long basename-start [^long score matching-indices]]
```
matgis, line 283: Renamed argument to `matching-indices`, which is the term we use for it everywhere else.

sprocketc (approved):
Works great!

I can't really comment on the `BitSet` optimization. It should be faster, and it appears to work as expected.

As already mentioned here https://github.com/defold/defold/pull/10983#issuecomment-3095684732, the empty list issue has been resolved.

PR #10988, opened 2025-07-20
Fix issue where the same image can't be used more than once in a cubemap

AGulev:
Fixed issue where attempting to bundle a project with a cubemap that uses the same image more than once resulted in a null pointer exception.

Fix https://github.com/defold/defold/issues/10987

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/CubemapBuilder.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/CubemapBuilder.java
@@ -17,6 +17,10 @@ package com.dynamo.bob.pipeline;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.util.EnumSet;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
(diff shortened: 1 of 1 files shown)


aglitchman:
😍

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/CubemapBuilder.java
```diff
@@ -38,26 +42,57 @@
...
+
         TaskBuilder taskBuilder = Task.newBuilder(this)
                 .setName(params.name())
-                .addInput(input)
-                .addInput(input.getResource(builder.getRight()))
-                .addInput(input.getResource(builder.getLeft()))
-                .addInput(input.getResource(builder.getTop()))
-                .addInput(input.getResource(builder.getBottom()))
-                .addInput(input.getResource(builder.getFront()))
-                .addInput(input.getResource(builder.getBack()))
```
AGulev, line 54: we can't add it like that because `addInput` deduplicate resource inputs, so I added logic that takes it into account.
new logic also helps to avoid multibuilding of the same textures many times

britzl (approved):

PR #10989, opened 2025-07-20
Resolve dependencies faster when bundling using Bob

AGulev:
With this change, Bob uses two threads to resolve dependencies. Two simultaneous connections to a remote server will not be treated as "Too many requests" but will double the speed of dependency resolution (it depends on the connection, but even when all the dependencies are in the cache, cache checking goes twice as fast).

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
@@ -2006,6 +2006,7 @@ public class Project {
 
     /**
      * Resolve (i.e. download from server) the stored lib URLs.
+     *
      * @throws IOException
      */
     public void resolveLibUrls(IProgress progress) throws IOException, LibraryException {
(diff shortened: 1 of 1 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -2021,137 +2022,167 @@ public void resolveLibUrls(IProgress progress) throws IOException, LibraryExcept
...
-                        String envKey = password.substring(2, password.length() - 2);
-                        String envValue = getSystemEnv(envKey);
-                        if (envValue != null) {
-                            basicAuthData = username + ":" + envValue;
-                        }
-                    }
-                }
+            // Use a fixed thread pool with 2 threads for parallel downloads
+            ExecutorService executor = Executors.newFixedThreadPool(2);
+            List<Future<Void>> futures = new ArrayList<>();
```
AGulev, line 2028: it looks like a lot of changes, but it's mostly code formatting. 
The only change is adding ExecutorService and run every download as a Future

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -2021,137 +2022,167 @@ public void resolveLibUrls(IProgress progress) throws IOException, LibraryExcept
...
+                        throw (LibraryException) cause;
+                    } else if (cause instanceof IOException) {
+                        throw (IOException) cause;
+                    } else {
+                        throw new LibraryException(cause.getMessage(), cause);
                     }
-                    subProgress.worked(1);
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                    throw new LibraryException("Download interrupted", e);
```
AGulev, line 2174: also better exception handling

britzl (approved):
Looks good. Approved with a question. Can we use more than two threads?

AGulev:
@britzl  I'm afraid that services (w.g. GitHub or gitlab) may decline it as "Too many connections", so I decided to keep it 2

PR #10993, opened 2025-07-21
Add gamepad mapping for 8BitDo Micro on MacOS

hunterloftis:
Adds support for receiving input from an [8BitDo Micro gamepad](https://www.8bitdo.com/micro/) on MacOS.

### Technical changes
* Adds gdc-created mapping to default.gamepads


britzl (approved):
Thanks

PR #11002, opened 2025-07-22
Fix issue with search in path cache.

AGulev:
Since we don't work with path themselves but with strings, the logic that search in cache strings should take into account corner cases like it isn't enough to check `.startwith` path `test` when you are checking `test_folder/path.file`.

Tests also added

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/BundleResourcesTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/BundleResourcesTest.java
@@ -191,7 +191,6 @@ public class BundleResourcesTest {
         assertTrue(resourceMap.containsKey("test.txt"));
 
         project.getProjectProperties().putStringValue("project", "bundle_resources", "/restest2");
-        project.cleanupResourcePathsCache();
         resourceMap = ExtenderUtil.collectBundleResources(project, Arrays.asList(Platform.getHostPlatform()));
         assertEquals(3, resourceMap.size());
     }
@@ -228,14 +227,12 @@ public class BundleResourcesTest {
 
         // Exclude the conflicting file from bundle_resources
         project.getProjectProperties().putStringValue("project", "bundle_exclude_resources", "/restest1/common/collision.txt");
-        project.cleanupResourcePathsCache();
         Map<String, IResource> resourceMap = ExtenderUtil.collectBundleResources(project, Arrays.asList(Platform.getHostPlatform()));
         assertEquals(3, resourceMap.size());
         assertTrue(resourceMap.containsKey("collision.txt"));
 
         // Exclude the conflicting file from extension
         project.getProjectProperties().putStringValue("project", "bundle_exclude_resources", "/extension1/res/common/collision.txt");
-        project.cleanupResourcePathsCache();
         resourceMap = ExtenderUtil.collectBundleResources(project, Arrays.asList(Platform.getHostPlatform()));
         assertEquals(3, resourceMap.size());
         assertTrue(resourceMap.containsKey("collision.txt"));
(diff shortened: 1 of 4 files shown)


com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/BundleResourcesTest.java
```diff
@@ -191,7 +191,6 @@ public void testcollectBundleResources() throws Exception {
         assertTrue(resourceMap.containsKey("test.txt"));
 
         project.getProjectProperties().putStringValue("project", "bundle_resources", "/restest2");
-        project.cleanupResourcePathsCache();
```
AGulev, line 194: it's not needed here

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/test/ProjectTest.java
```diff
@@ -122,6 +122,7 @@ public void setUp() throws Exception {
         libraryUrls.add(new URL("http://localhost:8081/test_lib2.zip"));
         libraryUrls.add(new URL("http://" + BASIC_AUTH + "@localhost:8081/test_lib5.zip"));
         libraryUrls.add(new URL("http://" + BASIC_AUTH_ENV_TOKEN + "@localhost:8081/test_lib6.zip"));
+        libraryUrls.add(new URL("http://localhost:8081/test.zip"));
```
AGulev, line 125: I added new mount for my test

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/test/ProjectTest.java
```diff
@@ -284,6 +285,30 @@ public void testFindResourceDirs() throws Exception {
...
+        assertEquals(0, pathResults.size());
+
+        ArrayList<String> pathResults1 = new ArrayList<String>();
+        project.findResourcePaths("/test/file.in", pathResults1);
+        assertEquals(1, pathResults1.size());
+
+        ArrayList<String> pathResults2 = new ArrayList<String>();
+        project.findResourcePaths("/test", pathResults2);
+        assertEquals(1, pathResults2.size());
+    }
```
AGulev, line 310: check that test situations like `test` isn't `test_1/`

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -2340,13 +2334,25 @@ public void handleFile(String filePath, Collection<String> results) {
...
+        if ((ext == null || ext.isEmpty()) && (_path == null || _path.isEmpty())) {
             result.addAll(allResourcePathsCache);
             TimeProfiler.stop();
             return;
         }
+        IResource res = fileSystem.get(_path);
+        try {
+            if ( (_path != null && !_path.isEmpty()) && (!res.isFile() || res.getContent() == null)) {
+                if (!_path.endsWith("/")) {
+                    _path += "/";
```
AGulev, line 2349: actual fix

JCash: Trailing `/` makes sense when it's not a file (i.e. dir), but when it's an empty file, do we get an empty byte array, or null? (I don't remember)

JCash (approved):

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
AGulev: It will be an empty byte array. 
I added this because of how MockFileSystem works: returns true for a folder when call `isFile()`
I wanted to fix MockFS, but then decided to add another check in cases some other filesystem works the same. In most of the cases this condition will not even be checked

PR #11021, opened 2025-07-28
Fix few `b2d.body` functions that always throw `vector3 expected, got userdata`.

### LINKED ISSUE #11017: get_linear_velocity_from_local_point expects first term to be both a body and vector3
Running the following code:
```
function init(self)
	local url = msg.url(nil, "go", "collisionobject")
	local body = b2d.get_body(url)

	b2d.body.get_linear_velocity_from_local_point(body,  vmath.vector3(1, 1, 1))
end
```
Generates the following error:
```

ERROR:SCRIPT: main/main.script:8: bad argument #1 to 'get_linear_velocity_from_local_point' (vector3 expected, got userdata)
stack traceback:
  [C]:-1: in function get_linear_velocity_from_local_point
  main/main.script:8: in function <main/main.script:1>
```
It seems that this is caused by  a bug in DEFOLD where both .get_linear_velocity_from_local_point and .get_linear_velocity_from_world_point expect the first term to be a body and a vector3

https://github.com/defold/defold/blob/d80ef0c20ad81b332e333f24907e801cd971a17a/engine/gamesys/src/gamesys/scripts/box2d/script_box2d_body.cpp#L450C1-L457C6

Platform:
Windows 11

Defold:
Version 1.10.3

AGulev:
Fix https://github.com/defold/defold/issues/11017

### DIFF
--- a/engine/gamesys/src/gamesys/scripts/box2d/script_box2d_body.cpp
+++ b/engine/gamesys/src/gamesys/scripts/box2d/script_box2d_body.cpp
@@ -406,7 +406,7 @@ namespace dmGameSystem
     {
         DM_LUA_STACK_CHECK(L, 1);
         b2BodyId* body = CheckBody(L, 1);
-        b2Vec2 p = CheckVec2(L, 1, GetPhysicsScale());
+        b2Vec2 p = CheckVec2(L, 2, GetPhysicsScale());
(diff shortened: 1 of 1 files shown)


ekharkunov (approved):
👍

PR #11022, opened 2025-07-28
Pan scene on right click down and update cursor while panning

### LINKED ISSUE #10996: Move editor focus by holding down the right mouse button
Is it a private platform specific feature? - No, it's common style of editor navigation.

Currently, you need to use the mouse scroll wheel to move along the X axis in the interface object editor. This is inconvenient because you need to calculate the force of the finger swing.

The solution is implemented in the Unity editor. The user holds down the right mouse button, moves the mouse cursor in space, and the editor content moves after the cursor. The mouse scroll is responsible for zooming.
See video below

https://youtu.be/hvdR5_Wfrsk

sprocketc:
Fixes #10996

### Technical changes
Testing how this would work on this PR, since I was already thinking about it while working on the scene context menu. We need to allow the action to propagate on right click, and move the selection and context menu trigger from `mouse-pressed` to `mouse-released`. Rendering a different cursor on pan is useful, especially when we use right click, because slightly moving the cursor on right click down, would cancel the context menu with no other indication. Cursor swapping should be triggered on mouse down when we use middle click or alt + left click to pan. When we use right click, we need to wait for a significant movement, because the context menu might be triggered instead.


editor/src/clj/editor/camera.clj
```diff
@@ -665,14 +668,19 @@
...
       :mouse-released (do
                         (g/user-data-swap! self ::ui-state assoc :last-x nil :last-y nil :movement :idle)
-                        (if (= movement :idle) action nil))
+                        (g/set-property! self :cursor-type Cursor/DEFAULT)
+                        (if (or (= movement :idle) is-secondary) action nil))
       :mouse-moved (if (not (= :idle movement))
                      (do
                        (g/user-data-swap! self ::ui-state assoc :last-x x :last-y y)
-                       nil)
+                       (g/set-property! self :cursor-type Cursor/MOVE)
```
sprocketc, line 682: I tried to use `Cursor/CLOSED_HAND`, but `CLOSED_HAND`, `OPEN_HAND_HAND`, `HAND` all render a pointer cursor on Linux for some reason.

editor/src/clj/editor/ui.clj
```diff
@@ -2421,3 +2421,11 @@
     (on-cancel! cancel-fn)
     (auto-commit! update-fn)
     (select-all-on-click!)))
+
+(defn set-cursor
+  [^Node node ^Cursor cursor-type]
+  ;; The cursor refresh appears buggy at the moment.
+  ;; Calling setCursor with DISAPPEAR before setting the cursor forces it to refresh.
```
sprocketc, line 2428: This does not appear to be true in my case so it could be fixed, but I don't want to risk removing `Cursor/DISAPPEAR` without additional confirmation. If we remove it, I suppose there is no need for this helper function.

editor/src/clj/editor/camera.clj
```diff
@@ -626,16 +628,28 @@
...
+        {:keys [x y type delta-y alt button]} action
+        is-secondary (= button :secondary)
         movement (if (= type :mouse-pressed)
                    (get movements-enabled (camera-movement action) :idle)
                    (:movement ui-state))
         camera (g/node-value self :camera)
+        is-significant-drag (or (not= (:button action) :secondary)
+                                (and initial-x
+                                     initial-y
+                                     (significant-drag? [x y] [initial-x initial-y])))
```
sprocketc, line 652: Check for `significant-drag?` on right click + drag only.

editor/src/clj/editor/scene_tools.clj
```diff
@@ -512,7 +512,8 @@
                            original-values (->> selected-renderables
                                                 scene-picking/top-nodes
                                                 (mapv original-values))]
-                       (when (not (empty? original-values))
+                       (when (and (not (empty? original-values))
+                                  (= (:button action) :primary))
```
sprocketc, line 516: Propagate action on scene tool controller right click. This should also fix the second issue mentioned here https://github.com/defold/defold/pull/10263#issuecomment-2678300884.

editor/src/clj/editor/scene.clj
```diff
@@ -939,14 +941,26 @@
       component (do (fx/delete-component component) (g/user-data! view-node key nil))
       desc (g/user-data! view-node key (fx/create-component desc)))))
 
+(defn cursor
+  "Maps inconsistent cursor types across platforms.
+  See https://bugs.openjdk.org/browse/JDK-8101062"
```
vlaaad, line 946: Sweet, good that you documented this and linked to the issue!

vlaaad (approved):
LGTM! Works great!

PR #11031, opened 2025-07-30
Fix remotery crash on reboot

AGulev:
This PR fixes a crash reported in Beta forum post: https://forum.defold.com/t/defold-1-10-4-beta/81007/38?u=agulev


ekharkunov (approved):
👍

PR #11033, opened 2025-07-31
Fixed crash when a material constant was used in a cloned node but the prototype node had been removed

### LINKED ISSUE #11011: Crash when deleting Gui node that has shader constant set and has been cloned.
**Describe the bug**
Deleting a gui node after setting a shader constant and cloning it, crashes the game. Does not matter if clone or original node is deleted.

**To Reproduce)**
1. Clone https://github.com/nog1potje/defold-gui-constants-crash
2. Run the project
3. Observe the crash

**Expected behavior**
No crash in this usecase.

**Defold version:**
  - Version 1.10.3

**Platforms:**
 - Platforms: Windows 11

See provided gihub repo

AGulev:
This fix creates a separate material constant buffer for cloned nodes. This not only fixes the crash when the prototype node is removed, but also makes it possible to use independent values per cloned node.

Fix https://github.com/defold/defold/issues/11011

### DIFF
--- a/engine/gui/src/gui.cpp
+++ b/engine/gui/src/gui.cpp
@@ -393,6 +393,7 @@ namespace dmGui
         scene->m_SetMaterialPropertyCallback = params->m_SetMaterialPropertyCallback;
         scene->m_SetMaterialPropertyCallbackContext = params->m_SetMaterialPropertyCallbackContext;
         scene->m_DestroyRenderConstantsCallback = params->m_DestroyRenderConstantsCallback;
+        scene->m_CloneRenderConstantsCallback = params->m_CloneRenderConstantsCallback;
         scene->m_OnWindowResizeCallback = params->m_OnWindowResizeCallback;
         scene->m_NewTextureResourceCallback = params->m_NewTextureResourceCallback;
         scene->m_DeleteTextureResourceCallback = params->m_DeleteTextureResourceCallback;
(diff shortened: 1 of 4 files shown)


engine/gui/src/gui.cpp
```diff
@@ -4269,6 +4270,19 @@ namespace dmGui
         out_n->m_Node.m_ResetPointProperties = 0;
```
AGulev, line 4270:
```
out_n->m_Node = n->m_Node;
```
Just copies a pointer.  
As a result, we have a shared buffer between the nodes (the prototype and the cloned node).  
If one of these nodes is removed, the engine crashes because the remaining pointers become invalid.

engine/gui/src/gui.cpp
```diff
@@ -4269,6 +4270,19 @@ namespace dmGui
...
+        
+        // Handle render constants - clone them if callback is available and source has them
+        if (n->m_Node.m_RenderConstants && scene->m_CloneRenderConstantsCallback)
+        {
+            out_n->m_Node.m_RenderConstants = scene->m_CloneRenderConstantsCallback(n->m_Node.m_RenderConstants);
+            out_n->m_Node.m_RenderConstantsHash = n->m_Node.m_RenderConstantsHash;
+        }
+        else
+        {
+            // Ensure cloned nodes don't share the pointer if no callback available
```
britzl: I am looking at this diff on a phone which is not optimal, but under what circumstances is no callback available?

AGulev: Yeah, I should probably change comment 
because it is two conditions
`n->m_Node.m_RenderConstants` exists
and
`scene->m_CloneRenderConstantsCallback` exists

We check if callbacks (we have a few) exist everywhere, so I do the same

AGulev: I removed this comments

AGulev: @britzl do you want me to remove it?

britzl (approved):

PR #11037, opened 2025-08-01
Enhance grid visibility on 3d mode

sprocketc:
Related to https://forum.defold.com/t/defold-1-10-4-beta/81007/28?u=sprocketc and https://github.com/defold/defold/pull/10345#pullrequestreview-2962225181.

### DIFF
--- a/editor/src/clj/editor/grid.clj
+++ b/editor/src/clj/editor/grid.clj
@@ -107,15 +107,14 @@
       (gl/gl-vertex-3d gl 0.0 0.0 (-> aabb types/max-p .z)))))
 
 (defn render-grid-sizes
-  [^GL2 gl ^doubles dir grids options]
+  [^GL2 gl ^doubles dir grids options is-2d]
   (let [{:keys [^double opacity color auto-scale]} options]
     (doseq [grid-index (range (if auto-scale 2 1))
             :let [^double fixed-axis (:plane grids)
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/grid.clj
```diff
@@ -127,16 +126,31 @@
...
+(defn- enable-fog
+  [^GL2 gl camera]
+  (let [min-fov (math/deg->rad (max ^double (:fov-x camera) ^double (:fov-y camera)))
+        fog-start (* min-fov ^double (:z-far camera))]
+    (doto gl
+      (.glEnable GL2/GL_FOG)
+      (.glFogi GL2/GL_FOG_MODE GL2/GL_LINEAR)
+      (.glFogfv GL2/GL_FOG_COLOR (float-array colors/scene-background) 0)
+      (.glFogf GL2/GL_FOG_START fog-start)
+      (.glFogf GL2/GL_FOG_END (* 2 fog-start)))))
```
sprocketc: I was trying to find a hardware accelerated way to fade out the grid on perspective mode.
Although using a fog got me close to what I wanted, a shader that makes the grid transparent depending on the distance from the camera might be a better option.

vlaaad (dismissed):
The user experience is much better now, let's get it into beta! 

Maybe the fog should be disabled only when it was enabled? Not sure if it matters...

vlaaad (approved):
Ship it!

PR #11046, opened 2025-08-02
Fixed issue when `sys.open_url()` reports false when a url is opened on Windows

### LINKED ISSUE #5549: sys.open_url reports false when a url is opened on Windows 10
On Windows 10, even though the url is opened successfully, it still reports as not successful
```
local success = sys.open_url("http://www.defold.com", {target = "_blank"})
if not success then
  print("failed to open url")
end
```

If a target can't detect that a URL was actually opened it should assumed it did assuming there are no other errors in the process

1.2.179

AGulev:
Fix https://github.com/defold/defold/issues/5549

### DIFF
--- a/engine/dlib/src/dlib/sys_posix.cpp
+++ b/engine/dlib/src/dlib/sys_posix.cpp
@@ -329,7 +329,7 @@ namespace dmSys
     Result OpenURL(const char* url, const char* target)
     {
         intptr_t ret = (intptr_t) ShellExecuteA(NULL, "open", url, NULL, NULL, SW_SHOWNORMAL);
-        if (ret == 32)
+        if (ret > 32)
         {
             return RESULT_OK;
         }


engine/dlib/src/dlib/sys_posix.cpp
```diff
@@ -329,7 +329,7 @@ namespace dmSys
     Result OpenURL(const char* url, const char* target)
     {
         intptr_t ret = (intptr_t) ShellExecuteA(NULL, "open", url, NULL, NULL, SW_SHOWNORMAL);
-        if (ret == 32)
+        if (ret > 32)
```
AGulev, line 332: >If the function succeeds, it returns a value greater than 32.

https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shellexecutea

ekharkunov (approved):
👍

PR #11048, opened 2025-08-04
Provide default grid size when tile dimensions are not available

sprocketc:


vlaaad (commented):
Shouldn't defaults be doubles?

vlaaad (approved):
LGTM!

PR #11049, opened 2025-08-04
Set center panel minimum width

### LINKED ISSUE #11020: Console area eats the game viewport area when its width is zero during editor resize
The bottom window (Console/Curve Editor/Build Errors/Search Results) goes to like 99% height when the editor window width changes to such a value that the “game view” area width becomes 0 and then returns to “normal” on Linux with Wayland (Hyprland).

When I briefly open another window next to the editor window and then I close that window I need to resize the console area every time.

It seems that manually changing the Assets and Outline panel widths to hide the game view area also tops the Console panel height when editor briefly resizes to w/e and back to normal.

**To Reproduce (REQUIRED)**
Steps to reproduce the behavior (tiling window manager):
1. Move the panel widths in such way that the game viewport area is not visible at all. [OR] Make the other panel widths large enough for the game viewport area width to go zero when the editor x size changes.
2. Open another window (terminal for example) so it opens next to the editor, changing its size.
3. Close the "another window".
4. See the console area with like 99% height (not 100% for some reason).

The console area not changing height.

  - Version 1.10.4 Beta
  - Version 1.10.3
  - Likely earlier versions too

**Platforms (REQUIRED):**
 - Platforms: Linux
 - OS: Arch Linux, Hyprland
 - Device: PC

**Video:**
https://github.com/user-attachments/assets/79c8b9d9-e553-456e-be8b-402c8a182426

sprocketc:
Set a minimum width for the center panel.

Fixes #11020

### Technical changes

I found a different way to work around the issue. By setting a minimum width, we avoid hiding the center panel without affecting the persisted height, and fix the mentioned issue indirectly. It also makes sense to have a min width set for the center panel in my opinion.

### DIFF
--- a/editor/resources/editor.fxml
+++ b/editor/resources/editor.fxml
@@ -70,7 +70,7 @@
                   <items>
                     <AnchorPane prefHeight="160.0" prefWidth="100.0">
                       <children>
-                        <SplitPane id="center-split" dividerPositions="0.5746666666666667" layoutX="24.0" layoutY="398.0" orientation="VERTICAL" prefHeight="200.0" prefWidth="160.0" AnchorPane.bottomAnchor="0.0" AnchorPane.leftAnchor="0.0" AnchorPane.rightAnchor="0.0" AnchorPane.topAnchor="0.0">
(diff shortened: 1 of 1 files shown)


vlaaad (approved):
LGTM!

PR #11050, opened 2025-08-04
Add `collision_type` editor script property for disambiguation

### LINKED ISSUE #11003: editor.tx.add "components" `type` value collides with `type` data
In Defold 1.10.4, the ability to add components to game objects via editor scripts ([docs](https://defold.com/manuals/editor-scripts/#editing-game-objects)) includes the ability to set properties like:

```lua
editor.tx.add(go_node, "components", {
        type = "collisionobject",
        id = component_id,
        collision_shape = "/" .. file_path,
        group = "ground",
        mask = "players",
    })
```

However, by using `type = "collisionobject"` to determine which type of component to add, it makes it impossible to set the `type` property on the component:

Image

Trying to set the collision object "type" to be static, for instance, yields:

```
ERROR:EXT: Create collision hulls's "run" in /main/levels/level_tiles.editor_script failed: /main/levels/level_tiles.editor_script:261 type is not component-reference, sound, factory, model, collisionobject, label, objectinterpolation, camera, sprite, mesh, collectionfactory or collectionproxy
```

**To Reproduce (REQUIRED)**

See that this call (modified from the doc examples) can never make sense as the keys collide:

```lua
    editor.tx.add("/npc.go", "components", {
        type = "collisionobject",
        id = "collision",
        type = "COLLISION_OBJECT_TYPE_STATIC",
        shapes = {
            {
                type = "shape-type-box",
                dimensions = {32, 32, 32}
            }
        }
    }),
```

When the UI says "available as 'type' in editor scripts", I should be able to get/set the "type". Or alternatively, maybe I misunderstand the docs & the UI and there's some other mechanism where "type" is allowed?

  - Version 1.10.4

**Platforms (REQUIRED):**
 - Platforms: MacOS
 - OS: Sonoma
 - Device: MBP

vlaaad:
Previously, the collision objects had a `type` property that could be shadowed by a collision component `type` property when the collision object is embedded in a game object. We added `collision_type` that should be used instead of `type` to refer to the collision object's physics object type.

Fixes #11003

### DIFF
--- a/editor/src/clj/editor/attachment.clj
+++ b/editor/src/clj/editor/attachment.clj
@@ -332,6 +332,14 @@
         (map gt/source-id)
         (filter #(= child-node-type (g/node-type* basis %)))))))
 
+(defn alternative
+  "Return alternative node id for a given node id"
+  [workspace node-id {:keys [basis] :as evaluation-context}]
+  (let [current-state (workspace/node-attachments basis workspace)
+        node-type (g/node-type* basis node-id)]
+    (when-let [alternative-fn (:alternative (clojure.core/get current-state node-type))]
+      (alternative-fn node-id evaluation-context))))
+
 ;; SDK api
 (defn nodes-getter
   "Node list getter that returns its :nodes output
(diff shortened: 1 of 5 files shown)


sprocketc (approved):
Works great!

hunterloftis:
Thanks for fixing this @vlaaad - I just updated the project using it. The way y'all are expanding editor scripting is brilliant.

PR #11052, opened 2025-08-04
Set PBR properties for shaders

Jhonnyg:
Material properties from a GLTF file are now exposed into shaders via a special hard-coded uniform:

```glsl
uniform PbrMaterial
{
    // Common properties
    vec4 pbrAlphaCutoffAndDoubleSidedAndIsUnlit; // R: alphaCutoff (Default=0.5), G: doubleSided (Default=false), B: unlit (Default=false)
    vec4 pbrCommonTextures;                      // R: use normalTexture, G: use occlusionTexture, B: use emissiveTexture
    PbrMetallicRoughness  pbrMetallicRoughness;
    PbrClearCoat pbrClearCoat;
   ... etc.
}
```

All properties that are available in our data format is now set in a model components material, if they exist in the shader. Here is the material specification from the engine source, which contains all the fields we currently support:
https://github.com/defold/defold/blob/23147f091896d079ec1cdba1e6648f2b31f3e1be/engine/rig/proto/rig/rig_ddf.proto#L154

To read more about how this works, head over to the [manual entry](https://defold.com/manuals/).

**Note: There is currently no PBR shader as part of the builtins. This will be included at a later time, but in the meanwhile you can use the Defold [PBR extension](https://github.com/defold/defold-pbr).**


engine/gamesys/src/gamesys/test/mesh_prebuilt_project/model/pbr.fp
```diff
@@ -0,0 +1,189 @@
+#version 140
```
Jhonnyg, line 1: This shader snippet should be the "truth" later on when we move this into builtins. But I don't want to add anything into builtins in this PR, so we'll leave that for later.

As the next step, we could add the shader inputs into a .glsl file into builtins that can then be included if people want to start using this right away?

engine/gamesys/src/gamesys/test/model/pbr.fp
```diff
@@ -0,0 +1,189 @@
+#version 140
```
Jhonnyg, line 1: This is a straight copy from the prebuilt project. The current structure kinda has to be like this I think, since the resource must exist in both places for the prebuilt stuff to work.

editor/src/clj/editor/model_scene.clj
```diff
@@ -184,9 +185,11 @@
                 vertex-attribute-bytes (assoc :vertex-attribute-bytes vertex-attribute-bytes))]
     (graphics/put-attributes! vbuf [mesh-renderable-data])))
 
+(def my-atom (atom 0))
+
 (defn- render-mesh-opaque-impl [^GL2 gl render-args renderable request-prefix override-shader override-vertex-description extra-render-args]
   (let [{:keys [node-id user-data ^Matrix4d world-transform]} renderable
-        {:keys [material-attribute-infos mesh-renderable-data textures vertex-attribute-bytes]} user-data
+        {:keys [material-attribute-infos material-data mesh-renderable-data textures vertex-attribute-bytes]} user-data
```
Jhonnyg, line 189: Maybe material-data should be named differently? Not sure 🤔

editor/src/clj/editor/model_scene.clj
```diff
@@ -300,23 +307,166 @@
       default-material-ids
       ret)))
 
-(defn- make-renderable-mesh [mesh mesh-material-index->material-name]
+(defn- vec->vector4d [v]
+  (Vector4d. (nth v 0) (nth v 1) (nth v 2) (nth v 3)))
```
Jhonnyg: I couldn't find a function that did this, but maybe there is one somewhere?

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -90,6 +114,7 @@ namespace dmGameSystem
         ModelResourceBuffers*       m_Buffers;
         dmRigDDF::Model*            m_Model;    // Used for world space materials
         dmRigDDF::Mesh*             m_Mesh;     // Used for world space materials
+        HComponentRenderConstants   m_RenderConstants; // Used for PBR properties, will be null if PBR data not needed.
```
Jhonnyg, line 117: Note for reviewer here (I'm open to changing this):

Right now we have a "shared" HComponentRenderConstants in the component that may or may not be null, which is mainly used for things like setting constants via go.set. This constant buffer is then merged with the potential render constant buffers that each individual render item may have. I did it this way since we use the constant buffer to marshall the PBR data from the rig material into the shader for each render item. 

While a bit extra to maintain, this could potentially be extended to then support setting individual parameters per render item down the line. I think we want to be able to address render items somehow later on, for example by chanrging a material on a part of a model and so on.

engine/gamesys/src/gamesys/components/comp_private.cpp
```diff
@@ -570,6 +569,25 @@ int AreRenderConstantsUpdated(HComponentRenderConstants constants)
     return constants->m_Updated ? 1 : 0;
 }
 
+// TODO: Can this be improved? The "SetRenderConstant" call feels extremely expensive.
+void CopyRenderConstants(HComponentRenderConstants dst, HComponentRenderConstants src)
```
Jhonnyg, line 573: I think we would like to support constant buffers as a graphics resource (similar to textures) on the render object. But before that, we should take a stab att refactoring the render object so it can support arbitrary render resources I think. For example, the "max 8 textures" is a hard limit right now which is very limiting :(

engine/gamesys/src/gamesys/test/mesh_prebuilt_project/model/pbr.fp
Jhonnyg: As the next step, or even in this PR, we could potentially add the shader inputs into a .glsl file into builtins that can then be included if people want to start using this right away. However, we might want to reconsider how the data is laid out once we can use uniform buffers as render resources (i.e refactor our constant buffer usage in the component), so perhaps we should wait with shipping something into builtins anyway?

engine/gamesys/src/gamesys/test/mesh_prebuilt_project/model/pbr.fp
```diff
@@ -0,0 +1,189 @@
...
+#define PBR_METALLIC_ROUGHNESS
+#define PBR_SPECULAR_GLOSSINESS
+#define PBR_CLEARCOAT
+#define PBR_TRANSMISSION
+#define PBR_IOR
+#define PBR_SPECULAR
+#define PBR_VOLUME
+#define PBR_SHEEN
+#define PBR_EMISSIVE_STRENGTH
+#define PBR_IRIDESCENCE
```
Jhonnyg, line 12: Most of these are extensions, and in a future PBR shader a user can trim the shader down by selecting a few of these before including the PBR shader code. Not all materials will use e.g iridescence or emissive strength, but if we leave the code on we have to still run that code in the shader.

JCash (dismissed):

editor/src/clj/editor/model_scene.clj
matgis: Yep! There's `math/clj->vecmath`. It has implementations for all the vector and matrix types we use.

editor/src/clj/editor/model_scene.clj
```diff
@@ -300,23 +302,166 @@
...
+          pbr-clear-coat (:clearcoat mesh-material-data)
+          pbr-transmission (:transmission mesh-material-data)
+          pbr-ior (:ior mesh-material-data)
+          pbr-specular (:specular mesh-material-data)
+          pbr-volume (:volume mesh-material-data)
+          pbr-sheen (:sheen mesh-material-data)
+          pbr-emissive-strength (:emissive-strength mesh-material-data)
+          pbr-iridescence (:iridescence mesh-material-data)
+          pbr-texture-index->value (fn ^double [^long ix]
+                                     (if (>= ix 0) 1.0 0.0))]
```
matgis, line 318: Not sure I understand this. Should it be the index instead of `1.0`?

editor/src/clj/util/digestable.clj
```diff
@@ -218,6 +219,10 @@
   (digest! [value writer opts]
     (digest-tagged! 'LayoutResult (str value) writer opts))
 
+  Graphics$ShaderDesc
+  (digest! [value writer opts]
+    (digest-tagged! 'ShaderDesc (str value) writer opts))
+
```
matgis, line 225: There are probably more efficient ways to do this, but I think it's fine for now.

editor/test/integration/material_test.clj
```diff
@@ -33,6 +33,24 @@
...
+          :has-volume true
+          :has-clearcoat true
+          :has-sheen true
+          :has-specular true
+          :has-transmission true
+          :has-specular-glossiness true
+          :has-ior true
+          :has-parameters true
+          :has-emissive-strength true}))))
+
```
matgis, line 53: This needs to be`(is (= pbr-parameters {...}))` to actually test equality. Also, it should be in `(= expected actual)` order to get reported correctly when it fails. When writing a test, it's a good idea to make it fail on purpose just to make sure it will actually fail when something is off.

editor/src/clj/editor/material.clj
```diff
@@ -123,6 +124,25 @@
...
+    MaterialBuilder$PbrParameterType/PBR_PARAMETER_TYPE_SHEEN :has-sheen
+    MaterialBuilder$PbrParameterType/PBR_PARAMETER_TYPE_EMISSIVE_STRENGTH :has-emissive-strength
+    MaterialBuilder$PbrParameterType/PBR_PARAMETER_TYPE_IRIDESCENCE :has-iridescene))
+
+(defn- build-target-pbr-params [shader-reflection]
+  (when-some [pbr-parameters (MaterialBuilder/getPbrParameters shader-reflection)]
+    (into {:has-parameters true}
+          (map #(vector (pb-parameter-type->pbr-parameter-type-pb-keyword %) true))
+          pbr-parameters)))
+
```
matgis, line 130: Suggestion: Why not have a static function in Bob that takes a `Graphics.ShaderDesc.ShaderReflection` and returns a populated `Material.MaterialDesc.PbrParameters` protobuf message? Then both Bob and the editor could use that.

For the editor, you can convert a protobuf message to a Clojure map using `protobuf/pb->map-with-defaults` or `pb->map-without-defaults`, depending on if you want the default values for the fields written to the binaries or not. We typically don't write default field values, but I saw the test you added expects them for some reason.

matgis (changes requested):
I put in some comments and suggestions on the editor-side changes, but I think we could merge this without addressing them. I did notice the editor test you added will not actually test the thing you want, so you'll need to fix that before merging.

editor/src/clj/editor/model_scene.clj
Jhonnyg: Hm no, because we can't really do anything with indices in the shader. We can't bind textures as arrays, so the value simply means if the value should be gathered from a specific sampler.

editor/src/clj/editor/material.clj
Jhonnyg: Not sure I understand, the test I wrote checks that all values have been set. I don't care about the default values, as they will be false.

matgis: Hmm, I might have misread the test. 😅

Anyway, I think we're doing the correct thing here by not including default pb-field values in the built binaries, so everything is fine here.

PR #11057, opened 2025-08-06
Allow easily bundling template projects with the editor.

### LINKED ISSUE #3930: Bundle downloaded zip archives of the new project templates (DEFEDIT-1368)
Currently the user requires internet to be able to create a new project. We should download the zip archives for the templates (empty project, mobile and desktop game) when the editor is built and ship them with the editor.

Forum: (LINK REMOVED)

vlaaad:
For starters, I bundle the empty project, as I assume it's the most used. It's just 2.5 kb. I decided not to bundle desktop/mobile templates since they are noticeably bigger (600 KB/400 KB), but with the pipeline in place, we can simply bundle more templates if needed by adding `:bundle true` in `welcome.edn`.

Fixes #3930

### DIFF
--- a/editor/.gitignore
+++ b/editor/.gitignore
@@ -18,6 +18,7 @@ pom.xml.asc
 /resources/dialogs.css
 /resources/editor.css
 /resources/splash.css
+/resources/template-projects
 /editor2.*.log
 /log.txt
 /.project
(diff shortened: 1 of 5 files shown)


vlaaad:
I tried new project creation on Windows and macOS; it would be nice to also check it on Linux.

sprocketc (approved):
Works fine on Linux!

PR #11058, opened 2025-08-06
Fixed an issue where a 16k x 16k atlas couldn’t be unpacked if the file size was bigger than 1 GB

### LINKED ISSUE #10404: 100% crash for 16K*16K atlas
The 16K by 16K atlas causes errors when the build starts, while it displays correctly in the editor.
This happens on all our Windows computers and laptops, regardless of the installed graphics cards.

Defold versions: 1.9.4 - 1.9.8

Repo-case project:
[__Crash_atlas_16Kx16K.zip](https://github.com/user-attachments/files/19513703/__Crash_atlas_16Kx16K.zip)

AGulev:
The 1 GB limitation for compressed files was doubled to make it possible to work with 16k x 16k atlases.

Fix https://github.com/defold/defold/issues/10404

### DIFF
--- a/engine/dlib/src/dlib/lz4.cpp
+++ b/engine/dlib/src/dlib/lz4.cpp
@@ -21,7 +21,6 @@ namespace dmLZ4
     Result DecompressBuffer(const void* buffer, uint32_t buffer_size, void* decompressed_buffer, uint32_t max_output, int* decompressed_size)
     {
         Result r;
-        // When we set an output size larger than 1G (or closer to 2G) lz4 fails for some reason...
         if(max_output <= DMLZ4_MAX_OUTPUT_SIZE)
         {
             *decompressed_size = LZ4_decompress_safe((const char*)buffer, (char *)decompressed_buffer, buffer_size, max_output);
--- a/engine/dlib/src/dlib/lz4.h
+++ b/engine/dlib/src/dlib/lz4.h
@@ -18,7 +18,7 @@
 #include <stdint.h>
 #include "shared_library.h"
 
-#define DMLZ4_MAX_OUTPUT_SIZE (1 << 30)
+#define DMLZ4_MAX_OUTPUT_SIZE (1LL << 31)  // 2GB limit instead of 1GB
 
 namespace dmLZ4
 {
--- a/engine/dlib/src/dlib/lz4.h
+++ b/engine/dlib/src/dlib/lz4.h
@@ -18,7 +18,7 @@
 #include <stdint.h>
 #include "shared_library.h"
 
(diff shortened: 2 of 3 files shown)


engine/dlib/src/dlib/lz4.h
```diff
@@ -18,7 +18,7 @@
 #include <stdint.h>
 #include "shared_library.h"
 
-#define DMLZ4_MAX_OUTPUT_SIZE (1 << 30)
+#define DMLZ4_MAX_OUTPUT_SIZE (1LL << 31)  // 2GB limit instead of 1GB
```
AGulev: But maybe we don't need it at all.. 
This workaround added many years ago, not sure if it is still make sense: 
https://github.com/defold/defold/blob/e7d9a387928d23790a199458274fa070a3bcaf88/engine/dlib/src/dlib/lz4.cpp#L24-L37

engine/resource/src/resource_archive.cpp
```diff
@@ -605,6 +605,10 @@ namespace dmResourceArchive
             dmLZ4::Result r = dmLZ4::DecompressBuffer(source_data, source_data_size, buffer, size, &decompressed_size);
             if (dmLZ4::RESULT_OK != r)
             {
+                dmLogError("LZ4 decompression failed: result=%d, expected size=%u, actual size=%d", r, size, decompressed_size);
```
AGulev, line 608: show a proper error

engine/dlib/src/dlib/lz4.h
britzl: If it works with a larger value then maybe no point having the check there at all? We have updated lz4 at some point I think? Or maybe we can fix the issue in lz4 if the problem still exists?

britzl (dismissed):

engine/dlib/src/dlib/lz4.h
AGulev: It works, but limited to two gigabytes anyway by build pipeline (java array size)

engine/dlib/src/dlib/lz4.cpp
```diff
@@ -21,7 +21,6 @@ namespace dmLZ4
     Result DecompressBuffer(const void* buffer, uint32_t buffer_size, void* decompressed_buffer, uint32_t max_output, int* decompressed_size)
     {
         Result r;
-        // When we set an output size larger than 1G (or closer to 2G) lz4 fails for some reason...
```
AGulev, line 24: not true anymore, but I left limit with another comment

britzl (approved):

PR #11064, opened 2025-08-07
Add a preference to disable auto-closing parens in the editor.

### LINKED ISSUE #10781: Editor setting needed to disable auto-inserting of closing parens in the code editor.

vlaaad:
New preference **Code → Auto-insert closing parens** allows turning off the auto-closing of parens in the editor.

Fixes #10781

### DIFF
--- a/editor/src/clj/editor/code/script.clj
+++ b/editor/src/clj/editor/code/script.clj
@@ -121,7 +121,10 @@
                                    "string.quoted.other.multiline.lua"
                                    "string.quoted.double.lua"
                                    "string.quoted.single.lua"
-                                   "constant.character.escape.lua"}
+                                   "constant.character.escape.lua"
+                                   "punctuation.definition.comment.lua"
+                                   "comment.block.lua"
+                                   "comment.line.double-dash.lua"}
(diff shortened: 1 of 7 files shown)


editor/src/clj/editor/code/script.clj
```diff
@@ -121,7 +121,10 @@
                                    "string.quoted.other.multiline.lua"
                                    "string.quoted.double.lua"
                                    "string.quoted.single.lua"
-                                   "constant.character.escape.lua"}
+                                   "constant.character.escape.lua"
+                                   "punctuation.definition.comment.lua"
+                                   "comment.block.lua"
+                                   "comment.line.double-dash.lua"}
```
vlaaad, line 127: Unrelated: I disabled auto-closing parens in comment blocks. Reason: it's common to type English text with single quotes in comments, i.e., words like `don't`.

sprocketc (approved):
Works great!

PR #11065, opened 2025-08-07
Make editor server port more accessible

### LINKED ISSUE #8737: Configurable port for Editor HTTP API
[Editor HTTP API](https://github.com/defold/defold/blob/dev/editor/doc/http-api.md)  is a great solution especially for 3rd party IDE integrations. I would like to use it with Sublime Text(or vscode)

Currently its [port is random](https://github.com/defold/defold/blob/dev/editor/src/clj/editor/boot_open_project.clj#L204) and there is only way to get the random port of the Editor's http-server is by looking at console output. There is a [workaround](https://forum.defold.com/t/on-reloaded-lifecycle-hook-in-editor/75667/2), but it is problematic and it is not a proper solution. 

it would be perfect, if the port number can be configured from editor’s preference window or before the Editor lauch(maybe by passing a param to Editor). But of course it requires a port check before starting http-server. If the port is in use it might use a random port

vlaaad:
When the editor opens a project, it will start a web server on a random port. This port is now written to the `.internal/editor.port` file.

Additionally, we added a command line option `--port` (or `-p`) to the editor executable, which allows specifying the port during launch.

Example use:
```shell
# on Windows
.\Defold.exe --port 8181

# on Linux:
./Defold --port 8181

# on macOS:
./Defold.app/Contents/MacOS/Defold --port 8181
```

Fixes #8737

### DIFF
--- a/editor/scripts/bundle.py
+++ b/editor/scripts/bundle.py
@@ -656,7 +656,7 @@ Commands:
 
     parser.add_option('--launcher', dest='launcher',
                       default = None,
-                      help = 'Specific local launcher to use when creating the bundle. Useful when testing.')
+                      help = 'Specific local launcher to use when creating the bundle, e.g. "../engine/tools/build/src/launcher/launcher". Useful when testing.')
 
     parser.add_option('--skip-tests', dest='skip_tests',
                       action = 'store_true',
--- a/editor/src/clj/editor.clj
+++ b/editor/src/clj/editor.clj
@@ -17,4 +17,4 @@
 
 (defn -main [& args]
   (println "Launching")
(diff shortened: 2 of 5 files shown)


sprocketc (dismissed):
Works great! It seems to gracefully handle various cases. 

The only thing that seemed weird, was that using an invalid port number (e.g. >65535) would throw a `port out of range` exception. It's obviously not a big deal, but I would expect your try/catch to handle this. It works fine when the port is already in use.

vlaaad:
Good catch!

AGulev:
it would be nice to update this doc as well https://github.com/defold/defold/blob/dev/editor/doc/http-api.md
btw, I see build command and console. but is there a way to get build errors?

PR #11072, opened 2025-08-09
New sound backend for iOS and macOS: AVAudio

AGulev:
This new backend replaces the deprecated OpenAL backend on iOS and macOS.

Fixes https://github.com/defold/defold/issues/10971

# Technical details

The implementation takes into account device switching, so the device restarts after an output change. For now, the sample rate is always, 48000, but in the future, it’s possible to make it dynamic.  
However, this would require a better way to notify the sound system that the sample rate has changed, or perhaps some kind of notification that the device should be recreated.

AVAudio isn’t as low-level as CoreAudio, but I’m not sure if we really need CoreAudio for now (We don't have to merge it if we decide we want CoreAudio instead, that's why this is a draft).


engine/sound/src/devices/device_avaudio.mm
```diff
@@ -0,0 +1,383 @@
...
+#endif
+        }
+    };
+
+    static uint32_t GetSystemSampleRate() {
+    // It's not used for now because we don't have a way to notify the sound system
+    // that the sample rate has changed in the middle of a session (e.g., when headphones are turned off).
+    // I’m keeping it here and forcing 48000 instead of dynamic detection for now.
+    // AVAudioEngine will resample to hardware.
+    double sr = 48000.0;
```
AGulev, line 69: For now, it forces using 48000, see comment

engine/sound/src/devices/device_avaudio.mm
```diff
@@ -0,0 +1,383 @@
...
+    //     sr = sysfmt ? sysfmt.sampleRate : 0.0;
+    // #endif
+
+    //     if (sr <= 0.0) {
+    //         sr = 44100.0;
+    //     }
+        return (uint32_t)sr;
+    }
+
+    static void DeviceAVAudioReconfigureIfNeeded(AVAudioDevice* device)
```
AGulev, line 86: device restart after device switch e.g. turn headphones on/off

AGulev:
I hear difference between OpenAL and AVAudio in tests, but I don't know what's the reason 
OpenAl:

https://github.com/user-attachments/assets/02500313-7d52-4bbd-8520-d77cb9ede11d

AVAudio:

https://github.com/user-attachments/assets/e04a496d-9d4a-406f-af43-eecf8014b66f

engine/engine/wscript
```diff
@@ -109,7 +109,7 @@ def configure(conf):
                 conf.env.append_value('LINKFLAGS', ['-l', 'c++'])
 
     elif target_os == TargetOS.IOS:
-        conf.env.append_value('LINKFLAGS', ['-framework', 'UIKit', '-framework', 'OpenGLES', '-framework', 'QuartzCore', '-framework', 'CoreGraphics', '-framework', 'OpenAL', '-framework', 'AudioToolbox'])
+        conf.env.append_value('LINKFLAGS', ['-framework', 'UIKit', '-framework', 'OpenGLES', '-framework', 'QuartzCore', '-framework', 'CoreGraphics', '-framework', 'AVFoundation'])
```
AGulev, line 112: Removed OpenAl and AudioToolbox, added AVFoundation

engine/sound/src/devices/device_avaudio.mm
AGulev: But I'm keeping this commented code here for the future when we decide to implement it

AGulev:
I tested it on both  ios and macos, didn't find any issues

engine/sound/src/devices/device_avaudio.mm
```diff
@@ -0,0 +1,392 @@
...
+        }
+    }
+
+    static void DeviceAVAudioStop(dmSound::HDevice _device)
+    {
+        assert(_device);
+        AVAudioDevice* device = (AVAudioDevice*)_device;
+        if (device->m_Started)
+        {
+            // Allow queued audio to finish similar to OpenAL stop behavior
```
AGulev, line 360: Maybe the difference that sounds like longer pause between tests between dev and this branch happens because of that (see video in the PR comment), but without that sounds obviously didn't play till the end, so I check what's OpenAL is doing to avoid it, and found that it continue consuming for a while or until the buffer fully consumed

JCash: Yeah, not sure what a best practice is here.

build_tools/waf_dynamo.py
```diff
@@ -502,7 +502,7 @@ def default_flags(self):
         if TargetOS.IOS == target_os:
             self.env.append_value('LINKFLAGS', ['-framework', 'UIKit', '-framework', 'SystemConfiguration', '-framework', 'AVFoundation'])
         else:
-            self.env.append_value('LINKFLAGS', ['-framework', 'AppKit'])
+            self.env.append_value('LINKFLAGS', ['-framework', 'AppKit', '-framework', 'AVFoundation'])
```
JCash: I think it's better to create an `FRAMEWORK_SOUND` that has this framework.
Adding it directly to linkflags, will add it to the headless engine as well.

While it matters little on iOS, since noone is going to use dmengine_headless there, I think it's good practice to show the way we like to define our features. One could imagine running a macOS server.

JCash (changes requested):
Small config change needed.
Also, don't we need AVFoundation on macOS in build_input.yml

build_tools/waf_dynamo.py
AGulev: I've made changes if I undertood it right

JCash (approved):

PR #11073, opened 2025-08-09
Fix multi-contour artifacts in DF font generator

### LINKED ISSUE #6577: DF font rendering causes artefacts when glyphs are composed of multiple vector shapes
Font consisting of multiple vector shapes:
image

Same font, with boolean union to form a single shape:
image

See this forum post for detailed writeup of the issue:
https://forum.defold.com/t/issues-with-df-fonts/63744/7

AGulev:
Fixed DF font artifacts that occur in some fonts with vector shapes overlapping each other:
CleanShot 2025-08-09 at 18 40 18@2x

Fix https://github.com/defold/defold/issues/6577


AGulev:
Few more tests:
CleanShot 2025-08-09 at 18 20 54@2x
CleanShot 2025-08-09 at 18 32 54@2x
CleanShot 2025-08-09 at 18 59 35@2x
CleanShot 2025-08-09 at 19 00 57@2x

JCash (approved):
Lgtm!

PR #11075, opened 2025-08-09
Fix selection context on outline label double click, and no selection npe

### LINKED ISSUE #11074: Double clicking on an outline label might open the last opened tab
> When I double click Gui/Nodes/Textures text to expand it, the previously open tab opens instead. 

Reported [here](https://forum.defold.com/t/defold-1-10-4-has-been-released/81160/11?u=sprocketc).

sprocketc:
Fixes #11074
Fixes #11076

### DIFF
--- a/editor/src/clj/editor/outline_view.clj
+++ b/editor/src/clj/editor/outline_view.clj
@@ -517,9 +517,9 @@
 (defn- get-selected-node-id
   [^TreeView tree-view]
   (let [selection-model (.getSelectionModel tree-view)
-        selected-items (.getSelectedItems selection-model)
-        ^TreeItem selected-item (first selected-items)]
-    (item->node-id selected-item)))
+        selected-items (.getSelectedItems selection-model)]
+    (when-let [^TreeItem selected-item (first selected-items)]
+      (item->node-id selected-item))))
 
 (defn- cancel-rename!
   [^TreeView tree-view]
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/outline_view.clj
```diff
@@ -541,7 +541,7 @@
         (< time-diff db-click-threshold)
         (do
           (cancel-rename! tree-view)
-          (ui/run-command (.getSource event) :file.open-selected))
+          (ui/run-command (.getSource event) :file.open-selected {} false (fn [] (.consume event))))
```
sprocketc, line 544: We previously used `ui/bind-double-click!` to trigger `:file.open-selected`. That handles this is a similar way, but we now also have to cancel the future event. Maybe it's possible to restructure this in order to use existing ui functions, but I am worried that it would introduce new issues.

editor/src/clj/editor/outline_view.clj
```diff
@@ -517,9 +517,9 @@
 (defn- get-selected-node-id
   [^TreeView tree-view]
   (let [selection-model (.getSelectionModel tree-view)
-        selected-items (.getSelectedItems selection-model)
-        ^TreeItem selected-item (first selected-items)]
-    (item->node-id selected-item)))
+        selected-items (.getSelectedItems selection-model)]
+    (when-let [^TreeItem selected-item (first selected-items)]
+      (item->node-id selected-item))))
```
sprocketc, line 522: To reproduce #11076, select the root node, and then use `Option+Click` to remove it from the current selection. Then click on the caret icon to collapse or expand a node that can be renamed (avoid selecting the node), and then click on its label.

vlaaad:
Ship it!

vlaaad (approved):

PR #11077, opened 2025-08-10
Make aspect ratio–related function in the `camera.*` module useful

### LINKED ISSUE #10632: Camera `auto aspect ratio` as field
The "auto aspect ratio" field affects camera behaviour, but is not available in the code. Aspect ratio is used in "screen to world" etc functions and it turns out that I can't know what behaviour the user wants and my functions won't work properly if the user ticks "auto aspect ratio" for a perspective camera.

The `camera.get_aspect_ratio()` function returns the value that is set in the component, but not automatically calculated.

- `camera.get_aspect_ratio()` returns autocalculated ratio if the box is checked.
- OR there are `camera.get_auto_aspect_ratio()/go.get(url, "auto_aspect_ratio")` functions.

Do not use the option at all and state in the docs of games/libs that the option is not supported/broken.

AGulev:
Changes in APIs:  
`camera.get_aspect_ratio()` – Now returns the effective aspect ratio (auto-calculated if auto is enabled, manual value if disabled)  
`camera.set_aspect_ratio()` – Sets the manual aspect ratio value, which is used only if auto is disabled  
`camera.get_auto_aspect_ratio()` – Returns whether auto-calculation is enabled  
`camera.set_auto_aspect_ratio()` –  Controls whether to use auto-calculation or the manual value

Fix https://github.com/defold/defold/issues/10632

### DIFF
--- a/engine/render/src/render/render_script_camera.cpp
+++ b/engine/render/src/render/render_script_camera.cpp
@@ -172,12 +172,22 @@ namespace dmRender
     }
 
     /*# get aspect ratio
+    * Gets the effective aspect ratio of the camera. If auto aspect ratio is enabled,
+    * returns the aspect ratio calculated from the current render target dimensions.
+    * Otherwise returns the manually set aspect ratio.
     *
     * @name camera.get_aspect_ratio
     * @param camera [type:url|number|nil] camera id
-    * @return aspect_ratio [type:number] the aspect ratio.
+    * @return aspect_ratio [type:number] the effective aspect ratio.
     */
-    GET_CAMERA_DATA_PROPERTY_FN(AspectRatio, lua_pushnumber);
+    static int RenderScriptCamera_GetAspectRatio(lua_State* L)
+    {
+        DM_LUA_STACK_CHECK(L, 1);
+        RenderCamera* camera = CheckRenderCamera(L, 1, g_RenderScriptCameraModule.m_RenderContext);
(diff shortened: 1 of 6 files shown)


engine/render/src/render/render_script_camera.cpp
```diff
@@ -274,6 +322,8 @@ namespace dmRender
         {"set_far_z",               RenderScriptCamera_SetFarZ},
         {"get_orthographic_zoom",   RenderScriptCamera_GetOrthographicZoom},
         {"set_orthographic_zoom",   RenderScriptCamera_SetOrthographicZoom},
+        {"get_auto_aspect_ratio",   RenderScriptCamera_GetAutoAspectRatio},
+        {"set_auto_aspect_ratio",   RenderScriptCamera_SetAutoAspectRatio},
```
AGulev, line 326: I though initially that it should be is_auto_aspect_ration because this is a checkbox? as in gui: https://defold.com/ref/gui/#gui.is_enabled:node-recursive

But then a conversation poped up in my head that when I implemented https://defold.com/ref/gui/#gui.get_visible:node 
we agree that we always use get, and this `is_` in gui is the only place

engine/render/src/render/render_script_camera.cpp
```diff
@@ -172,12 +172,22 @@ namespace dmRender
     }
 
     /*# get aspect ratio
+    * Gets the effective aspect ratio of the camera. If auto aspect ratio is enabled,
+    * returns the aspect ratio calculated from the current render target dimensions.
+    * Otherwise returns the manually set aspect ratio.
```
AGulev, line 177: better explanation what it returns

britzl (approved):

PR #11078, opened 2025-08-10
Show an error message if the decoder for a used sound isn't included in the App Manifest

AGulev:
Fixes an issue where the engine crashes if an `opus` file is used without including the OPUS decoder in the App Manifest. Now the engine shows an error to help figure out what's going on.

Fix https://github.com/defold/defold/issues/11069

### DIFF
--- a/engine/sound/src/sound.cpp
+++ b/engine/sound/src/sound.cpp
@@ -749,7 +749,7 @@ namespace dmSound
                 if (r == dmSoundCodec::RESULT_UNSUPPORTED) {
                     const char* name = dmHashReverseSafe64(sound_data->m_NameHash);
                     const char* format_str = dmSoundCodec::FormatToString(codec_format);
-                    dmLogWarning("Sound '%s' uses %s, but no decoder was found. Ensure the codec is included in your App Manifest.", name, format_str);
+                    dmLogError("Sound '%s' uses %s, but no decoder was found. Ensure the codec is included in your App Manifest.", name, format_str);
                 } else {
                     dmLogError("Failed to decode sound %s: (%d)", dmHashReverseSafe64(sound_data->m_NameHash), r);
                 }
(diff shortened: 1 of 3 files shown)


engine/sound/src/sound_decoder.cpp
```diff
@@ -72,8 +72,6 @@ namespace dmSoundCodec
 
             decoder = decoder->m_Next;
         }
-
-        assert(best != 0);
```
AGulev, line 76: we should handle it in `sound.cpp` instead of assert

engine/sound/src/sound.cpp
```diff
@@ -746,7 +746,13 @@ namespace dmSound
 
             dmSoundCodec::Result r = dmSoundCodec::NewDecoder(ss->m_CodecContext, codec_format, sound_data, &decoder);
             if (r != dmSoundCodec::RESULT_OK) {
-                dmLogError("Failed to decode sound %s: (%d)", dmHashReverseSafe64(sound_data->m_NameHash), r);
+                if (r == dmSoundCodec::RESULT_UNSUPPORTED) {
+                    const char* name = dmHashReverseSafe64(sound_data->m_NameHash);
+                    const char* format_str = dmSoundCodec::FormatToString(codec_format);
+                    dmLogWarning("Sound '%s' uses %s, but no decoder was found. Ensure the codec is included in your App Manifest.", name, format_str);
```
AGulev: I hope this message explains what's wrong, should it be an error?

JCash: I do think it should be an error.
If we get to this point, we've failed to alert them in the content pipeline.

JCash (dismissed):

engine/sound/src/sound.cpp
AGulev: Fixed!

AGulev:
@JCash could you please take a look, if it's fine I'll merge it.

JCash (approved):

PR #11083, opened 2025-08-11
Enable `go.exists()` to query the existence of game objects across collections

### LINKED ISSUE #10795: Enable `go.exists` to query the existence of game objects across collections
`go.exists` in most recent versions of defold silently returns `false` when asked to query another collection. Fixed here to behave like other `go.*` functions and throw an error:

https://github.com/defold/defold/issues/10652

However, `go.exists` would be far more useful if it could query other collections - is there a technical reason it can't? Quoting @britzl:

> It is questionable why it is not allowed to check for game objects in a different collection, but that is for another discussion.

AGulev:
With this fix, it will be possible to use the `go.exists()` function to check if a game object exists in another collection.

Fix https://github.com/defold/defold/issues/10795

### DIFF
--- a/engine/engine/src/test/collection_proxy/cross_collection_exists.script
+++ b/engine/engine/src/test/collection_proxy/cross_collection_exists.script
@@ -1,0 +1,58 @@
+-- Copyright 2020-2025 The Defold Foundation
+-- Copyright 2014-2020 King
+-- Copyright 2009-2014 Ragnar Svensson, Christian Murray
+-- Licensed under the Defold License version 1.0 (the "License"); you may not use
+-- this file except in compliance with the License.
+--
+-- You may obtain a copy of the License, together with FAQs at
+-- https://www.defold.com/license
+--
+-- Unless required by applicable law or agreed to in writing, software distributed
+-- under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+-- CONDITIONS OF ANY KIND, either express or implied. See the License for the
+-- specific language governing permissions and limitations under the License.
+
+function init(self)
+    self.test_phase = "loading"
+    msg.post("#target_proxy", "load")
+end
+
+function on_message(self, message_id, message, sender)
+    if message_id == hash("proxy_loaded") then
+        if self.test_phase == "loading" then
+            self.test_phase = "testing"
+            msg.post("#target_proxy", "enable")
+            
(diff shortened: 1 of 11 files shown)


engine/gameobject/src/gameobject/gameobject_script.cpp
```diff
@@ -2072,20 +2078,38 @@ namespace dmGameObject
...
-        dmGameObject::HInstance target_instance = dmGameObject::GetInstanceFromIdentifier(dmGameObject::GetCollection(instance), target.m_Path);
+        dmGameObject::HInstance target_instance = 0;
+        
+        // Check if target is in the same collection
+        if (receiver.m_Socket == dmGameObject::GetMessageSocket(instance->m_Collection->m_HCollection))
+        {
+            // Same collection - use current collection
+            target_instance = dmGameObject::GetInstanceFromIdentifier(dmGameObject::GetCollection(instance), target.m_Path);
+        }
+        else
```
AGulev, line 2098: The actual fix here

engine/gameobject/src/gameobject/test/script/exists.script
```diff
@@ -23,8 +23,6 @@ function init(self)
     if pcall(go.exists, nil) then
         assert(false, "Error expected when calling go.exists with nil")
     end
-    if pcall(go.exists, "foo:/bar") then
-        assert(false, "Error expected when calling go.exists with another collection")
-    end
+    assert(not go.exists("foo:/bar"), "Expected go 'foo:/bar' (non-existent collection) to not exist")
```
AGulev, line 26: Old test can check only negative case, I added a new one for positive case

engine/engine/src/test/collection_proxy/cross_collection_exists.script
```diff
@@ -0,0 +1,58 @@
...
+function init(self)
+    self.test_phase = "loading"
+    msg.post("#target_proxy", "load")
+end
+
+function on_message(self, message_id, message, sender)
+    if message_id == hash("proxy_loaded") then
+        if self.test_phase == "loading" then
+            self.test_phase = "testing"
+            msg.post("#target_proxy", "enable")
```
AGulev, line 24: Setup to test all the `go.exist()` cases including inter-collection checking

britzl (approved):

PR #11107, opened 2025-08-17
Add support for building Defold on Nix/OS

atomicptr:
NixOS differs significantly from other Linux distributions making it difficult to build the engine without providing some special configuration which is what this PR does.

To explain how Nix works simplified: It generally keeps different versions of dependencies around in `/nix/store` which are linked together per application in a unique environment depending on the applications need, which does however mean that even if you install certain dependencies they are not directly available to you in any path you'd expect (since NixOS doesn't support the default FHS ([Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html)) by default meaning in order to build the engine we have to make our own little environment with the required dependencies, effectively replicating what `./scripts/build.py shell` does. I was initially considering expanding the `./scripts/build.py shell` script, but deemed this to be a fairly invasive change for a platform that probably barely anyone here uses. 

This PR provides a nix-shell script that provides a working environment to build the engine without installing or setting up anything (assuming you have nix available), you can simply run:

```bash
$ ./scripts/nix/run.sh
$ ./scripts/build.py install_ext
$ ./scripts/build.py build_engine --skip-tests -- --skip-build-tests
$ cd editor
$ lein init <sha1>
$ lein run
```

This will build the engine, the editor and as far as I've seen so far everything works as expected

One remaining issue however, 1 test from build_engine fails and I haven't yet been able to figure out why:

```
...
dmSSDPTest took 0.551522 s
dmSSDPTest.Search failed
Ran 7 tests, with 14 assertions in 0.551522 s
6 tests passed, 0 skipped and 1 tests FAILED
test failed test_ssdp

[exec] /nix/store/yba3raw18d8vb5dnb4bczxd0063yxk81-python3-3.12.11-env/bin/python3.12 /home/christopher/dev/_clones/defold/tmp/dynamo_home/ext/bin/waf --prefix=/home/christopher/dev/_clones/defold/tmp/dynamo_home distclean configure build install --platform=x86_64-linux
Error:
Traceback (most recent call last):
  File "/home/christopher/dev/_clones/defold/./scripts/build.py", line 2803, in <module>
    f()
  File "/home/christopher/dev/_clones/defold/./scripts/build.py", line 1423, in build_engine
    self._build_engine_lib(args, lib, host)
  File "/home/christopher/dev/_clones/defold/./scripts/build.py", line 1369, in _build_engine_lib
    run.env_command(self._form_env(), args + plf_args + self.waf_options + skip_build_tests, cwd = cwd)
  File "/home/christopher/dev/_clones/defold/build_tools/run.py", line 98, in env_command
    return _exec_command(args, shell = False, stdout = None, env = env, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/christopher/dev/_clones/defold/build_tools/run.py", line 75, in _exec_command
    raise e
run.ExecException: 1
```

### DIFF
--- a/scripts/nix/shell.nix
+++ b/scripts/nix/shell.nix
@@ -1,0 +1,56 @@
+{
+  pkgs ? import <nixpkgs> { },
+}:
+
+pkgs.mkShell {
+  buildInputs = with pkgs; [
+    clang
+    cmake
+    freeglut
+    git
+    glib
+    leiningen
+    libGL
+    libGLU
+    ninja
+    openal
+    openjdk21
+    pkg-config
+    protobuf
+    xorg.libX11
+    xorg.libXcursor
+    xorg.libXi
+    xorg.libXinerama
+    xorg.libXrandr
+    xorg.libXtst
+    xorg.libXxf86vm
+
+    (python313.withPackages (
+      ps: with ps; [
+        pip
+        virtualenv
+      ]
+    ))
+  ];
+
+  shellHook = ''
+    export CC=clang
+    export CXX=clang++
+    export LD_LIBRARY_PATH=${
+      pkgs.lib.makeLibraryPath [
+        pkgs.freeglut
+        pkgs.glib
+        pkgs.libGL
+        pkgs.libGLU
+        pkgs.openal
+        pkgs.xorg.libX11
+        pkgs.xorg.libXcursor
+        pkgs.xorg.libXi
+        pkgs.xorg.libXinerama
+        pkgs.xorg.libXrandr
+        pkgs.xorg.libXtst
+        pkgs.xorg.libXxf86vm
+      ]
+    }:$LD_LIBRARY_PATH
+  '';
+}
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -631,7 +631,7 @@ class Configuration(object):
         if self.package_path is None:
             print("No package path provided. Use either --package-path option or DM_PACKAGES_URL environment variable")
             sys.exit(1)
-    
+
     def install_waf(self):
         def make_package_path(root, platform, package):
             return join(root, 'packages', package) + '-%s.tar.gz' % platform
@@ -646,7 +646,7 @@ class Configuration(object):
 
         def make_package_paths(root, platform, packages):
             return [make_package_path(root, platform, package) for package in packages]
-        
+
         self.install_waf()
 
         print("Installing common packages")
@@ -1881,7 +1881,10 @@ class Configuration(object):
 
         args = [SHELL, '-l']
 
-        process = subprocess.Popen(args, env=self._form_env(), shell=True)
+        if os.path.exists("/nix"):
+            args = ["nix-shell", os.path.join("scripts", "nix", "shell.nix"), "--run", " ".join(args)]
+
+        process = subprocess.Popen(args, env=self._form_env())
         try:
             output = process.communicate()[0]
         except KeyboardInterrupt as e:
@@ -2480,7 +2483,7 @@ class Configuration(object):
                     _threads[i].join()
                     self._log('Uploaded #%d %s -> %s' % (i + 1, path, url))
 
-                
+
                 if len(list(mp.parts.all())) == chunkcount:
                     try:
                         parts = []


scripts/nix/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+You need [Nix or NixOS](https://nixos.org/download/)
+
+Next run the nix shell using:
+
+```bash
+$ ./scripts/nix/run.sh
+```
+
+After that just use the `./scripts/build.py` scripts (except for shell) [as instructed in the build instructions](https://github.com/defold/defold/blob/dev/README_BUILD.md).
```
britzl: What would happen if I did run `./scripts/build.py shell` on Nix?

scripts/nix/shell.nix
```diff
@@ -0,0 +1,59 @@
...
+    openjdk21
+    pkg-config
+    protobuf
+    xorg.libX11
+    xorg.libXcursor
+    xorg.libXi
+    xorg.libXinerama
+    xorg.libXrandr
+    xorg.libXtst
+    xorg.libXxf86vm
```
britzl, line 26: How was this list produced?

scripts/nix/shell.nix
```diff
@@ -0,0 +1,59 @@
...
+        pkgs.libGL
+        pkgs.libGLU
+        pkgs.openal
+        pkgs.xorg.libX11
+        pkgs.xorg.libXcursor
+        pkgs.xorg.libXi
+        pkgs.xorg.libXinerama
+        pkgs.xorg.libXrandr
+        pkgs.xorg.libXtst
+        pkgs.xorg.libXxf86vm
```
britzl, line 52: And this one?

scripts/nix/README.md
atomicptr: Instead of the script I provided or in addition?

Initially I was looking into what would be needed to make the setup work and kept running into various errors, e.g. one of the first one you'll meet is for instance this one

image

Because pip is not available in the standard Python installation provided by Nix, this is also why I  added this configuration bit:

```nix
    (python312.withPackages (
      ps: with ps; [
        pip
        virtualenv
      ]
    ))
```

Although even if you manage to get past this, you will not be able to compile the game engine since pretty much no dependencies can be inferred by the build tool by default . (You also won't be able to start the editor because the X11 shared libs aren't accessible)

If you meant in addition:

The engine will build but we won't be able to start the editor because it will not find the required shared libraries (X11, etc) because the shell script ignores the LD_LIBRARY_PATH we set up from the nix shell script resulting in an error like this:

image

scripts/nix/shell.nix
atomicptr: Partially from reading the documentation of what libraries are necessary and partially from trial and error, e.g. I try to build the engine and it complains about not having library X, so I add library X and rebuild until I got it working

scripts/nix/shell.nix
atomicptr: Same as above, just this is for what libraries we'll be adding to the LD_LIBRARY_PATH

scripts/nix/README.md
britzl: > If you meant in addition

I meant this. I was just curious if it would make sense to add some check to the `shell` command if this was accidentally run from Nix. Perhaps there could be an early exit?

britzl: Maybe we could check if running from Nix and have the `shell` command do the equivalent of what `run.sh` does?

scripts/nix/shell.nix
britzl: Ok, somewhat similar to https://github.com/defold/defold/blob/dev/ci/ci.py#L188-L207 which we use on CI

scripts/nix/shell.nix
```diff
@@ -0,0 +1,59 @@
...
+    protobuf
+    xorg.libX11
+    xorg.libXcursor
+    xorg.libXi
+    xorg.libXinerama
+    xorg.libXrandr
+    xorg.libXtst
+    xorg.libXxf86vm
+
+    (python312.withPackages (
```
JCash: Python 3.12 seems quite old?

JCash (dismissed):

scripts/nix/shell.nix
atomicptr: I updated the Python version to 3.13

scripts/nix/README.md
atomicptr: I added a check for NixOS into the build.py file (checking if /nix exists) and execute nix-shell from there if we're on NixOS. This also allowed me to remove the special Defold stuff again from the shell.nix file

ekharkunov (approved):

britzl (approved):

PR #11108, opened 2025-08-18
Editor: Add fn/and and fn/or functions

matgis:
### Technical changes
* Add `fn/and` and `fn/or`, which are functions that behave exactly like the `and` and `or` macros in `clojure.core`.
* Update existing code to use these shared functions.


vlaaad (approved):
Ohhh that's nice!

PR #11114, opened 2025-08-18
Prevent duplicate connections from occurring on copy/pasting nodes

vlaaad:
Fixes #11103

### DIFF
--- a/editor/src/clj/editor/collection.clj
+++ b/editor/src/clj/editor/collection.clj
@@ -160,27 +160,29 @@
   ([self-id child-id]
    (tx-attach-go-referenced-go self-id child-id true))
   ([self-id child-id resolve-id]
-   (g/expand-ec
-     (fn tx-attach [evaluation-context]
-       (let [coll-id (core/scope-of-type (:basis evaluation-context) self-id CollectionNode)]
-         (concat
-           (when resolve-id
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/collection.clj
```diff
@@ -160,27 +160,29 @@
...
-           (child-go-go self-id child-id)))))))
+   (concat
+     (g/expand-ec
+       (fn tx-resolve-id [evaluation-context]
+         (let [coll-id (core/scope-of-type (:basis evaluation-context) self-id CollectionNode)]
+           (concat
+             (when resolve-id
+               (g/update-property child-id :id id/resolve (g/node-value coll-id :ids evaluation-context)))
+             (attach-coll-ref-go coll-id child-id)))))
+     (child-go-go self-id child-id))))
```
vlaaad, line 171: The solution is to move `child-go-go` out of `g/expand-ec` so that outline code can filter it out [here](https://github.com/defold/defold/blob/646a1acb3537f1c42c33da22f73f3fe04808f5dd/editor/src/clj/editor/outline.clj#L110). I checked if we do self<->child connection inside `g/expand-ec` inside `:tx-attach-fn` anywhere else, but this is the only place.

matgis (approved):
Nice find! ❤️

PR #11118, opened 2025-08-19
Editor: Pretty-print vecmath types and Arcs in Reveal

matgis:
### Technical changes
* Added Reveal stream handlers and readers for all floating-point `javax.vecmath` types.
* Added Reveal stream handler and reader for `gt/Arc`.


vlaaad (approved):
Nice!

PR #11133, opened 2025-08-25
Remove transparent grid pass

sprocketc:
### Technical changes
Now that the `infinity-grid` pass is on top of the `transparent` pass, I believe the remaining transparent pass of the grid doesn't make sense. It also seems to create this weird effect on 3D mode
Before
Screenshot From 2025-08-25 10-54-45
After
Screenshot From 2025-08-25 10-55-09


matgis (approved):
Nice to see this fixed! ❤️

PR #11142, opened 2025-08-26
Fix template property flattening regression when building

### LINKED ISSUE #7213: The GUI template location is not applied when changing the layout if there has been no change in the position of the parent node.

matgis:
Fixed a regression when building GUI scenes that would cause certain properties (such as transform-related ones) to not be correctly applied to the imported nodes unless they had layout overrides.

Fixes #7213
Fixes #11110

### Technical changes
* Added `update-vals` and `update-vals-kv` to `coll` module. Our `update-vals` retains the input collection type and accepts additional arguments to the transform function.
* Added `map-vals` and `map-vals-kv` to `coll` module.
* Added extensive tests around template gui node flattening during build.

### DIFF
--- a/editor/src/clj/editor/gui.clj
+++ b/editor/src/clj/editor/gui.clj
@@ -3209,24 +3209,28 @@
     :else
     (dissoc node-desc :overridden-fields)))
 
-(defn- make-rt-layout-desc [layout-name decorated-node-msgs]
+(defn- make-rt-layout-desc [layout-name decorated-node-msgs id->node-desc-for-default-layout]
+  {:pre [(map? id->node-desc-for-default-layout)]}
   (protobuf/make-map-without-defaults Gui$SceneDesc$LayoutDesc
     :name layout-name
     :nodes (coll/transfer decorated-node-msgs []
              (keep
                (fn [{:keys [layout->prop->value] :as decorated-node-msg}]
                  {:pre [(map? layout->prop->value)]}
-                 (let [default-prop->value (layout->prop->value "")
-                       prop->value (layout->prop->value layout-name)]
-                   (assert (map? default-prop->value))
+                 (let [prop->value (layout->prop->value layout-name)]
                    (assert (map? prop->value))
-                   (when (not= default-prop->value prop->value)
-                     (some->> (-> decorated-node-msg
-                                  (select-keys [:custom-type :id :parent :template-node-child :type])
-                                  (into (map prop-entry->pb-field-entry)
-                                        prop->value)
-                                  (node-desc->rt-node-desc layout-name))
-                              (protobuf/clear-defaults Gui$NodeDesc)))))))))
+                   (when-some [node-desc-for-layout
+                               (some->> (-> decorated-node-msg
+                                            (select-keys [:custom-type :id :parent :template-node-child :type])
+                                            (into (map prop-entry->pb-field-entry)
+                                                  prop->value)
+                                            (node-desc->rt-node-desc layout-name))
+                                        (protobuf/clear-defaults Gui$NodeDesc))]
+                     (let [id (:id node-desc-for-layout)
+                           node-desc-for-default-layout (id->node-desc-for-default-layout id)]
+                       (assert (map? node-desc-for-default-layout))
+                       (when (not= node-desc-for-default-layout node-desc-for-layout)
+                         node-desc-for-layout)))))))))
 
 (g/defnk produce-build-targets [_node-id build-errors resource pb-msg dep-build-targets template-build-targets layout-names node-msgs]
   ;; Built Gui$NodeDescs should be fully-formed, since no additional merging of
@@ -3244,10 +3248,6 @@
     (g/precluding-errors build-errors
       (let [template-build-targets (flatten template-build-targets)
 
-            rt-layout-descs
-            (mapv #(make-rt-layout-desc % node-msgs)
-                  layout-names)
-
             rt-node-descs
             (coll/transfer node-msgs []
               (keep
@@ -3255,6 +3255,12 @@
                   (-> decorated-node-msg
                       (dissoc :layout->prop->override :layout->prop->value)
                       (node-desc->rt-node-desc "")))))
+
+            id->node-desc-for-default-layout (coll/pair-map-by :id rt-node-descs)
+
+            rt-layout-descs
+            (mapv #(make-rt-layout-desc % node-msgs id->node-desc-for-default-layout)
+                  layout-names)
 
             rt-pb-msg
             (protobuf/assign-repeated pb-msg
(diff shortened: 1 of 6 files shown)


editor/src/clj/editor/gui.clj
```diff
@@ -3209,24 +3209,28 @@
...
+                                            (select-keys [:custom-type :id :parent :template-node-child :type])
+                                            (into (map prop-entry->pb-field-entry)
+                                                  prop->value)
+                                            (node-desc->rt-node-desc layout-name))
+                                        (protobuf/clear-defaults Gui$NodeDesc))]
+                     (let [id (:id node-desc-for-layout)
+                           node-desc-for-default-layout (id->node-desc-for-default-layout id)]
+                       (assert (map? node-desc-for-default-layout))
+                       (when (not= node-desc-for-default-layout node-desc-for-layout)
+                         node-desc-for-layout)))))))))
```
matgis, line 3233: The fix: We were only looking at property overrides to determine if we need to include a layout-specific `NodeDesc` for the node in the build output. However, template nodes apply their properties to the imported children before removing themselves from the built hierarchy, and we weren't seeing those property changes. In the new code, we apply the template flattening before this and compare the output of the layout-specific `NodeDesc` to that instead,

editor/src/clj/internal/node.clj
```diff
@@ -1029,7 +1029,7 @@
                                                                (update :default f)
 
                                                                (some? (-> propdef :dynamics))
-                                                               (update :dynamics (partial util/map-vals f)))]))
+                                                               (update :dynamics (partial coll/map-vals f)))]))
```
matgis, line 1032: I've moved `util/map-vals` into `util.coll`, but kept `util/map-vals` as a macro so this PR wouldn't baloon in size. However, we used it as a function in this one file.

editor/test/integration/gui_test.clj
```diff
@@ -535,13 +535,6 @@
   (-> (gui-node scene id)
       (g/node-value :text)))
 
-(defn- trans-x [root-id target-id]
-  (let [s (tree-seq fn/constantly-true :children (g/node-value root-id :scene))]
-    (when-let [^Matrix4d m (some #(and (= (:node-id %) target-id) (:transform %)) s)]
-      (let [t (Vector3d.)]
-        (.get m t)
-        (.getX t)))))
-
```
matgis, line 544: Removed unused function.

editor/test/integration/gui_test.clj
```diff
@@ -3026,3 +3019,549 @@
...
+
+(def ^:private gui-alpha-pb-field-index (gui/prop-key->pb-field-index :alpha))
+(def ^:private gui-enabled-pb-field-index (gui/prop-key->pb-field-index :enabled))
+(def ^:private gui-inherit-alpha-pb-field-index (gui/prop-key->pb-field-index :inherit-alpha))
+(def ^:private gui-layer-pb-field-index (gui/prop-key->pb-field-index :layer))
+(def ^:private gui-position-pb-field-index (gui/prop-key->pb-field-index :position))
+(def ^:private gui-rotation-pb-field-index (gui/prop-key->pb-field-index :rotation))
+(def ^:private gui-scale-pb-field-index (gui/prop-key->pb-field-index :scale))
+
+(deftest template-props-baked-into-imported-node-descs-test
```
matgis, line 3051: There are sooo many combinations to test! 😮‍💨

editor/src/clj/util/coll.clj
```diff
@@ -225,6 +225,50 @@
...
+   (when coll
+     (let [use-transient (supports-transient? coll)
+           rf (if use-transient
+                #(assoc! %1 %2 (f %3))
+                #(assoc %1 %2 (f %3)))
+           init (cond-> (empty coll)
+                        use-transient transient)]
+       (with-meta (cond-> (reduce-kv rf init coll)
+                          use-transient persistent!)
+                  (meta coll))))))
```
vlaaad: Why use this over `clojure.core/update-vals`? Is it to keep the type?

editor/src/clj/internal/node.clj
```diff
@@ -1017,7 +1017,7 @@
 
 (defn- update-fn-maps [tree f]
   (-> tree
-      (update :output (partial util/map-vals f))
+      (update :output (partial coll/map-vals f))
```
vlaaad, line 1020: If `coll` was a first argument to `map-vals` (like in `clojure.core/update-vals`), we wouldn't need to create a throw-away partial function; instead, we could use `(update :output coll/map-vals f)`

editor/src/clj/util/coll.clj
```diff
@@ -225,6 +225,50 @@
...
 
+(defn map-vals
+  "Applies f to all values in the supplied associative collection. Returns a new
+  associative collection of the same type with all the same keys, but the values
+  being the results of applying f to each previous value. Preserves metadata. If
+  coll is nil, returns nil without calling f."
+  ([f]
+   (map (fn [entry]
+          (pair (key entry)
+                (f (val entry))))))
```
vlaaad, line 281: Weird to have transducer arity here at all. It's unused, right? I prefer the semantics of `update-vals` more — it's a map-specific operation, not a collection processing operation.

vlaaad (dismissed):
LGTM! Minor nitpick about `map-vals` — I think it would be better if it didn't pretend to be a generic collection processor with a transducer arity and coll as a last argument. Its purpose is maps, right?

editor/src/clj/util/coll.clj
matgis: Ohh, I was not aware of `update-vals`! 👍 It suits my use-case perfectly, so I will change the test function to use it.

I still think `map-vals` and `map-vals-kv` has a place, though. The transducer arity is useful when you would use `(map (fn [[k v]] (pair k (f v)))`, which I've done plenty of times as part of a transformation.

However, looking at the source for `update-vals`, it also strikes me that while `map-vals` will retain the input map type, that probably doesn't work for records. I'll fix!

editor/src/clj/internal/node.clj
matgis: Yeah, `update-vals` looks like just what I need! I must have missed them adding it to `clojure.core`. It'll be a lot more readable that way. Thanks! ❤️

vlaaad (approved):
Awesome!

PR #11144, opened 2025-08-27
Fix transpilation in bob

vlaaad:


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1592,10 +1592,15 @@ private void transpileLua(IProgress monitor) throws CompileExceptionError, IOExc
...
-                            .collect(Collectors.toList());
+                    ArrayList<IResource> sources = new ArrayList<>();
+                    fileSystem.walk("", new FileSystemWalker() {
+                        @Override
+                        public void handleFile(String path, Collection<String> results) {
+                            if (path.endsWith(ext)) {
+                                sources.add(fileSystem.get(path));
+                            }
+                        }
+                    }, new ArrayList<>());
```
vlaaad, line 1603: We used to go over all `inputs` to collect the source files for transpilation, but now it's empty when we transpile. I think transpilation broke when we refactored bob from compiling everything to going through deps to find what to compile. Now I walk the fs to find the sources.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1654,7 +1653,6 @@ public void handleFile(String path, Collection<String> results) {
                                     }
                                 }
                             }, results);
-                            inputs.addAll(results);
```
vlaaad, line 1657: Similarly, I no longer force transpiled lua files to be compiled.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
AGulev, line 1603: I don't like us having this extra step outside our system with Builders, but if it's already implemented this way and works, let's keep it

AGulev (approved):

PR #11150, opened 2025-08-27
Remove hardcoded API doc paths

### LINKED ISSUE #11146: Defold c++ API Documentation Problems
I'm looking at the C++ API docs to see what's new, and I came across a few issues. I'm not sure if I should add an issue here or in the [docs](https://github.com/defold/doc) repo. 

- None of the include links are working anymore.
Example:
https://github.com/defold/defold/blob/dev/dmsdk/dlib/align.h

Image

- [dmGameSystem](https://defold.com/ref/stable/dmgamesystem-cpp) collection factory docs seem to be messed up. There are many font, material, and rendering-related entries in there. Is this normal? The title says: *"API for spawning collections from a collection factory component."*

https://defold.com/ref/stable/dmgamesystem-cpp/

Image

britzl:
Fixes #11146


build_tools/waf_dynamo.py
```diff
@@ -248,6 +251,7 @@ def _parse_source(source_path):
 
         elements = {}
         resource_path = resource.abspath()
+        relative_path = resource_path.replace(defold_root, "")[1:]
```
britzl, line 255: Get the resource path relative to the project root.

/users/bjornritzl/projects/defold/engine/dlib/src/dmsdk/align.h -> engine/dlib/src/dmsdk/align.h

engine/gamesys/src/dmsdk/gamesys/components/comp_factory.h
```diff
@@ -33,17 +32,6 @@
 
 namespace dmGameSystem
 {
-    /*# Factory component functions
```
britzl, line 36: This definition was duplicated a few lines above

engine/resource/src/dmsdk/resource/resource.h
```diff
@@ -290,13 +289,62 @@ ResourceResult ResourceRemoveFile(HResourceFactory factory, const char* path);
 ////////////////////////////////////////////////////////////////////////////////////////////////////////
 // Descriptor functions
 
+/*# get path hash of resource
```
britzl, line 292: These were previously in resource_doc.h

AGulev (dismissed):
Looks good

PR #11152, opened 2025-08-27
Disallow non-top-level go.property calls in scripts

vlaaad:
Fixes #11147

### DIFF
--- a/com.dynamo.cr/.idea/com.dynamo.cr.iml
+++ b/com.dynamo.cr/.idea/com.dynamo.cr.iml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module type="JAVA_MODULE" version="4">
-  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_17" inherit-compiler-output="true">
+  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_21" inherit-compiler-output="true">
     <exclude-output />
     <content url="file://$MODULE_DIR$">
       <sourceFolder url="file://$MODULE_DIR$/com.dynamo.cr.bob.test/src" isTestSource="true" />
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
@@ -559,4 +559,14 @@ public class LuaScannerTest {
         expected = " ";
         assertEquals(expected, scanner.getParsedLua());
     }
+
+    @Test
+    public void testGoPropertyNesting() {
+        var scanner = new LuaScanner();
+        scanner.parse("go.property('top-level', 1)\nfunction init() go.property('nested', 2) end");
+        List<Property> properties = scanner.getProperties();
+        assertEquals(2, properties.size());
+        assertPropertyStatus(properties, "top-level", Status.OK, 0);
+        assertPropertyStatus(properties, "nested", Status.INVALID_LOCATION, 1);
+    }
 }
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
@@ -14,23 +14,22 @@
 
 package com.dynamo.bob.pipeline;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import com.dynamo.bob.pipeline.LuaScanner.Property;
+import com.dynamo.bob.pipeline.LuaScanner.Property.Status;
+import org.apache.commons.io.IOUtils;
+import org.junit.Test;
 
+import javax.vecmath.Quat4d;
+import javax.vecmath.Vector3d;
+import javax.vecmath.Vector4d;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.List;
 
-import javax.vecmath.Quat4d;
-import javax.vecmath.Vector3d;
-import javax.vecmath.Vector4d;
-
-import org.apache.commons.io.IOUtils;
-import org.junit.Test;
-
-import com.dynamo.bob.pipeline.LuaScanner.Property;
-import com.dynamo.bob.pipeline.LuaScanner.Property.Status;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 public class LuaScannerTest {
 
@@ -42,19 +41,16 @@ public class LuaScannerTest {
     }
 
     private void assertValidRequire(String test, String expected) {
-        LuaScanner scanner = new LuaScanner();
-        scanner.parse(test);
-        List<String> modules = scanner.getModules();
-        assertTrue(modules.size() == 1);
-        assertEquals(expected, modules.get(0));
+        List<String> modules = LuaScanner.parse(test).modules();
+        assertEquals(1, modules.size());
+        assertEquals(expected, modules.getFirst());
     }
 
     @Test
     public void testScanner() throws Exception {
         String file = getFile("test_scanner.lua");
-        LuaScanner scanner = new LuaScanner();
-        scanner.parse(file);
-        List<String> modules = scanner.getModules();
+        var result = LuaScanner.parse(file);
+        List<String> modules = result.modules();
 
         assertEquals("a", modules.get(0));
         assertEquals("b", modules.get(1));
@@ -84,7 +80,7 @@ public class LuaScannerTest {
         assertEquals("foo14", modules.get(25));
         assertEquals("foo15", modules.get(26));
         assertEquals("foo16", modules.get(27));
-        assertTrue(modules.size() == 28);
+        assertEquals(28, modules.size());
 
         // test require detection with a trailing comment
         assertValidRequire("require \"foo.bar\" -- some comment", "foo.bar");
@@ -105,13 +101,13 @@ public class LuaScannerTest {
         assertValidRequire("require ('foo.bar') --[[ some comment]]--", "foo.bar");
 
         int linesInFile = file.split("\r\n|\r|\n").length;
-        int linesAfterScanner = scanner.getParsedLua().split("\r\n|\r|\n").length;
+        int linesAfterScanner = result.code().split("\r\n|\r|\n").length;
         assertEquals(linesInFile, linesAfterScanner);
     }
 
     private Property findProperty(List<Property> properties, String name) {
         for (Property p : properties) {
-            if (p.name != null && p.name.equals(name)) {
+            if (p.name() != null && p.name().equals(name)) {
                 return p;
             }
         }
@@ -120,26 +116,24 @@ public class LuaScannerTest {
 
     private void assertProperty(List<Property> properties, String name, Object value, int line) {
         Property property = findProperty(properties, name);
-        if (property != null && property.status == Status.OK && property.name.equals(name)) {
-            assertEquals(value, property.value);
-            assertEquals(line, property.line);
+        if (property != null && property.status() == Status.OK && property.name().equals(name)) {
+            assertEquals(value, property.value());
+            assertEquals(line, property.startLine());
             return;
         }
-        assertTrue(false);
+        fail();
     }
 
     private void assertPropertyStatus(List<Property> properties, String name, Status status, int line) {
         Property property = findProperty(properties, name);
-        assertTrue(property != null && property.status == status && property.line == line);
+        assertTrue(property != null && property.status() == status && property.startLine() == line);
     }
 
-    private List<Property> getPropertiesFromString(String s) throws IOException, LuaScanner.LuaScannerException {
-        LuaScanner scanner = new LuaScanner();
-        scanner.parse(s);
-        return scanner.getProperties();
+    private List<Property> getPropertiesFromString(String s) {
+        return LuaScanner.parse(s).properties();
     }
 
-    private List<Property> getPropertiesFromFile(String path) throws IOException, LuaScanner.LuaScannerException {
+    private List<Property> getPropertiesFromFile(String path) throws IOException {
         return getPropertiesFromString(getFile(path));
     }
 
@@ -153,7 +147,7 @@ public class LuaScannerTest {
         assertProperty(properties, "prop3", Double.valueOf(0), 14);
         assertProperty(properties, "prop4", Double.valueOf(0), 15);
         assertProperty(properties, "prop5", Double.valueOf(0), 16);
-        assertEquals(Status.INVALID_ARGS, properties.get(5).status);
+        assertEquals(Status.INVALID_ARGS, properties.get(5).status());
         assertPropertyStatus(properties, "three_args", Status.INVALID_VALUE, 19);
         assertPropertyStatus(properties, "unknown_type", Status.INVALID_VALUE, 20);
     }
@@ -161,27 +155,26 @@ public class LuaScannerTest {
     @Test
     public void testPropsStripped() throws Exception {
         String source = getFile("test_props.lua");
-        LuaScanner scanner = new LuaScanner();
-        source = scanner.parse(source);
-        List<Property> properties = scanner.getProperties();
+        var result = LuaScanner.parse(source);
+        source = result.code();
+        List<Property> properties = result.properties();
         assertEquals(8, properties.size());
         assertProperty(properties, "prop1", Double.valueOf(0), 10);
         assertProperty(properties, "prop2", Double.valueOf(0), 13);
         assertProperty(properties, "prop3", Double.valueOf(0), 14);
         assertProperty(properties, "prop4", Double.valueOf(0), 15);
         assertProperty(properties, "prop5", Double.valueOf(0), 16);
-        assertEquals(Status.INVALID_ARGS, properties.get(5).status);
+        assertEquals(Status.INVALID_ARGS, properties.get(5).status());
         assertPropertyStatus(properties, "three_args", Status.INVALID_VALUE, 19);
         assertPropertyStatus(properties, "unknown_type", Status.INVALID_VALUE, 20);
 
         int linesInSource = source.split("\r\n|\r|\n").length;
-        int linesAfterScanner = scanner.getParsedLua().split("\r\n|\r|\n").length;
+        int linesAfterScanner = result.code().split("\r\n|\r|\n").length;
         assertEquals(linesInSource, linesAfterScanner);
-        
+
         // parse the already stripped source
         // there should be no properties left
-        scanner.parse(source);
-        properties = scanner.getProperties();
+        properties = LuaScanner.parse(source).properties();
         assertEquals(0, properties.size());
     }
 
@@ -209,10 +202,9 @@ public class LuaScannerTest {
 
     @Test
     public void testPropsUrl() throws Exception {
-        LuaScanner scanner = new LuaScanner();
-        scanner.parse(getFile("test_props_url.lua"));
-        List<Property> properties = scanner.getProperties();
-        List<LuaScanner.LuaScannerException> exceptions = scanner.exceptions;
+        var result = LuaScanner.parse(getFile("test_props_url.lua"));
+        var properties = result.properties();
+        var errors = result.errors();
 
         assertEquals(11, properties.size());
         assertProperty(properties, "prop1", "url", 0);
@@ -223,31 +215,32 @@ public class LuaScannerTest {
         assertProperty(properties, "prop6", "socket:/path/to/object#fragment", 5);
         assertProperty(properties, "prop7", "socket-hash:/path/to/object-hash#fragment-hash", 6);
         assertProperty(properties, "prop8", ":#", 7);
-        // prop9, prop10, prop11 throw exception and line number starts from 1
-        assertEquals(3, exceptions.size());
-        assertEquals(9, exceptions.get(0).getLineNumber());
-        assertEquals(10, exceptions.get(1).getLineNumber());
-        assertEquals(11, exceptions.get(2).getLineNumber());
+
+        // prop9, prop10, prop11 throw exception and line number starts from 0
+        assertEquals(3, errors.size());
+        assertEquals(8, errors.get(0).startLine());
+        assertEquals(9, errors.get(1).startLine());
+        assertEquals(10, errors.get(2).startLine());
     }
 
     @Test
     public void testPropsVec3() throws Exception {
         List<Property> properties = getPropertiesFromFile("test_props_vec3.lua");
 
-        assertEquals(4, properties.size());
+        assertEquals(3, properties.size());
         assertProperty(properties, "prop1", new Vector3d(), 0);
         assertProperty(properties, "prop2", new Vector3d(1, 2, 3), 1);
-        assertProperty(properties, "prop4", new Vector3d(2, 2, 2), 3);
+        assertPropertyStatus(properties, "prop3", Status.INVALID_VALUE, 2);
     }
 
     @Test
     public void testPropsVec4() throws Exception {
         List<Property> properties = getPropertiesFromFile("test_props_vec4.lua");
 
-        assertEquals(4, properties.size());
+        assertEquals(3, properties.size());
         assertProperty(properties, "prop1", new Vector4d(), 0);
         assertProperty(properties, "prop2", new Vector4d(1, 2, 3, 4), 1);
-        assertProperty(properties, "prop4", new Vector4d(2, 2, 2, 2), 3);
+        assertPropertyStatus(properties, "prop3", Status.INVALID_VALUE, 2);
     }
 
     @Test
@@ -284,22 +277,19 @@ public class LuaScannerTest {
     @Test
     public void testCommentRemoving() throws Exception {
         // few comments
-        String luaCode = 
+        String luaCode =
             "local var = 1\n" +
             "-- comment 1\n" +
             "-- comment 2\n" +
             "-- comment 3\n" +
             "local var2 = 2\n";
-        LuaScanner scanner = new LuaScanner();
-        scanner.parse(luaCode);
-        String parsed = scanner.getParsedLua();
-        String expected = 
+        String expected =
             "local var = 1\n" +
             "\n" +
             "\n" +
             "\n" +
             "local var2 = 2\n";
-        assertEquals(expected, parsed);
+        assertEquals(expected, LuaScanner.parse(luaCode).code());
 
         // Function with comment
         luaCode =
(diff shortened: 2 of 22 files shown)


editor/src/clj/editor/code/script_compilation.clj
```diff
@@ -262,36 +278,36 @@
...
-                (map (fn [{:keys [name type value]}]
-                       (let [go-prop-type (script-property-type->go-prop-type type)
-                             go-prop-value (properties/clj-value->go-prop-value go-prop-type value)]
-                         {:id name
-                          :type go-prop-type
-                          :value go-prop-value
-                          :clj-value value}))
-                     preprocessed-script-properties)
+                  (lua-parser/lua-info workspace valid-resource-kind? reader))]
+            (g/precluding-errors (script-property-error _node-id resource (:script-properties preprocessed-lua-info))
```
vlaaad: Mostly indentation changes here because of this line

editor/src/clj/editor/lua_parser.clj
```diff
@@ -15,14 +15,52 @@
...
+                                      (.getCharPositionInLine start))
+                                    (data/->Cursor
+                                      (dec (.getLine stop))
+                                      (+ (.getCharPositionInLine stop) (.length (.getText stop)))))}))]
+      ; If there was a recognition error, we'll wrap the node in an ::error
+      ; and drop the exception in the metadata.
+      (if-let [e (.exception ^ParserRuleContext t)]
+        (with-meta (list :clj-antlr/error node)
+                   {:error (antlr.common/recognition-exception->map e)})
+        node))
```
vlaaad: I copied this code from clj-antlr, improved the performance a bit, and added cursor range metadata.

vlaaad (commented):
Both editor and Bob use the same approach to ensuring that go.property is on a top level — track the depth of `stat` block.

AGulev (dismissed):
Please add bob test as well, otherwise looks good

editor/src/clj/editor/lua_parser.clj
```diff
@@ -15,14 +15,52 @@
...
+                 (with-meta
+                   {:cursor-range (data/->CursorRange
+                                    (data/->Cursor
+                                      (dec (.getLine start))
+                                      (.getCharPositionInLine start))
+                                    (data/->Cursor
+                                      (dec (.getLine stop))
+                                      (+ (.getCharPositionInLine stop) (.length (.getText stop)))))}))]
+      ; If there was a recognition error, we'll wrap the node in an ::error
+      ; and drop the exception in the metadata.
```
matgis: Style: Use double-semicolon for comments (also below).

editor/src/clj/editor/lua_parser.clj
```diff
@@ -15,14 +15,52 @@
...
+  (if (instance? ParserRuleContext t)
+    (let [^ParserRuleContext t t
+          start (.getStart t)
+          stop (.getStop t)
+          node (cond->
+                 (cons
+                   (antlr.common/fast-keyword (antlr.common/parser-rule-name p (.getRuleIndex t)))
+                   (mapv (fn [^long i]
+                           (sexpr (.getChild t i) p))
+                         (range (.getChildCount t))))
```
matgis: I think we could get rid of the `cons` here? 🤔
```suggestion
                 (into [(antlr.common/fast-keyword (antlr.common/parser-rule-name p (.getRuleIndex t)))]
                       (map (fn [^long i]
                              (sexpr (.getChild t i) p)))
                       (range (.getChildCount t)))
```

AGulev (dismissed):

editor/src/clj/editor/lua_parser.clj
```diff
@@ -15,14 +15,52 @@
...
             [clojure.set :as set]
             [clojure.string :as string]
+            [editor.code.data :as data]
             [editor.math :as math]
             [editor.workspace :as workspace])
-  (:import [org.antlr.v4.runtime BufferedTokenStream Token]))
-
-(def real-lua-parser (antlr/parser (slurp (io/resource "Lua.g4")) {:throw? false}))
+  (:import [org.antlr.v4.runtime BufferedTokenStream Parser ParserRuleContext Token]
+           [org.antlr.v4.runtime.tree ParseTree]))
```
matgis: We should take this opportunity to add our usual compile directives and fix any performance warnings that crop up!
```clj
```

editor/test/editor/lua_parser_test.clj
```diff
@@ -407,9 +407,19 @@
...
              (src->properties workspace "go.property(\"\", 0.0)")))
-      (is (= [{:status :invalid-value :name "test"}]
-             (src->properties workspace "go.property(\"test\", \"foo\")"))))))
+      (is (= [{:status :invalid-value
+               :name "test"}]
+             (src->properties workspace "go.property(\"test\", \"foo\")")))
+      (is (= [{:name "nested"
+               :type :script-property-type-boolean
+               :value false
+               :status :invalid-location}]
```
matgis: Consistency: Move the `:status` entry to the top of the map like it is for the other test cases?

editor/test/editor/lua_parser_test.clj
```diff
@@ -253,9 +253,9 @@
   (let [fuzz-codes (take 100 (repeatedly #(RandomStringUtils/random (rand-int 400))))]
     (doall (map #(do (reset! last-fuzz %) (lua-info %)) fuzz-codes))
 
-  (let [original (slurp (io/resource "build_project/SideScroller/spaceship/spaceship.script"))
-        broken-versions (map soil (repeat 100 original))]
-    (doall (map #(do (reset! last-broken %) (lua-info %)) broken-versions)))))
+    (let [original (slurp (io/resource "build_project/SideScroller/spaceship/spaceship.script"))
+          broken-versions (map soil (repeat 100 original))]
+      (doall (map #(do (reset! last-broken %) (lua-info %)) broken-versions)))))
```
matgis: I think the intent was to have the `let` blocks separate? Seems the indentation was correct, but the parens were wrong? 🤔

matgis (changes requested):
While the editor seems to catch `go.property` declarations inside functions, it allows all of these at top-level without errors:
```lua
do
  go.property("atlas", resource.atlas("/main/game.atlas"))
end

while false do
  go.property("atlas", resource.atlas("/main/game.atlas"))
end

for name, path in pairs({atlas = "/main/game.atlas"}) do
  go.property(name, resource.atlas(path))
end
```
I've not done extensive testing across all block constructs, so there may be others. Maybe we need some more tests in the editor? These are all correctly reported as errors if I build with Bob, so that part works!

editor/src/clj/editor/lua_parser.clj
vlaaad: I wanted to preserve the data shape to prevent errors. Other code in this file checks if the node is `seq?`, for example.

com.dynamo.cr/.idea/com.dynamo.cr.iml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module type="JAVA_MODULE" version="4">
-  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_17" inherit-compiler-output="true">
+  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_21" inherit-compiler-output="true">
```
vlaaad, line 3: We are on JDK 21! I also had to change `compiler.xml` in this folder (it's gitignored) to use source/target language level 21 for compilation

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
```diff
@@ -42,19 +41,16 @@ String getFile(String file) throws IOException {
     }
 
     private void assertValidRequire(String test, String expected) {
-        LuaScanner scanner = new LuaScanner();
-        scanner.parse(test);
-        List<String> modules = scanner.getModules();
-        assertTrue(modules.size() == 1);
-        assertEquals(expected, modules.get(0));
+        List<String> modules = LuaScanner.parse(test).modules();
```
vlaaad, line 44: I heavily refactored LuaScanner. Now we use a static method that returns a result once instead of creating mutable instances of LuaScanner that change on every parsing call.

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
```diff
@@ -223,31 +215,32 @@ public void testPropsUrl() throws Exception {
...
-        assertEquals(3, exceptions.size());
-        assertEquals(9, exceptions.get(0).getLineNumber());
-        assertEquals(10, exceptions.get(1).getLineNumber());
-        assertEquals(11, exceptions.get(2).getLineNumber());
+
+        // prop9, prop10, prop11 throw exception and line number starts from 0
+        assertEquals(3, errors.size());
+        assertEquals(8, errors.get(0).startLine());
+        assertEquals(9, errors.get(1).startLine());
+        assertEquals(10, errors.get(2).startLine());
```
vlaaad, line 223: Error lines are 0-indexed

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
```diff
@@ -223,31 +215,32 @@ public void testPropsUrl() throws Exception {
...
     @Test
     public void testPropsVec3() throws Exception {
         List<Property> properties = getPropertiesFromFile("test_props_vec3.lua");
 
-        assertEquals(4, properties.size());
+        assertEquals(3, properties.size());
         assertProperty(properties, "prop1", new Vector3d(), 0);
         assertProperty(properties, "prop2", new Vector3d(1, 2, 3), 1);
-        assertProperty(properties, "prop4", new Vector3d(2, 2, 2), 3);
+        assertPropertyStatus(properties, "prop3", Status.INVALID_VALUE, 2);
```
vlaaad, line 233: There is a syntax error in prop3 go property declaration:
```
go.property("prop1", vmath.vector3())
go.property("prop2", vmath.vector3(1, 2, 3))
go.property("prop3", vmath.vector3(1a, 2, 3))
go.property("prop4", vmath.vector3(2))

```
`1a` is not a valid value, so there is an error at `1`, and then `a,2,3` are parsed as global variable declarations, but then there are more syntax errors because the code expects `=` (e.g.,`a,b,c=1,2,3`). Then, finally, `go.property` is parsed as a value inside the global var declaration, like `a,b,c = go.property(),2,3`, and it gets reported as an invalid location because it's not a top-level statement. But this error is bonkers, so I opted for stopping parsing `go.property` declarations after the first syntax error. Hence, now there are only 3 `go.property` declarations.

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
```diff
@@ -486,77 +443,173 @@ public void testRemoveEmptyLifecycleFunctions() throws Exception {
...
-        assertEquals(expected, scanner.getParsedLua());
+        assertEquals(expected, LuaScanner.parse(luaCode).code());
 
-        scanner = new LuaScanner();
-        scanner.setDebug();
-        scanner.parse(luaCode);
-        expected = " hash(\"value\")";
-        assertEquals(expected, scanner.getParsedLua());
+        expected = "                     hash(\"value\") ";
+        assertEquals(expected, LuaScanner.parse(luaCode, true).code());
```
vlaaad, line 478: In debug, we now replace `go.property` declarations with spaces, as we did in the editor. Better error reporting!

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
```diff
@@ -486,77 +443,173 @@ public void testRemoveEmptyLifecycleFunctions() throws Exception {
...
+        expected = "                                                                                          ";
+        assertEquals(expected, LuaScanner.parse(luaCode, true).code());
+    }
+
+    private static void assertCompilesToInDebug(String expected, String actual) {
+        assertEquals("`" + expected + "`", "`" + LuaScanner.parse(actual, true).code() + "`");
+    }
+
+    @Test
+    public void testGoPropertyWhitespace() {
```
vlaaad, line 501: This used to be an editor test. I moved it here, since it tests LuaScanner.

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/test_props.lua
```diff
@@ -20,8 +20,8 @@ go.property(invalid_string, 2)
 go.property("three_args", 1, 2)
 go.property("unknown_type", "hello")
 
-'go.property("no_prop", 123)'
-"go.property('no_prop', 123)"
+local s1 = 'go.property("no_prop", 123)'
+local s2 = "go.property('no_prop', 123)"
```
vlaaad, line 24: There was a problem with the previous test code. `go.property("unknown_type", "hello") 'go.property("no_prop", 123)'` is parsed as a nested function call, as if `go.property()` returns a function and it then gets called with a string arg, like this:
```lua
go.property("unknown_type", "hello")('go.property("no_prop", 123)')("go.property('no_prop', 123)")
```
This makes `go.property` not a top-level form, and it's reported as error, while the test is testing that `go.proeprty` calls within strings are not parsed as `go.property` invocations

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LuaBuilder.java
```diff
@@ -114,17 +112,12 @@ private LuaScanner getLuaScanner(IResource resource) throws IOException, Compile
...
 
-            luaScanner = new LuaScanner();
-            if (variant == Bob.VARIANT_DEBUG)
-            {
-                luaScanner.setDebug();
-            }
-            luaScanner.parse(script);
-            for(LuaScanner.LuaScannerException e: luaScanner.exceptions) {
-                throw new CompileExceptionError(resource, e.getLineNumber(), e.getErrorMessage());
+            luaScannerResult = LuaScanner.parse(script, Bob.VARIANT_DEBUG.equals(variant));
```
vlaaad, line 115: Debug is an argument

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LuaBuilder.java
```diff
@@ -114,17 +112,12 @@ private LuaScanner getLuaScanner(IResource resource) throws IOException, Compile
...
+            for (LuaScanner.ParseError error : luaScannerResult.errors()) {
+                throw new CompileExceptionError(resource, error.startLine() + 1, error.message());
```
vlaaad, line 117: Error lines are now 0-indexed

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LuaScanner.java
```diff
@@ -14,59 +14,103 @@
...
+     */
+    public record Property(Status status,
+                           int startLine,
+                           int startColumn,
+                           int endLine,
+                           int endColumn,
+                           String name,
+                           PropertyType type,
+                           Object value,
+                           String resourceKind) {
```
vlaaad, line 100: We now report resource kind too

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LuaScanner.java
```diff
@@ -91,233 +134,302 @@ public int getLineNumber() {
...
-            return Objects.equals(fnName, functionName);
-        }
+    public static Result parse(String code, boolean isDebug, Predicate<String> resourceKindPredicate) throws IOException {
+        return parse(CharStreams.fromString(code), isDebug, resourceKindPredicate);
     }
 
-    public LuaScanner() {}
+    public static Result parse(Reader reader, boolean isDebug, Predicate<String> resourceKindPredicate) throws IOException {
+        return parse(CharStreams.fromReader(reader), isDebug, resourceKindPredicate);
+    }
```
vlaaad, line 163: More overloads for the editor.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LuaScanner.java
```diff
@@ -91,233 +134,302 @@ public int getLineNumber() {
...
-             if (token.getChannel() == LuaLexer.COMMENTS) {
+            if (token.getChannel() == LuaLexer.COMMENTS) {
                 int type = token.getType();
                 if (type == LuaLexer.LINE_COMMENT) {
-                    // Single line comment
-                    rewriter.replace(token, System.lineSeparator());
-                }
-                else if (type == LuaLexer.COMMENT) {
+                    // Single line comment (might end with \n)
+                    rewriter.replace(token, token.getText().endsWith("\n") ? System.lineSeparator() : "");
```
vlaaad, line 190: A line comment on the last line in the code file does not end with `\n`. We shouldn't add empty lines to the rewritten code.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LuaScanner.java
```diff
@@ -91,233 +134,302 @@ public int getLineNumber() {
...
+                } else if (type == LuaLexer.COMMENT) {
                     // Multiline comment
-                    rewriter.replace(token, System.lineSeparator().repeat(token.getText().split("\r\n|\r|\n").length - 1));
+                    if (isDebug) {
+                        rewriter.replace(token, NON_LINE_BREAKS.matcher(token.getText()).replaceAll(" "));
```
vlaaad, line 194: In debug, we replace comments with spaces for better error reporting (we report precise locations for errors that include character positions, and the editor selects the right text) in the build errors view.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LuaScanner.java
```diff
@@ -91,233 +134,302 @@ public int getLineNumber() {
...
+            }
         }
 
         // parse code
-        LuaParser parser = new LuaParser(tokenStream);
-        ParseTreeWalker walker = new ParseTreeWalker();
-        walker.walk(this, parser.chunk());
-        String resultText = rewriter.getText();
+        var parser = new LuaParser(tokenStream);
+        parser.removeErrorListeners();
```
vlaaad, line 204: Remove the default error listener that prints to stderr; otherwise, editor tests are spammed with syntax errors. The errors are still collected and reported as errors!

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LuaScanner.java
```diff
@@ -366,225 +482,195 @@ private List<String> getAllStringArgs(LuaParser.ArgsContext argsCtx) {
...
+
+    private record InvalidValue(String message) implements ParsePropertyResult {
+    }
+
+    private record Success(ParserRuleContext resultContext,
+                           PropertyType type,
+                           Object value,
+                           String resourceKind) implements ParsePropertyResult {
+        public Success(ParserRuleContext resultContext, PropertyType type, Object value) {
+            this(resultContext, type, value, null);
```
vlaaad, line 565: Instead of throwing errors, we now return results when parsing the property values.

editor/src/clj/editor/code/script_compilation.clj
```diff
@@ -136,87 +134,20 @@
            :user-data (assoc (r/make-code-error-user-data filename line)
                         :message message)})))))
 
-(defn- line->whitespace [line]
-  (string/join (repeat (count line) \space)))
-
-(defn- go-property-declaration-cursor-ranges
```
vlaaad, line 142: Not needed: LuaScanner does the job too!

editor/src/clj/editor/code/script_intelligence.clj
```diff
@@ -1,27 +1,28 @@
...
             [editor.graph-util :as gu]
+            [editor.lua :as lua]
             [schema.core :as s]))
 
 (g/deftype ScriptCompletions {s/Str [{s/Keyword s/Any}]})
 
 (g/defnk produce-lua-completions
   [lua-completions]
-  (apply (partial merge-with into) lua-completions))
+  (lua/combine-completions (apply (partial merge-with into) lua-completions)))
```
vlaaad, line 25: We can generate built-in completions once for script-api files and defold APIs, and the rest is handled by LSP!

editor/src/clj/editor/lua.clj
```diff
@@ -230,52 +229,8 @@
...
-         (merge (:functions local-completion-info) (:local-functions local-completion-info))
-         nil)
-       (let [module->completion-info (into {} (map (juxt :module identity)) required-completion-infos)]
-         (eduction
-           (mapcat (fn [[alias module]]
-                     (let [completion-info (module->completion-info module)]
-                       (make-completion-info-completions (:vars completion-info)
-                                                         (:functions completion-info)
-                                                         alias))))
-           (:requires local-completion-info)))])))
```
vlaaad, line 275: The AST completions are no more!

editor/test/editor/code/script_test.clj
```diff
@@ -234,69 +236,3 @@
     ["    s = 'will end something'|"]
     ["    s = 'will end something'"
      "    |"]))
-
-(def strip-go-prop-declarations #'script-compilation/strip-go-property-declarations)
-
-(deftest strip-go-prop-declarations-test
```
vlaaad, line 240: I moved this test to LuaScannerTest.

editor/test/editor/lua_parser_test.clj
```diff
@@ -29,156 +29,44 @@
   ([workspace valid-resource-kind? code]
    (lp/lua-info workspace valid-resource-kind? code)))
 
-(deftest test-variables
```
vlaaad, line 32: Removed: we no longer parse Lua for completions

editor/test/editor/lua_parser_test.clj
```diff
@@ -29,156 +29,44 @@
...
+      (is (= ["deep_thought"] result))))
   (testing "require call as part of complex expression"
     (let [code (string/join "\n" ["state_rules[hash(\"main\")] = hash_rules("
                                   "    require \"main/state_rules\""
                                   ")"])
-          result (select-keys (lua-info code) [:vars :requires])]
-      (is (= {:vars #{"state_rules"}
-              :requires [[nil "main/state_rules"]]} result)))))
-
-(deftest test-functions
```
vlaaad, line 120: Removed: we no longer parse Lua for completions

editor/test/editor/lua_parser_test.clj
```diff
@@ -29,156 +29,44 @@
...
   (testing "require test file"
     (let [result (lua-info (slurp (io/resource "lua/require_test.lua")))]
-      (is (= {:vars #{"foo"}
-              :local-vars #{"bar"}
-              :functions {}
-              :local-functions {}
-              :script-properties []
-              :requires [[nil "mymath"] ["foo" "mymath"] ["bar" "mymath"]]} result)))))
-
-(deftest test-lua-info-with-spaceship
```
vlaaad, line 164: Removed: we no longer parse Lua for completions

editor/test/editor/lua_parser_test.clj
```diff
@@ -29,156 +29,44 @@
...
-              :script-properties []
-              :requires []} result))))
   (testing "require test file"
     (let [result (lua-info (slurp (io/resource "lua/require_test.lua")))]
-      (is (= {:vars #{"foo"}
-              :local-vars #{"bar"}
-              :functions {}
-              :local-functions {}
-              :script-properties []
-              :requires [[nil "mymath"] ["foo" "mymath"] ["bar" "mymath"]]} result)))))
```
vlaaad, line 162: Additionally, using Bob's parser fixed an issue where multiple requires of the same lua file resulted in a dependency being included multiple times. Now, it's only mentioned once (see line 63 below)

editor/test/editor/lua_parser_test.clj
```diff
@@ -251,31 +140,17 @@
...
 (deftest parsing-garbage-works
   (let [fuzz-codes (take 100 (repeatedly #(RandomStringUtils/random (rand-int 400))))]
-    (doall (map #(do (reset! last-fuzz %) (lua-info %)) fuzz-codes))
+    (run! #(do (reset! last-fuzz %) (lua-info %)) fuzz-codes)
 
-  (let [original (slurp (io/resource "build_project/SideScroller/spaceship/spaceship.script"))
-        broken-versions (map soil (repeat 100 original))]
-    (doall (map #(do (reset! last-broken %) (lua-info %)) broken-versions)))))
-
-(deftest string-as-function-parameter-is-ignored
```
vlaaad, line 260: Removed: we no longer parse Lua for completions

editor/test/editor/lua_parser_test.clj
```diff
@@ -287,13 +162,8 @@
     "local function\nfunction foo(a,b"
     "local function\nfunction foo(a,b)"))
 
-(deftest test-broken-table-def
```
vlaaad, line 290: Removed: we no longer parse Lua for completions

editor/test/editor/lua_parser_test.clj
```diff
@@ -407,9 +277,16 @@
              (src->properties workspace "go.property")))
       (is (= [{:status :invalid-args}]
              (src->properties workspace "go.property()")))
-      (is (= [{:status :invalid-args :name "test"}]
+      (is (= [{:status :invalid-value
+               :name "test"}]
              (src->properties workspace "go.property(\"test\")")))
-      (is (= [{:status :invalid-name :name "" :type :script-property-type-number :value 0.0}]
+      (is (= [{:status :invalid-args}]
```
vlaaad, line 283: Bob doesn't have an `invalid-name` error; there are only invalid args (this is also used when it can't parse name), invalid value, and invalid location.

editor/test/integration/extension_lua_preprocessor_test.clj
```diff
@@ -67,47 +67,47 @@
...
+             ""
              "                                                                                  "
-             "--#ENDIF"
              ""
-             "--#IF RELEASE"
+             ""
+             ""
              "                                                                                      "
-             "--#ENDIF"]
+             ""]
```
vlaaad, line 86: Comments are now removed in the editor, too!

editor/test/integration/script_test.clj
```diff
@@ -38,89 +38,37 @@
...
                                                  "z = 11")
           project          (test-util/setup-project! workspace [script-resource module1-resource module2-resource])
-          script-node      (project/get-resource-node project script-resource)
-          module1-node     (project/get-resource-node project module1-resource)
-          module2-node     (project/get-resource-node project module2-resource)]
+          script-node      (project/get-resource-node project script-resource)]
       (testing "are empty when no modules have been required"
         (is (not (g/error? (g/node-value script-node :build-targets))))
-        (is (empty? (:dynamic-deps (first (g/node-value script-node :build-targets)))))
-        (is (empty? (g/node-value script-node :module-completion-infos))))
```
vlaaad, line 47: Completion infos are no more!

editor/test/integration/unload_test.clj
```diff
@@ -544,22 +544,21 @@
...
               (let [node-load-info-tx-data-calls
                     (call-logged-transact!
                       (test-util/set-code-editor-lines
                         (project/get-resource-node project "/loaded_referencing_unloaded_lua.script")
                         ["local first_alias = require 'unloaded.unloaded'"
                          "local second_alias = require 'unloaded.unloaded'"]))]
 
-                (is (= ["/unloaded/unloaded.lua"
-                        "/unloaded/unloaded_math.lua"]
+                (is (= []
```
vlaaad, line 555: We no longer connect completions, and module deps are dynamic, so it seems unloaded Lua modules are no longer loaded.

AGulev (dismissed):
👍

editor/src/clj/editor/code/script_annotations.clj
```diff
@@ -0,0 +1,67 @@
...
+
+(set! *warn-on-reflection* true)
+(set! *unchecked-math* :warn-on-boxed)
+
+(g/deftype ResourceWithLines
+  {:resource (s/protocol resource/Resource)
+   :lines types/TLines})
+
+(g/defnk produce-sync-hash [_node-id root script-annotations]
+  (try
```
matgis, line 36: We should maybe return an error if `root` is an empty string or `nil` before we start to write to `.internal`? It likely wouldn't happen for users, but I can see myself causing it during development. 😅

editor/src/clj/editor/editor_extensions.clj
```diff
@@ -598,12 +600,16 @@
...
-    #{{:languages #{"lua"}
-       :watched-files [{:pattern "**/.luacheckrc"}]
-       :launcher {:command [(str lua-lsp-root "/bin/lua-language-server" (when (os/is-win32?) ".exe"))
-                            (str "--configpath=" lua-lsp-root "/config.json")]}}}))
+    {:languages #{"lua"}
+     :watched-files [{:pattern "**/.luacheckrc"}]
+     ;; We want to restart the language server every time annotations from
+     ;; dependencies change because lua language server does not watch
+     ;; `workspace.library` folder. Changing the map triggers the server restart
+     ::sync-hash (if (g/error-value? annotations-sync-hash) 0 annotations-sync-hash)
```
matgis, line 610: Is it intentional to use a `::` namespaced keyword  here? I can't find any usages of it.

editor/src/clj/editor/fs.clj
```diff
@@ -744,6 +744,10 @@
   ^long [p]
   (.toMillis (Files/getLastModifiedTime p empty-link-option-array)))
 
+(defn path-last-modified-time!
```
matgis: I would prefer if we renamed this to `set-path-last-modified-time!`. We try to avoid the setter-name-is-getter-name-with-bang convention these days.

editor/test/editor/code/script_test.clj
```diff
@@ -18,7 +18,9 @@
             [editor.code.data :as data]
             [editor.code.data-test :refer [c cr layout-info]]
             [editor.code.script :as script]
-            [editor.code.script-compilation :as script-compilation]))
+            [editor.code.script-compilation :as script-compilation]
+            [editor.lua-parser :as lua-parser])
+  (:import [com.dynamo.bob.pipeline LuaScanner]))
```
matgis: It looks like you added some unused imports here?

matgis (dismissed):
Fantastic work! LGTM!

editor/src/clj/editor/code/script_annotations.clj
vlaaad: If `root` is nil, `(fs/path nil ".internal" "lua-annotations")` will throw NPE. If it's an empty string, the path will be relative to the current working directory (cwd). In both these hypothetical scenarios, the behavior doesn't seem dangerous.

editor/src/clj/editor/editor_extensions.clj
vlaaad: Yeah, I made it namespaced so it doesn't clash with any keys that we might add to the lsp server data shape definition that uses non-namespaced keywords.

Issue #6557, opened 2022-04-28
Make editor suggest fetching libraries and reloading editor scripts


vlaaad:
**Is your feature request related to a problem? Please describe (REQUIRED):**
When changing the dependencies list, it's not clear that we should fetch libraries first to actually get dependencies. Same with editor scripts — changing the script does not immediately change the behavior of the editor (as it should!), but the editor could be more helpful by suggesting reloading the scripts.

**Describe the solution you'd like (REQUIRED):**
Show a notification popup that suggests reloading the scripts and fetching libraries

**Describe alternatives you've considered (REQUIRED):**
We could also fetch libraries automatically, and reload editor scripts on save.

britzl:
Decision: Any change to the libraries list (manually adding a dependency or switching branch) shows bubble popup in  lower right corner with option to Ignore or Fetch Libraries.

Same for editor scripts.

PR #11159, opened 2025-08-29
Suggest fetching libraries when modifying game.project dependencies

vlaaad:
We now show a notification when dependencies change:
Screenshot 2025-08-29 at 10 21 09

Related to #6557


matgis (approved):
Nice! ❤️

PR #11169, opened 2025-09-01
Issue 11145 simd support for box2d

deurell:
Add both simd and no_simd libs to all Box2D packages so the user can select if they want simd support when using Box2D 3.x.
Fixes #11145 

Technical changes
- GitHub action now builds and packages both simd and no_simd version of Box2D libs for all platforms.
- Removed explicit mfpu=vfp override in waf_dynamo.py to make it possible to set neon for armv7
- Added updated packages with simd/no_simd versions of Box2D for all platforms
- Added feature flag to select box2dv3 (--enable-feature=box2dv3)
- Box2Dv3 respects the --enable-feature=simd flag

Run local engine tests with selected Box2D version using:
`CXXFLAGS="-DDM_PHYSICS_BOX2D_V3" PREFIX="$DYNAMO_HOME" waf configure --enable-feature=box2dv3 --enable-feature=simd`


JCash (changes requested):
As discussed, (perhaps not clearly enough, sorry!), for html5 we should use _nosimd.a (same as for sound_nosimd.a
That makes it easy for them to use/test the new box2d for web.
Full support for simd on web won't arrive soon.
Search for "sound_nosimd" in project, and adjust accordingly (only small fixes needed imho)

Also, missing change in build_input.yml where "box2d_nosimd" library needs to be used.

Have you done testing with this defoldsdk and building a physics project using an app manifest?

JCash (dismissed):

JCash (approved):

PR #11170, opened 2025-09-01
Show literal labels for user-defined properties

### LINKED ISSUE #11158: Beautify names in the editor in a consistent way
**Is it a private platform specific feature? (i.e. would it break your NDA?)**
No

When Defold changes (beautifies) names in the editor, it does so in different ways.

I tested underscores and case for material names and script properties.

The names I used for testing:
test
Test
another_test
Another_Test

Material names and script properties in the editor:

Image

Material names have the underscores removed, the case is untouched.
For script properties, underscores are replaced with whitespace and all initial letters are capitalized. 

- beautifying should be consistent.
- lowercase initial letters should not be capitalized as this can lead to duplicate names in the editor and thus be a bit confusing.
- it might be a good idea to explain to the user how this beautifying works.
- esp. important imho would be to make clear that the beautifying is there for display purposes only and that names aren't actually changed in the   project files.

The explanations could for instance be added to the md file.

Leave the names as is if possible. Shorten them if they get too long for diplay in the editor. 
Maybe like so:
"This_is_a_terribly_terribly_long_name" would become "This_is_a_..."

matgis:
We now label user-defined properties in the Property Editor using the literal name the property was declared as.

Fixes #11158.

### DIFF
--- a/editor/src/clj/editor/properties_view.clj
+++ b/editor/src/clj/editor/properties_view.clj
@@ -930,6 +930,7 @@
         (doto (MenuButton.)
           (ui/add-style! "property-label")
           (ui/register-button-menu ::property-menu)
+          (.setMnemonicParsing false)
           (.setFocusTraversable false)
           (.setPrefWidth Region/USE_COMPUTED_SIZE)
           (.setMinWidth Region/USE_PREF_SIZE)
--- a/editor/test/integration/override_transfer_test.clj
+++ b/editor/test/integration/override_transfer_test.clj
@@ -139,7 +139,7 @@
     (test-util/with-changes-reverted project
       (let [source-node-id (test-util/resource-outline-node-id project "/shelf.collection" "referenced_book" "book_script")
             transfer-overrides-plans (pull-up-overrides-plan-alternatives source-node-id :all)]
-        (is (= [["Pull Up Text Override to 'book_script' in '/book.go'" :ok]]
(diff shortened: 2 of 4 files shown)


editor/test/integration/override_transfer_test.clj
```diff
@@ -139,7 +139,7 @@
     (test-util/with-changes-reverted project
       (let [source-node-id (test-util/resource-outline-node-id project "/shelf.collection" "referenced_book" "book_script")
             transfer-overrides-plans (pull-up-overrides-plan-alternatives source-node-id :all)]
-        (is (= [["Pull Up Text Override to 'book_script' in '/book.go'" :ok]]
+        (is (= [["Pull Up 'text' Override to 'book_script' in '/book.go'" :ok]]
```
matgis, line 142: It looked a bit weird to have a non-capitalized word here, so I decided to quote the property name like we already do for the file paths.

editor/src/clj/editor/properties_view.clj
```diff
@@ -930,6 +930,7 @@
         (doto (MenuButton.)
           (ui/add-style! "property-label")
           (ui/register-button-menu ::property-menu)
+          (.setMnemonicParsing false)
```
vlaaad, line 933: 🔥

vlaaad: Should be a default!

vlaaad (approved):
Ship iiiiiiit! 🚢 🚢 🚢

PR #11173, opened 2025-09-01
Fix built Gui layout NodeDescs missing non-property pb-fields

### LINKED ISSUE #11162: Spine not showing in the GUI when using layout in the editor
### Actual behaviour

Spine nodes disappear after adding any layout to the GUI.
Web and win bundles work fine. 
I have this problem only in the editor.

This problem on Defold 1.11.0.
Defold 1.10.4 work fine.

1. Without layout:
Image

2. With layout:
Image

### Steps to reproduce

1. Add spine to the GUI
2. Add any layout to the GUI

<hr/>

<table>
<tr><td>Build time</td><td>2025-08-29T07:51:31.129650</td></tr>
<tr><td>Defold editor sha</td><td>7e18fe7de0f5ff6d455e07e56241a791f58f98e6</td></tr>
<tr><td>Defold engine sha</td><td>7e18fe7de0f5ff6d455e07e56241a791f58f98e6</td></tr>
<tr><td>Defold version</td><td>1.11.0</td></tr>
<tr><td>GPU</td><td>Radeon RX 580 Series</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 Compatibility Profile Context 23.11.1.231017</td></tr>
<table>

matgis:
Fixed a regression where built Gui layout `NodeDescs` were missing all non-property Protobuf field values from the default layout `NodeDesc`, and as a result introducing a layout-specific override for the `NodeDesc` with bad values in the build output.

Fixes #11162.

### DIFF
--- a/editor/src/clj/editor/gui.clj
+++ b/editor/src/clj/editor/gui.clj
@@ -3221,7 +3221,7 @@
                    (assert (map? prop->value))
                    (when-some [node-desc-for-layout
                                (some->> (-> decorated-node-msg
-                                            (select-keys [:custom-type :id :parent :template-node-child :type])
+                                            (dissoc :layout->prop->override :layout->prop->value)
                                             (into (map prop-entry->pb-field-entry)
                                                   prop->value)
                                             (node-desc->rt-node-desc layout-name))
--- a/editor/test/integration/extension_spine_test.clj
+++ b/editor/test/integration/extension_spine_test.clj
@@ -25,6 +25,7 @@
             [editor.resource-node :as resource-node]
             [editor.settings-core :as settings-core]
             [editor.workspace :as workspace]
+            [integration.gui-test :as gui-test]
             [integration.test-util :as test-util]
             [support.test-support :as test-support]
             [util.coll :refer [pair]]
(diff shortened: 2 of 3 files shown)


editor/src/clj/editor/gui.clj
```diff
@@ -3221,7 +3221,7 @@
                    (assert (map? prop->value))
                    (when-some [node-desc-for-layout
                                (some->> (-> decorated-node-msg
-                                            (select-keys [:custom-type :id :parent :template-node-child :type])
+                                            (dissoc :layout->prop->override :layout->prop->value)
```
matgis, line 3224: The fix. Not sure why we were so aggressive about starting with a cleaned-out `NodeDesc` before we applied the layout property overrides. This change ensures the built `NodeDesc` will include field values that are not backed by a property on the specific `GuiNode`.

The problem before was that the `NodeDesc` produced by a `SpineNode` would include a `:size-mode :size-mode-auto` entry that was necessary for correct rendering in the runtime. But the `SpineNode` does not actually have a `:size-mode` property, so we wouldn't add this entry from `prop->value` below after we had stripped it out.

Anyway, we clear out any default from the resulting `NodeDesc` a few lines below, so I don't worry much about the fact that we're not starting from a clean slate after this change. Having `(dissoc :layout->prop->override :layout->prop->value)` as a starting point also mirrors what we do when producing the `NodeDesc` for the default layout, so it seems more correct.

vlaaad (approved):
LGTM!

PR #11178, opened 2025-09-02
Allow setting game.project properties using editor scripts

### LINKED ISSUE #11163: Add ability to set game.project properties using editor scripts
It turns out to get a list of dependencies:
`editor.get("/game.project", "project.dependencies")`

But there is no way to enter your data there (it does not work):
`editor.transact({editor.tx.set("/game.project", "project.dependencies", deps)})`

**Defold version: 1.10.4**

vlaaad:
Now, it's possible to set properties in the `game.project` file using editor scripts:
```lua
editor.transact({
    editor.tx.set("/game.project", "project.dependencies", {
        "https://github.com/defold/extension-spine/archive/refs/tags/3.10.0.zip"
    })
})
```

🚨 **Breaking change** 🚨 for extensions that are based on [extension-simpledata](https://github.com/defold/extension-simpledata): Now, the editor expects the code that is invoked when form values are set/cleared to return transaction steps instead of performing a transaction. See [this PR](https://github.com/defold/extension-simpledata/pull/19/files) for the changes you need to make.

Fixes #11163

### DIFF
--- a/editor/src/clj/editor/form.clj
+++ b/editor/src/clj/editor/form.clj
@@ -13,15 +13,25 @@
 ;; specific language governing permissions and limitations under the License.
 
 (ns editor.form
-  (:require [editor.util :as util])
+  (:require [dynamo.graph :as g]
+            [editor.util :as util])
   (:import [java.net URI]))
 
 (set! *warn-on-reflection* true)
 
+(defn- ensure-does-not-transact [ret f]
+  (when (g/tx-result? ret)
+    (throw
+      (IllegalStateException.
+        (str "Invalid implementation of form-ops\nFunction: "
+             (Compiler/demunge (.getName (class f)))
+             "\nIt should return transaction steps instead of performing a transaction. Use, e.g.:\n- `g/set-property` instead of `g/set-property!`\n- `g/clear-property` instead of `g/clear-property!`"))))
+  ret)
+
 (defn set-value
   "Returns transaction steps for setting the value"
   [{:keys [user-data set]} path value]
-  (set user-data path value))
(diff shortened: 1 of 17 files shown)


matgis (dismissed):
Nice!

I like that `:form-ops` now return transaction steps. However, I think you'll need to also update the behavior in `extension-simpledata`: https://github.com/defold/extension-simpledata/blob/ca1dbf55d2ad7ae9324206806141d474ee9893eb/defold-simpledata/editor/src/simpledata-ext.clj#L90

We also need to explicitly provide migration instructions in release notes, since game projects might have added Clojure code based on `extension-simpledata` to their projects.

vlaaad:
Simpledata PR: https://github.com/defold/extension-simpledata/pull/19

editor/src/clj/editor/form.clj
```diff
@@ -13,19 +13,33 @@
...
   (:import [java.net URI]))
 
 (set! *warn-on-reflection* true)
 
-(defn set-value! [{:keys [user-data set]} path value]
-  (set user-data path value))
+(defn- ensure-does-not-transact [ret f]
+  (when (g/tx-result? ret)
+    (throw
+      (IllegalStateException.
```
matgis, line 25: Is this the correct exception type to throw? Would `UnsupportedOperationException` be better?

matgis (approved):
Looks good to me!

editor/src/clj/editor/form.clj
vlaaad: I think exception type doesn't matter here as much as the message...

PR #11180, opened 2025-09-02
Vulkan improve texture upload performance

Jhonnyg:
Performance for dynamic texture uploads have been improved when using the Vulkan backend.

fixes https://github.com/defold/defold/issues/10921

### Technical changes

* All hard syncronization points using vkQueueWaitIdle have been removed, instead we use fences to check when an upload has finished
* The new fance based system is similar to our "deferred destroyal" arcitechture, but serves a slightly different purpose. Our usual resources are tied to an "in-flight" frame number, but the texture uploads are not. 
* We poll the fence as late as we can (i.e when the image handle is actually needed) so the GPU is free to process other tasks. This system can then be extended for device buffer (vertex/index buffers) use as well!


engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -1550,21 +1621,36 @@ namespace dmGraphics
...
             VkImageAspectFlags vk_aspect = 0;
-            if (flags & BUFFER_TYPE_DEPTH_BIT)
+
+            if ((flags & BUFFER_TYPE_DEPTH_BIT) && FormatHasDepth(depth_stencil_format))
             {
                 vk_aspect |= VK_IMAGE_ASPECT_DEPTH_BIT;
             }
 
-            if (flags & BUFFER_TYPE_STENCIL_BIT)
+            if ((flags & BUFFER_TYPE_STENCIL_BIT) && FormatHasStencil(depth_stencil_format))
```
Jhonnyg, line 1632: Fixes and issue where a depth and/or stencil clear was requested, but the render target doesn't have those attachments.

engine/graphics/src/vulkan/graphics_vulkan_private.h
```diff
@@ -309,18 +302,37 @@ namespace dmGraphics
...
         };
         VulkanResourceType m_ResourceType;
     };
 
+    struct FenceResourcesToDestroy
+    {
+        VkFence           m_Fence;
+        ResourceToDestroy m_Resources[MAX_FENCE_RESOURCES_TO_DESTROY_PER_ENTRY];
+        uint8_t           m_ResourcesCount;
+    };
```
Jhonnyg, line 334: It's intentionally generic, we can use the same syncronization structure for other types of CPU -> GPU uploads. E.g vertex/index buffer uploads.

engine/graphics/src/vulkan/graphics_vulkan_private.h
```diff
@@ -100,6 +92,7 @@ namespace dmGraphics
         VkImageUsageFlags m_UsageFlags;
         DeviceBuffer      m_DeviceBuffer;
         int32_atomic_t    m_DataState; // data state per mip-map (mipX = bitX). 0=ok, 1=pending
+        HOpaqueHandle     m_PendingUpload;
```
Jhonnyg, line 95: This is the upload marker for the GPU upload. It's a slightly different setup than our normal async uploads, this system is mostly for avoiding GPU stalls on the main thread.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -3729,24 +3832,6 @@ namespace dmGraphics
         g_VulkanContext->m_AssetHandleContainer.Release(texture);
     }
 
-    static inline uint32_t GetOffsetFromMipmap(VulkanTexture* texture, uint8_t mipmap)
```
Jhonnyg, line 3732: unused

engine/graphics/src/vulkan/graphics_vulkan_device.cpp
```diff
@@ -247,40 +247,6 @@ namespace dmGraphics
...
-        vk_submit_info.pCommandBuffers    = &m_CmdBuffer;
-
-        VkResult res = vkQueueSubmit(m_Context->m_LogicalDevice.m_GraphicsQueue, 1, &vk_submit_info, VK_NULL_HANDLE);
-        vkQueueWaitIdle(m_Context->m_LogicalDevice.m_GraphicsQueue);
-        vkFreeCommandBuffers(m_Context->m_LogicalDevice.m_Device, m_Context->m_LogicalDevice.m_CommandPool, 1, &m_CmdBuffer);
-
-        m_Context   = 0;
-        m_CmdBuffer = VK_NULL_HANDLE;
-        return res;
-    }
```
Jhonnyg, line 282: Replaced and/or removed. We don't do these hard syncronizations any longer (vkQueueWaitIdle basically stalls the entire GPU).

engine/dlib/src/dlib/opaque_handle_container.h
```diff
@@ -105,6 +105,21 @@ class dmOpaqueHandleContainer
...
+     * as denoted by the container capacity ([0...c-1] where c equals capacity).
+     *
+     * Note: this is indented for internal use only and relies on the fact that the backing
+     *       data structure for the container is a continuous array.
+     *
+     * @name IndexToHandle
+     * @param index [type:uint32_t] The index to query the container with
+     * @return [type:HOpaqueHandle] Return the constructed object handle
+     */
+    HOpaqueHandle IndexToHandle(uint32_t index);
```
Jhonnyg, line 121: We could potentially introduce an iterator instead and remove these "by index" functions. But for now it's fine I think?

AGulev:
is it fix for https://github.com/defold/defold/issues/10921 ?

Jhonnyg:
> is it fix for #10921 ?

Yop, added to description

JCash (approved):
Looks good. Are there any changes needed for Switch?

PR #11189, opened 2025-09-04
Decrease editor size by 12 MB

vlaaad:
Technical notes:
I removed clj-antler, which added a 12MB dependency on icu4j. It was only needed for generating the parser, but we can generate it at a build step.

It's good that we have a code completion test https://github.com/defold/defold/blob/dev/editor/test/editor/code_completion_test.clj


editor/src/clj/editor/code_completion.clj
```diff
@@ -147,6 +148,19 @@
...
+    (let [^ParserRuleContext ctx ctx
+          node (cons (keyword (aget ^Object/1 rule-names (.getRuleIndex ctx)))
+                     (into []
+                           (map (fn [^long i]
+                                  (->ast (.getChild ctx i) rule-names)))
+                           (range (.getChildCount ctx))))]
+      (if (.-exception ctx)
+        (list ::error node)
+        node))
+    (.getText ctx)))
```
vlaaad, line 162: This is a re-implementation of the clj-antler code that produces an AST. The intention with this PR is to reduce the editor build size, so I kept the behavior.

editor/src/clj/editor/code_completion.clj
```diff
@@ -197,7 +217,7 @@
                  ;; :snippet, :text, :text_esc and :err are unreachable because
                  ;; they only exist on the top level, while content-string is
                  ;; used to build string for choices, variables and placeholders
-                 :rule (content-string sb (first children))
+                 :syntax_rule (content-string sb (first children))
```
vlaaad, line 220: I had to rename `rule` to `syntax_rule`, otherwise antlr compiler complains that rule is a keyword...

editor/tasks/leiningen/init.clj
```diff
@@ -61,7 +61,7 @@
   [user_domain]
   (let [default_domain "d.defold.com"
         env_domain (System/getenv "DM_ARCHIVE_DOMAIN")]
-        (or user_domain env_domain default_domain)))
+    (or user_domain env_domain default_domain)))
```
vlaaad, line 64: Unrelated indentation fix

matgis (approved):
Let's gooo! 🥳🚀

PR #11190, opened 2025-09-04
Adding optional support for OpenAL based sound generation as an alterntive to the CPU based mixer

thengel01:
1) added alternative implementation of sound.cpp
2) alternative version is managed as it's own sound module library, which can be selected by --with-openal option during link
3) OpenAL version attempts to mix and decode sound using OpenAL, but offers fallback to WASM based code if features are lacking
4) For platforms that support it, sound updates can be run on a thread (using proxy to main event loop)


engine/sound/src/sound_private.h
```diff
@@ -18,6 +18,11 @@
 #include <dmsdk/dlib/configfile_gen.hpp>
 #include "sound.h"
 
+#if defined(__EMSCRIPTEN__) && !defined(DM_NO_THREAD_SUPPORT)
+#include <emscripten.h>
+#define SOUND_WASM_SUPPORT_THREADS 1
```
JCash: If it's not local to the file, we'd like a `DM_` prefix to the define.

engine/sound/src/decoders/decoder_opus.cpp
```diff
@@ -457,19 +465,37 @@ namespace dmSoundCodec
...
+                        uint32_t nc = stream_info->m_Info.m_Channels;
 
-                    for (uint32_t c = 0; c < nc; ++c)
+                        float* dest = (float*)buffer[0] + done_frames * nc;
+                        const float* out = (float*)(stream_info->m_LastOutput.Begin() + stride * stream_info->m_LastOutputOffset);
+                        uint32_t ns = nc * out_frames;
+                        for (uint32_t s = 0; s < ns; ++s)
+                        {
+                            *(dest++) = *(out++) * scale;
+                        }
```
JCash, line 479: Q: What does "interleaved" refer to here?
I would have assumed that interleaved is like: [l0,r0,l1,r1,...]

But this loop suggests that it's not interleaved? (whereas the loop on line 490 seem to step over an interleaved stream?)

engine/sound/src/sound_openal.cpp
```diff
@@ -0,0 +1,1348 @@
...
+ * Defold simple sound system
+ * NOTE: Must units is in frames, i.e a sample in time with N channels
+ *
+ */
+namespace dmSound
+{
+    using namespace dmVMath;
+
+    // Since using threads is optional, we want to make it easy to switch on/off the mutex behavior
+    struct OptionalScopedMutexLock
```
JCash: We already have `DM_MUTEX_OPTIONAL_SCOPED_LOCK(g_SoundSystem->m_Mutex);` in `dmsdk/dlib/mutex.h`

engine/sound/src/sound_openal.cpp
```diff
@@ -0,0 +1,1348 @@
...
+        };
+
+        dmAtomicStore32(&sound->m_IsRunning, 1);
+        dmAtomicStore32(&sound->m_Status, (int)RESULT_NOTHING_TO_PLAY);
+
+        sound->m_Mutex = 0;
+        sound->m_CondVar = 0;
+        sound->m_Thread = 0;
+
+#if SOUND_WASM_SUPPORT_THREADS
```
JCash: Can we use `DM_NO_THREAD_SUPPORT` instead, fo rconsistency?

engine/sound/src/sound_openal.cpp
```diff
@@ -0,0 +1,1348 @@
...
+
+    struct SoundSystem
+    {
+        dmSoundCodec::HCodecContext   m_CodecContext;
+
+        dmMutex::HMutex          m_Mutex;
+        dmConditionVariable::HConditionVariable m_CondVar;
+        dmThread::Thread        m_Thread;
+        int32_atomic_t          m_IsRunning;
+        int32_atomic_t          m_Status;
```
JCash: These are not needed, and can also be ifdef-ed with `DM_NO_THREAD_SUPPORT`

engine/sound/src/sound_openal.cpp
```diff
@@ -0,0 +1,1348 @@
...
+            uint32_t last = sound->m_InstancesActive.Size() - 1;
+            for (uint32_t i = 0; i < last; ++i) {
+                if (sound->m_InstancesActive[i] == sound_instance->m_Index) {
+                    if (i < last) {
+                        // swap to end of array so we can pop
+                        uint16_t tmp = sound->m_InstancesActive[last];
+                        sound->m_InstancesActive[last] = sound_instance->m_Index;
+                        sound->m_InstancesActive[i] = tmp;
+                    }
+                    sound->m_InstancesActive.Pop();
```
JCash: this looks like it could be substituted with `dmArray::EraseSwap()`:
```
sound->m_InstancesActive.EraseSwap(i);
```

engine/dlib/src/dlib/condition_variable.cpp
```diff
@@ -131,11 +131,18 @@ namespace dmConditionVariable
     {
         // dmLog uses dmMessage, which in turn uses dmConditionVariable (Wait & Signal).
         // We cannot place assertions here.
+    #if !defined(DM_NO_THREAD_SUPPORT)
```
JCash, line 134: Why is the define guard needed?
We shouldn't really be using this class if threads aren't supported?

engine/sound/src/decoders/decoder_opus.cpp
JCash: If I compare it to changes in decoder_stb_vorbis.cpp, they're sure do seem contrary?

engine/sound/src/sound_openal.cpp
```diff
@@ -0,0 +1,1348 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#include <stdint.h>
+#include <dlib/array.h>
+#include <dlib/atomic.h>
+#include <dlib/condition_variable.h>
```
JCash: I know a lot of it is a copy of the old sound.cpp, but this can probably avoid including the threading related includes as well. Every bit counts for compiling :)

engine/sound/src/sound_openal.cpp
```diff
@@ -0,0 +1,1348 @@
...
+        char*                   m_BufferData;
+
+        ALenum                  m_FormatMonoFloat32;
+        ALenum                  m_FormatStereoFloat32;
+        bool                    m_SourceSpatialize;
+
+        ALenum                  m_Eos;
+        ALenum                  m_DirectStreamingFormat[SOUND_DATA_TYPE_MAX];
+
+        Result initOpenal() {
```
JCash: Code style, we do `CamelCase()`.
As it's a new file, may be worth running our .clang_format on the entire file?

engine/sound/src/sound_openal.cpp
```diff
@@ -0,0 +1,1348 @@
...
+            result = sound->finalizeOpenal();
+            if (result != RESULT_OK) {
+                return result;
+            }
+        } else {
+            dmThread::Join(sound->m_Thread);
+            dmMutex::Delete(sound->m_Mutex);
+            dmConditionVariable::Delete(sound->m_CondVar);
+            result = (Result)dmAtomicGet32(&sound->m_Status);
+        }
```
JCash: I know this is copied from old sound.cpp, but I think I prefer ifdef-guards for this too.
I don't want our lower end systems to support cases that they're not really defined for (i.e. why use a mutex type when it's not even supported!).
I think such a strategy will help keep the code clean and understandable for others looking at the code.

JCash (changes requested):
Overall good.
POtential bug in the opus decoder.
Also, some smaller changes needed.

engine/sound/src/sound_private.h
thengel01: will do

engine/sound/src/decoders/decoder_opus.cpp
thengel01: There had been changes by Ben M to allow for easy feeding of data to the openAL backend from the fallback CPU side decoders. He had brought back the interleaved (LRLRLR...) data for Vorbis. I added the non-interleaved (slightly faster) version back in for the CPU mixer case there.
HERE we are looking at the Opus output of course. Ben M had forgotten to adjust it for the OpenAL backend at all. So I added in code to feed interleaved data when sound_openal.cpp is active & also use a different numerical range (-1..1 for AL; 32768..32767 for the CPU mixer)

engine/sound/src/sound_openal.cpp
thengel01: very true - will remove

engine/sound/src/sound_openal.cpp
thengel01: It currently is about the same. Thought was to possibly want a separate switch for just audio - should more systems become thread-aware for the wasm platform.

engine/sound/src/sound_openal.cpp
thengel01: Mutex we'll need - only to check for null - rest can indeed be optional.

engine/sound/src/sound_openal.cpp
thengel01: yup - "know your helper code"(tm)(

engine/dlib/src/dlib/condition_variable.cpp
thengel01: I wondered that, too. Copy-paste from the Linux version, I think... I can remove it.

engine/sound/src/sound_openal.cpp
thengel01: Sure... should cost zero bytes... but... a few ms in compile time maybe...

engine/sound/src/sound_openal.cpp
thengel01: k

engine/sound/src/sound_openal.cpp
thengel01: k - provided a non-threaded #if selected version, too

JCash (approved):

PR #11203, opened 2025-09-08
Clamp texture depth to 1 and guard against zero-sized Vulkan uploads.

### LINKED ISSUE #10975: Crash on resource.create_texture_async for basis compressed texture on Vulkan

Defold crash on MacOs 
`game_debug.appmanifest Graphics Osx` -> `Vulkan`
when try to do 

`resource.create_texture_async(texture_id, atlas_meta.texture, result.buffer, create_texture_cb)`

```
INFO:GRAPHICS: Transcoding: /__merge_item_assets_texture.texturec from 21 to 6 (TEXTURE_FORMAT_RGBA_BC7 -> cTFBC7_RGBA)
[mvk-error] VK_ERROR_OUT_OF_DEVICE_MEMORY: vkAllocateMemory(): Could not allocate a host-coherent or exportable VkDeviceMemory of size 0 bytes. The maximum memory-aligned size of a host-coherent VkDeviceMemory is 17179869184 bytes.
```

[carsh.txt](https://github.com/user-attachments/files/21294086/carsh.txt)

**To Reproduce (REQUIRED)**

call resource.create_texture_async using compressed texture 

No crash. Correct loaded texture.

  - Version 1.10.2

**Platforms (REQUIRED):**
 - Platforms: macOS

deurell:
Fix out-of-memory crash in Vulkan create_texture_async caused by zero-sized texture uploads

Fixes #10975

### Technical changes
- Clamp texture depth to at least 1 in `AsyncProcessCallback` to prevent invalid allocations
- Add guard to skip async upload if payload would be empty


engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -4126,15 +4126,24 @@ namespace dmGraphics
...
 
+            const bool no_payload =
+                (ap.m_Params.m_Width  == 0) ||
+                (ap.m_Params.m_Height == 0) ||
+                (tex_layer_count      == 0) ||
+                (tex_depth            == 0) ||
+                (tex_bpp              == 0) ||
+                (tex_data_size        == 0);
+
+            bool is_memoryless        = IsTextureMemoryless(tex) || no_payload;
```
Jhonnyg: Hm, doesn't this mean that we get a memoryless texture if the in parameters are invalid? 🤔 
I think that we should make sure the parameters are correct before we try to create a texture, if the width or height is zero the texture is invalid, even if it's created as memoryless.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -4126,15 +4126,24 @@ namespace dmGraphics
             VkCommandBuffer cmd_buffer = BeginSingleTimeCommands(context->m_LogicalDevice.m_Device, context->m_LogicalDevice.m_CommandPoolWorker);
 
             uint8_t tex_layer_count   = dmMath::Max(tex->m_LayerCount, ap.m_Params.m_LayerCount);
-            uint16_t tex_depth        = dmMath::Max(tex->m_Depth, ap.m_Params.m_Depth);
+            uint16_t tex_depth = dmMath::Max<uint16_t>( (uint16_t)1, dmMath::Max<uint16_t>(tex->m_Depth, ap.m_Params.m_Depth) );
```
Jhonnyg: You can omit the type if all arguments are correct, i.e

dmMath::Max((uint16_t) 1, dmMath::Max(tex->m_Depth, ap.m_Params.m_Depth));

Jhonnyg (changes requested):
Nice fix, but I think we might want to actually catch when the texture is trying to be created with incorrect in parameters. At least notify the users via a dmLogError or something

deurell:
Regarding the failing param the issue is that depth is always 0 on dynamic async created textures and the user wouldn't set depth explicitly on a 2d texture in lua right? Can it really be anything else than 1 in that case? I understand 3d textures but the most common one would be 2d right?

If we switch over to the non asynchronous create_texture it works perfectly fine with the exact same params.

Jhonnyg:
@deurell: Yeah that part of the fix is correct, but I'm thinking about this line:

`bool is_memoryless        = IsTextureMemoryless(tex) || no_payload;`

For example, if you pass in zero for width you say that it should be memoryless, but aren't we trying to create an invalid texture in that case?

engine/graphics/src/vulkan/graphics_vulkan.cpp
deurell: In retrospect looking at it, this was a bad idea. Removing it. 👍

deurell:
It became a oneliner in the end...

Jhonnyg (approved):
Thanks!

PR #11204, opened 2025-09-08
Vulkan fix validation errors

Jhonnyg:
Multiple validation errors have been fixed for the Vulkan graphics adapter. This essentially means that the Vulkan backend should be more stable and reliant than before. Please let us know if you encounter more validation errors!

### Technical Changes

* We now use the 'current frame in flight'-index when managing resources instead of the swap chain index. This means we can manage these independantly
* Each resource carries a number that indicates which frame-in-flight it was used in, so they can be correctly removed even if they are used outside of the VulkanBegin ... VulkanFlip flow
* Attachments for MSAA now have the correct initial state (clear), so we avoid the 'disco effect' of uninitialized memory
* Resources should be correctly cleaned now


engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -741,9 +821,11 @@ namespace dmGraphics
         VulkanTexture* depth_stencil_texture = &context->m_MainTextureDepthStencil;
         DestroyDeviceBuffer(vk_device, &depth_stencil_texture->m_DeviceBuffer.m_Handle);
         DestroyTexture(vk_device, &depth_stencil_texture->m_Handle);
-
         DestroySwapChain(vk_device, context->m_SwapChain);
 
+        // Reset the image layout
+        memset(depth_stencil_texture->m_ImageLayout, 0, sizeof(depth_stencil_texture->m_ImageLayout));
```
Jhonnyg, line 827: A bit of a special case here since we don't use the deferred deletion function that resets the layout automatically.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -959,15 +1041,9 @@ namespace dmGraphics
...
-        context->m_FragmentShaderInterlockFeatures       = {};
-        if (VulkanIsExtensionSupported((HContext) context, VK_EXT_FRAGMENT_SHADER_INTERLOCK_EXTENSION_NAME))
-        {
-            context->m_FragmentShaderInterlockFeatures.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FRAGMENT_SHADER_INTERLOCK_FEATURES_EXT;
-        }
-
         PhysicalDevice* device_list     = new PhysicalDevice[device_count];
         PhysicalDevice* selected_device = NULL;
-        GetPhysicalDevices(context->m_Instance, &device_list, device_count, &context->m_FragmentShaderInterlockFeatures);
+        GetPhysicalDevices(context->m_Instance, &device_list, device_count, NULL);
```
Jhonnyg, line 1046: The flow here was wrong before, we can't check if the extension is supported before we have a physical device. It's not perfect, but it's better now.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -3706,6 +3767,9 @@ namespace dmGraphics
             vk_copy_region.imageSubresource.layerCount     = 1;
         }
 
+        TouchResource(context, stage_buffer);
+        TouchResource(context, texture);
```
Jhonnyg, line 3771: Note: We have to tag all resources that are part of any vkCmd* calls

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -4026,9 +4086,16 @@ namespace dmGraphics
 
         DestroyDeviceBuffer(vk_device, &context->m_MainTextureDepthStencil.m_DeviceBuffer.m_Handle);
         DestroyTexture(vk_device, &context->m_MainTextureDepthStencil.m_Handle);
+        DestroyDeviceBuffer(vk_device, &context->m_DefaultTexture2D->m_DeviceBuffer.m_Handle);
```
Jhonnyg, line 4089: We never released the backing device buffers (the GPU memory) previously. It was not an issues since it's part of the shutdown, but still.

engine/graphics/src/vulkan/graphics_vulkan_device.cpp
```diff
@@ -860,7 +860,6 @@ namespace dmGraphics
         else if (tex_type == TEXTURE_TYPE_2D_ARRAY)
         {
             assert(imageLayers > 0);
-            vk_image_create_info.flags |= VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT;
```
Jhonnyg, line 863: This was apparently wrong!

engine/graphics/src/vulkan/graphics_vulkan.cpp
JCash: Yeah, it would have I guess, if a game would reboot (which they do on mobile)

JCash (approved):

PR #11206, opened 2025-09-08
Fix default build server in beta

### LINKED ISSUE #11174: Beta editors do not use the beta build server by default
Apparently the beta editor builds should use the beta build server by default as long the value is not set in the preferences. That is not my experience as I always had to set the build server to "https://build-stage.defold.com" when using beta editors.

**To Reproduce (REQUIRED)**
1. Go to editor preferences > Extensions in a beta editor.
2. See the build server input field with placeholder text "https://build.defold.com" instead of "https://build-stage.defold.com"

Beta editor uses the beta build server when the build server input field is empty (displays the placeholder text).

  - Version 1.11.0 BETA and earlier.

**Platforms (REQUIRED):**
 - Platforms: Windows /  Linux
 - OS: Windows 10 / Arch Linux
 - Device: PC

vlaaad:
Fixes #11174

### DIFF
--- a/editor/src/clj/editor/connection_properties.clj
+++ b/editor/src/clj/editor/connection_properties.clj
@@ -28,7 +28,7 @@
    :updater {:download-url-template "https://%s/archive/%s/%s/editor2/Defold-%s.zip"
              :update-url-template "https://%s/editor2/channels/%s/update-v4.json"}})
 
-(def ^:const defold-build-server-url
+(defn defold-build-server-url []
   (or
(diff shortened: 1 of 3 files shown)


vlaaad:
I checked that it works with defn and doesn't work with const by bundling the editor locally with this:
```
./scripts/bundle.py build bundle --platform=arm64-macos --version beta-channel --engine-artifacts 37c7e0f410a2affb6daeb454546682c2789e9a9f --channel=beta --skip-tests
```

matgis (approved):
LGTM!

PR #11213, opened 2025-09-09
Implement 10x faster lines reader in Java

vlaaad:
We get 10x improvement by using a Java implementation instead of a Clojure one that uses a proxy.

Benchmark:
```clj

(require 'criterium.core)

(defn- discard [^Reader reader]
  (let [buf (char-array 1024)]
    (while (not= -1 (.read reader buf)))))

(let [lines (dynamo.graph/node-value (dev/resource-node "/bob.editor_script") :lines)]
  ;; old impl
  (criterium.core/quick-bench (discard (lines-reader lines)))
  ;; new impl
  (criterium.core/quick-bench (discard (com.defold.editor.code.data.LinesReader. lines))))
```
Results:
```
WARNING: Final GC required 30.495942082913107 % of runtime
Evaluation count : 17298 in 6 samples of 2883 calls.
             Execution time mean : 35,407943 µs
    Execution time std-deviation : 1,156964 µs
   Execution time lower quantile : 34,636158 µs ( 2,5%)
   Execution time upper quantile : 37,337036 µs (97,5%)
                   Overhead used : 1,993880 ns

Found 1 outliers in 6 samples (16,6667 %)
    low-severe     1 (16,6667 %)
 Variance from outliers : 13,8889 % Variance is moderately inflated by outliers
WARNING: Final GC required 30.21848695903067 % of runtime
Evaluation count : 186792 in 6 samples of 31132 calls.
             Execution time mean : 3,228951 µs
    Execution time std-deviation : 23,331207 ns
   Execution time lower quantile : 3,207523 µs ( 2,5%)
   Execution time upper quantile : 3,255436 µs (97,5%)
                   Overhead used : 1,993880 ns
```

### DIFF
--- a/editor/src/clj/editor/code/data.clj
+++ b/editor/src/clj/editor/code/data.clj
@@ -18,8 +18,8 @@
             [editor.code.syntax :as syntax]
             [editor.code.util :as util]
             [util.coll :refer [pair]])
-  (:import [java.io IOException InputStream Reader Writer]
-           [java.nio CharBuffer]
+  (:import [com.defold.editor.code.data LinesReader]
+           [java.io InputStream Reader Writer]
            [java.util Collections]
            [java.util.regex MatchResult Pattern]
            [org.apache.commons.io.input ReaderInputStream]))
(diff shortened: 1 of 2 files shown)


vlaaad:
Ugh, unfortunately, there is no measurable impact on, e.g., big projects' load/build speed as a whole. Not worth it.

PR #11214, opened 2025-09-09
Remove open as code option from binary resources

sprocketc:
Remove code view option from binary resources.

Fixes #11192

### Technical changes
The resource is not available when we register the `placeholder-resource-type-ext`, so we need to filter out the option later. I think that `:view-types` could be a function that produces the type vector based on a resource input, but I am not sure if we want to change that.

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -2135,8 +2135,8 @@ If you do not specifically require different script states, consider changing th
 (defn view-types
   [resource]
   (cond->> (:view-types (resource/resource-type resource))
-    (text-util/binary? resource)
-    (e/filter #(not= :code (:id %)))))
+           (text-util/binary? resource)
+           (e/filter #(not= :code (:id %)))))
 
 (defn open-resource
   ([app-view prefs workspace project resource]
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/resource.clj
```diff
@@ -132,11 +132,6 @@
   (and (openable-resource? value)
        (true? (:editor-openable (resource-type value)))))
 
-(defn has-view-type? [resource view-type]
-  {:pre [(keyword? view-type)]}
-  (boolean (some #(= view-type (:id %))
-                 (:view-types (resource-type resource)))))
-
```
sprocketc, line 139: Unrelated unused fn.

editor/src/clj/editor/app_view.clj
```diff
@@ -2655,7 +2661,7 @@ If you do not specifically require different script states, consider changing th
 (defn- gen-tooltip [workspace project app-view resource]
   (when (resource/loaded? resource)
     (let [resource-type (resource/resource-type resource)
-          view-type (or (first (:view-types resource-type))
+          view-type (or (first (view-types resource))
```
matgis: I would prefer not to use the more sophisticated view-type check here, since it reads part of the file to determine if it is binary or not. The `gen-tooltip` function appears to be called for every resource presented in asset picker dialogs, so introducing a disk read per resource could slow things down. The `:code` view does not specify `:make-preview-fn` anyway.

matgis (changes requested):

matgis (approved):
LGTM! ❤️

PR #11218, opened 2025-09-10
New functions: `gui.set_layout()` and `gui.get_layouts()`

### LINKED ISSUE #5889: New function: `gui.set_layout()`
When you have two very similar but a bit different GUI (you can use the same code for logic, but a bit different positions or/and icons here and there), it would be easier to have two layoust for these GUI instead of creation two separate `gui` components.

Also, sometimes it's convenient to own conditions for layouts you wanna use like special layout only for some particular device or simple aspect ration calculation instead of current mechanism.

`gui.set_layout("my_layout")` or `gui.set_layout(hash("my_layout"))`

There is no alternatives for that.

AGulev:
Added new functions `gui.set_layout()` and `gui.get_layouts()`:

```lua
local ok = gui.set_layout("Portrait")
  if not ok then
      print("Portrait layout not found in this scene")
  end
```

```lua
local layouts = gui.get_layouts()
for id, size in pairs(layouts) do
    print(id, size.x, size.y)
end
```

A new checkbox, `Auto Layout Selection`, has been added to Display Profiles. It disables automatic layout detection in the project:  

Fix https://github.com/defold/defold/issues/5889

### DIFF
--- a/editor/src/clj/editor/display_profiles.clj
+++ b/editor/src/clj/editor/display_profiles.clj
@@ -17,6 +17,7 @@
             [editor.build-target :as bt]
             [editor.graph-util :as gu]
             [editor.protobuf :as protobuf]
+            [editor.protobuf-forms-util :as protobuf-forms-util]
             [editor.resource :as resource]
             [editor.resource-node :as resource-node]
             [editor.workspace :as workspace]
@@ -77,7 +78,10 @@
   {:navigation false
    :sections
    [{:title "Display Profiles"
-     :fields [{:path [:profiles]
+     :fields [{:path [:auto-layout-selection]
+               :label "Auto Layout Selection"
+               :type :boolean}
+              {:path [:profiles]
                :label "Profile"
                :type :2panel
                :panel-key {:path [:name] :type :string :default "New Display Profile"}
@@ -99,7 +103,11 @@
                                                    :type :string}]}]}]}}]}]})
 
 (g/defnk produce-form-data [_node-id form-data-desc form-values]
-  (assoc form-data-desc :values form-values))
+  (-> form-data-desc
+      (assoc :values form-values)
+      (assoc :form-ops {:user-data {:node-id _node-id}
+                        :set protobuf-forms-util/set-form-op
+                        :clear protobuf-forms-util/clear-form-op})))
 
 (defn- build-pb [resource dep-resources user-data]
   (let [pb  (:pb user-data)
@@ -123,13 +131,18 @@
 
 (g/defnode DisplayProfilesNode
   (inherits resource-node/ResourceNode)
+  (property auto-layout-selection g/Bool (default true)
+            (dynamic visible (g/constantly false)))
 
   (input profile-msgs g/Any :array)
-  (output save-value g/Any :cached (g/fnk [profile-msgs]
+  (output save-value g/Any :cached (g/fnk [profile-msgs auto-layout-selection]
                                      (protobuf/make-map-without-defaults Render$DisplayProfiles
-                                       :profiles profile-msgs)))
+                                       :profiles profile-msgs
+                                       :auto-layout-selection (boolean auto-layout-selection))))
   (input profile-form-values g/Any :array)
-  (output form-values g/Any (g/fnk [profile-form-values] {[:profiles] profile-form-values}))
+  (output form-values g/Any (g/fnk [profile-form-values auto-layout-selection]
+                                   {[:profiles] profile-form-values
+                                    [:auto-layout-selection] (boolean auto-layout-selection)}))
   (output form-data-desc g/Any :cached produce-form-data-desc)
   (output form-data g/Any :cached produce-form-data)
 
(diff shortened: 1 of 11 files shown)


editor/src/clj/editor/display_profiles.clj
```diff
@@ -123,13 +131,18 @@
 
 (g/defnode DisplayProfilesNode
   (inherits resource-node/ResourceNode)
+  (property auto-layout-selection g/Bool (default true)
+            (dynamic visible (g/constantly false)))
```
AGulev, line 135: When I started, I was sure that there should be a way to avoid property creation and operate directly with protobuf. But I didn't figure out how to do so. Please let me know if it can be simlified.

engine/gui/src/gui_script.cpp
```diff
@@ -2609,6 +2610,99 @@ namespace dmGui
...
+    /*# returns available layouts with sizes
+     *
+     * Returns a table mapping each layout id hash to a vector3(width, height, 0). For the default layout,
+     * the current scene resolution is returned. If a layout name is not present in the Display Profiles,
+     * the width/height pair is 0.
+     *
+     * @name gui.get_layouts
+     * @return [type:table] layout_id_hash -> vmath.vector3(width, height, 0)
+     */
+    static int LuaGetLayouts(lua_State* L)
```
AGulev, line 2667: The project contains many GUIs, and it might be hard to write a universal logic, that switches layouts without an ability to get all the layouts of the GUI component.

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -660,6 +660,11 @@ namespace dmGameSystem
 
     static void OnWindowResizeCallback(const dmGui::HScene scene, uint32_t width, uint32_t height)
     {
+        dmRender::HDisplayProfiles display_profiles = (dmRender::HDisplayProfiles) dmGui::GetDisplayProfiles(scene);
+        if (display_profiles && !dmRender::GetAutoLayoutSelection(display_profiles))
```
britzl: Can display_profiles really be null?

engine/gui/src/gui.cpp
```diff
@@ -395,6 +395,8 @@ namespace dmGui
         scene->m_DestroyRenderConstantsCallback = params->m_DestroyRenderConstantsCallback;
         scene->m_CloneRenderConstantsCallback = params->m_CloneRenderConstantsCallback;
         scene->m_OnWindowResizeCallback = params->m_OnWindowResizeCallback;
+        scene->m_ApplyLayoutCallback    = params->m_ApplyLayoutCallback;
```
britzl, line 398: Why are all of these called something with Callback? Seems strange... Shouldn't it be ApplyLayoutFn, SetMaterialPropertyFn etc?

engine/gamesys/src/gamesys/components/comp_gui.cpp
AGulev: we can't communicate `gui`<->`gamesys` another way

engine/gui/src/gui.cpp
AGulev: we can't communicate `gui`<->`gamesys` another way

AGulev: If you mean naming: I agree it should be `Fn`, but I in this PR I just adapt to what we already have

britzl: Yep, I was only referring to naming.

engine/gamesys/src/gamesys/components/comp_gui.cpp
britzl: Wrong comment?

AGulev: Fixed!

engine/gui/src/gui.cpp
AGulev: I can rename all the functions all over the code, ofc. but I'm not sure if we should do that as part of that PR

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -701,6 +705,68 @@ namespace dmGameSystem
...
+        memset(message, 0, sizeof(dmMessage::Message));
+        message->m_Sender = dmMessage::URL();
+        message->m_Receiver = dmMessage::URL();
+        message->m_Id = dmHashString64("layout_changed");
+        message->m_Descriptor = (uintptr_t)dmGuiDDF::LayoutChanged::m_DDFDescriptor;
+        message->m_DataSize = sizeof(dmGuiDDF::LayoutChanged);
+        dmGuiDDF::LayoutChanged* message_data = (dmGuiDDF::LayoutChanged*)message->m_Data;
+        message_data->m_Id = layout_id;
+        message_data->m_PreviousId = current_layout_id;
+        DispatchMessage(scene, message);
```
britzl: There's quite a bit of shared code with OnWindowResizeCallback. Should we move it to a function used in both places?

editor/src/clj/editor/display_profiles.clj
vlaaad: I think this can't be simplified; we typically assign all protobuf fields to node properties.

vlaaad (dismissed):
Editor changes look good!

engine/gamesys/src/gamesys/components/comp_gui.cpp
AGulev: True, I had plans to extract it, but forgot about it
Thank you. Now it is fixed.

britzl (approved):

PR #11222, opened 2025-09-12
Added orthographic camera projection modes 

### LINKED ISSUE #6849: Add orthographic camera projection settings
We've recently improved the camera component so that it can send an orthographic, centered and optionally zoomed projection to the render script. We should also consider adding additional configuration options to the orthographic projection, similar to what we have in the render script today.

* FIXED ZOOM - Use the predefined zoom
* AUTO_ZOOM - Use automatic zoom to always show the original display area (from game.project)
   * There are two different zoom options here: https://github.com/defold/defold/issues/6270

### LINKED ISSUE #10579: [SUGGESTION] Default render script should support alternative projections for all cameras
The default render script supports messages to change the camera projection. E.g. use_stretch_projection, use_fixed_fit_projection. However, this only works for the default camera. If you add a camera component, you will need to write your own render script to support alternative projections. Please update the default render script so that projections are applied to all cameras.

The default projection should also be fixed fit as this is how most games work. You always want to see the same content regardless of the resolution, except for variations in aspect ratio where you want to see either black bars or more content (e.g. 1280x800 vs 1280x720). This would also work better on mobile devices with similar aspect ratios. Stretch fit will distort the image, you want to scale the game to the device resolution while maintaining the aspect ratio.

AGulev:
New functionality for camera added:
- Added orthographic camera modes: ORTHO_MODE_FIXED, ORTHO_MODE_AUTO_FIT (contain), and ORTHO_MODE_AUTO_COVER (zoom).
- New Camera properties in Editor: “Orthographic Mode” (available only when Orthographic Projection is enabled).
- Lua API: `camera.get_orthographic_mode()` and `camera.set_orthographic_mode()` 

```lua
camera.get_orthographic_mode("#camera")
camera.set_orthographic_mode("#camera", camera.ORTHO_MODE_AUTO_FIT)
```

Fix https://github.com/defold/defold/issues/6849
Fix https://github.com/defold/defold/issues/10579

### DIFF
--- a/editor/resources/templates/template.camera
+++ b/editor/resources/templates/template.camera
@@ -5,4 +5,4 @@ far_z: 1000.0
 auto_aspect_ratio: 0
 orthographic_projection: 0
 orthographic_zoom: 1
-orthographic_zoom_mode: ORTHO_ZOOM_MODE_FIXED
+orthographic_mode: ORTHO_MODE_FIXED
--- a/editor/resources/templates/template.camera
+++ b/editor/resources/templates/template.camera
@@ -5,3 +5,4 @@ far_z: 1000.0
 auto_aspect_ratio: 0
 orthographic_projection: 0
 orthographic_zoom: 1
+orthographic_zoom_mode: ORTHO_ZOOM_MODE_FIXED
--- a/editor/src/clj/editor/camera_editor.clj
+++ b/editor/src/clj/editor/camera_editor.clj
@@ -18,6 +18,7 @@
             [dynamo.graph :as g]
             [editor.build-target :as bt]
             [editor.camera :as camera]
+            [editor.properties :as properties]
             [editor.colors :as colors]
             [editor.geom :as geom]
             [editor.gl :as gl]
@@ -28,10 +29,11 @@
             [editor.math :as math]
             [editor.outline :as outline]
             [editor.protobuf :as protobuf]
+            [editor.protobuf-forms :as protobuf-forms]
             [editor.protobuf-forms-util :as protobuf-forms-util]
             [editor.resource-node :as resource-node]
             [editor.workspace :as workspace])
-  (:import [com.dynamo.gamesys.proto Camera$CameraDesc]
+  (:import [com.dynamo.gamesys.proto Camera$CameraDesc Camera$OrthoZoomMode]
            [com.jogamp.opengl GL GL2]
            [javax.vecmath Matrix4d Quat4d Vector3d Vector4d]))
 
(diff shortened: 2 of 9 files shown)


engine/gamesys/proto/gamesys/camera_ddf.proto
```diff
@@ -17,6 +17,16 @@ option java_outer_classname = "Camera";
...
  */
 
+// NOTE: Enum values must correspond to the enum values in dmRender::OrthoZoomMode
+// Don't forget to change engine/render/src/render/render.h if you change here
+// Orthographic camera zoom modes
+enum OrthoZoomMode
+{
+    ORTHO_ZOOM_MODE_FIXED        = 0[(displayName) = "Fixed"]; // Use orthographic_zoom as-is
+    ORTHO_ZOOM_MODE_AUTO_FIT     = 1[(displayName) = "Auto Fit"]; // Fit original display area (fit)
+    ORTHO_ZOOM_MODE_AUTO_COVER   = 2[(displayName) = "Auto Cover"]; // Fill original display area (cover)
```
AGulev: In gui we call it `zoom`, should I use zoom here as well? but then it will be `ORTHO_ZOOM_MODE_AUTO_ZOOM` =/
 I didn't add stretch since we don't get what the usecase it.

engine/render/src/render/camera.cpp
```diff
@@ -119,7 +120,45 @@ namespace dmRender
...
+
+            // Fallbacks to avoid division by zero
+            if (proj_width <= 0.0f)
+            {
+            	proj_width = 1.0f;
+            }
+            if (proj_height <= 0.0f)
+            {
+            	proj_height = 1.0f;
+            }
```
AGulev, line 139: It happens from time to time and it's better to avoid it.

engine/gamesys/proto/gamesys/camera_ddf.proto
vlaaad: I think this is typically called "Fill" (it's even in the comment...), maybe we should call it `ORTHO_ZOOM_MODE_AUTO_FILL`?

AGulev: yeah, many names of the same are confusing (I called `fit` -> `contain` initially, forgot to change value in the comment)
I don't mind, whatever better for understanding. I don't like we use `zoom` for it in gui =/

AGulev: from css for image: 
CleanShot 2025-09-12 at 14 12 53@2x

vlaaad: Where is this from?

AGulev: https://www.w3schools.com/css/css3_object-fit.asp

editor/src/clj/editor/camera_editor.clj
```diff
@@ -55,51 +57,57 @@
...
-                        {:path [:auto-aspect-ratio]
-                         :label "Auto Aspect Ratio"
-                         :type :boolean}
-                        {:path [:orthographic-projection]
-                         :label "Orthographic Projection"
-                         :type :boolean}
-                        {:path [:orthographic-zoom]
-                         :label "Orthographic Zoom"
-                         :type :number}]}]
+               :fields (cond-> [{:path [:aspect-ratio]
```
vlaaad: Please remove useless `cond->`

vlaaad (dismissed):
Works great!

editor/src/clj/editor/camera_editor.clj
AGulev: thanks! fixed

vlaaad (dismissed):

editor/resources/templates/template.camera
```diff
@@ -4,4 +4,5 @@ near_z: 0.1
 far_z: 1000.0
 auto_aspect_ratio: 0
 orthographic_projection: 0
-orthographic_zoom: 1
\ No newline at end of file
+orthographic_zoom: 1
+orthographic_zoom_mode: ORTHO_ZOOM_MODE_FIXED
```
AGulev: maybe default should be Fixed FIt ?

engine/gamesys/proto/gamesys/camera_ddf.proto
AGulev: @britzl What’s your opinion on the naming? I’m not happy about introducing a new term that isn’t used anywhere else in the engine.

britzl: The terminology we have in the render script is actually not that great. We should probably have researched it better. But on the other hand there's several different naming schemes out there if you do the research. The CSS object fit property values linked above is the same as that which Rive decided to use for their layout Fit Modes:

* Contain (Default): The content scales to fit within the view's bounds while preserving its aspect ratio, leaving empty space if the ratios don't match. 
* Cover: The content scales to fill the view's bounds, maintaining the aspect ratio, but parts of the content may be cropped if there's a mismatch in ratios. 
* Fill: The content stretches to fill the entire view, changing the aspect ratio if necessary. 
* FitWidth: The content scales to match the view's width, potentially resulting in clipping or unused vertical space. 
* FitHeight: The content scales to match the view's height, potentially resulting in clipping or unused horizontal space.

The Contain mode from Rive and CSS Object Fit are the same as the Fixed Fit mode of our render script.

The Fill mode from Rive and CSS Object Fit are the same as our Stretch mode. Like mentioned it makes little sense to include it as it is rarely (never) useful to stretch all content. 

We don't have a Cover mode in our render script. Maybe we should have? And FitWidth and FitHeight are variations of Cover which might be nice to add at some point. 

We also have a fixed projection which renders with maintained aspect ratio and a fixed zoom. Rive doesn't have an equivalent. I suppose it is similar to `none` from CCS Object Fit?

AGulev: Yeah, I added Cover here for camera, because we have it in gui as 'zoom'

Should we rename the rest of modes?

britzl: By the way, does it actually have to be called ORTHO_ZOOM_MODE_XYZ? Can't it be named ORTHO_MODE_XYZ? It doesn't really have to do with zoom, even though we calculate the zoom to get the desired result.

britzl (changes requested):
See note about naming of ORTHO_ZOOM_MODE_XYZ. I propose to drop ZOOM from the name.

editor/resources/templates/template.camera
AGulev: @britzl what do you think about default? Should it be FIXED as it is now?

britzl (approved):
Looks good!

PR #11224, opened 2025-09-12
Add support for enabling file logging in release mode  

### LINKED ISSUE #10255: Allow enabling logging-to-file in release mode
I would like to have normal logging in release mode. Specifically, I would like to have my release-mode Steam (PC) game output to `log.txt` naturally using Defold's logging system just like it does in debug-mode, without having to write a bunch of file management code myself in C++.

This is *actively* and critically useful as a tool for me to debug ongoing issues with my users, especially ones during startup where I may not even have initialized all of my own code. I also have users actively requesting this for user instrumentation based on my logs.

In order of preference:
- A desktop-platform-specific option or appmanifest flag that turns on logging even in release mode (so that it's enabled from the moment the engine starts)
- A `dmLog*` function to turn logging on that works outside debug mode 
- Access to `dmLog::DoLogPlatform` in my C extension

I have a log listener in a C extension that I could- if needed- add my own code to log-to-text-file. I'd strongly rather not deal with managing platform-specific code to perform file operations when the ideal code is sitting right there in the engine, one boolean away.

I have perused the public interface to dmLog and come up with no other alternates

Currently the normal logging call is gated purely by `dLib::IsDebugMode()` as seen here:
- https://github.com/defold/defold/blob/4e8483c3143df92f66ae6086d4d2cfb0e843c8a7/engine/dlib/src/dlib/log.cpp#L634
- https://github.com/defold/defold/blob/4e8483c3143df92f66ae6086d4d2cfb0e843c8a7/engine/dlib/src/dlib/log.cpp#L689C9-L689C22

AGulev:
You can now enable file logging in Release mode.  
The previous `Write Log` option has been deprecated and replaced with a new `Write Log File` setting, which offers three choices: `Never`, `Debug`, and `Always`.  

CleanShot 2025-09-12 at 15 22 00@2x

Fix https://github.com/defold/defold/issues/10255

### DIFF
--- a/engine/dlib/src/dlib/log.cpp
+++ b/engine/dlib/src/dlib/log.cpp
@@ -320,12 +320,6 @@ static void DoLogPlatform(LogSeverity severity, const char* output, int output_l
 #elif !defined(ANDROID)
         fwrite(output, 1, output_len, stderr);
 #endif
-
-    if (dmLog::g_LogFile && dmLog::g_TotalBytesLogged < dmLog::MAX_LOG_FILE_SIZE) {
-        dmLog::g_TotalBytesLogged += output_len;
-        fwrite(output, 1, output_len, dmLog::g_LogFile);
-        fflush(dmLog::g_LogFile);
-    }
 }
 
 // Here we put logging that needs to be thread safe
@@ -630,9 +624,9 @@ void LogInternal(LogSeverity severity, const char* domain, const char* format, .
         return;
     }
 
-    // In release mode, if there are no custom listeners, we'll return here
+    // In release mode, if there are no custom listeners, and no log.txt file, we'll return here
     bool is_debug_mode = dLib::IsDebugMode();
-    if (!is_debug_mode && (dmAtomicGet32(&dmLog::g_ListenersCount) == 0))
+    if (!is_debug_mode && !dmLog::g_LogFile && (dmAtomicGet32(&dmLog::g_ListenersCount) == 0))
     {
         return;
     }
@@ -689,6 +683,13 @@ void LogInternal(LogSeverity severity, const char* domain, const char* format, .
     if (is_debug_mode)
     {
         dmLog::DoLogPlatform(severity, str_buf, actual_n);
+    }
+
+    if (dmLog::g_LogFile && dmLog::g_TotalBytesLogged < dmLog::MAX_LOG_FILE_SIZE)
+    {
+        dmLog::g_TotalBytesLogged += actual_n;
+        fwrite(str_buf, 1, actual_n, dmLog::g_LogFile);
+        fflush(dmLog::g_LogFile);
     }
 
     if (!dmLog::IsServerInitialized()) // in case the server lock isn't even created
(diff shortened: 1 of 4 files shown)


engine/dlib/src/dlib/log.cpp
```diff
@@ -320,12 +320,6 @@ static void DoLogPlatform(LogSeverity severity, const char* output, int output_l
 #elif !defined(ANDROID)
         fwrite(output, 1, output_len, stderr);
 #endif
-
-    if (dmLog::g_LogFile && dmLog::g_TotalBytesLogged < dmLog::MAX_LOG_FILE_SIZE) {
-        dmLog::g_TotalBytesLogged += output_len;
-        fwrite(output, 1, output_len, dmLog::g_LogFile);
-        fflush(dmLog::g_LogFile);
-    }
```
AGulev, line 328: Moved into `LogInternal` - now it's independent of the rest of logs.

engine/dlib/src/dlib/log.cpp
```diff
@@ -630,9 +624,9 @@ void LogInternal(LogSeverity severity, const char* domain, const char* format, .
         return;
     }
 
-    // In release mode, if there are no custom listeners, we'll return here
+    // In release mode, if there are no custom listeners, and no log.txt file, we'll return here
     bool is_debug_mode = dLib::IsDebugMode();
-    if (!is_debug_mode && (dmAtomicGet32(&dmLog::g_ListenersCount) == 0))
+    if (!is_debug_mode && !dmLog::g_LogFile && (dmAtomicGet32(&dmLog::g_ListenersCount) == 0))
```
AGulev, line 629: One more condition, if we have a `LogFile` - we want logs

engine/engine/src/engine.cpp
```diff
@@ -930,8 +930,12 @@ namespace dmEngine
...
         instance_index = dmConfigFile::GetInt(engine->m_Config, "project.instance_index", 0);
 #endif
-        int write_log = dmConfigFile::GetInt(engine->m_Config, "project.write_log", 0);
-        if (write_log) {
+        int write_log = dmConfigFile::GetInt(engine->m_Config, "project.write_log", 0); // Deprecated
+        int write_log_file = dmConfigFile::GetInt(engine->m_Config, "project.write_log_file", 0);
+        // for backward compatibility if write_log_file is 0, but write_log is 1
+        write_log_file = write_log_file == 0 ? write_log : write_log_file;
+        // 0 - no logs, 1 - debug only, 2 - always
+        if (write_log_file + dLib::IsDebugMode() >= 2) {
```
AGulev: 0 + 1 = 1 no logs
1 + 0 (is_debug) = 1 no logs
1 + 1 = 2 logs
2 + 1 = 3  logs

britzl: I get it but it makes my head hurt a bit. This on the other hand is immediately obvious (to me at least):

```cpp
if ((write_log_file == 2) || (write_log_file == 1 && dLib::IsDebugMode())) {
	
}
```

AGulev: Fixed!

britzl (approved):

PR #11227, opened 2025-09-15
Package simd enabled/disabled versions of Box2Dv3 the same way in local and CI builds.

deurell:
Package simd enabled/disabled versions of Box2Dv3 the same way in local and CI builds.

### Technical changes
* CI workflow and local waf install of Box2Dv3 now generates the same packages

### DIFF
--- a/external/box2d/Box2D/wscript
+++ b/external/box2d/Box2D/wscript
@@ -7,41 +7,56 @@ VERSION = '3.1.0'
 APPNAME = 'box2d'
 
 def configure(conf):
-    pass
+    plat = conf.env.PLATFORM
+    if not plat:
+        conf.fatal('Pass --platform=...')
+
+    conf.env.SIMD_CFLAGS,   conf.env.SIMD_CXXFLAGS   = [], []
+    conf.env.NOSIMD_CFLAGS, conf.env.NOSIMD_CXXFLAGS = [], []
+
+    if plat in ('win32', 'x86_64-win32'):
+        conf.env.STLIB_ST         = 'lib%s.lib'
+        conf.env.cstlib_PATTERN   = 'lib%s.lib'
+        conf.env.cxxstlib_PATTERN = 'lib%s.lib'
+        conf.env.append_unique('CFLAGS', ['/std:c11'])
+
+    if plat == 'armv7-android':
+        conf.env.SIMD_CFLAGS   += ['-mfpu=neon']
+        conf.env.SIMD_CXXFLAGS += ['-mfpu=neon']
+
+    if plat in ('wasm-web', 'wasm_pthread-web'):
+        conf.env.SIMD_CFLAGS   += ['-msimd128']
+        conf.env.SIMD_CXXFLAGS += ['-msimd128']
+    elif plat == 'js-web':
+        pass
 
 def build(bld):
-    # Build simd version
-    box2d = bld.stlib(features = 'c',
-                      includes = '.. include',
-                      source = bld.path.ant_glob('**/*.c'),
-                      target = 'box2d')
+    # Sources/Includes
+    srcs = bld.path.ant_glob('**/*.c')
+    incs = '.. include'
 
-    # Build simd-disabled version
-    box2d_nosimd = bld.stlib(features = 'c',
-                             includes = '.. include',
-                             source = bld.path.ant_glob('**/*.c'),
-                             target = 'box2d_nosimd')
+    box2d = bld.stlib(features='c',
+                      includes=incs,
+                      source=srcs,
+                      target='box2d',
+                      cflags=bld.env.SIMD_CFLAGS,
+                      cxxflags=bld.env.SIMD_CXXFLAGS)
 
-    # Add SIMD disable flag to the nosimd version
-    box2d_nosimd.env.append_unique('CFLAGS', '-DBOX2D_DISABLE_SIMD')
-    box2d_nosimd.env.append_unique('CXXFLAGS', '-DBOX2D_DISABLE_SIMD')
+    box2d_nosimd = bld.stlib(features='c',
+                             includes=incs,
+                             source=srcs,
+                             target='box2d_nosimd',
+                             defines=['BOX2D_DISABLE_SIMD'],
+                             cflags=bld.env.NOSIMD_CFLAGS,
+                             cxxflags=bld.env.NOSIMD_CXXFLAGS)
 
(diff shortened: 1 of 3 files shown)


external/box2d/Box2D/wscript
```diff
@@ -6,95 +6,107 @@ from waflib import Build
...
-                      source = bld.path.ant_glob('**/*.c'),
-                      target = 'box2d')
+    conf.env.SIMD_CFLAGS, conf.env.SIMD_CXXFLAGS = [], []
+    conf.env.NOSIMD_CFLAGS, conf.env.NOSIMD_CXXFLAGS = [], []
+
+    if plat in ('win32', 'x86_64-win32'):
+        conf.env.STLIB_ST = 'lib%s.lib'
+        conf.env.cstlib_PATTERN = 'lib%s.lib'
+        conf.env.cxxstlib_PATTERN = 'lib%s.lib'
+        conf.env.append_unique('CFLAGS', ['/std:c11'])
```
JCash, line 25: I just noticed this in some other wscript, but this shouldn't be needed. waf_dynamo.py does this.
That's the point of our external/ folder, that things should build just the same way as the rest of our buildsystem.

Compare with external/glfw/wscript that does `import waf_dynamo` and also `conf.load('waf_dynamo')`

external/box2d/Box2D/wscript
```diff
@@ -6,95 +6,107 @@ from waflib import Build
...
+    if not plat:
+        conf.fatal('Pass --platform=...')
 
-def build(bld):
-    box2d = bld.stlib(features = 'c',
-                      includes = '.. include',
-                      source = bld.path.ant_glob('**/*.c'),
-                      target = 'box2d')
+    conf.env.SIMD_CFLAGS, conf.env.SIMD_CXXFLAGS = [], []
+    conf.env.NOSIMD_CFLAGS, conf.env.NOSIMD_CXXFLAGS = [], []
```
JCash: To do idiomatic Waf, this would be:
```
confirming.env['CFLAGS_SIMD'] = ...
confirming.env['CXXFLAGS_SIMD'] = ...
```

and then you'd use the feature like so:
```python
bld.stlib(features='c',
    use = 'SIMD',
    includes=incs,
    source=srcs,
    target='box2d')
```

external/box2d/Box2D/wscript
```diff
@@ -6,95 +6,107 @@ from waflib import Build
...
+                          cflags=bld.env.SIMD_CFLAGS,
+                          cxxflags=bld.env.SIMD_CXXFLAGS)
+        bld.install_files(install_dir, bld.env.cxxstlib_PATTERN % 'box2d')
+
+    box2d_nosimd = bld.stlib(features='c',
+                             includes=incs,
+                             source=srcs,
+                             target='box2d_nosimd',
+                             cflags=bld.env.NOSIMD_CFLAGS + ['-DBOX2D_DISABLE_SIMD'],
+                             cxxflags=bld.env.NOSIMD_CXXFLAGS + ['-DBOX2D_DISABLE_SIMD'])
```
JCash: Defines go into the `defines = ['BOX2D_DISABLE_SIMD'` segment.

external/box2d/Box2D/wscript
```diff
@@ -6,95 +6,107 @@ from waflib import Build
...
+    bld.install_files(install_dir, bld.env.cxxstlib_PATTERN % 'box2d_nosimd')
+
+    if plat == 'js-web':
+        simd_name = bld.env.cxxstlib_PATTERN % 'box2d'
+        nosimd_name = bld.env.cxxstlib_PATTERN % 'box2d_nosimd'
+        nosimd_node = bld.path.get_bld().make_node(nosimd_name)
+        bld.install_as(install_dir + simd_name, nosimd_node)
```
JCash, line 72: I don't understand this part. Is this to make a copy, to make sure we have two libs?
Not sure I like it, as it is confusing. But we'll go with it for now I think.

JCash (changes requested):
overall good, some cleanup needed

external/box2d/Box2D/wscript
deurell: Since simd isn't available in plain javascript builds I had two options for this. Making an additional specific switch/logic not allowing simd just for the plain javascript/non wasm builds with additional logic or just package two nosimd versions for this platform. I went for this option in this case.

external/box2d/Box2D/wscript
deurell: Followed the pattern from earlier version and i think it was bullet that also used this way. If i remove it windows builds fail when i call it from CI/GH action even if i do a conf.load('waf_dynamo') in configure. I guess it should pick it up from default_flags in waf_dynamo but for some reason it doesn't. Will try to figure out why.

```
echo "DYNAMO_HOME=$DYNAMO_HOME"
python ./scripts/build.py install_waf
cd external/box2d
waf distclean || true
PREFIX="$DYNAMO_HOME" waf configure --platform=${{ matrix.platform }}
waf install
```

https://github.com/defold/defold/actions/runs/17761067614/job/50473757125

deurell: I checked the GLFW CI/GH build and it is broken on windows, could be related.
```
Waf: Leaving directory `D:\a\defold\defold\external\box2d\build'
source not found: 'box2d.lib' in bld(source=[], target='', meths=['process_install_task', 'process_rule', 'process_source'], features=['install_task'], path=D:\a\defold\defold\external\box2d\Box2D, idx=2, tg_idx_count=2, install_to='${PREFIX}/box2d-3.1.0/lib/win32/', install_from='box2d.lib', dest='${PREFIX}/box2d-3.1.0/lib/win32/', type='install_files', posted=True, install_task=
```

https://github.com/defold/defold/actions/runs/17761883825/job/50476331936

external/box2d/Box2D/wscript
deurell: Changing it to conf.env['CFLAGS_SIMD'] = ...

external/box2d/Box2D/wscript
deurell: just saw that `external/glfw/wscript` does
```
    if bld.env.PLATFORM in ('win32', 'x86_64-win32'):
        bld.env.cxxstlib_PATTERN = 'lib%s.lib'
```

in build. Missed that one earlier.

JCash (approved):

PR #11228, opened 2025-09-15
Move hid, input and platform libraries to use CMake+Ninja build system instead of Waf

JCash:
Note that we've updated our appmanifest tool in the editor. Windows libraries no longer need to be prefixed/suffixed in the form `lib<name>.lib`. We should now be able to start using the `common` platform for most such changes.

For backwards compatibility, the build server will still support old style appmanifests.

### Technical changes

**NOTE**: The wscript files are still there and intact, in case we need to temporarily "revert" this change, but for now there is no build option for it. Would be easy to add, but let's try to make cmake work first!

Added a `./scripts/cmake/README.md` as an entry point, explaining the current ideas behind the changes and code structure + flow.

#### Performance

* Locally on macOS
```
----------------------------------------
Waf
----------------------------------------

    Configure:
        0.62s user 0.33s system 159% cpu 0.600 total
        
    Configure + Build + Install:
        1.55s user 0.55s system 192% cpu 1.093 total

    Incremental:
        0.38s user 0.20s system 180% cpu 0.323 total


----------------------------------------
CMake + Ninja
----------------------------------------

    Configure:
        0.34s user 0.29s system 93% cpu 0.676 total

    Configure + Build + Install:
        1.11s user 0.29s system 135% cpu 1.039 total

    Incremental:
        0.01s user 0.01s system 76% cpu 0.038 total
```

* x86_64-win32 on CI shows a 1.5s increase for configuring each library. for 30 libs, that's an added 45 seconds. Of course, once they can all share the same config setup, it will go back down to only needing to configure once.
CMake works a certain way, but perhaps we can add some workarounds in the meantime.

```
Waf:
    check_sdk                0.00 s

    HID:

        'configure' finished successfully (0.180s)
        'build' finished successfully (1.288s)
        'install' finished successfully (0.029s)
          hid (x86_64-win32) completed in 1.68 s

    Platform

        Building platform for x86_64-win32
        'configure' finished successfully (0.178s)
        'build' finished successfully (1.627s)
        'install' finished successfully (0.031s)
          platform (x86_64-win32) completed in 2.03 s

CMake + Ninja

    check_sdk                0.86 s

    CMake configure platform            1.70 s
    Ninja build platform build_tests    1.88 s
    Ninja run_tests platform            0.05 s

    CMake configure hid                 1.62 s
    Ninja build hid build_tests         1.55 s
    Ninja run_tests hid                 0.03 s

    CMake configure input               1.61 s
    Ninja build input build_tests       2.31 s
    Ninja run_tests input               0.06 s

```

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+# Ensure our CMake modules are available
+include("${CMAKE_CURRENT_LIST_DIR}/scripts/cmake/defold.cmake")
+
+# we rename this project in build.py when we generate the solution
+project(defold_libraries LANGUAGES C CXX)
+
+# Add subprojects
+add_subdirectory(engine/hid)
+add_subdirectory(engine/input)
```
JCash: It's small, but it's a start, you can run:
```bash
./scripts/build.py make_solution --platform=arm64-macos
```
And you'll get an XCode solution with two libraries and tests that you can build incrementally and step into the test files.

build_tools/sdk.py
```diff
@@ -775,6 +775,8 @@ def get_host_platform():
             machine = 'arm64'
         return '%s-linux' % machine
     elif sys.platform == 'win32':
+        if machine == 'arm64':
+            machine = 'x86_64' # we don't support arm64 windows targets yet
```
JCash, line 779: Fix for building for windows, via a arm64 host Vm

build_tools/waf_dynamo.py
```diff
@@ -492,6 +492,16 @@ def default_flags(self):
...
 
+    # For 32-bit Windows, search both legacy 'win32' and tuple 'x86-win32' folders
+    if build_util.get_target_platform() == 'win32':
+        alias = 'x86-win32' # used by the new/CMake code path
+        # Includes under DYNAMO_HOME and DYNAMO_HOME/ext
+        self.env.append_value('INCLUDES', build_util.get_dynamo_home('include', alias))
+        self.env.append_value('INCLUDES', build_util.get_dynamo_ext('include', alias))
+        # Lib paths under DYNAMO_HOME and DYNAMO_HOME/ext
+        self.env.append_value('LIBPATH', build_util.get_dynamo_home('lib', alias))
+        self.env.append_value('LIBPATH', build_util.get_dynamo_ext('lib', alias))
```
JCash, line 510: Our old x86 platform is called "win32", but cmake calls it `x86-win32`, so we add both during the transition period.

build_tools/waf_dynamo.py
```diff
@@ -2138,6 +2150,15 @@ def detect(conf):
...
         conf.env['LINKFLAGS_DX12']      = ['D3D12.lib', 'DXGI.lib', 'D3Dcompiler.lib']
 
+    if conf.env.PLATFORM in ['win32', 'x86_64-win32']:
+        conf.env['LINKFLAGS_HID']      = 'hid.lib'
+        conf.env['LINKFLAGS_HID_NULL'] = 'hid_null.lib'
+        conf.env['LINKFLAGS_INPUT']    = 'input.lib'
+    else:
+        conf.env['STLIB_HID']          = 'hid'
+        conf.env['STLIB_HID_NULL']     = 'hid_null'
+        conf.env['STLIB_INPUT']        = 'input'
```
JCash: We're previously relied on the library pattern `lib%s.lib` which is non standard.
The easiest way out of that during the transition period, is to use linkflags, where we can specify the actual filename.

engine/dlib/src/dmsdk/dlib/log.h
```diff
@@ -225,6 +225,8 @@ void LogInternal(LogSeverity severity, const char* domain, const char* format, .
 #define dmLogOnceFatal(format, args... ) dmLogOnceInternal(dmLogFatal, format, ## args )
 #endif
 
+#endif // NDEBUG
```
JCash, line 228: Fix for when NDEBUG is used, it excluded too many functions.

engine/hid/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+add_library(hid_null STATIC
+    "${HID_SRC_DIR}/hid.cpp"
+    "${HID_SRC_DIR}/hid_null.cpp"
+)
+target_include_directories(hid_null PUBLIC "${HID_SRC_DIR}")
+
+
+# ===================== Tests =====================
+
+if(BUILD_TESTS)
```
JCash, line 50: This is always ON, as it's about _generating_ the build scripts.
The _actual running_ is controlled by how the Ninja build is invoked (i.e. with "build_tests" and "run_tests")

engine/hid/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+
+    # Do not build test executables by default (optional via explicit target)
+    set_target_properties(test_hid PROPERTIES EXCLUDE_FROM_ALL TRUE)
+
+    target_link_libraries(test_hid PRIVATE
+        ${defold_libprefix}dlib
+        ${defold_libprefix}profile_null
+        ${defold_libprefix}platform_null
+        ${defold_libprefix}testmain
+        hid_null
```
JCash: Example of how the library names differ.
We should probably use a helper function for that too.

engine/hid/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+
+    if (TARGET_PLATFORM MATCHES "arm64-linux|x86_64-linux")
+        find_package(Threads REQUIRED)
+        target_link_libraries(test_hid PRIVATE glfw3 Threads::Threads dl)
+
+    elseif (TARGET_PLATFORM MATCHES "x86-win32|x86_64-win32")
+        target_link_libraries(test_hid PRIVATE)
+    endif()
+
+    defold_target_link_socket(test_hid "${TARGET_PLATFORM}" PRIVATE)
```
JCash, line 68: Example of helper function adding correct link libraries for the target platform.

engine/hid/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+
+    # Register into global build_tests and run_tests
+    defold_register_test_target(test_hid)
```
JCash, line 71: As we have tests that we don't want to run automatically for various reasons, this is the easy option to register a test to run.

engine/hid/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+    elseif (TARGET_PLATFORM MATCHES "arm64-macos|x86_64-macos")
+        target_link_libraries(test_app_hid PRIVATE "-framework Cocoa" "-framework IOKit" "-framework CoreVideo")
+
+    elseif (TARGET_PLATFORM MATCHES "x86_64-win32|win32")
+        target_link_libraries(test_app_hid PRIVATE user32 gdi32 shell32)
+
+    elseif (TARGET_PLATFORM MATCHES "arm64-linux|x86_64-linux")
+        find_package(Threads REQUIRED)
+        target_link_libraries(test_app_hid PRIVATE Threads::Threads dl)
+    endif()
```
JCash, line 101: In the next PR (gamesys), I've added an "defold_target_link_app" helper function to add these.

engine/input/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+    ARCHIVE DESTINATION lib/${TARGET_PLATFORM}
+    LIBRARY DESTINATION lib/${TARGET_PLATFORM}
+    RUNTIME DESTINATION bin/${TARGET_PLATFORM}
+)
+
+# Install public headers
+install(FILES "${INPUT_SRC_DIR}/input.h" DESTINATION include/input)
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/input/input_ddf.h" DESTINATION include/input)
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/python/input/__init__.py" DESTINATION lib/python/input)
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/python/input/input_ddf_pb2.py" DESTINATION lib/python/input)
```
JCash, line 59: Necessary steps for installing the protobuf output files

engine/input/src/test/CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+
+#================================ Test Data ================================
+
+defold_protoc_encode(
+  "${CMAKE_CURRENT_BINARY_DIR}/build/test.input_bindingc"           # OUT_FILE
+  "${CMAKE_CURRENT_SOURCE_DIR}/test.input_binding"                  # SRC_FILE
+  "${CMAKE_CURRENT_SOURCE_DIR}/../../proto/input/input_ddf.proto"   # SCHEMA_PROTO
+  "dmInputDDF.InputBinding"                                         # MESSAGE_NAME
+  INCLUDES "${CMAKE_CURRENT_SOURCE_DIR}/../../proto"
+)
```
JCash, line 15: The protobuf plugins are a bit finicky, but I think they turned out ok.

engine/input/src/test/CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+add_executable(test_input "${CMAKE_CURRENT_LIST_DIR}/test_input.cpp")
+
+# make sure the test data is built
+add_custom_target(input_test_data DEPENDS
+    "${CMAKE_CURRENT_BINARY_DIR}/build/test.input_bindingc"
+    "${CMAKE_CURRENT_BINARY_DIR}/build/test2.input_bindingc"
+    "${CMAKE_CURRENT_BINARY_DIR}/build/test_text.input_bindingc"
+    "${CMAKE_CURRENT_BINARY_DIR}/build/combinations.input_bindingc"
+    "${CMAKE_CURRENT_BINARY_DIR}/build/test.gamepadsc")
+add_dependencies(input input_test_data)
```
JCash, line 61: And, how to make sure these custom targets build before the C++ code.

engine/input/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+
+target_include_directories(input PUBLIC ${INPUT_SRC_DIR} ${CMAKE_CURRENT_BINARY_DIR})
+# just to make sure the extra dependency is actually built
+add_dependencies(input input_proto_py)
+
+# Add GLFW3-specific compile define when applicable
+defold_get_glfw_version(_GLFW_VERSION "${TARGET_PLATFORM}")
+if (_GLFW_VERSION EQUAL 3)
+    target_compile_definitions(input PUBLIC DM_INPUT_USE_GLFW3)
+endif()
```
JCash, line 39: Hmm, as we have a `defold_target_link_glfw`, we might have a `defold_target_compile_glfw` as well 🤔

.github/workflows/engine-nightly.yml
```diff
@@ -21,6 +21,7 @@ jobs:
       { name: 'Install Python', uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c, with: { python-version: 3.12} },
       { name: 'XCode', uses: maxim-lobanov/setup-xcode@60606e260d2fc5762a71e64e74b2174e8ea3c8bd, with: { xcode-version: '15.4' } },
       { name: 'Install Java', uses: actions/setup-java@99b8673ff64fbf99d8d325f52d9a5bdedb8483e9, with: { java-version: '21.0.5+11.0.LTS', distribution: 'temurin'} },
+      { name: 'Install CMake', uses: lukka/get-cmake@latest },
```
JCash: Making sure we use CMake 4+

CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+# we rename this project in build.py when we generate the solution
+project(defold_libraries LANGUAGES C CXX)
+
+# Ensure our CMake modules are available (after project so compilers are known)
+include("${CMAKE_CURRENT_LIST_DIR}/scripts/cmake/defold.cmake")
+
+# Add subprojects
+add_subdirectory(engine/platform)
+add_subdirectory(engine/hid)
+add_subdirectory(engine/input)
```
JCash, line 19: This is the top script, used when one wants to create a solution.

editor/src/clj/editor/app_manifest.clj
```diff
@@ -56,19 +56,15 @@
     :js-web :wasm-web :wasm_pthread-web})
 
 (def custom-lib-names
-  {:x86-win32 {"vpx" "vpx"
-               "vulkan" "vulkan-1"}
-   :x86_64-win32 {"vpx" "vpx"
-                  "vulkan" "vulkan-1"}})
+  {:x86-win32 {"vulkan" "vulkan-1"}
+   :x86_64-win32 {"vulkan" "vulkan-1"}})
```
JCash, line 61: We should try to remove these as well in the long run.

build_tools/waf_dynamo.py
```diff
@@ -2138,6 +2149,22 @@ def detect(conf):
...
+        conf.env['LINKFLAGS_PLATFORM']          = ['platform.lib']
+        conf.env['LINKFLAGS_PLATFORM_VULKAN']   = ['platform_vulkan.lib']
+        conf.env['LINKFLAGS_PLATFORM_NULL']     = ['platform_null.lib']
+    else:
+        conf.env['STLIB_HID']               = ['hid']
+        conf.env['STLIB_HID_NULL']          = ['hid_null']
+        conf.env['STLIB_INPUT']             = ['input']
+        conf.env['STLIB_PLATFORM']          = ['platform']
+        conf.env['STLIB_PLATFORM_VULKAN']   = ['platform_vulkan']
+        conf.env['STLIB_PLATFORM_NULL']     = ['platform_null']
```
JCash, line 2173: During the transition period, we still have the STLIB_PATTERN which is "lib%s.lib", so we need to make it work with our newly renamed libraries.

engine/engine/wscript
```diff
@@ -142,15 +141,10 @@ def configure(conf):
     conf.env['STLIB_DMGLFW'] = glfw_lib
     conf.env['STLIB_GUI'] = 'gui'
     conf.env['STLIB_PARTICLE'] = 'particle'
-    conf.env['STLIB_PLATFORM'] = platform_lib
-    conf.env['STLIB_PLATFORM_NULL'] = 'platform_null'
```
JCash, line 146: moved the selection/setup to waf_dynamo.py

engine/hid/CMakeLists.txt
```diff
@@ -0,0 +1,139 @@
...
+    # Do not build test executables by default (optional via explicit target)
+    set_target_properties(test_hid PROPERTIES EXCLUDE_FROM_ALL TRUE)
+
+    target_link_libraries(test_hid PRIVATE
+        ${defold_libprefix}dlib
+        ${defold_libprefix}profile_null
+        ${defold_libprefix}testmain
+        platform_null
+        hid_null
+    )
```
JCash: this has been simplified in the next PR (gamesys library support)

engine/platform/src/test/CMakeLists.txt
```diff
@@ -0,0 +1,83 @@
+cmake_minimum_required(VERSION 4.0)
+
+# Early toolchain/SDK detection (for standalone configure)
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/../../../../scripts/cmake")
```
JCash, line 4: Using a relative path, avoids using environment variables.

engine/platform/src/test/CMakeLists.txt
```diff
@@ -0,0 +1,83 @@
...
+    ${defold_libprefix}profile_null
+    ${defold_libprefix}testmain
+    platform_null)
+
+defold_target_link_socket(test_platform "${TARGET_PLATFORM}" PRIVATE)
+
+# Ensure platform libs are built before tests when available
+if(TARGET platform_null)
+    add_dependencies(test_platform platform_null)
+endif()
```
JCash, line 28: needed extra dependencies, in order to make sure they're build in order 🤔

scripts/build.py
```diff
@@ -521,8 +523,94 @@ def __init__(self,
 
         self._create_common_dirs()
 
+    def make_solution(self, target_platform=None, build_dir=None):
```
JCash, line 526: A new function to generate a solution for an IDE:

Generate a solution for your default platform:
```
./scripts/build.py make_solution
```
or specify a platform
```
./scripts/build.py make_solution --platform=<platform>
```

The path to the solution is in `./solutions/<platform>/Defold.<suffix>`

scripts/build.py
```diff
@@ -521,8 +523,94 @@ def __init__(self,
...
+                    self._log(f"Warning: Failed to rename solution: {e}. Keeping default name: {old_path}")
+                    final_path = old_path
+            else:
+                final_path = new_path
+
+        self._log(f'Solution generated: {final_path}')
+
     def __del__(self):
-        if len(self.futures) > 0:
+        if len(getattr(self, "futures", [])) > 0:
```
JCash, line 607: fix for throwing an error when we're already quitting due to an error, thus hiding the correct log output

scripts/build.py
```diff
@@ -781,6 +869,13 @@ def check_sdk(self):
         if not result:
             self.fatal("Failed sdk check")
 
+        args = ["cmake", f"-DTARGET_PLATFORM={target_platform}", "-P", join(self.defold_root, "scripts/cmake/check_install.cmake")]
+        if self.verbose:
+            args.insert(1, '-DDEFOLD_VERBOSE=ON')
+
+        output = run.command(args)
+        self._log(output)
```
JCash, line 885: running our new `scripts/cmake/tools_*.cmake` during the `check_sdk` command, before proceeding.

scripts/build.py
```diff
@@ -1388,22 +1485,73 @@ def gen_sdk_source(self):
...
+        build_test = 'build_tests' if build_tests else ''
+        install = 'install'
+
+        # generate the build script
+        cmake_configure_args = f"cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE={build_type} -DTARGET_PLATFORM={platform} -DBUILD_TESTS=ON".split()
+        run.env_command(self._form_env(), cmake_configure_args, cwd = libdir)
+
+        # execute the build
+        ninja_build_args = f"ninja all {build_test} {test} {install} {verbose}".split()
+        run.env_command(self._form_env(), ninja_build_args, cwd = builddir)
```
JCash, line 1561: * We _always generate_ the full set of commands, but group them into different build targets. I.e the tests are already there.
* The ninja invocation is the one containing the optional flags for `all` (all libraries), and `build_tests` (building the tests`) and `run_tests`
  * This mimics the way we currently use the waf flags `--skip-tests` and `--skip-build-tests`

scripts/build.py
```diff
@@ -1388,22 +1485,73 @@ def gen_sdk_source(self):
...
+
+
+    def _build_engine_lib(self, args, lib, platform, skip_tests = False, directory = 'engine'):
+        self.build_tracker.start_component(lib, platform)
+        if lib in CMAKE_SUPPORT:
+            if platform == 'win32':
+                platform = 'x86-win32'
+            self._build_engine_lib_cmake(lib, platform, directory)
+        else:
+            self._build_engine_lib_waf(args, lib, platform, skip_tests, directory)
```
JCash, line 1584: with all wscripts intace, it would be easy to skip this if statement, in case we need to temporarily revert the use of cmake.

engine/input/src/test/test_input.cpp
```diff
@@ -51,16 +51,22 @@ class InputTest : public jc_test_base_class
         char buffer[128];
         #define HOSTPATH(_PATH) dmTestUtil::MakeHostPath(buffer, sizeof(buffer), _PATH)
 
-        dmDDF::Result result = dmDDF::LoadMessageFromFile(HOSTPATH("build/src/test/test.gamepadsc"), dmInputDDF::GamepadMaps::m_DDFDescriptor, (void**)&gamepad_maps);
+        #if defined(DM_USE_CMAKE)
+            #define BUILD_DIR "build/src/test/build"
+        #else
+            #define BUILD_DIR "build/src/test"
+        #endif
```
JCash, line 58: Another thing to consider while we're transitioning to cmake.
For running tests on different devices, we have DM_HOSTFS define, so we may add a DM_TEST_BUILD_DIR to use. Or possibly change the CMake output folder.

scripts/cmake/check_install.cmake
```diff
@@ -0,0 +1,42 @@
...
+if(NOT DEFINED TARGET_PLATFORM OR NOT TARGET_PLATFORM)
+  set(TARGET_PLATFORM "${HOST_PLATFORM}")
+endif()
+message(STATUS "check_install: TARGET_PLATFORM=${TARGET_PLATFORM}")
+
+# Load tool checks (Java, Ninja, Protoc) and validate versions/paths
+include(tools)
+
+# Load SDK detection for the chosen TARGET_PLATFORM
+include(sdk)
```
JCash, line 40: by simply including these, we should get informative errors if the setup is not ok for building Defold

scripts/cmake/defold.cmake
```diff
@@ -0,0 +1,143 @@
+message(DEBUG "defold.cmake:")
```
JCash, line 6: Each CMakeLists.txt should include this file, and then proceed to setup the library + tests.

build_tools/waf_dynamo.py
```diff
@@ -1728,16 +1741,19 @@ def detect(conf):
...
 
     if host_platform in ('x86_64-linux', 'arm64-linux', 'x86_64-win32', 'x86_64-macos', 'arm64-macos'):
         conf.env['IS_HOST_DESKTOP'] = 'true'
 
-    try:
-        build_util = create_build_utility(conf.env)
-    except BuildUtilityException as ex:
-        conf.fatal(ex.msg)
+    bindirs = [build_util.get_dynamo_ext('bin', host_platform)]
+    conf.find_program('glslang', var='GLSLANG', mandatory = True, path_list = bindirs)
```
JCash, line 1760: for some reasno, in my VM, it just wouldn't find "glslang", so it was better to use the waf build system to located it upfront.

engine/hid/CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+        profile_null
+        hid
+    )
+
+    defold_target_link_graphics(test_app_hid "${TARGET_PLATFORM}" PRIVATE)
+    defold_target_link_platform(test_app_hid "${TARGET_PLATFORM}" PRIVATE)
+    defold_target_link_socket(test_app_hid "${TARGET_PLATFORM}" PRIVATE)
+
+    # Register into global build_tests and run_tests
+    defold_register_test_target(test_app_hid FALSE)
```
JCash, line 114: as with our current build system, where we have tags "test" and "skip_test", we needed a way to do that, and the solution is to register the tests to a global list. `FALSE` here means that it won't be automatically run with the other tests (as it's not an actual unit test, but an app)

engine/input/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+add_custom_target(input_proto_py DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/python/input/input_ddf_pb2.py")
+
+defold_protoc_gen_cpp("${CMAKE_CURRENT_BINARY_DIR}/input/input_ddf.pb.cc"
+                          "${CMAKE_CURRENT_SOURCE_DIR}/proto/input/input_ddf.proto"
+                          INCLUDES "${CMAKE_CURRENT_SOURCE_DIR}/proto" "${DEFOLD_SDK_ROOT}/share/proto")
+
+add_library(input STATIC
+    "${INPUT_SRC_DIR}/input.cpp"
+    "${CMAKE_CURRENT_BINARY_DIR}/input/input_ddf.pb.cc"
+)
```
JCash, line 28: example of having separate functions for generating protoc bindings for Python and C++, and hooking in the .cc generated file into the library.

engine/input/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+target_compile_definitions(input PUBLIC "DLIB_LOG_DOMAIN=\"INPUT\"")
+
+target_include_directories(input PUBLIC ${INPUT_SRC_DIR} ${CMAKE_CURRENT_BINARY_DIR})
+# just to make sure the extra dependency is actually built
+add_dependencies(input input_proto_py)
```
JCash, line 33: Make sure to add the dependency, in order to make it build!

engine/input/src/test/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+
+defold_target_link_libraries(test_input "${TARGET_PLATFORM}"
+    dlib
+    mbedtls
+    ddf
+    profile_null
+    testmain
+    hid_null
+    input
+)
```
JCash, line 77: we have our own version of "target_link_libraries", as we need to prefix/not prefix some of our libraries during the transition period.

engine/input/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
+# Minimum CMake version and project
+cmake_minimum_required(VERSION 4.0)
+
+# Early toolchain/SDK detection (for standalone configure)
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/../../scripts/cmake")
+include(defold)
+
+project(defold_input LANGUAGES C CXX)
```
JCash: All sdk setup _must_ be done before the `project()` and seting languages is done.

scripts/build.py
```diff
@@ -1388,22 +1479,73 @@ def gen_sdk_source(self):
...
+    def _find_cmake_build_type(self, options):
+        opt_level = 2
+        for x in options:
+            if '--opt-level=' in x:
+                opt_level = x.replace('--opt-level=', '')
+                opt_level = int(opt_level)
+                break
+        if opt_level < 2:
+            return 'Debug'
+        return 'RelWithDebInfo'
```
JCash, line 1521: How we translate the opt level currently to "Debug" or "RelWithDebInfo" configurations.

scripts/build.py
```diff
@@ -1388,22 +1479,73 @@ def gen_sdk_source(self):
...
+        env = self._form_env()
+
+        is_verbose = ('-v' in self.waf_options) or ('--verbose' in self.waf_options)
+        verbose = '-v' if is_verbose else ''
+        test = '' if (self.skip_tests or not supports_tests) else 'run_tests'
+        build_test = 'build_tests' if build_tests else ''
+        install = 'install'
+
+        # generate the build script
+        cmake_configure_args = f"cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE={build_type} -DTARGET_PLATFORM={platform} -DBUILD_TESTS=ON".split()
```
JCash, line 1550: We need to translate more options here, like the desired opt level flag, or "with-asan" etc.

scripts/cmake/defold.cmake
```diff
@@ -0,0 +1,157 @@
...
+    if(DEFINED ${_cache_var})
+      set(_flags "${${_cache_var}}")
+    elseif(DEFINED ${_init_var})
+      set(_flags "${${_init_var}}")
+    endif()
+
+    set(${_cache_var} "${_flags}" CACHE STRING "${_cache_var}" FORCE)
+    set(${_init_var} "${_flags}")
+  endforeach()
+endforeach()
```
JCash, line 58: This is a weird part. The `-NDEBUG` (which we don't want, as it disables the asserts, which we want), cannot be removed before the compiler has been chosen.
And that step is done in the call to `project()`in the CMakeLists.txt.
However, it _also_ needs this preamble in order to do that 🤷

scripts/cmake/defold.cmake
```diff
@@ -0,0 +1,157 @@
...
+# Ensure the INTERFACE target exists early so platform modules can attach options
+if(NOT TARGET defold_sdk)
+  add_library(defold_sdk INTERFACE)
+endif()
+
+# verify our list of tools (e.g. java, ninja etc)
+include(tools)
+
+# list of toggleable features
+include(features)
```
JCash, line 99: I haven't really consolidated options yet, but I think it may be good to collect them into a single, easily discoverable file.

scripts/cmake/defold.cmake
```diff
@@ -0,0 +1,157 @@
...
+# Core include dirs (non-system)
+target_include_directories(defold_sdk INTERFACE
+  "${DEFOLD_INCLUDE_DIR}"
+  "${DEFOLD_DMSDK_INCLUDE_DIR}")
+# External/platform include dirs as SYSTEM to reduce warnings
+target_include_directories(defold_sdk SYSTEM INTERFACE
+  "${DEFOLD_EXT_INCLUDE_DIR}"
+  ${_DEFOLD_PLATFORM_INCLUDE_DIRS})
+# Library search directories
+target_link_directories(defold_sdk INTERFACE ${_DEFOLD_PLATFORM_LIB_DIRS})
```
JCash, line 142: The defold sdk is created as an INTERFACE.

scripts/cmake/defold.cmake
```diff
@@ -0,0 +1,157 @@
...
+  set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELWITHDEBINFO ON)
+  # Enable IPO on the defold_sdk usage target so consumers default to IPO
+  # (actual effect applies to real targets compiled/linked)
+  set_property(TARGET defold_sdk PROPERTY INTERPROCEDURAL_OPTIMIZATION TRUE)
+else()
+  defold_log("IPO not enabled: ${_DEFOLD_IPO_MSG}")
+endif()
+
+# Apply defold_sdk to all targets in this directory and below
+link_libraries(defold_sdk)
```
JCash, line 159: I wonder if the helper functions for e.g. glfw etc, could/should be rewritten as INTERFACE instead 🤔

scripts/cmake/exported_symbols.in.cpp
```diff
@@ -0,0 +1,4 @@
+@DM_EXPORTED_SYMBOL_DECLS@
+extern "C" void dmExportedSymbols() {
+@DM_EXPORTED_SYMBOL_CALLS@
+}
```
JCash, line 4: Template file for generating our exported symbols .cpp

scripts/cmake/functions_platform.cmake
```diff
@@ -0,0 +1,92 @@
...
+#
+# Selection rules:
+# - Web targets -> platform
+# - GLFW3 desktop targets (macOS, Linux, Win32, Nx64):
+#     - If WITH_VULKAN=ON OR platform in {arm64-macos,x86_64-macos,arm64-nx64}
+#       -> platform_vulkan
+#     - Else -> platform
+# - Other targets -> platform
+
+function(defold_target_link_platform target platform)
```
JCash, line 18: I cannot help but feel the code should be shorter and more concise, but my focus had been on the functionality.
But maintainability is of course a concern too.

scripts/cmake/functions_protoc.cmake
```diff
@@ -0,0 +1,281 @@
...
+    foreach(_inc IN LISTS _inc_dirs)
+        list(APPEND _inc_flags -I "${_inc}")
+    endforeach()
+
+    file(MAKE_DIRECTORY "${_out_dir}")
+
+    # Resolve DDF plugin path
+    if(NOT DEFOLD_SDK_ROOT)
+        message(FATAL_ERROR "defold_protoc_gen_cpp: DEFOLD_SDK_ROOT must be set to locate ddfc_cxx plugin")
+    endif()
```
JCash, line 61: I'm not a fan of these overly cautious extra checks.
* At this point, the DEFOLD_SDK_ROOT should already be set, or we should have failed a lot earlier
* Readability suffers

scripts/cmake/platform_android.cmake
```diff
@@ -0,0 +1,50 @@
+defold_log("platform_android.cmake:")
+
+# Try to resolve NDK root and sysroot
+set(_DEFOLD_ANDROID_NDK "${ANDROID_NDK}")
+if(NOT _DEFOLD_ANDROID_NDK AND DEFINED CMAKE_ANDROID_NDK)
+    set(_DEFOLD_ANDROID_NDK "${CMAKE_ANDROID_NDK}")
+endif()
```
JCash, line 7: At this point, `sdk_android.cmake` should already have been run.

scripts/cmake/sdk_android.cmake
```diff
@@ -0,0 +1,134 @@
...
+    set(ANDROID_ABI "armeabi-v7a" CACHE STRING "Android ABI" FORCE)
+    set(ANDROID_NATIVE_API_LEVEL 19 CACHE STRING "Android API Level" FORCE)
+    # Use the canonical Clang triplet for 32-bit ARM
+    set(ANDROID_TOOLCHAIN "armv7a-linux-androideabi${ANDROID_NATIVE_API_LEVEL}-clang" CACHE STRING "Android Toolchain" FORCE)
+
+endif()
+
+defold_log("ANDROID_ABI: ${ANDROID_ABI}")
+defold_log("ANDROID_NATIVE_API_LEVEL: ${ANDROID_NATIVE_API_LEVEL}")
+defold_log("ANDROID_TOOLCHAIN: ${ANDROID_TOOLCHAIN}")
```
JCash, line 130: As before, when running `./scripts/build.py check_sdk --platform=arm64-android --verbose` you'll get an output like this:
```bash
-- sdk.cmake:
-- sdk_android.cmake:
-- sdk_android: Using packaged Android NDK: /Users/mathiaswesterdahl/work/defold/tmp/dynamo_home/ext/SDKs/android-ndk-r25b
-- sdk_android: Toolchain file: /Users/mathiaswesterdahl/work/defold/tmp/dynamo_home/ext/SDKs/android-ndk-r25b/build/cmake/android.toolchain.cmake
-- ANDROID_ABI: arm64-v8a
-- ANDROID_NATIVE_API_LEVEL: 21
-- ANDROID_TOOLCHAIN: aarch64-linux-android21-clang
-- check_install: Environment looks OK for arm64-android
```

scripts/cmake/sdk_windows.cmake
```diff
@@ -0,0 +1,840 @@
+defold_log("sdk_windows.cmake:")
```
JCash, line 1: This one is very large, if feels excessive to be honest, and should probably be rewritten to make it more readable.

scripts/cmake/tools_protoc.cmake
```diff
@@ -0,0 +1,70 @@
+defold_log("tools_protoc.cmake:")
+
+# Ensure that protoc is the one shipped with the Defold SDK (tmp/dynamo_home)
+# Layout used by Defold packages:
+#   ${DEFOLD_SDK_ROOT}/ext/bin/<host-platform>/protoc
+#   ${DEFOLD_SDK_ROOT}/bin/ddfc_cxx (protoc plugin)
```
JCash, line 6: An example of our troubles in the past, users accidentally using their locally installed protoc.

scripts/cmake/sdk.cmake
```diff
@@ -0,0 +1,45 @@
...
+    elseif(EXISTS "${DEFOLD_HOME}/tmp/dynamo_home")
+        set(DEFOLD_SDK_ROOT "${DEFOLD_HOME}/tmp/dynamo_home" CACHE PATH "Path to Defold SDK (tmp/dynamo_home)" FORCE)
+    endif()
+endif()
+
+if (TARGET_PLATFORM MATCHES "arm64-macos|x86_64-macos|arm64-ios|x86_64-ios")
+    # NOTE: Minimum iOS-version is also specified in Info.plist-files
+    # (MinimumOSVersion and perhaps DTPlatformVersion)
+    defold_set_from_sdk_py(SOURCE VERSION_IPHONEOS_MIN TARGET SDK_VERSION_IPHONEOS_MIN)
+    defold_set_from_sdk_py(SOURCE VERSION_MACOSX_MIN  TARGET SDK_VERSION_MACOSX_MIN)
```
JCash, line 29: Reading our version numbers from sdk.py

scripts/cmake/sdk_android.cmake
```diff
@@ -0,0 +1,134 @@
...
+        get_filename_component(_ndk_grandparent "${_ndk_parent}" DIRECTORY)
+        set(_ANDROID_NDK_ROOT "${_ndk_grandparent}")
+    endif()
+
+    if(_SDKS_DIR AND _ANDROID_TOOLCHAIN_FILE MATCHES "^${_SDKS_DIR}.*")
+    defold_log("sdk_android: Using packaged Android NDK: ${_ANDROID_NDK_ROOT}")
+    else()
+    defold_log("sdk_android: Using local Android SDK NDK: ${_ANDROID_NDK_ROOT}")
+    endif()
+    defold_log("sdk_android: Toolchain file: ${_ANDROID_TOOLCHAIN_FILE}")
```
JCash, line 102: As long as we choose correct API level (below), I don't think it should matter if the local install gets a newer build system (than r25b that we prepackage). E.g. I'd like to be able to use the locally installed Android versions on CI as well. Having prepackaged versions is a nuisance so if we can avoid it...

vlaaad (dismissed):
Editor changes look good! It seems we can now move many lib names in app manifests from platform-specific configurations to common. Perhaps not in this PR, but we could make a task for that?

deurell (dismissed):
👍 Looks like a solid first step migration run. Small things like using find_package instead of direct linkage could easily be added in the future if needed. The only thing that could be a thing would be the caching and using FORCE to override user provided values that could be confusing for a user providing custom values. But could be fixed later if it's an issue. IDE's like CLion could get confused if multiple configure runs would fail based on cache deps like in the PR I made.

.github/workflows/engine-nightly.yml
ekharkunov: Maybe it's worth to:
1. Pin specific commit of the action to avoid any breaking changes in workflow
2. Pin specific version of Cmake (https://github.com/lukka/get-cmake?tab=readme-ov-file#if-you-want-to-pin-the-workflow-to-specific-range-of-versions-of-cmake-and-ninja)

build_tools/waf_dynamo.py
ekharkunov: I think it's worth to left "TODO" note in the code to mark which part of the code should be removed after transition period

engine/hid/CMakeLists.txt
```diff
@@ -0,0 +1,135 @@
+# Minimum CMake version and project
+cmake_minimum_required(VERSION 4.0)
+
+# Early toolchain/SDK detection (for standalone configure)
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/../../scripts/cmake")
+include(defold)
+
+project(defold_hid LANGUAGES C CXX)
```
ekharkunov: Q: Should we add `OBJCXX` because below we enumerate `*.mm` files?

engine/platform/CMakeLists.txt
```diff
@@ -0,0 +1,85 @@
...
+else()
+    defold_get_glfw_version(_GLFW_VERSION "${TARGET_PLATFORM}")
+    if (_GLFW_VERSION EQUAL 3)
+        set(PLATFORM_USE_GLFW3 ON)
+        set(_platform_glfw_sources "${PLAT_SRC_DIR}/platform_window_glfw3.cpp")
+        if (TARGET_PLATFORM MATCHES "arm64-macos|x86_64-macos")
+            list(APPEND _platform_glfw_sources "${PLAT_SRC_DIR}/platform_window_glfw3_osx.cpp")
+        elseif (TARGET_PLATFORM MATCHES "arm64-linux|x86_64-linux")
+            list(APPEND _platform_glfw_sources "${PLAT_SRC_DIR}/platform_window_glfw3_linux.cpp")
+        elseif (TARGET_PLATFORM MATCHES "arm64-win32|x86_64-win32|x86-win32")
```
ekharkunov, line 27: Q: We don't officially support `arm64-win32`, do we?

ekharkunov (dismissed):
Great job!

engine/platform/CMakeLists.txt
JCash: no. it's the chatbot that was a bit too helpful...

vlaaad (dismissed):
LGTM!

PR #11230, opened 2025-09-15
Added `sort_order` option to `render.draw()` in `render_script`

### LINKED ISSUE #3625: Control sort order of rendered primitives (DEF-2530)
Description:
In order to avoid overdraw when rendering things it is important to be able to use the correct sort order for each given scenario. Right now everything is sorted back-to-front, which is great for alpha blended sprites. But when doing opaque 3d you usually want to sort front-to-back to efficiently use the z-buffer. There currently is no way to do this in Defold at this point.

I would suggest adding something similar to this api call:

```
render.set_sort_order(render.SORT_FRONT_TO_BACK)
```

where possible options are:

* render.SORT_BACK_TO_FRONT (default),
* render.SORT_FRONT_TO_BACK
* render.SORT_NONE (for when sorting isn't needed and the extra cycles are better spent someplace else)

AGulev:
You can now choose from the following sort orders when using `render.draw()`:  
- `render.SORT_BACK_TO_FRONT` (default)  
- `render.SORT_FRONT_TO_BACK`  
- `render.SORT_NONE`

Fix https://github.com/defold/defold/issues/3625

### DIFF
--- a/engine/render/src/render/render.cpp
+++ b/engine/render/src/render/render.cpp
@@ -928,13 +928,15 @@ namespace dmRender
             }
         }
 
-        MakeSortBuffer(context, predicate?predicate->m_TagCount:0, predicate?predicate->m_Tags:0, sort_order);
+        SortOrder effective_sort_order = (sort_order == SORT_UNSPECIFIED) ? SORT_BACK_TO_FRONT : sort_order;
+
+        MakeSortBuffer(context, predicate?predicate->m_TagCount:0, predicate?predicate->m_Tags:0, effective_sort_order);
 
         if (context->m_RenderListSortBuffer.Empty())
             return RESULT_OK;
 
         {
-            if (sort_order != SORT_NONE)
+            if (effective_sort_order != SORT_NONE)
             {
                 DM_PROFILE("DrawRenderList_SORT");
                 RenderListSorter sort;
--- a/engine/render/src/render/render.cpp
+++ b/engine/render/src/render/render.cpp
@@ -587,6 +587,13 @@ namespace dmRender
         if (maxZW > minZW)
             rc = 1.0f / (maxZW - minZW);
 
+        const uint32_t ORDER_SCALE_MASK = 0xfffff0;
+        const uint32_t ORDER_BASE_FRONT = 0x000008;
+        const uint32_t ORDER_BASE_BACK = 0xfffff8;
+        const bool front_to_back = (sort_order == SORT_FRONT_TO_BACK);
+        const float order_multiplier = (front_to_back ? 1.0f : -1.0f) * ORDER_SCALE_MASK * rc;
+        const float order_bias = (front_to_back ? ORDER_BASE_FRONT : ORDER_BASE_BACK) - order_multiplier * minZW;
+
         for( uint32_t i = 0; i < num_ranges; ++i)
         {
             const RenderListRange& range = ranges[i];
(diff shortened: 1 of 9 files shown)


engine/render/src/render/render_script.cpp
```diff
@@ -1739,6 +1739,9 @@ namespace dmRender
      * `constants`
      * : [type:constant_buffer] optional constants to use while rendering
      *
+     * `sort_order`
+     * : [type:int] How to sort draw order for world-ordered entries. Default is render.SORT_BACK_TO_FRONT.
+     *
```
AGulev, line 1744: If I understand idea right, we want to specify sorting order as `render.draw()` function parameter: 
```lua
    draw_options_world.sort_order = render.SORT_FRONT_TO_BACK
    render.draw(predicates.model, draw_options_world)
```

aglitchman:
Many thanks 🙏

engine/render/src/render/render.cpp
```diff
@@ -607,7 +607,16 @@ namespace dmRender
...
+                    if (sort_order == SORT_FRONT_TO_BACK)
+                    {
+                        // Near-to-far
+                        sort_values[idx].m_Order = (uint32_t) (0x000008 + 0xfffff0 * rc * (z - minZW));
+                    }
+                    else
+                    {
+                        // Default: far-to-near
+                        sort_values[idx].m_Order = (uint32_t) (0xfffff8 - 0xfffff0 * rc * (z - minZW));
+                    }
```
JCash: Can we please make this branchless?
I.e. move out the Mask to outside of the loops instead.

engine/render/src/render/render_script.cpp
JCash: That was the idea as a first step, yes.
I'm not sure how other engines do it (e.g. on a per material basis), but I think this works as an "override" for now.

engine/render/src/render/render_script.cpp
```diff
@@ -1792,6 +1795,7 @@ namespace dmRender
         dmVMath::Matrix4* frustum_matrix = 0;
         dmRender::FrustumPlanes frustum_num_planes = dmRender::FRUSTUM_PLANES_SIDES;
         HNamedConstantBuffer constant_buffer = 0;
+        dmRender::SortOrder sort_order = dmRender::SORT_BACK_TO_FRONT;
```
JCash: Is there a point in having a dmRender::SORT_UNSPECIFIED ?
Then the renderer can use its default (back-to-front), or any other mechanic it chooses to (e.g. look at materials or what not)

I'm happy as it is, as I don't think we've painted ourselves into a corner, and can make such changes later on.
It'd be good for us to figure out the available options in other engines to control such a behavior.

JCash (dismissed):
lgtm

engine/render/src/render/render.cpp
AGulev: fixed! Thank you!

engine/render/src/render/render_script.cpp
AGulev: fully agree! Fixed!

engine/render/src/render/render.cpp
```diff
@@ -607,7 +614,7 @@ namespace dmRender
                 if (entry->m_MajorOrder == RENDER_ORDER_WORLD)
                 {
                     const float z = sort_values[idx].m_ZW;
-                    sort_values[idx].m_Order = (uint32_t) (0xfffff8 - 0xfffff0 * rc * (z - minZW));
+                    sort_values[idx].m_Order = (uint32_t) (order_bias + order_multiplier * z);
```
JCash, line 617: a single muladd is nice!

JCash (approved):

Issue #8989, opened 2024-05-27
Sometimes the build-time linting takes too much time


vlaaad:
**Describe the bug (REQUIRED)**
Sometimes, the development iteration process is slowed down by the project code linting. We should fix it.

**To Reproduce (REQUIRED)**
Would be nice to have these...

**Expected behavior (REQUIRED)**
Linting does not take too much time.

**Workaround (OPTIONAL):**
We could implement disabling the workspace linter, see #8974

jalhund:
Yes, linting definitely should be switchable. It's too annoying when you have to wait even if only one line of code was changed

Anisin:
We have a fairly large project, about 2400 lua files.
When we moved to 1.9.0 and I started the Linting Code stage (Ctrl+B) it took about 20-60 seconds to start the build.
When I first launched the project, the total time to Linting took 40-50 sec.
In other conditions Linting takes 10-20 sec. but sometimes it goes up to +/- 40 sec. again.
If I disable Linting Code functionality, the build starts instantly.

AGulev:
I turn it off by default for now https://github.com/defold/defold/pull/13283

PR #11234, opened 2025-09-16
Fixed code editor freeze when matching against very long strings

### LINKED ISSUE #11191: Language server hang when editing text
It's a 100% repro

Steps to reproduce:
* Open extension-rive
* Open loader.script
* At the top, add

    local myVar = "foo"

* Press save

* Try to rename variable to

    local myVarLongerName

* Observe a hang where the cursor going into "waiting" mode and never (?) recovers

I don't get any console info while it's happening, but when quitting the process in the terminal, I get this:

```
^C/Users/mathiaswesterdahl/work/defold/editor/tmp/unpack/arm64-macos/bin/lsp/lua/bin/lua-language-server: script/service/net.lua:270: interrupted!
stack traceback:
    [C]: in method 'wait'
    script/service/net.lua:270: in function 'service.net.update'
    script/service/service.lua:185: in local 'sleep'
    script/service/service.lua:197: in function 'service.eventLoop'
    script/service/service.lua:283: in function 'service.start'
    ...efold/editor/tmp/unpack/arm64-macos/bin/lsp/lua/main.lua:80: in main chunk
    (bootstrap.lua):85: in main chunk
    [C]: in ?
```

matgis:
* Rewrote the fuzzy text matching algorithm to use less memory and perform matches faster.
* Fixed a freeze that could occur in the code editor when attempting to perform a fuzzy text match against very long autocomplete suggestions originating from non-Lua files in some projects.
* Fixed an exception that would occur if a fuzzy text search term included the Turkish capital letter I (`İ`) on a machine running under a non-Turkish locale setting.

Fixes #11191.

### Technical changes
* Disabled workspace word autocomplete suggestions from the built-in Lua LSP server.

### DIFF
--- a/editor/src/clj/editor/fuzzy_text.clj
+++ b/editor/src/clj/editor/fuzzy_text.clj
@@ -64,6 +64,8 @@
 (set! *warn-on-reflection* true)
 (set! *unchecked-math* :warn-on-boxed)
 
+(def ^:private ^:const ^{:tag 'int} space-code-point (int \space))
+
 (defn- set-matching-index-bits-in-bit-set!
   "Sets bits in the supplied bit-set to true at each index where the specified
   code point is found in the string."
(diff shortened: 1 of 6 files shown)


editor/src/clj/editor/lsp/server.clj
```diff
@@ -263,6 +263,7 @@
                   workspace (g/node-value project :workspace evaluation-context)
                   root (g/raw-property-value (:basis evaluation-context) workspace :root)]
               {:runtime {:version "Lua 5.1" :pathStrict true}
+               :completion {:workspaceWord false}
```
matgis, line 266: When `true` or left at the default setting, the Lua language server will scan non-Lua files in the project for strings and suggest them as completions when there's no better alternative. This is probably not a great idea when the project directory can contain lots of large non-code files, but I'm open to leaving it on.

The alternative would be to explicitly list a bunch of file extensions that we know to cause trouble as exclusions, such as `.spinejson`, `.gltf`, `.dylib`, `.a`, `.so`, and so on.

editor/src/clj/editor/fuzzy_text.clj
```diff
@@ -15,7 +15,8 @@
 (ns editor.fuzzy-text
   (:require [clojure.string :as string]
             [util.bit-set :as bit-set]
-            [util.coll :as coll :refer [pair]]))
+            [util.coll :as coll :refer [pair]])
+  (:import [java.util.concurrent.atomic AtomicInteger]))
```
matgis, line 19: The fuzzy text matching algorithm has been rewritten in an imperative fashion so it will perform less allocations and be more efficient.

The scoring is still only performed once we have a fully formed candidate and uses the same code as before. As a future improvement, we could sum up the score as we build up the candidate so we can bail out earlier. However, re-implementing the algorithm so it would work exactly like before was already quite tricky, so I'll leave it like this for now.

editor/src/clj/editor/fuzzy_text.clj
```diff
@@ -242,31 +171,155 @@
...
+                  (recur (inc iteration)
+                         pending-tick-index-in-pattern))))))))))
 
 (defn prepare-pattern
   "Given a pattern string, returns a trimmed and case-agnostic prepared-pattern
   for use with the match functions."
   [^String pattern]
-  (let [trimmed-pattern (string/trim pattern)]
-    (pair (string/upper-case trimmed-pattern)
-          (string/lower-case trimmed-pattern))))
```
matgis, line 263: Fun fact: If you run `string/lower-case` on a string that contains the Turkish capital letter İ like `"İpek"` under a non-Turkish locale, you end up with a string that prints like `"ipek"`, but includes a hidden code point remnant and has a length of five.

Converting the code points individually solves the issue (it's not a surrugate-pair situation either - `İ` is a single code point in the input string, and its lower-case equivalent is also a single code point if converted under the Turkish locale).

editor/src/clj/editor/lsp/server.clj
vlaaad: Ohhhh, great find!

vlaaad (approved):
Wonderful work!

editor/src/clj/editor/lsp/server.clj
AGulev: hmm, maybe it's the reason of https://github.com/defold/defold/issues/8989

PR #11239, opened 2025-09-17
Added `start_time` and `start_frame` parameters for `sound.play()`

### LINKED ISSUE #9291: Sound: Add "start_time" parameter to sound.play()
This relates to issue https://github.com/defold/defold/issues/7835, where it was clear that the "set" part of the request is easier to implement than "get".

Use case:
We have three intensity variations of a music track. At certain points in the track, we check if we want to switch to another intensity level. The variations are not mapped 1:1 in terms of duration and position, so for example if we're at 00:12 of one track we might want to switch to 00:23 for the other.

It should be possible to do a workaround for "get" side in Lua (e.g. tracking the playtime using socket.gettime() or dt), but the "set" side needs an engine implementation.

AGulev:
You can now specify `start_time` or `start_frame` (audio frame) when calling `sound.play()`.

Fix https://github.com/defold/defold/issues/9291

### DIFF
--- a/engine/sound/src/sound.cpp
+++ b/engine/sound/src/sound.cpp
@@ -1094,6 +1094,70 @@ namespace dmSound
         return RESULT_OK;
     }
 
+    static inline uint32_t GetSkipStrideBytes(const dmSoundCodec::Info& info)
+    {
+        // Match stride logic used in mixing when calling Decode/Skip
+        // If decoder delivers float non-interleaved (or mono), use sizeof(float) per frame
+        // Otherwise use interleaved stride: channels * (bits/8)
+        if (info.m_BitsPerSample == 32 && (!info.m_IsInterleaved || info.m_Channels == 1))
+            return (uint32_t)sizeof(float);
+        return (uint32_t)(info.m_Channels * (info.m_BitsPerSample / 8));
+    }
+
+    static Result SkipToStartFrameNoLock(HSoundInstance sound_instance, uint64_t start_frame)
+    {
+        DM_PROFILE(__FUNCTION__);
+        dmSoundCodec::Info info;
+        dmSoundCodec::GetInfo(g_SoundSystem->m_CodecContext, sound_instance->m_Decoder, &info);
+
+        uint64_t total_bytes = start_frame * (uint64_t)GetSkipStrideBytes(info);
+        while (total_bytes > 0)
+        {
+            uint32_t chunk = (uint32_t)dmMath::Min<uint64_t>(total_bytes, 1ULL << 30);
+            uint32_t skipped = 0;
+            dmSoundCodec::Result r = dmSoundCodec::Skip(g_SoundSystem->m_CodecContext, sound_instance->m_Decoder, chunk, &skipped);
+            if (r == dmSoundCodec::RESULT_END_OF_STREAM)
+            {
+                break;
+            }
+            if (r != dmSoundCodec::RESULT_OK)
+            {
+                return RESULT_INVALID_STREAM_DATA;
+            }
+            if (skipped == 0)
+            {
+                break;
+            }
+            total_bytes -= skipped;
+        }
+        return RESULT_OK;
+    }
+
+    Result SetStartFrame(HSoundInstance sound_instance, uint32_t start_frame)
+    {
+        if (!g_SoundSystem)
+            return RESULT_OK;
+        DM_MUTEX_OPTIONAL_SCOPED_LOCK(g_SoundSystem->m_Mutex);
+        return SkipToStartFrameNoLock(sound_instance, (uint64_t)start_frame);
+    }
+
+    Result SetStartTime(HSoundInstance sound_instance, float start_time_seconds)
+    {
+        if (!g_SoundSystem)
+            return RESULT_OK;
+        if (start_time_seconds <= 0.0f)
+            return RESULT_OK;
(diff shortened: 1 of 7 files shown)


engine/sound/src/sound.cpp
```diff
@@ -1094,6 +1094,70 @@ namespace dmSound
...
+        DM_PROFILE(__FUNCTION__);
+        dmSoundCodec::Info info;
+        dmSoundCodec::GetInfo(g_SoundSystem->m_CodecContext, sound_instance->m_Decoder, &info);
+
+        uint64_t total_bytes = start_frame * (uint64_t)GetSkipStrideBytes(info);
+        while (total_bytes > 0)
+        {
+            uint32_t chunk = (uint32_t)dmMath::Min<uint64_t>(total_bytes, 1ULL << 30);
+            uint32_t skipped = 0;
+            dmSoundCodec::Result r = dmSoundCodec::Skip(g_SoundSystem->m_CodecContext, sound_instance->m_Decoder, chunk, &skipped);
```
AGulev: I'm not happy with that. It seems like we need `frame index -> page offset` search index (depends on a codec, needed for vorbis and opus) we prepare build-time for each sound and then use it in runtime.

engine/sound/src/test/test_sound.cpp
```diff
@@ -2351,3 +2351,164 @@ int main(int argc, char **argv)
...
+    dmSound::HSoundData sd = 0;
+    dmSound::NewSoundData(params.m_Sound, params.m_SoundSize, params.m_Type, &sd, 1234);
+
+    dmSound::HSoundInstance instance = 0;
+    r = dmSound::NewSoundInstance(sd, &instance);
+    ASSERT_EQ(dmSound::RESULT_OK, r);
+    ASSERT_NE((dmSound::HSoundInstance) 0, instance);
+
+    // Start at 1.585 seconds
+    r = dmSound::SetStartTime(instance, 1.618f);
```
AGulev, line 2488: we have sound where a guy say 1 2 3 4 5 6 7, this test keep only 7

engine/sound/src/sound.cpp
```diff
@@ -1094,6 +1094,70 @@ namespace dmSound
...
+        if (!g_SoundSystem)
+            return RESULT_OK;
+        if (start_time_seconds <= 0.0f)
+            return RESULT_OK;
+        DM_MUTEX_OPTIONAL_SCOPED_LOCK(g_SoundSystem->m_Mutex);
+
+        dmSoundCodec::Info info;
+        dmSoundCodec::GetInfo(g_SoundSystem->m_CodecContext, sound_instance->m_Decoder, &info);
+        double frames_d = (double)start_time_seconds * (double)info.m_Rate;
+        if (frames_d <= 0.0)
```
britzl: Can frames_d really be negative?

engine/sound/src/sound.cpp
```diff
@@ -1094,6 +1094,70 @@ namespace dmSound
...
+            return RESULT_OK;
+        uint64_t start_frame = (uint64_t)dmMath::Max(0.0, frames_d);
```
britzl: And if frames_d can't be negative why do a dmMath::Max with 0.0? start_frame is (uint64_t)frames_d always?

engine/sound/src/sound.cpp
```diff
@@ -1094,6 +1094,70 @@ namespace dmSound
...
+    static Result SkipToStartFrameNoLock(HSoundInstance sound_instance, uint64_t start_frame)
+    {
+        DM_PROFILE(__FUNCTION__);
+        dmSoundCodec::Info info;
+        dmSoundCodec::GetInfo(g_SoundSystem->m_CodecContext, sound_instance->m_Decoder, &info);
+
+        uint64_t total_bytes = start_frame * (uint64_t)GetSkipStrideBytes(info);
+        while (total_bytes > 0)
+        {
+            uint32_t chunk = (uint32_t)dmMath::Min<uint64_t>(total_bytes, 1ULL << 30);
```
britzl: What's going on here? What's 1ULL << 30? Why shift by 30?

engine/sound/src/sound.cpp
```diff
@@ -1094,6 +1094,70 @@ namespace dmSound
...
+    {
+        // Match stride logic used in mixing when calling Decode/Skip
+        // If decoder delivers float non-interleaved (or mono), use sizeof(float) per frame
+        // Otherwise use interleaved stride: channels * (bits/8)
+        if (info.m_BitsPerSample == 32 && (!info.m_IsInterleaved || info.m_Channels == 1))
+            return (uint32_t)sizeof(float);
+        return (uint32_t)(info.m_Channels * (info.m_BitsPerSample / 8));
+    }
+
+    static Result SkipToStartFrameNoLock(HSoundInstance sound_instance, uint64_t start_frame)
```
britzl, line 1107: What do we mean by calling it NoLock?

britzl (changes requested):
See code comments

engine/sound/src/sound.cpp
AGulev: That means caller should call DM_MUTEX_OPTIONAL_SCOPED_LOCK on needed scope himself (we do it in a couple of other places).

engine/sound/src/sound.cpp
AGulev: I'll remove extra defensive code

JCash (approved):

britzl (approved):

Issue #5453, opened 2021-01-06
Externalize editor localization text


subsoap:
**Is your feature request related to a problem? Please describe (REQUIRED):**
Currently interface text is baked into code. It makes adding translations of the editor to not be possible.

**Describe the solution you'd like (REQUIRED):**
Add a key system for grabbing localization strings based on locale. Default to English, have multiple places to toggle the language such as on the launcher / settings. Either allow hotswap on text / reload UI / or force editor restart to update language.

Should also support extra languages / language overrides within the .defold user folder. This is useful for people donating translations to easily see updates. The "Update CSS" button could be changed to mean everything in the .defold user folder.

All files related to language should be under a locale folder such as en-us with a folder above it have for example a language list file along with their short and long names along with which should be default.

**Describe alternatives you've considered (REQUIRED):**
People could fork Defold and change the English text to other languages. This should be avoided, it should be done in a clean and centralized way.

**Additional context (OPTIONAL):**
Some languages such as RTL would require extra care.

This should have more discussion about implementation / design before anyone starts on it. There needs to be proposals on exact implementation people can agree on since it needs to be as good as possible of a process for translators to contribute to.

subsoap:
I feel that text key strings should be written in English still.

```
:title "Unable to Open Resource"
```
-> 
```
:title System.getLocale("unable_to_open_resource")
```
I don't know enough about Clojure to know if this would work but something generally like this.

Potentially include the area the text string is located to help with organization

```
:title System.getLocale("app_view_unable_to_open_resource")
```

One trick we do to find text that is not setup right is make all calls to get localized strings to return something like ZZZZZ then text that still shows up in English obviously still needs to be moved over.

However the CSS updates should be used to update text too if that is possible. I don't know yet on that.

subsoap:
For key names, using dashes instead of underscores to separate words

```
"app-view-unable-to-open-resource"
```

instead of

```
"app_view_unable_to_open_resource"
```

seems would work better to fit existing style.

subsoap:
Seems like it could look something like this for getting. It seems a little annoying, but there is already loads of code with a similar pattern. 

```
 [editor.locale :as locale]
```
```
 (locale/get-locale locale "app-view-unable-to-open-resource")
```

Like I'd prefer just get and not get-locale but get-locale is closer to existing pattern.

Ref with prefs

https://github.com/defold/defold/blob/dev/editor/src/clj/editor/prefs.clj

[...]

PR #11241, opened 2025-09-19
Localize welcome screen and preferences dialog

vlaaad:
Now the editor has localization support!

Technical notes:
Lots more stuff to do still..
Related to #5453


build_tools/http_cache.py
```diff
@@ -78,7 +78,7 @@ def download(url, cb = None, cb_count = 10):
     hit = c.get(url)
     headers = {}
     if hit:
-        headers = {'If-None-Match' : '%s' % (hit[1])}
+        headers = {'If-None-Match' : '%s' % (hit[1].decode('utf-8'))}
```
vlaaad, line 81: Without this, the editor's bundle script keeps re-downloading the build JDK on every build...

editor/.idea/encodings.xml
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="Encoding" addBOMForNewFiles="with NO BOM" />
+  <component name="Encoding" defaultCharsetForPropertiesFiles="UTF-8" />
```
vlaaad, line 3: Editor localization files are properties files, and use UTF-8

editor/src/clj/editor/localization.clj
```diff
@@ -0,0 +1,514 @@
...
+;;
+;; You may obtain a copy of the License, together with FAQs at
+;; https://www.defold.com/license
+;;
+;; Unless required by applicable law or agreed to in writing, software distributed
+;; under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+;; CONDITIONS OF ANY KIND, either express or implied. See the License for the
+;; specific language governing permissions and limitations under the License.
+
+(ns editor.localization
```
vlaaad, line 15: This is the main thing of this PR!

editor/src/java/com/ibm/icu/text/NumberFormatServiceShim.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.Locale;
+
+/**
+ * Used dynamically in {@link NumberFormat#getShim()}
+ */
+public class NumberFormatServiceShim extends NumberFormat.NumberFormatShim {
+
+    private static final NumberFormat NUMBER_FORMAT = new DecimalFormat("#0.##", new DecimalFormatSymbols(ULocale.ROOT));
+    private static final NumberFormat PERCENT_FORMAT = new DecimalFormat("#.##%", new DecimalFormatSymbols(ULocale.ROOT));
```
vlaaad, line 18: We remove the original `NumberFormatServiceShim` class from the ICU package and replace it with our own to ensure numbers are formatted uniformly in any localization. Reason: localized number formatting is suitable for currency contexts, but not for programming contexts.

editor/styling/stylesheets/components/_button.scss
```diff
@@ -61,7 +61,7 @@
 
 .button-small {
   @extend .button;
-  -fx-padding: 2px 17px;
+  -fx-padding: 2px 8px;
```
vlaaad, line 64: For diff buttons, so localized buttons fit...

editor/styling/stylesheets/modules/_welcome-dialog.scss
```diff
@@ -194,11 +182,25 @@
     }
   }
 
+  // todo
```
vlaaad, line 185: Forgot to remove the todo, I'll remove it separately

editor/styling/stylesheets/modules/_welcome-dialog.scss
```diff
@@ -5,18 +5,6 @@
     -fx-spacing: 0; // Fight crazy non-zero default.
   }
 
-  .list-view {
-    -fx-background-color: transparent;
```
vlaaad, line 9: Problem: This CSS made all list views transparent, even the one used in a drop-down for the language selector.

editor/tasks/leiningen/local_jars.clj
```diff
@@ -30,6 +31,61 @@
     :jar-file "../com.dynamo.cr/com.dynamo.cr.bob/lib/openmali.jar"
     :version "1.0"}])
 
+(defn- clean-icu4j-data
```
vlaaad, line 36: Removing unnecessary data from ICU4J because it's quite big.

Here are some numbers. First, ICU library size:
1. Default jar size: `14_663_227` bytes.
2. Jar size if all data is removed (best possible result): `3_206_748` bytes.
3. Jar size after stripping: `4_859_556` bytes.

When it comes to the resulting editor bundle size, here are the numbers:
1. Without ICU4J: `281.75 MB`
2. With unstripped ICU4J: `295.64 MB`
3. With stripped ICU4J: `286.69 MB`

So, the bundle increased in size by around 5 MB.

vlaaad:
@JCash requesting you to look at [build_tools/http_cache.py](https://github.com/defold/defold/pull/11241/files#diff-90ab0078ab1a2f5fc93896e90ff24d24d31949e1a8988843e448bcfa6259a65d) changes.

editor/src/java/com/ibm/icu/text/NumberFormatServiceShim.java
matgis: It seems strange that we would have to do this? Typically these kind of enterprisey libraries let you inject your own custom implementations. If we're unable to do so for some reason, we should document why we're doing it like this here.

editor/resources/localization/en.editor_localization
```diff
@@ -0,0 +1,253 @@
...
+dialog.error.title = Error
+dialog.error.header = An error occurred
+dialog.error.footer = You can help us fix this problem by reporting it and providing more information about what you were doing when it happened.
+
+# GL support error dialog
+
+dialog.gl-support.title = Insufficient OpenGL Support
+dialog.gl-support.header = This is a very common issue. See if any of these instructions help:
+dialog.gl-support.linux = OpenGL on linux
+dialog.gl-support.glGenBuffers = glGenBuffers
```
matgis: Inconsistent casing. Should we enforce a standard?

editor/resources/localization/en.editor_localization
```diff
@@ -0,0 +1,253 @@
...
+welcome.new-project.error.no-internet-connection = No Internet Connection
+welcome.new-project.error.description.no-internet-connection = You must be connected to the internet to download project content
+welcome.new-project.error.host-unreachable = Host Unreachable
+welcome.new-project.error.description.host-unreachable = A firewall might be blocking network connections
+welcome.new-project.error.host-not-responding = Host Not Responding
+welcome.new-project.error.description.host-not-responding = The connection timed out
+welcome.new-project.error.ssl-connection-error = SSL Connection Error
+welcome.new-project.error.ssl-connection-error.description = Could not establish an SSL connection
+welcome.new-project.error.ssl-connection-error.common-causes = Common causes are:\n\u00A0\u00A0\u2022\u00A0 Antivirus software configured to scan encrypted connections\n\u00A0\u00A0\u2022\u00A0 Expired or misconfigured server certificate\n\u00A0\u00A0\u2022\u00A0 Untrusted server certificate\n\nThe following FAQ may apply:
+welcome.new-project.error.ssl-connection-error.pkix-faq-entry = PKIX path building failed
```
matgis, line 44: There's a lot of potential for merge conflicts in a file like this. It might make sense to enforce a more rigid structure. For example, refuse the file if the entries aren't sorted. If we start generating this from a tool, should there even be separate sections with comments?

editor/resources/welcome/home-pane.fxml
```diff
@@ -5,32 +5,32 @@
 <?import javafx.scene.layout.*?>
 <?import javafx.scene.Group?>
 
-<VBox id="home-pane" HBox.hgrow="ALWAYS" xmlns="http://javafx.com/javafx">
+<VBox id="home-pane" styleClass="welcome-dialog-content" HBox.hgrow="ALWAYS" xmlns="http://javafx.com/javafx">
     <HBox styleClass="header-pane">
-        <Label styleClass="df-header" text="Recent Projects" HBox.hgrow="ALWAYS" />
+        <Label styleClass="df-header" id="recent-projects-label" HBox.hgrow="ALWAYS" />
```
matgis, line 10: Is seems you can inline `%`-prefixed bundle key identifiers like `text="%welcome.recent-projects-label"`, and supply the `ResourceBundle` directly to the `FXMLLoader`. I think this would also let you see the localized strings if you're using Scene Builder to edit the `.fxml` file. I don't use it personally, but might be neat for contributors. Source: https://stackoverflow.com/a/26325566

After reading more of the PR, I understand we use the `id` to be able to update the UI elements when the `.editor_localization` changes, but perhaps the `localize!` function could get the bundle key identifier from the `text` (or counterpart) property on the `Label` instead of supplying it as an argument to `localize!`. Might be nice to be able to use Scene Builder or similar tools with the `.fxml` files? 🤔

editor/resources/welcome/welcome-dialog.fxml
```diff
@@ -9,6 +9,7 @@
 <?import javafx.scene.image.Image?>
 <?import javafx.scene.layout.*?>
 
+<?import javafx.scene.control.ComboBox?>
```
matgis: Keep sorted alongside the other imports?

editor/src/clj/editor/code/lang/java_properties.clj
```diff
@@ -0,0 +1,66 @@
...
+        (recur
+          ;; comment?
+          (if (re-find #"^\s*#" l)
+            acc
+            ;; key-value pair?
+            (if-let [[_ k v] (re-find #"^\s*([^=]+)\s*=\s*(.+)$" l)]
+              (assoc! acc (string/trimr k) (unescape v))
+              ;; something else, ignore it
+              acc)))
+        (persistent! acc)))))
```
matgis: Would it be better to use `java.util.PropertyResourceBundle` to parse this?

editor/src/clj/editor/localization.clj
```diff
@@ -0,0 +1,514 @@
...
+      :error-handler (fn report-localization-error [_ exception]
+                       (error-reporting/report-exception! exception)))))
+
+(defn make-editor
+  "Create localization configured for use in the editor"
+  [prefs]
+  (letfn [(generated-localization []
+            (->> {"prefs.extensions.build-server.prompt" (connection-properties/defold-build-server-url)}
+                 (e/map #(str (key %) "=" (val %)))
+                 (coll/join-to-string "\n")))
```
matgis: Should this really be a localization entry? It feels a bit strange if it could potentially be overridden from another `.editor_localization` file. Sometimes a string is just a string. 🤷

editor/src/clj/editor/ui/updater.clj
```diff
@@ -14,70 +14,72 @@
...
+                    {:title (localization (localization/message "updater.dialog.title"))
                      :icon :icon/circle-question
-                     :header "Install downloaded update and restart?"
-                     :buttons [{:text "Not Now"
+                     :header (localization (localization/message "updater.dialog.header"))
+                     :buttons [{:text (localization (localization/message "updater.dialog.button.not-now"))
                                 :cancel-button true
                                 :result false}
-                               {:text "Install and Restart"
+                               {:text (localization (localization/message "updater.dialog.button.install-and-restart"))
```
matgis, line 82: This feels a bit wordy in relation to how often we need to do this. Are there situations where the `localization` takes something that is not a `message`? Could it take the resource bundle key directly instead? Or maybe we should have a function that takes a `localization` and a resource key and returns a string directly?

editor/resources/localization/en.editor_localization
```diff
@@ -0,0 +1,253 @@
...
+pane.outline = Outline
+pane.properties = Properties
+pane.console = Console
+pane.curve-editor = Curve Editor
+pane.build-errors = Build Errors
+pane.search-results = Search Results
+
+# Console
+
+console.clear = Clear
```
matgis: I would like us to come up with a good set of rules around how we name these keys before it gets out of hand. Probably best to be quite specific? For example, should this be `console.clear-button`, `console.button.clear` or something else?

editor/src/clj/editor/localization.clj
```diff
@@ -0,0 +1,514 @@
...
+                                                (if (system/defold-dev?)
+                                                  (str "!" k "! `" v "`: " (or (ex-message e) (.getSimpleName (class e))))
+                                                  v)))))))
+                        (transient {}))
+                      persistent!)]
+    (assoc state
+      :messages messages
+      :date (DateFormat/getDateInstance DateFormat/SHORT u-locale)
+      :list-and (ListFormatter/getInstance u-locale ListFormatter$Type/AND ListFormatter$Width/WIDE)
+      :list-or (ListFormatter/getInstance u-locale ListFormatter$Type/OR ListFormatter$Width/WIDE)
```
matgis, line 166: I get reflection warnings for these two for some reason? 🤔

editor/src/clj/editor/prefs.clj
```diff
@@ -335,15 +312,12 @@
...
-(s/def ::description string?)
 (s/def ::multiline boolean?) ;; for string schemas
-(s/def ::prompt string?) ;; for textual inputs
 (s/def ::ui
-  (s/keys :opt-un [::label ::description ::multiline ::prompt ::type]))
+  (s/keys :opt-un [::multiline ::type]))
 (s/def ::default any?)
 (s/def ::scope #{:global :project})
-(s/def ::type #{:any :boolean :string :password :keyword :integer :number :array :set :object :object-of :enum :tuple})
+(s/def ::type #{:any :boolean :string :password :locale :keyword :integer :number :array :set :object :object-of :enum :tuple})
```
matgis, line 325: It feels a bit hacky to introduce a new type specifically for `:locale`? Can't it be an `:enum` that we update in the combined prefs schema?

editor/tasks/leiningen/local_jars.clj
```diff
@@ -30,6 +31,61 @@
...
+  ^Path [icu4j-config]
+  (let [{:keys [version]} icu4j-config
+        jar (Files/createTempFile "" "-icu4j.jar" (into-array FileAttribute []))]
+    (.deleteOnExit (.toFile jar))
+    (with-open [fs (FileSystems/newFileSystem
+                     (Files/copy
```
matgis, line 44: I get a reflection warning here.

editor/src/clj/editor/localization.clj
```diff
@@ -0,0 +1,514 @@
...
+  IRef
+  (deref [_] @agent)
+  (setValidator [_ vf] (.setValidator agent vf))
+  (getValidator [_] (.getValidator agent))
+  (getWatches [_] (.getWatches agent))
+  (addWatch [_ key callback] (.addWatch agent key callback))
+  (removeWatch [_ key] (.removeWatch agent key))
+  IFn
+  (invoke [_ v] (impl-format @agent v))
+  (applyTo [this args] (AFn/applyToHelper this args)))
```
matgis, line 183: Why do we want this to act like a function rather than having a function take it as an argument? Maybe we'd like to add additional functions in the future? I can't Find Usages on this thing. 🤔

editor/src/clj/editor/prefs_dialog.clj
```diff
@@ -34,6 +35,7 @@
             [editor.fxui :as fxui]
             [editor.handler :as handler]
             [editor.keymap :as keymap]
+            [editor.localization :as localization]
             [editor.prefs :as prefs]
             [editor.system :as system]
             [editor.util :as eutil]
```
matgis: `eutil` is now unused.

matgis (dismissed):
Incredible work! I did have some questions and suggestions, but I think we can merge it like this for now. I do think we want to make the language dropdown a bit nicer before releasing it to our users, though (maybe show the names of the languages instead of the locale key?). We probably also want to decide on some rules for the resource key names before excited users start to contribute localizations. We could hide the language selector in release builds until we have more coverage and feel a bit more confident, perhaps?

editor/src/java/com/ibm/icu/text/NumberFormatServiceShim.java
vlaaad: I've spent some time digging into ICU4J's sources, but this was the only way I found. Given the class name (shim), it seems to me this is the expected way to intercept formatting calls.

editor/resources/localization/en.editor_localization
vlaaad: Yep, I'll fix it

editor/resources/localization/en.editor_localization
vlaaad: We are not going to start generating it from a tool since many keys are going to be dynamic (think `(str "property." (name prop-kw))`, see also: prompts and descriptions in prefs dialogs). I used comments to help me find/create groups of keys. We will use Crowdin for localization contributions; I experimented with it — it preserves comments and typically does not create merge conflicts.

editor/resources/welcome/home-pane.fxml
vlaaad: Huh, TIL :)
I'd like us to remove FXML support completely in favor of cljfx, since it's more REPL-friendly. Your comment prompted me to look at Java's built-in property parser — I will replace the custom parser I initially implemented in java-properties ns with the one that's built into JDK. It's ~3 times faster!

editor/src/clj/editor/code/lang/java_properties.clj
vlaaad: Yes, and faster!

editor/src/clj/editor/localization.clj
vlaaad: Yeah, you are right.

editor/src/clj/editor/ui/updater.clj
vlaaad: I agree that this is too wordy. I was trying several ideas, but so far, a bit more verbosity seems like the least problematic approach. `localization` does take other things: in addition to message patterns (message, and-list, or-list, date), it may take any other object and will simply stringify it. This is very useful during the transition period when not every last bit of the editor is localized yet: new code may expect message pattern instances, but will continue to work with strings.

I don't think it's going to be useful to have a function that takes localization and a key and returns a string — in terms of verbosity, it will be about the same, i.e.:
```clj
;; current:
(localization (localization/message "updater.dialog.button.install-and-restart"))

;; vs custom function (about the same):
(localization/format localization "updater.dialog.button.install-and-restart")

;; vs shorter alias (shorter, but no longer nice for auto-completion when we can 
;; use the same symbol for ns and value):
(localization (l/message "updater.dialog.button.install-and-restart"))

;; l as a symbol feels weird, though nice for auto-completion
(l (l/message "updater.dialog.button.install-and-restart"))

;; tagged literal (can't find usages in IDEA though):
(localization #localization/message "updater.dialog.button.install-and-restart")

;; maybe l10n? it's a typical shorthand:
(l10n (l10n/message "updater.dialog.button.install-and-restart"))
```

For some reason, I think you will be against using `l10n`...

editor/src/clj/editor/localization.clj
vlaaad: Weird, I have no warnings, and neither are there warnings on CI. Did you do `lein init` when you checked out the branch?

editor/src/clj/editor/prefs.clj
vlaaad: This felt less hacky than doing a special case just for locale in the prefs dialog 🤷

editor/src/clj/editor/localization.clj
vlaaad: To reduce verbosity. For additional functions, there are public functions in the ns. You can search for `(localization ` to find usages. I'm open to suggestions here! This is kinda too long though:
```clj
(localization/format localization (localization/message "..."))
```

editor/resources/localization/en.editor_localization
vlaaad: Good point, I'll do a pass over keys and also document the approach in `en.editor_localization` file.

editor/src/clj/editor/ui/updater.clj
matgis: How about we have a `text` module that takes a `localization` argument?
```clj
(text/for localization "updater.dialog.button.install-and-restart")
```
I kind of dislike the use of the term "message" when a lot of the localized text will be shorter things like labels.

vlaaad: > I kind of dislike the use of the term "message"

"Message" / "message format" are standard localization terms in ICU: https://unicode-org.github.io/icu/userguide/format_parse/messages/

matgis: I think it's fine to have the record type be named `Message`, but what I'm suggesting is we could present a friendlier interface in something like a `text` utility module.

Here's a quite verbose example from the Welcome dialog:
```clj
;; Current:
(localization/localize! progress-header localization (localization/message header-text-key {"template" (localization/message template-name-key)}))

;; Suggestion - assume the argument is a resource-key followed by key-value pairs of injections.
;; Here, the `text/of` function is used to create a free-standing `Message`:
(text/bind! progress-header localization header-text-key :template (text/of template-name-key))

;; Just doing the lookup becomes:
(text/for localization header-text-key :template (text/of template-name-key))

;; The `of` function could take injections too:
(text/of template-name-key :person "James" :greeting (text/of greeting-key))
```

PR #11262, opened 2025-09-26
Localize notifications, progress and Load Project dialog

vlaaad:
Related to #5453


editor/src/clj/editor/localization.clj
```diff
@@ -74,6 +74,11 @@
 ;;        ;; - locale-specific date formatter (e.g. 2025-09-17 or 9/17/25)
 ;;        :date com.ibm.icu.text.DateFormat}))))
 
+(def javafx-executor
+  (reify Executor
+    (execute [_ command]
+      (Platform/runLater command))))
```
vlaaad, line 80: I moved the thing here because ui ns now depends on localization...

editor/resources/localization/en.editor_localization
```diff
@@ -318,6 +318,45 @@ dialog.fetch-libraries.host-unreachable.title = Fetch Failed
...
+
+# Asset browser notifications
+
+notification.asset-browser.circular-paste.error = Cannot paste folder {source} into its subfolder {target}.
+
+# Workspace notifications
+
+notification.resource-collision.directory.warning = Folder "{resource}" is shadowing a folder with the same name
+notification.resource-collision.builtins.warning = Folder "{resource}" is shadowing a folder with the same name in builtins
+notification.resource-collision.library.warning = Folder "{resource}" is shadowing a folder with the same name in a project dependency: {library}
```
matgis, line 342: Should these end with a period?

editor/resources/localization/en.editor_localization
```diff
@@ -506,6 +549,61 @@ prefs.keymap.warning = \n• {warning}
...
+progress.executing-bundle-hook = Executing bundle hook...
+progress.building-project = Building project...
+progress.installing-updated-libraries = Installing updated libraries...
+progress.initializing-editor = Initializing editor {lib}
+progress.loading-editor = Loading editor
+progress.loading = Loading
+progress.loading-resource = Loading {resource}
+progress.resolving-resource = Resolving {resource}
+progress.reading-resource = Reading {resource}
+progress.finalizing = Finalizing
```
matgis: Should it be "Finalizing..."?

editor/resources/localization/en.editor_localization
```diff
@@ -506,6 +549,61 @@ prefs.keymap.warning = \n• {warning}
...
+progress.downloading-archives = Downloading archives...
+progress.launching-engine = Launching engine...
+progress.launched-engine = Launched {engine}
+progress.engine-launch-failed = Launch failed
+progress.rebooting-engine = Rebooting {engine}...
+progress.rebooted-engine = Rebooted {engine}
+progress.engine-reboot-failed = Reboot failed
+progress.aborting = Aborting...
+progress.percentage = {percentage}%
+
```
matgis, line 605: Should we use ellipsis characters? I don't have a strong preference, but I think we do elsewhere, and maybe we should be consistent about it?

editor/src/clj/editor/notifications_view.clj
```diff
@@ -74,10 +74,12 @@
                                          :right (- horizontal-padding close-button-margin)}
                                :spacing spacing
                                :fill-height false
-                               :children [{:fx/type fxui/legacy-label
+                               :children [{:fx/type fxui/ext-localize
                                            :h-box/hgrow :always
-                                           :max-width Double/MAX_VALUE
-                                           :text text}
+                                           :localization localization
+                                           :pattern message
```
matgis: Would it be better to name it `:message`? Seems we call it that in most places?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/ConsoleProgress.java
```diff
@@ -78,8 +78,21 @@ private void printProgress() {
...
 
     @Override
-    public void beginTask(String name, int work) {
-        System.out.print(name);
+    public void beginTask(Task name, int work) {
+        System.out.print(switch (name) {
+            case BUNDLING -> "Bundling...";
+            case BUILDING_ENGINE -> "Building engine...";
+            case CLEANING_ENGINE -> "Cleaning engine";
+            case DOWNLOADING_SYMBOLS -> "Downloading symbols...";
```
matgis: If we wanted to, we could retain the original message string by injecting `work` into this: `"Downloading %s symbols..."`.

matgis (dismissed):
Looks good! We might want to have a shared `localization/empty-message`, since we seem to construct a lot of them?

editor/resources/localization/en.editor_localization
vlaaad: Great catch!

editor/resources/localization/en.editor_localization
vlaaad: Hmm yes, I think it'll be better.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/ConsoleProgress.java
vlaaad: Ohhh nice!

PR #11264, opened 2025-09-26
Fixing up sound openal builds by adding stub start offset/time calls

thengel01:
- Standard sound.cpp code added two new methods to setup a start offset into a sound asset before playback starts
- sound_openal.cpp was lackinig this and hence would fail during link
- implementation of the methods is added, but as "stub only" (returning proper errors) as OpenAL does not trivially support this


JCash:
Superceded by https://github.com/defold/defold/pull/11264

PR #11275, opened 2025-09-29
Minimize memory footprint of input bindings

deurell:
Minimize memory footprint of input bindings. 

Fixes #4127

### Technical changes
* Optimize Action and Touch structs
* Action hashmap only allocates space for actual item count in ddf

```
Input System Memory Optimization Results:
Struct Optimization:
• Action struct: 952 → 496 bytes (-48%)
• Touch struct: 44 → 28 bytes (-36%)
Dynamic Capacity Optimization:
• Old way (fixed 256 actions): 126,976 bytes
• New way (typical 16 actions): 7,936 bytes
• Memory saved: 119,040 bytes (93.8% reduction)

ActionAndTouchStructSizesAndTotalArraySize
[Struct Sizes]
sizeof(dmInput::Action): 496 bytes (baseline: 952, diff: -456)
sizeof(dmHID::Touch):    28 bytes (baseline: 44, diff: -16)
Total Action hashmap size: 126976 bytes (baseline: 243712, diff: -116736)
Total Touch array size: 78848 bytes (baseline: 123904, diff: -45056)
[End Struct Sizes]
ActionAndTouchStructSizesAndTotalArraySize PASS (7 µs)
DynamicCapacityOptimization
[Dynamic Capacity Optimization]
Action size: 496 bytes
Old way (fixed 256): 126976 bytes
New way (typical 16): 7936 bytes
Memory saved: 119040 bytes (93.8% reduction)
[End Dynamic Capacity]
DynamicCapacityOptimization PASS (6 µs)
```

### DIFF
--- a/engine/hid/src/dmsdk/hid/hid.h
+++ b/engine/hid/src/dmsdk/hid/hid.h
@@ -502,27 +502,27 @@ namespace dmHID
     struct Touch
     {
         /// Single-click, double, etc
-        int32_t m_TapCount;
+        uint16_t m_TapCount;
         /// Begin, end, etc
         Phase   m_Phase;
         /// Current x
-        int32_t m_X;
+        int16_t m_X;
         /// Current y
-        int32_t m_Y;
+        int16_t m_Y;
         /// Current x, in screen space
-        int32_t m_ScreenX;
+        int16_t m_ScreenX;
         /// Current y, in screen space
-        int32_t m_ScreenY;
+        int16_t m_ScreenY;
         /// Current dx
-        int32_t m_DX;
+        int16_t m_DX;
         /// Current dy
-        int32_t m_DY;
+        int16_t m_DY;
         /// Current dx, in screen space
-        int32_t m_ScreenDX;
+        int16_t m_ScreenDX;
         /// Current dy, in screen space
-        int32_t m_ScreenDY;
+        int16_t m_ScreenDY;
         /// Touch id
-        int32_t m_Id;
+        int16_t m_Id;
     };
 
     /*# gets a keyboard handle
--- a/engine/hid/src/dmsdk/hid/hid.h
+++ b/engine/hid/src/dmsdk/hid/hid.h
@@ -565,7 +565,7 @@ namespace dmHID
      * @param out [type: void**] Platform specific user id data
      * @return result [type: boolean] true if gamepad has a user id data assigned to it
      */
-    bool GetGamepadUserId(HContext context, HGamepad gamepad, uint32_t* out);
+    bool GetGamepadUserId(HContext context, HGamepad gamepad, uint16_t* out);
 
     /*# Adds a touch event touch.
      * @name AddTouch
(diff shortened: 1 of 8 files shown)


engine/hid/src/dmsdk/hid/hid.h
```diff
@@ -565,7 +565,7 @@ namespace dmHID
      * @param out [type: void**] Platform specific user id data
      * @return result [type: boolean] true if gamepad has a user id data assigned to it
      */
-    bool GetGamepadUserId(HContext context, HGamepad gamepad, uint32_t* out);
+    bool GetGamepadUserId(HContext context, HGamepad gamepad, uint16_t* out);
```
JCash: Headers in `dmsdk` are public, and we try hard not to change their api.
Is there a real gain in this function changing the api?
This is a breaking change and extensions will stop compiling.

engine/hid/src/dmsdk/hid/hid.h
```diff
@@ -502,27 +502,27 @@ namespace dmHID
...
+        int16_t m_DY;
         /// Current dx, in screen space
-        int32_t m_ScreenDX;
+        int16_t m_ScreenDX;
         /// Current dy, in screen space
-        int32_t m_ScreenDY;
+        int16_t m_ScreenDY;
         /// Touch id
-        int32_t m_Id;
+        int16_t m_Id;
```
JCash, line 525: I suspect less users will be affected by this breaking change, other than getting compile warnings.
But, in truth, it's one of the cases where we would like to add a getter/setter api around that structure.
But, I think we can leave it as-is for now.

engine/input/src/input.cpp
```diff
@@ -273,7 +272,17 @@ namespace dmInput
...
     {
+        uint32_t total_actions = ddf->m_KeyTrigger.m_Count +
+            ddf->m_MouseTrigger.m_Count +
+            ddf->m_TouchTrigger.m_Count +
+            ddf->m_TextTrigger.m_Count + 2;
+
+        uint32_t current_capacity = binding->m_Actions.Capacity();
+        uint32_t new_capacity = total_actions > current_capacity ? total_actions : current_capacity;
+        uint32_t bucket_count = dmMath::Max(4u, new_capacity / 4);
+        binding->m_Actions.SetCapacity(bucket_count, new_capacity);
```
JCash: Prefer "dmHashTable::OffsetCapacity()" to avoid having to do these calculations on your own.
Mostly for visibility to others, but still.

JCash (changes requested):
overall good.
Need to preserve the public api though.

engine/input/src/input.cpp
deurell: 👍 Missed that one. Tried to keep the bucket/element ratio from the 64/256 usage we had in the current setup.

engine/hid/src/dmsdk/hid/hid.h
deurell: will fix.

JCash (approved):
lgtm

PR #11288, opened 2025-10-02
Localize many menu items

vlaaad:
Related to #5453


editor/src/clj/editor/app_view.clj
```diff
@@ -1719,162 +1716,162 @@
             (process/start! {:dir install-dir} (system/defold-launcherpath)))))
 
 (handler/register-menu! ::menubar
-  [{:label "File"
+  [{:label (localization/message "command.file")
```
matgis: The menu labels themselves are arguably not commands. Should these keys be named differently? Not sure how I feel about this myself. 🤔

editor/src/clj/editor/atlas.clj
```diff
@@ -1050,7 +1050,7 @@
       (select! app-view image-nodes op-seq))))
 
 (handler/defhandler :edit.add-referenced-component :workbench
-  (label [] "Add Images...")
+  :label (localization/message "command.edit.add-referenced-component.variant.atlas")
```
matgis, line 1053: I would prefer a more descriptive key for commands associated with things like the reused "component"-related handlers.

The reason I'm not wild about reusing the `add-referenced-component` key is that it might make adding additional handlers that add new things awkward. I feel this is a quirk of how we overload these handlers to mean different things in different contexts, so they can share a keyboard shortcut. But there is no need for that to leak onto the localization keys.

Perhaps we could instead involve the `:resource-kind` in the key so extensions like `extension-texturepacker` can share it? I.e. something like `command.resource-kind.atlas.add-images`?

editor/src/clj/editor/collection.clj
```diff
@@ -804,8 +804,8 @@
   (active? [selection] (or (selection->collection selection)
                          (selection->game-object-instance selection)))
   (label [selection] (if (selection->collection selection)
-                       "Add Collection File"
-                       "Add Game Object File"))
+                       (localization/message "command.edit.add-secondary-referenced-component.variant.collection")
+                       (localization/message "command.edit.add-secondary-referenced-component.variant.collection-game-object")))
```
matgis, line 808: Again, I'd prefer to not expose the fact that we're sharing handler keys for a bunch of these into the localization keys.

editor/src/clj/editor/particlefx.clj
```diff
@@ -1198,18 +1205,19 @@
...
            (when (not user-data)
-             (let [self (selection->particlefx selection)]
-               (mapv (fn [[type data]] {:label (:label data)
-                                        :icon emitter-icon
-                                        :command :edit.add-embedded-component
-                                        :user-data {:emitter-type type}}) emitter-types)))))
+             (mapv (fn [[type data]] {:label (:label data)
+                                      :icon emitter-icon
+                                      :command :edit.add-embedded-component
+                                      :user-data {:emitter-type type}})
```
matgis: Let's put the resulting map on a new line from the `fn` and arglist while we're touching this!

editor/src/clj/editor/properties.clj
```diff
@@ -1198,78 +1199,43 @@
...
+                      (first (first target-aspects))
+                      (localization/vary-message-variables
+                        (let [target-kinds (into #{} (map second) target-aspects)]
+                          (if (= 1 (count target-kinds))
+                            (first target-kinds)
+                            (localization/message "override.aspect.generic.kind")))
+                        assoc "count" target-aspect-count))
+           "resource_count" (count target-proj-paths)
+           "resource" (first target-proj-paths)})
+        (localization/transform string/replace "_" "__"))))
```
matgis, line 1238: All that complexity condensed surprisingly nicely after all! ❤️

editor/src/clj/editor/tile_map.clj
```diff
@@ -28,6 +28,7 @@
             [editor.graph-util :as gu]
             [editor.grid :as grid]
             [editor.handler :as handler]
+            [editor.localization :as localization]
```
matgis: Let's sort the requires while we're here!

editor/src/clj/editor/ui.clj
```diff
@@ -1296,9 +1296,10 @@
        (when-let [user-data (some-> (select-items
                                       options
                                       {:title (handler/label handler-ctx)
-                                       :filter-on :label
-                                       :cell-fn (fn [{:keys [label icon]}]
-                                                  (cond-> {:text label}
+                                       :filter-on (fn [{:keys [label]} localization]
+                                                    (localization label))
+                                       :cell-fn (fn [{:keys [label icon]} localization]
+                                                  (cond-> {:text (localization label)}
```
matgis, line 1302: It might be better to generate the localized `:label` map entries earlier in the process so we don't have to realize the localized message string multiple times?

editor/resources/localization/en.editor_localization
```diff
@@ -603,6 +616,237 @@ progress.engine-reboot-failed = Reboot failed
...
+command.file.open-as.option.in-defold-editor = {view} in Defold Editor
+command.file.open-selected = Open
+command.file.load-external-changes = Load External Changes
+command.file.save-all = Save All
+command.file.save-and-upgrade-all = Upgrade File Formats…
+command.file.search = Search in Files…
+command.file.reopen-recent = Re-Open Closed File
+command.private.recent-files.option.entry = {path} • {view}
+command.private.recent-files = Recent Files
+command.private.recent-files.option.more = More...
```
matgis: The new messages contain a mix of three dots (`...`) and ellipsis (`…`). We should pick one!

matgis (changes requested):
Looks good overall!

I do not entirely agree with the approach taken around the overloaded command handlers like `:edit.add-referenced-component`. If I remember correctly, we share the handlers so that users will be able to configure a common keyboard shortcut for these handlers that will work across multiple situations. However, I think we should avold imposing this quirk onto the localization keys for these commands as well. Basically get rid of the `.variant.` keys and put in descriptive keys instead. Possibly, we could incorporate the `:resource-kind` into the key where applicable?

editor/src/clj/editor/app_view.clj
vlaaad: Perhaps `menu.file`? I'll rename them!

editor/src/clj/editor/ui.clj
vlaaad: The view is virtualized; we shouldn't generate localized labels for thousands of objects when only 20 are shown at a time.

vlaaad:
I agree that the variant situation is not perfect, but I think the consistency of localization naming (`command.<id>`) is more important here than hiding wrong handler abstractions. We definitely should improve how we define these handlers, and additionally, the "embedded/referenced/primary/secondary" component system is also a wrong abstraction, and all these commands should be rethought and renamed. I would prefer if we did this separately, and not expand the scope of this already gigantic task even more. I'm sure it would be easy to rename localization keys when the time comes.

matgis (approved):
Ok, let's get this merged, then! 👍

vlaaad:
I created an issue for the add command redesign https://github.com/defold/defold/issues/11303

PR #11289, opened 2025-10-02
Add scene-view-batches function to editor dev module

matgis:
Added a function to simulate the batches of renderables that will be drawn for a specific pass in the scene-view. Useful for debugging.

### DIFF
--- a/editor/src/clj/internal/util.clj
+++ b/editor/src/clj/internal/util.clj
@@ -413,14 +413,20 @@
     (map? coll)
     (if (sorted? coll)
       coll
-      (with-meta (into (sorted-map) coll)
-                 (meta coll)))
+      (try
+        (with-meta (into (sorted-map) coll)
+                   (meta coll))
+        (catch ClassCastException _
+          coll)))
 
     (set? coll)
     (if (sorted? coll)
       coll
-      (with-meta (into (sorted-set) coll)
-                 (meta coll)))
+      (try
+        (with-meta (into (sorted-set) coll)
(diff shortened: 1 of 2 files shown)


editor/src/clj/internal/util.clj
```diff
@@ -413,14 +413,20 @@
...
     (set? coll)
     (if (sorted? coll)
       coll
-      (with-meta (into (sorted-set) coll)
-                 (meta coll)))
+      (try
+        (with-meta (into (sorted-set) coll)
+                   (meta coll))
+        (catch ClassCastException _
+          coll)))
```
matgis, line 429: These threw an exception when it encountered map keys that aren't comparable. Now it will include the original unsorted collection instead.

This function is not used in production, only for test output and dev functions.

editor/src/dev/dev.clj
```diff
@@ -284,13 +284,17 @@
...
+                        (into (with-meta target-map
+                                         (meta value))
+                              (keep (fn [[k v]]
+                                      (when-some [v' (util/deep-keep-kv-helper deep-keep-finalize-coll-value-fn wrapped-value-fn k v)]
+                                        (pair k v'))))
+                              value))]
+                (try
+                  (finalize-into (sorted-map) value)
+                  (catch ClassCastException _
+                    (finalize-into {} value))))))]
```
matgis, line 297: Similar issue around non-comparable map keys.

vlaaad (approved):
Shipit!

PR #11294, opened 2025-10-04
Ddf recursive data formats support

Jhonnyg:
Adds support for creating recursive protouf messages, e.g:

```
message MessageRecursiveA
{
    optional int32 val_a = 1;
    optional MessageRecursiveB my_b = 2;
}

message MessageRecursiveB
{
    optional int32 val_b = 1;
    optional MessageRecursiveA my_a = 2;
}
```

This will yield the .h structs:

```
struct MessageRecursiveA
{
    int32_t                    m_ValA;
    TestDDF::MessageRecursiveB*m_MyB;
    // ...
};

struct MessageRecursiveB
{
    int32_t                    m_ValB;
    TestDDF::MessageRecursiveA m_MyA;
    // ...
};
```

(Note the pointer to m_MyB here)

This can then be used to represent arbitrary tree structures (or a full JSON implementation).


engine/ddf/src/ddf/ddf.cpp
```diff
@@ -126,21 +209,24 @@ namespace dmDDF
...
                         }
 
-                        Result e = CalculateRepeated(load_context, &sub_ib, field->m_MessageDescriptor);
-                        #else
-                        InputBuffer sub_ib = ib;
-                        sub_ib->m_End = sub_ib->m_Current + length;
-                        Result e = CalculateRepeated(load_context, &sub_ib, field->m_MessageDescriptor);
-                        #endif
+                        is_dynamic_type = is_dynamic_type || !field->m_FullyDefinedType;
+                        if (is_dynamic_type && *array_info_hash != 0)
```
Jhonnyg, line 219: Dynamic message elements that are part of an array needs to be placed in the dynamic portion of the message buffer, since repeated fields are stored sequentially in memory.

engine/ddf/src/ddf/ddf.cpp
```diff
@@ -81,7 +86,84 @@ namespace dmDDF
         return GetDescriptorFromHash(dmHashString64(name));
     }
 
-    static Result CalculateRepeated(LoadContext* load_context, InputBuffer* ib, const Descriptor* desc)
+    static Result CalculateDynamicDescriptorSize(LoadContext* load_context, InputBuffer* ib, const Descriptor* desc, bool is_dynamic_type)
```
JCash, line 89: As this affects all our messages in the engine, I wonder if we could save on this dryrun to figure out the correct buffer sizes, by storing these numbers in a header(or footer) instead?
Not sure what the implications are but I feel finding out how much memory was used when creating the message, should be easily stored and retrieved in the blob.

engine/ddf/src/ddf/ddf_save.cpp
```diff
@@ -134,18 +125,36 @@ namespace dmDDF
 
                     case TYPE_MESSAGE:
                     {
+                        const uint8_t* data_tmp = data;
```
JCash, line 128: this one feels redundant?

engine/ddf/src/ddf/ddf_save.cpp
```diff
@@ -134,18 +125,36 @@ namespace dmDDF
...
+
+                        // The data stored in the buffer for "not fully defined" fields is a pointer to a
+                        // separate dynamic memory area, which means that we need to resolve the pointer
+                        // to where the actual data resides.
+                        if (!field_desc->m_FullyDefinedType)
+                        {
+                            uintptr_t dynamic_ptr = *((uintptr_t*) data);
+                            if (dynamic_ptr == 0)
```
JCash, line 136: why not check `data == 0` ?

JCash (approved):
lgtm
minor code comment

bigger question is if we can store the decompressed size (for the load context), instead of calculating it each time? Should help with performance?

PR #11305, opened 2025-10-07
Localize many properties

vlaaad:
Related to #5453


editor/src/clj/dynamo/graph.clj
```diff
@@ -335,7 +335,7 @@
                              :type                                 s/Any
                              s/Keyword                             s/Any}}
      (s/optional-key :node-id) s/Int
-     (s/optional-key :display-order) [(s/conditional vector? [(s/one String "category") s/Keyword] keyword? s/Keyword)]})
+     (s/optional-key :display-order) [(s/conditional vector? [(s/one s/Any "category") s/Keyword] keyword? s/Keyword)]})
```
vlaaad, line 338: We now support localization message patterns too, but I don't want dynamo.graph to depend on localization ns that depends on JavaFX...

editor/resources/localization/en.editor_localization
```diff
@@ -850,7 +850,343 @@ override.aspect.generic.kind = {count, plural, one {# Aspect} other {# Aspects}}
...
+operation.collection.add-game-object = Add Game Object
+operation.game-object.add-component = Add Component
+operation.gui.add-gui-node = Add GUI Node
+operation.gui.add-layer = Add Layer
+operation.gui.add-layout = Add Layout
+operation.gui.add-resources = Add {type}
+operation.particlefx.add-emitter = Add Emitter
+operation.particlefx.add-modifier = Add Modifier
+operation.property.clear = Clear {property}
+operation.property.set = Set {property}
```
vlaaad, line 869: I localized operation labels too, because if properties are localized, operations need to be localized too, since they depend on properties.

editor/resources/localization/en.editor_localization
```diff
@@ -850,7 +850,343 @@ override.aspect.generic.kind = {count, plural, one {# Aspect} other {# Aspects}}
...
+property.atlas.margin = Margin
+property.atlas.margin.tooltip = Pixels inserted between each image on the atlas
+property.atlas.inner-padding = Inner Padding
+property.atlas.inner-padding.tooltip = Empty pixels padded around each image
+property.atlas.extrude-borders = Extrude Borders
+property.atlas.extrude-borders.tooltip = Copies edge pixels outward to avoid bleeding between images
+property.atlas.max-page-size = Max Page Size
+property.atlas.max-page-size.tooltip = Largest width and height for each page in a multi-page atlas
+property.atlas.rename-patterns = Rename Patterns
+property.atlas.rename-patterns.tooltip = Comma-separated search=replace patterns applied to each image base name; for example "hat=cat,_normal=" renames "hat_normal" to "cat", which helps match animations across atlases
```
vlaaad, line 945: I took the liberty of adding some tooltips.

editor/resources/localization/en.editor_localization
```diff
@@ -850,7 +850,343 @@ override.aspect.generic.kind = {count, plural, one {# Aspect} other {# Aspects}}
...
+property.tile-source.tile-spacing = Tile Spacing
+property.tile-source.tile-width = Tile Width
+
+# Resources
+
+resource.view.scene = Scene
+resource.view.html = HTML
+resource.view.form = Form
+resource.view.code = Code
+resource.view.text = Text
```
vlaaad, line 1189: Bonus: translated view names.

editor/src/clj/editor/core.clj
```diff
@@ -119,38 +118,5 @@ When a Scope is deleted, all nodes within that scope will also be deleted."
...
-  (output outline-label    g/Str :abstract)
-  (output outline-commands [types/OutlineCommand] (g/constantly []))
-  (output outline-tree     types/OutlineItem
-          (g/fnk [_node-id outline-label outline-commands outline-children]
-               {:label outline-label
-                ;; :icon "my type of icon"
-                :node-ref _node-id
-                :commands outline-commands
-                :children outline-children})))
-
```
vlaaad, line 154: Accidentally discovered some unused code.

editor/src/clj/editor/curve_view.clj
```diff
@@ -103,18 +103,18 @@
 (g/defnk produce-curve-renderables [visible-curves]
   (let [splines+colors (mapv (fn [{:keys [curve hue saturation]}]
                                (let [spline (->> curve
-                                              (mapv second)
-                                              (properties/->spline))
+                                                 (mapv second)
+                                                 (properties/->spline))
```
vlaaad, line 105: Most of the changes in this file are indentation.

editor/src/clj/editor/curve_view.clj
```diff
@@ -612,98 +612,98 @@
...
+                                   (g/connect view-id :_node-id app-view :scene-view-ids)
+                                   (g/connect view-id :viewport grid :viewport))))]
+     (when parent
+       (let [^Node pane (scene/make-gl-pane! node-id opts)]
+         (ui/context! parent :curve-view {:view-id node-id} (SubSelectionProvider. app-view))
+         (ui/fill-control pane)
+         (ui/children! view [pane])
+         (let [converter (DefoldStringConverter.
+                           #(if (:property %)
+                              (localization (properties/label (:property %)))
```
vlaaad, line 677: Non-indentation change.

editor/src/clj/editor/localization.clj
```diff
@@ -38,6 +39,14 @@
            [javafx.scene.text Text]
            [org.apache.commons.io FilenameUtils]))
 
+;; Next line of code makes sure JavaFX is initialized, which is required during
+;; compilation even when we are not actually running the editor. To properly
+;; generate reflection-less code, clojure compiler loads classes and searches
+;; for fitting methods while compiling it. Loading javafx.scene.control.Control
+;; class requires application to be running, because it sets default platform
+;; stylesheet, and this requires Application to be running.
+(PlatformImpl/startup (fn []))
```
vlaaad, line 48: Move this from `editor.ui` since this is now an earlier place in code that uses controls (for localizing them). Without this change, running e.g., `lein test editor.handler-test` would fail in initializing localization ns.

editor/src/clj/editor/curve_view.clj
```diff
@@ -261,16 +261,16 @@
       (g/set-property controller :initial-evaluation-context nil))))
 
 (defn handle-input [self action user-data]
-  (let [^Point3d start      (g/node-value self :start)
-        ^Point3d current    (g/node-value self :current)
-        op-seq     (g/node-value self :op-seq)
-        handle    (g/node-value self :handle)
+  (let [^Point3d start (g/node-value self :start)
+        ^Point3d current (g/node-value self :current)
```
matgis: Hmm, `current` seems unused?

editor/src/clj/editor/curve_view.clj
```diff
@@ -612,98 +612,98 @@
...
+           (.setCellFactory list
+                            (reify Callback
+                              (call ^ListCell [this list]
+                                (proxy [CheckBoxListCell] [selected-callback converter]
+                                  (updateItem [item empty]
+                                    (let [this ^CheckBoxListCell this]
+                                      (proxy-super updateItem item empty)
+                                      (when (and item (not empty))
+                                        (let [[r g b] (colors/hsl->rgb (:hue item) (:saturation item) 0.75)]
+                                          (proxy-super setStyle (format "-fx-text-fill: rgb(%d, %d, %d);" (int (* 255 r)) (int (* 255 g)) (int (* 255 b)))))))))))))))
```
matgis: Style: Let's move the `list` argument to its own line to avoid increasing the indent?

editor/src/clj/editor/properties.clj
```diff
@@ -594,10 +597,16 @@
...
+
+(defn tooltip-message [property-message-key]
+  (localization/message (str property-message-key ".tooltip") {} "none"))
 
 (defn tooltip [property]
-  (:tooltip property))
+  (or (:tooltip property)
+      (let [k (:key property)
+            k (if (vector? k) (last k) k)]
+        (tooltip-message (str "property." (name k))))))
```
matgis: Perhaps the `tooltip-message` function could take the resolved `k` as an argument instead? That way we could `fn/memoize` it to return the same message instance each time. I worry a bit about creating many instances of messages and strings that will add up to a lot of duplicated memory across thousands of objects.

editor/src/clj/editor/scene_selection.clj
```diff
@@ -133,7 +133,7 @@
       (concat
         (drop-fn resources)
         (g/operation-sequence op-seq)
-        (g/operation-label "Drop Resources")))))
+        (g/operation-label (localization/message "operation.drop"))))))
```
matgis, line 136: Was it intentional to change the message to "Drop" here?

matgis (dismissed):
I was kind of hoping we could avoid adding `label` and `tooltip` dynamics to properties everywhere, and instead infer the localization key from the combination of `node-type-kw` and `prop-kw`, but you already went ahead and did it.

I do feel this change adds quite a lot of additional noise to the `defnode` forms. Can we do better?

matgis:
When testing some more, I noticed that the Russian property labels can be quite a lot longer than the English ones, which causes the property editor to dedicate more than half of its width to the label column. We may want to try to shorten the a bit if possible. For example, have a look at the properties of a `.collisionobject`.

editor/src/clj/editor/scene_selection.clj
vlaaad: Yes, I thought it's unnecessarily specific for something we don't even show in the editor... at all...

vlaaad:
I tried my best to shorten the russian property translation. Regarding property key inference, in some cases, neither the namespace, the node name, nor the property name adds up to a good localization key. We could extend the defnode macro so it's parameterized by default localization "domain" like, e.g., `camera` in `CameraNode` from `camera-editor` ns, but I think we might be stretching the graph node type abstraction a bit too far when we add localization there. Perhaps a little repetition is better than a wrong abstraction. I refactored the label and tooltip dynamics to make them shorter though.

PR #11307, opened 2025-10-07
Fixes for spine extension

AGulev:
See https://github.com/defold/extension-spine/pull/236

### DIFF
--- a/editor/test/integration/save_data_test.clj
+++ b/editor/test/integration/save_data_test.clj
@@ -379,24 +379,24 @@
 
    ['dmGuiDDF.NodeDesc "[TYPE_TEXT]"]
    {:default
-   {"clipping_inverted" :unused
-    "clipping_mode" :unused
-    "clipping_visible" :unused
-    "custom_type" :unused
-    "innerRadius" :unused
-    "outerBounds" :unused
-    "particlefx" :unused
-    "perimeterVertices" :unused
-    "pieFillAngle" :unused
-    "size_mode" :unused
-    "slice9" :unused
-    "spine_create_bones" :unused
-    "spine_default_animation" :unused
-    "spine_node_child" :unused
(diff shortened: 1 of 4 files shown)


editor/test/integration/save_data_test.clj
```diff
@@ -272,6 +273,7 @@
      "size" :unused
      "size_mode" :unused
      "slice9" :unused
+     "spine_create_bones" :unused
```
matgis: This is for `TYPE_CUSTOM`, which is the NodeDesc we use for `SpineNode`. `spine_create_bones` should not be declared `:unused` here.

Instead, we want to set the field to a non-default value in `test/resources/save_data_project/checked01.gui` to verify that the value is saved correctly. The idea is to cover a non-default value under basic usage, and also overrides for it from a layout node and a template node.

The best way to do this is to open up `test/resources/save_data_project`, edit the Spine node in `checked01.gui`, save, and run the tests in `save_data_test.clj` with a REPL connected to the editor. The tests will pick up the saved changes between test runs, but large libraries like `extension-rive` won't have to be downloaded every attempt.

The other file, `checked02.gui`, exists to cover some situations that cannot be covered by a single file. I don't think you'll need to edit it for this added field. 🤔

matgis (changes requested):

editor/test/integration/save_data_test.clj
AGulev: I'll be able to fix it only when spine extension PR will be merged, so I can update spine dependency

britzl (dismissed):

editor/test/integration/save_data_test.clj
AGulev: @matgis  Tests have been fixed

matgis (approved):
LGTM! ❤️

PR #11309, opened 2025-10-08
Free camera mode in scene view + enhanced camera controls

### LINKED ISSUE #8840: Clipping issue in the editor viewport
In the Defold editor, if I import a 3d model and set the camera to Perspective mode, weird clipping applied that prevents all the models from being displayed properly. In Orthographic mode it works as expected. Please see the attached video.

**To Reproduce (REQUIRED)**
1. Create new project using basic 3D template
2. Import any model (tested with gltf/glb and dae)
3. Set viewport camera to perspective mode

Defold editor doesn't apply any specific clipping. For the example of expected behaviour please see attached videos captured from Blender and Godot 4.1 stable

  - Version 1.7.0

**Platforms (REQUIRED):**
 - Platforms: MacOS
 - OS: Sonoma 14.3
 - Device: MacBook Pro 15 late 2019

https://github.com/icewind/TestDefoldEditorClipping

Videos
Issue - https://github.com/defold/defold/assets/575985/42d3278f-9fa3-48bc-ab95-15b4a9b01f4b
Expected behavior from Blender -  https://github.com/defold/defold/assets/575985/45b24641-a1cb-485b-b0c4-79bc78fc4962
Expected behavior from Godot4 - https://github.com/defold/defold/assets/575985/50fbb207-8b1f-4862-b358-64704a3aad44

sprocketc:
We have added a new 3D perspective free camera mode for navigating scenes! Click and hold the right mouse button to activate and move around with the WASD keys. There is a scene view popup on the perspective camera button with additional controls for the camera.  The left mouse button controls (orbit, zoom, pan) now warp around the editor window to allow you to keep going forever. There are also several smaller improvements, like smoother zooming and localization for the grid menu.

https://github.com/user-attachments/assets/95d87155-2d0f-441c-a234-700b2a679ac2

https://github.com/user-attachments/assets/65e5a7b4-70ec-4ac4-8a43-5c9436792bd9

https://github.com/user-attachments/assets/c4f948d7-0312-4bdc-a458-9d65eca6016f

https://github.com/user-attachments/assets/cd02e73e-2cce-459a-9669-103313eae09f

https://github.com/user-attachments/assets/7f971d4a-bd6c-4852-a893-b515bc817233

https://github.com/user-attachments/assets/455e3412-a00d-4942-8048-6aae894caa28

Shoutout to @sprocketc who started the PR and @delinx who contributed his winapi know-how to the mouse_capture_windows.cpp module.

Fixes #11185
Fixes #8840

### Technical changes

Shoutout to @sprocketc who started the PR and @delinx who contributed his winapi know-how to the `mouse_capture_windows.cpp` module.

Several systems needed to be implemented, enhanced, or refactored to make this work properly.

1.) The major implementation is the native mouse capture controls. JavaFX doesn't support FPS-game style mouse capture, you get relative raw mouse deltas, so the [SDL](https://github.com/libsdl-org/SDL/) code was used a reference to create a cross-platform (Linux, Windows, Macos) mouse capture module using JNA for FFI, along with the CI changes to build and deploy.

3.) `scene.clj` and `input.clj` had a mechanism to accumulate input events into a queue and call input handlers so they could process them, however, it had no mechanism to call handlers at a consistent FPS where input state could be kept track of so the camera could do things like ignore keyboard repeat rate and have proper smoothing/interpolation. `SceneView` and the `input` module were enhanced to both accept `update-tick-handlers` inputs, process and update events into a persistent `input-state` that could be read by said update tick handlers.  `input.clj` also did not support keyboard events, so it was enhanced with functionality to track presses and releases along with modifiers.

4.) For the scene view popup, `grid.clj` implemented it's own system for displaying controls, while `scene_visibility.clj` implemented it's own system, sometimes duplicating behavior. Neither had a clean way to reuse for the camera popup controls so a refactor was done where a more data-oriented approach is now used to generalize the grid and camera scene popups, greatly reducing duplication, along with some but not all code from `scene-visibility`. Further work will be carried out in #12174 in order to finally fully unify all the popups, achieving feature parity between them, and make it easier to implement new popups.


Darkinggq:
Very cool and useful fly camera for 3D scene :)

engine/tools/src/mouse_capture/mouse_capture.h
```diff
@@ -0,0 +1,24 @@
+#ifndef DM_MOUSE_CAPTURE_H
```
JCash, line 15: The source files in our repo should have our LICENSE.txt at the top.
See e.g. `./apply_license.py`

engine/tools/src/mouse_capture/mouse_capture_linux.cpp
```diff
@@ -0,0 +1,179 @@
+#include "mouse_capture.h"
+#include <string.h>
+#include <X11/extensions/XInput2.h>
+#include <X11/extensions/Xfixes.h>
```
JCash, line 18: Will this always work on Wayland?

engine/tools/src/mouse_capture/mouse_capture_linux.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+        Window root = DefaultRootWindow(context->m_Display);
+        XWarpPointer(context->m_Display, None, root, 0, 0, 0, 0, context->m_SavedCursorX, context->m_SavedCursorY);
+        XFlush(context->m_Display);
+
+        return (context->m_AccumulatedDelta.dx != 0.0 ||
+                context->m_AccumulatedDelta.dy != 0.0);
+    }
+
+    void DestroyContext(HContext context)
```
JCash: Nitpick; prefer putting create/destroy close to each other in a file, to make it easier to read the respective create/destroy flows.

engine/tools/src/mouse_capture/mouse_capture_windows.cpp
```diff
@@ -0,0 +1,280 @@
...
+    // JavaFX was dispatching WM_INPUT events to our message window, so by adding this, we can catch the ones that were being
+    // dropped, but we also only handle WM_INPUT messages because we dispatch other messages we don't care about when we PollDelta
+    static LRESULT CALLBACK RawInputWndProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam)
+    {
+        if (msg == WM_INPUT)
+        {
+            Context*  context = reinterpret_cast<Context*>(GetWindowLongPtr(hwnd, GWLP_USERDATA));
+            UINT      size = sizeof(RAWINPUTHEADER) * 2;
+            RAWINPUT* raw = (RAWINPUT*)_alloca(size);
+            GetRawInputData((HRAWINPUT)lParam, RID_INPUT, raw, &size, sizeof(RAWINPUTHEADER));
```
JCash: I don't follow this allocation, could you elaborate?
Why is it the size of two (very small) [headers](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-rawinputheader), and not large enough to fit a [rawinput](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-rawinput) struct?

engine/tools/src/mouse_capture/mouse_capture_windows.cpp
```diff
@@ -0,0 +1,280 @@
...
+    // Just in case we accumulated some deltas between the stop and start capture, let's clear them
+    static void DiscardRawInput()
+    {
+        static constexpr size_t raw_input_buffer_size_in_u64s = (1024 * 6) / sizeof(uint64_t);
+        uint64_t                raw_input_buffer[raw_input_buffer_size_in_u64s] = {};
+        MouseDelta              frame_mouse_delta = {};
+
+        while (true)
+        {
+            PRAWINPUT rawinput_buffer_ptr = reinterpret_cast<PRAWINPUT>(&raw_input_buffer);
```
JCash: Style: we use C-style casts. `PRAWINPUT rawinput_buffer_ptr = &raw_input_buffer;`

engine/tools/src/mouse_capture/mouse_capture_windows.cpp
```diff
@@ -0,0 +1,280 @@
...
+    void WarpCursor(int x, int y)
+    {
+        SetCursorPos(x, y);
+    }
+
+    // Just in case we accumulated some deltas between the stop and start capture, let's clear them
+    static void DiscardRawInput()
+    {
+        static constexpr size_t raw_input_buffer_size_in_u64s = (1024 * 6) / sizeof(uint64_t);
+        uint64_t                raw_input_buffer[raw_input_buffer_size_in_u64s] = {};
```
JCash: This looks odd.
It looks like you want a byte buffer, but calculate it in terms of `uint64's`?

Why not
```c++
size_t buffer_size = 768;
uint8_t buffer[buffer_size];
```

Also, why 768? (or why the `* 6` ?)

engine/tools/src/mouse_capture/mouse_capture_windows.cpp
```diff
@@ -0,0 +1,280 @@
...
+            if (rawinput_count == ~UINT(0))
+            {
+                break;
+            }
+
+            PRAWINPUT rawinput_ptr = rawinput_buffer_ptr;
+            for (size_t i = 0; i < rawinput_count; i++)
+            {
+                rawinput_ptr = NEXTRAWINPUTBLOCK(rawinput_ptr);
+            }
```
JCash: According to the docs, `NEXTRAWINPUTBLOCK` has no side effects, so this loop doesn't really do anything?

engine/tools/src/mouse_capture/mouse_capture_windows.cpp
```diff
@@ -0,0 +1,280 @@
...
+            return false;
+
+        if (!EnsureWindowClass())
+            return false;
+
+        POINT pt;
+        GetCursorPos(&pt);
+
+        context->m_MessageWindow = CreateWindowEx(
+        0, TEXT("dmMouseCaptureRawInput"), NULL, 0, 0, 0, 0, 0, HWND_MESSAGE, NULL, GetModuleHandle(NULL), NULL);
```
JCash: style: formatting

engine/tools/src/mouse_capture/mouse_capture_windows.cpp
```diff
@@ -0,0 +1,280 @@
...
+            {
+                return true;
+            }
+        }
+        exhausted = true;
+        if (PeekMessage(msg, context->m_MessageWindow, WM_INPUT + 1, ~UINT(0), PM_REMOVE))
+        {
+            return true;
+        }
+        return false;
```
JCash, line 219: I think this logic could be documented a little bit.
The function name is clear, but the logic not as much.

engine/tools/src/mouse_capture/mouse_capture_linux.cpp
```diff
@@ -0,0 +1,179 @@
...
+    {
+        if (!context)
+            return;
+
+        if (context->m_Capturing)
```
JCash: Redundant double check here and in the StopCapture.
While this is an editor tool and we don't care about code size as much, we still try to keep the same code style for our C/C++ files in order to have it "documented" for other readers.

One such thing is deciding on when to check the values, and only do it once.

engine/tools/src/mouse_capture/mouse_capture_macos.mm
```diff
@@ -0,0 +1,114 @@
...
+
+        return had_motion;
+    }
+
+    void DestroyContext(HContext context)
+    {
+        if (!context)
+            return;
+
+        if (context->m_Capturing)
```
JCash: Redundant check. See comment above.

engine/tools/src/mouse_capture/mouse_capture_windows.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+        return (out_delta->dx != 0.0 || out_delta->dy != 0.0);
+    }
+
+    void DestroyContext(HContext context)
+    {
+        if (!context)
+            return;
+
+        if (context->m_Capturing)
```
JCash: Redundant check. See comment above.

engine/tools/src/mouse_capture/wscript
```diff
@@ -0,0 +1,47 @@
...
+def build(bld):
+    platform = bld.env.PLATFORM
+
+    if 'linux' in platform:
+        platform_source = 'mouse_capture_linux.cpp'
+        platform_libs   = []
+    elif 'darwin' in platform or 'macos' in platform:
+        platform_source = 'mouse_capture_macos.mm'
+        platform_libs   = []
+    elif 'win' in platform:
```
JCash: This builds it for `x86-win32` which isn't an editor platform.

engine/tools/src/mouse_capture/wscript
```diff
@@ -0,0 +1,47 @@
...
+import waflib.Options
+import sys
+
+def configure(conf):
+    pass
+
+def build(bld):
+    platform = bld.env.PLATFORM
+
+    if 'linux' in platform:
```
JCash: This builds it for `arm64-linux` which isn't an editor platform.

engine/tools/src/mouse_capture/wscript
```diff
@@ -0,0 +1,47 @@
...
+    elif 'darwin' in platform or 'macos' in platform:
+        platform_source = 'mouse_capture_macos.mm'
+        platform_libs   = []
+    elif 'win' in platform:
+        platform_source = 'mouse_capture_windows.cpp'
+        platform_libs   = ['user32']
+    else:
+        return
+
+    bld.add_group()
```
JCash: This is redundant.
The add group is for creating a manual sync point if one hasn't set up the dependencies fully.

engine/tools/src/mouse_capture/wscript
```diff
@@ -0,0 +1,47 @@
...
+
+    bld.shlib(features = 'cxx cshlib skip_asan',
+              includes = ['.'],
+              target   = 'mouse_capture_shared',
+              use      = 'DLIB_NOASAN SOCKET PLATFORM_NULL',
```
JCash: Is `PLATFORM_NULL` really needed?

engine/tools/src/mouse_capture/wscript
```diff
@@ -0,0 +1,47 @@
...
+              lib      = platform_libs,
+              source   = ['mouse_capture_exports.cpp', platform_source])
+
+    bld.install_files('${PREFIX}/include/mouse_capture', 'mouse_capture.h')
```
JCash: Currently no need?
There is no other dependency on this (shared) library?

scripts/build.py
JCash, line 1457: You made this variable into an unused one. Please remove it.

JCash (changes requested):
Overall good, but needs some cleanup.
Also some questions about logic and memory safety.

editor/src/clj/editor/keymap.clj
```diff
@@ -805,6 +826,11 @@
                          (.getName code)))))
      (.toString sb))))
 
+(defn shortcut-key-codes
+  "Returns a set of KeyCodes for a command's shortcuts, ignoring modifiers"
+  [keymap shortcuts]
+  (into #{} (map #(.getCode ^KeyCodeCombination %)) shortcuts))
```
vlaaad: This code assumes that every shortcut is `KeyCodeCombination`, while the rest of the code here supports both `KeyCodeCombination` and `KeyCharacterCombination`. Let's support `KeyCharacterCombination`s here too (e.g., by ignoring them instead of throwing exceptions).

editor/src/clj/editor/scene.clj
```diff
@@ -1031,8 +1032,36 @@
...
     :pan (if (os/is-mac-os?) Cursor/CLOSED_HAND Cursor/MOVE)
+    :none Cursor/NONE
     Cursor/DEFAULT))
 
+(defn- screen->world
+  ^Vector3d [camera viewport ^Vector3d screen-pos]
+  (let [w4 (c/camera-unproject camera viewport screen-pos)]
+    (Vector3d. (.x w4) (.y w4) (.z w4))))
+
+;; TODO: We have an evaluation-context provided usually when we call this, maybe change the behavior
```
vlaaad: Leftover todo, maybe `view->camera` should require a basis?

editor/src/clj/editor/scene.clj
```diff
@@ -1330,33 +1248,51 @@
...
+        (c/start-free-cam-mode! image-view camera (:cursor-pos input-state))))))
+
+;; NOTE: If we don't register these commands deleting the commands default shortcut makes the command unreachable unless
+;; we clear player prefs, so add these dummy commands to keep them active
+(handler/defhandler :scene.free-camera.up :free-camera (run []))
+(handler/defhandler :scene.free-camera.down :free-camera (run []))
+(handler/defhandler :scene.free-camera.left :free-camera (run []))
+(handler/defhandler :scene.free-camera.forward :free-camera (run []))
+(handler/defhandler :scene.free-camera.right :free-camera (run []))
+(handler/defhandler :scene.free-camera.backward :free-camera (run []))
```
vlaaad: Huh, interesting!

editor/src/clj/editor/scene.clj
```diff
@@ -1439,12 +1375,19 @@
...
 
-(defn dispatch-input [input-handlers action user-data]
+(defn dispatch-input [input-handlers input-state action user-data]
   (reduce (fn [action [node-id label]]
             (when action
-              ((g/node-value node-id label) node-id action user-data)))
+              ((g/node-value node-id label) node-id input-state action user-data)))
           action input-handlers))
 
+;; TODO: Can we merge this with dispatch-input?
```
vlaaad: Leftover todo, should we do something here?

editor/src/clj/editor/curve_view.clj
```diff
@@ -643,6 +644,7 @@
                                    (g/connect camera :camera view-id :camera)
                                    (g/connect camera :camera grid :camera)
                                    (g/connect camera :input-handler view-id :input-handlers)
+                                   (g/connect camera :update-tick-handler view-id :update-tick-handlers)
```
vlaaad, line 653: It looks like the curve view now "supports" free cam navigation mode. At least it tries to do it: if I open a curve view and do right-click + drag, I get this exception:
```
java.lang.NullPointerException: Cannot invoke "javafx.scene.Node.getBoundsInLocal()" because "image_view" is null
        at editor.camera$start_free_cam_mode_BANG_.invokeStatic(camera.clj:912)
        at editor.camera$start_free_cam_mode_BANG_.invoke(camera.clj:901)
        at editor.camera$handle_input.invokeStatic(camera.clj:988)
        at editor.camera$handle_input.invoke(camera.clj:942)
        at editor.scene$dispatch_input$fn__76531.invoke(scene.clj:1381)
        at clojure.lang.PersistentVector.reduce(PersistentVector.java:418)
        at clojure.core$reduce.invokeStatic(core.clj:6964)
        at clojure.core$reduce.invoke(core.clj:6947)
        at editor.scene$dispatch_input.invokeStatic(scene.clj:1379)
        at editor.scene$dispatch_input.invoke(scene.clj:1378)
        at editor.scene$update_image_view_BANG_$fn__76556$fn__76557.invoke(scene.clj:1429)
        at clojure.lang.PersistentVector.reduce(PersistentVector.java:418)
        at clojure.core$reduce.invokeStatic(core.clj:6964)
        at clojure.core$reduce.invoke(core.clj:6947)
        at editor.scene$update_image_view_BANG_$fn__76556.invoke(scene.clj:1427)
        at editor.scene$update_image_view_BANG_.invokeStatic(scene.clj:1424)
        at editor.scene$update_image_view_BANG_.invoke(scene.clj:1403)
        at editor.scene$refresh_scene_view_BANG_.invokeStatic(scene.clj:1074)
        at editor.scene$refresh_scene_view_BANG_.invoke(scene.clj:1065)
        at editor.app_view$refresh_scene_view_BANG_.invokeStatic(app_view.clj:2105)
        at editor.app_view$refresh_scene_view_BANG_.invoke(app_view.clj:2103)
        at editor.app_view$refresh_scene_views_BANG_.invokeStatic(app_view.clj:2113)
        at editor.app_view$refresh_scene_views_BANG_.invoke(app_view.clj:2109)
        at editor.app_view$make_app_view$fn__86784.invoke(app_view.clj:2239)
        at editor.ui$__GT_timer$fn__38909$fn__38910.invoke(ui.clj:2377)
        at clojure.lang.AFn.applyToHelper(AFn.java:152)
        at clojure.lang.AFn.applyTo(AFn.java:144)
        at clojure.core$apply.invokeStatic(core.clj:667)
        at clojure.core$with_bindings_STAR_.invokeStatic(core.clj:1990)
        at clojure.core$with_bindings_STAR_.doInvoke(core.clj:1990)
        at clojure.lang.RestFn.invoke(RestFn.java:428)
        at clojure.lang.AFn.applyToHelper(AFn.java:156)
        at clojure.lang.RestFn.applyTo(RestFn.java:135)
        at clojure.core$apply.invokeStatic(core.clj:671)
        at clojure.core$bound_fn_STAR_$fn__5837.doInvoke(core.clj:2020)
        at clojure.lang.RestFn.invoke(RestFn.java:400)
        at com.sun.javafx.application.PlatformImpl.lambda$runLater$0(PlatformImpl.java:424)
```

editor/test/integration/test_util.clj
```diff
@@ -753,7 +753,7 @@
                         {:type type :x x :y y :click-count click-count :button button}
                         modifiers)
          action (scene/augment-action view action)]
-     (scene/dispatch-input handlers action user-data))))
+     (scene/dispatch-input handlers {} action user-data))))
```
vlaaad: Is it okay to pass an empty map as an `input-state` there?

editor/src/clj/editor/scene.clj
```diff
@@ -1568,72 +1518,81 @@
...
+                        pos [x y 0.0]
+                        picking-rect (selection/calc-picking-rect pos pos)]
+                    (when (= :mouse-pressed (:type action))
+                      (.requestFocus parent)
+                      (.consume e))
+                    (g/user-data-swap! view-id ::input-action-queue conj action)
+                    (g/transact
+                      (concat
+                        (when screen-x
+                          (g/set-property view-id :cursor-pos [screen-x screen-y]))
```
vlaaad: Not sure if this is the right place, but it seems that the cursor position changed from `[scene-x scene-y]` to `[screen-x screen-y]`, yet, some code (like rulers) still assumes them to be scene-local. Because of that, rulers are shown in the wrong place:

https://github.com/user-attachments/assets/94841149-5286-480d-ab08-57fa99e05dda

(Note the green and red marks in the rulers)

vlaaad (changes requested):
Great work! I love the user experience, no comments here. 

I found a couple of problems; see the comments.

Additionally, I found one more behavior that we should address before merging. If I have a scene open, focus on the Outline, and then press <kbd>Shift `</kbd>, my mouse gets stuck in the Outline, and pressing escape does not exit this mode. I need to switch to another window, then move the mouse to the right place, then switch back to exit the WASD mode:

https://github.com/user-attachments/assets/2a6dfa4e-e9d7-4b6c-b9a7-297b54fe0860

scripts/build.py
JosephFerano: I think you're referring to `lib_dir` variable. If so, it's being used on line 1451. I changed it to follow the same pattern as the previous variable

```python
bin_dir = self.build_utility.get_binary_path()
lib_dir = self.build_utility.get_library_path()
```

engine/tools/src/mouse_capture/mouse_capture_linux.cpp
JosephFerano: Yes, I tested on Gnome Wayland and it seems to support these extensions with XWayland emulation. But that's Gnome, which is pretty well supported, I can't be sure of other less popular Wayland compositors as it'll depend on whether their X11 emulation is up to par.

engine/tools/src/mouse_capture/mouse_capture_linux.cpp
JosephFerano: Actually there really doesn't seem to be a need for Create/Start and Destroy/Stop pairs. They can be merged, much cleaner. I overcomplicated this.

engine/tools/src/mouse_capture/mouse_capture_windows.cpp
JosephFerano: Yeah, this looks wrong. It should be the size of RAWINPUT

engine/tools/src/mouse_capture/mouse_capture_windows.cpp
JosephFerano: The uint64 is to guarantee 8 byte alignment, which MSDN says is required on WOW64 to ensure proper behavior. I added a NOTE. `* 6` is indeed arbitrary, it was large enough to avoid multiple iterations in the while loop during testing.

engine/tools/src/mouse_capture/mouse_capture_windows.cpp
JosephFerano: This returns the pointer to the next RAWINPUT block.

engine/tools/src/mouse_capture/mouse_capture_windows.cpp
JosephFerano: Yeah, I'm not sure. I just replaced with a sizeof(RAWINPUT), I'll test to make sure it works fine.

editor/src/clj/editor/camera.clj
```diff
@@ -1033,27 +1044,37 @@
...
+  (if (and (not i/is-wayland) image-view screen-x last-x)
+    (let [win ^Window (.getWindow (ui/main-scene))
+          win-x (.getX win)
+          win-y (.getY win)
+          win-w (.getWidth win)
+          win-h (.getHeight win)
+          screen-x ^double screen-x
+          screen-y ^double screen-y
+          padding 5.0
+          _ (println screen-x screen-y win-x win-y win-w win-h)
```
vlaaad: leftover print

editor/src/clj/editor/camera.clj
```diff
@@ -1131,22 +1146,20 @@
...
-                               :pan
-                               :none))))))))
+                             :last-y mouse-y))))))
+
+;; NOTE: Merge this back in or use an fnk thingie?
+(defn- handle-input-fnk [self input-state action user-data]
+  (handle-input self input-state action user-data))
+
+(defn- handle-update-tick-fnk [self input-state dt]
+  (handle-update-tick self input-state dt))
```
vlaaad: Huh? Looks like these functions are completely unnecessary? Are they used for REPL reloadability?

editor/src/clj/editor/scene.clj
```diff
@@ -1277,12 +1337,16 @@
                  (= :perspective))))
 
 (handler/defhandler :scene.free-camera.activate :workbench
+  (enabled? [app-view evaluation-context]
+    (some-> (active-scene-view app-view)
+            (view->camera)
+            (g/node-value :movements-enabled)
```
vlaaad: Please use `evaluation-context` for `g/node-value` and `active-scene-view`, and its basis for `view->camera`.

editor/src/clj/editor/scene.clj
```diff
@@ -1435,7 +1492,10 @@
     (profiler/profile "update-tick" -1
       (let [update-tick-handlers (g/sources-of view-id :update-tick-handlers)
             input-state (g/user-data view-id ::input-state)]
-        (dispatch-update-tick update-tick-handlers input-state dt)))
+        (reduce (fn [input-state [node-id label]]
+                  (when input-state
+                    ((g/node-value node-id label) node-id input-state dt)))
```
vlaaad: Can we use a single `evaluation-context` here so we don't create a fresh one on every iteration?

editor/src/clj/editor/ui/popup.clj
```diff
@@ -34,3 +49,197 @@
...
+     (.hide popup)
+     (let [region (StackPane.)
+           popup (make-popup owner region)
+           anchor ^Point2D (pref-popup-position (.getParent owner))]
+       (.setPrefWidth region width)
+       (ui/children! region [(doto (Region.)
+                               (ui/add-style! "popup-shadow"))
+                             (doto (VBox. 10 (ui/node-array (settings prefs localization popup prefs-path setting-descriptors hidden-settings on-change)))
+                               (.setFocusTraversable true)
+                               (ensure-focus-traversable!)
```
vlaaad, line 239: This code will perform reflection at runtime because it calls `ensure-focus-traversable!` on a `VBox`, but `ensure-focus-traversable!` is type-hinted to `Control`, which `VBox` isn't. It's better to type-hint `ensure-focus-traversable!` to `Node` that defines focus traversable.

editor/src/clj/editor/input.clj
```diff
@@ -43,47 +82,118 @@
...
+  (not (.isModifierKey code)))
+
+(defn update-input-state [state action]
+  (let [modifiers (->> [:alt :shift :meta :control]
+                       (filter action)
+                       set)
+        cursor-pos (when (and (:screen-x action) (:screen-y action))
+                     [(:screen-x action) (:screen-y action)])
+        view-pos [(:x action) (:y action)]
+        state (assoc state :view-pos view-pos)]
```
vlaaad: `action-from-jfx` has no `:x` and `:y` for `:key-pressed` and `:key-released` actions. Here, we always set `:view-pos` in the state. Can it be a problem that it's set to `[nil nil]`?

vlaaad (commented):
LGTM! We are almost there! I have a couple of questions, and let's ensure we don't create too many throw-away evaluation contexts.

editor/src/clj/editor/camera.clj
JosephFerano: Yes, sorry, that was their intent, so I could evaluate the handlers and insta reload. I had cleaned this up in the last commit this morning, not sure why github still showed you this code as still being there. Weird. Same as above.

JosephFerano: Oh I see, I think you started the review and then added more to it and in between that, I had removed this. No worries.

vlaaad: Yeah, I started review on Friday, but didn't have time to finish it then.

editor/src/clj/editor/input.clj
JosephFerano: In practice it wasn't an issue, but it's still better to avoid an assoc if nil

JCash (dismissed):

editor/src/clj/editor/app_view.clj
```diff
@@ -536,21 +537,31 @@
...
 
-(defn- get-grid-settings-button
-  [^Tab tab]
+(defn- get-settings-button [^Tab tab button-id]
   (some-> tab
           .getContent
-          (.lookup "#show-grid-settings")))
+          (.lookup button-id)))
+
+(defn- show-settings-state [app-view scene-visibility button-id evaluation-context]
```
vlaaad: `scene-visibility` is unused.

editor/src/clj/editor/app_view.clj
```diff
@@ -536,21 +537,31 @@
...
+  (some-> (g/node-value app-view :active-tab evaluation-context)
+          (get-settings-button button-id)
+          (scene-visibility/settings-visible?)))
```
vlaaad, line 548: `scene-visibility/settings-visible?` check for visibility by looking up `:editor.scene-visibility/popup` userdata, while `popup/show-settings!` uses `:editor.ui.popup/popup` key (both look like `::popup` in code). Seems like it was broken before, too?

editor/src/clj/editor/input.clj
```diff
@@ -43,47 +82,122 @@
...
+
+(defn update-input-state [state action]
+  (let [modifiers (->> [:alt :shift :meta :control]
+                       (filter action)
+                       set)
+        cursor-pos (when (and (:screen-x action) (:screen-y action))
+                     [(:screen-x action) (:screen-y action)])
+        view-pos (when (and (:x action) (:y action))
+                   [(:x action) (:y action)])
+        state (assoc state :view-pos view-pos)]
```
vlaaad: We still unconditionally assoc `:view-pos` here.

editor/src/clj/editor/scene.clj
```diff
@@ -1633,72 +1589,82 @@
...
+                        (g/set-property view-id :tool-picking-rect picking-rect))))))
+              (catch Throwable error
+                (reset! process-events? false)
+                (error-reporting/report-exception! error)))))]
+    (ui/observe (.focusedProperty (ui/main-stage))
+      (fn [_ _ focused]
+        (when (and (not focused)
+                   (c/free-cam-mode-active? camera-id))
+          (c/stop-free-cam-mode! image-view camera-id)
+          (g/user-data! view-id ::input-state (i/make-input-state)))))
```
vlaaad: We are adding a scene focus listener, but never remove it on dispose. Does it cause a memory leak? Perhaps a cleaner solution would be to add a listener to the `sceneProperty` of the `parent`. When it's set to non-nil, we subscribe to the scene focus listener; when it's set to nil, we unsubscribe. This also removes the global dependency on the `ui/main-scene`.

editor/src/clj/editor/grid.clj
```diff
@@ -284,169 +272,14 @@
...
-      (doto popup
-        (.setAnchorLocation PopupWindow$AnchorLocation/CONTENT_TOP_RIGHT)
-        (ui/on-closed! (fn [_] (ui/user-data! owner ::popup nil)))
-        (.show owner (.getX anchor) (.getY anchor))))))
+        ignore-options (g/node-value grid :options)]
+    (popup/show-settings! owner prefs localization 220 [:scene :grid]
+                          [{:key :size :type :vec3-floats}
+                           {:key :active-plane :type :vec3-toggle :label "scene-popup.grid-plane"}
+                           {:key :color :type :color :label "scene-popup.opacity"}
+                           {:key :opacity :type :slider :label "scene-popup.opacity" :min 0.0 :max 1.0}]
```
vlaaad: "Opacity" is repeated twice:

Image

editor/src/clj/editor/scene.clj
```diff
@@ -1385,38 +1305,63 @@
...
-                          (realign-camera true)))
+  (run [app-view]
+    (g/with-auto-evaluation-context evaluation-context
+      (when-let [scene-view (active-scene-view app-view evaluation-context)]
+        (when-let [camera (view->camera (:basis evaluation-context) scene-view)]
+          (c/realign-camera camera true evaluation-context)))))
   (state [app-view evaluation-context]
-         (camera-2d? (active-scene-view app-view evaluation-context)
-                     evaluation-context)))
+    (c/camera-2d? (view->camera (active-scene-view app-view evaluation-context)) evaluation-context)))
```
vlaaad: `view->camera` needs a basis from `evaluation-context`.

editor/src/clj/editor/scene.clj
```diff
@@ -1385,38 +1305,63 @@
...
 ;; Used in the scene view tool bar.
 (handler/defhandler :scene.toggle-camera-type :workbench
   (active? [app-view evaluation-context]
-           (active-scene-view app-view evaluation-context))
+    (active-scene-view app-view evaluation-context))
   (enabled? [app-view evaluation-context]
-            (and (not (camera-2d? (active-scene-view app-view evaluation-context)
-                                  evaluation-context))
-                 (not (camera-animating? app-view evaluation-context))))
+    (and (not (c/camera-2d? (view->camera (active-scene-view app-view evaluation-context))
```
vlaaad: `view->camera` needs a basis from `evaluation-context`.

vlaaad (commented):
Found some small issues, see comments. Otherwise looks good!

editor/src/clj/editor/scene.clj
vlaaad: JavaFX nowadays also has some ways to make such subscriptions easier, see `.flatMap` uses in the codebase.

editor/src/clj/editor/app_view.clj
JosephFerano: Yes, it seems like it was broken before, good catch.

editor/src/clj/editor/input.clj
JosephFerano: Whoops, left that in there.

editor/src/clj/editor/app_view.clj
JosephFerano: Ok, so this one turned out to be a symptom of a larger issue. I tried to refactor things which would then fix the issue, but as I went along I found new issues and realized that this was going to turn into a change list worthy of its own PR. Because it's existing behavior, and this PR is so close, I would like to leave this as is to avoid conflicts. I created a new branch with the refactor and once this PR is merged, I'll cherry pick that into a new branch and start a new PR. Let me know if that would be ok :)

vlaaad: Makes sense. Please create an issue so we don't forget to fix it, and we'll leave it out of the scope of this PR.

JosephFerano: So technically it wasn't a bug

```clojure

;; scene-visibility.clj
(handler/defhandler :scene.visibility.show-settings :workbench
  (run [app-view scene-visibility]
    (when-let [btn (some-> ^Tab (g/node-value app-view :active-tab)
                           .getContent
                           (.lookup "#visibility-settings-graphic")
                           .getParent)]
      (scene-visibility/show-visibility-settings! app-view btn scene-visibility)))
  (state [app-view scene-visibility evaluation-context]
    (when-let [btn (some-> ^Tab (g/node-value app-view :active-tab evaluation-context)
                           .getContent
                           (.lookup "#visibility-settings-graphic")
                           .getParent)]
   ...)))

;; grid.clj
(handler/defhandler :scene.grid.show-settings :workbench
  (run [app-view scene-visibility prefs]
    (when-some [btn (some-> (g/node-value app-view :active-tab)
                            (get-grid-settings-button))]
      (grid/show-settings! app-view btn prefs)))))

```

It seems this is being saved on the button itself, so the namespaced qualifiers were acting independently. However, there are still other issues with the scene menu buttons and the work I started still applies. I created a task #12157

vlaaad (approved):
LGTM! Ship it! Great job!

PR #11311, opened 2025-10-08
Automatically save any changes in the editor whenever it loses focus

### LINKED ISSUE #7930: Auto-saves
Work with Defold editor is extremaly frustraiting if you using extra IDE (as VScode) and/or commiting something on git.

1. Adding autosave same as in VScode, that saves all your change instantly.
2. In addition to p.1. Could be cool to use system editor (or, perfectly, let user choose an editor) to handle hyperlinks in debug console.  

Ive been thinking to make some script that will push Ctrl+S once a sec or on window select/unselect :)

Here is 2 main "fail scenario":
1. Write something in Defold editor, then switches to this file in VScode... VScode has overwrite all youve done in Defold editor.
2. This is very strange to work in two IDE's with different save system. I regularly forget to push Ctrl+S  in Defold editor and commiting only part that was autosaved by VScode.

AGulev:
Introduced a new option that automatically saves all open files in the editor whenever it loses focus.

Fix https://github.com/defold/defold/issues/7930

### DIFF
--- a/editor/resources/localization/ru.editor_localization
+++ b/editor/resources/localization/ru.editor_localization
@@ -497,6 +497,7 @@ prefs.tab.keymap = Клавиши
 
 # General tab
 prefs.label.workflow.load-external-changes-on-app-focus = Загружать изменения при фокусе
+prefs.label.workflow.save-on-app-focus-lost = Сохранять всё при потере фокуса приложения
 prefs.label.bundle.open-output-directory = Открывать папку сборки
 prefs.label.build.open-html5-build = Открывать браузер после HTML5-сборки
 prefs.label.build.texture-compression = Сжатие текстур
--- a/editor/resources/localization/ru.editor_localization
+++ b/editor/resources/localization/ru.editor_localization
@@ -497,7 +497,7 @@ prefs.tab.keymap = Клавиши
 
 # General tab
 prefs.label.workflow.load-external-changes-on-app-focus = Загружать изменения при фокусе
-prefs.label.workflow.save-on-app-focus-lost = Сохранять всё при потере фокуса приложения
+prefs.label.workflow.save-on-app-focus-lost = Сохранять при потере фокуса
 prefs.label.bundle.open-output-directory = Открывать папку сборки
 prefs.label.build.open-html5-build = Открывать браузер после HTML5-сборки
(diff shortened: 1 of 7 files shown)


editor/src/clj/editor/prefs.clj
```diff
@@ -169,6 +169,8 @@
                :properties
                {:load-external-changes-on-app-focus {:type :boolean
                                                      :default true}
+                :save-on-app-focus-lost {:type :boolean
+                                         :default true}
```
vlaaad: Shouldn't the default be false?

editor/resources/localization/ru.editor_localization
```diff
@@ -497,6 +497,7 @@ prefs.tab.keymap = Клавиши
 
 # General tab
 prefs.label.workflow.load-external-changes-on-app-focus = Загружать изменения при фокусе
+prefs.label.workflow.save-on-app-focus-lost = Сохранять всё при потере фокуса приложения
```
vlaaad: Very long label...

editor/src/clj/editor/prefs.clj
AGulev: I don't know, to be honest

As a strange side effect, it saves on every popup "open"

editor/resources/localization/ru.editor_localization
AGulev: "Сохранять всё при потере фокуса"?

vlaaad: Or "Сохранять при потере фокуса"... we can add details in the tooltip.

matgis (approved):
Code changes look good to me! ❤️

vlaaad:
Wait, didn't you say it auto-saves on every dialog window open? Did you fix it?

vlaaad:
Also, what about the quit dialog? Shouldn't we auto-save on quit and never show this dialog when the auto-save pref is on?

matgis:
Hmm, yes. I was probably too hasty approving this change.

After a window has lost focus, we need to wait at least `application-unfocused-threshold-ms` before triggering the focus-lost callback. If we're `:focused` after the short time has elapsed, we didn't actually lose focus.

And I agree about the Quit dialog - we shouldn't even show it, just save, if auto-save is enabled.

PR #11315, opened 2025-10-09
Add support for running the Android tools in bob from the commandline

britzl:
Added the ability to run`aapt2` (Android Asset Packaging Tool), `bundletool` and `jarsigner` from the commandline.

```sh
java -cp bob.jar com.dynamo.bob.tools.AndroidTools aapt2 link --proto-format ....
java -cp bob.jar com.dynamo.bob.tools.AndroidTools jarsigner -verbose -keystore debug.keystore ...
java -cp bob.jar com.dynamo.bob.tools.AndroidTools bundletool build-apks --bundle= ...
```


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/AndroidBundler.java
```diff
@@ -72,136 +75,37 @@ public class AndroidBundler implements IBundler {
...
         byte[] encoded = Files.readAllBytes(Paths.get(path));
         return new String(encoded, StandardCharsets.UTF_8);
     }
 
-    private static void createFile(String path, String content) throws IOException {
+    private void createFile(String path, String content) throws IOException {
         FileUtils.writeStringToFile(new File(path), content);
     }
 
-    private static String getJavaBinFile(Project project, String file) {
```
britzl, line 100: Moved to ToolsHelper

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/AndroidBundler.java
```diff
@@ -72,136 +75,37 @@ public class AndroidBundler implements IBundler {
...
-    private static void extractFile(File zipFile, String fileName, File outputFile) throws IOException {
-        // Wrap the file system in a try-with-resources statement
-        // to auto-close it when finished and prevent a memory leak
-        try (FileSystem fileSystem = FileSystems.newFileSystem(zipFile.toPath(), new HashMap<>())) {
-            Path fileToExtract = fileSystem.getPath(fileName);
-            Files.copy(fileToExtract, outputFile.toPath());
-        }
-    }
-
-    private static String getAapt2Name()
```
britzl, line 120: Moved to AndroidTools.java

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/AndroidBundler.java
```diff
@@ -72,136 +75,37 @@ public class AndroidBundler implements IBundler {
         platformToLibMap.put(Platform.Arm64Android, "arm64-v8a");
     }
 
-    private static void logResourceMap(Map<String, IResource> map) {
```
britzl, line 75: Not sure why everything was static...

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -756,209 +729,232 @@ else if (path.endsWith(".html")) {
         }
 
         try {
-            TimeProfiler.start("ParseCommandLine");
```
britzl, line 669: Restructured Bob startup to 1) Parse command line options 2) Setup the Project instance 3) Run Bob commands

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -756,209 +729,232 @@ else if (path.endsWith(".html")) {
...
+                    AndroidTools.aapt2(commands);
+                }
+                else if (commands.contains("bundletool")) {
+                    commands.remove("bundletool");
+                    AndroidTools.bundletool(commands);
+                }
+                else if (commands.contains("jarsigner")) {
+                    commands.remove("jarsigner");
+                    AndroidTools.jarsigner(commands);
+                }
```
britzl: Special handling of running Android tools included in bob.jar. Here we just pass along all commands on the command line. To prevent bob.jar from parsing and failing due to unknown options we use the standard way of passing `--` to end option parsing and getting the "rest" as-is in `commands`. Examples:

```
java -jar bob.jar aapt2 -- link --proto-format ....
java -jar bob.jar jarsigner -- -verbose -keystore debug.keystore ...
java -jar bob.jar bundletool -- build-apks --bundle= ...
```

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -1042,13 +1038,4 @@ public static void main(String[] args) throws IOException, CompileExceptionError
             logErrorAndExit(e);
         }
     }
-
-    private static String getOptionsValue(CommandLine cmd, char o, String defaultValue) {
```
britzl, line 957: Not needed. We can achieve exactly the same thing using `cmd.getOptionValue(o, defaultValue)`

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
AGulev: It seems like it should be fine, but maybe it worth to check if bundle from the editor still works (and two bundles one after another)

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -756,209 +729,232 @@ else if (path.endsWith(".html")) {
...
-                    break;
+                if (cmd.hasOption("debug") && cmd.hasOption("strip-executable")) {
+                    System.out.println("-d (--debug) option is deprecated and can't be set together with option --strip-executable");
+                    throw new OptionValidationException(1);
                 }
-            }
 
-            boolean verbose = cmd.hasOption('v');
+                if (cmd.hasOption("exclude-build-folder")) {
+                    // Deprecated in 1.5.1. Just a message for now.
```
AGulev: I think we need to add a ticket fully remove deprecated (or very old deprecated) arguments, the next major update

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
AGulev: hmm. Do we really want it here? 

Maybe it's easier to run needed class separatly to avoid this logic in the main flow? For example:

```
java -cp bob.jar com.dynamo.bob.AndroidTools <args>
```

```
package com.dynamo.bob;

public class AndroidTools {
    public static void main(String[] args) throws Exception {
        // Example: forward args to a specific tool
        if (args.length == 0) {
            System.out.println("Usage: AndroidTools <aapt2|bundletool|...>");
            return;
        }

        switch (args[0]) {
            case "aapt2":
                runAapt2(args);
                break;
            case "bundletool":
                runBundletool(args);
                break;
            default:
                System.out.println("Unknown command: " + args[0]);
        }
    }

    private static void runAapt2(String[] args) throws Exception {
        // call AAPT2 linking logic
        System.out.println("Running AAPT2: " + String.join(" ", args));
    }

    private static void runBundletool(String[] args) throws Exception {
        // call BundleTool building logic
        System.out.println("Running BundleTool: " + String.join(" ", args));
    }
}
```

AGulev (dismissed):
Looks good overall, but please check my comments

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
britzl: I was briefly contemplating this, then discarded it, but now that you bring it up then maybe it is actually a better solution.... I will in this case revert all the changes to Bob and Project as that was mostly drive-by fixing

AGulev (approved):
👍

PR #11301, opened 2025-10-06
Fix Tile Source Collision Group colors in Outline

matgis:
Fixes #11290

### Technical changes
* Added some more info to the maps returned by the `dev/javafx-tree` function.

PR #11324, opened 2025-10-10
Fix lingering label colors in Outline

matgis:
Fix lingering label colors in Outline introduced by #11301.


vlaaad (approved):
Looks good!

PR #11325, opened 2025-10-10
Fix env patch for windows

AGulev:
Fix env patch on windows


DarkRavenStar (approved):
Looks good on my side. I have tested on my Windows environment and its works

PR #11327, opened 2025-10-11
Bypass the quit confirmation dialog when auto-save is turned on

AGulev:
Fix https://github.com/defold/defold/issues/11323


vlaaad (approved):
LGTM!

PR #11332, opened 2025-10-13
Issue 11298 resolve sound deadlock on Windows

### LINKED ISSUE #11298: Game freezes when audio device is disconnected during playback
When playing sound on Windows, if the active audio output device is disconnected (for example, unplugging wired headphones or Bluetooth headphones disconnecting), the game completely freezes. Windows automatically switches to another audio device, but the Defold game becomes unresponsive, i.e. freezes/hangs.

**To Reproduce (REQUIRED)**
1. Run a Defold game that continiously plays any sound like music.
2. While the sound is playing, disconnect the current audio device (e.g., unplug headphones or power off Bluetooth headphones).
3. Windows will automatically switch to another available audio device.
4. Observe that the game hangs and stops responding.

The game should not freeze when the audio device is disconnected. Instead, it should continue running normally and switch playback to the newly active audio device.

  - Version: 1.11.0, 1.11.1 BETA

**Platforms (REQUIRED):**
 - Platforms: Windows 11

Use the project from this comment - https://github.com/defold/defold/issues/9445#issuecomment-2612303987

Related:
https://github.com/defold/defold/issues/10258

If you ask why this wasn’t discovered in the previous issue or earlier — for example, I started using wireless headphones and happened to notice it by accident when the battery ran out 😄

deurell:
Resolve sound deadlock on Windows when disconnecting sound device. 

Fixes https://github.com/defold/defold/issues/11298

Technical changes
Fixed the WASAPI backend so a disappearing sound device no longer deadlocks the engine.

### DIFF
--- a/engine/sound/src/devices/device_wasapi.cpp
+++ b/engine/sound/src/devices/device_wasapi.cpp
@@ -499,6 +499,15 @@ namespace dmDeviceWasapi
         SoundDevice* device = (SoundDevice*) _device;
         if (device->m_DeviceInvalidated)
         {
+            if (device->m_AudioClient)
+            {
+                HRESULT hr = device->m_AudioClient->Stop();
+                if (FAILED(hr))
+                {
+                    dmLogError("Failed to stop audio client");
+                    CheckAndPrintError(hr);
+                }
+            }
             return;
         }
         if (device->m_AudioClient)
--- a/engine/sound/src/devices/device_wasapi.cpp
+++ b/engine/sound/src/devices/device_wasapi.cpp
@@ -288,7 +288,6 @@ namespace dmDeviceWasapi
 
         if (device->m_DeviceInvalidated)
         {
-            dmTime::Sleep(1000);
             return 0;
         }
 
@@ -343,7 +342,6 @@ namespace dmDeviceWasapi
 
         if (device->m_DeviceInvalidated)
         {
-            dmTime::Sleep(1000);
             return dmSound::RESULT_INIT_ERROR;
         }
 
--- a/engine/sound/src/devices/device_wasapi.cpp
+++ b/engine/sound/src/devices/device_wasapi.cpp
@@ -40,10 +40,12 @@ namespace dmDeviceWasapi
 
         uint32_t            m_Format;
         uint32_t            m_FrameCount;
+        bool                m_DeviceInvalidated;
 
         SoundDevice()
         {
             memset(this, 0, sizeof(*this));
+            m_DeviceInvalidated = false;
         }
     };
 
@@ -53,6 +55,20 @@ namespace dmDeviceWasapi
         {
             dmLogError("WASAPI error 0x%X: %s", hr, "");
         }
+    }
+
+    static void MarkDeviceInvalidated(SoundDevice* device, const char* context, HRESULT hr)
+    {
+        if (!device)
+            return;
+
+        if (!device->m_DeviceInvalidated)
+        {
+            dmLogError("WASAPI device lost in %s (hr=0x%08X)", context, hr);
+            CheckAndPrintError(hr);
+        }
+
+        device->m_DeviceInvalidated = true;
     }
 
     static void DeleteDevice(SoundDevice* device)
@@ -253,6 +269,7 @@ namespace dmDeviceWasapi
             return dmSound::RESULT_INIT_ERROR;
         }
 
+        device->m_DeviceInvalidated = false;
         *outdevice = device;
         return dmSound::RESULT_OK;
     }
@@ -269,6 +286,12 @@ namespace dmDeviceWasapi
         assert(_device);
         SoundDevice* device = (SoundDevice*) _device;
 
+        if (device->m_DeviceInvalidated)
+        {
+            dmTime::Sleep(1000);
+            return 0;
+        }
+
         HRESULT hr = WaitForSingleObject(device->m_ClientBufferEvent, 250);
         if (WAIT_OBJECT_0 != hr)
         {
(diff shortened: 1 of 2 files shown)


engine/sound/src/devices/device_wasapi.cpp
```diff
@@ -36,24 +36,40 @@ namespace dmDeviceWasapi
...
-        if (FAILED(hr))
-        {
-            dmLogError("WASAPI error 0x%X: %s", hr, "");
-        }
-    }
+        HANDLE              m_ClientBufferEvent;
+
+        uint32_t            m_Format;
+        uint32_t            m_FrameCount;
+        bool                m_DeviceInvalidated;
```
deurell, line 43: Add m_DeviceInvalidated and MarkDeviceInvalidated() to log first device invalidation and short circuit later calls to stop hammering WASAPI.

engine/sound/src/devices/device_wasapi.cpp
```diff
@@ -245,92 +261,131 @@ namespace dmDeviceWasapi
...
-        uint32_t frames_available = buffer_size - buffer_pos;
-
+    {
+        assert(_device);
+        SoundDevice* device = (SoundDevice*) _device;
+        const int16_t* samples = (const int16_t*)_samples;
+
+        if (device->m_DeviceInvalidated)
+        {
+            return dmSound::RESULT_INIT_ERROR;
```
deurell, line 345: Return RESULT_INIT_ERROR on AUDCLNT_E_DEVICE_INVALIDATED so UpdateInternal() marks the device stopped, m_DeviceInvalidated = true will short circuit future calls.

engine/sound/src/sound.cpp
```diff
@@ -1758,44 +1758,74 @@ namespace dmSound
             sound->m_IsDeviceStarted = true;
         }
 
-        DM_MUTEX_OPTIONAL_SCOPED_LOCK(g_SoundSystem->m_Mutex);
-
         uint32_t free_slots = sound->m_DeviceType->m_FreeBufferSlots(sound->m_Device);
```
deurell, line 1761: Query the device before locking the sound mutex

deurell:
Removed the recovery logic for lost device and user changed default device to make the PR cleaner, only keeping the dead lock fix.

engine/sound/src/sound.cpp
```diff
@@ -1758,44 +1758,74 @@ namespace dmSound
...
             }
 
-            sound->m_FrameCount = frame_count;
+            bool skip_queue = false;
+            uint32_t buffer_index = 0;
 
-            if (frame_count < SOUND_MAX_HISTORY)
-                continue;
+            {
+                DM_MUTEX_OPTIONAL_SCOPED_LOCK(g_SoundSystem->m_Mutex);
```
JCash, line 1787: I guess this could be moved to the else-statement?

engine/sound/src/sound.cpp
```diff
@@ -1758,44 +1758,74 @@ namespace dmSound
...
+
+                sound->m_FrameCount = frame_count;
+
+                if (frame_count < SOUND_MAX_HISTORY)
+                {
+                    skip_queue = true;
```
JCash: What is the benefit of this, over simply calling `continue` directly?

engine/sound/src/sound.cpp
```diff
@@ -1758,44 +1758,74 @@ namespace dmSound
...
+                }
+                else
+                {
+                    MixContext mix_context(current_buffer, total_buffers, frame_count);
+                    MixInstances(&mix_context);
+
+                    Master(&mix_context);
+
+                    buffer_index = sound->m_NextOutBuffer;
+                    sound->m_NextOutBuffer = (sound->m_NextOutBuffer + 1) % sound->m_OutBufferCount;
```
JCash: I think these should be moved to around line 1820, where the buffer is actually being used.
It's not needed before, and the conditions will be the same 🤔 

Apart from the mutex lock being moved from outside of the loop, I don't understand the benefits of the new if-else logic here.

engine/sound/src/devices/device_wasapi.cpp
```diff
@@ -391,49 +455,117 @@ namespace dmDeviceWasapi
...
-            }
-        }
-    }
+    {
+        assert(_device);
+        SoundDevice* device = (SoundDevice*) _device;
+        if (device->m_DeviceInvalidated)
+        {
+            return;
+        }
```
JCash, line 464: This function is part of the cleanup, i.e. when the game is shutting down.
While the device might be invalidated, isn't there anything that needs to be cleaned up?

JCash (changes requested):
I have some design questions in the sound.cpp
also, I think we're missing some device cleanup in the device Stop() function

engine/sound/src/devices/device_wasapi.cpp
deurell: I think this is regarding DeviceWasapiStop. The wasapi buffer is lost if we get AUDCLNT_E_DEVICE_INVALIDATED so not able to call ReleaseBuffer but added a call to stop the client as cleanup.

engine/sound/src/devices/device_wasapi.cpp
```diff
@@ -391,49 +455,126 @@ namespace dmDeviceWasapi
...
+
+    static void DeviceWasapiStop(dmSound::HDevice _device)
+    {
+        assert(_device);
+        SoundDevice* device = (SoundDevice*) _device;
+        if (device->m_DeviceInvalidated)
+        {
+            if (device->m_AudioClient)
+            {
+                HRESULT hr = device->m_AudioClient->Stop();
```
deurell, line 504: If device is invalidated we can't release the buffer without getting a AUDCLNT_E_DEVICE_INVALIDATED, but we can call stop on the audio client.

engine/sound/src/sound.cpp
deurell: removing and calling continue

engine/sound/src/sound.cpp
deurell: The thought I had was moving the m_NextOutBuffer bump inside the mutex so we can snapshot it before releasing the lock, then drop the lock before calling Queue to avoid the deadlock.

engine/sound/src/devices/device_wasapi.cpp
JCash: As long as we follow the recommended shut down steps, then it's good.

JCash (approved):

AGulev:
Please don’t forget to add the project to the task you’re closing so the fix ends up in the release notes (I added it for this one myself).

PR #11333, opened 2025-10-13
Fix autosave when a popup opened

AGulev:
Fix https://github.com/defold/defold/issues/11322#event-20240061211

### DIFF
--- a/editor/src/clj/editor/ui.clj
+++ b/editor/src/clj/editor/ui.clj
@@ -87,12 +87,8 @@
   (let [delay (/ (double application-unfocused-threshold-ms) 1000.0)
         timer (->future delay
                         (fn []
-                          (let [{stored-timestamp :timestamp} (get @application-unfocused-tasks key)
-                                current @focus-state]
-                            (when (and (= stored-timestamp timestamp)
-                                       current
-                                       (not (:focused current))
-                                       (= (:t current) timestamp))
+                          (let [{stored-timestamp :timestamp} (get @application-unfocused-tasks key)]
+                            (when (= stored-timestamp timestamp)
                               (swap! application-unfocused-tasks dissoc key)
                               (apply application-unfocused! args)))))]
     (swap! application-unfocused-tasks assoc key {:timer timer :timestamp timestamp})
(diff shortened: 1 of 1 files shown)


vlaaad (changes requested):
If you quickly Cmd+Tab to another app, then back, then another app, then back, then another app, you'll have the callback executed twice, but you only want it executed once. The code should hold a reference to a launched future and cancel it when the app becomes focused.

AGulev:
Nice one, I'll fix it, thank you!

editor/src/clj/editor/ui.clj
```diff
@@ -72,6 +72,31 @@
...
+  (cancel-application-unfocused-task! key)
+  (let [delay (/ (double application-unfocused-threshold-ms) 1000.0)
+        timer (->future delay
+                        (fn []
+                          (let [{stored-timestamp :timestamp} (get @application-unfocused-tasks key)
+                                current @focus-state]
+                            (when (and (= stored-timestamp timestamp)
+                                       current
+                                       (not (:focused current))
+                                       (= (:t current) timestamp))
```
vlaaad: Are these checks necessary? I've got the impression the timer is already getting stopped when the app becomes focused.

vlaaad (approved):
I'd prefer if we avoided introducing global mutable state when possible, but I don't see any obvious way to do it. Let's merge it.

editor/src/clj/editor/ui.clj
AGulev: If user comes back to the window, the same frame callback fires ...

vlaaad: Hmm, maybe I don't understand the code flow? I thought that when the user comes back to the window, we execute `(cancel-application-unfocused-task! key)` (lines 156-157), so this callback is not invoked?

vlaaad (changes requested):
After some more thought, I'd like you to remove unnecessary checks on lines 92-95 (and verify that they are indeed unnecessary :D). Both focus events and `->future` timer work on the UI thread, so it should not be possible for them to run in parallel.

vlaaad (approved):
LGTM!

PR #11343, opened 2025-10-15
Localize Outline

vlaaad:
Related to #5453

### DIFF
--- a/editor/resources/localization/ru.editor_localization
+++ b/editor/resources/localization/ru.editor_localization
@@ -1170,6 +1170,30 @@ property.tile-source.tile-margin = Отступ тайла
 property.tile-source.tile-spacing = Отступ между тайлами
 property.tile-source.tile-width = Ширина тайла
 
+# Outline
+
+outline.atlas.no-image = <Нет изображения>
+outline.atlas = Атлас
+outline.camera = Камера
+outline.collection = Коллекция
+outline.collection-proxy = Прокси коллекции
+outline.collision-object = Объект коллизий
+outline.game-object = Игровой объект
+outline.gui.fonts = Шрифты
+outline.gui.layouts = Макеты
+outline.gui.layers = Слои
+outline.gui.nodes = Элементы
+outline.gui.materials = Материалы
+outline.gui.textures = Текстуры
+outline.non-editable-collection = Нередактируемая коллекция
+outline.non-editable-game-object = Нередактируемый игровой объект
+outline.particlefx = ParticleFX
+outline.sound = Звук
+outline.tile-map = Тайловая карта
+outline.tile-source = Набор тайлов
+outline.unknown = Неизвестно
+outline.unnamed-collision-shape = <Безымянная форма: {shape}>
+
 # Resources
 
 resource.view.scene = Сцена
--- a/editor/src/clj/editor/collection.clj
+++ b/editor/src/clj/editor/collection.clj
@@ -493,7 +493,7 @@
                  (fn [localization-state _]
                    (into (localization/natural-sort-by-label localization-state coll-outlines)
                          (localization/natural-sort-by-label localization-state go-outlines)))
-                 (-> coll-outlines (or []) (into go-outlines)))
+                 (into (or coll-outlines []) go-outlines))
      :child-reqs [{:node-type ReferencedGOInstanceNode
                    :tx-attach-fn outline-tx-attach-coll-referenced-go}
                   {:node-type EmbeddedGOInstanceNode
--- a/editor/src/clj/editor/collection.clj
+++ b/editor/src/clj/editor/collection.clj
@@ -190,7 +190,11 @@
        :node-outline-key id
        :label id
        :icon (or (not-empty (:icon source-outline)) game-object-common/game-object-icon)
-       :children (into (outline/natural-sort child-outlines) (:children source-outline))
+       :children (localization/annotate-as-sorted
+                   (fn [localization-state _]
+                     (into (localization/natural-sort-by-label localization-state child-outlines)
+                           (localization/sort-if-annotated localization-state (:children source-outline))))
+                   (into child-outlines (:children source-outline)))
        :child-reqs [{:node-type ReferencedGOInstanceNode
                      :tx-attach-fn tx-attach-go-referenced-go}
                     {:node-type EmbeddedGOInstanceNode
@@ -480,13 +484,16 @@
     (tx-attach-coll-coll self-id child-id)))
 
 (g/defnk produce-coll-outline [_node-id child-outlines]
-  (let [[go-outlines coll-outlines] (let [outlines (group-by #(g/node-instance? CollectionInstanceNode (:node-id %)) child-outlines)]
-                                      [(get outlines false) (get outlines true)])]
+  (let [{go-outlines false coll-outlines true} (group-by #(g/node-instance? CollectionInstanceNode (:node-id %)) child-outlines)]
     {:node-id _node-id
      :node-outline-key "Collection"
-     :label "Collection"
+     :label (localization/message "outline.collection")
      :icon collection-common/collection-icon
-     :children (into (outline/natural-sort coll-outlines) (outline/natural-sort go-outlines))
+     :children (localization/annotate-as-sorted
+                 (fn [localization-state _]
+                   (into (localization/natural-sort-by-label localization-state coll-outlines)
+                         (localization/natural-sort-by-label localization-state go-outlines)))
+                 (-> coll-outlines (or []) (into go-outlines)))
      :child-reqs [{:node-type ReferencedGOInstanceNode
                    :tx-attach-fn outline-tx-attach-coll-referenced-go}
                   {:node-type EmbeddedGOInstanceNode
@@ -499,7 +506,7 @@
 
   (property name g/Str
             (dynamic error (g/fnk [_node-id name]
-                                 (validation/prop-error :warning _node-id :id validation/prop-contains-prohibited-characters? name "Name"))))
+                             (validation/prop-error :warning _node-id :id validation/prop-contains-prohibited-characters? name "Name"))))
 
   ;; This property is legacy and purposefully hidden
   ;; The feature is only useful for uniform scaling, we use non-uniform now
--- a/editor/src/clj/editor/gui.clj
+++ b/editor/src/clj/editor/gui.clj
@@ -2665,6 +2665,8 @@
                      [:resource :particlefx-resource]
                      [:build-targets :dep-build-targets]
                      [:scene :particlefx-scene])))
+            (dynamic label (properties/label-dynamic :gui :particlefx))
+            (dynamic tooltip (properties/tooltip-dynamic :gui :particlefx))
             (dynamic error (g/fnk [_node-id particlefx]
                                   (prop-resource-error _node-id :particlefx particlefx "Particle FX")))
             (dynamic edit-type (g/constantly
@@ -2742,9 +2744,11 @@
   (input child-indices NodeIndex :array)
   (output child-scenes g/Any (g/fnk [child-scenes] (vec (sort-by (comp :child-index :renderable) child-scenes))))
   (output node-outline outline/OutlineData :cached
-          (gen-outline-fnk "Nodes" nil 0 true (mapv (fn [type-info] {:node-type (:node-cls type-info)
-                                                                     :tx-attach-fn gui-node-attach-fn})
-                                                    (get-registered-node-type-infos))))
+          (gen-outline-fnk (localization/message "outline.gui.nodes") nil 0 true
+                           (mapv (fn [type-info]
+                                   {:node-type (:node-cls type-info)
+                                    :tx-attach-fn gui-node-attach-fn})
+                                 (get-registered-node-type-infos))))
 
   (output scene g/Any (g/fnk [_node-id child-scenes]
                         {:node-id _node-id
@@ -2837,8 +2841,9 @@
   (output texture-page-counts g/Any :cached (g/fnk [texture-page-counts]
                                               (into {} cat texture-page-counts)))
   (output node-outline outline/OutlineData :cached
-          (gen-outline-fnk "Textures" "Textures" 1 false [{:node-type TextureNode
-                                                           :tx-attach-fn (gen-outline-node-tx-attach-fn attach-texture)}]))
+          (gen-outline-fnk (localization/message "outline.gui.textures") "Textures" 1 false
+                           [{:node-type TextureNode
+                             :tx-attach-fn (gen-outline-node-tx-attach-fn attach-texture)}]))
   (output add-handler-info g/Any
           (g/fnk [_node-id]
                  [_node-id "Textures..." texture-icon add-textures-handler {}])))
@@ -2874,8 +2879,9 @@
   (input build-errors g/Any :array)
   (output build-errors g/Any (gu/passthrough build-errors))
   (output node-outline outline/OutlineData :cached
-          (gen-outline-fnk "Materials" "Materials" 1 false [{:node-type MaterialNode
-                                                             :tx-attach-fn (gen-outline-node-tx-attach-fn attach-material)}]))
+          (gen-outline-fnk (localization/message "outline.gui.materials") "Materials" 1 false
+                           [{:node-type MaterialNode
+                             :tx-attach-fn (gen-outline-node-tx-attach-fn attach-material)}]))
   (output add-handler-info g/Any
           (g/fnk [_node-id]
             [_node-id "Materials..." material-icon add-materials-handler {}])))
@@ -2916,8 +2922,9 @@
   (input build-errors g/Any :array)
   (output build-errors g/Any (gu/passthrough build-errors))
   (output node-outline outline/OutlineData :cached
-          (gen-outline-fnk "Fonts" "Fonts" 2 false [{:node-type FontNode
-                                                     :tx-attach-fn (gen-outline-node-tx-attach-fn attach-font)}]))
+          (gen-outline-fnk (localization/message "outline.gui.fonts") "Fonts" 2 false
+                           [{:node-type FontNode
+                             :tx-attach-fn (gen-outline-node-tx-attach-fn attach-font)}]))
   (output add-handler-info g/Any
           (g/fnk [_node-id]
                  [_node-id "Fonts..." font-icon add-fonts-handler {}])))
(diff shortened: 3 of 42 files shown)


editor/src/clj/editor/collection.clj
```diff
@@ -190,7 +190,11 @@
        :node-outline-key id
        :label id
        :icon (or (not-empty (:icon source-outline)) game-object-common/game-object-icon)
-       :children (into (outline/natural-sort child-outlines) (:children source-outline))
+       :children (localization/annotate-as-sorted
```
vlaaad, line 193: Localization affects sorting, and there are several solutions that can express it:
1. Wire the localization state into every node that produces a sorted outline, and sort in the output. I thought this would bloat the connections in large projects unnecessarily
2. Use a special localization message that takes items and returns sorted items when formatted. I thought this would be a weak abstraction for localization, since all the other message patterns return strings, not items. Additionally, if `:children` becomes a message pattern, it breaks all the code that expects children to be a vector of outline items.
3. The solution I used: annotate the `:children` vector with metadata that defines how they are sorted when presented to the user in UI. This avoids the graph connection bloat and keeps children as a vector of outlines.

editor/src/clj/editor/handler.clj
```diff
@@ -403,12 +403,13 @@
...
-         (e/filter #(= command (:command %)))
-         vec
-         coll/not-empty)))
+    (when-let [flat-opts (->> opts
+                              (e/mapcat flatten-menu-item-tree)
+                              (e/remove #(= :separator (:label %)))
+                              (e/filter #(= command (:command %)))
+                              vec
+                              coll/not-empty)]
+      (with-meta flat-opts (meta opts)))))
```
vlaaad, line 412: This preserves the designated sort function for options if it exists (e.g., when we show a dialog with a list of options for adding a component type to a game object or collision object)

editor/src/clj/editor/localization.clj
```diff
@@ -318,8 +319,9 @@
   ([object ^Localization localization pattern]
    {:pre [(localization? localization)]}
    (apply-localization object (impl-format @localization pattern))
-   (when (message-pattern? pattern)
-     (send-without-thread-binding-reset javafx-executor (.-agent localization) impl-localize! object pattern))
+   (if (message-pattern? pattern)
+     (send-without-thread-binding-reset javafx-executor (.-agent localization) impl-localize! object pattern)
+     (send-without-thread-binding-reset javafx-executor (.-agent localization) impl-unlocalize! object))
```
vlaaad, line 324: The previous code was wrong, actually. The thing is, we can `localize!` either with simple values like strings, or localization-aware values like message patterns. We added a listener only when we localized with message patterns. But in the outline view, the tree cells may be repeatedly localized with both strings and message patterns. So it's possible that when we call `localize!` with a string (which will apply the localization once), the same object was already localized with a message pattern, so changing the language would refresh the label. So, we need to remove a listener if it exists.

editor/src/clj/editor/outline.clj
```diff
@@ -328,9 +332,6 @@
   (let [names (id/resolve-all (map resource/base-name resources) taken-ids)]
     (map vector names resources)))
 
-(defn natural-sort [items]
-  (->> items (sort-by :label util/natural-order) vec))
```
vlaaad, line 332: Lots of pain because we can't use this small function anymore...

editor/src/clj/editor/outline_view.clj
```diff
@@ -40,159 +49,234 @@
...
+    (let [parents (map outline/parent item-iterators)]
+      (when (apply = parents)
+        (first parents)))))
+
+(defn- set-paste-parent! [root-its]
+  (alter-var-root #'*paste-into-parent* (constantly (some-> (single-parent-it root-its) outline/value :node-id))))
+
+(defn- update-selection! [app-view swap-state active-resource-node tree-items]
+  (set-paste-parent! nil)
+  ;; TODO - handle selection order
```
vlaaad, line 158: This isn't my todo, I moved the function up from below.

editor/src/clj/editor/outline_view.clj
```diff
@@ -40,159 +49,234 @@
...
+(defn decorate-outline [outline hidden-node-outline-key-paths outline-name-paths localization-state selection]
+  (letfn [(decorate [{:keys [node-id children] :as outline} node-id-path node-outline-key-path parent-reference?]
+            (let [node-id-path (conj node-id-path node-id)
+                  node-outline-key-path (if (empty? node-outline-key-path)
+                                          [node-id]
+                                          (if-some [node-outline-key (:node-outline-key outline)]
+                                            (conj node-outline-key-path node-outline-key)
+                                            node-outline-key-path))
+                  data (->> children
+                            (localization/sort-if-annotated localization-state)
```
vlaaad, line 81: I rewrote the outline tree view to be managed using cljfx. Main reason: sorting. Naive approach that I initially tried — using `localize!` with a custom message pattern that changes the order of tree view item children — was very inefficient, since there was a listener attached to every tree item that had children, and switching between a few collections, as well as switching the language a couple of times, was incredibly slow. When I rewrote it to cljfx, there is a single localization state listener in the outline view, and that's it. Works much better.

editor/src/clj/editor/outline_view.clj
```diff
@@ -40,159 +49,234 @@
...
+                            (persistent!
+                              (reduce-kv
+                                (fn [acc new-node-id-path _]
+                                  (assoc! acc new-node-id-path (boolean (or (old-node-id-path->expanded new-node-id-path)
+                                                                            (auto-expanded new-node-id-path)))))
+                                (transient old-node-id-path->expanded)
+                                new-node-id-path->expanded)))
+                          old-active-node-id->node-id-path->expanded
+                          active-node-id->node-id-path->expanded))))]
+    ret))
```
vlaaad, line 217: Kinda annoying... previously, the expanded state was managed implicitly: there was a separate tree view per active resource type, and changing selection just mutated the expanded state of new selection parents, while previously expanded tree items kept being expanded. Now that it's reactive, I have to manage it myself...

editor/src/clj/editor/outline_view.clj
```diff
@@ -207,22 +291,21 @@
...
   (input active-resource-node g/NodeID :substitute nil)
   (input open-resource-nodes g/Any :substitute [])
   (input selection g/Any :substitute [])
   (input hidden-node-outline-key-paths types/NodeOutlineKeyPaths)
   (input outline-name-paths scene-visibility/OutlineNamePaths)
 
-  (output root TreeItem :cached produce-tree-root)
   (output tree-view TreeView :cached update-tree-view)
   (output tree-selection g/Any :cached (g/fnk [tree-view] (ui/selection tree-view)))
-  (output tree-selection-root-its g/Any :cached (g/fnk [tree-view] (mapv ->iterator (ui/selection-root-items tree-view (comp :node-id-path item->value) (comp :node-id item->value)))))
```
vlaaad, line 221: I also got rid of ->iterator,  which was a useless wrapper around a tree item that implemented a protocol that the tree item can implement directly.

editor/test/integration/collection_test.clj
```diff
@@ -39,7 +40,7 @@
         ;; Two game objects under the collection
         (is (= 2 (count (:children outline))))
         ;; One component and game object under the game object
-        (is (= 2 (count (:children (second (:children outline)))))))))
+        (is (= 2 (count (:children (first (:children outline)))))))))
```
vlaaad, line 43: Now that outline items are not sorted in the graph, but only in the UI, a lot of tests needed to be fixed...

editor/test/integration/script_properties_test.clj
```diff
@@ -162,9 +162,13 @@
...
-    ;; [2 0] type faulty instance script, bad collection level override, fallback to script setting = 1.0
-    ;; [3 0] type faulty instance script, proper collection-level override = 3.0
-    (doseq [[resource path-vals] [["/collection/type_faulty_props.collection" [[[0 0] false 2.0] [[1 0] false 1.0] [[2 0] false 1.0]] [[3 0] true 3.0]]]
+    ;; [2 0] type faulty instance script, proper collection-level override = 3.0
+    ;; [3 0] type faulty instance script, bad collection level override, fallback to script setting = 1.0
+    (doseq [[resource path-vals] [["/collection/type_faulty_props.collection"
+                                   [[[0 0] false 2.0]
+                                    [[1 0] false 1.0]
+                                    [[2 0] true 3.0]
+                                    [[3 0] false 1.0]]]]
```
vlaaad, line 171: If you count brackets in the old code, you'll notice that `[[3 0] true 3.0]]` was never tested since it's outside the `path-vals` vector.

editor/resources/localization/ru.editor_localization
```diff
@@ -1170,6 +1170,30 @@ property.tile-source.tile-margin = Отступ тайла
...
+outline.game-object = Игровой объект
+outline.gui.fonts = Шрифты
+outline.gui.layouts = Макеты
+outline.gui.layers = Слои
+outline.gui.nodes = Элементы
+outline.gui.materials = Материалы
+outline.gui.textures = Текстуры
+outline.non-editable-collection = Нередактируемая коллекция
+outline.non-editable-game-object = Нередактируемый игровой объект
+outline.particlefx = ParticleFX
```
matgis, line 1190: Is there no russian word for Particle FX?

editor/src/clj/editor/collection.clj
```diff
@@ -480,13 +484,16 @@
...
      :node-outline-key "Collection"
-     :label "Collection"
+     :label (localization/message "outline.collection")
      :icon collection-common/collection-icon
-     :children (into (outline/natural-sort coll-outlines) (outline/natural-sort go-outlines))
+     :children (localization/annotate-as-sorted
+                 (fn [localization-state _]
+                   (into (localization/natural-sort-by-label localization-state coll-outlines)
+                         (localization/natural-sort-by-label localization-state go-outlines)))
+                 (-> coll-outlines (or []) (into go-outlines)))
```
matgis: Might be clearer if you lose the threading macro here?

editor/src/clj/editor/gui.clj
```diff
@@ -2665,6 +2665,8 @@
                      [:resource :particlefx-resource]
                      [:build-targets :dep-build-targets]
                      [:scene :particlefx-scene])))
+            (dynamic label (properties/label-dynamic :gui :particlefx))
+            (dynamic tooltip (properties/tooltip-dynamic :gui :particlefx))
```
matgis, line 2669: It seems the property label fed to `prop-resource-error` below should also be localized? Or maybe we're tackling those later?

editor/src/clj/editor/handler.clj
matgis: Note that both `coll/transform` or `into` with `(empty opts)` will preserve metadata as well if we want to shorten it.

editor/src/clj/editor/localization.clj
```diff
@@ -613,10 +615,36 @@
...
+    items    vector of items to annotate"
+  [f items]
+  {:pre [(vector? items)]}
+  (vary-meta items assoc ::sort f))
+
+(defn natural-sort-by-label
+  "Localization-aware function that sorts the items on their :label vals"
+  [localization-state items]
+  (->> items
+       (mapv #(coll/pair (localization-state (:label %)) %))
```
matgis, line 633: Should we `refer` `pair`?

editor/src/clj/editor/outline_view.clj
```diff
@@ -40,159 +49,234 @@
...
+                                          [node-id]
+                                          (if-some [node-outline-key (:node-outline-key outline)]
+                                            (conj node-outline-key-path node-outline-key)
+                                            node-outline-key-path))
+                  data (->> children
+                            (localization/sort-if-annotated localization-state)
+                            (mapv #(decorate %
+                                             node-id-path
+                                             node-outline-key-path
+                                             (or parent-reference? (:outline-reference? outline)))))
```
matgis, line 85: Shouldn't this be `(:outline-reference? %)`?

editor/src/clj/editor/outline_view.clj
```diff
@@ -40,159 +49,234 @@
...
+               :child-error (or (:child-error outline) (:outline-error? outline))
+               :child-overridden (or (:child-overridden outline) (:outline-overridden? outline))}))]
+    (let [outline (:outline (decorate outline [] [] (:outline-reference? outline)))
+          node-id->node-id-paths (util/group-into {} [] key val
+                                                  (coll/transfer [outline] :eduction
+                                                    (coll/tree-xf :children :children)
+                                                    (map (coll/pair-fn :node-id :node-id-path))))
+          default-selection (coll/transfer selection {}
+                              (keep #(when-let [node-id-paths (node-id->node-id-paths %)]
+                                       (coll/pair % #{(first node-id-paths)}))))
```
matgis, line 108: refer pair?

editor/src/clj/editor/outline_view.clj
```diff
@@ -40,159 +49,234 @@
...
+                                                :hidden
+                                                :visible))]
+              {:outline outline
+               :child-error (or (:child-error outline) (:outline-error? outline))
+               :child-overridden (or (:child-overridden outline) (:outline-overridden? outline))}))]
+    (let [outline (:outline (decorate outline [] [] (:outline-reference? outline)))
+          node-id->node-id-paths (util/group-into {} [] key val
+                                                  (coll/transfer [outline] :eduction
+                                                    (coll/tree-xf :children :children)
+                                                    (map (coll/pair-fn :node-id :node-id-path))))
```
matgis, line 105: Could use `coll/pair-map-by` here?

editor/src/clj/editor/outline_view.clj
```diff
@@ -40,159 +49,234 @@
...
+(def ^:private ext-with-tree-view-props
+  (fx/make-ext-with-props
+    (assoc fx.tree-view/props
+      :selected-node-id-paths
+      (fx.prop/make
+        (fx.mutator/setter
+          (fn set-selected-node-id-paths [^TreeView tree-view new-selected-node-id-paths]
+            (let [selection-model (.getSelectionModel tree-view)
+                  old-selected-node-id-paths (coll/transfer (.getSelectedItems selection-model) #{}
+                                               (keep #(:node-id-path (.getValue ^TreeItem %))))]
```
matgis, line 132: Not sure if it matters here, but the order of the selected items does have an effect on which item the Scene View gizmo is placed on, so using a set might not be entirely correct. I guess it used sets before too? 🤔

editor/src/clj/editor/outline_view.clj
```diff
@@ -40,159 +49,234 @@
...
+   :value outline
+   :expanded (node-id-path->expanded node-id-path)
+   :on-expanded-changed #(swap-state update :active-node-id->node-id-path->expanded
+                                     update active-resource-node
+                                     assoc node-id-path %)
+   :children (->> outline
+                  :children
+                  (mapv #(outline->tree-item % node-id-path->expanded swap-state active-resource-node)))})
+
+(defn- reset-tree-view-state [old {:keys [active-resource-node selection open-resource-nodes active-node-id->selected-node-id->node-id-paths active-node-id->node-id-path->expanded] :as new}]
```
matgis: `new` appears unused? Rename to `_new` or remove?

editor/src/clj/editor/outline_view.clj
matgis: Still, it feels better to not rely on that internal state and simply project The Truth onto the view. 👍

editor/src/clj/editor/tile_source.clj
```diff
@@ -343,13 +343,16 @@
...
      :node-outline-key "Tile Source"
-     :label "Tile Source"
+     :label (localization/message "outline.tile-source")
      :icon tile-source-icon
-     :children (into (outline/natural-sort coll-outlines) (outline/natural-sort anim-outlines))
+     :children (localization/annotate-as-sorted
+                 (fn [localization-state _]
+                   (into (localization/natural-sort-by-label localization-state coll-outlines)
+                         (localization/natural-sort-by-label localization-state anim-outlines)))
+                 (-> coll-outlines (or []) (into anim-outlines)))
```
matgis, line 355: Simply be removing the threading macro?

editor/test/integration/script_properties_test.clj
matgis: Nice find!

matgis (dismissed):
Seems to work! I noticed that the selection highlights a different node after the nodes are re-sorted when I toggled languages, but maybe that isn't a big deal?

editor/resources/localization/ru.editor_localization
vlaaad: We could use "Система частиц", i.e., "Particle System", but we use a different, peculiar term, so I decided not to translate it.

editor/src/clj/editor/gui.clj
vlaaad: We are doing this later, it would be too much to do everything everywhere all at once :)

editor/src/clj/editor/localization.clj
vlaaad: I'd prefer if we didn't `refer` things that can be required/aliased instead. Perhaps with the exception of `clojure.test`.

editor/src/clj/editor/outline_view.clj
vlaaad: Doesn't look like it should, see the previous line of code that recursively called `decorate`:
https://github.com/defold/defold/pull/11343/files#diff-4c381bb725e8f7d75b6700638fe4d057dbaa9372f3d8cdf7f5bb6c94f104b518L137

editor/src/clj/editor/outline_view.clj
vlaaad: Hmm, doesn't look like I can do that. 2-arity checks if the second arg is `fn?` to differentiate between a coll and a transducer behaviors, but keywords are not `fn`s.

editor/src/clj/editor/outline_view.clj
vlaaad: Yes, it did use a set before during selection sync, see:
https://github.com/defold/defold/pull/11343/files#diff-4c381bb725e8f7d75b6700638fe4d057dbaa9372f3d8cdf7f5bb6c94f104b518L111-L125

The behavior of selection sync is the same: for a set of selected nodes, check every displayed tree item and select it if the node id matches.

editor/src/clj/editor/outline_view.clj
vlaaad: Good catch, I used it for debugging when working on the code.

PR #11347, opened 2025-10-15
Shortcut displayed for tooltips when they have a keybinding mapped to them

### LINKED ISSUE #11345: Enhancement: Show keybinding in editor tooltips for commands
This PR helps display the active keybinding for tooltip pop-ups when UI elements have shortcut/keybindings mapped to a command/action.

JosephFerano:
Tooltips like the ones for the Move, Rotate, Scale buttons in the scene editor can now showcase the keybinding that's mapped to them, allowing new users to more easily discover hotkeys.

Fixes #11345


editor/src/clj/editor/ui.clj
```diff
@@ -522,8 +522,17 @@
...
+  ([^Control ctrl tip localization]
+   (.setTooltip ctrl (when tip (localization/localize! (Tooltip.) localization tip))))
+  ([^Control ctrl tip localization keymap command]
+   (.setTooltip ctrl (when tip
+                       (let [tooltip-obj (localization/localize! (Tooltip.) localization tip)
+                             shortcut (when (and keymap command)
+                                        (editor.keymap/display-text keymap command ""))]
+                         (when (seq shortcut)
+                           (.setText tooltip-obj (str (.getText tooltip-obj) " (" shortcut ")")))
+                         tooltip-obj)))))
```
vlaaad, line 537: From the abstraction point of view, I'm certain the tooltip function should not be concerned with keymaps and commands. Also, this code doesn't work: `localization/localize!` not only applies the localization immediately, but also adds a listener that will update the text when the user switches the language. This means changing the language will hide the shortcut part of a tooltip. Also, we cannot expect that every language will use the `{command} ({shortcut})` pattern. For example:
- Korean uses dashes: `이동 도구 - E`
- RTL languages (Arabic and Hebrew) don't use parens at all: `E أداة التحريك`

This should be done as a localization message instead of passing the keymap and command here.

editor/src/clj/editor/ui.clj
```diff
@@ -1889,11 +1898,11 @@
         hbox)
 
       :else
-      (let [{:keys [graphic-fn label icon tooltip more]} menu-item
+      (let [{:keys [graphic-fn label icon tooltip command more]} menu-item
             label (or (handler/label handler-ctx) label)
             button (doto (ToggleButton.)
                      (localization/localize! localization label)
-                     (tooltip! tooltip localization))]
+                     (tooltip! tooltip localization keymap command))]
```
vlaaad, line 1907: This place should check if there is a shortcut, and if there is, it should wrap the tooltip in a localization message, e.g.:
```clj
(tooltip! 
  (if shortcut 
    (localization/message "..." {"tooltip" tooltip "shortcut" shortcut}) 
    tooltip) 
  localization)
```
And then, in `en.editor_localization`, the entry for the key should be `{tooltip} ({shortcut})`.

Not sure what's a fitting localization message key here... Something like `blah.tooltip-with-shortcut`...

vlaaad (changes requested):

editor/src/clj/editor/ui.clj
JosephFerano: Yes, from the abstraction point of view, I was hesitant at first for the reasons you mentioned, but I noticed that this function is called in several places, figured that as might be better to centralize things. However, these things you mentioned about localizing shortcuts naturally changes everything, I'll investigate these things you mentioned.

editor/src/clj/editor/ui.clj
JosephFerano: Got it, will investigate.

PR #11349, opened 2025-10-15
Add logic to recover lost sound device on Windows

deurell:
Handle recovery of lost sound device on Windows.

### Technical changes
- Add NotifyDeviceInvalidated() so backends can queue a reset instead of silently stopping.
- Reopen the WASAPI device in ResetDevice() during the next update, reuse the original frame count/buffer count.
- Clamp GetAvailableFrames() to the existing/original allocation and log the mismatch when the driver reports a different frame count after lost device.


engine/sound/src/devices/device_wasapi.cpp
```diff
@@ -538,7 +539,7 @@ namespace dmDeviceWasapi
                 return;
             }
 
-            uint32_t frames_available = buffer_size;// buffer_size - buffer_pos;
+            uint32_t frames_available = buffer_size - buffer_pos;
```
deurell, line 544: Will trigger a AUDCLNT_E_BUFFER_TOO_LARGE on shutdown if not taking buffer_pos into account.

engine/sound/src/sound.cpp
```diff
@@ -273,6 +275,54 @@ namespace dmSound
...
+        }
+
+        sound->m_Device = new_device;
+
+        DeviceInfo device_info = {0};
+        sound->m_DeviceType->m_DeviceInfo(sound->m_Device, &device_info);
+
+        if (device_info.m_FrameCount && device_info.m_FrameCount != sound->m_DeviceFrameCount)
+        {
+            dmLogWarning("Audio device reported frame count %u after reset (previous %u). Clamping to previous allocation.", device_info.m_FrameCount, sound->m_DeviceFrameCount);
```
deurell: Log that new device frame count (after recovery) is different. It will be clamped to the smallest one.

engine/sound/src/sound.cpp
```diff
@@ -273,6 +275,54 @@ namespace dmSound
...
+        }
+
+        if (device_info.m_MixRate && device_info.m_MixRate != sound->m_MixRate)
+        {
+            dmLogWarning("Audio device reported mix rate %u after reset (previous %u).", device_info.m_MixRate, sound->m_MixRate);
```
deurell, line 317: log and update MixRate if different

engine/sound/src/sound.cpp
```diff
@@ -1779,6 +1842,8 @@ namespace dmSound
                 {
                     break;
                 }
+                // New device after reset/recovery can have a larger frame count than we allocated buffers for
+                frame_count = dmMath::Min(frame_count, sound->m_DeviceFrameCount);
```
deurell, line 1889: New device after reset/recovery can have a larger frame count than we allocated buffers for

engine/sound/src/devices/device_wasapi.cpp
```diff
@@ -66,6 +66,7 @@ namespace dmDeviceWasapi
         {
             dmLogError("WASAPI device lost in %s (hr=0x%08X)", context, hr);
             CheckAndPrintError(hr);
+            dmSound::NotifyDeviceInvalidated();
```
JCash: This is a new dependency making it a bit circular.
Previously, the devices doesn't have any knowledge about the sound system itself.

Also, another (old) problem we have is that there is no sound context being passed from sound.cpp to the device. So, you have to call a function with no arguments.
If we were to do that, I'd like to change it so that it _does_ take a sound context, and that the sound.cpp would have to pass the context to the device.

Another option would be to let the device return a "device lost" enum to the sound.cpp.
I think that is currently a better approach, avoiding the two issues I mentioned above.

engine/sound/src/sound.cpp
```diff
@@ -382,6 +432,8 @@ namespace dmSound
         sound->m_HasWindowFocus = true; // Assume we startup with the window focused
         sound->m_DeviceType = device_type;
         sound->m_Device = device;
+        sound->m_DeviceParams = device_params; // Stash frame and buffer count for potential device reset
+        sound->m_DeviceResetPending = false;
```
JCash, line 438: would it make sense to call ResetDevice() in the Initialze() function?
(To avoid duplicated code for setting up a device)

JCash (changes requested):
overall good
some rearrangement requests

engine/sound/src/devices/device_wasapi.cpp
deurell: Nice feedback. Thanks, will have a look and redesign it. 👍

deurell: Did an alternate implementation, skipping the NotifyDeviceInvalidated call in favor of returning RESULT_DEVICE_LOST. Do you think it's better this way or should I do it differently?

engine/sound/src/sound.cpp
deurell: Tried it but got a bit messy to merge them since I wanted my reset to be fast and reuse/reopen the existing backend and init has to set up everything so kept them separate. Can give it another go if important to have one path.

engine/sound/src/sound.cpp
JCash: Am I missing something, there's no clamping going on in this if-statement? 🤔

JCash (changes requested):
The message about clamping looks a bit off as there is no clamping going on?

engine/sound/src/sound.cpp
deurell: Maybe misleading comment. I log it here, and do a frame_count = dmMath::Min(frame_count, sound->m_DeviceFrameCount); @ row 1846

JCash (dismissed):

deurell:
This PR was approved previously but now had merge conflicts with dev. Reverse Integrated dev into feature branch. Tested and verified device lost and reconnect to default device.

JCash (approved):

Issue #11299, opened 2025-10-06
Editor exceptions shouldn't disable functionality permanently


matgis:
**Describe the bug (REQUIRED)**
When an exception happens in the editor, it can sometimes cause command handlers or views associated with editor tabs to stop responding. For example, it might cause the editor to no longer respond to changes in the Properties view.

**To Reproduce (REQUIRED)**
There are various ways that this can happen, but it would probably be simpler to simulate exceptions from different call sites. For example:
* Throw from `active` `enabled`, `options` and `run` in command handlers.
* Throw from an `AnimationTimer` used to refresh a view.

**Expected behavior (REQUIRED)**
If an exception forces us to disable editor functionality, switching to a different editor tab should re-enable disabled functionality.

matgis:
Possibly related: #3883

PR #11353, opened 2025-10-15
Re-enable stopped timers and add dev helper function

JosephFerano:
This PR will add a private atom of a set called `stopped-timers` which will keep track of them so we can then restart them from a debug/dev function. More work can be done to do this automatically for users.

Related #11299
Possibly related: #3883


matgis (approved):
Looks good to me! Please remove `Fixes #11299` from the PR description before merging, though. We don't want to close it until we've fixed the user-facing issue as well!

PR #11357, opened 2025-10-15
Fixed Crash when using raycasts under certain conditions

### LINKED ISSUE #10312: Crash when using raycasts under certain conditions
Then please report your issue in the corresponding extension repository!_

Using physics.raycast() sometimes causes a crash.

**To Reproduce (REQUIRED)**
This code:

```
	local from = vmath.vector3(0, -13200000, 0)
	local to = from + vmath.vector3(0, -1, 0)
	local groups = {hash("test")}
	physics.raycast(from, to, groups)
```

Together with the game.project setting:

```
[physics]
scale = 0.01
```

Causes a crash with this message:

`Assertion failed: (r.LengthSquared() > 0.0f), function RayCast, file b2DynamicTree.h, line 232.`

Usually when a ray’s vector has 0 length, this warning is displayed instead:

`WARNING:PHYSICS: Ray had 0 length when ray casting, ignoring request.`

It could have something to do with floating point precision, because of the extreme numbers, coupled with a short ray.

The crash can be avoided, but it may be better to catch it and display the above message instead?

1.9.7

**Platforms (REQUIRED):**
macOS Sonoma (14.6.1)

[RaycastCrash.zip](https://github.com/user-attachments/files/19144151/RaycastCrash.zip)

**Workaround (OPTIONAL):**
Avoid using raycasts for extreme numbers.

[Add any other context about the problem here.](https://forum.defold.com/t/raycast-crash/79990)

DarkRavenStar:
This fixes a crash when the length of a 2D or 3D raycast is zero.
This fix specifically takes into account the case when the ray length becomes zero after applying the `physics.scale` setting in game.project.

Fixes #10312

### Technical notes
Added extra dmLogWarning if ray had 0 length after applying physics scale
Added Unit Test to test for zero length raycast before and oafter applying scale

### DIFF
--- a/engine/physics/src/physics/box2d/box2d_physics.cpp
+++ b/engine/physics/src/physics/box2d/box2d_physics.cpp
@@ -2105,6 +2105,12 @@ namespace dmPhysics
         ToB2(from2d, from, scale);
         b2Vec2 to;
         ToB2(to2d, to, scale);
+        
+        if (b2LengthSquared((to - from)) <= 0.0f)
+        {
+            dmLogWarning("Ray had 0 length when ray casting after applying physics scale, ignoring request.");
+            return;
+        }
 
         // Box2d V3 requires a translation vector, not a point
         b2Vec2 translate = b2Sub(to, from);
--- a/engine/physics/src/physics/box2d/box2d_physics.cpp
+++ b/engine/physics/src/physics/box2d/box2d_physics.cpp
@@ -2093,11 +2093,6 @@ namespace dmPhysics
 
         const Point3 from2d = Point3(request.m_From.getX(), request.m_From.getY(), 0.0);
         const Point3 to2d = Point3(request.m_To.getX(), request.m_To.getY(), 0.0);
-        if (lengthSqr(to2d - from2d) <= 0.0f)
-        {
-            dmLogWarning("Ray had 0 length when ray casting, ignoring request.");
-            return;
-        }
 
         float scale = world->m_Context->m_Scale;
 
--- a/engine/physics/src/physics/box2d_defold/box2d_defold_physics.cpp
+++ b/engine/physics/src/physics/box2d_defold/box2d_defold_physics.cpp
@@ -1396,18 +1396,26 @@ namespace dmPhysics
         }
 
         float scale = world->m_Context->m_Scale;
+        b2Vec2 from;
+        ToB2(from2d, from, scale);
+        b2Vec2 to;
+        ToB2(to2d, to, scale);
+        
+        if ((to - from).LengthSquared() <= 0.0f)
+        {
+            dmLogWarning("Ray had 0 length when ray casting after applying physics scale, ignoring request.");
+            return;
+        }
+
         ProcessRayCastResultCallback2D query;
         query.m_Request = &request;
         query.m_ReturnAllResults = request.m_ReturnAllResults;
         query.m_Context = world->m_Context;
         query.m_Results = &results;
-        b2Vec2 from;
-        ToB2(from2d, from, scale);
-        b2Vec2 to;
-        ToB2(to2d, to, scale);
         query.m_IgnoredUserData = request.m_IgnoredUserData;
         query.m_CollisionMask = request.m_Mask;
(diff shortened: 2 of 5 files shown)


engine/physics/src/physics/box2d/box2d_physics.cpp
```diff
@@ -2105,6 +2105,12 @@ namespace dmPhysics
         ToB2(from2d, from, scale);
         b2Vec2 to;
         ToB2(to2d, to, scale);
+        
+        if (b2LengthSquared((to - from)) <= 0.0f)
+        {
+            dmLogWarning("Ray had 0 length when ray casting after applying physics scale, ignoring request.");
+            return;
+        }
```
JCash, line 2108: Do we really need two checks?
I suggest we remove the first check, and keep this one.

engine/physics/src/physics/box2d_defold/box2d_defold_physics.cpp
```diff
@@ -1396,18 +1396,26 @@ namespace dmPhysics
...
+        b2Vec2 from;
+        ToB2(from2d, from, scale);
+        b2Vec2 to;
+        ToB2(to2d, to, scale);
+        
+        if ((to - from).LengthSquared() <= 0.0f)
+        {
+            dmLogWarning("Ray had 0 length when ray casting after applying physics scale, ignoring request.");
+            return;
+        }
```
JCash, line 1404: Same here

engine/physics/src/physics/physics_3d.cpp
```diff
@@ -1213,6 +1213,12 @@ namespace dmPhysics
         ToBt(request.m_From, from, scale);
         btVector3 to;
         ToBt(request.m_To, to, scale);
+        
+        if ((to - from).length2() <= 0.0f)
+        {
+            dmLogWarning("Ray had 0 length when ray casting after applying physics scale, ignoring request.");
+            return;
+        }
```
JCash, line 1215: and here

JCash (changes requested):

engine/physics/src/physics/box2d/box2d_physics.cpp
britzl: What do you mean by two checks? Where is the first one?

DarkRavenStar: @britzl, I think @JCash is refering(few lines above my newly added warning message), same with other 2 files
[L2096](https://github.com/defold/defold/blob/8be1bf350159f8fe755602528237b8c400f7389d/engine/physics/src/physics/box2d/box2d_physics.cpp#L2096)

@JCash I agree with you. I will make it into 1 check only

engine/physics/src/physics/box2d_defold/box2d_defold_physics.cpp
DarkRavenStar: Fixed

engine/physics/src/physics/physics_3d.cpp
DarkRavenStar: Fixed

JCash (changes requested):
lgtm. Only thing missing is a small addition to our unit tests.

engine/physics/src/physics/test/test_physics.cpp
```diff
@@ -1034,6 +1035,34 @@ TYPED_TEST(PhysicsTest, EmptyRayCasting)
...
+    // A zero length raycast after applying scale
+    hits.SetSize(0);
+    request.m_From = Point3(0.0f, -13200000.0f, 0.0f);
+    request.m_To = Point3(0.0f, -13200001.0f, 0.0f);
+    request.m_ReturnAllResults = 0;
+    (*TestFixture::m_Test.m_RayCastFunc)(TestFixture::m_World, request, hits);
+
+    ASSERT_EQ(0u, hits.Size());
+}
+
```
JCash, line 1065: 👍

JCash (dismissed):

JCash (approved):

PR #11359, opened 2025-10-15
Enhanced the orthographic camera behavior in the editor for better usability and accuracy

### LINKED ISSUE #8495: Orthographic Zoom should affect frustum box visualization for camera component
Currently Orthographic Zoom does not affect the frustum visualization for camera component in Defold Editor. The Near and Far planes are affecting it, but not zoom.

Width and height of frustum box for orthographic camera representation is affected by Orthographic Zoom property of camera component.

-

Related: https://github.com/defold/defold/issues/3227

### LINKED ISSUE #11007: Orthographic gizmo size is strange
Orthographic gizmo size
https://forum.defold.com/t/orthographic-gizmo-size/80907

The camera gizmo size in orthographic mode is not displaying what is expected.
While manipulating the `Fov` can make it correct, since `Fov` values should be ignored in orthographic mode (according to the manual), I would like this to be fixed to the correct behavior.

>Fov
>(Perspective camera only) - The vertical camera field of view expressed in radians. The wider the field of view, the more the camera will see.

Best regards.

AGulev:
- Orthographic camera gizmo size in the scene view reflects `game.project` display width/height and Orthographic Zoom.
- Orthographic camera FOV is now disabled when Orthographic Projection is enabled.
- Perspective-only fields are disabled in orthographic mode.
- Orthographic Zoom must be greater than zero; invalid values show an error.

Fix https://github.com/defold/defold/issues/8495
Fix https://github.com/defold/defold/issues/11007


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/meta.properties
```diff
@@ -62,10 +62,12 @@ group = Runtime
 width.type = integer
 width.help = the width in pixels of the application window, 960 by default
 width.default = 960
+width.minimum = 1
```
AGulev, line 65: make sure user can't set display width and height 0 or smaller

editor/src/clj/editor/camera_editor.clj
```diff
@@ -218,12 +219,20 @@
     (.setColumn m 3 (.x p) (.y p) (.z p) 1.0)
     m))
 
-(defn- camera-projection-matrix [is-orthographic near far fov-deg aspect-ratio]
-  ;; TODO: Derive aspect-ratio from display setting in game.project when auto-aspect-ratio is enabled.
+(defn- camera-projection-matrix
+  [is-orthographic near far fov-deg aspect-ratio display-width display-height orthographic-zoom orthographic-mode]
```
AGulev, line 222: Use proper calculations for ortographic camera

editor/src/clj/editor/camera_editor.clj
```diff
@@ -273,30 +282,39 @@
...
-    fov :fov
-    near-z :near-z
-    far-z :far-z
-    auto-aspect-ratio (protobuf/int->boolean :auto-aspect-ratio)
-    orthographic-projection (protobuf/int->boolean :orthographic-projection)
-    orthographic-zoom :orthographic-zoom
-    orthographic-mode :orthographic-mode))
+  (concat
+    (g/connect project :display-width self :project-display-width)
+    (g/connect project :display-height self :project-display-height)
```
AGulev, line 301: connect nodes from game project

editor/src/clj/editor/camera_editor.clj
```diff
@@ -273,30 +282,39 @@
...
+      orthographic-mode :orthographic-mode)))
 
 (g/defnode CameraNode
   (inherits resource-node/ResourceNode)
 
   (property aspect-ratio g/Num ; Required protobuf field.
             (dynamic label (properties/label-dynamic :camera :aspect-ratio))
-            (dynamic tooltip (properties/tooltip-dynamic :camera :aspect-ratio)))
+            (dynamic tooltip (properties/tooltip-dynamic :camera :aspect-ratio))
+            (dynamic read-only? (g/fnk [orthographic-projection] orthographic-projection)))
```
AGulev, line 318: better UX for perspective/orthographic mode (show and hide fields depends on mode)

editor/src/clj/editor/camera_editor.clj
```diff
@@ -317,7 +336,11 @@
   (property orthographic-zoom g/Num (default (protobuf/default Camera$CameraDesc :orthographic-zoom))
             (dynamic label (properties/label-dynamic :camera :orthographic-zoom))
             (dynamic tooltip (properties/tooltip-dynamic :camera :orthographic-zoom))
-            (dynamic read-only? (g/fnk [orthographic-projection orthographic-mode] (not (and orthographic-projection (= orthographic-mode :ortho-mode-fixed))))))
+            (dynamic read-only? (g/fnk [orthographic-projection orthographic-mode] (not (and orthographic-projection (= orthographic-mode :ortho-mode-fixed)))))
+            (dynamic error (validation/prop-error-fnk :fatal validation/prop-zero-or-below? orthographic-zoom)))
```
AGulev, line 345: validate that zoom is positive

editor/src/clj/editor/defold_project.clj
```diff
@@ -1606,6 +1606,10 @@
                                                                  (not (resource/read-only? resource))))))
                                                    save-data)))
   (output settings g/Any (g/fnk [settings] (or settings gpc/default-settings)))
+  (output display-width g/Num (g/fnk [settings]
+                                 (double (or (get settings ["display" "width"]) 0))))
```
AGulev, line 1610: is ok to cast value to doule here?

editor/src/clj/editor/camera_editor.clj
```diff
@@ -27,6 +27,7 @@
             [editor.graph-util :as gu]
             [editor.localization :as localization]
             [editor.math :as math]
+            [editor.validation :as validation]
```
matgis: We like to keep requires and imports sorted.

editor/src/clj/editor/camera_editor.clj
```diff
@@ -235,10 +244,10 @@
...
             cg (get color 1)
             cb (get color 2)
-            {:keys [is-orthographic near-z far-z fov aspect-ratio]} (:user-data renderable)
+            {:keys [is-orthographic near-z far-z fov aspect-ratio display-width display-height orthographic-zoom orthographic-mode]} (:user-data renderable)
             world-translation (:world-translation renderable)
             world-rotation (:world-rotation renderable)
-            ^Matrix4d proj-matrix (camera-projection-matrix is-orthographic near-z far-z fov aspect-ratio)
+            ^Matrix4d proj-matrix (camera-projection-matrix is-orthographic near-z far-z fov aspect-ratio display-width display-height orthographic-zoom orthographic-mode)
             ^Matrix4d view-matrix (camera-view-matrix world-translation world-rotation)
             ^Matrix4d inv-view-matrix (math/inverse view-matrix)
```
matgis: This will throw an exception if Orthographic Zoom is zero. It gets printed to the console, and the entire camera will disappear from view. We should probably just hide the frustum preview in that case instead?

editor/src/clj/editor/defold_project.clj
matgis: It should be OK, I think. I mean, users can hand-edit `game.project` and put a non-number value in there, but I don't think we should deal with that situation here - it should be validated by the `game.project` node if anything.

matgis (dismissed):
Nice! ❤️

I found that the entire camera would disappear from view due to an exception from a non-invertible matrix if I set Orthographic Zoom to 0.0 using the Property Editor. We might want to check for this and not try to render the frustum part so the camera wireframe is still visible.

editor/src/clj/editor/camera_editor.clj
AGulev: hm, I thought I protected against it by 
>(dynamic error (validation/prop-error-fnk :fatal validation/prop-zero-or-below? orthographic-zoom)))

I'll fix it

editor/src/clj/editor/camera_editor.clj
AGulev: I hope IDEA has something for it, I'll fix it

matgis: It's under Code -> Optimize Imports in the menu. 👍

editor/src/clj/editor/camera_editor.clj
matgis: Yeah, the field shows up as an error in the property editor. However, it can still be in this broken state before the user can fix it, so we need to protect against its usage too.

AGulev: Fixed!

matgis (approved):

PR #11361, opened 2025-10-16
Report debugger attaching exception through a notification

JosephFerano:
We are presenting a Dialog Window when the user receives an exception because the debugger cannot connect. This PR will change it to a notification for better UX since it doesn't seem to affect anything, it's a stray attach attempt from hitting the <F5> key too quickly in succession.

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -55,6 +55,7 @@
             [editor.lsp :as lsp]
             [editor.lua :as lua]
             [editor.menu-items :as menu-items]
+            [editor.notifications :as notifications]
             [editor.os :as os]
             [editor.pipeline :as pipeline]
             [editor.pipeline.bob :as bob]
@@ -1405,9 +1406,20 @@
       (let [main-scene (.getScene ^Stage main-stage)
             render-build-error! (make-render-build-error main-scene tool-tab-pane build-errors-view)]
         (build-errors-view/clear-build-errors build-errors-view)
-        (if (debug-view/can-attach? prefs)
-          (attach-debugger! workspace project prefs debug-view render-build-error!)
-          (run-with-debugger! workspace project prefs debug-view render-build-error! web-server))))))
+        (try
+          (if (debug-view/can-attach? prefs)
+            (attach-debugger! workspace project prefs debug-view render-build-error!)
+            (run-with-debugger! workspace project prefs debug-view render-build-error! web-server))
+          (catch Exception e
+            (notifications/show!
+             (workspace/notifications workspace)
+             {:type :error
+              :id ::debugger-connection-error
+              :message (localization/message "notification.debug-view.connect-failed.error" {"error" "Testing"})})
+            (throw (ex-info "Failed to attach to debugger"
+                            {:type ::cannot-check-attach
+                             :suppressed? true}
+                            e))))))))
 
 (def ^:private clean-build-dialog-info
   {:title (localization/message "dialog.clean-build.title")
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -1418,7 +1418,7 @@
               :message (localization/message "notification.debug-view.connect-failed.error" {"error" "Testing"})})
             (throw (ex-info "Failed to attach to debugger"
                             {:type ::cannot-check-attach
-                             :suppressed? true}
+                             :suppress-dialog-window? true}
                             e))))))))
 
 (def ^:private clean-build-dialog-info
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -96,7 +96,7 @@
            [com.dynamo.bob Platform]
            [com.sun.javafx.scene NodeHelper]
            [java.io File PipedInputStream PipedOutputStream]
-           [java.net URL SocketTimeoutException]
+           [java.net SocketTimeoutException URL]
            [java.util Arrays Collection List]
            [java.util.concurrent ExecutionException]
            [javafx.beans.value ChangeListener]
@@ -873,7 +873,7 @@
 (defn- on-service-url-found [prefs target]
   (engine/apply-simulated-resolution! prefs target))
 
-(defn- launch-built-project! [project workspace engine-descriptor project-directory prefs web-server debug?]
+(defn- launch-built-project! [project engine-descriptor project-directory prefs web-server debug?]
   (let [selected-target (targets/selected-target prefs)
         launch-new-engine! (fn []
                              (targets/kill-launched-targets!)
@@ -919,7 +919,7 @@
               (reboot-engine! selected-target web-server debug?))
             (launch-new-engine!))))
       (catch SocketTimeoutException e
-        (debug-view/show-connect-failed-info! e workspace))
+        (debug-view/show-connect-failed-info! e (project/workspace project)))
       (catch Exception e
         (log/warn :exception e)
         (let [localization (g/with-auto-evaluation-context evaluation-context
@@ -1324,7 +1324,7 @@
                                (when (handle-build-results! workspace render-build-error! build-results)
                                  (when (or engine skip-engine)
                                    (show-console! main-scene tool-tab-pane)
-                                   (launch-built-project! project workspace engine project-directory prefs web-server false)))))))
+                                   (launch-built-project! project engine project-directory prefs web-server false)))))))
 
 (handler/defhandler :project.build :global
   (enabled? [] (not (build-in-progress?)))
@@ -1375,7 +1375,7 @@
                   :result-fn (fn [{:keys [engine] :as build-results}]
                                (when (handle-build-results! workspace render-build-error! build-results)
                                  (when (or engine skip-engine)
-                                   (when-let [target (launch-built-project! project workspace engine project-directory prefs web-server true)]
+                                   (when-let [target (launch-built-project! project engine project-directory prefs web-server true)]
                                      (when (nil? (debug-view/current-session debug-view))
                                        (debug-view/start-debugger! debug-view project (:address target "localhost") (:instance-index target 0))))))))))
 
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -1410,16 +1410,8 @@
           (if (debug-view/can-attach? prefs)
             (attach-debugger! workspace project prefs debug-view render-build-error!)
             (run-with-debugger! workspace project prefs debug-view render-build-error! web-server))
-          (catch Exception e
-            (notifications/show!
-             (workspace/notifications workspace)
-             {:type :error
-              :id ::debugger-connection-error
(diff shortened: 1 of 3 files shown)


editor/src/clj/editor/app_view.clj
```diff
@@ -1405,9 +1406,20 @@
...
+          (catch Exception e
+            (notifications/show!
+             (workspace/notifications workspace)
+             {:type :error
+              :id ::debugger-connection-error
+              :message (localization/message "notification.debug-view.connect-failed.error" {"error" "Testing"})})
+            (throw (ex-info "Failed to attach to debugger"
+                            {:type ::cannot-check-attach
+                             :suppress-dialog-window? true}
+                            e))))))))
```
matgis: Why do you need to throw the exception after you've caught it? Isn't it enough to just show the notification in this case?

editor/src/clj/editor/app_view.clj
```diff
@@ -1405,9 +1406,20 @@
...
+        (try
+          (if (debug-view/can-attach? prefs)
+            (attach-debugger! workspace project prefs debug-view render-build-error!)
+            (run-with-debugger! workspace project prefs debug-view render-build-error! web-server))
+          (catch Exception e
+            (notifications/show!
+             (workspace/notifications workspace)
+             {:type :error
+              :id ::debugger-connection-error
+              :message (localization/message "notification.debug-view.connect-failed.error" {"error" "Testing"})})
```
matgis: Why does it say `{"error" "Testing"}` here? It shows up in the notification for me.

Also style: We indent using two spaces.

Actually, I found an existing function `debug-view/show-connect-failed-info!`, which notifies about connection issues later in the process -  I think it would be better to simply call that one from here?

editor/src/clj/editor/app_view.clj
```diff
@@ -1405,9 +1406,20 @@
...
             render-build-error! (make-render-build-error main-scene tool-tab-pane build-errors-view)]
         (build-errors-view/clear-build-errors build-errors-view)
-        (if (debug-view/can-attach? prefs)
-          (attach-debugger! workspace project prefs debug-view render-build-error!)
-          (run-with-debugger! workspace project prefs debug-view render-build-error! web-server))))))
+        (try
+          (if (debug-view/can-attach? prefs)
+            (attach-debugger! workspace project prefs debug-view render-build-error!)
+            (run-with-debugger! workspace project prefs debug-view render-build-error! web-server))
+          (catch Exception e
```
matgis: It may be better to only catch the specific type of exception thrown in the situation we want to catch?

editor/src/clj/editor/error_reporting.clj
```diff
@@ -105,8 +105,9 @@
    (report-exception! exception (Thread/currentThread)))
   ([^Throwable exception thread]
    (try
-     (let [{:keys [ex-map suppressed?]} (record-exception! exception)]
-       (if suppressed?
+     (let [{:keys [ex-map]} (record-exception! exception)
+           suppress-dialog-window? (some #(-> % :data :suppress-dialog-window?) (:via ex-map))]
+       (if suppress-dialog-window?
```
matgis: If we don't throw the exception from the catch clause, we won't need this.

matgis (changes requested):

JosephFerano:
Ok, will take a look

editor/src/clj/editor/app_view.clj
JosephFerano: The reasoning I had was that in the `report-exception!` function, it was doing some additional exception handling, with analytics and logging. So I rethrow with the `:suppress-dialog-window?` key to let that function handle all the other tasks.

However, I just realized now I made a mistake with where I put the `if` form, it should only wrap the section.

```clojure
             (let [sentry-id-promise (sentry-reporter exception thread)]
               (exception-notifier ex-map sentry-id-promise)
               nil))))
```

Originally I wanted to handle the notification in the `report-exception!` function, so this could would switch on a key for which GUI system to use, so any exception can be caught and with a key decide whether to use notifications or the dialog window, but I wasn't sure how to resolve the `workspace` dependency that the notifications module needed.

However, I am making an assumption that might not be the desired behavior. If we don't want this exception to be treated as one at all, meaning don't bother reporting with analytics or logging, then yes, it makes complete sense to not rethrow at all.

editor/src/clj/editor/app_view.clj
JosephFerano: Yes, sorry, this was intended to be a WIP PR because I wasn't sure if my approach was the right way to go. I will look into this function and correct the spacing.

JosephFerano:
@matgis I made the changes, that function you mentioned simplified things. I also had the idea of introducing a "cooldown" for the Start/Attach command, so that accidental double taps don't trigger the error, but I don't know how involved that would be. It does seem like the editor ignores the command once it's attached, but just not during that small window where it's attaching. The only way to consistently reproduce is to spam F5.

matgis (changes requested):
The changes so far look good! However, while testing it out, I managed to trigger a yet another exception dialog of this type! This time, it was coming from the `app-view/reboot-engine!` function. 😬 Maybe we're catching the Exception in the wrong place? 🤔 Or we just need to catch it here as well?

To clarify, the editor keeps track of the engine process it started, and if it is still alive when you Build and Run, it sends it a message to `reboot` itself with the new content instead of starting up a fresh engine. It saves a bit of load time.

```
java.net.SocketTimeoutException: Read timed out
        at java.base/sun.nio.ch.NioSocketImpl.timedRead(NioSocketImpl.java:278)
        at java.base/sun.nio.ch.NioSocketImpl.implRead(NioSocketImpl.java:304)
        at java.base/sun.nio.ch.NioSocketImpl.read(NioSocketImpl.java:346)
        at java.base/sun.nio.ch.NioSocketImpl$1.read(NioSocketImpl.java:796)
        at java.base/java.net.Socket$SocketInputStream.read(Socket.java:1099)
        at java.base/java.io.BufferedInputStream.fill(BufferedInputStream.java:291)
        at java.base/java.io.BufferedInputStream.read1(BufferedInputStream.java:347)
        at java.base/java.io.BufferedInputStream.implRead(BufferedInputStream.java:420)
        at java.base/java.io.BufferedInputStream.read(BufferedInputStream.java:399)
        at java.base/sun.net.www.http.HttpClient.parseHTTPHeader(HttpClient.java:827)
        at java.base/sun.net.www.http.HttpClient.parseHTTP(HttpClient.java:759)
        at java.base/sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1705)
        at java.base/sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1614)
        at editor.engine$reboot_BANG_.invokeStatic(engine.clj:122)
        at editor.engine$reboot_BANG_.invoke(engine.clj:103)
        at editor.app_view$reboot_engine_BANG_.invokeStatic(app_view.clj:853)
        at editor.app_view$reboot_engine_BANG_.invoke(app_view.clj:849)
        at editor.app_view$launch_built_project_BANG_.invokeStatic(app_view.clj:919)
        at editor.app_view$launch_built_project_BANG_.invoke(app_view.clj:876)
        at editor.app_view$run_with_debugger_BANG_$fn__100917.invoke(app_view.clj:1376)
        at editor.app_view$async_build_BANG_$finish_with_result_BANG___100684.invoke(app_view.clj:1134)
        at editor.app_view$async_build_BANG_$phase_7_await_lint_BANG___100686$process_lint_results_on_ui_thread_BANG___100689.invoke(app_view.clj:1174)
        at editor.app_view$async_build_BANG_$run_on_background_thread_BANG___100678$fn__100679$fn__100680.invoke(app_view.clj:1116)
        at com.sun.javafx.application.PlatformImpl.lambda$runLater$10(PlatformImpl.java:456)
        at com.sun.javafx.application.PlatformImpl.lambda$runLater$11(PlatformImpl.java:455)
```

editor/src/clj/editor/app_view.clj
matgis: Yeah, so because it is an error we know can happen, and that relates to the user's setup, we want to notify them about it without making an error report. In retrospect I could have been clearer about this when describing the issue. 👍

JosephFerano:
I'll take a look

JosephFerano:
> The changes so far look good! However, while testing it out, I managed to trigger a yet another exception dialog of this type! This time, it was coming from the `app-view/reboot-engine!` function. 😬 Maybe we're catching the Exception in the wrong place? 🤔 Or we just need to catch it here as well?
> 
> To clarify, the editor keeps track of the engine process it started, and if it is still alive when you Build and Run, it sends it a message to `reboot` itself with the new content instead of starting up a fresh engine. It saves a bit of load time.
> 
>...

How can I reproduce this?

matgis:
Build and Run, then once the game is running, quickly press `F5` a bunch from the editor is how it happed for me.

JosephFerano:
> Build and Run, then once the game is running, quickly press `F5` a bunch from the editor is how it happed for me.

Ok, so I had to do an extra step to reproduce;

https://github.com/defold/defold/blob/dev/editor/src/clj/editor/debug_view.clj#L527-L531

This function is returning false for you. I could not get this to return false for me, I had to force it. What do I have to do in my environment to trigger this to return false?

And you were correct, there's a separate exception handling path;

https://github.com/defold/defold/blob/dev/editor/src/clj/editor/app_view.clj#L918-L939

For the quick fix, I can leave the `catch Exception e` block alone, and just add a case for `SocketTimeoutException` that just rethrows so it's caught by ex handler that triggers a notification.

Perhaps an improvement here could be to refactor all paths that invoke dialogs and replace them with a rethrown ex-info with data that can inform the central ex handler which options to use for reporting the error, like whether to use dialog vs notification, log type vs no log, and report analytics vs no report. If this is feasible and worth doing, I can do it for this one instance and add TODOs at all the call sites and refactor the rest when it's convenient.

matgis:
Personally I'd prefer to do the simple fix now, and revisit the idea of building a new system around error reporting when we encounter more situations like this.

JosephFerano:
Ok, made the simple fix.

editor/src/clj/editor/app_view.clj
```diff
@@ -96,7 +96,7 @@
            [com.dynamo.bob Platform]
            [com.sun.javafx.scene NodeHelper]
            [java.io File PipedInputStream PipedOutputStream]
-           [java.net URL]
+           [java.net URL SocketTimeoutException]
```
matgis: Keep imported symbols sorted.

editor/src/clj/editor/app_view.clj
```diff
@@ -873,7 +873,7 @@
 (defn- on-service-url-found [prefs target]
   (engine/apply-simulated-resolution! prefs target))
 
-(defn- launch-built-project! [project engine-descriptor project-directory prefs web-server debug?]
+(defn- launch-built-project! [project workspace engine-descriptor project-directory prefs web-server debug?]
```
matgis: You can get the `workspace` from the `project` using `project/workspace`, so you don't need to add a `workspace` argument here.

matgis (dismissed):

matgis (approved):
LGTM! ❤️

PR #11367, opened 2025-10-17
WelcomeScreen: Enter keybinding to open selected project

### LINKED ISSUE #11184: Open the selected recent project when pressing Enter
I noticed that when the recent projects window opens I can use arrow keys (up/down) to select a project from the list, but there seems to be no way to open the selected project without using mouse to press the button.

Pressing the Enter key should open the selected project.

Could open the top most recent project with CTRL+Enter (even when not selected in the list).

Having to reach for the mouse because of "little things" is always a pain. 😄

JosephFerano:
Currently pressing ENTER in the welcome screen to select a project doesn't do anything. This PR adds the ENTER keybinding so it opens the currently selected project.

Fixes #11184

### DIFF
--- a/editor/src/clj/editor/welcome.clj
+++ b/editor/src/clj/editor/welcome.clj
@@ -412,7 +412,10 @@
               {:graphic (make-recent-project-entry recent-project recent-projects-list prefs localization)}))
           (.setOnMouseClicked (ui/event-handler event
                                 (when (= 2 (.getClickCount ^MouseEvent event))
-                                  (open-selected-project!)))))))
+                                  (open-selected-project!))))
+          (.setOnKeyPressed (ui/event-handler event
+                      (when (= javafx.scene.input.KeyCode/ENTER (.getCode ^KeyEvent event))
+                        (open-selected-project!)))))))
     home-pane))
 
 ;; -----------------------------------------------------------------------------
@@ -473,7 +476,7 @@
   (ui/with-controls root [^ButtonBase create-new-project-button
                           new-project-location-field
                           ^TextField new-project-title-field
-                          ^ListVew template-list
+                          ^ListView template-list
                           new-project-title-label
                           new-project-location-label]
     (.setText new-project-title-field (localization (localization/message "welcome.new-project.default-name")))
@@ -486,7 +489,11 @@
       (b/bind! (location-field-title-property new-project-location-field) sanitized-title-property))
     (doto template-list
       (ui/cell-factory! (fn [project-template]
-                          {:graphic (make-template-entry project-template localization)})))
+                          {:graphic (make-template-entry project-template localization)}))
+      (.setOnKeyPressed (ui/event-handler event
+                                            (when (= javafx.scene.input.KeyCode/ENTER (.getCode ^KeyEvent event))
+                                              (when-some [project-template (first (ui/selection template-list))]
+                                                (.fire create-new-project-button))))))
     (when (some? templates)
       (ui/items! template-list templates)
       (when (seq templates)
@@ -815,7 +822,7 @@
                                 (let [key-event ^KeyEvent event
                                       selected-pane-button (.getSelectedToggle pane-buttons-toggle-group)]
                                   (when (and (.isShortcutDown key-event)
-                                             (= "r" (.getText key-event)))
+                                             (= javafx.scene.input.KeyCode/R (.getCode key-event)))
                                     (ui/close! stage)
                                     (show-welcome-dialog! prefs localization updater open-project-fn
                                                           {:x (.getX stage)
--- a/editor/src/clj/editor/welcome.clj
+++ b/editor/src/clj/editor/welcome.clj
@@ -51,7 +51,7 @@
            [javafx.event Event]
            [javafx.geometry Pos]
            [javafx.scene Node Parent Scene]
-           [javafx.scene.control Button ButtonBase ComboBox Hyperlink Label ListView OverrunStyle ProgressBar RadioButton TextArea TextField ToggleGroup]
+           [javafx.scene.control Button ButtonBase ComboBox Hyperlink Label ListCell ListView OverrunStyle ProgressBar RadioButton TextArea TextField ToggleGroup]
            [javafx.scene.image Image ImageView]
            [javafx.scene.input KeyEvent MouseEvent]
            [javafx.scene.layout HBox Priority Region StackPane VBox]
@@ -408,14 +408,20 @@
           (.setFixedCellSize 56.0)
           (ui/items! (recent-projects prefs))
           (ui/cell-factory!
-            (fn [recent-project]
-              {:graphic (make-recent-project-entry recent-project recent-projects-list prefs localization)}))
+           (fn [recent-project]
+             {:graphic (make-recent-project-entry recent-project recent-projects-list prefs localization)}))
           (.setOnMouseClicked (ui/event-handler event
-                                (when (= 2 (.getClickCount ^MouseEvent event))
-                                  (open-selected-project!))))
+                                (let [^MouseEvent mouse-event event
+                                      ^Node target (.getTarget mouse-event)]
+                                  (when (and (instance? javafx.scene.control.ListCell target)
+                                             (nil? (.getItem ^ListCell target)))
+                                    (.clearSelection (.getSelectionModel recent-projects-list)))
+                                  (when (and (= 2 (.getClickCount mouse-event))
+                                             (not-empty (ui/selection recent-projects-list)))
+                                    (open-selected-project!)))))
           (.setOnKeyPressed (ui/event-handler event
-                      (when (= javafx.scene.input.KeyCode/ENTER (.getCode ^KeyEvent event))
-                        (open-selected-project!)))))))
+                              (when (= javafx.scene.input.KeyCode/ENTER (.getCode ^KeyEvent event))
+                                (open-selected-project!)))))))
     home-pane))
 
 ;; -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


vlaaad (changes requested):
Please remove the commented-out code before finishing the work on a PR.

editor/src/clj/editor/welcome.clj
```diff
@@ -412,7 +412,10 @@
               {:graphic (make-recent-project-entry recent-project recent-projects-list prefs localization)}))
           (.setOnMouseClicked (ui/event-handler event
                                 (when (= 2 (.getClickCount ^MouseEvent event))
-                                  (open-selected-project!)))))))
+                                  (open-selected-project!))))
+          (.setOnKeyPressed (ui/event-handler event
+                      (when (= javafx.scene.input.KeyCode/ENTER (.getCode ^KeyEvent event))
+                        (open-selected-project!)))))))
```
vlaaad: Please keep the code formatted. Here `(when ...)` should be indented more than `(ui/event-handler ...)` since it's nested inside it (this will break my parinfer setup...)

editor/src/clj/editor/welcome.clj
```diff
@@ -486,7 +489,11 @@
       (b/bind! (location-field-title-property new-project-location-field) sanitized-title-property))
     (doto template-list
       (ui/cell-factory! (fn [project-template]
-                          {:graphic (make-template-entry project-template localization)})))
+                          {:graphic (make-template-entry project-template localization)}))
+      (.setOnKeyPressed (ui/event-handler event
+                                            (when (= javafx.scene.input.KeyCode/ENTER (.getCode ^KeyEvent event))
```
vlaaad: Indentation seems wrong here too...

JosephFerano:
@vlaaad Fixed indentation and I added the mouse double click functionality. Should work now. If everything looks good the way I did it, I think this PR is ready.

editor/src/clj/editor/welcome.clj
```diff
@@ -408,11 +408,20 @@
           (.setFixedCellSize 56.0)
           (ui/items! (recent-projects prefs))
           (ui/cell-factory!
-            (fn [recent-project]
-              {:graphic (make-recent-project-entry recent-project recent-projects-list prefs localization)}))
+           (fn [recent-project]
```
matgis: Style: We indent with two spaces when the arguments are on a separate line from the function:
```clj
(ui/cell-factory!
  (fn [recent-project]
    ,,,)
```

editor/src/clj/editor/welcome.clj
```diff
@@ -408,11 +408,20 @@
...
+             {:graphic (make-recent-project-entry recent-project recent-projects-list prefs localization)}))
```
matgis: Style: The indentation changed on these two lines, but it was correct before. We use two spaces to indent function arguments on their own line.
```suggestion
            (fn [recent-project]
              {:graphic (make-recent-project-entry recent-project recent-projects-list prefs localization)}))
```

editor/src/clj/editor/welcome.clj
```diff
@@ -486,7 +495,20 @@
...
+                                  ^Node target (.getTarget mouse-event)]
+                              (when (and (instance? javafx.scene.control.ListCell target)
+                                         (nil? (.getItem ^ListCell target)))
+                                (.clearSelection (.getSelectionModel template-list)))
+                              (when (and (= 2 (.getClickCount mouse-event))
+                                         (not-empty (ui/selection template-list)))
+                                (.fire create-new-project-button)))))
+      (.setOnKeyPressed (ui/event-handler event
+                          (when (= javafx.scene.input.KeyCode/ENTER (.getCode ^KeyEvent event))
+                            (when-some [project-template (first (ui/selection template-list))]
```
matgis: `project-template` is unused, so you could combine the check with the outer `when`?

matgis (changes requested):
Nice, I can now press enter to create a new project when the input focus is on the Template list!

However, I'd also like to be able to press enter when focus is on the Title text field, since I would typically want to type in a name for the new project before creating it. I think this could be achieved by setting `defaultButton="true"` on the `create-new-project-button` directly in the `fxml` file?

JosephFerano:
Fixed with improvements

editor/src/clj/editor/welcome.clj
```diff
@@ -486,7 +498,20 @@
...
+                              (when (and (instance? javafx.scene.control.ListCell target)
+                                         (nil? (.getItem ^ListCell target)))
+                                (.clearSelection (.getSelectionModel template-list)))
+                              (when (and (= 2 (.getClickCount mouse-event))
+                                         (not-empty (ui/selection template-list)))
+                                (.fire create-new-project-button)))))
+      (.setOnKeyPressed (ui/event-handler event
+                          (when (and (= javafx.scene.input.KeyCode/ENTER (.getCode ^KeyEvent event))
+                                     (first (ui/selection template-list)))
+                              (.fire create-new-project-button)))))
```
matgis: Indentation fix:
```suggestion
                            (.fire create-new-project-button)))))
```

matgis (dismissed):
Works great! It's great that it focuses on the first project in the list, too! ❤️

matgis (dismissed):
LGTM!

editor/src/clj/editor/welcome.clj
```diff
@@ -411,8 +411,20 @@
             (fn [recent-project]
               {:graphic (make-recent-project-entry recent-project recent-projects-list prefs localization)}))
           (.setOnMouseClicked (ui/event-handler event
-                                (when (= 2 (.getClickCount ^MouseEvent event))
-                                  (open-selected-project!)))))))
+                                (let [^MouseEvent mouse-event event
+                                      ^Node target (.getTarget mouse-event)]
+                                  (when (and (instance? javafx.scene.control.ListCell target)
```
vlaaad: Minor: since you import it, you can use the short form here:
```suggestion
                                  (when (and (instance? ListCell target)
```

editor/src/clj/editor/welcome.clj
```diff
@@ -411,8 +411,20 @@
...
+                                             (nil? (.getItem ^ListCell target)))
+                                    (.clearSelection (.getSelectionModel recent-projects-list)))
+                                  (when (and (= 2 (.getClickCount mouse-event))
+                                             (not-empty (ui/selection recent-projects-list)))
+                                    (open-selected-project!)))))
+          (.setOnKeyPressed (ui/event-handler event
+                              (when (= javafx.scene.input.KeyCode/ENTER (.getCode ^KeyEvent event))
```
vlaaad: We should also import `javafx.scene.input.KeyCode`.

editor/src/clj/editor/welcome.clj
```diff
@@ -486,7 +498,20 @@
...
+                              (when (and (instance? javafx.scene.control.ListCell target)
+                                         (nil? (.getItem ^ListCell target)))
+                                (.clearSelection (.getSelectionModel template-list)))
+                              (when (and (= 2 (.getClickCount mouse-event))
+                                         (not-empty (ui/selection template-list)))
+                                (.fire create-new-project-button)))))
+      (.setOnKeyPressed (ui/event-handler event
+                          (when (and (= javafx.scene.input.KeyCode/ENTER (.getCode ^KeyEvent event))
+                                     (first (ui/selection template-list)))
+                            (.fire create-new-project-button)))))
```
vlaaad: Similar import issues in this hunk...

editor/src/clj/editor/welcome.clj
```diff
@@ -815,7 +840,7 @@
                                 (let [key-event ^KeyEvent event
                                       selected-pane-button (.getSelectedToggle pane-buttons-toggle-group)]
                                   (when (and (.isShortcutDown key-event)
-                                             (= "r" (.getText key-event)))
+                                             (= javafx.scene.input.KeyCode/R (.getCode key-event)))
```
vlaaad: And here!

vlaaad (commented):
Now double-click on a template also cause the editor to create a project and launch it. Not sure we want this — user should probably pick a name for it. Also, if you restart the editor, and then double-click on the same template, it will attempt to create another project at the same path and will fail, so double-click will lead to an error dialog. I'd limit double-click behavior only to the project list.

JosephFerano:
@vlaaad I fixed the imports :)

> Now double-click on a template also cause the editor to create a project and launch it. Not sure we want this — user should probably pick a name for it. Also, if you restart the editor, and then double-click on the same template, it will attempt to create another project at the same path and will fail, so double-click will lead to an error dialog. I'd limit double-click behavior only to the project list.

I agree that the flow is a little off, but I think the double-click you're describing is orthogonal. Right now, you can also press Enter to open a project and run into the same issue. I can remove the double click, not sure if we need to remove Enter for the same reasons (except for the input text box, I think that one makes sense).

I have an idea though that I think would address your concern better, just requires a little bit of extra work. This feels like what we want is to draw the user's attention to the steps required to create a valid new templated project. For that, I think a modal window that asks the user to confirm the project name and directory location might be better. We then remove the form fields from the bottom. While it's an extra step for the user, it removes confusion and then unifies the interfaces between recent projects and new projects a bit more (since recent projects doesn't have these fields). Thoughts?

@matgis

vlaaad:
Maybe we could focus on the project title text field when double-clicking or pressing enter in the template list?

JosephFerano:
> Maybe we could focus on the project title text field when double-clicking or pressing enter in the template list?

That's not a bad idea, yeah, it kind of does a similar thing to the modal window without needing a modal window, so that's good.

JosephFerano:
Actually, the one issue would be, will the user notice that that's what happened? Like if their expectation when pressing enter was to open the project because they didn't know they  were supposed to fill out a text field, all of a sudden if the focus changes, they would need some kind of visual indicator that this happened. Usually editors that do this sort of thing will have a flash or animation to draw the user's attention to where the selection is focused.

vlaaad:
There appears a bold orange underline on a text field when it becomes focused.

JosephFerano:
Seems to stand out enough, let me know how that looks

vlaaad (approved):
LGTM!

PR #11375, opened 2025-10-17
Ensure the camera component icon (gizmo) maintains a pixel-stable size on screen

### LINKED ISSUE #7713: The camera component has a fixed presentation size in the editor, regardless of the in-game scale.
I tried the new camera component introduced in Defold 1.4.6, and I'm not sure if the representation in the Editor is okay. The camera component looks too big with the 1 point per meter level scale in a 3d game.

I don’t know if I’m right about using 1 point per meter, but I found it handy for 3D space. If I use 1 centimeter as a point, then on a large level I have to operate with coordinates in the form of tens or hundreds of thousands of points.

Perhaps the camera could be better made more independent of the scale in the game. For example, remove the camera model or replace it with a sprite that is always rendered at the same size? I don't know exactly what's easier to do.

**To Reproduce (REQUIRED)**
1. Place any model with scale of 1 unit per 1 meter in the collection.
2. Place the camera component in the collection.
3. See how big the camera component is.

The camera looks the same size, regardless of the in-game scale.

  - Version 1.4.6

**Screenshots (OPTIONAL):**
55b16493190e08b937dba299821451e9a828d433
b54dbaec7c127e26dc3b3a1ee406076fe8580e4b

AGulev:
Ensure the camera component’s icon in the editor stays a constant size regardless of scene scale.

Fix https://github.com/defold/defold/issues/7713

### DIFF
--- a/editor/src/clj/editor/camera_editor.clj
+++ b/editor/src/clj/editor/camera_editor.clj
@@ -18,16 +18,14 @@
             [dynamo.graph :as g]
             [editor.build-target :as bt]
             [editor.camera :as camera]
-            [editor.types :as types]
             [editor.colors :as colors]
             [editor.geom :as geom]
             [editor.gl :as gl]
             [editor.gl.pass :as pass]
             [editor.gl.shader :as shader]
             [editor.gl.vertex :as vtx]
-            [editor.scene-tools :as scene-tools]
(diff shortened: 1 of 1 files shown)


matgis (approved):
Very nice! ❤️

PR #11376, opened 2025-10-18
Remove empty property from the `game.projectc`

AGulev:
In `meta.properties` we have :
```
[]
group-order = Main,Platforms,Components,Runtime,Distribution
default-category = project
```

It's needed for the editor. But it's not needed it in the output file. This PR removes it from the output file.

Fix https://github.com/defold/defold/issues/11364


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/BobProjectProperties.java
```diff
@@ -36,7 +36,6 @@
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.io.IOUtils;
 
-import com.dynamo.bob.util.StringUtil;
```
AGulev, line 39: this import isn't used

JCash (approved):

PR #11384, opened 2025-10-20
Added an ability to specify `library.defold_min_version` for extensions to prevent using incompatible extensions

### LINKED ISSUE #11276: Decide on extension editor versioning system
In order to allow the editor show better user facing messages about keeping versions in sync.

As a first step, we need to decide on how to put the required Defold version in some info in an extension.

NOTE: Not all extensions are on github, so we shouldn't rely on github functionality for the info.

Q: Should the file be at the root, or in the "library" folder?
Q: We like to display the Defold version in the extension release notes. Can we automate our releases to read the defold version info from the file, so we don't have to do it manually?

AGulev:
Extensions that define `library.defold_min_version` are now validated against the current Defold version.
If the extension requires a newer version than the one being used, the build will fail with an error message.

Fix https://github.com/defold/defold/issues/11276


JCash (dismissed):

editor/src/clj/editor/library.clj
```diff
@@ -247,6 +248,68 @@
...
+      (if (< i len)
+        (let [sa (get a i "0")
+              sb (get b i "0")
+              ia (parse-leading-int sa)
+              ib (parse-leading-int sb)
+              d (- ia ib)]
+          (if (zero? d)
+            (recur (unchecked-inc i))
+            d))
+        0))))
```
vlaaad: I think we should avoid duplicating code between bob/java and editor/clojure whenever possible. Here it would be quite easy to do: make `LibraryUtil.compareVersions` public, and use it instead of these functions.

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -225,20 +225,36 @@
...
-                (let [failed-tasks (filterv #(not (.isOk ^TaskResult %)) (.-taskResults result))]
-                  (if (coll/empty? failed-tasks)
-                    {:error (g/map->error {:message "Bob failed, see console output for more details" :severity :fatal})}
-                    {:error {:causes (engine-build-errors/failed-tasks-error-causes project evaluation-context failed-tasks)}})))))))
+
+          ;; Pre-build guard: fail fast if any dependency requires a newer Defold version.
+          (let [deps (g/node-value workspace :dependencies evaluation-context)
+                incompatible (into [] (filter #(and (= :error (:status %)) (= :defold-min-version (:reason %)))) deps)]
+            (if (seq incompatible)
+              (let [loc (workspace/localization workspace evaluation-context)
```
vlaaad: `loc` -> `localization`

editor/src/clj/editor/pipeline/bob.clj
```diff
@@ -225,20 +225,36 @@
...
-                           internal-options
-                           (into-array String cli-args))]
-              (if (.-success result)
-                nil
-                (let [failed-tasks (filterv #(not (.isOk ^TaskResult %)) (.-taskResults result))]
-                  (if (coll/empty? failed-tasks)
-                    {:error (g/map->error {:message "Bob failed, see console output for more details" :severity :fatal})}
-                    {:error {:causes (engine-build-errors/failed-tasks-error-causes project evaluation-context failed-tasks)}})))))))
+
+          ;; Pre-build guard: fail fast if any dependency requires a newer Defold version.
```
vlaaad: Is this necessary? I thought Bob would perform this check? If so, why use it here? I would prefer if Bob bundling from the editor would stay as simple as possible, i.e., a straightforward invoke -> get results. We don't localize any other bob error, so we shouldn't do this either.

editor/src/clj/editor/settings.clj
```diff
@@ -169,6 +170,26 @@
                       error))))
           setting-values)))
 
+(defn get-dependencies-setting-errors
+  "Return build errors derived from project dependencies (e.g., incompatible
+  library.defold_min_version). Expects a Project node id and an
+  evaluation-context."
+  [project evaluation-context]
+  (let [ws (project/workspace project evaluation-context)
```
vlaaad: `ws` -> `workspace`

editor/src/clj/editor/settings.clj
```diff
@@ -169,6 +170,26 @@
...
+        deps (g/node-value ws :dependencies evaluation-context)
+        incompatible (into [] (filter #(and (= :error (:status %))
+                                            (= :defold-min-version (:reason %)))) deps)
+        loc (workspace/localization ws evaluation-context)]
```
vlaaad: `loc` -> `localization`

editor/src/clj/editor/settings.clj
```diff
@@ -231,8 +252,10 @@
   (output form-data g/Any :cached produce-form-data)
 
   (output save-value g/Any (gu/passthrough merged-raw-settings))
-  (output setting-errors g/Any :cached (g/fnk [form-data]
-                                              (get-settings-errors form-data)))
+  (output setting-errors g/Any :cached (g/fnk [form-data project]
+                                         (let [base-errors (get-settings-errors form-data)]
+                                           (g/with-auto-evaluation-context evaluation-context
+                                             (into base-errors (get-dependencies-setting-errors project evaluation-context))))))
```
vlaaad: It is forbidden to use `g/with-evaluation-context` from within node outputs because it breaks the dependencies and invalidation. If you want more stuff from the graph, you have to wire it.

editor/src/clj/editor/workspace.clj
```diff
@@ -540,14 +540,24 @@ ordinary paths."
   (let [escaped-name (protobuf/escape-string name)]
     (string/replace template "{{NAME}}" escaped-name)))
 
+(defn- dependency-error-line [loc {:keys [uri reason required current]}]
```
vlaaad: `loc` -> `localization`

editor/src/clj/editor/workspace.clj
```diff
@@ -540,14 +540,24 @@ ordinary paths."
...
-  (let [{:keys [error missing]} (->> lib-states
-                                     (eduction
-                                       (keep (fn [{:keys [status file uri]}]
-                                               (cond
-                                                 (= status :error) (pair :error uri)
-                                                 (nil? file) (pair :missing uri)))))
-                                     (iutil/group-into {} [] key val))
+  (let [missing (into [] (keep (fn [{:keys [file uri]}] (when (nil? file) uri))) lib-states)
+        error (g/with-auto-evaluation-context evaluation-context
+                (let [loc (localization workspace evaluation-context)]
```
vlaaad: `loc` -> `localization`

editor/src/clj/editor/workspace.clj
```diff
@@ -540,14 +540,24 @@ ordinary paths."
...
+                  (into []
+                        (keep (fn [s]
+                                (when (= :error (:status s))
+                                  (dependency-error-line loc s))))
```
vlaaad: Actually, this approach is wrong: you apply localization at the time of construction. This means the text will be translated only once. Instead, notifications expect localization messages before the localization is applied, so that the notification will update itself if the users switches language while the notification is open.

editor/test/editor/library_min_version_test.clj
```diff
@@ -0,0 +1,73 @@
...
+        bad (make-lib-zip "bad" too-high)
+        states [{:status :up-to-date :file ok :uri (java.net.URI. "https://example.com/ok.zip")}
+                {:status :up-to-date :file bad :uri (java.net.URI. "https://example.com/bad.zip")}]
+        validated (library/validate-updated-libraries states)
+        [ok-state bad-state] validated]
+    (is (not= :error (:status ok-state)))
+    (is (= :error (:status bad-state)))
+    (is (= :defold-min-version (:reason bad-state)))
+    (is (= too-high (:required bad-state)))
+    (is (= current (:current bad-state)))))
```
vlaaad: I guess we won't need these tests if we just use the library version comparator function from bob.

vlaaad (changes requested):
Please deduplicate the version comparison between bob and editor. It's also forbidden to use `g/with-auto-evaluation-context` from graph outputs. You probably don't even need it, because you do it to get a localization instance, but you shouldn't be doing it in the first place. We don't localize errors from bob. When we localize notifications, we do it inside the notification system, and you should only produce localization messages that don't need, and shouldn't use, the localization instance. We don't localize errors in the editor yet, but when we'll do that, we will use an approach similar to notifications.

editor/src/clj/editor/pipeline/bob.clj
AGulev: it's needed because bob check it on `resolve` command which we don't do when build from the editor

editor/src/clj/editor/library.clj
AGulev: it makes sense

editor/src/clj/editor/pipeline/bob.clj
AGulev: but I shouldn't localize anything at all, right?

editor/src/clj/editor/workspace.clj
AGulev: >We don't localize errors from bob.

that means I don't need to localize it at all for now, right?

vlaaad: Yes!

editor/src/clj/editor/pipeline/bob.clj
vlaaad: I still think this approach is wrong. This function is not only used for bundling. We expose it as a generic bob invocation to editor scripts. The editor scripts might use it for whatever, including just doing `--help` — we don't want to check dependency versions in that case. Perhaps we should introduce another command to Bob that checks, but does not resolve, dependencies? Or perform this check in Bob, not at a resolution step, but when trying to build?

editor/src/clj/editor/pipeline/bob.clj
vlaaad: Yeah :)

editor/src/clj/editor/library.clj
```diff
@@ -298,15 +331,26 @@
     render-progress!))
 
 (defn validate-updated-libraries
-  "Validate newly downloaded libraries (:status is :stale).
+  "Validate libraries after fetching updates.
```
AGulev: @vlaaad I've changed the way validation implemented. Now I do it separately as part of  libraries validation:
now error injected via Settings → GameProject, so it shows immediately for both Cmd+B and Cmd+M.

editor/src/clj/editor/library.clj
AGulev: fixed

editor/src/clj/editor/pipeline/bob.clj
AGulev: now I validate `dependencies` filed in the game.project, so it's not needed anymore

editor/src/clj/editor/settings.clj
AGulev: fixed

editor/src/clj/editor/settings.clj
AGulev: fixed

editor/src/clj/editor/library.clj
```diff
@@ -247,6 +249,37 @@
...
+          (let [base (library-base-path file)
+                entry-path (if (str/blank? base)
+                             "game.project"
+                             (str base "/game.project"))
+                entry (.getEntry zip entry-path)]
+            (if (some? entry)
+              (with-open [r (io/reader (.getInputStream zip entry))]
+                (let [settings (settings-core/parse-settings r)
+                      minv (some-> (settings-core/get-setting settings ["library" "defold_min_version"]) str)]
+                  (if (and minv (not (str/blank? minv)) (LibraryUtil/isCurrentEngineOlderThan minv))
```
vlaaad: `(and minv (not (str/blank? minv)))` check is also performed in `LibraryUtil/isCurrentEngineOlderThan` function. Remove it here? Or there?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/LibraryUtil.java
```diff
@@ -151,6 +203,15 @@ public static Set<String> readIncludeDirsFromArchive(String includeBaseDir, ZipF
                 BobProjectProperties properties = new BobProjectProperties();
                 properties.load(is);
                 String dirs = properties.getStringValue("library", "include_dirs", "");
+                // Validate minimum required Defold/Bob version if specified
+                String minVersion = properties.getStringValue("library", "defold_min_version", "");
+                if (minVersion != null && !minVersion.isEmpty()) {
+                    if (compareVersions(EngineVersion.version, minVersion) < 0) {
```
vlaaad: Should this place also call `isCurrentEngineOlderThan`? This is the same check it seems.

editor/src/clj/editor/library.clj
```diff
@@ -247,6 +249,37 @@
...
+  [lib-state file-key]
+  (let [^File file (get lib-state file-key)]
+    (if (and file (.isFile file))
+      (try
+        (with-open [zip (ZipFile. file)]
+          (let [base (library-base-path file)
+                entry-path (if (str/blank? base)
+                             "game.project"
+                             (str base "/game.project"))
+                entry (.getEntry zip entry-path)]
```
vlaaad: Something strange is going on here... `library-base-path` already opens the zip file and searches for the `game.project` file. Issues:
1. It seems unnecessary to open the zip twice: first, to find the entry, and second, to read it. Locating a entry in a zip file requires a full scan, and reading zips might be slow, so we should avoid unnecessary scans.
2. If `library-base-path` returns nil, there is no point in invoking `.getEntry` to find it: it's not there.

editor/src/clj/editor/library.clj
```diff
@@ -247,6 +249,37 @@
...
+            (if (some? entry)
+              (with-open [r (io/reader (.getInputStream zip entry))]
+                (let [settings (settings-core/parse-settings r)
+                      minv (some-> (settings-core/get-setting settings ["library" "defold_min_version"]) str)]
```
vlaaad: Why the `(some-> ... str)` part? It seems the setting is already a string?

editor/src/clj/editor/settings.clj
```diff
@@ -169,6 +169,19 @@
...
+  library.defold_min_version)."
+  [dependencies]
+  (let [min-version (into [] (filter #(and (= :error (:status %))
+                                           (= :defold-min-version (:reason %)))) dependencies)]
+    (mapv (fn [{:keys [uri required current]}]
+            (g/map->error
+              {:severity :fatal
+               :message (format "Library '%s' requires Defold %s or newer (bob.jar is %s). Update Defold or check older extension versions for compatibility."
+                                (str uri) (str required) (str current))}))
+          min-version)))
```
vlaaad: Should be something like:
```clj
(into []
  (comp
    (filter #(and ...))
    (map (fn ...)))
  dependencies)
```
You unnecessarily create an intermediate vector (`min-version`) that is immediately transformed to return value.

editor/src/clj/editor/workspace.clj
```diff
@@ -549,6 +549,29 @@ ordinary paths."
...
+        (let [message (format "Library '%s' requires Defold %s or newer (bob.jar is %s). Update Defold or check older extension versions for compatibility."
+                          (str uri) (str required) (str current))]
+          (notifications/show!
+            notifications
+            {:id (notification-id uri)
+             :type :error
+             :message (reify editor.localization.MessagePattern
+                        (format [_ _] message))
+             :actions [{:message (reify editor.localization.MessagePattern
+                                   (format [_ _] "Open game.project"))
```
vlaaad: Here, we actually need localizations. And we don't reify `editor.localization.MessagePattern` willy nilly, please use `(localization/message "...")` instead.

vlaaad (changes requested):
I like this approach much better, but there are still some issues.

editor/src/clj/editor/library.clj
AGulev: I didn't find a nice way to avoid it. 
validation happening before `resource_watch/load-library-zip` happens

If you can suggest how to make it, I'll be happy to implement this.

editor/src/clj/editor/library.clj
AGulev: 👍

editor/src/clj/editor/library.clj
AGulev: 👍

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/LibraryUtil.java
AGulev: 👌

editor/src/clj/editor/settings.clj
AGulev: 👍

editor/src/clj/editor/workspace.clj
AGulev: only for `"Open game.project"` ?

vlaaad: For every :message here

editor/src/clj/editor/settings.clj
```diff
@@ -169,6 +169,21 @@
...
+  [dependencies]
+  (into []
+        (comp
+          (filter #(and (= :error (:status %))
+                        (= :defold-min-version (:reason %))))
+          (map (fn [{:keys [uri required current]}]
+                 (g/map->error
+                   {:severity :fatal
+                    :message (format "Library '%s' requires Defold %s or newer (bob.jar is %s). Update Defold or check older extension versions for compatibility."
+                                     (str uri) (str required) (str current))}))))
```
AGulev, line 184: this is a build error, as I understand we don't localize it yet

editor/src/clj/editor/app_view.clj
```diff
@@ -2938,6 +2938,7 @@
                   (disk/async-reload! render-install-progress! workspace [] changes-view
                                       (fn [success]
                                         (when success
+                                          (workspace/update-dependencies-min-version-from-cache! workspace)
```
vlaaad, line 2941: I'm not sure I support the existence of `update-dependencies-min-version-from-cache!` function; this place is an example of why. First, we do `(workspace/install-validated-libraries! workspace lib-states)` on line 2937, which sets project libraries to some values that may be incorrect, but will only be marked as such eventually, once the reload is complete.

editor/src/clj/editor/defold_project.clj
```diff
@@ -1876,6 +1879,7 @@
     (render-progress! (swap! progress progress/advance 4 (localization/message "progress.syncing-resources")))
     (du/log-time "Initial resource sync"
       (workspace/resource-sync! workspace-id [] (progress/nest-render-progress render-progress! @progress)))
+    (workspace/update-dependencies-min-version-from-cache! workspace-id)
```
vlaaad, line 1882: Here, we set project dependencies on lines 1876 or 1877, depending on internet availability, and the values may still be incorrect, but we will know about it later, after the sync. Introducing this "eventual correctness" seems unnecessary, because we already read the zips in `fetch-and-validate-libraries` where we locate, but currently don't read the `game.project` entry (`library-base-path`).

vlaaad (changes requested):
`update-dependencies-min-version-from-cache!` should not exist — it adds "eventual correctness" to library states, where at different points in time we may have an installed list of libraries that is reported as good, but will become bad later.

The proper solution would be to:
1. update `validate-updated-library` fn so that instead of `library-base-path` it uses another function that both locates the `game.project` entry and reads the minimum version from it.
2. In the case when the internet is not available when we launch the project (defold_project.clj, line 1877 in your PR), we should also validate the minimum version of libraries returned by `current-library-state`

AGulev:
editor implementation is too complicated, it would be better if someone from the editor side will take a look into the issue

PR #11385, opened 2025-10-20
Save and Restore open tabs

JosephFerano:
Currently, when a user has open tabs and assigned to different panes, when they close the editor, they are all closed out and require opening them up again. This PR will restore all previous open tabs, in the correct order, and in the proper splits.

Fixes #3868


maxim-yudin:
Thank you. I'll be waiting for this feature.

vlaaad:
Hmm, was it necessary to move `make-tab!` and related functions to another ns?

JosephFerano:
> Hmm, was it necessary to move `make-tab!` and related functions to another ns?

This is what I had mentioned about moving code that's in app-view into the new editor-tabs module. But maybe I didn't communicate that properly as I had mentioned consolidating recent-files only. I have a stash where I moved even more code (everything related to editor tabs that's in app-view) to the new ns. But I wanted to merge this and use another PR for that. If this is not desired I can stash these few functions I moved as well.

vlaaad:
Yeah, I thought, consolidating recent-files ns with open tabs makes sense, but I don't like that we move `make-tabs!` there...

JosephFerano:
> Yeah, I thought, consolidating recent-files ns with open tabs makes sense, but I don't like that we move `make-tabs!` there...

Roger, I'll undo

JosephFerano:
@vlaaad So the reason I had created `editor-tabs` ns was because the idea was to move code from app-view that is tab related into `editor-tabs` and pass functions as args when needed. But now it just makes sense to move the little bit of prefs related code I had to recent-files ns, since it's pretty similar. Let me know what you think.

vlaaad:
I think it's fine if it's in recent-files or app-view ns.

editor/src/clj/editor/app_view.clj
```diff
@@ -2010,29 +2010,35 @@
...
-                    tab-panes (.getItems editor-tabs-split)]
-                (when (< 1 (count tab-panes))
-                  (.remove tab-panes tab-pane)
-                  (let [remaining-tab-pane (.get tab-panes 0)
-                        selected-tab (ui/selected-tab remaining-tab-pane)
-                        resource-node (tab->resource-node selected-tab)
-                        view-type (tab->view-type selected-tab)]
-                    (.requestFocus ^TabPane remaining-tab-pane)
-                    (on-selected-tab-changed! app-view app-scene selected-tab resource-node view-type)))))))))
+       (let [save-scheduled (atom false)]
```
vlaaad: Is this necessary? It's not expensive to set prefs values since `prefs/set!` only updates an in-memory state. It's also thread-safe, if that matters...

editor/src/clj/editor/app_view.clj
```diff
@@ -2281,6 +2289,51 @@
                                                                           {"error" msg})}}))))
              false)))))))
 
+(defn- open-tabs-from-prefs [app-view prefs localization workspace project evaluation-context tab-panes-to-restore]
+  (into []
+        (for [[pane-num pane] (map-indexed vector tab-panes-to-restore)
+              [proj-path view-type-id] pane
+              :let [tab-pane (g/node-value app-view :active-tab-pane)
```
vlaaad: 1. Use `evaluation-context` as an arg to `g/node-value`.
2. Move it outside the loop — we only need to do it once per `open-tabs-from-prefs` fn call.

editor/src/clj/editor/app_view.clj
```diff
@@ -2281,6 +2289,51 @@
...
+                                           {:selected-view-type view-type
+                                            :use-custom-editor false})]
+              :when opened?]
+          {:pane-num pane-num
+           :tab (ui/selected-tab tab-pane)})))
+
+(defn restore-tabs-from-prefs! [app-view prefs localization workspace project evaluation-context]
+  (when-let [tab-panes-to-restore (seq (prefs/get prefs [:workflow :open-tabs]))]
+    (let [{:keys [selected-pane tab-selection-by-pane]} (prefs/get prefs [:workflow :last-selected-tabs])
+          editor-tabs-split ^SplitPane (g/node-value app-view :editor-tabs-split)
```
vlaaad: Use `evaluation-context` as an arg to `g/node-value`.

editor/src/clj/editor/app_view.clj
```diff
@@ -2281,6 +2289,51 @@
...
+          opened-tabs (open-tabs-from-prefs app-view prefs localization workspace
+                                            project evaluation-context tab-panes-to-restore)
+          tab-panes (.getItems editor-tabs-split)
+          first-tab-pane (.get tab-panes 0)
+          tabs-to-move (map :tab (filter #(= 1 (:pane-num %)) opened-tabs))
```
vlaaad: Minor nitpick: This is probably not going to be a hot loop, but we generally try to avoid lazy sequences in favor of transducers because they are much faster. You could do:
```clj
(into []
  (comp
    (filter ...)
    (map ...))
  opened-tabs)
```
...or:
```clj
(->> opened-tabs
     (e/filter ...)
     (e/map ...)
     vec)
```
... or maybe `coll/transfer` which does the same, but with a more intuitive order.

editor/src/clj/editor/app_view.clj
```diff
@@ -2281,6 +2289,51 @@
...
+          ;; NOTE: We're just assuming there's only ever going to be two max splits, certainly would
+          ;; need to change if we made a more elaborate window tiling system.
+          second-tab-pane (when (seq tabs-to-move)
```
vlaaad: For the same performance/memory pressure reasons, we typically avoid `seq` in favor of `coll/empty?` and `coll/not-empty` checks.

editor/src/clj/editor/boot_open_project.clj
```diff
@@ -285,7 +285,9 @@
 
       (ui/run-later
         (app-view/restore-split-positions! scene prefs)
-        (app-view/restore-hidden-panes! scene prefs))
+        (app-view/restore-hidden-panes! scene prefs)
+        (g/with-auto-evaluation-context evaluation-context
```
vlaaad: 👍

editor/src/clj/editor/recent_files.clj
```diff
@@ -106,4 +106,53 @@
...
+    :resource))
+
+(defn- serialize-open-tabs [app-view]
+  (let [editor-tabs-split ^SplitPane (g/with-auto-evaluation-context ec
+                                       (g/node-value app-view :editor-tabs-split ec))]
+    (mapv (fn [^TabPane tab-pane]
+            (mapv (fn [tab]
+                    [(resource/proj-path (tab->resource tab))
+                     (-> tab
+                         tab->resource
```
vlaaad: It's probably better to call `tab->resource` once in this fn.

editor/src/clj/editor/recent_files.clj
```diff
@@ -106,4 +106,53 @@
...
+    (g/node-value :view-data)
+    second
+    :resource))
+
+(defn- serialize-open-tabs [app-view]
+  (let [editor-tabs-split ^SplitPane (g/with-auto-evaluation-context ec
+                                       (g/node-value app-view :editor-tabs-split ec))]
+    (mapv (fn [^TabPane tab-pane]
+            (mapv (fn [tab]
+                    [(resource/proj-path (tab->resource tab))
```
vlaaad: `resource/proj-path` will throw if it receives a `nil` argument, but `some->` in `tab->resource` implies it might return nil. Probably we should replace `some->` with `->` and assert that the returned value is not `nil`.

editor/src/clj/editor/recent_files.clj
```diff
@@ -25,7 +25,7 @@
             [editor.ui :as ui]
             [editor.ui.fuzzy-choices :as fuzzy-choices]
             [editor.workspace :as workspace])
-  (:import [javafx.scene.control TabPane]))
+  (:import [javafx.scene.control SplitPane Tab TabPane]))
```
vlaaad, line 28: Let's add `(set! *warn-on-reflection* true)` after the `ns` declaration and ensure there are no reflection warnings.

editor/src/clj/editor/boot_open_project.clj
```diff
@@ -173,7 +173,7 @@
           workbench            (.lookup root "#workbench")
           notifications        (.lookup root "#notifications")
           scene-visibility     (scene-visibility/make-scene-visibility-node! *view-graph*)
-          app-view             (app-view/make-app-view *view-graph* project stage menu-bar editor-tabs-split tool-tabs prefs localization)
+          app-view             (app-view/make-app-view *view-graph* workspace project stage menu-bar editor-tabs-split tool-tabs prefs localization)
```
vlaaad: Typically, if you have a `project` node, you don't need to also pass the `workspace` node there since you can get it using `(project/workspace project evaluation-context)` fn.

vlaaad (dismissed):
Code is good, but I think we should tidy it up a bit more before merging. I left comments about it.

Regarding behavior, I found 2 issues:
1. A second split sometimes is not visible after reopening:

   https://github.com/user-attachments/assets/e8543965-f67d-418e-8b1e-db8cc8cf02d6
3. I've got this exception after opening an app when I had a lot of windows open with scene views:
```
java.lang.NullPointerException: Cannot invoke "javafx.scene.Node.getParent()" because "p1__27799_SHARP_" is null
	at editor.ui$nodes_along_path_QMARK_$fn__27806.invoke(ui.clj:233)
	at clojure.lang.Iterate.first(Iterate.java:49)
	at clojure.lang.RT.first(RT.java:712)
	at clojure.core$first__5468.invokeStatic(core.clj:55)
	at clojure.core$take_while$fn__6005.invoke(core.clj:2923)
	at clojure.lang.LazySeq.force(LazySeq.java:50)
	at clojure.lang.LazySeq.realize(LazySeq.java:89)
	at clojure.lang.LazySeq.seq(LazySeq.java:106)
	at clojure.lang.Cons.next(Cons.java:41)
	at clojure.lang.RT.next(RT.java:733)
	at clojure.core$next__5470.invokeStatic(core.clj:64)
	at clojure.core$some.invokeStatic(core.clj:2718)
	at editor.ui$nodes_along_path_QMARK_.invokeStatic(ui.clj:228)
	at editor.ui$refresh_toolbars_BANG_.invokeStatic(ui.clj:2051)
	at editor.ui$refresh.invokeStatic(ui.clj:2065)
	at editor.app_view$refresh_menus_and_toolbars_BANG_.invokeStatic(app_view.clj:1957)
	at editor.app_view$make_app_view$fn__72262.invoke(app_view.clj:2094)
	at editor.ui$__GT_timer$fn__29273$fn__29274.invoke(ui.clj:2176)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$10(PlatformImpl.java:456)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$11(PlatformImpl.java:455)
```

editor/src/clj/editor/app_view.clj
JosephFerano: Ok, perfect, yes the reason I did it was because I thought it be writing to disk. If it's not expensive, I'll get rid of it.

editor/src/clj/editor/recent_files.clj
JosephFerano: Ah good catch, I missed that

editor/src/clj/editor/boot_open_project.clj
JosephFerano: Actually turns out I didn't even need this arg. I think I changed the code but I forgot to removed it. I'm going to be adding clj-kondo to my workflow soon to catch these issues.

editor/src/clj/editor/app_view.clj
JosephFerano: The reason I did it was that in batch operations when you were opening say 10 tabs at once, or calling tab close-all, then it would call prefs/set! for however many tabs were being opened/closed. It's still an optimization, but if you think it's really not worth it to introduce the atom + `compare-and-set!`, I'll remove it.

JosephFerano:
> Code is good

:man_dancing: 

> Regarding behavior, I found 2 issues:
> 
> 1. A second split sometimes is not visible after reopening:
>    

Oh that's interesting, I wonder if it's an order of execution thing? It never happened to me. Perhaps it's restoring the tabs first before restoring the pane positions?

> 2. I've got this exception after opening an app when I had a lot of windows open with scene views:

I'll try to reproduce, if you have any further info or the project you used, that would be helpful. Let me check first if I can get the exception though.

JosephFerano:
> Oh that's interesting, I wonder if it's an order of execution thing? It never happened to me. Perhaps it's restoring the tabs first before restoring the pane positions?

Yes, this seems to be exactly the problem. I was able to replicate, what's happening is that when I call the restore here;

```clojure
      (ui/run-later
        (app-view/restore-split-positions! scene prefs)
        (app-view/restore-hidden-panes! scene prefs)
          (app-view/restore-tabs-from-prefs! app-view prefs localization workspace project evaluation-context)))
```

This is calling this;

```clojure
(defn restore-split-positions! [^Scene scene prefs]
  (let [split-positions (stored-split-positions prefs)
        split-panes (existing-split-panes scene)]
    ;; The nested run-later fixes restore on Linux, by forcing an initial rendering pass. 
    (ui/run-later
      (;; do the thing))
```

There's a nested ui/run-later which is then likely causing a race condition, I thought those functions would run one after the other but clearly not the case. I'm thinking I would need a "post-editor-load" to be sure that all the run-laters have completed before triggering the restoring tabs. There's a possibility that the other exception, which I have yet to replicate, is also related.

JosephFerano:
I finally got the same exception

```
java.lang.NullPointerException: Cannot invoke "javafx.scene.Node.getParent()" because "p1__28609_SHARP_" is null
	at editor.ui$nodes_along_path_QMARK_$fn__28616.invoke(ui.clj:233)
	at clojure.lang.Iterate.first(Iterate.java:49)
	at clojure.lang.RT.first(RT.java:712)
	at clojure.core$first__5468.invokeStatic(core.clj:55)
	at clojure.core$take_while$fn__6005.invoke(core.clj:2923)
	at clojure.lang.LazySeq.force(LazySeq.java:50)
	at clojure.lang.LazySeq.realize(LazySeq.java:89)
	at clojure.lang.LazySeq.seq(LazySeq.java:106)
	at clojure.lang.Cons.next(Cons.java:41)
	at clojure.lang.RT.next(RT.java:733)
	at clojure.core$next__5470.invokeStatic(core.clj:64)
	at clojure.core$some.invokeStatic(core.clj:2718)
	at clojure.core$some.invoke(core.clj:2709)
	at editor.ui$nodes_along_path_QMARK_.invokeStatic(ui.clj:235)
	at editor.ui$nodes_along_path_QMARK_.invoke(ui.clj:228)
	at editor.ui$refresh_toolbars_BANG_.invokeStatic(ui.clj:2051)
	at editor.ui$refresh_toolbars_BANG_.invoke(ui.clj:2041)
	at editor.ui$refresh.invokeStatic(ui.clj:2068)
	at editor.ui$refresh.invoke(ui.clj:2060)
	at editor.app_view$refresh_menus_and_toolbars_BANG_.invokeStatic(app_view.clj:1972)
	at editor.app_view$refresh_menus_and_toolbars_BANG_.invoke(app_view.clj:1970)
	at editor.app_view$make_app_view$fn__102070.invoke(app_view.clj:2110)
	at editor.ui$__GT_timer$fn__30217$fn__30218.invoke(ui.clj:2176)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$0(PlatformImpl.java:424)
	at com.sun.glass.ui.gtk.GtkApplication._runLoop(Native Method)
	at com.sun.glass.ui.gtk.GtkApplication.lambda$runLoop$0(GtkApplication.java:240)
	at java.base/java.lang.Thread.run(Thread.java:1474)
```

matgis (approved):
Works great! Let's get this merged! ❤️

I pushed a commit with the following small fixes, just so you're aware of them:
* There was a `g/node-value` call inside the `restore-tabs-from-prefs!` function that didn't use the supplied `evaluation-context`.
* Moved the `evaluation-context` argument the end of the arglist of the `open-tabs-from-prefs` function. We typically put the `evaluation-context` last.
* Fixed formatting in a few places.

PR #11407, opened 2025-10-23
Check ENV and validate DM_SERVICE_PORT

### LINKED ISSUE #9813: Support engine port configuration via DM_SERVICE_PORT environment variable in editor
**Is your feature request related to a problem? Please describe:**
When running from the editor, where DM_SERVICE_PORT is set to "dynamic", the engine uses a random port, but the system firewall blocks that port. This is a problem for content-less build, where the editor needs to communicate with the engine.

**Describe the solution you'd like:**
Add a check for the DM_SERVICE_PORT environment variable:
1. If DM_SERVICE_PORT exists and contains a specific port number, use that port
2. If DM_SERVICE_PORT is not set, fall back to current behavior of using a random port

**Describe alternatives you've considered:**
Using a bob.jar with DM_SERVICE_PORT environment variable set to a specific port number

https://forum.defold.com/t/solved-problem-with-contentless-android-app-build/75743/4

### LINKED ISSUE #11334: Configure editor to follow DM_SERVICE_PORT
Hi

I’m running defold on Linux, the problem is variable `DM_SERVICE_PORT` is ignored when opening the editor

I’ve set export DM_SERVICE_PORT=8001. But when running the editor, the editor still runs on random port
```
util.http-server - {:line 376, :msg "Http server running", :local-url "http://localhost:44085"}
```
Because of firewall i need the editor to keep using the same port.

Could i request that the editor to honor the `DM_SERVICE_PORT` variable?

Thank you

JosephFerano:
Editor used to ignore `DM_SERVICE_PORT` and always use a dynamic port. Now the editor can read the port when the environment variable supplied.

Fixes #11334
Fixes #9813

### DIFF
--- a/editor/src/clj/editor/engine.clj
+++ b/editor/src/clj/editor/engine.clj
@@ -276,6 +276,13 @@
       (copy-dmengine-dependencies! engine-dir extender-platform)
       engine-file)))
 
+(defn- validate-service-port [port-str]
+  (try
+    (let [port (Integer/parseInt port-str)]
+      (when (<= 0 port 65535)
+        (str port)))
+    (catch Exception _ false)))
+
 (defn launch! [^File engine project-directory prefs debug? instance-index]
   (let [defold-log-dir (some-> (System/getProperty "defold.log.dir")
                                (File.)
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/engine.clj
```diff
@@ -276,6 +276,13 @@
       (copy-dmengine-dependencies! engine-dir extender-platform)
       engine-file)))
 
+(defn- validate-service-port [port-str]
+  (try
+    (let [port (Integer/parseInt port-str)]
+      (when (<= 0 port 65535)
+        (str port)))
+    (catch Exception _ false)))
```
vlaaad: Nitpick: I think, it would be better from a semantics perspective if this function returned `string or nil` instead of `string or nil or false`. You can simply omit `false` in the catch branch.

vlaaad (dismissed):
LGTM!

editor/src/clj/editor/engine.clj
JosephFerano: Yes good point

vlaaad (approved):

PR #11410, opened 2025-10-23
Include oneof member number in message payload

Jhonnyg:
### Technical changes
This PR fixes and issue where oneof constructs are used in protobuf message definitions. Basically, the old parser would mutate the static descriptors we generate when oneof values are set. See the comment for more details.

### DIFF
--- a/engine/ddf/src/ddf/ddf_load.cpp
+++ b/engine/ddf/src/ddf/ddf_load.cpp
@@ -179,7 +179,7 @@ namespace dmDDF
                     if (field->m_OneOfIndex != DDF_NO_ONE_OF_INDEX)
                     {
                         FieldDescriptor* field_non_const = (FieldDescriptor*) field;
-                        field_non_const->m_OneOfSet = 1;
+                        //field_non_const->m_OneOfSet = 1;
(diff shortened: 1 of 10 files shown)


engine/ddf/src/ddf/ddf_load.cpp
```diff
@@ -178,8 +178,7 @@ namespace dmDDF
 
                     if (field->m_OneOfIndex != DDF_NO_ONE_OF_INDEX)
                     {
-                        FieldDescriptor* field_non_const = (FieldDescriptor*) field;
-                        field_non_const->m_OneOfSet = 1;
+                        message->SetOneOf(desc, field);
```
Jhonnyg, line 181: This is basically where the issue is/was. In the old code, we mutated the FieldDescriptor which indeed should be static in memory. I.e, if we have a message containing a oneof block and we use that in more than one serialized message where different oneof values are set, we would effectively mark more than one of these fields as "in use". 

The fix now is that we actually store this information in the messages/structs, so we can keep track of which oneof member is set.

JCash (approved):

PR #11412, opened 2025-10-24
Repaint up to last visible minimap row

### LINKED ISSUE #11377: Code map not working properly under Linux
I have this bug, the code map on the right side of the code editor of the Defold IDE seems not to work properly under Linux (Honestly, I didn't check on Windows).

The first time I open a script file, only the code under the window of the code map is represented using the colors used in the editor, the other portion has grey lines of code that get the color as the window scrolls over the code.

**To Reproduce (REQUIRED)**
1. Open the Defold IDE.
2. Open a script code
3. Scroll the window of the code map.
4. You can see the window coloring the grey code

I expect the preview of the code map is rendered correctly since the opening of the code editor

  - Version 1.11.1

**Platforms (REQUIRED):**
 - Platforms: Linux
 - OS: Manjaro Linux
 - DE: KDE Plasma 6.3.6
 - Kernel: 6.17.1

Image

JosephFerano:
The minimap in the code editor wasn't syntax highlighting all the lines visible in larger files. Now the minimap will have syntax colors all the way through.

Fixes #11377

### DIFF
--- a/editor/src/clj/editor/code/view.clj
+++ b/editor/src/clj/editor/code/view.clj
@@ -3532,7 +3532,7 @@
     ;; Repaint canvas if needed.
     (when-not (identical? prev-canvas-repaint-info canvas-repaint-info)
       (g/user-data! view-node :canvas-repaint-info canvas-repaint-info)
-      (let [row (data/last-visible-row (:layout canvas-repaint-info))]
+      (let [row (data/last-visible-row (:minimap-layout canvas-repaint-info))]
         (repaint-canvas! canvas-repaint-info (get-valid-syntax-info resource-node canvas-repaint-info row))))
 
     (when editable


JosephFerano:
The fix involves getting syntax-info for lines up to the drawn minimap lines. I tried opening an 8k LOC file and didn't experience any slowdowns, so I believe it's doing it's job. It can potentially increase the the total number of lines it has to check syntax for by an order of magnitude (50-ish lines up to 500-ish lines), but it seems to get the job done without requiring run-later. I can still do this;

```clojure
    (when-not (identical? prev-canvas-repaint-info canvas-repaint-info)
      (ui/run-later
        (g/user-data! view-node :canvas-repaint-info canvas-repaint-info)
        (let [row (data/last-visible-row (:minimap-layout canvas-repaint-info))]
          (repaint-canvas! canvas-repaint-info (get-valid-syntax-info resource-node canvas-repaint-info row)))))
```

Didn't commit it but I also experimented with `ui/run-later` and didn't notice any major difference (assuming this is the correct way to run in the background).

vlaaad (approved):
LGTM! Works great! My test 150k line Lua file has the same performance characteristics both with and without this change.

matgis (approved):
Huh, that was a lot simpler than I would have expected. 😅 LGTM! ❤️

PR #11418, opened 2025-10-24
Localize some missed places in the editor UI

vlaaad:
We now localize:
- Find/Replace UI in the code editor.
- About dialog (bonus: Esc closes it now).
- Diff viewer title.


matgis (approved):
Ship it!

PR #11427, opened 2025-10-27
Added sound group mute functionality to C++ dmSDK

### LINKED ISSUE #8763: Add API to mute/unmute game sound from dmsdk
It would be useful to have an ability to mute/unmute game sound system in extensions. It would make using of many SDKs easier.
Most of the time it's related to showing ADS using external SDK:
- Poki SDK;
- any ADS SDK (admob, applovin, ironsource, unity ads etc)

A function in dmsdk to mute/unmute sound system

Suggest users in each manual to do it themselves integrating SDK

deurell:
Add functionality to mute sound groups from extensions.

Fixes #8763 

### Technical changes
- Expose sound muting functions to dmsdk
- Support both muting master and individual sound groups

### DIFF
--- a/engine/sound/src/dmsdk/sound/sound.h
+++ b/engine/sound/src/dmsdk/sound/sound.h
@@ -15,6 +15,8 @@
 #ifndef DMSDK_SOUND_SOUND_H
 #define DMSDK_SOUND_SOUND_H
 
+#include <dmsdk/dlib/hash.h>
+
 /*# Sound API documentation
  *
  * Functions for controlling the engine sound mixer from native extensions.
@@ -27,7 +29,6 @@
 
 namespace dmSound
 {
-    // Forward declaration supplied by the runtime.
     enum Result : int;
 
     /*# Set master mute state
@@ -38,6 +39,32 @@ namespace dmSound
      * @return result [type:Result] RESULT_OK on success
      */
     Result SetMasterMute(bool mute);
+
+    /*# Set mute state for a mixer group
+     * Temporarily mute or restore an individual mixer group.
+     *
+     * @name SetGroupMute
+     * @param group [type:dmhash_t] hash of the mixer group (e.g. `hash("master")`)
+     * @param mute [type:bool] `true` to mute, `false` to restore audio
+     * @return result [type:Result] RESULT_OK on success
+     */
+    Result SetGroupMute(dmhash_t group, bool mute);
+
+    /*# Toggle mute state for a mixer group
+     * Convenience toggle for `SetGroupMute`.
+     *
+     * @name ToggleGroupMute
+     * @param group [type:dmhash_t] hash of the mixer group (e.g. `hash("master")`)
+     * @return result [type:Result] RESULT_OK on success
+     */
+    Result ToggleGroupMute(dmhash_t group);
+
+    /*# Query group mute state
+     * @name IsGroupMuted
+     * @param group [type:dmhash_t] hash of the mixer group (e.g. `hash("master")`)
+     * @return muted [type:bool] `true` if the mixer group is muted
+     */
+    bool IsGroupMuted(dmhash_t group);
 
     /*# Toggle master mute
      * Toggle the master mixer group mute state.
--- a/engine/sound/src/dmsdk/sound/sound.h
+++ b/engine/sound/src/dmsdk/sound/sound.h
@@ -29,16 +29,30 @@
 
 namespace dmSound
 {
-    enum Result : int;
-
-    /*# Set master mute state
-     * Mute or unmute the master mixer group.
-     *
-     * @name SetMasterMute
-     * @param mute [type:bool] `true` to mute, `false` to restore audio
-     * @return result [type:Result] RESULT_OK on success
-     */
-    Result SetMasterMute(bool mute);
+    enum Result
+    {
+        RESULT_OK                 =  0,
+        RESULT_PARTIAL_DATA       =  1,
+        RESULT_OUT_OF_SOURCES     = -1,
+        RESULT_EFFECT_NOT_FOUND   = -2,
+        RESULT_OUT_OF_INSTANCES   = -3,
+        RESULT_RESOURCE_LEAK      = -4,
+        RESULT_OUT_OF_BUFFERS     = -5,
+        RESULT_INVALID_PROPERTY   = -6,
+        RESULT_UNKNOWN_SOUND_TYPE = -7,
+        RESULT_INVALID_STREAM_DATA= -8,
+        RESULT_OUT_OF_MEMORY      = -9,
+        RESULT_UNSUPPORTED        = -10,
+        RESULT_DEVICE_NOT_FOUND   = -11,
+        RESULT_OUT_OF_GROUPS      = -12,
+        RESULT_NO_SUCH_GROUP      = -13,
+        RESULT_NOTHING_TO_PLAY    = -14,
+        RESULT_INIT_ERROR         = -15,
+        RESULT_FINI_ERROR         = -16,
+        RESULT_NO_DATA            = -17,
+        RESULT_END_OF_STREAM      = -18,
+        RESULT_UNKNOWN_ERROR      = -1000,
+    };
 
     /*# Set mute state for a mixer group
      * Temporarily mute or restore an individual mixer group.
@@ -66,19 +80,6 @@ namespace dmSound
      */
     bool IsGroupMuted(dmhash_t group);
 
-    /*# Toggle master mute
-     * Toggle the master mixer group mute state.
-     *
-     * @name ToggleMasterMute
-     * @return result [type:Result] RESULT_OK on success
-     */
-    Result ToggleMasterMute();
-
-    /*# Query master mute state
-     * @name IsMasterMuted
-     * @return muted [type:bool] `true` if the mixer master group is muted
-     */
-    bool IsMasterMuted();
 }
 
 #endif // DMSDK_SOUND_SOUND_H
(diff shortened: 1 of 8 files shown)


engine/sound/src/dmsdk/sound/sound.h
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * @document
+ * @name Sound
+ * @namespace dmSound
+ * @language C++
+ */
+
+namespace dmSound
+{
+    enum Result : int;
```
JCash: We don't use this feature elsewhere, so this is a first.
While it is technically a C++11 feature, I'd prefer to keep even the C++ api's as "C like" as possible, to aid with keeping parsing/generating as easy as possible.

Is the feature fixing an issue?

engine/sound/src/dmsdk/sound/sound.h
```diff
@@ -0,0 +1,84 @@
...
+
+    /*# Set master mute state
+     * Mute or unmute the master mixer group.
+     *
+     * @name SetMasterMute
+     * @param mute [type:bool] `true` to mute, `false` to restore audio
+     * @return result [type:Result] RESULT_OK on success
+     */
+    Result SetMasterMute(bool mute);
```
JCash: Since we can use `SetGroupMute(hash("master") ...)` we should be able to skip this function?
(Every bit counts in keeping the engine small)

engine/sound/src/dmsdk/sound/sound.h
```diff
@@ -0,0 +1,84 @@
...
+     * @name ToggleMasterMute
+     * @return result [type:Result] RESULT_OK on success
+     */
+    Result ToggleMasterMute();
+
+    /*# Query master mute state
+     * @name IsMasterMuted
+     * @return muted [type:bool] `true` if the mixer master group is muted
+     */
+    bool IsMasterMuted();
```
JCash: Same with these functions, they feel a bit redundant imho.

engine/sound/src/sound.h
```diff
@@ -144,6 +144,12 @@ namespace dmSound
     Result AddGroup(const char* group);
     Result SetGroupGain(dmhash_t group_hash, float gain);
     Result GetGroupGain(dmhash_t group_hash, float* gain);
+    Result SetGroupMute(dmhash_t group_hash, bool mute);
+    Result ToggleGroupMute(dmhash_t group_hash);
+    bool IsGroupMuted(dmhash_t group_hash);
+    Result SetMasterMute(bool mute);
+    Result ToggleMasterMute();
+    bool IsMasterMuted();
```
JCash: Duplicates?

engine/sound/src/sound.h
JCash, line 21: Include `dmsdk/sound/sound.h` here so that our internal code doesn't have to change.

engine/sound/src/sound.cpp
```diff
@@ -457,12 +461,16 @@ namespace dmSound
         sound->m_GroupMap.SetCapacity(MAX_GROUPS * 2 + 1, MAX_GROUPS);
         for (uint32_t i = 0; i < MAX_GROUPS; ++i) {
             memset(&sound->m_Groups[i], 0, sizeof(SoundGroup));
+            sound->m_Groups[i].m_GainBeforeMute = 1.0f;
+            sound->m_Groups[i].m_IsMuted = false;
```
JCash: Isn't this already handled by GetOrCreateGroup() ?
I'm not even sure why the array is memset(), as it's just redundant code imho.

engine/sound/src/sound.cpp
```diff
@@ -908,6 +926,77 @@ namespace dmSound
...
+                if (group->m_GainParameter > 0.0f)
+                {
+                    group->m_GainBeforeMute = group->m_GainParameter;
+                }
+                group->m_IsMuted = true;
+                target_gain = 0.0f;
+            }
+            else
+            {
+                target_gain = group->m_GainBeforeMute > 0.0f ? group->m_GainBeforeMute : 1.0f;
```
JCash: Can `m_GainBeforeMute` ever be 0 here?

JCash (changes requested):
overall good, but some requests

engine/sound/src/dmsdk/sound/sound.h
deurell: Removed

engine/sound/src/dmsdk/sound/sound.h
deurell: Removed

engine/sound/src/dmsdk/sound/sound.h
deurell: And removed... :)

engine/sound/src/sound.cpp
deurell: Correct, removed.

engine/sound/src/sound.cpp
deurell: unneccecary check, removed.

engine/sound/src/dmsdk/sound/sound.h
```diff
@@ -0,0 +1,85 @@
...
+ * @document
+ * @name Sound
+ * @namespace dmSound
+ * @language C++
+ */
+
+namespace dmSound
+{
+    enum Result
+    {
```
JCash, line 58: Undocumented

engine/sound/src/sound_openal.cpp
```diff
@@ -343,11 +347,16 @@ namespace dmSound
...
+            sound->m_Groups[i].m_Gain = 1.0f;
+            sound->m_Groups[i].m_GainBeforeMute = 1.0f;
+            sound->m_Groups[i].m_IsMuted = false;
         }
 
         int master_index = GetOrCreateGroup("master");
         SoundGroup* master = &sound->m_Groups[master_index];
         master->m_Gain = master_gain;
+        master->m_GainBeforeMute = master_gain > 0.0f ? master_gain : 1.0f;
+        master->m_IsMuted = master_gain <= 0.0f;
```
JCash: Same comment as above.

engine/sound/src/sound.cpp
```diff
@@ -455,14 +459,13 @@ namespace dmSound
...
-        for (uint32_t i = 0; i < MAX_GROUPS; ++i) {
-            memset(&sound->m_Groups[i], 0, sizeof(SoundGroup));
-        }
 
         int master_index = GetOrCreateGroup("master");
         SoundGroup* master = &sound->m_Groups[master_index];
         master->m_GainParameter = master_gain;
         master->m_Gain.Reset(GainToScale(master->m_GainParameter));
+        master->m_GainBeforeMute = master->m_GainParameter > 0.0f ? master->m_GainParameter : 1.0f;
+        master->m_IsMuted = master->m_GainParameter <= 0.0f;
```
JCash: This is the Initialize() function, isn't it enough to just set it to `master->m_GainParameter` + `false`?

engine/sound/src/sound.cpp
```diff
@@ -891,6 +894,16 @@ namespace dmSound
...
         group->m_GainParameter = gain;
+        if (gain > 0.0f)
+        {
+            group->m_GainBeforeMute = gain;
+            group->m_IsMuted = false;
+        }
+        else
+        {
+            group->m_IsMuted = true;
+        }
```
JCash: Overall, I feel that the if-statements surrounding this logic (>0.0f) is a bit redundant, and hinders readability. `gain = 0` is a valid value.

E.g. what about:
```c++
group->m_GainBeforeMute = gain;
group->m_IsMuted = group->m_GainBeforeMute > 0.0f;
```

engine/sound/src/sound.cpp
```diff
@@ -181,6 +181,8 @@ namespace dmSound
         dmhash_t m_NameHash;
         Value    m_Gain;
         float    m_GainParameter;
+        float    m_GainBeforeMute;
+        bool     m_IsMuted;
```
JCash: The more I think about it, we shouldn't need the `m_GainBeforeMute` variable?

In our implementation, muting isn't stopping the sound, so the gain may be animated up/down during gameplay. Muting is an override. Setting gain from Lua shouldn't unmute the sounds?

I believe, In terms of the final mixing, we should look at the m_IsMuted, and then only mix in zeroes. See `bool is_muted = dmSound::IsMuted(instance);` in MixInstance()

engine/sound/src/sound.cpp
```diff
@@ -891,6 +894,16 @@ namespace dmSound
         SoundGroup* group = &sound->m_Groups[*index];
         group->m_Gain.Set(GainToScale(gain), reset);
         group->m_GainParameter = gain;
+        if (gain > 0.0f)
+        {
+            group->m_GainBeforeMute = gain;
+            group->m_IsMuted = false;
+        }
```
JCash: If this is being called from Lua, it will effectively unmute the group?
I don't feel that it's the intended behavior?

JCash (changes requested):
I feel we've missed some things during the design phase of this task.
Let's discuss in slack.

engine/sound/src/sound.cpp
```diff
@@ -1291,14 +1331,37 @@ namespace dmSound
...
+                s_ScratchStorage.SetCapacity(needed);
+            }
+            s_ScratchStorage.SetSize(needed);
+
+            float* scratch_buffers[SOUND_MAX_MIX_CHANNELS];
+            float* scratch_base = s_ScratchStorage.Begin();
+            for (uint32_t c = 0; c < SOUND_MAX_MIX_CHANNELS; ++c)
+            {
+                scratch_buffers[c] = scratch_base + c * mix_count;
+            }
```
JCash: I don't know why it's needed?

If the `is_muted` flag is set (`bool is_muted = dmSound::IsMuted(instance);`), then we read/forward the sounds, and set the read data as all zeroes. So muted sounds are already supported.

deurell: I got into trouble with skipping and a glitch when not doing it. I'll do another run and see if i can get rid of it. I'd love to remove it.

deurell: Removed it.

JCash (approved):
lgtm

PR #11431, opened 2025-10-28
Added support for disabling rendering with sys.set_render_enabled()

### LINKED ISSUE #8309: Pause rendering for battery saving
My mobile app does not change anything on the screen most of the time, but sometimes it needs 60 fps to render a smooth animation. This no-change situation could be taken advantage of to save CPU/GPU cycles and thus save battery.

When forcing a lower frame rate, the app starts to become slow to respond to user input.

In order for the app to remain responsive with a zero frame rate, the input system (touch) has to be decoupled from the rendering loop.

The new battery saving feature can either be an automatic system which detects that nothing changes, and/or a manual system where the the frame rate can be set to zero, or rendering paused.

If an automatic change detection is used, it should work with Rive.

JCash:
This allows games that know they don't need to refresh the frame too often to manually enable or disable the rendering.
When disabled, the screen isn't cleared. In fact, no render code path is invoked, avoiding calculations on both CPU+GPU.

_NOTE:_ Regular updates are still run, so if you load/unload collections, these resources will be uploaded to the GPU if required.

```lua
sys.set_render_enabled(false) -- disable rendering
sys.set_render_enabled(true) -- enable rendering
``` 

Test Project:
[RenderDisable.zip](https://github.com/user-attachments/files/23205046/RenderDisable.zip)

Fixes https://github.com/defold/defold/issues/8309

### Technical details

Design: https://docs.google.com/document/d/1Qh0mQJwezarKDLZuYeFRbecGEOlVkkMmmLu73fgyfIQ/edit?usp=sharing

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/engine/engine/src/engine.cpp
+++ b/engine/engine/src/engine.cpp
@@ -285,7 +285,8 @@ namespace dmEngine
     }
 
     Stats::Stats()
-    : m_FrameCount(0)
+    : m_UpdateCount(0)
+    , m_RenderCount(0)
     , m_TotalTime(0.0f)
     {
 
@@ -2057,6 +2058,7 @@ bail:
                 }
 
                 dmGraphics::Flip(engine->m_GraphicsContext);
+                ++engine->m_Stats.m_RenderCount;
 
                 RecordData* record_data = &engine->m_RecordData;
                 if (record_data->m_Recorder)
@@ -2082,7 +2084,7 @@ bail:
         }
         ProfileFrameEnd(profile);
 
-        ++engine->m_Stats.m_FrameCount;
+        ++engine->m_Stats.m_UpdateCount;
         engine->m_Stats.m_TotalTime += dt;
     }
 
@@ -2389,7 +2391,7 @@ bail:
 
     uint32_t GetFrameCount(HEngine engine)
     {
-        return engine->m_Stats.m_FrameCount;
+        return engine->m_Stats.m_UpdateCount;
     }
 
     void GetStats(HEngine engine, Stats& stats)
--- a/engine/engine/src/engine.cpp
+++ b/engine/engine/src/engine.cpp
@@ -62,6 +62,7 @@
 #include "engine_service.h"
 #include "engine_version.h"
 #include "physics_debug_render.h"
+#include "script/script_engine.h"
 
 #ifdef __EMSCRIPTEN__
     #include <emscripten/emscripten.h>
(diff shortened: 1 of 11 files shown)


engine/engine/src/script/script_engine.cpp
```diff
@@ -0,0 +1,91 @@
...
+        return DM_LUA_ERROR("Expected boolean as first argument");
+    }
+
+    dmEngine::SetRenderEnable(g_Engine, enable);
+    return 0;
+}
+
+static const luaL_reg EngineSys_methods[] =
+{
+    {"set_render_enable", EngineSys_SetRenderEnable},
```
britzl: Shouldn't it be `set_render_enabled()`?

In other APIs we use 'enabled', not 'enable'. For example gui.set_enabled() and model.set_mesh_enabled()

engine/engine/src/engine.cpp
```diff
@@ -1930,9 +1949,11 @@ namespace dmEngine
 
                 dmSound::Update();
 
-                // Don't render while iconified
-                if (!dmGraphics::GetWindowStateParam(engine->m_GraphicsContext, dmPlatform::WINDOW_STATE_ICONIFIED)
-                    && !dmRender::IsRenderPaused(engine->m_RenderContext))
+                // Don't render while iconified, or when device is lost, or when manually disabled
+                bool skip_render = dmGraphics::GetWindowStateParam(engine->m_GraphicsContext, dmPlatform::WINDOW_STATE_ICONIFIED)
+                                || dmRender::IsRenderPaused(engine->m_RenderContext)
+                                || !render_enabled;
```
britzl, line 1956: I don't suppose we want to merge these two concepts? Or do we want to reserve dmRender::IsRenderPaused() and dmRender::SetRenderPaused() for when the context is lost?

britzl (commented):
See code comments

engine/engine/src/script/script_engine.cpp
JCash: Yes, it's a typo. I searched the other functions and still managed to misspell it :/

engine/engine/src/engine.cpp
JCash: I thought about it too, but I'm not sure. The pause concept is a boolean inside dmRender, that controls the engine, which feels a bit wrong. Not sure it should be called "paused" either. We can improve on it later if we feel like it.

engine/engine/src/script/script_engine.cpp
britzl: Still enable, not enabled

britzl (approved):

PR #11438, opened 2025-10-29
Editor: Fix page count mismatch detection and error message for sprites

matgis:
Fix issue where paged Materials were erroneously detected as unpaged when inspecting Sprite properties.

### Technical changes

Also fixes an unreleased regression where the string `editor.localization.SimpleMessage` appears in the property error tooltip when there is a page count mismatch between the assigned Atlas and Material of a Sprite.

### DIFF
--- a/editor/src/clj/editor/sprite.clj
+++ b/editor/src/clj/editor/sprite.clj
@@ -276,7 +276,8 @@
 (g/defnk produce-build-targets [_node-id resource textures texture-binding-infos default-animation material material-attribute-infos material-max-page-count material-samplers material-shader blend-mode size-mode manual-size slice9 offset playback-rate vertex-attribute-bytes vertex-attribute-overrides]
   (g/precluding-errors
     (let [sampler-name->texture-binding-info (coll/pair-map-by :sampler texture-binding-infos)
-          is-paged-material (shader/is-using-array-samplers? material-shader)
+          is-paged-material (and (shader/shader-lifecycle? material-shader)
+                                 (shader/is-using-array-samplers? material-shader))
           unassigned-default-animation-error (validation/prop-error :fatal _node-id :default-animation validation/prop-empty? default-animation "Default Animation")]
       (concat
         ;; Validate material assignment.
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/sprite.clj
```diff
@@ -276,7 +276,8 @@
 (g/defnk produce-build-targets [_node-id resource textures texture-binding-infos default-animation material material-attribute-infos material-max-page-count material-samplers material-shader blend-mode size-mode manual-size slice9 offset playback-rate vertex-attribute-bytes vertex-attribute-overrides]
   (g/precluding-errors
     (let [sampler-name->texture-binding-info (coll/pair-map-by :sampler texture-binding-infos)
-          is-paged-material (shader/is-using-array-samplers? material-shader)
+          is-paged-material (and (shader/shader-lifecycle? material-shader)
+                                 (shader/is-using-array-samplers? material-shader))
```
matgis, line 280: Unrelated: Added `shader/shader-lifecycle?` call for consistency with similar check elsewhere in the file.

editor/src/clj/editor/sprite.clj
```diff
@@ -381,7 +382,7 @@
 
 (g/defnk produce-properties [_declared-properties _node-id default-animation material-attribute-infos material-max-page-count material-samplers material-shader texture-binding-infos vertex-attribute-overrides]
   (let [extension (workspace/resource-kind-extensions (project/workspace (project/get-project _node-id)) :atlas)
-        is-paged-material (and (shader/shader-lifecycle? material-samplers)
+        is-paged-material (and (shader/shader-lifecycle? material-shader)
```
matgis, line 385: Fix: Previously, `is-paged-material` would always be `false` here because `material-samplers` is certainly not a `ShaderLifecycle`. 😬

editor/src/clj/editor/sprite.clj
```diff
@@ -396,8 +397,10 @@
              sort
              (mapv
                (fn [[sampler-name order :as name+order]]
-                 (let [property-key (keyword (str "__sampler__" sampler-name "__" order))
-                       label (if (= 1 (count combined-name+indices)) (localization/message "property.image") sampler-name)]
+                 (let [has-single-sampler (= 1 (count combined-name+indices))
+                       property-key (keyword (str "__sampler__" sampler-name "__" order))
+                       property-name (if has-single-sampler "Image" sampler-name)
+                       label (if has-single-sampler (localization/message "property.image") sampler-name)]
```
matgis, line 403: We currently don't localize the error tooltips from invalid property values. Concatenating the `label`, which is an `editor.localization.SimpleMessage` directly into the error string caused the string `editor.localization.SimpleMessage` to appear in the tooltip. We now use the hard-coded english property name instead. This is how it still is everywhere else in the codebase until we have a good way of localizing this category of strings.

AGulev (approved):
lgtm!

PR #11442, opened 2025-10-29
Produce user-friendly build errors from invalid Tile Sources

### LINKED ISSUE #11166: Crash when setting Tile Margin to a negative value
### Steps to reproduce

Set Tile Margin on a tilesource to -1

<hr/>

<table>
<tr><td>Build time</td><td>2025-08-04T15:29:15.191640</td></tr>
<tr><td>Defold editor sha</td><td>1aafd0a262ff40214ed7f51302d92fa587c607ef</td></tr>
<tr><td>Defold engine sha</td><td>1aafd0a262ff40214ed7f51302d92fa587c607ef</td></tr>
<tr><td>Defold version</td><td>1.10.4</td></tr>
<tr><td>Error</td><td><a href='https://sentry.io/organizations/defold/issues/?query=id%3A"f2ad2d1e088c4c09aeb0d90e0cfd86c5"'>f2ad2d1e088c4c09aeb0d90e0cfd86c5</a></td></tr>
<tr><td>GPU</td><td>AMD Radeon RX 5700 XT</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 Compatibility Profile Context 25.8.1.250723</td></tr>
<table>

matgis:
* All invalid properties in Tile Sources now generate clickable build errors.
* Highlight Tile Source Animations with invalid properties in the Outline View.
* Render errors in the Scene View will now report the source file when the error originates from a substructure node.
* Fixed Exception in the editor when setting the Tile Source Tile Margin to a negative value.
* Tile Source Tile Width and Height must be positive.
* The error tooltip for the Sprite Default Animation field now refers to the correct Image property label.

Fixes #11166

### Technical changes
We will not attempt to render anything in the Scene View until the user has assigned an Image to the Tile Source. It seems the `ToolController` we use for Tile Sources isn't well-behaved when there is no Image assigned, so we'll hopefully eliminate Exceptions from it as a welcome side-effect.

### DIFF
--- a/editor/src/clj/editor/scene.clj
+++ b/editor/src/clj/editor/scene.clj
@@ -103,7 +103,7 @@
                     {:value value}))))
 
 (defn- get-resource-name [node-id]
-  (when-let [resource (resource-node/as-resource node-id)]
+  (when-let [resource (some-> node-id resource-node/owner-resource)]
     (resource/resource-name resource)))
 
 (defn- error-message-lines
--- a/editor/src/clj/editor/sprite.clj
+++ b/editor/src/clj/editor/sprite.clj
@@ -508,8 +508,9 @@
             (dynamic error (g/fnk [_node-id textures primary-texture-binding-info default-animation]
                              (when (pos? (count textures))
                                (or (validation/prop-error :info _node-id :default-animation validation/prop-empty? default-animation "Default Animation")
-                                   (let [{:keys [anim-data sampler]} primary-texture-binding-info]
-                                     (validation/prop-error :fatal _node-id :default-animation validation/prop-anim-missing-in? default-animation anim-data sampler))))))
+                                   (let [{:keys [anim-data sampler]} primary-texture-binding-info
+                                         image-property-label (if (= 1 (count textures)) "Image" sampler)]
+                                     (validation/prop-error :fatal _node-id :default-animation validation/prop-anim-missing-in? default-animation anim-data image-property-label))))))
             (dynamic edit-type (g/fnk [anim-ids] (properties/->choicebox anim-ids))))
 
   (property material resource/Resource ; Default assigned in load-fn.
--- a/editor/src/clj/editor/tile_source.clj
+++ b/editor/src/clj/editor/tile_source.clj
@@ -221,7 +221,39 @@
 
 (defn- prop-tile-range? [max v name]
   (when (or (< v 1) (< max v))
-    (format "%s is outside the tile range (1-%d)" name max)))
+    (format "'%s' is outside the tile range (1-%d)" name max)))
+
+(defn- validate-animation-id [node-id id]
+  (validation/prop-error :fatal node-id :id validation/prop-empty? id "Id"))
+
+(defn- validate-animation-start-tile [tile-count node-id start-tile]
+  (validation/prop-error :fatal node-id :start-tile (partial prop-tile-range? tile-count) start-tile "Start Tile"))
+
+(defn- validate-animation-end-tile [tile-count node-id end-tile]
+  (validation/prop-error :fatal node-id :end-tile (partial prop-tile-range? tile-count) end-tile "End Tile"))
+
+(defn- validate-animation-fps [node-id fps]
+  (validation/prop-error :fatal node-id :fps validation/prop-negative? fps "FPS"))
+
+(def ^:private protobuf-animation-defaults
+  "Default field values declared in the Tile$Animation protobuf message. Fields
+  that match these values will be excluded from the saved project files."
+  (protobuf/default-value Tile$Animation))
+
+(defn- animation-ddf-errors [tile-count node-id animation-ddf]
+  {:pre [(g/node-id? node-id)
+         (map? animation-ddf)]} ; Tile$Animation in map format.
+  (->> [[:id validate-animation-id]
+        [:start-tile (partial validate-animation-start-tile tile-count)]
+        [:end-tile (partial validate-animation-end-tile tile-count)]
+        [:fps validate-animation-fps]]
+       (keep (fn [[pb-field validation-fn]]
+               (let [pb-value (get animation-ddf pb-field ::not-found)
+                     pb-value (if (= ::not-found pb-value)
+                                (get protobuf-animation-defaults pb-field)
+                                pb-value)]
+                 (validation-fn node-id pb-value))))
+       (not-empty)))
 
 (defn render-animation
   [^GL2 gl render-args renderables n]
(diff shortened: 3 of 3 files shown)


editor/src/clj/editor/tile_source.clj
```diff
@@ -318,6 +318,28 @@
...
+
+(defn- validate-tile-spacing [node-id tile-spacing]
+  (validation/prop-error :fatal node-id :tile-spacing validation/prop-negative? tile-spacing "Tile Spacing"))
+
+(defn- validate-extrude-borders [node-id extrude-borders]
+  (validation/prop-error :fatal node-id :extrude-borders validation/prop-negative? extrude-borders "Extrude Borders"))
+
+(defn- validate-inner-padding [node-id inner-padding]
+  (validation/prop-error :fatal node-id :inner-padding validation/prop-negative? inner-padding "Inner Padding"))
+
```
matgis, line 377: Broke these property validation checks out into functions so they can be used from the `tile-source-attributes` output as well. The `tile-source-attributes` will now produce an `ErrorValue` if any of these properties are invalid. This results in clickable build errors with friendlier error messages from improperly configured Tile Sources.

editor/src/clj/editor/tile_source.clj
```diff
@@ -543,9 +564,15 @@
...
-      (validation/prop-error :fatal _node-id :image validation/prop-resource-not-exists? image-resource "Image")
+  [_node-id image-resource image-size tile-width tile-height tile-margin tile-spacing extrude-borders inner-padding collision-size sprite-trim-mode image-dim-error tile-width-error tile-height-error]
+  (or (g/flatten-errors
+        (validate-image-resource _node-id image-resource)
+        (validate-tile-width _node-id tile-width)
+        (validate-tile-height _node-id tile-height)
+        (validate-tile-margin _node-id tile-margin)
+        (validate-tile-spacing _node-id tile-spacing)
+        (validate-extrude-borders _node-id extrude-borders)
+        (validate-inner-padding _node-id inner-padding))
```
matgis, line 610: `flatten-errors` returns `nil` if there are no errors, or an aggregate `ErrorValue` containing all the errors if one or more of these functions return an `ErrorValue`.

JosephFerano (commented):
image

So this seems to work, I can't run the game. I tested on 1.11.1 and it was letting me run the game despite the errors, so this fix seems to have done it's job.

However, I'm seeing some odd behavior

image

This was present before, so it's not a regression. It's hard for me to make out what the problem is, I have to lower the width and height to make this error go away. If this is how it's supposed to behave, I think it would help if the errors were clearer but I suspect this might not be being handled properly because of the "unknown" in the error message.

Also, to note, I tried reproducing the original exception from #11166, but never got the exception. Not even on 1.10.4

matgis:
Hmm, nice find. It is a separate bug, but I'll have a look! It would be nice to address it as part of this PR as well.

Regarding the exception from #11166, I was able to reproduce it. I think I had to interact with the viewport in some way, I can't remember exactly, but it should not be able to happen anymore after this fix.

editor/src/clj/editor/scene.clj
```diff
@@ -103,7 +103,7 @@
                     {:value value}))))
 
 (defn- get-resource-name [node-id]
-  (when-let [resource (resource-node/as-resource node-id)]
+  (when-let [resource (some-> node-id resource-node/owner-resource)]
```
matgis, line 106: This ensures we report the source file of an error in the Scene View when the error originates from a substructure node and not the resource node itself.

editor/src/clj/editor/sprite.clj
```diff
@@ -508,8 +508,9 @@
             (dynamic error (g/fnk [_node-id textures primary-texture-binding-info default-animation]
                              (when (pos? (count textures))
                                (or (validation/prop-error :info _node-id :default-animation validation/prop-empty? default-animation "Default Animation")
-                                   (let [{:keys [anim-data sampler]} primary-texture-binding-info]
-                                     (validation/prop-error :fatal _node-id :default-animation validation/prop-anim-missing-in? default-animation anim-data sampler))))))
+                                   (let [{:keys [anim-data sampler]} primary-texture-binding-info
+                                         image-property-label (if (= 1 (count textures)) "Image" sampler)]
+                                     (validation/prop-error :fatal _node-id :default-animation validation/prop-anim-missing-in? default-animation anim-data image-property-label))))))
```
matgis, line 513: When there is only one texture sampler in the Material referenced by the Sprite, the property label is "Image". The error tooltip for the Sprite Default Animation field now refers to the correct Image property label.

JosephFerano:
> Hmm, nice find. It is a separate bug, but I'll have a look! It would be nice to address it as part of this PR as well.

Sounds good, ping me and I'll test again.

> Regarding the exception from #11166, I was able to reproduce it. I think I had to interact with the viewport in some way, I can't remember exactly, but it should not be able to happen anymore after this fix.

:+1:

editor/src/clj/editor/tile_source.clj
```diff
@@ -221,7 +221,39 @@
 
 (defn- prop-tile-range? [max v name]
   (when (or (< v 1) (< max v))
-    (format "%s is outside the tile range (1-%d)" name max)))
+    (format "'%s' is outside the tile range (1-%d)" name max)))
```
matgis, line 224: Quoted the property label to be consistent with the other error messages.

editor/src/clj/editor/tile_source.clj
```diff
@@ -308,16 +342,39 @@
...
   (input gpu-texture g/Any)
 
-  (output node-outline outline/OutlineData :cached (g/fnk [_node-id id]
+  (output node-outline outline/OutlineData :cached (g/fnk [_node-id ddf-message id tile-count]
                                                      {:node-id _node-id
                                                       :node-outline-key id
                                                       :label id
-                                                      :icon animation-icon}))
+                                                      :icon animation-icon
+                                                      :outline-error? (some? (animation-ddf-errors tile-count _node-id ddf-message))}))
```
matgis, line 350: Highlight Tile Source Animations with invalid properties in the Outline View.

editor/src/clj/editor/tile_source.clj
```diff
@@ -1024,7 +1088,8 @@
         inner-padding :inner-padding
         sprite-trim-mode :sprite-trim-mode))))
 
-(def ^:private default-animation
+(def ^:private new-animation-defaults
+  "Default field values for added Tile$Animation instances."
```
matgis, line 1092: Renamed this and added a clarifying comment since we now also have `protobuf-animation-defaults`.

JosephFerano (approved):
image

Yes, now it's reporting the correct resource. I still find the error message a little confusing, but I guess I understand what it's saying. For example, why are they repeated?

matgis:
Maybe we could group the errors by their resource origin? Anyway, I'm going to merge this for now, but we could certainly improve things further!

PR #11445, opened 2025-10-30
Added engine throttle performance Lua api

### LINKED ISSUE #8309: Pause rendering for battery saving

JCash:
This api allows the developer to skip engine updates altogether, except input detection.
* No update or rendering will occur once the engine is throttled
* Any input will wake up the engine again
* The engine will be throttled again after the cooldown period

Api:
```lua
sys.set_engine_throttle(false) -- No throttling, full engine updates + rendering (default)
sys.set_engine_throttle(true, 1.0) -- Throttling with a cooldown. Engine will wake up for 1 second after each input
sys.set_engine_throttle(true, 0.0) -- Throttling with a cooldown. Engine will wake up for 1 frame after each input.
```

Fixes https://github.com/defold/defold/issues/8309

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


engine/engine/src/engine.cpp
```diff
@@ -1860,6 +1884,11 @@ namespace dmEngine
 #endif
                 dmJobThread::Update(engine->m_JobThreadContext, jobthread_max_time_us);
 
+                {
+                    DM_PROFILE("Resource");
+                    dmResource::UpdateFactory(engine->m_Factory); // Process Reload event
+                }
```
JCash, line 1895: This function only processes the "Reload" event, any jobs it has are handled by the jobthread

engine/engine/src/engine.cpp
```diff
@@ -1950,11 +1979,9 @@ namespace dmEngine
...
                 dmSound::Update();
 
-                // Don't render while iconified, or when device is lost, or when manually disabled
-                bool skip_render = dmGraphics::GetWindowStateParam(engine->m_GraphicsContext, dmPlatform::WINDOW_STATE_ICONIFIED)
-                                || dmRender::IsRenderPaused(engine->m_RenderContext)
-                                || !render_enabled;
-                if (!skip_render)
+                // Don't render while iconified
+                if (!dmGraphics::GetWindowStateParam(engine->m_GraphicsContext, dmPlatform::WINDOW_STATE_ICONIFIED)
+                    && !dmRender::IsRenderPaused(engine->m_RenderContext))
```
JCash, line 1989: Revert from previous change.

engine/hid/src/hid.cpp
```diff
@@ -574,4 +574,17 @@ namespace dmHID
...
+        HashState64 state;
+        dmHashInit64(&state, false);
+        dmHashUpdateBuffer64(&state, &context->m_Keyboards[0], MAX_KEYBOARD_COUNT * sizeof(Keyboard));
+        dmHashUpdateBuffer64(&state, &context->m_Mice[0], MAX_MOUSE_COUNT * sizeof(Mouse));
+        dmHashUpdateBuffer64(&state, &context->m_Gamepads[0], MAX_GAMEPAD_COUNT * sizeof(Gamepad));
+        dmHashUpdateBuffer64(&state, &context->m_TouchDevices[0], MAX_TOUCH_DEVICE_COUNT * sizeof(TouchDevice));
+        dmHashUpdateBuffer64(&state, &context->m_TextPacket, sizeof(TextPacket));
+        dmHashUpdateBuffer64(&state, &context->m_AccelerationPacket, sizeof(context->m_AccelerationPacket));
+        return dmHashFinal64(&state);
+    }
```
JCash, line 589: I found this to be the easiest way to detect if anything has changed between two updates.

engine/engine/src/engine.cpp
```diff
@@ -1784,6 +1790,24 @@ namespace dmEngine
...
+            return false;
+
+        // If cooldown is 0, we want 1 frame of update
+        bool skip = engine->m_ThrottleCooldown > engine->m_ThrottleCooldownMax;
+
+        engine->m_ThrottleCooldown += dt;
+
+        // We have new input, so reset the cooldown
+        if (has_input)
+            engine->m_ThrottleCooldown = 0;
```
britzl, line 1804: Might as well do this check early and return false if we have input

engine/engine/src/script/script_engine.cpp
```diff
@@ -27,46 +27,57 @@ namespace dmEngine
     dmEngine::HEngine g_Engine = 0;
 
 /*#
- *
- * @name sys.set_render_enabled
- * @param enable [type:bool] true if rendering should be enabled
+ * Enables engine throttling.
```
britzl, line 30: I think we need a bit more documentation here. What does it mean to throttle the engine? We do write "it skips the entire update+render loop" which is good, but maybe we should clarify that timers will not fire, http requests will continue in the background but not invoke callbacks, sounds will play(?) (and what about streaming?).

britzl (changes requested):
See comments about documentation and early out if we have input

britzl (dismissed):

PR #11451, opened 2025-11-02
MSL (metal shading language) support in shader pipeline

Jhonnyg:
This PR adds support for cross-compiling shaders into MSL via the shader pipeline as well as bob via the `--debug-output-msl` bob flag (they are not automatically included). This also updates spirv-cross to the latest commit.


JCash (approved):
nice!

PR #11457, opened 2025-11-03
Localize resource type labels

vlaaad:
Related to #5453


editor/src/clj/editor/font.clj
```diff
@@ -795,7 +795,6 @@
   (input material-shader ShaderLifecycle)
   (input font-resource-map g/Any)
 
-  (output outline g/Any :cached (g/fnk [_node-id] {:node-id _node-id :label "Font" :icon font-icon}))
```
vlaaad, line 798: Unused output. The actual output name for the outline is `node-outline`.

editor/src/clj/editor/game_project.clj
```diff
@@ -210,9 +211,6 @@
                          :message (ex-message error)
                          :severity :fatal}))))))
 
-  (output outline g/Any :cached
-          (g/fnk [_node-id] {:node-id _node-id :label "Game Project" :icon game-project-icon}))
```
vlaaad, line 214: Unused output. The actual output name for the outline is `node-outline`.

matgis (approved):
LGTM!

Just a reminder that we have a bunch of extensions registering resource types as well that we need to get to eventually.

PR #11468, opened 2025-11-06
Editor: Restored tabs right-click context menuu bug fixed

### LINKED ISSUE #11467: Right clicking fails on restored tabs
With the new Save/Restore Open Tabs feature, a bug was introduced where if a tab was opened during the restoring process, the right click context menu now appears empty.

JosephFerano:
This PR moves a selection process that syncs some state so that the context menu doesn't fail

Fixes #11467


vlaaad (approved):
Works great, good catch!

PR #11477, opened 2025-11-08
Add Move Step tab with regular and Shift move step options

ufgo:
This pull request introduces a new "Move Step" tab in the Editor preferences.

Users can now configure two separate movement step values:
* Regular move step – used for standard object movement.
* Shift move step – used when holding the Shift key to move objects with a different step size.

This feature improves precision and flexibility when positioning objects in the scene.

https://github.com/user-attachments/assets/1a2c0215-1098-4ed4-8eb4-7f718f62ecd3

### DIFF
--- a/editor/resources/localization/en.editor_localization
+++ b/editor/resources/localization/en.editor_localization
@@ -507,7 +507,7 @@ prefs.tab.general = General
 prefs.tab.code = Code
 prefs.tab.extensions = Extensions
 prefs.tab.tools = Tools
-prefs.tab.move-step = MoveStep
+prefs.tab.move-step = Move Step
 prefs.tab.dev = Dev
 prefs.tab.keymap = Keymap
 
@@ -549,7 +549,7 @@ prefs.help.tools.ios-deploy-path = Path to ios-deploy command that might be used
 
 # MoveStep tab
 prefs.label.move-step.move-step = Move Step
-prefs.label.move-step.fine-move-step = Fine Move Step
+prefs.label.move-step.fine-move-step = Shift move step
 
 # Dev tab
 prefs.label.dev.custom-engine = Custom Engine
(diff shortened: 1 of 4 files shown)


Demonevil1:
Nice, plan to add rotation (default  degrees 15%) and scale ?

matgis:
While the feature is nice, it is implemented here on top of the Move Whole Pixels setting, which is intended to keep finely positioned 2D objects on the same relative sub-pixel offset when moved. The result is that you're not really snapping objects to a grid, but simply moving them in relative increments from their original position. This feature was intended to work in conjunction with snapping, not as an implementation of it. Unfortunately, it sort of behaves like a relative snapping feature for 3D objects since there the unit scale system is comparatively tiny to the pixel unit scale used in 2D scenes, which confuses things.

We understand snapping is a crucial feature that is sorely missing from the Scene View, but we want it to work more in line with how typical 2D and 3D authoring applications do it, which is a bit more involved. We want a system that supports various kinds of snapping, not only snapping the center point of an object to a grid. We also want these settings easily available from the toolbar, so the user does not have to open the Preferences to change it. In many ways, it is as much a UX design task as it is a programming task. We have a lot of ideas around this, and in fact had a member of the editor team working on it, but unfortunately they left before they could complete the task.

We'd prefer to revive and continue their work on snapping once we can dedicate the resources, so I'm closing this PR. Please understand, it's just very difficult for us to coordinate with outside contributors on this until we have the relevant systems in place.

ufgo:
Thanks for the detailed feedback! I completely understand your reasoning and appreciate the clarification about how the Move Whole Pixels setting was intended to work. I’m glad to hear snapping is already on your roadmap — it’s definitely an important feature for many Defold users. Looking forward to seeing it in future editor updates!

PR #11487, opened 2025-11-10
Localize editor scripts and bundle dialogs

vlaaad:
Technical notes:
Related to #5453


editor/scripts/bundle.py
```diff
@@ -368,7 +368,7 @@ def check_reflections(jdk_path):
         exit(1)
 
 def write_docs(docs_dir, jdk_path=None):
-    invoke_lein(['with-profile', '+docs', 'run', '-m', 'editor.docs', docs_dir], jdk_path)
+    invoke_lein(['with-profile', 'docs', 'run', '-m', 'editor.docs', docs_dir], jdk_path)
```
vlaaad, line 371: Unrelated: we had internal dev-only editor script documentation (`editor.bundle.*`) visible to the users: https://defold.com/ref/stable/editor-lua/#editor.bundle.assoc:table-key-value

editor/src/clj/editor/localization.clj
```diff
@@ -132,25 +130,24 @@
     (.format ^MessagePattern v state)
     (str v)))
 
-(defonce/record LocalizationState [locale bundles listeners messages available-locales list-and list-or date]
+(defonce/type LocalizationState [locale bundles listeners messages available-locales list-and list-or date]
```
vlaaad, line 138: Changed record to type so that equality check is an identity check. Reason: We now use the localization state as a cache key in editor script UI components, so it's more efficient to use an identity check and skip recreating the state when unnecessary.

editor/src/clj/editor/localization.clj
```diff
@@ -132,25 +130,24 @@
...
   IFn
   (invoke [this v] (impl-format this v))
   (applyTo [this args] (AFn/applyToHelper this args)))
 
 (defn- localization-state? [x]
   (instance? LocalizationState x))
 
-(defn- refresh-messages! [^LocalizationState state]
-  (let [u-locale (ULocale/createCanonical ^String (.-locale state))
+(defn- make-localization-state [^String locale bundles listeners]
```
vlaaad, line 146: A pure function, actually.

editor/src/clj/editor/localization.clj
```diff
@@ -209,16 +210,17 @@
 (defn- localization? [x]
   (instance? Localization x))
 
-(defn- impl-set-bundle! [state key bundle]
-  (-> state
-      (update :bundles assoc key bundle)
-      refresh-messages!))
+(defn- impl-set-bundle [^LocalizationState state key bundle]
```
vlaaad, line 218: A pure function too.

editor/test/editor/git_test.clj
```diff
@@ -20,9 +20,13 @@
...
             [util.text-util :as text-util])
-  (:import [java.io File]
+  (:import [ch.qos.logback.classic Level Logger]
+           [java.io File]
            [org.eclipse.jgit.api Git]
-           [org.eclipse.jgit.revwalk RevCommit]))
+           [org.eclipse.jgit.revwalk RevCommit]
+           [org.slf4j LoggerFactory]))
+
+(.setLevel ^Logger (LoggerFactory/getLogger "org.eclipse.jgit.util.FS") Level/ERROR)
```
vlaaad, line 29: Unrelated: reduce log spam in tests.

editor/test/editor/process_test.clj
```diff
@@ -44,7 +44,7 @@
 
 (deftest redirect-test
   (when-not (os/is-win32?)
-    (is (nil? (process/exec! "bash" "-c" "ls >&2")))
+    (is (nil? (process/exec! {:err :discard} "bash" "-c" "ls >&2")))
```
vlaaad, line 47: Unrelated: reduce log spam in tests.

editor/test/integration/test_util.clj
```diff
@@ -89,6 +89,7 @@
 (set! *warn-on-reflection* true)
 
 (.setLevel ^Logger (LoggerFactory/getLogger "org.eclipse.jetty") Level/ERROR)
+(.setLevel ^Logger (LoggerFactory/getLogger "cognitect.aws.credentials") Level/ERROR)
```
vlaaad, line 92: Unrelated: reduce log spam in tests.

editor/resources/localization/en.editor_localization
```diff
@@ -472,6 +472,95 @@ dialog.open-resource.open-file-failed.detail = This can happen if the file type
...
+dialog.bundle.error.keystore.invalid = Invalid keystore.
+dialog.bundle.error.keystore.required-with-pass = Keystore must be set if keystore password is specified.
+dialog.bundle.error.keystore-pass.not-found = Keystore password file not found.
+dialog.bundle.error.keystore-pass.invalid = Invalid keystore password file.
+dialog.bundle.error.keystore-pass.required-with-keystore = Keystore password must be set if keystore is specified.
+dialog.bundle.error.key-pass.not-found = Key password file not found.
+dialog.bundle.error.key-pass.invalid = Invalid key password file.
+dialog.bundle.error.key-pass.required-with-keystore = Keystore must be set if key password is specified.
+dialog.bundle.error.bundle-format.required = No bundle format selected.
+
```
matgis, line 564: Replace `...` with `…` everywhere?

editor/src/java/com/defold/editor/luart/DynamicToStringUserdata.java
```diff
@@ -0,0 +1,18 @@
+package com.defold.editor.luart;
```
matgis, line 15: Missing copyright header.

editor/src/common/com/defold/editor/localization/MessagePattern.java
```diff
@@ -0,0 +1,5 @@
+package com.defold.editor.localization;
```
matgis, line 15: Missing copyright header.

editor/src/clj/editor/editor_extensions/localization.clj
```diff
@@ -0,0 +1,46 @@
+(ns editor.editor-extensions.localization
```
matgis, line 15: Missing copyright header and
```
```

editor/src/clj/editor/localization.clj
matgis: Does that mean that we won't invalidate cache entries correctly if something like the locale changes?

If so, we could implement custom equality based on a `version` `long` field?

Otherwise, I think we should add a comment about the decision to use `deftype` so someone doesn't "fix" it to use `defrecord` in the future.

matgis (changes requested):
Amazing work! ❤️

Some of the added files are missing copyright headers and reflection checks, and the new strings introduce triple-dots `...` into the translations again. I also think we should add a comment about `LocalizationState` being a `deftype` on purpose.

editor/src/clj/editor/localization.clj
vlaaad: No, we will still invalidate cache entries, but it will be faster (reference equality check instead of structural equality check). I'll add a comment.

PR #11490, opened 2025-11-11
Report user id to Sentry

vlaaad:
We now report users:
Screenshot 2025-11-11 at 13 46 01


editor/src/clj/editor/sentry.clj
```diff
@@ -93,11 +93,12 @@
     value))
 
 (defn make-event
-  [^Exception ex ^Thread thread]
+  [^Exception ex ^Thread thread user]
```
matgis, line 96: We should probably validate the `user` argument to be what we expect here?

matgis (dismissed):

PR #11508, opened 2025-11-14
Improve Model rendering performance in Scene View

matgis:
The editor responsiveness in scenes that feature Models has been drastically improved, especially when per-instance vertex attributes are involved.

Partially adresses #9817.

### Technical changes
* Added a new `editor.gl.attribute` module for working with static vertex attribute buffers and per-attribute bindings.
* Refactored all the various ways we used to represent a description of a vertex attribute into a single `attribute-info` representation, with a sibling type `attribute-reflection-info` extending upon the `attribute-info` with a few fields only known about the attribute after the shader has compiled.
* We now use the vertex attribute reflection data from the shader transpiler whenever we have it, which lets us get `attribute-reflection-info` from the `ShaderLifecycle` without needing an active OpenGL context. This enables us to store both the prepared `AttributeBuffers` and `AttributeBufferBindings` in the graph before rendering, which makes the scene cache dirty check a simple pointer comparison.
* Index buffer functionality was removed from `gl.vertex` and `gl.vertex2`. We use the new `gl.attribute/make-index-buffer` function to manage index buffers now.
* The `with-gl-bindings` macro was updated to properly call `unbind!` on all bound `GLBindables` if something throws an exception.
* The `with-gl-bindings` macro will now recurse into and call `bind!` on values inside collections and maps supplied to it.
* Started moving internal editor shaders from our custom DSL into GLSL resources inside the `.jar` file. The goal is to have all internal shaders go through the shader transpiler we use for project shaders referenced from a Material. A library of internal shaders in various permutations have been added to `editor/resources/shaders`, and are available in both world-space and local-space flavors from the new `editor.shaders` module. There are still many places where legacy `defshader` clauses remain in the editor codebase, but all the usages from editor extensions have been removed, so we'll be able to remove the `defshader` macro without updating the extensions once we've addressed them all.
* Refactored types and conversion functions into a new `graphics.types` module.
* Renamed the `gl.protocols` module into `gl.types` and expanded on it. The `GlBind` protocol was renamed to `GLBinding`, and a `!` was suffixed to its methods.

Depends on PR counterparts:
* defold/extension-rive#225
* defold/extension-spine#249
* defold/extension-texturepacker#21


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderUtil.java
```diff
@@ -42,7 +42,7 @@ public static class Common {
         public static String        includeDirectiveBaseStr              = "^\\s*\\#include\\s+(?:<(?<pathbrackets>[^\"<>|\b]+)>|\"(?<pathquotes>[^\"<>|\b]+)\")\\s*(?://.*)?$";
         public static final Pattern includeDirectivePattern              = Pattern.compile(includeDirectiveBaseStr);
         public static final Pattern arrayArraySamplerPattern             = Pattern.compile("^\\s*uniform(?<qualifier>.*)sampler2DArray\\s+(?<uniform>\\w+);$");
-        public static final Pattern regexVersionStringPattern            = Pattern.compile("^\\h*#\\h*version\\h+(?<version>\\d+)(\\h+(?<profile>\\S+))?\\h*\\n");
+        public static final Pattern regexVersionStringPattern            = Pattern.compile("^\\h*#\\h*version\\h+(?<version>\\d+)(\\h+(?<profile>\\S+))?\\h*\\r?\\n");
```
matgis, line 45: Bugfix: Correctly disambiguate between old shaders and new shaders when the string contains CRLF line endings.

editor/src/clj/editor/gl/buffer.clj
```diff
@@ -1,146 +0,0 @@
...
-(ns editor.gl.buffer
-  (:require [editor.buffers :as buffers]
-            [editor.gl :as gl]
-            [editor.gl.protocols :refer [GlBind]]
-            [editor.scene-cache :as scene-cache]
-            [internal.java :as java]
-            [util.defonce :as defonce])
-  (:import [clojure.lang IHashEq]
-           [com.jogamp.opengl GL2]
-           [java.nio Buffer]))
```
matgis, line 24: This module was refactored into specialized types for various buffers in the `gl.attribute` module.

editor/src/clj/editor/gl/protocols.clj
```diff
@@ -1,21 +0,0 @@
...
-;; CONDITIONS OF ANY KIND, either express or implied. See the License for the
-;; specific language governing permissions and limitations under the License.
-
-(ns editor.gl.protocols)
-
-(set! *warn-on-reflection* true)
-
-(defprotocol GlBind
-  (bind [this gl render-args] "Bind this object to the GPU context.")
-  (unbind [this gl render-args] "Unbind this object from the GPU context."))
```
matgis, line 21: Moved into `gl.types`.

editor/src/clj/editor/gl/shader.clj
```diff
@@ -287,28 +294,29 @@ These forms should be quoted, as if they came from a macro."
...
   (set-uniform-array [this gl name count val]))
 
-(defn attribute-locations [shader-variables gl attribute-names]
-  (let [attribute-infos-by-name (attribute-infos shader-variables gl)]
-    (coll/transfer attribute-names (vector-of :int)
-      (map (fn [attribute-name]
-             (if-some [attribute-info (get attribute-infos-by-name attribute-name)]
-               (:index attribute-info)
-               -1))))))
-
```
matgis, line 311: Moved into the `ShaderVariables` protocol.

editor/src/clj/editor/gl/shader.clj
```diff
@@ -357,22 +365,6 @@ This must be submitted to the driver for compilation before you can use it. See
...
-(defn- shader-type->gl-shader-type
-  ^long [shader-type]
-  ;; Enum values for compute shaders exist in Protobuf but are not supported by
-  ;; the OpenGL profile we use in the editor. Uncomment if we update.
-  (case shader-type
-    :shader-type-vertex GL2/GL_VERTEX_SHADER
-    :shader-type-fragment GL2/GL_FRAGMENT_SHADER
-    #_#_ :shader-type-compute GL3/GL_COMPUTE_SHADER))
-
```
matgis, line 375: Moved to new `gl.types` module.

editor/src/clj/editor/gl/shader.clj
```diff
@@ -450,7 +441,7 @@ This must be submitted to the driver for compilation before you can use it. See
 (defn- set-uniform-impl! [gl program uniform-infos uniform-name uniform-value]
   (when-some [uniform-info (uniform-infos uniform-name)]
     (try
-      (set-uniform-at-index gl program (:index uniform-info) uniform-value)
```
matgis, line 453: `uniform-info` `:index` has been renamed to `:location` for clarity and consistency with `attribute-info`.

editor/src/clj/editor/gl/shader.clj
```diff
@@ -624,31 +782,6 @@ This must be submitted to the driver for compilation before you can use it. See
...
-    GL2/GL_INT_VEC4 :int-vec4
-    GL2/GL_BOOL :bool
-    GL2/GL_BOOL_VEC2 :bool-vec2
-    GL2/GL_BOOL_VEC3 :bool-vec3
-    GL2/GL_BOOL_VEC4 :bool-vec4
-    GL2/GL_FLOAT_MAT2 :float-mat2
-    GL2/GL_FLOAT_MAT3 :float-mat3
-    GL2/GL_FLOAT_MAT4 :float-mat4
-    GL2/GL_SAMPLER_2D :sampler-2d
-    GL2/GL_SAMPLER_CUBE :sampler-cube))
```
matgis, line 645: Moved to `gl.types`.

editor/src/clj/editor/gl/shader.clj
```diff
@@ -624,31 +782,6 @@ This must be submitted to the driver for compilation before you can use it. See
...
-
-(defn- sampler-uniform-type? [uniform-type]
-  (case uniform-type
-    (:sampler-2d :sampler-cube) true
-    false))
-
```
matgis, line 651: Moved to `graphics.types`.

editor/src/clj/editor/gl/shader.clj
```diff
@@ -660,40 +793,53 @@ This must be submitted to the driver for compilation before you can use it. See
...
+            array-size (aget out-array-size 0)
             ;; 1. strip brackets from uniform name, i.e "uniform_name[123]"
             sanitized-name (string/replace name name-array-suffix-pattern "")]
         {:name sanitized-name
-         :index location
-         :type type
-         :count count}))))
+         :uniform-type uniform-type
+         :location base-location
+         :array-size array-size}))))
```
matgis, line 803: The `uniform-info` now uses the same keys as `attribute-info`.

editor/src/clj/editor/gl/shader.clj
```diff
@@ -660,40 +793,53 @@ This must be submitted to the driver for compilation before you can use it. See
...
+              {:name name
+               :name-key name-key
+               :semantic-type inferred-semantic-type
+               :vector-type vector-type
+               :data-type data-type
+               :normalize false
+               :coordinate-space :coordinate-space-default
+               :step-function :vertex-step-function-vertex
+               :location base-location
+               :array-size array-size})))))))
```
matgis, line 833: This is the `attribute-reflection-info` we obtain using the OpenGL context. It has been updated to conform to the `attribute-info` maps we can customize in the Material, but has a few extra attributes. Basically, an `attribute-reflection-info` is an `attribute-info` with extra `:location` and `:array-size` fields.

editor/src/clj/editor/gl/shader.clj
```diff
@@ -764,12 +912,20 @@ This must be submitted to the driver for compilation before you can use it. See
...
+    ;; reflection info during the shader transpilation step, and we could also
+    ;; move the uniform values out to a separate kind of GLBinding for uniforms.
+    ;; The attribute reflection info is already available in any ShaderLifecycle
+    ;; that is not created from a defshader expression.
     {:program gl-program
      :uniform-infos uniform-infos
-     :attribute-infos attribute-infos
+     :attribute-reflection-infos attribute-reflection-infos
+     :name-key->attribute-location name-key->attribute-location
+     :semantic-type->attribute-locations semantic-type->attribute-locations
```
matgis, line 919: This is the stuff we return from the `scene-cache` after we've requested a shader from the OpenGL context.

editor/src/clj/editor/gl/vertex.clj
```diff
@@ -344,7 +346,7 @@ the `do-gl` macro from `editor.gl`."
   ([capacity layout vx-size set-fn get-fn]
    (let [buffer-starts (buffer-starts capacity layout)
          byte-size (* capacity vx-size)
-         storage (b/new-byte-buffer byte-size :byte-order/little-endian)
+         storage (b/new-byte-buffer byte-size :byte-order/native)
```
matgis, line 349: I think this was just wrong, but we're not building the editor for any big-endian architectures at the moment.

editor/src/clj/editor/gl/vertex.clj
```diff
@@ -434,31 +436,34 @@ the `do-gl` macro from `editor.gl`."
...
+    (mapv (fn [^long offset attrib ^long loc]
+            (when (not= -1 loc)
+              (let [[_nm ^int sz tp & more] attrib
+                    ^boolean norm (if (not (nil? (first more))) (first more) false)
+                    ^int gl-type (gl-types tp)]
+                (gl/gl-vertex-attrib-pointer gl loc sz gl-type norm stride offset))))
           offsets
-          attribs)))
+          attribs
+          attribute-locations)))
```
matgis, line 466: This function was rewritten to call `shader/attribute-locations` instead of `shader/attribute-infos`.

editor/src/clj/editor/gl/vertex.clj
```diff
@@ -501,40 +491,24 @@ the `do-gl` macro from `editor.gl`."
...
-(defrecord VertexIndexBufferShaderLink [request-id ^PersistentVertexBuffer vertex-buffer ^"[I" index-buffer shader]
-  GlBind
-  (bind [_this gl render-args]
-    (bind-vertex-buffer-with-shader! gl request-id vertex-buffer shader)
-    (bind-index-buffer! gl request-id index-buffer))
-
-  (unbind [_this gl render-args]
-    (unbind-vertex-buffer-with-shader! gl vertex-buffer shader)
-    (unbind-index-buffer! gl)))
-
```
matgis, line 528: This needlessly tied the index buffer to the shader. We now the `GLBinding` returned from `gl.attribute/make-index-buffer` to manage index buffers. I don't think this one in `gl.vertex` was used anywhere, and the one in `gl.vertex2` was only used from `extension-rive`.

editor/src/clj/editor/gl/vertex2.clj
```diff
@@ -16,75 +16,19 @@
...
-    :binormal :semantic-type-binormal
-    :color :semantic-type-color
-    (:texcoord :texcoord0) :semantic-type-texcoord
-    :page-index :semantic-type-page-index
-    :blend-indices :semantic-type-blend-indices
-    :blend-weights :semantic-type-blend-weights
-    :mtx-world :semantic-type-world-matrix
-    :mtx-normal :semantic-type-normal-matrix
-    :semantic-type-none))
-
```
matgis, line 87: Moved to `graphics.types` module.

editor/src/clj/editor/gl/vertex2.clj
```diff
@@ -16,75 +16,19 @@
...
-   :ushort  GL/GL_UNSIGNED_SHORT
-   :short   GL/GL_SHORT
-   :uint    GL/GL_UNSIGNED_INT
-   :int     GL2/GL_INT
-   :float   GL/GL_FLOAT
-   :double  GL2/GL_DOUBLE})
-
-(def usage-types
-  {:static GL2/GL_STATIC_DRAW
-   :dynamic GL2/GL_DYNAMIC_DRAW})
```
matgis, line 46: Moved to new `gl.types` module.

editor/src/clj/editor/gl/vertex2.clj
```diff
@@ -118,16 +62,19 @@
...
 
 (defn wrap-buf
   ^ByteBuffer [^bytes byte-array]
-  (buffers/wrap-byte-array byte-array :byte-order/little-endian))
+  (buffers/wrap-byte-array byte-array :byte-order/native))
 
 (defn make-buf
   ^ByteBuffer [byte-capacity]
-  (buffers/new-byte-buffer byte-capacity :byte-order/little-endian))
+  (buffers/new-byte-buffer byte-capacity :byte-order/native))
```
matgis, line 77: More places where we assumed little-endian architecture. Again, this change doesn't really change anything since we're only building the editor for big-endian.

editor/src/clj/editor/gl/vertex2.clj
```diff
@@ -145,113 +96,89 @@
...
 
-(definline buf-put! [buf byte-offset data-type normalize numbers]
-  `(buffers/put! ~buf ~byte-offset ~data-type ~normalize ~numbers))
-
-(defn put!
-  ^VertexBuffer [^VertexBuffer vbuf byte-offset data-type normalize numbers]
-  (buf-put! (.buf vbuf) byte-offset data-type normalize numbers)
-  vbuf)
+(definline buf-put! [buf byte-offset buffer-data-type normalize numbers]
+  `(buffers/put! ~buf ~byte-offset ~buffer-data-type ~normalize ~numbers))
```
matgis, line 100: Removed unused `put!` function and renamed `data-type` argument to `buffer-data-type` to differentiate it from the vertex attribute `data-type` that is present in many places where we call this from.

editor/src/clj/editor/gl/vertex2.clj
```diff
@@ -145,113 +96,89 @@
...
+(defn- attribute-components [attribute-infos]
+  (for [{attribute-name :name :keys [name-key vector-type] :as attribute-info} attribute-infos
+        n (range (graphics.types/vector-type-component-count vector-type))]
+    (let [suffix (nth ["-x" "-y" "-z" "-w"] n)
+          attribute-name (str attribute-name suffix)
+          attribute-key (keyword (str (name name-key) suffix))]
+      (assoc attribute-info
+        :name attribute-name
+        :name-key attribute-key
+        :vector-type :vector-type-scalar))))
```
matgis, line 119: We used to have many different representations of vertex attribute descriptions. Now, we return something that conforms to `graphics.types/attribute-info?` everywhere.

editor/src/clj/editor/gl/vertex.clj
```diff
@@ -434,31 +436,34 @@ the `do-gl` macro from `editor.gl`."
...
 
+;; TODO(instancing): Produce proper attribute-infos from the defvertex macro.
+(defn- fake-attribute-info [[name-sym :as _attrib]]
+  {:pre [(symbol? name-sym)]}
+  (let [name (name name-sym)
+        name-key (graphics.types/attribute-name-key name)]
+    {:name name
+     :name-key name-key
+     :semantic-type (graphics.types/infer-semantic-type name-key)}))
+
```
matgis, line 447: This is the one place left where we don't return a fully-formed `attribute-info` for a vertex attribute. This is enough to be compatible with the use case. Returning a full `attribute-info` would require us to update the `defvertex` macro. Might do this in a separate PR at some point.

editor/src/clj/editor/gl/vertex2.clj
```diff
@@ -145,113 +96,89 @@
...
-(defn- vertex-size
-  ^long [attributes]
-  (transduce (map attribute-size) + 0 attributes))
-
-(defn make-vertex-description
-  [name attributes]
-  {:name name
-   :attributes attributes
-   :size (vertex-size attributes)})
-
```
matgis, line 182: Moved to `graphics.types` module.

editor/src/clj/editor/gl/vertex2.clj
```diff
@@ -145,113 +96,89 @@
...
 (definline buf-push-floats! [buffer numbers]
   `(buffers/push-floats! ~buffer ~numbers))
 
-(definline buf-push! [buf data-type normalize numbers]
-  `(buffers/push! ~buf ~data-type ~normalize ~numbers))
-
-(defn push!
-  ^VertexBuffer [^VertexBuffer vbuf data-type normalize numbers]
-  (buf-push! (.buf vbuf) data-type normalize numbers)
-  vbuf)
```
matgis, line 165: Unused `push!` function removed.

editor/src/clj/editor/gl/vertex2.clj
```diff
@@ -145,113 +96,89 @@
...
-                           4 :vector-type-vec4)]
-    (assert num-components (str type " is not a valid type name. It must start with vec1, vec2, vec3, or vec4."))
-    (assert (get gl-types suffix) (str type " is not a valid type name. It must end with byte, short, int, float, or double. (Defaults to float if no suffix.)"))
+        attribute-key    (graphics.types/attribute-name-key attribute-name)
+        semantic-type    (graphics.types/infer-semantic-type attribute-key)
+
+        ;; TODO: You can't define a local-space attribute using defvertex.
+        coordinate-space (case semantic-type
+                           (:semantic-type-position :semantic-type-normal :semantic-type-tangent) :coordinate-space-world
+                           :coordinate-space-local)]
```
matgis, line 166: I believe all `defvertex` that remain should be in world-space now.

editor/src/clj/editor/gl/vertex2.clj
```diff
@@ -145,113 +96,89 @@
...
+     :data-type data-type
      :semantic-type semantic-type}))
 
 (defmacro defvertex
   [name & attribute-definitions]
   (let [attributes         (mapv parse-attribute-definition attribute-definitions)
-        vertex-description (make-vertex-description name attributes)
+        vertex-description (-> attributes
+                               (graphics.types/make-vertex-description)
+                               (assoc :defvertex-name-sym name)) ;
```
matgis, line 181: This used to be called `:name`. It only appears to be used by the tests at the moment.

editor/src/clj/editor/gl/vertex2.clj
```diff
@@ -345,55 +270,38 @@
...
-  (clear-attributes! gl expanded-attribute-locations)
-  (gl/gl-bind-buffer gl GL/GL_ARRAY_BUFFER 0))
-
-(defn- bind-index-buffer! [^GL2 gl request-id ^IntBuffer index-buffer]
-  ;; TODO: Feed in actual version. Perhaps deftype IndexBuffer with IntBuffer + version?
-  (let [ibo (scene-cache/request-object! ::ibo2 request-id gl {:index-buffer index-buffer :version 0})]
-    (gl/gl-bind-buffer gl GL/GL_ELEMENT_ARRAY_BUFFER ibo)))
-
-(defn- unbind-index-buffer! [^GL2 gl]
-  (gl/gl-bind-buffer gl GL/GL_ELEMENT_ARRAY_BUFFER 0))
```
matgis, line 360: Like in `gl.vertex`, index buffer related functionality was removed. We now use the `GLBinding` returned from `gl.attribute/make-index-buffer` to manage index buffers. This was only used from `extension-rive`.

editor/src/clj/editor/pipeline/shader_gen.clj
```diff
@@ -14,38 +14,18 @@
...
-    :shader-type-vertex Graphics$ShaderDesc$ShaderType/SHADER_TYPE_VERTEX
-    :shader-type-fragment Graphics$ShaderDesc$ShaderType/SHADER_TYPE_FRAGMENT
-    :shader-type-compute Graphics$ShaderDesc$ShaderType/SHADER_TYPE_COMPUTE))
-
-(defn filename->pb-shader-type
-  ^Graphics$ShaderDesc$ShaderType [^String filename]
-  (-> filename
-      filename->shader-type
-      shader-type->pb-shader-type))
-
```
matgis, line 48: Moved to `graphics.types` module.

editor/src/clj/editor/atlas.clj
```diff
@@ -74,50 +74,6 @@
...
-
-(shader/defshader outline-fragment-shader
-  (varying vec4 var_color)
-  (defn void main []
-    (setq gl_FragColor var_color)))
-
-; TODO - macro of this
-; JG: This is copied from sprite, but does the same thing - merge them?
-(def outline-shader (shader/make-shader ::outline-shader outline-vertex-shader outline-fragment-shader))
-
```
matgis, line 120: We use the shared shaders and vertex-descriptions in the new `editor.shaders` module now.

editor/src/clj/editor/atlas.clj
```diff
@@ -593,81 +535,20 @@
...
-                :tags #{:atlas}
-                :passes [pass/transparent]}
-   :children [{:aabb aabb
-               :renderable {:render-fn render-atlas-outline
-                            :tags #{:atlas :outline}
-                            :passes [pass/outline]}}]})
+(defn- make-page-scene
+  [layout-width layout-height page-index gpu-texture]
+  (let [page-offset-transform (get-rect-transform layout-width page-index)]
+    (render-util/make-outlined-textured-quad-scene #{:atlas} page-offset-transform layout-width layout-height gpu-texture page-index)))
```
matgis, line 541: Atlas now shares this utility function with `extension-texturepacker`.

editor/src/clj/editor/atlas.clj
```diff
@@ -593,81 +535,20 @@
...
-        child-renderables (into [] (for [[page-index page-rects] pages] (produce-page-renderables aabb width height page-index page-rects gpu-texture)))]
-    {:aabb aabb
-     :info-text (format "%d x %d (%s profile)" width height (:name texture-profile))
-     :children (into child-renderables
-                     child-scenes)}))
+        page-scenes (mapv (fn [^long page-index]
+                            (make-page-scene width height page-index gpu-texture))
+                          (range (count pages)))]
+    {:info-text (format "%d x %d (%s profile)" width height (:name texture-profile))
+     :children (into page-scenes child-scenes)}))
```
matgis, line 551: Removed unused `:aabb` from parent scene. It only included the first page anyway.

editor/src/clj/editor/buffers.clj
```diff
@@ -94,6 +109,55 @@
...
+          (.rewind view)
+          (loop [result init]
+            (if (.hasRemaining view)
+              (let [element (next-element view)
+                    result (rf result element)]
+                (if (reduced? result)
+                  (unreduced result)
+                  (recur result)))
+              result)))))))
+
```
matgis, line 160: You can now reduce over the contents of a `Buffer`. Used for Reveal and tests.

editor/src/clj/editor/buffers.clj
```diff
@@ -208,6 +349,79 @@
           byte-order-comparison
           (compare (total-byte-size a) (total-byte-size b)))))))
 
+;; A wrapper around Buffer with equality semantics based on the topology of the
+;; Buffer and an explicit data version number.
+(defonce/type BufferData
+  [^Buffer data
+   ^int data-version
+   ^int topology-hash]
```
matgis, line 359: We use this wrapper around `Buffer` to avoid expensive equality tests when you need to supply a buffer as part of the `request-data` provided to `scene-cache` requests. The `Buffer` implementations do proper content equality, bless them, but we'd rather manage equality ourselves by comparing a version number.

editor/src/clj/editor/buffers.clj
```diff
@@ -284,10 +498,17 @@
                   (.putInt buffer byte-offset (num/normalized->uint n))
                   (+ byte-offset Integer/BYTES))
                 byte-offset
+                numbers)
+
+        :long
+        (reduce (fn [^long byte-offset n]
+                  (.putLong buffer byte-offset (num/normalized->long n))
+                  (+ byte-offset Long/BYTES))
+                byte-offset
```
matgis, line 509: We didn't have support for `LongBuffers` before. And we don't really use them yet, but I added it for completeness.

editor/src/clj/editor/gl.clj
```diff
@@ -237,26 +241,277 @@
...
+                   (body-fn!))]
+      (reduce #(gl.types/unbind! %2 gl render-args)
+              nil
+              (rseq bound-items))
+      (if exception
+        (throw exception)
+        result))))
+
+(defmacro with-gl-bindings [gl-expr render-args-expr bindable-items-expr & body]
+  `(with-gl-bindings-impl ~gl-expr ~render-args-expr ~bindable-items-expr (fn ~'body-fn! [] ~@body)))
```
matgis, line 508: The `with-gl-bindings` macro was updated to properly call `unbind!` on all bound `GLBindables` if something throws an exception. It will also recurse into and call `bind!` on values inside collections and maps supplied to it.

editor/src/clj/editor/graphics.clj
```diff
@@ -38,39 +38,6 @@
...
-    0.0 1.0 0.0
-    0.0 0.0 1.0))
-
-(def ^:private mat4-identity
-  (vector-of :double
-    1.0 0.0 0.0 0.0
-    0.0 1.0 0.0 0.0
-    0.0 0.0 1.0 0.0
-    0.0 0.0 0.0 1.0))
-
```
matgis, line 73: A lot of stuff was moved from here to `graphics.types` module so we can access it from elsewhere, like `editor.shader-gen`.

editor/src/clj/editor/material.clj
```diff
@@ -417,7 +398,7 @@
                 :type-unsigned-short num/normalized->ushort-double
                 :type-int num/normalized->int-double
                 :type-unsigned-int num/normalized->uint-double))]
-        (update new-attribute :values #(into (empty %) (map coerce-fn) %)))
+        (update new-attribute :values coll/transform (map coerce-fn)))
```
matgis, line 401: `coll/transform` is like `coll/transfer`, but transfers into an empty collection of the same type as the input collection.

editor/src/clj/editor/material.clj
```diff
@@ -434,15 +415,12 @@
...
-            (map-indexed (fn [index new-attribute]
-                           (if-some [old-attribute (get old-attributes index)]
-                             (coerce-attribute new-attribute old-attribute)
-                             new-attribute)))
+    ;; updated data and vector type.
+    (let [old-attributes-by-name (coll/pair-map-by :name (:attributes user-data))]
+      (mapv (fn [new-attribute]
+              (if-some [old-attribute (get old-attributes-by-name (:name new-attribute))]
+                (coerce-attribute new-attribute old-attribute)
+                new-attribute))
```
matgis, line 423: I realized we will want to add the ability to reorder attributes in the view quite soon since it affects the shader location bindings. Updated this code to no longer assume that values will not be reordered. Instead, it assumes the name and the type cannot change at the same time, which seems more sensible here.

editor/src/clj/editor/math.clj
```diff
@@ -529,78 +549,94 @@
...
+  (try
+    (let [normal-transform (Matrix3d.)]
+      (.getRotationScale transform normal-transform)
+      (.invert normal-transform)
+      (.transpose normal-transform)
+      (doto (Matrix4d.)
+        (.setRotationScale normal-transform)
+        (.setM33 1.0)))
+    (catch SingularMatrixException _
+      identity-mat4)))
```
matgis, line 561: Give up gracefully if matrices cannot be inverted. The normals probably don't matter much in this case since the positions will not be in view either.

editor/src/clj/editor/math.clj
```diff
@@ -529,78 +549,94 @@
...
+    :projection
+    :texture
+    :view
+    :view-proj
+    :world
+    :world-rotation
+    :world-view
+    :world-view-proj})
+
+(fn/defamong render-transform-key? render-transform-keys)
```
matgis, line 639: `fn/defamong` converts a collection into a `case` function. It's fast and stuff.

editor/src/clj/editor/mesh.clj
```diff
@@ -343,14 +366,10 @@
...
-                   :renderable {:render-fn render-outline
-                                :tags #{:model :outline}
-                                :batch-key _node-id
-                                :select-batch-key _node-id
-                                :passes [pass/outline]}}]})))
+                   :renderable (render-util/make-aabb-outline-renderable #{:model})}]})))
 
 (g/defnk produce-aabb [streams position-stream]
-  (if-some [{:keys [count data]} (util/first-where #(position-stream-name? (:name %) position-stream) streams)]
+  (if-some [{:keys [count data]} (coll/first-where #(position-stream-name? (:name %) position-stream) streams)]
```
matgis, line 372: Moved `first-where` into `coll`. `util/first-where` exists as a macro and will be removed later.

editor/src/clj/editor/mesh.clj
```diff
@@ -569,16 +588,6 @@
...
-(defn- world-transform->normal-transform
-  ^Matrix4d [^Matrix4d world-transform]
-  (let [normal-transform (Matrix3d.)]
-    (.getRotationScale world-transform normal-transform)
-    (.invert normal-transform)
-    (.transpose normal-transform)
-    (doto (Matrix4d.)
-      (.setIdentity)
-      (.setRotationScale normal-transform))))
-
```
matgis, line 581: This was the same as `math/derive-normal-transform`.

editor/src/clj/editor/model_scene.clj
```diff
@@ -14,32 +14,34 @@
...
-            [internal.graph.error-values :as error-values]
             [util.coll :as coll]
             [util.num :as num])
   (:import [com.google.protobuf ByteString]
            [com.jogamp.opengl GL GL2]
-           [editor.gl.vertex2 VertexBuffer]
-           [java.nio ByteOrder]
-           [javax.vecmath Matrix4d Vector4d]))
+           [java.nio ByteOrder FloatBuffer]
+           [javax.vecmath Vector4d]))
```
matgis, line 44: This is a full rewrite of everything related to rendering for Model Scenes. Improvements:
* Use attribute reflection info from the shader transpiler so we don't need to defer vertex buffer creation until we have an OpenGL context. Previously a single interleaved attribute buffer would be created inside the `update-fn` associated with the `scene-cache` object, and expensive mesh data comparisons would be performed each frame in case the data changed. Using the attribute reflection info from the shader transpiler lets us create and cache the attribute buffers in the graph instead.
* Each vertex attribute now resides in its own buffer. This lets us selectively update just the positions, normals, or tangents for world-space vertex buffers, but leave other attributes intact. This also saves us from having to write matrix attributes for each vertex into the interleaved vertex buffer when a material uses per-instance attributes. We don't yet batch instances, but we can assign each matrix attribute from a constant `AttributeValueBinding` instead of writing them to the interleaved vertex buffer.
* Instead of duplicating vertices in the interleaved vertex buffer, we create and cache an `IndexBufferLifecycle` in the graph and use it to refer to the sparse vertices in the attribute buffers.

editor/src/clj/editor/particle_lib.clj
```diff
@@ -160,52 +161,19 @@
...
-(defn- vector-type->int [vertex-vector-type]
-  (case vertex-vector-type
-    :vector-type-scalar Graphics$VertexAttribute$VectorType/VECTOR_TYPE_SCALAR_VALUE
-    :vector-type-vec2 Graphics$VertexAttribute$VectorType/VECTOR_TYPE_VEC2_VALUE
-    :vector-type-vec3 Graphics$VertexAttribute$VectorType/VECTOR_TYPE_VEC3_VALUE
-    :vector-type-vec4 Graphics$VertexAttribute$VectorType/VECTOR_TYPE_VEC4_VALUE
-    :vector-type-mat2 Graphics$VertexAttribute$VectorType/VECTOR_TYPE_MAT2_VALUE
-    :vector-type-mat3 Graphics$VertexAttribute$VectorType/VECTOR_TYPE_MAT3_VALUE
-    :vector-type-mat4 Graphics$VertexAttribute$VectorType/VECTOR_TYPE_MAT4_VALUE))
-
```
matgis, line 202: Moved to `graphics.types` module.

editor/src/clj/editor/render.clj
```diff
@@ -1,128 +0,0 @@
...
-(defn gen-outline-vb [renderables n]
-  (let [vbuf (->vtx-pos-col (* n 2 (count line-order)))]
-    (persistent! (reduce renderable->aabb-box! vbuf renderables))))
-
-(defn render-aabb-outline
-  [^GL2 gl render-args request-id renderables n]
-  (let [vbuf (gen-outline-vb renderables n)
-        outline-vertex-binding (vtx/use-with request-id vbuf shader-outline)]
-    (gl/with-gl-bindings gl render-args [shader-outline outline-vertex-binding]
-      (gl/gl-draw-arrays gl GL/GL_LINES 0 (count vbuf)))))
```
matgis, line 128: This file was renamed to `render_util.clj`.

editor/src/clj/editor/scene.clj
```diff
@@ -216,15 +216,24 @@
...
+                (assoc
+                  :actual/world-rotation shared-world-rotation
+                  :world-rotation shared-world-rotation)
+                (coll/merge
+                  (math/derive-render-transforms
+                    shared-world-transform
+                    (:view render-args)
+                    (:projection render-args)
+                    (:texture render-args))))]
+
```
matgis, line 236: Added `:world-rotation` quaternion to the `render-args`. We use it to transform normals & tangents in some cases for efficiency.

editor/src/clj/editor/scene.clj
```diff
@@ -216,15 +216,24 @@
...
-                                                                     (:projection render-args)
-                                                                     (:texture render-args)))]
+      (let [shared-world-transform (:world-transform first-renderable math/identity-mat4) ; rulers apparently don't have world-transform
+            shared-world-rotation (:world-rotation first-renderable math/identity-quat)
+
+            shared-render-args
+            (-> render-args
+                (assoc
+                  :actual/world-rotation shared-world-rotation
+                  :world-rotation shared-world-rotation)
```
matgis, line 229: When a function such as `math/derive-render-transforms` replaces the actual world-space transforms with identity matrices because the attribute coordinate space is set to `:coordinate-space-world`, and that's just what we decided to do at some point, An `:actual/` prefixed transform is still available. We use these to transform the attribute data so we can supply world-space attributes to the shader while also supplying an identity matrix for the world-space transform component of all the world-space matrix attributes and uniforms.

Yes, there are annoyed comments about this situation added by me elsewhere in this PR.

editor/src/clj/editor/scene.clj
```diff
@@ -704,10 +708,12 @@
...
   (let [old-node-id (:node-id scene)
-        children (:children scene)]
+        children (:children scene)
+        finalize-claim-fn (:finalize-claim-fn scene)]
     (cond-> (assoc scene :node-id new-node-id :node-outline-key new-node-outline-key)
-            children (assoc :children (mapv (partial map-scene (partial claim-child-scene old-node-id new-node-id new-node-outline-key))
-                                            children)))))
+            children (assoc :children (mapv (partial map-scene #(claim-child-scene % old-node-id new-node-id new-node-outline-key))
+                                            children))
+            finalize-claim-fn (finalize-claim-fn old-node-id new-node-id))))
```
matgis, line 716: Nodes can now specify a `:finalize-claim-fn` in their `scene` node output. Whenever the ownership of the scene data is claimed by the `claim-scene` function, we will call this function on the result. This allows us to replace the owning `scene-node-id` in the `scene-cache` `request-ids` for each `TransformedAttrubuteBufferLifecycle`. If we do not do this, the attribute buffers will be continuously overwritten as each instance is rendered, which is a lot less efficient than keeping them individual to each instance. See the `gl.attribute/claim-transformed-attribute-buffer-bindings` function for details.

editor/src/clj/editor/scene_cache.clj
```diff
@@ -27,14 +27,6 @@
...
 (defonce ^:private object-caches-atom (atom {}))
 
-(defn valid-request-id? [value]
-  ;; TODO(instancing): A good request-id should implement efficient hashing.
-  ;; Enable stricter check and fix call sites in all extensions.
-  #_ (or (number? value)
-         (keyword? value)
-         (instance? IHashEq value))
-  (any? value))
-
```
matgis, line 37: Moved to `graphics.types` module.

editor/src/clj/editor/scene_cache.clj
```diff
@@ -151,8 +143,7 @@
   nil)
 
 (defn request-object! [cache-id request-id context request-data]
-  {:pre [(valid-request-id? request-id)]}
-  ;; TODO(instancing): Should we change the argument order?
```
matgis, line 155: Turned out we shouldn't! 😄

editor/src/clj/editor/scene_cache.clj
```diff
@@ -163,7 +154,6 @@
       (if-some [[old-object old-request-data]
                 (some-> (get-in cache-meta [:caches context])
                         (cache/lookup request-id))]
-        ; TODO(instancing): Potentially expensive comparison. Can we do better?
```
matgis, line 166: The new `BufferData` class helps with this for the Model scenes that had this problem.

editor/src/clj/editor/texture_set.clj
```diff
@@ -152,11 +153,16 @@
...
+  (vec3 position)
+  (vec2 texcoord)
+  (vec1 page_index))
+
 (defn- gen-vertex
   [^Matrix4d world-transform x y u v page-index]
   (let [p (Point3d. x y 0.0)]
     (.transform world-transform p)
-    (vector-of :double (.x p) (.y p) (.z p) 1.0 u v page-index)))
+    (vector-of :double (.x p) (.y p) (.z p) u v page-index)))
```
matgis, line 162: Remove `1.0` for unused `w` component of positions.

editor/src/clj/util/coll.clj
```diff
@@ -767,3 +898,48 @@
...
+      (primitive-types-by-array-manager-id array-manager-id))))
+
+(defn mapv>
+  "Like core.mapv, but takes the input sequence as the first argument and
+  supplies any arguments following the transform function to it after the item
+  argument. Useful with various core functions such as update."
+  ([coll f]
+   (mapv f coll))
+  ([coll f & args]
+   (mapv #(apply f % args) coll)))
```
matgis, line 945: This turned out really useful in conjunction with `update`. Maybe we should have more of these? 🤔

editor/src/clj/util/ensure.clj
```diff
@@ -0,0 +1,57 @@
...
+  [argument-sym expected-class-sym]
+  {:pre [(symbol? argument-sym)
+         (symbol? expected-class-sym)
+         (class? (resolve expected-class-sym))]}
+  `(argument
+     ~argument-sym
+     #(instance? ~expected-class-sym %)
+     "%s argument must be %s - got %s"
+     (.getName ~expected-class-sym)
+     (some-> ~argument-sym class .getName)))
```
matgis, line 57: `ensure` are like `assert`, but are still enabled when you turn asserts off. Not entirely sure about these after using them for a bit. Main issue is that you tend to want to include a friendlier error message describing the problem.

editor/test/editor/graphics_test.clj
```diff
@@ -37,18 +32,6 @@
...
-          :vector-type-vec2 2
-          :vector-type-vec3 3
-          :vector-type-vec4 4
-          :vector-type-mat2 4
-          :vector-type-mat3 9
-          :vector-type-mat4 16}
-         (into {}
-               (map (juxt identity graphics/vector-type->component-count))
-               vector-types))))
-
```
matgis, line 51: Moved to `graphics.types-test`.

editor/test/integration/collada_scene_test.clj
```diff
@@ -1,60 +0,0 @@
...
-          vb (-> (->vtx-pos-nrm-tex vertex-count)
-                 (model-scene/populate-vb! world-transform normal-transform vertex-attribute-bytes false false mesh-renderable-data))]
-      (is (= vertex-count (count vb))))))
-
-(deftest invalid-scene
-  (test-util/with-loaded-project
-    (let [node-id (test-util/resource-node project "/mesh/invalid.dae")
-          scene (log/without-logging
-                  (g/node-value node-id :scene))]
-      (is (g/error? scene)))))
```
matgis, line 60: This was a copy of `model_scene_test.clj` for some reason.

editor/test/internal/util_test.clj
```diff
@@ -93,39 +92,6 @@
...
-  (is (nil? (util/first-index-where nil? nil)))
-  (is (nil? (util/first-index-where even? nil)))
-  (is (nil? (util/first-index-where even? [])))
-  (is (nil? (util/first-index-where even? [1 3 5])))
-
-  (testing "stops calling pred after first true"
-    (let [pred (fn/make-call-logger fn/constantly-true)]
-      (is (= 0 (util/first-index-where pred (range 10))))
-      (is (= 1 (count (fn/call-logger-calls pred)))))))
-
```
matgis, line 128: Moved to `util.coll-test`.

editor/test/resources/test_project/materials/test_pbr_materials.material
```diff
@@ -1,4 +1,4 @@
 name: "test_pbr_materials"
 tags: "test_pbr_materials"
-vertex_program: "/materials/test.vp"
```
matgis, line 3: Fixes a warning about mixing old and new shaders from the tests.

editor/src/clj/editor/buffers.clj
```diff
@@ -157,6 +234,19 @@
...
 
+(definline item-count [^Buffer buffer]
+  `(int
+     (if-let [^Buffer buffer# ~buffer]
+       (- (.limit buffer#)
+          (.position buffer#))
+       0)))
+
+(definline empty? [^Buffer buffer]
+  `(pos? (item-count ~buffer)))
```
vlaaad: `(pos? (item-count ...))` seems like the opposite of `empty?`

editor/src/clj/editor/buffers.clj
```diff
@@ -170,7 +260,58 @@
...
+    (:byte :ubyte)
+    (let [size (item-count buffer)
+          array (byte-array size)]
+      (.get buffer array)
+      array)))
+
+(defn data-hash
+  ^long [^Buffer buffer]
+  (if (.hasArray buffer)
+    (array/hash-code (.array buffer) (.arrayOffset buffer) (.limit buffer))
```
vlaaad: I'm not 100% sure, but it's possible you need to use `(+ (.arrayOffset buffer) (.limit buffer))` instead of `(.limit buffer)`.

```clj
(let [buf (-> (ByteBuffer/wrap
                (byte-array [1 2 3 4 5 6])
                2 3)
              (.position 2)
              (.slice)
              (.limit 3))]
  {:as-vec (vec (reducible buf))
   :data-hash (data-hash buf)
   :array-hash (array/hash-code (byte-array [1 2 3 4 5 6]) 2 5)})
=> {:as-vec [3 4 5]
    :data-hash 34 ;; 31 (seed) + 3, implies only hashed the first element
    :array-hash 32803}
```

editor/src/clj/editor/gl/shader.clj
```diff
@@ -577,16 +646,116 @@ This must be submitted to the driver for compilation before you can use it. See
...
+    (throw (IllegalArgumentException. "At least one shader-path must be supplied."))
+    (let [resources-by-proj-path (g/valid-node-value workspace :resource-map)
+
+          shader-proj-path->source
+          (fn proj-path->source [proj-path]
+            (if-some [resource (resources-by-proj-path proj-path)]
+              (slurp resource)
+              (throw (FileNotFoundException. (format "'%s' was not found in the workspace." proj-path)))))]
+
+      (read-shader :workspace-shader shader-proj-paths opts shader-proj-path->source))))
```
vlaaad: Then delete it?

editor/src/clj/editor/math.clj
vlaaad: Should we return a new Identity matrix instead? I couldn't quite understand if the return value is modified afterwards.

editor/src/clj/editor/mesh.clj
```diff
@@ -290,19 +311,21 @@
...
-     :semantic-type (let [semantic-attribute-key (condp = attribute-key
-                                                   position-stream-name :position
-                                                   normal-stream-name :normal
-                                                   attribute-key)]
-                      (vtx/attribute-key->semantic-type semantic-attribute-key))}))
+     :coordinate-space :coordinate-space-default
+     :step-function :vertex-step-function-vertex
+     :semantic-type (condp = attribute-key
+                      position-stream-name :semantic-type-position
+                      normal-stream-name :semantic-type-normal
```
vlaaad: It seems that `attribute-key` is always a keyword, and `position-stream-name`/`normal-stream-name` are always strings.

editor/src/clj/editor/shaders.clj
```diff
@@ -0,0 +1,193 @@
...
+     :uniforms {"mtx_view_proj" :view-proj}}
+    "shaders/selection-instance.vp"
+    "shaders/selection-instance.fp"))
+
+(def selection-instance-paged-local-space
+  (editor-shader
+    {:coordinate-space :coordinate-space-local
+     :uniforms {"mtx_view_proj" :view-proj}}
+    "shaders/selection-instance-paged.vp"
+    "shaders/selection-instance-paged.fp"))
```
vlaaad, line 160: Should it also provide `:max-page-count shader/max-array-samplers`?

editor/src/clj/editor/model_scene.clj
```diff
@@ -48,189 +50,188 @@
...
+                        :mesh mesh})])
+                  (keep (fn [[attribute-pb-field ^long attribute-count]]
+                          (when (not= position-count attribute-count)
+                            (ex-info
+                              "Attribute count mismatch."
+                              {:position-count position-count
+                               :attribute-count attribute-count
+                               :attribute-pb-field attribute-pb-field
+                               :mesh-request-id mesh-request-id
+                               :mesh mesh}))))
```
vlaaad: Shouldn't it be `g/error-fatal` instead?

editor/src/clj/editor/model_scene.clj
```diff
@@ -48,189 +50,188 @@
...
+
+                          (pos? tangent-count)
+                          (assoc :semantic-type-tangent [tangents])
+
+                          (pos? color-count)
+                          (assoc :semantic-type-color [colors])
+
+                          (or (pos? texcoord0-count)
+                              (pos? texcoord1-count))
+                          (assoc :semantic-type-texcoord (cond-> [(when (pos? texcoord0-count) texcoord0s)]
```
vlaaad, line 223: Is it okay that we may end up with a `nil` in the vector here?
```clj
[(when (pos? texcoord0-count) texcoord0s)]
```

editor/src/clj/editor/model_util.clj
```diff
@@ -0,0 +1,141 @@
...
+                   (if (< channel attribute-buffer-count)
+                     (let [attribute-buffer (attribute-buffers channel)]
+                       (make-attribute-buffer-binding attribute-buffer attribute-info scene-node-id))
+                     (let [attribute-bytes (or (name-key->attribute-bytes name-key) ; From the referencing component.
+                                               (:bytes attribute-info))] ; From the Material.
+                       (make-attribute-value-binding attribute-bytes attribute-info)))))))))
+
+(defn make-attribute-bindings
+  ([semantic-type->attribute-buffers combined-attribute-infos]
+   (make-attribute-bindings semantic-type->attribute-buffers combined-attribute-infos {} nil))
```
vlaaad: This place passes `nil` as a `scene-node-id`, but as it gets passed along, at some point the code asserts `(gt/node-id? scene-node-id)` (in `make-transformed-attribute-buffer`)

editor/src/clj/internal/util.clj
```diff
@@ -156,6 +157,7 @@
   (reduce-kv (fn [m k v] (assoc m (f k) v)) (empty m) m))
 
 (defmacro map-vals [f m]
+  "DEPRECATED. Use coll/map-vals."
```
vlaaad, line 160: Docstring should be before args.

editor/src/clj/util/array.clj
```diff
@@ -13,27 +13,84 @@
...
 
+(defonce ^boolean/1 empty-boolean-array (empty-of-type Boolean/TYPE))
+(defonce ^char/1 empty-char-array (empty-of-type Character/TYPE))
+(defonce ^byte/1 empty-byte-array (empty-of-type Byte/TYPE))
+(defonce ^short/1 empty-short-array (empty-of-type Short/TYPE))
+(defonce ^int/1 empty-int-array (empty-of-type Integer/TYPE))
+(defonce ^long/1 empty-long-array (empty-of-type Long/TYPE))
+(defonce ^float/1 empty-float-array (empty-of-type Float/TYPE))
+(defonce ^double/1 empty-double-array (empty-of-type Double/TYPE))
 (defonce ^"[Ljava.lang.Object;" empty-object-array (empty-of-type Object))
```
vlaaad: `Object/1`?

editor/src/clj/util/array.clj
```diff
@@ -13,27 +13,84 @@
...
+  (condp = (class array)
+    boolean/1 (alength ^boolean/1 array)
+    char/1 (alength ^char/1 array)
+    byte/1 (alength ^byte/1 array)
+    short/1 (alength ^short/1 array)
+    int/1 (alength ^int/1 array)
+    long/1 (alength ^long/1 array)
+    float/1 (alength ^float/1 array)
+    double/1 (alength ^double/1 array)
+    (alength ^"[Ljava.lang.Object;" array)))
```
vlaaad: `^Object/1`?

editor/src/clj/util/array.clj
```diff
@@ -154,3 +218,30 @@
...
+     (condp = (class array)
+       boolean/1 (Arrays/hashCode (Arrays/copyOfRange ^booleans array start end))
+       char/1 (Arrays/hashCode (Arrays/copyOfRange ^chars array start end))
+       byte/1 (Arrays/hashCode (Arrays/copyOfRange ^bytes array start end))
+       short/1 (Arrays/hashCode (Arrays/copyOfRange ^shorts array start end))
+       int/1 (Arrays/hashCode (Arrays/copyOfRange ^ints array start end))
+       long/1 (Arrays/hashCode (Arrays/copyOfRange ^longs array start end))
+       float/1 (Arrays/hashCode (Arrays/copyOfRange ^floats array start end))
+       double/1 (Arrays/hashCode (Arrays/copyOfRange ^doubles array start end))
+       (Arrays/hashCode (Arrays/copyOfRange ^"[Ljava.lang.Object;" array start end))))))
```
vlaaad: Wouldn't it be too inefficient to create new arrays when doing hashing?

editor/src/clj/util/coll.clj
```diff
@@ -719,6 +802,54 @@
...
+(defn first-where
+  "Returns the first element in coll where pred returns true, or nil if there
+  was no matching element. If coll is a map, the elements are key-value pairs."
+  [pred coll]
+  (reduce
+    (fn [_ item]
+      (when (pred item)
+        (reduced item)))
+    nil
+    coll))
```
vlaaad, line 836: Looks a lot like `some`, though I like that you don't have to do the `(when (pred %) (reduced %))` when using it.

editor/src/clj/util/coll.clj
```diff
@@ -767,3 +898,48 @@
    (transduce identity str-rf coll))
   ([sep coll]
    (transduce (interpose (str sep)) str-rf coll)))
+
+(defn consensus
```
vlaaad: That's an unexpected name... I don't think the operation is generic enough to be in `coll`, maybe leave it in `shader-gen` where it's used?

editor/src/clj/util/coll.clj
vlaaad: I generally agree that Clojure using collections as a last arg was a meh design choice. I guess it's fine... it's just... we have so many collection transformation functions that do the same thing, differently... I mean it's fine...

editor/src/clj/util/debug_util.clj
```diff
@@ -287,3 +291,78 @@
...
+  [object seen return-raw?]
+  (when (some? object)
+    (let [class (class object)
+          key-namespace (.getSimpleName class)
+          object-id (System/identityHashCode object)]
+      (if (contains? seen object-id)
+        (array-map (keyword key-namespace "id") object-id)
+        (let [seen (conj seen object-id)]
+          (cond
+            (or (coll? class)
```
vlaaad: class is never a coll, no?

editor/src/clj/util/ensure.clj
```diff
@@ -0,0 +1,57 @@
...
+(set! *unchecked-math* :warn-on-boxed)
+
+(defmacro argument
+  ([argument-sym pred]
+   `(argument ~argument-sym ~pred "Invalid %s argument: %s" ~argument-sym))
+  ([argument-sym pred message-fmt & extra-message-fmt-args]
+   {:pre [(symbol? argument-sym)
+          (string? message-fmt)]}
+   `(if (~pred ~argument-sym)
+      nil
```
vlaaad, line 27: If we introduce some `assert` replacement, I'd prefer it to be threadable (`->`), so it returns the validated argument.

editor/src/clj/util/ensure.clj
```diff
@@ -0,0 +1,57 @@
...
+     ~argument-sym
+     #(satisfies? ~expected-protocol-sym %)
+     "%s argument must satisfy the protocol %s - got %s"
+     (.getName ^Class (:on-interface ~expected-protocol-sym))
+     (some-> ~argument-sym class .getName)))
+
+(defmacro argument-type
+  [argument-sym expected-class-sym]
+  {:pre [(symbol? argument-sym)
+         (symbol? expected-class-sym)
```
vlaaad, line 50: But if we do it threadable, then it doesn't have to be a symbol. I think it's useful to be able to write code like this:
```clj
(-> (load-stuff)
    :key
    (ensure/argument-type String)
    (do-stuff))
```

But it's not safe to not use symbols here — they are embedded into different places in code, which will lead to double evaluation. I dunno. I guess it's fine for a very limited use... I would also like to be able to use function predicates here.

editor/test/util/coll_test.clj
```diff
@@ -35,6 +36,119 @@
...
+      (testing "No transducer."
+        (doseq [coll colls]
+          (testing (.getSimpleName (class coll))
+            (let [transformed-coll (coll/transform coll)]
+              (is (identical? coll transformed-coll))))))
+
+      (testing "Single transducer."
+        (doseq [coll colls]
+          (testing (.getSimpleName (class coll))
+            (tap> (.getSimpleName (class coll)))
```
vlaaad: Leftover `tap>`

editor/test/util/coll_test.clj
```diff
@@ -35,6 +36,119 @@
...
+            (let [transformed-coll (coll/transform coll
+                                     (map (fn [entry]
+                                            (tap> entry)
```
vlaaad: Leftover `tap>`

editor/test/editor/graphics/types_test.clj
```diff
@@ -0,0 +1,185 @@
...
+  (is (= {:byte :type-byte
+          :ubyte :type-unsigned-byte
+          :short :type-short
+          :ushort :type-unsigned-short
+          :int :type-int
+          :uint :type-unsigned-int
+          :float :type-float}
+         (into {}
+               (map (juxt identity graphics.types/buffer-data-type-data-type))
+               (disj editor.buffers/buffer-data-types :double :long)))))
```
vlaaad: `editor.buffers` should be `:require`-d

editor/test/editor/graphics/types_test.clj
```diff
@@ -0,0 +1,185 @@
...
+          :type-short Short/BYTES
+          :type-unsigned-short Short/BYTES
+          :type-int Integer/BYTES
+          :type-unsigned-int Integer/BYTES
+          :type-float Float/BYTES}
+         (into {}
+               (map (juxt identity graphics.types/data-type-byte-size))
+               graphics.types/data-types))))
+
+(pb-ints-by-val Graphics$CoordinateSpace)
```
vlaaad: Leftover form?

editor/src/clj/util/num.clj
```diff
@@ -177,3 +201,8 @@
 
 (definline normalized->uint [num]
   `(uint (normalized->uint-double ~num)))
+
+(definline normalized->long [num]
+  (if *unchecked-math*
```
vlaaad, line 206: When we `(set! *unchecked-math* :warn-on-boxed)`, does it also mean we make math ops unchecked? Looks like we are, huh:
```clj
(add-doc-and-meta *unchecked-math*
  "While bound to true, compilations of +, -, *, inc, dec and the
  coercions will be done without overflow checks. While bound
  to :warn-on-boxed, same behavior as true, and a warning is emitted
  when compilation uses boxed math. Default: false."
  {:added "1.3"})
```
I always assumed we use `(set! *unchecked-math* :warn-on-boxed)` only to avoid boxing, but it looks like it also turns off overflow checking... Not sure if we ever want this?

editor/test/util/coll_test.clj
```diff
@@ -1584,6 +1710,128 @@
...
+
+  (testing "stops calling pred after first true"
+    (let [pred (fn/make-call-logger fn/constantly-true)]
+      (is (= 0 (coll/first-where pred (range 10))))
+      (is (= 1 (count (fn/call-logger-calls pred)))))))
+
+(deftest first-index-where-test
+  (is (= 1 (coll/first-index-where even? (range 1 4))))
+  (is (= 2 (coll/first-index-where nil? [:a :b nil :d])))
+  (is (= 3 (coll/first-index-where #(^[char] Character/isDigit %) "abc123def")))
```
vlaaad: BTW you can just do `(first-index-where ^[char] Character/isDigit "abc123def")`

editor/src/clj/editor/render_util.clj
```diff
@@ -0,0 +1,203 @@
...
+   :batch-key gpu-texture
+   :passes [pass/transparent]
+   :user-data {:x x
+               :y y
+               :width width
+               :height height
+               :page-index page-index
+               :gpu-texture gpu-texture}})
+
+(defn make-outlined-textured-quad-scene
```
vlaaad, line 195: Mark as SDK API?

editor/src/clj/editor/render_util.clj
```diff
@@ -0,0 +1,203 @@
...
+  {:pre [(set? tags)
+         (every? keyword? tags)]}
+  {:render-fn render-aabb-outline
+   :tags (coll/merge #{:outline} tags)
+   :batch-key ::aabb-outline
+   :select-batch-key :not-rendered
+   :passes [pass/outline]})
+
+(def ^{:arglists '([tags])} make-aabb-outline-renderable
+  (fn/memoize make-aabb-outline-renderable-raw))
```
vlaaad, line 119: Mark as SDK API?

vlaaad (changes requested):
The improvements are great! I found some smaller issues, and I also have some questions; please see the comments.

editor/src/clj/editor/buffers.clj
matgis: Um. Yes. I'll fix!

editor/src/clj/editor/buffers.clj
matgis: Yeah, I think you're right! Nice catch. I'll fix.

editor/src/clj/editor/gl/shader.clj
matgis: I guess I will. 🤷

editor/src/clj/editor/mesh.clj
matgis: Indeed, you are right. I think I intended to change them to keywords, but got distracted. I'll just change it to use the name for now.

editor/src/clj/editor/shaders.clj
matgis: Yes it should. Nice catch! I'll fix.

editor/src/clj/editor/model_scene.clj
matgis: Yep! Must have missed that one! Thanks!

editor/src/clj/editor/model_scene.clj
matgis: Well, it's correct to put in a `nil` here if there is no data for the UV channel, but we weren't handling that correctly in `model-util/make-attribute-binding-entries`, so thanks for highlighting this!

editor/src/clj/editor/model_util.clj
matgis: There is a specific code path where a `scene-node-id` is required, but it is unused otherwise. Specifically, we need to assign a unique `request-id` to each `TransformedAttributeBufferLifecycle` we create. That code path is never hit from the places where the two-argument arity is called. But yeah, it feels a bit like laying a trap for our future selves. I think I'll make all arguments mandatory.

editor/src/clj/internal/util.clj
matgis: Indeed it should!

editor/src/clj/util/ensure.clj
matgis: Makes sense!

editor/src/clj/util/ensure.clj
matgis: Hmm, you make a good point about how it would be more useful if it returned the value. However, as you noted, it currently assumes symbols for simplicity. I think we should leave it for now and make a separate PR with this improvement when the need arises. For now, at least the asserts prevent misuse that would lead to bugs until we can improve it.

editor/test/util/coll_test.clj
matgis: Ouch. Nice catches!

editor/test/editor/graphics/types_test.clj
matgis: You're damn right it should! 😅

editor/test/editor/graphics/types_test.clj
matgis: Hmm, yes. Honestly, I left a bit of a mess all around this PR. Not sure what happened!

editor/src/clj/util/num.clj
matgis: I think we want overflow checks disabled for most of what we're doing, honestly. I feel like bugs resulting from overflow are rare, and we want the code to be as performant as possible. I'm quite happy to give up overflow checks to know where boxed math is happening.

editor/test/util/coll_test.clj
matgis: Nice!

editor/src/clj/editor/render_util.clj
matgis: I suppose so!

editor/src/clj/editor/render_util.clj
matgis: Yep!

editor/src/clj/util/debug_util.clj
matgis: Whoops.

editor/src/clj/util/coll.clj
matgis: Is it? There are multiple definitions, but these seem quite close to what I was going for:
* An agreement between two or more parties on an issue
* General voice speaking as one accord

I think it has utility as a general function in `coll`, but the name could maybe be better.
* `unanimity` is "A state of being in unanimous agreement by all involved parties".
* `unity` is, amongst other things, "The state or condition of being equal, identical or alike".
* `unison` is "The state of acting together, at the same time, or as one", "The quality of forming a pleasing and consistent whole", "The state or quality of being identical", and "General voice speaking as one accord". This one seems pretty good to me. 🤔

vlaaad: Everything is equal... Socialism 🤪

editor/src/clj/util/coll.clj
matgis: Yeah, that was the idea. I feel this is mostly what I use `some` for. But we already had this function. I just moved it to `coll`.

editor/src/clj/util/array.clj
matgis: It's not ideal, but I think the hashing part would actually be quite efficient. The allocation and copying part, not so much. We could do better for sure, but we'd likely need different implementations for each primitive type. I think we may be hitting the happy path of hashing the entire range for our current usage anyway. I'll add a TODO comment, but I suggest we keep it like this for now.

editor/src/clj/util/coll.clj
matgis: `coll/socialism` is tempting, but I think I'll go with `unanimous-value` for now! 😅

vlaaad: I think `unify` is a good name, and it's similar to `unify-values` in properties ns that does the same thing!

editor/src/clj/editor/buffers.clj
```diff
@@ -242,7 +242,7 @@
        0)))
 
 (definline empty? [^Buffer buffer]
-  `(pos? (item-count ~buffer)))
+  `(not (pos? (item-count ~buffer))))
```
vlaaad: `zero?`?

editor/src/clj/editor/math.clj
matgis: It should be fine to return a shared instance of the identity matrix. We shouldn't mutate these matrices in-place, and we are not currently doing that as far as I can tell.

vlaaad (approved):
Ship it!

editor/src/clj/editor/tile_source.clj
```diff
@@ -293,7 +260,7 @@
                             :tile-source-attributes tile-source-attributes
                             :anim-data   (get anim-data id)
                             :start-tile  start-tile}
-                :passes    [pass/outline pass/overlay pass/selection]}
```
AGulev, line 296: removing of the selection pass in tile source broke animation play on `spacebar`:
https://forum.defold.com/t/defold-1-12-0-beta/81981/63?u=agulev

matgis: A fix has been merged.

PR #11518, opened 2025-11-18
Build runtime-generated fonts from the editor

matgis:
The editor now respects the `font.runtime_generation` setting in `game.project`. Previously only Bob would produce runtime-generated fonts when this setting was enabled for the project.

Fixes #10999.

### DIFF
--- a/editor/src/clj/editor/font.clj
+++ b/editor/src/clj/editor/font.clj
@@ -33,6 +33,7 @@
             [editor.properties :as properties]
             [editor.protobuf :as protobuf]
             [editor.resource :as resource]
+            [editor.resource-io :as resource-io]
             [editor.resource-node :as resource-node]
             [editor.scene-cache :as scene-cache]
             [editor.validation :as validation]
@@ -60,6 +61,7 @@
 
 (def ^:private font-file-extensions ["ttf" "otf" "fnt"])
 (def ^:private font-icon "icons/32/Icons_28-AT-Font.png")
+(def ^:private font-label (localization/message "resource.type.font"))
 
 (def ^:private alpha-slider-edit-type {:type :slider
                                        :min 0.0
@@ -539,7 +541,7 @@
   (let [base-path (resource/parent-proj-path (resource/proj-path font))]
     (fn [path]
       (let [proj-path (str base-path "/" path)]
-        (resource-map proj-path)))))
+        (get resource-map proj-path)))))
 
 (g/defnk produce-font-map [_node-id font type font-resource-map save-value]
   ;; TODO(save-value-cleanup): make-font-map expects all values to be present.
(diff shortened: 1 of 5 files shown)


editor/src/clj/editor/font.clj
```diff
@@ -539,7 +541,7 @@
   (let [base-path (resource/parent-proj-path (resource/proj-path font))]
     (fn [path]
       (let [proj-path (str base-path "/" path)]
-        (resource-map proj-path)))))
+        (get resource-map proj-path)))))
```
matgis, line 544: `resource-map` can now be `nil`, depending on what kind of `FontSourceNode` is connected to the `FontNode`.

editor/src/clj/editor/font.clj
```diff
@@ -606,33 +615,47 @@
...
+                                (FilenameUtils/normalize)
+                                (Paths/get (into-array String []))
+                                (.getFileName)
+                                (.toString))
+          texture-resource (workspace/resolve-resource resource texture-file-name)]
+      (concat
+        (g/set-property self :texture texture-resource)
+        (when disk-sha256
+          (let [workspace (resource/workspace resource)]
+            (workspace/set-disk-sha256 workspace self disk-sha256)))))))
```
matgis, line 643: Only an indentation change from removing the surrounding file extension check. We now have different `FontSourceNode` implementations for `.fnt`, `.ttf`, and `.otf` resources.

editor/src/clj/editor/font.clj
```diff
@@ -606,33 +615,47 @@
...
+          (g/fnk [_node-id project-settings resource]
+            (when (get project-settings ["font" "runtime_generation"])
+              (resource-io/with-error-translation resource _node-id :runtime-generation-build-target
+                (pipeline/make-source-bytes-build-target _node-id resource))))))
+
+(defn load-true-type-font-source [project self _resource]
+  (g/connect project :settings self :project-settings))
+
+(g/defnode OpenTypeFontSourceNode
+  (inherits resource-node/ResourceNode))
```
matgis, line 658: We currently only use the `resource` output for `.otf` resources, but we might add runtime generation support for these in the future, pending engine support.

editor/src/clj/editor/font.clj
```diff
@@ -893,10 +917,25 @@
...
+      :label font-label
+      :node-type TrueTypeFontSourceNode
+      :load-fn load-true-type-font-source
+      :stateless? true
+      :icon font-icon
+      :view-types [:default])
+    (workspace/register-resource-type workspace
+      :ext "otf"
+      :label font-label
+      :node-type OpenTypeFontSourceNode
```
matgis, line 938: Note: Both the `.ttf` and `.otf` `resource-types` are `:stateless?`. The `.otf` `resource-type` will be implicitly `:stateless?`, since it doesn't specify a `:load-fn`.

editor/src/clj/editor/game_project.clj
```diff
@@ -106,14 +99,6 @@
...
   (make-writer        [this opts] (io/writer resource)))
 
-(defn- make-custom-build-target [node-id resource]
-  (bt/with-content-hash
-    {:node-id node-id
-     :resource (workspace/make-build-resource (CustomResource. resource))
-     :build-fn build-custom-resource
-     ;; NOTE! Break build cache when resource content changes.
-     :user-data {:hash (murmur/hash64-bytes (resource-content resource))}}))
-
```
matgis, line 116: Moved to `pipeline/make-source-bytes-build-target`.

editor/src/clj/editor/pipeline.clj
```diff
@@ -99,6 +99,26 @@
...
+        source-bytes (resource/resource->bytes source-resource)]
+    {:resource build-resource
+     :content source-bytes}))
+
+(defn make-source-bytes-build-target [node-id source-resource]
+  ;; Warning: May throw IOException.
+  ;; We hash the source resource contents to ensure the build target is
+  ;; invalidated from the on-disk build cache when the contents change.
+  (let [build-resource (workspace/make-build-resource source-resource)
+        source-hash (resource/resource->path-inclusive-sha1-hex source-resource)]
```
matgis, line 113: Note: We include the source path in the hash here, since it is possible to act on these resources at runtime. We don't want to fuse the build targets.

vlaaad (approved):
LGTM!

PR #11519, opened 2025-11-18
ASTC encoder fixes

AGulev:
- Updated the ASTC encoder to version 5.3.0
- Enabled the release flag to ensure full performance
- Fixed the ASTC 4x4 RGB build


.github/workflows/build-astc.yml
```diff
@@ -0,0 +1,148 @@
+name: Build ASTC
```
AGulev, line 1: new workflow to build ASTCen lib for all the platforms

share/ext/astcenc/build_astcenc.sh
```diff
@@ -36,22 +35,30 @@ CONFIG=Release
 
 CMAKE_FLAGS="-DASTCENC_CLI=OFF ${CMAKE_FLAGS}"
 CMAKE_FLAGS="-DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded ${CMAKE_FLAGS}"
+# The default Unix Makefile/Ninja generators build Debug unless build type is set explicitly.
+# Our macOS and Linux builds use single-config generators, so make sure we get an optimized lib.
+if [ "$PLATFORM" != "x86_64-win32" ]; then
+    CMAKE_FLAGS="-DCMAKE_BUILD_TYPE=${CONFIG} ${CMAKE_FLAGS}"
+fi
```
AGulev: the main fix

share/ext/astcenc/build_astcenc.sh
```diff
@@ -36,22 +35,30 @@ CONFIG=Release
...
 
 case $PLATFORM in
     arm64-macos)
         CMAKE_FLAGS="-DCMAKE_OSX_ARCHITECTURES=arm64 ${CMAKE_FLAGS}"
         CMAKE_FLAGS="-DCMAKE_OSX_DEPLOYMENT_TARGET=${OSX_MIN_SDK_VERSION} ${CMAKE_FLAGS}"
+        CMAKE_FLAGS="-DASTCENC_ISA_NEON=ON -DASTCENC_PACKAGE=arm64 ${CMAKE_FLAGS}"
```
AGulev: make sure all the flags used as in https://github.com/ARM-software/astc-encoder/blob/main/.github/workflows/post_weekly_release.yaml

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/TextureGenerator.java
```diff
@@ -203,11 +203,6 @@ else if (componentCount == 3)
                     return TextureFormat.TEXTURE_FORMAT_RGB_BC1;
                 return TextureFormat.TEXTURE_FORMAT_RGBA_ETC2;
             }
-            case TEXTURE_FORMAT_RGBA_ASTC_4X4 -> {
-                if (componentCount < 4)
-                    return TextureFormat.TEXTURE_FORMAT_RGB_BC1;
-                return TextureFormat.TEXTURE_FORMAT_RGBA_ASTC_4X4;
-            }
```
AGulev, line 210: we don't support BC1 so ASTC_4X4 is fine for it

share/ext/astcenc/build_astcenc.sh
```diff
@@ -34,36 +60,93 @@ OSX_MIN_SDK_VERSION=$VERSION_MACOSX_MIN
...
+        CMAKE_FLAGS+=(-DASTCENC_PACKAGE=x64)
         # Need to setup clang on linux
         cmi_setup_cc $PLATFORM
         ;;
     x86_64-win32)
-
-        # NOTE! For windows we need to use the exact same version of the toolset as we use for the engine
-        #       However, I could not get the cmake setup to work, so I just installed the same version via the visual studio installer.
-        #       Leaving the almost-working-code here for posterity..
-        #
```
AGulev, line 63: Pay attention this comment. I can't just build ASTC lib, it has to be the exact version of the compiler we uses for the engine, otherwise it fails on ASTC test (can't load library?). It took me some time to figure out that and rewrite build to use the same VSStudio

Build should be green now

share/ext/astcenc/build_astcenc.sh
```diff
@@ -94,10 +177,10 @@ case $PLATFORM in
         LIB_NAME=libastcenc-neon-static.a
         ;;
     x86_64-linux)
-        LIB_NAME=libastcenc-native-static.a
+        LIB_NAME=libastcenc-avx2-static.a
```
JCash: Is the avx2 version fallback compatible?
As not all devices have avx2, it may be problematic otherwise?

JCash (dismissed):
lgtm
small question about avx2 support on devices

share/ext/astcenc/build_astcenc.sh
AGulev: I checked docs and their releases, for linux they build 3 separate binaries so users can pick depends on their CPU

So for that PR I suggest using sse2 - the slowest (~30% slower than avx2) but the most supported lib. Then in the future we may consider building all 3 and load one depends on user's machine

AGulev: or maybe sse4.1 like on windows

AGulev: Here is task https://github.com/defold/defold/issues/11521

share/ext/astcenc/build_astcenc.sh
JCash: Why not use the same setup for all platforms?
Less code, and it shouldn't hurt?

JCash (dismissed):
lgtm. small question about the config setup

share/ext/astcenc/build_astcenc.sh
AGulev: DCMAKE_BUILD_TYPE ignored on windows, only `--config $CONFIG`
but we can probably remove condition and keep just both (as for `--config $CONFIG`)

JCash (approved):

PR #11537, opened 2025-11-24
Add multithreaded encoding for ASTC encoder

AGulev:
Speed up the ASTC encoder using a few threads per image (in addition to encoding a few images at the same time)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,5 @@ editor/.lsp/sqlite.db
 
 # ChatGPT
 AGENTS.md
+/share/ext/astcenc/build/
+/share/ext/astcenc/source
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/defold/extension/pipeline/texture/TextureCompressorASTC.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/defold/extension/pipeline/texture/TextureCompressorASTC.java
@@ -16,6 +16,7 @@ package com.defold.extension.pipeline.texture;
 
 import com.dynamo.bob.pipeline.Texc;
 import com.dynamo.bob.pipeline.TexcLibraryJni;
+import com.dynamo.bob.pipeline.TextureGenerator;
 import com.dynamo.graphics.proto.Graphics;
 import com.dynamo.graphics.proto.Graphics.TextureImage;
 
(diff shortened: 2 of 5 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/defold/extension/pipeline/texture/TextureCompressorBasisU.java
```diff
@@ -105,7 +106,7 @@ public byte[] compress(TextureCompressorPreset preset, TextureCompressorParams p
         settings.outPixelFormat = Texc.PixelFormat.fromValue(params.getPixelFormatOut());
         settings.colorSpace = Texc.ColorSpace.fromValue(params.getColorSpace());
 
-        settings.numThreads = 4;
+        settings.numThreads = TextureGenerator.maxThreads;
```
AGulev: I found that we use hardcoded value instead of a value were setup for it

.gitignore
```diff
@@ -81,3 +81,5 @@ editor/.lsp/sqlite.db
 
 # ChatGPT
 AGENTS.md
+/share/ext/astcenc/build/
+/share/ext/astcenc/source
```
AGulev, line 85: Ignore folders used to build astc locally

com.dynamo.cr/com.dynamo.cr.bob/src/com/defold/extension/pipeline/texture/TextureCompressorASTC.java
```diff
@@ -138,11 +139,7 @@ public byte[] compress(TextureCompressorPreset preset, TextureCompressorParams p
         settings.height = params.getHeight();
         settings.pixelFormat = Texc.PixelFormat.fromValue(params.getPixelFormat());
         settings.colorSpace = Texc.ColorSpace.fromValue(params.getColorSpace());
-        // The library doesn't create threads for us. To use the multithreaded version, we need to spawn our own threads or workers.
-        // Because we encode many images in parallel, we can stick to single threaded encoding and still achieve parallelisation
-        // by processing multiple images simultaneously.
-         // https://github.com/ARM-software/astc-encoder/blob/701503966b1ac2ebd2616cba94adee5ae8ba6363/Source/astcenc.h#L37C2-L78C41
-        settings.numThreads = 1;
+        settings.numThreads = TextureGenerator.maxThreads;
```
JCash, line 142: I think this is fine for now, but I think in this setting should come from the TextureCompressorParams, as it (imho) should be configurable in the presets.

JCash (approved):

PR #11540, opened 2025-11-25
Add support for http proxy connections

### LINKED ISSUE #11447: Defold native extension, to respect and apply OS HTTP Proxy settings for the network calls
It would be nice to create a new Defold native extension, to act as an intermediary between the client networking code and defold.http.lua.

This would greatly aid developers who're debugging their network traffic with proxy / monitoring tools (eg Charles).

The idea is to create an extension that the client code could use for networking. This extension would be polling the corresponding OS APIs (eg _CFNetworkCopyProxiesForAutoConfigurationScript_, in case of iOS) for the HTTP proxy host and port specified by the user in OS System Settings. Then the extension would reroute any incoming HTTP calls to that proxy (specifying the real destination address as HTTP Host header value, etc).

That would allow to dynamically configure network traffic proxying without needing to rebuild the client app, which would facilitate debugging.

Right now, one'd have to use reverse-DNS approach, because simply specifying an HTTP Proxy IP address in OS settings doesn't make Defold http.request() to route API calls through the proxy address.

britzl:
HTTP requests can now specify a proxy server to use when connecting to the destination server. When a proxy is used the connection to the destination server is established using an a HTTP tunnel through the proxy. The HTTP tunnel is established using the [CONNECT HTTP method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/CONNECT).

```lua
	-- connect to www.defold.com via localhost proxy on port 8888
	local url = "https://www.defold.com:443"
	local method = "GET"
	local headers = {}
	local post_data = nil
	local options = {
		proxy = "https://127.0.0.1:8888"
	}
	http.request(url, method, function(self, id, response)
		pprint(response)
	end, headers, post_data, options)
```

Fixes #11447


engine/dlib/src/dlib/connection_pool.cpp
```diff
@@ -301,8 +301,28 @@ namespace dmConnectionPool
...
+    {
+        DM_MUTEX_SCOPED_LOCK(pool->m_Mutex);
+
+        Connection* c = GetConnection(pool, connection);
+
+        return CreateSSLSocket(c->m_Socket, host, timeout, &c->m_SSLSocket);
+    }
+
+    static Result Connect(HPool pool, const char* host, dmSocket::Address address, uint16_t port, int timeout,
+                                    dmSocket::Socket* socket, dmSocket::Result* sr)
```
britzl: The Connect() function will only create the normal TCP socket. Creation of the SSL socket now happens in a separate function.

For all connections when no proxy is involved a call to Connect() will be immediately followed by a call to CreateSSLSocket() (assuming that https is used).

When a proxy is used a call to Connect() is made to get TCP socket and do a first header exchange. If the connection is successfully established and https is used a call to CreateSSLSocket() is made before communication through the tunnel.

engine/dlib/src/dlib/connection_pool.cpp
```diff
@@ -313,26 +333,13 @@ namespace dmConnectionPool
...
-        uint64_t handshakestart = dmTime::GetMonotonicTime();
-        if( timeout > 0 && (handshakestart - connectstart) > (uint64_t)timeout )
+        uint64_t now = dmTime::GetMonotonicTime();
+        if( timeout > 0 && (now - connectstart) > (uint64_t)timeout )
         {
             dmSocket::Delete(*socket);
             *socket = dmSocket::INVALID_SOCKET_HANDLE;
             return RESULT_SOCKET_ERROR;
         }
-
```
britzl, line 323: Here's the part split to CreateSSLSocket()

engine/dlib/src/dlib/connection_pool.cpp
```diff
@@ -415,15 +431,19 @@ namespace dmConnectionPool
     Result Dial(HPool pool, const char* host, uint16_t port, bool ssl, int timeout, int* cancelflag, HConnection* connection, dmSocket::Result* sock_res)
     {
         // try connecting to the host using ipv4 first
+        bool ipv4 = true;
```
britzl, line 442: For the sake of clarity

engine/dlib/src/dlib/connection_pool.cpp
```diff
@@ -301,8 +301,37 @@ namespace dmConnectionPool
         return RESULT_OK;
     }
 
-    static Result Connect(HPool pool, const char* host, dmSocket::Address address, uint16_t port, bool ssl, int timeout,
-                                    dmSocket::Socket* socket, dmSSLSocket::Socket* sslsocket, dmSocket::Result* sr)
+
+    static Result CreateSSLSocket(dmSocket::Socket socket, const char* host, int timeout, dmSSLSocket::Socket* sslsocket, dmSocket::Result* sock_res)
```
britzl, line 305: Splitting the Socket creation from the SSL Socket creation. For proxy connections you'd like to start with a CONNECT request using a normal TCP socket and then once the tunnel is established you can proceed to set up the SSL Socket and communicate with the server through the tunnel.

engine/dlib/src/dlib/connection_pool.cpp
```diff
@@ -500,7 +528,13 @@ namespace dmConnectionPool
                     // Due to threaded behavior, the connection may be in use,
                     // but the socket's haven't been assigned yet
                     if(c->m_Socket != dmSocket::INVALID_SOCKET_HANDLE)
-                        dmSocket::Shutdown(c->m_Socket, how);
+                    {
+                        dmSocket::Result r = dmSocket::Shutdown(c->m_Socket, how);
+                        if (r == dmSocket::RESULT_OK)
+                        {
+                            c->m_Socket = dmSocket::INVALID_SOCKET_HANDLE;
+                        }
```
britzl: We never cleared the socket handle after successful shutdown

engine/dlib/src/dlib/connection_pool.cpp
```diff
@@ -518,12 +552,9 @@ namespace dmConnectionPool
         uint32_t n = pool->m_Connections.Size();
         for (uint32_t i=0; i != n; i++) {
             Connection* c = &pool->m_Connections[i];
-            if (c->m_State == STATE_CONNECTED)
```
britzl, line 521: We should do cleanup always, regardless of connection state

engine/dlib/src/dlib/configfile.cpp
```diff
@@ -695,7 +695,7 @@ namespace dmConfigFile
         params.m_Userdata = &context;
         params.m_HttpContent = &HttpContent;
         params.m_HttpHeader = &HttpHeader;
-        dmHttpClient::HClient client = dmHttpClient::New(&params, uri_parts.m_Hostname, uri_parts.m_Port, strcmp(uri_parts.m_Scheme, "https") == 0, 0);
+        dmHttpClient::HClient client = dmHttpClient::New(&params, (dmURI::Parts*)&uri_parts, 0);
```
JCash: No need to cast to `(dmURI::Parts*)` here, as it's already what it is. I think the cast is actually confusing: "Wait, so it _isn't_  a dmURI:Parts struct?"

engine/dlib/src/dlib/connection_pool.cpp
JCash: It's likely part of the paradigm of "initialize when starting to use", as opposed to setting every member on _both_ destroy/create events. (Less code=smaller engine and all that).

Is the if-statement necessary?
If the shutdown fails, is the handle still valid?

engine/resource/src/providers/provider_http.cpp
```diff
@@ -173,7 +173,7 @@ static dmResourceProvider::Result Mount(const dmURI::Parts* uri, dmResourceProvi
     http_params.m_HttpWriteHeaders = &HttpWriteHeaders;
     http_params.m_Userdata = archive;
     http_params.m_HttpCache = archive->m_HttpCache;
-    archive->m_HttpClient = dmHttpClient::New(&http_params, uri->m_Hostname, uri->m_Port, strcmp(uri->m_Scheme, "https") == 0, 0);
+    archive->m_HttpClient = dmHttpClient::New(&http_params, (dmURI::Parts*)uri, 0);
```
JCash: Same thing here with the casting.

engine/dlib/src/dlib/http_client.h
```diff
@@ -183,22 +183,29 @@ namespace dmHttpClient
...
      * @param params Parameters
-     * @param hostname Hostname
-     * @param port Port number
-     * @param secure TLS/SSL or not
+     * @param hostURI Host URI (hostname, port, scheme etc)
      * @param if non null and set, aborts the call as soon as possible
      * @return HTTP-client handle on success. 0 on failure.
      */
-    HClient New(const NewParams* params, const char* hostname, uint16_t port, bool secure, int* cancelflag);
+    HClient New(const NewParams* params, dmURI::Parts* hostURI, int* cancelflag);
```
JCash: Regarding the casting, I believe it's because it's not "const dmURI::Parts*" here.
By setting it to const here, the callers don't have to cast it.

JCash (dismissed):
Nice!

Only a small issue with the casting of the dmURI::Parts*.

engine/dlib/src/dlib/configfile.cpp
britzl: Ah, ok, the problem was that I got an issue with "losing const qualifier" when passing to dmHttpClient::New()

engine/dlib/src/dlib/connection_pool.cpp
britzl: > Is the if-statement necessary? If the shutdown fails, is the handle still valid?

You're right. No need to check the result.

engine/dlib/src/dlib/configfile.cpp
britzl: Fixed.

engine/dlib/src/dlib/http_client.h
britzl: Yep that was it.

engine/dlib/src/dlib/connection_pool.cpp
```diff
@@ -530,10 +530,7 @@ namespace dmConnectionPool
                     if(c->m_Socket != dmSocket::INVALID_SOCKET_HANDLE)
                     {
                         dmSocket::Result r = dmSocket::Shutdown(c->m_Socket, how);
-                        if (r == dmSocket::RESULT_OK)
-                        {
-                            c->m_Socket = dmSocket::INVALID_SOCKET_HANDLE;
-                        }
+                        c->m_Socket = dmSocket::INVALID_SOCKET_HANDLE;
```
JCash, line 533: now the "r" is likely generating a warning of being "unused" ?

JCash (dismissed):
lgtm.
nitpick about adding the warning

PR #11541, opened 2025-11-25
Performance optimisations for atlas creation

AGulev:
These improvements reduced the total atlas build time in Bob.jar from ~2850ms down to ~1100ms for a tested atlas (8192x4096, 4818 images on MBP M1 Max), including a drop in layout calculation time from ~1300ms down to ~490ms. With the reduced algorithmic complexity, adding more images will affect performance far less than before.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/TextureGenerator.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/TextureGenerator.java
@@ -259,14 +259,28 @@ public class TextureGenerator {
 
     private static List<Long> GenerateImages(long image, int width, int height, boolean generateMipChain) throws TextureGeneratorException {
         List<Long> images = new ArrayList<>();
-        int mipWidth = width;
-        int mipHeight = height;
-        int mipLevel = 0;
-        long prevImage = image;
+        int baseWidth = TexcLibraryJni.GetWidth(image);
+        int baseHeight = TexcLibraryJni.GetHeight(image);
+        boolean baseMatches = baseWidth == width && baseHeight == height;
 
-        while (mipWidth != 0 || mipHeight != 0) {
-            mipWidth = Math.max(mipWidth, 1);
-            mipHeight = Math.max(mipHeight, 1);
+        // Use the provided image as mip0 if it matches the requested size; otherwise resize once to seed the chain.
+        if (baseMatches) {
+            images.add(image);
+        } else {
+            long resizedBase = TexcLibraryJni.Resize(image, width, height);
+            if (resizedBase == 0) {
+                throw new TextureGeneratorException("Failed to create mipmap 0");
+            }
+            images.add(resizedBase);
+        }
+        if (!generateMipChain) {
+            return images;
+        }
+
+        long prevImage = images.get(images.size() - 1);
+        for (int mipLevel = 1, nextWidth = width / 2, nextHeight = height / 2; nextWidth > 0 || nextHeight > 0; mipLevel++, nextWidth /= 2, nextHeight /= 2) {
+            int mipWidth = Math.max(nextWidth, 1);
+            int mipHeight = Math.max(nextHeight, 1);
             long resizedImage;
 
             TimeProfiler.start("ResizeMipLevel" + mipLevel);
@@ -278,16 +292,7 @@ public class TextureGenerator {
 
             images.add(resizedImage);
 
-            // If we don't need all the mipmap images, just return the first.
-            if (!generateMipChain) {
-                return images;
-            }
-
             prevImage = resizedImage;
-
-            mipLevel++;
-            mipWidth /= 2;
-            mipHeight /= 2;
         }
 
         return images;
@@ -472,7 +477,10 @@ public class TextureGenerator {
 
             // Cleanup the texture images
             for (Long mipImage : mipImages) {
-                TexcLibraryJni.DestroyImage(mipImage);
+                // Avoid double-destroying the original image handle (cleaned up in the finally block).
+                if (mipImage != textureImage) {
+                    TexcLibraryJni.DestroyImage(mipImage);
+                }
             }
 
             return compressedMipImageDatas;
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/MaxRectsLayoutStrategy.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/MaxRectsLayoutStrategy.java
@@ -62,7 +62,7 @@ public class MaxRectsLayoutStrategy implements TextureSetLayoutStrategy {
         }
 
         ArrayList<Page> pages = new ArrayList<Page>();
-        while (srcNodes.size() > 0) {
+        while (!srcNodes.isEmpty()) {
             Page result = packPage(srcNodes);
             pages.add(result);
             srcNodes = result.remainingRects;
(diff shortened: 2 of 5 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/TextureGenerator.java
```diff
@@ -259,14 +259,28 @@ private static ITextureCompressor getDefaultTextureCompressor() {
...
-        long prevImage = image;
-
-        while (mipWidth != 0 || mipHeight != 0) {
-            mipWidth = Math.max(mipWidth, 1);
-            mipHeight = Math.max(mipHeight, 1);
+        int baseWidth = TexcLibraryJni.GetWidth(image);
+        int baseHeight = TexcLibraryJni.GetHeight(image);
+        boolean baseMatches = baseWidth == width && baseHeight == height;
+
+        // Use the provided image as mip0 if it matches the requested size; otherwise resize once to seed the chain.
```
AGulev, line 266: if provided image == mip0, use it instead of calling `TexcLibraryJni.Resize` which makes buffer copy and for tested image it takes about 100ms

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/MaxRectsLayoutStrategy.java
```diff
@@ -174,7 +182,11 @@ private Page packPage(ArrayList<RectNode> inputRects) {
     /** @param fully If true, the only results that pack all rects will be considered. If false, all results are considered, not all
      *           rects may be packed.
      **/
-    private Page packAtSize(boolean fully, int width, int height, ArrayList<RectNode> inputRects) {
+    private Page packAtSize(boolean fully, int width, int height, ArrayList<RectNode> inputRects, RectStats rectStats) {
+        if (!canFitInPage(width, height, rectStats, fully)) {
+            return null;
+        }
```
AGulev, line 188: We know the area of all images. If they can't fit the current atlas size, just skip the heuristics for that size.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/MaxRectsLayoutStrategy.java
```diff
@@ -295,6 +334,8 @@ class MaxRects {
         private final ArrayList<RectNode> usedRectangles = new ArrayList<RectNode>();
         private final ArrayList<RectNode> freeRectangles = new ArrayList<RectNode>();
         private final ArrayList<RectNode> newFreeRectangles = new ArrayList<RectNode>();
+        private final HashMap<Integer, ArrayList<RectNode>> verticalEdges = new HashMap<Integer, ArrayList<RectNode>>();
+        private final HashMap<Integer, ArrayList<RectNode>> horizontalEdges = new HashMap<Integer, ArrayList<RectNode>>();
```
AGulev, line 332: Most of the changes in this file is Implementation edge-indexed contact scoring to avoid scanning all placed rectangles on every contact-point evaluation, except a small fix for image rotation

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/MaxRectsLayoutStrategy.java
```diff
@@ -452,6 +517,8 @@ private RectNode findPositionForNewNodeBottomLeft (int width, int height, int ro
             bestNode.rect = new Rect(null, 0,0,0,0,0);
             bestNode.score1 = Integer.MAX_VALUE; // best y, score2 is best x
 
+            boolean doRotate = rotate && width != height;
```
AGulev, line 514: There is no sense to rotate image if width == height

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/TextureSetGenerator.java
```diff
@@ -746,12 +746,9 @@ private static Pair<TextureSet.Builder, List<UVTransform>> buildData(int width,
 
         AnimDesc animDesc = null;
         while ((animDesc = iterator.nextAnim()) != null) {
-            String animId = animDesc.getId();
```
AGulev, line 749: unused code

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/PackedResources.java
```diff
@@ -18,6 +21,12 @@ public class PackedResources {
 
     private static void runUnpackAllLibsAsync(Platform platform) throws IOException {
         TimeProfiler.start("runUnpackAllLibsAsync");
+        // unpacking for these happens in static initializers
+        new TexcLibraryJni();
+        new ShadercJni();
+        new ModelImporterJni();
```
AGulev: These JNI wrapper classes contain a static initializer that unpacks shared libraries. It’s better to do this in a separate thread at the same place we unpack all the other libs so we save time on iterative builds.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/TextureUtil.java
```diff
@@ -48,6 +48,11 @@
 import static com.dynamo.bob.util.MiscUtil.concatenateArrays;
 
 public class TextureUtil {
+
+    static {
+        ImageIO.setUseCache(false);
```
AGulev, line 53: Using this helped speed up image loading about 6 times: ~1025ms -> ~175ms.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/MaxRectsLayoutStrategy.java
```diff
@@ -217,6 +229,21 @@ static int getExponentNextOrMatchingPowerOfTwo(int value) {
         return exponent;
     }
 
+    private boolean canFitInPage(int width, int height, RectStats stats, boolean fully) {
+        long availableArea = (long)width * (long)height;
+        if (fully && availableArea < stats.totalArea) {
+            return false;
+        }
+
+        if (!settings.rotation) {
```
JCash: would it make sense moving the "settings.rotation" to an argument?
I.e. making this a function, and not a method?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/PackedResources.java
JCash: Looking at the synchronization regarding this function (`PackedResources.waitForuUpackAllLibsAsync()`), the usage of that is a little bit scattered.

I think we need a better and more robust approach.
Currently, we call it via Bob.getExeWithExtension() + EngineArtifactProvider.getExeWithExtension() (yes, they are the same!) and Bob.initLua().

Bob.initLua() is called in the LuaBuilder, but this PR doesn't add any similar call in any texture, shader or model builders, so I think this highlights the need of a more robust approach, as it's easy to miss.

E.g.call the PackedResources.unpackAllLibsAsync() at the start of the Project.build(), and wait for the thread, before starting to create tasks.

I think there is a tradeoff between performance and readability/maintainability here, and I'm not exactly sure where that line goes, but currently, it seems too easy to create a race condition.

JCash (changes requested):
Overall very nice.
Needs some attention to thread synching I believe.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/MaxRectsLayoutStrategy.java
AGulev: fixed! thank you!

AGulev:
Regarding https://github.com/defold/defold/pull/11541#discussion_r2564070289

I removed it from this PR and will prepare a new one with this change

JCash (approved):

PR #11559, opened 2025-12-02
Fix issues around textures shared between Model Scenes

### LINKED ISSUE #9401: Improve editor memory usage when image files are used as textures
**Task (REQUIRED):**
The `gpu-texture-generator` output on `ImageNode` is not optimal. It is currently used by `model/TextureBinding`. Because it is not cached, every model that references an `ImageNode` as a texture will allocate its own GPU texture in VRAM. On top of that, the output uses the `texture-image` output of `ImageNode`, which loads the image data into RAM, even though we don't need it until the `gpu-texture-generator` is invoked.

Improved memory utilization in the editor.

### LINKED ISSUE #9702: Project with a large number of models with textures opens infinitely in IDE
I was trying to make a game with building destruction. Buildings are pre-cut into pieces . It turned out that the engine loads 1000+ objects per scene quite well (in modern large games there are tens of thousands of objects on the scene, so I did not doubt Defold), but the game itself cannot be developed further because of an issue with the IDE.

The point: the project takes an enormously huge amount of time to load in IDE, if the project has a lot of models with large textures.

Texture 128x128:

https://github.com/user-attachments/assets/364826ff-fb0d-4939-aaf4-f5a0982d24ea

Texture 4096x4096 (not finished):

https://github.com/user-attachments/assets/014cb9ae-2d4b-4e15-974d-ce26db296435

**To Reproduce (REQUIRED)**
1. Open the project.
2. Go to '/assets'.
3. Rename 'texture_4096.png' to 'texture.png'.
4. Open the project.

IDE should load the project quickly. 1000 objects and models is not that much!

  - Version: 1.9.4

**Platforms (REQUIRED):**
 - Platforms: Windows 10.

[issue_models_with_4k_texture.zip](https://github.com/user-attachments/files/17650272/issue_models_with_4k_texture.zip)

**Workaround (OPTIONAL):**
Reduce the texture to 128x128, and when building from CI return back to 4096x4096.

matgis:
Texture resources shared across multiple Model Scenes (`.gltf` files) will now properly share VRAM resources.

Fixes #9401.
Fixes #9702.

### Technical changes
After the Scene View camera mode and Outline View object visibility toggles were implemented, some `defhandlers` were evaluating `state` clauses dependent on the `scene` output of the active resource tab. Since `state` was not using the shared `evaluation-context` we create when we refresh the UI, multiple versions of the same `scene` output could end up being produced during refresh. This resulted in multiple instances of `TextureData` being produced for the same `request-id` (based on the source image), and ending up across multiple cached `scene` outputs. The `scene-cache` would then assume the textures needed to be constantly re-uploaded to VRAM as different objects were drawn in the Scene View.

This PR addresses various issues to prevent this from happening:
* All `defhandler` clauses except `run` will now use the shared `evaluation-context` during UI refresh. We were missing `state`, `options`, and `label`.
* We also use the same shared `evaluation-context` when evaluating the `auto-pulls` and refreshing the app title.
* `app-view/refresh-scene-views!` was updated to not perform its graph mutations inside an `auto-evaluation-context` block, and now happens last in the UI update cycle.
* A `texture-util` module was added to consolidate the process of creating `TextureLifecycles` from `content-generators`. All `gpu-texture` outputs have been updated to use the `texture-util` module. This is a non-breaking change as far as our extensions are concerned, but there will be an upcoming PR to `extension-texturepacker`.
* We now wrap the `jogamp` `TextureData` instances in a `TextureRequestData` type with equality semantics based on manually managed `data-version` and `topology-hash`, just like we do for `java.nio.Buffer` with `BufferData`.

In addition, the following improvements were made to the Reveal Defold integration:
* Reveal actions that operate on `Endpoints` or `node-ids` will now try to coerce one into another when possible. For example, you can activate `defold:successors` on a `:label` in the `defold:node-tree` view, or use `defold:node-tree` on an `Endpoint`.
* Added a `defold:predecessors` action to Reveal. It works similarly to `defold:successors`, but shows a reverse tree of inputs affecting a particular `Endpoint`.
* Endpoints in the `defold:successors` or `defold:predecessors` views now show `self` instead of the `node-id` and `NodeType` if they belong to the same node as the queried `Endpoint`.

After this is merged, we have a corresponding code-cleanup PR in defold/extension-texturepacker#22.

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -84,6 +84,7 @@
             [service.log :as log]
             [service.smoke-log :as slog]
             [util.coll :as coll :refer [pair]]
+            [util.defonce :as defonce]
             [util.eduction :as e]
             [util.fn :as fn]
             [util.http-server :as http-server]
@@ -406,15 +407,15 @@
 
 (handler/defhandler :scene.select-move-tool :workbench
   (run [app-view] (g/transact (g/set-property app-view :active-tool :move)))
-  (state [app-view] (= (g/node-value app-view :active-tool) :move)))
+  (state [app-view evaluation-context] (= (g/node-value app-view :active-tool evaluation-context) :move)))
 
 (handler/defhandler :scene.select-scale-tool :workbench
   (run [app-view] (g/transact (g/set-property app-view :active-tool :scale)))
-  (state [app-view] (= (g/node-value app-view :active-tool) :scale)))
+  (state [app-view evaluation-context] (= (g/node-value app-view :active-tool evaluation-context) :scale)))
 
 (handler/defhandler :scene.select-rotate-tool :workbench
   (run [app-view] (g/transact (g/set-property app-view :active-tool :rotate)))
-  (state [app-view] (= (g/node-value app-view :active-tool) :rotate)))
+  (state [app-view evaluation-context] (= (g/node-value app-view :active-tool evaluation-context) :rotate)))
 
 (handler/defhandler :scene.visibility.show-settings :workbench
   (run [app-view scene-visibility]
@@ -424,8 +425,8 @@
                            (.lookup "#visibility-settings-graphic")
                            .getParent)]
       (scene-visibility/show-visibility-settings! app-view btn scene-visibility)))
-  (state [app-view scene-visibility]
-    (when-let [btn (some-> ^TabPane (g/node-value app-view :active-tab-pane)
+  (state [app-view scene-visibility evaluation-context]
+    (when-let [btn (some-> ^TabPane (g/node-value app-view :active-tab-pane evaluation-context)
                            ui/selected-tab
                            .getContent
                            (.lookup "#visibility-settings-graphic")
@@ -433,25 +434,29 @@
       ;; TODO: We have no mechanism for updating the style nor icon on
       ;; on the toolbar button. For now we piggyback on the state
       ;; update polling to set a style when the filters are active.
-      (if (scene-visibility/filters-appear-active? scene-visibility)
+      (if (scene-visibility/filters-appear-active? scene-visibility evaluation-context)
         (ui/add-style! btn "filters-active")
         (ui/remove-style! btn "filters-active"))
       (scene-visibility/settings-visible? btn))))
 
 (defn get-grid-settings-button
-  [app-view]
-  (some-> ^TabPane (g/node-value app-view :active-tab-pane)
+  [^TabPane tab-pane]
+  (some-> tab-pane
           ui/selected-tab
           .getContent
           (.lookup "#show-grid-settings")))
 
 (handler/defhandler :scene.grid.show-settings :workbench
   (run [app-view scene-visibility prefs]
-       (when-let [btn (get-grid-settings-button app-view)]
-         (grid/show-settings! app-view btn prefs)))
-  (state [app-view scene-visibility]
-         (when-let [btn (get-grid-settings-button app-view)]
-           (scene-visibility/settings-visible? btn))))
+       (let [active-tab-pane (g/node-value app-view :active-tab-pane)
+             btn (get-grid-settings-button active-tab-pane)]
+         (when btn
+           (grid/show-settings! app-view btn prefs))))
+  (state [app-view scene-visibility evaluation-context]
+         (let [active-tab-pane (g/node-value app-view :active-tab-pane evaluation-context)]
+           (some-> active-tab-pane
+                   (get-grid-settings-button)
+                   (scene-visibility/settings-visible?)))))
 
 (def ^:private eye-icon-svg-path
   (ui/load-svg-path "scene/images/eye_icon_eye_arrow.svg"))
@@ -1909,7 +1914,7 @@
    menu-items/pull-up-overrides
    menu-items/push-down-overrides])
 
-(defrecord SelectionProvider [app-view]
+(defonce/record SelectionProvider [app-view]
   handler/SelectionProvider
   (selection [_] (g/node-value app-view :selected-node-ids))
   (succeeding-selection [_] [])
@@ -1956,22 +1961,25 @@
         "Defold"))
   ([project-title] (str project-title " - " (make-title))))
 
-(defn- refresh-app-title! [^Stage stage project]
-  (let [settings      (g/node-value project :settings)
-        project-title (settings ["project" "title"])
-        new-title     (make-title project-title)]
+(defn- refresh-app-title! [^Stage stage project evaluation-context]
+  (let [project-title (some-> (g/maybe-node-value project :settings evaluation-context)
+                              (get ["project" "title"]))
+        new-title (if project-title
+                    (make-title project-title)
+                    (make-title))]
     (when (not= (.getTitle stage) new-title)
       (.setTitle stage new-title))))
 
-(defn- refresh-menus-and-toolbars! [app-view ^Scene scene]
-  (ui/user-data! scene :keymap (g/node-value app-view :keymap))
-  (ui/refresh scene))
+(defn- refresh-menus-and-toolbars! [app-view ^Scene scene evaluation-context]
+  (ui/user-data! scene :keymap (g/node-value app-view :keymap evaluation-context))
+  (ui/refresh scene evaluation-context))
 
-(defn- refresh-views! [app-view]
-  (let [auto-pulls (g/node-value app-view :auto-pulls)]
-    (doseq [[node label] auto-pulls]
-      (profiler/profile "view" (:name @(g/node-type* node))
-        (g/node-value node label)))))
+(defn- refresh-views! [app-view evaluation-context]
+  (let [basis (:basis evaluation-context)
+        auto-pulls (g/node-value app-view :auto-pulls evaluation-context)]
+    (doseq [[node-id label] auto-pulls]
+      (profiler/profile "view" (:name @(g/node-type* basis node-id))
+        (g/node-value node-id label evaluation-context)))))
 
 (defn- refresh-scene-view! [scene-view-id dt]
   (try
@@ -2089,22 +2097,24 @@
       (let [prev-localization-bundle (volatile! nil)
             refresh-timer (ui/->timer
                             "refresh-app-view"
-                            (fn [_ elapsed dt]
+                            (fn [_animation-timer _elapsed dt]
                               (when-not (ui/ui-disabled?)
                                 (let [refresh-requested? (ui/user-data app-scene ::ui/refresh-requested?)]
                                   (when refresh-requested?
+                                    (ui/user-data! app-scene ::ui/refresh-requested? false)
                                     (g/with-auto-evaluation-context evaluation-context
                                       (let [localization-bundle (-> project
                                                                     (project/editor-localization-bundle evaluation-context)
                                                                     (editor-localization-bundle/bundle evaluation-context))]
                                         (when-not (identical? @prev-localization-bundle localization-bundle)
                                           (vreset! prev-localization-bundle localization-bundle)
-                                          (localization/set-bundle! localization ::project localization-bundle))))
-                                    (ui/user-data! app-scene ::ui/refresh-requested? false)
-                                    (refresh-menus-and-toolbars! app-view app-scene)
-                                    (refresh-views! app-view))
-                                  (refresh-scene-views! app-view dt)
-                                  (refresh-app-title! stage project)))))]
+                                          (localization/set-bundle! localization ::project localization-bundle)))
+                                      (refresh-menus-and-toolbars! app-view app-scene evaluation-context)
+                                      (refresh-views! app-view evaluation-context)
+                                      (refresh-app-title! stage project evaluation-context)))
+                                  ;; Scene views are always refreshed, since they may play animations.
+                                  ;; This performs graph mutations, so needs to manage its own evaluation-contexts.
+                                  (refresh-scene-views! app-view dt)))))]
         (ui/timer-stop-on-closed! stage refresh-timer)
         [app-view refresh-timer]))))
 
@@ -2886,23 +2896,23 @@
       (search-results-view/show-search-in-files-dialog! search-results-view project prefs localization show-search-results-tab!))))
 
 (handler/defhandler :project.bundle :global
-  (options [user-data _context]
+  (options [user-data _context evaluation-context]
     (when-not user-data
       (let [contexts [_context]]
         (into []
               (keep
                 (fn [{:keys [command]}]
                   (when command
-                    (when-let [handler+context (handler/active command contexts true)]
+                    (when-let [handler+context (handler/active command contexts true evaluation-context)]
                       {:command :project.bundle
-                       :label (handler/label handler+context)
+                       :label (handler/label handler+context evaluation-context)
                        :user-data {:command command
                                    :handler+context handler+context}}))))
               (handler/realize-menu :editor.bundle/menu)))))
   (run [prefs user-data]
     (let [{:keys [command handler+context]} user-data]
       (prefs/set! prefs [:bundle :last-bundle-command] (if (handler/synthetic-command? command) nil command))
-      (when (handler/enabled? handler+context)
+      (when (handler/enabled? handler+context) ; Safe to not supply evaluation-context - we're executing a command.
         (handler/run handler+context)))))
 
 (handler/defhandler :project.rebundle :global
--- a/editor/src/clj/editor/atlas.clj
+++ b/editor/src/clj/editor/atlas.clj
@@ -46,6 +46,7 @@
             [editor.scene-tools :as scene-tools]
             [editor.shaders :as shaders]
             [editor.texture-set :as texture-set]
+            [editor.texture-util :as texture-util]
             [editor.types :as types]
             [editor.validation :as validation]
             [editor.workspace :as workspace]
@@ -61,7 +62,6 @@
            [com.dynamo.gamesys.proto AtlasProto$Atlas AtlasProto$AtlasAnimation AtlasProto$AtlasImage TextureSetProto$TextureSet Tile$Playback]
            [com.jogamp.opengl GL GL2]
            [editor.types Animation Image]
-           [java.awt.image BufferedImage]
            [java.nio ByteBuffer]
            [java.util List]
            [javax.vecmath AxisAngle4d Matrix4d Point3d Vector3d]))
@@ -556,15 +556,12 @@
     (catch Exception error
       (g/->error error-node-id :max-page-size :fatal nil (.getMessage error)))))
 
-(defn- call-generator [generator]
-  ((:f generator) (:args generator)))
-
 (defn- generate-packed-page-images [{:keys [digest-ignored/error-node-id image-resources layout-data-generator]}]
   (let [buffered-images (mapv #(resource-io/with-error-translation % error-node-id nil
                                  (image-util/read-image %))
                               image-resources)]
     (g/precluding-errors buffered-images
-      (let [layout-data (call-generator layout-data-generator)]
+      (let [layout-data (texture-util/call-generator layout-data-generator)]
         (g/precluding-errors layout-data
           (let [id->image (zipmap (map resource/proj-path image-resources) buffered-images)]
             (texture-set-gen/layout-atlas-pages (:layout layout-data) id->image)))))))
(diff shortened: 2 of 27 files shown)


AGulev:
Does it also fix this one https://github.com/defold/defold/issues/9401 ?

editor/src/clj/editor/gl/texture.clj
```diff
@@ -301,67 +432,62 @@
...
+           GL/GL_UNSIGNED_BYTE ; gl type
+           false               ; compressed?
+           false               ; flip vertically?
+           mipmap-buffers
+           nil)
+
+         topology-hash (texture-data-topology-hash texture-data)]
+
+     (TextureRequestData. texture-data (int data-version) topology-hash))))
+
```
matgis, line 471: Indentation change due to added arity with `data-version` argument.

editor/src/clj/editor/gl/texture.clj
```diff
@@ -383,9 +509,18 @@
...
 
-(defn tex-sub-image [^GL2 gl ^TextureLifecycle texture page-index data x y w h data-format]
+(defonce placeholder
+  (delay
+    (image-texture
+      ::placeholder
+      nil
+      (assoc default-image-texture-params
+        :min-filter GL2/GL_NEAREST
+        :mag-filter GL2/GL_NEAREST))))
```
matgis, line 519: We already had `texture/white-pixel` and `texture/black-pixel`. And now we also have `texture/placeholder` - a very pink "missing texture" look.

editor/src/clj/editor/gl/texture.clj
```diff
@@ -419,21 +578,15 @@
 
 (scene-cache/register-object-cache! ::texture make-texture update-texture destroy-textures)
 
-(def ^:private cubemap-targets
-  {:px GL/GL_TEXTURE_CUBE_MAP_POSITIVE_X
-   :nx GL/GL_TEXTURE_CUBE_MAP_NEGATIVE_X
-   :py GL/GL_TEXTURE_CUBE_MAP_POSITIVE_Y
-   :ny GL/GL_TEXTURE_CUBE_MAP_NEGATIVE_Y
-   :pz GL/GL_TEXTURE_CUBE_MAP_POSITIVE_Z
-   :nz GL/GL_TEXTURE_CUBE_MAP_NEGATIVE_Z})
```
matgis, line 428: Moved up in the file so we can use it to validate the arguments to `make-gpu-cubemap-texture`.

editor/src/clj/editor/gl/types.clj
```diff
@@ -28,8 +28,14 @@
...
-(defn gl-binding? [value]
-  (satisfies? GLBinding value))
+(def
+  ^{:doc "Returns true if the value satisfies the GLBinding protocol."
+    :arglists '([value])}
+  gl-binding?
+  ;; This is quite a bit faster than the satisfies? function.
+  (let [gl-binding-class (:on-interface GLBinding)]
+    (fn gl-binding? [value]
+      (instance? gl-binding-class value))))
```
matgis, line 38: Unrelated: `satisfies?` was actually showing up fairly prominently in the profiler when viewing a complex scene in the Scene View. 😬

editor/src/clj/editor/pipeline/tex_gen.clj
```diff
@@ -77,17 +78,18 @@
    (make-texture-image image texture-profile compress? true))
   (^TextureGenerator$GenerateResult [^BufferedImage image texture-profile compress? flip-y?]
    (let [^Graphics$TextureProfile texture-profile-data (some->> texture-profile (protobuf/map->pb Graphics$TextureProfile))
-         texture-generate-result (TextureGenerator/generate image texture-profile-data ^boolean compress? (if ^boolean flip-y? (EnumSet/of Texc$FlipAxis/FLIP_AXIS_Y) (EnumSet/noneOf Texc$FlipAxis)))]
-     texture-generate-result)))
+         texture-generator-result (TextureGenerator/generate image texture-profile-data ^boolean compress? (if ^boolean flip-y? (EnumSet/of Texc$FlipAxis/FLIP_AXIS_Y) (EnumSet/noneOf Texc$FlipAxis)))]
+     texture-generator-result)))
```
matgis, line 82: Renamed for consistency with everywhere else.

editor/src/clj/editor/app_view.clj
```diff
@@ -424,34 +425,38 @@
...
+      (if (scene-visibility/filters-appear-active? scene-visibility evaluation-context)
         (ui/add-style! btn "filters-active")
         (ui/remove-style! btn "filters-active"))
       (scene-visibility/settings-visible? btn))))
 
 (defn get-grid-settings-button
-  [app-view]
-  (some-> ^TabPane (g/node-value app-view :active-tab-pane)
+  [^TabPane tab-pane]
+  (some-> tab-pane
```
matgis, line 444: Argument changed from `app-view` `node-id` to `TabPane` to avoid creating an `evaluation-context` arity.

editor/src/clj/editor/app_view.clj
```diff
@@ -1956,22 +1961,25 @@
...
-(defn- refresh-app-title! [^Stage stage project]
-  (let [settings      (g/node-value project :settings)
-        project-title (settings ["project" "title"])
-        new-title     (make-title project-title)]
+(defn- refresh-app-title! [^Stage stage project evaluation-context]
+  (let [project-title (some-> (g/maybe-node-value project :settings evaluation-context)
+                              (get ["project" "title"]))
+        new-title (if project-title
+                    (make-title project-title)
+                    (make-title))]
```
matgis, line 1969: Note: We now use `maybe-node-value` to evaluate the project title. This avoids a `NullPointerException` if `game.project` is corrupted.

editor/src/clj/editor/atlas.clj
```diff
@@ -556,15 +556,12 @@
     (catch Exception error
       (g/->error error-node-id :max-page-size :fatal nil (.getMessage error)))))
 
-(defn- call-generator [generator]
-  ((:f generator) (:args generator)))
-
```
matgis, line 561: We had these in a couple of places. Now in `texture-util/call-generator`.

editor/src/clj/editor/atlas.clj
```diff
@@ -768,18 +763,13 @@
...
-                                                           (texture/texture-images->gpu-texture
-                                                             _node-id
-                                                             page-texture-images+texture-bytes
-                                                             {:min-filter gl/nearest
-                                                              :mag-filter gl/nearest}))))
+  (output gpu-texture g/Any :cached
+          (g/fnk [_node-id packed-page-images-generator texture-profile]
+            (-> (texture-util/construct-gpu-texture _node-id packed-page-images-generator texture-profile)
+                (texture/set-params {:min-filter gl/nearest
+                                     :mag-filter gl/nearest}))))
```
matgis, line 772: All `gpu-texture` outputs will now use either `texture-util/construct-gpu-texture` or `texture-util/generate-gpu-texture`. The difference being the latter takes a `gpu-texture-generator` instead of a `content-generator`. In addition, all `gpu-texture-generator` outputs will now use `texture-util/make-gpu-texture-generator`.

These will early-out and return any `ErrorValues` returned by called generators. `texture/set-params` will return `ErrorValues` unaltered.

We are quite liberal about what we accept as a `content-generator`. It can be either a single generator that may return a single `BufferedImage` or a sequence of `BufferedImages`, or a sequence of generators return one `BufferedImage` each.

editor/src/clj/editor/buffers.clj
```diff
@@ -317,9 +317,10 @@
 
 (defn topology-hash
   ^long [^Buffer buffer]
-  (-> (hash (class buffer))
-      (Util/hashCombine (.hashCode (byte-order buffer)))
-      (Util/hashCombine (Murmur3/hashLong (total-byte-size buffer)))))
+  (java/combine-hashes
+    (hash (class buffer))
+    (.hashCode (byte-order buffer))
+    (Murmur3/hashLong (total-byte-size buffer))))
```
matgis, line 323: Unrelated: Use `java/combine-hashes`.

editor/src/clj/editor/cubemap.clj
```diff
@@ -86,14 +85,10 @@
...
-    (gl/gl-draw-arrays gl GL/GL_TRIANGLES 0 (* 6 (* sphere-lats sphere-longs)))
-    (gl/gl-disable gl GL/GL_CULL_FACE)))
-
-(g/defnk produce-gpu-texture
-  [_node-id gpu-texture-generator]
-  ((:f gpu-texture-generator) (:args gpu-texture-generator) _node-id texture/default-cubemap-texture-params 0))
+    (gl/gl-enable gl GL2/GL_CULL_FACE)
+    (gl/gl-cull-face gl GL2/GL_BACK)
+    (gl/gl-draw-arrays gl GL2/GL_TRIANGLES 0 (* 6 (* sphere-lats sphere-longs)))
+    (gl/gl-disable gl GL2/GL_CULL_FACE)))
```
matgis, line 91: `GL` -> `GL2` to avoid importing `GL`.

editor/src/clj/editor/font.clj
```diff
@@ -837,7 +837,7 @@
                                                    h (:cache-height font-map)
                                                    channels (:glyph-channels font-map)
                                                    data-format (glyph-channels->data-format channels)]
-                                               (texture/empty-texture _node-id w h data-format
+                                               (texture/empty-texture _node-id data-format w h
```
matgis, line 840: Changed argument order for consistency.

editor/src/clj/editor/mesh.clj
```diff
@@ -148,11 +150,14 @@
...
-                        texture    (tex-fn tex-args request-id params unit-index)]
-                    [(:name sampler) texture]))
+  (into {} (map (fn [unit-index sampler gpu-texture-generator]
+                  (let [gpu-texture (texture-util/generate-gpu-texture gpu-texture-generator)]
+                    (pair (:name sampler)
+                          (-> (if (g/error-value? gpu-texture)
+                                @texture/placeholder
+                                gpu-texture)
+                              (texture/set-params (material/sampler->tex-params sampler))
+                              (texture/set-base-unit unit-index)))))
```
matgis, line 160: Previously we weren't handling `ErrorValues` returned from the generator here.

editor/src/clj/editor/model.clj
```diff
@@ -146,16 +147,19 @@
...
-                                            texture (tex-fn tex-args request-id params unit-index)]
-                                        [name texture]))))
+                                    (when-let [gpu-texture-generator (sampler-name->gpu-texture-generator name)]
+                                      (let [gpu-texture (texture-util/generate-gpu-texture gpu-texture-generator)]
+                                        (pair name
+                                              (-> (if (g/error-value? gpu-texture)
+                                                    @texture/placeholder
+                                                    gpu-texture)
+                                                  (texture/set-params (material/sampler->tex-params sampler))
+                                                  (texture/set-base-unit unit-index)))))))
```
matgis, line 162: Previously we weren't handling `ErrorValues` returned from the generator here.

editor/src/clj/editor/scene.clj
```diff
@@ -959,26 +959,28 @@
...
+          (advance-user-data-component!
+            node-id :overlay-anchor-pane
+            {:fx/type fxui/ext-with-anchor-pane-props
+             :props overlay-anchor-pane-props
+             :desc {:fx/type fxui/ext-value
+                    :value overlay-anchor-pane}}))))))
 
 (defn dispose-scene-view! [node-id]
-  (when-let [scene (g/node-by-id node-id)]
+  (when (g/node-by-id node-id)
```
matgis, line 983: `scene` was unused.

editor/src/clj/editor/scene.clj
```diff
@@ -1146,25 +1163,24 @@
       (= (:type camera-a) :perspective)
       (c/camera-orthographic->perspective c/fov-y-35mm-full-frame))))
 
-(defn- get-3d-camera
-  [camera]
-  (or (g/node-value camera :cached-3d-camera)
-      (c/tumble (g/node-value camera :local-camera) 200.0 -100.0)))
-
```
matgis, line 1153: Inlined at call site.

editor/src/clj/editor/scene.clj
```diff
@@ -1357,19 +1377,22 @@
...
+             (g/node-value view-id :active-updatables evaluation-context)
+             (g/node-value view-id :updatable-states evaluation-context)])
+
           has-active-updatables (not (coll/empty? active-updatables))
-          updatable-states (g/node-value view-id :updatable-states evaluation-context)
           new-updatable-states (if has-active-updatables
                                  (profiler/profile "updatables" -1 (update-updatables updatable-states play-mode active-updatables dt))
                                  updatable-states)
-          renderables (g/node-value view-id :all-renderables evaluation-context)
+          renderables (g/node-value view-id :all-renderables)
```
matgis, line 1395: Don't use previous `evaluation-context` here in case the update calls above does something graph-related.

editor/src/clj/editor/texture_util.clj
```diff
@@ -0,0 +1,286 @@
...
+;;
+;; You may obtain a copy of the License, together with FAQs at
+;; https://www.defold.com/license
+;;
+;; Unless required by applicable law or agreed to in writing, software distributed
+;; under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+;; CONDITIONS OF ANY KIND, either express or implied. See the License for the
+;; specific language governing permissions and limitations under the License.
+
+(ns editor.texture-util
```
matgis, line 15: I introduced a new module for this to avoid bringing in graph-related stuff into `editor.gl.texture`.

editor/src/clj/editor/tile_source.clj
```diff
@@ -724,28 +721,30 @@
...
+                      (digestable/sha1-hash
+                        {:image-sha1 image-sha1
+                         :tile-source-attributes tile-source-attributes
+                         :type :packed-tile-source-image})]
+                  {:f generate-packed-image
+                   :sha1 packed-image-sha1
+                   :args {:digest-ignored/error-node-id _node-id
+                          :layout-result layout-result
+                          :image-resource image-resource
+                          :tile-source-attributes tile-source-attributes}})))))
```
matgis, line 747: Indentation change from handling IO errors from sha1 hash calculation.

editor/src/clj/editor/ui.clj
```diff
@@ -1606,7 +1607,7 @@
         command-contexts (node-contexts node true)
         handler-ctx (handler/active command command-contexts user-data)
         enabled (and handler-ctx
-                     (handler/enabled? handler-ctx))]
+                     (handler/enabled? handler-ctx))] ; TODO: This really should supply the same evaluation-context we use to refresh the menus and so on.
```
matgis, line 1610: This is in `refresh-bound-action-enabled!`. It doesn't matter for how we currently use it.

editor/src/clj/editor/ui.clj
```diff
@@ -2059,14 +2060,16 @@
...
+  ([^Scene scene]
+   (g/with-auto-or-fake-evaluation-context evaluation-context
+     (refresh scene evaluation-context)))
+  ([^Scene scene evaluation-context]
+   (let [keymap (or (user-data scene :keymap) keymap/empty)
+         localization (user-data scene :localization)]
+     (assert (some? localization))
+     (refresh-accelerators! scene keymap)
+     (refresh-menus! scene keymap localization evaluation-context)
+     (refresh-toolbars! scene localization evaluation-context))))
```
matgis, line 2072: Indentation change from added arity taking an `evaluation-context`.

editor/src/clj/internal/system.clj
```diff
@@ -41,7 +42,7 @@
 (defn- new-history []
   {:tape (conj (h/paper-tape history-size-max) [])})
 
-(defrecord HistoryState [label graph sequence-label cache-keys])
+(defonce/record HistoryState [label graph sequence-label cache-keys])
```
matgis, line 45: Unrelated, but I was on a roll.

editor/src/reveal/editor/reveal.clj
```diff
@@ -50,6 +52,36 @@
...
+
+(defn- as-node-id+label [value annotation]
+  (if (gt/endpoint? value)
+    [(g/endpoint-node-id value)
+     (g/endpoint-label value)]
+    (let [[node-id label :as node-id+label] (::node-id+label annotation)]
+      (when (and (g/node-id? node-id)
+                 (keyword? label))
+        node-id+label))))
+
```
matgis, line 84: Reveal actions that operate on `Endpoints` or `node-ids` will now try to coerce one into another when possible. For example, you can activate `defold:successors` on a `:label` in the `defold:node-tree` view, or use `defold:node-tree` on an `Endpoint`.

editor/test/editor/pipeline/tex_gen_test.clj
```diff
@@ -53,18 +53,21 @@
...
+                            :platforms [{:os :os-id-generic
+                                         :formats [{:format :texture-format-rgb
+                                                    :compression-level :best}
+                                                   {:format :texture-format-rgba
+                                                    :compression-level :best}
+                                                   {:format :texture-format-luminance
+                                                    :compression-level :best}]
+                                         :mipmaps false
+                                         :premultiply-alpha true}]}
+                           true)
```
matgis, line 70: Formatting. Only change is adding `true` for the `flip-y` argument at the end.

editor/test/integration/resource_watch_test.clj
```diff
@@ -165,10 +165,10 @@
       (workspace/set-project-dependencies! workspace [{:uri imagelib1-uri}])
       (workspace/resource-sync! workspace)
       (let [lib1-paddle (project/get-resource-node project "/images/paddle.png")]
-        (is (not (g/error? (g/node-value lib1-paddle :content))))
+        (is (not (g/error? (g/node-value lib1-paddle :size))))
         (workspace/set-project-dependencies! workspace [{:uri imagelib2-uri}])
         (workspace/resource-sync! workspace)
-        (is (g/error? (g/node-value lib1-paddle :content))))))) ; removed, should emit errors
+        (is (g/error? (g/node-value lib1-paddle :size))))))) ; removed, should emit errors
```
matgis, line 171: The `content` output was removed from `ImageNode`, but `size` works fine for this test.

editor/src/clj/editor/gl/types.clj
vlaaad: Yeah, it's slow, that's why there is, e.g., a special cache for satisfies checks for resources (see `resource/resource?`). Maybe soon we will want to make a generic performant `satisfies?` wrapper...

editor/src/clj/editor/image.clj
```diff
@@ -128,17 +120,6 @@
                                    :uv-transforms [(TextureSetGenerator$UVTransform.)])}))
 
   (output texture-page-count g/Int (g/constantly texture/non-paged-page-count))
-
-  (output gpu-texture g/Any :cached (g/fnk [_node-id texture-image]
-                                      (texture/texture-image->gpu-texture _node-id
-                                                                          texture-image
-                                                                          {:min-filter gl/nearest
-                                                                           :mag-filter gl/nearest})))
```
vlaaad, line 136: We used to use nearest filtering, but now we are using `texture/default-image-texture-params`, which is linear; is this intentional?

editor/src/clj/editor/texture_util.clj
```diff
@@ -0,0 +1,286 @@
...
+                   {:pre [(content-generator? content-generator)]}
+                   (let [content (call-generator content-generator)]
+                     (cond
+                       (g/error-value? content)
+                       content
+
+                       (instance? BufferedImage content)
+                       (let [texture-image (tex-gen/make-preview-texture-image content texture-profile flip-y)
+                             data-version (hash (:sha1 content-generator))]
+                         (texture/texture-image->texture-request-data texture-image data-version))
```
vlaaad, line 118: Does `sha1` include information about, e.g., `flip-y`? Or is it unnecessary? Would (and should) a change to flip-y result in a different `data-version`?

editor/src/clj/editor/mesh.clj
```diff
@@ -148,11 +150,14 @@
             :deps dep-build-targets})])))
 
 (g/defnk produce-gpu-textures [_node-id samplers gpu-texture-generators]
-  (into {} (map (fn [unit-index sampler {tex-fn :f tex-args :args}]
-                  (let [request-id [_node-id unit-index]
-                        params     (material/sampler->tex-params sampler)
-                        texture    (tex-fn tex-args request-id params unit-index)]
-                    [(:name sampler) texture]))
+  (into {} (map (fn [unit-index sampler gpu-texture-generator]
+                  (let [gpu-texture (texture-util/generate-gpu-texture gpu-texture-generator)]
```
vlaaad, line 154: I couldn't figure out how to repro it, but `texture-util/generate-gpu-texture` asserts that its argument is `gpu-texture-generator?`. Is it possible that a cubemap will end up there? If so, cubemap produces a value that satisfies `cubemap-gpu-texture-generator?`, and this is NOT the same as `gpu-texture-generator?`.

vlaaad (dismissed):
LGTM! I left a couple of questions.

editor/src/clj/editor/image.clj
matgis: Oops. No, it wasn't intentional. I'll fix!

editor/src/clj/editor/texture_util.clj
matgis: It should, yes. I added the `flip-y` argument quite late, so I missed including it in the hash. Nice catch!

editor/src/clj/editor/mesh.clj
matgis: I think you're right - we shouldn't assume the type of texture the generator will return here. I'll split out the `texture-lifecycle-generator?` part of `gpu-texture-generator?` predicate and use that to validate the `gpu-texture-generator` here.

Issue #10485, opened 2025-04-13
Gamepads - use a centralised mapping database


8bitskull:
### The problem

Supporting a gamepad requires a mapping to be generated for that specific model, for each operating system. New models are not automatically supported. The Defold community is relatively small, so it is not guaranteed that we have coverage of all gamepads. It is hard to convince customers to download and run mapping tools - they might not want to, or lack the technical know-how. If the developer wants to do it, they need to buy the gamepad in question **and** run multiple operating systems (Windows, Mac, Linux at a minimum).

![Image](https://github.com/user-attachments/assets/539b744e-f015-454b-beb2-a893b51691b0)
_This user does not want to download download gdc to provide a gamepad mapping. Do I have to buy the two controllers they are using and do it myself?_

### What does a solution look like?

Defold's gamepad mapping project is community driven. The issue is that the community is small, so there are gaps in support. Perhaps the solution is still relying on community, just expanding the size of that community?

There must be community gamepad mapping projects (here's one [example](https://github.com/mdqinc/SDL_GameControllerDB) - though I don't know if it's the one to go for) that could be accessed and parsed by Defold. As a second line of defense, if the mapping database we access does not have a controller, we could fall back to the manual mappings we already have. If that fails too then we're back to raw.

A quick search seems to imply that [Godot uses SDL](https://docs.godotengine.org/en/stable/tutorials/inputs/controllers_gamepads_joysticks.html#supporting-universal-input) (the example linked above) and a [Godot specific database](https://github.com/godotengine/godot/blob/master/core/input/godotcontrollerdb.txt) (which is reminiscent of what we have right now).

### Conclusion

Gamepad support is a big pain point that can generate poor user experience, which results in things like this:

![Image](https://github.com/user-attachments/assets/73936935-b46c-40b0-a219-bc68e91b13fc)

I don't know if my proposed solution is the way to go, so perhaps this issue can serve as a starting point for discussion on how/if to proceed.

aglitchman:
Related:
- https://github.com/defold/defold/issues/9877

aglitchman:
Btw, GLFW 3 has gamepad mappings db too and it's included in every build, but I don't know if it's used or not by Defold - https://github.com/glfw/glfw/blob/master/src/mappings.h

```bash
$ strings ./build/x86_64-win32/dmengine.exe | grep 3DRUDDER
[...]

PR #11565, opened 2025-12-03
Improved GamePad mapping for Defold supporting SDL controller db

deurell:
Supporting gamepads in Defold requires a mapping to be generated for that specific gamepad. New gamepads are not automatically supported. The Defold community is relatively small, so it is not guaranteed that we have coverage of all gamepads. It is hard to convince developers to download and run mapping tools or manually add mappings. How can we improve this? We've looked at two ways of doing this.

- Add support for the SDL controller database by converting and integrating it to Defold-
- Delegate controller implementation to GLFW

After analysing the two options, delegating controller logic to GLFW by hooking up GetControllerState is not ideal due to consoles not using GLFW so this document is based on the SDL controller database mapping option.

We suggest rewriting Defolds controller logic so every gamepad is resolved by hardware GUID first, using a single SDL-style source database that we convert into the existing protobuf format (glfw.gamepadsc). An optional GUID property is added to the input_ddf and we add a way for the engine to query gamepad hardware GUIDs via GLFW while still providing a fallback to use hashed controller names like before. This improves the name-only matches and also adds additional controller support by pulling the complete list of gamepads from the SDL community controller database. 
Reference: Issue #10485

Design document available here: 
https://docs.google.com/document/d/1iF4IRrBZSr0P5eQ1fi9_-f-ZP24UaBUQ2XLFEc6hIBM/edit?usp=sharing

**This is a draft PR with code backing the design document used to develop the feature. It is not a PR that is supposed to be merged with dev.**


AGulev:
it's already implemented

PR #11568, opened 2025-12-04
Revamp the Properties view

vlaaad:
We have now refreshed the Properties view.

https://github.com/user-attachments/assets/4c447b4a-ae3f-4de8-a64d-108a67e3ff02

It looks better than it was before; it shows the updates immediately, and its styling is now more consistent with the rest of the editor, particularly with the editor scripts UI.

### Technical changes
Related to #10741 — we now update the values in the Properties view immediately, though we still want to do it a bit differently, so we are not closing the issue.


editor/resources/messages.properties
```diff
@@ -1,133 +0,0 @@
-BuilderUtil_EMPTY_RESOURCE=you must specify a {0}
```
vlaaad, line 1: Unused file

editor/src/clj/editor/atlas.clj
```diff
@@ -709,14 +709,17 @@
...
             (dynamic label (properties/label-dynamic :atlas :margin))
             (dynamic tooltip (properties/tooltip-dynamic :atlas :margin))
             (dynamic error (g/fnk [_node-id margin] (validate-margin _node-id margin))))
   (property inner-padding g/Int (default (protobuf/default AtlasProto$Atlas :inner-padding))
+            (dynamic edit-type (g/constantly {:type g/Int :min 0}))
             (dynamic label (properties/label-dynamic :atlas :inner-padding))
             (dynamic tooltip (properties/tooltip-dynamic :atlas :inner-padding))
             (dynamic error (g/fnk [_node-id inner-padding] (validate-inner-padding _node-id inner-padding))))
   (property extrude-borders g/Int (default (protobuf/default AtlasProto$Atlas :extrude-borders))
+            (dynamic edit-type (g/constantly {:type g/Int :min 0}))
```
vlaaad, line 722: While experimenting with the scrubber, I got annoyed by how easy it was to make it throw errors when setting invalid values.

editor/src/clj/editor/editor_extensions/ui_components.clj
```diff
@@ -452,44 +434,48 @@
...
-                    {:fx/type fx.combo-box/lifecycle
-                     :value value
-                     :pseudo-classes (or (some-> issue :severity hash-set) #{})
-                     :on-key-pressed on-select-box-key-pressed
-                     :items options
-                     :disable (not enabled)}
-                    tooltip issue localization-state)}
-     :props (cond-> {} on_value_changed (assoc :on-value-changed (make-event-handler-1 :window :value rt "on_value_changed" on_value_changed)))}
+            :key :to-string
+            :desc (-> {:fx/type fxui.combo-box/view
```
vlaaad, line 470: Now, both editor scripts and properties view use the same combo-box component.

editor/src/clj/editor/field_expression.clj
```diff
@@ -14,10 +14,7 @@
 
 (ns editor.field-expression
   (:require [editor.math :as math])
-  (:import [java.math MathContext]
-           [java.text DecimalFormat DecimalFormatSymbols]
-           [java.util Locale]
-           [net.objecthunter.exp4j ExpressionBuilder]))
+  (:import [net.objecthunter.exp4j ExpressionBuilder]))
```
vlaaad, line 17: Unrelated import cleanup

editor/src/clj/editor/fxui.clj
```diff
@@ -346,6 +324,27 @@
     (delete [_ component opts]
       (fx.lifecycle/delete lifecycle (:child component) opts))))
 
+(def ext-dedupe-identical-desc
+  "Extension lifecycle that skips advancing if new desc is identical to old one
```
vlaaad, line 327: This one makes the Properties view noop refresh really fast.

editor/src/clj/editor/git_credentials.clj
```diff
@@ -1,100 +0,0 @@
-;; Copyright 2020-2025 The Defold Foundation
```
vlaaad, line 1: Unused file

editor/src/clj/editor/graphics.clj
```diff
@@ -644,7 +644,7 @@
              :clear-fn attribute-clear-fn}
 
             (= semantic-type :semantic-type-color)
-            (assoc :ignore-alpha? (not= :vector-type-vec4 vector-type)))))
+            (assoc :ignore-alpha (not= :vector-type-vec4 vector-type)))))
```
vlaaad, line 647: I renamed a keyword.

editor/src/clj/editor/jfx.clj
```diff
@@ -1,46 +0,0 @@
...
-         cached-image (get @cached-images image-key)]
-     (cond
-       cached-image cached-image
-       url          (let [image (if size
-                                  (Image. (str url) size size true true)
-                                  (Image. (str url)))]
-                        (swap! cached-images assoc image-key image)
-                        image)))))
-
-(defn get-image-view
```
vlaaad, line 39: There was a nearly identical ns `editor.icons` that was much more widely used, and had more features.

editor/src/clj/editor/messages.clj
```diff
@@ -1,36 +0,0 @@
-;; Copyright 2020-2025 The Defold Foundation
```
vlaaad, line 1: Unused.

editor/src/clj/editor/web_profiler.clj
```diff
@@ -1,36 +0,0 @@
-;; Copyright 2020-2025 The Defold Foundation
```
vlaaad, line 1: Unused file.

editor/src/java/com/defold/control/ButtonWithCappedMinWidth.java
```diff
@@ -0,0 +1,30 @@
...
+public class ButtonWithCappedMinWidth extends Button {
+    private final double minWidthCap;
+
+    public ButtonWithCappedMinWidth(double minWidthCap) {
+        this.minWidthCap = minWidthCap;
+    }
+
+    @Override
+    protected double computeMinWidth(double v) {
+        return Math.max(super.computePrefWidth(v), minWidthCap);
```
vlaaad, line 28: Somehow, this was the easiest way to make a JavaFX button whose min width is a maximum of its pref width and some constant. It's not the same thing as setting min-width to a constant value, because if pref-width is bigger than the constant, then the button width may be reduced to the constant value, which will make, e.g., the ellipses appear on a button label.

editor/styling/stylesheets/modules/_properties.scss
```diff
@@ -1,7 +1,4 @@
 #properties {
-    HBox {
-        -fx-spacing: 0;
-    }
```
vlaaad, line 4: I was going crazy because of this at some point. I wanted spacing, but there was none, even though I had CSS style classes that only added spacing. But yeah, you can target Java classes with CSS...

editor/src/clj/editor/fxui.clj
```diff
@@ -945,37 +1119,187 @@
...
+      (do (.stop animation)
+          (.setTranslateX node 0.0)
+          (.playFromStart animation))
+      (let [animation (SequentialTransition.
+                        node
+                        (into-array Animation [(doto (TranslateTransition. (Duration. 30.0)) (.setByX 4.0))
+                                               (doto (TranslateTransition. (Duration. 30.0)) (.setByX -8.0))
+                                               (doto (TranslateTransition. (Duration. 30.0)) (.setByX 7.0))
+                                               (doto (TranslateTransition. (Duration. 30.0)) (.setByX -4.0))
+                                               (doto (TranslateTransition. (Duration. 30.0)) (.setByX 1.0))]))]
```
matgis, line 1261: Move out `Animation` array to global scope for sharing?

editor/src/clj/editor/fxui.clj
```diff
@@ -1034,13 +1393,15 @@
   Supports all :text-field props, plus:
     :value               the edited value
     :on-value-changed    value change callback
+    :on-invalid-value    called when invalid value is submitted, wiggles the
+                         view by default
```
matgis: Document the args that will be supplied to these callbacks?

editor/src/clj/editor/editor_extensions/ui_components.clj
```diff
@@ -403,19 +391,13 @@
...
   (let [message (:message issue)]
-    (cond-> props (or message tooltip) (assoc :tooltip (localization-state
-                                                         (if (and message tooltip)
-                                                           (localization/join "\n\n" [message tooltip])
-                                                           (or message tooltip)))))))
+    (cond-> props (or message tooltip) (assoc :tooltip {:severity (:severity issue :info)
+                                                        :message (localization-state
+                                                                   (if (and message tooltip)
+                                                                     (localization/join "\n\n" [message tooltip])
+                                                                     (or message tooltip)))}))))
```
matgis: We could take this opportunity to split this expression into multiple lines to fight the rather severe indentation here.

editor/src/clj/editor/fxui/combo_box.clj
```diff
@@ -0,0 +1,272 @@
...
+(defn- set-filter-text [state text]
+  (assoc state :filter-text text))
+
+(defn- set-selected-item [state item]
+  (assoc state :selected-item item))
+
+(defn- handle-mouse-pressed [swap-state ^MouseEvent e]
+  (.requestFocus ^Node (.getSource e))
+  (.consume e)
+  (swap-state toggle))
```
matgis, line 79: There are a couple of places in this file where we return the value returned by a user-provided callback. Is that OK, or should we make sure to return `nil` instead?

editor/src/clj/editor/fxui/combo_box.clj
```diff
@@ -0,0 +1,272 @@
...
+      (update :pseudo-classes conj color)
+
+      showing
+      (assoc
+        prop-popup
+        (let [option->text #(to-string (% 0))
+              filtered-item+matching-indices (->> items
+                                                  (mapv vector)
+                                                  (fuzzy-choices/filter-options option->text option->text filter-text)
+                                                  (e/map #(coll/pair (% 0) (:matching-indices (meta %))))
```
matgis: Should we `:refer` `coll/pair`?

editor/src/clj/editor/fxui.clj
```diff
@@ -498,29 +497,157 @@
...
+        (when showing-property
+          (.addListener showing-property showing-property-listener))
+        #(do
+           (.removeEventHandler node MouseEvent/MOUSE_ENTERED on-enter)
+           (.removeEventHandler node MouseEvent/MOUSE_EXITED on-exit)
+           (when showing-property
+             (.removeListener showing-property showing-property-listener)))))
+    fx.lifecycle/dynamic))
+
+
```
matgis: Style: Extra newline.

editor/src/clj/editor/fxui.clj
```diff
@@ -1065,6 +1426,87 @@
...
+   (let [nr (Math/round (* (float 255.0) (float (.getRed color))))
+         ng (Math/round (* (float 255.0) (float (.getGreen color))))
+         nb (Math/round (* (float 255.0) (float (.getBlue color))))
+         na (Math/round (* (float 255.0) (float (.getOpacity color))))]
+     (if (or ignore-alpha (= 255 na))
+       (format "#%02x%02x%02x" nr ng nb)
+       (format "#%02x%02x%02x%02x" nr ng nb na)))))
+
+(defn- web-string->color [^String s]
+  (try (Color/valueOf s) (catch IllegalArgumentException _)))
```
matgis: Do we need to turn semi-transparent colors opaque here if `ignore-alpha` is true for the property?

editor/src/clj/editor/fxui.clj
```diff
@@ -1065,6 +1426,87 @@
...
+  ([color]
+   (color->web-string false color))
+  ([ignore-alpha ^Color color]
+   (let [nr (Math/round (* (float 255.0) (float (.getRed color))))
+         ng (Math/round (* (float 255.0) (float (.getGreen color))))
+         nb (Math/round (* (float 255.0) (float (.getBlue color))))
+         na (Math/round (* (float 255.0) (float (.getOpacity color))))]
+     (if (or ignore-alpha (= 255 na))
+       (format "#%02x%02x%02x" nr ng nb)
+       (format "#%02x%02x%02x%02x" nr ng nb na)))))
```
matgis, line 1420: I don't think we want to strip out the alpha part, even if the color is opaque. To me, it signals that a property supports semi-transparent colors. Not all of them do.

editor/src/clj/editor/properties_view.clj
```diff
@@ -1019,89 +210,533 @@
...
+         :style-class "ext-resource-picker-field"}
+        {:fx/type fxui/button
+         :graphic {:fx/type ui/image-icon
+                   :path "icons/32/Icons_S_14_linkarrow.png"
+                   :size 16.0}
+         :style-class "ext-resource-picker-open-button"
+         :disable (not openable)
+         :on-action #(when openable (ui/run-command (.getSource ^Event %) :file.open value))}
+        {:fx/type fxui/button
+         :text "..."
```
matgis, line 548: Should it use an ellipsis character?

editor/src/clj/editor/properties_view.clj
```diff
@@ -1019,89 +210,533 @@
...
+      (resolve-validation property localization-state)))
+
+(defmethod make-control-view :default [property _context localization-state]
+  (-> {:fx/type fxui/value-field
+       :to-string str
+       :editable false}
+      (resolve-value property)
+      (resolve-validation property localization-state)))
+
+;; TODO: profile
```
matgis: I did some testing and used the scale tool with ~1000 objects selected. It was slightly slower in this case, but the old implementation would not refresh the properties view at all during drag. I think it's fine. It would be nice to not have to evaluate the `:_properties` outputs at all during the drag operation, though. A later improvement for sure!

matgis (changes requested):
Incredible work! And it fixes a lot of small annoyances in the Properties View! ❤️

I did, however, notice a few issues:
* Previously, if I pressed Escape from a field I was editing, the edit was canceled, and the field lost focus. Now the edit is canceled, but the field retains focus and its contents are selected. Other applications appear to do the former, whereas I've not seen the latter behavior elsewhere. It can be a bit annoying for numeric fields because the adjustment drag handle only appears when the focus leaves the field. Similarly, color fields only show the eyedropper when they aren't focused. I think we should change the Escape key behavior back to how it used to work.
* I'm not seeing the error tooltip for the Gui Box Node Material Property if I try to use a paged Atlas with a non-paged Gui material. Other errors do show up. Could it be specific to dropdowns?
* We now strip the alpha value from the hex string shown for colors when the color is opaque. I don't think we want to. To me, it signals that a property supports semi-transparent colors. Not all of them do.

editor/src/clj/editor/fxui.clj
vlaaad: No, `TranslateTransition` is a mutable object that refers to a target `Node` when played; we shouldn't share it.

editor/src/clj/editor/fxui/combo_box.clj
vlaaad: Values returned by event handlers are silently dropped since the `EventHandler` interface returns `void` from its `handle` method.

editor/src/clj/editor/fxui/combo_box.clj
vlaaad: I don't see a need to make an exception for this particular function among all the other frequently used functions.

editor/src/clj/editor/properties_view.clj
vlaaad: I started with an ellipsis character, but didn't like how it looked.
Ellipsis:
Screenshot 2025-12-08 at 10 17 46

Triple dots:
Screenshot 2025-12-08 at 10 16 27

editor/src/clj/editor/fxui.clj
vlaaad: It won't make a difference in our current use, because the value change listener in the properties view goes through all selected Colors and keeps alpha as is when it's ignored (instead of making it opaque)... but from the semantics standpoint, I think it makes sense to make the reported changed color opaque if alpha is ignored. I'll change it.

vlaaad:
I have now applied the requested changes. One note about losing focus on <kbd>Escape</kbd>. I think it's useful not to lose focus when you want to cancel an edit if you realize you are editing the wrong field (e.g., `Y` position instead of `Z`), and want to use <kbd>Tab</kbd> navigation to move to the next field after cancel. As a compromise, I made first <kbd>Escape</kbd> clear the change, and the second lose the focus. Also, I checked vscode, Godot, and Unity, and neither of them gives up focus completely on <kbd>Escape</kbd>...

editor/src/clj/editor/fxui.clj
```diff
@@ -502,32 +501,51 @@
...
-        (.put (.getProperties node) custom-tooltip-node-showing-key value))
-      (fn [^Node node _]
-        (.remove (.getProperties node) custom-tooltip-node-showing-key)))
-    javafx-property-lifecycle))
+    (fx.mutator/setter
+      (fn set-custom-tooltip-node-showing [^Node node value]
+        (.put (.getProperties node) custom-tooltip-node-showing-key value)))
+    fx.lifecycle/scalar))
+
+(defn- observable-map+key->observable-val
```
matgis: Naming: We typically use `something+other` to refer to tuples. Maybe it would be clearer as something like `make-observable-value-for-key`?

However, I think `(Bindings/valueAt m k)` might be doing exactly this, so perhaps we don't need it? 🤔

editor/src/clj/editor/properties_view.clj
```diff
@@ -662,6 +660,10 @@
    :color :error
    :text (or (.getMessage exception) (.getSimpleName (class exception)))})
 
+(defn- handle-grid-key-pressed [^KeyEvent e]
+  (when (= KeyCode/ESCAPE (.getCode e))
+    (.requestFocus ^Node (.getSource e))))
+
```
matgis, line 667: Should we also consume the event here?

matgis (dismissed):
I think the new Escape key behavior works well. Nicely done! LGTM! ❤️

editor/src/clj/editor/fxui.clj
vlaaad: OMG, `Bindings/valueAt` is exactly what I need, how could I not find it before!

matgis: I really had to look hard to find it myself! I felt very strongly that there should be some built-in function that does this, but it took me nearly 10 minutes of searching to find it. For future reference, the takeaway is that they put a lot of utility functions in the `Bindings` class, so it's a good place to look!

PR #11570, opened 2025-12-04
Improve project open performance in large projects

vlaaad:
A project now opens approximately 15% faster (as measured on a large project with thousands of assets: time to open reduced from 3m36s to 3m02s on my machine).

Technical notes:
Related to #10166


matgis (approved):
Ship it! ❤️

PR #11574, opened 2025-12-04
Improved resource management for glyph generation

JCash:
### Technical changes

* The dmJobThread now handles cancelling a job with children, where some children has finished.
* The res_font.cpp now handles reference counts of resources for each pending job
* The editor doesn't generate a cache-width/cache-height, but respect the user values.

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/editor/src/clj/editor/font.clj
+++ b/editor/src/clj/editor/font.clj
@@ -563,7 +563,7 @@
      :build-fn build-glyph-bank
      :user-data user-data}))
 
-(g/defnk produce-build-targets [_node-id resource font-map material dep-build-targets runtime-generation-build-target]
+(g/defnk produce-build-targets [_node-id resource cache-width cache-height font-map material dep-build-targets runtime-generation-build-target]
   (or (when-let [errors (->> [(validation/prop-error :fatal _node-id :material validation/prop-nil? material "Material")
                               (validation/prop-error :fatal _node-id :material validation/prop-resource-not-exists? material "Material")]
                              (remove nil?)
@@ -586,8 +586,8 @@
                      :render-mode (:render-mode font-map)
                      :all-chars (:all-chars font-map)
                      :characters (:characters font-map)
-                     :cache-width (:cache-width font-map)
-                     :cache-height (:cache-height font-map)
+                     :cache-width cache-width
+                     :cache-height cache-height
                      :sdf-spread (:sdf-spread font-map)
                      :sdf-outline (:sdf-outline font-map)
                      :sdf-shadow (:sdf-shadow font-map))
--- a/engine/dlib/src/dlib/job_thread.cpp
+++ b/engine/dlib/src/dlib/job_thread.cpp
@@ -177,27 +177,27 @@ static void RemoveChildFromParent(JobThreadContext* ctx, HJob hchild)
     }
 }
 
-static void PrintParent(JobThreadContext* ctx, JobItem* parent)
-{
-    if (!parent)
-        return;
+// static void PrintParent(JobThreadContext* ctx, JobItem* parent)
+// {
+//     if (!parent)
+//         return;
 
-    printf("Parent state: nc: %d  done: %d\n", dmAtomicGet32(&parent->m_NumChildren), dmAtomicGet32(&parent->m_NumChildrenCompleted));
-    HJob hchild = parent->m_FirstChild;
-    while (hchild != INVALID_JOB)
-    {
-        JobItem* child = GetJobItem(ctx, hchild);
-        if (child == 0)
-        {
-            printf("  child == 0!!! \n");
-            break; // We cannot iterate further
-        }
+//     printf("Parent state: nc: %d  done: %d\n", dmAtomicGet32(&parent->m_NumChildren), dmAtomicGet32(&parent->m_NumChildrenCompleted));
+//     HJob hchild = parent->m_FirstChild;
+//     while (hchild != INVALID_JOB)
+//     {
+//         JobItem* child = GetJobItem(ctx, hchild);
+//         if (child == 0)
+//         {
+//             printf("  child == 0!!! \n");
+//             break; // We cannot iterate further
+//         }
 
-        printf("  child: %llx (%u %u)  s: %d\n", hchild, ToGeneration(hchild), ToIndex(hchild), child->m_Status);
+//         printf("  child: %llx (%u %u)  s: %d\n", hchild, ToGeneration(hchild), ToIndex(hchild), child->m_Status);
 
-        hchild = child->m_Sibling;
-    }
-}
+//         hchild = child->m_Sibling;
+//     }
+// }
 
 static void FreeJob(JobThreadContext* ctx, HJob hjob)
 {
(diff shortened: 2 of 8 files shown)


editor/src/clj/editor/font.clj
```diff
@@ -586,8 +586,8 @@
                      :render-mode (:render-mode font-map)
                      :all-chars (:all-chars font-map)
                      :characters (:characters font-map)
-                     :cache-width (:cache-width font-map)
-                     :cache-height (:cache-height font-map)
+                     :cache-width cache-width
+                     :cache-height cache-height
```
JCash, line 590: The default values are 0, which means, "grow dynamically" at runtime.
If the user wishes a certain size, they can set the values explicitly.

engine/dlib/src/dlib/job_thread.cpp
```diff
@@ -109,48 +110,109 @@ static inline uint32_t ToIndex(HJob job)
...
+    JobItem* item = GetJobItem(ctx, hjob);
     assert(item != 0); // Generation differed!
     return item;
 }
 
-static JobItem* GetJobItem(JobThreadContext* ctx, HJob hjob)
+// ***********************************************************************************
+// Jobs
+
+static void RemoveChildFromParent(JobThreadContext* ctx, HJob hchild)
```
JCash, line 133: Bug fix: Added this as the linked list otherwise would contain dangling "references", in the middle of the list, and the rest of the list couldn't be traversed.

engine/dlib/src/dmsdk/dlib/object_pool.h
```diff
@@ -203,6 +204,20 @@ class dmObjectPool
...
+     * @param index [type: uint32_t] index of the object
+     * @return object [type: T*] a pointer to the object. Null if the logical index wasn't valid
+     */
+    T* GetPtr(uint32_t index)
+    {
+        Entry* e = &m_Entries[index];
+        if (e->m_Physical >= m_Objects.Size())
+            return 0;
+        return &m_Objects[e->m_Physical];
+    }
```
JCash, line 219: Added this as a way to see if an index was invalid.
Avoids an asset, and lets us use "handles" more conveniently in the dmJobThread.

engine/gamesys/src/gamesys/resources/res_font_private.h
```diff
@@ -31,6 +31,12 @@ namespace dmGameSystem
     struct MaterialResource;
     struct GlyphBankResource;
 
+    struct FontJobResourceInfo
+    {
+        dmJobThread::HJob   m_Job;
+        dmArray<void*>      m_Resources; // the resources that are incref'ed for this job
+    };
```
JCash, line 38: Keep the associated resources incref'd during the job.
Note that between each started job, the actual list of resources may change (adding/removing fonts to this font collection), so we need to keep track of it per-job.

engine/gamesys/src/gamesys/resources/res_font.cpp
```diff
@@ -102,7 +105,11 @@ namespace dmGameSystem
     {
         for (uint32_t i = 0; i < font->m_PendingJobs.Size(); ++i)
         {
-            dmJobThread::HJob hjob = font->m_PendingJobs[i];
+            FontJobResourceInfo* job_info = font->m_PendingJobs[i];
+            dmJobThread::HJob hjob = job_info->m_Job;
+
+            DecRefJobResourceInfo(font->m_Factory, font, job_info);
+            delete job_info;
```
vlaaad: As per codex:
**P1** Don’t free font job resources before cancel completes — `engine/gamesys/src/gamesys/resources/res_font.cpp:104-120`  
In `CancelPendingJobs`, `DecRefJobResourceInfo` and `delete job_info` run before `dmJobThread::CancelJob` finishes. Worker tasks still hold `FontJobResourceInfo` via `FontJobContextWrapper::m_JobInfo` and release the same resources in `PrewarmTextCallbackWrapper`/`CacheMissGlyphCallback` when the job finishes or is canceled, causing use-after-free/double-release. Pending jobs must stay referenced (and their resources incref’d) until the callback runs.

engine/dlib/src/dlib/job_thread.cpp
```diff
@@ -273,20 +347,31 @@ static JobResult CancelJobInternal(JobThreadContext* ctx, HJob hjob)
...
+        }
+
         JobItem* child = GetJobItem(ctx, hchild);
-        if (childresult == JOB_RESULT_PENDING)
+        if (child == 0)
+        {
+            break; // We cannot iterate further
+        }
+
+        if (childresult == JOB_RESULT_PENDING || childresult == JOB_STATUS_PROCESSING)
```
vlaaad: As per codex:

**P1** Wrong enum comparison in `CancelJobInternal` — `engine/dlib/src/dlib/job_thread.cpp:329-375`  
The child pending check uses `childresult == JOB_RESULT_PENDING || childresult == JOB_STATUS_PROCESSING`, but `childresult` is a `JobResult` and `JOB_STATUS_PROCESSING` is a `JobStatus`, so the branch never fires. Parents are reported canceled while children still run, enabling premature resource teardown. Compare against the child job’s status (or use another `JOB_RESULT_PENDING`) instead of a different enum.

vlaaad (dismissed):
The editor changes look good to me. I ran your PR through Codex, and it reported 2 serious issues, though I can't vouch for them — feel free to disregard my comments.

engine/gamesys/src/gamesys/resources/res_font.cpp
JCash: nice catch!
I should also write a script to do this.
While the code would have worked, the suggestion is ofc much cleaner.

engine/dlib/src/dlib/job_thread.cpp
JCash: Nice!
YEah, I blame too verbose build logs, and the warning was buried deep within... :(

engine/gamesys/src/gamesys/resources/res_font.cpp
vlaaad: FYI, I didn't write any scripts to review the code, but used built-in `/review` command.

engine/dlib/src/dlib/job_thread.cpp
```diff
@@ -109,48 +110,109 @@ static inline uint32_t ToIndex(HJob job)
...
+//         {
+//             printf("  child == 0!!! \n");
+//             break; // We cannot iterate further
+//         }
+
+//         printf("  child: %llx (%u %u)  s: %d\n", hchild, ToGeneration(hchild), ToIndex(hchild), child->m_Status);
+
+//         hchild = child->m_Sibling;
+//     }
+// }
```
britzl, line 200: What's the general thought around these kinds of helpers that you may need while developing stuff? Debug printing of the Lua stack, dumping a scene graph or printing a job queue? I like that we have them around like this, but maybe we could have them ifdef'd to be always available in debug builds at no cost so that we still build the debug code and make sure it's up to date and working.

britzl (approved):

engine/dlib/src/dlib/job_thread.cpp
JCash: Yeah, I've thought the same, but haven't acted on it.
Different systems have internal defines for this, but we should come up with a more common practice.

PR #11584, opened 2025-12-07
Add support for mounting live update archives from Android assets

britzl:
It is now possible to mount a live update archive located inside the assets folder of an Android bundle by using a specific prefix to the archive path:

```lua
liveupdate.add_mount("my_mount", "zip:/android_asset/myassets.zip", priority, callback)
```

The archive can be added to the Android bundle using the bundle resources option or as an install-time Play Asset Delivery pack.

Fixes #11577

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/AndroidBundler.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/AndroidBundler.java
@@ -589,7 +589,7 @@ public class AndroidBundler implements IBundler {
             File aabDir = new File(outDir, "aab");
             File baseConfig = new File(aabDir, "BundleConfig.json");
             try (BufferedWriter writer = new BufferedWriter(new FileWriter(baseConfig))) {
-                String uncompressedGlob = "\"assets/game.arcd\"";
+                String uncompressedGlob = "\"assets/game.arcd\",\"assets/**/*.zip\"";
                 if (!extractNativeLibs) {
                     uncompressedGlob += ",";
                     uncompressedGlob += "\"lib/**/*.so\"";
--- a/engine/dlib/src/dlib/zip.cpp
+++ b/engine/dlib/src/dlib/zip.cpp
@@ -24,6 +24,12 @@ Result Open(const char* path, HZip* zip)
     return *zip != 0 ? RESULT_OK : RESULT_NO_SUCH_ENTRY;
 }
 
+Result OpenStream(const char *stream, uint32_t size, HZip* zip)
+{
+    *zip = zip_stream_open(stream, size, 9, 'r');
+    return *zip != 0 ? RESULT_OK : RESULT_NO_SUCH_ENTRY;
+}
+
 void Close(HZip zip)
 {
     if (zip)
--- a/engine/resource/src/mount/mount_android.cpp
+++ b/engine/resource/src/mount/mount_android.cpp
@@ -41,7 +41,7 @@ namespace dmResource
         uint32_t index_length;
     };
 
-    Result MapAsset(const char* path, void*& out_asset, uint32_t& out_size, void*& out_map)
+    Result MapAsset(const char* path, void*& out_asset, void*& out_map, uint32_t& out_size)
     {
         AAssetManager* am = g_AndroidApp->activity->assetManager;
 
@@ -133,7 +133,7 @@ namespace dmResource
         bool mem_mapped_data = false;
         if (strcmp(data_path, "game.arcd") == 0)
         {
-            r = MapAsset(data_path, (void*&)data_asset, data_length, data_map);
+            r = MapAsset(data_path, (void*&)data_asset, data_map, data_length);
             if (r != RESULT_OK)
             {
                 dmLogError("Error when mapping data file '%s', result = %i", data_path, r);
@@ -154,7 +154,7 @@ namespace dmResource
         bool mem_mapped_index = false;
         if (strcmp(index_path, "game.arci") == 0)
         {
-            r = MapAsset(index_path, (void*&)index_asset, index_length, index_map);
+            r = MapAsset(index_path, (void*&)index_asset, index_map, index_length);
             if (r != RESULT_OK)
             {
                 if (mem_mapped_data)
--- a/engine/resource/src/mount/mount_android.cpp
+++ b/engine/resource/src/mount/mount_android.cpp
@@ -89,7 +89,7 @@ namespace dmResource
         return RESULT_OK;
     }
 
-    static Result UnmapAsset(void*& asset, uint32_t size)
+    Result UnmapAsset(void*& asset, uint32_t size)
     {
         if (asset != 0x0)
         {
@@ -160,7 +160,7 @@ namespace dmResource
                 if (mem_mapped_data)
                     UnmapFile(data_map, data_length);
                 else
-                    UnmapAsset(data_asset);
+                    UnmapAsset((void*&)data_asset, data_length);
                 dmLogError("Error when mapping index file, result: %i", r);
                 return RESULT_IO_ERROR;
             }
@@ -173,7 +173,7 @@ namespace dmResource
                 if (mem_mapped_data)
                     UnmapFile(data_map, data_length);
                 else
-                    UnmapAsset(data_asset);
+                    UnmapAsset((void*&)data_asset, data_length);
                 dmLogError("Error mapping liveupdate index file, result = %i", r);
                 return RESULT_IO_ERROR;
             }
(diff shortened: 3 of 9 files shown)


engine/dlib/src/dlib/zip.cpp
```diff
@@ -24,6 +24,12 @@ Result Open(const char* path, HZip* zip)
     return *zip != 0 ? RESULT_OK : RESULT_NO_SUCH_ENTRY;
 }
 
+Result OpenStream(const char *stream, uint32_t size, HZip* zip)
```
britzl, line 27: Open a zip archive from a stream (memory)

engine/resource/src/mount/mount_android.cpp
```diff
@@ -41,8 +41,10 @@ namespace dmResource
         uint32_t index_length;
     };
 
-    Result MapAsset(AAssetManager* am, const char* path,  void*& out_asset, uint32_t& out_size, void*& out_map)
+    Result MapAsset(const char* path, void*& out_asset, void*& out_map, uint32_t& out_size)
     {
+        AAssetManager* am = g_AndroidApp->activity->assetManager;
```
britzl, line 46: Made MapAsset more generic by not requiring AAssetManager as an argument

engine/resource/src/mount/mount_android.cpp
```diff
@@ -87,11 +89,11 @@ namespace dmResource
         return RESULT_OK;
     }
 
-    static Result UnmapAsset(AAsset*& asset)
+    Result UnmapAsset(void*& asset, uint32_t size)
```
britzl, line 92: Again, making things generic

engine/resource/src/providers/provider_zip.cpp
```diff
@@ -196,28 +204,56 @@ static dmResourceProvider::Result Mount(const dmURI::Parts* uri, dmResourceProvi
     memcpy(&archive->m_BaseUri, uri, sizeof(dmURI::Parts));
 
     char path[1024];
-    dmSnPrintf(path, sizeof(path), "%s", uri->m_Path);
-    dmPath::Normalize(path, path, sizeof(path));
 
-    char mount_path[1024];
-    if (dmSys::RESULT_OK != dmSys::ResolveMountFileName(mount_path, sizeof(mount_path), path))
+    // starts with /android_asset/ ?
```
britzl, line 215: I was debating wether I should have a separate scheme like `assetzip:/foo` or use the same scheme and check the path instead `zip:/android_asset/foo.zip`. The use of `/android_asset/` is not new and it is actually the way you can access an asset from a webview. Using the same kind of path here seemed to make sense.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/AndroidBundler.java
```diff
@@ -589,7 +589,7 @@ private File createBundle(Project project, File outDir, File baseZip, ICanceled
             File aabDir = new File(outDir, "aab");
             File baseConfig = new File(aabDir, "BundleConfig.json");
             try (BufferedWriter writer = new BufferedWriter(new FileWriter(baseConfig))) {
-                String uncompressedGlob = "\"assets/game.arcd\"";
+                String uncompressedGlob = "\"assets/game.arcd\",\"assets/**/*.zip\"";
```
britzl, line 592: Make sure any zip archive in assets is uncompressed

AGulev: As I remember you re-pack android manually using a script (in assets bundle repo), both zip and arcdshould be taken into account there as well

AGulev: I mean https://github.com/defold/extension-pad/blob/master/bundle.sh

britzl: I do not repack, but I add more files to the existing .aab file. And these are added with compression level 0.

https://github.com/defold/extension-pad/blob/dev-gradle/bundle.sh#L40

engine/resource/src/providers/provider_zip.cpp
```diff
@@ -47,6 +51,8 @@ struct ZipProviderContext
 {
     dmURI::Parts                m_BaseUri;
     dmZip::HZip                 m_Zip;
+    void*                       m_ZipAsset;
```
JCash, line 59: Perhaps some code documentation is needed, as it's unclear why we have both a HZip and a void*

engine/resource/src/providers/provider_zip.cpp
```diff
@@ -36,6 +37,9 @@ namespace dmResourceProviderZip
 
 const char* LIVEUPDATE_ARCHIVE_MANIFEST_FILENAME = "liveupdate.game.dmanifest";
 
+const char* ANDROID_ASSET_PATH  = "/android_asset/";
+const uint32_t ANDROID_ASSET_PATH_LENGTH = strlen(ANDROID_ASSET_PATH);
```
JCash: tiny opt: `strlen()` will run when running the game, `sizeof(ANDROID_ASSET_PATH)` will be calculated when compiling the .cpp file.

engine/resource/src/providers/provider_zip.cpp
```diff
@@ -196,28 +204,56 @@ static dmResourceProvider::Result Mount(const dmURI::Parts* uri, dmResourceProvi
...
+            dmLogError("Could not open zip file '%s'", mount_path);
+            DeleteZipArchiveInternal(archive);
+            return dmResourceProvider::RESULT_NOT_FOUND;
+        }
     }
 
     dmResourceProvider::Result result = LoadManifest(archive->m_Zip, LIVEUPDATE_ARCHIVE_MANIFEST_FILENAME, &archive->m_Manifest);
     if (dmResourceProvider::RESULT_OK != result)
     {
+        dmLogInfo("Mount LoadManifest error");
```
JCash: If we don't have the string, output the result value instead.

engine/resource/src/providers/provider_zip.cpp
```diff
@@ -196,28 +204,56 @@ static dmResourceProvider::Result Mount(const dmURI::Parts* uri, dmResourceProvi
...
+        {
+            dmLogError("Could not resolve a mount path '%s'", path);
+            DeleteZipArchiveInternal(archive);
+            return dmResourceProvider::RESULT_NOT_FOUND;
+        }
+
+        dmZip::Result zr = dmZip::Open(mount_path, &archive->m_Zip);
+        if (dmZip::RESULT_OK != zr)
+        {
+            dmLogError("Could not open zip file '%s'", mount_path);
```
JCash: Old code (I know), but it's probably good to output the error result as an integer.

JCash (changes requested):
Overall good.
Small doc suggestion.
Mostly, I think we can improve on the error message reporting.

engine/resource/src/providers/provider_zip.cpp
```diff
@@ -50,9 +50,16 @@ struct EntryInfo
 struct ZipProviderContext
 {
     dmURI::Parts                m_BaseUri;
+    
+    // handle to the zip archive
     dmZip::HZip                 m_Zip;
+
+    // pointer to the asset associated with the mounted zip archive
+    // will only be set when the archive was mapped from an asset and not from a file
```
JCash, line 58: 👍

engine/resource/src/providers/provider_zip.cpp
```diff
@@ -77,7 +84,7 @@ static void DeleteZipArchiveInternal(dmResourceProvider::HArchiveInternal _archi
     if (archive->m_Zip)
         dmZip::Close(archive->m_Zip);
     if (archive->m_ZipAsset)
-        dmResource::UnmapAsset(archive->m_ZipAsset, archive->m_ZipLength);
+        dmResource::UnmapAsset(archive->m_ZipAsset, archive->m_ZipAssetLength);
```
JCash, line 87: 👍

engine/resource/src/providers/provider_zip.cpp
```diff
@@ -234,17 +241,19 @@ static dmResourceProvider::Result Mount(const dmURI::Parts* uri, dmResourceProvi
         dmPath::Normalize(path, path, sizeof(path));
 
         char mount_path[1024];
-        if (dmSys::RESULT_OK != dmSys::ResolveMountFileName(mount_path, sizeof(mount_path), path))
+
+        dmSys::Result rr = dmSys::ResolveMountFileName(mount_path, sizeof(mount_path), path)
```
JCash: the compile error of missing ";"

JCash (dismissed):
looks good

PR #11593, opened 2025-12-10
Editor: Optimize atlas preview generation

### LINKED ISSUE #10688: Editor renders empty sprite in paged atlas
No

If paged atlas has more than 8 pages, 9th and next paged contains empty sprites.
Image

**To Reproduce (REQUIRED)**
1. Open attached project
2. Open `test.atlas`
3. See error

Editor renders all sprites correctly
or
Editor shows warning with useful information

  - Version 1.10.1

**Platforms (REQUIRED):**

[pivot-points.zip](https://github.com/user-attachments/files/20272556/pivot-points.zip)

**Logs (OPTIONAL):**

**Workaround (OPTIONAL):**

**Screenshots (OPTIONAL):**

JosephFerano:
Working with Atlases now have better performance characteristics. Changes to image and animation properties that do not affect the atlas texture are now near instant, and changes to Atlas properties that do require recalculation perform faster. Finally, an error is now shown in the Scene View when texture page count exceeds 8 pages, since that is the maximum supported amount.

Fixes https://github.com/defold/defold/issues/10688

### Technical changes

This PR helps address some performance issues with the Atlas workflow

- Speed up reading in individual sprites from disk by parallelizing with pmap
- Check paged-atlas sha1 and return a cached version when we don't need to regenerate, so now editing Atlas image properties doesn't trigger redrawing the atlas (still has more future work)
- Add error message when texture page count exceeds 8 - Fixes #10688

As discussed with @matgis, I reverted some code so that this PR can be merged and we can trickle in the optimizations and fixes instead of having one large PR.


editor/src/clj/editor/atlas.clj
```diff
@@ -505,6 +506,10 @@
 (defn- validate-max-page-size [node-id page-size]
   (validation/prop-error :fatal node-id :validate-max-page-size max-page-size-error-message page-size))
 
+(defn- texture-page-count-error-message [x]
+  (when (> x 8)
+    (format "Atlas pages count (%d) cannot exceed 8 pages per atlas" x)))
```
JosephFerano: If we want to move this magic number somewhere else, let me know. It seems self-explanatory here but if it's bad form for the project, I'd be happy to move it wherever it makes sense

editor/src/clj/editor/atlas.clj
```diff
@@ -589,17 +600,22 @@
...
-  [_node-id extrude-borders image-resources inner-padding margin layout-data-generator]
+  [_node-id extrude-borders image-resources inner-padding margin layout-data-generator max-page-size]
   (let [flat-image-resources (filterv some? (flatten image-resources))
         image-sha1s (mapv (fn [resource]
                             (resource-io/with-error-translation resource _node-id nil
                               (resource/resource->path-inclusive-sha1-hex resource)))
                           flat-image-resources)]
     (g/precluding-errors image-sha1s
-      (let [packed-image-sha1 (digestable/sha1-hash
+      (let [layout-data (texture-util/call-generator layout-data-generator)
```
JosephFerano: I profiled this and it doesn't seem to be too much overhead

editor/src/clj/editor/atlas.clj
```diff
@@ -557,15 +562,21 @@
       (g/->error error-node-id :max-page-size :fatal nil (.getMessage error)))))
 
 (defn- generate-packed-page-images [{:keys [digest-ignored/error-node-id image-resources layout-data-generator]}]
-  (let [buffered-images (mapv #(resource-io/with-error-translation % error-node-id nil
-                                 (image-util/read-image %))
-                              image-resources)]
+  (let [buffered-images (->> image-resources
+                             (pmap #(resource-io/with-error-translation % error-node-id nil
+                                      (image-util/read-image %)))
+                             (vec))]
```
JosephFerano, line 568: Informal benchmarking from the REPL; with the mapv, for 100 images, I'm getting around ~500ms, with pmap, I'm getting ~200ms. From the flamegraph, most of that remaining 200ms is coming from the layout-atlas-pages call.

Testing the pmap vs mapv parts I go from like 350-400ms to ~60ms

I tried this command to clear my disk caches on linux.

```
sudo sync && echo 3 | sudo tee /proc/sys/vm/drop_caches
```

Didn't seem to do much

editor/src/clj/editor/atlas.clj
```diff
@@ -760,19 +776,31 @@
...
   (output gpu-texture g/Any :cached
           (g/fnk [_node-id packed-page-images-generator texture-profile]
-            (-> (texture-util/construct-gpu-texture _node-id packed-page-images-generator texture-profile)
-                (texture/set-params {:min-filter gl/nearest
-                                     :mag-filter gl/nearest}))))
+            ;; NOTE: Temporary fix until we can properly disconnect the image nodes that invalidate this
+            (if (and (= (:sha1 packed-page-images-generator) (g/user-data _node-id :gpu-texture-sha1))
+                     (g/user-data _node-id :gpu-texture-cached))
+              (when-let [texture-ref (g/user-data _node-id :gpu-texture-cached)]
+                (.get texture-ref))
```
matgis: I think this will return `nil` if the `gpu-texture` referenced by the `WeakReference` has been garbage-collected? We should treat this the same as a sha1 mismatch, right?

editor/src/clj/editor/atlas.clj
matgis: I don't think we want to call the `layout-data-generator` here. As you can see,  previously we would defer layout generation until the `packed-page-images-generator` was called. Instead, we should return the `ErrorValue` from `generate-packed-page-images`, where we need to call the `layout-data-generator`.

editor/src/clj/editor/atlas.clj
matgis: -> "Atlas page count"

matgis (changes requested):

editor/src/clj/editor/atlas.clj
JosephFerano: Yes, sorry, this was a little careless on my part.

editor/src/clj/editor/atlas.clj
JosephFerano: I don't quite understand what you mean here. Do you mean the ErrorValue related to the 9+ pages validation error? This is a different fix/optimization I introduced that will check if the atlas page count changed in a way where the atlas needs to be regenerated. First a fix because since the page count isn't part of the SHA, then when pages were added or removed, the atlas wasn't being regenerated with this new SHA check optimization. Before though, in 1.11.2, any change to the W or H of `max-page-size`, regardless of whether it needed a regeneration, would still trigger one. So to improve that, I introduced an optimization by making it part of the SHA so that even if the W or H change, it's only if the page count differs that the the SHA produced will invalidate the cached one, thus triggering a regeneration.

The downside is that we need to calculate the layout-data so that we can then call `calculate-texture-page-count` and pass that to the SHA generation. I profiled the `call-generator` call and it seems to take around 10ms, which isn't great, but it seemed like a worthwhile trade-off to avoid regeneration when it wasn't needed.

AGulev:
Please don't forget to update the PR's title and text to explain it in terms of "user facing changes" so it can be used in Release Notes (automatically by script)

All the technical details for the other engine devs can be explained after `### Technical changes` and will be excluded from the release notes automatically by script, see this PR as an example:
https://github.com/defold/defold/pull/11604

JosephFerano:
@AGulev Let me know if that's good.

AGulev:
Usually we don't mention PR or something like that for our users (in a section before `Technical changes`).
It is end up in posts like this one: https://forum.defold.com/t/defold-1-12-0-beta/81981 - so imagine how a Defold user read your post on forum in release notes in such release post (not as a PR on github)

AGulev:
CleanShot 2025-12-15 at 12 38 49@2x

JosephFerano:
@AGulev Ah great, thank you.

JosephFerano:
Ok improved it again

matgis (approved):
Ship it! ❤️

PR #11595, opened 2025-12-10
Added an ability to specify `library.defold_min_version` for extensions to prevent using incompatible extensions 

### LINKED ISSUE #11276: Decide on extension editor versioning system

AGulev:
Extensions that define `library.defold_min_version` are now validated against the current Defold version.
If the extension requires a newer version than the one being used, the build will fail with an error message.

Fix https://github.com/defold/defold/issues/11276


editor/src/clj/editor/library.clj
```diff
@@ -249,14 +257,25 @@
...
+                         {:status :error
+                          :reason :defold-min-version
+                          :required min-version
+                          :current (str EngineVersion/version)})))
+                   {:status :error
+                    :reason :missing-game-project})))
+             (catch Exception e
+               {:status :error
+                :reason :invalid-zip
+                :exception e})))))
```
AGulev, line 278: I can't restore old PR, so I make a new one

This is the main change, single zip reading to check for both possible errors

editor/src/clj/editor/settings.clj
```diff
@@ -172,6 +172,21 @@
...
+        (comp
+          (filter #(and (= :error (:status %))
+                        (= :defold-min-version (:reason %))))
+          (map (fn [{:keys [uri required current]}]
+                 (g/map->error
+                   {:severity :fatal
+                    :message (format "Library '%s' requires Defold %s or newer (bob.jar is %s). Update Defold or check older extension versions for compatibility."
+                                     (str uri) (str required) (str current))}))))
+        dependencies))
+
```
AGulev, line 189: not sure if I can use here something like :
```
:severity :fatal
                    :message (localization/message
                               "notification.fetch-libraries.min-version.error"
                               {"uri" (str uri)
                                "required" (str required)
                                "current" (str current)})}
                                ```

Comment (vlaaad) [reply editor/src/clj/editor/settings.clj:189]: Unfortunately, it's currently not possible since we haven't localized the error views. We'll do it later.

File: editor/src/clj/editor/workspace.clj [hunk 3]
```diff
@@ -564,20 +587,23 @@ ordinary paths."
...
-                    "notification.fetch-libraries.dependencies-error.error"
-                    {"dependencies" (coll/join-to-string "\n" (e/map dialogs/indent-with-bullet error))})
-         :actions [{:message (localization/message "notification.fetch-libraries.dependencies-error.action.open-game-project")
-                    :on-action #(ui/execute-command
-                                  (ui/contexts (ui/main-scene))
-                                  :file.open
-                                  "/game.project")}]})
-      (notifications/close! notifications ::dependencies-error))))
+    (let [min-version-states (into [] (filter #(= :defold-min-version (:reason %))) lib-states)
+          failing-uris (into #{} (map :uri) min-version-states)
```
vlaaad: These 2 lines are repeated (here and on lines 554-555); we shouldn't do that since it's unnecessary.

editor/src/clj/editor/workspace.clj
```diff
@@ -550,6 +550,29 @@ ordinary paths."
...
+                       "required" (str required)
+                       "current" (str current)})
+           :actions [{:message (localization/message "notification.fetch-libraries.dependencies-error.action.open-game-project")
+                      :on-action #(ui/execute-command
+                                    (ui/contexts (ui/main-scene))
+                                    :file.open
+                                    "/game.project")}]}))
+      ;; Close for those no longer failing
+      (doseq [uri (set/difference all-uris failing-uris)]
+        (notifications/close! notifications (notification-id uri))))
```
vlaaad: Perhaps we should only use a single notification for library version errors? It seems to me that `(set/difference all-uris failing-uris)` is not enough in this scenario:
1. A user adds a dependency with an invalid min version -> notification is shown
2. The user removes this dependency -> notification is **not** hidden because it's not in `all-uris` at all since it's removed.

vlaaad (changes requested):
Good job! I think we should avoid repeating the same calculations and ensure a notification is hidden when you add a library with an invalid version and then remove it.

editor/src/clj/editor/workspace.clj
AGulev: fixed

editor/src/clj/editor/workspace.clj
AGulev: fixed

vlaaad (approved):
LGTM!

PR #11596, opened 2025-12-10
Refactored editor project loading for simpler profiling

matgis:
* Make it easier to call individual post-transaction steps from `load_project.clj`.
* Gather stats around `:scene` output dependencies in `load_project.clj`.
* Add `localization/set-message-key`, for when you want to alter the message key but keep the arguments of a message pattern.
* Add `dev/run-with-terminal-progress` for reporting progress from tasks without involving JavaFX.
* Removed progress reporting from `load_project.clj`, since it was interfering with timings.


vlaaad (approved):
LGTM!

PR #11600, opened 2025-12-11
Skip unnecessary traversal in update-overrides

matgis:
Skipping some graph traversal in the `update-overrides` step that concludes transactions is now around 14.7 times faster. In the large project we were testing with the `update-overrides` step during the initial project loading went from taking 1:43 to 6.8 seconds, resulting in a 27% project load time improvement.


vlaaad (approved):
Ship it!

vlaaad:
It would also be nice to know total improvement, e.g. if time to open went from 10m to 9m, the improvement is (10-9)/10=10%

PR #11602, opened 2025-12-11
Add support for setting and getting custom attribute data for model components

### LINKED ISSUE #9746: Setting/getting instanced attributes for models
I'd like to change the values of instanced attributes in runtime to animate models, but changing them is currently not available via `go.set`/`go.animate`:

image

image

image

Add support for setting/getting instanced attributes for models via the go.set.

Do not use instancing for those models.

- https://github.com/defold/defold/pull/8743
- https://github.com/defold/defold/pull/8562

Jhonnyg:
`go.get`, `go.set` and `go.animate` can now be used to query or set custom vertex data for models. This will only be supported for the first mesh in the component, if the model file contains multiple meshes there is currently no way of indexing or referencing those. This will be added at a later time!

Note: that this can be an expensive operation depending on how large the mesh is and if you are setting an instanced attribute or not. To learn more about instancing read [here](https://defold.com/manuals/material/#instancing), and to learn more about custom vertex formats read [here](https://defold.com/manuals/material/#setting-custom-vertex-attribute-data).

This example uses go.animate on the separate cubes to animate a custom vertex stream called "custom_color":

```lua
for i = 1, 16 do
    local rnd_color = vmath.vector4(math.random(), math.random(), math.random(), 1)
    local rnd_time = math.random(2, 7)
    go.animate("/go" .. i .. "#model", "custom_color", go.PLAYBACK_LOOP_PINGPONG, rnd_color, go.EASING_INQUAD, rnd_time)
end
```

instancing_custom_vertex_formats

Test project: 
[issue-9746-model-set-custom-attributes.zip](https://github.com/user-attachments/files/24735176/issue-9746-model-set-custom-attributes.zip)

Fixes #9746


flametime:
Hello sir!

Since you already did this for 3D models, maybe you could also add the same thing for gui.set()? 🙏

Jhonnyg:
> Hello sir!
> 
> Since you already did this for 3D models, maybe you could also add the same thing for gui.set()? 🙏

Hi, models are not supported by gui components afaik, can you explain a bit about what issue you are facing?

flametime:
@Jhonnyg 

Sure, sorry for the confusion.

The issue is not about models in GUI, but about batching

If I use gui.set() to pass per-node data (for example via constants), it breaks batching/instancing

To keep batching, I need per-instance data via vertex attributes, but GUI nodes don’t support custom vertex attributes right now

My use case is a GUI shader with rounded corners, where each corner radius is controlled individually per box node

I’m doing this purely in the shader so the corners stay perfectly smooth and don’t get pixelated

Jhonnyg:
@flametime thanks, I understand better now! Yeah, GUIs doesn't use instancing at all currently, this is something I'd have to look at separately. Could you open up a ticket and explain your usecase in the ticket? you can just copy-paste what you wrote here. Just so we can keep track of it, it will get lost in here otherwise :)

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -118,9 +118,11 @@ namespace dmGameSystem
         uint32_t                    m_InstanceRenderHash;
         uint32_t                    m_BoneIndex;
         uint32_t                    m_MaterialIndex;
-        uint32_t                    m_Enabled                     : 1;
-        uint32_t                    m_AttributeRenderDataIndex    : 16;
-        uint32_t                    m_PerInstanceCustomAttributes : 1;
+        uint32_t                    m_Enabled                      : 1;
+        uint32_t                    m_AttributeRenderDataIndex     : 16;
```
JCash: Probably should move into a separate uint16_t, to avoid bit shifts when reading the variable.

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -777,7 +785,111 @@ namespace dmGameSystem
...
+                non_default_attribute.m_Infos[attribute_index]              = attr_model;
+                non_default_attribute.m_Infos[attribute_index].m_VectorType = attr_material.m_VectorType;
+                non_default_attribute.m_VertexStride                       += value_byte_size;
+            }
+        }
+
+        uint32_t vertex_count     = render_item->m_Buffers->m_VertexCount;
+        uint32_t vertex_data_size = non_default_attribute.m_VertexStride * vertex_count;
+        void* attribute_data      = malloc(vertex_data_size);
+        memset(attribute_data, 0, vertex_data_size);
```
JCash: Note that there is also `calloc()` that returns zeroed memory.

Question is though, why is the clearing needed?
Isn't WriteMeshAttributes() writing all data to the buffer?

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1744,7 +1780,19 @@ namespace dmGameSystem
 
                 if (!attribute_rd->m_Initialized)
                 {
-                    SetupMeshAttributeRenderData(render_context,
+                    SetupMeshAttributeRenderData(world, component,
+                        render_context,
+                        render_material,
+                        render_item,
+                        component->m_Resource->m_Materials[material_index].m_Attributes,
+                        component->m_Resource->m_Materials[material_index].m_AttributeCount,
```
JCash, line 1786: This should be possible to put outside of the loop, right?
Materials shouldn't change within a batch?

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1744,7 +1780,19 @@ namespace dmGameSystem
...
+                        attribute_rd);
+                }
+                
+                if (render_item->m_DynamicVertexAttributesDirty)
+                {
+                    SetMeshAttributeRenderData(world, component,
+                        render_context,
                         render_material,
                         render_item,
                         component->m_Resource->m_Materials[material_index].m_Attributes,
```
JCash, line 1796: Same here I think?

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1937,7 +1985,8 @@ namespace dmGameSystem
         if (has_custom_attributes)
         {
             dmGraphics::VertexAttributeInfos attribute_infos;
-            FillAttributeInfos(0, INVALID_DYNAMIC_ATTRIBUTE_INDEX, // Not supported yet
+            FillAttributeInfos(&world->m_DynamicVertexAttributePool,
+                render_item->m_DynamicVertexAttributeIndex,
                 c->m_Resource->m_Model->m_Materials[material_index].m_Attributes.m_Data,
```
JCash, line 1988: I guess we can cleanup the world space code path later, but feels like we should pass the material to this function?

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -2577,6 +2626,32 @@ namespace dmGameSystem
...
+            return 0;
+        return &component->m_RenderItems[render_item_index];
+    }
+
+    static bool CompModelGetMaterialAttributeCallback(void* user_data, dmhash_t name_hash, const dmGraphics::VertexAttribute** attribute)
+    {
+        ModelComponent* component = (ModelComponent*) user_data;
+
+        // TODO: Support for multiple materials
+        MaterialInfo* material_info = &component->m_Resource->m_Materials[0];
```
JCash, line 2637: I think this should handle if the material is overridden.
E.g.
```c++
MaterialResource* resource = GetMaterialResource(component, component->m_Resource, 0);
MaterialInfo* material_info = &resource->m_Materials[0];
```

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -2612,7 +2687,20 @@ namespace dmGameSystem
...
+        dmRender::HMaterial material = GetComponentMaterial(component, component->m_Resource, 0);
+
+        if (GetMaterialConstant(material, params.m_PropertyId, params.m_Options.m_Index, out_value, false, CompModelGetConstantCallback, component) == dmGameObject::PROPERTY_RESULT_OK)
+        {
+            return dmGameObject::PROPERTY_RESULT_OK;
+        }
+
+        // TODO: This will use the dynamic vertex attribute index for the first render item only.
+        //       We need a way to address sub-components to support custom attributes for sub-models
+        MeshRenderItem* render_item = GetMeshRenderItem(component, 0);
```
JCash, line 2696: Since we _do_ get to this point, what stops us from using the `params.m_Options.m_Index` value here?

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -2674,7 +2762,32 @@ namespace dmGameSystem
...
-        return SetMaterialConstant(GetComponentMaterial(component, component->m_Resource, 0), params.m_PropertyId, params.m_Value, params.m_Options.m_Index, CompModelSetConstantCallback, component);
+
+        dmRender::HMaterial material = GetComponentMaterial(component, component->m_Resource, 0);
+        dmGameObject::PropertyResult res = SetMaterialConstant(material, params.m_PropertyId, params.m_Value, params.m_Options.m_Index, CompModelSetConstantCallback, component);
+
+        // Only check attributes if the constant property was not found
+        if (res == dmGameObject::PROPERTY_RESULT_NOT_FOUND)
+        {
+            const dmGraphics::VertexAttribute* attribute = 0;
+            MeshRenderItem* render_item = GetMeshRenderItem(component, 0);
```
JCash, line 2769: Same question as above

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -2674,7 +2762,32 @@ namespace dmGameSystem
...
+            if (res == dmGameObject::PROPERTY_RESULT_OK)
+            {
+                // Mark the render item as dirty if the overridden attribute is not an instanced attribute
+                // For instanced attributes, we will copy the data anyway so this just makes sure we don't
+                // copy the base vertex data when we don't need to.
+                if (attribute->m_StepFunction != dmGraphics::VERTEX_STEP_FUNCTION_INSTANCE && render_item)
+                {
+                    // TODO: This will mark only the first render item as dirty.
+                    //       We need a way to address sub-components to support custom attributes for sub-models
+                    render_item->m_DynamicVertexAttributesDirty = 1;
```
JCash: The dirty flag is correct though? For the _selected_ `render_item`, we mark it as dirty?
It feels like the comment is about how we _select_ the render item to modify? (i.e. the comment should be moved up?)

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -2453,7 +2453,7 @@ namespace dmGameSystem
         // Only check attributes if the constant property was not found
         if (res == dmGameObject::PROPERTY_RESULT_NOT_FOUND)
         {
-            return SetMaterialAttribute(sprite_world->m_DynamicVertexAttributePool, &component->m_DynamicVertexAttributeIndex, material, set_property, params.m_Value, CompSpriteGetMaterialAttributeCallback, component);
+            return SetMaterialAttribute(sprite_world->m_DynamicVertexAttributePool, &component->m_DynamicVertexAttributeIndex, material, set_property, params.m_Value, CompSpriteGetMaterialAttributeCallback, component, 0);
```
JCash, line 2456: Same as model, we do have an `params.m_Options.m_Index`, should that be used?

JCash (changes requested):
Overall good
Seems like some potential improvements could be made?

engine/gamesys/src/gamesys/components/comp_model.cpp
Jhonnyg: But, we don't use the index for other properties that we _could_ use the index for. A bit further up we have:

```
else if (params.m_PropertyId == PROP_MATERIAL)
{
    return GetResourceProperty(dmGameObject::GetFactory(params.m_Instance), GetMaterialResource(component, component->m_Resource, 0), out_value);
}
```        
So I'd rather pick one or the other here and not mix supporting one or the other. I think we can support indexes for poperties (but wouldn't we rather support setting/getting via material names?). Either way, I think we can do that in a separate PR.

engine/gamesys/src/gamesys/components/comp_model.cpp
Jhonnyg: Yes, you are right :)

engine/gamesys/src/gamesys/components/comp_sprite.cpp
Jhonnyg: It's not an index, it's a pointer to the attribute stream we have set. It's so that we can mark render items dirty if the attribute isn't instanced:

```
const dmGraphics::VertexAttribute* attribute = 0;
MeshRenderItem* render_item = GetMeshRenderItem(component, 0);
uint16_t* dynamic_attribute_index_ptr = render_item ? &render_item->m_DynamicVertexAttributeIndex : 0;

res = SetMaterialAttribute(world->m_DynamicVertexAttributePool, dynamic_attribute_index_ptr, material, params.m_PropertyId, params.m_Value, CompModelGetMaterialAttributeCallback, component, &attribute);
if (res == dmGameObject::PROPERTY_RESULT_OK)
{
    // Mark the render item as dirty if the overridden attribute is not an instanced attribute
    // For instanced attributes, we will copy the data anyway so this just makes sure we don't
    // copy the base vertex data when we don't need to.
    if (attribute->m_StepFunction != dmGraphics::VERTEX_STEP_FUNCTION_INSTANCE && render_item)
    {
        render_item->m_DynamicVertexAttributesDirty = 1;
    }
}
```

engine/gamesys/src/gamesys/components/comp_model.cpp
Jhonnyg: Materials doesn't change, but the m_Materials is MaterialInfo structs that can contain overridden attributes, which isn't part of the hash for non-instanced local materials. 

I _think_ this is the case (I guess we could verify perhaps):
Two components can be part of the same batch based on the dmRender::Material, but each model component can have different attribute overrides in the MaterialInfo struct (res_model.h). For non-instanced components, the HashRenderItem function only includes the material but not the attributes, so I think this makes sense.

engine/gamesys/src/gamesys/components/comp_model.cpp
Jhonnyg: We already cover this by passing in the overridden material to the GetMaterialAttribute function. If the callback returns false (i.e there's no component-level override), that function will take the data from the material instead.

PR #11609, opened 2025-12-14
Improved atlas texture creation speed in the editor

AGulev:


JosephFerano:
On my machine, the Encode operation that was eliminated that was taking around 1 sec, is gone.

AGulev:
Thank you, @JosephFerano ! Mac M1 has pretty fast memory, so it wasn't that noticeable on my machine where it takes 120~150ms for the same atlas

JCash (approved):

PR #11611, opened 2025-12-15
Added a persistent Discord banner to the Defold Editor

### LINKED ISSUE #5451: Enable Discord presence
Much like Steam presence on friends list, Discord is also a powerful marketing tool to spread awareness of software and games.

Add Discord presence. It has a new API that will probably be easy to implement into all platform versions.

Defold team will need to setup official Defold app on Discord. Community can help implement the integration.

Supporting Discord presence is popular in Software, not just games. 90% of the time this is my presence since I always have it open, but Discord also detects top window so it would change to Defold to more than 50% of the time at least since I use Defold more. 😎
discord

AGulev:
Defold Editor will be shown in Discord :
CleanShot 2025-12-15 at 12 01 34@2x

Fix https://github.com/defold/defold/issues/5451

### DIFF
--- a/editor/src/java/com/defold/editor/DiscordRichPresence.java
+++ b/editor/src/java/com/defold/editor/DiscordRichPresence.java
@@ -1,0 +1,92 @@
+// Copyright 2020-2025 The Defold Foundation
+// Copyright 2014-2020 King
+// Copyright 2009-2014 Ragnar Svensson, Christian Murray
+// Licensed under the Defold License version 1.0 (the "License"); you may not use
+// this file except in compliance with the License.
+//
+// You may obtain a copy of the License, together with FAQs at
+// https://www.defold.com/license
+//
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+package com.defold.editor;
+
+import com.google.gson.JsonObject;
+import com.jagrosh.discordipc.IPCClient;
+import com.jagrosh.discordipc.IPCListener;
+import com.jagrosh.discordipc.entities.ActivityType;
+import com.jagrosh.discordipc.entities.Packet;
+import com.jagrosh.discordipc.entities.RichPresence;
+import com.jagrosh.discordipc.entities.User;
+import com.jagrosh.discordipc.exceptions.NoDiscordClientException;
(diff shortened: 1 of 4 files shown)


editor/src/java/com/defold/editor/DiscordRichPresence.java
```diff
@@ -0,0 +1,92 @@
...
+final class DiscordRichPresence {
+
+    private static final Logger logger = LoggerFactory.getLogger(DiscordRichPresence.class);
+    private static final long CLIENT_ID = 1450057060514402396L;
+    public static final String STATE = "Making a game";
+
+    static void startAsync() {
+        Thread thread = new Thread(DiscordRichPresence::connect, "discord-rich-presence");
+        thread.setDaemon(true);
+        thread.start();
```
vlaaad: I think we should use a virtual thread (which is a daemon by default) instead of a daemon platform thread to make it a bit more efficient.

Maybe something like
```java
Thread.ofVirtual().name("discord-rich-presence").start(DiscordRichPresence::connect);
```

vlaaad (commented):
Should we perhaps also add a runtime shutdown hook that disconnects the client?

vlaaad (dismissed):
Works great BTW! Maybe it's not necessary to properly clean up the connection on shutdown

editor/src/java/com/defold/editor/DiscordRichPresence.java
AGulev: Fixed.
Disconnect isn't needed

vlaaad (approved):
Ship it!

AGulev:
We agreed that this should be a preference option and that it should be disabled by default.

Jerakin:
Configs are nice, but disabled by default feels slightly odd. 

* Discord itself has an option to opt out of "presence". To me it feels like anyone who _doens't_ want people to know what they are doing have opt out there already.
* Defold is a "tame" software.
* In general I think that software that implements "presence" have it on by default, most probably doesn't even have any config to turn it off but expects users to do so globally on Discords side.

AGulev:
@Jerakin maybe, but that's what @britzl and @matgis insisted on

britzl:
I know we talked about it a great deal. I am undecided. I think @Jerakin provides some good arguments for why it should be on by default.

AGulev:
This PR doesn't contain such an option at all, but I can add it. We need to make a decision about what the default should be here.

vlaaad:
What was the argument against? Privacy? I remember there was some outage online about discord around the time this was considered?

PR #11612, opened 2025-12-15
Update Android SDK to version 36.1.0 and targetSdkVersion to 36

### LINKED ISSUE #11554: Update Android SDK to 36.x
**Task (REQUIRED):**
It’s time to update to the latest SDK so we can transition early, before Google eventually requires developers to do it.

https://developer.android.com/about/versions/16/behavior-changes-16

AGulev:
Android SDK updated to the latest stable version 36.1.0

Fix https://github.com/defold/defold/issues/11554


AGulev:
I built an apk/aab locally with a local extender and tested on my device, also uploaded to google play. It works.

PR #11617, opened 2025-12-16
Editor: Optimize Atlas Preview Image Generation for faster iterations

### LINKED ISSUE #8463: Long time waiting while working with animations
Describe the bug (REQUIRED)
When working with animations in Defold, the system experiences significant performance issues. Actions such as copying, deleting, and moving elements upwards cause the computer to freeze for a few seconds.

To Reproduce (REQUIRED)

Open a project in Defold.
Work with animations (e.g., create/edit animations).
Perform actions like copying, deleting, or moving elements.
Observe the system lagging or freezing.
Expected behavior (REQUIRED)
The expected behavior is smooth and responsive performance when working with animations, without causing the system to freeze or lag.

Defold version (REQUIRED):

Version 1.6.3
Platforms (REQUIRED): Windows 10

Platforms: [Please list the platforms you are experiencing this issue on, e.g., Windows, macOS]
OS: Windows 10
Device: Desktop PC

Workaround (OPTIONAL):

https://github.com/defold/defold/assets/27631293/36ea8940-bb7a-4514-a7b5-46d04e20d704

JosephFerano:
The time it takes to generate a new atlas preview after changing its properties was reduced significantly. It should now be around 6x faster on slower machines. As well, dragging/dropping images or moving them up/down in the Outline view of an Atlas when working with Animations will no longer trigger a recalculation.

Fixes #8463

### Technical changes

1.) Bob and texc have a generalized approach to texture generation, allowing Defold to pick and choose which operations to execute based on configuration. However, for the purposes of generating a preview texture, this configuration will be constant, and it's currently blocking the UI thread as it calculates an atlas. By creating a new code path, several things were accomplished;

- Existing path requires that texture data be extracted into a collection of mipmaps. The texc library was creating 1 allocation for the output data, but then the mipmap collection created a new data buffer to extract that data and move it to the Java side. The new path passes a Java byte array that the C++ code can write directly to, avoiding the extra allocation.
- By using `ScopedByteArrayCritical` instead of `ScopedByteArray` we avoid another allocation on the C++ side. Great find @AGulev 
- The preview texture required several passes; ABGR -> RGBA conversion, a Y flip, and pre-multiplying the alpha. These 3 operations were being done in separate passes. The new path merges all 3 operations into a single for loop.

2.) Reordering images for animations in the outline view for an Atlas was producing a different SHA, by sorting the hashes before calculating the final hash, we can avoid a recalculation. The sort takes about ~10 ms for 400 images, an nlogn sort should keep this in the same order of magnitude up to a few thousands of images.


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/TextureGenerator.java
```diff
@@ -589,6 +584,38 @@ private static TextureImage.CompressionType textureCompressorToCompressionType(S
         return TextureImage.CompressionType.COMPRESSION_TYPE_DEFAULT;
     }
 
+    public static GenerateResult generateAtlasPreview(BufferedImage image) throws TextureGeneratorException {
+        int width = image.getWidth();
+        int height = image.getHeight();
+        byte[] inputBuffer = getByteBuffer(image).array();
+        byte[] outputBuffer = new byte[width * height * 4];
```
JosephFerano, line 593: I went with this option + `ScopedByteArrayCritical` which calls `GetPrimitiveArrayCritical` because I couldn't find a way to have a direct byte buffer return the underlying byte array.

editor/src/clj/editor/atlas.clj
```diff
@@ -602,14 +603,17 @@
 (g/defnk produce-packed-page-images-generator
   [_node-id extrude-borders image-resources inner-padding margin layout-data-generator max-page-size]
   (let [flat-image-resources (filterv some? (flatten image-resources))
-        image-sha1s (mapv (fn [resource]
+        image-sha1s (pmap (fn [resource]
                             (resource-io/with-error-translation resource _node-id nil
```
JosephFerano, line 606: This went from 10 ms to sub ms

editor/src/clj/editor/atlas.clj
```diff
@@ -602,14 +603,17 @@
...
                               (resource/resource->path-inclusive-sha1-hex resource)))
-                          flat-image-resources)]
+                          flat-image-resources)
+        ;; Note: Dragging or reordering images up and down the outline for animations results in a different order,
+        ;; which produces a different hash, which triggers an unnecessary atlas regeneration, so hash unordered
+        images-sha1 (Murmur3/hashUnordered image-sha1s)]
```
JosephFerano: Thanks @vlaaad

engine/texc/src/texclibrary_jni.cpp
```diff
@@ -42,7 +42,7 @@ JNIEXPORT jlong JNICALL Java_TexcLibraryJni_CreateImage(JNIEnv* env, jclass cls,
         dmJNI::ScopedString j_path(env, _path);
         const char* path = strdup(j_path.m_String ? j_path.m_String : "null");
 
-        dmJNI::ScopedByteArray j_array(env, array);
+        dmJNI::ScopedByteArrayCritical j_array(env, array);
```
JosephFerano, line 45: I checked this thorough stackoverflow answer
https://stackoverflow.com/questions/23258357/whats-the-trade-off-between-using-getprimitivearraycritical-and-getprimitivety

It seems pretty safe to do this change considering how we're calling this library.

engine/texc/src/texc_private.cpp
```diff
@@ -336,6 +336,32 @@ namespace dmTexc
...
+                uint8_t r = *(src++);
+
+                *(dst++) = (uint8_t)((r * a) / 255);
+                *(dst++) = (uint8_t)((g * a) / 255);
+                *(dst++) = (uint8_t)((b * a) / 255);
+                *(dst++) = a;
+            }
+        }
+    }
+
```
AGulev, line 364: It seems like rewriting of the `/255` may help with auto-vectorization: https://godbolt.org/z/6Go58Gbzd

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/TextureGenerator.java
```diff
@@ -589,6 +584,38 @@ private static TextureImage.CompressionType textureCompressorToCompressionType(S
         return TextureImage.CompressionType.COMPRESSION_TYPE_DEFAULT;
     }
 
+    public static GenerateResult generateAtlasPreview(BufferedImage image) throws TextureGeneratorException {
```
AGulev, line 589: Could you please add a comment that this functions used from the editor?

engine/texc/src/texc_private.cpp
AGulev: I don't insist we have to do it, just took a look out of curiosity

matgis (dismissed):
Code changes look good to me! Great investigation and nice optimizations! ❤️

editor/src/clj/editor/atlas.clj
vlaaad: Just a note, `Murmur3/hashUnordered` is a java hash (i.e. an int, 32 bits), not sha1 (160 bits). But maybe you can write a function that procudes an unordered sha1 hash from a list of sha1 hashes, similar to what `Murmur3/hashUnordered` does?

engine/texc/src/texc_private.cpp
JosephFerano: This is a really interesting find. I just checked a bunch of code with armv8, and it turns out that most of the code is being vectorized for that platform, both the unmerged and the new merged path. But not for x64. That explains at least some of the performance differences we've been seeing between the M-series and my AMD. I introduced this pragma;

```c++
            #pragma clang loop vectorize(enable) interleave(enable)
            for (uint32_t x = 0; x < width; ++x)
            {
```

And this was enough to vectorize it. It generated SSE instructions. If we're already compiling with SSE then it should be fine to leave this pragma in, I hope. I get about a 60-80ms speed-up. I'll check the code in, if we need to remove it, let me know.

editor/src/clj/editor/atlas.clj
JosephFerano: Ok, so please take a look at the latest commit. I went with a simple XOR on the bytes because the collection of filenames should be unique and we don't need to be cryptographically secure for this. Other things I tried were adding sum of squares and sum of mods but just feels like overkill. Worst case on a rare collision is just a recalculation. Thoughts?

JosephFerano: It's about ~1.3ms on my machine for 400 images. Which doesn't seem all that great.

vlaaad:
I now get reflection warnings when starting the editor, though it seems it's caused by a previous change:
```
Reflection warning, editor/atlas.clj:793:19 - reference to field get can't be resolved.
```

editor/src/clj/editor/atlas.clj
```diff
@@ -602,14 +602,17 @@
...
-        image-sha1s (mapv (fn [resource]
+        ;; Note: Dragging or reordering images up and down the outline for animations results in a different order,
+        ;; which produces a different hash, which triggers an unnecessary atlas regeneration, so hash unordered
+        image-sha1s (pmap (fn [resource]
                             (resource-io/with-error-translation resource _node-id nil
-                              (resource/resource->path-inclusive-sha1-hex resource)))
-                          flat-image-resources)]
+                              (resource/resource->path-inclusive-sha1 resource)))
+                          flat-image-resources)
+        images-sha1 (digestable/unordered-sha1s->sha1-hash image-sha1s)]
```
vlaaad: Maybe name it `sha1s->unordered-sha1`? Or something like that... The inputs are ordered after all, and it's the output hash that's unordered.

I think there is a bug on this line: the hash should be computed AFTER `g/precluding-errors` — `image-sha1s` might include `ErrorValue` instances if loading failed, and this will cause a class cast exception in `unordered-sha1s->sha1-hash`. You should move `digestable/unordered-sha1s->sha1-hash` invocations so it's called after `precluding-errors`.

vlaaad (changes requested):
I observed 18% performance improvements when opening a bunch of atlases on macOS (3.9s on dev, 3.3s on this branch). It's good, but the improvement is even bigger on Linux, right?

The code changes look good to me, though I find a small bug we should fix before merging.

JosephFerano:
> I observed 18% performance improvements when opening a bunch of atlases on macOS (3.9s on dev, 3.3s on this branch). It's good, but the improvement is even bigger on Linux, right?

Yes, I can run a before/after on this project if you want. The TL;DR is that 1.) clang was able to auto-vectorize the C++ codepaths for armv8 but not for x64, per godbolt. 2.) malloc is slow for large buffers; for a 16k * 8k RGBA buffer it took 200+ ms, where as it seems like macOS's architecture is more performant. So I focused on removing extra allocations and merging C++ code that were likely already fast on macOS

editor/src/clj/editor/atlas.clj
```diff
@@ -602,15 +602,15 @@
...
                             (resource-io/with-error-translation resource _node-id nil
                               (resource/resource->path-inclusive-sha1 resource)))
-                          flat-image-resources)
-        images-sha1 (digestable/unordered-sha1s->sha1-hash image-sha1s)]
+                          flat-image-resources)]
     (g/precluding-errors image-sha1s
-      (let [packed-image-sha1 (digestable/sha1-hash
+      ;; Note: Dragging or reordering images up and down the outline for animations results in a different order,
+      ;; which produces a different hash, which triggers an unnecessary atlas regeneration, so hash unordered
+      (let [images-sha1 (digestable/sha1s->unordered-sha1-hex image-sha1s)
```
vlaaad, line 612: By the way, are images deduplicated, or is it possible they are repeated multiple times here? Asking because the XOR of an item repeated an even number of times cancels itself out...

vlaaad (dismissed):
LGTM! I only have a concern about using xor instead of a sum for unordered hashing, though I'm not sure how important it is...

vlaaad:
Maybe at least the `sha1s->unordered-sha1-hex` should mention that it's sensitive to duplicates...

engine/texc/src/texc_private.cpp
```diff
@@ -336,6 +336,32 @@ namespace dmTexc
         }
     }
 
+    // NOTE: This is used by the Defold Editor when Generating a Preview Image for atlases
+    // We merge all the required steps to optimize since this blocks the editor UI while this is generated
+    void ConvertPremultiplyAndFlip_ABGR8888ToRGBA8888(const uint8_t* input_data, uint8_t* output_data, const uint32_t width, const uint32_t height)
```
JCash, line 341: Code style: we have a .clang-format in the repo.
E.g. we don't use "const" on numbers ("const uint32_t")

editor/src/clj/editor/pipeline/texture_set_gen.clj
```diff
@@ -153,7 +153,10 @@
 
 (defn atlas->texture-set-data
   [animations images margin inner-padding extrude-borders max-page-size]
-  (let [sprite-geometries (mapv make-image-sprite-geometry images)]
+  ;; NOTE: Images order matters when generating the layouts, especially if they are of the same size.
+  ;; Since the SHA1 for the packed-page-images-generator is insensitive to order, things could get out of sync
+  (let [images (sort-by #(-> % :path :project-path) images)
```
matgis: The `:path` seems to be a `Resource`? In that case, we should use `resource/resource->proj-path` (returns `""` for `nil`) instead of accessing the `:project-path` field of the resource directly. A `ZipResource` does not even have this field, so we must use the methods on the `Resource` protocol whenever we want information from a `Resource`.

matgis (changes requested):

editor/src/clj/editor/pipeline/texture_set_gen.clj
JosephFerano: Yes, you are right. It didn't have the type annotation when I was inspecting the data, so it looked like a regular old map.

I looked around to see if it was safe to assume it'll always be a resource, there is an assert in `make-image-sprite-geometry` and in the `AtlasImage`, the `image-resource` output validates that it is a Resource. So hopefully no checks are need before calling `resource/resource->proj-path`. Tested and it works.

matgis (dismissed):
Ship it! ❤️

JCash (dismissed):

editor/src/clj/editor/pipeline/texture_set_gen.clj
```diff
@@ -153,7 +153,10 @@
 
 (defn atlas->texture-set-data
   [animations images margin inner-padding extrude-borders max-page-size]
-  (let [sprite-geometries (mapv make-image-sprite-geometry images)]
+  ;; NOTE: Images order matters when generating the layouts, especially if they are of the same size.
+  ;; Since the SHA1 for the packed-page-images-generator is insensitive to order, things could get out of sync
+  (let [images (sort-by #(-> % :path resource/proj-path) images)
```
vlaaad, line 158: Should we do it after `g/precluding-errors`? I'm not sure if `images` may be an error value here.

editor/src/clj/util/digestable.clj
```diff
@@ -248,3 +249,28 @@
...
+              (fn [acc ba]
+                (-> (BigInteger. 1 ^bytes ba)
+                    (.add acc)
+                    (.mod modulus)))
+              BigInteger/ZERO
+              byte-arrays)
+        result-bytes ^bytes (.toByteArray ^BigInteger sum)
+        ;; NOTE: BigInteger.toByteArray() returns variable length:
+        ;; 21 bytes if high bit set (adds sign byte), <20 if leading zeros, else 20
+        normalized (let [len (alength result-bytes)]
```
vlaaad: No need to introduce a let here for `len` — we can reuse the let above.

editor/src/clj/editor/pipeline/texture_set_gen.clj
vlaaad: Disregard this, I wasn't thinking straight here...

vlaaad (approved):
LGTM!

PR #11623, opened 2025-12-16
SDK build speedup

AGulev:
The previous implementation extracted every platform SDK zip into the same temp directory, repeatedly overwriting shared files (headers, protos, configs, etc.). This is I/O heavy and dominated by extraction time, making SDK builds slow. The new manifest-based merge extracts each unique file once and avoids redundant work.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/gradle/wrapper/gradle-wrapper.properties
+++ b/com.dynamo.cr/com.dynamo.cr.bob/gradle/wrapper/gradle-wrapper.properties
@@ -1,7 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 distributionUrl=https\://services.gradle.org/distributions/gradle-9.1.0-bin.zip
-networkTimeout=10000
+networkTimeout=20000
 validateDistributionUrl=true
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -1973,20 +1973,37 @@ class Configuration(object):
 
         platform_zips = []  # (platform, zip_path)
         try:
-            for platform in platforms:
-                prefix = os.path.join(base_prefix, 'engine', platform, 'defoldsdk.zip')
-                entry = bucket.Object(prefix)
+            DOWNLOAD_WORKERS = 4
+            download_pool = ThreadPool(DOWNLOAD_WORKERS)
 
-                platform_sdk_zip = tempfile.NamedTemporaryFile(delete = False)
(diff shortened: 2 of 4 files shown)


com.dynamo.cr/com.dynamo.cr.bob/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 distributionUrl=https\://services.gradle.org/distributions/gradle-9.1.0-bin.zip
-networkTimeout=10000
+networkTimeout=20000
```
AGulev, line 4: sometimes build fails because of timeout: https://github.com/defold/defold/actions/runs/20275338456/job/58222288081

scripts/build.py
```diff
@@ -1943,20 +1943,169 @@ def build_sdk(self):
...
 
-        for platform in platforms:
-            prefix = os.path.join(base_prefix, 'engine', platform, 'defoldsdk.zip')
-            entry = bucket.Object(prefix)
-
-            platform_sdk_zip = tempfile.NamedTemporaryFile(delete = False)
-            print ("Downloading", entry.key)
-            entry.download_file(platform_sdk_zip.name)
-            print ("Downloaded", entry.key, "to", platform_sdk_zip.name)
+        # We merge platform SDK zips by building a per-zip index and selecting a single source
```
AGulev: New approach decide in advance what should be extracted instead of extracting all the archives + parallel downloading

with this approach build takes 4m 
CleanShot 2025-12-16 at 18 30 15@2x instead of 8 and sometimes 10+ and timeout

JCash (dismissed):
Only comment I have, is to perhaps move this into a separate python file, as this functionality is quite large now. This would clean up build.py a little.

scripts/build.py
```diff
@@ -1328,9 +1329,10 @@ def _findfiles(directory, exts):
 
                 self._add_files_to_zip(zip, protobuf_files, self.dynamo_home, topfolder)
 
-                # bob pipeline classes
-                bob_light = os.path.join(self.dynamo_home, 'share/java/bob-light.jar')
-                self._add_files_to_zip(zip, [bob_light], self.dynamo_home, topfolder)
+                # bob pipeline classes include only in one sdk
+                if platform in ('x86_64-linux'):
+                    bob_light = os.path.join(self.dynamo_home, 'share/java/bob-light.jar')
+                    self._add_files_to_zip(zip, [bob_light], self.dynamo_home, topfolder)
```
AGulev, line 1335: include `bob-light.jar` only into `x86_64-linux` sdk

JCash (approved):

PR #11630, opened 2025-12-18
Decrease editor memory pressure from transaction steps

matgis:
Refactored the data structures we use to represent atomic changes to the editor graph. This yields a small improvement to peak memory usage when loading the project.

For comparison, in our large test project, the transaction steps needed to load the project went from `1.02 GB` to `695 MB`. In practice, the improvement will not be directly reflected as a corresponding decrease in peak memory usage, but the change puts less pressure on the garbage collector and enables further optimizations.

### Technical changes
* We now use plain Java objects to represent transaction steps. A `TransactionStep` interface was introduced to cover the functionality we need from all transaction steps.
* New functions have been added to the `dynamo.graph` module for querying data from a sequence of transaction steps:
  * `g/tx-data-step-types`
  * `g/tx-data-added-arcs`
  * `g/tx-data-added-nodes`
  * `g/tx-data-added-node-ids` (was renamed from `g/tx-data-nodes-added`)
* Code that used to access the transaction step data structures directly have been updated to use the new `g/tx-data-*` functions.
* Some previously deprecated arities of the `g/set-properties` functions were removed.
* The measurements in `load_project.clj` have been updated to include total heap usage and a total time measurement without GC overhead.

Counterpart PRs for the Editor API changes:
* defold/extension-rive#231
* defold/extension-spine#257


editor/src/clj/editor/collection.clj
```diff
@@ -852,11 +853,23 @@
...
+          child->parent (-> {}
+                            (into (map #(pair (key %) nil))
+                                  id->nid)
+                            (into
+                              (comp cat
+                                    (mapcat (fn [{:keys [children id]}]
+                                              (map #(pair % id)
+                                                   children))))
+                              (pair (:instances collection)
+                                    (:embedded-instances collection))))]
```
matgis, line 872: Use `(g/tx-data-added-nodes)`, transducers and split over multiple lines.

editor/src/clj/editor/curve_view.clj
```diff
@@ -489,7 +490,7 @@
   (property list ListView)
   (property hidden-curves g/Any)
   (property input-action-queue g/Any (default []))
-  (property updatable-states g/Any (default (atom {})))
+  (property updatable-states g/Any)
```
matgis, line 493: Unrelated fix: If an `atom` is used in a `(default)` clause, all instances of the Node will use the same `atom`. In this case there will only be one `CurveViewNode`, so it didn't matter.

editor/src/clj/editor/gui.clj
```diff
@@ -4625,7 +4619,7 @@
                 :custom-type custom-type
                 :type node-type)]
     (concat
-      (apply g/set-property child-node-id (coll/mapcat identity props))
+      (apply g/set-properties child-node-id (coll/mapcat identity props))
```
matgis, line 4622: I've removed the deprecated arities of the `g/set-properties` functions that previously accepted multiple kv-pairs.

editor/src/clj/editor/outline.clj
```diff
@@ -109,13 +111,12 @@
...
-                                                             tgt-serial-id (node-id->serial-id (:target-id tx-step))]
-                                                         (when (and src-serial-id tgt-serial-id)
-                                                           [src-serial-id (:source-label tx-step) tgt-serial-id (:target-label tx-step)]))))
-                                                   (flatten tx-data)))))))
+                                             (map (fn [arc]
+                                                    (let [src-serial-id (node-id->serial-id (gt/source-id arc))
+                                                          tgt-serial-id (node-id->serial-id (gt/target-id arc))]
+                                                      (when (and src-serial-id tgt-serial-id)
+                                                        [src-serial-id (gt/source-label arc) tgt-serial-id (gt/target-label arc)])))
+                                                  (g/tx-data-added-arcs tx-data)))))))
```
matgis: Use `g/tx-data-added-arcs` instead of the internal transaction step fields.

editor/src/clj/editor/workspace.clj
```diff
@@ -933,11 +933,11 @@ ordinary paths."
...
   (property dependencies Dependencies)
-  (property opened-files g/Any (default (atom #{})))
-  (property resource-snapshot g/Any)
-  (property resource-listeners g/Any (default (atom [])))
+  (property opened-files g/Any)
+  (property resource-snapshot g/Any (default resource-watch/empty-snapshot))
+  (property resource-listeners g/Any)
   (property disk-sha256s-by-node-id g/Any (default {}))
-  (property view-types g/Any)
+  (property view-types g/Any (default {:default {:id :default}}))
```
matgis, line 940: Unrelated fix: Moved `atoms` out of `(default)` clauses, and some constants into `(default)` clauses.

editor/src/clj/internal/transaction.clj
```diff
@@ -448,14 +285,10 @@
...
-      (apply-tx ctx' (init-fn (in/custom-evaluation-context {:basis (:basis ctx') :tx-data-context (:tx-data-context ctx')})
-                              original-node-id->override-node-id)))))
-
-(defmethod metrics-key :override
-  [{:keys [root-id]}]
-  root-id)
+          (apply-tx ctx' (concat new-override-nodes-tx-data
+                                 new-override-tx-data))
+          (apply-tx ctx' (init-fn (in/custom-evaluation-context {:basis (:basis ctx') :tx-data-context (:tx-data-context ctx')})
+                                  original-node-id->override-node-id)))))
```
matgis, line 291: Indentation fix.

editor/src/clj/internal/transaction.clj
```diff
@@ -734,9 +551,21 @@
...
+      ctx
+      (let [is-override-node (some? (gt/original node))
+            is-dynamic (not (contains? (in/all-properties (gt/node-type node)) property-label))
+
+            ;; Use a custom evaluation-context since we're inside a transaction
+            ;; and cannot use the cache.
+            old-value (let [tx-data-context (:tx-data-context ctx)
+                            evaluation-context (in/custom-evaluation-context {:basis basis :tx-data-context tx-data-context})]
+                        (in/node-property-value node property-label evaluation-context))]
+        (invoke-setter ctx node-id node property-label old-value new-value is-override-node is-dynamic)))))
```
matgis, line 568: `ctx-set-property` now has its own distinct code path alongside `ctx-update-property`. It does effectively the same thing, but I've been experimenting with not producing the `old-value` in some situations where it would be safe for a slight load-time improvement. That's a story for another time, though!

editor/src/dev/load_project.clj
```diff
@@ -211,11 +211,13 @@
...
-              (it/apply-tx-label transaction-context)
-              (it/finalize-update transaction-context))
+              (when transaction-context
+                (it/trace-dependencies transaction-context)
+                (it/apply-tx-label transaction-context)
+                (it/finalize-update transaction-context)))
 
-        _ (g/commit-tx-result! tx-result transact-opts)
+        _ (when tx-result
+            (g/commit-tx-result! tx-result transact-opts))
```
matgis, line 220: These were wrapped in `when` to make stuff work OK when the `final-task` is defined to something earlier than the input they're expecting.

editor/src/clj/editor/outline.clj
```diff
@@ -109,13 +111,12 @@
...
-                                                       (let [src-serial-id (node-id->serial-id (:source-id tx-step))
-                                                             tgt-serial-id (node-id->serial-id (:target-id tx-step))]
-                                                         (when (and src-serial-id tgt-serial-id)
-                                                           [src-serial-id (:source-label tx-step) tgt-serial-id (:target-label tx-step)]))))
-                                                   (flatten tx-data)))))))
+                                             (map (fn [arc]
+                                                    (let [src-serial-id (node-id->serial-id (gt/source-id arc))
+                                                          tgt-serial-id (node-id->serial-id (gt/target-id arc))]
+                                                      (when (and src-serial-id tgt-serial-id)
+                                                        [src-serial-id (gt/source-label arc) tgt-serial-id (gt/target-label arc)])))
```
vlaaad: why use map instead of keep if fn can still return nil?

editor/src/clj/editor/tile_source.clj
```diff
@@ -1007,14 +1007,8 @@
...
-  (let [collision-group->node-id
-        (into {}
-              (keep (fn [tx]
-                      (let [{:keys [_node-id id]} (:node tx)]
-                        (when id
-                          (pair id _node-id)))))
-              collision-group-nodes-tx-data)]
-
+  (let [collision-group-nodes (g/tx-data-added-nodes collision-group-nodes-tx-data)
+        collision-group->node-id (coll/pair-map-by :id g/node-id collision-group-nodes)]
```
vlaaad, line 1011: The previous code checked id for nil, the new one doesn't; intentional?

editor/src/clj/internal/transaction.clj
```diff
@@ -23,17 +23,17 @@
...
 
-;; ---------------------------------------------------------------------------
-;; Configuration parameters
-;; ---------------------------------------------------------------------------
-(def maximum-retrigger-count 100)
-(def maximum-graph-coloring-recursion 1000)
+(defonce/interface TransactionStep
+  (step_type [] "Returns a keyword uniquely identifying the type of transaction step.")
+  (metrics_key [] "Returns a key which identifies the subject of the transaction step in metrics reports.")
+  (perform [ctx] "Returns a new ctx with changes applied."))
```
vlaaad: This is not a valid Clojure code. It compiles for sure, but `definterface` doesn't support defining docstrings because docstrings are for Clojure Vars, while definterface defines a Java interface with methods.

editor/src/clj/dynamo/graph.clj
```diff
@@ -237,15 +238,35 @@
...
+(defn tx-data-step-types
+  "Given a sequence of transaction steps, returns a sequence of keywords
+  identifying the type of each pending transaction step. Used in tests."
+  [txs]
+  (map it/tx-step-type
+       (flatten txs)))
+
+(defn tx-data-added-arcs
+  "Given a sequence of transaction steps, returns a sequence of Arcs that will
+  be added by the transaction."
```
vlaaad: I think this docstring is confusing because we have an `expand-ec` transaction step that creates new — any! —transaction steps during application. So, if we have a transaction steps seq with `expand-ec`, it might create more arcs. This situation sounds like a caveat worth mentioning (both here and in dosctrings below)

editor/src/clj/dynamo/graph.clj
```diff
@@ -712,14 +734,8 @@
...
-  ([node-id property-label value]
-   (it/set-property node-id property-label value nil))
-  ([node-id property-label value opts]
-   (it/set-property node-id property-label value opts))
-  ([node-id property-label value another-property-label another-value & more]
-   ;; Deprecated. To set multiple properties at once, use set-properties.
-   ;; Kept for compatibility with editor plugins until we can update them.
-   (apply set-properties node-id property-label value another-property-label another-value more)))
+  [node-id property-label value]
+  (it/set-property node-id property-label value))
```
vlaaad, line 744: Example in the docstring is now incorrect (uses no longer existing arity)

editor/src/clj/dynamo/graph.clj
```diff
@@ -237,15 +238,35 @@
...
 ;; ---------------------------------------------------------------------------
-(defn tx-data-nodes-added
-  "Returns a list of the node-ids added given a list of transaction steps, (tx-data)."
+
+(defn tx-data-step-types
+  "Given a sequence of transaction steps, returns a sequence of keywords
+  identifying the type of each pending transaction step. Used in tests."
+  [txs]
+  (map it/tx-step-type
+       (flatten txs)))
```
vlaaad: Hmm, `flatten` was there before, but it's actually concerning. Transactions explicitly support nils (`apply-tx` skips them), but `flatten` preserves nils. Maybe use `coll/flatten-xf` and transducers for better semantics (`flatten-xf` skips nils) and performance?

vlaaad (changes requested):
We should fix the `definterface` call so that it's a valid Clojure (it currently compiles, but it's not a valid Clojure since `definterface` doesn't support docstrings). Also I left some questions... Otherwise the changes look good!

We didn't allocate 1GB for tx steps before, right? Because transaction steps are typically lazy, and tx application doesn't keep a reference to the head of the collection. My understanding is that this change mostly affects GC pressure (i.e. less garbage to collect while the tx is applied). Did you see any reductions in load times?

editor/src/clj/editor/outline.clj
matgis: Whoops! Yep, this still needs to use `keep`. Fixing.

editor/src/clj/editor/tile_source.clj
matgis: Yes. The previous code was operating on a mix of `:create-node` and other transaction steps, whereas now `collision-group-nodes` will be a sequence of `NodeImpl` records. I think the filtering in the previous code served to only include `:create-node` steps for consideration in a somewhat brittle way.

editor/src/clj/internal/transaction.clj
matgis: Huh. You're right. This used to be a protocol before I converted it into an interface. I'll move the "docstrings" to comments.

editor/src/clj/dynamo/graph.clj
matgis: Good point. I'll update the comment so it is clear it only considers specific transaction step types for this.

editor/src/clj/dynamo/graph.clj
matgis: Hmm, yes. I don't think the use of `flatten` was a problem before, since the only function that existed prior to this would filter out the `nils` in the `keep` clause. However, it can be a problem here where we use `map`. Nice catch!

editor/src/clj/dynamo/graph.clj
matgis: Ah, I missed this. Fixed!

matgis (commented):
Added two missing responses to review comments already addressed in last commit.

vlaaad (approved):
LGTM!

PR #11642, opened 2025-12-20
Add a window title to splash, welcome, and project loading

### LINKED ISSUE #11607: Add window title to the recent projects window
When the recent projects window spawns it has no title. This is a problem for (Linux) Hyprland users when comes to window rules.

The recent projects window should have an actual window title.

.

`hyprctl clients` relevant output of the recent projects window:
```
	class: com.defold.editor.Start
	title: 
	initialClass: com.defold.editor.Start
	initialTitle: 
```

JosephFerano:
Defold now has an OS window title while starting up, showing the welcome screen, and during the project loading window.

Fixes #11607

### Technical changes

While the user reported this as a problem on Linux, this should affect all operating systems.

### DIFF
--- a/editor/src/java/com/defold/editor/Splash.java
+++ b/editor/src/java/com/defold/editor/Splash.java
@@ -130,7 +130,11 @@ public class Splash {
         String channelName = System.getProperty("defold.channel");
         String sha1 = System.getProperty("defold.editor.sha1");
 
-        versionString = (versionString == null || versionString.isEmpty()) ? "No version" : versionString;
+        boolean hasVersion = true;
+        if (versionString == null || versionString.isEmpty()) {
+            versionString = "No version";
+            hasVersion = false;
+        }
         channelName = (channelName == null || channelName.isEmpty()) ? "No channel" : channelName;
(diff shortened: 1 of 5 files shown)


editor/src/java/com/defold/editor/Splash.java
```diff
@@ -142,6 +142,11 @@ public void show() throws IOException {
         channelLabel.setText(versionString + " (" + sha1 + ")" + channelName);
         channelLabel.setVisible(true);
 
+        String windowTitle = "Defold";
+        if (!versionString.equals("No version")) {
```
matgis: This seems a bit brittle. Maybe it would be better to add a `hasVersion` `bool` right above where we assign the `versionString` and check that here as well?

matgis (dismissed):
It works!

editor/src/java/com/defold/editor/Splash.java
JosephFerano: Sure, good point, I introduced a bool

matgis (approved):

PR #11645, opened 2025-12-20
Asset Browser: Group by resources for "New..." menu

### LINKED ISSUE #10761: Organize items within the 'New' context menu in the asset pane
A long, unorganized list in the context menu makes it hard to find options quickly. The items in the context menu are already color-coded, but the list is still unorganized. This makes it hard to find options quickly, and users often have to scroll through the long list, which can be hard when the menu isn’t fully visible.

While the context menu items are already color-coded, they are not grouped together by similar functionality. Grouping these related items more explicitly would improve clarity and navigation. Additional horizontal separators could be used to visually divide the groups, making the menu easier to scan and use.

I’m not a fan of nested context menus, so having a larger window with all options visible at once seems like a better choice. I've added some missing icons to the context menu, but I haven't yet come up with suitable names for grouping the items as I'm still learning the Defold philosophy.

Image
Image

There are no alternatives, as this is a core feature of the editor.

JosephFerano:
The "New..." right-click context menu in the Asset Browser was getting a bit tall, so now it is formatted as a grid with sub-categories that groups items together for more compact organization. Supports arbitrary group names and uncategorized resources.

image

Fixes #11597
Fixes #10761


editor/src/clj/editor/asset_browser.clj
```diff
@@ -511,10 +511,11 @@
                        (localization/message "command.file.new")
                        (let [rt (:resource-type user-data)]
                          (or (:label rt) (:ext rt)))))
-  (active? [selection selection-context] (or (= :global selection-context) (and (= :asset-browser selection-context)
-                                                                                (= (count selection) 1)
-                                                                                (not= nil (some-> (handler/adapt-single selection resource/Resource)
-                                                                                            resource/abs-path)))))
+  (active? [selection selection-context] (or (= :global selection-context)
```
JosephFerano: Formatting

editor/src/clj/editor/asset_browser.clj
```diff
@@ -547,22 +548,31 @@
...
+                          ["resource.category.resources"]
+                          ["resource.category.editor" "resource.category.project_settings" "resource.category.other"]]
+            predefined-categories (into #{} cat base-columns)
+            all-items (keep (fn [[_ext resource-type]]
+                              (when (workspace/has-template? workspace resource-type evaluation-context)
+                                {:label (or (:label resource-type) (:ext resource-type))
+                                 :icon (:icon resource-type)
+                                 :category (or (some-> resource-type :category :k)
+                                               (:category resource-type)
+                                               "resource.category.other")
```
JosephFerano: This is kinda funky, I'm extracting the string key, basically undoing the localization message, both for the grouping here, but also because in the new `ui/make-grid-menu` function, I'm using `internal.util/group-into` which requires MessagePattern to implement java.lang.Comparable. So not sure what the best approach here is. Should categories even be an instance of MessagePattern? Implement Comparable?

editor/src/clj/editor/ui.clj
```diff
@@ -1470,6 +1478,82 @@
     (user-data! menu-item ::menu-user-data user-data)
     menu-item))
 
+(defn- make-grid-menu
```
JosephFerano, line 1550: I attempted to make this function at least a little generalized, so that other context menus could use this sort of multi-column layout. More work can be done to support more scenarios (like category header vs no header). It does at least supports catching uncategorized children, which go into the "Other" group.

editor/src/clj/editor/asset_browser.clj
```diff
@@ -547,22 +548,31 @@
...
+                                 :style (resource/type-style-classes resource-type)
+                                 :command :file.new
+                                 :user-data {:resource-type resource-type}}))
+                            (resource/resource-types-by-type-ext (:basis evaluation-context) workspace :editable))
+            unlisted-categories (vec (remove predefined-categories (into #{} (map :category all-items))))
```
JosephFerano: This all catches resources with categories that are not part of the base ones. If an extension registers a resource with an unlisted category, it'll show up in a new 5th column. It could also just be added to the 4th column (last one) instead of creating a new column, let me know if y'all have a stronger preference for having more or less columns.

editor/src/clj/editor/ui.clj
```diff
@@ -1470,6 +1478,82 @@
...
+                    group or a localization key as a string
+
+  Returns:
+    A JavaFX custom menu item containing the grid layout"
+  [^Scene scene localization grid-config command-contexts evaluation-context]
+  (let [{:keys [columns children]} grid-config
+        items-by-category (-> (util/group-into {} []
+                                #(or (:category %)
+                                     "resource.category.other")
+                                children)
```
JosephFerano: This is expecting the categories to be strings, even if they are localization keys. As mentioned, previous, this is because of needing to implement java.lang.Comparable. Is there a better approach?

editor/src/clj/editor/ui.clj
```diff
@@ -1470,6 +1478,82 @@
...
+               :spacing 2.0
+               :children
+               (mapcat
+                 (fn [category-key]
+                   (when-let [category-items (seq (get items-by-category category-key))]
+                     (concat
+                       [{:fx/type fx.h-box/lifecycle
+                         :alignment :center-left
+                         :children [{:fx/type fx.label/lifecycle
+                                     :text (localization (localization/message category-key {} category-key))
```
JosephFerano: Support both strings and localization keys for category names

editor/src/clj/editor/asset_browser.clj
JosephFerano: This was removed to avoid conflict with Mats' evaluation-context PR.

editor/src/clj/editor/asset_browser.clj
```diff
@@ -547,22 +547,31 @@
...
+            columns (cond-> base-columns
+                      (seq unlisted-categories)
+                      (conj (vec (sort unlisted-categories))))]
+        [{:layout :grid
+          :columns columns
+          :children all-items}]))))
```
vlaaad: Options are not **only** used for context menus. They are also used to populate the option list dialog, and the list dialog flattens the tree of options using arrows as separators. Since this root option item has no label, it's shown as an empty string in the dialog:
Screenshot 2026-01-12 at 10 10 23
The dialog should probably show the list of options without any `->` separators... Perhaps the data should be structured in some other way, so that this fn returns `all-items` with layout/columns metadata?

editor/src/clj/editor/asset_browser.clj
```diff
@@ -547,22 +547,31 @@
...
-                        {:label (or (:label resource-type) (:ext resource-type))
-                         :icon (:icon resource-type)
-                         :style (resource/type-style-classes resource-type)
-                         :command :file.new
-                         :user-data {:resource-type resource-type}})))
-              (resource/resource-types-by-type-ext (:basis evaluation-context) workspace :editable))))))
+      (let [base-columns [["resource.category.objects" "resource.category.scripts" "resource.category.shaders"]
+                          ["resource.category.components"]
+                          ["resource.category.resources"]
+                          ["resource.category.editor" "resource.category.project_settings" "resource.category.other"]]
```
vlaaad: Shouldn't columns be localization messages?

editor/src/clj/editor/asset_browser.clj
vlaaad: Why do they need to implement comparable? I think we should allow extensions that register resource types to define categories as simple strings, and they should not be interpreted as localization message keys.

editor/src/clj/editor/asset_browser.clj
vlaaad: Please avoid lazy sequences in favor of transducers. Here, `into #{}` can probably be replaced with a `distinct` transducer.

editor/src/clj/editor/ui.clj
vlaaad: It doesn't need to be Comparable; it works fine with a mix of string and MessagePattern instances:
```clj
(util/group-into
  {}
  []
  :key
  [{:key "a"}
   {:key (localization/message "a")}])
=> {"a" [{:key "a"}]
    #editor.localization.SimpleMessage {:k "a"
                                        :m {}
                                        :fallback nil} [{:key #editor.localization.SimpleMessage {:k "a"
                                                                                                  :m {}
                                                                                                  :fallback nil}}]}
```

vlaaad (changes requested):
Works great, but we should work with MessagePatterns and strings properly, not trying to mix them into one thing. MessagePattern is localizable/translatable, string is literal and should not be used as a localization message key. Also, you should create PRs for our extensions that define new resource types, i.e. [spine](https://github.com/defold/extension-spine/tree/main/defold-spine/editor/src), [rive](https://github.com/defold/extension-rive/blob/main/defold-rive/editor/src/rive.clj#L865), and [texturepacker](https://github.com/defold/extension-texturepacker/blob/main/texturepacker/editor/src/texturepacker.clj#L1289).

editor/src/clj/editor/ui.clj
JosephFerano: I think I stumbled on a false positive. I'm not sure what I was doing now to trigger it, but I was getting an exception when attempting to use group-by and group-into saying that Comparable wasn't implemented. Now I can't replicate the exception :/ At least it works now and I make this better!

JosephFerano:
@vlaaad I will go through all the extensions and update them once this PR is approved. Let me know if it's ok to merge first or if we should try to merge them all at the same time. This should handle uncategorized items gracefully at least.

editor/src/clj/editor/asset_browser.clj
```diff
@@ -549,29 +549,31 @@
...
+                              (resource/resource-types-by-type-ext (:basis evaluation-context) workspace :editable)))
+            unlisted-categories (vec (remove predefined-categories (distinct (map :category all-items))))
             columns (cond-> base-columns
                       (seq unlisted-categories)
                       (conj (vec (sort unlisted-categories))))]
-        [{:layout :grid
-          :columns columns
-          :children all-items}]))))
+        (with-meta
+          (localization/natural-sort-by-label localization all-items)
```
vlaaad: Type error that works by accident: `natural-sort-by-label` expects `localization-state`, not `localization`; you can obtain it from `localization` using `@`.

Side note: it would be nice to add `{:pre [(localization-state? localization-state)]}` to `natural-sort-by-label` to guard against such errors in the future.

editor/src/clj/editor/ui.clj
```diff
@@ -1470,6 +1478,84 @@
...
+
+  Returns:
+    A JavaFX custom menu item containing the grid layout"
+  [^Scene scene localization items command-contexts evaluation-context]
+  (let [columns (:columns (meta items))
+        items-by-category (-> (util/group-into {} []
+                                #(or (:category %)
+                                     (localization/message "resource.category.other"))
+                                items)
+                              (update-vals #(localization/natural-sort-by-label localization %)))]
```
vlaaad: Similar type error that works by accident here.

editor/src/clj/editor/asset_browser.clj
```diff
@@ -547,22 +547,33 @@
...
+                        (keep (fn [[_ext resource-type]]
+                                (when (workspace/has-template? workspace resource-type evaluation-context)
+                                  {:label (or (:label resource-type) (:ext resource-type))
+                                   :icon (:icon resource-type)
+                                   :category (or (:category resource-type)
+                                                 (localization/message "resource.category.other"))
+                                   :style (resource/type-style-classes resource-type)
+                                   :command :file.new
+                                   :user-data {:resource-type resource-type}}))
+                              (resource/resource-types-by-type-ext (:basis evaluation-context) workspace :editable)))
```
vlaaad: Please avoid lazy sequences in favor of transducers, e.g. using `into` or `coll/transfer`

editor/src/clj/editor/asset_browser.clj
```diff
@@ -547,22 +547,33 @@
...
+            unlisted-categories (vec (remove predefined-categories (distinct (map :category all-items))))
```
vlaaad: Please avoid lazy sequences in favor of transducers, e.g.:
```clj
(coll/transfer all-items []
  (map :category)
  (distinct)
  (remove predefined-categories))
```

editor/src/clj/editor/asset_browser.clj
```diff
@@ -547,22 +547,33 @@
...
+            columns (cond-> base-columns
+                      (seq unlisted-categories)
```
vlaaad: Please avoid using `seq` since it creates garbage, use e.g. `(not (coll/empty? unlisted-categories))`

editor/src/clj/editor/ui.clj
```diff
@@ -1470,6 +1478,84 @@
...
+             :min-width 1.0
+             :max-width 1.0
+             :style-class ["grid-menu-column-separator"]}
+            (for [column columns]
+              {:fx/type fx.v-box/lifecycle
+               :spacing 2.0
+               :children
+               (mapcat
+                 (fn [category-key]
+                   (when-let [category-items (seq (get items-by-category category-key))]
```
vlaaad: Please avoid using `seq` since it creates garbage. I may be mistaken, but since `items-by-category` is created using `util/group-into`, any existing key in this map is already guaranteed to point to a non-empty coll of items, so just `(get items-by-category category-key)` may be enough?

editor/src/clj/editor/ui.clj
```diff
@@ -1470,6 +1478,84 @@
...
+                     (concat
+                       [{:fx/type fx.h-box/lifecycle
+                         :alignment :center-left
+                         :children [{:fx/type fx.label/lifecycle
+                                     :text (if (string? category-key)
+                                             category-key
+                                             (localization category-key))
```
vlaaad: BTW, localization accepts any argument, not only MessagePattern instances, for easier support of mixing localized and unlocalized items, so `(localization category-key)` is semantically valid and will work fine.

vlaaad (changes requested):
Hmm... It works great! I feel like I'm mostly nitpicking now about lazy sequences; @matgis do you think it's worth it to ensure we don't introduce them in new code, even if they are most probably not going to be in hot loops? 

Regardless, I think we should fix the type error in the `natural-sort-by-label` use.

By the way, one behavior difference I noticed is that in the old component list, I could use keyboard arrows + <kbd>Enter</kbd> to select an item; now it's no longer possible. Can we add keyboard support here, e.g. using focus on grid items?

editor/src/clj/editor/ui.clj
```diff
@@ -1470,6 +1478,84 @@
...
+                              (update-vals #(localization/natural-sort-by-label localization %)))]
+    (fx/instance
+      (fx/create-component
+        {:fx/type fx.custom-menu-item/lifecycle
+         :hide-on-click false
+         :style-class ["grid-menu"]
+         :content
+         {:fx/type fx.h-box/lifecycle
+          :spacing 10.0
+          :padding (Insets. 5.0)
```
matgis: Nitpick: Padding can be specified directly as a number. It will be coerced into an `Insets` instance automatically.

editor/src/clj/editor/asset_browser.clj
```diff
@@ -547,22 +547,33 @@
             (when (resource/loaded? resource)
               (app-view/open-resource app-view prefs localization workspace project resource))
             (select-resource! asset-browser resource))))))
-  (options [workspace selection user-data evaluation-context]
+  (options [workspace selection user-data localization evaluation-context]
```
matgis: `selection` appears unused. Remove?

editor/src/clj/editor/asset_browser.clj
```diff
@@ -547,22 +547,33 @@
...
+                                   :category (or (:category resource-type)
+                                                 (localization/message "resource.category.other"))
+                                   :style (resource/type-style-classes resource-type)
+                                   :command :file.new
+                                   :user-data {:resource-type resource-type}}))
+                              (resource/resource-types-by-type-ext (:basis evaluation-context) workspace :editable)))
+            unlisted-categories (vec (remove predefined-categories (distinct (map :category all-items))))
+            columns (cond-> base-columns
+                      (seq unlisted-categories)
+                      (conj (vec (sort unlisted-categories))))]
```
matgis: Do we need to use localization-aware sorting here?

editor/src/clj/editor/ui.clj
```diff
@@ -1470,6 +1478,84 @@
...
+                                 {:fx/type fx.button/lifecycle
+                                  :text (localization label)
+                                  :disable (not enabled?)
+                                  :on-action (fn [_] (invoke-handler (contexts scene) command user-data))
+                                  :style-class (into ["grid-menu-item"] child-style)
+                                  :graphic {:fx/type fx.image-view/lifecycle
+                                            :image (icons/get-image child-icon 18)}}))))
+                         category-items)
+                       [{:fx/type fx.region/lifecycle
+                         :pref-height 20}])))
```
matgis: It would be nicer if this spacing was between the categories rather than simply added after each category. Right now there's too much space at the bottom of the resulting popup.

editor/styling/stylesheets/components/_context-grid-menu.scss
```diff
@@ -0,0 +1,39 @@
...
+
+    @include extensions(grid);
+}
+
+.grid-menu-group-label {
+    -fx-font-size: 11;
+    -fx-padding: 0 4 0 0;
+    -fx-font-style: italic;
+    -fx-text-fill: derive(-df-text-dark, 20%);
+}
```
matgis, line 21: I think the group headers might be a bit hard to read on non-retina displays. Here's how it looks when I run it through the xScope deretinalizer: 

Image

Perhaps we should just keep it at the default font size and remove the italics?

editor/src/clj/editor/asset_browser.clj
JosephFerano: Actually, now that I think about it, maybe we shouldn't be sorting here. Reason being is that Objects category is the nearest to the top left, and I think that's on purpose since these are some of the most used. In other languages, it might shuffle it to the bottom.What do you think?

editor/styling/stylesheets/components/_context-grid-menu.scss
JosephFerano: I removed the italics, and increased the font size to 12. Let me know if that improves it, I increase the font size further.

editor/src/clj/editor/ui.clj
JosephFerano: Used interleave with a region to add a small about of space, let me know if this is works better.

JosephFerano:
I added some styling in case the the resource is not enabled from the `handler/enabled?` check, I'm not sure when this would happen, but at least it'll dim the buttons.

editor/src/clj/editor/ui.clj
```diff
@@ -1470,6 +1478,90 @@
...
+                                    child-label (:label child)
+                                    child-icon (:icon child)
+                                    child-style (:style child)]
+                                (when-let [handler-ctx (handler/active command command-contexts user-data evaluation-context)]
+                                  (let [label (or (handler/label handler-ctx evaluation-context) child-label)
+                                        enabled? (handler/enabled? handler-ctx evaluation-context)]
+                                    {:fx/type fx.button/lifecycle
+                                     :text (localization label)
+                                     :disable (not enabled?)
+                                     :on-action (fn [_] (invoke-handler (contexts scene) command user-data))
```
vlaaad: On macOS, pressing <kbd>Enter</kbd> doesn't invoke the command. But this does:
```clj
:on-key-pressed (fn [^KeyEvent e]
                  (when (= KeyCode/ENTER (.getCode e))
                    (.consume e)
                    (invoke-handler (contexts scene) command user-data)))
```

editor/src/clj/editor/ui.clj
```diff
@@ -56,7 +64,7 @@
            [javafx.fxml FXMLLoader]
            [javafx.geometry Orientation Point2D]
            [javafx.scene Cursor Group Node Parent Scene]
-           [javafx.scene.control Button ButtonBase Cell CheckBox CheckMenuItem ChoiceBox ColorPicker ComboBox ComboBoxBase ContextMenu Control Label Labeled ListView Menu MenuBar MenuButton MenuItem MultipleSelectionModel ProgressBar SelectionMode SelectionModel Separator SeparatorMenuItem Tab TabPane TableView TextArea TextField TextInputControl Toggle ToggleButton Tooltip TreeItem TreeTableView TreeView]
+           [javafx.scene.control Button ButtonBase Cell CheckBox CheckMenuItem ChoiceBox ColorPicker ComboBox ComboBoxBase ContextMenu Control Label Label Labeled ListView Menu MenuBar MenuButton MenuItem MultipleSelectionModel ProgressBar SelectionMode SelectionModel Separator SeparatorMenuItem Tab TabPane TableView TextArea TextField TextInputControl Toggle ToggleButton Tooltip TreeItem TreeTableView TreeView]
```
vlaaad: `Label` is repeated :D

editor/src/clj/editor/asset_browser.clj
```diff
@@ -547,22 +547,35 @@
...
+                                   :category (or (:category resource-type)
+                                                 (localization/message "resource.category.other"))
+                                   :style (resource/type-style-classes resource-type)
+                                   :command :file.new
+                                   :user-data {:resource-type resource-type}}))))
+            unlisted-categories (coll/transfer all-items []
+                                  (map :category)
+                                  (distinct)
+                                  (remove predefined-categories))
+            columns (into base-columns unlisted-categories)]
```
vlaaad: I think there is a bug. `base-columns` is a vector of vectors of categories, i.e., `(MessagePattern|String)[][]`. `unlisted-categories` is a vector of categories, i.e., `(MessagePattern|String)[]`. Once we start getting unlisted categories, things will begin breaking, since suddenly you'll start getting `MessagePattern|String` where a `(MessagePattern|String)[]` is expected.

vlaaad (changes requested):
Found a couple of small issues; otherwise, it's solid!

editor/src/clj/editor/ui.clj
JosephFerano: Also works on Linux

editor/src/clj/editor/ui.clj
JosephFerano: Huh, that's odd. I ran the "clean namespace" thing with LSP

editor/src/clj/editor/asset_browser.clj
JosephFerano: Ah yes you're right, I made a mistake when removing the sorting. Here I'm assoc/dissoc the categories for testing

image

Works now

editor/styling/stylesheets/components/_context-grid-menu.scss
```diff
@@ -0,0 +1,33 @@
+.grid-menu-item-base {
+    -fx-padding: 4 10 4 10;
+    -fx-alignment: center-left;
+    -fx-max-width: infinity;
+    -fx-text-fill: -df-text;
+}
+
+.grid-menu-item-enabled {
+    &:focused, {
```
vlaaad: nitpick: comma shouldn't be here (though it's nice that it still works)

vlaaad (dismissed):
LGTM! Great work! 

Please remove a dangling comma from scss before merging (no need to re-request a review for that change)

matgis (changes requested):
I just noticed the icons in the popup menu appear very blurry compared to their counterparts on the icons in the Assets panel. I think part of it is that we appear to generate 18-pixel icons for this menu, whereas we use 16 pixels for the Asset icons. However, even after I change it to 16 pixels in `ui.clj`, the icons still appear blurry to me, so there must be something else going on as well. 🤔

JosephFerano:
> I just noticed the icons in the popup menu appear very blurry compared to their counterparts on the icons in the Assets panel. I think part of it is that we appear to generate 18-pixel icons for this menu, whereas we use 16 pixels for the Asset icons. However, even after I change it to 16 pixels in `ui.clj`, the icons still appear blurry to me, so there must be something else going on as well. 🤔

Good catch! I had to glue my face to my monitor see the difference but sure enough. I used an existing function to render the icon and now it seems to match what we see in the asset browser. Let me know if that looks better.

matgis (dismissed):
Looks great! Now the icons are nice and crisp. Ship it! ❤️

JosephFerano:
Sorry, I found a minor issue, let me see if I can fix it tomorrow.

editor/src/clj/editor/ui.clj
```diff
@@ -1481,42 +1489,130 @@
...
+          (if (or (nil? options)
+                  (and key-combo (not (:expand item))))
+            (make-menu-command scene id label localization icon style key-combo user-data command enabled? check)
+            (if (some-> options meta :layout (= :grid))
+              (let [grid-menu (make-grid-menu scene localization options command-contexts evaluation-context)]
+                (make-submenu id label localization icon style [grid-menu] #(focus-first-grid-menu-item grid-menu)))
+              (make-submenu id label localization icon style
+                            (make-menu-items scene (localization/sort-if-annotated @localization options)
+                                             command-contexts keymap localization evaluation-context)
+                            on-submenu-open))))))))
```
JosephFerano, line 1664: This is what I changed.

matgis (approved):
Works great! ❤️

PR #11653, opened 2025-12-22
Use injected evaluation context in editor handlers

matgis:
Fix a few places where we were not using the `evaluation-context` that is automatically injected into handlers.


vlaaad (dismissed):
Very nice!

PR #11655, opened 2025-12-22
Exclude insignificant entries when applying gui property overrides in the editor

matgis:
When applying property overrides to Gui nodes, filter out entries that have no significance. This significantly shrinks the list of affected nodes and properties that will need to be applied when loading, which saves around 82 MB of memory and a few seconds of load time in the large project we're testing with.

### DIFF
--- a/editor/src/clj/editor/gui.clj
+++ b/editor/src/clj/editor/gui.clj
@@ -1205,7 +1205,7 @@
 
   (input node-overrides g/Any :array)
   (output node-overrides g/Any :cached (g/fnk [node-overrides id _overridden-properties]
-                                         (into {id _overridden-properties}
+                                         (into {id (dissoc _overridden-properties :layout->prop->override)}
                                                node-overrides)))
   (output layout->prop->value g/Any :cached
           (g/fnk [^:unsafe _evaluation-context _this layout->prop->override trivial-gui-scene-info]
@@ -2281,8 +2281,8 @@
                                                       (cond-> (empty? (:parent %)) (assoc :parent id))))
                                             scene-node-msgs))))
   (output node-overrides g/Any :cached (g/fnk [id _overridden-properties template-overrides]
-                                              (-> {id _overridden-properties}
-                                                (merge template-overrides))))
+                                         (coll/merge {id (dissoc _overridden-properties :layout->prop->override)}
+                                                     template-overrides)))
   (output aabb g/Any (g/fnk [template-scene]
                        (if (some? template-scene)
                          (:aabb template-scene)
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/gui.clj
```diff
@@ -1205,7 +1205,7 @@
 
   (input node-overrides g/Any :array)
   (output node-overrides g/Any :cached (g/fnk [node-overrides id _overridden-properties]
-                                         (into {id _overridden-properties}
+                                         (into {id (dissoc _overridden-properties :layout->prop->override)}
```
matgis, line 1208: This does not need to be among the properties returned from the `node-overrides` output. Layout property overrides take a different path through the graph.

editor/src/clj/editor/gui.clj
```diff
@@ -3897,14 +3897,20 @@
                 {}
                 (:layouts scene))
 
+        with-layout-prop-overrides
+        (fn with-layout-prop-overrides [node-desc id]
+          (let [layout->prop->override (get node->layout->prop->override id)]
+            (cond-> node-desc
+                    (coll/not-empty layout->prop->override)
+                    (assoc :layout->prop->override layout->prop->override))))
+
```
matgis, line 3906: When loading the gui scene, we no longer include an empty map for each `:layout->prop->override` property when setting the properties.

editor/src/clj/editor/gui.clj
```diff
@@ -3918,21 +3924,26 @@
...
+                        (into {}
+                              (keep (fn [[id tmpl]]
+                                      (when-some [prop->override (coll/not-empty (:data tmpl))]
+                                        (let [imported-id (subs id (inc (count importing-id)))]
+                                          (pair imported-id prop->override)))))
+                              (rest (tree-seq fn/constantly-true
+                                              (comp tmpl-children first)
+                                              (pair importing-id nil))))]
+                    (when (pos? (count imported-id->prop->override))
+                      (pair importing-id imported-id->prop->override))))))
```
matgis, line 3946: We no longer include `[imported-id prop->override]` pairs when the resulting `prop->override` map is empty when loading the gui scene.

editor/src/clj/editor/gui.clj
```diff
@@ -3918,21 +3924,26 @@
...
+          (keep (fn [importing-id]
+                  (let [imported-id->prop->override
+                        (into {}
+                              (keep (fn [[id tmpl]]
+                                      (when-some [prop->override (coll/not-empty (:data tmpl))]
+                                        (let [imported-id (subs id (inc (count importing-id)))]
+                                          (pair imported-id prop->override)))))
+                              (rest (tree-seq fn/constantly-true
+                                              (comp tmpl-children first)
+                                              (pair importing-id nil))))]
```
vlaaad, line 3944: We have `coll/tree-xf` BTW.

vlaaad (approved):
LGTM!

PR #11657, opened 2025-12-22
CI stability improvements: turn off ccache + stop gradle daemon

AGulev:
Stop gradle daemon so it can be cached
Turn off ccache on CI sinse there is no benefit out of it and it fixes https://github.com/defold/defold/actions/runs/20430930119/job/58701234892?pr=11653

### DIFF
--- a/.github/workflows/main-ci.yml
+++ b/.github/workflows/main-ci.yml
@@ -61,6 +61,7 @@ jobs:
         shell: bash,
         run: 'ci/ci.sh --platform=${{ matrix.platform }} --archive --skip-builtins --skip-docs engine'
       },
+      { name: 'Stop Gradle Daemon', run: '.\\com.dynamo.cr\\com.dynamo.cr.bob\\gradlew.bat --stop' },
       {
         name: 'Notify if build status changed',
         uses: homoluctus/slatify@cd4b4a1158cfb3e26fe1ee35c1cd4f0247dfbf96,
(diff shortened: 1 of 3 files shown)


ci/ci.py
```diff
@@ -237,6 +237,8 @@ def build_engine(platform, channel, with_valgrind = False, with_asan = False, wi
     waf_opts = []
 
     opts.append('--platform=%s' % platform)
+    # ccache isn't needed on CI
+    opts.append('--disable-ccache')
```
AGulev, line 241: main fix

build_tools/waf_dynamo.py
```diff
@@ -1730,7 +1730,8 @@ def detect(conf):
     if Options.options.with_valgrind:
         conf.find_program('valgrind', var='VALGRIND', mandatory = False)
 
-    conf.find_program('ccache', var='CCACHE', mandatory = False)
+    if not Options.options.disable_ccache:
+        conf.find_program('ccache', var='CCACHE', mandatory = False)
```
AGulev, line 1734: we don't need to search for ccache if it's off

.github/workflows/main-ci.yml
```diff
@@ -61,6 +61,7 @@ jobs:
         shell: bash,
         run: 'ci/ci.sh --platform=${{ matrix.platform }} --archive --skip-builtins --skip-docs engine'
       },
+      { name: 'Stop Gradle Daemon', run: '.\\com.dynamo.cr\\com.dynamo.cr.bob\\gradlew.bat --stop' },
```
AGulev, line 64: stop gradle serveice after build to fix cache issues on windows : https://github.com/actions/setup-java/issues/633#issuecomment-2219579714

build_tools/waf_dynamo.py
JCash: I was surprised to learn that windows has ccache installed. Feels like it shouldn't be needed for CI at all, and then one could install it oneself 🤷 

"it fixes https://github.com/defold/defold/actions/runs/20430930119/job/58701234892?pr=11653"
In the log I don't see anything ccache related, just that it's the last successful step in finding it? 🤔

JCash (approved):

build_tools/waf_dynamo.py
AGulev: It's a bit of speculation from my side, of course, I don't know for sure. Something has crashed on Windows during ccache search. not sure if waf call ccache during search or not, but if we don't need it, it's better to fully skip it

PR #11659, opened 2025-12-23
Protobuf struct support

Jhonnyg:
We now have support for our own protobuf struct type:
https://protobuf.dev/reference/protobuf/google.protobuf/#struct

Note that this is a copy-paste (with slight modifications) of the original google protobuf struct source.

Fixes #11647


engine/ddf/src/test/test_ddf.cpp
```diff
@@ -42,6 +42,16 @@
...
+
+// NOTE:
+// Ideally we should be able to import the ddf_struct.proto from test_ddf_proto.proto,
+// but doing so will generate build and linker errors due to including both our generated
+// header files as well as the protoc generated header files.
+//
+// To circumvent this, we include the ddfc generated header inside a different module called "test_ddf_struct.cpp"
+// and perform the ddf struct tests inside that module instead of here.
+#include "ddf/ddf_struct.pb.h"
+
```
Jhonnyg, line 54: This is not exclusive to the ddf_struct header, we get the same issue if we import ddf_math.

JCash (approved):

PR #11662, opened 2025-12-23
feat: Add editor preferences api endpoint

atomicptr:
This PR adds an API endpoint allowing external code editors to query and update (a subset of) the editor preferences. This specifically addresses [defold.nvim #34](https://github.com/atomicptr/defold.nvim/issues/34).

Why this is needed: Currently, in order to automatically set the external code editor I have to modify prefs.editor_settings directly, this requires me to ship an EDN parser and also a full editor restart for the changes to apply.

## Changes

- Added a GET /prefs endpoint to read current editor settings.
- Added a POST /prefs endpoint to update settings dynamically without a restart.

This allows me to significantly improve the plug and play experience in defold.nvim and would probably also benefit other external editors

## Alternatives

An alternative approach would be to add a new command request that allows me to get the editor to reload the written prefs.editor_settings alleviating the need to restart it

### DIFF
--- a/editor/src/clj/editor/boot_open_project.clj
+++ b/editor/src/clj/editor/boot_open_project.clj
@@ -43,6 +43,7 @@
             [editor.outline-view :as outline-view]
             [editor.pipeline.bob :as bob]
             [editor.prefs :as prefs]
+            [editor.prefs-api :as prefs-api]
             [editor.properties-view :as properties-view]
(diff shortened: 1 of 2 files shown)


vlaaad:
Thank you for your contribution @atomicptr! I've cleaned up the code and opened a new PR based on your initial implementation. Closing this in favor of #11690

PR #11664, opened 2025-12-23
Replaced the deprecated `ALooper_pollAll` call with `ALooper_pollOnce`

AGulev:
`ALooper_pollAll` has been deprecated due to [a potential race condition](https://github.com/android/ndk-samples/issues/999). It has been replaced with `ALooper_pollOnce` in the engine code.

Fix https://github.com/defold/defold/issues/9505


external/glfw/lib/android/android_window.c
```diff
@@ -486,27 +486,26 @@ void _glfwPlatformPollEvents( void )
 // Called from the looper thread
 void glfwAndroidPollEvents()
```
AGulev, line 487: this function called from the looper thread in a loop: https://github.com/defold/defold/blob/2b2b6c302c9645d1ad28f95a6cd1be41324dda7c/engine/engine/src/engine_main.cpp#L131-L140

That's why I didn't add one more internal loop here

JCash (dismissed):

PR #11666, opened 2025-12-24
Lua script builds faster for large files

### LINKED ISSUE #6490: Editor freezes when loading a project
If a script has too many properties listed (10000 in example), the editor will freeze when loading such a project.

**To Reproduce:**
1. Add the script from the archive to any project (you can add it to an empty project).
2. Start the editor, select this project.

**Expected behavior:**
The editor should load in adequate time.

**Defold version:**
  - any version

**Platforms:**
 - Platforms: Windows 10

It's not a project, it's just a script and a set of propperty files. You can add this script to any project. The result will always be the same.
[script.zip](https://github.com/defold/defold/files/8336671/-.zip)

AGulev:
Large Lua scripts now process faster during builds, reducing time spent scanning and stripping properties/comments without changing output behavior.

### Technical changes

It partially fixes https://github.com/defold/defold/issues/6490 - loading time now much faster and the main script file loads 200ms instead of 70s (350 times faster). This improvement speed up Editor loading time, building time from the editor and bundling time from bob.

Even if a project doesn't use go.property() heavily, I expect speed up in big projects for Lua parsing because of much faster comments stripping.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LuaScanner.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LuaScanner.java
@@ -387,9 +387,27 @@ public class LuaScanner {
 
     // replace the token with an empty string
     private static void removeTokens(TokenStreamRewriter rewriter, List<Token> tokens, boolean isDebug) {
-        for (Token token : tokens) {
-            removeToken(rewriter, token, isDebug);
+        if (tokens.isEmpty()) {
+            return;
         }
+        int rangeStartIndex = tokens.getFirst().getTokenIndex();
+        int rangeStopIndex = rangeStartIndex;
+        int rangeLength = tokens.getFirst().getText().length();
+        for (int i = 1; i < tokens.size(); i++) {
+            Token token = tokens.get(i);
+            int tokenIndex = token.getTokenIndex();
+            if (tokenIndex == rangeStopIndex + 1) {
+                rangeStopIndex = tokenIndex;
+                rangeLength += token.getText().length();
+            } else {
+                removeTokenRange(rewriter, rangeStartIndex, rangeStopIndex, rangeLength, isDebug);
+                rangeStartIndex = tokenIndex;
+                rangeStopIndex = tokenIndex;
+                rangeLength = token.getText().length();
+            }
+        }
+        removeTokenRange(rewriter, rangeStartIndex, rangeStopIndex, rangeLength, isDebug);
+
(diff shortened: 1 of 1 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LuaScanner.java
```diff
@@ -386,31 +385,93 @@ private static List<Token> getTokens(CommonTokenStream tokenStream, ParserRuleCo
...
+                rangeStopIndex = tokenIndex;
+                rangeLength += token.getText().length();
+            } else {
+                removeTokenRange(tokenEditor, rangeStartIndex, rangeStopIndex, rangeLength, isDebug);
+                rangeStartIndex = tokenIndex;
+                rangeStopIndex = tokenIndex;
+                rangeLength = token.getText().length();
+            }
+        }
+        removeTokenRange(tokenEditor, rangeStartIndex, rangeStopIndex, rangeLength, isDebug);
```
AGulev, line 408: Removing a token by token isn't efficient, only this improvement without replacement of the `TokenStreamRewriter` reduced time from 70s to 2s, but it's still too much, since `TokenStreamRewriter` calls `reduceToSingleOperationPerIndex` - not important in our case, we do it by design.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LuaScanner.java
```diff
@@ -386,31 +385,93 @@ private static List<Token> getTokens(CommonTokenStream tokenStream, ParserRuleCo
...
+    }
+
+    private static void removeToken(TokenEditor tokenEditor, Token token, boolean isDebug) {
+        tokenEditor.replaceRange(token.getTokenIndex(), token.getTokenIndex(), isDebug ? " ".repeat(token.getText().length()) : "");
+    }
+
+    private record RangeEdit(int startIndex, int stopIndex, String replacement) {
+    }
+
+    private static final class TokenEditor {
```
AGulev, line 436: very simple implementation of our replacement for `TokenStreamRewriter`

vlaaad (approved):
LGTM, but maybe we should have some tests for this? Or is this behavior covered by existing tests?

AGulev:
yes, it's covered by existing tests 
Basically, I just rewrite the way we modify file (and by pass heavy logic which calculates operations overlaping we can't really have in our case ) as a string, but don't really touch logic part of it

PR #11680, opened 2025-12-29
Use the latest version of the `android_native_app_glue`

### LINKED ISSUE #8377: Use default `android_native_app_glue.c` instead of custom one
**Task (REQUIRED):**
`android_native_app_glue.c` we currently use contains a workaround for an old issue on old android version we don't support anymore. So it would be nice to use default one from NDK to make sure we don't need to patch anything in our custom version.

Less custom code in the engine we should support + better handling of the native events, which maybe a reason of some ANR in case we don't handle something properly (less likely).

### LINKED ISSUE #8825: window.get_size() return unreliable value when changing screen resolution
When changing screen resolution on Android phone, I see the `window.get_size()` returns unreliable value.
On WFHD+, app in portrait mode, I've got the window height sometimes 3040 (the correct value) and mostly 2280.
On FHD+, app in portrait mode, I've got  the window height sometimes 1710  and mostly 2280 (the correct value).

**To Reproduce (REQUIRED)**
1. Create a minimum project with a label showing the current window height
2. Build & run it on an Android phone (mine is Samsung Galaxy Note 10 Plus)
3. On the phone, open Settings -> Display -> Screen Resolution and change to the resolution we want
fc9c2f771f10bba27955bbb37edc0d3d1a55db26_2_354x750
4. Reopen the app to see its window height

Get the right window size value, should be equal to screen resolution.

  - Version 1.7.0

**Platforms (REQUIRED):**
 - Platforms: Android
 - OS: Android 12
 - Device: Samsung Galaxy Note 10 Plus

AGulev:
Fix https://github.com/defold/defold/issues/8377

I assume it also fixes https://github.com/defold/defold/issues/8825


external/glfw/lib/android/android_native_app_glue.c
```diff
@@ -12,17 +12,17 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *
  */
 
+#include "android_native_app_glue.h"
```
AGulev, line 17: copy-pasted from our current NDK (25b)

external/glfw/lib/android/android_window.c
```diff
@@ -290,14 +291,17 @@ static void _glfwPlatformSwapBuffersNoLock( void )
...
+     Handle orientation/size changes when signaled by APP_CMD_CONFIG_CHANGED,
+     APP_CMD_WINDOW_RESIZED or APP_CMD_CONTENT_RECT_CHANGED. Some devices
+     report the old EGL size at the time of the event, so we defer the query
+     until a swap occurs.
      */
-    update_width_height_info(&_glfwWin, &_glfwWinAndroid, 0);
+    if (g_PendingResize)
+    {
+        update_width_height_info(&_glfwWin, &_glfwWinAndroid, 1);
+        g_PendingResize = 0;
```
AGulev, line 303: it's enough to do it when we know something has changed. We don't have to do it every swap

JCash (approved):

PR #11690, opened 2026-01-02
Add `/prefs/...` editor HTTP server enpoints

vlaaad:
Now, the editor's HTTP server allows querying and modifying the preferences using its `/prefs/...` endpoints.

Use a `GET` request for reading all preferences:
```sh
$ curl http://localhost:$(cat .internal/editor.port)/prefs | jq
{
  ...big JSON blob...
}
```

Or a group of settings:
```sh
$ curl http://localhost:$(cat .internal/editor.port)/prefs/build | jq
{
  "lint-code": true,
  "texture-compression": true,
  "open-html5-build": true
}
```
Or a single setting:
```
$ curl http://localhost:$(cat .internal/editor.port)/prefs/build/texture-compression | jq
true
```

Use a `POST` request for writing:
```sh
$ curl http://localhost:$(cat .internal/editor.port)/prefs/build/texture-compression -X POST -d 'false'
```

### Technical changes
This PR refactors externally contributed PR #11662 to support any types of prefs, plus adds tests.


editor/src/clj/editor/editor_extensions/prefs_docs.clj
```diff
@@ -36,7 +36,11 @@
   (coerce/wrap-with-pred coerce/userdata editor-script-defined-schema? "is not a schema"))
 
 (definline allowed-keyword-character? [ch]
-  `(or (Character/isLetterOrDigit (char ~ch)) (= \- ~ch) (= \. ~ch) (= \_ ~ch)))
+  `(or (Character/isLetterOrDigit (char ~ch))
+       (= \- ~ch)
+       (= \. ~ch)
+       (= \_ ~ch)
+       (= \+ ~ch)))
```
vlaaad, line 43: I added `+` as a supported character for prefs for extensions because we already use prefs with such keys, but it wasn't possible to read them.

editor/src/clj/editor/http_server/prefs.clj
```diff
@@ -0,0 +1,99 @@
...
+        (http-server/response 400 (.getMessage e))
+        (throw e)))))
+
+(defn- json->value
+  "Coerce JSON value with string keys to prefs value"
+  [json schema]
+  (case (:type schema)
+    (:any :string :password :locale :boolean :integer) json
+    :keyword (cond-> json (string? json) keyword)
+    :number (cond-> json (number? json) double)
```
vlaaad: We don't do a proper validation here since the prefs system is going to validate the values anyway.

editor/src/clj/editor/http_server/prefs.clj
```diff
@@ -0,0 +1,99 @@
...
+      (http-server/response 200))
+    (catch LuaError e
+      (http-server/response 400 (.getMessage e)))
+    (catch ExceptionInfo e
+      (if (::prefs/error (ex-data e))
+        (http-server/response 400 (.getMessage e))
+        (throw e)))))
+
+(defn routes [prefs]
+  {"/prefs/{path}" {"GET" (partial #'handle-get-prefs prefs)
```
JosephFerano: My initial instinct to read nested values was to try to use slashes, so `GET ../prefs/workflow/open-tabs` instead of `GET ../prefs/workflow.open-tabs`, so when I got a 404 I thought it wasn't working. Would it make sense to use a wildcard for the remaining path and convert to the dot format before parsing? If we prefer the dots, then I think we should update `http-api.md` to explain the API

editor/src/clj/editor/http_server/prefs.clj
```diff
@@ -0,0 +1,99 @@
...
+
+(defn- handle-get-prefs [prefs request]
+  (try
+    (http-server/json-response
+      (prefs/get prefs (prefs-functions/parse-dot-separated-path (-> request :path-params :path))))
+    (catch LuaError e
+      (http-server/response 400 (.getMessage e)))
+    (catch ExceptionInfo e
+      (if (::prefs/error (ex-data e))
+        (http-server/response 400 (.getMessage e))
```
JosephFerano, line 54: This one could maybe be a 422, but I'm not sure, would depend on what gets thrown.

JosephFerano (changes requested):
The 400s seem appropriate, if anything a 422 somewhere might be better if it's validation related, but not sure.

vlaaad:
I realized I needed to do a bit more to make it deliverable. The problem was that we had `:any` schema that used Clojure data structures (i.e., keywords) values, but since JSON can't encode them, reading and then writing back would corrupt the preferences. My initial attempt to work around this problem was to require the user to provide a specific path, so it's less likely that the preference values will be corrupted. But this makes preferences really hard to discover: the user has to check the source code first to see which prefs exist. So it's very useful to be able to read all prefs just by accessing the `/prefs` endpoint, to see available prefs. And now we have a proper fix instead of a workaround: all prefs are properly typed, so that changing them via JSON wouldn't corrupt the data.

editor/src/clj/editor/app_view.clj
```diff
@@ -572,17 +567,12 @@
         split-ids))
 
 (defn- stored-split-positions [prefs]
-  (if-some [split-positions (prefs/get prefs prefs-split-positions)]
-    (if (vector? split-positions) ; Legacy preference format
-      (zipmap (map keyword legacy-split-ids)
-              split-positions)
-      split-positions)
-    {}))
+  (prefs/get prefs prefs-split-positions))
```
vlaaad, line 570: Now that we use explicit types for prefs values, I have decided to discontinue support for the legacy format of window split positions (we started automatic conversion from it in 2016, so like 10 years ago...)

editor/src/clj/editor/prefs.clj
```diff
@@ -221,21 +241,22 @@
...
     :dev {:type :object
           :properties
-          {:custom-engine {:type :any
+          {:custom-engine {:type :one-of
+                           :schemas [{:type :enum :values [nil]}
+                                     {:type :string}]
                            :ui {:type :string}}}}
-    :git {:type :object
-          :properties
-          {:credentials {:type :any :scope :project}}}
```
vlaaad, line 235: Removed since it was never used (a while ago I deleted an unused ns that accessed this pref...)

editor/test/editor/prefs_test.clj
```diff
@@ -81,21 +77,29 @@
...
+  (prop/for-all [x gen/small-integer]
     (prefs/valid? {:type :integer} x)))
 
 (defspec integer-schema-invalid-spec 10
   (prop/for-all [x (gen/such-that (complement int?) gen/any)]
     (not (prefs/valid? {:type :integer} x))))
 
 (defspec number-schema-valid-spec 100
-  (prop/for-all [x (gen/one-of [gen/double gen/int])]
+  (prop/for-all [x (gen/one-of [gen/double gen/small-integer])]
```
vlaaad, line 88: `gen/int` is deprecated and simply refers to `gen/small-integer`, which should be used instead.

editor/test/integration/web_server_test.clj
```diff
@@ -158,6 +162,97 @@
...
+                    (testing "write a different value"
+                      (let [new-value (different-value (prefs/get prefs path) schema)]
+                        (when (is (= 200 (:status @(http/request path-url :method "POST" :body (json/write-str new-value)))))
+                          (let [second-get @(http/request path-url :as :string)]
+                            (when (is (= 200 (:status second-get)))
+                              (let [second-json (json/read-str (:body second-get))
+                                    second-clj (prefs/get prefs path)]
+                                (is (not= first-json second-json))
+                                (is (not= first-clj second-clj))
+                                (is (= new-value second-clj))))))))))))))))))
```
vlaaad, line 254: This test ensures that any existing preference is correctly handled via JSON API, so the prefs can't be accidentally corrupted.

editor/src/clj/editor/prefs.clj
```diff
@@ -221,21 +241,22 @@
...
     :welcome {:type :object
               :properties
-              {:last-opened-project-directory {:type :any}
+              {:last-opened-project-directory {:type :one-of
+                                               :schemas [{:type :enum :values [nil]}
```
matgis, line 258: `{:type :enum :values [nil]}` seems fairly common. As a future improvement, we might want to support `{:type :nil}` in schema declarations? 🤔

matgis (approved):
Looks good to me! I think Joe's point about which HTTP error codes we return is worth considering, though. I must confess not knowing much about what is best practice around that.

editor/src/clj/editor/prefs.clj
vlaaad: Good idea for the future!

vlaaad:
Regarding response codes, I looked at [code 422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/422) and thought it doesn't fit in our case: we don't do any checks beyond input syntax, but 422 is for valid syntax + invalid semantics. We really are fine as long as the response code is 4XX (the error is on the client side, not ours).

PR #11693, opened 2026-01-03
Add an embedded sound component instead of using a raw sound

AGulev:
When a raw sound component is added to a GameObject, it is automatically wrapped and converted into an embedded sound component.

Fix https://github.com/defold/defold/issues/4959

### DIFF
--- a/editor/src/clj/editor/game_object.clj
+++ b/editor/src/clj/editor/game_object.clj
@@ -53,6 +53,8 @@
 
 (def unknown-icon "icons/32/Icons_29-AT-Unknown.png")
 
+(declare add-embedded-component)
+
 (defn- gen-ref-ddf [id position rotation scale source-resource ddf-properties]
   (-> (protobuf/make-map-without-defaults GameObject$ComponentDesc
         :id id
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/game_object.clj
```diff
@@ -503,14 +505,31 @@
...
 (defn add-component-handler [workspace project go-id select-fn]
   (when-let [resources (resource-dialog/make
                          workspace project
                          {:ext (get-all-comp-exts workspace)
                           :title (localization/message "dialog.select-component-file.title")
                           :selection :multiple})]
     (doseq [resource resources]
-      (add-referenced-component! go-id resource select-fn))))
+      (if (raw-audio-resource? resource)
+        (add-embedded-sound-component! go-id resource select-fn)
```
AGulev, line 531: I think it should be more data driven and specified in such a component itself, but on the other hand I don't see any other cases like this...

matgis (approved):
It works! ❤️

We probably also want to sanitize data when loading projects that contain existing references to audio files like this. But it's good that this change only touches the code path that adds new components. Sanitation happens at the read stage, so it should not involve the code that adds new components.

It's not a big task to implement sanitation, but we need to add test content to `integration.save-data-test/silent-migrations-test` and so on, so we might as well treat it as a separate task.

AGulev:
@matgis if you don't mind I'll merge it and create a separate issue for migration

matgis:
Sounds good to me!

PR #11700, opened 2026-01-05
Select the first element by default in forms where choosing an option reveals additional settings

### LINKED ISSUE #8442: In editor UI make texture profiles selected by default
Now it's not obvious that texture profile (Default, Generic) is clickable to reveal more options. Especially for newcomers.
Instead it would be nice to have these additional options to be visible from the beginning by having the Default texture profile and Generic platform selected by default.

image

AGulev:
In some forms, it’s not immediately clear that menu selections reveal additional options. To improve this, those options are selected by default.

Here’s an example showing how the same file looks to the user immediately after opening it, before and after the changes:
image

Fix https://github.com/defold/defold/issues/8442

### DIFF
--- a/editor/src/clj/editor/cljfx_form_view.clj
+++ b/editor/src/clj/editor/cljfx_form_view.clj
@@ -36,6 +36,8 @@
             [cljfx.fx.text-formatter :as fx.text-formatter]
             [cljfx.fx.tooltip :as fx.tooltip]
             [cljfx.fx.v-box :as fx.v-box]
+            [cljfx.lifecycle :as fx.lifecycle]
+            [cljfx.mutator :as fx.mutator]
             [clojure.set :as set]
             [clojure.string :as string]
             [dynamo.graph :as g]
@@ -555,9 +557,9 @@
     ;; default
     (add-list-elements [(form/field-default element)] map-event)))
 
-(defmethod handle-event :remove-list-selection [{:keys [on-removed state-path ui-state]}]
+(defmethod handle-event :remove-list-selection [{:keys [on-removed state-path ui-state selected-indices]}]
   (let [indices-path (conj state-path :selected-indices)]
-    [[:dispatch (assoc on-removed :fx/event (set (get-in ui-state indices-path)))]
+    [[:dispatch (assoc on-removed :fx/event (set selected-indices))]
      [:set-ui-state (-> ui-state
                         (assoc-in indices-path [])
                         (update-in state-path dissoc :edit))]]))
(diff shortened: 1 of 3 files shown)


editor/src/clj/editor/cljfx_form_view.clj
```diff
@@ -1202,12 +1202,23 @@
                                             state-path
                                             panel-key
                                             resource-string-converter
-                                            localization-state]
+                                            localization-state
+                                            auto-select-first?]
```
vlaaad: Please use `?` suffix only for predicate functions, but not for boolean values. See https://vlaaad.github.io/2019-03-30/question-marks-in-clojure

editor/src/clj/editor/cljfx_form_view.clj
```diff
@@ -1202,12 +1202,23 @@
...
+        auto-select-index (when (and auto-select-first?
+                                     (nil? selected-index)
+                                     (not (contains? state :key))
+                                     (seq value))
+                            0)
+        selected-index (or selected-index auto-select-index)
+        key-state (cond-> (:key state)
+                          auto-select-index (assoc :selected-indices [auto-select-index]))
+        state (cond-> state
+                      auto-select-index (assoc :key key-state))
```
vlaaad: It seems to me this can be simplified (we only need to check for `auto-select-index` once) to something like:
```clj
state (cond-> state auto-select-index (update :key assoc :selected-indices [auto-select-index]))
```

vlaaad (changes requested):
LGTM, but please rename `:auto-select-first?` to `:auto-select-first` before merging.

editor/src/clj/editor/cljfx_form_view.clj
AGulev: ah, ok
thanks

vlaaad (changes requested):
The code changes seem to be good, but the behavior is wrong. If you open a view that performs this automatic selection, and then press the `-` button (which deletes the selected item), it will only clear the selection. Meanwhile, if you change a selection explicitly and then press `-`, it will delete the item as expected:

https://github.com/user-attachments/assets/d84b58eb-d0de-463e-951a-e890ba1bd739

It seems that your changes are, unfortunately, not enough to enable auto-selection of items.

AGulev:
hmm, yeah I missed that, initially tried to update `ui-state` to make everything more idiomatic, but it didn't select fields then.
Ofc. I can patch remove hadler, but then we will have two sources of truth which we don't want to have.
Let me check cljfx selection example again

AGulev:
it seems like this approach works, maybe last time I tested it I missed something 🤔

vlaaad:
I noticed some weirdness with the selection after deletion (i.e., it disappeared) and took a shot at fixing it.

matgis (approved):
Code looks good to me!

AGulev:
> Code looks good to me!

And some kind of cljfx magic to me =)

PR #11701, opened 2026-01-06
Move project dependencies into their own separate category in the Assets panel

AGulev:
In projects with many dependencies, having them all in the main folder can be disruptive. To improve this, all dependencies are moved into a separate Dependencies folder at the same root level as the project.

Fix https://github.com/defold/defold/issues/4156


editor/src/clj/editor/asset_browser.clj
```diff
@@ -698,31 +760,37 @@
...
-        db (.startDragAndDrop ^Node (.getSource e) (into-array TransferMode [mode]))
-        content (ClipboardContent.)]
-    (when (= 1 (count resources))
-      (.setDragView db (icons/get-image (workspace/resource-icon (first resources)) 16)
-                    0 16))
-    (.putFiles content files)
-    (.putString content paths)
-    (.setContent db content)
-    (.consume e)))
+  (when-not (some asset-group-resource? selection)
```
AGulev: the only change here is this line

vlaaad (changes requested):
This approach does not work because the grouping object **must not** be a resource. A resource is something with a path that can be referenced by the path from the workspace. Fake paths cannot work here — they get exposed, but don't exist in the workspace. We have dozens of menu commands that adapt selection to a resource. Editor scripts define commands that work on selected resources. This check for "is a grouping resource?" will have to spread throughout the rest of the editor to every resource handler. One example of a problem with creating this fake resource kind is an editor script command:
```lua
        {
            label = "Format JSON",
            locations = {"Assets"},
            query = {selection = {type = "resource", cardinality = "one"}},
            active = function(opts)
                local path = editor.get(opts.selection, "path")
                return path:match(".json$") ~= nil
            end,
            run = function(opts)
            end
        }
```
This command will throw an error on right-click on the `Dependencies` tree item, because it will receive `/__editor__dependencies` as a path, and then will use to look up its properties, which will then fail, since no such file exists.

A better solution that does not affect the rest of the editor is to make a custom `(defonce/record AssetGroup [localization-key children])` instead, which is not a resource, and make the TreeItem-building function aware that its input type is not `Resource`, but `Resource|AssetGroup`.

BTW, regarding localization. Currently, changing the language in prefs does not immediately update the `Dependencies` text as it updates the rest of the editor.

editor/src/clj/editor/app_view.clj
```diff
@@ -365,25 +365,36 @@
...
-     (when (resource/openable-resource? resource)
-       resource)))
+   (context-openable-resource app-view selection nil nil))
   ([app-view selection evaluation-context]
-   (when-let [resource (or (selection->single-resource selection)
-                           (g/node-value app-view :active-resource evaluation-context))]
+   (context-openable-resource app-view selection nil evaluation-context))
+  ([app-view selection selection-context evaluation-context]
+   (let [resource (or (selection->single-resource selection)
+                      (when (allow-active-resource? selection selection-context)
```
AGulev, line 380: Here we have fallback for the context menu: if there is nothing is selection, we show context menu for the opened file. That means, if a user click on our new Dependencies node, he will see a context menu for the opened file tab. These changes in `app_view.clj` needed to prevent it. I didn't find a better way to do it.

(I don't like this fallback at all and I don't get what the goal to have it, but maybe I missed something and it's critical so context menu for tabs work idk)

editor/src/clj/editor/asset_browser.clj
```diff
@@ -55,15 +57,84 @@
 (defn- ->path [s]
   (Paths/get s empty-string-array))
 
+(defonce/record AssetGroup [localization-key icon children])
+
+(defn- asset-group? [node]
+  (instance? AssetGroup node))
+
+(defn- make-asset-tree-cell
```
AGulev, line 65: I didn't find a better way to make localization hot-reloading than having a custom factory, as we have for outline.
(using a default one required the three rebuild to show localization, I also tried a hacky way with a custom label and node on top and it had ugly artifacts and also looks like ugly hack overall)

editor/src/clj/editor/asset_browser.clj
```diff
@@ -813,12 +939,15 @@
 (defrecord SelectionProvider [asset-browser]
   handler/SelectionProvider
   (selection [this]
-    (ui/selection (g/node-value asset-browser :tree-view)))
+    (let [selection (ui/selection (g/node-value asset-browser :tree-view))]
+      (if (some asset-group? selection)
```
AGulev, line 945: if Dependencies in the selection, all the context menu options should be inactive

editor/src/clj/editor/asset_browser.clj
```diff
@@ -612,24 +693,63 @@
...
+        local-resources (remove dependency-resource? children)
+        dependencies-group (when (not (coll/empty? dependencies))
+                             (->AssetGroup "dialog.dependencies.title"
+                                           "icons/32/Icons_03-Builtins.png"
+                                           (vec dependencies)))
+        project-root (assoc resource-tree :children (vec local-resources))
+        top-children (cond-> []
+                             dependencies-group (conj dependencies-group)
+                             true (conj project-root))]
+    (->AssetGroup nil nil (vec top-children))))
```
vlaaad: Perhaps, `(localization/message "pane.assets")` here as a first argument? It's useful to have IDs for items...

vlaaad:
Closing this in favor of #11748

Issue #11587, opened 2025-12-08
Button for collapsing all folders (except root) in the asset window


NaakkaDev:
**Is your feature request related to a problem? Please describe (REQUIRED):**
Would be nice to have a quick way to collapse the open folders in the asset panel. Except the root folder (why is it even there?).

**Describe the solution you'd like (REQUIRED):**
A button next to the "Assets" title, like in VS Code.
Example:
<img width="399" height="84" alt="Image" src="https://github.com/user-attachments/assets/a146d430-0c5f-4c2b-8d54-6b3adc2070e5" />

**Describe alternatives you've considered (REQUIRED):**
Some key combination, but both would be best imho.

awulkan:
This would be nice to have in the outline pane as well.

Insality:
option + left click will do the trick :)

awulkan:
@Insality Thanks, but even if it technically works I think it could be vastly improved by adding a visual button toggle from a UI/UX perspective. :)

It's not obvious at all to a user that you can even do that. Plus it forces you to use both your mouse and keyboard simultaneously just to toggle the hierarchy.

PR #11703, opened 2026-01-06
Add "Open File" and "New File" buttons to the Assets panel header to improve discoverability

### LINKED ISSUE #6831: Assets window search/filter box
That would be nice to have a search (filter) box for assets (and for outline as well)
telegram-cloud-photo-size-2-5402090894770945148-y

AGulev:
We discovered that some users are not aware of the "New File" (Cmd/Ctrl + N) and "Open File" (Cmd/Ctrl + P) popups or the shortcuts that trigger them. To improve discoverability, we added these buttons to the Assets panel header, along with tooltips that display the corresponding shortcuts.

assets_panel

Fix https://github.com/defold/defold/issues/6831


editor/src/clj/editor/boot_open_project.clj
```diff
@@ -236,7 +237,43 @@
           port-file (doto (io/file project-path ".internal" "editor.port")
                       (io/make-parents)
                       (spit port-file-content))]
-      (localization/localize! (.lookup root "#assets-pane") localization (localization/message "pane.assets"))
+      (let [^TitledPane assets-pane (.lookup root "#assets-pane")
```
AGulev, line 296: @vlaaad is it OK to keep it in JavaFX, or should I rewrite it in cljfx?

vlaaad: It's fine!

AGulev: too late. it is in cljfx already

editor/src/clj/editor/boot_open_project.clj
```diff
@@ -238,42 +294,13 @@
...
-        (ui/on-action! search-button (fn [_] (ui/execute-command (ui/contexts (.getScene search-button)) :file.open nil)))
-        (HBox/setHgrow spacer Priority/ALWAYS)
+            header-desc {:fx/type assets-pane-header
+                         :localization localization
+                         :keymap (g/node-value app-view :keymap)}]
         (.setText assets-pane "")
-        (.setGraphic assets-pane header)
-        (.addAll (.getChildren header) (ui/node-array [assets-title spacer new-file-button search-button]))
-        (localization/localize! assets-title localization (localization/message "pane.assets")))
+        (fxui/advance-ui-user-data-component! assets-pane ::assets-pane-header header-desc)
```
vlaaad: Cljfx only really makes sense when there are updates of UI state from one to another, otherwise it's just overhead... You also don't need to save the cljfx state on the assets pane node since you are not going to use it later; in this case, `fx/create-component` is preferred over `fxui/advance-ui-user-data-component!`

AGulev: should I revert cljfx then, or it's fine if I use fx/create-component ?

vlaaad: It's fine to use `fx/create-component`, it's better because it uses a single localization subscription for both tooltips!

editor/src/clj/editor/boot_open_project.clj
```diff
@@ -293,7 +356,7 @@
       (ui/on-closing! stage (fn [_]
                               (let [dirty-save-data (project/dirty-save-data project)
                                     auto-save-on-quit? (prefs/get prefs [:workflow :save-on-app-focus-lost])]
-                                (if (and auto-save-on-quit? (seq dirty-save-data))
+                                (if (and auto-save-on-quit? (not (coll/empty? dirty-save-data)))
```
AGulev, line 359: I didn't touch it, but now I'm checking every file I work in ... =)

editor/resources/localization/zh_CN.editor_localization
```diff
@@ -739,6 +739,8 @@ menu.help = 帮助
 
 # Commands
 
+command.tooltip = {command}{shortcut, select, none {} other { ({shortcut})}}
```
AGulev, line 742: missed in prev. PR, the labels reused from commands

AGulev:
Related, but can be fixed separately later: https://github.com/defold/defold/issues/11587

editor/resources/localization/zh_CN.editor_localization
vlaaad: In this case, it's probably fine, but as a general rule, please avoid adding a translation unless you are a native speaker of the language. If the translation is omitted, you'll get the English version as a fallback, so it will continue to work fine, but now the translators might miss an opportunity to add a better version. For example, some asian languages prefer using a single space or a dash instead of wrapping the shortcut in parentheses.

vlaaad (commented):
Not sure whether to approve or reject this PR, I'll let @matgis decide. The buttons work great most of the time. There is one problem, though. The buttons represent commands, and commands might be in active/inactive states (not visible to the user) and enabled/disabled states (visible to the user, but not available). The Open asset command is always active/enabled, so it's all fine. The New file command, on the other hand, is sometimes disabled, e.g., during resource sync or save. This is really noticeable on big projects during resource sync: you focus on the editor, click the New File icon, select the file type to create, and then nothing will happen. This is probably an undesired behavior, and we should enable/disable the icon, or maybe we shouldn't bother, since it's a small annoyance. I don't know.

matgis (changes requested):
It's a good start, but this PR feels kind of half-baked to me.

* As Vlad already pointed out, the buttons need to be enabled or disabled to reflect their availability. Like Vlad brought up, we should disallow creating new resources while bundling the project, and I'd also like the Add Asset button to be disabled when multiple destinations are selected or a read-only destination such as builtins is selected in the tree. But I'm not sure if we currently have a good bridge between Cljfx-based views and the enabled state of the handlers. Would we need to built that? 🤔
* The buttons appear too large to me in their current state. Maybe they should be around half the size they are currently?
* Instead of adding more `.png` icon files to the repo, I'd like to see the SVG path data added to `fxui/icon-svg-path-data` so we can use `fxui/icon-graphic` to create the icons in whatever dimensions we want.
* Why are we showing a resource picker where you select the type of resource to add? Shouldn't we show the new categorized resource type menu instead?
* The discrepancy in behavior between these two buttons feels a bit strange. The Add button operates on the current selection inside the Assets panel, so it makes sense that it is connected to it. But the Open button simply shows a popup that is also available from the menu bar. I would maybe expect it to open the selected Asset, since it is part of the Assets panel. But that would be of little use. To me, it feels like it belongs more to the Tab Bar, maybe? I'm open to discussion, though.
* I feel if we should have an Add button in the Assets Panel, we might also want one in the Outline panel? But there it may be a bit tricker since we have primary and secondary add actions.
* I have some concerns that adding these buttons might conflict with our plans to make the tree filterable. I.e. will it end up cluttered as we add more stuff to it?

So overall, I'm a bit reserved. I can see arguments for adding these buttons, but it doesn't feel like an obvious slam-dunk to me. There are trade-offs. But if we do decide to go ahead with them, we need to make the enabled states work and preferably lose the `.png` icons in favor of scalable SVG paths.

AGulev:
I don't know if we ever added a filter there. From previous conversations, I remember that we intentionally don't want to filter in tree components.

This change is a small discoverability improvement I decided to add, knowing that filtering is not planned and after hearing multiple reactions like "oh, wow, I didn't know Defold has it". For example, @dragosha discovered `cmd+P` only recently (around two years ago), even though he has been using the editor for many years.

At the same time, I'm not ready to spend a lot of time defending it. Since it's a new type of user UI and communication with the user, maybe we need to agree on what we want to see UI wise in the long run, so we move toward it and avoid conflicts between "fast fixes" and a long-term vision.

I'm closing this Pr for now until we decide what we want and need.

PR #11704, opened 2026-01-06
Added "Close" button to the animation preview in atlas

### LINKED ISSUE #8960: Animation preview interferes with viewing the atlas
### Steps to reproduce

<hr/>

<table>
<tr><td>Build time</td><td>2024-05-15T14:56:13.376883</td></tr>
<tr><td>Defold editor sha</td><td>c99258874cf2e19c04a1025974a403df8c3f4dfc</td></tr>
<tr><td>Defold engine sha</td><td>9141d9d3605e3f5d51c71293116d769da2613d39</td></tr>
<tr><td>Defold version</td><td>1.8.0</td></tr>
<tr><td>GPU</td><td>AMD Radeon(TM) Vega 8 Graphics</td></tr>
<tr><td>GPU Driver</td><td>4.6.13587 Compatibility Profile Context 19.40.54.01 26.20.14054.1001</td></tr>
<table>

[issue.zip](https://github.com/defold/defold/files/15374616/issue.zip)

AGulev:
It wasn’t obvious how to close the animation preview in the atlas, so a Close button was added along with a tooltip reminding users of the shortcut:

Fix https://github.com/defold/defold/issues/8960

### DIFF
--- a/editor/resources/localization/en.editor_localization
+++ b/editor/resources/localization/en.editor_localization
@@ -924,7 +924,6 @@ command.debugger.stop = Stop Debugger
 command.debugger.toggle-breakpoint = Toggle Breakpoint
 command.debugger.toggle-breakpoint-enabled = Toggle Breakpoint Enabled
 command.debugger.edit-breakpoint = Edit Breakpoint
-command.scene.close-animation-preview = Close preview
 command.scene.set-manipulator-space = Set Manipulator Space
 command.scene.set-manipulator-space.option.world = World Space
 command.scene.set-manipulator-space.option.local = Local Space
--- a/editor/resources/localization/en.editor_localization
+++ b/editor/resources/localization/en.editor_localization
@@ -763,6 +763,8 @@ menu.help = Help
 
 # Commands
 
+command.tooltip = {command}{shortcut, select, none {} other { ({shortcut})}}
+
 command.file.new = New…
 command.file.new.option.any-file = File
 command.file.new.any-file = New File
@@ -922,6 +924,7 @@ command.debugger.stop = Stop Debugger
 command.debugger.toggle-breakpoint = Toggle Breakpoint
 command.debugger.toggle-breakpoint-enabled = Toggle Breakpoint Enabled
 command.debugger.edit-breakpoint = Edit Breakpoint
+command.scene.close-animation-preview = Close preview
 command.scene.set-manipulator-space = Set Manipulator Space
 command.scene.set-manipulator-space.option.world = World Space
 command.scene.set-manipulator-space.option.local = Local Space
@@ -967,11 +970,6 @@ command.help.open-donations = Development Fund
 command.app.about = About
 command.vcs.diff = View Diff
 command.vcs.revert = Revert
-
-# Scene
-
-scene.preview.close = Close preview
-scene.preview.close.tooltip = Close preview ({shortcut})
 
 # Code editor – find bar
 
--- a/editor/resources/localization/en.editor_localization
+++ b/editor/resources/localization/en.editor_localization
@@ -968,6 +968,11 @@ command.app.about = About
 command.vcs.diff = View Diff
 command.vcs.revert = Revert
 
+# Scene
+
+scene.preview.close = Close preview
+scene.preview.close.tooltip = Close preview ({shortcut})
+
 # Code editor – find bar
 
 code.find.word = Word
--- a/editor/resources/localization/pt_BR.editor_localization
+++ b/editor/resources/localization/pt_BR.editor_localization
@@ -943,6 +943,11 @@ command.app.about = Sobre
 command.vcs.diff = Ver Diff
 command.vcs.revert = Desfazer
 
+# Scene
+
+scene.preview.close = Fechar visualização
+scene.preview.close.tooltip = Fechar visualização ({shortcut})
+
 # Code editor – find bar
 
 code.find.word = Palavra
--- a/editor/resources/localization/pt_BR.editor_localization
+++ b/editor/resources/localization/pt_BR.editor_localization
@@ -943,11 +943,6 @@ command.app.about = Sobre
 command.vcs.diff = Ver Diff
 command.vcs.revert = Desfazer
 
-# Scene
-
-scene.preview.close = Fechar visualização
-scene.preview.close.tooltip = Fechar visualização ({shortcut})
-
 # Code editor – find bar
 
 code.find.word = Palavra
--- a/editor/resources/localization/ru.editor_localization
+++ b/editor/resources/localization/ru.editor_localization
@@ -952,6 +952,11 @@ command.app.about = О программе
 command.vcs.diff = Сравнить
 command.vcs.revert = Откатить
 
+# Scene
+
+scene.preview.close = Закрыть предпросмотр
+scene.preview.close.tooltip = Закрыть предпросмотр ({shortcut})
+
 # Code editor – find bar
 
 code.find.word = Слово
--- a/editor/resources/localization/ru.editor_localization
+++ b/editor/resources/localization/ru.editor_localization
@@ -908,7 +908,6 @@ command.debugger.detach = Отсоединить отладчик
 command.debugger.stop = Остановить отладчик
 command.debugger.toggle-breakpoint = Переключить точку останова
 command.debugger.edit-breakpoint = Изменить точку останова
-command.scene.close-animation-preview = Закрыть просмотр
 command.scene.set-manipulator-space = Задать пространство манипулятора
 command.scene.set-manipulator-space.option.world = Глобальное пространство
 command.scene.set-manipulator-space.option.local = Локальное пространство
--- a/editor/resources/localization/ru.editor_localization
+++ b/editor/resources/localization/ru.editor_localization
@@ -748,6 +748,8 @@ menu.help = Справка
 
 # Commands
 
+command.tooltip = {command}{shortcut, select, none {} other { ({shortcut})}}
+
 command.file.new = Создать…
 command.file.new.option.any-file = Файл
 command.file.new.any-file = Новый файл
@@ -906,6 +908,7 @@ command.debugger.detach = Отсоединить отладчик
 command.debugger.stop = Остановить отладчик
 command.debugger.toggle-breakpoint = Переключить точку останова
 command.debugger.edit-breakpoint = Изменить точку останова
+command.scene.close-animation-preview = Закрыть просмотр
 command.scene.set-manipulator-space = Задать пространство манипулятора
 command.scene.set-manipulator-space.option.world = Глобальное пространство
 command.scene.set-manipulator-space.option.local = Локальное пространство
@@ -951,11 +954,6 @@ command.help.open-donations = Фонд разработки
 command.app.about = О программе
 command.vcs.diff = Сравнить
 command.vcs.revert = Откатить
-
-# Scene
-
-scene.preview.close = Закрыть предпросмотр
-scene.preview.close.tooltip = Закрыть предпросмотр ({shortcut})
 
 # Code editor – find bar
 
--- a/editor/resources/localization/zh_CN.editor_localization
+++ b/editor/resources/localization/zh_CN.editor_localization
@@ -943,11 +943,6 @@ command.app.about = 关于
 command.vcs.diff = 查看差异
 command.vcs.revert = 还原
 
-# Scene
-
-scene.preview.close = 关闭预览
-scene.preview.close.tooltip = 关闭预览（{shortcut}）
-
 # Code editor – find bar
 
 code.find.word = 全词匹配
--- a/editor/resources/localization/zh_CN.editor_localization
+++ b/editor/resources/localization/zh_CN.editor_localization
@@ -943,6 +943,11 @@ command.app.about = 关于
 command.vcs.diff = 查看差异
 command.vcs.revert = 还原
 
+# Scene
+
+scene.preview.close = 关闭预览
+scene.preview.close.tooltip = 关闭预览（{shortcut}）
+
 # Code editor – find bar
 
 code.find.word = 全词匹配
--- a/editor/src/clj/editor/scene.clj
+++ b/editor/src/clj/editor/scene.clj
@@ -785,8 +785,8 @@
            :max-width preview-close-button-size
            :max-height preview-close-button-size
            :graphic {:fx/type fxui/icon-graphic
-                     :type :icon/cross
-                     :size 10.0
+                     :type :icon/cross-thick
+                     :size 8.0
                      :style-class "scene-preview-close-icon"}
            :tooltip (stop-preview-tooltip localization keymap)
            :on-action (fn [^ActionEvent event]
--- a/editor/src/clj/editor/scene.clj
+++ b/editor/src/clj/editor/scene.clj
@@ -1717,6 +1717,8 @@
   (input input-handlers Runnable :array)
   (input active-tool g/Keyword)
   (input manip-space g/Keyword)
+  (input localization g/Any)
+  (input keymap g/Any)
 
   (input hidden-renderable-tags types/RenderableTags)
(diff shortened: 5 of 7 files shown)


editor/src/clj/editor/scene.clj
```diff
@@ -1059,6 +1134,9 @@
                     (play-handler view))))
 
 (defn- stop-handler [view-id]
+  (when-let [image-view ^ImageView (g/node-value view-id :image-view)]
+    ;; Force a redraw on next refresh without requiring input.
+    (ui/user-data! image-view ::last-frame-version nil))
```
AGulev, line 1140: This helps prevent an issue where the preview remains visible after clicking the button until the cursor is moved.

editor/src/clj/editor/scene.clj
```diff
@@ -758,7 +764,66 @@
   (assert (= (count buf) (* picking-drawable-size picking-drawable-size)) "picking buf of unexpected size")
   (map (partial aget buf) picking-buf-spiral-indices))
 
-(g/defnk produce-overlay-anchor-pane-props [scene ^:try tool-info-text]
+(def ^:private preview-close-button-size 18.0)
```
AGulev, line 767: The button scaled together with the window and since we use it in calculations, it can't be moved into `scss`

editor/resources/localization/pt_BR.editor_localization
```diff
@@ -947,6 +947,11 @@ command.app.about = Sobre
 command.vcs.diff = Ver Diff
 command.vcs.revert = Desfazer
 
+# Scene
```
AGulev: we don't have localizations for scene, it seems, so I added a new category

editor/src/clj/editor/scene.clj
```diff
@@ -758,7 +764,66 @@
...
+           :style-class "scene-preview-close-button"
+           :min-width preview-close-button-size
+           :min-height preview-close-button-size
+           :pref-width preview-close-button-size
+           :pref-height preview-close-button-size
+           :max-width preview-close-button-size
+           :max-height preview-close-button-size
+           :graphic {:fx/type fxui/icon-graphic
+                     :type :icon/cross
+                     :size 10.0
```
matgis: The cross icon looks a bit thin to me. I did an experiment and put in `:type :icon/cross-thick` and `:size 8.0`, which makes it look more like the cross icons on the editor tabs. It also gives it a little bit more space so it doesn't feel so close to the corner.
```suggestion
                     :type :icon/cross-thick
                     :size 8.0
```

editor/resources/localization/zh_CN.editor_localization
```diff
@@ -947,6 +947,11 @@ command.app.about = 关于
 command.vcs.diff = 查看差异
 command.vcs.revert = 还原
 
+# Scene
+
+scene.preview.close = 关闭预览
+scene.preview.close.tooltip = 关闭预览（{shortcut}）
+
```
matgis: I think maybe we aren't supposed to add string to the non-english localization files and instead letting the tooling handle it, but I defer to @vlaaad if there are special circumstances around adding a category? 🤔

matgis (changes requested):
Works great! I posted a suggestion to tweak the visuals a bit to be more in line with the rest of the editor, and I'm unsure if we're allowed to add non-english localization strings directly, but apart from that this looks like a nice feature to me! ❤️

editor/src/clj/editor/scene.clj
AGulev: I don't mind, I decided on this one just because if it's small pixelart animation a big cross it cover the animation too much

editor/resources/localization/en.editor_localization
```diff
@@ -968,6 +968,11 @@ command.app.about = About
 command.vcs.diff = View Diff
 command.vcs.revert = Revert
 
+# Scene
+
+scene.preview.close = Close preview
+scene.preview.close.tooltip = Close preview ({shortcut})
```
vlaaad: There are other icon buttons that we want to show with a tooltip. We should make a generic shortcut tooltip localization key and use it instead.

I think the localization message should be like this:
```properties
command.tooltip = {command}{shortcut, select, none {} other { ({shortcut})}}
```

Then, when we use the localization, we don't pass `nil` as a not-found shortcut text, but instead use `"none"`, e.g.:
```clj
(localization/message 
  "command.tooltip" 
  {"command" (localization/message "command.scene.stop") 
   "shortcut" (keymap/display-text keymap :scene.stop "none")})
```

editor/resources/localization/ru.editor_localization
```diff
@@ -952,6 +952,11 @@ command.app.about = О программе
 command.vcs.diff = Сравнить
 command.vcs.revert = Откатить
 
+# Scene
+
+scene.preview.close = Закрыть предпросмотр
+scene.preview.close.tooltip = Закрыть предпросмотр ({shortcut})
```
vlaaad: In other places, it's called `просмотр`, not `предпросмотр`.

editor/resources/localization/pt_BR.editor_localization
```diff
@@ -943,6 +943,11 @@ command.app.about = Sobre
...
+
+scene.preview.close = Fechar visualização
```
vlaaad: Some other translations of `preview` in ptBR use `Prévia` instead. I'd suggest removing ptBR/Chinese translations so that native speakers can select what fits best.

editor/src/clj/editor/scene.clj
```diff
@@ -758,7 +764,66 @@
   (assert (= (count buf) (* picking-drawable-size picking-drawable-size)) "picking buf of unexpected size")
   (map (partial aget buf) picking-buf-spiral-indices))
 
-(g/defnk produce-overlay-anchor-pane-props [scene ^:try tool-info-text]
+(def ^:private preview-close-button-size 18.0)
+
+(defn- stop-preview-tooltip [localization keymap]
+  (let [shortcut (when keymap
```
vlaaad: I don't think `keymap` should be nilable here. It's always present in the AppView that you wire from the graph.

editor/src/clj/editor/scene.clj
```diff
@@ -771,21 +836,29 @@
...
+            close-button (when-let [anim-data (and (seq active-updatable-ids)
+                                                   (active-animation-anim-data updatables active-updatable-ids))]
+                           (stop-preview-button
+                             (animation-preview-anchor-props camera viewport anim-data)
+                             localization
+                             keymap))
+            children (cond-> []
+                             info-text (conj (info-label info-text))
+                             close-button (conj close-button))]
+        (if (seq children)
```
vlaaad: Please don't use `seq` for a non-empty check. It does empty checks by creating garbage objects. Instead, use `(not (coll/empty? children))`. Perhaps, you might have checked the `empty?` docstring and got the impression that `seq` is an appropriate way to check for non-emptyness:
```
(defn empty?
  "Returns true if coll has no items. To check the emptiness of a seq,
  please use the idiom (seq x) rather than (not (empty? x))"
  ...)
```
Unfortunately, this docstring is wrong, and the advice is garbage (literally).

editor/src/clj/editor/scene.clj
```diff
@@ -771,21 +836,29 @@
...
+          (cond-> {:pick-on-bounds false
+                   :children children}
+                  (nil? close-button) (assoc :mouse-transparent true)
+                  close-button (assoc :mouse-transparent false))
```
vlaaad:
```clj
{:pick-on-bounds false
 :children children
 :mouse-transparent (nil? close-button)}

File: editor/src/clj/editor/scene.clj [hunk 12]
```diff
@@ -758,7 +764,66 @@
...
+
+(defn- stop-preview-tooltip [localization keymap]
+  (let [shortcut (when keymap
+                   (keymap/display-text keymap :scene.stop nil))]
+    (if (and shortcut (not (string/blank? shortcut)))
+      (localization (localization/message "scene.preview.close.tooltip" {"shortcut" shortcut}))
+      (localization (localization/message "scene.preview.close")))))
+
+(defn- stop-preview-button [anchor-props localization keymap]
+  (-> (fxui/button
```
vlaaad: We don't use `fxui/button` as a function; instead, we use it as a map with `:fx/type fxui/button` key. This is done to minimize invocations of these UI description functions, where some of them might perform some "heavy" computations.

editor/src/clj/editor/scene.clj
```diff
@@ -758,7 +764,66 @@
...
+        (fxui/resolve-tooltip
```
vlaaad: `resolve-tooltip` is accidentally public; it shouldn't be used outside of `fxui` ns. I think it's better to add `resolve-tooltip` to the `button` component in the `fxui` ns.

editor/src/clj/editor/scene.clj
```diff
@@ -758,7 +764,66 @@
...
 
-(g/defnk produce-overlay-anchor-pane-props [scene ^:try tool-info-text]
+(def ^:private preview-close-button-size 18.0)
+
+(defn- stop-preview-tooltip [localization keymap]
+  (let [shortcut (when keymap
+                   (keymap/display-text keymap :scene.stop nil))]
+    (if (and shortcut (not (string/blank? shortcut)))
+      (localization (localization/message "scene.preview.close.tooltip" {"shortcut" shortcut}))
+      (localization (localization/message "scene.preview.close")))))
```
vlaaad: This code applies the localization once on rendering, and doesn't refresh it on language change in prefs. Instead, `stop-preview-button` should be a cljfx component function (not invoked directly, but used in a map on `:fx/type` key), and it should be an `fxui/defc` component that watches the localization state so the tooltip is refreshed on language change. See `fxui/defc breakpoints-view` for an example of dynamic localization. It's almost always wrong to invoke `localization` as a function!

editor/src/clj/editor/scene.clj
```diff
@@ -771,21 +836,29 @@
...
-                 tool-info-text
-                 (let [scene-info-text (:info-text scene)]
-                   (when (and (string? scene-info-text)
-                              (pos? (count scene-info-text)))
-                     scene-info-text)))]
-        {:mouse-transparent true
-         :children [{:fx/type fx.label/lifecycle
-                     :style-class "info-label"
-                     :anchor-pane/top 15
-                     :anchor-pane/left 30
```
vlaaad, line 786: The code that assigned `:anchor-pane/*` keys was better before, because these keys are only valid on maps that are items of the `:children` array, not some nested components. It's better to keep them in the scope where the children are created.

editor/src/clj/editor/scene.clj
```diff
@@ -771,21 +836,29 @@
...
+            (if (and (string? tool-info-text)
+                     (pos? (count tool-info-text)))
+              tool-info-text
+              (let [scene-info-text (:info-text scene)]
+                (when (and (string? scene-info-text)
+                           (pos? (count scene-info-text)))
+                  scene-info-text)))
+            close-button (when-let [anim-data (and (seq active-updatable-ids)
+                                                   (active-animation-anim-data updatables active-updatable-ids))]
+                           (stop-preview-button
```
vlaaad: This is a good example for assigning `:anchor-pane/*` keys here instead of passing them to the `stop-preview-button` function. `stop-preview-button` should be a cljfx component function (not invoked directly, but created as a map with `:fx/type stop-preview-button` key), because it should be an `fxui/defc` component that watches the localization so it's properly reapplied on language change. But when it's used as a proper cljfx component, passing `:anchor-pane/*` will not work because these keys are only valid on maps that are used as `:children` items. So it should be something like:
```clj
(merge 
  {:fx/type stop-preview-button :localization localization :keymap keymap}
  (animation-preview-anchor-props camera viewport anim-data))
```

vlaaad (changes requested):
Works great, though I have many notes about the code 😀

AGulev:
It feels like drawing a button in OpenGL is much simpler than adding it using a UI framework that supposed to help to make it easy

editor/resources/localization/en.editor_localization
AGulev: this is a good one

editor/resources/localization/pt_BR.editor_localization
AGulev: yes, I removed it 
btw should I leave empty keys or fully remove it?

editor/resources/localization/ru.editor_localization
AGulev: fixed

editor/src/clj/editor/scene.clj
AGulev: fixed

editor/src/clj/editor/scene.clj
AGulev: fixes

editor/src/clj/editor/scene.clj
AGulev: fixed, and it's private now

editor/src/clj/editor/scene.clj
AGulev: fixed

editor/src/clj/editor/scene.clj
AGulev: 🤯

AGulev:
@vlaaad 
I believe all your feedback has been addressed, unless I missed something.  
Thank you for the detailed feedback. I’m still a bit confused about certain parts and why things are so complicated, but I really appreciate the time and effort you put into it.

editor/src/clj/editor/scene.clj
```diff
@@ -758,7 +764,68 @@
...
   (map (partial aget buf) picking-buf-spiral-indices))
 
-(g/defnk produce-overlay-anchor-pane-props [scene ^:try tool-info-text]
+(def ^:private preview-close-button-size 18.0)
+
+(defn- close-preview-tooltip [localization-state keymap]
+  (localization-state
+    (localization/message
+      "command.tooltip"
+      {"command" (localization/message "command.scene.close-animation-preview")
```
vlaaad: You should either use `"command.scene.stop"` here (since `scene.stop` is the ID of the invoked command), or rename the `:scene.stop` command to `:scene.close-animation-preview`. This is important for consistency. In the future, we will show the command name localization in the Keymap panel of the preferences, and since the command is currently `scene.stop`, it will be presented with a localization key `command.scene.stop`, not `command.scene.close-animation-preview`

editor/resources/localization/en.editor_localization
```diff
@@ -922,6 +924,7 @@ command.debugger.stop = Stop Debugger
 command.debugger.toggle-breakpoint = Toggle Breakpoint
 command.debugger.toggle-breakpoint-enabled = Toggle Breakpoint Enabled
 command.debugger.edit-breakpoint = Edit Breakpoint
+command.scene.close-animation-preview = Close preview
```
vlaaad: In English translation, we use Title Case, i.e., "Close Preview".

editor/src/clj/editor/scene.clj
```diff
@@ -771,21 +838,28 @@
...
-        {:visible false}))))
+      (let [info-text
+            (if (and (string? tool-info-text)
+                     (pos? (count tool-info-text)))
+              tool-info-text
+              (let [scene-info-text (:info-text scene)]
+                (when (and (string? scene-info-text)
+                           (pos? (count scene-info-text)))
+                  scene-info-text)))
+            close-button (when-let [anim-data (and (seq active-updatable-ids)
```
vlaaad: Please don't use `seq` for non-empty checks.

vlaaad (changes requested):
This is much better, great work! We are getting close to merging it!

editor/src/clj/editor/scene.clj
AGulev: it's strange this command has "stop" name
it is closing preview, spacebar used to stop it

vlaaad: I think spacebar means "pause", i.e., it "stops", but does not rewind.

vlaaad (approved):
LGTM!

PR #11708, opened 2026-01-07
Add `zlib` module to editor scripts

### LINKED ISSUE #11676: Add zlib (deflate/inflate) support to Editor Scripts
Is your feature request related to a problem? Please describe (REQUIRED):

Currently, Editor Scripts in Defold do not have access to low-level compression utilities such as `deflate` and `inflate`. While the editor already provides `zip.pack()` and `zip.unpack()` functions, these operate only at the ZIP archive level and do not expose raw compression/decompression capabilities.

This limitation makes it difficult to directly read, manipulate, validate, or generate compressed data inside editor scripts, especially when working with custom asset pipelines or editor-side tooling.

Describe the solution you'd like (REQUIRED):

Expose the `zlib-lua` API (https://defold.com/ref/stable/zlib-lua/) to Editor Scripts, allowing functions such as `deflate` and `inflate` to be used inside the Defold editor scripting environment.

This would enable editor scripts to perform compression and decompression operations without relying exclusively on ZIP archives, providing a simpler and more flexible way to handle compressed data during editor workflows.

Describe alternatives you've considered (REQUIRED):

- Using `zip.pack()` and `zip.unpack()`, which only works with full ZIP archives and does not allow access to raw compressed data.
- Performing compression/decompression outside the editor using external tools, which breaks editor automation and complicates the workflow.
- Implementing custom compression logic in Lua, which is inefficient and unnecessary since Defold already includes zlib support at runtime.

Additional context (OPTIONAL):

The Defold editor already supports ZIP-related functionality, indicating that compression utilities are available internally. Making `zlib-lua` accessible to Editor Scripts would be a natural extension of the existing editor API and would significantly improve editor tooling and asset pipeline flexibility.

vlaaad:
Now, the editor scripts' Lua runtime has `zlib` module — with the same semantics as the engine's `zlib` module.

Fixes #11676

### DIFF
--- a/editor/src/clj/editor/editor_extensions.clj
+++ b/editor/src/clj/editor/editor_extensions.clj
@@ -36,6 +36,7 @@
             [editor.editor-extensions.tile-map :as tile-map]
             [editor.editor-extensions.ui-components :as ui-components]
             [editor.editor-extensions.zip :as zip]
+            [editor.editor-extensions.zlib :as zlib]
             [editor.error-reporting :as error-reporting]
             [editor.fs :as fs]
             [editor.future :as future]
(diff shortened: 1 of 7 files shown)


matgis (approved):
Ship it!

PR #11709, opened 2026-01-07
Added Safe Area related functionality

### LINKED ISSUE #11627: Investigate Android Warnings
1. 

Image

2.

Image

3.

Image

AGulev:
Safe Area support for Android and iOS phones is now available in the engine out of the box.
You can control how the GUI respects the safe area using `gui.safe_area_mode = none|long|short|both` in `game.project`, with the option to override this per GUI scene at runtime via `gui.gui.set_safe_area_mode(gui.SAFE_AREA_*)`.

CleanShot 2026-01-08 at 11 48 20@2x

This allows UI elements to stay within safe zones on devices with notches, while still using the full screen when no cutout is present (On platforms without cutouts, the inset values are returned as 0).
The new `window.get_safe_area()` API provides the current safe area rectangle along with inset values, making it possible to use this data in custom layout and positioning logic.

example

Fix https://github.com/defold/defold/issues/11627


engine/gui/src/gui_script.cpp
```diff
@@ -3671,6 +3671,34 @@ namespace dmGui
         return 0;
     }
 
+    /*# sets the safe area mode for the gui scene
+     *
+     * Sets how the safe area is applied to this gui scene.
+     *
+     * @name gui.set_safe_mode
```
AGulev: this one is needed so user can switch mode picked as default in `game.project` to something else: e.g. a background implemented as gui, or window shadowing.

AGulev: reviewing it I'm not sure anymore that making it shorter with `gui.set_safe_mode` instead of `gui.set_safe_area_mode` is a good idea

engine/gamesys/src/gamesys/scripts/script_window.cpp
```diff
@@ -282,6 +282,80 @@ static int GetSize(lua_State* L)
     return 2;
 }
 
+/*# get the safe area
+ *
+ * This returns the safe area rectangle (x, y, width, height) and the inset
+ * values relative to the window edges. On platforms without a safe area,
+ * this returns the full window size and zero insets.
```
AGulev, line 289: Usually we make a function unavailable on platforms it's not supported, but in this case it's calculations related, so not having safe area is actually 0, if user take this calculation into account like this, if in the future we will add new platforms (e.g. we will find a way to do it on html5 or macos) it will work out of the box

engine/platform/CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ else()
     else()
         set(_platform_glfw_sources "${PLAT_SRC_DIR}/platform_window_glfw.cpp")
         if (TARGET_PLATFORM MATCHES "arm64-ios|x86_64-ios")
-            list(APPEND _platform_glfw_sources "${PLAT_SRC_DIR}/platform_window_glfw_ios.cpp")
+            list(APPEND _platform_glfw_sources "${PLAT_SRC_DIR}/platform_window_glfw_ios.mm")
```
AGulev, line 33: changed to MM for ios/macos to be able to use ios/macos SDK APIs

engine/platform/src/platform_window_glfw_android.cpp
```diff
@@ -54,4 +58,58 @@ namespace dmPlatform
...
+            out->m_InsetLeft = inset_left;
+            out->m_InsetTop = inset_top;
+            out->m_InsetRight = inset_right;
+            out->m_InsetBottom = inset_bottom;
+            return true;
+        }
+
+        android_app* app = glfwGetAndroidApp();
+        // ARect content depends on window settings we swt in DefoldActivity.java
+        const ARect rect = app->contentRect;
```
AGulev, line 91: Rect is really a strange one. It returns insets only in some cases (depends on window parameters like cutout and immersive mode and so on), that's why I kept it only as a fallback, but now I think it would be clearer to just remove it... hmm...

external/glfw/java/com/dynamo/android/DefoldActivity.java
```diff
@@ -84,19 +88,43 @@ public enum GLFWKeyboardType {
...
-                | View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY);
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                window.setDecorFitsSystemWindows(false);
+
+                WindowInsetsController controller = window.getInsetsController();
+                if (controller != null) {
+                    controller.hide(WindowInsets.Type.systemBars());
+                    controller.setSystemBarsBehavior(
+                            WindowInsetsController.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE
+                    );
```
AGulev, line 101: `setSystemUiVisibility` is deprecated, so use new way of doing it now

external/glfw/java/com/dynamo/android/DefoldActivity.java
```diff
@@ -84,19 +88,43 @@ public enum GLFWKeyboardType {
...
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
             if (mDisplayCutout) {
                 WindowManager.LayoutParams layoutParams = getWindow().getAttributes();
-                layoutParams.layoutInDisplayCutoutMode = WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_SHORT_EDGES;
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                    // Android 11+ also has ALWAYS, and on Android 15 SHORT_EDGES is treated like ALWAYS for non-floating windows
+                    layoutParams.layoutInDisplayCutoutMode = WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_ALWAYS;
+                } else {
+                    // Android 9-10
+                    layoutParams.layoutInDisplayCutoutMode = WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_SHORT_EDGES;
```
AGulev, line 126: deprecated and now it's always `LAYOUT_IN_DISPLAY_CUTOUT_MODE_ALWAYS`, so I add this condition.

On the other hand taking into account that we add handling of this situation, maybe it's better to fully remove this else branch and keep only `WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_ALWAYS` ?

external/glfw/java/com/dynamo/android/DefoldActivity.java
```diff
@@ -258,6 +287,16 @@ public void onBackInvoked() {
         }
 
         nativeOnCreate(this);
+
+        View decorView = getWindow().getDecorView();
+        decorView.setOnApplyWindowInsetsListener(new View.OnApplyWindowInsetsListener() {
```
AGulev, line 292: there is no way to handle this even using NDK. So we handle it here and rise flag, that we have to update window.

external/glfw/lib/ios/app/ViewController.m
```diff
@@ -212,6 +212,7 @@ float _glfwPlatformGetDisplayScaleFactor()
     {
         _glfwPlatformAcquireAuxContextOpenGL();
     }
+    return 0;
```
AGulev, line 215: just to be able to compile it using new xcode and compiler

engine/gui/src/gui_script.cpp
britzl: I agree. set_safe_area_mode() is better

external/glfw/java/com/dynamo/android/DefoldActivity.java
britzl: Yeah, why not!

britzl (dismissed):

engine/gui/src/gui_script.cpp
AGulev: fixed

external/glfw/java/com/dynamo/android/DefoldActivity.java
AGulev: ah, no. it's availible only from API 30 (edge from 28). I keep it as is

britzl (approved):

PR #11715, opened 2026-01-08
Do not rename items on double-click in the outline (keep only F2 for renaming)

### LINKED ISSUE #11396: Turn on/off double click in Editor Outline to rename stuff.
Editor possibility to detect two delayed clicks for changing name/id in Editor Outline for anything is causing a lot of troubles. It feels too unbearable and detects almost all my clicks as those those delayed double clicks, when I really rarely need to change the id, I rather click there with an intention to select, that is too many times misinterpreted by the Editor. I like the option to rename things in Outline quickly itself, but I just prefer using F2 shortcut for this and I think it's enough.

Have a possibility to turn on/off this option among preferences.

Resign of double delayed click for renaming.

https://github.com/user-attachments/assets/8861d096-5755-4cd2-a752-175b6c16c21e

AGulev:
Fix https://github.com/defold/defold/issues/11396

### DIFF
--- a/editor/src/clj/editor/outline_view.clj
+++ b/editor/src/clj/editor/outline_view.clj
@@ -595,9 +595,8 @@
 
 (defn- cancel-rename!
   [^TreeView tree-view]
-  (when-let [future (ui/user-data tree-view ::future-rename)]
-    (ui/cancel future)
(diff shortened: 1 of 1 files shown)


vlaaad (approved):
Works great! Ship it!

PR #11716, opened 2026-01-08
Resolve lein init SHA argument as git-ref

matgis:
You can now use a shortened commit SHA, as well as local tags and branch names for the SHA argument to `lein init`.

### DIFF
--- a/editor/tasks/leiningen/init.clj
+++ b/editor/tasks/leiningen/init.clj
@@ -55,7 +55,7 @@
       "dynamo-home"     nil ; for symmetry
       "archived-stable" (sha-from-version-file)
       "archived"        (sha-from-ref "HEAD")
-      version)))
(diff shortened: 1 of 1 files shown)


editor/tasks/leiningen/init.clj
```diff
@@ -55,7 +55,7 @@
       "dynamo-home"     nil ; for symmetry
       "archived-stable" (sha-from-version-file)
       "archived"        (sha-from-ref "HEAD")
-      version)))
+      (sha-from-ref version))))
```
matgis, line 58: This will use `git rev-parse` to resolve the argument into a full commit SHA. If the `version` string does not resolve to an existing commit, throws an exception.

JosephFerano (approved):
Works! The exception it throws is pretty clear what the problem is, so LGTM.

PR #11721, opened 2026-01-08
Strict evaluation-context scope checks in the editor

matgis:
This PR introduces a flag that turns on `evaluation-context` scope violations in the editor. Any calls to `(g/now)` or `(g/make-evaluation-context)` from inside an auto `evaluation-context` scope will be logged or trip an `assert`, depending on the setting of the `defold.graph.strict-evaluation-context-scopes` JVM property. Valid settings are `throw`, `log`, or `log-once`. If not defined, the generated code will not perform any checks or incur any additional performance overhead.

This flag is currently disabled by default, but can be set to `log-once` by enabling the `strict-ec-scopes` `lein` profile.

Alongside the addition of the flag, a lot of `evaluation-context` scope violations were fixed as part of this PR, but some work still remains.

Also of interest is the addition of the `g/let-ec` macro. It works like a regular `let` expression, but creates and binds an `evaluation-context` that can be used by the `let` bindings. The `evaluation-context` is then closed and merged into the system cache before the body of the `g/let-ec` expression is executed. It is a convenient alternative to `g/with-auto-evaluation-context` when you need to write code in a gather-then-act style.

Counterpart in extension:
* defold/extension-texturepacker#23


editor/src/clj/editor/code/script.clj
```diff
@@ -226,13 +226,14 @@
...
   (when-not deleted?
-    (let [project (project/get-project _node-id)
-          workspace (project/workspace project)
+    (let [basis (:basis _evaluation-context)
+          project (project/get-project basis _node-id)
+          workspace (project/workspace project _evaluation-context)
           prop-kw (properties/user-name->key name)
           prop-type (script-compilation/script-property-type->property-type type)
-          edit-type (script-compilation/script-property-edit-type workspace prop-type resource-kind type)
+          edit-type (script-compilation/script-property-edit-type workspace prop-type resource-kind type _evaluation-context)
```
matgis, line 236: We have a couple of places where we bypass the graph invalidation system when accessing resource type information from the `workspace`. We should probably just expose an output on the `Workspace` node and have this information travel through graph connections.

editor/src/clj/editor/code/script_compilation.clj
```diff
@@ -69,7 +69,7 @@
 (def resource-kind->workspace->extensions
   "Declares which file extensions are valid for different kinds of resource
   properties. This affects the Property Editor, but is also used for validation."
-  {"atlas"        #(workspace/resource-kind-extensions % :atlas)
+  {"atlas"        #(workspace/resource-kind-extensions %1 :atlas %2)
```
matgis, line 72: We now call this with an `evaluation-context`.

editor/src/clj/editor/code/view.clj
```diff
@@ -4271,13 +4266,14 @@
...
-(defn- focus-view! [view-node opts]
+(defn- focus-view! [view-node opts done-fn]
   (.requestFocus ^Node (g/node-value view-node :canvas))
   (when-some [cursor-range (:cursor-range opts)]
     (set-properties! view-node :navigation
                      (data/select-and-frame (get-property view-node :lines)
                                             (get-property view-node :layout)
-                                            cursor-range))))
+                                            cursor-range)))
+  (done-fn))
```
matgis, line 4276: The `focus-view!` function is supplied as the `:focus-fn` when registering the view type with the `workspace`. We now call the `:focus-fn` with a `done-fn` continuation, because sometimes we need to defer steps until the next JavaFX tick. You'll see this in the `:focus-fn` for `SceneView`, which will call the `done-fn` in a `run-later` block. Here, we can just call it directly, though.

editor/src/clj/editor/editor_extensions/commands.clj
```diff
@@ -82,17 +83,23 @@
...
-                                    (handler/adapt-every outline/OutlineNode)
-                                    (node-ids->lua-selection q))]
-               (cont assoc :selection res))))
+             (let [evaluation-context (or (:evaluation-context env) (g/make-evaluation-context))]
+               (when-let [res (some-> (:selection env)
+                                      (handler/adapt-every outline/OutlineNode evaluation-context)
+                                      (node-ids->lua-selection q))]
+                 (when-not (:evaluation-context env)
+                   (g/update-cache-from-evaluation-context! evaluation-context))
+                 (cont assoc :selection res)))))
```
matgis, line 92: I needed an `evaluation-context` for these two cases. I copied how it was done elsewhere.

editor/src/clj/editor/app_view.clj
```diff
@@ -210,6 +210,26 @@
...
+  ([^Tab tab]
+   (g/with-auto-evaluation-context evaluation-context
+     (tab->resource-node tab evaluation-context)))
+  ([^Tab tab evaluation-context]
+   (some-> tab
+           tab->view-id
+           (g/node-value :view-data evaluation-context)
+           second
+           :resource-node)))
+
```
matgis: Moved up in the file. There were a few places where we were accessing the user-data directly. Now everything goes through these functions.

editor/src/clj/editor/app_view.clj
```diff
@@ -317,15 +336,16 @@
   (output open-dirty-views g/Any :cached (g/fnk [open-dirty-views] (into #{} (keep #(when (second %) (first %))) open-dirty-views)))
   (output hidden-renderable-tags types/RenderableTags (gu/passthrough hidden-renderable-tags))
   (output hidden-node-outline-key-paths types/NodeOutlineKeyPaths (gu/passthrough hidden-node-outline-key-paths))
+  (output active-tab-pane TabPane (g/fnk [^Tab active-tab ^SplitPane editor-tabs-split]
+                                    (or (some-> active-tab .getTabPane)
+                                        (first (.getItems editor-tabs-split)))))
```
matgis, line 324: `active-tab-pane` is no longer a property, but an output dependent on the state of the `active-tab` property.

editor/src/clj/editor/app_view.clj
```diff
@@ -1574,8 +1577,8 @@
         (remove-tab! tab-pane tab)))))
 
 (defn- editor-tab-pane
-  "Returns the editor TabPane that is above the Node in the scene hierarchy, or
-  nil if the Node does not reside under an editor TabPane."
+  "Returns the closest editor TabPane at or above the Node in the scene
+  hierarchy, or nil if the Node is not an editor TabPane or resides under one."
```
matgis, line 1600: Clarified comment to make it clearer that this will return the `TabPane` if called with the editor `TabPane` itself.

editor/src/clj/editor/app_view.clj
```diff
@@ -396,16 +414,6 @@
...
-(defn- on-selected-tab-changed! [app-view app-scene tab resource-node view-type]
-  (g/transact
-    (concat
-      (replace-connection resource-node :node-outline app-view :active-outline)
-      (if (= :scene view-type)
-        (replace-connection resource-node :scene app-view :active-scene)
-        (disconnect-sources app-view :active-scene))
-      (g/set-property app-view :active-tab tab)))
-  (ui/user-data! app-scene ::ui/refresh-requested? true))
-
```
matgis, line 408: The editor tab situation was quite complex before, and has been simplified a bit in this PR. There were a lot of complex interactions between JavaFX controls and event handlers. We have now split the restoration of the open editor tabs stored in the preferences into a gather and an act phase. First, we gather all the stuff we need from the preferences and the graph using a single `evaluation-context`, then we act on the information to apply the changes to the view hierarchy.

The reason is that we want to close the `evaluation-context` and commit any cached values from it back into the system graph before we cause further changes to the graph from triggering UI event handlers. Changes like that are safe to do, but are suboptimal since they will cause any affected outputs we have cached values for in the `evaluation-context` to be thrown away rather that being committed into the system cache.

In line with this, the events we observe have also been simplified a bit. The goal is to minimize updates to the graph caused by changes to the UI.

editor/src/clj/editor/app_view.clj
```diff
@@ -1989,16 +1992,6 @@
...
-(defn- tab->resource-node [^Tab tab]
-  (some-> tab
-    (ui/user-data ::view)
-    (g/node-value :view-data)
-    second
-    :resource-node))
-
-(defn- tab->view-type [^Tab tab]
-  (some-> tab (ui/user-data ::view-type) :id))
-
```
matgis, line 2001: Moved up in the file.

editor/src/clj/editor/app_view.clj
```diff
@@ -2031,48 +2030,53 @@
...
     (.setTitle stage (ui/make-title))
-    (let [editor-tab-pane (TabPane.)
-          keymap (keymap/from-prefs prefs)
+    (.add (.getItems editor-tabs-split) editor-tab-pane)
+    (let [keymap (keymap/from-prefs prefs)
           app-view (first (g/tx-nodes-added (g/transact (g/make-node view-graph AppView
                                                                      :stage stage
                                                                      :scene app-scene
                                                                      :editor-tabs-split editor-tabs-split
-                                                                     :active-tab-pane editor-tab-pane
```
matgis, line 2068: Property removed, now an output dependent on the `active-tab` property.

editor/src/clj/editor/app_view.clj
```diff
@@ -2119,11 +2123,15 @@
     (.addAll  (.getChildren info-panel) (Arrays/asList (into-array Node [left-label spacer right-link])))
     info-panel))
 
-(declare open-resource)
+(declare open-resource!)
```
matgis, line 2149: Renamed to reflect it is mutating.

editor/src/clj/editor/app_view.clj
```diff
@@ -3133,25 +3207,3 @@
...
-    (let [resource (workspace/find-resource (dev/workspace) "/scripts/knight_copy.script" ec)]
-      (resource/openable-resource? resource)))
-  (ui/run-later
-    (g/with-auto-evaluation-context ec
-      (clojure.pprint/pprint
-        (open-tabs-from-prefs (dev/app-view) (dev/prefs) (dev/localization) (dev/workspace) (dev/project)
-                              '([["/scripts/game.script" :code] ["/scripts/knight.script" :code]]
-                                [["/scripts/utils.lua" :code]])
-                              ec))))
-  ,)
```
matgis, line 3160: I think if we want to have these sort of dev-time code blocks in the code base, we need to make them useful to everyone. This one appears to depend on a local project, so I don't consider it generally useful.

editor/src/clj/internal/node.clj
```diff
@@ -1350,20 +1351,8 @@
...
-  (boolean (-> description :output output :annotations argument :raw)))
-
-(defn- desc-try-argument? [description output argument]
-  {:pre [(keyword? output) (keyword? argument)]}
-  (boolean (-> description :output output :annotations argument :try)))
-
-(defn- desc-unsafe-argument? [description output argument]
-  {:pre [(keyword? output) (keyword? argument)]}
-  (boolean (-> description :output output :annotations argument :unsafe)))
-
```
matgis, line 1364: Now unused. We supply `argument-annotations` to `fnk-argument-form` instead, which is the same as `(-> description :output output :annotations)` above for outputs, but different for property dynamics. This change was made to allow for `^:unsafe _evaluation-context` use from property dynamics.

editor/src/dev/dev.clj
```diff
@@ -230,7 +231,7 @@
   (prefs/project (workspace/project-directory (workspace))))
 
 (defn localization []
-  (some #(-> % :env :localization) (ui/contexts (ui/main-scene))))
+  (some #(-> % :env :localization) (ui/contexts (ui/main-scene) true)))
```
matgis, line 235: The `all-selections?` argument to `ui/contexts` have been made explicit now that it takes an optional `evaluation-context`.

editor/test/integration/undo_test.clj
```diff
@@ -71,9 +71,9 @@
   (property counter g/Any)
 
   (output outline g/Any :cached
-          (g/fnk [_node-id outline]
-                 (swap! (g/node-value _node-id :counter) inc)
-                 outline)))
```
matgis, line 76: We were doing some messed-up stuff in the tests before, man. 🤷

editor/test/internal/copy_paste_test.clj
```diff
@@ -284,8 +284,8 @@
 
 (g/defnode SetterNode
   (property producer g/NodeID
-    (value (g/fnk [_node-id]
-             (g/node-feeding-into _node-id :value)))
```
matgis, line 288: Again, messed-up stuff going on in the tests here. Not changing it, though. 🤷

matgis:
@JosephFerano This PR includes some changes to the restoration of open tabs from preferences, specifically it splits up the work into a `plan` phase and a `perform` phase to facilitate focused `evaluation-context` use. I also tried to simplify the event flow a bit. Could I ask you to take a look at these changes in `app_view.clj` and `recent_files.clj` in particular?

editor/src/clj/dynamo/graph.clj
```diff
@@ -195,6 +278,12 @@
         (aset-long tps-counts 0 0)))
     tps-counts))
 
+(defn eager-tx-data
+  "Given a sequence of possibly nested sequences of transaction steps, force its
+  evaluation into a flat sequence of realized transaction steps."
+  [txs]
+  (doall (filter some? (flatten txs))))
```
vlaaad: Why not `(into [] coll/flatten-xf txs)`? What's the purpose of this fn? Wouldn't it cause huge spikes in memory use when opening big projects?

Judging by how it's used, I'd prefer if we created non-lazy sequences of transaction steps instead of making them lazily and then realizing them using this function. I'd consider every use of this function a code smell... If you don't want laziness, create a vector.

editor/src/clj/dynamo/graph.clj
```diff
@@ -985,16 +1073,38 @@
 
 (defmacro with-auto-evaluation-context [ec & body]
   `(let [~ec (make-evaluation-context)
-         result# (do ~@body)]
+         result# (do-strict-evaluation-context-scope-body ~@body)]
      (update-cache-from-evaluation-context! ~ec)
      result#))
 
+(defmacro let-ec [binding-forms & body]
```
vlaaad: Maybe add a docstring with an example use? It would be particularly useful to document where the `evaluation-context` symbol is available (i.e., in the let bindings block, but not after it)

editor/src/clj/dynamo/graph.clj
```diff
@@ -985,16 +1073,38 @@
...
+         (even? (count binding-forms))]}
+  (if strict-evaluation-context-scopes
+    (let [binding-syms (mapv first (partition 2 binding-forms))]
+      (list*
+        `let
+        [binding-syms
+         `(with-auto-evaluation-context ~'evaluation-context
+            (let ~binding-forms
+              ~binding-syms))]
+        body))
```
vlaaad: It looks like we can avoid creating a vector for destructuring here if we avoid `with-auto-evaluation-context`, and instead make a context and then update the cache from the context, as we do below.

editor/src/clj/editor/editor_extensions/commands.clj
vlaaad: Looks good!

editor/src/clj/editor/particlefx.clj
```diff
@@ -1225,7 +1226,7 @@
                :command :edit.add-secondary-embedded-component
                :user-data {:modifier-type type}})
             mod-types))))
-
+halt-when
```
vlaaad: huh?

editor/test/integration/undo_test.clj
vlaaad: LOL

editor/src/clj/dynamo/graph.clj
```diff
@@ -985,16 +1073,38 @@
...
+    (list*
+      `let
+      (vec
+        (concat
+          ['evaluation-context `(make-evaluation-context)]
+          binding-forms
+          ['_ `(update-cache-from-evaluation-context! ~'evaluation-context)
+           'evaluation-context nil]))
```
vlaaad: Hmm, ideally, `evaluation-context` should not even be present in the local scope of the `let` body, though I'm not sure it's possible to do so without throw-away vectors for destructuring...

editor/src/clj/editor/recent_files.clj
```diff
@@ -112,15 +112,15 @@
...
-      (ui/user-data :editor.app-view/view)
-      (g/node-value :view-data)
-      second
-      :resource))
+   :post [(or (nil? %)
+              (resource/resource? %))]}
+  (some-> (ui/user-data tab :editor.app-view/view)
+          (g/node-value :view-data)
+          second
+          :resource))
```
vlaaad: How did it become nilable?

editor/src/clj/editor/gui.clj
```diff
@@ -4196,26 +4205,36 @@
...
+      "material"
+      (g/let-ec [[target-node name] (drop-target-node-id+unique-name gui-scene :materials-node base-name evaluation-context)]
+        (add-material gui-scene target-node resource name))
 
-      :else
-      nil)))
+      ;; else
+      (g/let-ec [atlas-exts (workspace/resource-kind-extensions workspace :atlas evaluation-context)
+                 target-node+name (when (contains? atlas-exts ext)
+                                    (drop-target-node-id+unique-name gui-scene :textures-node base-name evaluation-context))]
```
vlaaad, line 4235: Seems to be a bug here... Workspace's `resource-kind-extensions` output looks like this:
```
{:atlas ["atlas" "tilesource" "tpatlas"]}
```
This means `atlas-exts` is `["atlas" "tilesource" "tpatlas"]`, and `contains?` checks for index, not value inclusion.

editor/src/clj/editor/app_view.clj
```diff
@@ -2009,16 +2002,23 @@
...
   (-> tab-pane
       (.getSelectionModel)
-      (.selectedItemProperty)
+      (.selectedIndexProperty)
       (^[ChangeListener] ObservableValue/.addListener
-        (fn [_observable _old-val new-val]
-          (recent-files/save-tab-selections prefs app-view)
-          (on-selected-tab-changed! app-view app-scene new-val (tab->resource-node new-val) (tab->view-type new-val)))))
+        (fn [_observable _old-val _new-val]
+          (recent-files/save-tab-selections prefs app-view))))
```
vlaaad: Not sure if the problem is exactly here... On this branch, if I have gamepads and render files open and switch between them, the outline doesn't update:

https://github.com/user-attachments/assets/14c83c9e-40b8-4c60-9efc-12e9abe1620d

This problem does not exist in `dev`. Probably related to the fact that `active-tab-pane` is now an output and not a property?

editor/src/clj/dynamo/graph.clj
```diff
@@ -985,16 +1073,38 @@
...
-         result# (do ~@body)]
+         result# (do-strict-evaluation-context-scope-body ~@body)]
      (update-cache-from-evaluation-context! ~ec)
      result#))
 
+(defmacro let-ec [binding-forms & body]
+  {:pre [(vector? binding-forms)
+         (even? (count binding-forms))]}
+  (if strict-evaluation-context-scopes
+    (let [binding-syms (mapv first (partition 2 binding-forms))]
```
vlaaad: This code expects that odd items in the let binding vector are symbols, but `let` also allows using destructuring, e.g., `(let [{:keys [a b]} (get-map...)] ...)`. It seems that code that uses destructuring instead of symbols will work when `strict-evaluation-context-scopes` is off, but will break when it's on...

editor/src/clj/dynamo/graph.clj
```diff
@@ -230,8 +319,10 @@
   ([opts txs]
    (when *tps-debug*
      (send-off tps-counter tick (System/nanoTime)))
-   (let [transaction-context (make-transaction-context opts)
-         tx-result (it/transact* transaction-context txs)]
+   (let [txs (cond-> txs strict-evaluation-context-scopes eager-tx-data)
```
vlaaad, line 336: Why do we use `eager-tx-data` here? It changes the semantics of the executed code when `strict-evaluation-context-scopes` is on! I thought we wanted to keep tx steps lazy to reduce memory load, e.g., when opening big projects?

vlaaad (changes requested):
I found a couple of bugs. Overall, I'm very happy with how you tightened up our use of evaluation contexts. 

I'm only worried that we introduced the `eager-tx-data` function. It looks like a band-aid or something... Perhaps I'm not understanding its use? But to me, it seems that transaction steps should probably stay lazy to reduce memory load. Or they should be created as eager vectors. But I'd say they shouldn't be created lazily, to then be immediately converted to eager sequences. I also find it weird that the tx sequence is always realized only when we use strict evaluation context scopes — it's supposed to be a helper that detects incorrect use, but here it changes the semantics of when evaluation context used inside lazy tx sequences are created, so to me it seems it will report the wrong places, no?

editor/src/clj/dynamo/graph.clj
matgis: We do want to keep the tx-steps lazy, but when strict `evaluation-context` scope checks are enabled we also want to verify that, for example, `g/node-value` isn't being used to evaluate an out-of-transaction value from something like a property setter or override `:traverse-fn`. However, it is perfectly valid to evaluate on the out-of-transaction basis to generate the transaction steps themselves. Since we use a lot of lazy sequences in functions that return transaction steps, we must realize the lazy sequence outside of the `do-strict-evaluation-context-scope-body` block to avoid false positives. I see how this can appear strange at first glance. Thanks for bringing it to my attention. I'll add a comment to clarify things!

editor/src/clj/dynamo/graph.clj
matgis: We do realize lazy sequences of transaction steps into eager sequences in a few places. I thought it would be preferable to use a single function that could be easily found to make it explicit where this is happening.

I'm not sure I agree it is a code smell that we need to do this. When your function is supposed to return a sequence of transaction steps, it may often be preferable to realize the sequence instead of holding on to the resources required to produce it.

So the intent was to have a function for those places where we would previously create a vector of flattened transaction steps. But I also wanted to do this in `g/transact` when strict evaluation-context scope checks are enabled (as I explain below), and was worried realizing everything into a huge vector would be problematic. But perhaps it is better than realizing the seq. 🤔

I think I'll keep the function, but change it to use `(into [] coll/flatten-xf txs)` instead, and see where that takes us.

editor/src/clj/dynamo/graph.clj
matgis: Good idea!

editor/src/clj/dynamo/graph.clj
matgis: Ohh, good catch! I'll fix!

editor/src/clj/editor/app_view.clj
JosephFerano: I was able to replicate

JosephFerano (dismissed):
I tested a few "restore tabs" scenarios and everything seems to work. I went through the code and things are easier to follow with the plan/perform separation and case forms. Seems like a great application of DoD. I took notes 😄

editor/src/clj/dynamo/graph.clj
matgis: The problem in this case is that we need to have the `binding` scope that rebinds `g/now` and `g/make-evaluation-context` around the `let` bindings. The `binding` macro restores the bindings in a `finally` clause, which we very much want to happen, so I don't think we can avoid returning a vector and destructuring, and we might as well use `g/with-auto-evaluation-context` here.

editor/src/clj/dynamo/graph.clj
matgis: Good point. I fixed it by using a `gensym` for the `evaluation-context` and replacing all references to `evaluation-context` in the `init-exprs` in the `let` bindings.

editor/src/clj/editor/particlefx.clj
matgis: Oops. I have this bad habit of typing symbols into files so I can navigate to their implementation. I guess I was reviewing a different PR that used a core function I hadn't encountered yet. 😅

editor/src/clj/editor/recent_files.clj
matgis: I ran into this during testing. When you use the `:window.tab.swap-with-other-group` command, it temporarily inserts a dummy `Tab` in place of the `Tab` it is moving over. This is later replaced with the `Tab` that is swapped in from the other tab group. But since we respond to list change events, there is a moment where we see the dummy tab in the list. On top of that, we explicitly set this specific `ui/user-data` key to `nil` when closing tabs, so it was always nilable. I've gone over the usages and attempted to handle the possible `nil` cases now.

editor/src/clj/editor/gui.clj
matgis: Oh, you're right! Nice catch!

editor/src/clj/editor/app_view.clj
matgis: Yep, me as well. Seems I was a bit too aggressive in my efforts to minimize updates. I've had another stab at it now and it appears to work again. Sorry for not testing properly.

editor/src/clj/editor/particlefx.clj
vlaaad: Hehe me too!

editor/src/clj/dynamo/graph.clj
```diff
@@ -1076,28 +1087,74 @@
...
+      (partition-all 2)
+      (mapcat (fn [[binding-form init-expr]]
+                ;; Replace all references to `evaluation-context` in the
+                ;; init-exprs with our generated symbol. Using a generated
+                ;; symbol ensures the evaluation-context cannot be
+                ;; referenced from the body after it has been closed.
+                (pair binding-form
+                      (walk/postwalk-replace
+                        {'evaluation-context evaluation-context-sym}
+                        init-expr)))))
```
vlaaad: I think we can make the relaxed form more reliable if we avoid walking Clojure code, since it might replace too much:
```clj
(g/let-ec [foo '(evaluation-context)]
  foo)
=> (evaluation-context__140052)
```
I believe it should be possible by transforming e.g. this:
```clj
(g/let-ec [a b...]
  c...)
```
into this:
```
(let [ec# (make-evaluation-context)
      a (let [~'evaluation-context ec#]
          b...)]
  (update-cache-from-evaluation-context! ec#)
  c...)
```

editor/src/clj/editor/app_view.clj
```diff
@@ -403,16 +388,50 @@
...
+             ^SplitPane editor-tabs-split (g/node-value app-view :editor-tabs-split evaluation-context)
+             ^Scene app-scene (g/node-value app-view :scene evaluation-context)
+             editor-tab-pane (some-> new-active-tab .getTabPane)
+             new-resource-node-id (some-> new-active-tab (editor-tab/resource-node-id evaluation-context))
+
+             tx-data
+             (when (and (not= old-active-tab new-active-tab)
+                        (some-> editor-tab-pane .isFocusWithin))
+               (g/eager-tx-data
+                 (concat
```
vlaaad, line 417: Why should we use `(g/eager-tx-data (concat ...))` instead of using `[...]` here? Similarly, we could use `[...]` instead of `(concat ...)` in `replace-connection`...

editor/src/clj/editor/app_view.clj
```diff
@@ -346,65 +351,87 @@
...
+  (g/let-ec [basis (:basis evaluation-context)
+             ^Tab old-active-tab (g/node-value app-view :active-tab evaluation-context)
+             ^SplitPane editor-tabs-split (g/node-value app-view :editor-tabs-split evaluation-context)
+             ^Scene app-scene (g/node-value app-view :scene evaluation-context)
+             editor-tab-pane (some-> new-active-tab .getTabPane)
+             new-resource-node-id (some-> new-active-tab (editor-tab/resource-node-id evaluation-context))
+
+             tx-data
+             (when (and (not= old-active-tab new-active-tab)
+                        (some-> editor-tab-pane .isFocusWithin))
```
vlaaad: Not sure if this place is the source of the problem, but when I close the last open tab, the outline keeps showing it:

https://github.com/user-attachments/assets/88b79436-a959-4e23-b67f-ab1ca62daa0a

editor/src/clj/editor/editor_tab.clj
```diff
@@ -0,0 +1,98 @@
...
+              (resource/resource? %))]}
+  (some-> (resource-node-id editor-tab evaluation-context)
+          (g/node-value :resource evaluation-context)))
+
+(defn prefs-data
+  "Returns the preference data associated with the editor tab, if any."
+  [^Tab editor-tab evaluation-context]
+  (when-let [view-type-id (view-type-id editor-tab)]
+    (when-let [resource (resource editor-tab evaluation-context)]
+      (let [proj-path (resource/proj-path resource)]
```
vlaaad, line 96: Minor nitpick, feel free to ignore: this `let` seems really unnecessary.

editor/src/clj/dynamo/graph.clj
vlaaad: Can you give me a code example of an issue that gets caught by eagerly evaluating txs here outside of `do-strict-evaluation-context-scope-body` block?

vlaaad (changes requested):
I found a bug with an active tab not updating when closing the last tab, see the comment. I also think we can easily make `g/let-ec` more reliable by avoiding walking the code.

editor/src/clj/editor/app_view.clj
matgis: I'm pretty sure it is, yeah. Argh, it never ends. I regret trying to clean this up. I'll fix.

editor/src/clj/editor/app_view.clj
matgis: I didn't want to keep a reference to the `basis` from the hidden `evaluation-context` into the body of the `g/let-ec` expression. I think it's both cleaner and clearer to realize the `tx-data` into a flat vector eagerly. The `disconnect-sources` function returns a lazy sequence after all, and this way will also work as expected if future changes require us to supply the `evaluation-context` to the functions we call.

vlaaad: Makes sense! 👍

editor/src/clj/dynamo/graph.clj
matgis: I like it! I'm going to give it a try.

editor/src/clj/editor/editor_tab.clj
matgis: It is absolutely unnecessary, but I think it clarifies the shape of the returned data.

editor/src/clj/dynamo/graph.clj
matgis: Well, the problem is kind of the opposite. Unless we realize the sequence, we get all sorts of issues reported from various places in `:load-fns` that implicitly invoke `g/now` or `g/make-evaluation-context` in order to produce the transaction steps. Most of these functions return lazy sequences, and around all that we are iterating over an eduction that calls the `:load-fn` for each resource. It's not a problem that `g/now` is called to produce transaction steps, but unless we realize the entire sequence beforehand, `g/now` might be called as an effect of producing more of the lazily evaluated sequence, and it would be doing so from inside the `strict-evaluation-context-scope` block. This will be reported as a violation when it is really not. What *is* a violation, though, is if someone is calling `g/now` while we are performing the transaction steps themselves. Then you're looking at an out-of-transaction state from within the transaction. These are the errors I want to catch, and realizing the sequence allows me to turn on the `strict-evaluation-context-scope` checks around `it/transact*`.

editor/src/clj/dynamo/graph.clj
```diff
@@ -59,11 +59,99 @@
...
+;;
+;; We're still in the process of fixing all the various places where this rule
+;; is violated. When complete, we should enable strict evaluation-context scope
+;; checks by default.
+;;
+;; Until that day, you can turn on these checks using the :strict-ec-scopes lein
+;; profile when running tasks: `lein with-profile +strict-ec-scopes <task>`.
+(def ^:private strict-evaluation-context-scopes
+  (let [strict-evaluation-context-scopes-setting (and *assert* (some-> (System/getProperty "defold.graph.strict-evaluation-context-scopes") keyword))]
+    (assert #{nil :log :log-once :throw} strict-evaluation-context-scopes-setting)
```
vlaaad: This assertion does nothing (tests that a set is truthy, which is always true)

editor/src/clj/dynamo/graph.clj
```diff
@@ -985,16 +1087,92 @@
...
+    (list*
+      `let
+      [binding-forms
+       (list
+         `with-auto-evaluation-context 'evaluation-context
+         (list
+           `let
+           private-bindings
+           init-expr-results))]
+      body)))
```
vlaaad, line 1115: I think I found an issue with this macro. It performs destructuring twice, which can trigger unexpected side effects. If I understand it right, here is how it happens twice: first, explicitly, in your `destructure` call, and second time, in the `let`.

You can reproduce the issue using this form:
```
(g/let-ec [{:keys [foo], :or {foo (println "calc-or")}} {}])
```
When the `strict-ec-scopes` profile is enabled, `calc-or` is printed twice. Otherwise, it's printed once as expected.

vlaaad (dismissed):
I found 2 small issues with the code, but otherwise everything works perfectly fine! Not sure how important the double-destructuring problem is, if the issue will only appear in the dev move, but it might lead to subtle bugs.

editor/src/clj/dynamo/graph.clj
matgis: Doh! Fixing.

editor/src/clj/dynamo/graph.clj
matgis: You're right. I'm fixing it and adding a test.

PR #11728, opened 2026-01-09
Fix memory leak / fragmentation in PreStepEmscripten

meehaul:
Fixed a memory leak on every frame with Emscripten builds, resulting in significant heap fragmentation and usage.

### Technical changes

The PreStepEmscripten function was requesting argc/argv from dmEngineGetResult on every frame, though these values are only used when handling a reboot. When argc > 0, this caused unnecessary malloc + strdup allocations that were never freed.

The other various call sites have it freed correctly, so this just impacted web builds.

Tested by monitoring the state of the Emscripten heap, observed that this fix prevents the leaky allocation.


JCash (approved):

JosephFerano:
Just a heads up, `engine_loop.cpp` got converted to CRLF

PR #11738, opened 2026-01-13
Add support for ASAN and UBSAN for wasm builds

AGulev:
This PR adds support for compiler flags needed for ASAN and UBSAN and makes it possible to run tests for modules built with `cmake`.
Also, a small fix to skip tests for host platforms libs e.g. to be able to build wasm we build `dlib`, and before this fix this dlib build run tests as well as target platform tests later.

But tests will need to be fixed before running them by default or in nightly builds. For now there are many fails.

### DIFF
--- a/build_tools/waf_dynamo.py
+++ b/build_tools/waf_dynamo.py
@@ -623,13 +623,17 @@ def default_flags(self):
         emflags_compile = zip(['-s'] * len(emflags_compile), emflags_compile)
         emflags_compile =[j for i in emflags_compile for j in i]
 
+        initial_memory = 'INITIAL_MEMORY=33554432'
+        if Options.options.with_asan and target_arch == 'wasm':
+            initial_memory = 'INITIAL_MEMORY=67108864'
+
         emflags_link = [
             'DISABLE_EXCEPTION_CATCHING=1',
             'ALLOW_UNIMPLEMENTED_SYSCALLS=0',
             'EXPORTED_RUNTIME_METHODS=["ccall","UTF8ToString","callMain","HEAPU8","stringToNewUTF8"]',
             'EXPORTED_FUNCTIONS=_main,_malloc,_free',
             'ERROR_ON_UNDEFINED_SYMBOLS=1',
-            'INITIAL_MEMORY=33554432',
+            initial_memory,
             'MAX_WEBGL_VERSION=2',
             'GL_SUPPORT_AUTOMATIC_ENABLE_EXTENSIONS=0',
             'IMPORTED_MEMORY=1',
@@ -669,6 +673,8 @@ def default_flags(self):
             emflags_link += ['PTHREAD_POOL_SIZE=1']
 
         if 'wasm' == target_arch:
+            if Options.options.with_ubsan:
+                emflags_link += ['ASSERTIONS=1']
             emflags_link += ['WASM=1', 'ALLOW_MEMORY_GROWTH=1']
             if int(opt_level) < 2:
                 flags += ['-gseparate-dwarf', '-gsource-map']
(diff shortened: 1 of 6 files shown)


build_tools/waf_dynamo.py
```diff
@@ -623,13 +623,17 @@ def default_flags(self):
         emflags_compile = zip(['-s'] * len(emflags_compile), emflags_compile)
         emflags_compile =[j for i in emflags_compile for j in i]
 
+        initial_memory = 'INITIAL_MEMORY=33554432'
+        if Options.options.with_asan and target_arch == 'wasm':
+            initial_memory = 'INITIAL_MEMORY=67108864'
```
AGulev, line 628: for ASAN it's better to use more memory initially

build_tools/waf_dynamo.py
```diff
@@ -669,6 +673,8 @@ def default_flags(self):
             emflags_link += ['PTHREAD_POOL_SIZE=1']
 
         if 'wasm' == target_arch:
+            if Options.options.with_ubsan:
+                emflags_link += ['ASSERTIONS=1']
```
AGulev, line 677: More info is available with ASSERTIONS

engine/dlib/src/test/wscript
```diff
@@ -133,7 +136,7 @@ def build(bld):
     skip_threads = False
     if bld.env.PLATFORM in ('js-web', 'wasm-web', 'wasm_pthread-web'):
         skip_http = True
-        if bld.env in ('js-web', 'wasm-web'):
+        if bld.env.PLATFORM in ('js-web', 'wasm-web'):
```
AGulev, line 139: typo

engine/resource/src/test/test_resource_archive.cpp
```diff
@@ -464,7 +464,7 @@ This test is failing intermittenly on Linux. Typical output from a failed test:
 #if !defined(__linux__)
 TEST(dmResourceArchive, ManifestSignatureVerification)
 {
-    dmResource::HManifest manifest = new dmResource::Manifest();
+    dmResource::HManifest manifest = 0;
```
AGulev, line 467: I fixed only this test for now. 
`LoadManifestFromBuffer` creates a new manifest and return into `manifest`. So we don't need 

>dmResource::HManifest manifest = new dmResource::Manifest();

ASAN detects it as leak

scripts/cmake/functions_test.cmake
```diff
@@ -8,6 +8,47 @@ defold_log("functions_test.cmake:")
 #               adds it to global run_tests
 #   - run_workdir: optional working directory for executing the test
 
+function(_defold_find_nodejs out_var)
```
AGulev, line 11: find node.js in installed packages

scripts/cmake/functions_test.cmake
```diff
@@ -49,15 +90,33 @@ function(defold_register_test_target target_name)
   if(_RUN_TEST)
     set(_run_target "run_${target_name}")
     if(NOT TARGET ${_run_target})
+      if(TARGET_PLATFORM MATCHES "js-web|wasm-web|wasm_pthread-web")
```
AGulev, line 93: CMake equivalent of the waf web test setup

scripts/cmake/platform_html5.cmake
```diff
@@ -35,6 +35,11 @@ if(_DEFOLD_WITH_PTHREAD)
 endif()
 
 # Link options base
+set(_DEFOLD_INITIAL_MEMORY 33554432)
+if(WITH_ASAN AND TARGET_PLATFORM MATCHES "^wasm")
+  set(_DEFOLD_INITIAL_MEMORY 67108864)
+endif()
```
AGulev, line 41: CMake equivalent of the waf_dynamo changes

JCash (approved):

PR #11740, opened 2026-01-13
Report old and new shader combinations

### LINKED ISSUE #11739: Report old shaders when there is a warning about potential performance problems.
Description:
When the warning appears:
”Warning: Mixing old shaders with new shaders is slow. Consider migrating old shaders to using the new shader pipeline.”

It would be good to provide a list of the specific materials or shaders that are causing this warning. 
When you have alot of shaders from an older version of Defold and you upgrade, then its quite a substantial amount of time to go through each shader/material to resolve this. 

Potential solution:
Add reporting or logging that lists the old materials or shaders. 

An awesome solution:
Auto update the old shaders :) 

Reference discussion on forum:
https://forum.defold.com/t/mixing-old-shaders-with-new-shaders-warning/82223

Jhonnyg:
When mixing old and new shader pipelines, we currently show a warning but don't tell the users which shaders is causing it. Instead, we now show which shader resources that are causing this to happen.

Fixes #11739


AGulev (approved):
👍

PR #11746, opened 2026-01-14
Fix Android resource detection

ekharkunov:
Fixed issue when bundle resources doesn't detected as valid Android resources because of additional .DS_Store files


AGulev (approved):
👍

PR #11748, opened 2026-01-15
Move project dependencies into their own separate category in the Assets panel

vlaaad:
In projects with many dependencies, having them all in the main folder can be disruptive. To improve this, all dependencies are moved into a separate Dependencies folder at the same root level as the project.

Fix #4156

### DIFF
--- a/editor/resources/localization/en.editor_localization
+++ b/editor/resources/localization/en.editor_localization
@@ -583,6 +583,10 @@ pane.build-errors = Build Errors
 pane.search-results = Search Results
 pane.breakpoints = Breakpoints
 
+# Assets
+
+assets.dependencies = Dependencies
+
 # Console
 
 console.button.clear = Clear
--- a/editor/src/clj/editor/asset_browser.clj
+++ b/editor/src/clj/editor/asset_browser.clj
@@ -39,11 +39,10 @@
             [util.defonce :as defonce]
             [util.eduction :as e]
             [util.fn :as fn])
-  (:import [editor.resource FileResource]
+  (:import [com.defold.control LazyTreeItem]
+           [editor.resource FileResource]
            [java.io File]
            [java.nio.file Path Paths]
-           [java.util Collection]
-           [javafx.collections ObservableList]
            [javafx.scene.input Clipboard ClipboardContent]
            [javafx.scene.input DragEvent MouseEvent TransferMode]
            [javafx.scene Node]
@@ -69,18 +68,18 @@
 (defn- asset-group? [x]
   (instance? AssetGroup x))
 
+(defn- list-children [asset-group-or-resource]
+  (e/filter
+    #(or (asset-group? %)
+         (and (resource/loaded? %)
+              (not (resource/internal? %))))
+    (if (asset-group? asset-group-or-resource)
+      (:children asset-group-or-resource)
+      (resource/children asset-group-or-resource))))
+
 (defn tree-item
   ^TreeItem [asset-group-or-resource]
-  (let [children (if (asset-group? asset-group-or-resource)
-                   (:children asset-group-or-resource)
-                   (resource/children asset-group-or-resource))
-        child-items (coll/transfer children []
-                      (filter #(or (asset-group? %)
-                                   (and (resource/loaded? %)
-                                        (not (resource/internal? %)))))
-                      (map tree-item))]
-    (doto (TreeItem. asset-group-or-resource)
-      (-> .getChildren (^[Collection] ObservableList/.setAll child-items)))))
+  (LazyTreeItem. asset-group-or-resource list-children))
 
 (handler/register-menu! ::resource-menu
   [menu-items/open-selected
--- a/editor/src/clj/editor/asset_browser.clj
+++ b/editor/src/clj/editor/asset_browser.clj
@@ -139,12 +139,9 @@
        (not (resource/read-only? x))
        (not (fixed-resource-paths (resource/proj-path x)))))
 
-(defn- resource-roots
-  "Accepts selection (Resources and AssetGroups), returns resource roots only"
-  [selection]
-  (let [path->resource (->> selection
-                            (e/filter resource/resource?)
-                            (coll/pair-map-by (comp ->path resource/proj-path)))
+(defn- roots [resources]
+  {:pre [(coll/every? resource/resource? resources)]}
+  (let [path->resource (coll/pair-map-by (comp ->path resource/proj-path) resources)
         roots (loop [paths (keys path->resource)
                      roots []]
                 (if-let [^Path path (first paths)]
@@ -193,7 +190,7 @@
   (enabled? [selection]
     (coll/any? resource/resource? selection))
   (run [selection]
-    (copy (-> selection resource-roots fileify-resources!))))
+    (copy (fileify-resources! (roots (filterv resource/resource? selection))))))
 
 (defn- select-resource!
   ([asset-browser resource]
@@ -209,7 +206,10 @@
          tree-items (ui/tree-item-seq (.getRoot tree-view))
          path (resource/resource->proj-path resource)]
      (when-let [tree-item (coll/first-where
-                            #(and (resource/resource? %) (= path (resource/proj-path %)))
+                            (fn [^TreeItem tree-item]
+                              (let [item (.getValue tree-item)]
+                                (and (resource/resource? item)
+                                     (= path (resource/proj-path item)))))
                             tree-items)]
        (doto (.getSelectionModel tree-view)
          (.clearSelection)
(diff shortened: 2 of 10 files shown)


editor/src/clj/editor/asset_browser.clj
```diff
@@ -206,26 +204,31 @@
...
   (run [selection selection-provider asset-browser]
     (let [next (-> (handler/succeeding-selection selection-provider)
                    (handler/adapt-single resource/Resource))
-          cut-files-directory (fs/create-temp-directory! "asset-cut")]
-      (copy (mapv #(temp-resource-file! cut-files-directory %) (roots selection)))
-      (delete selection)
+          cut-files-directory (fs/create-temp-directory! "asset-cut")
+          resources (roots (filterv deletable-resource? selection))]
+      (copy (mapv #(temp-resource-file! cut-files-directory %) resources))
+      (delete resources)
```
vlaaad, line 231: There is a small semantics change without a behavior change. Previously, we deleted `selection`, now we delete the **roots** of deletable resources. It was possible to select `/path/to/dir` and `/path/to/dir/file.txt`, and deletion would attempt to delete both, and the process would silently fail when deleting `/path/to/dir/file.txt` because the parent dir would already be deleted. Now, since we use `roots` instead, we would only attempt to delete the `/path/to/dir`.

editor/src/clj/editor/asset_browser.clj
```diff
@@ -32,59 +36,50 @@
...
+           [javafx.scene.control SelectionMode TreeCell TreeItem TreeView]
            [javafx.scene.input KeyCode KeyEvent MouseEvent]
            [javafx.stage Stage]
            [org.apache.commons.io FilenameUtils]))
 
-(set! *warn-on-reflection* true)
+;; Implementation note: tree item values are either:
+;; - AssetGroup instances: grouping folders that are not implied to be resources
+;; - Resource instances: resources
+;; This means the selection is not guaranteed to only contain resources.
```
vlaaad, line 57: I carefully went through this file and, hopefully, found all places that needed updating to reflect this change.

editor/src/clj/editor/asset_browser.clj
```diff
@@ -804,27 +841,57 @@
...
   (selection [this]
     (ui/selection (g/node-value asset-browser :tree-view)))
   (succeeding-selection [this]
-    (let [tree-view (g/node-value asset-browser :tree-view)
-          path-fn (comp #(string/split % #"/") item->path)]
-      (->> (ui/selection-root-items tree-view path-fn item->path)
-        (ui/succeeding-selection tree-view))))
+    (let [tree-view (g/node-value asset-browser :tree-view)]
+      (->> (ui/selection-root-items tree-view asset-group-or-resource->id)
+           (ui/succeeding-selection tree-view))))
```
vlaaad, line 851: I refactored `ui/selection-root-items` so it does not need a "path" of an item (which was used to check if one selected item is a parent of another item by checking if the path is a subsequence of another path). This is because previously, you could only have resources selected, so you could split a `proj-path` by `/` to get this "path", but now we might have AssetGroups that are identified by MessagePatterns. We actually don't need this path, since the code in `selection-root-items` could walk TreeItem parents and only look up their ids to build such a "path" — and this is what it does now.

editor/src/clj/editor/ui.clj
```diff
@@ -555,7 +555,7 @@
            TreeView
            ;; Only count double-clicks on selected tree items. Ignore disclosure arrow clicks.
            (when-some [clicked-node (some-> event .getPickResult .getIntersectedNode)]
-             (when-some [^TreeCell tree-cell (closest-node-of-type TreeCell clicked-node)]
+             (when-some [^javafx.scene.control.TreeCell tree-cell (closest-node-of-type javafx.scene.control.TreeCell clicked-node)]
```
vlaaad, line 558: Funny code change. We actually have `TreeCell` imported, but it's not `javafx.scene.control.TreeCell`; instead, it's `com.defold.control.TreeCell`, which is used in other places. But now, the asset browser uses JavaFX TreeCells. Anyway, this code should work on `javafx.scene.control.TreeCell`.

editor/test/integration/asset_browser_test.clj
```diff
@@ -198,7 +198,7 @@
           [read-only-file-resource] false
           [writable-dir-resource] true
           [writable-file-resource] true
-          [writable-file-resource read-only-dir-resource] false
+          [writable-file-resource read-only-dir-resource] true
```
vlaaad, line 201: I changed the semantics of cut/copy/delete operations to much Intellij IDEA: they are now enabled if at least one selected item is a valid target for the operation, and they filter selection in the `run` handler. So, if a writable and a read-only resource are selected, the delete operation is available, and will only delete the writable resource.

editor/resources/localization/en.editor_localization
```diff
@@ -583,6 +583,10 @@ pane.build-errors = Build Errors
 pane.search-results = Search Results
 pane.breakpoints = Breakpoints
 
+# Assets
+
+assets.dependencies = Dependencies
+
```
matgis, line 589: I feel somewhat uncomfortable about how the Dependencies folder name is capitalized next to the other top-level folder named after my project directory, which happens to be in lower-case. Maybe we should change the project directory to appear simply as "Project" as part of this change so the two top-level items are consistent in their capitalization? 🤔

editor/src/clj/editor/asset_browser.clj
```diff
@@ -189,9 +186,10 @@
     (.setContent cb content)))
 
 (handler/defhandler :edit.copy :asset-browser
-  (enabled? [selection] (not (empty? selection)))
+  (enabled? [selection]
+    (coll/any? resource/resource? selection))
   (run [selection]
-       (copy (-> selection roots fileify-resources!))))
+    (copy (fileify-resources! (roots (filterv resource/resource? selection))))))
```
matgis, line 192: There are a few places in this PR where the `selection` is treated as a sequence of `Resources`. But elsewhere we seem to use the `handler/adapt` functions rather than assuming the selection will consist of `Resources` directly. I suppose it's fine, since the handlers are in the `:asset-browser` context where we supply our own `SelectionProvider` that guarantees this? 🤔

editor/src/clj/editor/asset_browser.clj
```diff
@@ -611,95 +622,120 @@
...
+      (when (get expanded (tree-item->id item))
+        (.setExpanded item true)))))
+
+(def ^:private assets-message (localization/message "pane.assets"))
+(def ^:private dependencies-message (localization/message "assets.dependencies"))
+
+(defn- with-dependencies-subtree [resource-tree]
+  (let [children (:children resource-tree)
+        file-resources (filterv resource/file-resource? children)
+        dep-resources (into [] (remove resource/file-resource?) children)]
```
matgis: I think we could use `coll/separate-by` here?

editor/src/clj/editor/asset_browser.clj
```diff
@@ -611,95 +622,120 @@
       (app-view/show-asset-browser! (.getScene ^Stage main-stage))
       (select-resource! asset-browser r {:scroll? true}))))
 
-(defn- item->path [^TreeItem item]
-  (-> item (.getValue) (resource/proj-path)))
+(defn- asset-group-or-resource->id [asset-group-or-resource]
```
matgis, line 625: We could maybe add a `{:post [(some? %)]}` check here to signal this is expected to return a non-`nil` value. We appear to rely on it at the call sites.

editor/src/clj/editor/asset_browser.clj
```diff
@@ -804,27 +841,57 @@
...
+(def ^:private ext-with-tree-view-props
+  (fx/make-ext-with-props fx.tree-view/props))
+
+(fxui/defc asset-tree-view
+  {:compose [{:fx/type fx/ext-watcher :ref (:localization props) :key :localization-state}]}
+  [{:keys [tree-view localization-state on-drag-dropped]}]
+  {:fx/type ext-with-tree-view-props
+   :desc {:fx/type fxui/ext-value :value tree-view}
+   :props {:cell-factory {:fx/cell-type fx.tree-cell/lifecycle
+                          :describe (fn/partial #'describe-tree-cell localization-state on-drag-dropped)}}})
```
matgis, line 881: Will the equality semantics of `fn/partial` work correctly when `on-drag-dropped` appears to itself be a lambda here?

matgis (dismissed):
Seems to work, and a nice cleanup of the code as well!

I had some questions, but I don't see anything that needs to be addressed before merging.

editor/src/clj/editor/asset_browser.clj
vlaaad: I was actually considering refactoring the whole `adapt` thing because it has some weird limitations and assumptions, but I didn't want to expand the scope of this task too much since I found myself doing it by a complete accident :D 

One of such weird assumptions is that it expects the whole selection to be uniform/heterogeneous; e.g., if at least one item is a Resource, then all of them must be, but here, I specifically have a different use case, where I have a mix of Resources and other things in a selection, and I want to filter out stuff that doesn't fit — currently it can't be achieved using `adapt`. It's fine, though, since it's for this context only.

editor/src/clj/editor/asset_browser.clj
vlaaad: Yes, `on-drag-dropped` is created once in `setup-asset-browser` and then the instance is always reused here.

Issue #9161, opened 2024-07-03
Animating tint values using go.animate causes flickering when using Vulkan


selimanac:
OpenGL

https://github.com/defold/defold/assets/1324635/1f025cba-834a-4298-840e-bcb6b800af93

Vulkan

https://github.com/defold/defold/assets/1324635/2b6983f1-e499-4562-aa3a-8b4f93dc30b8

Min repo case:
[tint-vulkan.zip](https://github.com/user-attachments/files/16078229/tint-vulkan.zip)

<hr/>

<table>
<tr><td>Build time</td><td>2024-07-03T06:10:19.654069</td></tr>
<tr><td>Defold channel</td><td>alpha</td></tr>
<tr><td>Defold editor sha</td><td>5627038e48711ee6f148b32fea72d6d9781770d5</td></tr>
<tr><td>Defold engine sha</td><td>5627038e48711ee6f148b32fea72d6d9781770d5</td></tr>
<tr><td>Defold version</td><td>1.9.1</td></tr>
<tr><td>GPU</td><td>Apple M2 Pro</td></tr>
<tr><td>GPU Driver</td><td>2.1 Metal - 88.1</td></tr>
<tr><td>Java version</td><td>17.0.5+8</td></tr>
<tr><td>OS arch</td><td>aarch64</td></tr>
<tr><td>OS name</td><td>Mac OS X</td></tr>
<tr><td>OS version</td><td>14.5</td></tr>
<table>

PR #11751, opened 2026-01-15
Uniform buffer as first class object

Jhonnyg:
The engine can now create, bind and manipulating uniform buffer objects in runtime. These can then be used for immutable data storage inside the engine for storing things like PBR materials and light component data. There is no user facing code involved here yet, but it is a requirement for the upcoming light component refactor.


engine/graphics/src/dmsdk/graphics/graphics.h
```diff
@@ -50,7 +50,7 @@ namespace dmGraphics
      * @typedef
      * @name HRenderTarget
      */
-    typedef uint64_t HRenderTarget; // Where is this currently used?
+    typedef uint64_t HRenderTarget;
```
Jhonnyg, line 53: Rive extension!

engine/graphics/src/graphics.h
```diff
@@ -63,6 +63,7 @@ namespace dmGraphics
     // Note: that we can only use a total of 53 bits out of the 64 due to how we expose the handles
     //       to the users via lua: http://lua-users.org/wiki/NumbersTutorial
     typedef uint64_t HAssetHandle;
+    typedef uint64_t HUniformBuffer;
```
Jhonnyg: Don't need it in DMSDK yet.

engine/graphics/src/graphics.h
```diff
@@ -236,6 +237,39 @@ namespace dmGraphics
...
+        uint32_t                    m_Offset;
+    };
+
+    struct ShaderResourceTypeInfo
+    {
+        char*                 m_Name;
+        dmhash_t              m_NameHash;
+        ShaderResourceMember* m_Members;
+        uint32_t              m_MemberCount;
+    };
```
Jhonnyg: Moved from graphics_private.h

engine/graphics/src/graphics.cpp
```diff
@@ -1551,7 +1735,24 @@ namespace dmGraphics
...
+    void GetUniformBufferLayout(uint32_t root_type_index, const ShaderResourceTypeInfo* types, uint32_t num_types, UniformBufferLayout* layout_desc)
+    {
+        HashState32 hash_state;
+        dmHashInit32(&hash_state, false);
+        assert(root_type_index < num_types);
+
+        dmArray<bool> visited;
+        visited.SetCapacity(num_types);
+        visited.SetSize(num_types);
+        memset(visited.Begin(), 0, sizeof(bool) * num_types);
```
JCash: I think heap allocation is not necessary, as it should be a small number of actual types?
E.g.:
```c++
#include <dlib/dalloca.h>
bool* visited = (bool*)dmAlloca(sizeof(bool) * num_types);
memset(visited, 0, sizeof(bool) * num_types);
```

engine/graphics/src/graphics.cpp
```diff
@@ -1400,12 +1449,147 @@ namespace dmGraphics
...
         {
+            program->m_UniformBufferLayouts.SetCapacity(program->m_ShaderMeta.m_UniformBuffers.Capacity());
+
             FillProgramResourceBindings(program, program->m_ShaderMeta.m_UniformBuffers, program->m_ShaderMeta.m_TypeInfos, bindings, ubo_alignment, ssbo_alignment, info);
             FillProgramResourceBindings(program, program->m_ShaderMeta.m_StorageBuffers, program->m_ShaderMeta.m_TypeInfos, bindings, ubo_alignment, ssbo_alignment, info);
             FillProgramResourceBindings(program, program->m_ShaderMeta.m_Textures, program->m_ShaderMeta.m_TypeInfos, bindings, ubo_alignment, ssbo_alignment, info);
         }
     }
 
+    uint32_t GetStd140BaseAlignment(dmGraphics::ShaderDesc::ShaderDataType type)
```
JCash: I don't see it being used outside of this file?
Can it be made "static" and removed from the header?

engine/graphics/src/graphics.h
JCash: I don't see it used outside of the graphics library?
What is the use case of having it accessible from other libraries?

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2304,77 +2414,74 @@ static void LogFrameBufferError(GLenum status)
...
         uint32_t num_ubos = program->m_BaseProgram.m_ShaderMeta.m_UniformBuffers.Size();
         uint32_t ubo_binding = 0;
         program->m_UniformBuffers.SetCapacity(num_ubos);
         program->m_UniformBuffers.SetSize(num_ubos);
 
-        memset(program->m_UniformBuffers.Begin(), 0, sizeof(OpenGLUniformBuffer) * num_ubos);
+        memset(program->m_UniformBuffers.Begin(), 0, sizeof(OpenGLScratchUniformBuffer) * num_ubos);
 
-        for (uint32_t i = 0; i < num_shaders; ++i)
+        for (uint32_t j = 0; j < program->m_BaseProgram.m_ShaderMeta.m_UniformBuffers.Size(); ++j)
```
JCash: While this is during building, the performance isn't as critical, but I think it also sets a standard.
Here it's 4 fetches, and then a function call, for each turn.
I'd rather have it stored upfront in a variable "uint32_t num_buffers`.

It also highlights the contents of the loop, should we store a UniformBuffer* outside of the loop, to further reduce the number of fetches for each case?
(Since they're pointers, the program has to traverse them all as it cannot be certain none of them has changed since last time)

The OpenGL backend is on its way out of course, so it's not about this backend in particular, but our code style in general, and having "code as documentation" so that when people read it, they know how to do things.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -1774,23 +1774,95 @@ namespace dmGraphics
...
-        // so for now we always mark the old buffer for destruction when updating the data.
-    #ifndef __MACH__
-        if (size != buffer->m_MemorySize)
-    #endif
+        if (offset == 0)
         {
-            DestroyResourceDeferred(context, buffer);
+            // Coherent memory writes does not seem to be properly synced on MoltenVK,
+            // so for now we always mark the old buffer for destruction when updating the data.
+        #ifndef __MACH__
```
JCash, line 1788: Does this handle when using `with_vulkan_validation` ?
I.e. is it a MACH thing or a MoltenVK thing?

Currently, when we have MACH and with_vulkan_validation, we suddenly get the extra if-statement, but the comment suggests that was only supposed to happen if we use MoltenVK, which we don't in that case.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -2628,14 +2737,14 @@ namespace dmGraphics
 
     static inline VkDescriptorType GetDescriptorType(const ShaderResourceBinding& res)
     {
-        if (ShaderResourceBinding::BINDING_FAMILY_GENERIC)
+        if (BINDING_FAMILY_GENERIC)
             return (VkDescriptorType) -1;
```
JCash: Not sure what this is supposed to do?
Should be fixed or removed, as it's alwas false, and never executed.
Probably outputs a warning when building?

engine/graphics/src/webgpu/graphics_webgpu.cpp
```diff
@@ -2256,56 +2333,82 @@ static void WebGPUUpdateBindGroups(WebGPUContext* context)
...
+                    {
+                        entries[desc.entryCount].buffer = bound_ubo->m_Buffer;
+                        entries[desc.entryCount].offset = 0;
+                        entries[desc.entryCount].size   = bound_ubo->m_BaseUniformBuffer.m_Layout.m_Size;
+                    }
+                    else
+                    {
+                        if (context->m_CurrentScratchUniforms.m_Allocs.Size() == 0 ||
+                            context->m_CurrentScratchUniforms.m_Allocs[context->m_CurrentScratchUniforms.m_Alloc]->m_Size <
+                            context->m_CurrentScratchUniforms.m_Allocs[context->m_CurrentScratchUniforms.m_Alloc]->m_Used + pgm_res.m_Res->m_BindingInfo.m_BlockSize)
```
JCash, line 2375: While it's a bit of inherited code, it shows the same type of pattern as above with lots of pointer chasing.
Imho, it also distracts when reading.

Perhaps something like this would help:
```c++
Alloc* alloc = context->m_CurrentScratchUniforms.m_Allocs.Empty() ? 0 : &context->m_CurrentScratchUniforms.m_Allocs[context->m_CurrentScratchUniforms.m_Alloc];
if (alloc && alloc->m_Size < (alloc->m_Used + pgm_res.m_Res->m_BindingInfo.m_BlockSize))
{
    ...
}
```

engine/graphics/src/webgpu/graphics_webgpu.cpp
```diff
@@ -2256,56 +2333,82 @@ static void WebGPUUpdateBindGroups(WebGPUContext* context)
...
+                                    context->m_CurrentScratchUniforms.m_Allocs.OffsetCapacity(4);
+                                }
+                                context->m_CurrentScratchUniforms.m_Alloc = context->m_CurrentScratchUniforms.m_Allocs.Size();
+                                context->m_CurrentScratchUniforms.m_Allocs.Push(alloc);
                             }
-                            context->m_CurrentUniforms.m_Alloc = context->m_CurrentUniforms.m_Allocs.Size();
-                            context->m_CurrentUniforms.m_Allocs.Push(alloc);
                         }
+                        entries[desc.entryCount].buffer = context->m_CurrentScratchUniforms.m_Allocs[context->m_CurrentScratchUniforms.m_Alloc]->m_Buffer;
+                        entries[desc.entryCount].offset = context->m_CurrentScratchUniforms.m_Allocs[context->m_CurrentScratchUniforms.m_Alloc]->m_Used;
```
JCash, line 2405: Future improvements: While this is the same code as before, it is a bit crazy to me to do this many lookups (5) to get the Alloc*, and then do it twice (once again for the second line). And as it's in a Update type of function, I think this is worth taking into account in later improvements.

JCash (changes requested):
Overall good, but left some notes

engine/graphics/src/graphics.h
Jhonnyg: Yeah it was used outside at first, I think. But it's basically only for tests, so I moved it back to private :)

engine/graphics/src/opengl/graphics_opengl.cpp
Jhonnyg: Yep I agree, easy thing to forget :) Fixed.

engine/graphics/src/vulkan/graphics_vulkan.cpp
Jhonnyg: I haven't really had time to figure out why the de-sync is happening. It's basically because of this issue:
https://github.com/defold/defold/issues/9161

It doesn't happen on windows+vulkan, so it's something metal related, but I don't know more I'm afraid..

engine/graphics/src/webgpu/graphics_webgpu.cpp
Jhonnyg: I will not touch this right now, but I'll add a todo

PR #11752, opened 2026-01-16
Improved scrolling behavior in Assets Browser and Outline Pane

### LINKED ISSUE #11711: GUI outliner cutting to current selection
At random occasions, the GUI outliner (I believe this also happens or at least similarly in collections, but I have no proof, will follow up if I can docuemnt it) cuts off the entire list, making the currently selected point the top one. Scrolling or selecting in the outliner does not affect it. Clicking into the editor fixes it and brings back the outline list.

**To Reproduce (REQUIRED)**
1. Make a list of nodes in a .gui
2. Wait for it to happen, I guess? I could not find a rule to it.

The outline shouldn't be clipped.

1.11.2

**Platforms (REQUIRED):**
Windows 10

Image

JosephFerano:
- When dragging and dropping, scroll just enough to make the items visible, rather than always at the top, which should help track movement of files/nodes
- Center a resource when opening with right-click "Show in Asset Browser", rather than up top
- Fix bug when auto-scrolling downwards with mouse drag-n-drop where it would scroll by a whole page, rather than gradually
- Auto-scrolling now has variable speed depending on how close to the edge you move the mouse while dragging. It's also smoother!

Fixes #11711


editor/src/java/com/defold/control/ExtendedTreeViewSkin.java
```diff
@@ -19,11 +19,15 @@
 import javafx.scene.control.skin.TreeViewSkin;
 import javafx.scene.control.skin.VirtualFlow;
 
-public class OutlineTreeViewSkin<T> extends TreeViewSkin<T> {
-    public OutlineTreeViewSkin(TreeView treeView) {
+public class ExtendedTreeViewSkin<T> extends TreeViewSkin<T> {
```
JosephFerano, line 22: Now assets_browser uses the same so renamed to something more general.

editor/src/clj/editor/outline_view.clj
```diff
@@ -486,33 +483,28 @@
...
-                                 (root-iterators outline-view)
-                                 [])]
-            (when (outline/drop? project item-iterators (.getTreeItem cell)
-                                 (.getContent db (data-format-fn)))
-              (let [modes (if (ui/drag-internal? e)
-                            [TransferMode/MOVE]
-                            [TransferMode/COPY])]
-                (.acceptTransferModes e (into-array TransferMode modes)))
-              (.consume e))))))))
+  (when-let [^TreeCell cell (find-tree-cell-ancestor (.getTarget e))]
```
JosephFerano, line 493: I removed `(Event/fireEvent parent (.copyFor e (.getSource e) parent))` because it was triggering the dragOver event extra times for the outline view, causing faster scrolling. So instead I just try and find the parent recursively. Everything seems to work the same.

editor/src/clj/editor/asset_browser.clj
```diff
@@ -779,14 +778,6 @@
     (when-let [^TreeCell cell (target (.getTarget e))]
       (when (and (not (.isEmpty cell))
                  (.hasFiles db))
-       ;; Auto scrolling
```
JosephFerano, line 783: Handle auto-scrolling with the new editor-ui functions and on-drag-over on the tree-view

editor/src/clj/editor/ui.clj
```diff
@@ -1146,11 +1144,53 @@
...
+         flow (.getVirtualFlowInstance skin)
+         last-visible-idx (.getIndex (.getLastVisibleCell flow))]
+     (when (.shouldScrollTo skin first-index)
+       (if (> last-index last-visible-idx)
+         (let [fixed-cell-size (.getHeight (.getCell flow first-index))
+               cells-to-scroll (+ scroll-padding-cells (- last-index last-visible-idx))]
+           (.scrollPixels flow (* cells-to-scroll fixed-cell-size)))
+         (.scrollTo tree-view (dec first-index)))))))
+
+(defn scroll-tree-view-to-center-item! [^TreeView tree-view index]
```
JosephFerano: This one is used by right click "Show in Asset Browser". Now it centers instead of scrolling the item to the top.

editor/src/clj/editor/ui.clj
```diff
@@ -1146,11 +1144,53 @@
...
+  (let [skin ^ExtendedTreeViewSkin (.getSkin tree-view)
+        flow (.getVirtualFlowInstance skin)]
+    (.scrollPixels flow pixels)))
+
+(defn handle-tree-view-scroll-on-drag! [^TreeView tree-view ^DragEvent e]
+  (let [view-y (.getY (.sceneToLocal tree-view (.getSceneX e) (.getSceneY e)))
+        height (.getHeight (.getBoundsInLocal tree-view))
+        scroll-zone 40
+        max-speed 16]
+    (cond
```
JosephFerano, line 1193: This adds variable speed scrolling. If you wiggle your mouse horizontally at the innermost extent of the scroll zone, it scrolls more slowly than if you wiggle closer to the edge of the zone.

editor/src/clj/editor/asset_browser.clj
```diff
@@ -692,11 +692,10 @@
         (auto-expand (.getChildren root) selected-ids)
         (let [count (.getExpandedItemCount tree-view)
               selected-indices (filterv #(selected-ids (tree-item->id (.getTreeItem tree-view %))) (range count))]
-          (when (not (empty? selected-indices))
+          (when (not (coll/empty? selected-indices))
             (ui/select-indices! tree-view selected-indices))
           (when-not (= old-tree-view-ids selected-ids)
-            (when-some [first-item (first (.getSelectedItems selection-model))]
-              (ui/scroll-to-item! tree-view first-item))))))))
+            (ui/scroll-tree-view-to-encompass-items! tree-view (first selected-indices) (peek selected-indices))))))))
```
vlaaad: Should we guard against empty selection here? The old code did it...

editor/src/clj/editor/ui.clj
```diff
@@ -1146,11 +1144,53 @@
...
+  "Scrolls tree-view to show all items between first-index and last-index,
+  with optional padding cells above and below."
+  ([^TreeView tree-view first-index last-index]
+   (scroll-tree-view-to-encompass-items! tree-view first-index last-index 2))
+  ([^TreeView tree-view first-index last-index scroll-padding-cells]
+   {:pre [(instance? ExtendedTreeViewSkin (.getSkin tree-view))]}
+   (let [skin ^ExtendedTreeViewSkin (.getSkin tree-view)
+         flow (.getVirtualFlowInstance skin)
+         last-visible-idx (.getIndex (.getLastVisibleCell flow))]
+     (when (.shouldScrollTo skin first-index)
```
vlaaad: Previous code skipped scrolling to the item when the index was `-1`, should we still do it here? Also, it seems to me that in all cases where `scroll-tree-view-to-encompass-items!` fn is used, the first and last indices are the first and last indices of the selection of this tree view. Maybe we shouldn't even pass the indices here, and rename fn to `scroll-tree-view-to-encompass-selection!`, and lookup selected indices, and do the empty checks here?

editor/src/clj/editor/ui.clj
vlaaad: Should we guard against `-1` index here as well?

editor/src/clj/editor/ui.clj
```diff
@@ -1146,11 +1144,53 @@
...
+       (if (> last-index last-visible-idx)
+         (let [fixed-cell-size (.getHeight (.getCell flow first-index))
+               cells-to-scroll (+ scroll-padding-cells (- last-index last-visible-idx))]
+           (.scrollPixels flow (* cells-to-scroll fixed-cell-size)))
+         (.scrollTo tree-view (dec first-index)))))))
```
vlaaad: Looks like a potential invalid scroll target when `first-index` is zero. Maybe `(max 0 (dec first-index))`?

editor/src/clj/editor/asset_browser.clj
```diff
@@ -39,14 +39,14 @@
...
            [editor.resource FileResource]
            [java.io File]
            [java.nio.file Path Paths]
-           [javafx.scene.input Clipboard ClipboardContent]
-           [javafx.scene.input DragEvent MouseEvent TransferMode]
            [javafx.scene Node]
            [javafx.scene.control SelectionMode TreeCell TreeItem TreeView]
+           [javafx.scene.input Clipboard ClipboardContent]
+           [javafx.scene.input DragEvent MouseEvent TransferMode]
            [javafx.scene.input KeyCode KeyEvent MouseEvent]
```
vlaaad: Huh? 2 lines for importing classes from `javafx.scene.input` package with repeated `MouseEvent`?

vlaaad (changes requested):
Seems to work great! 

I think there are some places in the code that might not be guarding enough, see comments...

editor/src/clj/editor/asset_browser.clj
JosephFerano: Oh this might be a setting in clojure-lsp's clean-ns code action, let me go read the docs, because I think it did it in another PR.

JosephFerano:
How does the auto-scroll speed feel? Fast enough?

vlaaad:
Feels great, better than before!

editor/src/clj/editor/ui.clj
JosephFerano: I had tested it and it seems like javafx does bounds checking so I decided not to do it here. However, if you think it would be better form, I can add it

vlaaad: But if the first index is 0 (i.e., it's the first item in a tree view), we still want to actually scroll to it, not let JavaFX perform a bounds check and do nothing?

JosephFerano: Sorry, I called it bounds checking but that's the wrong term. I meant clamping. JavaFX will clamp and scroll to the item.

editor/src/clj/editor/ui.clj
JosephFerano: For the latter point, I agree, I moved the `first` and `peek`. By also putting it inside the empty check, it shouldn't fail.

In regards to checking for -1, I tested the code by forcing a -1 and some super high number like 1000 as well. JavaFX seems to handle it gracefully. I'm not sure what the edge case for getting a -1 would be (perhaps a temporarily detached item?). I can add a `(>= last-index first-index 0)` just in case.

editor/src/clj/editor/ui.clj
vlaaad: Huh, TIL. Sorry, disregard my comment then. Maybe we could add a comment in the code that explains the clamping?

editor/src/clj/editor/asset_browser.clj
```diff
@@ -45,9 +45,7 @@
            [java.nio.file Path Paths]
            [javafx.scene Node]
            [javafx.scene.control SelectionMode TreeCell TreeItem TreeView]
-           [javafx.scene.input Clipboard ClipboardContent]
-           [javafx.scene.input DragEvent MouseEvent TransferMode]
-           [javafx.scene.input KeyCode KeyEvent MouseEvent]
+           [javafx.scene.input Clipboard ClipboardContent DragEvent KeyCode KeyEvent MouseEvent MouseEvent TransferMode]
```
vlaaad: Two `MouseEvent` imports 😄

editor/src/clj/editor/ui.clj
```diff
@@ -1144,21 +1144,25 @@
...
-(defn scroll-tree-view-to-encompass-items!
+(defn scroll-tree-view-to-encompass-selection!
   "Scrolls tree-view to show all items between first-index and last-index,
   with optional padding cells above and below."
-  ([^TreeView tree-view first-index last-index]
-   (scroll-tree-view-to-encompass-items! tree-view first-index last-index 2))
-  ([^TreeView tree-view first-index last-index scroll-padding-cells]
+  ([^TreeView tree-view selected-indices]
+   (scroll-tree-view-to-encompass-selection! tree-view selected-indices 2))
+  ([^TreeView tree-view selected-indices scroll-padding-cells]
```
vlaaad: Do we need to pass `selected-indices` as an argument here? Or is it expected that they are always the same as `(.getSelectedIndices (.getSelectionModel tree-view))`?

vlaaad (changes requested):
Looks good, but lets get rid of boxed math warnings that we get on startup before merging:
```
Boxed math warning, editor/ui.clj:1160:12 - call: public static boolean clojure.lang.Numbers.gt(java.lang.Object,long).
Boxed math warning, editor/ui.clj:1162:56 - call: public static java.lang.Number clojure.lang.Numbers.unchecked_minus(java.lang.Object,long).
Boxed math warning, editor/ui.clj:1162:32 - call: public static java.lang.Number clojure.lang.Numbers.unchecked_add(java.lang.Object,java.lang.Object).
Boxed math warning, editor/ui.clj:1163:32 - call: public static double clojure.lang.Numbers.unchecked_multiply(java.lang.Object,double).
Boxed math warning, editor/ui.clj:1166:31 - call: public static java.lang.Number clojure.lang.Numbers.unchecked_dec(java.lang.Object).
Boxed math warning, editor/ui.clj:1177:34 - call: public static java.lang.Number clojure.lang.Numbers.unchecked_minus(java.lang.Object,long).
Boxed math warning, editor/ui.clj:1177:27 - call: public static java.lang.Object clojure.lang.Numbers.max(long,java.lang.Object).
```

editor/src/clj/editor/asset_browser.clj
JosephFerano: :man_facepalming:

vlaaad (approved):
LGTM!

One interesting behavior difference I noticed is that scrolling on drag-over. If I open an atlas and try to drag an item from the atlas over the asset browser:
- The old version didn't scroll because it had a guard that only did auto-scroll if the dragboard had files.
- The new version now scrolls.

To be honest, I don't think this is important :D

PR #11753, opened 2026-01-16
Limit max height of editor script dialogs

### LINKED ISSUE #11532: Defold Editor script dialog height issue
In some cases, the height of the dialog window becomes larger than the display height. I'm using a scroll container.

The window should adapt to the available height, or there should be a way to adjust the size of the dialog window

Latest

**Platforms (REQUIRED):**
- It happend on the linux and windows. On MacOS I have all sizes correct on macbook and exterrnal displays

https://github.com/Insality/asset-store/issues/1

Image

Image

vlaaad:
Now, editor dialogs are limited in height. Users are encouraged to use the `editor.ui.scroll()` container for dialogs that may display a lot of content.

Fix #11532


JosephFerano (approved):
Tested with the test script. Works

PR #11769, opened 2026-01-20
Add `editor.ui.image()` editor script function

### LINKED ISSUE #10220: Custom images for Editor UI icons and buttons with icons.
Currently we can use in Editor UI [icon](https://defold.com/ref/beta/editor/#editor.ui.icon:props) component with few predefined images, same for buttons with icons. It would open many possibilities to be able to define somehow custom images for icons.

One possibility is to add ability to add custom images as "icons" to be used in existing editor icon component/button with icon, even statically, defined in some catalog in game project or near editor_script.

Next is to allow to somehow define path to such image instead and/or select image from Defold project image related resources (atlases and tilesources) as icon.

None

[Add any other context or screenshots about the feature request here.
](https://forum.defold.com/t/editor-scripts-ui-images/79717)

vlaaad:
Now, editor scripts can show images, supporting both project files and external URLs, e.g.:
```lua
editor.ui.show_dialog(editor.ui.dialog({
    title = "Images",
    content = editor.ui.horizontal({
        padding = editor.ui.PADDING.LARGE,
        children = {
            editor.ui.image({image = "/builtins/assets/images/logo/logo_256.png", width = 64, height = 64}),
            editor.ui.image({image = "https://defold.com/images/assets/monarch-hero.jpg"})
        }
    })
}))
```
Screenshot 2026-01-20 at 14 10 30

Fixes #10220

### DIFF
--- a/editor/src/clj/editor/editor_extensions/ui_components.clj
+++ b/editor/src/clj/editor/editor_extensions/ui_components.clj
@@ -21,11 +21,13 @@
             [cljfx.fx.label :as fx.label]
             [cljfx.fx.region :as fx.region]
             [cljfx.fx.row-constraints :as fx.row-constraints]
+            [cljfx.fx.stack-pane :as fx.stack-pane]
             [cljfx.fx.stage :as fx.stage]
             [cljfx.fx.v-box :as fx.v-box]
             [cljfx.lifecycle :as fx.lifecycle]
             [cljfx.mutator :as fx.mutator]
             [cljfx.prop :as fx.prop]
+            [clojure.java.io :as io]
             [clojure.spec.alpha :as s]
             [clojure.string :as string]
             [dynamo.graph :as g]
@@ -44,10 +46,12 @@
             [editor.resource :as resource]
             [editor.resource-dialog :as resource-dialog]
             [editor.ui :as ui]
+            [editor.workspace :as workspace]
             [internal.util :as iutil]
             [util.coll :as coll :refer [pair]]
             [util.fn :as fn])
   (:import [com.defold.editor.luart DefoldLuaFn]
+           [java.net URI URISyntaxException]
            [java.nio.file Path]
            [java.util Collection List]
            [javafx.beans.property ReadOnlyProperty]
@@ -64,6 +68,16 @@
 ;; See also:
 ;; - Components doc: https://docs.google.com/document/d/1e6kmVLspQEoe17Ys1nmbbf54cqUn2fNPZowIe7JBwl4/edit
 ;; - Reactive UI doc: https://docs.google.com/document/d/1cgqFPLUc3YQih2tsEA067Pz9Q0nWlKoVflqLEVX8CPY/edit
+
+;; The value gets bound to a volatile with evaluation context. When the context
+;; becomes invalid (fills the cache after use), the volatile is set to nil to
+;; signal that this context can't be used anymore. The default value is a
+;; reduced instance to support `deref`, but not `vreset!`
+(def ^:private ^:dynamic *lifecycle-evaluation-context-ideref* (reduced nil))
+
+(defn- lifecycle-evaluation-context []
+  {:post [(some? %)]}
+  @*lifecycle-evaluation-context-ideref*)
 
 (s/def ::create ifn?)
 (s/def ::advance ifn?)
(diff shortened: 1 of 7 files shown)


editor/src/clj/editor/editor_extensions/ui_components.clj
```diff
@@ -65,6 +70,16 @@
...
 
+;; The value gets bound to a volatile with evaluation context. When the context
+;; becomes invalid (fills the cache after use), the volatile is set to nil to
+;; signal that this context can't be used anymore. The default value is a
+;; reduced instance to support `deref`, but not `vreset!`
+(def ^:private ^:dynamic *lifecycle-evaluation-context-ideref* (reduced nil))
+
+(defn- lifecycle-evaluation-context []
+  {:post [(some? %)]}
+  @*lifecycle-evaluation-context-ideref*)
```
vlaaad, line 81: I moved this stuff up from below

editor/src/clj/editor/markdown.clj
```diff
@@ -389,26 +388,31 @@
...
-  (or
-    (let [src (coll/not-empty (.attr node "src"))]
-      (when-let [^URI uri (when src
-                            (try (URI. src) (catch URISyntaxException _)))]
-        (when-let [image (if (or (.getAuthority uri) (.getScheme uri))
-                           {:url src :background-loading true}
-                           (when-let [base-resource (:base-resource ctx)]
-                             (let [resource (workspace/resolve-resource base-resource (.getPath uri))]
-                               (when (resource/exists? resource)
-                                 {:is (io/input-stream resource)
```
vlaaad, line 405: There was a bug I discovered while implementing the editor script version: every time we showed an image view, we created a fresh `InputStream` that was never closed since JavaFX Image does not close the input streams it receives. Also, InputStreams are handled synchronously, so `:background-loading` did nothing. Background loading of InputStreams (with closing them afterwards) was actually implemented in JavaFX recently (see https://github.com/openjdk/jfx/pull/1875), but I tried version 25.0.1, and the change wasn't there...

editor/src/java/com/defold/control/ResizableImageView.java
```diff
@@ -78,6 +78,16 @@ public void resize(double width, double height) {
...
+    @Override
+    public double maxWidth(double v) {
+        return Double.MAX_VALUE;
+    }
+
+    @Override
+    public double maxHeight(double v) {
+        return Double.MAX_VALUE;
+    }
+
```
vlaaad, line 90: I added an unconstrained max size for ResizableImageView so it can be resized to a bigger size than it originally had (it wasn't possible in Markdown view, but it's now possible to upscale an image using editor scripts)

editor/src/clj/editor/markdown.clj
```diff
@@ -389,26 +388,31 @@
...
+                (Image. is)))))))))
+
+(fxui/defc image-view-impl
+  {:compose [{:fx/type fxui/ext-memo
+              :fn construct-image
+              :args [(:src props) (:base-resource props)]
+              :key :image}]}
+  [{:keys [image]}]
+  (if image
+    {:fx/type fx.h-box/lifecycle :children [{:fx/type fxui/resizable-image :image image}]}
```
vlaaad, line 409: A non-obvious change: previous code didn't wrap the image in `HBox`. Now we need to do it since we allowed the image to increase its size above 100% for editor scripts (in `ResizableImageView.java`). Without HBox, images began to grow to fill the whole available space horizontally...

editor/src/clj/editor/markdown.clj
```diff
@@ -389,26 +388,31 @@
...
 
-(def ^:private ext-with-image-view-props
-  (fx/make-ext-with-props fx.image-view/props))
+(defn- construct-image [src base-resource]
+  (when-not (coll/empty? src)
+    (when-let [^URI uri (try (URI. src) (catch URISyntaxException _))]
+      (if (or (.getAuthority uri) (.getScheme uri))
+        (Image. src #_background-loading true)
+        (when-let [base-resource base-resource]
+          (let [resource (workspace/resolve-resource base-resource (.getPath uri))]
```
matgis, line 397: We may want to supply an `evaluation-context` here?

Not sure if that is easy to do given how all this works, but I suspect we may be violating an existing `evaluation-context` scope here. 🤔

No need to fix it now if it is tricky, though.

editor/test/resources/editor_extensions/ui_project/test.editor_script
```diff
@@ -47,6 +48,68 @@ function M.get_commands()
...
+                                            },
+                                            {
+                                                -- invalid URLs
+                                                editor.ui.image({
+                                                    image = "https://defold.com/"
+                                                }),
+                                                editor.ui.image({
+                                                    image = "https://probably-does-not-exist.com/"
+                                                }),
+                                            }
```
matgis, line 118: Should we add some tests for invalid `width` and `height` too?

editor/src/clj/editor/editor_extensions/ui_docs.clj
```diff
@@ -256,6 +256,12 @@
 (def ^:private icon-props
   (into icon-specific-props common-props))
 
+(def ^:private image-props
+  (into [(make-prop :image :coerce coerce/string :required true :doc "either a resource path (starts with <code>/</code>), or an URL")
+         (make-prop :width :coerce coerce/number :doc "width of the image")
+         (make-prop :height :coerce coerce/number :doc "height of the image")]
```
matgis: These docs could be a bit clearer, maybe?

I suppose these are really the dimensions of the resulting view and the image will be uniformly scaled to fit inside it? Or do we non-uniformly stretch it? 🤔

We could add an optional prop to control this behavior later, but what should the default be? JavaFX defaults to `false` for `isPreserveRatio`, but maybe `true` is better for our use case?

matgis (dismissed):
Changes look good to me. I think we could clarify the docs a bit, perhaps?

editor/src/clj/editor/editor_extensions/ui_docs.clj
vlaaad: ResizableImageView sets `preserveRatio` to true, so we don't distort the image. I'll add a clarification!

editor/src/clj/editor/markdown.clj
vlaaad: That would be tricky, but ultimately, this is the end goal with UI: a cljfx tree with a shared evaluation context throughout the thing.

Insality:
Some ui.images in action :)


https://github.com/user-attachments/assets/2a98e583-26e8-4e5a-908d-4b24a356f90f

vlaaad:
Uhh what? How? It's not even in beta! Looks great!

PR #11775, opened 2026-01-21
Move Path-related editor functions from fs module to new util.path module

matgis:
The `fs` module had a lot of variant functions that operate on either `File` or `Path` objects. This PR moves most of the `Path` functions to their own `util.path` module, and ensures we have counterparts for all the methods on the `Path` class. As a result, a few modules are no longer dependent on the larger `fn` module.

The `util.path` module will try to coerce its arguments into `Path` objects, so it can operate on `File` objects, `URI` objects, strings and so on.

We also now implement the `as-path` coercion on all our `Resource` types, and extend the `clojure.java.io/Coercions` protocol methods `as-file` and `as-url` on the `Path` class.

### Technical changes
* The return value of `fs/create-parent-directories!` has changed from returning the `File` denoting the innermost created directory to simply returning `nil`. The old return value was not used anywhere.

### DIFF
--- a/editor/src/clj/util/path.clj
+++ b/editor/src/clj/util/path.clj
@@ -1,0 +1,371 @@
+;; Copyright 2020-2026 The Defold Foundation
+;; Copyright 2014-2020 King
+;; Copyright 2009-2014 Ragnar Svensson, Christian Murray
+;; Licensed under the Defold License version 1.0 (the "License"); you may not use
+;; this file except in compliance with the License.
+;;
+;; You may obtain a copy of the License, together with FAQs at
+;; https://www.defold.com/license
+;;
+;; Unless required by applicable law or agreed to in writing, software distributed
+;; under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+;; CONDITIONS OF ANY KIND, either express or implied. See the License for the
+;; specific language governing permissions and limitations under the License.
+
+(ns util.path
+  (:require [clojure.java.io :as io]
+            [util.defonce :as defonce])
+  (:import [clojure.lang IReduceInit]
+           [java.io BufferedInputStream BufferedOutputStream File]
+           [java.net URI URL]
+           [java.nio.charset Charset StandardCharsets]
+           [java.nio.file FileVisitResult FileVisitor Files LinkOption NoSuchFileException OpenOption Path StandardOpenOption]
+           [java.nio.file.attribute BasicFileAttributes FileAttribute FileTime]
+           [java.util Set]))
+
+(set! *warn-on-reflection* true)
+(set! *unchecked-math* :warn-on-boxed)
+
+(defonce ^:private ^FileAttribute/1 empty-file-attribute-array (make-array FileAttribute 0))
+(defonce ^:private ^LinkOption/1 empty-link-option-array (make-array LinkOption 0))
+(defonce ^:private ^String/1 empty-string-array (make-array String 0))
+(defonce ^:private ^OpenOption/1 append-open-options (into-array OpenOption [StandardOpenOption/WRITE StandardOpenOption/CREATE StandardOpenOption/APPEND]))
+(defonce ^:private ^OpenOption/1 overwrite-open-options (into-array OpenOption [StandardOpenOption/WRITE StandardOpenOption/CREATE StandardOpenOption/TRUNCATE_EXISTING]))
+
+(defonce/protocol Coercions
+  "Convert to Path objects."
+  (^Path as-path [this] "Coerce argument to a Path."))
+
+(extend-protocol Coercions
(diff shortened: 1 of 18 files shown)


editor/src/clj/util/path.clj
```diff
@@ -0,0 +1,371 @@
...
+
+(defn relativize
+  "Coerces both arguments to java.nio.file.Path and returns a relative Path
+  that, when resolved against this path, yields a path that locates the same
+  file as the other path. For example, on UNIX, if this path is `/a/b` and the
+  other path is `/a/b/c/d`, then the resulting relative path would be `c/d`."
+  ^Path [this other]
+  (.relativize (as-path this) (as-path other)))
+
+(defn resolve
```
vlaaad, line 157:
```
WARNING: resolve already refers to: #'clojure.core/resolve in namespace: util.path, being replaced by: #'util.path/resolve
```

editor/src/clj/util/path.clj
```diff
@@ -0,0 +1,371 @@
...
+  paths do not exist."
+  [x y]
+  (let [a (as-path x)
+        b (as-path y)]
+    (try
+      (Files/isSameFile a b)
+      (catch NoSuchFileException _
+        false))))
+
+(defn compare
```
vlaaad, line 270:
```
WARNING: compare already refers to: #'clojure.core/compare in namespace: util.path, being replaced by: #'util.path/compare
```

editor/src/clj/util/path.clj
```diff
@@ -0,0 +1,371 @@
...
+   (as-path x))
+  (^Path [parent child]
+   (let [child-path ^Path (as-path child)]
+     (if (.isAbsolute child-path)
+       (throw (IllegalArgumentException. (str child " is not a relative path")))
+       (.resolve ^Path (as-path parent) child-path))))
+  (^Path [parent child & children]
+   (reduce path (path parent child) children)))
+
+(defn normalized-path
```
vlaaad: Bike shedding: rename to `normalize`?

editor/src/clj/util/path.clj
```diff
@@ -0,0 +1,371 @@
...
+  "Returns a normalized java.nio.file.Path, passing each argument to as-path.
+  When called with multiple arguments, the first argument is treated as the
+  parent and later arguments as children relative to the parent."
+  (^Path [x]
+   (.normalize (as-path x)))
+  (^Path [x & xs]
+   (.normalize ^Path (apply path x xs))))
+
+(defn absolute-path
```
vlaaad: Bike shedding: rename to `absolute`?

editor/src/clj/util/path.clj
```diff
@@ -0,0 +1,371 @@
...
+  "Returns an absolute java.nio.file.Path, passing each argument to as-path.
+  When called with multiple arguments, the first argument is treated as the
+  parent and later arguments as children relative to the parent."
+  (^Path [x]
+   (.toAbsolutePath (as-path x)))
+  (^Path [x & xs]
+   (.toAbsolutePath ^Path (apply path x xs))))
+
+(defn real-path
```
vlaaad: Bike shedding: rename to `real`?

editor/src/clj/util/path.clj
```diff
@@ -0,0 +1,371 @@
...
+  ^Set [x]
+  (Files/getPosixFilePermissions (as-path x) empty-link-option-array))
+
+(defn exists?
+  "Coerces the argument to a java.nio.file.Path and returns true if a
+  corresponding file system entry exists."
+  [x]
+  (Files/exists (as-path x) empty-link-option-array))
+
+(defn existing-directory?
```
vlaaad: Bike shedding: rename to `directory?`? If a path is a directory, it already implies it exists, because without existing, you can't say if it's a file or a directory.

editor/src/clj/util/path.clj
```diff
@@ -0,0 +1,371 @@
...
+  "Coerces the argument to a java.nio.file.Path and returns true if a
+  corresponding file system entry exists and is a directory."
+  [x]
+  (Files/isDirectory (as-path x) empty-link-option-array))
+
+(defn existing-file?
```
vlaaad: Bike shedding: rename to `file?`? If a path is a file, it already implies it exists, because without existing, you can't say if it's a file or a directory.

vlaaad (changes requested):
Looks good, but let's get rid of warnings before merging.

There are some bike-shedding comments; feel free to ignore these. Most of these are about removing `path` from fn names, since the word `path` is going to be there in the ns name (e.g.,`path/normalize` vs `path/normalized-path`)

vlaaad (approved):
Ship it! 🚀 🚀 🚀

PR #11776, opened 2026-01-22
Hotfix: Return nil when count is not 1, check if extensions is a string or coll

JosephFerano:
Hotfix for some cases I didn't check when dragging and dropping a resource into the properties control view.


vlaaad (approved):
LGTM!

PR #11787, opened 2026-01-23
Significantly improve load time and memory consumption of large projects

vlaaad:
On some large projects:
- The time to open the project was reduced by 50% (4m00s->2m01s on an example project)
- Memory consumption after load was reduced by 71% (10.59GB->3.04Gb on the same project)
- The time to perform a full build improved by 40% (13m50s->8m11s). 

This was achieved by delaying the computation of data that is used for applying changes in the editor until it is actually needed, rather than computing it eagerly when loading the project. 

The downside is that in some cases, the first time a change to a particular element is made, it will take longer than before (worst case: 6s->35s). It is noticeable only in some rare cases in large projects when changing resources that are widely used throughout the project (e.g., when changing a GUI button layout when the button is referenced in hundreds/thousands of places). Subsequent changes to the same element will be faster than before. 

Overall, the tradeoff is worth it, because previously, every user had to always pay this cost when opening the project, while now, the cost is only paid when actually making certain changes to edge case elements, and even then, the wait is significantly reduced (30s instead of 2m).

#### Load time before vs after
```sh
# before
2026-01-21 15:54:52.922 INFO  default    editor.defold-project - {:line nil, :message "Project loading completed in 4:00"}
# after 
2026-01-21 15:48:38.405 INFO  default    editor.defold-project - {:line nil, :message "Project loading completed in 2:01"}
```
Significant improvement 🚀 
#### Memory use before vs after
```sh
# before 
2026-01-21 15:55:05.791 INFO  default    editor.defold-project - {:line nil, :message "Project loaded", :allocated "10.59 GB"}
# after 
2026-01-21 15:48:39.419 INFO  default    editor.defold-project - {:line nil, :message "Project loaded", :allocated "3.04 GB"}
```
Significant improvement 🚀 
#### Worst case edit (first change)
```sh
# before
"Elapsed time: 6496.768416 msecs"
# after 
"Elapsed time: 35279.083125 msecs"
```
Got worse 😢 Still, waiting 30s in some rare cases is better than always waiting for 2 extra minutes during project load
#### More representative case edit (first change)
```sh
# before 
"Elapsed time: 2.802042 msecs"
# after
"Elapsed time: 5.771083 msecs"
```
Got worse 😢 But it's not noticeable, it's less than a frame.
#### Worst case edit (subsequent changes)
```sh
# before
"Elapsed time: 6516.027458 msecs"
# after 
"Elapsed time: 2212.596625 msecs"
```
Significant improvement, though still bad.
#### More representative case edit (subsequent changes)
```sh 
# before
"Elapsed time: 1.4385 msecs"
# after 
"Elapsed time: 0.873125 msecs"
```
Improvement, but not really noticeable.
#### Build time
```sh
# before 
2026-01-22 11:47:51.715 INFO  default    editor.build - {:line nil, :message "Build", :elapsed "13:50", :allocated "3.76 GB", :heap "14.53 GB"}
# after 
2026-01-22 11:25:38.114 INFO  default    editor.build - {:line nil, :message "Build", :elapsed "8:11", :allocated "4.36 GB", :heap "7.60 GB"}
```
Significant improvement 🚀 

Related to #10166

### DIFF
--- a/editor/src/clj/editor/defold_project.clj
+++ b/editor/src/clj/editor/defold_project.clj
@@ -1784,26 +1784,26 @@
               (g/make-nodes plugin-graph [code-transpilers code.transpilers/CodeTranspilersNode]
                 (g/connect code-preprocessors :lua-preprocessors code-transpilers :lua-preprocessors)))))
         project-id
-    (second
-          (g/tx-nodes-added
-            (g/transact
-              (g/make-nodes graph
-                  [script-intelligence si/ScriptIntelligenceNode
-                   project [Project :workspace workspace-id]
-                   script-annotations script-annotations/ScriptAnnotations
-                   editor-localization-bundle editor-localization-bundle/EditorLocalizationBundle]
-                (g/connect workspace-id :root script-annotations :root)
-                (g/connect script-annotations :_node-id project :script-annotations)
-                (g/connect editor-localization-bundle :_node-id project :editor-localization-bundle)
-                (g/connect extensions :_node-id project :editor-extensions)
-                (g/connect script-intelligence :_node-id project :script-intelligence)
-                (g/connect workspace-id :build-settings project :build-settings)
-                (g/connect workspace-id :dependencies project :dependencies)
-                (g/connect workspace-id :resource-list project :resources)
-                (g/connect workspace-id :resource-map project :resource-map)
-                (g/set-graph-value graph :project-id project)
-                (g/set-graph-value graph :lsp (lsp/make project get-resource-node))
-                (g/set-graph-value graph :code-transpilers transpilers-id)))))]
+        (second
+              (g/tx-nodes-added
+                (g/transact
+                  (g/make-nodes graph
+                      [script-intelligence si/ScriptIntelligenceNode
+                       project [Project :workspace workspace-id]
+                       script-annotations script-annotations/ScriptAnnotations
+                       editor-localization-bundle editor-localization-bundle/EditorLocalizationBundle]
+                    (g/connect workspace-id :root script-annotations :root)
+                    (g/connect script-annotations :_node-id project :script-annotations)
+                    (g/connect editor-localization-bundle :_node-id project :editor-localization-bundle)
+                    (g/connect extensions :_node-id project :editor-extensions)
+                    (g/connect script-intelligence :_node-id project :script-intelligence)
+                    (g/connect workspace-id :build-settings project :build-settings)
+                    (g/connect workspace-id :dependencies project :dependencies)
+                    (g/connect workspace-id :resource-list project :resources)
+                    (g/connect workspace-id :resource-map project :resource-map)
+                    (g/set-graph-value graph :project-id project)
+                    (g/set-graph-value graph :lsp (lsp/make project get-resource-node))
+                    (g/set-graph-value graph :code-transpilers transpilers-id)))))]
     (reload-plugins! project-id (g/node-value project-id :resources))
     (workspace/add-resource-listener! workspace-id 1 (ProjectResourceListener. project-id))
     (g/reset-undo! graph)
--- a/editor/src/clj/internal/graph.clj
+++ b/editor/src/clj/internal/graph.clj
@@ -13,11 +13,11 @@
 ;; specific language governing permissions and limitations under the License.
 
 (ns internal.graph
-  (:require [clojure.core.reducers :as r]
-            [clojure.data.int-map :as int-map]
+  (:require [clojure.data.int-map :as int-map]
             [internal.graph.types :as gt]
             [internal.node :as in]
             [internal.util :as util]
+            [util.array :as array]
             [util.coll :as coll :refer [pair]]
             [util.defonce :as defonce])
   (:import [clojure.lang IPersistentSet Indexed]
@@ -374,11 +374,22 @@
   (and (p (.source-id arc) (.source-label arc))
        (p (.target-id arc) (.target-label arc))))
 
+;; Referentially transparent cache that supports explicit invalidation
+;; and on-demand computation.
+;; `cache` is an atom for the "current generation" of successors: when
+;; invalidated, a new atom + new Successors instance is created.
+;; `cache` value is a map: {node-id {output Endpoint/1}}
+(defonce/type Successors [cache])
+
+(defn- make-successors
+  ^Successors ([] (make-successors {}))
+  ^Successors ([m] (->Successors (atom m))))
+
 (defn empty-graph
   []
   {:nodes (int-map/int-map)
    :sarcs {}
-   :successors {}
+   :successors (make-successors)
    :tarcs {}
    :tx-id 0})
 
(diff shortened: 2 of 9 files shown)


editor/src/clj/editor/defold_project.clj
```diff
@@ -1784,26 +1784,26 @@
...
+                    (g/connect editor-localization-bundle :_node-id project :editor-localization-bundle)
+                    (g/connect extensions :_node-id project :editor-extensions)
+                    (g/connect script-intelligence :_node-id project :script-intelligence)
+                    (g/connect workspace-id :build-settings project :build-settings)
+                    (g/connect workspace-id :dependencies project :dependencies)
+                    (g/connect workspace-id :resource-list project :resources)
+                    (g/connect workspace-id :resource-map project :resource-map)
+                    (g/set-graph-value graph :project-id project)
+                    (g/set-graph-value graph :lsp (lsp/make project get-resource-node))
+                    (g/set-graph-value graph :code-transpilers transpilers-id)))))]
```
vlaaad, line 1806: Unrelated: indentation.

editor/src/clj/internal/graph.clj
```diff
@@ -837,6 +848,75 @@
...
+                       ;; The connected nodes and their outputs.
+                       (doseq [^Arc outgoing-arc outgoing-arcs
+                               :let [target-id (.target-id outgoing-arc)
+                                     target-label (.target-label outgoing-arc)]
+                               dep-label (get (input-deps basis target-id) target-label)]
+                         (.add deps (gt/endpoint target-id dep-label)))
+
+                       (if (.isEmpty deps)
+                         empty-endpoints-array
+                         (.toArray deps ^Endpoint/1 empty-endpoints-array)))
```
vlaaad, line 918: Side note: previous version of this code looked like this:
```clj
(let [^"[Linternal.graph.types.Endpoint;" storage (make-array Endpoint (.size deps))
      endpoint-array (.toArray deps storage)] ; Use a typed array to aid memory profilers.
  ...)
```
It used to create an array of a target size, while now I use an empty array as an argument to `.toArray` fn. This is because it's more performant: JVM does not spend time creating an array and filling it with zeroes; instead, it immediately fills it with the right data.

editor/src/clj/editor/defold_project.clj
matgis: But it's still wrong? By all means, fix it, but let's fully fix it. 😅

editor/src/clj/internal/graph.clj
```diff
@@ -374,11 +374,22 @@
   (and (p (.source-id arc) (.source-label arc))
        (p (.target-id arc) (.target-label arc))))
 
+;; Referentially transparent cache that supports explicit invalidation
+;; and on-demand computation.
+;; `cache` is an atom for the "current generation" of successors: when
+;; invalidated, a new atom + new Successors instance is created.
+;; `cache` value is a map: {node-id {output Endpoint/1}}
```
matgis, line 381: Maybe clarify `output` to `output-label` in this comment?

editor/src/clj/internal/graph.clj
```diff
@@ -374,11 +374,22 @@
...
+(defonce/type Successors [cache])
+
+(defn- make-successors
+  ^Successors ([] (make-successors {}))
+  ^Successors ([m] (->Successors (atom m))))
```
matgis: Maybe add a `(map? m)` precondition here?

editor/src/clj/internal/graph.clj
```diff
@@ -837,6 +848,75 @@
...
+                (if (nil? outputs)
+                  (dissoc! acc node-id)
+                  (let [output->endpoints (reduce dissoc! (transient output->endpoints) outputs)]
+                    (if (zero? (count output->endpoints))
+                      (dissoc! acc node-id)
+                      (assoc! acc node-id (persistent! output->endpoints))))))))
+          (transient current-state)
+          changes)))))
+
+(defn- input-deps [basis node-id] (some-> (gt/node-by-id-at basis node-id) gt/node-type in/input-dependencies))
```
matgis: Style: `def` is fine on a single line, but we don't do that for `defn`.

editor/src/clj/internal/graph.clj
```diff
@@ -837,6 +848,75 @@
                  result')
           (persistent! result'))))))
 
+(defn- update-graph-successors
```
matgis: Should we rename this to `invalidate-graph-successors` now for clarity?

editor/src/clj/internal/graph.clj
```diff
@@ -1159,89 +1238,12 @@
...
-          changes+old-node-successors)]
-    (persistent!
-      (reduce dissoc!
-              (transient (into old-successors new-successors))
-              removed-successor-entries))))
-
 (defn update-successors
   [basis changes]
-  ;; changes = {node-id #{outputs}}
+  ;; changes = {node-id #{outputs}|nil}
```
matgis: Perhaps we should take this opportunity to clarify the comment here and in `update-graph-successors` above. In particular, it's not clear what `#{outputs}` means. Is it a set of `Endpoints`? A set of output labels? Or something else? I think this should say something like `{node-id #{output-label ...}|nil}`.

editor/src/clj/internal/graph.clj
```diff
@@ -1159,89 +1238,12 @@
...
   (reduce (fn [basis [graph-id changes]]
-            (if-let [graph (get (:graphs basis) graph-id)]
-              (update-in basis [:graphs graph-id :successors] update-graph-successors basis graph changes)
+            (if (contains? (:graphs basis) graph-id)
+              (update-in basis [:graphs graph-id :successors] update-graph-successors changes)
               basis))
           basis
-          (util/group-into (comp gt/node-id->graph-id first) changes)))
+          (util/group-into {} {} (comp gt/node-id->graph-id first) changes)))
```
matgis: We now group the changes into a hash-map instead of a vector, and use `reduce-kv` inside `update-graph-successors`, but is this really faster than reducing over a vector of `node-id+outputs` pairs? I don't think we should change this unless we see a performance improvement. I fear bulk invalidations resulting from a large resource-sync (i.e. switching branches) might be slower as a result of this change.

editor/src/clj/internal/graph.clj
```diff
@@ -837,6 +848,75 @@
...
+(defn- update-graph-successors
+  ^Successors [^Successors successors changes]
+  ;; changes = {node-id #{outputs}|nil}
+  ;; when changes value is nil, it means that every output was invalidated
+  (let [current-state @(.-cache successors)]
+    (make-successors
+      (persistent!
+        (reduce-kv
+          (fn [acc node-id outputs]
+            (let [output->endpoints (acc node-id ::not-found)]
```
matgis: I'm not sure, but it may be faster to look up `output->endpoints` in `current-state` rather than in `acc`, since we're modifying `acc` in the process. 🤔Given how the input is structured, the outcome should be the same.

editor/test/integration/gui_test.clj
```diff
@@ -1110,7 +1110,7 @@
    (has-successor? (g/now) source-id+source-label target-id+target-label))
   ([basis [source-id source-label] [target-id target-label]]
    (let [graph-id (g/node-id->graph-id source-id)
-         successor-endpoint-array (get-in basis [:graphs graph-id :successors source-id source-label])
+         successor-endpoint-array (#'internal.graph/query-successors (get-in basis [:graphs graph-id :successors]) basis source-id source-label)
```
matgis: We seem to do this in a few places, so why not add a helper function to `g/` instead of calling this private internal function?

matgis (changes requested):
A very nice improvement! However, I found two places where `:successors` is still used as a regular `hash-map` that we should fix before merging:
* The private `endpoint-successors` function in `editor.reveal`.
* The private `successor-pairs` function in `dev`.

vlaaad:
Aww shucks, I forgot to update the dev tools! Good catch!

editor/test/integration/gui_test.clj
vlaaad: Because it is private and internal to the implementation. Maybe it's okay to use it in tests or dev tools, but definitely not in the production code...

editor/src/clj/internal/graph.clj
vlaaad: I'm not sure either. Doing lookups on one big map might be faster than doing them on two different maps from the memory cache perspective. Not sure it's worth investigating here...

editor/test/integration/gui_test.clj
vlaaad: I'll add a `g/successors` fn that states that it's only for tests/introspection.

matgis (approved):
Ship it! ❤️

PR #11795, opened 2026-01-26
Fix http.request regression

britzl:
Fixes #11790

### DIFF
--- a/engine/dlib/src/dlib/http_client.cpp
+++ b/engine/dlib/src/dlib/http_client.cpp
@@ -305,6 +305,10 @@ namespace dmHttpClient
         {
             memcpy(&client->m_ProxyURI, proxyURI, sizeof(*proxyURI));
         }
+        else
+        {
+            memset(&client->m_ProxyURI, 0, sizeof(client->m_ProxyURI));
+        }
 
         return client;
     }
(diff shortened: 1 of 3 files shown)


engine/dlib/src/dlib/http_client.cpp
```diff
@@ -305,6 +305,10 @@ namespace dmHttpClient
         {
             memcpy(&client->m_ProxyURI, proxyURI, sizeof(*proxyURI));
         }
+        else
```
britzl: The fix

AGulev (dismissed):
LGTM

engine/dlib/src/dlib/http_client.cpp
JCash: Client is an old struct, and "new Client()" suggests it has a constructor, but it doesn't :/
One could also imagine creating a constructor that does a memset(0) on the whole struct. It's a pattern we use a lot.

JCash (dismissed):

engine/dlib/src/dlib/http_client.cpp
britzl: That is a good point. Maybe I should add:

```cpp
        Client()
        {
            memset(this, 0, sizeof(*this));
        }
```

To the Client struct and get rid of the explicit memset of `client->m_Statistics` and `client->m_ProxyURI` ?

JCash (approved):

aglitchman:
Will be the fix included in the beta?.. https://github.com/defold/defold/commits/beta (not cherry-picked, not merged)

AGulev:
> Will be the fix included in the beta?.. https://github.com/defold/defold/commits/beta (not cherry-picked, not merged)

Sorry for that, it's building now

PR #11797, opened 2026-01-26
Throw an error when the `build` folder inside the project folder is used for bundling

AGulev:
Notify the user that the `build` folder inside the project folder cannot be used for bundling.

Fix https://github.com/defold/defold/issues/11771

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
@@ -728,6 +728,9 @@ public class Project {
      */
     public List<TaskResult> build(IProgress monitor, String... commands) throws IOException, CompileExceptionError, MultipleCompileException {
         try {
+            if (isBundleCommandRequested(commands)) {
+                getBundleOutputDirectory();
+            }
             TimeProfiler.start("loadProjectFile");
             loadProjectFile(true);
             TimeProfiler.stop();
@@ -753,6 +756,18 @@ public class Project {
         } catch (Throwable e) {
             throw new CompileExceptionError(null, 0, e.getMessage(), e);
         }
+    }
+
+    private boolean isBundleCommandRequested(String... commands) {
+        if (commands == null) {
+            return false;
+        }
+        for (String command : commands) {
+            if ("bundle".equals(command)) {
+                return true;
+            }
+        }
+        return false;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -728,6 +728,9 @@ public List<IResource> getPropertyFilesAsResources() {
      */
     public List<TaskResult> build(IProgress monitor, String... commands) throws IOException, CompileExceptionError, MultipleCompileException {
         try {
+            if (isBundleCommandRequested(commands)) {
```
AGulev: check to fail earlier.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/AndroidBundler.java
```diff
@@ -594,7 +594,7 @@ private File createAppBundleBaseZip(Project project, File outDir, File apk, ICan
             BundleHelper.throwIfCanceled(canceled);
             return baseZip;
         } catch (Exception e) {
-            throw new CompileExceptionError("Failed creating AAB base.zip", e);
+            throw new CompileExceptionError("Failed creating AAB base.zip. Cause: " + String.valueOf(e.getMessage()), e);
```
AGulev, line 597: More info about the actual fail

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -755,6 +758,18 @@ public List<TaskResult> build(IProgress monitor, String... commands) throws IOEx
         }
     }
 
+    private boolean isBundleCommandRequested(String... commands) {
+        if (commands == null) {
```
britzl: Can commands be nil? In Bob.java we default to ["build"].

Also, we do a similar search of commands in Bob.java where we search the commands for "resolve". Perhaps this could be moved to a shared utility function?

Also there's `Arrays.asList(array).contains(x);`.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -917,17 +932,27 @@ private void bundle(IProgress monitor) throws IOException, CompileExceptionError
         m.done();
     }
 
-    private File getBundleOutputDirectory() {
+    private File getBundleOutputDirectory() throws IOException, CompileExceptionError {
```
britzl: I don't like that we call this function getBundleOutputDirectory() since it now does more than get the bundle output directory. Could we have a separate function to check if the folder is valid and then throw? Something like this:

```java
File output = getBundleOutputDirectory();
if (!isValidBundleOutputDirectory(output)) {
    throw new CompileExceptionError("Build folder '" + output + "' in the project folder can't be used for bundling as this folder is reserved for Defold build system.");
}
```

britzl: Shouldn't it be "Bundle folder" and not "Build folder" in the message?

britzl (changes requested):
See comments

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
AGulev: Fully agree

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
AGulev: yes, that will be much better

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -917,17 +921,26 @@ private void bundle(IProgress monitor) throws IOException, CompileExceptionError
...
+            bundleDir = new File(FilenameUtils.concat(getRootDirectory(), getBuildDirectory()));
         }
         return bundleDir;
     }
 
+    private void validateBundleOutputDirectory(File bundleDir) throws IOException, CompileExceptionError {
+        File buildDir = new File(FilenameUtils.concat(getRootDirectory(), "build"));
+        Path bundlePath = bundleDir.getCanonicalFile().toPath();
+        Path buildPath = buildDir.getCanonicalFile().toPath();
+        if (bundlePath.startsWith(buildPath) && !bundlePath.toString().contains(getBuildDirectory())) {
```
AGulev, line 939: an extra condition to make sure html5 build is fine

AGulev:
Ready for review (should I pick it into beta?)

britzl (approved):

PR #11799, opened 2026-01-27
Improve project load time on big projects by 9%

vlaaad:
Now the editor opens big projects a bit faster. As measured on an example big project, the load time reduced from 2m01s to 1m50s.

Technical notes:

Related to #10166

### DIFF
--- a/editor/src/clj/editor/defold_project.clj
+++ b/editor/src/clj/editor/defold_project.clj
@@ -91,9 +91,6 @@
 
 (defn- load-resource-node [project resource-node-id resource source-value transpiler-tx-data-fn]
   (try
-    ;; TODO(save-value-cleanup): This shouldn't be able to happen anymore. Remove this check after some time in the wild.
-    (assert (and (not= :folder (resource/source-type resource))
-                 (resource/exists? resource)))
     (let [{:keys [read-fn load-fn] :as resource-type} (resource/resource-type resource)
           transpiler-tx-data (transpiler-tx-data-fn resource-node-id resource)]
       (cond-> []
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/defold_project.clj
```diff
@@ -91,9 +91,6 @@
 
 (defn- load-resource-node [project resource-node-id resource source-value transpiler-tx-data-fn]
   (try
-    ;; TODO(save-value-cleanup): This shouldn't be able to happen anymore. Remove this check after some time in the wild.
-    (assert (and (not= :folder (resource/source-type resource))
-                 (resource/exists? resource)))
```
vlaaad, line 96: I removed 1 out of 2 such asserts. I checked in Sentry: both errors get regularly reported. I don't think it's possible to avoid triggering these asserts since you can't prevent users from removing files on the disc while the editor opens the project...

The other assertion didn't contribute much to the load time because the code that performs an assertion is parallelized over resources.

editor/src/clj/editor/defold_project.clj
```diff
@@ -1785,25 +1782,25 @@
...
+                (g/connect editor-localization-bundle :_node-id project :editor-localization-bundle)
+                (g/connect extensions :_node-id project :editor-extensions)
+                (g/connect script-intelligence :_node-id project :script-intelligence)
+                (g/connect workspace-id :build-settings project :build-settings)
+                (g/connect workspace-id :dependencies project :dependencies)
+                (g/connect workspace-id :resource-list project :resources)
+                (g/connect workspace-id :resource-map project :resource-map)
+                (g/set-graph-value graph :project-id project)
+                (g/set-graph-value graph :lsp (lsp/make project get-resource-node))
+                (g/set-graph-value graph :code-transpilers transpilers-id)))))]
```
vlaaad, line 1803: Unrelated: indentation :D

editor/src/clj/editor/dialogs.clj
```diff
@@ -86,7 +89,7 @@
       (dissoc :size :header :content :footer :root-props)
       (assoc :fx/type fxui/dialog-stage
              :scene {:fx/type fx.scene/lifecycle
-                     :stylesheets [(str (io/resource "dialogs.css"))]
+                     :stylesheets [@dialogs-css-delay]
```
vlaaad, line 92: This was invoked on every dialog refresh; for a big project, looking up a resource on a classpath contributed around 3 seconds out of 2 minutes...

matgis: This sounds a bit strange. Wouldn't it be the same for a small project? `io/resource` should only look for resources on the classpath. Is there something deeper going on here? 🤔

editor/src/clj/editor/defold_project.clj
matgis: Yeah, I'm pretty confident we can remove this now. I've also discovered these asserts can be tripped by symlinks that refer to missing file system entries, which could explain some of the Sentry reports.

matgis (approved):
Nice finds! LGTM! ❤️

editor/src/clj/editor/dialogs.clj
vlaaad: The function was called on every data update, e.g., when we want to show a new progress message (`"Reading /foo..."`, etc.), so performance suffered in proportion to the number of resources.

PR #11801, opened 2026-01-27
Custom game.project fields now show the correct UI after editor restart

### LINKED ISSUE #11293: ext.properties "resource" not keeping input controls after restart
Properties defined in an ext.properties with `type = resource` will lose the controls after an editor restart.

**To Reproduce (REQUIRED)**
1. Create an `ext.properties`.
2. Define a property.
```
[test]
my_property.type = resource
```
4. Open your game.project.
5. Insert a resource into the field.
6. Restart the editor.
7. Open your game.project, there are no longer any input controls for the property.

Input controls for properties should work after a restart

vlaaad:
Previously, custom values defined in `game.project` via the `game.properties` or `ext.properties` would always be shown as strings upon restarting the editor. Now they are displayed with their correct types (e.g., checkboxes for booleans, list views for lists).

### Technical changes

I implemented 2 changes:
1. Made custom settings override unknown string settings, so the UI shows the expected UI
2. Added guards for situations where existing `ResourceSettingNode` connections don't match the settings meta, i.e., we might have a node for a setting, but change its type to string, or we might have no node, and we changed the setting type to be a resource.

Fixes #11793
Fixes #11293

### DIFF
--- a/editor/src/clj/editor/settings.clj
+++ b/editor/src/clj/editor/settings.clj
@@ -23,7 +23,8 @@
             [editor.settings-core :as settings-core]
             [editor.validation :as validation]
             [editor.workspace :as workspace]
-            [util.coll :as coll]))
+            [util.coll :as coll]
+            [util.eduction :as e]))
 
 (g/defnode ResourceSettingNode
   (property resource-connections g/Any) ; [target-node-id [connections]]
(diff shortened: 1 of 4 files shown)


AGulev:
I think it also fixes https://github.com/defold/defold/issues/11293 ?

editor/src/clj/editor/settings.clj
```diff
@@ -189,14 +190,28 @@
...
+                                  (fn [{:keys [path value] :as setting}]
+                                    ;; We resolve raw string values to resources
+                                    ;; so the form gets typed values; this may
+                                    ;; happen when a :resource setting is
+                                    ;; defined via, e.g., game.properties,
+                                    ;; without a corresponding
+                                    ;; ResourceSettingNode in the graph.
+                                    (cond-> setting
+                                            (and (string? value) (= :resource (:type (meta-settings-map path))))
+                                            (assoc :value (workspace/resolve-resource owner-resource value))))))
```
matgis: We should probably inject `^:unsafe evaluation-context` and supply it to `workspace/resolve-resource` here.

I can't help but feel we need to do a better take on this at some point - really, we have a dependency on the `resource-map` output of the `workspace` that we're not expressing in the graph. And it's also messed up that we don't establish connections between the resource nodes and the `SettingsNode` - we don't currently update the resource references when you move a resource.

Overall, I don't think this is the correct place to do this, but doing it correctly feels like a much larger task. At least it fixes a bug! 🤷

matgis (dismissed):

editor/src/clj/editor/settings.clj
vlaaad: Yes, exactly, I considered doing it the right way, but it involves loading all `.properties` files before `game.project`, and re-creating the resource setting nodes on typing in these files, which is much, much more involved, so I thought doing a quick fix for the most common use case is better here...

PR #11803, opened 2026-01-27
Limit max dialog height to the height of the screen

vlaaad:
Dialog height is now limited to the height of the screen to prevent dialog content from overflowing outside of the visible area.

### Technical changes

We recently limited dialog heights to some reasonable value. It was not enough for computers with very small screens. This change ensures we don't exceed the height of the screen visual bounds. 

Fixes #11786

### DIFF
--- a/editor/src/clj/editor/dialogs.clj
+++ b/editor/src/clj/editor/dialogs.clj
@@ -55,7 +55,7 @@
            [javafx.event Event]
            [javafx.scene.control ListView TextField]
            [javafx.scene.input KeyCode KeyEvent MouseButton MouseEvent]
-           [javafx.stage DirectoryChooser FileChooser FileChooser$ExtensionFilter Stage Window]
+           [javafx.stage DirectoryChooser FileChooser FileChooser$ExtensionFilter Screen Stage Window]
            [org.apache.commons.io FilenameUtils]))
 
 (set! *warn-on-reflection* true)
@@ -65,6 +65,15 @@
 
 (def ^:private dialogs-css-delay
   (delay (str (io/resource "dialogs.css"))))
+
+(defn max-dialog-stage-height []
+  (let [screens (Screen/getScreens)
+        n (.size screens)]
+    (loop [i 0
+           min-height Double/MAX_VALUE]
+      (if (= i n)
+        min-height
+        (recur (unchecked-inc i) (min min-height (.getHeight (.getVisualBounds ^Screen (.get screens i)))))))))
 
 (defn dialog-stage
   "Dialog `:stage` that manages scene graph itself and provides layout common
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/dialogs.clj
```diff
@@ -66,6 +66,15 @@
...
 
+(defn max-dialog-stage-height []
+  (let [screens (Screen/getScreens)
+        n (.size screens)]
+    (loop [i 0
+           min-height Double/MAX_VALUE]
+      (if (= i n)
+        min-height
+        (recur (unchecked-inc i) (min min-height (.getHeight (.getVisualBounds ^Screen (.get screens i)))))))))
+
```
matgis, line 77: I'm unsure how Cljfx works internally. Will this get called a lot, or will the `dialog-stage` only be called if its inputs change?

If it gets called a lot and querying the screen dimensions is somewhat costly, we might want to put this in a delay instead. Seems like it would be pretty uncommon for a new smaller screen to appear? 🤔

editor/src/clj/editor/dialogs.clj
```diff
@@ -88,6 +97,7 @@
   (-> props
       (dissoc :size :header :content :footer :root-props)
       (assoc :fx/type fxui/dialog-stage
+             :max-height (max-dialog-stage-height)
```
matgis, line 100: Should we really use the `:max-height`? Could we use `:pref-height` instead so that I could at least adjust the height after I move the dialog to a larger screen?

editor/src/clj/editor/dialogs.clj
```diff
@@ -66,6 +66,15 @@
 (def ^:private dialogs-css-delay
   (delay (str (io/resource "dialogs.css"))))
 
+(defn max-dialog-stage-height []
+  (let [screens (Screen/getScreens)
```
matgis, line 70: An unfortunate side effect of this approach is that all dialogs will be adjusted to the height of the smallest screen. I think it is a fairly common use case to maybe have a 13-inch laptop connected to a 27-inch monitor. I would feel pretty frustrated about cramped dialogs on my main screen as a side effect of using my laptop monitor as a side screen.

I think we should either limit the height to the largest screen or to the screen we're about to show the dialog on. Perhaps we could use `Screen/getScreensForRectangle` instead of `Screen/getScreens` to achieve this somehow?

matgis (changes requested):
I think this change will be annoying to users with a mix of large and small displays. I put some suggestions in the code comments.

editor/src/clj/editor/dialogs.clj
vlaaad: Only when inputs change. I measured this function invocation; it can run 10 million times in 200 ms, it's mostly a dirty flag check.

editor/src/clj/editor/dialogs.clj
vlaaad: Good point, I think it makes sense to limit it to the largest screen.

editor/src/clj/editor/dialogs.clj
vlaaad: The execution time mean is only 20 nanoseconds

matgis (approved):

PR #11804, opened 2026-01-27
Handle proxy unload and load during the same frame

### LINKED ISSUE #11764: Error when loading same collection proxy after receiving "unload"

When you unload a collection proxy, and receive the "unload" message, you expect the collection to be fully unloaded by that point such that if you want to attempt loading the same collection proxy again at that time, it should work.  But instead, the engine will crash with the error: `The collection 'test collection' could not be created since there is already a socket with the same name.`

If you wait a few frames, you can re-attempt the load successfully, but it suggests the order of the messages and the actual deletion of the game object are not what they should be.  See discussion thread https://forum.defold.com/t/cant-load-after-proxy-unloaded-since-updating-from-1-9-to-1-12/82244/2

**To Reproduce (REQUIRED)**
The attached project simply loads a collection, unloads it, and upon receiving the "unloaded" message,  attempts to load the same collection again.  This will cause the crash.  There is a commented line that adds a delay before loading, which if you switch to that line instead, will avoid the crash.

1. Run the attached project
2. You will get the crash, see the console.

If the crash is fixed (or you use the "delay" load instead), you will get an infinite loop of loading and unloading.

  - Version 1.12.0

**Platforms (REQUIRED):**
 - Platforms: Linux, but probably all
 - OS: EndeavourOS, but probably all
 - Device: Desktop, but probably all

Attached [proxy_unload_crash.zip](https://github.com/user-attachments/files/24727297/proxy_unload_crash.zip)

**Logs (OPTIONAL):**

**Workaround (OPTIONAL):**
There is a commented line that adds a delay before loading, which if you switch to that line instead, will avoid the crash.

**Screenshots (OPTIONAL):**

@mathiaswking suggested this is due to the recent changes made to the update functions (eg. the addition of `late_update` and the changing of the order of `update` and `fixed_update`. (see [discussion thread](https://forum.defold.com/t/cant-load-after-proxy-unloaded-since-updating-from-1-9-to-1-12/82244/2) )

britzl:
This fixes an issue when unloading and loading the same collection in a single frame.

Fixes #11764

### DIFF
--- a/engine/gameobject/src/gameobject/gameobject.cpp
+++ b/engine/gameobject/src/gameobject/gameobject.cpp
@@ -422,6 +422,27 @@ namespace dmGameObject
         delete collection;
     }
 
+    // Remove a to-be-deleted collection that still owns the given socket so a new
+    // collection can re-use the name immediately.
+    static bool TryDeletePendingCollectionForSocket(HRegister regist, const char* socket_name)
+    {
+        dmhash_t socket_hash = dmHashString64(socket_name);
+        HCollection existing = GetCollectionByHash(regist, socket_hash);
+        if (existing == 0)
+        {
+            return false;
+        }
+
+        Collection* old_collection = existing->m_Collection;
+        if (old_collection == 0 || !old_collection->m_ToBeDeleted)
+        {
+            return false;
+        }
+
+        DeleteCollection(old_collection);
+        return true;
+    }
+
     Result AttachCollection(Collection* collection, const char* name, dmResource::HFactory factory, HRegister regist, HCollection hcollection)
     {
         collection->m_HCollection = hcollection;
@@ -437,6 +458,14 @@ namespace dmGameObject
         for (int i = 0; i < 2; ++i)
         {
             dmMessage::Result result = dmMessage::NewSocket(socket_names[i], sockets[i]);
+            if (result == dmMessage::RESULT_SOCKET_EXISTS)
+            {
+                if (TryDeletePendingCollectionForSocket(regist, socket_names[i]))
+                {
+                    result = dmMessage::NewSocket(socket_names[i], sockets[i]);
+                }
+            }
+
             if (result != dmMessage::RESULT_OK)
             {
                 if (result == dmMessage::RESULT_SOCKET_EXISTS)
--- a/engine/gameobject/src/gameobject/gameobject.cpp
+++ b/engine/gameobject/src/gameobject/gameobject.cpp
@@ -450,6 +450,9 @@ namespace dmGameObject
         hcollection->m_Collection = collection;
         collection->m_Factory = factory;
 
+        // if there exists a collection with the same name and that collection
+        // is to be deleted we immediately delete it so that we can attach the
+        // the new one
         HCollection existing = GetCollectionByHash(regist, dmHashString64(name));
         if (existing && existing->m_Collection->m_ToBeDeleted)
         {
--- a/engine/gameobject/src/gameobject/gameobject.cpp
+++ b/engine/gameobject/src/gameobject/gameobject.cpp
@@ -450,6 +450,12 @@ namespace dmGameObject
         hcollection->m_Collection = collection;
         collection->m_Factory = factory;
 
+        HCollection existing = GetCollectionByHash(regist, dmHashString64(name));
+        if (existing && existing->m_Collection->m_ToBeDeleted)
+        {
+            DeleteCollection(existing->m_Collection);
+        }
+
         char name_frame[128];
         dmStrlCpy(name_frame, name, sizeof(name_frame));
         dmStrlCat(name_frame, "_frame", sizeof(name_frame));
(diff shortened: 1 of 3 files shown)


engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -429,6 +429,15 @@ namespace dmGameObject
         hcollection->m_Collection = collection;
         collection->m_Factory = factory;
 
+        // if there exists a collection with the same name and that collection
+        // is to be deleted we immediately delete it so that we can attach the
+        // the new one
+        HCollection existing = GetCollectionByHash(regist, dmHashString64(name));
+        if (existing && existing->m_Collection->m_ToBeDeleted)
+        {
+            DeleteCollection(existing->m_Collection);
```
britzl, line 438: The question is if deleting the collection here has some unforeseen side-effects? It's usually deleted in PostUpdate().

One might also ask why this this m_ToBeDeleted is needed. Why not immediately delete it in dmGameObject::DeleteCollection(HCollection) instead of just raising the m_ToBeDeleted flag?

JCash: Perhaps due to code flow.
Calling DeleteCollection() will call FinalCollection() and then FinalComponents()

* Do we have a check that FinalCollection isn't called twice?
(comp_collection_proxy.cpp calls `dmGameObject::Final(coll)`)

* Do we accidentally change the order of Lua calls? E.g. skipping a last update before the `final()` call?
Cannot say I have all implications clear in my head, but that's part of the task I imagine.

JCash (changes requested):
I think it'd be good to have a unit test of sorts for this?

engine/gameobject/src/gameobject/gameobject.cpp
britzl: > Do we have a check that FinalCollection isn't called twice?
>  (comp_collection_proxy.cpp calls `dmGameObject::Final(coll)`)

`comp_collection_proxy.cpp` will only call `dmGameObject::Final(coll)` when `proxy->m_Collection != 0` and `proxy->m_Initialized == 1`:

* https://github.com/defold/defold/blob/dev/engine/gamesys/src/gamesys/components/comp_collection_proxy.cpp#L238-L243
* https://github.com/defold/defold/blob/dev/engine/gamesys/src/gamesys/components/comp_collection_proxy.cpp#L293-L297
* https://github.com/defold/defold/blob/dev/engine/gamesys/src/gamesys/components/comp_collection_proxy.cpp#L567-L571

Both m_Collection and m_Initialized are set to 0 immediately when unload starts:

https://github.com/defold/defold/blob/dev/engine/gamesys/src/gamesys/components/comp_collection_proxy.cpp#L507-L508

britzl: > * Do we accidentally change the order of Lua calls? E.g. skipping a last update before the `final()` call?
>   Cannot say I have all implications clear in my head, but that's part of the task I imagine.

Let me check that!

britzl: So this is what it currently looks like if I log from the lifecycle functions of a script component loaded and unloaded from a collection proxy:

```
DEBUG:SCRIPT: loader: loading #collectionproxy
DEBUG:SCRIPT: loader: received proxy_loaded
DEBUG:SCRIPT: proxy: init
DEBUG:SCRIPT: proxy: script fixed_update
DEBUG:SCRIPT: proxy: script update
DEBUG:SCRIPT: proxy: script late_update
DEBUG:SCRIPT: loader: unloading #collectionproxy
DEBUG:SCRIPT: loader: received proxy_unloaded
DEBUG:SCRIPT: loader: loading #collectionproxy
ERROR:GAMEOBJECT: The collection 'proxy' could not be created since there is already a socket with the same name.
ERROR:GAMEOBJECT: AcquireResources NewCollection RESULT_OUT_OF_RESOURCES
WARNING:RESOURCE: Unable to create resource: /main/proxy.collectionc: OUT_OF_RESOURCES
ERROR:GAMESYS: The collection /main/proxy.collectionc could not be loaded.
DEBUG:SCRIPT: proxy: script final
```

And this is what it looks like with this fix applied:

```
DEBUG:SCRIPT: loader: loading #collectionproxy
DEBUG:SCRIPT: loader: received proxy_loaded
DEBUG:SCRIPT: proxy: init
DEBUG:SCRIPT: proxy: script fixed_update
DEBUG:SCRIPT: proxy: script update
DEBUG:SCRIPT: proxy: script late_update
DEBUG:SCRIPT: loader: unloading #collectionproxy
DEBUG:SCRIPT: loader: received proxy_unloaded
DEBUG:SCRIPT: loader: loading #collectionproxy
DEBUG:SCRIPT: proxy: script final
DEBUG:SCRIPT: loader: received proxy_loaded
DEBUG:SCRIPT: proxy: init
DEBUG:SCRIPT: proxy: script fixed_update
DEBUG:SCRIPT: proxy: script update
DEBUG:SCRIPT: proxy: script late_update
```

So the order does not seem to be affected.

JCash (approved):

PR #11806, opened 2026-01-27
Get and set properties from ParticleFX emitters

### LINKED ISSUE #11051: Change material on a particle effect
It is currently not possible to change material on a particle effect component at runtime,

An API to change material. Maybe:

```lua
go.property("mymaterial", resource.material("/path/to/my.material"))

function init(self)
    -- one of
    particlefx.set_material("#mypfx","myemitter", self.mymaterial)
    go.set("#mypfx", "material", self.mymaterial, { emitter = "myemitter" })
end
```

Duplicate versions of the same particle effect but with different materials

Jhonnyg:
The first iteration of getting and setting properties for particlefx emitters via `go.get` / `go.set` is now available:

```lua
go.property("my_material", resource.material())
function init(self)
    local cone_material  = go.get("#particlefx", "material", { keys = {{ emitter_id = "cone_emitter" }}})
    -- change the material for the next time the particlefx is played
    go.set("#particlefx", "material", self.my_material,  { keys = {{ emitter_id = "cone_emitter" }}})
end
```

Important note: The changes invoked via `go.set` will NOT affect the _currently_ playing particlefx, they will only have an effect when playing the fx next time. For example, if you want to change material in an input handler, you need to first stop the particlefx and then play it again:

```lua
function on_input(self, action_id, action)
    if action_id == hash("space") and action.released then
        particlefx.stop("#particlefx")
        go.set("#particlefx", "material", self.my_material, {keys = {{ emitter_id = "cone_emitter" }}})
        particlefx.play("#particlefx")
    end
end
```

The properties we currently support for emitters are:
* "material" - get and set the emitter material
* "image" - get and set the emitter image, i.e atlas or tile source used for rendering
* "animation" - get and set the animation (the animation must exist in the image!)

As this is the first iteration of this feature, very few properties are available. In the future, we hope to support more properties and even curved properties!

Fixes #3417 
Fixes #11051


engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -200,7 +200,6 @@ namespace dmGameSystem
     static const uint8_t VX_DECL_CUSTOM_BUFFER      = 2;
 
     static const dmhash_t PROP_SKIN          = dmHashString64("skin");
-    static const dmhash_t PROP_ANIMATION     = dmHashString64("animation");
```
Jhonnyg, line 203: moved to gamesys.cpp

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
```diff
@@ -44,30 +47,46 @@ DM_PROPERTY_U32(rmtp_ParticleVertexSizeGPU, 0, PROFILE_PROPERTY_FRAME_RESET, "si
...
+    {
+        MaterialResource*   m_Material;
+        TextureSetResource* m_TextureSet;
+        dmhash_t            m_Animation;
+    };
+
+    struct ParticleFXPrototypeOverrides
+    {
+        dmArray<ParticleFXEmitterOverride> m_EmitterOverrides;
+    };
```
Jhonnyg, line 66: We could skip the intermediate ParticleFXPrototypeOverrides struct here, but I'm certain we will build upon this later for MPV2+

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
```diff
@@ -727,6 +846,206 @@ namespace dmGameSystem
...
+                        const char* new_animation_name = dmHashReverseSafe64Alloc(&hash_ctx, current_animation);
+                        dmLogWarning("Atlas doesn't contains animation '%s'. Animation '%s' will be used", error_message, new_animation_name);
+
+                        AddOverrideAnimation(dmGameObject::GetFactory(params.m_Instance), prototype, emitter_index, current_animation);
+                    }
+                    else
+                    {
+                        dmLogWarning("Atlas doesn't contain animation '%s'. No animation will be used", error_message);
+                    }
+                    // else anim_id still == 0
```
Jhonnyg, line 1001: Basically copy-paste from comp_sprite.

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
```diff
@@ -603,17 +653,86 @@ namespace dmGameSystem
...
+                    }
+                }
+                component->m_Overrides = component_overrides;
+            }
+
+            InstanceUserData* instance_user_data = new InstanceUserData;
+            instance_user_data->m_World       = world;
+            instance_user_data->m_ComponentId = component_id;
+            instance_user_data->m_Overrides   = component->m_Overrides;
+            dmParticle::SetInstanceUserData(world->m_ParticleContext, component->m_ParticleInstance, instance_user_data);
```
Jhonnyg: Overrides is mostly a question about ownership, and as it stands now the gamesys/component level manages the override data, which means we need to be able to get hold of some context in the callback that resolves the resources during update.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -196,7 +196,6 @@ namespace dmGameSystem
 
     static const dmhash_t SPRITE_PROP_CURSOR        = dmHashString64("cursor");
     static const dmhash_t SPRITE_PROP_PLAYBACK_RATE = dmHashString64("playback_rate");
-    static const dmhash_t SPRITE_PROP_ANIMATION     = dmHashString64("animation");
```
Jhonnyg, line 199: Merged with comp_model into gamesys.cpp

engine/gameobject/src/gameobject/gameobject_script_util.cpp
```diff
@@ -153,7 +266,7 @@ namespace dmGameObject
             }
         case dmGameObject::PROPERTY_RESULT_RESOURCE_NOT_FOUND:
             {
-                if (has_key)
+                if (key_requsted)
```
JCash: typo: key_requsted

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
JCash: the "m_World" nor "m_ComponentId" doesn't seem to be used?
Would it be enough to pass "component" as the user data?
That would avoid the dynamic allocation, and removed some extra code.

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
```diff
@@ -727,6 +847,206 @@ namespace dmGameSystem
...
 
+    static bool ResolvePropertyEmitter(dmGameObject::HPropertyOptions options, dmhash_t* emitter_id)
+    {
+        uint32_t property_count = dmGameObject::GetPropertyOptionsCount(options);
+        for (int i = 0; i < property_count; ++i)
+        {
+            dmhash_t hash = 0;
+            dmGameObject::GetPropertyOptionsKey(options, i, &hash);
+
+            if (hash == HASH_EMITTER_ID)
```
JCash: This feels a bit strange to me.
We designed the input options to be an ordered list.
And, I'd also assume these would be in order of "importance", as we do most things when minimizing scope.
I.e. to get an emitter property, I'd expect the emitter first, on index 0.

What's the reasoning here? (maybe I forgot something from our discussions?)

engine/particle/src/particle.cpp
```diff
@@ -1972,7 +2014,15 @@ namespace dmParticle
         dmVMath::Matrix4 world = dmTransform::ToMatrix4(transform);
         dmParticle::EmitterPrototype* emitter_proto = &instance->m_Prototype->m_Emitters[emitter_index];
 
-        render_emitter_callback(user_context, emitter_proto->m_Material, emitter->m_AnimationData.m_Texture, world, emitter_proto->m_BlendMode, emitter->m_VertexIndex, emitter->m_VertexCount, emitter->m_RenderConstants.Begin(), emitter->m_RenderConstants.Size());
+        render_emitter_callback(user_context,
+            emitter_proto->m_Material,
```
JCash, line 2018: don't we want to pass in the potentially overridden material here?

JCash (changes requested):
Overall good.
Some question regarding changing user data allocaton.
Some question about emitter key index.

engine/gamesys/src/gamesys/components/comp_particlefx.cpp
Jhonnyg: Nice spotted! Yeah, not sure what i was thinking. Maybe it was used for something else earlier.

engine/particle/src/particle.cpp
Jhonnyg: This is only used in the editor, which doesn't have overrides.

AGulev:
isit also fixes https://github.com/defold/defold/issues/11051 ?

PR #11811, opened 2026-01-29
Backspace key (⌫) now performs a Delete action on macOS

vlaaad:
Pressing Backspace (⌫) now performs a Delete action on macOS.

Technical notes:

I verified it does not interfere with the code editor.

Fix #11000


matgis (approved):
Ship it!

PR #11821, opened 2026-01-31
Filter and search implementation for Outline in editor

kindablissy:
It's only a draft and I am very new to clojure and to the codebase So I don't know what side effects this might cause. 

closes #4098 #3926 #3925


vlaaad:
Hi @kindablissy, thank you for considering a contribution to the editor! It's fine that you are new to the code base. For features that are not immediately obvious, we need a design document that thoroughly considers the design space before we consider a PR. (see the relevant section in the [contribution guide](https://github.com/defold/defold/blob/dev/CONTRIBUTING.md#new-features)).

When it comes to search/filter in the Outline, here are the first things that come to mind that need evaluation and discussion first:
1. Filter vs Select: Should the feature work like a search, where you jump between matches (e.g., using up/down), or as a filter hiding irrelevant subtrees?
2. What to filter on: only labels of the tree view items, or on properties too? See the comment here https://github.com/defold/defold/issues/3925#issuecomment-2472893425 that requests filtering by property, presumably using syntax like `@prop:value`.

So, for features, you need to create a document with the following high-level structure:
1. Problem Statement.
2. Related information (e.g., how other IDEs approach this).
3. Alternative solutions: at least 2, every solution must list both pros and cons.
4. The suggested approach.

As an example, [here is a design doc for the Keymap configuration](https://docs.google.com/document/d/1gva_ksRDkb9UhxY0R5bfiYyyRtQxE2U7s41zkpYqdGE/edit?tab=t.0#heading=h.rk7i6ldgxut8).

Then, we'll discuss the document, add comments, and decide which approach we are comfortable with implementing/accepting a PR for/maintaining. Then it's time to create a PR.

AGulev:
It seems like this PR is outdated, and the requested changes will not be made.
Should we close it, @vlaaad?

PR #11824, opened 2026-02-01
Fix physics.set_event_listener documentation example

njg7194:
## Summary
This PR fixes the documentation example for `physics.set_event_listener` as requested in #11265.

## Changes
Fixed several issues in the example code:

1. **Lua syntax error**: Changed `for _,event in ipairs(events):` to `for _, event in ipairs(events) do`
2. **Inconsistent variable usage**: All event type comparisons now consistently use `event_type` instead of mixing `event` and `event_type`
3. **Missing loop end**: Added the missing `end` statement for the `for` loop

## Before
```lua
for _,event in ipairs(events):
    ...
    elseif event == hash("collision_event") then  -- inconsistent
    ...
end  -- missing for loop end
```

## After
```lua
for _, event in ipairs(events) do
    ...
    elseif event_type == hash("collision_event") then  -- consistent
    ...
  end  -- for loop end
end
```

Fixes #11265

### DIFF
--- a/engine/gamesys/src/gamesys/scripts/script_physics.cpp
+++ b/engine/gamesys/src/gamesys/scripts/script_physics.cpp
@@ -1594,7 +1594,7 @@ namespace dmGameSystem
      *     --    relative_velocity = vmath.vector3(0, 0, 0),
      *     --  },
      *     -- }
-     *   elseif event == hash("collision_event") then
+     *   elseif event_type == hash("collision_event") then
      *     pprint(data)
      *     -- {
      *     --  a = {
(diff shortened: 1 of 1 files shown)


engine/gamesys/src/gamesys/scripts/script_physics.cpp
```diff
@@ -1594,7 +1594,7 @@ namespace dmGameSystem
      *     --    relative_velocity = vmath.vector3(0, 0, 0),
      *     --  },
      *     -- }
-     *   elseif event == hash("collision_event") then
+     *   elseif event_type == hash("collision_event") then
```
britzl, line 1597: This is the old deprecated function. There is no event_type in this example.

britzl (changes requested):
You made incorrect changes to the old deprecated function too

AGulev:
it fixed already, but thank you for your effort

PR #11834, opened 2026-02-02
Fix issues where DF fonts appear blurry in some cases.

### LINKED ISSUE #9560: Sdf font is blurry on some rotation.
I use sdf gui node in 3D world. On some angles
euler.y == 90 and euler.y == 270 sdf font is blurry.

https://forum.defold.com/t/billboard-shader-model-matrix/12983/35

### Expected behaviour

Sdf font render not blurry

### Actual behaviour

Sdf font render not blurry

### Steps to reproduce
1)Create an SDF GUI node in a 3D scene.
2)Rotate the node such that euler.y is 90 or 270 degrees.
3)Set camera to see that node
4)Observe the rendered text becoming blurry.

This is also repro case from totebo

[Billboard.zip](https://github.com/user-attachments/files/17270603/Billboard.zip)

https://github.com/user-attachments/assets/fc0fbe69-a562-46d8-9bfd-3b7181ad7e25

I believe the issue might be related to how the scale is extracted from the transformation matrix in the SDF font rendering code. Specifically, this part of the code may be problematic:

https://github.com/defold/defold/blob/8a9df4c2f84c0398829cfca03a62d4837aa50c43/engine/render/src/render/font_renderer.cpp#L849C5-L859C81

```c++
const Vector4 r0 = te.m_Transform.getRow(0);
const float sdf_edge_value = 0.75f;
float sdf_world_scale = sqrtf(r0.getX() * r0.getX() + r0.getY() * r0.getY());
```

It seems that both the scale and rotation are affecting the value extracted from the transformation, which could be causing the blurriness in certain rotations.

<hr/>

<table>
<tr><td>Build time</td><td>2024-10-04T11:53:41.364164</td></tr>
<tr><td>Defold editor sha</td><td>ef2ce600e47a7840cca1342ab3653b25a8b0f5fe</td></tr>
<tr><td>Defold engine sha</td><td>e4aaff11f49c941fde1dd93883cf69c6b8abebe4</td></tr>
<tr><td>Defold version</td><td>1.9.3</td></tr>
<tr><td>GPU</td><td>Intel(R) RaptorLake-S Mobile Graphics Controller</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 - Build 31.0.101.5186</td></tr>
<table>

### LINKED ISSUE #9568: Bluring fonts on Go.Labels when Scale and Zoom
When I add a Label in a Collection with other(s) objects, then zoom :
`go.set("#camera", "orthographic_zoom", 1*(math.pow(zoom_gap, zoom_state)))`
and apply a scale on the Label (and only it) of the exact same value, but opposite (if I zoom the camera X2, I scale the Label 0.5)
`Camera zoom : 1*(math.pow(zoom_gap, zoom_state))'
then 
'go.set(tmp_Label, "scale.y", 1*(math.pow(zoom_gap, -zoom_state)))
go.set(tmp_Label, "scale.x", 1*(math.pow(zoom_gap, -zoom_state)))`

Thread in Defold board : 
[https://forum.defold.com/t/math-question-about-fonts-and-orthographic-zoom/78331/2](url)

**To Reproduce (REQUIRED)**
Do as the example, or see my project on github : 
[https://github.com/Vampiloup/syzygie?tab=readme-ov-file](url)

The Label should not blur.
I also tried by making a PNG of a word with transparent background, and using THAT PNG as sprite in place of the Label, and it not blured.
Aren't Font also PNG ? Shouldn't they react in the same manner ?

  - Version 1.9.3 - Latest updae

**Platforms (REQUIRED):**
 - Platforms: Linux
 - OS: Manjaro
 - Device: Desktop

[https://github.com/Vampiloup/syzygie?tab=readme-ov-file](url)

**Workaround (OPTIONAL):**
Trying to use Gui label node, but I never can make it appear at the same position than the Object. Even when the two elements have the same get.position result (yes, after a pass in worldtoscreen for the GO).

I need to make appear some words on the same position than objects made by factories. With a camera moving in a world bigger than the screen, and zooming in and out.
That's for a game like Master of Orion (llok at the third image : 
[https://androidarts.com/spaceconquest/gamedesign.htm](url)

AGulev:
DF smoothing for distance-field fonts is now computed in screen space rather than world space. The renderer derives a pixels-per-local-unit scale from the current view-projection matrix and viewport, ensuring stable edge thickness under camera zoom and perspective.

b_a_a

Fix https://github.com/defold/defold/issues/9568
Fix https://github.com/defold/defold/issues/11779
Fix https://github.com/defold/defold/issues/9560

### DIFF
--- a/engine/render/src/render/font/default/font_default_vertex.cpp
+++ b/engine/render/src/render/font/default/font_default_vertex.cpp
@@ -324,7 +324,7 @@ void GetTextMetrics(HFontRenderBackend backend, HFontMap font_map, const char* t
 }
 
 
-uint32_t CreateFontVertexData(HFontRenderBackend backend, HFontMap font_map, uint32_t frame, const char* text, const TextEntry& te, float recip_w, float recip_h, uint8_t* _vertices, uint32_t num_vertices)
+uint32_t CreateFontVertexData(HFontRenderBackend backend, HFontMap font_map, uint32_t frame, const char* text, const TextEntry& te, float sdf_screen_scale, float recip_w, float recip_h, uint8_t* _vertices, uint32_t num_vertices)
 {
     DM_PROFILE(__FUNCTION__);
     (void)backend;
@@ -339,17 +339,22 @@ uint32_t CreateFontVertexData(HFontRenderBackend backend, HFontMap font_map, uin
     const Vector4 outline_color = dmGraphics::UnpackRGBA(te.m_OutlineColor);
     const Vector4 shadow_color  = dmGraphics::UnpackRGBA(te.m_ShadowColor);
 
-    // No support for non-uniform scale with SDF so just peek at the first
-    // row to extract scale factor. The purpose of this scaling is to have
-    // world space distances in the computation, for good 'anti aliasing' no matter
-    // what scale is being rendered in.
-    const Vector4 r0 = te.m_Transform.getRow(0);
     const float sdf_edge_value = 0.75f;
-    float sdf_world_scale = sqrtf(r0.getX() * r0.getX() + r0.getY() * r0.getY());
+    float sdf_scale = sdf_screen_scale;
+    if (sdf_scale <= 0.0f)
+    {
+        // Fallback to local scale when screen scale is invalid.
+        const Vector4 r0 = te.m_Transform.getRow(0);
+        sdf_scale = sqrtf(r0.getX() * r0.getX() + r0.getY() * r0.getY());
+    }
+    if (sdf_scale < 1e-6f)
+    {
+        sdf_scale = 1e-6f;
+    }
     float sdf_outline = font_map->m_SdfOutline;
     float sdf_shadow  = font_map->m_SdfShadow;
     // For anti-aliasing, 0.25 represents the single-axis radius of half a pixel.
-    float sdf_smoothing = 0.25f / (font_map->m_SdfSpread * sdf_world_scale);
+    float sdf_smoothing = 0.25f / (font_map->m_SdfSpread * sdf_scale);
 
     HFontCollection font_collection = font_map->m_FontCollection;
 
--- a/engine/render/src/render/font/font_renderer.cpp
+++ b/engine/render/src/render/font/font_renderer.cpp
@@ -20,6 +20,7 @@
 #include <dlib/hashtable.h>         // for dmHashTable32
 #include <dlib/log.h>               // for dmLogError, dmLogOnceError, dmLog...
 #include <dlib/profile.h>           // for DM_PROFILE, DM_PROPERTY_*
+#include <dlib/math.h>              // for dmMath::Max
 #include <dlib/zlib.h>              // for InflateBuffer
 #include <dmsdk/dlib/intersection.h>// for dmIntersection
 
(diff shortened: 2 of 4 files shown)


engine/render/src/render/font/default/font_default_vertex.cpp
```diff
@@ -339,17 +339,22 @@ uint32_t CreateFontVertexData(HFontRenderBackend backend, HFontMap font_map, uin
     const Vector4 outline_color = dmGraphics::UnpackRGBA(te.m_OutlineColor);
     const Vector4 shadow_color  = dmGraphics::UnpackRGBA(te.m_ShadowColor);
 
-    // No support for non-uniform scale with SDF so just peek at the first
-    // row to extract scale factor. The purpose of this scaling is to have
-    // world space distances in the computation, for good 'anti aliasing' no matter
-    // what scale is being rendered in.
```
AGulev, line 345: World space scale doesn't fit for labels, also it may be not perfect when render gui with non standart projections. So it's better if `sdf_screen_scale` used instead of `sdf_world_scale`

engine/render/src/render/font/default/font_default_vertex.cpp
```diff
@@ -339,17 +339,22 @@ uint32_t CreateFontVertexData(HFontRenderBackend backend, HFontMap font_map, uin
...
-    const Vector4 r0 = te.m_Transform.getRow(0);
     const float sdf_edge_value = 0.75f;
-    float sdf_world_scale = sqrtf(r0.getX() * r0.getX() + r0.getY() * r0.getY());
+    float sdf_scale = sdf_screen_scale;
+    if (sdf_scale <= 0.0f)
+    {
+        // Fallback to local scale when screen scale is invalid.
+        const Vector4 r0 = te.m_Transform.getRow(0);
+        sdf_scale = sqrtf(r0.getX() * r0.getX() + r0.getY() * r0.getY());
```
AGulev, line 348: in some cases view may be reported as 0 0 0 0, which might result in a crash
so, I decided that it's better to have such a fallback here, but maybe I'm over defensive?

engine/render/src/render/font/default/font_default_vertex.cpp
```diff
@@ -339,17 +339,22 @@ uint32_t CreateFontVertexData(HFontRenderBackend backend, HFontMap font_map, uin
...
+    }
+    if (sdf_scale < 1e-6f)
+    {
+        sdf_scale = 1e-6f;
+    }
```
AGulev, line 353: avoid situations when sdf_scale is too small.

engine/render/src/render/font/font_renderer.cpp
```diff
@@ -133,6 +134,69 @@ namespace dmRender
 
     static dmhash_t g_TextureSizeRecipHash = dmHashString64("texture_size_recip");
 
+    static float CalcSdfScreenScale(HRenderContext render_context, const Matrix4& world_transform)
```
AGulev, line 137: This function calculated pixels per local unit for a text object and then screen spaced scale based on it and window size

engine/render/src/render/font/font_renderer.cpp
```diff
@@ -133,6 +134,69 @@ namespace dmRender
...
+        Vector4 col3 = m.getCol(3);
+
+        Vector4 c0 = col3;
+        Vector4 cx = col0 + col3;
+        Vector4 cy = col1 + col3;
+
+        const float kMinW = 1e-6f;
+        if (fabsf(c0.getW()) < kMinW || fabsf(cx.getW()) < kMinW || fabsf(cy.getW()) < kMinW)
+        {
+            return 0.0f;
```
AGulev, line 154: If clip w is near 0, the perspective divide would explode.

engine/render/src/render/font/font_renderer.cpp
```diff
@@ -133,6 +134,69 @@ namespace dmRender
...
+        float half_h = 0.5f * (float)vp_h;
+
+        float dx = ndc_dx_x * half_w;
+        float dy = ndc_dx_y * half_h;
+        float len_sq_x = dx * dx + dy * dy;
+        dx = ndc_dy_x * half_w;
+        dy = ndc_dy_y * half_h;
+        float len_sq_y = dx * dx + dy * dy;
+
+        float scale = sqrtf(dmMath::Max(len_sq_x, len_sq_y));
```
AGulev, line 192: it's basically a uniform scale where max is taken

engine/render/src/render/font/font_renderer.cpp
```diff
@@ -133,6 +134,69 @@ namespace dmRender
...
+        float ndc_dx_y = cx.getY() * inv_wx - ndc0_y;
+        float ndc_dy_x = cy.getX() * inv_wy - ndc0_x;
+        float ndc_dy_y = cy.getY() * inv_wy - ndc0_y;
+
+        dmGraphics::HContext gc = GetGraphicsContext(render_context);
+        int32_t vp_x = 0;
+        int32_t vp_y = 0;
+        uint32_t vp_w = 0;
+        uint32_t vp_h = 0;
+        dmGraphics::GetViewport(gc, &vp_x, &vp_y, &vp_w, &vp_h);
```
AGulev, line 173: get actual pixels and use it to calculate screen space scale (using normalized coordinates calculated from view projection and world transform)

JCash (approved):

PR #11836, opened 2026-02-02
Fix blurry distance field font in HTML5 when `High DPI` is off

### LINKED ISSUE #8998: Blurry distance field font in html5
A user reported that they distance field font is crisp and easy to read when building for desktop, but as soon as the project is built or bundled for html5 the font gets very blurry:

HTML5:
Screenshot 2024-05-29 at 11 28 15

Desktop (macOS):
Screenshot 2024-05-29 at 11 28 06

Attached sample project.

[Uploading font_issue.zip…]()

AGulev:
Distance-field font smoothing is now clamped at runtime to prevent text from becoming overly soft or gray when it becomes too small in pixel size. This typically happens when High DPI is turned off, the camera is heavily zoomed out, or the font scale is reduced too much.

CleanShot 2026-02-02 at 23 01 28@2x

Fix https://github.com/defold/defold/issues/8998

### DIFF
--- a/engine/render/src/render/font/default/font_default_vertex.cpp
+++ b/engine/render/src/render/font/default/font_default_vertex.cpp
@@ -340,16 +340,22 @@ uint32_t CreateFontVertexData(HFontRenderBackend backend, HFontMap font_map, uin
     const Vector4 shadow_color  = dmGraphics::UnpackRGBA(te.m_ShadowColor);
 
     const float sdf_edge_value = 0.75f;
+    const float kMinSdfScreenScale = 0.5f;
+    const float kMinSdfScale = 1e-6f;
     float sdf_scale = sdf_screen_scale;
-    if (sdf_scale <= 0.0f)
+    if (sdf_scale > 0.0f)
+    {
+        sdf_scale = fmaxf(sdf_scale, kMinSdfScreenScale);
+    }
+    else
     {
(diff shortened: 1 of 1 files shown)


engine/render/src/render/font/default/font_default_vertex.cpp
```diff
@@ -340,16 +340,22 @@ uint32_t CreateFontVertexData(HFontRenderBackend backend, HFontMap font_map, uin
     const Vector4 shadow_color  = dmGraphics::UnpackRGBA(te.m_ShadowColor);
 
     const float sdf_edge_value = 0.75f;
+    const float kMinSdfScreenScale = 0.5f;
```
AGulev: I tried 1.0, but it was too sharp and pixilated, so after experiments I decided to keep 0.5

JCash (dismissed):

engine/render/src/render/font/default/font_default_vertex.cpp
JCash: we don't normally use "k" as prefix. nor do we use camel case. Looks like the agent didn't read the code style.
We just use "min_sdk_screen_scale"

AGulev: Fixed

PR #11849, opened 2026-02-04
Light component refactor

### LINKED ISSUE #11724: Refactor light component to use new struct type
**Task (REQUIRED):**
Design: https://docs.google.com/document/d/1aTUFwBqc3m3VbYkmsV9jHO4-CPToRfRG3VzYF78tPu0/edit?usp=sharing

Jhonnyg:
The light component has been refactored to use the new `dmStructDDF.Struct` data type and a new dmRender::HLightPrototype / dmRender::HLightInstance API that can be used to marshall data from the light component into shaders. Keep in mind that this is a work-in-progress, which means that we could potentially restructure this feature until it has been properly released.

Fixes #11724


engine/render/src/render/material.cpp
```diff
@@ -25,10 +25,16 @@
 #include "render.h"
 #include "render_private.h"
 
+// TODO: Should we move the internals into an api or move to graphics.h instead?
+#include "../../../graphics/src/graphics_private.h"
```
Jhonnyg, line 29: I currently need to dig out binding information about where the uniform buffer is bound (via ProgramResourceBindingIterator). I guess we _could_ replace the iterator down below with some API calls in graphics.h, with some new structs and functions perhaps.

engine/render/src/render/render.h
```diff
@@ -118,6 +120,22 @@ namespace dmRender
...
+    enum LightType
+    {
+        LIGHT_TYPE_DIRECTIONAL = 0,
+        LIGHT_TYPE_POINT       = 1,
+        LIGHT_TYPE_SPOT        = 2,
+    };
+
+    // NOTE: These enum values are duplicated in gamesys camera DDF (camera_ddf.proto)
+    // Don't forget to change dmGamesysDDF::OrthoZoomMode if you change here
+    enum OrthoZoomMode
```
Jhonnyg, line 132: I just moved to live with the other enums

engine/render/src/render/render.h
```diff
@@ -118,6 +120,22 @@ namespace dmRender
         SORT_NONE          = 3
     };
 
+    enum LightType
```
Jhonnyg, line 123: In the future we can have some sort of "user" type here that we can derive from the dynamic protobuf. But for this is enough to move on with the other parts of light support.

engine/gamesys/src/gamesys/resources/res_light.cpp
```diff
@@ -12,45 +12,169 @@
...
+    }
+
+    static void GetVector3(const dmStructDDF::Struct* s, const char* key, dmVMath::Vector3* out)
+    {
+        dmVMath::Vector4 v4;
+        GetVector4(s, key, &v4);
+        *out = dmVMath::Vector3(v4.getX(), v4.getY(), v4.getZ());
+    }
+
+    static void DDFToLightParams(const dmGameSystemDDF::LightDesc* light_desc, dmRender::LightPrototypeParams& params)
```
Jhonnyg, line 102: This relies on a fixed layout currently, but since we now use the "struct" data type we can easily extend this to pick up generic vector4 as properties that can be mapped to shaders, or something like that.

engine/graphics/src/graphics.h
```diff
@@ -249,6 +249,33 @@ namespace dmGraphics
...
+        uint32_t                    m_Offset;
+    };
+
+    struct ShaderResourceTypeInfo
+    {
+        char*                 m_Name;
+        dmhash_t              m_NameHash;
+        ShaderResourceMember* m_Members;
+        uint32_t              m_MemberCount;
+    };
```
Jhonnyg, line 277: These are now used by dmRender to construct a UniformBufferLayout for the light buffer. I am a bit on the fence if we should have these "public" to the rest of the engine or not..

engine/engine/src/engine.cpp
```diff
@@ -1484,9 +1489,10 @@ namespace dmEngine
         if (fact_result != dmResource::RESULT_OK)
             goto bail;
 
-        go_result = dmGameSystem::RegisterComponentTypes(engine->m_Factory, engine->m_Register, engine->m_RenderContext, physics_context, &engine->m_ParticleFXContext, &engine->m_SpriteContext,
-                                                                                                &engine->m_CollectionProxyContext, &engine->m_FactoryContext, &engine->m_CollectionFactoryContext,
-                                                                                                &engine->m_ModelContext, &engine->m_LabelContext, &engine->m_TilemapContext);
+        go_result = dmGameSystem::RegisterComponentTypes(engine->m_Factory, engine->m_Register,
+            engine->m_RenderContext, physics_context, &engine->m_ParticleFXContext, &engine->m_SpriteContext,
+            &engine->m_CollectionProxyContext, &engine->m_FactoryContext, &engine->m_CollectionFactoryContext,
+            &engine->m_ModelContext, &engine->m_LabelContext, &engine->m_TilemapContext, &engine->m_LightContext);
```
JCash: As discussed on the offsite, we want to make our scripts/components more modular.
And I think we should do it now with the light component, rather than postponing it to a future task.

See how comp_mesh.cpp registers its own type, and creates its own context and so on.
This will affect the link steps (have to register exported symbols to link with), and also build_input.yml.

But the good news it that then it's truly an extension, and the engine/gamesystem doesn't have to know about it.

engine/gamesys/src/gamesys/components/comp_light.cpp
```diff
@@ -29,50 +31,61 @@ namespace dmGameSystem
...
-        {
-            m_Instance = instance;
-            m_LightResource = light_resource;
-            m_AddedToUpdate = 1;
-        }
+        dmGameObject::HInstance  m_Instance;
+        LightResource*           m_LightResource;
+        dmRender::HLightInstance m_LightInstance;
+        uint16_t                 m_AddedToUpdate : 1;
+        uint16_t                 m_Padding : 15;
```
JCash: Probably less confusing here, but if it was a name clash with "padding", (like in fonts), one can also write it anonymously:
```c++
uint16_t m_AddedToUpdate : 1;
uint16_t : 15;
```

engine/gamesys/src/gamesys/resources/res_light.h
```diff
@@ -17,12 +17,20 @@
...
+#include <render/render.h>
+
 namespace dmGameSystem
 {
-    dmResource::Result ResLightCreate(const dmResource::ResourceCreateParams* params);
+    struct LightResource
+    {
+        dmRender::HLightPrototype   m_LightPrototype;
+        dmGameSystemDDF::LightDesc* m_DDF;
+    };
```
JCash, line 28: I know that the a lot of the resource types are like this as well, but I think we should continue anonymizing the structs.

I.e. make this "LightResource*" to be anonymous.
Instead add an acessor function `dmRender::HLightPrototype ResLightGetPrototype(LightResource*)` and so on.

Create a res_light.h in gamesys/src/dmsdk/resoruces/res_light.h where the public part should go.

engine/gamesys/src/gamesys/resources/res_light.cpp
```diff
@@ -12,45 +12,169 @@
...
+                    : 1.0f;
+
+        *out = dmVMath::Vector4(r, g, b, a);
+    }
+
+    static void GetVector3(const dmStructDDF::Struct* s, const char* key, dmVMath::Vector3* out)
+    {
+        dmVMath::Vector4 v4;
+        GetVector4(s, key, &v4);
+        *out = dmVMath::Vector3(v4.getX(), v4.getY(), v4.getZ());
```
JCash:
```c++
*out = v4.getXYZ();
```

engine/gamesys/src/gamesys/resources/res_light.cpp
```diff
@@ -12,45 +12,169 @@
...
+        const dmStructDDF::Struct::FieldsEntry* type_field = FindField(data, "type");
+        if (type_field)
+        {
+            const char* type_str = type_field->m_Value->m_Kind.m_StringValue;
+            if (type_str)
+            {
+                if (strcmp(type_str, "directional") == 0)
+                {
+                    params.m_Type = dmRender::LIGHT_TYPE_DIRECTIONAL;
+                    GetVector3(data, "direction", &params.m_Direction);
```
JCash: What do we want for error handling, or reporting things for the user?
E.g. if there is a "directional" light, but no "direction", or if the type is wrong.

engine/gamesys/src/gamesys/resources/res_light.cpp
```diff
@@ -12,45 +12,169 @@
...
+        memset(light_resource, 0, sizeof(LightResource));
+
+        dmRender::HRenderContext render_context = (dmRender::HRenderContext) params->m_Context;
+
+        light_resource->m_DDF = (dmGameSystemDDF::LightDesc*) params->m_PreloadData;
+
+        dmRender::LightPrototypeParams light_params;
+        DDFToLightParams(light_resource->m_DDF, light_params);
+
+        light_resource->m_LightPrototype = dmRender::NewLightPrototype(render_context, light_params);
```
JCash, line 166: Beyond this point, is the ddf really needed?

engine/gamesys/src/gamesys/resources/res_light.cpp
```diff
@@ -12,45 +12,169 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
-#include "res_light.h"
+#include <dmsdk/dlib/vmath.h>
+
+#include <gamesys/gamesys.h>
 
-#include <gamesys/gamesys_ddf.h>
+#include "res_light.h"
```
JCash, line 19: We should move the res_light to an extension as well.
See e.g. res_gameobject.cpp.

engine/gamesys/src/gamesys/resources/res_light.cpp
```diff
@@ -12,45 +12,169 @@
...
             return dmResource::RESULT_FORMAT_ERROR;
         }
-        dmGameSystemDDF::LightDesc** light_resource = (dmGameSystemDDF::LightDesc**) dmResource::GetResource(params->m_Resource);
-        dmDDF::FreeMessage(*light_resource);
-        *light_resource = light_desc;
+
+        LightResource* light_resource = (LightResource*) dmResource::GetResource(params->m_Resource);
+        dmRender::HRenderContext render_context = (dmRender::HRenderContext) params->m_Context;
+        ReleaseResources(params->m_Factory, render_context, light_resource);
+        *light_resource = tmp_light_resource;
```
JCash: At this point the old light probe is released, and no new one is created?

engine/graphics/src/graphics.h
JCash: As long as it's still internal in the engine, it's still possible to change later to something else.

engine/render/src/render/light.cpp
```diff
@@ -0,0 +1,338 @@
...
+        light_instance->m_Direction        = light_prototype->m_Direction;
+        light_instance->m_LightPrototype   = light_prototype;
+        light_instance->m_LightBufferIndex = render_context->m_RenderLightsIndices.Pop();
+
+        uint32_t scratch_size = light_instance->m_LightBufferIndex+1;
+        if (scratch_size > render_context->m_LightBufferScratch.Capacity())
+        {
+            render_context->m_LightBufferScratch.SetCapacity(scratch_size);
+            render_context->m_LightBufferScratch.SetSize(scratch_size);
+        }
```
JCash: I think it would fit better to do this check/reallocation after line 74, where we already know the new max size?

engine/render/src/render/light.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+    ////////////////////////////////
+    // Light instance
+    ////////////////////////////////
+
+    HLightInstance NewLightInstance(HRenderContext render_context, HLightPrototype light_prototype)
+    {
+        if (render_context->m_RenderLightsIndices.Remaining() == 0)
+        {
+            render_context->m_RenderLightsIndices.SetCapacity(render_context->m_RenderLightsIndices.Capacity() + 4);
```
JCash: If someone started to use lights, 4 is a quite small number.
Perhaps better with 32 or 64 ?

engine/render/src/render/render.h
```diff
@@ -195,6 +204,19 @@ namespace dmRender
...
+        LightPrototypeParams();
+
+        LightType        m_Type;
+        dmVMath::Vector4 m_Color;
+        dmVMath::Vector3 m_Direction;
+        float            m_Intensity;
+        float            m_Range;
+        float            m_InnerConeAngle;
+        float            m_OuterConeAngle;
+    };
```
JCash, line 218: Following your comment above, I guess we'll replace this structure with something more generic in our MVP2, in order to support any type of light data?
Essentially a ddf structure, but more densely packed.

engine/render/src/render/render.h
JCash: I assume we won't have this enum at all in MVP2, as it should all be data driven?

JCash (changes requested):
Overall good, but I'd like to see the component/resource type moved to use the extension api's.

Also some other comments.

engine/gamesys/src/gamesys/components/comp_light.cpp
Jhonnyg: It was some old code, but it's fixed now!

engine/render/src/render/render.h
Jhonnyg: Yes, but I am not sure what that would look like yet.

PR #11851, opened 2026-02-04
Make DF fonts sharp in the editor

AGulev:
Improve DF font rendering in the Editor

CleanShot 2026-02-04 at 16 23 04@2x


editor/src/clj/editor/font.clj
```diff
@@ -413,26 +495,32 @@
...
+   (request-vertex-buffer gl request-id font-data text-entries nil))
+  ([^GL2 gl request-id font-data text-entries render-args]
+   (let [text-entries (add-sdf-screen-scale render-args font-data text-entries)
+         glyph-cache (scene-cache/request-object! ::glyph-caches (:texture font-data) gl
+                                                  (select-keys font-data [:font-map :texture]))
+         layer-count (count-layers-in-mask (:layer-mask (:font-map font-data)))]
+     (scene-cache/request-object! ::vb [request-id (:type font-data) (vertex-count text-entries layer-count)] gl
+                                  {:font-data font-data
+                                   :text-entries text-entries
+                                   :glyph-cache glyph-cache}))))
```
matgis, line 523: There is a potential optimization we probably want to try out here.

This map fed in as the `request-data` argument to `scene-cache/request-object!` will be structurally compared against the `request-data` argument from the previous call for the same `request-id` supplied to `request-object!`. If it equals the previous value, we skip updating the vertex buffer in VRAM.

If we could introduce another `scene-object-cache` using `scene-cache/register-object-cache!`, similar to what we have for `::glyph-caches`, for `::text-entries`, we can get the same `:text-entries` object back every time we request them, and avoid re-generating both the `:text-entries` and the vertex buffer when the camera hasn't moved. This hinges on moving the call to `add-sdf-screen-scale` into the `update-fn` we register for the new `::text-entries` `scene-object-cache`.

Unfortunately until we make this change, we will be continuously calling the `add-sdf-screen-scale` function every frame. It would be best to not include the entire `render-args` in the `request-data` argument in our call to `scene-cache/request-object!`, but only the parts we need from it alongside the non-updated `text-entries`. This ensures we only re-calculate and update the SDF screen scale if any of the relevant inputs change.

editor/src/clj/editor/font.clj
```diff
@@ -100,10 +101,89 @@
...
+  [^Matrix4d view-proj ^Region viewport ^Matrix4d world-transform]
+  (if (or (nil? view-proj) (nil? viewport) (nil? world-transform))
+    0.0
+    (let [m (doto (Matrix4d. view-proj) (.mul world-transform))
+          col0 (Vector4d.)
+          col1 (Vector4d.)
+          col3 (Vector4d.)]
+      (.getColumn m 0 col0)
+      (.getColumn m 1 col1)
+      (.getColumn m 3 col3)
```
matgis: Just double-checking: Is it correct that we read from columns 0, 1, and 3 here, but not 2?

editor/src/clj/editor/font.clj
```diff
@@ -100,10 +101,89 @@
...
+              (let [half-w (* 0.5 vp-w)
+                    half-h (* 0.5 vp-h)
+                    dx (* ndc-dx-x half-w)
+                    dy (* ndc-dx-y half-h)
+                    len-sq-x (+ (* dx dx) (* dy dy))
+                    dx2 (* ndc-dy-x half-w)
+                    dy2 (* ndc-dy-y half-h)
+                    len-sq-y (+ (* dx2 dx2) (* dy2 dy2))
+                    scale (Math/sqrt (Math/max len-sq-x len-sq-y))]
+                (if (<= scale 0.0) 0.0 scale)))))))))
```
matgis: There's quite a lot of primitive math going on in this file now. We should probably add `(set! *unchecked-math* :warn-on-boxed)` below the existing `(set! *warn-on-reflection* true)` compiler directive we have at the top of the file to catch any boxed math that might have slipped in. It can cause significant slowdown. Unfortunately you might expose some existing boxed math warnings in other places in the file if you do this as well. 😬 We really should fix all of them, but you can turn it on temporarily and just make sure the code you've added is free of boxed math if you want.

editor/src/clj/editor/font.clj
```diff
@@ -100,10 +101,89 @@
...
+  [^Matrix4d transform]
+  (let [row0 (Vector4d.)]
+    (.getRow transform 0 row0)
+    (Math/sqrt (+ (* (.x row0) (.x row0))
+                  (* (.y row0) (.y row0))))))
+
+(defn- calc-sdf-screen-scale
+  "Return pixels per local unit for the given transform, or 0.0 if invalid."
+  [^Matrix4d view-proj ^Region viewport ^Matrix4d world-transform]
+  (if (or (nil? view-proj) (nil? viewport) (nil? world-transform))
```
matgis: We don't even call this if `view-proj` or `viewport` is `nil`, so we don't need this check. I'm pretty sure we're also guaranteed to have a `world-transform` as well, since I believe we add an identity transform to each renderable when we flatten the scene hierarchy. 🤔

editor/src/clj/editor/font.clj
```diff
@@ -392,9 +475,8 @@
 (defn- fill-vertex-buffer
   [^GL2 gl vbuf {:keys [type font-map texture] :as font-data} text-entries glyph-cache]
   (let [put-glyph-quad-fn (make-put-glyph-quad-fn font-map)
-        put-pos-uv-fn (cond-> put-pos-uv!
-                              (= type :distance-field)
-                              (wrap-with-sdf-params font-map))
+        distance-field? (= type :distance-field)
```
matgis: Style: We tend to use the `?`-suffix for predicates, i.e. functions that check something. For booleans, we typically name them something like `is-distance-field` instead.

matgis (commented):
It looks so much better! ❤️

I think the code changes look mostly alright, but I did find a potential optimization that I've tried to explain in the comments. Basically, we could maybe avoid re-calculating the SDF screen scale for all `text-entries` if neither the camera nor the text objects have moved since the last frame.

I want to try this PR out in a scene with a bunch of text, though. Maybe it doesn't slow things down that much in practical terms.

editor/src/clj/editor/font.clj
matgis: Hmm, on further examination there might not be that much to gain from doing this. I initially thought we calculated the SDF screen scale per glyph, but it appears to be per text field. Let's not put time into this yet.

editor/src/clj/editor/font.clj
AGulev: we don't need Z in these calculations of the screen scale, only X, Y and origin

editor/src/clj/editor/font.clj
AGulev: true, that makes sense (fixed)

editor/src/clj/editor/font.clj
AGulev: nice one! I didn't know about (should add it into my forum post I use as a cheat sheet)

editor/src/clj/editor/font.clj
```diff
@@ -413,26 +497,32 @@
                           (- 0 (* (:padding font-map) 0.5))
                           0)
                      :y 0}]
-    (when (and layer-mask-enabled shadow-enabled) (fill-vertex-buffer-quads vbuf text-entries put-pos-uv-fn line-height char->glyph glyph-cache put-glyph-quad-fn [0 0 1] shadow-offset alpha outline-alpha shadow-alpha))
-    (when (and layer-mask-enabled outline-enabled) (fill-vertex-buffer-quads vbuf text-entries put-pos-uv-fn line-height char->glyph glyph-cache put-glyph-quad-fn [0 1 0] font-offset alpha outline-alpha shadow-alpha))
-    (fill-vertex-buffer-quads vbuf text-entries put-pos-uv-fn line-height char->glyph glyph-cache put-glyph-quad-fn face-mask font-offset alpha outline-alpha shadow-alpha)))
+    (when (and layer-mask-enabled shadow-enabled) (fill-vertex-buffer-quads vbuf text-entries font-map  is-distance-field put-pos-uv-fn line-height char->glyph glyph-cache put-glyph-quad-fn [0 0 1] shadow-offset alpha outline-alpha shadow-alpha))
```
matgis: Style: Remove double-space.
```suggestion
    (when (and layer-mask-enabled shadow-enabled) (fill-vertex-buffer-quads vbuf text-entries font-map is-distance-field put-pos-uv-fn line-height char->glyph glyph-cache put-glyph-quad-fn [0 0 1] shadow-offset alpha outline-alpha shadow-alpha))
```

editor/src/clj/editor/font.clj
```diff
@@ -413,26 +497,32 @@
...
+    (when (and layer-mask-enabled outline-enabled) (fill-vertex-buffer-quads vbuf text-entries font-map  is-distance-field put-pos-uv-fn line-height char->glyph glyph-cache put-glyph-quad-fn [0 1 0] font-offset alpha outline-alpha shadow-alpha))
```
matgis: Style: Remove double-space.
```suggestion
    (when (and layer-mask-enabled outline-enabled) (fill-vertex-buffer-quads vbuf text-entries font-map is-distance-field put-pos-uv-fn line-height char->glyph glyph-cache put-glyph-quad-fn [0 1 0] font-offset alpha outline-alpha shadow-alpha))
```

editor/src/clj/editor/font.clj
```diff
@@ -413,26 +497,32 @@
...
+    (fill-vertex-buffer-quads vbuf text-entries font-map  is-distance-field put-pos-uv-fn line-height char->glyph glyph-cache put-glyph-quad-fn face-mask font-offset alpha outline-alpha shadow-alpha)))
```
matgis: Style: Remove double-space.
```suggestion
    (fill-vertex-buffer-quads vbuf text-entries font-map is-distance-field put-pos-uv-fn line-height char->glyph glyph-cache put-glyph-quad-fn face-mask font-offset alpha outline-alpha shadow-alpha)))
```

editor/src/clj/editor/font.clj
```diff
@@ -100,10 +101,91 @@
...
+              (if (<= scale 0.0) 0.0 scale))))))))
+
+(defn- add-sdf-screen-scale
+  "Annotate text entries with :sdf-screen-scale when rendering SDF fonts."
+  [render-args font-data text-entries]
+  (if (and (= :distance-field (:type font-data))
+           (some? render-args))
+    (let [camera (:camera render-args)
+          ^Matrix4d view-proj (or (when camera (camera/camera-view-proj-matrix camera))
+                                  (:view-proj render-args))
```
matgis: I don't see why we're querying the `camera` directly here?

As I understand it, you will always have a valid `:view-proj` matrix in the `render-args` derived from the `camera` and projection setting of the `RenderPass`.

matgis (dismissed):
Looks good! Just fix the introduced double-spaces before merging! :heart:

I did have a question about getting the View Projection Matrix from the `camera` directly instead of using the `:view-proj` from the `render-args`. If we don't need to do it, we should maybe just  use `(:view-proj render-args)`?

matgis (approved):
Ship it! ❤️

PR #11858, opened 2026-02-07
Fix or suppress Java-related warnings in the engine build process

AGulev:


engine/engine/src/test/wscript
```diff
@@ -32,6 +32,8 @@ def run_bob(task):
     output = run.env_command(env, args.split())
 
 def build(bld):
+    java_runtime_flags = os.environ.get('DM_JAVA_RUNTIME_FLAGS', '')
```
JCash, line 35: Shouldn't it be bld.env.JAVA_RUNTIME_FLAGS ? (seen from waf_dynamo.py where you set it up)

engine/engine/src/test/wscript
```diff
@@ -84,7 +86,7 @@ def build(bld):
 
     # Clean folder
     bld(rule = run_bob,
-        run_args = f'java -cp {classpath} com.dynamo.bob.Bob -root ./src/test clean',
+        run_args = f'java {java_runtime_flags} -cp {classpath} com.dynamo.bob.Bob -root ./src/test clean',
```
JCash, line 89: Haven't tried, but might be that waf does variable substitution on these as well (i.e. using the configured variables)
```python
run_args = f'{JAVA} {JAVA_RUNTIME_FLAGS} -cp {classpath} com.dynamo.bob.Bob -root ./src/test clean'
```

editor/bundle-resources/config
```diff
@@ -18,7 +18,7 @@ java = ${launcher.jdk}/bin/java
...
+vmargs = -Dfile.encoding=UTF-8,-Djna.nosys=true,-Ddefold.launcherpath=${bootstrap.launcherpath},-Ddefold.resourcespath=${bootstrap.resourcespath},-Ddefold.version=${build.version},-Ddefold.editor.sha1=${build.editor_sha1},-Ddefold.engine.sha1=${build.engine_sha1},-Ddefold.buildtime=${build.time},-Ddefold.channel=${build.channel},-Ddefold.archive.domain=${build.archive_domain},-Djava.net.preferIPv4Stack=true,-Dsun.net.client.defaultConnectTimeout=30000,-Dsun.net.client.defaultReadTimeout=30000,-Djogl.texture.notexrect=true,-Dglass.accessible.force=false,-Ddefold.schema.check.disable=true,--add-opens=java.base/java.lang=ALL-UNNAMED,--add-opens=java.desktop/sun.awt=ALL-UNNAMED,--add-opens=java.desktop/sun.java2d.opengl=ALL-UNNAMED,--add-opens=java.xml/com.sun.org.apache.xerces.internal.jaxp=ALL-UNNAMED,--add-opens=java.base/sun.nio.fs=ALL-UNNAMED,--add-opens=java.desktop/sun.awt.image=ALL-UNNAMED,--enable-native-access=ALL-UNNAMED,-XX:+UseCompactObjectHeaders,--sun-misc-unsafe-memory-access=allow
```
JCash, line 21: I'm less sure about this one, as the bundled editor is the released product.
Is the warnings coming from the tests or the actual editor runtime?

JCash (dismissed):
Engine part looks good, but I am less sure of the editor parts as it affects the released product.

editor/bundle-resources/config
vlaaad: I think it makes sense since these warnings only confuse our users if they try to read logs.

editor/project.clj
```diff
@@ -159,6 +159,8 @@
                       ;; used in editor.scene$read_to_buffered_image
                       "--add-opens=java.desktop/sun.awt.image=ALL-UNNAMED"
                       "--enable-native-access=ALL-UNNAMED"
+                      "-XX:+IgnoreUnrecognizedVMOptions"
```
vlaaad: Can you remove `"-XX:+IgnoreUnrecognizedVMOptions"`? I don't see a reason why we would want to silently ignore unrecognized options.

engine/dlib/src/java/com/dynamo/upnp/SSDP.java
```diff
@@ -75,14 +77,16 @@ public void disconnect() throws IOException {
...
+                mcastSocket.leaveGroup(new InetSocketAddress(SSDP_MCAST_ADDR, 0), networkInterface);
                 mcastSocket.close();
                 mcastSocket = null;
+                networkInterface = null;
             }
         }
 
         private MulticastSocket mcastSocket;
         private List<DatagramSocket> sockets;
+        private NetworkInterface networkInterface;
```
vlaaad, line 89: I know this isn't the editor code, but there is also a similar class in the editor code — https://github.com/defold/defold/blob/dev/editor/src/java/com/dynamo/upnp/SSDP.java, so I'm just curious what the reason for these changes is, as I don't see any warnings in the editor's SSDP class.

vlaaad (commented):
I think it's useful to disable the unsafe memory access warning for our users, but I really don't want to remove arg validation unless it is absolutely necessary.

engine/dlib/src/java/com/dynamo/upnp/SSDP.java
AGulev: I'm fixing this warnings here as well:
```
warning: [options] bootstrap class path is not set in conjunction with -source 8
  not setting the bootstrap class path may lead to class files that cannot run on JDK 8
    --release 8 is recommended instead of -source 8 -target 1.8 because it sets the bootstrap class path automatically
warning: [options] source value 8 is obsolete and will be removed in a future release
warning: [options] target value 8 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
Note: ../src/java/com/dynamo/upnp/SSDP.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
```

vlaaad (dismissed):
The editor changes look good!

PR #11869, opened 2026-02-09
Avoid unnecessary resize calls when the size hasn't changed.

### LINKED ISSUE #11844: `resize_window_callback` always triggers due to locally scoped variables in dmloader.js
The standard `resize_window_callback` function is executed every time a resize event occurs, even when the window size has not actually changed. This happens because the variables  
```js
var prevInnerWidth = -1;
var prevInnerHeight = -1;
```  
are declared locally inside the function in https://github.com/defold/defold/blob/dev/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js , causing them to reset on each callback execution.

**To Reproduce (REQUIRED)**  
1. Add logging to `resize_window_callback`
2. Run an HTML5 build using Defold.  
3. Open the application in a browser.  
4. Resize the browser window slightly, or even trigger the resize event without changing the actual size.  
5. Observe that `resize_window_callback` is fired every time.

The callback should only execute when the window size changes compared to the previous stored values, not on every resize event.

- Version: 1.12.0 and older

**Platforms (REQUIRED):**  
- Platforms: HTML5

AGulev:
This fixes a problem where the HTML5 `resize_window_callback` function is executed every time a resize event occurs, even when the window size has not actually changed. The callback is now only executed when the window size changes compared to the previous stored values, not on every resize event.

Fix https://github.com/defold/defold/issues/11844

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/resources/jsweb/dmloader.js
@@ -65,11 +65,11 @@ var CUSTOM_PARAMETERS = {
     },
     update_imports: function(imports) {
     },
+    resize_window_prev_inner_width: -1,
+    resize_window_prev_inner_height: -1,
     resize_window_callback: function() {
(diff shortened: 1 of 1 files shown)


ekharkunov (approved):
👍

PR #11871, opened 2026-02-09
Added CompModelPlayAnimation to dmSDK

britzl:
This change adds a `dmGameSystem::CompModelPlayAnimation()` API function to dmSDK. The function is also used from `script_model.cpp` instead of sending a message.


engine/graphics/src/null/graphics_null.cpp
```diff
@@ -101,6 +101,7 @@ namespace dmGraphics
         m_TextureFormatSupport |= 1 << TEXTURE_FORMAT_RGB_16BPP;
         m_TextureFormatSupport |= 1 << TEXTURE_FORMAT_RGBA_16BPP;
         m_TextureFormatSupport |= 1 << TEXTURE_FORMAT_RGB_ETC1;
+        m_TextureFormatSupport |= 1 << TEXTURE_FORMAT_RGBA32F;
```
britzl, line 104: Needed for the bone pose cache

engine/gamesys/src/gamesys/test/test_gamesys.cpp
```diff
@@ -5943,6 +5943,26 @@ TEST_F(ModelTest, GetAABB)
     ASSERT_TRUE(dmGameObject::Final(m_Collection));
 }
 
+TEST_F(ModelTest, PlayAnim)
```
britzl, line 5953: This test is for some reason failing on CI on macOS... It works locally on macOS. And the test works on WIndows and Linux on CI...

engine/gamesys/src/dmsdk/gamesys/components/comp_model.h
```diff
@@ -0,0 +1,55 @@
...
+    typedef struct ModelWorld* HModelWorld;
+    typedef struct ModelComponent* HModelComponent;
+
+    /*# 
+     * Play a model animation.
+     * @name CompModelPlayAnimation
+     * @param world [type: HModelWorld] Model world.
+     * @param component [type: HModelComponent] Model component.
+     * @param anim_id [type: dmhash_t] Animation to play.
+     * @param playback [type: uint32_t] Playback mode.
```
JCash: The playback mode alternatives should be an enum, or at least clearly stated what the accepted values are.

We should be able to use the correct enum for the playback mode as it's already part of the dmsdk/rig/rig.h.

engine/gamesys/src/gamesys/test/test_gamesys.cpp
britzl: The dt calculation in the test helper was not correct

engine/gamesys/src/gamesys/scripts/script_model.cpp
```diff
@@ -194,14 +196,24 @@ namespace dmGameSystem
             }
         }
 
-        dmModelDDF::ModelPlayAnimation msg;
```
britzl, line 197: Should we remove the ModelPlayAniomation message from model_ddf.proto?

engine/gamesys/src/gamesys/scripts/script_model.cpp
```diff
@@ -194,14 +196,24 @@ namespace dmGameSystem
...
-        msg.m_PlaybackRate = playback_rate;
+        ModelWorld* world;
+        ModelComponent* component;
+        dmScript::GetComponentFromLua(L, 1, MODEL_EXT, (dmGameObject::HComponentWorld*)&world, (dmGameObject::HComponent*)&component, 0);
+        if (!component)
+        {
+            return luaL_error(L, "the component '%s' could not be found", lua_tostring(L, 1));
+        }
+
+        dmRig::Result result = dmGameSystem::CompModelPlayAnimation(world, component, anim_id, playback, blend_duration, offset, playback_rate, sender, functionref);
```
JCash: Note that switching from message to synchronous animation change might be a breaking change, as it will change _when_ things happen (i.e. the completed event).

engine/gamesys/src/dmsdk/gamesys/components/comp_model.h
```diff
@@ -0,0 +1,55 @@
...
+     * @param blend_duration [type: float] Duration of a linear blend between the current and new animation.
+     * @param offset [type: float] The normalized initial value of the animation cursor when the animation starts playing.
+     * @param playback_rate [type: float] The rate with which the animation will be played. Must be positive.
+     * @param listener [type: dmMessage::URL] Optional URL to send play events to.
+     * @param functionref [type: int] Optional Lua function reference to call with play events.
+     * @return result [type: dmRig::Result] Result of the operation.
+     */
+    dmRig::Result CompModelPlayAnimation(HModelWorld world, HModelComponent component, dmhash_t anim_id, uint32_t playback, float blend_duration, float offset, float playback_rate, dmMessage::URL listener, int functionref);
```
JCash: In making this a C++ api, I'd like to see if we can keep Lua/Scripting out of it.
E.g. if you want to completely control the events from the C++ side? How to get a callback?

I think the CompModelPlayAnimation() could take a C callback and a user context, and the caller can decide on what to do.

The problem is that this small task grows a little a bit, as such an api is more complicated, as we have few use cases, and the data payload differs on event type. (See CompModelEventCallback in comp_model.cpp)

E.g. in comp_model.h
```c++
#include <dmsdk/rig/rig.h> // enums and types

typedef void (*FModelAnimationCallback)(void* user_ctx, dmRig::RigEventType event, void* event_data);
```

C++ Example: `CompModelPlayAnimation(..., callback, some_callback_data)`

And, in the case of script_model.cpp, it might look like:
```c++
FModelAnimationCallback callback = ScriptModelAnimationCallback
dmScript::LuaCallbackInfo* luacbk = dmScript::CreateCallback(L, 4);

CompModelPlayAnimation(..., callback, luacbk);
```

Internally, it would change from the m_Listener/m_FunctionRef to also have a m_Callback+m_CallbackContext.
If m_Callback is set, call that one first, otherwise post message to m_Listener if it's set.

JCash (changes requested):
overall good, but let's see if we can separate Lua from the component some more.

engine/gamesys/src/dmsdk/gamesys/components/comp_model.h
britzl: Yeah, I realised that! Fixed!

engine/gamesys/src/dmsdk/gamesys/components/comp_model.h
britzl: Let's do it!

engine/gamesys/src/dmsdk/gamesys/components/comp_model.h
```diff
@@ -0,0 +1,66 @@
...
+     * @param playback [type: dmRig::RigPlayback] Playback mode.
+     * @param blend_duration [type: float] Duration of a linear blend between the current and new animation.
+     * @param offset [type: float] The normalized initial value of the animation cursor when the animation starts playing.
+     * @param playback_rate [type: float] The rate with which the animation will be played. Must be positive.
+     * @param listener [type: dmMessage::URL] Optional URL to send play events to.
+     * @param callback [type: FModelAnimationCallback] Optional function callback to send play events to.
+     * @param callback_ctx [type: void*] Optional function callback context (only when callback is set)
+     * @return result [type: dmRig::Result] Result of the operation.
+     */
+    dmRig::Result CompModelPlayAnimation(HModelWorld world, HModelComponent component, dmhash_t anim_id, dmRig::RigPlayback playback, float blend_duration, float offset, float playback_rate, dmMessage::URL listener, FModelAnimationCallback callback, void* callback_ctx);
```
JCash: The "dmMessage::URL listener" is part of the Lua callback mechanism,
It should be enough for the caller to put the listener in the callback_ctx if they want to?

Or am I missing something?

JCash (changes requested):
The message (lua) listener is still part of the api.

engine/gamesys/src/dmsdk/gamesys/components/comp_model.h
```diff
@@ -0,0 +1,65 @@
...
+     * @param anim_id [type: dmhash_t] Animation to play.
+     * @param playback [type: dmRig::RigPlayback] Playback mode.
+     * @param blend_duration [type: float] Duration of a linear blend between the current and new animation.
+     * @param offset [type: float] The normalized initial value of the animation cursor when the animation starts playing.
+     * @param playback_rate [type: float] The rate with which the animation will be played. Must be positive.
+     * @param callback [type: FModelAnimationCallback] Optional function callback to send play events to.
+     * @param callback_ctx [type: void*] Optional function callback context (only when callback is set)
+     * @return result [type: dmRig::Result] Result of the operation.
+     */
+    dmRig::Result CompModelPlayAnimation(HModelWorld world, HModelComponent component, dmhash_t anim_id, dmRig::RigPlayback playback, float blend_duration, float offset, float playback_rate, FModelAnimationCallback callback, void* callback_ctx);
```
JCash, line 62: 👍

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -2526,25 +2509,7 @@ namespace dmGameSystem
         }
         else if (params.m_Message->m_Descriptor != 0x0)
         {
-            if (params.m_Message->m_Id == dmModelDDF::ModelPlayAnimation::m_DDFDescriptor->m_NameHash)
```
JCash, line 2529: Not being able to send a message to play animation, is this a breaking change?
E.g. is someone did `msg.post("#model", "play_animation")` ?

(I'm all for it, I just want us to be clear on what it means for us/the users)

engine/gamesys/src/gamesys/scripts/script_model.cpp
```diff
@@ -146,6 +148,82 @@ namespace dmGameSystem
...
+                    }
+                }
+                dmScript::TeardownCallback(cbctx->m_LuaCallback);
+            }
+            dmScript::DestroyCallback(cbctx->m_LuaCallback);
+        }
+        else
+        {
+            switch (event_type) {
+                case dmRig::RIG_EVENT_TYPE_COMPLETED:
```
JCash: iirc, the curly braces end up on a new line.
ask the agent to add .clang_format to any changes.

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -133,7 +133,8 @@ namespace dmGameSystem
         ModelResource*              m_Resource;
         dmRig::HRigInstance         m_RigInstance;
         dmMessage::URL              m_Listener;
-        int                         m_FunctionRef;
+        FModelAnimationCallback     m_Callback;
+        void*                       m_CallbackContext;
```
JCash: Is this m_Listener needed anymore? 🤔

JCash (changes requested):
seems the m_Listener in comp_model.cpp isn't used anymore?

also could probably benefit from using .clang_format on the changes.

engine/gamesys/src/dmsdk/gamesys/components/comp_model.h
britzl: Removed and handled in script_model.coo

engine/gamesys/src/gamesys/components/comp_model.cpp
britzl: This is not part of our public API and we have no examples of using a "play_animation" message. It seems to be used only internally between script and component.

engine/gamesys/src/gamesys/components/comp_model.cpp
britzl: Removed!

engine/gamesys/src/gamesys/scripts/script_model.cpp
britzl: I added BraceWrapping AfterCaseLabel to clang-format to handle this case.

Issue #11408, opened 2025-10-23
Undo history is cleared when a text file is created during game launch


ufgo:
Describe the bug (REQUIRED)
The Defold Editor clears the Undo/Redo history if a new text file is created in the project folder while running the game from the editor.

To Reproduce (REQUIRED)
	1.	Use a library such as imgui, which creates an imgui.ini file when the game runs.
	2.	Launch the game from the editor (Cmd+B / Ctrl+B).
	3.	Stop the game.
	4.	Observe that the Undo/Redo history is completely cleared.

Expected behavior (REQUIRED)
The Undo/Redo history should remain intact even if a temporary or configuration file (like imgui.ini) is created during runtime.

Defold version (REQUIRED):
	•	Version: 1.11.1

Platforms (REQUIRED):
	•	Platforms: macOS, Windows (probably all)
	•	OS: macOS
	•	Device: Desktop

Workaround (OPTIONAL):
Create a .defignore file in the project root and add the following line:

`/imgui.ini`

Additional context (OPTIONAL):
This issue reproduces consistently every time the game is launched from the editor while imgui.ini is not listed in .defignore.

JCash:
For the implementer in the editor, we have some logic regarding the log.txt file.
It may reside in different folders:

https://github.com/defold/defold/blob/dev/engine/engine/src/engine.cpp#L943-L982

PR #11875, opened 2026-02-10
Include engine log files among defignored patterns in the editor

matgis:
The editor will now ignore log files written by the engine runtime when scanning the project for changes. Previously, changes to these log files would cause the editor to reload external changes into the graph and clear the undo queue unless the log file patterns were explicitly listed in a `.defignore` file or configured to be written outside the project directory.

Partially addresses #11408.

### Technical changes
The `defignore-pred` has been updated to ignore engine log files. For this, it needs to take into account:
  * The `proj-path` patterns listed in the `.defignore` file.
  * The hidden `project.log_dir` setting in `game.project`.
  * The `defold.log.dir` JVM property.
  
Since we need to obtain the `.defignore` patterns before we populate the `workspace`, the new implementation relies heavily on memoization to avoid parsing these files over and over. Every `project-directory` root is associated with several memoized functions limited to a single cached result based on the `last-modified-time` of each input file. These are in turn composed into "umbrella" memoized functions that produce the list of `proj-path-patterns` and finally the `defignore-pred` itself.

Extensive tests have been added to `resource_test.clj` to ensure the caching works as expected.

* Added optional `opts` map argument to `fn/memoize`, allowing you to specify a `:limit` to the number of cached calls.
* Added `path/resolve-normalized`, which is probably what you want most of the time. Updated `path/resolve` docs to reflect that it does not normalize the resulting path.
* Added `test-support/with-cleared-system-properties!` macro so the state of the `defold.log.dir` JVM property does not interfere with the tests.

### Testing notes
If you test the functionality in a `dev` build of the editor, make sure to comment out the `"-Ddefold.log.dir="` line under `:jvm-opts` in `project.clj` first. This setting causes the editor to override `project.log_dir` to the current working directory, which will be outside of the project directory.

### Intermittent Spec error logged during tests
This change appears to have unearthed an issue around how the `lsp` system works with the editor tests. You will occasionally see a "Spec assertion failed" error logged, caused by a message posted to the `lsp` async channel, which doesn't get handled until the originating test has deleted the project directory.

Running this command, I can get it to happen about half the time on my local machine:
```
lein test :only integration.script-properties-test/zip-resource-reference-remains-valid-after-script-reload-test integration.sound-test
```

The error does not cause a test failure - it just gets printed to `stdout`. I think we should investigate it as a separate problem rather than attempting to fix it as part of this PR.

### DIFF
--- a/editor/src/clj/editor/resource.clj
+++ b/editor/src/clj/editor/resource.clj
@@ -14,12 +14,16 @@
 
 (ns editor.resource
   (:require [clojure.java.io :as io]
+            [clojure.spec.alpha :as s]
             [clojure.string :as string]
             [cognitect.transit :as transit]
             [dynamo.graph :as g]
             [editor.core :as core]
             [editor.fs :as fs]
-            [schema.core :as s]
+            [editor.settings-core :as settings-core]
+            [editor.system :as system]
+            [schema.core :as schema]
+            [service.log :as log]
             [util.coll :as coll :refer [pair]]
             [util.defonce :as defonce]
             [util.digest :as digest]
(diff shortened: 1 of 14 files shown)


editor/src/clj/util/fn.clj
```diff
@@ -138,15 +170,22 @@
...
+  ([ifn]
+   (memoize nil ifn))
+  ([opts ifn]
+   (if (::memoize-cache (meta ifn))
+     ifn ; Already memoized.
+     (let [arity (max-arity ifn)]
+       (case arity
+         1 (memoize-one opts ifn)
+         2 (memoize-two opts ifn)
+         (memoize-any opts ifn arity))))))
```
matgis, line 188: Indentation from added arity, and supply `opts` to all functions.

editor/src/clj/util/path.clj
```diff
@@ -173,11 +173,22 @@
   coerced Path. If other is an empty path, returns the Path coerced from the
   first argument. Otherwise, we consider this path to be a directory and
   resolve the other path against this path. For example, on UNIX, if this path
-  is `/a/b` and the other path is `../c/d`, then the resulting resolved path would
-  be `/a/c/d`."
+  is `/a/b` and the other path is `../c/d`, then the resulting resolved path
+  would be `/a/b/../c/d`."
```
matgis, line 177: This is the docstring for `path/resolve`, which was previously incorrect, since the resulting path wasn't normalized.

editor/test/editor/resource_test.clj
```diff
@@ -13,56 +13,353 @@
...
+  (testing "Trailing slashes are stripped from patterns."
+    (is (= ["/a" "/b" "/c"]
+           (resource/lines->proj-path-patterns ["/a" "/b" "/b/" "/a/" "/c/"])))))
+
+(deftest make-proj-path-patterns-pred-test
   (testing "Returns fn/constantly-false when there are no valid patterns."
-    (is (identical? fn/constantly-false (resource/lines->proj-path-patterns-pred nil)))
-    (is (identical? fn/constantly-false (resource/lines->proj-path-patterns-pred [])))
-    (is (identical? fn/constantly-false (resource/lines->proj-path-patterns-pred ["a"])))
-    (is (identical? fn/constantly-false (resource/lines->proj-path-patterns-pred ["a" "b"]))))
```
matgis, line 31: Instead of testing the predicate, we test the pattern filtering that happens inside `resource/lines->proj-path-patterns` separately.

editor/test/editor/resource_test.clj
```diff
@@ -13,56 +13,353 @@
...
+    (is (identical? fn/constantly-false (resource/make-proj-path-patterns-pred-raw nil)))
+    (is (identical? fn/constantly-false (resource/make-proj-path-patterns-pred-raw []))))
 
   (testing "Returns functioning predicate when there are some valid patterns."
-    (let [proj-path-patterns-pred (resource/lines->proj-path-patterns-pred ["a" "/b"])]
+    (let [proj-path-patterns-pred (resource/make-proj-path-patterns-pred-raw ["/b"])]
```
matgis, line 55: `resource/make-proj-path-patterns-pred` was changed to take a source `proj-path-patterns-file` argument so we can more efficiently cache the contents per file. It was simpler to make the tests use the raw implementation than adding a dummy file argument to each call.

editor/src/clj/editor/resource.clj
```diff
@@ -163,58 +167,226 @@
...
+(s/def ::log-directory-override-pathname (s/nilable string?))
+
+(s/def ::defignore-patterns-key
+  (s/keys :req-un [::defignore-mtime
+                   ::project-directory
+                   (or ::log-directory-override-pathname ::game-project-mtime)]))
+
+(defn- make-defignore-patterns-key [^File project-directory]
+  {:pre [(s/assert ::project-directory project-directory)]
+   :post [(s/assert ::defignore-patterns-key %)]}
```
vlaaad, line 210: Just so you are aware, spec asserts are pretty slow and therefore are not compiled in the release build. Instead of an assertion, the form gets replaced with the value, basically acting as a truthy check. It's fine here!

editor/src/clj/editor/resource.clj
```diff
@@ -163,58 +167,226 @@
...
+(defn- make-defignore-patterns-raw [defignore-patterns-key]
+  {:pre [(s/assert ::defignore-patterns-key defignore-patterns-key)]
+   :post [(s/assert ::proj-path-patterns %)]}
+  (let [{:keys [defignore-mtime project-directory]} defignore-patterns-key
+        defignore-file (io/file project-directory ".defignore")]
+    (into []
+          (distinct)
+          (concat
+            (log-file-proj-path-patterns defignore-patterns-key)
+            (read-proj-path-patterns defignore-file defignore-mtime)))))
```
vlaaad: Laziness?? In my transducer paradise???

vlaaad (dismissed):
LGTM! It would be cool if we made text files reloadable with an undo step.

editor/src/clj/editor/resource.clj
matgis: I'll fix!

Issue #11876, opened 2026-02-10
Label component throws exceptions when font or material are nil


vlaaad:
**Describe the bug (REQUIRED)**
The editor throws different errors when a label component has Font and/or Material set to nil

**To Reproduce (REQUIRED)**
1. Create `repro.label` component
2. Set `font` (or `material`) property to an empty string

**Expected behavior (REQUIRED)**
No error dialogs appear

**Defold version (REQUIRED):**
  - `dev`

**Errors:**

When the font is set to nil:
```
java.lang.NullPointerException: Cannot invoke "java.lang.Number.doubleValue()" because "y" is null
        at clojure.lang.Numbers.divide(Numbers.java:3899)
        at editor.font$get_texture_recip_uniform.invokeStatic(font.clj:450)
        at editor.font$get_texture_recip_uniform.invoke(font.clj:447)
        at editor.label$fn__125695.invokeStatic(label.clj:205)
        at editor.label$fn__125695.invoke(label.clj:200)
        at clojure.lang.AFn.applyToHelper(AFn.java:154)
        at clojure.lang.AFn.applyTo(AFn.java:144)
        at clojure.lang.AFunction$1.doInvoke(AFunction.java:33)
        at clojure.lang.RestFn.invoke(RestFn.java:411)
        at clojure.lang.Var.invoke(Var.java:386)
        at editor.label$LabelNode$behavior$scene$fn__126099.invoke(label.clj:253)
        at internal.node$trace_expr.invokeStatic(node.clj:46)
        at internal.node$trace_expr.invoke(node.clj:35)
        at editor.label$LabelNode$behavior$scene.invokeStatic(label.clj:253)
        at editor.label$LabelNode$behavior$scene.invoke(label.clj:253)
        at clojure.lang.Var.invoke(Var.java:395)
        at internal.node.NodeImpl.produce_value(node.clj:338)
        at internal.node$pull_first_input_value.invokeStatic(node.clj:1229)
        at internal.node$pull_first_input_value.invoke(node.clj:1224)
        at editor.app_view$AppView$behavior$active_scene$fn__83985.invoke(app_view.clj:293)
        at internal.node$trace_expr.invokeStatic(node.clj:46)
        at internal.node$trace_expr.invoke(node.clj:35)
        at editor.app_view$AppView$behavior$active_scene.invokeStatic(app_view.clj:293)
        at editor.app_view$AppView$behavior$active_scene.invoke(app_view.clj:293)
[...]

PR #11878, opened 2026-02-10
Localize error messages in the editor

### LINKED ISSUE #5453: Externalize editor localization text

vlaaad:
Error messages are now localized in the editor.

Technical notes:
I used Codex a lot to extract hardcoded strings into localization files. The hardest part was checking the errors in the UI, but I think I managed to reproduce more than 95% of the localized errors: 
Screenshot 2026-02-11 at 08 31 05

I created [a follow-up issue](https://github.com/defold/defold/issues/11876), so we can properly support invalid values in the label component.

The essential things I did myself: changes in `build-errors-view` ns, `error-values` ns, error rendering in the scene view, etc.

Fix #5453


editor/src/clj/editor/disk.clj
```diff
@@ -226,58 +226,59 @@
...
-              (future
-                (try
-                  (render-save-progress! (progress/make-indeterminate (localization/message "progress.reading-timestamps")))
-                  (project/reload-plugins! project written-resources)
-                  (lsp/touch-resources! (lsp/get-node-lsp project) written-resources)
-                  (cond
-                    reload-required
-                    (complete! (blocking-reload! render-reload-progress! workspace [] changes-view))
+  (g/let-ec [workspace (project/workspace project evaluation-context)
+             localization (workspace/localization workspace evaluation-context)]
```
vlaaad, line 230: Mostly indentation to pass around localization, I recommend using the "Hide Whitespace" setting in this PR.

editor/src/clj/editor/label.clj
```diff
@@ -283,7 +284,6 @@
             (dynamic label (properties/label-dynamic :label :pivot))
             (dynamic tooltip (properties/tooltip-dynamic :label :pivot)))
   (property blend-mode g/Any (default (protobuf/default Label$LabelDesc :blend-mode))
```
vlaaad, line 285: First of all, we never used `tip` dynamic; it's called a `tooltip`. Second, it's about the deprecated blend mode (`:blend-mode-add-alpha`) that doesn't even exist in the label protobuf definition.

editor/src/clj/editor/particlefx.clj
```diff
@@ -918,11 +921,11 @@
                                  {:type resource/Resource
                                   :ext ["material"]}))
             (dynamic error (g/fnk [_node-id material material-max-page-count material-shader texture-page-count]
-                             (prop-resource-error :fatal _node-id :material material "Material")
+                             (prop-resource-error :fatal _node-id :material material material-message)
                              (validate-material _node-id material material-max-page-count material-shader texture-page-count))))
 
   (property blend-mode g/Keyword (default (protobuf/default Particle$Emitter :blend-mode))
-            (dynamic tip (validation/blend-mode-tip blend-mode Particle$BlendMode))
+            (dynamic tooltip (validation/blend-mode-tip blend-mode Particle$BlendMode))
```
vlaaad, line 928: Yeah, it's a `tooltip`.

editor/resources/localization/en.editor_localization
```diff
@@ -2195,3 +2211,86 @@ quick-help.reopen-closed-file = Reopen Closed File
...
+error.gui.material-not-found-in-scene = Material "{material}" does not exist in the scene
+error.gui.texture-not-found-in-scene = Texture "{texture}" does not exist in the scene
+error.gui.font-not-found-in-scene = Font "{font}" does not exist in the scene
+error.gui.particlefx-not-found-in-scene = ParticleFX "{particlefx}" does not exist in the scene
+error.gui.template-resource-type.material = Material
+error.gui.template-resource-type.font = Font
+error.gui.template-resource-type.particlefx = ParticleFX
+error.gui.template-resource-type.custom-resource = Custom resource
+error.gui.template-resource-conflicting-values = {type} "{resource}" has conflicting values in templates: {values}
+error.gui.max-nodes-exceeded = The actual number of nodes ({count}) exceeds "{property}" ({max})
```
matgis: Most of these are sentences. Some of them are several sentences. Perhaps we should make sure all of them ends with a period before we ship it out for localization?

editor/src/clj/editor/font.clj
```diff
@@ -725,8 +735,8 @@
                                     [:samplers :material-samplers]
                                     [:shader :material-shader])))
     (dynamic error (g/fnk [_node-id material-resource]
-                          (or (validation/prop-error :fatal _node-id :font validation/prop-nil? material-resource "Material")
-                              (validation/prop-error :fatal _node-id :font validation/prop-resource-not-exists? material-resource "Material"))))
+                          (or (validation/prop-error :fatal _node-id :font validation/prop-nil? material-resource material-message)
+                              (validation/prop-error :fatal _node-id :font validation/prop-resource-not-exists? material-resource material-message))))
```
matgis: These errors are reported on the `:font` property, but this is the `:material` property. It was like that before your changes, but let's fix it!

editor/src/clj/editor/gui.clj
```diff
@@ -582,26 +582,54 @@
...
+  "Use [[validate-optional-gui-resource-localized]] instead."
+  [fmt prop-kw node-id coll key]
   (when-not (empty? key)
     (validate-gui-resource :fatal fmt prop-kw node-id coll key)))
 
+;; SDK api
+(defn validate-optional-gui-resource-localized [message-fn prop-kw node-id coll key]
+  (when-not (empty? key)
+    (validate-gui-resource-localized :fatal message-fn prop-kw node-id coll key)))
+
```
matgis, line 633: It would be nice if we could revert to using the old names for these once we've updated the extensions.

editor/resources/localization/en.editor_localization
```diff
@@ -2195,3 +2211,86 @@ quick-help.reopen-closed-file = Reopen Closed File
...
+error.attribute-components-must-be-between = Components in vertex attribute "{attribute}" must be between {min} and {max}
+error.vertex-attribute = Vertex attribute "{attribute}": {error}
+error.vertex-attribute-normalize-float-data-type = Vertex attribute "{attribute}" uses Normalize with Float data type
+error.atlas-page-count-cannot-exceed = Atlas page count ({count}) cannot exceed {max} pages per atlas
+error.cubemap-image-sizes-differ = Cubemap image sizes differ:\n{right} = {right_size}\n{left} = {left_size}\n{top} = {top_size}\n{bottom} = {bottom_size}\n{front} = {front_size}\n{back} = {back_size}
+error.lua-preprocessing-failed = Lua preprocessing failed.\n{error}
+error.transpiler-compilation-failed = Compilation failed: {error}
+error.animation-set-build-failed = Failed to build {resource}: {error}
+error.blend-mode-add-alpha-replaced = "{old}" has been replaced by "{new}"
+error.duplicate-instance-id = "{id}" is in use by another instance
```
matgis: This message is used in the Gui to report Gui resource name conflicts. It might be nice to have it worded like `"{id}" is in use by another Texture`. Should we make `instance` an argument while we're at it?

editor/src/clj/editor/gui.clj
```diff
@@ -2675,7 +2738,7 @@
             (dynamic label (properties/label-dynamic :gui :particlefx))
             (dynamic tooltip (properties/tooltip-dynamic :gui :particlefx))
             (dynamic error (g/fnk [_node-id particlefx]
-                                  (prop-resource-error _node-id :particlefx particlefx "Particle FX")))
+                                  (prop-resource-error _node-id :particlefx particlefx particlefx-message)))
```
matgis: Style: Fix indentation while we're at it?

editor/src/clj/editor/gui.clj
```diff
@@ -2856,7 +2919,7 @@
                              :tx-attach-fn (gen-outline-node-tx-attach-fn attach-texture)}]))
   (output add-handler-info g/Any
           (g/fnk [_node-id]
-                 [_node-id "Textures..." texture-icon add-textures-handler {}])))
+                 [_node-id (localization/message "command.edit.add-embedded-component.variant.gui.option.textures") texture-icon add-textures-handler {}])))
```
matgis: Style: Fix indentation while we're at it?

editor/src/clj/editor/gui.clj
```diff
@@ -2994,7 +3057,7 @@
                              :tx-attach-fn (gen-outline-node-tx-attach-fn attach-layer :ordered-layer-names)}]))
   (output add-handler-info g/Any
           (g/fnk [_node-id]
-                 [_node-id "Layer" layer-icon add-layer-handler {}])))
+                 [_node-id (localization/message "command.edit.add-embedded-component.variant.gui.option.layer") layer-icon add-layer-handler {}])))
```
matgis: Style: Fix indentation while we're at it?

editor/src/clj/editor/gui.clj
```diff
@@ -2937,7 +3000,7 @@
                              :tx-attach-fn (gen-outline-node-tx-attach-fn attach-font)}]))
   (output add-handler-info g/Any
           (g/fnk [_node-id]
-                 [_node-id "Fonts..." font-icon add-fonts-handler {}])))
+                 [_node-id (localization/message "command.edit.add-embedded-component.variant.gui.option.fonts") font-icon add-fonts-handler {}])))
```
matgis: Style: Fix indentation while we're at it?

editor/resources/localization/en.editor_localization
```diff
@@ -2195,3 +2211,86 @@ quick-help.reopen-closed-file = Reopen Closed File
...
+error.gui.layer-from-template-not-found-in-scene = Layer "{layer}" from template scene does not exist in the scene; the parent layer will be used
+error.gui.adjust-mode-stretch-not-supported-for-particlefx = Adjust mode "Stretch" is not supported for ParticleFX nodes
+error.gui.material-not-found-in-scene = Material "{material}" does not exist in the scene
+error.gui.texture-not-found-in-scene = Texture "{texture}" does not exist in the scene
+error.gui.font-not-found-in-scene = Font "{font}" does not exist in the scene
+error.gui.particlefx-not-found-in-scene = ParticleFX "{particlefx}" does not exist in the scene
+error.gui.template-resource-type.material = Material
+error.gui.template-resource-type.font = Font
+error.gui.template-resource-type.particlefx = ParticleFX
+error.gui.template-resource-type.custom-resource = Custom resource
```
matgis: It's kind of weird to present this as a "Custom resource" to the user. It is basically a resource of a type registered by a plugin, like a Spine Scene. It might be better to simply refer to it as a "Resource"? I did notice that we refer to them as "Custom resource" in other localization labels in this file as well. Maybe this is a good time to fix it?

editor/src/clj/editor/gui.clj
```diff
@@ -4516,33 +4583,33 @@
...
     :defaults shape-base-node-defaults}
    {:node-type :type-text
     :node-cls TextNode
-    :display-name "Text"
+    :display-name outline-text-message
     :custom-type 0
     :icon text-icon
     :defaults (assoc visual-base-node-defaults
                 :manual-size default-manual-size
                 :text "<text>")}
```
matgis, line 4606: Should we localize the default string used for new text fields?

editor/src/clj/editor/label.clj
```diff
@@ -298,8 +299,8 @@
                                             [:font-data :font-data]
                                             [:build-targets :dep-build-targets])))
             (dynamic error (g/fnk [_node-id font]
-                                  (or (validation/prop-error :info _node-id :image validation/prop-nil? font "Font")
-                                      (validation/prop-error :fatal _node-id :image validation/prop-resource-not-exists? font "Font"))))
+                                  (or (validation/prop-error :info _node-id :image validation/prop-nil? font font-message)
+                                      (validation/prop-error :fatal _node-id :image validation/prop-resource-not-exists? font font-message))))
```
matgis: Another case where the property keyword does not match the property. These should be reported on `:font`, not `:image`. Let's fix it!

editor/src/clj/editor/label.clj
```diff
@@ -314,8 +315,8 @@
                                             [:samplers :material-samplers]
                                             [:build-targets :dep-build-targets])))
             (dynamic error (g/fnk [_node-id material]
-                                  (or (validation/prop-error :info _node-id :image validation/prop-nil? material "Material")
-                                      (validation/prop-error :fatal _node-id :image validation/prop-resource-not-exists? material "Material"))))
+                                  (or (validation/prop-error :info _node-id :image validation/prop-nil? material material-message)
+                                      (validation/prop-error :fatal _node-id :image validation/prop-resource-not-exists? material material-message))))
```
matgis: `:image` -> `:material`

editor/resources/localization/en.editor_localization
```diff
@@ -2195,3 +2211,86 @@ quick-help.reopen-closed-file = Reopen Closed File
...
+error.render.item = - {resource}: {error}
+error.render.unknown-resource = unknown
+error.render.and-more = …and more
+error.gui.layer-not-found-in-scene = Layer "{layer}" does not exist in the scene
+error.gui.layer-from-template-not-found-in-scene = Layer "{layer}" from template scene does not exist in the scene; the parent layer will be used
+error.gui.adjust-mode-stretch-not-supported-for-particlefx = Adjust mode "Stretch" is not supported for ParticleFX nodes
+error.gui.material-not-found-in-scene = Material "{material}" does not exist in the scene
+error.gui.texture-not-found-in-scene = Texture "{texture}" does not exist in the scene
+error.gui.font-not-found-in-scene = Font "{font}" does not exist in the scene
+error.gui.particlefx-not-found-in-scene = ParticleFX "{particlefx}" does not exist in the scene
```
matgis: There's quite a few messages like "TYPE ITEM not found in LIST" if you search for `not-found` in this file. Maybe we should try to consolidate them before we send this off for localization?

editor/src/clj/editor/particlefx.clj
matgis: Should we rename the weirdly named macro as well?

editor/src/clj/editor/particlefx.clj
```diff
@@ -68,6 +68,9 @@
 (def modifier-icon "icons/32/Icons_19-ParicleFX-Modifier.png")
 
 (def particlefx-ext "particlefx")
+(def image-message (localization/message "property.image"))
+(def material-message (localization/message "property.material"))
+(def animation-message (localization/message "property.particlefx.animation"))
```
matgis: I just noticed we have a `properties/memoized-label-message` function that returns messages such as these. Should we be using it every time we do this?

editor/src/clj/editor/resource_io.clj
```diff
@@ -40,14 +39,10 @@
...
-                     (resource/proj-path resource)
-                     (if-some [message (ex-message exception)]
-                       (str ": " message)
-                       "."))
-             {:type :invalid-content
-              :resource resource
-              :exception exception}))
+             (localization/message "error.resource-not-loaded"
+                                   {"resource" (resource/proj-path resource)
+                                    "error" (or (ex-message exception) "none")})
```
matgis: Should we have a separate localization key for "unknown error" or something?

editor/src/clj/editor/scene.clj
```diff
@@ -109,44 +108,46 @@
...
           (map (fn [{:keys [_node-id message]}]
-                 (let [resource-name (get-resource-name _node-id)]
-                   (format "- %s: %s"
-                           (or resource-name "unknown")
-                           message)))))]
+                 (localization-state
+                   (localization/message "error.render.item"
+                                         {"resource" (or (get-resource-name _node-id)
+                                                         (localization-state error-render-unknown-resource-message))
+                                          "error" message})))))]
```
matgis: I think I've brought this up before, but I'm not a huge fan of having the bullet be part of the localized message. At some point we might want to change the formatting to use something like an error icon instead.

I don't think this needs to be a localized message at all?

editor/src/clj/editor/scene.clj
```diff
@@ -109,44 +108,46 @@
...
+  [gl render-args _renderables _nrenderables]
   (when (= pass/overlay (:pass render-args))
-    (let [error-values (e/map renderable->error-value renderables)
-          error-message-lines (error-message-lines error-values)]
-      (->> error-message-lines
-           (e/map-indexed coll/pair)
-           (run!
-             (fn [[index error-message-line]]
-               (scene-text/overlay gl error-message-line 24.0 (- -22.0 (* 14 index)))))))))
+    (scene-text/overlay gl "RENDER ERROR" 24.0 -22.0)))
```
matgis, line 148: I think we render errors to the cljfx-based `overlay-anchor-pane` nowadays. I don't think we should even have this anymore. Better to make sure these `ErrorValues` end up in the `overlay-anchor-pane`.

editor/src/clj/editor/localization.clj
```diff
@@ -673,6 +673,4 @@
     items))
 
 ;; TODO:
-;;  - build errors
-;;  - missed things...
 ;;  - extensions: resource types, properties, outlines...
```
matgis, line 676: I found some more untranslated stuff: Scene editor toolbar and related popups, dropdowns based on Protobuf enums, the Breakpoints tab, "no content in table" in the form view, Add New Resource menu categories. Perhaps some of these have just not been translated by the community yet?

editor/src/clj/editor/tile_source.clj
```diff
@@ -86,6 +86,21 @@
...
+(def ^:private start-tile-message (localization/message "property.tile-source.start-tile"))
+(def ^:private end-tile-message (localization/message "property.tile-source.end-tile"))
+(def ^:private fps-message (localization/message "property.tile-source.fps"))
+(def ^:private tile-width-message (localization/message "property.tile-source.tile-width"))
+(def ^:private tile-height-message (localization/message "property.tile-source.tile-height"))
+(def ^:private tile-margin-message (localization/message "property.tile-source.tile-margin"))
+(def ^:private tile-spacing-message (localization/message "property.tile-source.tile-spacing"))
+(def ^:private extrude-borders-message (localization/message "property.tile-source.extrude-borders"))
+(def ^:private inner-padding-message (localization/message "property.tile-source.inner-padding"))
+(def ^:private more-than-16-collision-groups-in-use-message (localization/message "error.tile-source.more-than-16-collision-groups-in-use"))
```
matgis: It might be a bad idea to encode the current limit directly into the localization key. What if we increase it someday? I think the limit should really be a parameter.

editor/test/integration/script_properties_test.clj
```diff
@@ -231,6 +232,16 @@
...
 (def ^:private texture-resource-property? (partial resource-kind-property? "texture"))
+(def ^:private texture-resource-exts [".cubemap" ".jpeg" ".jpg" ".png" ".render_target"])
+
+(defn- resource-not-found-message [property resource]
+  (localization/message "error.resource-assignment-not-found" {"property" property
+                                                               "resource" resource}))
+
+(defn- resource-not-of-type-message [property resource exts]
+  (localization/message "error.resource-assignment-not-of-type" {"property" property
+                                                                 "resource" resource
```
matgis: Style: Break argument maps onto own line?

matgis (dismissed):
Works great overall! Changing the Language preference and seeing all my Build Errors get re-translated feels like magic! ❤️

I do think we should make the `error.tile-source.more-than-16-collision-groups-in-use` message take the current limit as a parameter before merging, though!

editor/src/clj/editor/particlefx.clj
vlaaad: I wanted to do it, but it's used by spine, so renaming a macro will introduce a breaking change.

editor/resources/localization/en.editor_localization
vlaaad: Makes sense!

editor/src/clj/editor/font.clj
vlaaad: Yes!

editor/resources/localization/en.editor_localization
vlaaad: Hmmph, in English it's going to be easy to add, but in other languages with gendered words (e.g., Russian), it would mean that the word `another` will need to change depending on the value we start to pass there, or we will have to use some unwieldy sentence construction like "by another object of type {type}"..

editor/resources/localization/en.editor_localization
vlaaad: Huh, I noticed that these keys are not even used. I think Codex added them, but then I made it reuse the existing ones and forgot to delete the added keys. Your comment makes sense, though! I'll rename it to just Resource.

editor/src/clj/editor/gui.clj
vlaaad: Nah :P

editor/resources/localization/en.editor_localization
vlaaad: Good idea!

editor/src/clj/editor/particlefx.clj
vlaaad: Yes!

editor/src/clj/editor/resource_io.clj
vlaaad: Hmm yes!

editor/src/clj/editor/scene.clj
vlaaad: Makes sense!

editor/src/clj/editor/scene.clj
vlaaad: I agree, but this is used in a graph `:substitute` input, I'm not sure it's safe to just remove the substitute...

editor/src/clj/editor/localization.clj
vlaaad: Some of the things you mention have translation keys already, and the new resource categories are already translated: 
Screenshot 2026-02-12 at 11 52 23

There will be missed things, but I'm not sure there needs to be this to-do.

PR #11882, opened 2026-02-11
Queue sound->Stop() for threaded OpenAL to prevent cross-thread crashes

meehaul:
When sound runs on a worker thread (DM_SOUND_WASM_SUPPORT_THREADS), Stop() now queues a command instead of calling OpenAL directly. The queued commands are processed in UpdateInternal on the sound thread.

This prevents crashes caused by calling OpenAL functions from multiple threads simultaneously, as some OpenAL implementations are not thread-safe for operations on non-main.

Ran a bunch of tests with logs to validate:

- One sound: sound plays to end on its own, stops cleanly
- Stop & delete: create object, play, stop, delete, ensured that path was stable
- Simultaneous play/stop: play & stop many different sounds individually

Verified that only Stop caused this crash

### DIFF
--- a/engine/sound/src/sound_openal.cpp
+++ b/engine/sound/src/sound_openal.cpp
@@ -87,6 +87,19 @@ namespace dmSound
         FSoundDataGetData   m_GetData;
     };
 
+#if DM_SOUND_WASM_SUPPORT_THREADS
+    enum CommandType
+    {
+        COMMAND_STOP
+    };
+
+    struct Command
+    {
+        CommandType     m_Type;
+        HSoundInstance  m_Instance;
+    };
+#endif
+
     struct SoundData
     {
         dmhash_t            m_NameHash;
@@ -142,6 +155,7 @@ namespace dmSound
 #if DM_SOUND_WASM_SUPPORT_THREADS
         dmConditionVariable::HConditionVariable m_CondVar;
         dmThread::Thread        m_Thread;
+        dmArray<Command>        m_CommandQueue;
 #endif
         int32_atomic_t          m_IsRunning;
         int32_atomic_t          m_Status;
@@ -502,6 +516,7 @@ namespace dmSound
 #if DM_SOUND_WASM_SUPPORT_THREADS
         sound->m_Thread = 0;
         sound->m_CondVar = 0;
+        sound->m_CommandQueue.SetCapacity(64);
 
         if (params->m_UseThread) {
             dmLogInfo("sound updates on worker thread");
@@ -600,6 +615,8 @@ namespace dmSound
         float extra_percent = 1.05f;
         return ceilf(f_frame_count * extra_percent) * 32;
     }
+
+    static void StopInternal(SoundSystem* sound, HSoundInstance sound_instance);
 
     static inline const char* GetSoundName(SoundSystem* sound, SoundInstance* instance)
     {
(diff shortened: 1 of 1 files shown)


engine/sound/src/sound_openal.cpp
```diff
@@ -87,6 +87,19 @@ namespace dmSound
...
+#if DM_SOUND_WASM_SUPPORT_THREADS
+    enum CommandType
+    {
+        COMMAND_STOP
+    };
+
+    struct Command
+    {
+        CommandType     m_Type;
+        HSoundInstance  m_Instance;
```
JCash, line 99: For communication purposes, we usually put the largest types first, as it's good rule-of-thumb.
It may not help in a lot of cases, but having developers think in such terms is always good.

engine/sound/src/sound_openal.cpp
```diff
@@ -142,6 +155,7 @@ namespace dmSound
 #if DM_SOUND_WASM_SUPPORT_THREADS
```
JCash, line 155: While it's "old" code, I'm generally not a fan of having too much if-def logic going on, as it gets harder to road/follow imho, as seen in this PR with the if-def segments added.

engine/sound/src/sound_openal.cpp
```diff
@@ -142,6 +155,7 @@ namespace dmSound
...
         dmConditionVariable::HConditionVariable m_CondVar;
         dmThread::Thread        m_Thread;
+        dmArray<Command>        m_CommandQueue;
```
JCash, line 158: In my mind, this grouping is strange:

* if we support threads, let's assume it doesn't support threads and stop/delete on the main thread
* if we don't support threads, let's stop/delete on the main thread.

Doesn't this tell us that we should always stop/delete on the main thread?
(And as such, the if-defs could be simplified)
Or am I missing something?

To be clear, it seems to me that we should always do the delay delete mechanic, in order to keep code complexity low, and also get same behavior on the platforms.

engine/sound/src/sound_openal.cpp
```diff
@@ -1189,6 +1194,31 @@ namespace dmSound
...
+            if (sound->m_CommandQueue.Remaining() == 0) {
+                dmLogError("Sound command queue full (%u), stop command dropped", sound->m_CommandQueue.Capacity());
+                return RESULT_OUT_OF_INSTANCES;
+            }
+            Command cmd;
+            cmd.m_Type = COMMAND_STOP;
+            cmd.m_Instance = sound_instance;
+            sound->m_CommandQueue.Push(cmd);
+            return RESULT_OK;
+        }
```
JCash, line 1217: I think this deserves to be in its own function now. E.g.  `StopDelayed()`

engine/sound/src/sound_openal.cpp
```diff
@@ -1273,6 +1303,19 @@ namespace dmSound
 
         DM_MUTEX_OPTIONAL_SCOPED_LOCK(sound->m_Mutex);
 
+#if DM_SOUND_WASM_SUPPORT_THREADS
+        // Process queued commands from other threads
+        for (uint32_t cmd_idx = 0; cmd_idx < sound->m_CommandQueue.Size(); ++cmd_idx)
```
JCash, line 1308: for communication/performance purposes, moving the size variable outside of the loop.
```c++
uint32_t num_commands = sound->m_CommandQueue.Size();
for (...)
```

engine/sound/src/sound_openal.cpp
```diff
@@ -1189,6 +1194,31 @@ namespace dmSound
...
+    Result Stop(HSoundInstance sound_instance)
+    {
+        SoundSystem* sound = g_SoundSystem;
+
+#if DM_SOUND_WASM_SUPPORT_THREADS
+        if (sound->m_Thread != 0) {
+            // Queue the stop command for execution on the audio thread
+            // to avoid cross-thread OpenAL calls which can crash
+            DM_MUTEX_SCOPED_LOCK(sound->m_Mutex);
+            if (sound->m_CommandQueue.Remaining() == 0) {
```
JCash, line 1208: While it's "old" code, it'd be good to see changes use our `.clang_format` file.

JCash (changes requested):
I think the logic seems a bit overcomplicated, and could be much more streamlined.

Also, some smaller formatting issues.

PR #11884, opened 2026-02-11
Fix LSP errors in tests that deleted the project dir on completion

vlaaad:
The errors in the test output were caused by LSP asynchronously looking up resources after the project dir was deleted. I think the easiest fix would be to simply delete these folders on exit, but I think it's not unreasonable to want to make some tests "tidy" and "isolated". I also considered disabling the LSP system completely in tests by default, but it led to some tests being suddenly stuck (the tests that built a project, which included requesting an LSP lint...) without any feedback, which I think is worse than an error in the log — it makes it much harder to figure out what's wrong...

### DIFF
--- a/editor/src/clj/editor/lsp.clj
+++ b/editor/src/clj/editor/lsp.clj
@@ -13,6 +13,7 @@
 ;; specific language governing permissions and limitations under the License.
 
 (ns editor.lsp
+  (:refer-clojure :exclude [await])
   (:require [clojure.core.async :as a :refer [<! >!]]
             [clojure.set :as set]
             [clojure.spec.alpha :as s]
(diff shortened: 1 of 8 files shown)


matgis (approved):
One concern with the approach taken here is that the `lsp/await` function will not be called if an exception is thrown that aborts the test.

It would be nice if we could combine it with the `directory-deleter` somehow, but it is tricky because a lot of tests don't create the `project` until we're already inside the `directory-deleter` scope.

But it is still an improvement!

vlaaad:
Yes, I initially wanted to maybe allow passing some cleanup callback to the deleter, but as you also noticed, unfortunately, it's not always possible...

PR #11888, opened 2026-02-12
Support more than 8 vertex streams

### LINKED ISSUE #9547: Allow more than 8 vertex / instancing stream slots
I decided to add instancing to my complex material, but it turns out that Defold strictly limits the number of attributes for instancing to 8 slots. The error is:

`Unable to add vertex stream 'position', stream declaration has no slots left (max: 8)`
etc

In fact, only 3 slots are left for my own needs, because the other 5 are already ‘filled’ by default:
- position
- normal
- uv
- mtx_world
- mtx_normal

This is very few. I want more!

Add option to set the value of "max vertex stream count".

Use `mat4` instead of `vec4`.

Jhonnyg:
Vertex shaders can now use more than 8 inputs, which was the previous limit. Now there is no limitation from the engine to how many input streams you can use, but please keep in mind that there are still GPU / driver limitations.

Fixes #9547 

### Technical changes

* The engine / renderer now uses only the dmGraphics::VertexAttributeInfo representation instead of the protobuf/ddf version
* Vertex stream declarations and vertex declarations have dynamic memory for storing the stream descsriptors
* Graphics utility functions have been moved from graphics.h to graphics_util.h in an attempt to clean up graphics.h a little bit


engine/gamesys/src/gamesys/gamesys_private.cpp
```diff
@@ -23,6 +23,20 @@ namespace dmGameSystem
...
+    dmGraphics::VertexAttributeInfos* GetScratchVertexAttributeInfos(uint32_t stream_count)
+    {
+        if (g_VertexAttributeInfoStreams.Size() < stream_count)
+        {
+            g_VertexAttributeInfoStreams.SetCapacity(stream_count);
+            g_VertexAttributeInfoStreams.SetSize(stream_count);
+        }
+        g_VertexAttributeInfos.m_Infos = g_VertexAttributeInfoStreams.Begin();
+        return &g_VertexAttributeInfos;
+    }
```
Jhonnyg, line 40: This might be the "iffy" part of this feature. But I want to avoid allocating memory all the time for temporary structs that live just a short while. I guess an alternative would be to use alloca instead everywhere this is used

engine/graphics/src/test/test_graphics.cpp
```diff
@@ -12,6 +12,10 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
+#if defined(_WIN32)
+#include <winsock2.h>
+#endif
```
Jhonnyg: This is a bit annoying. It's because graphics_util.h drags in windows headers via the dlib/endian.h header. But I wanted to clean up graphics.h a little bit, hence this dance of includes.. perhaps there's a better way?

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -2106,6 +2112,14 @@ namespace dmGraphics
         context->m_CurrentVertexDeclaration[binding_index]             = &context->m_MainVertexDeclaration[binding_index];
         context->m_CurrentVertexBufferOffset[binding_index]            = base_offset;
 
+        if (context->m_MainVertexDeclarationStreams[binding_index].Size() < vertex_declaration->m_StreamCount)
+        {
+            context->m_MainVertexDeclarationStreams[binding_index].SetCapacity(vertex_declaration->m_StreamCount);
+            context->m_MainVertexDeclarationStreams[binding_index].SetSize(vertex_declaration->m_StreamCount);
+        }
+        memset(context->m_MainVertexDeclarationStreams[binding_index].Begin(), 0, sizeof(VertexDeclaration::Stream) * vertex_declaration->m_StreamCount);
+        context->m_MainVertexDeclaration[binding_index].m_Streams = context->m_MainVertexDeclarationStreams[binding_index].Begin();
```
Jhonnyg, line 2117: Note to self: We need to do this for vendor adapters as well.

engine/graphics/src/graphics.h
```diff
@@ -648,6 +485,24 @@ namespace dmGraphics
...
+
+    static inline AssetType GetAssetType(HAssetHandle asset_handle)
+    {
+        return (AssetType) (asset_handle >> 32);
+    }
+
+    static inline HOpaqueHandle GetOpaqueHandle(HAssetHandle asset_handle)
+    {
+        return (HOpaqueHandle) asset_handle & 0xFFFFFFFF;
+    }
```
Jhonnyg, line 508: These couldn't be moved unfortunately.. I ended up in an include order hell when I tried, so best to avoid for now.

engine/render/src/render/material.cpp
```diff
@@ -207,6 +215,20 @@ namespace dmRender
...
+        info.m_SemanticType     = attribute.m_SemanticType;
+        info.m_DataType         = attribute.m_DataType;
+        info.m_CoordinateSpace  = attribute.m_CoordinateSpace;
+        info.m_VectorType       = attribute.m_VectorType;
+        info.m_Normalize        = attribute.m_Normalize;
+        info.m_StepFunction     = attribute.m_StepFunction;
+        info.m_ValueVectorType  = attribute.m_VectorType;
+        uint32_t tmp;
+        GetMaterialVertexAttributeValues(m, attribute, material_attribute, &info.m_ValuePtr, &tmp);
+    }
```
Jhonnyg: Maybe VertexAttribute and VertexAttributeInfo can be merged into a single struct 🤔

engine/render/src/render/material.cpp
```diff
@@ -52,78 +53,80 @@ namespace dmRender
         return (dmGraphics::VertexAttribute::VectorType) -1;
     }
 
-    static void SetVertexAttributeDefaultSettings(dmGraphics::VertexAttribute* attribute, dmhash_t name_hash, dmGraphics::Type graphics_type, uint32_t element_count, bool instancing_supported)
+    static void SetVertexAttributeInfoDefaultSettings(dmGraphics::VertexAttributeInfo* info, dmhash_t name_hash, dmGraphics::Type graphics_type, uint32_t element_count, bool instancing_supported, const uint8_t* value_ptr)
```
Jhonnyg, line 55: We now use the VertexAttributeInfo* in the material instead of storing both attribute and attributeinfo structs.

engine/render/src/render/render_private.h
```diff
@@ -85,24 +85,24 @@ namespace dmRender
...
+        dmArray<MaterialAttribute>                  m_MaterialAttributes;
+        dmArray<uint8_t>                            m_MaterialAttributeValues;
+        dmArray<RenderConstant>                     m_Constants;
+        dmArray<Sampler>                            m_Samplers;
+        dmRenderDDF::MaterialDesc::PbrParameters    m_PbrParameters;
+        uint32_t                                    m_TagListKey; // the key to use with GetMaterialTagList()
+        dmRenderDDF::MaterialDesc::VertexSpace      m_VertexSpace;
+        uint8_t                                     m_InstancingSupported : 1;
+        uint8_t                                     m_HasSkinnedAttributes : 1;
+        uint8_t                                     m_HasSkinnedMatrixCache : 1;
```
Jhonnyg, line 105: Mostly indentation, I got a stroke from the previous mixed indentation :)

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -809,19 +842,22 @@ namespace dmGameSystem
             if (!IsDefaultStream(attr_model.m_NameHash, attr_material.m_SemanticType, attr_material.m_StepFunction))
             {
                 uint32_t value_byte_size = dmGraphics::VectorTypeToElementCount(attr_model.m_VectorType) * dmGraphics::DataTypeToByteWidth(attr_model.m_DataType);
-                uint32_t attribute_index = non_default_attribute.m_NumInfos++;
-                non_default_attribute.m_Infos[attribute_index]              = attr_model;
-                non_default_attribute.m_Infos[attribute_index].m_VectorType = attr_material.m_VectorType;
-                non_default_attribute.m_VertexStride                       += value_byte_size;
+                uint32_t attribute_index = non_default_attribute->m_NumInfos++;
+
+                dmGraphics::VertexAttributeInfo* writable_attribute_info = (dmGraphics::VertexAttributeInfo*) &non_default_attribute->m_Infos[attribute_index];
```
JCash, line 846: It should already be the `dmGraphics::VertexAttributeInfo` type? 
The cast makes you doubt that fact.

engine/gamesys/src/gamesys/gamesys_private.cpp
JCash: Yes, those are my concerns as well.
I think we eventually will try to do things multithreaded, and then things like this will be a sync point.
The stack allocation approach was my first thought too.
But this will do for the MVP1 I think.

engine/gamesys/src/gamesys/gamesys_private.cpp
```diff
@@ -281,17 +296,49 @@ namespace dmGameSystem
...
 
+    void CopyAttributeInfos(dmGraphics::VertexAttributeInfos* dst, dmGraphics::VertexAttributeInfos* src, dmGraphics::CoordinateSpace default_coordinate_space)
+    {
+        memcpy((void*) dst->m_Infos, (void*) src->m_Infos, sizeof(dmGraphics::VertexAttributeInfo) * src->m_NumInfos);
+
+        dst->m_NumInfos = src->m_NumInfos;
+        dst->m_VertexStride = src->m_VertexStride;
+        dst->m_StructSize = src->m_StructSize;
+
+        for (int i = 0; i < dst->m_NumInfos; ++i)
```
JCash: can grab the "num_infos" to put into a local variable, to avoid unnecessary dereferencing the pointer for each access.

engine/graphics/src/test/test_graphics.cpp
JCash: I don't understand, endian_posix.h already includes "winsock2.h"
Including it twice shouldn't be a solution, which makes me think something else is the cause?

Having to do this seems like a code smell. It "worked before" so to speak, so I think we shouldn't need this now either?

JCash (changes requested):
Overall good as the MVP1.
* I share the concern of the the scratch buffer allocations, and would like to see those updates in an MVP2.
* Make sure we update the console backends with this change as well.

However, new include required in the test suggests a code smell, which I'm less comfortable with. I'd like to know the actual cause, before moving on.

engine/gamesys/src/gamesys/components/comp_model.cpp
Jhonnyg: the m_Infos is a const so i need to un-const it, which is fine in this case since I it comes from the scratch buffer

JCash (approved):

PR #11891, opened 2026-02-13
Improved stability on Android and reduced the ANR rate

AGulev:
Reduced the scope of the `g_EventLock` spinlock to prevent potential locking scenarios that could lead to ANRs on Android.

### Technical changes

I assume that `android_app_set_window()` blocks the UI thread until APP_CMD_TERM_WINDOW finishes.
`post_exec` runs only after `onAppCmd ` returns.
`onAppCmd` (`glfwAndroidHandleCommand`) locks `g_EventLock`, but engine thread may also hold the same lock while flushing commands/inputs stack:

```
Looper thread (looper_main):
ALooper_pollOnce(...)
 -> source->process(...)
   -> process_cmd(...) / process_input(...)           // android_native_app_glue.c
      -> app->onAppCmd(...) = glfwAndroidHandleCommand(...)  // use `g_EventLock` here
      -> app->onInputEvent(...) = glfwAndroidHandleInput(...)   // and use `g_EventLock` here as well
```

```
engine_main loop
 -> dmHID::Update(...)
   -> dmPlatform::PollEvents(...)
     -> glfwPollEvents()
       -> _glfwPlatformPollEvents()
         -> glfwAndroidFlushEvents(...)  // android_window.c - holds `g_EventLock` while process events
            -> _glfwAndroidHandleInput(...)
            -> surface/focus handling
```

If `engine_main` loop stuck with processing of somithing under the `g_EventLock` spinlock scope then `looper_main` can't process events.

Fix: reduce `g_EventLock` scope just to a small memcopy  of the commands/inputs stacks so we don't block the `looper_main` thread.


external/glfw/lib/android/android_window.c
```diff
@@ -379,11 +379,42 @@ static void CreateGLSurface()
...
+    int local_num_input_events = 0;
+
     spinlock_lock(&g_EventLock);
 
-    for (int i = 0; i < g_NumAppCommands; ++i)
+    local_num_app_commands = g_NumAppCommands;
+    if (local_num_app_commands > 0)
+    {
+        memcpy(local_app_commands, g_AppCommands, local_num_app_commands * sizeof(int));
+    }
```
JCash: This looks like AI "safety" imho which makes it less readable.
I suggest:
```c++
int num_app_commands =  g_NumAppCommands;
memcpy(app_commands, g_AppCommands, num_app_commands * sizeof(int)); 
```

external/glfw/lib/android/android_window.c
```diff
@@ -379,11 +379,42 @@ static void CreateGLSurface()
 
 void glfwAndroidFlushEvents()
 {
+    static struct InputEvent* flush_input_events = 0;
+    static int flush_input_events_capacity = 0;
+
+    int local_app_commands[MAX_APP_COMMANDS];
+    int local_num_app_commands = 0;
+    struct InputEvent* local_input_events = 0;
+    int local_num_input_events = 0;
```
JCash: Code style:
"Local" isn't necessary here, it's already implied.
Also, we avoid C89 where all variables are defined at the top, but instead where they are first used.

external/glfw/lib/android/android_window.c
```diff
@@ -379,11 +379,42 @@ static void CreateGLSurface()
...
+        memcpy(local_app_commands, g_AppCommands, local_num_app_commands * sizeof(int));
+    }
+    g_NumAppCommands = 0;
+
+    local_num_input_events = g_NumAppInputEvents;
+    if (local_num_input_events > flush_input_events_capacity)
     {
-        int cmd = g_AppCommands[i];
+        flush_input_events_capacity = local_num_input_events;
+        flush_input_events = realloc(flush_input_events, local_num_input_events * sizeof(struct InputEvent));
```
JCash: This is a memory leak, and while we don't currently have such runtime tests for Android, we may well have "soon", and then I'd like to avoid having leaks reported.

Is there a reasonable upper limit per frame?
What is the size of an InputEvent?
Can we use alloca() to allocate the copy on the stack?

JCash (changes requested):
Overall good idea.
However, I'd like to see a fix for the memory management.

external/glfw/lib/android/android_window.c
AGulev: This all manually written after simliar code we have here 
https://github.com/defold/defold/blob/ea9558f72593d325e5e27ba524f8c803b832c6e0/external/glfw/lib/android/android_init.c#L865-L871
(with both a check and realloc)

external/glfw/lib/android/android_window.c
AGulev: why it's a leak? 

>The realloc() function shall deallocate the old object pointed to by ptr and return a pointer to a new object that has the size specified by size.

https://pubs.opengroup.org/onlinepubs/9699919799/functions/realloc.html

AGulev: this logic repeats logic after, to make sure input events fit: https://github.com/defold/defold/blob/ea9558f72593d325e5e27ba524f8c803b832c6e0/external/glfw/lib/android/android_init.c#L865-L871

AGulev: I added a separate task to investigate it https://github.com/defold/defold/issues/11904

PR #11895, opened 2026-02-14
Moved texture transcoding to a thread during the resource preload step

### LINKED ISSUE #11825: Move BasisU transcoding out of the main thread
**Task (REQUIRED):**
For now BasisU transcoding happens on the main thread which is a reason of shutters during collection loading (if collections contain textures)  

Image

BasisU transcoding should happen on a job thread

britzl:
Texture transcoding is now done on a thread during the resource preload step.

Fixes #11825 

## Technical changes

Moved transcoding logic to `ResTexturePreload()` and store the result in the ImageDesc for later use in `ResTextureCreate()`.

CleanShot 2026-02-14 at 15 51 30@2x

Before:

https://github.com/user-attachments/assets/8a12a0a7-25a0-4c3f-a076-f19020aa8b01

After:

https://github.com/user-attachments/assets/3dc0029a-34be-498e-8f7e-9b1a674c70b2

### DIFF
--- a/engine/gamesys/src/gamesys/resources/res_texture.cpp
+++ b/engine/gamesys/src/gamesys/resources/res_texture.cpp
@@ -33,11 +33,15 @@ namespace dmGameSystem
 
     struct ImageDesc
     {
-        uint8_t*                  m_Memory; // the memory from the resource system, if we've taken ownership
-        dmGraphics::TextureImage* m_DDFImage;
-        uint8_t*                  m_DDFImageBytes;
-        uint8_t*                  m_DecompressedData[MAX_MIPMAP_COUNT];
-        uint32_t                  m_DecompressedDataSize[MAX_MIPMAP_COUNT];
+        uint8_t*                         m_Memory; // the memory from the resource system, if we've taken ownership
+        dmGraphics::TextureImage*        m_DDFImage;
+        uint8_t*                         m_DDFImageBytes;
+        uint8_t*                         m_DecompressedData[MAX_MIPMAP_COUNT];
+        uint32_t                         m_DecompressedDataSize[MAX_MIPMAP_COUNT];
+        dmGraphics::TextureImage::Image* m_Image;
+        uint8_t*                         m_ImageData;
+        dmGraphics::TextureFormat        m_OutputFormat;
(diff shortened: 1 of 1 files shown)


engine/gamesys/src/gamesys/resources/res_texture.cpp
```diff
@@ -174,228 +178,239 @@ namespace dmGameSystem
...
-                dmGraphics::TextureType texture_type = TextureImageToTextureType(image_desc->m_DDFImage->m_Type);
-                output_format = dmGraphics::GetSupportedCompressionFormatForType(context, output_format, image->m_Width, image->m_Height, texture_type);
-
-                if (!dmGraphics::Transcode(path, image, image_desc->m_DDFImage->m_Count, image_data_alternative, output_format, image_desc->m_DecompressedData, image_desc->m_DecompressedDataSize, &num_mips))
-                {
-                    dmLogError("Failed to transcode %s", path);
-                    continue;
-                }
+                dmGraphics::TextureCreationParams creation_params;
+                creation_params.m_Type = dmGraphics::TEXTURE_TYPE_2D;
```
JCash: While it's old code, I do wonder if we don't know the type we want? (2d vs 3d)
Since the code focuses on "surviving", then I'd think using `TextureImageToTextureType(image_desc->m_DDFImage->m_Type);` is something we may want here?

JCash (commented):
lgtm.

AGulev:
btw, as I understand we do load one texture (one resource) at a time? if there are 4 textures, all of them will be loaded and transcoded in a separate thread one after another?

britzl:
> btw, as I understand we do load one texture (one resource) at a time? if there are 4 textures, all of them will be loaded and transcoded in a separate thread one after another?

Yes, see load_queue_threaded.cpp where we create one thread "AsyncLoad" to load resources in parallel. I suppose this could be changed to use the job system maybe and multiple threads?

AGulev:
> Yes, see load_queue_threaded.cpp where we create one thread "AsyncLoad" to load resources in parallel. I suppose this could be changed to use the job system maybe and multiple threads?

yes, maybe

PR #11924, opened 2026-02-19
Moved doc creation fully to build_docs

britzl:


build_tools/waf_dynamo.py
```diff
@@ -319,9 +319,26 @@ def extract_docs(bld, src):
         for s in src:
             elements = _parse_source(s)
             for k,v in elements.items():
-                # turn path into key which will later be used as the
-                # build target filename
-                key = "-".join(os.path.normpath(s).split(os.sep))
+                # remove initial part of the path up to src/
```
britzl: Since we are not running extract_docs() from within each lib anymore we need to trim away the initial part of the path so that the final filename is the same as before.

I'm not sure this is entirely necessary, the API reference on defold.com/ref will still work, but some URLs will be different unless we do this.

engine/docs/src/script_doc.py
```diff
@@ -540,19 +540,19 @@ def parse_document(doc_str):
                         if param.name == "...":
                             continue
                         if not validate_lua_type(t, doc):
-                            errors.append("'%s' has unknown type '%s' for parameter '%s'" % (element.name, t, param.name))
+                            errors.append("'%s' has unknown type '%s' for parameter '%s' in file %s" % (element.name, t, param.name, file))
```
britzl, line 543: It was a bit annoying to read API warnings and errors and trying to figure out from exactly which file it originated

engine/docs/wscript
```diff
@@ -8,9 +8,11 @@ blddir = 'build'
 
 import waflib.Utils
 import sys, os, re, fnmatch
+import shutil
 import waflib.Options
 import waf_dynamo
 import TestContext
+from waf_dynamo import apidoc_extract_task
```
JCash: Is anyone else really using "apidoc_extract_task" from another place?
Perhaps it makes sense to move that function here?

engine/docs/wscript
```diff
@@ -36,22 +38,60 @@ def build(bld):
...
+            for f in files:
+                if os.path.splitext(f)[1] in exts:
+                    paths.append(os.path.join(root, f))
+        return paths
+
+    apidoc_extract_task(bld, ["../crash/src/script_crash.cpp"])
+    apidoc_extract_task(bld, _findfiles("../ddf/src/dmsdk", exts))
+    apidoc_extract_task(bld, _findfiles("../dlib/src/dmsdk", exts))
+    apidoc_extract_task(bld, _findfiles("../dlib/src/cs/dmsdk", exts))
+    apidoc_extract_task(bld, ["../engine/proto/engine/engine_ddf.proto", "../engine/src/engine_doc.h", "../engine/src/script/script_engine.cpp"])
```
JCash: To make fixes even more easy-to-use, I'd try to make it a bit closer to generic:

* Get the ENGINE_LIBS from build.py
* make a loop over the library folders and known patterns:
```python
for lib in ENGINE_LIBS:
    apidoc_extract_task(bld, _findfiles(f"../{lib}/src/dmsdk", exts))
    apidoc_extract_task(bld, _findfiles(f"../{lib}/src/cs/dmsdk", exts))
    ...
```

E.g., if a file has documentation, doesn't it deserve to be included?
This hard coded list is easy to forget about.

JCash (dismissed):
Lgtm.
Can probably move the helper function into the script_doc.py file.

PR #11926, opened 2026-02-19
Produce better local space vertices for sprites and particles

Jhonnyg:
Currently when writing custom vertex positions in the "local" coordinate space for sprites and particles, the size of the sprite/particle is applied to both the position _and_ the world transform. This means that you cannot render the primitive correctly when applying the world transform to the local position in a shader for example:

```glsl
// 1. the position attribute is in local space (i.e pixel local coordinate space)
gl_Position = view_proj * mtx_world * vec4(position.xyz, 1.0);
// 2. the position attribute is in world space
gl_Position = view_proj * vec4(position.xyz, 1.0);
```

These two statements should produce identical results, but doesn't since the size is applied to the world transform for both sprites and particles. This has now been fixed, which means that we can use custom vertex formats to properly do billboarding properly.

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_sprite.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_sprite.cpp
@@ -804,6 +804,7 @@ namespace dmGameSystem
         uint8_t* vertices,
         uint8_t* indices,
         bool is_indices_16_bit,
+        bool has_world_position_attribute,
         bool has_local_position_attribute,
         const Matrix4& world_matrix,
         uint32_t vertex_offset,
@@ -938,7 +939,11 @@ namespace dmGameSystem
         ys[1] = sy * slice9.getW();
         ys[2] = 1 - sy * slice9.getY();
 
-        EnsureSize(*scratch_positions_world, SPRITE_VERTEX_COUNT_SLICE9);
+        if (has_world_position_attribute)
+        {
+            EnsureSize(*scratch_positions_world, SPRITE_VERTEX_COUNT_SLICE9);
+        }
+
         if (has_local_position_attribute)
         {
             EnsureSize(*scratch_positions_local, SPRITE_VERTEX_COUNT_SLICE9);
@@ -971,18 +976,25 @@ namespace dmGameSystem
         {
             for (int x=0; x<4; x++)
             {
-                // convert from [0,1] to [-0.5, 0.5]
-                float px = xs[x] - 0.5f;
-                float py = ys[y] - 0.5f;
-                Point3 p = Point3(px - pivot_x, py - pivot_y, 0);
-
-                // Local space has size applied; world = mtx_world * local (mtx_world has no scale)
-                Vector4 local_pos(p.getX() * sp_width, p.getY() * sp_height, 0.0f, 1.0f);
-                if (has_local_position_attribute)
+                if (has_local_position_attribute || has_world_position_attribute)
                 {
-                    (*scratch_positions_local)[vertex_index] = local_pos;
+                    // convert from [0,1] to [-0.5, 0.5]
+                    float px = xs[x] - 0.5f;
+                    float py = ys[y] - 0.5f;
+                    Point3 p = Point3(px - pivot_x, py - pivot_y, 0);
+
+                    // Local space has size applied; world = mtx_world * local (mtx_world has no scale)
+                    Vector4 local_pos(p.getX() * sp_width, p.getY() * sp_height, 0.0f, 1.0f);
+
+                    if (has_local_position_attribute)
+                    {
+                        (*scratch_positions_local)[vertex_index] = local_pos;
+                    }
+                    if (has_world_position_attribute)
+                    {
+                        (*scratch_positions_world)[vertex_index] = world_matrix * local_pos;
+                    }
                 }
-                (*scratch_positions_world)[vertex_index] = world_matrix * local_pos;
 
                 vertices = dmGraphics::WriteAttributes(vertices, vertex_index++, 1, params);
             }
@@ -1321,8 +1333,8 @@ namespace dmGameSystem
         return anim_data;
     }
 
-    static void CreateVertexData(SpriteWorld* sprite_world, dmGraphics::VertexAttributeInfos* material_attribute_info, 
-        bool has_local_position_attribute, uint8_t** vb_where, uint8_t** ib_where, dmRender::RenderListEntry* buf, uint32_t* begin, uint32_t* end)
+    static void CreateVertexData(SpriteWorld* sprite_world, dmGraphics::VertexAttributeInfos* material_attribute_info,
+        uint8_t** vb_where, uint8_t** ib_where, dmRender::RenderListEntry* buf, uint32_t* begin, uint32_t* end)
     {
         DM_PROFILE("CreateVertexData");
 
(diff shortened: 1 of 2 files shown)


engine/particle/src/particle.cpp
```diff
@@ -1334,31 +1334,27 @@ namespace dmParticle
...
+            position_local_flat[0] = Vector3(-hx, -hy, 0.0f);
+            position_local_flat[1] = Vector3(-hx,  hy, 0.0f);
+            position_local_flat[2] = Vector3( hx,  hy, 0.0f);
+            position_local_flat[3] = position_local_flat[2];
+            position_local_flat[4] = Vector3( hx, -hy, 0.0f);
+            position_local_flat[5] = position_local_flat[0];
+
+            dmTransform::Transform particle_transform_no_scale = particle_transform;
+            particle_transform_no_scale.SetScale(Vector3(1.0f, 1.0f, 1.0f));
+            world_matrix = dmTransform::ToMatrix4(particle_transform_no_scale);
```
Jhonnyg, line 1349: This looks confusing, but scale / size works a bit different for particles it seems. The hierachy scale (from gameobject world) is being applied via:

```cpp
//particle.cpp:900
emitter_transform = dmTransform::Mul(instance->m_WorldTransform, emitter_transform);
//particle.cpp:964
particle->SetSourceSize(emitter_properties[EMITTER_KEY_PARTICLE_SIZE] * emitter_transform.GetScale());
```

And then this source size (which contains the hierarchy scale) is used to produce local positions which are then transformed by a world transform to produce world positions. This basically means that the local positions are not really 100% correct since the scale is still baked into local positions, but at least now multiplying the world matrix with the local positions will generate the correct world transform (since the scale of the matrix is 1,1,1).

I guess we could fix this by changing how the scale is supposed to work, but that will be a behaviour change in the particle library (i.e tests are failing), so I am a bit unsure if we should do this change or not. As it stands now, the billboading task doesn't really require it so I am a bit unsure if we should make this change or not...

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1500,19 +1510,27 @@ namespace dmGameSystem
...
                         positions_local[2] = Vector4(x1 * sp_width, y1 * sp_height, 0.0f, 1.0f);
                         positions_local[3] = Vector4(x1 * sp_width, y0 * sp_height, 0.0f, 1.0f);
+
+                        positions_world[0] = world_matrix * positions_local[0];
+                        positions_world[1] = world_matrix * positions_local[1];
+                        positions_world[2] = world_matrix * positions_local[2];
+                        positions_world[3] = world_matrix * positions_local[3];
+                    }
+                    else
+                    {
```
JCash: In the code above, we always write the world positions (line 985, line 1429)
Is this discrepancy intentional?

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1500,19 +1510,27 @@ namespace dmGameSystem
...
+                        positions_world[0] = world_matrix * Vector4(x0 * sp_width, y0 * sp_height, 0.0f, 1.0f);
+                        positions_world[1] = world_matrix * Vector4(x0 * sp_width, y1 * sp_height, 0.0f, 1.0f);
+                        positions_world[2] = world_matrix * Vector4(x1 * sp_width, y1 * sp_height, 0.0f, 1.0f);
+                        positions_world[3] = world_matrix * Vector4(x1 * sp_width, y0 * sp_height, 0.0f, 1.0f);
```
JCash: In the code seen on line 985, it's of the form `positions_world = world_matrix *  local_pos`.
Is that something we can/wish to do here as well?

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1960,9 +1977,10 @@ namespace dmGameSystem
             if (!component->m_Enabled || !component->m_AddedToUpdate)
                 continue;
             UpdateTransform(component, sub_pixels);
-            // we need to consider the full scale here
-            // I.e. we want the length of the diagonal C, where C = X + Y
-            float radius_sq = dmVMath::LengthSqr((component->m_World.getCol(0).getXYZ() + component->m_World.getCol(1).getXYZ()) * 0.5f);
+            // Bounding radius: world matrix has no scale, so incorporate size
+            Vector3 size = dmVMath::MulPerElem(component->m_Size, component->m_Scale);
+            Vector3 half_diagonal = (component->m_World.getCol(0).getXYZ() * size.getX() + component->m_World.getCol(1).getXYZ() * size.getY()) * 0.5f;
+            float radius_sq = dmVMath::LengthSqr(half_diagonal);
```
JCash, line 2001: if `size` now is the final size, do we really need to multiply it with a unit vector to find the length again?
Isn't it enough with:
```
float max_elem = dmMath::Max(size.getX(), dmMath::Max(size.getY(), size.getZ()));
float radius_sq = max_elem * max_elem;
```

(dmVMath doesn't currently expose the Vectormath::Aos::maxElem() function, so the manual approach works for now imho, but also easy to add in our vmath.h!)

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -1988,8 +2006,9 @@ namespace dmGameSystem
             const AnimationData* anim_data = GetOrCreateAnimationData(sprite_world, component);
             float pivot_x = 0.f, pivot_y = 0.f;
             GetPivot(anim_data, &pivot_x, &pivot_y);
-            Point3 pivot(-pivot_x, -pivot_y, 0.f);
-            Vector3 world_pos = (component->m_World * pivot).getXYZ();
+            Vector3 size = dmVMath::MulPerElem(component->m_Size, component->m_Scale);
+            Point3 pivot_scaled(-pivot_x * size.getX(), -pivot_y * size.getY(), 0.f);
+            Vector3 world_pos = (component->m_World * Vector4(pivot_scaled.getX(), pivot_scaled.getY(), 0.f, 1.f)).getXYZ();
```
JCash: The point already has 1.0f as the W component.
And, we create another duplicate Vector4 here.
I think this could be:
```
Vector3 world_pos = (component->m_World * pivot_scaled).getXYZ();
```

engine/particle/src/particle.cpp
```diff
@@ -1334,31 +1334,27 @@ namespace dmParticle
...
-                position_local_flat[0] = -x_local_pixel - y_local_pixel;
-                position_local_flat[1] = -x_local_pixel + y_local_pixel;
-                position_local_flat[2] = x_local_pixel + y_local_pixel;
-                position_local_flat[3] = position_local_flat[2];
-                position_local_flat[4] = x_local_pixel - y_local_pixel;
-                position_local_flat[5] = position_local_flat[0];
+                for (int i = 0; i < 6; ++i)
+                {
+                    Vector4 p_local(position_local_flat[i].getX(), position_local_flat[i].getY(), position_local_flat[i].getZ(), 1.0f);
+                    position_world_flat[i] = (world_matrix * p_local).getXYZ();
```
JCash: Some alternatives:
1) Change to 
```
Point3 p_local(position_local_flat[i]);
position_world_flat[i] = Vector3(world_matrix * p_local); // constructor accepts a point3
```

1) Change type of position_local_flat and position_world_flat to `Point3`.
```
position_world_flat[i] = world_matrix * p_local;
```

Granted, I believe this will leave the w=1.0f in these arrays, but also, they _are_ points, so is there a drawback to that?

JCash (changes requested):
Overall good.
Can probably simplify calculations a bit.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
Jhonnyg: Size is just the final size for the sprite component, but we need to factor in the world transform scale, right? The size is stripped from the world transform, but the scale isn't

engine/gamesys/src/gamesys/components/comp_sprite.cpp
Jhonnyg: Good catch!

engine/gamesys/src/gamesys/components/comp_sprite.cpp
JCash: that is true, I was distracted by the component's scale.

JCash (approved):
lgtm!

PR #11931, opened 2026-02-20
Show function list in the right sidebar

### LINKED ISSUE #6260: Functions list
Please add "Functions list" for better navigation bitween function
functions
s

### LINKED ISSUE #11886: When open collection becomes broken, I get this error
### Expected behaviour
Outline/Properties don't break until restart.
### Actual behaviour
Outline/Properties doesn't break until restart.

### Steps to reproduce
1. Create a collection with an embedded go
2. Change the file externally to make the collection protobuf invalid
3. Focus on editor
<hr/>

<table>
<tr><td>Build time</td><td></td></tr>
<tr><td>Defold channel</td><td></td></tr>
<tr><td>Defold editor sha</td><td>dc80a805a306becaf6ec371e76157f05dfe08f9a</td></tr>
<tr><td>Defold engine sha</td><td>9eeb3c81ff4bb115fbde5ccb001ab04af0522869</td></tr>
<tr><td>Defold version</td><td>dev</td></tr>
<tr><td>Error</td><td><a href='https://sentry.io/organizations/defold/issues/?query=id%3A"be79860584b446d38dd0c852913098d2"'>be79860584b446d38dd0c852913098d2</a></td></tr>
<tr><td>GPU</td><td>Apple M1 Pro</td></tr>
<tr><td>GPU Driver</td><td>2.1 Metal - 89.4</td></tr>
<tr><td>Java version</td><td>25+36-LTS</td></tr>
<tr><td>OS version</td><td>15.6.1</td></tr>
<table>

vlaaad:
We now show the function list in the right sidebar:
Screenshot 2026-02-20 at 08 37 39

Technical notes:
This was a moderately sized refactoring, and I'm surprised that Codex was effective in helping with it. See (now deleted) [REFACTOR.md](https://github.com/defold/defold/pull/11931/changes/48c59853e15257e54f88d09d470d031e4ace95d0#diff-52c180073019ecc41032561097ab5a0905cd41f977cf89dd978049877db33d16) if you are curious. My approach was: 
1. Dump the problem description on Codex and let him propose a refactoring plan. 
2. Iterate on the refactoring plan until I'm satisfied with it.
3. Split the plan into multiple small, manageable phases and implement it

So the refactoring went like this:
1. Refactor the right sidebar so the app view manages the shown items (always Outline and Properties), but driven by `sidebar-panes` output
2. Small visible showcase of the new functionality: form views only show Outline
3. Debugger (previously it had its own split, and it toggled visibility of right split/debugger split)
4. LSP document structure sync
5. Show Structure in the code view
6. Cleanup/bug fixes: LSP request retries (lua language server frequently responded with a retryable error on startup if we had open code files), Outline active tracking (since some globally available commands expected the Outline to always be up to date), split prefs persistence, maybe some more minor things...

Fix #6260
Fix #11886


editor/src/clj/editor/code/view.clj
```diff
@@ -1268,6 +1271,156 @@
...
 ;; endregion
 
+;; region code icons
+
+;; Completion icons by Microsoft (CC BY 4.0)
+;; https://github.com/microsoft/vscode-icons
+
+(def ^:private ^:const code-type-icon-size 12.0)
+
+(defn- code-type-icon [{:keys [type]}]
```
vlaaad, line 1283: Moved this code from below, added missing icons

editor/src/clj/editor/code/view.clj
```diff
@@ -3077,10 +3187,6 @@
 (defn- set-bar-ui-type! [ui-type]
   (case ui-type (:hidden :goto-line :find :replace) (.setValue bar-ui-type-property ui-type)))
 
-(defn- focus-code-editor! [view-node]
-  (let [^Canvas canvas (g/node-value view-node :canvas)]
-    (.requestFocus canvas)))
```
vlaaad, line 3082: Moved this higher.

editor/src/clj/editor/collection_proxy.clj
```diff
@@ -43,8 +43,12 @@
                :fields [{:path [:collection]
                          :localization-key "collectionproxy.collection"
                          :type :resource
-                         :filter "collection"}]}]
-   :values {[:collection] collection-resource}})
+                         :filter "collection"}
+                        {:path [:exclude]
+                         :localization-key "collectionproxy.exclude"
+                         :type :boolean}]}]
```
vlaaad, line 49: Added a missing `exclude` property

editor/src/clj/editor/lsp.clj
```diff
@@ -205,33 +205,117 @@
...
+              ;; Also, the server might be restarted by now, which means it's
+              ;; not guaranteed that it will have responses-ch in its requests
+              ;; map, but it's not a problem for dissoc
+              (update-in state [:server->server-state server :requests] dissoc responses-ch)
+              state))
+          (assoc state :requests (dissoc requests responses-ch))
+          servers))
       state)))
 
+(defn- send-requests! [state responses-ch & {:keys [requests requests-fn capabilities-pred language timeout-ms retries]
```
vlaaad: Moved from below, added retries.

editor/src/clj/editor/lsp.clj
```diff
@@ -205,33 +205,117 @@
...
+                                         (let [remaining-server-responses (dec (state-requests responses-ch))]
+                                           (if (zero? remaining-server-responses)
+                                             (dissoc state-requests responses-ch)
+                                             (assoc state-requests responses-ch remaining-server-responses)))))))]
+            (if error
+              state
+              (server-response-update-state result server state)))))
+      state)))
+
+(defn- cancel-requests [responses-ch servers]
```
vlaaad, line 262: Moved from below.

editor/src/clj/editor/lsp.clj
```diff
@@ -205,33 +205,117 @@
...
+  "See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#errorCodes
+
+  The server detected that the document content has changed during result
+  computation, we can simply retry the request"
+  -32801)
+
+(defn- retriable-error? [response]
+  (= content-modified-error-code (-> response :error :code)))
+
+(defn- on-server-response [server responses-ch {:keys [request response retries retry-delay-ms]}]
```
vlaaad, line 219: A lot of indentation changes, plus added retries.

editor/src/clj/editor/ui.clj
```diff
@@ -385,7 +386,7 @@
 
 (defmacro run-now
   [& body]
-  `(do-run-now (bound-fn [] ~@body)))
+  `(do-run-now (fn [] ~@body)))
```
vlaaad, line 390: Minot perf improvement that I discovered while profiling the editor during right sidebar interaction. `ui/ui-disabled?` is checked every frame with run-now, and for bound-fns, it's somewhat expensive to compute the binding map. But we don't need to do it, if we are on the ui thread already, so now I avoid it unless it's necessary.

matgis:
Bug: Immediately after opening my project and restoring the open tabs, the Outline view appears cropped for the Atlas that ended up as the selected tab. The root `TreeItem` and the top part of the tree is outside the view, as if it was scrolled down. However, there are no scroll bars, and I can't correct it by scrolling the mouse wheel. I can move the selection using the arrow keys, but it won't reframe the view when I move outside the visible region. Switching back and forth between two open editor tabs fixed the issue.

image

matgis:
The view no longer aligns vertically with the Asset browser. It would be even nicer if they both aligned with the scene view. The Outline is very close now, but is still a point higher than it should be in comparison to the main panel. image

editor/resources/editor.fxml
```diff
@@ -134,48 +134,9 @@
...
-                          </AnchorPane>
-                          <AnchorPane prefHeight="100.0" prefWidth="160.0">
-                            <TitledPane animated="false" collapsible="false" text="Variables" AnchorPane.bottomAnchor="0.0" AnchorPane.leftAnchor="0.0" AnchorPane.rightAnchor="0.0" AnchorPane.topAnchor="0.0">
-                              <TreeView id="debugger-variables" />
-                            </TitledPane>
-                          </AnchorPane>
-                        </SplitPane>
+                          <SplitPane id="right-split" dividerPositions="0.4" layoutX="50.0" layoutY="182.0" orientation="VERTICAL" prefHeight="200.0" prefWidth="160.0" AnchorPane.bottomAnchor="0.0" AnchorPane.leftAnchor="0.0" AnchorPane.rightAnchor="0.0" AnchorPane.topAnchor="0.0" >
+                              <AnchorPane/>
+                          </SplitPane>
```
matgis: Style: Indentation is inconsistent with the rest of the file.

vlaaad:
@matgis I addressed your comments

editor/src/clj/editor/code/view.clj
```diff
@@ -1268,6 +1271,156 @@
...
+                            :variable 11
+                            (:method :function :constructor) 17
+                            (:value :enum) 13
+                            :enum-member 13
+                            :constant 12
+                            :boolean 11
+                            :folder 13
+                            :struct 13
+                            :type-parameter 10
+                            16)
```
matgis, line 1302: I realize this was simply moved up in the file, but just so you know, we use `fxui/icon-graphic` elsewhere to uniformly fit SVG icons into a 16x16 view box.  Maybe it would make sense to update this code to use it at some point.

editor/src/clj/editor/code/view.clj
```diff
@@ -1268,6 +1271,156 @@
...
+      {:graphic {:fx/type fxui/horizontal
+                 :alignment :left
+                 :spacing :small
+                 :children (cond-> [{:fx/type code-type-icon :type kind}
+                                    {:fx/type fxui/label :text name}]
+                                   (not (coll/empty? detail))
+                                   (conj {:fx/type fxui/label
+                                          :text (summarize-document-symbol-detail detail)
+                                          :color :hint}))}})
+    {}))
```
matgis: Style: Invert `if` to avoid orphan `{}`?

editor/src/clj/editor/code/view.clj
```diff
@@ -1268,6 +1271,156 @@
...
+    (focus-code-editor! view-node)
+    (.consume event)))
+
+(defn- handle-structure-pane-mouse-clicked! [view-node ^MouseEvent e]
+  (when (ui/double-click-event? e)
+    (let [^TreeItem selected-item (-> e ^TreeView (.getSource) .getSelectionModel .getSelectedItem)]
+      (if (and selected-item (not (.isLeaf selected-item)) (not (.isExpanded selected-item)))
+        (.setExpanded selected-item true)
+        (focus-code-editor! view-node)))
+    (.consume e)))
```
matgis, line 1400: The behavior is a bit weird for expandable `TreeItems`. The first double-click appears to expand it in the `TreeView`, then the second time I double-click it I focus on the Code View. Maybe we need to handle and consume the event on `TreeItem` itself instead of the Pane? Or maybe even use `{:double-click-expand false}` for the `TreeItems`?

editor/src/clj/editor/code/view.clj
```diff
@@ -1268,6 +1271,156 @@
...
+
+(def ^:private structure-pane-message (localization/message "pane.structure"))
+
+(fxui/defc structure-pane
+  {:compose [{:fx/type fx/ext-watcher :ref (:localization props) :key :localization-state}]}
+  [{:keys [document-symbols localization-state view-node]}]
+  {:fx/type fxui/titled-pane
+   :title (localization-state structure-pane-message)
+   :content {:fx/type fx.ext.tree-view/with-selection-props
+             :props {:on-selected-item-changed #(navigate-to-document-symbol! view-node %)}
```
matgis, line 1410: I think we should maybe make it so that the current line indicator in the gutter shows up in the slightly dimmed color when the code view is unfocused to make it more clear where the cursor is when I navigate without focusing on the code editor view?

editor/src/clj/editor/code/view.clj
```diff
@@ -1346,6 +1501,15 @@
...
 
+  (output sidebar-panes g/Any :cached (g/fnk [_node-id document-symbols localization node-id+type+resource]
+                                        (cond-> [{:fx/type fxui/ext-dedupe-identical-desc
+                                                  :desc {:fx/type structure-pane
+                                                         :document-symbols document-symbols
+                                                         :localization localization
+                                                         :view-node _node-id}}]
+                                                (and node-id+type+resource (resource/overridable? (node-id+type+resource 2)))
+                                                (conj :properties))))
+
```
matgis, line 1513: Style: These days we tend to move the `(g/fnk ...)` expression to its own line do keep the indentation under control.

editor/src/clj/editor/fxui.clj
```diff
@@ -1713,3 +1726,12 @@
     ;; ResizableImageView uses fitToWidth and fitToHeight internally during
     ;; resize, ignoring externally set values
     :props (dissoc fx.image-view/props :fit-to-width :fit-to-height)))
+
+
```
matgis: Style: Extra newline?

editor/src/clj/editor/lsp.clj
```diff
@@ -205,33 +205,117 @@
...
+  {:pre [(not= (some? requests) (some? requests-fn))
+         (nat-int? retries)
+         (pos-int? timeout-ms)
+         (not (contains? (:requests state) responses-ch))]}
+  (let [requests-fn (or requests-fn (constantly requests))
+        retry-delay-ms (if (pos? retries)
+                         ;; additionally divide the retry delay by 2 since the
+                         ;; server needs time to compute the results
+                         (max 1 (quot timeout-ms (* 2 retries)))
+                         0)
```
matgis, line 318: This is a bit confusing. `retry-delay-ms` appears to be used as a timeout, not a delay. And shouldn't we *double* the timeout duration if the server needs more time to compute the results? 🤔 Maybe I don't understand completely what is going on, and the comment just needs to be clarified?

editor/src/clj/editor/lsp.clj
```diff
@@ -205,33 +205,117 @@
...
+        server->requests (->> state
+                              :server->server-state
+                              (eduction
+                                (mapcat
```
matgis: Isn't this just `e/mapcat`?

editor/src/clj/editor/lsp.clj
```diff
@@ -205,33 +205,117 @@
...
+                                  (fn [[{:keys [languages] :as server} {:keys [status capabilities] :as server-state}]]
+                                    (when (and (= :running status)
+                                               (capabilities-pred capabilities)
+                                               (or (nil? language) (contains? languages language)))
+                                      (eduction
+                                        (map #(pair server %))
```
matgis: `e/map`?

editor/src/clj/editor/lsp.clj
```diff
@@ -338,6 +422,43 @@
...
+    (a/go (>! in (set-view-document-symbols resource (<! (a/into [] responses-ch)))))
+    (send-requests!
+      state responses-ch
+      :capabilities-pred :document-symbol
+      :language (resource/language resource)
+      :requests [(lsp.server/document-symbols resource)]
+      :timeout-ms 3000)))
+
+(defn- schedule-debounce [state id timeout-ms f]
+  {:pre [(some? id) (pos-int? timeout-ms) (ifn? f)]}
```
matgis: Style: Please don't use single-letter symbols. Maybe `state-fn`?

editor/src/clj/editor/outline_view.clj
```diff
@@ -739,14 +734,14 @@
...
-  (succeeding-selection [_this evaluation-context]
-    (g/node-value outline-view :succeeding-tree-selection evaluation-context))
-  (alt-selection [_this evaluation-context]
-    (g/node-value outline-view :alt-tree-selection evaluation-context)))
+  (selection [_this _evaluation-context]
+    (ui/selection tree-view))
+  (succeeding-selection [_this _evaluation-context]
+    (ui/succeeding-selection tree-view (ui/selection-root-items tree-view :node-id)))
+  (alt-selection [_this _evaluation-context]
+    (alt-selection (ui/selection tree-view))))
```
matgis, line 744: These were previously `:cached` outputs. Are they expensive to produce?

editor/src/clj/editor/outline_view.clj
```diff
@@ -787,15 +787,19 @@
...
-(defn make-outline-view [view-graph project tree-view app-view localization]
-  (let [outline-view (first
+      (ui/context! :outline {:outline-view outline-view} (->SelectionProvider tree-view) {} {Long :node-id
+                                                                                             resource/Resource :link}))))
+
+(defn make-outline-view [view-graph project app-view localization]
+  (let [tree-view (doto (ExtendedTreeView.)
+                    (.setId "outline")
+                    (.setPrefWidth 269.0)
+                    (.setPrefHeight 326.0))
```
matgis, line 797: Why do we set the dimensions here? Isn't this in an `AnchorPane` or something anyway?

editor/src/clj/editor/ui.clj
```diff
@@ -1267,7 +1271,7 @@
   [^TreeView tree-view opts]
   (.addEventFilter tree-view KeyEvent/KEY_PRESSED custom-tree-view-key-pressed-event-filter)
   (.addEventFilter tree-view MouseEvent/MOUSE_PRESSED custom-tree-view-mouse-pressed-event-filter)
-  (when-not (:double-click-expand? opts)
+  (when-not (:double-click-expand opts)
```
matgis, line 1274: We should update the docstring too!

editor/src/clj/editor/view.clj
```diff
@@ -15,17 +15,23 @@
 (ns editor.view
   (:require [dynamo.graph :as g]))
 
+(def default-sidebar-panes [:outline :properties])
+
```
matgis, line 19: There are already over 360 usages of the keyword `:properties` in various meanings. Let's use more specific keywords for these when we have the opportunity! Both `:properties-pane` and `:outline-pane` are unused and available!

matgis (dismissed):
Very nice! I think we should rename the `view/default-sidebar-panes` keywords and maybe fix the branch name before merging. Git history is forever after all!

editor/src/clj/editor/code/view.clj
vlaaad: Good to know, can definitely be unified in the future!

editor/src/clj/editor/code/view.clj
vlaaad: We already use `{:double-click-expand false}` for this tree view... But I agree, the behavior is a bit weird still. I will make double-click navigate only, and expansions will be handled by the arrow icon. I think this makes sense because for navigation, it's usually enough to have top-level functions.

editor/src/clj/editor/lsp.clj
vlaaad: I think there are no style issues in code that uses single-letter names, especially `f`, `n`, `x`, `m`, `a`/`b`/`c`... I'll change it here though, `state-fn` fits well here.

editor/src/clj/editor/lsp.clj
vlaaad: For all requests to LSP servers, we use a timeout, so that lagging LSP servers don't cause the editor to appear waiting for too long. E.g., for document symbols, we wait at most 3000ms. So if the server can't respond within 3 seconds, we report to UI that we got nothing. When we are talking about retries, these are retries **within** that 3-second budget. So, if we have 3 retries for 3 seconds total budget, we shouldn't wait in between retries for more than a second. A second is already too long, since the server also needs some additional time to compute the results. So we reduce the wait by 2, e.g., when we have a 3-second budget and 3 retries, we will retry every 500 ms, assuming that a server might need another 500ms or less to compute the results.

editor/src/clj/editor/outline_view.clj
vlaaad: They were cached outputs that depended on the `tree-view` output that used to perform a cljfx update to produce results, forcing a refresh of the view if somehow it wasn't up to date. But now it's impossible to do in the graph since the tree view is managed in the app-view ns. They are pretty cheap to produce, though.

editor/src/clj/editor/outline_view.clj
vlaaad: To preserve behavior. Previously, the same thing was done in `editor.fxml` (now deleted):
```
<TreeView id="outline" prefHeight="326.0" prefWidth="269.0" AnchorPane.bottomAnchor="0.0" AnchorPane.leftAnchor="0.0" AnchorPane.rightAnchor="0.0" AnchorPane.topAnchor="0.0" />
```

editor/src/clj/editor/lsp.clj
vlaaad: I think this code was written before the introduction of `util.eduction` ns.

editor/src/clj/editor/lsp.clj
matgis: To me, every symbol name is an opportunity to convey information to the reader. Choosing a single-letter name like `f` communicates that this can be any function - it can take any arguments and return whatever it chooses. It's a good choice for an argument name for general-purpose higher-order functions like the ones in the `util.fn` module. But I see this choice in lots of places where argument is expected to be something particular.

I think subconsciously functional programming aficionados pick this up while learning the paradigm, because a lot of the material are around higher-order functions where it makes sense. But that context is lost somewhere along the way, and instead we start using `f` whenever something is a function, which isn't really helpful.

vlaaad:
BTW, git branch name is a mutable reference to a commit, it's not preserved in history in any way (unless its name is included in a commit, which does not seem to happen when we squash-merge PRs into dev)

editor/src/clj/editor/code/view.clj
```diff
@@ -1268,6 +1271,156 @@
...
+        width (* code-type-icon-size (double (/ width-multiplier 16)))
+        height (* code-type-icon-size (double (/ height-multiplier 16)))]
+    {:fx/type fx.region/lifecycle
+     :style {:-fx-background-color :-df-text-dark}
+     :min-width width
+     :max-width width
+     :min-height height
+     :max-height height
+     :shape {:fx/type fx.svg-path/lifecycle
+             :content (case type
```
matgis, line 1312: We should probably add a default case here so we won't throw an exception in case the language server returns a type we don't know about?

editor/src/clj/editor/lsp.clj
matgis: That's good info! Let's put in in the code comment, as I'm sure others will be as confused as I was!

PR #11957, opened 2026-02-27
Cache texture sampler state for OpenGL

### LINKED ISSUE #8163: Graphics renderer sets texture parameters two times per draw call
This is not a critical flaw and does not affect engine's functioning in any way.

Graphics renderer sets texture parameters two times per draw call if you use a texture sampler in a material. What it looks like in the case of OpenGL and `sprite.material`:

### built-in sprite.material:
image

### the same material, but with a custom texture sampler to set your own texture parameters:
image

You see, the renderer always resets the default texture parameters first, and then sets the ones specified in the material.

**To Reproduce (REQUIRED)**
- Add a texture sampler to a material.

Graphics renderer shouldn't set texture parameters two times per draw call.

  - Version: <1.5.0, 1.6.0, 1.6.1.

**Platforms (REQUIRED):**
 - Platforms: all

Jhonnyg:
We currently set the texture sampler state every time we bind a texture as well as when `dmGraphic::SetTextureParams` is called when textures are bound in the renderer. This has now been changed so that we cache the sampler settings and only apply these when the sampler state has changed. 

In this sample scene the number OpenGL calls is almost half and the fps diff is ~370 vs 340.

Left: Caching of OpenGL texture state (not applying filter if we don't have to)
Right: Current dev

texturesamplestate

Fixes #8163

### DIFF
--- a/engine/graphics/src/opengl/graphics_opengl.cpp
+++ b/engine/graphics/src/opengl/graphics_opengl.cpp
@@ -2196,6 +2196,84 @@ static void LogFrameBufferError(GLenum status)
                 }
             }
         }
+
+    #if !defined(GL_ES_VERSION_3_0) && defined(GL_ES_VERSION_2_0) && !defined(__EMSCRIPTEN__)  && !defined(ANDROID)
+        glEnable(GL_TEXTURE_2D);
+        CHECK_GL_ERROR;
+    #endif
+
+        for (int unit = 0; unit < DM_MAX_TEXTURE_UNITS; ++unit)
+        {
+            if (context->m_CurrentTextures[unit].m_Texture)
+            {
+                // Can we use the actual texture pointer instead of extracting this from the asset container every time?
+                DM_MUTEX_OPTIONAL_SCOPED_LOCK(context->m_AssetHandleContainerMutex);
+                OpenGLTexture* tex = GetAssetFromContainer<OpenGLTexture>(context->m_AssetHandleContainer, context->m_CurrentTextures[unit].m_Texture);
+                if (!tex)
+                {
+                    continue;
+                }
+
+                glActiveTexture(TEXTURE_UNIT_NAMES[unit]);
+                CHECK_GL_ERROR;
+
+                uint8_t id_index = context->m_CurrentTextures[unit].m_TextureIdIndex;
+                bool bind_as_texture = true;
+                if (tex->m_Type == TEXTURE_TYPE_IMAGE_2D || tex->m_Type == TEXTURE_TYPE_IMAGE_3D)
+                {
+                    bind_as_texture = !BindComputeImage(context, tex, unit, id_index);
+                }
+
+                if (bind_as_texture)
+                {
+                    glBindTexture(GetOpenGLTextureType(tex->m_Type), GetGLHandle(context, tex->m_TextureIds[id_index]));
+                    CHECK_GL_ERROR;
+
+                    GLenum gl_type = GetOpenGLTextureType(tex->m_Type);
+
+                    if (tex->m_Sampler.m_MinFilter != tex->m_SamplerDirty.m_MinFilter)
+                    {
+                        GLenum gl_min_filter = GetOpenGLTextureFilter(minfilter == TEXTURE_FILTER_DEFAULT ? context->m_DefaultTextureMinFilter : minfilter);
+
+                        // Using a mipmapped min filter without any mipmaps will break the sampler
+                        if (tex->m_MipMapCount <= 1)
+                        {
+                            gl_min_filter = GetNonMipMapVersionOfFilter(gl_min_filter);
+                        }
+
+                        glTexParameteri(gl_type, GL_TEXTURE_MIN_FILTER, gl_min_filter);
+                        CHECK_GL_ERROR;
+                    }
+
+                    if (tex->m_Sampler.m_MagFilter != tex->m_SamplerDirty.m_MagFilter)
+                    {
+                        GLenum gl_mag_filter = GetOpenGLTextureFilter(magfilter == TEXTURE_FILTER_DEFAULT ? context->m_DefaultTextureMagFilter : magfilter);
+                        glTexParameteri(gl_type, GL_TEXTURE_MAG_FILTER, gl_mag_filter);
+                        CHECK_GL_ERROR;
+                    }
+
+                    if (tex->m_Sampler.m_AddressModeU != tex->m_SamplerDirty.m_AddressModeU)
+                    {
+                        glTexParameteri(gl_type, GL_TEXTURE_WRAP_S, GetOpenGLTextureWrap(uwrap));
+                        CHECK_GL_ERROR
+                    }
+
+                    if (tex->m_Sampler.m_AddressModeV != tex->m_SamplerDirty.m_AddressModeV)
+                    {
+                        glTexParameteri(gl_type, GL_TEXTURE_WRAP_T, GetOpenGLTextureWrap(vwrap));
+                        CHECK_GL_ERROR
+                    }
+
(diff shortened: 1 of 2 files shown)


engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -1545,7 +1551,7 @@ static void LogFrameBufferError(GLenum status)
             }
         }
 
-        if (glGenVertexArrays)
+        if (&glGenVertexArrays)
```
Jhonnyg: This was always true before.

engine/graphics/src/opengl/graphics_opengl_private.h
```diff
@@ -41,11 +41,23 @@ namespace dmGraphics
         DEVICE_BUFFER_TYPE_VERTEX  = 1,
     };
 
+    // TODO: Merge with vulkan, dx12, webgpu!
```
Jhonnyg, line 44: Not in this PR however! But these settings are identical to the other adapters, with the exception that dx12,vulkan,webgpu also stores a sampler object that can be bound. It should work similar to what we have for programs:

```
struct Sampler
{
    // shared settings..
};

struct VulkanSampler
{
    Sampler m_BaseSampler;
    VkSampler m_VulkanSampler;
};
...
```

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2196,6 +2202,99 @@ static void LogFrameBufferError(GLenum status)
                 }
             }
         }
+
+    #if !defined(GL_ES_VERSION_3_0) && defined(GL_ES_VERSION_2_0) && !defined(__EMSCRIPTEN__)  && !defined(ANDROID)
+        glEnable(GL_TEXTURE_2D);
+        CHECK_GL_ERROR;
+    #endif
```
Jhonnyg: No sure if this is applicable any longer? The only platforms we have es2 support for is webgl and android.

engine/graphics/src/opengl/graphics_opengl.cpp
JCash: This looks strange?

Previously, the `glGenVertexArrays` would be a function pointer, that could be 0.
While that may have changed, I'm not sure taking address would be better, will this not be always true as well?

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2144,6 +2144,263 @@ static void LogFrameBufferError(GLenum status)
...
+    {
+        const GLenum texture_filter_lut[] = {
+            0,
+            GL_NEAREST,
+            GL_LINEAR,
+            GL_NEAREST_MIPMAP_NEAREST,
+            GL_NEAREST_MIPMAP_LINEAR,
+            GL_LINEAR_MIPMAP_NEAREST,
+            GL_LINEAR_MIPMAP_LINEAR,
+        };
```
JCash: If TextureFilter changes, we'd have to make sure all backend' lists are updated, like this place.
Would it make more sense to use a switch-statement, that default-asserts if the enum isn't specified?

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2196,6 +2453,99 @@ static void LogFrameBufferError(GLenum status)
...
+
+                        GLenum gl_mag_filter = GetOpenGLTextureFilter(magfilter);
+                        glTexParameteri(gl_type, GL_TEXTURE_MAG_FILTER, gl_mag_filter);
+                        CHECK_GL_ERROR;
+                    }
+
+                    if (tex->m_Sampler.m_AddressModeU != tex->m_SamplerDirty.m_AddressModeU)
+                    {
+                        glTexParameteri(gl_type, GL_TEXTURE_WRAP_S, GetOpenGLTextureWrap(tex->m_SamplerDirty.m_AddressModeU));
+                        CHECK_GL_ERROR
```
JCash: missing `;` after statement

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2196,6 +2453,99 @@ static void LogFrameBufferError(GLenum status)
...
+                    }
+
+                    if (tex->m_Sampler.m_AddressModeV != tex->m_SamplerDirty.m_AddressModeV)
+                    {
+                        glTexParameteri(gl_type, GL_TEXTURE_WRAP_T, GetOpenGLTextureWrap(tex->m_SamplerDirty.m_AddressModeV));
+                        CHECK_GL_ERROR
```
JCash: missing `;` after statement

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2196,6 +2453,99 @@ static void LogFrameBufferError(GLenum status)
...
+                    }
+
+                    if (context->m_AnisotropySupport && tex->m_Sampler.m_MaxAnisotropy != tex->m_SamplerDirty.m_MaxAnisotropy)
+                    {
+                        float max_anisotropy = tex->m_SamplerDirty.m_MaxAnisotropy;
+                        glTexParameterf(gl_type, GL_TEXTURE_MAX_ANISOTROPY_EXT, dmMath::Min(max_anisotropy, context->m_MaxAnisotropy));
+                        CHECK_GL_ERROR
```
JCash: missing `;` after statement

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -2196,6 +2453,99 @@ static void LogFrameBufferError(GLenum status)
...
+                    if (tex->m_Sampler.m_AddressModeV != tex->m_SamplerDirty.m_AddressModeV)
+                    {
+                        glTexParameteri(gl_type, GL_TEXTURE_WRAP_T, GetOpenGLTextureWrap(tex->m_SamplerDirty.m_AddressModeV));
+                        CHECK_GL_ERROR
+                    }
+
+                    if (context->m_AnisotropySupport && tex->m_Sampler.m_MaxAnisotropy != tex->m_SamplerDirty.m_MaxAnisotropy)
+                    {
+                        float max_anisotropy = tex->m_SamplerDirty.m_MaxAnisotropy;
+                        glTexParameterf(gl_type, GL_TEXTURE_MAX_ANISOTROPY_EXT, dmMath::Min(max_anisotropy, context->m_MaxAnisotropy));
```
JCash: Imho, I think the dmMath::Min() should be applied earlier, when the value is actually set?

JCash (changes requested):
Overall good.
Some smaller changes, but also some questions regarding the actual if-statement

engine/graphics/src/opengl/graphics_opengl.cpp
Jhonnyg: Hm maybe it's platform dependant, because the original code is giving us warnings about it, this is from arm64-macos on CI:

2026-03-01T23:24:34.2569610Z ../src/opengl/graphics_opengl.cpp:1548:13: warning: address of function 'glGenVertexArrays' will always evaluate to 'true' [-Wpointer-bool-conversion]
2026-03-01T23:24:34.2570260Z  1548 |         if (glGenVertexArrays)
2026-03-01T23:24:34.2570600Z       |         ~~  ^~~~~~~~~~~~~~~~~
2026-03-01T23:24:34.2571180Z ../src/opengl/graphics_opengl.cpp:1548:13: note: prefix with the address-of operator to silence this warning
2026-03-01T23:24:34.2571720Z  1548 |         if (glGenVertexArrays)
2026-03-01T23:24:34.2572040Z       |             ^
2026-03-01T23:24:34.2572340Z       |             &

JCash (approved):

aglitchman:
🥰🥰🥰🥰🥰

AGulev:
@Jhonnyg this PR isn't in release notes

PR #11958, opened 2026-02-27
Fix strange sprite behavior when change atlas in runtime

### LINKED ISSUE #11930: The textures are behaving very strangely after resource.set_atlas. Beta, Alpha versions
The problem occurs after replacing sd textures with hd textures using resource.set_atlas(sd, hd_atlas)

All Defold 1.12+ versions:
Image

Defold 1.11.2 (Correct behavior):
Image

**To Reproduce**
1. Find file build_bat/html5_zip_playgama.sh
2. Change path_to_bundle (output dir) and bob variables
3. Run script and build the project. The build will happen for the html5 platform. Liveupdate resources will be included in the build
4. Launch the build using the web server. You will see that the sd textures look fine. As soon as they are replaced by the HD version, they begin to flip over and change positions

**Platforms:**
All

https://drive.google.com/file/d/1CVzgoZqLmDC8Ay6CKVDWbKPtu3rDAD5J/view?usp=sharing

**Additional context:**
This may have happened after you optimized texture loading.

Files and scripts that you may need:
collections/loader/load_hd.script (loads liveupdate)
collections/hd/hd.script (replaces atlas with another one by resource.set_atlas(sd, hd_atlas))

ekharkunov:
Fix sprite strange behavior when sprite atlas was changed in runtime.

Fixes #11930 

### Technical changes
Reason:
Because pointer to `TextureSetResource` isn't changed when user updates atlas in runtime no sprite rehashing triggers. As a result after texture update sprite component used cached data about geometry, etc.
Fix:
Added resource reload callback to react to `TextureSetResource` changes. Also added `m_TexturesGeneration` field to `TextureSetResource` and use it to calculate hash to keep track that underlying texture resource was updated.

## PR checklist

* [x] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/engine/gamesys/src/dmsdk/gamesys/resources/res_sprite.h
+++ b/engine/gamesys/src/dmsdk/gamesys/resources/res_sprite.h
@@ -30,7 +30,6 @@ namespace dmGameSystem
     {
         dmhash_t            m_SamplerNameHash;
         TextureSetResource* m_TextureSet;
-        uint8_t             m_TexturesGeneration;
     };
 
     struct SpriteResource
--- a/engine/gamesys/src/dmsdk/gamesys/resources/res_sprite.h
+++ b/engine/gamesys/src/dmsdk/gamesys/resources/res_sprite.h
@@ -30,13 +30,14 @@ namespace dmGameSystem
     {
         dmhash_t            m_SamplerNameHash;
         TextureSetResource* m_TextureSet;
+        uint8_t             m_TexturesGeneration;
     };
 
     struct SpriteResource
     {
         dmGameSystemDDF::SpriteDesc* m_DDF;
-        MaterialResource*           m_Material;
-        dmhash_t                    m_DefaultAnimation;
+        MaterialResource*            m_Material;
+        dmhash_t                     m_DefaultAnimation;
 
         // Sorted by the order of occurrance of samplers in the .material file
         SpriteTexture* m_Textures;
--- a/engine/gamesys/src/gamesys/components/comp_sprite.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_sprite.cpp
@@ -19,7 +19,6 @@
 #include <algorithm>
 
 #include <dlib/array.h>
-#include <dlib/set.h>
 #include <dlib/hash.h>
 #include <dlib/double_linked_list.h>
 #include <dlib/log.h>
@@ -484,17 +483,6 @@ namespace dmGameSystem
         return texture ? texture->m_TextureSet : 0;
     }
 
-    static SpriteTexture* GetSpriteTextureByIndex(const SpriteComponent* component, uint32_t index)
-    {
-        SpriteResourceOverrides* overrides = component->m_Overrides;
-        SpriteTexture* texture = 0x0;
-        if (overrides && index < overrides->m_Textures.Size())
-            texture = &overrides->m_Textures[index];
-        if (!texture || !texture->m_TextureSet)
-            texture = &component->m_Resource->m_Textures[index];
-        return texture;
-    }
-
     static inline TextureSetResource* GetTextureSetByHash(const SpriteComponent* component, dmhash_t sampler_name_hash)
     {
         if (component->m_Overrides)
@@ -535,6 +523,20 @@ namespace dmGameSystem
                 return overrides->m_Textures[texture_unit].m_TextureSet->m_Texture;
         }
         return component->m_Resource->m_Textures[texture_unit].m_TextureSet->m_Texture;
+    }
+
+    uint8_t GetTextureResourceGeneration(const SpriteComponent* component, uint32_t texture_unit)
+    {
+        if(texture_unit >= component->m_Resource->m_NumTextures)
+            return 0;
+
+        const SpriteResourceOverrides* overrides = component->m_Overrides;
+        if (overrides && texture_unit < overrides->m_Textures.Size())
+        {
+            if (overrides->m_Textures[texture_unit].m_TextureSet)
+                return overrides->m_Textures[texture_unit].m_TextureSet->m_TexturesGeneration;
+        }
+        return component->m_Resource->m_Textures[texture_unit].m_TextureSet->m_TexturesGeneration;
     }
 
     dmGraphics::HTexture GetMaterialTexture(const SpriteComponent* component, uint32_t texture_unit)
@@ -649,6 +651,12 @@ namespace dmGameSystem
         }
 
         dmHashUpdateBuffer32(&state, resource->m_Textures, sizeof(SpriteTexture) * resource->m_NumTextures);
+        for (size_t idx = 0; idx < resource->m_NumTextures; ++idx)
+        {
+            uint8_t generation = GetTextureResourceGeneration(component, idx);
+            dmHashUpdateBuffer32(&state, &generation, sizeof(generation));
+        }
+        dmHashUpdateBuffer32(&state, resource->m_Textures->m_TextureSet, sizeof(resource->m_Textures->m_TextureSet));
         dmHashUpdateBuffer32(&state, resource->m_Material, sizeof(MaterialResource*));
 
         HashResourceOverrides(&state, component->m_Overrides);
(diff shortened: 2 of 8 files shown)


engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -2656,6 +2673,43 @@ namespace dmGameSystem
...
+                        SpriteTexture* sprite_texture = GetSpriteTextureByIndex(component, texture_idx);
+                        if (!updated_sprite_textures.Contains(sprite_texture))
+                        {
+                            sprite_texture->m_TexturesGeneration++;
+                            if (updated_sprite_textures.Full())
+                            {
+                                updated_sprite_textures.OffsetCapacity(32);
+                            }
+                            updated_sprite_textures.Add(sprite_texture);
+                        }
```
JCash: The `m_ReHash=1` part is clear to me, but I don't understand the part of `m_TexturesGeneration++`?
`m_TexturesGeneration` isn't used anywhere?

JCash (changes requested):
Overall good, but the texture generation counter seems redundant

engine/gamesys/src/gamesys/components/comp_sprite.cpp
ekharkunov: `m_ReHash` isn't enough to fix initial issue because even if we force hash recalculation - the result hash will be the same as it was before atlas change. And as result old geometry information will be used next frame.
`m_TexturesGeneration` is used to change result here https://github.com/defold/defold/blob/dev/engine/gamesys/src/gamesys/components/comp_sprite.cpp#L635. Counter help to track change of underlying `TextureSetResource` without going depp inside structure and without changing hash calculation logic.

JCash: Ah, we hash the entire struct as-is, that's why I didn't find a reference to m_TexturesGeneration.

engine/gamesys/src/dmsdk/gamesys/resources/res_sprite.h
```diff
@@ -30,13 +30,14 @@ namespace dmGameSystem
     {
         dmhash_t            m_SamplerNameHash;
         TextureSetResource* m_TextureSet;
+        uint8_t             m_TexturesGeneration;
```
JCash: I think I'd rather have the "generation" on the actual resource that changed, the TextureSetResource*.
This of course would means that the sprite's Rehash() function would change a little.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -2656,6 +2673,43 @@ namespace dmGameSystem
         pit->m_FnIterateNext = CompSpriteIterPropertiesGetNext;
     }
 
+    static void ResourceReloadedCallback(const dmResource::ResourceReloadedParams* params)
+    {
```
JCash, line 2685: You can use `dmhash_t dmResource::ResourceTypeGetNameHash(params.m_Type);` to compare compare and early out if it's not a `texturesetc` resource.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
JCash: If the generation is moved to the TextureSetResource, then this logic isn't needed.

JCash (changes requested):
Thanks for clarifying.
I do think that the generation counter belongs with the texture set resource, as it's universal, and not unique to the sprite usage.

engine/gamesys/src/gamesys/components/comp_sprite.cpp
ekharkunov: As I understand reload callback called only for requested resource type so no need to filter it additionally.
https://github.com/defold/defold/blob/dev/engine/resource/src/resource.cpp#L1267-L1289

engine/gamesys/src/gamesys/resources/res_textureset.cpp
```diff
@@ -126,14 +126,16 @@ namespace dmGameSystem
...
 
     static uint32_t GetResourceSize(TextureSetResource* res, uint32_t ddf_size)
     {
         uint32_t size = sizeof(TextureSetResource);
         size += ddf_size;
-        size += res->m_HullCollisionGroups.Capacity()*sizeof(dmhash_t);
-        size += res->m_AnimationIds.Capacity()*(sizeof(dmhash_t)+sizeof(uint32_t));
+        size += res->m_HullCollisionGroups.Capacity() * sizeof(dmhash_t);
+        size += res->m_AnimationIds.Capacity() * (sizeof(dmhash_t) + sizeof(uint32_t));
+        size += res->m_FrameIds.Capacity() * (sizeof(dmhash_t) + sizeof(uint32_t));
```
ekharkunov, line 138: I assumed that it was missed.

engine/gamesys/src/gamesys/resources/res_textureset.cpp
```diff
@@ -194,14 +196,17 @@ namespace dmGameSystem
...
         {
+            uint8_t current_generation = tile_set->m_TexturesGeneration;
             ReleaseResources(params->m_Factory, tile_set);
 
             tile_set->m_TextureSet = tmp_tile_set.m_TextureSet;
             tile_set->m_Texture = tmp_tile_set.m_Texture;
             tile_set->m_HullCollisionGroups.Swap(tmp_tile_set.m_HullCollisionGroups);
             tile_set->m_HullSet = tmp_tile_set.m_HullSet;
             tile_set->m_AnimationIds.Swap(tmp_tile_set.m_AnimationIds);
+            tile_set->m_FrameIds.Swap(tmp_tile_set.m_FrameIds);
```
ekharkunov, line 207: I assumed that we should also swap `FrameIds`. Let me know if it was done by design

engine/gamesys/src/gamesys/components/comp_sprite.cpp
JCash: You call the rgister function above:
```c++
dmResource::RegisterResourceReloadedCallback(sprite_context->m_Factory, ResourceReloadedCallback, sprite_world);
```

There is no additional information there, it gets called for all reloaded resources.
E.g. if someone reloads or sets a sound file?

engine/gamesys/src/gamesys/resources/res_textureset.cpp
JCash: this looks like a good improvement. Was this the final issue?

JCash (changes requested):
overall good
see note about resource reload check

engine/gamesys/src/gamesys/resources/res_textureset.cpp
ekharkunov: No, it doesn't relate to issue with wrong atlas animation. Just notice and add it.

engine/resource/src/resource.cpp
```diff
@@ -1348,6 +1349,7 @@ Result SetResource(HFactory factory, uint64_t hashed_name, void* message)
                 params.m_Resource = rd;
                 params.m_Filename = 0;
                 params.m_FilenameHash = hashed_name;
+                params.m_Type = resource_type;
```
ekharkunov, line 1352: One more missed thing

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -2683,6 +2683,11 @@ namespace dmGameSystem
 
     static void ResourceReloadedCallback(const dmResource::ResourceReloadedParams* params)
     {
+        dmhash_t name_hash = ResourceTypeGetNameHash(params->m_Type);
+        if (name_hash != TEXTURE_SET_EXT_HASH)
+        {
+            return;
+        }
```
JCash, line 2690: 👍

JCash (approved):
lgtm!

PR #11963, opened 2026-03-02
Re-create controls on selection change

### LINKED ISSUE #11948: Wrong values shown for the component's properties in the editor.
There is a special scenario in which the editor shows wrong values for the component's properties.

**To Reproduce (REQUIRED)**
1. Create a collection with a GO and two collisionobject components. It's important that the components are identical(same values for the properties).
2. Edit the numerical value of one of the properties of the 1st component but don't confirm the value(don't enter and don't click another text box).
3. Click on the other component.
4. See that the clicked component doesn't show the initial value for the property but the value entered for the modified component.

The editor should always show the correct value(the value from *.collection file).

  - Version 1.12.1 , Editor SHA1: 16c6fd602f32de4814660672c38ce3ccbbc1fb59

**Platforms (REQUIRED):**
 - Platforms: Windows
 - OS: Windows 10
 - Device: PC

**Logs (OPTIONAL):**

**Workaround (OPTIONAL):**

**Screenshots (OPTIONAL):**

Image showing the initial values of the component's properties:
Image

Image showing modification of one of the component's properties and then the values shown for the other components (that were not modified):
Image

Image showing that after clicking again on the modified component, the values are shown correctly:
Image

vlaaad:
Fix #11948

### DIFF
--- a/editor/src/clj/editor/properties_view.clj
+++ b/editor/src/clj/editor/properties_view.clj
@@ -708,8 +708,8 @@
      (->> properties
           category-property-edit-types
           (e/mapcat
-            (fn [[category-title properties]]
-              (when-not (coll/empty? properties)
+            (fn [[category-title category-properties]]
+              (when-not (coll/empty? category-properties)
                 (-> []
                     (cond->
                       category-title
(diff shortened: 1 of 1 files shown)


JosephFerano (commented):
The fix isn't working for me.

https://github.com/user-attachments/assets/697de1b5-b610-4273-8c05-16c3473f5d6b

vlaaad:
Huh, you are right! I swear it worked... what the..

vlaaad:
@JosephFerano, there was a shadowing issue I missed, try again :)

JosephFerano (approved):
It works!

PR #11965, opened 2026-03-02
Incorrect attribute values for instancing

### LINKED ISSUE #9722: Switching the material loses instanced vertex attribute values
I started using instanced rendering for optimisation. Also in game I have all models rendered to depth buffer for shadow mapping. It turned out that on the second call to `render.draw` instanced models do not get the correct vertex attributes values.

On the first `render.draw` call, the instanced vertex attributes are present:
image
image

On the second `render.draw` call, some instanced vertex attributes are mssing (i.e. the second instance value is `1.0`, but should be `30.0`):
image
image

And it happens in rare cases, so I was able to repeat the result only on models from the working project. I apologise for the ugly picture!

**To Reproduce (REQUIRED)**
1. Add a new material to render script
2. In render script: enable the material, draw instanced models, disable the material, draw instanced models.
3. The instanced values are lost on the second render call.

On the screen we should see 4 models, i.e. I have coloured in red the place where there should be a second mesh (with `piece_num = 30.0`).

Radeon 6600 XT:
image

On Geforce RTX 3090 we see only the result of the first call, ie on the second call instanced values are absent (I have not checked what they are exactly, it is clear that this is a bug):
image

  - Version: 1.9.4, 1.9.5 BETA

**Platforms (REQUIRED):**
 - Platforms: Windows 10
 - Device: Radeon RX 6600 XT, Geforce RTX 3090

[issue_gpu_instancing.zip](https://github.com/user-attachments/files/17709173/issue_gpu_instancing.zip)

Jhonnyg:
Using materials with custom vertex formats for models together with overriding materials in a render script can now render correctly.

Fixes #9722


JCash (approved):
lgtm

PR #11968, opened 2026-03-02
Support new Protobuf features in the editor

matgis:
This PR adds support for missing Protobuf features to the editor. We now support `map<TKey, TValue>` fields and self-referencing message types. Both of these are required for Json-like messages where objects and arrays of objects can be nested inside each other. It also fixes a few uncovered corner cases around `oneof` fields and adds a whole host of tests.

Required for #11724.

### Technical changes
* Protobuf `field-info` no longer specifies a `:field-rule` as `:required`, `:optional`, or `:repeated`. Instead, a we have a `:field-kind` that is considered to be either `required`, `optional`, `list`, or `map`. This was done to distinguish between `list` and `map` fields, which need to be handled differently, but are both `repeated` fields internally. The new values are namespaced keywords under `:pb-field-kind/*`.
* The old method of extracting resource field path has been removed. Previously you would obtain a `resource-field-path-fn` from a Protobuf `Message` subclass that could apply to Clojure map representations of those messages. This is no longer possible when we have self-referencing types. Instead, you now supply the `pb-class` and Clojure map directly to the `protobuf/resource-field-value-paths` function.
* `protobuf/default-value` was renamed to `optional-field-defaults` for clarity.
* `protobuf/required-default-value` was renamed to `required-field-defaults` for clarity.
* `protobuf/default-message` was made private. Callers should call either `optional-field-defaults` or `required-field-defaults`, depending on their needs.
* Added `memoize-late-bound` function to `fn` module.

### DIFF
--- a/editor/src/clj/editor/protobuf.clj
+++ b/editor/src/clj/editor/protobuf.clj
@@ -30,7 +30,8 @@ Macros currently mean no foreseeable performance gain, however."
             [util.digest :as digest]
             [util.fn :as fn]
             [util.text-util :as text-util])
-  (:import [com.dynamo.proto DdfExtensions DdfMath$Matrix4 DdfMath$Point3 DdfMath$Quat DdfMath$Vector3 DdfMath$Vector3One DdfMath$Vector4 DdfMath$Vector4One DdfMath$Vector4WOne]
+  (:import [clojure.lang IPending]
+           [com.dynamo.proto DdfExtensions DdfMath$Matrix4 DdfMath$Point3 DdfMath$Quat DdfMath$Vector3 DdfMath$Vector3One DdfMath$Vector4 DdfMath$Vector4One DdfMath$Vector4WOne]
            [com.google.protobuf DescriptorProtos$FieldOptions Descriptors$Descriptor Descriptors$EnumDescriptor Descriptors$EnumValueDescriptor Descriptors$FieldDescriptor Descriptors$FieldDescriptor$JavaType Descriptors$FieldDescriptor$Type Descriptors$FileDescriptor Message Message$Builder ProtocolMessageEnum TextFormat]
            [java.io ByteArrayOutputStream StringReader]
            [java.lang.reflect Method]
@@ -61,6 +62,44 @@ Macros currently mean no foreseeable performance gain, however."
 (defonce/protocol PbConverter
   (msg->vecmath [^Message pb v] "Return the javax.vecmath equivalent for the Protocol Buffer message")
   (msg->clj [^Message pb v]))
+
+(defn- as-late-bound-fn [fn-or-promise]
+  (cond
+    (fn? fn-or-promise)
+    fn-or-promise
+
+    (instance? IPending fn-or-promise)
+    (fn late-fn [& args]
+      (let [fn (deref fn-or-promise)]
+        (apply fn args)))))
+
+(defn- memoize-late-bound [higher-order-fn]
+  (let [cache-atom (atom {})]
+    (fn memoized-higher-order-fn [& args]
+      (let [cache-key (vec args)
+            cache-value (get @cache-atom cache-key)]
+        (or (as-late-bound-fn cache-value)
+            (let [late-fn-promise (promise)
+
+                  cache-value
+                  (-> cache-atom
+                      (swap! update cache-key fn/or late-fn-promise)
+                      (get cache-key))]
+
+              (if (identical? late-fn-promise cache-value)
+                ;; We just inserted our promise into the cache, so we're
+                ;; responsible for delivering the result.
+                (let [late-fn (apply higher-order-fn args)]
+                  (assert (ifn? late-fn) "The higher-order-fn must return a function.")
+                  (deliver late-fn-promise late-fn)
+                  (swap! cache-atom assoc cache-key late-fn)
+                  late-fn)
+
+                ;; Found an already-existing promise or function in the cache.
+                ;; Return the function unaltered, or a function that
+                ;; dereferences the promise when called.
+                (or (as-late-bound-fn cache-value)
+                    (assert false "Unexpected value found in cache.")))))))))
 
 (def ^:private upper-pattern (re-pattern #"\p{javaUpperCase}"))
 
@@ -712,7 +751,7 @@ Macros currently mean no foreseeable performance gain, however."
                                (let [field-pb-value (.invoke field-get-method pb java/no-args-array)]
                                  (pb-value->clj field-pb-value))))))))))))))
 
-(def ^:private pb->clj-fn (fn/memoize pb->clj-fn-raw))
+(def ^:private pb->clj-fn (memoize-late-bound pb->clj-fn-raw))
 
 (def ^:private pb->clj-with-defaults-fn (fn/memoize #(pb->clj-fn % #{:pb-field-kind/optional :pb-field-kind/required})))
 
--- a/editor/src/clj/editor/protobuf.clj
+++ b/editor/src/clj/editor/protobuf.clj
@@ -95,6 +95,15 @@ Macros currently mean no foreseeable performance gain, however."
     Descriptors$FieldDescriptor$JavaType/ENUM (desc->pb-class (.getEnumType field-desc))
     Descriptors$FieldDescriptor$JavaType/MESSAGE (desc->pb-class (.getMessageType field-desc))))
 
+(defn pb-field-desc-field-kind
+  [^Descriptors$FieldDescriptor field-desc]
+  (cond
+    (.isMapField field-desc) :pb-field-kind/map
+    (.isRepeated field-desc) :pb-field-kind/list
+    (.isOptional field-desc) :pb-field-kind/optional
+    (.isRequired field-desc) :pb-field-kind/required
+    :else (assert false)))
+
 (defn- new-builder
   ^Message$Builder [^Class cls]
   (java/invoke-no-arg-class-method cls "newBuilder"))
@@ -336,14 +345,7 @@ Macros currently mean no foreseeable performance gain, however."
     (into {}
           (map (fn [^Descriptors$FieldDescriptor field-desc]
                  (let [field-key (field->key field-desc)
-
-                       field-kind
-                       (cond
-                         (.isMapField field-desc) :pb-field-kind/map
-                         (.isRepeated field-desc) :pb-field-kind/list
-                         (.isOptional field-desc) :pb-field-kind/optional
-                         (.isRequired field-desc) :pb-field-kind/required
-                         :else (assert false))
+                       field-kind (pb-field-desc-field-kind field-desc)
 
                        [key-info value-info]
                        (if (.isMapField field-desc)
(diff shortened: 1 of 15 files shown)


editor/src/clj/editor/protobuf.clj
```diff
@@ -223,47 +271,65 @@ Macros currently mean no foreseeable performance gain, however."
...
+        (case field-kind
+          :pb-field-kind/list
+          (str "get" java-name "List")
+
+          :pb-field-kind/map
+          (str "get" java-name "Map")
+
+          (:pb-field-kind/optional :pb-field-kind/required)
+          (str "get" java-name))]
+    (lookup-method pb-class get-method-name)))
```
matgis, line 317: Before this, we were being a bit clever about trying to share code paths between the various field kinds. It made it quite difficult to reason about the differences between them when implementing support for `map` fields.

As part of this change, I've split the code paths apart so it is more explicit what is going on. Thus, we might be repeating ourselves in a few places, but in exchange it is now clearer what is going on and it's a lot easier to fix issues around differences in behavior.

editor/src/clj/editor/protobuf.clj
```diff
@@ -279,214 +345,153 @@ Macros currently mean no foreseeable performance gain, however."
...
+                         {:field-kind field-kind
+                          :is-oneof-field (some? (.getRealContainingOneof field-desc))
                           :java-name (underscores-to-camel-case (.getName field-desc))
-                          :field-type-key field-type-key
-                          :field-rule field-rule
-                          :default default
-                          :declared-default declared-default
-                          :options (options (.getOptions field-desc))}))))
+                          :options (options (.getOptions field-desc))})]
+                   (pair field-key field-info))))
```
matgis, line 375: We now have information for both the `key` and the `value`. Non-map fields only have the value information. I've removed the confusing `:default` and `:declared-default` entries. `:value-default` is now the `:declared-default` for values so it is among the `:value-*` prefixed keys. There is no corresponding `:key-default` because you can't declare a default for a `map` field.

editor/src/clj/editor/protobuf.clj
```diff
@@ -279,214 +345,153 @@ Macros currently mean no foreseeable performance gain, however."
...
 (defn resource-field? [field-info]
-  (and (= :string (:field-type-key field-info))
-       (true? (:resource (:options field-info)))))
+  (case (:field-kind field-info)
+    (:pb-field-kind/list :pb-field-kind/optional :pb-field-kind/required)
+    (and (= :string (:value-type-kw field-info))
+         (true? (:resource (:options field-info))))
+
+    (:pb-field-kind/map)
+    false))
```
matgis, line 387: Likewise, a `map` field cannot currently be annotated as a `resource` field, because it's unclear if it refers to the `key` or the `value`.

editor/src/clj/editor/protobuf.clj
```diff
@@ -279,214 +345,153 @@ Macros currently mean no foreseeable performance gain, however."
...
 
 (defn message-field? [field-info]
-  (= :message (:field-type-key field-info)))
+  (case (:field-kind field-info)
+    (:pb-field-kind/list :pb-field-kind/optional :pb-field-kind/required)
+    (= :message (:value-type-kw field-info))
+
+    (:pb-field-kind/map)
+    false))
```
matgis, line 395: Technically `map` fields are repeated message fields of `KeyValuePair`-like messages, but that is not what you're asking about with this function.

editor/src/clj/editor/protobuf.clj
```diff
@@ -279,214 +345,153 @@ Macros currently mean no foreseeable performance gain, however."
...
-  (resource-field-path-specs-raw ResourceDefaulted) => [ [{:image '/default.png'}] ]
-  (resource-field-path-specs-raw ResourceRepeated)  => [ [[:images]] ]
-
-  (resource-field-path-specs-raw ResourceSimpleNested)    => [ [:simple :image] ]
-  (resource-field-path-specs-raw ResourceDefaultedNested) => [ [:defaulted {:image '/default.png'}] ]
-  (resource-field-path-specs-raw ResourceRepeatedNested)  => [ [:repeated [:images]] ]
-
-  (resource-field-path-specs-raw ResourceSimpleRepeatedlyNested)    => [ [[:simples] :image] ]
-  (resource-field-path-specs-raw ResourceDefaultedRepeatedlyNested) => [ [[:defaulteds] {:image '/default.png'}] ]
-  (resource-field-path-specs-raw ResourceRepeatedRepeatedlyNested)  => [ [[:repeateds] [:images]] ]"
```
matgis, line 379: We no longer use this way of declaring where to look for `resource` values, because you don't know the path beforehand when messages can be self-referencing.

Instead, you simply call `(protobuf/resource-field-value-paths pb-class pb-map)` directly.

editor/src/clj/editor/protobuf.clj
```diff
@@ -499,51 +504,68 @@ Macros currently mean no foreseeable performance gain, however."
          (persistent!)
          (msg->clj pb))))
 
-(defn- default-message-raw [^Class class default-included-field-rules]
-  (let [pb->clj (pb->clj-fn class default-included-field-rules)]
+(defn- default-message-raw [^Class class default-included-field-kinds]
+  (let [pb->clj (pb->clj-fn class default-included-field-kinds)]
```
matgis, line 508: "Field rule" is a Protobuf term that signifies the difference between a `required`, `optional`, or `repeated` field. We longer use field rules to distinguish between these. Instead, we classify field kinds as `required`, `optional`, `list`, or `map`, and the new term for this is `field-kind`.

editor/src/clj/editor/protobuf.clj
```diff
@@ -499,51 +504,68 @@ Macros currently mean no foreseeable performance gain, however."
...
+                       :pb-field-kind/optional
+                       (if include-defaults
+                         (if (message-field? field-info)
+                           (let [field-has-value? (field-has-value-fn pb-class java-name field-kind)
+                                 field-pb-class (:value-class field-info)]
+                             (fn pb->field-clj-value [pb]
+                               (if (field-has-value? pb)
+                                 (let [field-pb-value (.invoke field-get-method pb java/no-args-array)]
+                                   (pb-value->clj field-pb-value))
+                                 (default-message field-pb-class default-included-field-kinds)))) ; Call from inside the fn to support self-referential types.
```
matgis, line 545: This was outside the `fn` definition before. We can't call it from there when dealing with self-referencing types, so I had to move it inside the returned `fn`. It's memoized, so it shouldn't be an issue.

editor/src/clj/editor/protobuf.clj
```diff
@@ -662,22 +705,32 @@ Macros currently mean no foreseeable performance gain, however."
...
+          package (if (.hasJavaPackage options)
+                    (.getJavaPackage options)
+                    (let [full (full-name desc)
+                          name (desc-name desc)]
+                      (subs full 0 (- (count full) (count name) 1))))
+          outer-class-name (if (.hasJavaOuterClassname options)
+                             (.getJavaOuterClassname options)
+                             (->CamelCase (FilenameUtils/getBaseName (.getName file))))
+          inner-class-name (desc-name desc)]
+      (.intern (str package "." outer-class-name "$" inner-class-name)))))
```
matgis, line 725: Indentation change from breaking out the `class-name` part (most of the function) into a separately memoized function.

editor/src/clj/editor/protobuf.clj
```diff
@@ -793,7 +858,7 @@ Macros currently mean no foreseeable performance gain, however."
       (comp builder-fn vector->map)
       builder-fn)))
 
-(def ^:private pb-builder (fn/memoize pb-builder-raw))
+(def ^:private pb-builder (fn/memoize-late-bound {:timeout-ms 1000} pb-builder-raw))
```
matgis, line 863: `fn/memoize-late-bound` allows `pb-builder` to be called with an already-visited `Class` from inside `pb-builder-raw`.

editor/src/clj/editor/resource_node.clj
```diff
@@ -213,14 +213,13 @@
...
-              (filter seq)
-              (distinct))
-            (get-fields source-value)))))
+(defn- make-ddf-dependencies-fn-raw [^Class pb-class]
+  (fn ddf-dependencies-fn [source-value]
+    (coll/into->
+      (protobuf/resource-field-value-paths pb-class source-value) []
+      (map second) ; => proj-paths
+      (remove coll/empty?)
+      (distinct))))
```
matgis, line 222: `protobuf/resource-field-value-paths` returns pairs of `[path-to-value, value]`.

editor/src/clj/util/fn.clj
```diff
@@ -40,9 +40,25 @@
...
+  ([a b] (if a b a))
+  ([a b & more] (if a (reduce and-fn b more) a)))
+
+;; Exposed publicly as fn/or at the bottom of this file.
+(defn- or-fn
+  "Like clojure.core/or, but is eager and can be used as a function."
+  ([] nil)
+  ([a] a)
+  ([a b] (if a a b))
+  ([a b & more] (reduce or-fn (if a a b) more)))
```
matgis, line 57: These were previously at the bottom of the file and named `and` and `or`, respectively. We want to expose them as `fn/and` and `fn/or`.

But now I needed to call `or-fn` internally, so I moved them up and expose them under an alias at the bottom of the file.

editor/src/clj/util/fn.clj
```diff
@@ -40,9 +40,25 @@
...
+
+(definline ^:private with-memoize-info [memoized-fn original-fn cache arity]
   `(with-meta ~memoized-fn
-              {::memoize-original ~memoized-fn
+              {::memoize-original ~original-fn
```
matgis, line 61: Unrelated fix: Previously `::memoize-original` erroneously referred to the `memoized-fn`.

editor/src/dev/dev.clj
```diff
@@ -1326,26 +1323,48 @@
...
+                                                 value-message (assoc :value-message value-message))}))
+
+                        field-name (.getName field-desc)
+                        field-kind (protobuf/pb-field-desc-field-kind field-desc)
+                        field-info (coll/merge {:field-kind field-kind}
+                                               key-info
+                                               value-info)]
+                    (when (field-info-predicate field-info)
+                      (pair field-name field-info)))))
+          (.getFields desc))))
```
matgis, line 1359: Changed to include map key info.

editor/src/clj/editor/protobuf.clj
```diff
@@ -731,17 +781,32 @@ Macros currently mean no foreseeable performance gain, however."
...
+
+                                   field-builder
+                                   (if (= Descriptors$FieldDescriptor$JavaType/MESSAGE (.getJavaType fd))
+                                     (let [^Method field-get-method
+                                           (get methods
+                                                (if is-map-field
+                                                  (str "get" java-name "OrThrow")
+                                                  (str "get" java-name)))]
+                                       (pb-builder (.getReturnType field-get-method)))
+                                     (primitive-builder fd))
```
vlaaad: Not sure if this is the right place, but shouldn't the following code work?
```clj
(require '[editor.protobuf :as protobuf])
(import '[com.defold.editor.test TestDdf$MappedPrimitive])
(protobuf/map->pb TestDdf$MappedPrimitive {:string-to-number {"a" 1.0}})
```
It gives me this:
```
java.lang.AssertionError: Assert failed: (pb-class? cls)
  editor.protobuf/pb-class->descriptor (protobuf.clj:338)
  editor.protobuf/pb-builder-raw (protobuf.clj:768)
  clojure.core/apply (core.clj:667)
  util.fn/memoize-late-bound/memoized-fn/fn (fn.clj:262)
  util.fn/memoize-late-bound/memoized-fn (fn.clj:261)
  clojure.lang.AFunction$1.doInvoke (AFunction.java:33)
  editor.protobuf/pb-builder-raw/fn (protobuf.clj:802)
  clojure.core/map/fn/fn (core.clj:2759)
  clojure.core.protocols/iterator-reduce! (protocols.clj:42)
  clojure.core.protocols/iter-reduce (protocols.clj:52)
  clojure.core.protocols/fn (protocols.clj:74)
  clojure.core.protocols/fn/G (protocols.clj:13)
  clojure.core/transduce (core.clj:7026)
  clojure.core/into (core.clj:7042)
  editor.protobuf/pb-builder-raw (protobuf.clj:782)
  clojure.core/apply (core.clj:667)
  util.fn/memoize-late-bound/memoized-fn/fn (fn.clj:262)
  util.fn/memoize-late-bound/memoized-fn (fn.clj:261)
  clojure.lang.AFunction$1.doInvoke (AFunction.java:33)
  editor.resource/eval (resource.clj:950)
  clojure.lang.Compiler.eval (Compiler.java:7700)
  clojure.lang.Compiler.eval (Compiler.java:7655)
  clojure.core/eval (core.clj:3232)
  vlaaad.reveal.repl/eval-in-eval-file-metadata-namespace (repl.clj:129)
  vlaaad.reveal.repl/wrap-eval/fn (repl.clj:141)
  clojure.main/repl/read-eval-print/fn (main.clj:437)
  clojure.main/repl/read-eval-print (main.clj:437)
  clojure.main/repl/fn (main.clj:459)
  clojure.main/repl (main.clj:459)
  clojure.core/apply (core.clj:667)
  vlaaad.reveal.repl/repl (repl.clj:168)
  vlaaad.reveal/repl (reveal.clj:78)
  ... 5 more
```

vlaaad (commented):
LGTM, but it's possible there is a bug with the handling of primitive types in maps, see comment.

matgis:
Nice find! I fixed the bug and added a test for it.

PR #11970, opened 2026-03-03
Add `"text"` property to textual resources in editor scripts

vlaaad:
We added `"text"` properties to textual resources besides code resources, e.g.:
```lua
print(editor.get("/builtins/render/default.render", "text"))
-- script: "/builtins/render/default.render_script"
```

Fix #11969

### DIFF
--- a/editor/AGENTS.md
+++ b/editor/AGENTS.md
@@ -2,8 +2,7 @@
 
 ## Coding Style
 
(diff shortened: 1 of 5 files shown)


JosephFerano (approved):
image

Let me know if there's anything else related to the workflow you'd like me to test

PR #11977, opened 2026-03-03
Avoid surprisingly expensive file system access during project read

matgis:
Avoid calling `Path.toRealPath` on stateless resources when reading project data. This method appears to be surprisingly slow on certain file systems. We now only call this method for symlinks to stateless resources, where the existence of the file is important to determine at read-time.

This change resulted in a 24% project load time improvement in the large project we were testing in, but it will vary depending on the project.


vlaaad (approved):
"Fix performance regression" is really underselling it. It also improved the project load performance! On a big project I tried, it now takes 1m24s to load the project, down from 1m50s before regression!

PR #11980, opened 2026-03-04
Add "Duplicate Selection" code action

### LINKED ISSUE #8642: Duplicate line/selection in code editor
I find myself trying to duplicate a line often. (Used to it from PyCharm, if an example help)
Being able to duplicate a line would increase the speed of some coding tasks.

I would like if `CTRL+d` (you can pick any shortcut) duplicating the current line if there is no selection and current selection if something is selected. If no selection the duplicated line is placed on the next line, if there is a selection it's placed after the cursor.

Copy -> Paste -> Enter -> Paste -> Enter -> Paste

Or be real fiddly in your selection and select the line you want to copy _and_ the new line character.

vlaaad:
The text editor now supports a Duplicate Selection code action (<kbd>⌘⇧D</kbd> on macOS, <kbd>Ctrl Shift D</kbd> on Windows and Linux):

https://github.com/user-attachments/assets/936e49f2-1de1-453b-a79e-ae0a3c8c281b

Fix #8642


AGulev:
the only q. I have: are you sure it should be Cmd+Shift+D by default?

vlaaad:
I think it's usually Cmd+D, but our text editor already uses this shortcut for selecting the next occurrence.

AGulev:
https://www.perplexity.ai/search/what-is-the-shorcut-for-a-line-mLu7.Xe5SLiH9OmM1OR8Ww
CleanShot 2026-03-04 at 12 31 48@2x

AGulev:
I don't use it myself, so I don't really know what's better here

vlaaad:
Duplicate Line is not the same thing as Duplicate Selection. For example, vscode has both Duplicate Line and Duplicate Selection commands.

vlaaad:
To be more precise, vscode does NOT have a Duplicate Line command; it has Copy Line Up and Copy Line Down, which perplexity presented as Duplicate Line commands. So, for copying lines in a particular direction, it's common to use modifiers + arrows, while for duplicating a selection, it's modifiers and D.

matgis (dismissed):
Works great! Nice work getting it to work correctly with all the weird ways cursors can be placed, too! ❤️

I found a small discrepancy when testing, but I don't think it is very important to address: If I have a file with a trailing newline and I duplicate the last line prior to the trailing newline, the trailing newline is removed. If it is an easy fix, we could fix it before merging, otherwise I don't really mind it.

vlaaad:
Good catch! The issue was a minor bug in the `empty-subsequence?` predicate, which returned true for a subsequence that represents a single newline.

PR #11993, opened 2026-03-05
Expose engine time to shaders

### LINKED ISSUE #11987: Add support for built-in "time" and "delta-time" uniform data in shaders
Similar to Unitys built-in shader constants, we should automatically add uniform data for time. 

Proposal:

```glsl
uniform my_uniforms
{
    vec4 time;
    // or 
    vec4 time_data;
    ... other uniforms
};
```

Jhonnyg:
There is now a new constant type added `CONSTANT_TYPE_TIME` that can be used by shaders to automatically recieve time data from the engine:

```glsl
uniform my_uniforms
{
    vec4 my_time; // Set to constant type "Time" in the material editor!
};
```

The current data is exposed:
my_time.x : time since engine start
my_time.y : delta time from last frame (same as being passed into script updates)

This change means that you don't need to keep track of these values manually for certain effects like scrolling textures or other shader based animations!

Fixes #11987

### DIFF
--- a/engine/render/proto/render/material_ddf.proto
+++ b/engine/render/proto/render/material_ddf.proto
@@ -22,6 +22,7 @@ message MaterialDesc
         CONSTANT_TYPE_WORLDVIEW = 7;
         CONSTANT_TYPE_WORLDVIEWPROJ = 8;
         CONSTANT_TYPE_USER_MATRIX4 = 9;
+        CONSTANT_TYPE_TIME = 10;
     }
 
     message Constant
--- a/editor/src/clj/editor/material.clj
+++ b/editor/src/clj/editor/material.clj
@@ -164,6 +164,7 @@
                                     (.setElement 1 0 y)
                                     (.setElement 2 0 z)
                                     (.setElement 3 0 w)))
+    :constant-type-time (Vector4d. 0.0 0.0 0.0 0.0)
     :constant-type-viewproj :view-proj
     :constant-type-world :world
     :constant-type-texture :texture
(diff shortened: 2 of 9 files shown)


engine/render/proto/render/material_ddf.proto
```diff
@@ -22,6 +22,7 @@ message MaterialDesc
         CONSTANT_TYPE_WORLDVIEW = 7;
         CONSTANT_TYPE_WORLDVIEWPROJ = 8;
         CONSTANT_TYPE_USER_MATRIX4 = 9;
+        CONSTANT_TYPE_TIME = 10;
```
Jhonnyg, line 25: This is something to think about / discuss perhaps. The PR is setup to use a constant type to dictate when and where to put the time data, but another alternative would be to:

Detect the feature based on shader uniform name ONLY. I.e, if there's a uniform somewhere in the shader we can pick that up and set the data automatically:

```
uniform my_uniforms
{
    vec4 time; // this would get the time data automatically
    mat4 mtx_world;
    mat4 mtx_normal;
    ... etc
};
```

Constant type:
Pros - doesn't clash with old projects. it's reasonable to think that there are projects that have a uniform named time already. It's also very obvious from the project what data this gives you.
Cons - extra steps to configure, you need to add this constant type for all material that should have the time uniform (which I guess is similar to other uniforms, like tint or transforms..).

JCash: Regarding the "you need to add this constant type for all material", I don't really see it as a burden. How often do you really need to add/remove those?

JCash (approved):
Nice!

engine/render/proto/render/material_ddf.proto
Jhonnyg: True! It was my only "concern" about the feature so I thought I'd raise it!

Issue #8166, opened 2023-10-20
Optimise OpenGL renderer by utilizing OpenGL's state machine


aglitchman:
**Is your feature request related to a problem? Please describe (REQUIRED):**
By learning the OpenGL renderer of the Defold engine, I came to the conclusion that it is designed as if it was created according to the OpenGL handbook (I hope that didn't sound rude!). I mean that any draw call starts with activation (or bind) of an object/texture/program and ends with deactivation (or unbind). Even if the next draw call uses the same texture, the same program, the same matrices, etc.

As we know, OpenGL is stateful. Therefore, you can specify texture parameters once, set constants to the program once, and so on.

How to use this fact? In modern games 50-100 draw calls per frame is becoming normal, because it is impossible to draw shadows, outlines, other effects and objects themselves in 1-3 draw calls. And I have found that in my game for about 100 draw calls the Defold engine makes about **5000 API calls**. In a single frame! That may not be so much, but modern OpenGL is:
- on desktop runs as a client-server, commands are passed asynchronously. So graphics API calls do not go directly to the graphics card and passing any commands most likely takes extra CPU time.
- on web each API call is a long chain of calls: Wasm -> JavaScript -> Browser -> [OpenGL] or [ANGLE -> DirectX] or [ANGLE -> Metal] (depends on the platform).

And I came to the point that we can reduce the number of OpenGL API calls if we:
- Do not set the same texture filtering parameters every draw call.
- Cache values of uniform constants and update them only when they are really changed. This is especially noticeable when rendering models, when you need to draw 50 models with the same material that has 20 constants with almost the same values.

**Describe the solution you'd like (REQUIRED):**
Reduce the number of OpenGL API calls by setting texture parameters and constants once, and so on.

aglitchman:
Later I will make a PR where the constants will be cached. To discuss it and see if this optimization makes sense or if it's an overcomplication 🙌🏻

aglitchman:
I found an Emscripten Optimizing WebGL article that describes a similar idea. That it's overhead to make any OpenGL calls and you need to reuse OpenGL state as much as possible. That is, don't reset buffers and textures after each draw if the same buffers and textures are used in the next draw call. And so on.

https://emscripten.org/docs/optimizing/Optimizing-WebGL.html#avoid-redundant-calls

Jhonnyg:
At the moment, there's not that much more work to do that has more benefits than drawbacks (as per #11998). Closing.

PR #11998, opened 2026-03-05
Cache the pipeline state

Jhonnyg:
We now cache the internal pipeline state of the OpenGL graphics layer. 

Save ~120 API calls by skipping unbinding of vertex buffers and similar:

skip-unbind

Save ~300 API calls by caching sampler bindings. Very small performance gains.

cache-samplerbindings

Sub-task of #8166

### DIFF
--- a/engine/graphics/src/opengl/graphics_opengl.cpp
+++ b/engine/graphics/src/opengl/graphics_opengl.cpp
@@ -568,7 +568,7 @@ static void LogFrameBufferError(GLenum status)
         DM_STATIC_ASSERT(sizeof(m_TextureFormatSupport) * 8 >= TEXTURE_FORMAT_COUNT, Invalid_Struct_Size );
     }
 
-    static GLenum GetOpenGLCompareFunc(CompareFunc func)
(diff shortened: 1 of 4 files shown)


engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -558,62 +558,339 @@ static void LogFrameBufferError(GLenum status)
...
+            case COMPARE_FUNC_GREATER:  return GL_GREATER;
+            case COMPARE_FUNC_GEQUAL:   return GL_GEQUAL;
+            case COMPARE_FUNC_EQUAL:    return GL_EQUAL;
+            case COMPARE_FUNC_NOTEQUAL: return GL_NOTEQUAL;
+            case COMPARE_FUNC_ALWAYS:   return GL_ALWAYS;
+            default:
+                assert(0 && "Unsupported compare func");
+                return GL_ALWAYS;
+        }
     }
```
Jhonnyg: We did a similar change for when we were caching the texture filtering settings, so I thought we should probably do the same for all of these "lut"-style functions to use the same setup.

JCash (approved):

PR #12014, opened 2026-03-09
Add `parent` property to editor script outline nodes

### LINKED ISSUE #5370: editor.get(id, "parent")

### LINKED ISSUE #11898: Access parent node from outline selection in editor scripts

When writing editor scripts that operate on outline components (e.g., a Sprite), there is no API to navigate from a child component node to its parent game object node. This makes it impossible to programmatically add sibling components based on the selected component's data.

My use case: I want to right-click a Sprite component in the Outline, generate a `.convexshape` collision hull from the sprite's image (by reading its atlas and animation), and then automatically create a `collisionobject` component as a sibling on the same game object - all in one action.

I can successfully query the sprite's properties (`id`, `url`, `default_animation`, `__sampler__texture_sampler__0`) but cannot determine or access its parent game object to call `editor.tx.add(parent, "components", {...})`.

Here is what the available properties return for different outline nodes:

| Context | `path` | `id` | `url` |
|---|---|---|---|
| Root node of a `.go` file | `"/assets/images/logo.go"` | N/A | N/A |
| `go` (game object in `.collection`) | nil | `"go"` | `"/go"` |
| Sprite inside `.go` file | nil | `"sprite"` | `"#sprite"` |
| Sprite inside `go` in `.collection` | nil | `"sprite"` | `"/go#sprite"` |

While the `url` property hints at the parent (e.g., `/go#sprite` → parent is `go`), there is no way to resolve that into a node reference that can be passed to `editor.tx.add()`.

A way to navigate from an outline node to its parent. For example:

- A `"parent"` property: `editor.get(sprite_node, "parent")` → returns the parent game object node (userdata), or
- A function like `editor.get_parent(node)` → returns the parent node

This would allow:

```lua
local parent_go = editor.get(sprite_node, "parent")
editor.transact({
    editor.tx.add(parent_go, "components", {
        type = "collisionobject",
        id = "collisionobject",
        collision_shape = "/assets/player.convexshape"
    })
})
```

1. **Parsing the `url` property** - e.g., extracting `go` from `/go#sprite`. But there is no way to resolve the game object instance id back to a node reference, and for `.go` files the url is just `#sprite` with no file path information.
2. **Using `editor.ui.show_resource_dialog()`** - asking the user to manually select the `.go` file to add the component to. This works but adds friction and defeats the purpose of a one-click workflow.
3. **Creating a standalone `.collisionobject` file** instead of an embedded component - the user would then have to manually add it as a referenced component. This works but is a degraded experience.

This came up while building an editor script that generates convex hull collision shapes from sprite images. The full workflow I want to achieve:

1. User right-clicks a Sprite in the Outline → "Generate Convex Shape..."
2. Script reads the sprite's atlas (`__sampler__texture_sampler__0`) and animation (`default_animation`)
3. Script resolves the atlas → finds the source PNG/JPG image
4. Script runs a Python tool to generate a `.convexshape` file from the image silhouette
5. Script creates a `collisionobject` component **on the same game object** with `collision_shape` pointing to the generated `.convexshape`

Steps 1-4 work perfectly. Step 5 is blocked because there is no way to get from the sprite node to its parent game object.

vlaaad:
Editor scripts can now access outline parents using `editor.can_get(node, "parent")` and `editor.get(node, "parent")` in commands that query outline selection.

Fix #5370
Fix #11898

### DIFF
--- a/editor/test/integration/editor_extensions_test.clj
+++ b/editor/test/integration/editor_extensions_test.clj
@@ -28,6 +28,7 @@
             [editor.graph-util :as gu]
             [editor.handler :as handler]
             [editor.os :as os]
+            [editor.outline-view :as outline-view]
             [editor.pipeline.bob :as bob]
             [editor.prefs :as prefs]
             [editor.process :as process]
@@ -38,6 +39,7 @@
             [editor.workspace :as workspace]
             [integration.test-util :as test-util]
             [support.test-support :as test-support]
+            [util.coll :as coll]
             [util.diff :as diff]
             [util.http-server :as http-server]
             [util.path :as path])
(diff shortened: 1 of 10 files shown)


editor/test/integration/editor_extensions_test.clj
```diff
@@ -309,38 +311,48 @@
...
+    (reduce (fn [outline-selection index]
+              (nth (:children outline-selection) index))
+            decorated-outline
+            outline-path)))
+
 (deftest editor-scripts-commands-test
   (test-util/with-loaded-project "test/resources/editor_extensions/commands_project"
-    (let [sprite-outline (:node-id (test-util/outline (test-util/resource-node project "/main/main.collection") [0 0]))]
+    (let [resource-node (test-util/resource-node project "/main/main.collection")
+          sprite-outline (decorated-outline resource-node [0 0])
```
matgis, line 325: No need to change this, but we also have `test-util/outline-node-id` and `resource-outline-node-id`, which take proj-paths and string labels instead of indices. Sometimes this can make the tests clearer.

matgis (approved):

Jerakin:
Holy moly this is huge!

<sup>(so excited had to add a useless comment)</sup>

PR #12015, opened 2026-03-09
Add API for creating and manipulating lights to dmRender

Jhonnyg:
This adds a new (private for now) API for creating and manipulating light sources in the renderer. There are no user facing changes involved here, but this is needed for the new light component to create a glue between components and shaders.

Partial fix for #11724

### Technical changes

These changes are essentially the same as in the previous PR #11849 but with all the component level code stripped.

### DIFF
--- a/engine/graphics/src/graphics.cpp
+++ b/engine/graphics/src/graphics.cpp
@@ -1353,22 +1353,12 @@ namespace dmGraphics
     void IterateProgramResourceBindings(HProgram prog, ShaderResourceBindingFamily family, IterateProgramResourceBindingsCallback callback, void* user_data)
     {
         Program* program = (Program*) prog;
-        if (!program || !callback)
-        {
-            return;
-        }
-
         ProgramResourceBindingIterator it(program);
 
         const ProgramResourceBinding* binding;
         while ((binding = it.Next()))
         {
-            if (!binding->m_Res)
-            {
-                continue;
-            }
-
-            if (binding->m_Res->m_BindingFamily != family)
+            if (!binding->m_Res || binding->m_Res->m_BindingFamily != family)
             {
                 continue;
             }
@@ -1381,11 +1371,7 @@ namespace dmGraphics
             }
 
             const ShaderResourceTypeInfo* root_type = &type_infos[root_type_index];
-
-            callback(binding->m_Res->m_Set,
-                     binding->m_Res->m_Binding,
-                     root_type,
-                     user_data);
+            callback(binding->m_Res->m_Set, binding->m_Res->m_Binding, root_type, user_data);
         }
     }
 
--- a/engine/graphics/src/graphics.cpp
+++ b/engine/graphics/src/graphics.cpp
@@ -1185,7 +1185,7 @@ namespace dmGraphics
             free(meta.m_TypeInfos[i].m_Name);
             for (int j = 0; j < meta.m_TypeInfos[i].m_MemberCount; ++j)
             {
-                free(meta.m_TypeInfos[i].m_Members[j].m_Name);
+                free((char*) meta.m_TypeInfos[i].m_Members[j].m_Name);
             }
 
             delete[] meta.m_TypeInfos[i].m_Members;
--- a/engine/graphics/src/graphics.cpp
+++ b/engine/graphics/src/graphics.cpp
@@ -1350,7 +1350,7 @@ namespace dmGraphics
         }
     }
 
-    void IterateProgramResourceBindings(HProgram prog, ShaderResourceBindingFamily family, ProgramResourceBindingCallback callback, void* user_data)
+    void IterateProgramResourceBindings(HProgram prog, ShaderResourceBindingFamily family, IterateProgramResourceBindingsCallback callback, void* user_data)
     {
         Program* program = (Program*) prog;
         if (!program || !callback)
--- a/engine/graphics/src/graphics.cpp
+++ b/engine/graphics/src/graphics.cpp
@@ -1350,6 +1350,45 @@ namespace dmGraphics
         }
     }
 
+    void IterateProgramResourceBindings(HProgram prog, ShaderResourceBindingFamily family, ProgramResourceBindingCallback callback, void* user_data)
+    {
+        Program* program = (Program*) prog;
+        if (!program || !callback)
+        {
+            return;
+        }
+
+        ProgramResourceBindingIterator it(program);
+
+        const ProgramResourceBinding* binding;
+        while ((binding = it.Next()))
+        {
+            if (!binding->m_Res)
+            {
+                continue;
+            }
+
+            if (binding->m_Res->m_BindingFamily != family)
+            {
+                continue;
+            }
+
+            const dmArray<ShaderResourceTypeInfo>& type_infos = *binding->m_TypeInfos;
+            uint32_t root_type_index = binding->m_Res->m_Type.m_TypeIndex;
+            if (root_type_index >= (uint32_t) type_infos.Size())
+            {
+                continue;
+            }
+
+            const ShaderResourceTypeInfo* root_type = &type_infos[root_type_index];
+
+            callback(binding->m_Res->m_Set,
+                     binding->m_Res->m_Binding,
+                     root_type,
+                     user_data);
+        }
+    }
+
     void BuildUniforms(Program* program)
     {
         uint32_t uniform_count = 0;
@@ -1594,7 +1633,10 @@ namespace dmGraphics
             if (use_type_index)
             {
                 uint32_t child_type = member.m_Type.m_TypeIndex;
-                dmHashUpdateBuffer32(hash_state, &child_type, sizeof(child_type));
+                // Hash the type's name so the layout hash is independent of type array order
+                // (shader reflection may order types differently than manually built layouts).
+                dmhash_t child_name_hash = types[child_type].m_NameHash;
+                dmHashUpdateBuffer32(hash_state, &child_name_hash, sizeof(child_name_hash));
 
                 // Recurse into referenced type
                 HashTypeRecursive(child_type, types, num_types, hash_state, visited);
--- a/engine/graphics/src/graphics.h
+++ b/engine/graphics/src/graphics.h
@@ -271,17 +271,11 @@ namespace dmGraphics
     };
 
     // Callback invoked for each resource binding in a program that matches the specified family.
-    // The callback receives the descriptor set and binding indices for the resource, and the
-    // root type description for the bound resource.
-    typedef void (*ProgramResourceBindingCallback)(
-        uint16_t                      set,
-        uint16_t                      binding,
-        const ShaderResourceTypeInfo* root_type,
-        void*                         user_data);
+    typedef void (*IterateProgramResourceBindingsCallback)(uint16_t set, uint16_t binding, const ShaderResourceTypeInfo* root_type, void* user_data);
 
     // Iterate over all resource bindings for the given program that belong to the specified
     // binding family and invoke the supplied callback for each binding.
-    void IterateProgramResourceBindings(HProgram program, ShaderResourceBindingFamily family, ProgramResourceBindingCallback callback, void* user_data);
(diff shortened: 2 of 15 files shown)


engine/graphics/src/graphics.cpp
```diff
@@ -1350,6 +1350,45 @@ namespace dmGraphics
         }
     }
 
+    void IterateProgramResourceBindings(HProgram prog, ShaderResourceBindingFamily family, ProgramResourceBindingCallback callback, void* user_data)
```
Jhonnyg: This is new in this PR as opposed to the old PR:

In the old PR we included graphics_private.h from material.cpp, but it felt dirty so I decided to add a single function that iterates the shader resources instead of exposing *all* the structs and such to graphics.h

engine/render/src/render/render.h
```diff
@@ -119,6 +121,22 @@ namespace dmRender
...
+    enum LightType
+    {
+        LIGHT_TYPE_DIRECTIONAL = 0,
+        LIGHT_TYPE_POINT       = 1,
+        LIGHT_TYPE_SPOT        = 2,
+    };
+
+    // NOTE: These enum values are duplicated in gamesys camera DDF (camera_ddf.proto)
+    // Don't forget to change dmGamesysDDF::OrthoZoomMode if you change here
+    enum OrthoZoomMode
```
Jhonnyg, line 133: Moved from below so it lives next to all other enums.

engine/render/src/render/light.cpp
```diff
@@ -0,0 +1,348 @@
...
+            vec4 params;          // x: type (0 dir, 1 point, 2 spot; matches dmRender::LightType)
+                                  // y: intensity
+                                  // z: innerConeAngle (radians, spot only)
+                                  // w: outerConeAngle (radians, spot only)
+        };
+        uniform LightBuffer
+        {
+            vec4  lights_count;   // x: number of active lights
+            Light lights[MAX_LIGHTS];
+        };
```
Jhonnyg, line 154: This is the current fixed layout. We might have to change this later in future PRs if we want to support ES2, which I think we will (otherwise no editor support).

engine/graphics/src/graphics.cpp
```diff
@@ -1350,6 +1350,45 @@ namespace dmGraphics
         }
     }
 
+    void IterateProgramResourceBindings(HProgram prog, ShaderResourceBindingFamily family, IterateProgramResourceBindingsCallback callback, void* user_data)
+    {
+        Program* program = (Program*) prog;
+        if (!program || !callback)
+        {
+            return;
+        }
```
JCash: This looks LLM generated to me.
Will we ever call this without a program or callback?
I guess we could assert instead, but that too feels a bit redundant imo.

engine/render/src/render/light.cpp
```diff
@@ -0,0 +1,348 @@
...
+    ////////////////////////////////
+
+    LightPrototypeParams::LightPrototypeParams()
+    : m_Type(LIGHT_TYPE_POINT)
+    , m_Color(1.0f, 1.0f, 1.0f, 1.0f)
+    , m_Direction(0.0f, 0.0f, -1.0f)
+    , m_Intensity(1.0f)
+    , m_Range(10.0f)
+    , m_InnerConeAngle(0.0f)
+    , m_OuterConeAngle(3.1415926535f / 4.0f)
```
JCash: You can use the `M_PI` constants from `<math.h>`: `M_PI_4` is `PI/4`

engine/render/src/render/light.cpp
```diff
@@ -0,0 +1,348 @@
...
+        /*
+        struct Light
+        {
+            vec4 position;        // xyz: position, w: unused
+            vec4 color;           // RGBA (matches LightParams order)
+            vec4 direction_range; // xyz: normalized direction; w: range
+            vec4 params;          // x: type (0 dir, 1 point, 2 spot; matches dmRender::LightType)
+                                  // y: intensity
+                                  // z: innerConeAngle (radians, spot only)
+                                  // w: outerConeAngle (radians, spot only)
```
JCash, line 148: Look good. I believe this will work well for area lights too. E.g. sphere, disc, rectangle, cylinder.
They will usually need 1-2 arguments, which fits nicely here.

E.g. https://seblagarde.wordpress.com/wp-content/uploads/2015/07/course_notes_moving_frostbite_to_pbr_v32.pdf (page 40+)

engine/graphics/src/graphics.h
```diff
@@ -234,7 +234,48 @@ namespace dmGraphics
...
+        {
+            ShaderDesc::ShaderDataType m_ShaderType;
+            uint32_t                   m_TypeIndex;
+        };
+        uint8_t m_UseTypeIndex : 1;
+    };
+
+    struct ShaderResourceMember
+    {
+        char*                       m_Name;
```
JCash: const char*?

engine/render/src/render/light.cpp
```diff
@@ -0,0 +1,348 @@
...
+        light_types[1].m_Name        = (char*)"Light";
+        light_types[1].m_NameHash    = dmHashString64("Light");
+        light_types[1].m_Members     = light_members;
+        light_types[1].m_MemberCount = DM_ARRAY_SIZE(light_members);
+
+        // Generate a new layout for the light buffer
+        dmGraphics::UpdateShaderTypesOffsets(light_types, DM_ARRAY_SIZE(light_types));
+        dmGraphics::GetUniformBufferLayout(0, light_types, DM_ARRAY_SIZE(light_types), &g_LightBufferContext.m_LightBufferLayout);
+
+        g_LightBufferContext.m_Initialized = true;
```
JCash: Is the boolean needed?
Isn't it enough with "m_MaxLightCount > 0" ?

engine/render/src/render/light.cpp
```diff
@@ -0,0 +1,348 @@
...
+        light_std140.m_Position = dmVMath::Vector4(instance->m_Position);
+        light_std140.m_Color = instance->m_LightPrototype->m_Color;
+        light_std140.m_DirectionRange = dmVMath::Vector4(
+            instance->m_Direction,
+            instance->m_LightPrototype->m_Range);
+        light_std140.m_Params = dmVMath::Vector4(
+            instance->m_LightPrototype->m_Type,
+            instance->m_LightPrototype->m_Intensity,
+            instance->m_LightPrototype->m_InnerConeAngle,
+            instance->m_LightPrototype->m_OuterConeAngle);
```
JCash: perhaps lay the code foundation for the type check, and which data to upload?
E.g. seems a little presumptious to always pass cone info.

Btw, regarding "intensity" or other naming, did we do a design to compare what other engines provide and name things?

engine/render/src/render/light.cpp
JCash: Also, I don't see a "destruction", which means it will remain initialized even when rebooting the game.
I think I'd feel better to handle a create/destroy scenario.
E.g. as "max_light" is an input, if may actually change after a reboot (e.g. a performance config setting by the user to use fewer lights).

engine/render/src/render/light.cpp
```diff
@@ -0,0 +1,348 @@
...
+        }
+
+        if (!g_LightBufferContext.m_Initialized)
+        {
+            GenerateDefaultUniformBufferLayout(max_light_count);
+        }
+        else if (max_light_count != g_LightBufferContext.m_MaxLightCount)
+        {
+            dmLogOnceWarning("Using different sizes of light buffers is not supported. You should use the same size everywhere for the uniform buffer!");
+        }
```
JCash: If they cannot mismatch, why does the material have this setting? (material->m_LightBufferLightsCount)
E.g. would it be better if it was a project property?

engine/render/src/render/light.cpp
```diff
@@ -0,0 +1,348 @@
...
+        {
+            uint32_t min_size = dmMath::Max(1u, render_context->m_LightBufferScratch.Size());
+            ResizeLightBuffer(render_context, min_size);
+        }
+
+        // The light count has changed, we need to write that separately
+        if (render_context->m_LightBufferDirtyCount)
+        {
+            float count = (float) render_context->m_RenderLightsIndices.Size();
+            dmGraphics::SetUniformBuffer(render_context->m_GraphicsContext, render_context->m_LightUniformBuffer, 0, sizeof(float), &count);
```
JCash: So, even if the number of actual lights in the buffer hasn't changed, we update the first float (counter) to "change" the size of the data?

Are there any drawbacks to this?

engine/render/src/render/light.cpp
```diff
@@ -0,0 +1,348 @@
...
+            // The lightbuffer needs to match the material
+            if (material->m_LightBufferLightsCount != render_context->m_LightBufferScratch.Size())
+            {
+                ResizeLightBuffer(render_context, material->m_LightBufferLightsCount);
+            }
+
+            if (IsLightBufferDirty(render_context))
+            {
+                WriteLightInstanceData(render_context);
+            }
```
JCash: E.g. If a render.draw() call batches multiple materials interleaved, and the materials have different number of lights (e.g. for performance reasons), then we'll resize for each batch swap.

Is that costly?

engine/render/src/render/render_private.h
```diff
@@ -268,6 +273,34 @@ namespace dmRender
         uint8_t          m_Enabled    : 1;
     };
 
+    struct LightPrototype
+    {
+        LightType        m_Type;
+        dmVMath::Vector4 m_Color;
+        dmVMath::Vector3 m_Direction;
```
JCash: Type after m_Direction ? (i.e. sort by size to keep it smaller)

JCash (changes requested):
Some smaller changes.

Also some questions regarding the user facing properties ("Intensity" etc).
And some questions regarding the per-material max lights workflow.

engine/render/src/render/light.cpp
Jhonnyg: Yes, very true! I'm removing the static context and moving the data next to the render context instead.

engine/render/src/render/light.cpp
Jhonnyg: Hm well, I didn't present naming from other engines in a table, but I did look at what is common for the light types we want to support and went by that as base. I can add a table in the ddoc perhaps.

engine/render/src/render/light.cpp
Jhonnyg: We keep track of this now.

engine/render/src/render/light.cpp
Jhonnyg: Made the buffer static in size.

engine/render/src/render/light.cpp
JCash: Unity/Unreal seem to use "intensity".
I believe we also need to think about the units moving forward.
E.g. I can imagine small changes in our calculations, affecting carefully lit scenes.
How to best test for regressions? 

I honestly didn't anticipate the differences betwen light types (or engines):
https://www.perplexity.ai/search/in-unity-unreal-and-godot-and-JV5jg1JsQj2UzKExycnv7g#1

engine/graphics/src/graphics.cpp
```diff
@@ -1185,7 +1185,7 @@ namespace dmGraphics
             free(meta.m_TypeInfos[i].m_Name);
             for (int j = 0; j < meta.m_TypeInfos[i].m_MemberCount; ++j)
             {
-                free(meta.m_TypeInfos[i].m_Members[j].m_Name);
+                free((char*) meta.m_TypeInfos[i].m_Members[j].m_Name);
```
JCash, line 1188: technically it accepts a `void*`

JCash (approved):
lgtm!

engine/render/src/render/light.cpp
Jhonnyg: There doesn't seem to be a real consensus on what the values mean, but I think we should try to match gltf/blender unit spaces: https://claude.ai/share/97501c81-cb66-4b93-b6fd-e6c37aab9e00

PR #12035, opened 2026-03-13
Remove unnecessary resource sync on project open

vlaaad:
We had a useless resource sync before the real resource sync when opening a project. Removing it improves the open project time by 5% on big projects (on an example project: 1m51s -> 1m45s)

Related to https://github.com/defold/defold/issues/11989


matgis (dismissed):

matgis:
We still perform this extra resource-sync in `test-util/setup-workspace!`, which we call from `dev/load_project.clj`. We should perhaps update `load_project.clj` to not perform the unnecessary `resource-sync`, since we frequently use its more controlled environment in measurements?

PR #12037, opened 2026-03-13
Fixed issue where the HTML5 bundle always displayed the `DOMToGLFWKeyCode...` message in the console

AGulev:
Fixed an issue where every keyboard key press produced a console log in HTML5 bundles, even in release builds.

Fix https://github.com/defold/defold/issues/12017


AGulev:
It seems like logs were added for tests, but when removed GLFW wasn't rebuilt

britzl (approved):

PR #12043, opened 2026-03-16
Invalidate all successors on project load

vlaaad:
This change improves project load time on big projects by 13% (on a tested project: 1m46s -> 1m32s).

Related to #11989

### DIFF
--- a/editor/src/clj/dynamo/graph.clj
+++ b/editor/src/clj/dynamo/graph.clj
@@ -304,21 +304,57 @@
   (when (and (not (:dry-run transact-opts))
              (= :ok (:status tx-result)))
     (swap! *the-system* is/merge-graphs (get-in tx-result [:basis :graphs]) (:graphs-modified tx-result) (:outputs-modified tx-result) (:nodes-deleted tx-result))
+    (when-not (:track-changes transact-opts true)
+      (clear-system-cache!))
     nil))
 
 (defn transact
-  "Provides a way to run a transaction against the graph system.  It takes a list of transaction steps.
+  "Runs a transaction against the graph system.
+
+  Args:
+    opts    optional map with transaction settings:
+              :dry-run
+                Returns the tx report without committing to *the-system*.
+
+              :metrics
+                The metrics collector; results are returned in :metrics.
+
+              :track-changes
+                Defaults to true. When false, disables precise change tracking
+                and uses coarse cache invalidation instead of incremental
+                updates. The system cache is cleared automatically after
+                commit, but older evaluation contexts may still be stale and
+                must not be written back into the cache. Undo history is not
+                tracked accurately in this mode, so callers that need
+                consistent undo semantics must reset or otherwise manage
+                history explicitly.
+
+              :tx-data-context-map
+                Initial transaction data context map. The final value is
+                returned as :tx-data-context-map.
+    txs     sequence of transaction steps
 
   Example:
 
       (g/transact
-         [(g/connect n1 output-name n :xs)
+        [(g/connect n1 output-name n :xs)
          (g/connect n2 output-name n :xs)])
 
-  It returns the transaction result, (tx-result),  which is a map containing keys about the transaction.
-  Transaction result-keys:
-  `[:status :basis :graphs-modified :nodes-added :nodes-modified :nodes-deleted :outputs-modified :label :sequence-label]`
-  "
+  Returns a transaction result, a map with the following keys:
+    :status                :empty if no transaction steps completed, otherwise
+                           :ok
+    :basis                 transaction basis after applying the transaction
+    :graphs-modified       modified graph ids, most useful when change tracking
+                           is enabled
+    :nodes-added           added node ids
+    :nodes-modified        modified node ids when change tracking is enabled
+    :nodes-deleted         deleted nodes by node id
+    :outputs-modified      modified endpoints when change tracking is enabled
+    :label                 transaction label, if any
+    :sequence-label        transaction sequence label, if any
+    :tx-data-context-map   final transaction context map
+    :metrics               transaction metrics, when metrics collection is
+                           enabled"
   ([txs]
    (transact nil txs))
   ([opts txs]
--- a/editor/src/clj/dynamo/graph.clj
+++ b/editor/src/clj/dynamo/graph.clj
@@ -296,15 +296,15 @@
         override-id-generator (is/override-id-generator system)
         tx-data-context-map (or (:tx-data-context-map opts) {})
         metrics-collector (:metrics opts)
-        track-changes (:track-changes opts true)]
-    (it/new-transaction-context basis id-generators override-id-generator tx-data-context-map metrics-collector track-changes)))
+        full-invalidation (:full-invalidation opts)]
+    (it/new-transaction-context basis id-generators override-id-generator tx-data-context-map metrics-collector full-invalidation)))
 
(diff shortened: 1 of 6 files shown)


editor/src/clj/editor/defold_project.clj
```diff
@@ -1026,14 +1026,6 @@
...
              (load-nodes! project prelude-tx-data node-load-infos render-progress! resource-metrics transact-opts)))]
 
-     ;; When we're not tracking changes, we will not evict stale values from the
-     ;; system cache. This means subsequent graph queries won't see the changes
-     ;; from the transaction if a value was previously cached. To be on the safe
-     ;; side, we clear the cache after each transaction we perform with change
-     ;; tracking disabled.
-     (when-not change-tracked-transact
-       (g/clear-system-cache!))
-
```
vlaaad, line 1036: Now it's done automatically in `g/transact`.

editor/src/clj/internal/graph.clj
```diff
@@ -1261,3 +1261,9 @@
               basis))
           basis
           (util/group-into (comp gt/node-id->graph-id first) changes)))
+
+(defn invalidate-all-successors
+  [basis]
+  (update basis :graphs #(coll/map-vals (fn [graph]
+                                          (assoc graph :successors (make-successors)))
+                                        %)))
```
matgis: Would `update-vals` work here?
```clj
(update basis :graphs update-vals assoc :successors (make-successors))
```

editor/src/clj/internal/transaction.clj
```diff
@@ -704,13 +708,14 @@
...
-                                 (e/map #(pair % source-label)
-                                        (ig/get-overrides (:basis ctx) source-id))))
+      (cond-> (:track-changes ctx)
+        ;; When updating the successors, we must also consider any override nodes
+        ;; of the source node, since these will inherit an implicit connection
+        ;; between them and the corresponding override nodes of the target node.
+        (flag-successors-changed (e/cons
+                                   (pair source-id source-label)
+                                   (e/map #(pair % source-label)
+                                          (ig/get-overrides (:basis ctx) source-id)))))
```
matgis: Style: Let's add a line break to the `flag-successors-changed` call like you did above to keep things consistent. 😊

editor/src/clj/internal/transaction.clj
```diff
@@ -1108,9 +1113,17 @@
             override-nodes-affected-ordered)))
 
 (defn update-successors
-  [{:keys [successors-changed] :as ctx}]
+  [{:keys [completed-action-count successors-changed] :as ctx}]
```
matgis: Type hint `^long completed-action-count` to avoid boxed math `zero?` check below?

editor/test/internal/transaction_test.clj
```diff
@@ -202,6 +202,70 @@
...
+              graph-b-successors-before (graph-successors-cache graph-b)]
+          (g/transact {:track-changes false}
+            (g/set-property resource-a :marker (int 1)))
+
+          (let [graph-a-successors-after (graph-successors-cache graph-a)
+                graph-b-successors-after (graph-successors-cache graph-b)]
+            (is (not (identical? graph-a-successors-before graph-a-successors-after)))
+            (is (not (identical? graph-b-successors-before graph-b-successors-after)))
+            (is (= #{(g/endpoint receiver-b :passthrough)}
+                   (set (g/successors (g/now) resource-b :b)))))))))
```
matgis, line 243: I don't quite understand this test. Why do we want this to happen in particular? Are we testing the wrong thing here? Otherwise this could use an explaining comment.

editor/test/internal/transaction_test.clj
```diff
@@ -202,6 +202,70 @@
         (is (= #{:_declared-properties :_properties :a-property :ordinary :self-dependent}
                (into #{} (map gt/endpoint-label) outputs-modified)))))))
 
+(defn- graph-successors-cache
+  [graph-id]
+  (get-in @g/*the-system* [:graphs graph-id :successors]))
+
+(deftest transact-without-change-tracking-test
+  (testing "invalidates all successors after a property update"
```
matgis: Nit-pick: The test runner concatenates nested `testing` comments with single spaces when reporting errors, so it's nicer if they are properly formatted sentences. E.g:
```
"Invalidates all successors after a property update."
```

editor/src/clj/editor/defold_project.clj
matgis: Don't forget to update `dev/load_project.clj` as well!

editor/src/clj/dynamo/graph.clj
```diff
@@ -304,21 +304,57 @@
...
+
+              :track-changes
+                Defaults to true. When false, disables precise change tracking
+                and uses coarse cache invalidation instead of incremental
+                updates. The system cache is cleared automatically after
+                commit, but older evaluation contexts may still be stale and
+                must not be written back into the cache. Undo history is not
+                tracked accurately in this mode, so callers that need
+                consistent undo semantics must reset or otherwise manage
+                history explicitly.
```
matgis, line 330: I think the new behavior of automatically clearing the cache and invalidating all successors is better since it is less error-prone. However, I can't help but feel that supplying `{:track-changes false}` and having it clear the cache is a bit unexpected. Maybe we should rename the parameter to something a bit less innocuous. Something like `{:apocalyptic true}`, perhaps? 😅

matgis (changes requested):

editor/src/clj/editor/defold_project.clj
vlaaad: Argh, I forgot about `load-project` again! Good catch!

editor/src/clj/internal/transaction.clj
vlaaad: Okay!

editor/test/internal/transaction_test.clj
vlaaad: Successors is a mutable object (an atom wrapper), so `not identical?` ensures the successors cache is different. I'll add a comment.

editor/src/clj/internal/graph.clj
vlaaad: That would be nice, but not quite like you propose, since `(make-successors)` is a mutable cache (basically an atom wrapper) — we don't want to share these between graphs. I'll simplify the code though!

editor/src/clj/dynamo/graph.clj
vlaaad: I'll go with `:full-invalidation` 😅

editor/src/clj/internal/transaction.clj
```diff
@@ -213,17 +214,18 @@
...
+            ;; new override node into account.
+            (flag-all-successors-changed all-originals)
 
-        ;; Similarly, so must the source outputs of any arcs that target any of
-        ;; the original nodes.
-        (flag-successors-changed (e/mapcat #(gt/sources basis %) all-originals)))))
+            ;; Similarly, so must the source outputs of any arcs that target any of
+            ;; the original nodes.
+            (flag-successors-changed (e/mapcat #(gt/sources basis %) all-originals))))
+      ctx)))
```
matgis: Style: I think this would read better if we swap the branches and use regular `if`. Seems more natural now that `:full-invalidation` defaults to `false`, too.  🤔

matgis (dismissed):

PR #12044, opened 2026-03-16
Use more efficient parallelization for reading resources

vlaaad:
This change improves project open performance on large projects by 4% (1m32s -> 1m28s).

Related to #11989

### DIFF
--- a/editor/src/clj/util/coll.clj
+++ b/editor/src/clj/util/coll.clj
@@ -1047,20 +1047,35 @@
    (mapv #(apply f % args) coll)))
 
 (defn pmapv
-  "Like core.pmap, but eagerly returns a vector and uses virtual threads. Fails
-  fast by cancelling remaining tasks on the first error."
+  "Like core.pmap, but eagerly returns a vector, uses virtual threads, and
+  keeps a bounded number of tasks in flight. Fails fast by cancelling
+  remaining tasks on the first error."
   ([f coll]
    (let [items (if (vector? coll) coll (vec coll))
          item-count (count items)]
      (if (zero? item-count)
        []
        (let [binding-frame (Var/cloneThreadBindingFrame)
-             results (object-array item-count)]
+             results (object-array item-count)
+             next-index (AtomicInteger. 0)
+             worker-count (-> (.availableProcessors (Runtime/getRuntime))
+                              (* 2)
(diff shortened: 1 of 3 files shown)


vlaaad:
Huh, that's actually interesting. I can't reproduce the issue by default because the JVM sets the max file descriptor limit to >1000000 by default, but if the limit is lowered, we can hit the error. So this might happen if the big project grows 10x more. I'll keep iterating on it.

vlaaad:
I have now changed the `pmapv` to use bounded parallelism — no change in performance.

editor/src/clj/util/coll.clj
```diff
@@ -1044,3 +1045,40 @@
...
    (mapv #(apply f % args) coll)))
+
+(defn pmapv
+  "Like core.pmap, but eagerly returns a vector, uses virtual threads, and
+  keeps a bounded number of tasks in flight. Fails fast by cancelling
+  remaining tasks on the first error."
+  ([f coll]
+   (let [items (if (vector? coll) coll (vec coll))
+         item-count (count items)]
+     (if (zero? item-count)
```
matgis: Might be nice to have a special case for a `coll` with a single element as well and avoid the thread setup ceremony? E.g:
```clj
(case item-count
  0 []
  1 [(f (items 0))]
  ...)

File: editor/src/clj/util/coll.clj [hunk 2]
```diff
@@ -1044,3 +1045,40 @@
...
+       []
+       (let [binding-frame (Var/cloneThreadBindingFrame)
+             results (object-array item-count)
+             next-index (AtomicInteger. 0)
+             worker-count (-> (.availableProcessors (Runtime/getRuntime))
+                              (* 2)
+                              (min (long item-count))
```
matgis: `long` cast seems unnecessary here?

editor/src/clj/util/coll.clj
```diff
@@ -1044,3 +1045,40 @@
...
+                              (max 4))]
```
matgis: Shouldn't the `(max 4)` that ensures we get a minimum of 4 virtual threads go before the `(min (long item-count))` that ensures we don't get more threads than we have items?

matgis (dismissed):

PR #12051, opened 2026-03-17
Shader precision options

### LINKED ISSUE #6519: Option to set GLSL ES precision for the whole shader

Jhonnyg:
Two new game.project options have been added that controls how the default type precision for floats and ints are generated for GLSL ES shaders:

```yaml
shader.glsl_es_default_precision_float = highp | mediump
shader.glsl_es_default_precision_int = highp | mediump
```

These will affect how the global type qualifiers are generated in the cross-compiled shaders:

```glsl
#version 300 es
precision highp float;
precision mediump int;
```

Fixes #6519

### DIFF
--- a/editor/src/clj/editor/code/shader_compilation.clj
+++ b/editor/src/clj/editor/code/shader_compilation.clj
@@ -23,7 +23,7 @@
             [editor.workspace :as workspace]
             [util.array :as array]
             [util.eduction :as e])
-  (:import [com.dynamo.bob.pipeline ShaderProgramBuilderEditor]
+  (:import [com.dynamo.bob.pipeline ShaderProgramBuilderEditor Shaderc$ShaderPrecision]
            [com.dynamo.bob.pipeline.shader ShaderCompilePipeline$ShaderModuleDesc]
            [com.dynamo.graphics.proto Graphics$ShaderDesc Graphics$ShaderDesc$Language]))
 
@@ -59,6 +59,13 @@
     (set! (. shader-module-desc type) pb-shader-type)
     shader-module-desc))
 
+(defn- precision-string->enum
+  ^Shaderc$ShaderPrecision [s default-enum]
+  (cond
+    (nil? s) default-enum
+    (= "highp" (str s)) Shaderc$ShaderPrecision/SHADER_PRECISION_HIGHP
+    :else Shaderc$ShaderPrecision/SHADER_PRECISION_MEDIUMP))
+
 (defn- error-string->error-value [^String error-string]
   (g/error-fatal (string/trim error-string)))
 
@@ -70,7 +77,7 @@
       {:resource build-resource
        :content (protobuf/pb->bytes shader-desc)})))
 
-(defn make-shader-build-target [node-id shader-source-infos max-page-count exclude-gles-sm100]
+(defn make-shader-build-target [node-id shader-source-infos max-page-count exclude-gles-sm100 & [settings]]
   {:pre [(g/node-id? node-id)
          (vector? shader-source-infos)
          (pos? (count shader-source-infos))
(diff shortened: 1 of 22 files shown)


editor/src/clj/editor/defold_project.clj
```diff
@@ -1634,6 +1632,8 @@
   (output display-height g/Num (g/fnk [settings]
                                   (double (or (get settings ["display" "height"]) 0))))
   (output exclude-gles-sm100 g/Any (g/fnk [settings] (get settings ["shader" "exclude_gles_sm100"])))
+  (output glsl-es-default-precision-float g/Any (g/fnk [settings] (get settings ["shader" "glsl_es_default_precision_float"])))
+  (output glsl-es-default-precision-int g/Any (g/fnk [settings] (get settings ["shader" "glsl_es_default_precision_int"])))
```
Jhonnyg, line 1636: I made these to match how the "exclude-gles-sm100" flag is propagated through the editor.

editor/src/clj/editor/code/shader_compilation.clj
```diff
@@ -96,8 +103,9 @@
         pb-shader-languages (if exclude-gles-sm100
                               pb-shader-languages-without-gles-sm100
                               pb-default-shader-languages)
-
-        shader-desc-build-result (ShaderProgramBuilderEditor/makeShaderDescWithVariants build-resource-path shader-module-descs-array pb-shader-languages (int max-page-count))
+        float-precision (glsl-precision-string->enum glsl-es-default-precision-float)
+        int-precision (glsl-precision-string->enum glsl-es-default-precision-int)
```
vlaaad: Should we fail the compilation if the value is different? Can it be different here? E.g., if the user edits game.project manually to use something else...

editor/src/clj/editor/pipeline/shader_gen.clj
```diff
@@ -135,38 +135,49 @@
...
+  when nil, mediump float and highp int are used."
+  [^String shader-path ^String shader-source max-page-count float-precision-str int-precision-str]
   {:pre [(string? shader-path)
          (pos? (count shader-path))
          (string? shader-source)
          (pos? (count shader-source))]}
   (let [shader-type (graphics.types/filename-shader-type shader-path)
         pb-shader-type (graphics.types/shader-type-pb-shader-type shader-type)
+        float-precision (precision-string->enum float-precision-str Shaderc$ShaderPrecision/SHADER_PRECISION_MEDIUMP)
+        int-precision (precision-string->enum int-precision-str Shaderc$ShaderPrecision/SHADER_PRECISION_HIGHP)
```
vlaaad: Should we throw instead of falling back to the default if there is some invalid value?

editor/src/clj/editor/pipeline/shader_gen.clj
```diff
@@ -135,38 +135,49 @@
     ::shader-transpile-error true
     false))
 
+(defn- precision-string->enum
+  "Maps game.project precision string to Shaderc.ShaderPrecision. default-enum used when s is nil."
+  ^Shaderc$ShaderPrecision [s default-enum]
+  (cond
+    (= "highp" (str s)) Shaderc$ShaderPrecision/SHADER_PRECISION_HIGHP
+    (= "mediump" (str s)) Shaderc$ShaderPrecision/SHADER_PRECISION_MEDIUMP
+    :else default-enum))
```
vlaaad: We could also use case here...

vlaaad (dismissed):
LGTM.

Not sure about error handling... Is it possible to edit `game.project` to use invalid settings like `lowp`, and if it is, should we perhaps fail builds in this case? We typically try to catch invalid data errors instead of pretending it's fine.

AGulev (dismissed):
LGTM

editor/src/clj/editor/code/shader_compilation.clj
```diff
@@ -64,7 +64,9 @@
   (case s
     "highp" Shaderc$ShaderPrecision/SHADER_PRECISION_HIGHP
     "mediump" Shaderc$ShaderPrecision/SHADER_PRECISION_MEDIUMP
-    nil))
+    (throw (IllegalArgumentException.
+             (format "Invalid shader precision '%s'. Expected \"highp\" or \"mediump\"."
+                     (str s))))))
```
vlaaad: This code is invoked inside graph outputs. This means this code should NOT throw exceptions, and instead it should return error values.

editor/src/clj/editor/pipeline/shader_gen.clj
```diff
@@ -161,8 +163,8 @@
          (pos? (count shader-source))]}
   (let [shader-type (graphics.types/filename-shader-type shader-path)
         pb-shader-type (graphics.types/shader-type-pb-shader-type shader-type)
-        float-precision (precision-string->enum float-precision-str Shaderc$ShaderPrecision/SHADER_PRECISION_MEDIUMP)
-        int-precision (precision-string->enum int-precision-str Shaderc$ShaderPrecision/SHADER_PRECISION_HIGHP)
+        float-precision (precision-string->enum float-precision-str)
+        int-precision (precision-string->enum int-precision-str)
```
vlaaad: This code now throws, which is fine in this context, but the other thrown exceptions are decorated using `decorate-transpile-error`. Maybe move function invocations inside the try block that does the decoration?

vlaaad (changes requested):
Yes, but not like that :P

PR #12053, opened 2026-03-17
Fix `model_animation_done` callback

AGulev:
Fix https://github.com/defold/defold/issues/12048


engine/gamesys/src/gamesys/scripts/script_model.cpp
```diff
@@ -428,7 +428,7 @@ namespace dmGameSystem
 
         AnimationCallbackContext* callback_ctx = new AnimationCallbackContext();
         callback_ctx->m_LuaCallback = 0;
-        callback_ctx->m_Listener = receiver;
+        callback_ctx->m_Listener = sender;
```
AGulev, line 431: This is the fix, the rest just a test

britzl (approved):

PR #12059, opened 2026-03-18
New light component in engine

Jhonnyg:
A light component has been added that uses the .light resource as input and is essentially a .data resource in disguise. This is the last piece of the engine implementation for the light data support.

Engine implementation for #11724


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/meta.properties
```diff
@@ -484,6 +484,14 @@ max_count.type = integer
 max_count.help = max number of mesh components, 128 by default
 max_count.default = 128
 
+[light]
+help = Light related settings
+group = Components
+
+max_count.type = integer
+max_count.help = max number of light components, 64 by default
+max_count.default = 64
```
Jhonnyg, line 501: Not exposed to editor yet. But needed for testing.

engine/engine/src/engine.cpp
```diff
@@ -1470,6 +1471,7 @@ namespace dmEngine
             engine->m_ResourceTypeContexts.Put(dmHashString64("guic"), engine->m_GuiContext);
         }
         engine->m_ResourceTypeContexts.Put(dmHashString64("fontc"), engine->m_RenderContext);
+        engine->m_ResourceTypeContexts.Put(dmHashString64("lightc"), engine->m_RenderContext);
```
Jhonnyg, line 1479: The .lightc is the engine resource type for the lights, the actual generated data is still a ".data" type. The actual engine resource creates and manages a HLightPrototype (e.g the render 'resource', similar to how particle prototypes work) that can then be used to spawn light components from.

engine/engine/src/engine.cpp
```diff
@@ -1282,6 +1282,7 @@ namespace dmEngine
         render_params.m_MaxDebugVertexCount = 0;
 #endif
         render_params.m_MaxBatches = (uint32_t) dmConfigFile::GetInt(engine->m_Config, "graphics.max_font_batches", 128);
+        render_params.m_MaxLights = (uint32_t) dmMath::Max(0, dmConfigFile::GetInt(engine->m_Config, "light.max_count", 64));
```
JCash: In the spirit of decoupling components/resources from the engine setup, Could this be setup using a `dmRender::SetMaxNumberLights()` instead?
Then it can be called from the component function, and the component can read the config setting and so on....

engine/gamesys/src/gamesys/gamesys.cpp
```diff
@@ -299,6 +299,8 @@ namespace dmGameSystem
                 0, 0,
                 0);
 
+        // prio: 1000 comp_light.cpp
```
JCash, line 310: Is there a specific strategy to why the priority is in between model and sprite?
If so, perhaps it's good to have a oneliner comment about it?

engine/engine/src/engine.cpp
JCash: This is necessary for the time being, but I want to design away this connection as well, in order for the resource plugins decide what they want to get/register to these stores.

Basically, I want something like the context store we use for components (as I mentioned above)

JCash (changes requested):
Overall good.
Would like to see the render configuration done by the light component though.

engine/gamesys/src/gamesys/gamesys.cpp
Jhonnyg: I don't actually know the reason for the priority. It was 1000 with the old component, so I figured I wouldn't change it needlessly.

PR #12061, opened 2026-03-18
Fix build error source links for embedded resources

### LINKED ISSUE #11697: Improve "Unknown source" error reporting for missing collectionfactory collections
When a collectionfactory references a non-existent collection file (e.g. `/main/test.collection`), the IDE shows a vague build error: "Unknown source - The file '/main/test.collection' could not be found." without specifying the problematic collectionfactory or gameobject. 

Enhance the Defold IDE's build error reporting to include the exact source of the issue, such as the filename of a collection or a gameobject. This would allow quick navigation to the problem in the editor.

Manual full-project text search for the missing collection filename across .collection and .go

Image

Test project:
[missing_collection.zip](https://github.com/user-attachments/files/24435021/missing_collection.zip)

AGulev:
Build errors coming from embedded resources now consistently open the owning file instead of ending up with an unknown or missing source. 

Fix https://github.com/defold/defold/issues/11697

### Technical Changes

The build errors view now resolves the owner resource through `owner-resource-node-id` before building the error tree, instead of assuming the failing node is itself a top-level ResourceNode. I also added an integration test covering an embedded collection proxy with a missing collection reference to verify the parent error item links to /main/main.collection.

### DIFF
--- a/editor/src/clj/editor/build_errors_view.clj
+++ b/editor/src/clj/editor/build_errors_view.clj
@@ -43,9 +43,8 @@
   (conj PersistentQueue/EMPTY item))
 
 (defn- openable-resource [evaluation-context node-id]
-  (let [basis (:basis evaluation-context)
-        owner-resource-node-id (resource-node/owner-resource-node-id basis node-id)]
-    (when (g/node-instance? basis resource/ResourceNode owner-resource-node-id)
+  (let [basis (:basis evaluation-context)]
+    (when-let [owner-resource-node-id (resource-node/owner-resource-node-id basis node-id)]
       (when-some [resource (g/node-value owner-resource-node-id :resource evaluation-context)]
         (when (resource/openable-resource? resource)
           {:node-id owner-resource-node-id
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/build_errors_view.clj
```diff
@@ -42,10 +43,13 @@
   (conj PersistentQueue/EMPTY item))
 
 (defn- openable-resource [evaluation-context node-id]
-  (when (g/node-instance? (:basis evaluation-context) resource/ResourceNode node-id)
-    (when-some [resource (g/node-value node-id :resource evaluation-context)]
-      (when (resource/openable-resource? resource)
-        resource))))
+  (let [basis (:basis evaluation-context)
+        owner-resource-node-id (resource-node/owner-resource-node-id basis node-id)]
+    (when (g/node-instance? basis resource/ResourceNode owner-resource-node-id)
```
vlaaad: `resource-node/owner-resource-node-id` returns either `nil` or a node that is already validated to be a `resource/ResourceNode`. If it returns nil, this code will throw NPE. If it returns a resource node, it will perform a useless check. I recommend using `(when-let [owner-resource-node-id (resource-node/owner-resource-node-id basis node-id)] ...)` here instead.

vlaaad (changes requested):
LGTM, but there is a small issue we should fix before merging.

vlaaad (approved):
Ship it!

Issue #11923, opened 2026-02-19
Error building project after version upgrade - "unable to resolve build target from value {:FileResource ...}"


swajj:
### Actual behaviour
Error building my project from Defold editor 1.12.2 beta
This appears to also reproduce in 1.12.1, but not in 1.12.0. The screenshot/bugreport were produced by 1.12.2 beta.

<img width="3440" height="1440" alt="Image" src="https://github.com/user-attachments/assets/220cff30-8801-4f3f-944e-653c3f8c3da6" />

### Steps to reproduce
Load my project in Defold 1.12.1 or 1.12.2 beta
Build
Get error

<hr/>

<table>
<tr><td>Build time</td><td>2026-02-18T09:22:48.667685</td></tr>
<tr><td>Defold channel</td><td>beta</td></tr>
<tr><td>Defold editor sha</td><td>31e1a6b7a7b1d217e277a5a2956c18dfacff1618</td></tr>
<tr><td>Defold engine sha</td><td>31e1a6b7a7b1d217e277a5a2956c18dfacff1618</td></tr>
<tr><td>Defold version</td><td>1.12.2</td></tr>
<tr><td>Error</td><td><a href='https://sentry.io/organizations/defold/issues/?query=id%3A"d2c7cb1dcb5b4fab97dce011b1672fe8"'>d2c7cb1dcb5b4fab97dce011b1672fe8</a></td></tr>
<tr><td>GPU</td><td>Intel(R) UHD Graphics 770</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 - Build 32.0.101.6987</td></tr>
<tr><td>Java version</td><td>25+36-LTS</td></tr>
<tr><td>OS arch</td><td>amd64</td></tr>
<tr><td>OS name</td><td>Windows 11</td></tr>
<tr><td>OS version</td><td>10.0</td></tr>
<table>

drajjon:
Obviously happy to try things / provide logs / etc., I haven't tried to narrow down a repro / hunt for workarounds yet. We do use relatively large atlases compared to most projects (I assume) - not sure if that's at all related here or not.

drajjon:
This appears to be because we have atlases with >8 pages.

Reducing all of my atlases down to 8 or less pages appears to fix the build.

I could create a repro project if useful, but I'm 99% sure it's just "have an atlas with 9+ pages".

This is a pretty big breaking/blocking change for us, reworking our art system to use more atlases is on the roadmap but it's a multi-week project.

[...]

PR #12065, opened 2026-03-19
Editor fix: Allow shaders to compile with page count > 8

JosephFerano:
Allow the compilation of shaders when a paged material exceeds the maximum allowed pages currently set to 8 when "Exclude GLES 2.0" is not checked in game.project > Shader setting.

### Technical changes

Further work on #11923 which was closed

The user was able to get around the issue by using their own shader which wasn't causing compilation issues, and I originally missed the building errors when using their repro project. We now pass the exclude gles 2.0 project setting to the graph nodes that validate the page count; SpriteNode, GuiSceneNode, and EmitterNode.


JosephFerano:
@matgis I wanted to write tests for this, but I think it has to be a standalone project since it'll usually fail to compile given that the default project setting is not to exclude GLES 2.0. Is this worth creating those tests? I have to setup a project anyway to test this, cause I don't have paged materials for some things.

JosephFerano:
Well, I added the test with a test project.

editor/test/resources/max_paged_count_project/main/main.collection
```diff
@@ -0,0 +1,55 @@
...
+  "far_z: 1000.0\\n"
+  "\"\n"
+  "}\n"
+  ""
+  position {
+    x: 344.0
+    y: 280.0
+    z: 339.0
+  }
+}
```
matgis: Tiny nit-pick: I suppose we don't really need the Camera to be present for this test, so we could remove it.

matgis (dismissed):

matgis (approved):

AGulev:
Should it be fixed in beta? I don't know if it's a hotfix of something we expected to be fixed in beta

JosephFerano:
@AGulev It could be merged, it was part of another hotfix, but since it unblocked the user, I think it's ok?

AGulev:
@JosephFerano if it was promised to be fixed in 1.12.3, then it's better to cherry-pick it into beta

JosephFerano:
Let's do it

PR #12088, opened 2026-03-23
Replace SSDP/UPnP target discovery with mDNS/DNS-SD for Defold debug targets

### LINKED ISSUE #8570: device does not appear in the Targets menu [macOS]
The device does not appear in the Targets menu,

1) certificate iOS Development, profile iOS Development (the application opens successfully on the device)

2) iOS developer mode is enabled on device.

3) device is connected to the same wifi network as your computer (checked also via `ios-deploy -c`, device is connected via USB and via wif-fi)
```
[....] Found 00008120-0001451C0A70201E (D73AP, iPhone 14 Pro, iphoneos, arm64e, 17.3.1, ) a.k.a. 'iPhone (Nikita)' connected through WIFI.
[....] Found 00008120-0001451C0A70201E (D73AP, iPhone 14 Pro, iphoneos, arm64e, 17.3.1, 21D61) a.k.a. 'iPhone (Nikita)' connected through USB.
```
4) the dev app is built in Debug mode

5) I tried also `Project -> "Enter Target IP"` - without success. `no route to host (connect failed)`

  - Version 1.6.4

**Platforms (REQUIRED):**
 - Platforms: macOS
 - OS: Monterey 12.7.3
 - Device: iPhone 14 Pro iOS 17.3.1

AGulev:
Debug engine builds now advertise themselves over mDNS/DNS-SD using the `_defold._tcp` service and the editor now discovers debug targets through mDNS instead of SSDP/UPnP. iOS debug builds now publish the discovery service through Bonjour, so they can be discovered by the editor over the local network without an extra entitlements.

It fixes the issue that iOS targets can't be discovered by the editor.

⚠️If you are using a custom `Info.plist` in your project, please make sure it is updated with the changes introduced in this version (new section):

```xml
        {{^variant_release}}
        <key>NSBonjourServices</key>
        <array>
                <string>_defold._tcp</string>
        </array>

        <key>NSLocalNetworkUsageDescription</key>
        <string>Discover Defold targets on the local network.</string>
        {{/variant_release}}
```

Fix https://github.com/defold/defold/issues/8570
Fix https://github.com/defold/defold/issues/3660


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/BundleHelper.java
```diff
@@ -286,6 +286,9 @@ public void updateTemplateProperties() throws CompileExceptionError, IOException
         String exeName = BundleHelper.projectNameToBinaryName(title);
         this.templateProperties.put("exe-name", exeName);
         this.templateProperties.put("build-timestamp", String.valueOf(System.currentTimeMillis() / 1000));
+        if (Bob.VARIANT_RELEASE.equals(project.option("variant", Bob.VARIANT_RELEASE))) {
+            this.templateProperties.put("variant_release", Boolean.TRUE);
+        }
```
AGulev, line 291: I was surprised we don't expose this flag anywhere for processing. It is needed for `plist` now

engine/dlib/src/test/wscript
```diff
@@ -128,9 +128,9 @@ def build(bld):
     if 'win32' in bld.env['PLATFORM']:
         skip_http_run = True
 
-    skip_ssdp = False
-    if 'win32' in bld.env['PLATFORM'] or sdk.is_wsl():
-        skip_ssdp = True
+    skip_mdns = False
+    # if 'win32' in bld.env['PLATFORM'] or sdk.is_wsl():
+    #     skip_mdns = True
```
AGulev: windows should work now

engine/engine/content/builtins/manifests/ios/Info.plist
```diff
@@ -115,6 +115,16 @@
...
 
+        {{^variant_release}}
+        <key>NSBonjourServices</key>
+        <array>
+                <string>_defold._tcp</string>
+        </array>
+
+        <key>NSLocalNetworkUsageDescription</key>
+        <string>Discover Defold targets on the local network.</string>
+        {{/variant_release}}
```
AGulev, line 126: since it is debug only funcitonality we need this debug only section

editor/src/clj/editor/targets.clj
```diff
@@ -168,122 +162,144 @@
   nil)
 
 (def ^:private update-targets-context
-  {:targets-atom ssdp-targets
+  {:targets-atom mdns-targets
    :log-fn log
-   :fetch-url-fn http-get
    :on-targets-changed-fn invalidate-target-menu!})
 
-(defn- device->target [{:keys [fetch-url-fn]} device]
```
AGulev, line 176: I provide log-posr in the TXT record to simplify this logic

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/BundleHelper.java
vlaaad: Should the template property instead be "variant" with "release"/"debug"/"headless" options?

editor/src/java/com/dynamo/discovery/MDNSServiceInfo.java
```diff
@@ -0,0 +1,79 @@
...
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+package com.dynamo.discovery;
+
+import java.util.Collections;
+import java.util.Map;
+
+public class MDNSServiceInfo {
```
vlaaad: Can we use Java records to avoid all this verbosity?

editor/test/editor/mdns_test.clj
```diff
@@ -0,0 +1,288 @@
...
+  ([^MDNS mdns ^bytes packet]
+   (parse-and-rebuild! mdns packet "127.0.0.1" "127.0.0.1"))
+  ([^MDNS mdns ^bytes packet ^String local-address ^String remote-address]
+   (invoke-private! mdns "parsePacket" [(Class/forName "[B") Integer/TYPE String String] packet (alength packet) local-address remote-address)
+   (invoke-private! mdns "rebuildDiscovered" [])))
+
+(defn- expire-and-rebuild!
+  [^MDNS mdns]
+  (invoke-private! mdns "expireEntries" [])
+  (invoke-private! mdns "rebuildDiscovered" []))
```
vlaaad: Is this really necessary? Usually, tests verify publicly observable behaviors, not private implementation details...

AGulev:
I tested 
Windows to win, android, iOS target
Linux to linux, android, iOS target
Mac to mac, android, iOS target

editor/src/java/com/dynamo/discovery/MDNS.java
```diff
@@ -0,0 +1,760 @@
...
+import java.util.Locale;
+import java.util.Map;
+
+public class MDNS {
+
+    public interface Logger {
+        void log(String msg);
+    }
+
+    private static class HostAddress {
```
vlaaad: Please use java records

editor/src/java/com/dynamo/discovery/MDNS.java
```diff
@@ -0,0 +1,760 @@
...
+            HostAddress hostAddress = new HostAddress();
+            hostAddress.address = address;
+            hostAddress.expires = expires;
+            hosts.put(hostKey, hostAddress);
+        }
+    }
+
+    private static void voidUnused(Object ignored) {
+        // Explicitly consume vars to silence static analyzers where applicable.
+    }
```
vlaaad: WAT... why not avoid unused objects altogether?

editor/src/java/com/dynamo/discovery/MDNSServiceInfo.java
```diff
@@ -0,0 +1,79 @@
...
+            MDNSServiceInfo s = (MDNSServiceInfo) other;
+            return serviceName.equals(s.serviceName)
+                    && address.equals(s.address)
+                    && (host != null ? host.equals(s.host) : s.host == null)
+                    && (localAddress != null ? localAddress.equals(s.localAddress) : s.localAddress == null)
+                    && port == s.port
+                    && txt.equals(s.txt);
+        }
+        return false;
+    }
```
vlaaad, line 61: `equals()` and `hashCode()` ignore id and instanceName. Intentional?

editor/src/java/com/dynamo/discovery/MDNS.java
```diff
@@ -0,0 +1,760 @@
...
+
+    private static int writeName(byte[] out, int offset, String name) {
+        String[] labels = name.split("\\.");
+        for (String label : labels) {
+            if (label.isEmpty()) {
+                continue;
+            }
+            out[offset++] = (byte) label.length();
+            byte[] bytes = label.getBytes();
```
vlaaad: Missing charset in `.getBytes`, may blow up on some weird user locales. Can you explain the encoding situation from the protocol POV?

editor/test/editor/mdns_test.clj
```diff
@@ -0,0 +1,288 @@
...
+          (recur (inc i)))))))
+
+(deftest mdns-discovers-service-from-dns-sd-records
+  (let [mdns (MDNS. (dummy-logger))
+        service-type MDNS/MDNS_SERVICE_TYPE
+        full-name (str "TargetOne." service-type)
+        host-name "target-host.local"
+        packet (make-response-packet (service-records service-type full-name host-name 8123 120))]
+    (parse-and-rebuild! mdns packet)
+    (let [devices ^"[Lcom.dynamo.discovery.MDNSServiceInfo;" (.getDevices mdns)]
```
vlaaad: For array type hints, we use `^com.dynamo.discovery.MDNSServiceInfo/1`

editor/test/editor/mdns_test.clj
```diff
@@ -0,0 +1,288 @@
...
+
+(deftest mdns-discovers-service-when-records-arrive-out-of-order
+  (let [mdns (MDNS. (dummy-logger))
+        service-type MDNS/MDNS_SERVICE_TYPE
+        full-name (str "TargetReordered." service-type)
+        host-name "target-reordered.local"
+        records (service-records service-type full-name host-name 8124 120)
+        packet (make-response-packet [(second records) (nth records 2) (nth records 3) (first records)])]
+    (parse-and-rebuild! mdns packet)
+    (let [devices ^"[Lcom.dynamo.discovery.MDNSServiceInfo;" (.getDevices mdns)]
```
vlaaad: `^com.dynamo.discovery.MDNSServiceInfo/1`

vlaaad (changes requested):
I think the changes are good and work fine. I'd like to see more tests of our MDNS implementation. I don't know the mDNS protocol enough to see which tests are missing, so I hope you can come up with some good examples here.


I find the equality semantics of `MDNSServiceInfo` puzzling, it's either a bug, or (probably) an intentional choice that needs to be documented and explained in code.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/BundleHelper.java
AGulev: no, we allow only basic mustache templater functionality which means options should be true or false (not having an option is `false`)

editor/test/editor/mdns_test.clj
AGulev: I'll avoid reflection here, but keep a hook of the `parsePacket` (but with a special class) for tests to make sure I can tests state/parser behaviour without real network (there is a real network test exist separatly). It's really hard to take into account all the cases using a real network and be sure that enviroment is stable enough so test will not endup being noisy

Some examples:
https://github.com/orgs/community/discussions/170669
https://github.com/orgs/community/discussions/148693

editor/src/java/com/dynamo/discovery/MDNS.java
AGulev: done

editor/src/java/com/dynamo/discovery/MDNSServiceInfo.java
AGulev: done

editor/src/java/com/dynamo/discovery/MDNS.java
AGulev: temp code I didn't noticed on final review. thx

editor/src/java/com/dynamo/discovery/MDNSServiceInfo.java
AGulev: yes, I added a comment

editor/test/editor/mdns_test.clj
AGulev: thanks!

editor/test/editor/mdns_test.clj
AGulev: Thanks!

editor/src/java/com/dynamo/discovery/MDNS.java
AGulev: now it is replcaced with a name created from prefix+ip+port e.g. `defold-192-168-0-2-8001`
so only ANCII expected here

editor/src/clj/editor/targets.clj
```diff
@@ -26,40 +26,49 @@
...
 (def ^:const max-log-entries 512)
 
 (defn- clear-selected-target-hint! []
   (reset! selected-target-atom ::undefined))
 
+(defn- destroy-launched-target-process! [^Process process]
+  (when (.isAlive process)
+    (.destroy process)))
+
 (defn kill-launched-target! [target]
```
AGulev, line 62: I updated this funcitonality to properly exit the instance using "exit" message we send to "@system", otherwise target will not be removed for 120s (max, it is TTL we use for our DNS entry). So it's not critical, just better user expirience in case of using `Cmd+J` to close instances

editor/bundle-resources/Info.plist
```diff
@@ -24,6 +24,12 @@
 	<string>????</string>
 	<key>NSHighResolutionCapable</key>
 	<true/>
+	<key>NSBonjourServices</key>
+	<array>
+		<string>_defold._tcp</string>
+	</array>
+	<key>NSLocalNetworkUsageDescription</key>
+	<string>Defold uses local network access to discover and connect to running Defold targets for debugging, profiling, hot reload, and log streaming on devices on your LAN.</string>
```
AGulev, line 32: When macOS show a popup that requests a local network access this text will be show to explain to a user why we request it.

AGulev (commented):
I think I fixed alll your findings, Vlad. Thank you!

vlaaad (dismissed):
Editor changes look good!

engine/dlib/src/dlib/mdns.h
```diff
@@ -0,0 +1,113 @@
...
+//
+// You may obtain a copy of the License, together with FAQs at
+// https://www.defold.com/license
+//
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#ifndef DM_MDNS
```
JCash: minor style: `DM_MDNS_H`

engine/dlib/src/dlib/mdns.h
```diff
@@ -0,0 +1,113 @@
...
+    {
+        RESULT_OK = 0,
+        RESULT_NETWORK_ERROR = -1,
+        RESULT_ALREADY_REGISTERED = -2,
+        RESULT_NOT_REGISTERED = -3,
+        RESULT_OUT_OF_RESOURCES = -4,
+        RESULT_INVALID_ARGS = -5
+    };
+
+    struct TxtEntry
```
JCash, line 36: style: In an api, we try to avoid unnecessary abbreviations. -> `TextEntry` is just fine.

engine/dlib/src/dlib/mdns.h
```diff
@@ -0,0 +1,113 @@
...
+    struct ServiceDesc
+    {
+        const char*      m_Id;
+        const char*      m_InstanceName;
+        const char*      m_ServiceType; // e.g. "_defold._tcp"
+        const char*      m_Host;        // optional, nullptr means auto
+        uint16_t         m_Port;
+        const TxtEntry*  m_Txt;
+        uint32_t         m_TxtCount;
+        uint32_t         m_Ttl;
```
JCash, line 51: style: readable/understandable variable names. -> "m_Text..."

engine/dlib/src/dlib/mdns.h
```diff
@@ -0,0 +1,113 @@
...
+        const char*     m_Id;
+        const char*     m_InstanceName;
+        const char*     m_ServiceType;
+        const char*     m_Host;
+        uint16_t        m_Port;
+        const TxtEntry* m_Txt;
+        uint32_t        m_TxtCount;
+    };
+
+    typedef void (*ServiceCallback)(void* context, const ServiceEvent* event);
```
JCash: style: Naming -> `FServiceCallback`

engine/dlib/src/dlib/mdns.cpp
```diff
@@ -0,0 +1,1632 @@
...
+
+    static bool EndsWithLocal(const char* value)
+    {
+        uint32_t n = (uint32_t) strlen(value);
+        static const char suffix[] = ".local";
+        uint32_t suffix_n = (uint32_t) strlen(suffix);
+        if (n < suffix_n)
+            return false;
+        return dmStrCaseCmp(value + n - suffix_n, suffix) == 0;
+    }
```
JCash, line 338: this looks a bit convoluted.
I think something like this should work:

```c++
const char* suffix = strrchr(value); // find the last "."
return suffix && dmStrCaseCmp(".local", suffix) == 0;
```

engine/dlib/src/dlib/mdns.cpp
```diff
@@ -0,0 +1,1632 @@
+// Copyright 2020-2026 The Defold Foundation
```
JCash, line 1: Overall I think this code could also use some documentation of some of the steps.
At the very least links to the specification/algorithm used etc.

engine/engine/src/engine_service.cpp
```diff
@@ -30,39 +29,84 @@
...
+    static void BuildServiceInstanceName(const char* local_address, const char* port_text, char* out, uint32_t out_size)
+    {
+        if (out_size == 0)
+            return;
+
+        char sanitized_address[128];
+        const char* source = (local_address && local_address[0]) ? local_address : "localhost";
+        uint32_t sanitized_length = 0;
+        bool last_was_dash = true;
+        while (*source && sanitized_length + 1 < sizeof(sanitized_address))
```
JCash: Style: Sanitizing a path seems like it deserves it's own function.

engine/engine/src/engine_service.cpp
```diff
@@ -30,39 +29,84 @@
...
 
     static const char INTERNAL_SERVER_ERROR[] = "(500) Internal server error";
     const char* const FOURCC_RESOURCES = "RESS";
 
+    static void BuildServiceInstanceName(const char* local_address, const char* port_text, char* out, uint32_t out_size)
+    {
+        if (out_size == 0)
+            return;
+
+        char sanitized_address[128];
```
JCash: Given all the constants or dynamic calculations, it's a bit surprising to find the 128 here?
Is there a constant to use, or can it be arbitrary?
If it's a dynamic string, perhaps use `dmAlloca` from our dalloca.h

engine/engine/src/engine_service.cpp
```diff
@@ -30,39 +29,84 @@
...
+
+        static const char prefix[] = "defold-";
+        const uint32_t max_label_length = dmMath::Min((uint32_t) out_size - 1, MDNS_MAX_LABEL_LENGTH);
+        const uint32_t prefix_length = (uint32_t) strlen(prefix);
+        const uint32_t suffix_length = (uint32_t) strlen(suffix);
+        const uint32_t max_address_length = prefix_length + suffix_length < max_label_length ? max_label_length - prefix_length - suffix_length : 0;
+
+        uint32_t address_length = (uint32_t) strlen(sanitized_address);
+        if (address_length > max_address_length)
+            address_length = max_address_length;
```
JCash: The check for truncation seems a bit convoluted.
Perhaps documentation would help reading the requirements?

Also, is it safe to truncate the address?
If it get's truncated, the "out" will be "defold-". Is that a valid address?

Sometimes, the StrlCpy + StrlCat becomes hard to read, and can be replaced with the pattern:
```c++
int length = dmSnPrintf(out, out_size, "defold-%s%s", sanitized_address, suffix);
if (length < 0)
{
    // handle truncation issues...
}

File: engine/engine/src/engine_service.cpp [hunk 24]
```diff
@@ -401,49 +384,40 @@ namespace dmEngineService
...
-            dmStrlCat(m_DeviceDesc.m_UDN, m_LocalAddress, sizeof(m_DeviceDesc.m_UDN));
-            dmStrlCat(m_DeviceDesc.m_UDN, ":", sizeof(m_DeviceDesc.m_UDN));
+            // DNS-SD instance labels are protocol identifiers, while the
+            // human-readable target name is published separately in TXT.
+            BuildServiceInstanceName(m_LocalAddress, m_PortText, m_ServiceInstanceName, sizeof(m_ServiceInstanceName));
+
+            // Service id must be unique and this scheme is probably unique enough.
+            dmStrlCpy(m_ServiceId, "defold-", sizeof(m_ServiceId));
+            dmStrlCat(m_ServiceId, m_LocalAddress, sizeof(m_ServiceId));
+            dmStrlCat(m_ServiceId, ":", sizeof(m_ServiceId));
```
JCash: Can be replaced with a single dmSnPrintf
```c++
dmSnPrintf(str, sizeof(str), "defold-%s:%s-%s", m_LocalAddress, m_PortText, info.m_DeviceModel);
```

engine/engine/src/engine_service_discovery.cpp
```diff
@@ -0,0 +1,172 @@
...
+            --offset;
+
+        if (offset == 0)
+        {
+            dmStrlCpy(out, "defold", out_size);
+            return;
+        }
+
+        if (out[0] == '-')
+            out[0] = 'd';
```
JCash: What's this? Perhaps deserves a code comment?

engine/engine/src/engine_service_discovery.cpp
```diff
@@ -0,0 +1,172 @@
...
+        out[offset] = 0;
+    }
+
+    HDiscoveryService DiscoveryServiceNew(const char* service_id, const char* instance_name, uint16_t port, const DiscoveryTxtEntry* txt_entries, uint32_t txt_count)
+    {
+        if (service_id == 0 || instance_name == 0 || port == 0 || txt_entries == 0)
+            return 0;
+
+        dmMDNS::TxtEntry mdns_txt_entries[16];
+        const uint32_t mdns_txt_entries_count = sizeof(mdns_txt_entries) / sizeof(mdns_txt_entries[0]);
```
JCash: Use `mdns_txt_entries_count = DM_ARRAY_SIZE(mdns_txt_entries)` from our array.h

engine/engine/src/engine_service_discovery.cpp
```diff
@@ -0,0 +1,172 @@
...
+        if (txt_count > mdns_txt_entries_count)
+        {
+            dmLogWarning("Unable to register mDNS service, too many TXT entries (%u)", txt_count);
+            return 0;
+        }
+
+        for (uint32_t i = 0; i < txt_count; ++i)
```
JCash, line 73: can txt_count be larger than the array size?
If so, this is a buffer overrun.

engine/engine/src/engine_service_discovery.cpp
```diff
@@ -0,0 +1,172 @@
...
+        {
+            mdns_txt_entries[i].m_Key = txt_entries[i].m_Key;
+            mdns_txt_entries[i].m_Value = txt_entries[i].m_Value;
```
JCash, line 76: no need to loop here?
A simple memcpy should work since they're both the same type?

JCash (changes requested):
overall good, but could use some cleanup

engine/dlib/src/dlib/mdns.h
AGulev: There are a few entry types in DNS: A/AAAA (IPv4/IPv6 address), CNAME (domain alias), MX (mail handling), TXT (text verification/email security), and NS (authoritative name server). Usually it's caled like that in all dashboards and documentation. I would prefer to keep it like this

engine/dlib/src/dlib/mdns.h
AGulev: the same, TXT, TTL are DNS related terms

engine/dlib/src/dlib/mdns.cpp
AGulev: done

engine/engine/src/engine_service.cpp
AGulev: you right, it's should be MDNS_MAX_LABEL_LENGTH + 1
(added comment there shy)

our label looks like `defold-192-168-0-12-8001` etc

engine/engine/src/engine_service_discovery.cpp
AGulev: added explanation + a test

engine/engine/src/engine_service_discovery.cpp
AGulev: no, I check it a few linew above

engine/engine/src/engine_service_discovery.cpp
AGulev: the layout of these two structs the same, but we have them as two separate structs to split logic of the implementation from the discovery service. I can use one type for both - couplic of the service with mdns implementatioon or just rely that layout stay the same - a bit shaky. 

it's jsut 16 elemnets top, not that big a perf. win here

AGulev:
@codex review it again

AGulev:
@codex review

AGulev:
Working on the review fixes, I found a few more things and incosistancies I decided to fix to make implementation less noisy for the local network, fix tiny issues, and make sure it's in align with specs

AGulev:
Now PR is ready for review again.

AGulev:
> I tested
> Windows to win, android, iOS target
> Linux to linux, android, iOS target
> Mac to mac, android, iOS target

retested on all the platforms again

engine/dlib/src/test/wscript
JCash: Then I suggest removing the commented out code.

JCash (dismissed):

editor/src/java/com/dynamo/discovery/MDNS.java
```diff
@@ -0,0 +1,1145 @@
...
+        // Query either when the exponential backoff fires or when a cached
+        // record reaches its half-TTL refresh point, whichever comes first.
+        long due = nextRefreshAt();
+        if (nextQueryAt == 0 || (due != 0 && due < nextQueryAt)) {
+            due = nextQueryAt == 0 ? due : Math.min(nextQueryAt, due);
+        } else if (due == 0) {
+            due = nextQueryAt;
+        }
+        return due == 0 || now >= due;
+    }
```
vlaaad, line 729: Is the logic here correct? I don't quite get it. It's supposed to use the earlier deadline (either `nextRefreshAt` or `nextQueryAt`), right? This code seems to use `nextRefreshAt` if `nextQueryAt` is earlier. Does this code look more correct to you: 
```java
private boolean shouldQuery(long now) {
    long refreshAt = nextRefreshAt();
    long due = nextQueryAt == 0 ? refreshAt
            : refreshAt == 0 ? nextQueryAt
            : Math.min(nextQueryAt, refreshAt);
    return due == 0 || now >= due;
}
```
?

editor/src/java/com/dynamo/discovery/MDNSServiceInfo.java
```diff
@@ -0,0 +1,61 @@
...
+                              String logPort,
+                              Map<String, String> txt) {
+
+    public MDNSServiceInfo {
+        txt = Collections.unmodifiableMap(txt);
+    }
+
+    // Equality is used to detect meaningful discovery changes in MDNS.
+    // id, instanceName, and logPort are derived from txt/serviceName, while
+    // expires is expected to change as records are refreshed.
```
vlaaad: Do we actually use `MDNSServiceInfo` instances as keys anywhere? If not, do we need these equality semantics?

editor/src/clj/editor/targets.clj
```diff
@@ -168,122 +173,145 @@
...
-(defn- search-interval [^SSDP ssdp]
-  (if (.isConnected ssdp)
+(defn- search-interval [^MDNS mdns]
+  (if (.isConnected mdns)
     search-interval-connected
     search-interval-disconnected))
 
-(defn- device->map [^DeviceInfo device]
-  {:address (.address device)
+(defn- device->map [^MDNSServiceInfo device]
```
vlaaad, line 273: `MDNSServiceInfo` is already an immutable value with proper equality semantics, do we need `device->map` still?

vlaaad (commented):
LGTM, but I have a couple of questions.

editor/src/java/com/dynamo/discovery/MDNS.java
AGulev: Nice one, thanks!
I'm fixing it and will add a test for it.

editor/src/java/com/dynamo/discovery/MDNSServiceInfo.java
AGulev: yes, see `rebuildDiscovered()` in `MDNS.java` - > `next.equals(discoveredServices)`

editor/src/clj/editor/targets.clj
AGulev: it's not like we "need" it, but it will be a big refactor all over the places to replace it. also, user may manually add targets using `Enter Target IP` which isn't the same as discoverable target, then it will endup only in this Clojure map

AGulev:
Thank you for the review, it has been addressed.

vlaaad (approved):
LGTM!

alexzheng:
target running in iOS device can not be connected after screen locked and unlocked again.
steps
1. install a contentless bundle on iOS device, I call it a defold player.
2. project->target, select the iOS target, cmd+B run on that target, it runs OK.
3. lock my device, the device will disappear from project->target, it's OK
4. unlock my device, the defold player is still running in foreground and it appear in the target list. 
5. select the iOS targer, cmd + B to run on that targer,it failed with a dialog popup:
Launching localhost-192.168.5.5-iPhone OS-192.168.5.5:8001 failed. If the engine is already running, shut down the process manually and retry.

vlaaad:
Please open an issue

PR #12092, opened 2026-03-23
Improved Scene View manipulator performance

### LINKED ISSUE #11828: Implement efficient scene view updates
Related to #10741

**Task (REQUIRED):**
We should implement dragging in the scene view as a single transaction with an ephemeral in-flight state on top for better performance. This state should apply to both the scene view and the properties view.

Scene and property views are updated live while dragging, while having a better performance (no transaction churn).

matgis:
We now defer committing changes to the graph until the drag operation has completed when using Scene View manipulators. This can significantly improve responsiveness when the manipulated objects are referenced from a large number of scenes. An unfortunate downside of this new approach is that you won't see the drag reflected across multiple open tabs in Split View until you complete the drag.

Fixes #11828.

### Technical changes
* The `WorkbenchView` associated with the active Editor Tab now receives the `selected-node-properties` from the `AppView` and has control of what is shown in the `PropertiesView` from its `displayed-node-properties` output.
* Manipulators now operate in two different phases:
  1. `:manip-phase/preview` maintains a map of `node-id->prop-kw->override-value` in the `SceneView` that will augment the `displayed-node-properties` while dragging. Overrides to transform properties are applied automatically. Overrides to other properties can augment the `visibility-aabb` and `user-data` associated with the `renderable` by specifying a `:preview-fn`.
  2. `:manip-phase/commit` concludes with applying a series of `tx-steps` to the graph.
* The `:render-key` of each `renderable` no longer takes the perspective projection into account. Objects now are sorted by the world-space distance from the camera along its Z axis, whereas previously they were sorted by what would have been their resulting depth in a 32-bit depth buffer. In practice, the distinction shouldn't matter, and it allows us to defer calculation of the camera clip planes until after we've flattened the scene.


editor/src/clj/editor/game_object.clj
```diff
@@ -244,34 +244,6 @@
...
-      (scene-tools/manip-scale evaluation-context embedded-resource-id delta)
-
-      (contains? (g/node-value node-id :transform-properties evaluation-context) :scale)
-      (scene/manip-scale-scene-node evaluation-context node-id delta)
-
-      :else
-      nil)))
-
-;; -----------------------------------------------------------------------------
-
```
matgis, line 274: We no longer have a need for this. It was only used for the Label component, which used to store the Scale in the `.label` resource. We already had migration code to move the scale value to the `ComponentDesc`.

editor/src/clj/editor/scene.clj
```diff
@@ -600,34 +682,15 @@
...
-                           (geom/empty-aabb? local-aabb))
-                     union-aabb
-                     (-> local-aabb
-                         (geom/aabb-transform world-transform)
-                         (geom/aabb-union union-aabb)))]
-    (reduce (fn [^AABB union-aabb child-scene]
-              (calculate-scene-aabb union-aabb world-transform child-scene))
-            union-aabb
-            (:children scene))))
-
```
matgis, line 623: Now happens when we flatten the scene as part of the `scene-render-data` output.

editor/src/clj/editor/scene_tools.clj
```diff
@@ -25,38 +26,51 @@
...
+(s/def ::prop-kw keyword?)
+(s/def ::prop-value any?)
+(s/def ::tx-data g/tx-data?)
+(s/def ::prop-kw->override-value (s/map-of ::prop-kw ::prop-value))
+(s/def ::node-id->prop-kw->override-value (s/map-of ::node-id ::prop-kw->override-value))
+(s/def :manip/tx-data ::tx-data)
+(s/def :manip/node-id ::node-id)
+(s/def :manip/prop-kw->override-value ::prop-kw->override-value)
+(s/def ::manipulation (s/keys :req [(or :manip/tx-data
+                                        :manip/prop-kw->override-value)]))
```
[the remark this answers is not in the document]
matgis: The `or` clause is actually interpreted by the `s/keys` macro as "we require either key to be present" in this case. See: https://clojuredocs.org/clojure.spec.alpha/keys

editor/src/clj/editor/scene_tools.clj
```diff
@@ -479,79 +497,133 @@
...
                                   (= (:button action) :primary))
-                         (g/transact
-                            (concat
-                              (g/set-property self :start-action action)
-                              (g/set-property self :prev-action action)
-                              (g/set-property self :original-values original-values)
-                              (g/set-property self :initial-evaluation-context (atom evaluation-context))
-                              (g/set-property self :active-manip manip)
-                              (g/set-property self :hot-manip nil)
-                              (g/set-property self :op-seq (gensym)))))
```
matgis, line 526: A lot of these properties have been combined into a single `drag-start-state` property.

editor/test/editor/scene_test.clj
```diff
@@ -0,0 +1,221 @@
...
+                       {:node-id :house-node-id
+                        :node-outline-key "house-node-outline-key"
+                        :picking-node-id :new-node-id
+                        :children [{:node-id :door-node-id
+                                    :node-outline-key "door-node-outline-key"
+                                    :picking-node-id :new-node-id
+                                    :children [{:node-id :door-handle-node-id
+                                                :node-outline-key "door-handle-node-outline-key"
+                                                :picking-node-id :new-node-id}]}]}]}
+           (scene/claim-scene scene :new-node-id "new-node-outline-key")))))
```
matgis, line 75: Moved here from `integration.scene-test`, since it is more of a unit test.

editor/test/integration/scene_test.clj
```diff
@@ -24,45 +24,47 @@
...
+
+                 "/switcher/switcher.atlas"
+                 (fn [_node-id view-id]
+                   (is (= (g/node-value view-id :scene-aabb) (geom/coords->aabb [0 0] [2048 1024]))))}]
+
+      (test-util/with-loaded-project
+        (doseq [[path test-fn] cases]
+          (let [[node-id view-id] (test-util/open-scene-view! project app-view path 128 128)]
+            (is (not (nil? node-id)) (format "Could not find '%s'" path))
+            (test-fn node-id view-id)))))))
```
matgis, line 67: Mostly indentation, but we now evaluate the `scene-aabb` output instead of calculating the resulting AABB separately.

editor/test/integration/scene_test.clj
```diff
@@ -263,43 +307,6 @@
...
-                        :node-outline-key "house-node-outline-key"
-                        :picking-node-id :new-node-id
-                        :children [{:node-id :door-node-id
-                                    :node-outline-key "door-node-outline-key"
-                                    :picking-node-id :new-node-id
-                                    :children [{:node-id :door-handle-node-id
-                                                :node-outline-key "door-handle-node-outline-key"
-                                                :picking-node-id :new-node-id}]}]}]}
-           (scene/claim-scene scene :new-node-id "new-node-outline-key")))))
-
```
matgis, line 302: Moved to `editor.scene-test` since it is more of a unit test than an integration test.

editor/src/clj/editor/app_view.clj
```diff
@@ -483,6 +486,8 @@
                  (concat
                    (g/set-property app-view :active-tab new-active-tab)
                    (replace-connection basis new-resource-node-id :node-outline app-view :active-outline)
+                   (when properties-view
+                     (replace-connection basis new-view-node-id :displayed-node-properties properties-view :displayed-node-properties))
```
vlaaad, line 490: Is it possible to have no properties view?

vlaaad (approved):
LGTM!

editor/src/clj/editor/app_view.clj
matgis: Not sure if it is a real-world concern, but you can certainly create an `AppView` node without connecting anything to the `properties-view` input. 🤔

vlaaad: Wouldn't that be an invalid use of AppView?

PR #12095, opened 2026-03-24
Fix Box2D 3.1 trigger enter/exit when one collision object overlaps several triggers

### LINKED ISSUE #12093: Box2D 3.1 triggers report wrong enter/exit
When using the new Box2D 3.1 physics engine in Defold, multiple triggers colliding with each other report incorrect enter/exit events.

**To Reproduce (REQUIRED)**
1. Launch the project
2. Observe the log output - you will get a message indicating an exit from `/go3`, although the trigger is still overlapping with the other trigger:

```
DEBUG:SCRIPT: go2_enter url: [main:/go2#collisionobject] hash: [/go3]
DEBUG:SCRIPT: go2_enter url: [main:/go2#collisionobject] hash: [/go1]
DEBUG:SCRIPT: go2_exit url: [main:/go2#collisionobject] hash: [/go1]
DEBUG:SCRIPT: go2_exit url: [main:/go2#collisionobject] hash: [/go3]
```

The `enter=true` and `enter=false` (exit) events should be triggered exactly at the moments when triggers start or stop overlapping, and in the correct order. In the described scenario, only three events should be reported:

```
DEBUG:SCRIPT: go2_enter	url: [main:/go2#collisionobject]	hash: [/go3]
DEBUG:SCRIPT: go2_enter	url: [main:/go2#collisionobject]	hash: [/go1]
DEBUG:SCRIPT: go2_exit	url: [main:/go2#collisionobject]	hash: [/go1]
```

* Version: 1.12.2, 1.12.3 BETA

**Platforms (REQUIRED):**
* Platforms: All

[trigger bug.zip](https://github.com/user-attachments/files/26195206/trigger.bug.zip)

Related:
* https://github.com/defold/defold/issues/10606

aglitchman:
2D physics with Box2D 3.1 could report wrong trigger enter/exit when one object overlapped more than one trigger at the same time. Exits were applied incorrectly to all overlaps of a body instead of only the pair that ended. This change fixes that tracking and adds a regression test. Sensor overlap queries used for collision callbacks also read overlap counts from Box2D correctly.

Fixes https://github.com/defold/defold/issues/12093

### Technical changes

* `GetSensorOverlapBuffer`: treat `b2Shape_GetSensorCapacity` as capacity only; call `b2Shape_GetSensorOverlaps`, then set scratch buffer size to the returned overlap count.  
* `OverlapCacheDecreaseCount` now takes `(object_a, object_b)` and decrements the overlap from **a -> b** only; We call it twice with arguments swapped so both directed edges stay consistent.  
* New test: `TriggerThreeOverlapping` (two overlapping triggers + third body entering/leaving both - the same as the re-pro case from the issue).

### DIFF
--- a/engine/physics/src/physics/box2d/box2d_physics.cpp
+++ b/engine/physics/src/physics/box2d/box2d_physics.cpp
@@ -269,19 +269,19 @@ namespace dmPhysics
 
     static dmArray<b2ShapeId>& GetSensorOverlapBuffer(HWorld2D world, b2ShapeId shapeId)
     {
-        int num_overlaps = b2Shape_GetSensorCapacity(shapeId);
+        int capacity = b2Shape_GetSensorCapacity(shapeId);
 
-        if (world->m_GetSensorOverlapsScratchBuffer.Capacity() < num_overlaps)
+        if (world->m_GetSensorOverlapsScratchBuffer.Capacity() < capacity)
         {
-            world->m_GetSensorOverlapsScratchBuffer.SetCapacity(num_overlaps);
+            world->m_GetSensorOverlapsScratchBuffer.SetCapacity(capacity);
         }
 
+        int num_overlaps = 0;
+        if (capacity > 0)
+        {
+            num_overlaps = b2Shape_GetSensorOverlaps(shapeId, world->m_GetSensorOverlapsScratchBuffer.Begin(), capacity);
+        }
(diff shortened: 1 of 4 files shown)


aglitchman:
Sorry for the force-pushes - commited with the wrong email first.

Jhonnyg (approved):
LGTM!

JCash (approved):

PR #12101, opened 2026-03-25
Add module size and build ID to crash dumps

### LINKED ISSUE #12006: Add size and path to crash.modules()
It's hard to debug engine or native extension crashes without good tools to collect and symbolicate them. Services like [Backtrace.io](https://backtrace.io) are made for this, but the information in `_crash` isn't sufficient to implement a full integration with Backtrace.io's stack symbolication.

I would like to add two fields to the structure returned by `crash.modules()`: `size` and `path`, with the size of the mapped module and the full path to it, rather than simply the file name. These fields are easy to fill in from data already at hand in `engine/crash/src/load_addrs_*.cpp` and they make it easy for Lua code to integrate with Backtrace.io or similar services. Backtrace.io in particular refuses to symbolicate without both module addresses and sizes. The full path to loaded images allows integration code to read those images and extract additional info (like Build ID) that may be needed for symbolication or debugging.

Without Defold's `crash` module adding this info, integration with a crash dump service would need to be implemented as a native extension. That seems overcomplicated when the existing code already loads most of the information.

It looks like module size may have been planned to be included already, since `struct AppStateV3` includes `m_ModuleSize`, which is already filled in on Windows. Full paths would require creating `struct AppStateV4`. That type would be 128KiB larger than V3 if we continue to use a fixed-size struct and the existing constant values for module number and file path length.

mstetson:
Crash dumps now contain enough information for automatic symbolication by tools like Backtrace.io. This required adding size and build_id entries to the crash.get_modules table.

Closes #12006

### Technical changes
* ELF, Mach-O, and PE files already include size and build ID information. This change finds that info in memory and stores it in the crash dump.
* On Linux & Android, /proc/self/smaps is no longer used to find modules. The information that was being loaded from that file is available in memory via the same API needed to find the build ID.
* The ELF headers do not contain the image name for the main executable. For functions like `backtrace_symbols`, the C library saves a copy of `argv[0]` and uses that for the module name when the module name is blank, as it's only blank for the main executable. In `load_addrs_elf.cpp`, I've set the main module name to "dmengine", but if you want it to have the actual executable name, `argv[0]` will need to be made available to the crash code.

### Outstanding issues
* I haven't been able to build on Windows yet. I don't have a Windows dev machine to hand, so the changes in load_addrs_win32.cpp are likely buggy.
* I don't have the ability to test on iOS or Android, but I believe the Mach-O and ELF support should work just fine there.

## PR checklist

* [x] Code
	* [ ] Ensure all CI tests pass
* [x] Documentation
	* [x] Make sure that manuals are updated (in github.com/defold/doc) **no update needed**
	* [x] Pull request - Link the pull request to the issue(s) it is closing. Use one of the [approved closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
	* [x] Affected issue - Assign the "breaking change" label to the issue if introducing a breaking change. **not a breaking change**
	* [x] Affected issue - Assign the "skip release notes" is the issue should not be included in the generated release notes. **include in release notes**


engine/crash/src/load_addrs_mach.cpp
```diff
@@ -14,29 +14,60 @@
...
+            {
+                struct uuid_command* uuid = (struct uuid_command*)cmd;
+                uuid_unparse_upper(uuid->uuid, state->m_ModuleBuildID[mod]);
+            }
+            else if (cmd->cmd == LC_SEGMENT_64)
+            {
+                struct segment_command_64* seg = (struct segment_command_64*)cmd;
+                if (strcmp(seg->segname, "__TEXT") != 0)
+                {
+                    state->m_ModuleSize[mod] = (uint32_t)seg->vmsize;
```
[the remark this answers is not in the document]
mstetson: This was a typo. Please recheck.

engine/crash/src/load_addrs_elf.cpp
```diff
@@ -0,0 +1,109 @@
...
+    static int eachModule(struct dl_phdr_info* info, size_t size, void* data)
+    {
+        AppState* state = (AppState*)data;
+        uint32_t  count = state->m_ModuleCount;
+        if (count == AppState::MODULES_MAX)
+        {
+            return 0;
+        }
+        state->m_ModuleCount = count + 1;
+        state->m_ModuleAddr[count] = (void*)info->dlpi_addr;
```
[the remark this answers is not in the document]
mstetson: I've fixed this, but the number of times I've iterated on this ELF code is starting to make me nervous. I begin to wonder if it would be better to go back to using /proc/self/maps except for the build ID itself. I've been using the man pages and commentary from Grok AI to check my work here, but I'm not an expert in this area. It would be good to have a human more experienced with ELF loading review the code.

engine/crash/src/load_addrs_elf.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include <dlib/dstrings.h>
+#include "crash_private.h"
+
+#define NT_GNU_BUILD_ID 3
+
+namespace dmCrash
+{
+    static const char* hexdigit = "0123456789ABCDEF";
+
+    static bool  isBuildID(ElfW(Nhdr) * note)
```
JCash: Code style: -> `IsBuildID()`

engine/crash/src/load_addrs_elf.cpp
```diff
@@ -0,0 +1,115 @@
...
+    {
+        return (note->n_type == NT_GNU_BUILD_ID &&
+                note->n_descsz != 0 &&
+                note->n_namesz == 4 &&
+                memcmp((char*)(note + 1), "GNU", 4) == 0);
+    }
+
+    static void copyBuildIDNote(char* id, struct dl_phdr_info* info)
```
JCash: Code style: `CopyBuildIDNote()`

Also, I'd like to see an argument "uint32_t id_len" to make sure we don't write out-of-bounds.

engine/crash/src/load_addrs_elf.cpp
```diff
@@ -0,0 +1,115 @@
...
+            {
+                ElfW(Nhdr)* note = (ElfW(Nhdr)*)(note_addr);
+                if (isBuildID(note))
+                {
+                    uint8_t* build_id = (uint8_t*)(note+1) + DM_ALIGN(note->n_namesz, 4);
+                    ElfW(Word) len = note->n_descsz;
+                    if (len >= AppState::MODULE_BUILD_ID_SIZE / 2)
+                    {
+                        len = AppState::MODULE_BUILD_ID_SIZE / 2 - 1;
+                    }
```
JCash: If we need to clamp on this, isn't the premise wrong?
E.g. we have a struct that doesn't fit the data you needed to begin with?

engine/crash/src/load_addrs_elf.cpp
```diff
@@ -0,0 +1,115 @@
...
+                    return;
+                }
+                note_addr += (sizeof(*note) +
+                         DM_ALIGN(note->n_namesz, 4) +
+                         DM_ALIGN(note->n_descsz, 4));
+            }
+        }
+    }
+
+    static int eachModule(struct dl_phdr_info* info, size_t size, void* data)
```
JCash: Style...

engine/crash/src/load_addrs_mach.cpp
```diff
@@ -14,29 +14,60 @@
...
 #include "crash_private.h"
 
+#include <mach-o/loader.h>
 #include <mach-o/dyld.h>
+#include <uuid/uuid.h>
 #include <dlib/dstrings.h>
 
 namespace dmCrash
 {
+    static void walkMachCommands(AppState* state, uint32_t mod, const struct mach_header* hdr)
```
JCash: Code style.

engine/crash/src/test/test_crash.cpp
```diff
@@ -103,15 +103,20 @@ TEST_F(dmCrashTest, TestLoad)
     {
         const char *name = dmCrash::GetModuleName(d, i);
         void *addr = dmCrash::GetModuleAddr(d, i);
+        uint32_t size = dmCrash::GetModuleSize(d, i);
+        const char *buildid = dmCrash::GetModuleBuildID(d, i);
```
JCash, line 107: Codex: ```Stop requiring build_id for every loaded module in the test — /Users/mathiaswesterdahl/work/defold/engine/crash/src/test/test_crash.cpp:108-
    114
    This assertion turns an optional field into a hard requirement: GetModuleBuildID() intentionally returns nil when a module has no embedded build
    ID, which is common for system DLLs and some third-party binaries. In those cases TestLoad will fail even though the public API documents build_id
    as "if available".```

engine/crash/src/test/test_crash.cpp
```diff
@@ -103,15 +103,20 @@ TEST_F(dmCrashTest, TestLoad)
...
         if (!name)
         {
             break;
         }
 
         ASSERT_NE((void*) 0, addr);
+        ASSERT_NE((uint32_t) 0, size);
+        ASSERT_NE((char*) 0, buildid);
```
JCash: Like code, mentions, are we guarantueed that our platforms always returns a valid string? 🤔

engine/crash/src/load_addrs_win32.cpp
```diff
@@ -14,39 +14,75 @@
...
 
+#include <dlib/dstrings.h>
 #include <dlib/log.h>
 #include <windows.h>
 #include <psapi.h>
 
 namespace dmCrash
 {
-    void SetLoadAddrs(AppState* state)
+    static copyPdbGuid(char* id, HMODULE hModule)
```
JCash: Missing return type.

Code style

engine/crash/src/load_addrs_win32.cpp
```diff
@@ -14,39 +14,75 @@
...
+        IMAGE_DOS_HEADER*    dos = (IMAGE_DOS_HEADER*)hModule;
+        IMAGE_NT_HEADERS*    nt = (IMAGE_NT_HEADERS*)((BYTE*)hModule + dos->e_lfanew);
+        IMAGE_DATA_DIRECTORY debugDir = nt->OptionalHeader.DataDirectory[IMAGE_DIRECTORY_ENTRY_DEBUG];
+
+        if (debugDir.VirtualAddress == 0 || debugDir.Size == 0)
+        {
+            return;
+        }
+
+        IMAGE_DEBUG_DIRECTORY* debugEntries = (IMAGE_DEBUG_DIRECTORY*)((BYTE*)hModule + debugDir.VirtualAddress;
```
JCash: Missing parenthesis.

engine/crash/src/load_addrs_win32.cpp
```diff
@@ -14,39 +14,75 @@
...
+            {
+                dmLogWarning("Failed to read name of module");
+            }
+            MODULEINFO info;
+            if (GetModuleInformation(process, mods[i], &info, sizeof(info)))
+            {
+                state->m_ModuleAddr[i] = info.lpBaseOfDll;
+                state->m_ModuleSize[i] = info.SizeOfImage;
+            }
+            copyPdbGuid(state->m_ModuleBuildID[i], mods[i])
```
JCash: Missing `;`

engine/crash/src/script_crash.cpp
```diff
@@ -175,12 +182,24 @@ namespace dmCrash
...
+                lua_settable(L, -3);
+            }
+
             char str[64];
             dmSnPrintf(str, sizeof(str), "%p", addr);
             lua_pushliteral(L, "address");
             lua_pushstring(L, str);
             lua_settable(L, -3);
 
+            dmSnPrintf(str, sizeof(str), "%d", size);
```
JCash: Format: `%u`

engine/crash/src/script_crash.cpp
```diff
@@ -175,12 +182,24 @@ namespace dmCrash
...
+            lua_pushliteral(L, "size");
+            lua_pushstring(L, str);
+            lua_settable(L, -3);
```
JCash: While this is in style with the old code, nowadays we promote the shorter, and more easily readable `lua_setfield`:
```c
lua_pushstring(L, str);
lua_setfield(L, -2, "size");
```

engine/crash/src/load_addrs_elf.cpp
```diff
@@ -0,0 +1,115 @@
...
+//
+// You may obtain a copy of the License, together with FAQs at
+// https://www.defold.com/license
+//
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#include <link.h>
```
JCash, line 15: I like the shorter list of includes!

JCash (changes requested):
Overall good, but needs a few fixes
* Compile issues
* Code style

engine/crash/src/load_addrs_elf.cpp
mstetson: A correct ELF build ID note is either a 16-byte MD5/UUID or a 20-byte SHA1, but the format allows an arbitrary length specified in `note->n_descsz`. The division by 2 is because the note is binary, but the output is a hex string, so 2 bytes out for every byte in. Since `AppState::MODULE_BUILD_ID_SIZE` is 64, we can accommodate either correct format (32+1 or 40+1 output bytes), but we need to truncate in case a binary's `n_descsz` is large for some reason.

engine/crash/src/test/test_crash.cpp
mstetson: I'm not certain we will always get a valid string. I believe we will, but I have no guarantee, so I've altered the test to allow a missing build id.

JCash (approved):

JCash:
I prematurely merged this as I misread the "checks passed" after I have assigned our "ok to test" label.

There are some issues:
* armv7-android NDK api 19 is not supported by your PR. This is still a very important platform to support.
* Windows doesn't compile

See build logs here:
https://github.com/defold/defold/actions/runs/25275211087/job/74103921440

PR #12102, opened 2026-03-25
Editor: Fix duplicate atlas image animation

### LINKED ISSUE #7403: Using the same image inside an atlas with different Sprite Trim modes results in duplicates in the atlas image and inability to select the duplicate.
### Expected behaviour
When copying an image frame inside an animation and applying a different sprite trim mode the atlas image should remain the same since no extra images are used.

### Actual behaviour
When copying an image frame inside an animation and applying a different sprite trim mode a duplicate frame is added to the atlas image. It results in an increase of the build size and inability to select the duplicate image inside the atlas layout view.

### Steps to reproduce
Copy a frame inside an animation, applying a different sprite trim mode.
Here is a video of the issue.
https://www.youtube.com/watch?v=3K8HpqlY3_A

<hr/>

<table>
<tr><td>Build time</td><td>2023-02-18T08:42:50.880700</td></tr>
<tr><td>Defold editor sha</td><td>8eaab6b1281ce492163428e0e7b2e0fa247a0a93</td></tr>
<tr><td>Defold engine sha</td><td>8eaab6b1281ce492163428e0e7b2e0fa247a0a93</td></tr>
<tr><td>Defold version</td><td>1.4.3</td></tr>
<tr><td>GPU Driver</td><td>2.1 Metal - 83</td></tr>
<tr><td>Java version</td><td>11.0.15+10-LTS</td></tr>
<tr><td>OS arch</td><td>x86_64</td></tr>
<tr><td>OS version</td><td>13.1</td></tr>
<table>

JosephFerano:
Fixed a bug where duplicate atlas image animations would be created with different properties (pivot and/or trim mode) and the sprite would not be interactable in the scene view. Also fixed the fact that it wasn't triggering a re-generation.

Fixes #7403


editor/src/clj/editor/pipeline/texture_set_gen.clj
```diff
@@ -153,10 +153,7 @@
 
 (defn atlas->texture-set-data
   [animations images margin inner-padding extrude-borders max-page-size]
-  ;; NOTE: Images order matters when generating the layouts, especially if they are of the same size.
-  ;; Since the SHA1 for the packed-page-images-generator is insensitive to order, things could get out of sync
-  (let [images (sort-by #(-> % :path resource/proj-path) images)
```
JosephFerano, line 158: I felt like moving this sort for when we produce images might be better than doing it randomly here while everywhere else consumes a non-sorted version.

editor/src/clj/editor/atlas.clj
```diff
@@ -615,7 +616,7 @@
          :args augmented-args})))
 
 (g/defnk produce-packed-page-images-generator
-  [_node-id extrude-borders image-resources inner-padding margin layout-data-generator max-page-size texture-page-count]
+  [_node-id extrude-borders image-resources all-atlas-images inner-padding margin layout-data-generator max-page-size texture-page-count]
```
JosephFerano, line 618: This fnk get's re-evaluated with or without all-atlas-images.

editor/src/clj/editor/atlas.clj
```diff
@@ -780,8 +791,15 @@
...
                                   (tex-gen/match-texture-profile texture-profiles (resource/proj-path resource))))
 
-  (output all-atlas-images [Image] :cached (g/fnk [animation-images]
-                                             (into [] (distinct) (flatten animation-images))))
+  (output all-atlas-images [Image] :cached
+          (g/fnk [animation-images]
+            ;; NOTE: Images order matters when generating the layouts, especially if they are of the same size.
+            ;; Since the SHA1 for the packed-page-images-generator is insensitive to order, things could get out of sync
+            ;; If not, you will get rects rendered laying on top of the wrong atlas images
+            (vec (sort-by #(-> % :path resource/proj-path)
```
JosephFerano, line 800: Is there a way to get this sort-by to be performed during the coll/into->? I read that "It cannot be used directly as a transducer".

editor/src/clj/editor/pipeline/texture_set_gen.clj
```diff
@@ -181,12 +178,18 @@
                               (reset! anim-imgs-atom [])))
             rects (mapv texture-set-layout-rect images)
             use-geometries (if (every? #(= :sprite-trim-mode-off (:sprite-trim-mode %)) images) 0 1)
+            image-keys (mapv (fn [img]
+                               [(resource/proj-path (:path img))
+                                [(:pivot-x img) (:pivot-y img)]
+                                (:sprite-trim-mode img)])
```
matgis: We should probably have a `make-image-key` function since we also construct a `image-key` for lookup later in the file?

editor/src/clj/editor/atlas.clj
matgis: Transducers process items one-by-one. There's not really a way to make a `sort` transducer without realizing the entire sequence before passing it on to the next steps. Initially I thought you could use a `(sorted-set)` with a custom comparator for the `into` target, but you have different criteria for the sort key and distinct key here, so probably better to leave it as-is.

matgis (changes requested):
I think this change fixes the wrong bug? The report states that the expected behavior is that the packed image should only need to contain the pixels for a particular source image once. Two animations that refer to the same source image should be able to share the same pixel data in the packed image, but have individual vertex data for the trimmed geometry.

I'm not entirely sure the code we share with Bob supports this scenario at the moment, though. 🤔

I think you fixed a bug with the viewport selection behavior, but I think we probably shouldn't include the `pivot` coordinates for consideration when we refresh the packed image. They should only really affect the selection and outline passes in the `scene` output, if anything.

Reviewing the data dependencies, I feel that there's a few strange things going on in `atlas.clj` already before your changes. 😞

JosephFerano:
@matgis 
> I think this change fixes the wrong bug? The report states that the expected behavior is that the packed image should only need to contain the pixels for a particular source image once. Two animations that refer to the same source image should be able to share the same pixel data in the packed image, but have individual vertex data for the trimmed geometry.
>
> I'm not entirely sure the code we share with Bob supports this scenario at the moment, though. 🤔
>
> I think you fixed a bug with the viewport selection behavior...

Thank you for reviewing the changes.

I tried to fix the bug based on the behavior we are currently supporting, if we would like to support avoiding the duplicate when the trim mode changes, I could either keep working in this same branch, or merge it without closing this issue and then starting a new PR to close this one.

I have two questions if we were to remove duplication;
1.) Should both the pivot and trim mode avoid duplicates? Meaning, changing either property does not trigger a duplicate.
2.) How could we go about resolving which animation image to select if we try to click on it through the scene view. If it's just a trim mode change, then clicking outside the trimmed area should select the one behind, but I'd have to review the code to ensure we can sort that correctly. If it's only the pivot mode, then perhaps a modifier key while clicking could do the job of cycling.

> but I think we probably shouldn't include the pivot coordinates for consideration when we refresh the packed image. They should only really affect the selection and outline passes in the scene output, if anything.

Changing the pivot only triggers a regen when an animation image is duplicated. This is a good test to include, so I just added it.

> Reviewing the data dependencies, I feel that there's a few strange things going on in atlas.clj already before your changes. 😞

That's unfortunate, let me know if you would like to discuss further on a call.

matgis:
> 1.) Should both the pivot and trim mode avoid duplicates? Meaning, changing either property does not trigger a duplicate.

Since the pixels added for each instance would be identical, neither property discrepancy should cause duplicate pixels to be added to the packed image.

> 2.) How could we go about resolving which animation image to select if we try to click on it through the scene view.

It's a good question. Ideally, all the overlapping images would be selected if I click on pixels that are shared between them. It might be tricky to achieve this, however. I think as a first iteration, just selecting the topmost one would be fine.

To me, this situation feels a bit contrived. Why would I want to have multiple instances of the same image with different pivot points and sprite trim settings? 🤔

Trouble it, you immediately run into this conflict if you have edited these properties on an `AtlasImage` and then add that same source `ImageNode` to another `AtlasAnimation` in the same `AtlasNode`. The new `AtlasImage` created for the twice-referenced `ImageNode` will have defaults that differ from the existing `AtlasImage`.

If I were to start over, I think I would model this as having one unique `AtlasImage` per source `ImageNode` so that you edit these properties for all instances of a source image, and then have the `AtlasAnimation` simply reference these images.

JosephFerano:
> Since the pixels added for each instance would be identical, neither property discrepancy should cause duplicate pixels to be added to the packed image.

Roger, I'll take a look at bob first and make sure we can get it to support this.

> It's a good question. Ideally, all the overlapping images would be selected if I click on pixels that are shared between them. It might be tricky to achieve this, however. I think as a first iteration, just selecting the topmost one would be fine.

:+1: 

> To me, this situation feels a bit contrived. Why would I want to have multiple instances of the same image with different pivot points and sprite trim settings? 🤔

It does seem contrived. Even in the user's video, taking a look at it more carefully, I get the impression that they just recorded the video to report the edge case, not necessarily displaying an issue in their actual workflow. They are free to correct me of course :)

> Trouble it, you immediately run into this conflict if you have edited these properties on an `AtlasImage` and then add that same source `ImageNode` to another `AtlasAnimation` in the same `AtlasNode`. The new `AtlasImage` created for the twice-referenced `ImageNode` will have defaults that differ from the existing `AtlasImage`.
>
> If I were to start over, I think I would model this as having one unique `AtlasImage` per source `ImageNode` so that you edit these properties for all instances of a source image, and then have the `AtlasAnimation` simply reference these images.

Yes, if I understand correctly, this would eliminate the ability to set different properties for an `AtlasImage` that references the same `ImageNode`, so if I modify the properties of an `AtlasImage` that is duplicated for both "AnimA" and "AnimB", it should update the properties for both `AtlasAnimation`s. If this is the more correct solution, I'll try to make it this way first.

JosephFerano:
Going to close this in favor of a new PR to create a more appropriately named branch name and undo all the commits.

PR #12107, opened 2026-03-25
Exclude unused code from the HTML5 build to reduce binary size

AGulev:
Even though `http_service.cpp` and `provider_http.cpp` were unused in HTML5, they were accidentally included in the HTML5 binary. 

Excluding them reduces the HTML5 release build from 2 604 862 bytes to 2 469 942 bytes, saving 134 920 bytes (~5.2%)

### DIFF
--- a/engine/dlib/src/dlib/configfile.cpp
+++ b/engine/dlib/src/dlib/configfile.cpp
@@ -23,15 +23,12 @@
 #include "uri.h"
 
 #include "configfile.h"
+#include "configfile_http.h"
 #include "array.h"
 #include "dstrings.h"
 #include "math.h"
 #include "sys.h"
 #include "static_assert.h"
-
-#if !defined(__EMSCRIPTEN__)
-#include "http_client.h"
-#endif
 
 struct ConfigFileEntry
 {
@@ -503,51 +500,6 @@ namespace dmConfigFile
         }
     }
 
-#if !defined(__EMSCRIPTEN__)
-    struct HttpContext
-    {
-        HttpContext()
-        {
-        }
-        dmArray<char> m_Buffer;
-    };
-
-    static void HttpHeader(dmHttpClient::HResponse response, void* user_data, int status_code, const char* key, const char* value)
-    {
-        (void) response;
-        (void) user_data;
-        (void) status_code;
-        (void) key;
-        (void) value;
-    }
-
-    static void HttpContent(dmHttpClient::HResponse response, void* user_data, int status_code, const void* content_data, uint32_t content_data_size, int32_t content_length,
-                            uint32_t range_start, uint32_t range_end, uint32_t document_size,
-                            const char* method)
-    {
-        (void) method; // unused
-        HttpContext* context = (HttpContext*) user_data;
-        if (status_code != 200)
-            return;
-
-        if (!content_data && !content_data_size)
-        {
-            context->m_Buffer.SetSize(0);
-            return;
-        }
-
-        dmArray<char>& buffer = context->m_Buffer;
-        if (buffer.Remaining() < content_data_size)
-        {
-            uint32_t new_capacity = buffer.Capacity() + dmMath::Max(4U * 1024U, content_data_size);
-            context->m_Buffer.SetCapacity(new_capacity);
-        }
-
-        assert(content_data);
-        buffer.PushArray((const char*) content_data, content_data_size);
-    }
-#endif
-
     static Result LoadFromBufferInternal(const char* url, const char* buffer, uint32_t buffer_size, int argc, const char** argv, HConfig* config)
     {
         Context context;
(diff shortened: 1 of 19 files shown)


engine/engine/src/engine.cpp
```diff
@@ -71,6 +69,9 @@
 
 #if defined(__EMSCRIPTEN__)
     #include "engine_web.h"
+#else
+	#include <dlib/http_client.h>
+	#include <dlib/sslsocket.h>
 #endif
```
JCash, line 72: In the future, I think we should use clearer "feature" defines, like e.g. `DM_HAVE_HTTP_SUPPORT` and so on.

AGulev:
I tried to avoid using even header files, so I fully excluded all the calls for emscripten, if you think it is too dirty, I'll add null implementation instead. but I think it will add a bit to the build

engine/dlib/src/dlib/configfile.cpp
```diff
@@ -30,6 +29,10 @@
 #include "sys.h"
 #include "static_assert.h"
 
+#if !defined(__EMSCRIPTEN__)
+#include "http_client.h"
+#endif
```
JCash: I think all these "sprinkles" of platform defines will be harder and harder to follow.
The engine is built around the concept of "features" and I believe that's the way forward.
Often we use a "null" library, but sometimes we also have defines, although their naming or usage isn't super clean yet.

E.g. in this case, what is the feature not supported on HTML5?
I suggest adding that as a define in the waf_dynamo.py, e.g. `DM_HAVE_X_SUPPORT` or `DM_NO_X_SUPPORT`, and if that is or isn't defined, the code here would be toggled on or off.

Potentially, we may have a "dconfig.h" for each platform in the future, in order to avoid the long command lines. That is just an idea at this stage.

engine/dlib/src/dlib/configfile.cpp
```diff
@@ -741,7 +748,12 @@ namespace dmConfigFile
         {
             if (strcmp(uri_parts.m_Scheme, "http") == 0 || strcmp(uri_parts.m_Scheme, "https") == 0)
             {
+#if defined(__EMSCRIPTEN__)
+                dmLogError("HTTP(S) config loading is not supported on HTML5: %s", url);
+                return RESULT_INVALID_URI;
```
JCash: Tbh, I see no real reason it _shouldn't_ work? (i.e. I dislike platform specific code like this)
However, if the feature is only supposed to be used in debug builds, then this function should probably live elsewhere.

engine/engine/src/engine.cpp
JCash, line 72: I'm not a fan of all these extra define-sprinkles in this file.
At the very least, use a feature-define for it.

engine/engine/src/engine.cpp
```diff
@@ -1140,7 +1149,7 @@ namespace dmEngine
         SetUpdateFrequency(engine, dmConfigFile::GetInt(engine->m_Config, "display.update_frequency", 0));
 
         engine->m_HttpCache = 0;
-#if !defined(DM_NO_HTTP_CACHE)
+#if !defined(DM_NO_HTTP_CACHE) && !defined(__EMSCRIPTEN__)
```
JCash: Here we already have a feature-define.
Would it be enough to define `DM_NO_HTTP_CACHE` for emscripten to disable this section?

engine/engine/src/engine.cpp
```diff
@@ -2524,7 +2533,9 @@ void dmEngineInitialize()
     dmCrash::Init(dmEngineVersion::VERSION, dmEngineVersion::VERSION_SHA1);
     dmDDF::RegisterAllTypes();
     dmSocket::Initialize();
+#if !defined(__EMSCRIPTEN__)
     dmSSLSocket::Initialize();
+#endif
```
JCash: I think it makes sense to simply implement sslsocket_null.cpp (like we usually do when toggling features in the engine)?

engine/resource/src/test/wscript
```diff
@@ -124,9 +126,13 @@ def build(bld):
     # ******************************************************************************************************************************
     # Resource providers
     skip_http_test = []
-    if bld.env.PLATFORM in ['wasm-web','wasm_pthread-web','js-web']:
+    if is_web:
         skip_http_test = ['skip_test']
 
+    resource_http_provider_symbols = []
+    if not is_web:
+        resource_http_provider_symbols.append('ResourceProviderHttp')
```
JCash: Should ProviderHttp be excluded from all platforms by default?
And if they need it, they can add it via an app manifest?

I'm not sure I like having platform specific setups, as each such thing makes it all harder to reason about.
Imho, it should work in a similar fashion on all platforms?

JCash (changes requested):
While I approve of the idea, I also believe that we should take the time to do it "the Defold way".
I.e. use common setup for all platforms (defines, libraries), avoid platform specific defines if necessary and use empty base implementation libraries to opt out of features.

engine/engine/src/engine.cpp
AGulev: fixed

engine/dlib/src/dlib/http_client.cpp
```diff
@@ -299,11 +299,17 @@ namespace dmHttpClient
...
-        client->m_IgnoreCache = params->m_HttpCache != 0 ? 0 : 1;
         client->m_CancelFlag = cancelflag;
 
+#if defined(DM_NO_HTTP_CACHE)
+        client->m_HttpCache = 0;
+        client->m_IgnoreCache = 1;
+#else
+        client->m_HttpCache = params->m_HttpCache;
+        client->m_IgnoreCache = params->m_HttpCache != 0 ? 0 : 1;
+#endif
```
JCash, line 311: 👍

JCash (approved):

PR #12108, opened 2026-03-25
Fix editor performance regression introduced by symlink support

matgis:
The recent addition of better support for symlinks introduced a performance regression in file existence checks. This had a significant impact on build times in larger projects.

### Technical changes
* We now validate `FileResource` `proj-paths` against the `resource-map` in the `Workspace`. To facilitate this, The `resource-map` was converted from an `output` to a `property`, along with all other outputs that are downstream of the `resource-snapshot` property. During `resource-sync`, we eagerly evaluate the resulting collections and assign them to the corresponding `Workspace` properties. This lets us cheaply access the `resource-map` - which is a map of canonical `proj-path` to `Resource` - from the `FileResource` `resource/exists?` implementation using `g/raw-property-value`.
* The various resource-resolving functions in the `workspace` module now take a `basis` instead of an `evaluation-context`.
* Added tests for `path/actual-cased` and `path/real`.
* Avoid calling the expensive `path/actual-cased` function whenever possible.

### DIFF
--- a/editor/src/clj/editor/collection.clj
+++ b/editor/src/clj/editor/collection.clj
@@ -859,52 +859,54 @@
 
 (defn load-collection [project self resource collection]
   {:pre [(map? collection)]} ; GameObject$CollectionDesc in map format.
-  (concat
-    (gu/set-properties-from-pb-map self GameObject$CollectionDesc collection
-      name :name
-      scale-along-z (protobuf/int->boolean :scale-along-z))
-    (let [tx-go-creation (flatten
-                           (concat
-                             (for [game-object (:instances collection)
-                                   :let [source-resource (workspace/resolve-resource resource (:prototype game-object))]]
-                               (make-ref-go self source-resource (:id game-object) game-object nil (:component-properties game-object) nil))
-                             (for [embedded (:embedded-instances collection)]
-                               (do
-                                 ;; Note: We only need to check that the
-                                 ;; EmbeddedInstanceDesc has been string-decoded
-                                 ;; here. Any EmbeddedComponentDescs inside will
-                                 ;; be validated by the game-object :load-fn.
-                                 (collection-string-data/verify-string-decoded-embedded-instance-desc! embedded resource)
-                                 (make-embedded-go self project (:data embedded) (:id embedded) embedded nil nil)))))
-          id->nid (-> tx-go-creation
-                      (g/tx-data-added-nodes)
-                      (coll/into-> {}
-                        (filter #(g/node-instance*? GameObjectInstanceNode %))
-                        (map (fn [node]
-                               (pair (:id node)
-                                     (g/node-id node))))))
-          child->parent (-> {}
-                            (into (map #(pair (key %) nil))
-                                  id->nid)
-                            (into
-                              (comp cat
-                                    (mapcat (fn [{:keys [children id]}]
-                                              (map #(pair % id)
-                                                   children))))
-                              (pair (:instances collection)
-                                    (:embedded-instances collection))))]
-      (validate-child-references! id->nid child->parent)
-      (concat
-        tx-go-creation
-        (for [[child parent] child->parent
-              :let [child-id (id->nid child)
-                    parent-id (if parent (id->nid parent) self)]]
-          (if parent
-            (child-go-go parent-id child-id)
-            (child-coll-any self child-id)))))
-    (for [coll-instance (:collection-instances collection)
-          :let [source-resource (workspace/resolve-resource resource (:collection coll-instance))]]
-      (make-collection-instance self source-resource (:id coll-instance) coll-instance (:instance-properties coll-instance) nil))))
+  (let [basis (g/now)
+        resolve-resource #(workspace/resolve-resource basis resource %)]
+    (concat
+      (gu/set-properties-from-pb-map self GameObject$CollectionDesc collection
+        name :name
+        scale-along-z (protobuf/int->boolean :scale-along-z))
+      (let [tx-go-creation (flatten
+                             (concat
+                               (for [game-object (:instances collection)
+                                     :let [source-resource (resolve-resource (:prototype game-object))]]
+                                 (make-ref-go self source-resource (:id game-object) game-object nil (:component-properties game-object) nil))
+                               (for [embedded (:embedded-instances collection)]
+                                 (do
+                                   ;; Note: We only need to check that the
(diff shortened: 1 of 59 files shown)


editor/src/clj/editor/collection.clj
```diff
@@ -859,52 +859,54 @@
...
-              :let [child-id (id->nid child)
-                    parent-id (if parent (id->nid parent) self)]]
-          (if parent
-            (child-go-go parent-id child-id)
-            (child-coll-any self child-id)))))
-    (for [coll-instance (:collection-instances collection)
-          :let [source-resource (workspace/resolve-resource resource (:collection coll-instance))]]
-      (make-collection-instance self source-resource (:id coll-instance) coll-instance (:instance-properties coll-instance) nil))))
+  (let [basis (g/now)
+        resolve-resource #(workspace/resolve-resource basis resource %)]
```
matgis, line 863: Indentation change from this surrounding `let`. Other than that, we use this `resolve-resource` function everywhere instead of `workspace/resolve-resource`.

editor/src/clj/editor/graph_util.clj
```diff
@@ -51,91 +71,6 @@
...
-         owner-resource-node-id (try
-                                  (resource-node/owner-resource-node-id basis node-id)
-                                  (catch Exception _
-                                    nil))
-         owner-resource-node-type-kw (some->> owner-resource-node-id (g/node-type-kw basis))]
-     {:node-type-kw node-type-kw
-      :node-debug-label-path node-debug-label-path
-      :owner-resource-node-id owner-resource-node-id
-      :owner-resource-node-type-kw owner-resource-node-type-kw})))
-
```
matgis, line 138: Moved to `node_util.clj` so we can exclude a lot of require dependencies from this file and use it from `workspace.clj`.

editor/src/clj/editor/node_util.clj
```diff
@@ -0,0 +1,112 @@
...
+         node-debug-label-path (node-debug-label-path node-id evaluation-context)
+         owner-resource-node-id (try
+                                  (resource-node/owner-resource-node-id basis node-id)
+                                  (catch Exception _
+                                    nil))
+         owner-resource-node-type-kw (some->> owner-resource-node-id (g/node-type-kw basis))]
+     {:node-type-kw node-type-kw
+      :node-debug-label-path node-debug-label-path
+      :owner-resource-node-id owner-resource-node-id
+      :owner-resource-node-type-kw owner-resource-node-type-kw})))
```
matgis, line 112: Moved here from `graph_util.clj`.

editor/src/clj/editor/workspace.clj
```diff
@@ -198,18 +199,6 @@ ordinary paths."
...
-    (resource/make-file-resource _node-id root (io/as-file root) (:resources resource-snapshot) editable-proj-path? unloaded-proj-path?)))
-
-(g/defnk produce-resource-list [resource-tree]
-  (vec (sort-by resource/proj-path util/natural-order (resource/resource-seq resource-tree))))
-
-(g/defnk produce-resource-map [resource-list]
-  (into {}
-        (map #(pair (resource/proj-path %) %))
-        resource-list))
-
```
matgis, line 212: This now happens in the `resource-snapshot` property setter.

editor/src/clj/editor/workspace.clj
```diff
@@ -955,31 +934,56 @@ ordinary paths."
     s/Keyword s/Any}])
 
 (g/defnode Workspace
-  (property root g/Str)
+  (property root g/Str
+            (set (gu/immutable-property-setter root)))
```
matgis, line 938: Instead of throwing an exception, we could update the same properties as we do in the `resource-snapshot` property setter from these immutable properties, but they never change, so why bother?

editor/src/clj/editor/resource.clj
```diff
@@ -168,11 +172,11 @@
     (subs proj-path 0 last-slash)))
 
 (defn project-directory? [^File value]
-  ;; The project directory must exist on disk and exactly match the real casing.
+  ;; The project directory must be a real and existing path on disk.
```
vlaaad, line 175: Real means resolved to the actual file path if a symlink? Do we resolve the open project dir to the real path when starting the editor?

editor/src/clj/editor/resource.clj
```diff
@@ -441,22 +445,10 @@
...
-        (and (.exists file)
-             (not (ignored-project-path? (io/file root) project-path))
-             (string/ends-with? (->unix-seps (str (path/actual-cased file))) project-path)))
-      (catch IOException _
-        false)
-      (catch SecurityException _
-        false)))
+    (and (proj-path-exists? (g/unsafe-basis) workspace project-path)
+         (if (path/symlink? this)
+           (.exists (io/file this))
```
vlaaad, line 450: Why not `path/exists?`?

editor/src/clj/editor/fs.clj
```diff
@@ -313,7 +313,7 @@
 
 (def ^:private fs-temp-dir (create-temp-directory! "moved"))
 
-(def case-sensitive?
+(def is-case-sensitive
```
vlaaad, line 316: 👍

vlaaad (approved):
LGTM! Please cherry-pick it into `beta`. I also measured a 10% project load time improvement on a big project (1m29s -> 1m20s).

editor/src/clj/editor/resource.clj
matgis: Yes, we call `getCanonicalFile` on the specified `project-path` when we create the workspace in `workspace/make-workspace`, which does follow symlinks. The old way of doing it may have been wrong, actually. 🤔

editor/src/clj/editor/resource.clj
matgis: Hmm. No real reason. It would have to coerce `this` to a `Path` instead of a `File`, so there's no real gain either. I think I'll leave it like this for now, since I'm a bit afraid the `Path` existence check might have have subtle differences from the `File` implementation or something. I feel a bit gun-shy after the surprising implications of using `toRealPath`. 😬

matgis:
Since we want this fix to make it into the 1.12.3 Beta, the changes will be merged into `beta` in PR #12116. We will then cherry-pick the commit into `dev` after fixing merge issues.

matgis:
#12116 has been merged into `beta`. I will cherry-pick the commit into `dev`. Closing this one.

PR #12113, opened 2026-03-26
Reject loading of glTF files that have more than one joint or weight channels

### LINKED ISSUE #11420: Incorrect rendering of animated FBX model exported from Houdini to GLB
An animated FBX model imported into SideFX Houdini 21 and exported to GLB format is incorrectly rendered in Defold. The model appears improperly skinned, despite having no validation errors. The same GLB file displays correctly in external viewers such as [PlayCanvas](https://playcanvas.com/viewer) and the [GLTF Khronos Viewer](https://github.khronos.org/glTF-Sample-Viewer-Release/), confirming the file integrity.

_For clarity:_ the model was originally created in another 3D software and only imported into Houdini for export. I mention FBX and Houdini because _FBX uses centimeters as its default unit_, and I suspect this may cause scale or transformation matrix discrepancies in the resulting GLTF file. However, since all external viewers display the model correctly, the issue appears specific to how Defold interprets the exported data.

**To Reproduce (REQUIRED)**
1. Open the test project.
2. Observe incorrect mesh deformation.

The animated GLB model should render correctly in Defold.

  - Version: 1.11.1, 1.11.2 BETA

**Platforms (REQUIRED):**
 - Platforms: Windows 11

[issue_houdini_fbx_glb_octopus.zip](https://github.com/user-attachments/files/23131786/issue_houdini_fbx_glb_octopus.zip)

**Screenshots (OPTIONAL):**
How it should look:
Image

How it looks now:

Image

Jhonnyg:
Defold currently only supports one weight and one joint channel (maps to attributes/inputs to vertex shaders) with four components each. Using .gltf content that has more than one of these channels will cause the rendering to be incorrect. In the future we might support the generic case of unlimited such channels, but in the meantime the editor (and bob) will raise an error when trying to load the content. Currently, the simplest fix is to re-export the offending files via Blender.

Fixes #11420

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ModelUtilTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ModelUtilTest.java
@@ -14,6 +14,12 @@
 
 package com.dynamo.bob.pipeline;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
+
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
@@ -39,8 +45,6 @@ import com.dynamo.bob.util.MurmurHash;
 
 import com.dynamo.proto.DdfMath.Vector3;
 import com.dynamo.rig.proto.Rig;
-
-import static org.junit.Assert.*;
 
 public class ModelUtilTest {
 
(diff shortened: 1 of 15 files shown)


com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ModelUtilTest.java
```diff
@@ -287,7 +291,7 @@ private Modelimporter.Scene loadBuiltScene(String path,
      * Tests a collada file with fewer, and more, than 4 bone influences per vertex.
      */
     @Test
-    public void testBoneInfluences() throws Exception {
```
Jhonnyg, line 290: None of these actually throws exceptions.

engine/modelc/src/modelimporter.cpp
```diff
@@ -48,14 +48,33 @@ struct ModelImporterInitializer
     }
 } g_ModelImporterInitializer;
 
-
 namespace dmModelImporter
 {
 
+static char g_ModelLoadError[512];
```
JCash: I do worry a little bit about multi threaded loading where multiple failures write to the same field.
A mutex would help with writing, but not with reading which may give wrong error.

Can we store the error with the Scene* pointer?
And, in the jni.cpp, we see that it has a load failure:
* build Java Exception / or save error string
* delete Scene*
* throw Java Excption

JCash (changes requested):
Overall good, but I think loading should be thread safe which is currently isn't.

PR #12117, opened 2026-03-26
Minor memory use and load time improvement

vlaaad:
Now the editor loads 2% faster and uses 2% less memory (on a big project I tried, memory use after open reduced from 2.45 Gb to 2.39 Gb, and load time reduced from 1:16 to 1:14).

Technical notes:
A lot of nodes have "parallel" connections, meaning that between 2 nodes of a particular type, there always exist the same sets of arcs. Most of them cannot be merged together, because these connections carry invalidation and error propagation info that should not be intertwined. But some of them ultimately mean the same thing. In this PR, I merged such connections for the most common types in big projects.

Related to #11989

### DIFF
--- a/editor/src/clj/editor/atlas.clj
+++ b/editor/src/clj/editor/atlas.clj
@@ -165,7 +165,12 @@
 (defn- atlas-rect->editor-rect [rect]
   (types/->Rect (:path rect) (:x rect) (:y rect) (:width rect) (:height rect)))
 
-(g/defnk produce-scene-info [layout-size image-path->rect updatable :as ret] ret)
+(g/defnk produce-atlas-scene-info [layout-size image-path->rect]
+  {:layout-size layout-size
+   :image-path->rect image-path->rect})
+
+(g/defnk produce-animation-scene-info [atlas-scene-info updatable]
+  (assoc atlas-scene-info :updatable updatable))
 
 (g/defnk produce-image-scene [_node-id image-resource order scene-info]
   (let [{:keys [layout-size image-path->rect updatable]} scene-info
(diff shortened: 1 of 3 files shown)


editor/src/clj/editor/atlas.clj
```diff
@@ -165,9 +165,11 @@
 (defn- atlas-rect->editor-rect [rect]
   (types/->Rect (:path rect) (:x rect) (:y rect) (:width rect) (:height rect)))
 
-(g/defnk produce-image-scene
-  [_node-id image-resource order layout-size image-path->rect animation-updatable]
```
vlaaad, line 169: `layout-size`, `image-path->rect`, and `updatable` outputs are all connected to `produce-image-scene`, so they can be safely merged. Changes in this file do exactly that — merge them into `scene-info`.

editor/src/clj/editor/defold_project.clj
```diff
@@ -926,7 +926,6 @@
       (g/construct node-type
         :_node-id node-id
         :resource resource))
-    (g/connect node-id :_node-id project :nodes)
```
vlaaad, line 929: All resource nodes always had 2 connections to the project: `:_node-id`-> `:nodes` (for ownership) and `:node-id+resource` -> `:node-id+resources` (for useful work). I made ownership tracked through `node-id+resources` input.

editor/src/clj/editor/atlas.clj
```diff
@@ -165,9 +165,11 @@
...
-  (let [path (resource/proj-path image-resource)
+(g/defnk produce-scene-info [layout-size image-path->rect updatable :as ret] ret)
```
matgis: I think we may be able to do even better here. I don't think we should share this production-fn between `AtlasNode` and `AtlasAnimation`. Instead, we should remove the `updatable` output from the `AltasNode` (which always produces `nil`), connect the `AtlasNode` `scene-info` to an `atlas-scene-info` input on `AtlasAnimation` (saves us from having the separate connections for `layout-size` and `image-path->rect` from the `AtlasNode`), and assoc the `AtlasAnimation` `updatable` into its `scene-info` output.

matgis (dismissed):

PR #12118, opened 2026-03-26
Untrack bob libexec

Jhonnyg:
### Technical changes
All files from bob's libexec folder have been removed in favor of putting them into .tar.gz files inside the packages folder per platform. We now run the copy script every time we build bob (we previously did that for bob-light only in the build.py script).


JCash (approved):
Bravo!! 👏

Issue #9034, opened 2024-06-05
Refactor basic graphic adapters structures for reusability and simplifcation


Jhonnyg:
**Task (REQUIRED):**

We currently have 5 graphics adapters (null, opengl, vulkan + 2 private console) with at least 2 more on the way (DX12, WebGPU). The setup we currently have in place implements a lot of individual functions for each adapter that could be generalized, which would remove a lot of the current code size as well as simplify making changes to the graphics API. For example, getting information about a texture requires each adapter to implement all sorts of functions:

```
uint32_t    GetTextureResourceSize(HTexture texture);
uint16_t    GetTextureWidth(HTexture texture);
uint16_t    GetTextureHeight(HTexture texture);
uint16_t    GetTextureDepth(HTexture texture);
...

Each adapter implements as:
uint32_t NullGetTextureResourceSize(HTexture texture);
uint32_t OpenGLGetTextureResourceSize(HTexture texture);
uint32_t VulkanGetTextureResourceSize(HTexture texture);
....
```

This is because the private structures representing the basic objects (textures, render targets, shaders and so forth) need to store slightly different data depending on the API. But in general, all of the basic objects share at least some amount of data between them, which means that we could use mostly the same set of functions on then instead of specific individual implementations.
To achieve this, we can utilise a "soft inheritance" structure in C that separates the base data from the specific data:

```
struct Texture
{
    int m_Width;
    int m_Height;
    int m_Flags;
    ...
};

struct OpenGLTexture
{
    Texture m_Base;
    GLuint  m_Id;
    ...
};

struct VulkanTexture
{
    Texture m_Base;
    VkImage m_Image;
    ...
};
```

If we do this for all APIs, we would be able to simplify the implementation quite a lot!

**Expected outcome (REQUIRED):**
Describe the expected outcome of the task (new issue, documentation, code change etc)

PR #12123, opened 2026-03-27
Refactor adapter implementations to use a shared GraphicsContext struct

Jhonnyg:
All graphics adapter contexts now has a shared "GraphicsAdapter" struct.

Sub-part of #9034

### DIFF
--- a/engine/graphics/src/dx12/graphics_dx12_private.h
+++ b/engine/graphics/src/dx12/graphics_dx12_private.h
@@ -235,6 +235,7 @@ namespace dmGraphics
 
     struct DX12Context
     {
+        GraphicsContext                    m_BaseContext;
         ID3D12Device*                      m_Device;
 
 #if defined(DM_PLATFORM_VENDOR)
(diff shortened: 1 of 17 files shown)


engine/graphics/src/dx12/graphics_dx12_private.h
```diff
@@ -235,6 +235,7 @@ namespace dmGraphics
 
     struct DX12Context
     {
+        GraphicsContext                    m_BaseContext;
```
JCash, line 238: Nice.
This what I thought of as well, but I limited the scope of my GraphicsContext PR :)

engine/graphics/src/graphics_private.h
```diff
@@ -133,6 +133,20 @@ namespace dmGraphics
...
+    {
+        HWindow                            m_Window;
+        dmOpaqueHandleContainer<uintptr_t> m_AssetHandleContainer;
+        uint64_t                           m_TextureFormatSupport;
+        TextureFilter                      m_DefaultTextureMinFilter;
+        TextureFilter                      m_DefaultTextureMagFilter;
+        uint32_t                           m_Width;
+        uint32_t                           m_Height;
+        uint32_t                           m_VerifyGraphicsCalls : 1;
+        uint32_t                           m_PrintDeviceInfo     : 1;
```
JCash, line 147: Should validation layer bool be here as well?
And, what about other common settings, like msaa or similar?
Not critical at this stage, as this struct is private, but something to think about.

JCash (dismissed):

engine/graphics/src/graphics_private.h
Jhonnyg: Yes, I thought about those as well, but I wanted to keep the scope down first. The current changes are just the ones that are used for _all_ platforms, we can add the most common settings later.

PR #12127, opened 2026-03-28
Remove unsupported luasocket functionality from HTML5 build

AGulev:
TCP/UDP isn't supported on the HTML5 platform with luasockets. Removing this functionality from the binary helps reduce the WASM binary size slightly (~1%)

### Technical changes
Removing unused code helps to reduce dmengine_release.wasm from 2445043 bytes to 2420425 bytes

### DIFF
--- a/build_tools/waf_dynamo.py
+++ b/build_tools/waf_dynamo.py
@@ -546,6 +546,13 @@ def default_flags(self):
             if int(opt_level) < 2:
                 flags += ['-gseparate-dwarf', '-gsource-map']
                 linkflags += ['-gseparate-dwarf', '-gsource-map']
+            if Options.options.wasm_size_analysis:
+                # Keep source attribution outside the main wasm so size measurements
+                # still reflect the optimized binary while enabling deeper analysis.
+                flags += ['-gline-tables-only']
+                for flag in ['-gsource-map', '-gseparate-dwarf']:
+                    if flag not in linkflags:
+                        linkflags += [flag]
         else:
             emflags_link += ['WASM=0', 'LEGACY_VM_SUPPORT=1']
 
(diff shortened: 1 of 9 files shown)


scripts/web/wasm_symbol_sizes.py
```diff
@@ -0,0 +1,217 @@
+#!/usr/bin/env python3
```
AGulev, line 1: it's not related directly to the PR
more like a helper script

build_tools/waf_dynamo.py
```diff
@@ -546,6 +546,13 @@ def default_flags(self):
             if int(opt_level) < 2:
                 flags += ['-gseparate-dwarf', '-gsource-map']
                 linkflags += ['-gseparate-dwarf', '-gsource-map']
+            if Options.options.wasm_size_analysis:
```
JCash: Does the name need to be "wasm" specific?
I think we should use platform agnostic naming for such options.

build_tools/waf_dynamo.py
```diff
@@ -2138,6 +2145,7 @@ def options(opt):
     opt.add_option('--with-dx12', action='store_true', default=False, dest='with_dx12', help='Enables DX12 as a graphics backend')
     opt.add_option('--with-opus', action='store_true', default=False, dest='with_opus', help='Enable Opus audio codec support in runtime')
     opt.add_option('--with-webgpu', action='store_true', default=False, dest='with_webgpu', help='Enables WebGPU as graphics backend')
+    opt.add_option('--wasm-size-analysis', action='store_true', default=False, dest='wasm_size_analysis', help='Emit extra wasm-web analysis artifacts such as source maps and separate DWARF')
```
JCash: We have `--static-analyze`, so perhaps `-size-analyze` ?

engine/script/src/luasocket/luasocket_html5.c
```diff
@@ -0,0 +1,157 @@
...
+
+static int global_unsupported(lua_State *L) {
+    (void) L;
+    lua_pushnil(L);
+    lua_pushliteral(L, "LuaSocket networking is not supported on HTML5");
+    return 2;
+}
+
+static int global_dns_unsupported(lua_State *L) {
+    (void) L;
```
JCash: redundant, since it's actually used below.

JCash (changes requested):
Overall good, but would like a change of name of the argument

AGulev:
Thank you for review! All The comments addressed.

JCash (approved):

PR #12132, opened 2026-03-30
Fix custom resources not rebuilding after changes in the editor

matgis:
A regression was introduced in #12116 that caused custom resources to not refresh between builds in the editor. This PR addresses the issue and updates the build tests that failed to catch the problem.

Fixes #12126.

### Technical changes
* The build tests have been updated to cache graph evaluations in the system cache, just like the editor does. The fact that we were bypassing the system cache made it possible for this regression to slip through in the first place.
* There is now an explicit connection between the `Workspace` `resource-snapshot` and the `GameProjectNode` that emits `build-targets` for the custom resources. This data dependency already existed in the graph before the regression was introduced in #12116. In the old code, the `Workspace` `resource-map` output (used by the `custom-build-targets` output of `GameProjectNode`) depended on `resource-snapshot`. The new code introduced in #12116 only invalidates the `resource-map` output if there are actual changes to the produced `resource-map` resulting from changes to the `resource-snapshot`. With this PR, the data dependency on the `resource-snapshot` is made explicit.
* The `resource-map` `passthrough` output on `Project` has been removed, and all use sites have been updated to use the `Workspace` `resource-map` output instead. This was done to avoid having the `Project` node also having to `passthrough` the `resource-snapshot` from the `Workspace`.
* The `ErrorValue` messages resulting from a faulty `network.ssl_certificates` setting now differs from the messages resulting from faulty `project.custom_resources`.

### DIFF
--- a/editor/src/clj/editor/game_project.clj
+++ b/editor/src/clj/editor/game_project.clj
@@ -29,6 +29,7 @@
             [editor.settings :as settings]
             [editor.settings-core :as settings-core]
             [editor.workspace :as workspace]
+            [util.coll :as coll :refer [pair]]
             [util.defonce :as defonce]))
 
 (set! *warn-on-reflection* true)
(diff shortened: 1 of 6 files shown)


editor/src/clj/editor/game_project.clj
```diff
@@ -174,27 +164,80 @@
...
+            (let [status-map (:status-map resource-snapshot)
+
+                  directory-resources
+                  (cond-> custom-resources-directory-resources
+                          ssl-certificates-directory-resource (conj ssl-certificates-directory-resource))
+
+                  custom-resources
+                  (coll/into-> directory-resources []
+                    (map resource/resource-seq)
+                    coll/flatten-xf
```
vlaaad, line 214: Looks like `coll/tree-xf`...

vlaaad (approved):
LGTM!

PR #12136, opened 2026-03-31
Cache the text layout to speed up text rendering

AGulev:
Text layout calculation is pretty expensive. To reduce its cost, we cache it once and reuse it until layout-related parameters change. This significantly speeds up text rendering (1.5x-4.5x, depending on the selected font system, classic or dynamic fonts), but increases memory consumption.

### Technical changes

TextLayout cached in Labels, GuiTextNodes and partially in profiler (for labels it is cached, but for numbers we create it once for the frame to avoid double calculations).

For my test on my M1 Max
500 labels (2 fonts)
500 gui text nodes (2 fonts)

Sim ~9.5ms -> ~1.9ms
- Render ~4ms -> ~0.3 ms
— guic ~2.1ms -> ~0.22ms
— labelc ~2ms -> ~0.1ms
- UpdateRSI ~5ms -> ~1.26ms
- RenderProfiler ~0.62ms -> 0.37ms

test project
[aLotOfText.zip](https://github.com/user-attachments/files/26411599/aLotOfText.zip)

 It comes with a memory cost:
  +16 bytes per GUI node (any type)
  +12 bytes per label component
  +heap TextLayout (every label, and only text nodes): ~104 bytes + 40 bytes * glyphs + 8 * lines + allocator overhead


engine/gamesys/src/gamesys/components/comp_label.cpp
```diff
@@ -694,6 +788,7 @@ namespace dmGameSystem
                 return dmGameObject::PROPERTY_RESULT_TYPE_MISMATCH;
             }
             component->m_Leading = params.m_Value.m_Number;
+            InvalidateTextLayout(component);
```
JCash, line 763: I think this is good for now.
Ofc one could create a hash of the inputs affecting the layout, but we can do that later if necessary.

engine/gamesys/src/gamesys/components/comp_label.cpp
```diff
@@ -576,7 +658,14 @@ namespace dmGameSystem
...
     {
-        dmRender::HFontMap font_map = GetFontMap(component, component->m_Resource);
+        LabelComponent* mutable_component = const_cast<LabelComponent*>(component);
+        dmRender::HFontMap font_map = GetFontMap(mutable_component, mutable_component->m_Resource);
+        HTextLayout layout = GetOrCreateTextLayout(mutable_component);
+        if (layout)
+        {
+            dmRender::GetTextMetrics(font_map, layout, &metrics);
+            return;
+        }
```
JCash: Is there a case where we need the "else" clause here?
if we couldn't make a HTextLayout, why would the next code block produce better results? 🤔

engine/gamesys/src/gamesys/resources/res_font.cpp
```diff
@@ -867,7 +883,12 @@ namespace dmGameSystem
             return dmResource::RESULT_RESOURCE_NOT_FOUND;
         }
 
-        return AddFontInternal(factory, resource, ttfresource, ttf_hash);
+        r = AddFontInternal(factory, resource, ttfresource, ttf_hash);
+        if (r == dmResource::RESULT_OK)
+        {
+            ++resource->m_Version;
+        }
```
JCash, line 882: 👍

engine/gamesys/src/gamesys/components/comp_label.cpp
```diff
@@ -486,6 +565,7 @@ namespace dmGameSystem
 
             dmRender::DrawTextParams text_params;
             CreateDrawTextParams(component, text_params);
+            text_params.m_TextLayout = GetOrCreateTextLayout(component);
```
JCash: This should be moved to `CreateDrawTextParams()` for consistency
and DrawTextParams.m_Text should set to `0`, as the m_Text should no longer be needed for rendering the text.

engine/render/src/render/font/fontmap.cpp
```diff
@@ -331,6 +331,23 @@ namespace dmRender
...
 
+    void GetTextMetrics(HFontMap font_map, HTextLayout layout, TextMetrics* metrics)
+    {
+        DM_MUTEX_SCOPED_LOCK(font_map->m_Mutex);
+
+        metrics->m_Width = 0.0f;
+        metrics->m_Height = 0.0f;
+        metrics->m_MaxAscent = font_map->m_MaxAscent;
+        metrics->m_MaxDescent = font_map->m_MaxDescent;
+        metrics->m_LineCount = 0;
```
JCash: To minimize locked scope:
```c++
void GetTextMetrics(HFontMap font_map, HTextLayout layout, TextMetrics* metrics)
{
    {
        DM_MUTEX_SCOPED_LOCK(font_map->m_Mutex);
        metrics->m_Width = 0.0f;
        ...
    }

    if (layout)
    { ... }
```

engine/render/src/render/render_private.h
```diff
@@ -155,6 +156,7 @@ namespace dmRender
         HConstant           m_RenderConstants[MAX_TEXT_RENDER_CONSTANTS];
         HFontMap            m_FontMap;
         HMaterial           m_Material;
+        HTextLayout         m_TextLayout;
```
JCash, line 159: 👍

JCash (changes requested):
Overall good, and the same as I envisioned it.
I see some room for improvement in this step as well, as the m_Text shouldn't be necessary if a layout is passed in the drawtextparams struct.

engine/font/src/text_layout.cpp
```diff
@@ -46,6 +47,24 @@ void TextLayoutFree(HTextLayout layout)
...
+    }
+    codepoints.SetSize(0);
+
+    const char* cursor = safe_text;
+    while (uint32_t c = dmUtf8::NextChar(&cursor))
+    {
+        codepoints.Push(c);
+    }
+    return len;
+}
```
AGulev, line 66: we had it in a few places including tests. It's better to have one function and reuse it everywhere

engine/gamesys/src/gamesys/resources/res_font.cpp
```diff
@@ -670,23 +679,7 @@ namespace dmGameSystem
                 return dmResource::RESULT_NOT_SUPPORTED;
             }
 
-            if (font->m_TTFResources.Full())
```
AGulev, line 673: moved into a function to fix hotreload

engine/profiler/src/profiler_render.cpp
```diff
@@ -138,12 +140,141 @@ namespace dmProfileRender
...
+        dmArray<uint32_t> m_TextCodePoints;
+        dmHashTable64<HTextLayout> m_LabelTextLayouts;
+        dmRender::HFontMap m_TextLayoutFontMap;
 
         uint32_t m_IncludeFrameWait : 1;
 
         RenderProfile(float fps, uint64_t ticks_per_second, uint64_t lifetime_in_milliseconds);
     };
 
+    void ClearTransientTextLayouts(HRenderProfile render_profile)
```
AGulev, line 153: two types of text layout cache:
`m_TransientTextLayouts` -  is a per-frame owner list for layouts that are expected to change all the time. Avoids double layout calculation per frame in `dmRender::DrawText()` and in `CreateFontVertexData()` but since numbers are dynamic, it's better to cleanup it each frame (FrameArena basically)
`m_LabelTextLayouts` - is a long-lived hash table for stable text that repeats across frames.

engine/render/src/render/font/default/font_default_vertex.cpp
```diff
@@ -12,12 +12,12 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
+#include <assert.h>                        // for assert
 #include <stdint.h>                         // for uint32_t, int16_t
 #include <dlib/align.h>                     // for DM_ALIGNED
 #include <dlib/log.h>                       // for dmLog*
 #include <dlib/profile.h>                   // for DM_PROFILE, DM_PROPERTY_*
 #include <dlib/vmath.h>                     // for Vector4
-#include <dlib/time.h>                      // for dmTime::GetMonotonicTime()
```
AGulev, line 20: isn't used

engine/render/src/render/font/default/font_default_vertex.cpp
```diff
@@ -426,6 +392,7 @@ uint32_t CreateFontVertexData(HFontRenderBackend backend, HFontMap font_map, uin
     }
 
     int32_t dir = 1;//layout->m_Direction == TEXT_DIRECTION_RTL ? -1 : 1;
+    (void)dir;
```
AGulev, line 382: isn't needed for now, but I think maybe it's been left for a reason of some upcoming changes, so I'm keeping it here

engine/render/src/render/font/default/font_default_vertex.cpp
```diff
@@ -356,38 +354,6 @@ uint32_t CreateFontVertexData(HFontRenderBackend backend, HFontMap font_map, uin
     // and not generated from the visual bounds (see Fontc.java)
     bool is_metrics_ttf = font_map->m_IsDynamic;
 
-    HFontCollection font_collection = font_map->m_FontCollection;
```
AGulev, line 359: moved into a function below

engine/render/src/render/font/fontmap.cpp
```diff
@@ -729,7 +747,7 @@ namespace dmRender
         return result;
     }
 
-    void AddGlyphToCache(HFontMap font_map, uint32_t frame, uint64_t glyph_key, FontGlyph* glyph, int32_t g_offset_y)
+    CacheGlyph* AddGlyphToCache(HFontMap font_map, uint32_t frame, uint64_t glyph_key, FontGlyph* glyph, int32_t g_offset_y)
```
AGulev, line 750: I'm trying to avoid multiple array data fetching since it's a hotpath. See the call place for the explanation

engine/render/src/render/font/default/font_default_vertex.cpp
```diff
@@ -486,14 +455,13 @@ uint32_t CreateFontVertexData(HFontRenderBackend backend, HFontMap font_map, uin
             if (glyph && glyph->m_Bitmap.m_Width > 0) // only add glyphs with a size (image) to the glyph cache
             {
                 uint64_t glyph_key = dmRender::MakeGlyphIndexKey(font, glyph_index);
-                if (!IsInCache(font_map, glyph_key))
```
AGulev, line 489: since this is a hotpath and `IsInCache()` is basically `GetFromCache()` but without  returning the value, I avoid calling it. It helps to save some signifficant amount of time on text rich scenes.

engine/render/src/render/font/default/font_default_vertex.cpp
```diff
@@ -486,14 +455,13 @@ uint32_t CreateFontVertexData(HFontRenderBackend backend, HFontMap font_map, uin
...
+                CacheGlyph* cache_glyph = GetFromCache(font_map, glyph_key, frame);
+                if (!cache_glyph)
                 {
                     // Calculate y-offset in cache-cell space by moving glyphs down to baseline
                     int16_t px_cell_offset_y = font_map->m_CacheCellMaxAscent - (int16_t)glyph->m_Ascent;
-                    AddGlyphToCache(font_map, frame, glyph_key, glyph, px_cell_offset_y);
+                    cache_glyph = AddGlyphToCache(font_map, frame, glyph_key, glyph, px_cell_offset_y);
```
AGulev, line 450: the same reason as above. If we know what we have added, we can reuse it and avoid an extra `GetFromCache`

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -1271,6 +1271,82 @@ namespace dmGameSystem
         return 0;
     }
 
+    static float CanonicalizeCacheKeyFloat(float value)
+    {
+        return value == 0.0f ? 0.0f : value;
```
JCash: While I understand what it's supposed to do, it only helps with +/-0, not with other rounding errors like 0.9999/1.00001 etc.

Isn't it enough to just hash the value as-is?
If it changed, it _did_ change after all? Why hide that?

engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -1271,6 +1271,82 @@ namespace dmGameSystem
...
+
+        dmHashUpdateBuffer64(&key_state, &font_resource, sizeof(font_resource));
+        dmHashUpdateBuffer64(&key_state, &font_version, sizeof(font_version));
+        dmHashUpdateBuffer64(&key_state, &text_hash, sizeof(text_hash));
+        dmHashUpdateBuffer64(&key_state, &canonical_width, sizeof(canonical_width));
+        dmHashUpdateBuffer64(&key_state, &canonical_leading, sizeof(canonical_leading));
+        dmHashUpdateBuffer64(&key_state, &canonical_tracking, sizeof(canonical_tracking));
+        dmHashUpdateBuffer64(&key_state, &line_break_flag, sizeof(line_break_flag));
+        return dmHashFinal64(&key_state);
+    }
```
JCash, line 1289: 👍

engine/gui/src/gui.h
```diff
@@ -258,6 +260,12 @@ namespace dmGui
         /// Max descent of font, positive value
         float m_MaxDescent;
     };
+
+    struct TextLayoutCache
```
JCash: Perhaps the name isn's the best.
Imho a "cache" is a whole cache with multiple entries, whereas this is just a textlayout.
I think simply `TextLayout` is just fine.

engine/gui/src/gui.h
```diff
@@ -1065,6 +1073,9 @@ namespace dmGui
 
     Result GetTextMetrics(HScene scene, const char* text, const char* font_id, float width, bool line_break, float leading, float tracking, TextMetrics* metrics);
     Result GetTextMetrics(HScene scene, const char* text, dmhash_t font_id, float width, bool line_break, float leading, float tracking, TextMetrics* metrics);
+    void GetNodeTextLayoutCache(HScene scene, HNode node, TextLayoutCache* out_cache);
+    void SetNodeTextLayoutCache(HScene scene, HNode node, const TextLayoutCache& cache);
+    void ClearNodeTextLayoutCache(HScene scene, HNode node);
```
JCash: re code size:
This one feels redundant?
I imagine `SetNodeTextLayout(scene, node, 0)` is sufficient?

engine/profiler/src/profiler_render.cpp
JCash: How much is saved in this system?

Using the full layout system for this seems like an extra burden imho.
E.g. if the `TextLayoutSettings` has an option to use a the TextLayoutType::TEXT_LAYOUT_TYPE_LEGACY in text_layout.cpp : TextLayoutCreate().
E.g. then I'd imagine the profile text would become much faster (more or less same as before)

To be clear, I'm not pushing for that now, it's just a thought.

engine/render/src/render/font/default/font_default_vertex.cpp
JCash: Yeah, I left it as a reminder or a FYI, that someone may want this in the future. Not quite a TODO, since I'm not sure what how to use it or expose it to the user.

engine/render/src/render/font/font_renderer.cpp
```diff
@@ -44,6 +45,19 @@ namespace dmRender
...
+    {
+#ifndef NDEBUG
+        if (layout)
+        {
+            assert(layout->m_FontCollection == GetFontCollection(font_map));
+        }
+#else
+        (void)font_map;
+        (void)layout;
+#endif
```
JCash: This is not usually how we do things.

When using defines to turn things off, we'd rather remove it altogether (not relying on compiler/linker).
i.e. by not emitting any code at all:
```c++
#if !defined(NDEBUG)
    static inline void AssertTextLayoutMatchesFontMap() { ... }
#else
    #define AssertTextLayoutMatchesFontMap(_ARG_A, _ARG_B)
#endif
```

However, we don't generally do ifdef on `NDEBUG` either, as we tend to ship it.
And `assert()`is already depending on NDEBUG. 
The remaining part is the `if (layout)`.
Is that necessary?
It seems it's already verified outside of the function?
```
if (layout)
{
   AssertTextLayoutMatchesFontMap(layout, ...);
```

So, my suggestion is simply to use it straight up:
```c++
if (layout)
{
    assert(layout->m_FontCollection == GetFontCollection(font_map));
    GetTextMetrics(font_map, text_layout, &metrics);
}
```


We avoid defensive coding, as it grows the executable size.
No need to check values more than once.
And LLM's are notoriously bad at this, we need to stay on top of it, but I see it in so many PR's now.

JCash (changes requested):
Overall good.
* I'd like some renaming
* I'd also like to try to keep the defensive coding to a minimum, to avoid growing the executable size

engine/profiler/src/profiler_render.cpp
AGulev: 1.12.3 beta:
1. if I turn off `Runtime Generation` in `game.project` and in AppManifest I get 0.63ms for `RenderProfiler` scope (I assume it should go legacy path in such case?)
2. it's about the same if I turn it on (maybe it shouldn't be measured on my mac)

in my branch
1. 0.43ms
2. 0.45ms

AGulev: yes, maybe.

AGulev: I downloaded 1.11.2 (before layout changes) and there profiler render in the same example takes about 0.5-0..6ms.
In my branch it is 0.37-0.45ms

JCash (approved):

PR #12137, opened 2026-03-31
Store frustum has in render items to avoid already calculated culling

### LINKED ISSUE #8088: Frustum culling call when it not needed
I try to optimize my game. I look at profile and find that frustum culling for labels calls: 264 

Also i find that frustum culling for mesh and model calls(11) equals number of render.draw with frustum matrix

https://github.com/defold/defold/blob/acd26d2bf4f4e37d14b9c28837b51e0a03ce8725/engine/render/src/render/render.cpp#L695

So it looks like you update frustum for all objects in render list, before check predicates. So when i render models you check frustum for all objects(models,labels, ...)
```lua
    render.draw(self.model_pred,frusutm_draw_opts)
```

**To Reproduce (REQUIRED)**
[TestFrustum.zip](https://github.com/defold/defold/files/12807582/TestFrustum.zip)

```lua
  render.draw(self.model_pred,frusutm_draw_opts)
  render.draw(self.model2_pred,frusutm_draw_opts)
  render.draw(self.tile_pred,frusutm_draw_opts)
```
So 3 render.draw with frustum.
3 for Model and Sprite
6 for Label(2 gui)

1. Do not update label frustum when i draw model.
2. Update frustum only for objects that i draw(by predicate)

  - Version 1.6.0

**Platforms (REQUIRED):**
ALL

Jhonnyg:
The amount of time spent on frustum culling has been reduced when issuing multiple `render.draw` calls during a single frame. 

Fixes #8088

### DIFF
--- a/engine/render/src/dmsdk/render/render.h
+++ b/engine/render/src/dmsdk/render/render.h
@@ -239,11 +239,11 @@ namespace dmRender
      * Each callback then represents a draw call, and will register a RenderObject
      * @name RenderListEntry
      * @param m_WorldPosition [type: dmVMath::Point3] the world position of the object
-     * @param m_FrustumHash [type: dmhash_t] The hash of the frustum matrix used for frustum culling
+     * @param m_UserData [type: uint64_t] user data (available in the render dispatch callback)
      * @param m_Order [type: uint32_t] the order to sort on (used if m_MajorOrder != RENDER_ORDER_WORLD)
      * @param m_BatchKey [type: uint32_t] the batch key to sort on (note: only 48 bits are currently used by renderer)
      * @param m_TagListKey [type: uint32_t] the key to the list of material tags
-     * @param m_UserData [type: uint64_t] user data (available in the render dispatch callback)
+     * @param m_FrustumHash [type: uint32_t] Last combined frustum cull key (note: engine internal use only!)
      * @param m_MinorOrder [type: uint32_t:4] used to sort within a batch
      * @param m_MajorOrder [type: uint32_t:2] If RENDER_ORDER_WORLD, then sorting is done based on the world position.
                                               Otherwise the sorting uses the m_Order value directly.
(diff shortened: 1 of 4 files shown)


engine/render/src/dmsdk/render/render.h
```diff
@@ -252,6 +253,7 @@ namespace dmRender
     struct RenderListEntry
     {
         dmVMath::Point3 m_WorldPosition;
+        dmhash_t m_FrustumHash;
```
JCash: Perhaps comment it as "internal", as the caller shouldn't modify this member.

JCash: This is a somewhat "hot" struct.
Is it necessary with a 64 bit value? Is it enough with 32 bit? (I totally chose the "safe" option as it was a member of the context struct)
Is it likely we'll get many collisions from the inputs?

engine/render/src/render/render.cpp
```diff
@@ -55,6 +55,7 @@ namespace dmRender
     const dmhash_t VERTEX_STREAM_ANIMATION_DATA       = dmHashString64("animation_data");
     const dmhash_t VERTEX_STREAM_TEXTURE_TRANSFORM_2D = dmHashString64("texture_transform_2d");
     const dmhash_t SAMPLER_POSE_MATRIX_CACHE          = dmHashString64("pose_matrix_cache");
+    const dmhash_t FRUSTUM_HASH_UNINITIALIZED         = 0xFFFFFFFF;
```
JCash: 1) It's a bit weird that I used this 32bit constant, when the variable is currently 64 bit
2) Also strange that I didn't opt for `0` as the invalid number. I intentionally opted for `VISIBILITY_NONE = 0` to make it easier to clear/check data.

engine/render/src/render/render.cpp
```diff
@@ -252,10 +261,11 @@ namespace dmRender
         uint32_t size = render_list.Size();
         render_list.SetSize(size + entries);
 
-        // If we push new items after the last frustum culling, we need to reevaluate it
-        render_context->m_FrustumHash = 0xFFFFFFFF;
+        // If we push new items after the last frustum culling, we need to reevaluate them.
+        RenderListEntry* start = render_list.Begin() + size;
+        SetVisibility(entries, start, dmRender::VISIBILITY_NONE, FRUSTUM_HASH_UNINITIALIZED);
```
JCash: Here, I think we can improve on speed.
If we change defaults to 0 (e.g. FRUSTUM_HASH_UNINITIALIZED = 0), we can simply memset this entire array to 0:
```c++
memset(start, 0, size * sizeof(RenderListEntry));
```

JCash (changes requested):
Overall good, but we should probably look to minimize struct size for the cache utilization.

JCash (approved):

PR #12149, opened 2026-04-02
Avoid deadlocks when the profiler stack is unbalanced

### LINKED ISSUE #8899: On-screen profiler becomes unresponsive and no longer updates
It has been observed that the on-screen profiler can become unresponsive and no longer update with new frame data. This has happened after unloading and loading collections, but requires more investigation and a solid repro case.

AGulev:
Fixes an issue where the engine may freeze if a user calls `profiler.scope_begin("my_scope")` without later closing it with `profiler.scope_end()`, or vice versa: closing without opening. In such cases, an error will be shown and scopes will be auto-closed at the moment of detection.

Also fixes occasional freezes when using `profiler.dump_frame()`.

Fix https://github.com/defold/defold/issues/8899

### DIFF
--- a/engine/modelc/src/modelimporter_gltf.cpp
+++ b/engine/modelc/src/modelimporter_gltf.cpp
@@ -22,6 +22,7 @@
 // when locale LC_NUMERIC use commas instead of periods for decimal points
 #if defined(__APPLE__) || defined(__linux__)
 #ifdef __APPLE__
+#include <stdlib.h>
 #include <xlocale.h>
 #else
 #include <locale.h>
--- a/engine/profiler/src/profiler.cpp
+++ b/engine/profiler/src/profiler.cpp
@@ -14,6 +14,7 @@
 
 #include "profiler.h"
 
+#include <dlib/array.h>
 #include <dlib/dlib.h>
 #include <dlib/hash.h>
 #include <dlib/log.h>
@@ -27,6 +28,7 @@
 #include "profiler_render.h"
 
 #include <algorithm> // std::sort
+#include <string.h>
 
 #include <dmsdk/dlib/vmath.h>
 #include <dmsdk/extension/extension.h>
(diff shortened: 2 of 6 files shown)


engine/modelc/src/modelimporter_gltf.cpp
```diff
@@ -22,6 +22,7 @@
 // when locale LC_NUMERIC use commas instead of periods for decimal points
 #if defined(__APPLE__) || defined(__linux__)
 #ifdef __APPLE__
+#include <stdlib.h>
```
AGulev, line 25: unrelated. It's needed to build engine oon xcode 26.4

engine/profiler/src/profiler.cpp
```diff
@@ -835,8 +998,18 @@ static dmExtension::Result UpdateProfiler(dmExtension::Params* params)
     return dmExtension::RESULT_OK;
 }
 
+static ExtensionResult PreRenderProfiler(dmExtension::Params* params)
+{
+    (void) params;
+    AutoCloseLuaProfilerScopes();
```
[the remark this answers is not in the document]
AGulev: I think it's not that important

engine/profiler/src/profiler.cpp
```diff
@@ -60,8 +62,176 @@ static dmMutex::HMutex                  g_ProfilerMutex = 0;
 static dmHashTable64<int>               g_ProfilerThreadSortOrder;
 static bool                             g_ProfilerDumpNextFrame = false;
 
+struct LuaProfilerScope
+{
+    uint32_t    m_NameOffset;
+    uint64_t    m_NameHash;
```
JCash: style: Make it a habit of putting the largest items first, so that packing can more easily occur.
```c++
uint64_t    m_NameHash;
uint32_t    m_NameOffset;
```

It's likely it'll still be 2x8 bytes due to the uint64_t, but it would be possible to shrink it using "pragma pack", whereas in it's current state, it cannot.

JCash (dismissed):
looks good.

PR #11807, opened 2026-01-28
Update MbedTLS to v3

ekharkunov:
Updated MbedTLS to v3. Introduced new option in App Manifest: "Use fat MbedTLS". If that option is on - more features are available:
* MBEDTLS_FS_IO
* MBEDTLS_PK_WRITE_C
* MBEDTLS_SSL_SRV_C
* MBEDTLS_X509_CREATE_C
* MBEDTLS_X509_CRT_WRITE_C

Any additional features can be added by request.

Fixes #9527
Fixes #11259 

### Technical changes
Two additional engine libas were introduced: `mbedtls_fat` and `dlib_fat`. It was done in such way because depends on feature set in `mbedtls_config.h` inner MbedTLS structures can have different size. 

Added threaded layer. By default MbedTLS provides lock primitives only for Unix platform. So our threaded layer based on engine primitives to be cross-platform. It's needed because at least RSA context creation should be synchronized. More information https://github.com/Mbed-TLS/mbedtls-docs/blob/e137d16cca66d1a00f1bedeb6ec02d9bafaeb46c/kb/development/thread-safety-and-multi-threading.md#status-of-various-contexts-and-associated-functions

There are 2 configs which controls feature set: `mbedtls_confog_min.h` (minimal set of features required by the engine) and `mbedtls_confog_fat.h` (extended set of features). Switch between two configs is controlled by `DEFOLD_MBEDTLS_FAT` macro.

Docs PR https://github.com/defold/doc/pull/608

## PR checklist

* [x] Code
	* [x] Add engine and/or editor unit tests.
	* [x] New and changed code follows the overall code style of existing code
	* [x] Add comments where needed
* [x] Documentation
	* [x] Make sure that API documentation is updated in code comments
	* [x] Make sure that manuals are updated (in github.com/defold/doc)
* [x] Prepare pull request and affected issue for automatic release notes generator
	* [x] Pull request - Write a message that explains what this pull request does. What was the problem? How was it solved? What are the changes to APIs or the new APIs introduced? This message will be used in the generated release notes. Make sure it is well written and understandable for a user of Defold.
	* [x] Pull request - Write a pull request title that in a sentence summarises what the pull request does. Do not include "Issue-1234 ..." in the title. This text will be used in the generated release notes.
	* [x] Pull request - Link the pull request to the issue(s) it is closing. Use on of the [approved closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
[...]

PR #12153, opened 2026-04-02
Deprecate the legacy LiveUpdate APIs

### LINKED ISSUE #12032: Strip liveupdate content from game.dmanifest
In a game with a lot of liveupdate content, `game.dmanifest` size varies with amount of liveupdate content. It affects an apk/aab/ipa size and makes the size unpredictable.

`game.dmanifest` should contain information about assets included in the build, but not into liveupdate.

- One alternative is to build the game 2 times: 1 time with liveupdate content excluded from collections hierarchy to receive clean `game.dmanifest` and second time with liveupdate content included, to receive just liveupdate pack.
- Second alternative is to get `game.dmanifest` from resulted apk/aab/ipa, strip references to liveupdate content from `game.dmanifest`, put it back into apk/aab/ipa and resign.
Both alternatives require more efforts than stripping of `game.dmanifest` at build time

AGulev:
⚠️ We are deprecating old LiveUpdate functionality in favor of mount/archive-based Live Update workflows.

The following Lua API has been deprecated:
```lua
liveupdate.get_current_manifest()
liveupdate.store_resource()
liveupdate.store_manifest()
liveupdate.store_archive()
liveupdate.is_using_liveupdate_data()
collectionproxy.missing_resources()
```
`resource.*` aliases for these APIs have been removed:
```lua
resource.get_current_manifest()
resource.is_using_liveupdate_data()
resource.store_resource()
resource.store_manifest()
resource.store_archive()
```

The deprecated API still works, giving users of the old Live Update time to migrate before it is removed later this year. However, it has been removed from the documentation, and some functionality has changed.

**First**, manifest signing keys are no longer part of the pipeline, which means:
* `bob.jar` no longer accepts `--manifest-private-key` or `--manifest-public-key`
* `game.public.der` is no longer generated or bundled, and the `publickey` and `privatekey` fields in `liveupdate.settings` are now deprecated and unused
* At runtime, deprecated Live Update manifest and archive validation now check only supported engine versions instead of using bundled public/private key signatures

This change reduces the engine's build size (for example, from 2,339,290 bytes to 2,423,041 bytes for a WASM release build, about ~3.6%).

**Second**, archive builds now include a new Live Update setting, `Strip Live Update Entries from Main Manifest` in `game.project` (`liveupdate.exclude_entries_from_main_manifest`), enabled by default.

When this option is enabled for archive builds that publish Live Update content, liveupdate-only resources are excluded from the bundled `game.dmanifest`, while the published `liveupdate.game.dmanifest` still contains the full resource list required for remote content delivery. Disable this setting only if you need the deprecated behavior where excluded entries remain in the bundled `game.dmanifest`. This reduces build size and runtime memory usage, especially for projects with a large number of resources in Live Update archives.

This option is enabled by default because mount/archive-based Live Update workflows are highly dynamic and users can replace archive content at any time, manifest data cannot be relied upon for resources excluded during the bundle build.
For the same reason, `collectionproxy.missing_resources()` is now deprecated, as it depends on manifest data to detect excluded resources. It can be partially replaced by `collectionproxy.get_resources()`, which now returns an empty table `{}` when no collection exists in any mounted archive and `Strip Live Update Entries from Main Manifest` checked, indicating that the required archive has not been mounted yet.

Fix https://github.com/defold/defold/issues/12032


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ManifestBuilder.java
```diff
@@ -296,7 +120,6 @@ public static PublicKey loadPublicKey(String filepath, SignAlgorithm algorithm)
     private HashMap<String, ResourceEntry> urlToResource = new HashMap<>();
     private Set<String> supportedEngineVersions = new HashSet<String>();
     private Set<ResourceEntry> resourceEntries = new TreeSet<ResourceEntry>(new Comparator<ResourceEntry>() {
-        // We need to make sure the entries are sorted properly in order to do the binary search
```
JCash, line 299: This is still true for our base archive.
Probably good to keep the comment?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ManifestBuilder.java
```diff
@@ -309,17 +132,13 @@ private int compare(byte[] left, byte[] right) {
         }
 
         public int compare(ResourceEntry s1, ResourceEntry s2) {
-            // We want a set sorted on hashes, so we can do binary search at runtime
```
JCash, line 312: Same

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ManifestBuilder.java
```diff
@@ -397,35 +200,23 @@ public void addSupportedEngineVersion(String version) {
...
-           on are added to the list of dependants. If a CollectionProxy is
-           found that resource itself is added to the list of dependants, but
-           it is seen as a leaf and the Collection that it points to is ignored.
-
-           The reason children of a CollectionProxy is ignored is that they are
-           not required to load the parent Collection. This allows us to
-           exclude an entire Collection that is loaded through a CollectionProxy
-           and thus create a partial archive that has to be updated (through
-           LiveUpdate) before that CollectionProxy can be loaded.
-        */
```
JCash, line 427: This is also still true?

editor/resources/localization/pt_BR.editor_localization
```diff
@@ -2034,10 +2034,10 @@ form.label.liveupdate.liveupdate.output-folder-name = Pasta de destino
 form.help.liveupdate.liveupdate.output-folder-name = Nome da pasta de destino
 form.label.liveupdate.manifest = Manifesto
 form.label.liveupdate.liveupdate.publickey = Chave Pública
-form.help.liveupdate.liveupdate.publickey = Chave pública para assinatura de manifesto
+form.help.liveupdate.liveupdate.publickey = Esta configuração está obsoleta. Chaves públicas de manifesto não são mais usadas.
 form.title.liveupdate.liveupdate.publickey = Selecione a Chave Pública do Manifesto
 form.label.liveupdate.liveupdate.privatekey = Chave Privada
-form.help.liveupdate.liveupdate.privatekey = Chave privada para assinatura de manifesto
+form.help.liveupdate.liveupdate.privatekey = Esta configuração está obsoleta. Chaves privadas de manifesto não são mais usadas.
```
JCash: Is this our recommended workflow?
I though we should only commit english, and then let community translate the rest? 🤔

engine/dlib/src/dlib/crypt.cpp
```diff
@@ -12,28 +12,17 @@
...
 
-#include <stdio.h>
 #include <string.h>
-#include <stdlib.h>
 #include <assert.h>
 #include "shared_library.h"
 #include "crypt.h"
 
 #include <dlib/endian.h>
 #include <mbedtls/md5.h>
```
JCash, line 21: I don't recall how mbedtls is configured exactly (E.g. defines)
But, I'd look into how mbedtls source is added to the build. I.e. if we know we don't need some .c files, let's not even include them!
I think it may be possible to save more on size!

JCash (changes requested):
Nice one.
But I think we should keep most of the code comments?

engine/dlib/src/dlib/crypt.cpp
AGulev: I removed all the unused imports and it helps to save these 3.6% 
The rest are used.

AGulev: we have undocumented  `hashmd5()` (a separate conversation needed) function and also in `CreateResourceHash()` cpp function

AGulev:
Thank you for review! Fixed!

engine/dlib/src/dlib/crypt.cpp
JCash: > I removed all the unused imports

Well, it depends on the linker for each platform, if we use dead-stripping, and if it succeeds properly.
The best way to be sure is simply not add code that we know we don't need.
I'd still try experimenting with removing the actual .c files from the build that we know we don't need. (or the other way around, only include the .c files we need)

JCash (approved):

engine/dlib/src/dlib/crypt.cpp
AGulev: I investigated it a bit and maybe it should be done as part of https://github.com/defold/defold/pull/11807 instead?

PR #12163, opened 2026-04-05
Update GUI nodes if the texture has been replaced from a script

AGulev:
Fixed an issue where, when updating a texture in GUI, `go.set()` updates node parameters in debug builds but doesn't update them in release builds.

Fix https://github.com/defold/defold/issues/11572

### DIFF
--- a/engine/gamesys/src/gamesys/test/gui/texture_reload_override.script
+++ b/engine/gamesys/src/gamesys/test/gui/texture_reload_override.script
@@ -1,0 +1,30 @@
+-- Copyright 2020-2026 The Defold Foundation
+-- Copyright 2014-2020 King
+-- Copyright 2009-2014 Ragnar Svensson, Christian Murray
+-- Licensed under the Defold License version 1.0 (the "License"); you may not use
+-- this file except in compliance with the License.
+--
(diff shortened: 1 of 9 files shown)


engine/gamesys/src/gamesys/test/gui/texture_reload_override.script
```diff
@@ -0,0 +1,30 @@
...
+local ATLAS_B = "/gui/texture_setter_override_b.t.texturesetc"
+
+function init(self)
+    self.updated = false
+end
+
+function update(self, dt)
+    if self.updated then
+        return
+    end
```
Jhonnyg, line 25: Not sure I understand why this and the init is updated - is update called before init is called? 🤔

Jhonnyg (dismissed):

engine/gamesys/src/gamesys/test/gui/texture_reload_override.script
AGulev: it's here from another test design 
initally it was a project test in `engine` and I added this flag to run update exatly once, then i moved test but forgot to fix this sprite

it's not needed, I'm removing it

PR #12164, opened 2026-04-05
Set light buffer for compute shaders

Jhonnyg:
The light buffer uniform buffer was never applied to compute shaders.


engine/render/src/render/compute.cpp
```diff
@@ -55,6 +55,14 @@ namespace dmRender
 
         SetProgramConstantValues(render_context->m_GraphicsContext, program->m_Program, total_constants_count, program->m_NameHashToLocation, program->m_Constants, program->m_Samplers);
 
+        bool has_light_buffer;
+        uint8_t light_buffer_set;
+        uint8_t light_buffer_binding;
```
JCash: uiint8_t here, and uint16_t later on in LightBufferBindingCallback() ?

JCash (dismissed):
Overall good, but the uint8_t / uint16_t mismatch seems like an oversight?

JCash (approved):

PR #12174, opened 2026-04-07
Scene view toolbar button popup enhancements

### LINKED ISSUE #10963: Grid opacity slider exception
When we try to use the opacity slider of grid options, we might get this exception
```
Exception in thread "JavaFX Application Thread" java.lang.NullPointerException: Cannot invoke "javafx.geometry.Point2D.getX()" because "this.dragStart" is null
        at javafx.scene.control.skin.SliderSkin.lambda$initialize$14(SliderSkin.java:393)
        at com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:86)
        at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:232)
        at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:189)
        at com.sun.javafx.event.CompositeEventDispatcher.dispatchBubblingEvent(CompositeEventDispatcher.java:59)
        at com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:58)
        at com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
        at com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
        at com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
        at com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
        at com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
        at com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
        at com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
        at com.sun.javafx.event.EventUtil.fireEventImpl(EventUtil.java:74)
        at com.sun.javafx.event.EventUtil.fireEvent(EventUtil.java:54)
        at javafx.event.Event.fireEvent(Event.java:199)
        at javafx.scene.Scene$MouseHandler.process(Scene.java:3987)
        at javafx.scene.Scene.processMouseEvent(Scene.java:1893)
        at javafx.scene.Scene$ScenePeerListener.mouseEvent(Scene.java:2711)
        at com.sun.javafx.tk.quantum.GlassViewEventHandler$MouseEventNotification.run(GlassViewEventHandler.java:411)
        at com.sun.javafx.tk.quantum.GlassViewEventHandler$MouseEventNotification.run(GlassViewEventHandler.java:301)
        at com.sun.javafx.tk.quantum.GlassViewEventHandler.lambda$handleMouseEvent$2(GlassViewEventHandler.java:450)
        at com.sun.javafx.tk.quantum.QuantumToolkit.runWithoutRenderLock(QuantumToolkit.java:430)
        at com.sun.javafx.tk.quantum.GlassViewEventHandler.handleMouseEvent(GlassViewEventHandler.java:449)
        at com.sun.glass.ui.View.handleMouseEvent(View.java:560)
        at com.sun.glass.ui.View.notifyMouse(View.java:946)
        at com.sun.glass.ui.gtk.GtkApplication._runLoop(Native Method)
        at com.sun.glass.ui.gtk.GtkApplication.lambda$runLoop$10(GtkApplication.java:264)
        at java.base/java.lang.Thread.run(Thread.java:1583)
```

I cannot consistently reproduce this, but once again it seems to be related to drag events within a popup. `consumeAutoHidingEvents` might also affect the behavior of the slider, because some events are consumed. It looks similar to https://bugs.openjdk.org/browse/JDK-8095717

### LINKED ISSUE #12157: Unify scene menu button popup buttons
Currently, `grid` and `scene-visibility` modules are re-implementing the same functionality to display a popup menu to the user. #11185 introduces a new button for the perspective camera which provides a more fleshed out `popup` module to generalize functionality. However, `scene-visibility` requires a little more thorough work to merge into the new system. They have also started to implement different features for themselves and not sharing them.

This task will unify the popup system, reducing code surface and allowing all buttons to share the same feature set, such as visually unifying toggles, displaying shortcuts, localization, and allow the ability to track open buttons across the whole menu for more UX-friendly opening/closing of other buttons.

JosephFerano:
The user experience and feature set were unified between the current three scene view button popups; grid, camera, and scene visibility.

- Shortcuts can now be assigned to and their binding displayed for camera toggles, while reacting live.
- Sliders now have a text field showing their current value
- Better color picker for the grid
- Scene Visibility now has focus traversal (tab or shift+tab) and is now localized
- Miscellaneous bug fixes, styling and UX improvements

https://github.com/user-attachments/assets/817234c7-1193-464d-af85-f592e7ac2bda

Fixes #12157
Fixes #10963
Fixes #10853

### Technical changes

`scene-visibility` was re-implementing a lot of the same UI code, but it also had some features that were not yet implemented for grid/camera popovers. Moreover, fxui also implemented features that we could reuse like the color picker. cljfx also provided a way to to pass any data backing store for the settings; grid and camera use prefs while scene visibility uses a graph node.

Also added two new features to existing systems;
- `handler.clj` now has an event listener system so interested parties can listen without coupling
- `prefs.clj` now has a reset-path function to do a clean reset for a path from storage

Things were also renamed a bit better to match intent with conventions which unfortunately made git think that settings_popup is a new file, but it's just the old popup.clj.


editor/src/clj/editor/handler.clj
```diff
@@ -290,6 +291,19 @@
...
+  "Register a listener that fires after a handler runs for [command context].
+  listener-id should be unique per listener."
+  [listener-id command callback-fn]
+  (swap! state-atom assoc-in [:listeners command listener-id] callback-fn)
+  nil)
+
+(defn remove-listener!
+  "Remove a previously registered listener."
+  [listener-id command]
+  (swap! state-atom util/dissoc-in [[:listeners command listener-id]])
```
vlaaad: Shouldn't it be `[:listeners command listener-id]` instead?

editor/src/clj/editor/scene_visibility.clj
```diff
@@ -228,181 +230,146 @@
...
+        scene-vis-store (->SceneVisibilityStore scene-visibility)
+        ;; HACK: There's a visual bug where if you're hovering over the SplitPane next to Outline, if you move the
+        ;; cursor into the popup, JavaFX doesn't receive a mouse-move inside the scene view, so once you
+        ;; enter the popup, the H_RESIZE cursor stays active. As a hack, just move the scene visibility to the left by
+        ;; 13 pixels, that seems to be enough to allow the cursor to get reset.
+        controls (settings-popup/show! owner keymap localization scene-vis-store 230 -13.0 setting-descriptors nil
+                                         (fn [_]
+                                           (sync-filter-checkboxes! scene-visibility)
+                                           ;; NOTE: On close, free the references to the GUI nodes
+                                           (g/set-property! scene-visibility :ui-check-boxes nil)))]
```
vlaaad, line 328: Filter controls are not actually disabled; instead, the labels are:

Image

editor/src/clj/editor/scene_visibility.clj
```diff
@@ -228,181 +230,146 @@
...
+     (sync-filter-checkboxes! scene-visibility evaluation-context)))
+  ([scene-visibility evaluation-context]
+   (sync-filter-button-style! (g/node-value scene-visibility :app-view) scene-visibility evaluation-context)
+   (when-let [check-boxes (:filter-check-boxes (g/node-value scene-visibility :ui-check-boxes evaluation-context))]
+     (when (seq check-boxes)
+       (let [enabled? (g/node-value scene-visibility :visibility-filters-enabled? evaluation-context)]
+         (doseq [cb check-boxes]
+           (ui/enable! cb enabled?)))))))
+
+(defrecord SceneVisibilityStore [scene-visibility]
```
vlaaad: `defonce/record`

editor/src/clj/editor/ui/settings_popup.clj
```diff
@@ -0,0 +1,307 @@
...
+           [javafx.scene.control Button CheckBox Control Label PopupControl Separator Skin Slider TextField ToggleButton ToggleGroup]
+           [javafx.scene.layout HBox Priority Region StackPane VBox]
+           [javafx.scene.paint Color]
+           [javafx.stage PopupWindow$AnchorLocation]))
+
+(set! *warn-on-reflection* true)
+
+(defonce ^List axes [:x :y :z])
+
+(defprotocol SettingsStore
```
vlaaad: defonce/protocol

editor/src/clj/editor/ui/settings_popup.clj
```diff
@@ -0,0 +1,307 @@
...
+  (get-value [this key])
+  (get-default-value [this key])
+  (set-value! [this key v]))
+
+(defrecord PrefsStore [prefs prefs-prefix setting-descriptors hidden-settings on-change-fn]
```
vlaaad: defonce/record

editor/src/clj/editor/ui/settings_popup.clj
```diff
@@ -0,0 +1,307 @@
...
+      (ensure-focus-traversable!))
+    [label text-field]))
+
+(defn- make-vec3-floats-row [_popup settings-store {:keys [key]}]
+  (wrap-in-hbox (into []
+                      (comp (map (partial make-vec3-group settings-store key))
+                            (mapcat identity))
+                      axes)))
+
+;; TODO: Rename this cause plane is too specific
```
vlaaad: leftover todo?

editor/src/clj/editor/ui/settings_popup.clj
```diff
@@ -0,0 +1,307 @@
...
+                                           string/lower-case
+                                           keyword)]
+                      (set-value! settings-store key active-plane))
+                    (.setSelected old-value true))))
+    (wrap-in-hbox (concat [label] buttons))))
+
+(defn- make-color-row [popup settings-store {:keys [key label]}]
+  (let [text-field (TextField.)
+        [r g b a] (get-value settings-store key)
+        color (->> (Color. r g b a) (.toString) nnext (drop-last 2) (apply str "#"))
```
vlaaad, line 192: We have `color->web-string`... Actually... we have a much better color control in fxui ns! Maybe it does make sense to move this to cljfx after all?..

editor/src/clj/editor/camera.clj
```diff
@@ -1201,9 +1205,20 @@
...
+         {:key :speed :type :slider :label "scene-popup.camera.move-speed" :min 0.5 :max 2.0
+          :snap-to 0.25
+          :slider-value->string (fn [^double v] (str (math/round-with-precision v 0.01) "x"))}
+         {:key :look-sensitivity :type :slider :label "scene-popup.camera.look-sensitivity" :min 0.02 :max 0.4
+          :slider-value->string (fn [^double v]
+                                  (let [scaled (+ 1.0 (* 9.0 (/ (- v 0.02) (- 0.4 0.02))))]
+                                    (str (math/round-with-precision scaled 0.1))))}
+         ;; TODO: Fix ortho <-> persp fov dolly to be able to enable this
+         #_{:key :fov :type :slider :label "scene-popup.camera.fov" :min 15.0 :max 120.0
+          :slider-value->string (fn [^double v] (str (Math/round v) "°"))}
```
vlaaad: leftover code?

editor/src/clj/editor/prefs.clj
```diff
@@ -250,7 +250,8 @@
              :perspective-camera {:type :object
                                   :scope :project
                                   :properties {:speed {:type :number :default 1.0}
-                                               :look-sensitivity {:type :number :default 0.15}
+                                               :look-sensitivity {:type :number :default 0.145}
+                                               :fov {:type :number :default 37.8}
```
vlaaad: We don't have fov yet, though?

editor/src/clj/editor/prefs.clj
```diff
@@ -792,6 +792,17 @@
...
+  "Returns the default value at a prefs path, recursively constructing
+   defaults for :object types from their properties."
+  [prefs path]
+  (let [s (schema prefs path)]
+    (if (and (= :object (:type s)) (not (contains? s :default)))
+      (coll/into-> (:properties s) {}
+                   (map (fn [[prop-key _]]
+                          [prop-key (default-value-at prefs (conj path prop-key))])))
+      (default-value s))))
+
```
vlaaad, line 859: I don't like this. I think the idea of resetting to the default value is not to set the preference to the value that is currently reported as default, but instead to remove any explicit value, so that the default would be used instead. This removes unnecessary data from prefs files, and this enables us to change defaults and expect the users to get new defaults if they don't have an override. We use this approach in overrides, and in `game.project` form, for example. If you want to implement a reset, you should instead introduce a `reset!` function that clears the user-defined pref.

editor/src/clj/editor/scene_visibility.clj
```diff
@@ -17,42 +17,53 @@
...
+            [schema.core :as s]
+            [util.coll :as coll])
   (:import [javafx.geometry Point2D Pos]
+           [javafx.css PseudoClass]
            [javafx.scene Parent]
-           [javafx.scene.control CheckBox Label PopupControl Separator]
-           [javafx.scene.layout HBox Priority Region StackPane VBox]))
+           [javafx.scene.control CheckBox Label PopupControl Separator Tab ToggleButton]
+           [javafx.scene.layout HBox Priority Region StackPane VBox]
+           [javafx.stage PopupWindow$AnchorLocation]))
```
vlaaad: There are many unused imports, including added `localization`

vlaaad (changes requested):
It seems that styling somehow became less consistent... The distance between the pop-up and menubar is different now:

Image

Image

The spirit of unification actually got me thinking that we should use cljfx views, since we are unifying controls around them — they are already used in editor scripts, and the properties view. For example, there is a good implementation of a color picker that we may be able to use.

Also, I'm not convinced that `SettingsStore` is a useful abstraction. It's very hard to explain such abstract semantics... But it feels like... it could be simpler if the unification just used a completely different "axis" or "pathway". For example, the settings pop-up could define public functions to build the ui elements (cljfx...) in terms of value, on-value-changed callback, and maybe reset callback, and then this thing would be configured at `show!` call sites using prefs, etc. We already have various places, like the properties view, that work in these terms, and they don't need such store abstractions... I just feel that value/on-value-changed/on-reset is more useful and versatile than store key that reimplements it.

vlaaad:
Also, there is some leftover incomplete fov code; let's clean it up before merging.

JosephFerano:
> It seems that styling somehow became less consistent... The distance between the pop-up and menubar is different now:

That one was done on purpose;
https://github.com/defold/defold/pull/12174/changes#diff-cabde542cdbf7a84d79c525afe839aa9301c15551fac26b971a917bd83759739R324-R327
```
;; HACK: There's a visual bug where if you're hovering over the SplitPane next to Outline, if you move the
;; cursor into the popup, JavaFX doesn't receive a mouse-move inside the scene view, so once you
;; enter the popup, the H_RESIZE cursor stays active. As a hack, just move the scene visibility to the left by
;; 13 pixels, that seems to be enough to allow the cursor to get reset.
```
I can remove the shift, I just never found a good solution to this problem. But I believe everything else looks consistent right?

> The spirit of unification actually got me thinking that we should use cljfx views, since we are unifying controls around them — they are already used in editor scripts, and the properties view. For example, there is a good implementation of a color picker that we may be able to use.
> 
> Also, I'm not convinced that `SettingsStore` is a useful abstraction. It's very hard to explain such abstract semantics... But it feels like... it could be simpler if the unification just used a completely different "axis" or "pathway". For example, the settings pop-up could define public functions to build the ui elements (cljfx...) in terms of value, on-value-changed callback, and maybe reset callback, and then this thing would be configured at `show!` call sites using prefs, etc. We already have various places, like the properties view, that work in these terms, and they don't need such store abstractions... I just feel that value/on-value-changed/on-reset is more useful and versatile than store key that reimplements it.

Yes, I was hoping you would say this. I got the feeling I was re-implementing things. At several points I felt like it could be general enough where I could hook into existing parts of the UI. I also agree with your thoughts on SettingStore; it felt awkward to use throughout but given what the system was, it kinda felt like an ok compromise.  I'll take a look at properties and editor scripts, if I can delete all this code that would be a huge win!

editor/src/clj/editor/prefs.clj
JosephFerano: I see what you mean, good observation.

editor/src/clj/editor/ui/settings_popup.clj
JosephFerano: Let's do it!

editor/src/clj/editor/scene_visibility.clj
JosephFerano: Whoops! I had it working but I think I know which commit botched that...

editor/src/clj/editor/prefs.clj
```diff
@@ -770,6 +770,59 @@
...
+  (if (and (= :object (:type schema)) (coll/empty? path))
+    (e/mapcat
+      (fn [e]
+        (let [k (key e)
+              property-schema (val e)]
+          (reset-path-entries scopes property-schema [k])))
+      (:properties schema))
+    [[(-> schema :scope scopes) path]]))
+
+(defn reset-path!
```
JosephFerano, line 787: `sync-state!` takes events and writes changes to disk. However I could not find a mechanism to remove an existing value from disk, only add or overwrite. I thought I could create an event with a `::dissoc` sentinel, so it could process that when it was time to sync and write to the config, however, there is an edge case;

If I reset a value, let's say `[:scene :grid]`, if I then immediately set `[:scene :grid: :size]` to another value, because `:events` is a map with no time ordering, the `::dissoc` could occur either before or after the `:size` value event.

This can cause unexpected behavior when the timer finally writes to disk. We could keep track of any `::dissoc` events and when new events come in, we check if this new event is either a sub-path and if it is then discard the dissoc. But each new entry pays this tax and it's additional state to track, as well as needing to add plumbing to more functions to get it working.

I opted instead to write to disk immediately and remove any existing keys that are in the same path as the setting.

Now that I understand the problem better, the `::dissoc` sentinel seems somewhat viable as well, if we really prefer keeping IO on the 30 second timer, then I can remove the write to config and add the sentinel check.

vlaaad: We really prefer to keep prefs interaction non-blocking; it's expected to be usable from the UI thread, both in our code and in editor extensions.

JosephFerano: I figured out a simpler solution that doesn't require immediate IO and doesn't have the ordering issue: I now track `:removals` as a separate set. Whenever a removal is added, I search for existing events with sub-paths of the key and remove them, since they should be cleared, but any new events added after the removal should not be. Then when `sync-state!` gets called, we first process all removals, then process events, maintaining the invariant that events occurring after a removal should persist.

JosephFerano:
@vlaaad

I was able to find fixes for all the little annoying bugs, including the Color picker issue, which turned out to be unrelated to the existing issues. Please let me know if the workarounds aren't too funky or if there are other ways of going about them

JosephFerano:
I had a design proposal, doesn't have to be done now, but I think it would make the UX more consistent.

image

The general idea is to remove the little arrow icon from the eye, and also remove the "Visibility Filters" toggle. The logic being that the eye button (scene vis) has a different UX than the two other buttons with popups. The grid/camera buttons behavior is if you press the icon, you toggle the overall behavior associated with the button. In this case, the scene vis button could perform the exact behavior as the "Visibility Filters" toggle. The scene vis button would then have the same arrow icon as grid and camera (maybe style it so they're all a bit more cohesive rather than look like two separate graphical elements).

The only downside to this approach is that you lose the accelerator (shortcut tip) label you see saying you can press Ctrl+Shift+I to toggle. The workaround there would be for me to revisit one of my very first PRs where I was trying to add the shortcut to the tooltips. Now I understand how to work with both keymap and localization a bit better so I could tackle that again and that would solve the issue of shortcut discoverability, albeit still not as nice. Perhaps a label at the top of the popup? Like a header with a title, instead of it just being a bunch of controls, then the shortcut accelerator can be to the side.

Finally, now that everything is unified, we can also add the "Reset to Defaults" to scene visibility, probably trivially.

Would love to get your thoughts.

Update: There's an issue for it now #12520

editor/src/clj/editor/ui/settings_popup.clj
```diff
@@ -0,0 +1,394 @@
...
+;; The setAutoHide part: The popup loses focus when we open the "Custom Color..." color picker window, so
+;; we apply the same workaround the slider's get by disabling auto-hide and enabling it once it closes.
+(defn- ext-color-picker-focus-traversable [{:keys [desc popup]}]
+  {:fx/type fx/ext-on-instance-lifecycle
+   :on-created (fn [^javafx.scene.Node node]
+                 (when-let [tf (.lookup node ".ext-color-picker-field")]
+                   (ui/observe (.focusTraversableProperty tf)
+                     (fn [_ _ _]
+                       (.setFocusTraversable tf true)))
+                   (.setFocusTraversable tf true))
```
vlaaad: Huh, isn't it already focus-traversable?

editor/src/clj/editor/fxui.clj
```diff
@@ -150,6 +150,17 @@
    :on-created focus-when-on-scene!
    :desc desc})
 
+;; Workaround for JavaFX PopupWindow focus-traversal behavior: when controls live inside a PopupWindow, their
+;; `focusTraversable` property tends to get reset to false, breaking Tab navigation. Wrapping a control description in
+;; this component observes the property and forces it back to true. See `editor.settings-popup` for usage.
```
vlaaad: Huh? So JavaFX disables focus-traversability in popups?

editor/src/clj/editor/grid.clj
```diff
@@ -272,14 +272,33 @@
...
         grid (g/node-value scene-view-id :grid)
-        ignore-options (g/node-value grid :options)]
-    (popup/show-settings! owner prefs localization 220 [:scene :grid]
-                          [{:key :size :type :vec3-floats}
-                           {:key :active-plane :type :vec3-toggle :label "scene-popup.grid.plane"}
-                           {:key :color :type :color :label "scene-popup.grid.color"}
-                           {:key :opacity :type :slider :label "scene-popup.grid.opacity" :min 0.0 :max 1.0}]
-                          ignore-options
-                          #(invalidate-grids! app-view))))
+        ;; TODO JOE: Investigate this because maybe this broke?
```
vlaaad: leftover todo?

editor/src/clj/editor/keymap.clj
```diff
@@ -319,6 +321,7 @@
            ["Q" :scene.free-camera.down]
            ["E" :scene.free-camera.up]
            ["Shift+Back Quote" :scene.free-camera.activate]
+           ["Alt+W" :scene.free-camera.walking-mode]
```
vlaaad, line 318: Wrong locations for these shortcuts - we keep the list sorted for easier discoverability.

editor/src/clj/editor/grid.clj
```diff
@@ -272,14 +272,33 @@
...
+        ignore-options (g/node-value grid :options)
```
vlaaad: seems unused...

editor/src/clj/editor/handler.clj
```diff
@@ -376,7 +390,10 @@
 
 (defn run [[handler command-context]]
   (analytics/track-screen! (ctx->screen-name command-context))
-  (invoke-fnk handler :run command-context nil))
+  (let [result (invoke-fnk handler :run command-context nil)]
```
vlaaad: When invoke-fnk throws, it returns nil here. Should we still notify listeners when the command failed?

editor/src/clj/editor/handler.clj
```diff
@@ -376,7 +390,10 @@
...
+    (doseq [[_ callback-fn] (get-in @state-atom [:listeners (:command handler)])]
+      (callback-fn))
```
vlaaad: Listener exceptions are not isolated; we probably want to catch and report all exceptions here instead of throwing.

editor/src/clj/editor/ui/settings_popup.clj
```diff
@@ -0,0 +1,394 @@
...
+    localization         localization watcher ref used for label translation
+    state                initial state map, keyed by the :key of each descriptor.
+                         May include a :disabled set of keys for toggle-rows that are
+                         grayed out
+    width                preferred width of the popup content in pixels
+    x-offset             horizontal offset applied when positioning the popup
+    setting-descriptors  sequence of row descriptor maps (see below)
+    hidden-settings      optional set of :key values whose rows are omitted
+    on-closed            optional 0-or-1-argument callback invoked when the popup
+                         is closed
```
vlaaad: 0 or 1 arg? It's always called with 1 arg...

vlaaad (changes requested):
Looking good, but I found some issues.

Some prefs reset semantics are incorrect. Here are the failing tests that I think should succeed:
```clj

;; Resetting an object path should also clear nested values that are stored in a
;; different scope file.
(deftest reset-path-nested-scope-test
  (with-schemas {::nested-scope
                 {:type :object
                  :properties {:scene {:type :object
                                       :properties {:move-whole-pixels {:type :boolean :default true}
                                                    :grid {:type :object
                                                           :scope :project
                                                           :properties {:opacity {:type :number :default 0.25}}}}}}}}
    (let [global-file (fs/create-temp-file! "global" "test.editor_settings")
          project-file (fs/create-temp-file! "project" "test.editor_settings")
          p (prefs/make :scopes {:global global-file :project project-file}
                        :schemas [::nested-scope])]
      (prefs/set! p [:scene :move-whole-pixels] false)
      (prefs/set! p [:scene :grid :opacity] 0.75)
      (prefs/sync!)
      (is (= {:scene {:move-whole-pixels false}}
             (edn/read-string (slurp global-file))))
      (is (= {:scene {:grid {:opacity 0.75}}}
             (edn/read-string (slurp project-file))))

      (prefs/reset-path! p [:scene])

      (is (= {:move-whole-pixels true
              :grid {:opacity 0.25}}
             (prefs/get p [:scene])))
      (is (not (prefs/set? p [:scene :move-whole-pixels])))
      (is (not (prefs/set? p [:scene :grid :opacity])))
      (is (not (prefs/set? p [:scene])))
      (prefs/sync!)
      (is (= {} (edn/read-string (slurp global-file))))
      (is (= {} (edn/read-string (slurp project-file)))))))

;; Resetting an already-default path should be a no-op even when the backing
;; prefs file does not exist yet.
(deftest reset-path-missing-storage-test
  (with-schemas {::missing-storage {:type :object
                                    :properties {:opacity {:type :number :default 0.25}}}}
    (let [file (fs/create-temp-file! "missing" "test.editor_settings")
          _ (fs/delete-file! file)
          p (prefs/make :scopes {:global file}
                        :schemas [::missing-storage])]
      (is (= 0.25 (prefs/get p [:opacity])))
      (is (nil? (prefs/reset-path! p [:opacity])))
      (is (= 0.25 (prefs/get p [:opacity])))
      (is (not (prefs/set? p [:opacity]))))))

;; Syncing a reset after the backing file was deleted should write an empty
;; prefs map, not the internal missing-file sentinel.
(deftest reset-path-sync-missing-file-test
  (with-schemas {::sync-missing-file {:type :object
                                      :properties {:opacity {:type :number :default 0.25}}}}
    (let [file (fs/create-temp-file! "missing-sync" "test.editor_settings")
          p (prefs/make :scopes {:global file}
                        :schemas [::sync-missing-file])]
      (prefs/set! p [:opacity] 0.75)
      (prefs/sync!)
      (is (= {:opacity 0.75} (edn/read-string (slurp file))))
      (fs/delete-file! file)
      (prefs/reset-path! p [:opacity])
      (prefs/sync!)
      (is (= {} (edn/read-string (slurp file)))))))
```

If I were designing prefs reset from scratch, I would not introduce a separate removals key. Instead, I'd model reset as:
1. Resolving the schema at path
2. Walking the schema tree to collect all stored prefs paths
3. For each leaf, emitting a `::not-found` write
Perhaps there will be a need to do some special-casing of not-found sentinel during the write.

I haven't tried to figure out if this approach will work; this is just an intuition/input for your consideration.


There is something wrong with the color picker:
Image
1. It does not show the color dropper when hovered. Intentional?
2. There is an unexpected spacing between the text field and the icon.

editor/src/clj/editor/keymap.clj
JosephFerano: Ah sorry, I had not noticed, I'll fix all the camera ones as well

editor/src/clj/editor/ui/settings_popup.clj
JosephFerano: I'll answer in the fxui.clj comment

editor/src/clj/editor/fxui.clj
JosephFerano: Yes, I have tried searching for the exact cause of this but I haven't found much and the LLMs just say that JavaFX "resets them to false". This does seem to be the behavior though, and the previous implementation had this very workaround. `scene-visibility.clj` did not have focus traversability at all and `grid.clj` used this function;

```clojure
(defn- ensure-focus-traversable!
  [^Control control]
  (ui/observe (.focusTraversableProperty control)
    (fn [_ _ _]
      (.setFocusTraversable control true))))
```

Where every component had this in a `doto` or the like. Indeed, if I remove any of these focus traversal extensions, including the custom one for the color picker, tabbing will skip the component. I guess it's because popups in JavaFX have their own focusing rules given that it's intended to be more of a transient control.

vlaaad: Could it be caused by this? https://github.com/defold/defold/blob/f1a3237dd54b6fb271a77186a2311aaa0b036924/editor/styling/stylesheets/modules/_scene-view.scss#L3

JosephFerano: Ah that explains it. The LLM was hallucinating then. I now understand why this was implemented this way; this CSS property disables focus traversal on the toolbar because there seems to be some weird behavior when traversing the buttons (the whole panel hides sometimes). But then I tried adding `-fx-focus-traversable: true` to the controls, but this was really finicky. With some controls I got it to work with others I tried a bunch of things and it didn't stick. So perhaps they added the `ensure-focus-traversable!` since it works each time.

I'll update the comments.

vlaaad: I was thinking, maybe we should remove that css? Or replace it with something more precise? Or explicitly disable focus traversing for the menu bar? That css looks too broad, and I don't like that we need hacks to compensate for it being too broad

JosephFerano: For sure, I'll try and figure out the right combo.

JosephFerano: Wow, yeah turns out none of these "ensure focus traversal" hacks were needed! Great catch.

editor/src/clj/editor/grid.clj
JosephFerano: I went through the commit history and I think this is a feature that wasn't fully implemented as it wasn't being used
```clj
  (output options g/Any (g/constantly nil))
```
It's just nil. I couldn't figure out what the intent was, as in, what the desired feature enhancement was that required implementing filtering like this.

Either way, given that we refactored this to cljfx, filtering options or modifying their state should now be handled in the idiomatic cljfx way at the caller, settings-popup should just render what is given to it.

editor/src/clj/editor/handler.clj
JosephFerano: Ok, I added the same mechanism as invoke-fnk, but it will continue on to invoke other callbacks.

editor/src/clj/editor/handler.clj
JosephFerano: Right, I'm going to assume we don't want to notify listeners. However, this one is tricky because `invoke-fnk` returns `nil` for multiple kinds of results. I checked some code paths and there are enough callers to `invoke-fnk-ec` which calls `invoke-fnk` where it might be a bad idea to not return `nil` from an exception, as I found one or two that are calling `when` or `if` on the results. Given this, ~I think inlining the functionality is the better way to go, that way `handler/run` can differentiate between `nil` as a return value from the fnk vs an exception.~ we can add an additional arity that allows us to return a different value from exceptions

JosephFerano:
> Some prefs reset semantics are incorrect. Here are the failing tests that I think should succeed:
>
> If I were designing prefs reset from scratch, I would not introduce a separate removals key. Instead, I'd model reset as:
> 
> 1. Resolving the schema at path
> 2. Walking the schema tree to collect all stored prefs paths
> 3. For each leaf, emitting a `::not-found` write
>    Perhaps there will be a need to do some special-casing of not-found sentinel during the write.
> 
> I haven't tried to figure out if this approach will work; this is just an intuition/input for your consideration.

I went ahead and incorporated your version, as it made the tests pass. I added the tests you provided, as well as some for `incorporate-updated-storage`, everything is passing. Let me know if that's what you had in mind.

editor/src/clj/editor/prefs.clj
```diff
@@ -792,6 +846,17 @@
   (swap! global-state update :registry dissoc id)
   nil)
 
+(defn default-value-at
```
vlaaad: Seems unused now.

editor/test/editor/prefs_test.clj
```diff
@@ -607,4 +607,295 @@
            (repeatedly thread-count)
            (vec)
            (run! deref))
-      (is (= (* thread-count inc-count-per-thread) (prefs/get p [:counter]))))))
\ No newline at end of file
+      (is (= (* thread-count inc-count-per-thread) (prefs/get p [:counter]))))))
+
+(deftest incorporate-updated-storage-test
```
vlaaad: I don't like when tests test random subsets of a system that happened to be extracted into a private function. That private function — `incorporate-updated-storage` — exists within a larger private context (`sync-state!`) that may or may not impose additional constraints on the function. Testing a slice of the internal, whole, system suddenly encodes some subset of it as some sort of contract that has to be preserved. While semantically, the whole thing, and not some sub-slice of it, is a contract. We already have a test-specific function — `sync!` — that sidesteps the timer aspect of the system for performance. Let's try to test the real contracts of the system using its public APIs and backing files where possible.

editor/test/editor/prefs_test.clj
```diff
@@ -607,4 +607,295 @@
...
+    (let [file (fs/create-temp-file! "missing-sync" "test.editor_settings")
+          p (prefs/make :scopes {:global file}
+                        :schemas [::sync-missing-file])]
+      (prefs/set! p [:opacity] 0.75)
+      (prefs/sync!)
+      (is (= {:opacity 0.75} (edn/read-string (slurp file))))
+      (fs/delete-file! file)
+      (prefs/reset-path! p [:opacity])
+      (prefs/sync!)
+      (is (= {} (edn/read-string (slurp file)))))))
```
vlaaad, line 887: I guess the implementation does not say it out loud, but one of the semantics/constraints of the prefs system is that it does not require the root pref element to be an object. The other tests in this ns verify that, and functions like `safe-assoc-in` actually intentionally work on non-map values given empty paths, which, combined with schema validation, makes it all work fine, but now, with the introduction of `reset!`, it's in a half-broken state. Here are tests that I would expect to succeed that currently fail in this branch:

```clj
(deftest reset-path-root-non-object-schema-test
  (testing "string root schema"
    (with-schemas {::root-string {:type :string :default "default"}}
      (let [file (fs/create-temp-file! "root-string" "test.editor_settings")
            p (prefs/make :scopes {:global file}
                          :schemas [::root-string])]
        (prefs/set! p [] "custom")
        (prefs/sync!)
        (is (= "custom" (edn/read-string (slurp file))))

        (prefs/reset-path! p [])

        (is (= "default" (prefs/get p [])))
        (is (not (prefs/set? p [])))
        (prefs/sync!)
        (is (not= "custom" (edn/read-string (slurp file)))))))

  (testing "password root schema"
    (with-schemas {::root-password {:type :password :default "default"}}
      (let [file (fs/create-temp-file! "root-password" "test.editor_settings")
            p (prefs/make :scopes {:global file}
                          :schemas [::root-password])]
        (prefs/set! p [] "custom")
        (prefs/sync!)
        (is (= "custom" (prefs/get p [])))

        (prefs/reset-path! p [])

        (is (= "default" (prefs/get p [])))
        (is (not (prefs/set? p [])))
        (prefs/sync!)
        (is (not= "custom" (prefs/get p [])))))))
```

It's a good question what the prefs system should do when the root non-object value is reset. I'd say delete the file...

editor/src/clj/editor/grid.clj
```diff
@@ -246,14 +246,11 @@
...
 (g/defnk produce-merged-options
-  [prefs camera options]
+  [prefs camera]
   (cond-> (if prefs (prefs/get prefs [:scene :grid]) {})
           :always
           (assoc :auto-scale true)
 
-          options
-          (merge options)
-
```
vlaaad, line 256: Hmm, why was this removed? And the other stuff, like `(output options g/Any (g/constantly nil))`? As I understand, it defined a "default method" that was overridden by `CurveGrid` and `TileMapGrid`, but those overrides still remain. And it seems that it was used in the old code... but not in the new one? What should we do here?

vlaaad (commented):
Found a couple of concerning places

editor/src/clj/editor/grid.clj
JosephFerano: Ah I see now, I apologize, I missed that Grid was being inherited from. I reverted the deletion and added a filter.

editor/test/editor/prefs_test.clj
JosephFerano: I agree with this practice but I wanted to make sure the incorporation was being handled correctly since I modified that code. In this particular case, because `sync!` is performing side-effects and I'm testing a race condition, it made this particular test tricky. So I took the simpler path of testing the private function directly. I ended up modifying it with this pattern, which is still poking around internals, but at least it's testing the public contract.

```clojure
(testing "events arriving during sync I/O are not lost"
      (let [file (fs/create-temp-file! "race" "test.editor_settings")
            p (prefs/make :scopes {:global file}
                          :schemas [::race])
            real-read-config! @#'prefs/read-config!]
        (prefs/set! p [:x] 1)
        (with-redefs [prefs/read-config!
                      (fn [path]
                        (real-read-config! path)
                        (prefs/set! p [:y] 2))]
          (prefs/sync!))
        (is (= {:x 1} (edn/read-string (slurp file))))
        (is (= 1 (prefs/get p [:x])))
        (is (= 2 (prefs/get p [:y])))
        (prefs/sync!)
        (is (= {:x 1 :y 2} (edn/read-string (slurp file))))))
```

If you prefer to delete these tests, I'm ok with that as well, I now have the peace of mind that I didn't break something while modifying it :)

vlaaad (commented):
The code looks good 👍 

Here are some comments about UI/UX...

There is a missing border around the color picker icon:
Screenshot 2026-05-20 at 12 08 21

I find this toggling behavior strange; it doesn't look like it does anything:

https://github.com/user-attachments/assets/9228eb22-64b1-455f-ad98-51cd184548ff

editor/src/clj/editor/prefs.clj
```diff
@@ -770,6 +792,49 @@
...
+                                               (and (map? file-storage)
+                                                    (not (identical? ::not-found
+                                                                     (get-in file-storage config-path ::not-found)))))))))]
+               (reduce
+                 (fn [acc [file-path config-path _ :as event]]
+                   (-> acc
+                       (update :storage (fn [s]
+                                          (if (coll/empty? config-path)
+                                            (dissoc s file-path)
+                                            (update s file-path util/dissoc-in  config-path))))
```
vlaaad: Hmm... is it correct to `dissoc` the file path?

We have lookups like this:
```clj
(-> schema :scope scopes storage (get-in path ::not-found))
```

If `path` is empty, we now will get nil instead of `::not-found`:
```clj
(get-in nil [] ::not-found)
=> nil
```

vlaaad:
Regarding `dissoc`ing, it seems we should not do it after all; this test currently fails:
```clj
(deftest reset-path-root-nil-valid-schema-test
  (with-schemas {::root-nil-valid {:type :enum :values [nil "custom"]}}
    (let [file (fs/create-temp-file! "root-nil-valid" "test.editor_settings")
          p (prefs/make :scopes {:global file}
                        :schemas [::root-nil-valid])]
      (prefs/set! p [] "custom")
      (prefs/sync!)
      (is (= "custom" (prefs/get p [])))
      (is (prefs/set? p []))

      (prefs/reset-path! p [])

      (is (nil? (prefs/get p [])))
      (is (not (prefs/set? p [])))
      (prefs/sync!)
      (is (nil? (prefs/get p [])))
      (is (not (prefs/set? p [])))
      (is (not (.exists file))))))
```

editor/src/clj/editor/ui/settings_popup.clj
```diff
@@ -210,6 +210,10 @@
                             :style-class ["toggle-button" "plane-toggle" "spaced"]
                             :text (string/upper-case (name axis))
                             :selected (= axis (key state))
+                            :on-action (fn [^ActionEvent e]
+                                         (let [btn ^ToggleButton (.getSource e)]
+                                           (when-not (.isSelected btn)
+                                             (.setSelected btn true))))
```
JosephFerano, line 215: I couldn't find a better way to do this

vlaaad (changes requested):
When I hover the focused color picker, its orange border disappears. Do we really need to restyle it?

https://github.com/user-attachments/assets/d6c3f509-4c3d-4441-baa9-55148f8361ef


Prefs reset is still wrong. My example of a failing `reset-path-root-nil-valid-schema-test` was just an example of a single particular scenario. A better example that cover more types would be this test:

```clj
(deftest reset-path-root-all-schema-types-test
  (doseq [[schema-type {:keys [schema custom expected]}]
          {:boolean {:schema {:type :boolean :default true}
                     :custom false
                     :expected true}
           :string {:schema {:type :string :default "default"}
                    :custom "custom"
                    :expected "default"}
           :password {:schema {:type :password :default "default"}
                      :custom "custom"
                      :expected "default"}
           :locale {:schema {:type :locale :default "sv"}
                    :custom "en"
                    :expected "sv"}
           :keyword {:schema {:type :keyword :default :default}
                     :custom :custom
                     :expected :default}
           :integer {:schema {:type :integer :default 1}
                     :custom 2
                     :expected 1}
           :number {:schema {:type :number :default 1.0}
                    :custom 2.0
                    :expected 1.0}
           :one-of {:schema {:type :one-of
                             :schemas [{:type :object-of
                                        :key {:type :string}
                                        :val {:type :integer}
                                        :default {"default" 1}}
                                       {:type :string}]}
                    :custom "custom"
                    :expected {"default" 1}}
           :array {:schema {:type :array
                            :item {:type :string}
                            :default ["default"]}
                   :custom ["custom"]
                   :expected ["default"]}
           :set {:schema {:type :set
                          :item {:type :string}
                          :default #{"default"}}
                 :custom #{"custom"}
                 :expected #{"default"}}
           :object {:schema {:type :object
                             :properties {:x {:type :integer :default 1}}}
                    :custom {:x 2}
                    :expected {:x 1}}
           :object-of {:schema {:type :object-of
                                :key {:type :string}
                                :val {:type :integer}
                                :default {"default" 1}}
                       :custom {"custom" 2}
                       :expected {"default" 1}}
           :enum {:schema {:type :enum
                           :values [{} "custom"]}
                  :custom "custom"
                  :expected {}}
           :tuple {:schema {:type :tuple
                            :items [{:type :string} {:type :integer}]
                            :default ["default" 1]}
                   :custom ["custom" 2]
                   :expected ["default" 1]}}]
    (testing (name schema-type)
      (let [schema-id (keyword "editor.prefs-test" (str "root-" (name schema-type)))
            file (fs/create-temp-file! (str "root-" (name schema-type)) "test.editor_settings")]
        (prefs/register-schema! schema-id schema)
        (try
          (let [p (prefs/make :scopes {:global file}
                              :schemas [schema-id])]
            (prefs/set! p [] custom)
            (prefs/sync!)
            (is (= custom (prefs/get p [])))
            (is (prefs/set? p []))

            (prefs/reset-path! p [])

            (is (= expected (prefs/get p [])))
            (is (not (prefs/set? p [])))
            (prefs/sync!)
            (is (= expected (prefs/get p [])))
            (is (not (prefs/set? p []))))
          (finally
            (prefs/unregister-schema! schema-id)))))))
```

It currently has 10 failures. The issue is that `reset-path!` now uses `{}` as the in-memory replacement when resetting `[]`. That fixes root schemas where `nil` is a valid/default value, but it changes semantics for root schemas where `{}` is itself a valid stored value. It looks like using `::not-found` instead of `{}` as a not-found value sentinel for `[]` path fixes all these issues.

JosephFerano:
> When I hover the focused color picker, its orange border disappears. Do we really need to restyle it?

The reason I have to style it is that the border color for the color picker matches the background color for the settings popup, so the border wasn't missing it was just camouflaged. However, I messed up and have now applied the correct styling. It should behave correctly now. If there's a better way to do this, please let me know and I'll iterate on it.

> It currently has 10 failures. The issue is that `reset-path!` now uses `{}` as the in-memory replacement when resetting `[]`. That fixes root schemas where `nil` is a valid/default value, but it changes semantics for root schemas where `{}` is itself a valid stored value. It looks like using `::not-found` instead of `{}` as a not-found value sentinel for `[]` path fixes all these issues.

Ah I see, thank you for providing a test, these have been really helpful. Fixed.

editor/src/clj/editor/prefs.clj
```diff
@@ -460,7 +460,17 @@
...
+                                     (fn [c path val]
+                                       (cond
+                                         (and (identical? ::not-found val) (coll/empty? path)) ::not-found
+                                         (identical? ::not-found val) (if (map? c) (util/dissoc-in c path) c)
+                                         :else (safe-assoc-in c path val)))
+                                     (or (clojure.core/get acc file-path) {})
+                                     path->val)]
+                       (if (identical? ::not-found new-cfg)
+                         (dissoc acc file-path)
+                         (assoc acc file-path new-cfg))))
```
vlaaad: This whole change is unnecessary. If you roll it back to `(update acc file-path #(reduce-kv safe-assoc-in % path->val))`, everything will continue to work.

vlaaad (dismissed):
LGTM. Please consider reverting changes to `incorporate-updated-storage` before merging — they are completely unnecessary, since everything continues to work as before with the old, much simpler, code.

vlaaad (approved):
Shipit!

PR #12176, opened 2026-04-07
Light component support in editor

### LINKED ISSUE #11724: Refactor light component to use new struct type

Jhonnyg:
The light component has now been fully refactored to use the new `data resource format`, and the editor can now create, manipulate and preview light components. Currently, four light component types are supported:

image

If a `LightBuffer` struct is present (and has the correct layout) in any of the shaders, the engine will automatically bind it with the populated light data from the light components. 

Screen Recording 2026-05-04 at 15 27 56

To use the `LightBuffer` in custom shaders, please refer to `builtins/materials/lighting.glsl` for the backing layout and `builtins/materials/model_lit.fp` for how to use it. The layout of the `LightBuffer` _must_ have the exact same layout as specified in the lighting.glsl include, otherwise the engine will not be able to bind the internal uniform buffer to this resource. Also note that the `MAX_LIGHT_COUNT` must be equal or less than the `Light->Max Count` setting in the game.project.

NOTE a: There are four separate light resource types instead of a single .light resource type. This is an intentional design from our side which will eventually enable users to write custom light components that can hook into the light buffer and light systems automatically!

NOTE b: The editor can only support at most 8 lights right now. Since we are not using uniform buffers and there is no render script support (i.e no other light representation than using standard forward lighting), we need some conservative default value. 

NOTE c: The light component will not work on WebGL1 / ES2 OpenGL contexts.

Fixes #11724

### Technical changes

**Editor**
The editor marshals light preview data by storing a normalized `:editor-preview-light` payload on each light scene renderable, flattening those renderables into a pass-level :editor/preview-lights array, then packing each light into the same four-vec4 shape the runtime LightBuffer uses:

```glsl
struct Light
{
    vec4 position;        // xyz: position, w: unused
    vec4 color;           // RGBA (matches LightParams order)
    vec4 direction_range; // xyz: normalized direction; w: range
    vec4 params;          // x: type (0 dir, 1 point, 2 spot; matches dmRender::LightType)
                          // y: intensity
                          // z: innerConeAngle (radians, spot only)
                          // w: outerConeAngle (radians, spot only)
};
```

During shader reflection, we mark materials that declare a compatible LightBuffer via type introspection, and at draw time the `bind-preview-lights-for-shader!` function uploads lights_count plus lights[i].* uniforms to the active program, clamped to the smaller of the editor max and the shader’s reflected light array capacity. This is conceptually matching to what the engine is doing, which is ensuring that the uniform data is commited to the uniform buffer and bound to the correct uniform binding before a draw call happens.

**Engine**
* Uniform buffer layout is now just a hash of the shader "struct" that is used to map the buffer to. Previously it had both hash + size of the buffer for compatibility checks, but it's perfectly valid to bind a larger buffer to a smaller one so now the layout is just a hash. 
* The light buffer capacity is no longer tied to the game.project setting, they can be individually sized so you can use a smaller light buffer for slower devices

**Data changes**
* Light direction (for spot and directional lights) are no longer stored in .light file. The direction will always be derived from the parent GameObject orientation. This removes one level of complexity, as it is less likely that the direction on the component itself will be modified.


engine/gamesys/src/gamesys/resources/res_light.cpp
```diff
@@ -101,18 +102,6 @@ namespace dmGameSystem
         return LIGHT_PARSE_RESULT_OK;
     }
 
-    static LightParseResult GetVector3(const dmStructDDF::Struct* s, const char* key, dmVMath::Vector3* out)
```
Jhonnyg, line 104: Directional parameter is removed from the source files, since it comes from the rotation of the game object instead. There's no point in supporting both

editor/src/clj/editor/gl/light.clj
```diff
@@ -0,0 +1,151 @@
...
+            [editor.math :as math]
+            [editor.scene-cache :as scene-cache])
+  (:import [com.jogamp.opengl GL2]
+           [editor.gl.shader ShaderLifecycle]
+           [javax.vecmath Matrix4d Vector3d Vector4d]))
+
+(set! *warn-on-reflection* true)
+(set! *unchecked-math* :warn-on-boxed)
+
+(def ^:const default-max-preview-lights 8)
```
Jhonnyg, line 28: Editor caveat: 
We currently only bind at most the first 8 lights in the buffer. Since we don't have uniform buffer support, we need to have some conservative default value I think. I guess we could put whatever number of lights we have set in the game.project, but the editor will never be able to work with that amount of data with the current approach.

editor/src/clj/editor/label.clj
```diff
@@ -147,6 +148,7 @@
               shader (or material-shader shader)
               vertex-binding (vtx/use-with ::tris vb shader)]
           (gl/with-gl-bindings gl render-args [shader vertex-binding gpu-texture]
+            (light/bind-preview-lights-for-shader! gl shader render-args)
```
Jhonnyg, line 151: I don't know how useful it is to light labels, but if a shader wants the light buffer it should have it. In the engine, the light buffer is more agnostic in this sense, it is basing this decision from what the shader requests.

editor/src/clj/editor/light.clj
```diff
@@ -0,0 +1,970 @@
...
+
+(def ^:private light-icon-shader shaders/basic-texture-color-world-space)
+
+(def ^:private ^:const billboard-circle-segments 32)
+(def ^:private ^:const gizmo-target-pixels 100.0)
+;; Screen-space half-extent (pixels) for the origin icon quad; world size = scale-factor * this (see camera preview mesh).
+(def ^:private ^:const origin-marker-pixels 8.0)
+(def ^:private ^:const max-spot-cone-angle 180.0)
+
+(def ^:private outline-icon "icons/64/Icons_21-Light.png")
```
Jhonnyg, line 70: Maybe we need a new resource icon as well? This is the old one, and it is now used for all three light types.

Jhonnyg:
@codex review pr

editor/src/clj/editor/light.clj
```diff
@@ -0,0 +1,932 @@
...
+  (resource-node/register-ddf-resource-type workspace
+    :ext light-component-type-ext
+    :node-type LightNode
+    :ddf-type DataProto$Data
+    :load-fn load-light
+    ;; After :read-fn (read-map-without-defaults), normalize to the same map
+    ;; produce-save-value emits. Empty/sparse files become {} in the first step;
+    ;; without this, :source-value (disk) and :save-value (graph) disagree and
+    ;; sparse-protobuf-test / dirty checks fail. Not for migrating legacy formats.
+    :sanitize-fn canonicalize-light-read
```
Jhonnyg: I think this needs some attention of the seasoned editor devs, I don't know if this is the way to go. But since this is the first real use of the map structure maybe we are missing something in the loading process? Tests will fail if we omit this.

editor/src/clj/editor/material.clj
```diff
@@ -213,7 +213,8 @@
     (:shader-type+source-pairs combined-shader-info)
     (:location+attribute-name-pairs combined-shader-info)
     (:array-sampler-name->slice-sampler-names combined-shader-info)
-    (:strip-resource-binding-namespace-regex-str combined-shader-info)))
+    (:strip-resource-binding-namespace-regex-str combined-shader-info)
+    (:uses-preview-light-buffer combined-shader-info)))
```
Jhonnyg: Similar to the engine, we keep track of if any of the shaders require light data. We do this bookkeeping so we can skip even attempting to bind the light buffer if the material+shaders doesn't require it.

editor/src/clj/editor/model_scene.clj
```diff
@@ -240,6 +241,7 @@
         (shader/set-samplers-by-name shader gl name (:texture-units t)))
       (doseq [[name v] material-data]
         (shader/set-uniform shader gl name v))
+      (light/bind-preview-lights-for-shader! gl shader render-args)
```
Jhonnyg, line 244: We don't have any shared renderables processor in the editor for such "global" render data, so we need to bind a light buffer for each component that could support it (which would be all of the visible ones at least). Maybe there's a better option for this? 🤔

editor/src/clj/editor/gl/light.clj
```diff
@@ -0,0 +1,183 @@
...
+         (Math/abs (.y ws))
+         (Math/abs (.z ws)))
+    1.0))
+
+(defn renderable->std140-light [renderable]
+  (let [light-data (get-in renderable [:user-data :editor-preview-light])
+        ^Vector3d translation (or (:world-translation renderable)
+                                     (Vector3d. 0.0 0.0 0.0))
+        ^Matrix4d transform (or (:world-transform renderable)
+                                      (doto (Matrix4d.) (.setIdentity)))
```
Jhonnyg: Are we being _too_ defensive here? I would expect :world-translation to always have some amount of data before being passed into the user-data.

editor/src/clj/editor/gl/light.clj
```diff
@@ -0,0 +1,183 @@
...
+        light-type (:light-type light-data)
+        light-color (:color light-data)
+        light-intensity (or (:intensity light-data) 1.0)
+        light-range (double (or (:range light-data) 10.0))
+        scaled-range (max 0.01 (* light-range (preview-renderable-min-scale renderable)))
+        translation-v4 (Vector4d. (.x translation) (.y translation) (.z translation) 1.0)
+        color-v4 (Vector4d. (nth light-color  0) (nth light-color  1) (nth light-color  2) (nth light-color  3))
+        type-index (engine-light-type-index light-type)]
```
Jhonnyg, line 75: Maybe the naming for these temporaries can be improved 🤔

engine/shaderc/src/shaderc_spvc.cpp
```diff
@@ -241,6 +241,13 @@ namespace dmShaderc
             resource.m_Location         = spvc_compiler_get_decoration(compiler, list[i].id, SpvDecorationLocation);
             resource.m_StageFlags       = (int) stage;
 
+            resource.m_Type.m_ArraySize = 1;
+            unsigned num_array_dimensions = spvc_type_get_num_array_dimensions(type);
+            if (num_array_dimensions > 0)
+            {
+                resource.m_Type.m_ArraySize = spvc_type_get_array_dimension(type, 0);
+            }
```
Jhonnyg, line 249: Array count was not exposed for _instance_ objects. I.e:

```
uniform UniformType
{
    // members..
}  uniform_member[4]; // the size here is not exposed correctly.
```

editor/src/clj/editor/gl/light.clj
JCash: Q: Are we grabbing the 8 closest to the current camera?
Otherwise it'd be quite hard to work with in a scene imho.

engine/gamesys/src/gamesys/components/comp_light.cpp
```diff
@@ -148,8 +148,13 @@ namespace dmGameSystem
 
             dmVMath::Point3 position = dmGameObject::GetPosition(light->m_Instance);
             dmVMath::Quat rotation = dmGameObject::GetRotation(light->m_Instance);
+            dmVMath::Vector3 world_scale = dmGameObject::GetWorldScale(light->m_Instance);
+            float scale_x = dmMath::Abs(world_scale.getX());
+            float scale_y = dmMath::Abs(world_scale.getY());
+            float scale_z = dmMath::Abs(world_scale.getZ());
+            float scale = dmMath::Min(scale_x, dmMath::Min(scale_y, scale_z));
```
JCash, line 208: Seeing the abs here, I was reminded that we should also think of "negative lighting" in the future. But I think it may suffice with a boolean to enable that for a light.

engine/gamesys/src/gamesys/resources/res_light.cpp
```diff
@@ -155,29 +144,41 @@ namespace dmGameSystem
...
             HANDLE_LIGHT_PARSE_RES("spot.range", res);
+            params.m_Range = dmMath::Max(0.0f, params.m_Range);
             res = GetNumber(&light_data, "inner_cone_angle", &params.m_InnerConeAngle);
             HANDLE_LIGHT_PARSE_RES("spot.inner_cone_angle", res);
             res = GetNumber(&light_data, "outer_cone_angle", &params.m_OuterConeAngle);
             HANDLE_LIGHT_PARSE_RES("spot.outer_cone_angle", res);
+            params.m_OuterConeAngle = dmMath::Clamp(params.m_OuterConeAngle, 0.0f, 180.0f);
+            params.m_InnerConeAngle = dmMath::Clamp(params.m_InnerConeAngle, 0.0f, params.m_OuterConeAngle);
+            params.m_InnerConeAngle = params.m_InnerConeAngle * ((float) M_PI / 180.0f);
+            params.m_OuterConeAngle = params.m_OuterConeAngle * ((float) M_PI / 180.0f);
```
JCash: Hmm, at this point, we should have radians?
The editor should ofc display Euler, but the engine should use radians when possible.

engine/gamesys/src/gamesys/resources/res_light.cpp
```diff
@@ -155,29 +144,41 @@ namespace dmGameSystem
 
         res = GetNumber(&light_data, "intensity", &params.m_Intensity);
         HANDLE_LIGHT_PARSE_RES("intensity", res);
+        params.m_Intensity = dmMath::Max(0.0f, params.m_Intensity);
```
JCash: This type of check for loading content, belongs in the pipeline

engine/gamesys/src/gamesys/resources/res_light.cpp
```diff
@@ -155,29 +144,41 @@ namespace dmGameSystem
...
+            params.m_InnerConeAngle = params.m_InnerConeAngle * ((float) M_PI / 180.0f);
+            params.m_OuterConeAngle = params.m_OuterConeAngle * ((float) M_PI / 180.0f);
         }
     #undef HANDLE_LIGHT_PARSE_RES
 
+        // Local light forward axis in component space; world direction comes from GO rotation.
+        if (type == dmRender::LIGHT_TYPE_DIRECTIONAL || type == dmRender::LIGHT_TYPE_SPOT)
+        {
+            params.m_Direction = dmVMath::Vector3(0.0f, 0.0f, -1.0f);
+        }
```
JCash: At first, I'd argue to always set `params.m_Direction` to `-Z`, but then I also realized, if it's always constant, why do we have it? 🤔

engine/render/src/render/render_private.h
```diff
@@ -294,6 +294,7 @@ namespace dmRender
         dmVMath::Point3       m_Position;
         dmVMath::Vector3      m_Direction;
         const LightPrototype* m_LightPrototype;
+        float                 m_Scale;
         uint16_t              m_LightBufferIndex;
```
JCash: Since this is a future "hot path" struct, where each instance costs memory, I think it may be good to think about sizes 

Since m_LightBufferIndex is good for 65k instances, that's what we should count with:
* 2**16 * 48 bytes = 3mb for light info.

Note that these are also backed by the comp_light in some way, which further adds to it.

I wonder if we don't need m_Position/m_Direction at all in this struct? The actual payload is the backing buffer.
Since they seem to be only needed at the time of `SetLightInstance()`, we can still calculate a `needs_commit`, and then commit the data. No need to store that information?

And, if we use a handle for the lightprototype, we'd get a slimmer struct of 8 bytes, which becomes .5 mb
We could also encode it as a `uint64_t` meaning we don't have to do `new LightInstance;` for each instance:

```c++
struct LightInstance // 8 bytes
{
    uint32_t    m_LightPrototypeHandle;
    uint16_t     m_LightBufferIndex;
    uint16_t     :16;
};
```

or

```c++
uint64_t MakeLightInstance(uint32_t hprototype, uint16_t buffer_index);
```

JCash (changes requested):
Overall good.
However, I'd like to see some changes for pipeline and also use radians at runtime (not Euler)
Also, some design may be needed for the storage of the light information at runtime.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ProtoBuilders.java
```diff
@@ -62,6 +65,8 @@
 
 public class ProtoBuilders {
 
+    private static final double MAX_LIGHT_CONE_ANGLE_DEGREES = 180.0;
```
Jhonnyg: We do more validation now during project build.

editor/src/clj/editor/gl/light.clj
Jhonnyg: We are now, yes.

editor/src/clj/editor/light.clj
```diff
@@ -0,0 +1,935 @@
...
+    "point_light"))
+
+(defn- get-number [fields key default]
+  (double (or (get-in fields [key :number]) default)))
+
+(defn- get-vec4 [fields key]
+  (let [vals (get-in fields [key :list :values])]
+    (if (seq vals)
+      (mapv #(double (or (:number %) 0.0)) vals)
+      [1.0 1.0 1.0 1.0])))
```
Jhonnyg: Maybe the default value should depend on what kind of data we are expecting, a color vs a position is two different things. Maybe we shouldn't even return a value if the field is missing.

editor/src/clj/editor/light.clj
```diff
@@ -0,0 +1,935 @@
...
+              (resource-node/register-ddf-resource-type workspace
+                :ext ext
+                :node-type LightNode
+                :ddf-type DataProto$Data
+                :load-fn (partial load-light light-type)
+                ;; After :read-fn (read-map-without-defaults), normalize to the same map
+                ;; produce-save-value emits. Empty/sparse files become {} in the first step;
+                ;; without this, :source-value (disk) and :save-value (graph) disagree and
+                ;; sparse-protobuf-test / dirty checks fail. Not for migrating legacy formats.
+                :sanitize-fn (partial canonicalize-light-read light-type)
```
Jhonnyg: Pay a bit of extra attention here, the tests don't work if we don't sanitize the data here. Maybe we are hitting a new case with the protobuf map type?

Jhonnyg:
@codex review pr

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ProtoBuilders.java
```diff
@@ -446,10 +553,12 @@ public static class DisplayProfilesBuilder extends ProtoBuilder<DisplayProfiles.
     @BuilderParams(name="Data", inExts=".data", outExt=".datac")
     public static class DataBuilder extends ProtoBuilder<Data.Builder> {}
 
-    // .light is the same backing data type as the data resource.
-    // Using an alias approach has mostly semantic value for users,
-    // as the intent is more clear.
     @ProtoParams(srcClass = Data.class, messageClass = Data.class)
-    @BuilderParams(name="Light", inExts=".light", outExt=".lightc")
-    public static class LightBuilder extends ProtoBuilder<Data.Builder> {}
+    @BuilderParams(name="Light", inExts={".point_light", ".directional_light", ".spot_light"}, outExt=".lightc")
```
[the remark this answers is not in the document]
Jhonnyg: There are no existing projects, so this is moot.

Jhonnyg: We still use a shared light resource type in the engine to simplify things. The pipeline will insert the corresponding tags to each light type based on the input extension.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ProtoBuilders.java
```diff
@@ -446,10 +446,4 @@ public static class DisplayProfilesBuilder extends ProtoBuilder<DisplayProfiles.
     @BuilderParams(name="Data", inExts=".data", outExt=".datac")
     public static class DataBuilder extends ProtoBuilder<Data.Builder> {}
 
-    // .light is the same backing data type as the data resource.
-    // Using an alias approach has mostly semantic value for users,
-    // as the intent is more clear.
-    @ProtoParams(srcClass = Data.class, messageClass = Data.class)
-    @BuilderParams(name="Light", inExts=".light", outExt=".lightc")
-    public static class LightBuilder extends ProtoBuilder<Data.Builder> {}
```
Jhonnyg, line 454: It got more complicated due to validation, so I moved it to its own LightBuilder.java

editor/src/clj/editor/gl/light.clj
Jhonnyg: I changed the behaviour regarding light selection a little bit:

* Directional lights have priority, we pick at most 8 lights
* If we have room for more lights, we pick the lights that are closest to the camera

My thinking is that directional lights don't have positions, they are global. If we also pick them by position, there will be more global flickering in the scene preview which will likely annoy people. It did me anyways.

engine/gamesys/src/gamesys/components/comp_light.cpp
```diff
@@ -164,6 +169,7 @@ namespace dmGameSystem
         dmRender::SetLightBufferCount(light_context->m_RenderContext, light_context->m_MaxLightCount);
 
         ComponentTypeSetPrio(type, 1000);
+        ComponentTypeSetReadsTransforms(type, true);
```
Jhonnyg, line 227: Reasoning:

The flag is needed because light late-update reads the cached world transform and commits it into the render light buffer. Without m_ReadsTransforms, a script or animation could change scale earlier in the frame, mark transforms dirty, and then the light component would read the previous cached world scale. The engine does flush dirty transforms after all updates, but that is too late: the light payload was already written to render.

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ParseUtil.java
```diff
@@ -247,6 +247,12 @@ public Message parse(byte[] content) throws InvalidProtocolBufferException {
                 return DataProto.Data.parseFrom(content);
             }
         });
+        parseMap.put("lightc", new IParser() {
```
JCash, line 250: Note to self: Seems strange to have these hardcoded in ParseUtil, and not registered via `*Builder.java` file.
I.e. we want data oriented approach in the future (not hard coded type names)

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LightBuilder.java
```diff
@@ -0,0 +1,170 @@
...
+                return;
+            }
+            case "spot_light":
+            {
+                double range = Math.max(0.0, getRequiredNumberField(resource, lightData, "range"));
+                double outerConeAngleDegrees = Math.max(0.0, Math.min(MAX_LIGHT_CONE_ANGLE_DEGREES, getRequiredNumberField(resource, lightData, "outer_cone_angle")));
+                double innerConeAngleDegrees = Math.max(0.0, Math.min(outerConeAngleDegrees, getRequiredNumberField(resource, lightData, "inner_cone_angle")));
+                setNumberField(lightDataBuilder, "range", range);
+                setNumberField(lightDataBuilder, "inner_cone_angle", Math.toRadians(innerConeAngleDegrees));
+                setNumberField(lightDataBuilder, "outer_cone_angle", Math.toRadians(outerConeAngleDegrees));
```
JCash, line 122: 👍

JCash (dismissed):
lgtm!

matgis (dismissed):

editor/resources/templates/template.model
```diff
@@ -1,3 +1,6 @@
 mesh: ""
-material: ""
 name: "{{NAME}}"
+materials {
+  name: "default"
+  material: "/builtins/materials/model_lit.material"
+}
```
Jhonnyg: To reviewers:
We didn't use a default material here before, and we said previously that we should now use the _lit variant.

But one caveat to this is that model materials are named, so when you add a new model and want to use a mesh that doesn't have a material named "default" you will need to set the material(s) and remove the "default" binding. Perhaps it's better to not set any default material at all?

editor/src/clj/editor/gl/light.clj
```diff
@@ -175,11 +205,12 @@
                           :direction-range "direction_range"
                           :params "params")))
 
-(def ^:private lights-count-uniform-name "lights_count")
+(def ^:private light-info-uniform-name "light_info")
```
Jhonnyg, line 202: I now store the ambient light + light count in an "light_info" vec4 insted of just using light_info.x for the light count.

editor/src/clj/editor/light.clj
```diff
@@ -562,20 +586,23 @@
...
 ;; -----------------------------------------------------------------------------
-;; DirectionalLightNode
+;; AmbientLightNode
 ;; -----------------------------------------------------------------------------
 
 (defn- validate-intensity [node-id intensity]
   (validation/prop-error :fatal node-id :intensity validation/prop-negative? intensity intensity-message))
 
-(g/defnode DirectionalLightNode
+(g/defnode AmbientLightNode
```
Jhonnyg, line 606: Ambient light is basically direct light without direction vector, so I made that the base class that the directional light can inherit from.

editor/resources/templates/template.model
Jhonnyg: Decided against this as default.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/ComponentsCounter.java
```diff
@@ -327,7 +333,7 @@ public static void copyDataToBuilder(Storage storage, Project project, Collectio
         for (Map.Entry<String, Integer> entry : components.entrySet()) {
             // different input component names may have the same output name
             // for example wav and sound both are soundc
-            String name = ResourceUtil.getOutputExt("." + entry.getKey()).substring(1);
+            String name = getComponentTypeName(entry.getKey());
```
Jhonnyg, line 336: The build output generates resource_name.<light_type>.lightc, which the previous implementation didn't support.

editor/resources/templates/template.model
```diff
@@ -1,3 +1,2 @@
 mesh: ""
-material: ""
```
Jhonnyg, line 2: deprecated!

engine/engine/content/builtins/materials/model.vp
```diff
@@ -27,9 +28,9 @@ void main()
 {
     vec4 p = mtx_worldview * vec4(position.xyz, 1.0);
     var_light = mtx_view * vec4(light.xyz, 1.0);
+    var_view = mtx_view;
     var_position = p;
     var_texcoord0 = texcoord0;
     var_normal = normalize((mtx_normal * vec4(normal, 0.0)).xyz);
     gl_Position = mtx_proj * p;
 }
```
Jhonnyg, line 36: One thing to note here that we should think about:
This shader is used by the model_lit.fp as well, which requires the new var_view output. Furthermore it specifies a "light" constant in the uniform buffer that is only used by the legacy model.fp file. The optimization pass will resolve some of the redundancy, but not everything. I don't necessarily see it as a big deal, but if we want correctness here we might need new .vp files for each model_xxx variant.

engine/engine/content/builtins/materials/model_lit.fp
```diff
@@ -0,0 +1,28 @@
...
+#include "/builtins/materials/lighting.glsl"
+
+void main()
+{
+    // Texture profiles can premultiply color textures.
+    // We need to premultiply the tint as well so tint alpha scales both color and alpha.
+    vec4 tint_pm = vec4(tint.xyz * tint.w, tint.w);
+    vec4 color = texture(tex0, var_texcoord0.xy) * tint_pm;
+    vec3 lighting = diffuse_lambert(normalize(var_normal), var_position.xyz);
+    out_fragColor = vec4(color.rgb * lighting, color.a);
```
Jhonnyg, line 27: Note the alpha output here! This is a change of behaviour between model.fp and model_lit.fp. By default, the render script does not enable alpha blending for the models so setting this value will not have an effect on the output. But if you want transparent objects, it should be enough to enable blending in a custom render script, but you don't need to create a new shader that writes the alpha. So it's a bit of an improvement to model rendering I think.

matgis (dismissed):

engine/gamesys/src/gamesys/components/comp_light.cpp
```diff
@@ -61,10 +61,59 @@ namespace dmGameSystem
         LightWorld* world = new LightWorld;
         uint32_t comp_count = dmMath::Min(params.m_MaxComponentInstances, context->m_MaxLightCount);
         world->m_Components.SetCapacity(comp_count);
+        world->m_MaxComponentInstances = params.m_MaxComponentInstances;
```
JCash: 1) why isn't this using `comp_count` ?
2) what is the use for m_MaxComponentInstances? Isn't it enough that we have world->m_Components.Capacity() ?

engine/gamesys/src/gamesys/components/comp_light.cpp
```diff
@@ -61,10 +61,59 @@ namespace dmGameSystem
...
+    static dmVMath::Vector3 AmbientContribution(LightContext* context, LightResource* light_resource)
+    {
+        dmRender::HLightPrototype prototype = GetLightPrototype(light_resource);
+        if (dmRender::GetLightType(context->m_RenderContext, prototype) != dmRender::LIGHT_TYPE_AMBIENT)
+        {
+            return dmVMath::Vector3(0.0f, 0.0f, 0.0f);
+        }
+
+        dmVMath::Vector4 color = dmRender::GetLightColor(context->m_RenderContext, prototype);
+        return dmVMath::Vector3(color.getX(), color.getY(), color.getZ()) * dmRender::GetLightIntensity(context->m_RenderContext, prototype);
```
JCash: You can use `color.getXYZ()`

JCash (dismissed):
Small comment about the comp_light components array.

PR #12179, opened 2026-04-08
Fix occasional freezes on load in debug builds

### LINKED ISSUE #12154: Launching Defold game on Android results in black screen 10-20% of the time
We're seeing an issue repeatedly popping up while testing our game on Android (not sure if it's Android-specific, since we didn't test other platforms yet).

What happens is that 80-90% of app launches work fine (the game loads and starts rendering), but in the remaining 10-20% of launches we only see a black screen.

We tried changing the clear color to red in `game.project > Render` and in the successful launches, we see a flash of red just before we start rendering our loading screen. However, when an unsuccessful launch happens we don't even see the red flash, we just see a completely black screen.

We looked at the `adb` logs and - as far as we can tell - it seems to be an issue related to `GLFW`. I've attached the (filtered) logs of a successful and an unsuccessful launch, so you can see the difference between the `GLFW` output. 

**To Reproduce (REQUIRED)**
1. Bundle game for Android
2. Open and quit (terminate the application) the game multiple times
3. In some of the app launches, the game will only show a black a screen instead of rendering anything

The game should not get stuck on a black screen when launching.

  - Version 1.12.2

**Platforms (REQUIRED):**
 - Platforms: Android
 - OS: Android 14, Android 11
 - Device: Samsung Galaxy A13, ZTE Blade A52

**Logs (OPTIONAL):**
https://gist.github.com/plotlessplot/92fc994a465c09ce676fb0d24573447d

**Workaround (OPTIONAL):**
I pasted the full adb logs into a few LLMs and asked what could possibly be causing this. Most of them suggested that it could be a race condition resulting in GLFW attempting to render before a window is available. Gemini specifically then suggested that I make a few additions/changes to the AndroidManifest.xml file.

The following activity tags:
```
android:configChanges="fontScale|keyboard|keyboardHidden|locale|mcc|mnc|navigation|orientation|screenLayout|screenSize|smallestScreenSize|touchscreen|uiMode|density|layoutDirection"
android:windowDisablePreview="true" 
android:hardwareAccelerated="true"
android:launchMode="singleTop"
```

And this meta-data element:
```
<meta-data android:name="android.window.layout_in_display_cutout_mode" android:value="shortEdges" />
```

It might be completely unrelated, but after these changes I am unable to reproduce the issue.

AGulev:
Fixes an issue where the profiler may get stuck in a deadlock in some cases, which is easier to reproduce on Android.

Fix https://github.com/defold/defold/issues/12154

### DIFF
--- a/engine/dlib/src/dlib/profile/profile.cpp
+++ b/engine/dlib/src/dlib/profile/profile.cpp
@@ -119,17 +119,39 @@ void ProfileRegisterProfiler(const char* name, ProfileListener* profiler)
         }
     }
 
-    DM_MUTEX_OPTIONAL_SCOPED_LOCK(g_ProfileLock);
-
-    dmSnPrintf(profiler->m_Name, sizeof(profiler->m_Name), "%s", name);
-    profiler->m_Next = g_ProfileListeners; // it's ok to add it first
+    bool initialized = IsProfileInitialized();
     profiler->m_Ctx = 0;
-    g_ProfileListeners = profiler;
 
-    if (IsProfileInitialized())
+    // Backend callback: do not hold g_ProfileLock here
+    if (initialized)
     {
         CreateListener(profiler);
+    }
+
+    {
+        // g_ProfileLock may still be null before first PropertyInitialize()
+        DM_MUTEX_OPTIONAL_SCOPED_LOCK(g_ProfileLock);
+
+        dmSnPrintf(profiler->m_Name, sizeof(profiler->m_Name), "%s", name);
+        profiler->m_Next = g_ProfileListeners;
+
+        // Publish as disabled so the listener cannot receive property callbacks
+        // before the existing property table has been replayed
+        if (initialized)
+        {
+            profiler->m_Disabled = true;
+        }
+
+        g_ProfileListeners = profiler;
+    }
+
+    // Backend callback: do not hold g_ProfileLock here either.
+    if (initialized && profiler->m_Ctx != 0)
+    {
         CreateListenerProperties(profiler);
+
+        DM_MUTEX_SCOPED_LOCK(g_ProfileLock);
+        profiler->m_Disabled = false;
     }
 
     dmLogDebug("Registered profiler '%s'", profiler->m_Name);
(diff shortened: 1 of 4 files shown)


engine/dlib/src/dlib/profile/profile.cpp
```diff
@@ -119,17 +119,39 @@ void ProfileRegisterProfiler(const char* name, ProfileListener* profiler)
         }
     }
 
-    DM_MUTEX_OPTIONAL_SCOPED_LOCK(g_ProfileLock);
-
-    dmSnPrintf(profiler->m_Name, sizeof(profiler->m_Name), "%s", name);
-    profiler->m_Next = g_ProfileListeners; // it's ok to add it first
+    bool initialized = IsProfileInitialized();
```
AGulev, line 122: The fix. I added comment right there since it's a bit tricky ccase : 

Image

scripts/mobile/android-repack-asan.sh
```diff
@@ -29,10 +29,12 @@ function terminate() {
 }
 
 function terminate_usage() {
```
AGulev, line 31: this is unrelated, just improvement to test android faster. I use just:

>./scripts/mobile/android-repack.sh ~/Downloads/TMP/arm64-android/AndroidTest/AndroidTest.apk --reinstall

and it creates temp keys, and reintall final app. Old way when keys provided also works

engine/dlib/src/dlib/profile/profile.cpp
```diff
@@ -119,17 +119,39 @@ void ProfileRegisterProfiler(const char* name, ProfileListener* profiler)
...
+            profiler->m_Disabled = true;
+        }
+
+        g_ProfileListeners = profiler;
+    }
+
+    // Backend callback: do not hold g_ProfileLock here either.
+    if (initialized && profiler->m_Ctx != 0)
+    {
         CreateListenerProperties(profiler);
```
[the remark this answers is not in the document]
AGulev: arr, it is right

JCash (dismissed):

AGulev:
@codex review PR

PR #12180, opened 2026-04-08
Fix crash when trying to open the resource profiler on an engine after building to a target

AGulev:


engine/engine/src/engine_service.cpp
```diff
@@ -840,30 +861,39 @@ namespace dmEngineService
         dmWebServer::Send(request, PROFILER_HTML, PROFILER_HTML_SIZE);
     }
 
+    static void AddProfilerHandler(dmWebServer::HServer web_server, const char* path, const dmWebServer::HandlerParams* params)
+    {
+        dmWebServer::Result result = dmWebServer::AddHandler(web_server, path, params);
+        if (result != dmWebServer::RESULT_OK && result != dmWebServer::RESULT_HANDLER_ALREADY_REGISTRED)
```
JCash, line 867: 👍

JCash (approved):

PR #12181, opened 2026-04-08
Display a warning when a profiler scope overflow occurs

AGulev:
The app used to hang if the user made a mistake and created too many scopes. That made it hard to debug. Now the profiler warns the user about the problem.

⚠️ This introduces a breaking change to the DMSDK API:
```cpp
void ProfileSetThreadName(const char* name);
void ProfileScopeBegin(const char* name, uint64_t name_hash);
void ProfileFrameEnd(HProfile profile);
void ProfileScopeEnd(const char* name, uint64_t name_hash);
void ProfileLogText(const char* text, ...);
```
replaced with 
```cpp
ProfileResult ProfileSetThreadName(const char* name);
ProfileResult ProfileScopeBegin(const char* name, uint64_t name_hash);
ProfileResult ProfileFrameEnd(HProfile profile);
ProfileResult ProfileScopeEnd(const char* name, uint64_t name_hash);
ProfileResult ProfileLogText(const char* text, ...);
// The result can be one of the following values:
// PROFILE_RESULT_OK
// PROFILE_RESULT_NOT_INITIALIZED
// PROFILE_RESULT_OUT_OF_SAMPLES
```

### DIFF
--- a/engine/dlib/src/dmsdk/dlib/profile.h
+++ b/engine/dlib/src/dmsdk/dlib/profile.h
@@ -89,6 +89,20 @@ enum ProfilePropertyFlags
     PROFILE_PROPERTY_FRAME_RESET = 1  // reset the property each frame
 };
 
+/*# Enum to describe the result of starting a profile scope
+ * @enum
+ * @name ProfileScopeResult
+ * @member PROFILE_SCOPE_RESULT_OK
+ * @member PROFILE_SCOPE_RESULT_NOT_INITIALIZED
+ * @member PROFILE_SCOPE_RESULT_OUT_OF_SAMPLES
+ */
+enum ProfileScopeResult
+{
+    PROFILE_SCOPE_RESULT_OK = 0,
+    PROFILE_SCOPE_RESULT_NOT_INITIALIZED = 1,
+    PROFILE_SCOPE_RESULT_OUT_OF_SAMPLES = 2,
+};
+
 /*# Index constant to mark a a property as invalid
  * @enum
  * @name PROFILE_PROPERTY_INVALID_IDX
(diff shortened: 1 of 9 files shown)


engine/profiler/src/profiler.cpp
```diff
@@ -192,9 +192,14 @@ static void PushLuaProfilerScope(lua_State* L, const char* name, uint32_t name_l
     LuaProfilerScope scope;
     scope.m_NameOffset = state->m_Names.Size();
     uint64_t name_hash = 0;
-    ProfileScopeBegin(name, &name_hash);
+    ProfileScopeResult result = ProfileScopeBegin(name, &name_hash);
     scope.m_NameHash = name_hash;
 
+    if (result == PROFILE_SCOPE_RESULT_OUT_OF_SAMPLES)
+    {
+        dmLogWarning("Lua profiler scope '%s' exceeded the profiler sample limit for the current thread/frame. Additional Lua profiler scopes will be dropped until the stack unwinds.", name);
```
AGulev, line 200: I thought initially about a lua error, but it feels like profiler shouldn't break the logic flow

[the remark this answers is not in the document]
AGulev: yes, it's intended.
I want it to be noisy + sending just one scope doesn't provide enough info to find the issue.

engine/dlib/src/dmsdk/dlib/profile.h
```diff
@@ -216,7 +230,7 @@ typedef void (*ProfileDestroyListenerFn)(void* ctx);
 
 typedef void (*ProfileFrameBeginFn)(void* ctx);
 typedef void (*ProfileFrameEndFn)(void* ctx);
-typedef void (*ProfileScopeBeginFn)(void* ctx, const char* name, uint64_t name_hash);
+typedef ProfileScopeResult (*ProfileScopeBeginFn)(void* ctx, const char* name, uint64_t name_hash);
 typedef void (*ProfileScopeEndFn)(void* ctx, const char* name, uint64_t name_hash);
 typedef void (*ProfileSetThreadNameFn)(void* ctx, const char* name);
 typedef void (*ProfileLogTextFn)(void* ctx, const char* text);
```
JCash: I think then we should allow them all to return an result?

JCash (changes requested):
I like it, but I think it makes sense to have the same expectation of a result value for the other callbacks then, to make it more future proof (avoid breaking changes in the future)

engine/dlib/src/dmsdk/dlib/profile.h
AGulev: fixed!

AGulev:
@codex review

JCash (dismissed):

PR #12184, opened 2026-04-09
Switching the atlas in a sprite no longer throws an error if the current animation doesn't exist in the new atlas

### LINKED ISSUE #11913: Switch sprite atlas (almost) always causes an error to be logged.
**Describe the bug**

Cannot switch between sprite images (atlases) without triggering the following error:
```
ERROR:GAMESYS: Atlas doesn't contains animation 'white_pixel'. Animation 'stairs_middle_01' will be used
```

**To Reproduce**
1. Create a sprite with atlas `A` and animation `1`
1. Create an atlas `B` with animation `2` in there.
1. Make a script that switches the sprite's atlas and animation to `B` and `2`.

**Expected behavior**
There should be a way to switch atlases without triggering an error in the logs.

**Defold version:**
  - Version 1.12.X

**Platforms:**
 - Platforms: All

**Workaround:**
- Filter these error
- Have every atlas have one shared animation (empty) that is first selected.

Discord conversation: https://discord.com/channels/250018174974689280/817015275982815242/1473261141428932620

AGulev:
Switching the atlas previously showed an error when an animation didn't exist in the new atlas:

>ERROR:GAMESYS: Atlas doesn't contain animation 'anim'. Animation 'new_0' will be used

This used to happen even when the next line after changing the atlas switched the animation, which made the error unnecessary and annoying.

Fix https://github.com/defold/defold/issues/11913

### DIFF
--- a/engine/gamesys/src/gamesys/components/comp_sprite.cpp
+++ b/engine/gamesys/src/gamesys/components/comp_sprite.cpp
@@ -591,6 +591,21 @@ namespace dmGameSystem
         }
     }
 
+    static void ClearCurrentAnimation(SpriteComponent* component, TextureSetResource* texture_set)
+    {
+        component->m_AnimationPlayback = dmGameSystemDDF::PLAYBACK_NONE;
+        component->m_IsPlaying = 0;
+        component->m_CurrentAnimation = 0x0;
+        component->m_CurrentAnimationFrame = 0;
+
+        if (!texture_set || (uint16_t)texture_set->m_TextureSet->m_Animations.m_Count <= component->m_AnimationID)
+        {
+            component->m_AnimationID = 0;
+        }
+
+        component->m_AnimationReHash = 1;
+    }
+
     static bool PlayAnimation(SpriteComponent* component, dmhash_t animation, float offset, float playback_rate)
     {
         TextureSetResource* texture_set = GetFirstTextureSet(component);
@@ -632,11 +647,7 @@ namespace dmGameSystem
         }
         else
         {
-            component->m_AnimationReHash |= component->m_CurrentAnimation != 0x0 && component->m_CurrentAnimationFrame != 0;
-            // TODO: Why stop the current animation? Shouldn't it continue playing the current animation?
-            component->m_IsPlaying = 0;
-            component->m_CurrentAnimation = 0x0;
-            component->m_CurrentAnimationFrame = 0;
+            ClearCurrentAnimation(component, texture_set);
             dmLogError("Unable to play animation '%s' from texture '%s' since it could not be found.", dmHashReverseSafe64(animation), dmHashReverseSafe64(texture_set->m_TexturePath));
         }
         return anim_id != 0;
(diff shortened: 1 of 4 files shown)


engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -591,6 +591,21 @@ namespace dmGameSystem
...
+    static void ClearCurrentAnimation(SpriteComponent* component, TextureSetResource* texture_set)
+    {
+        component->m_AnimationPlayback = dmGameSystemDDF::PLAYBACK_NONE;
+        component->m_IsPlaying = 0;
+        component->m_CurrentAnimation = 0x0;
+        component->m_CurrentAnimationFrame = 0;
+
+        if (!texture_set || (uint16_t)texture_set->m_TextureSet->m_Animations.m_Count <= component->m_AnimationID)
+        {
+            component->m_AnimationID = 0;
```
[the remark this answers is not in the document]
AGulev: @codex review

engine/gamesys/src/gamesys/components/comp_sprite.cpp
```diff
@@ -591,6 +591,17 @@ namespace dmGameSystem
...
 
+    static void ClearCurrentAnimation(SpriteComponent* component, TextureSetResource* texture_set)
+    {
+        component->m_AnimationPlayback = dmGameSystemDDF::PLAYBACK_NONE;
+        component->m_IsPlaying = 0;
+        component->m_CurrentAnimation = 0x0;
+        component->m_CurrentAnimationFrame = 0;
+        component->m_AnimationID = 0;
+
+        component->m_AnimationReHash = 1;
```
ekharkunov, line 602: Does it make sense to rehash only if 
```cpp
component->m_CurrentAnimation != 0x0 && component->m_CurrentAnimationFrame != 0
```
as it was before?

ekharkunov (dismissed):

engine/gamesys/src/gamesys/components/comp_sprite.cpp
AGulev: only m_AnimationID was under this condition, not m_AnimationReHash

I'm not sure why this condition is needed, correct me if I'm wrong: but here we change atlas, usually it happens together with changing animations:

```lua
go.set("#sprite", "image", ...)
sprite.play_flipbook(...)
```

in this case rehash needed anywais an it will be rised when `sprite.play_flipbook(...)`. But in case user set just new image without animations we have two cases:
1. it's a mistake and we endup in this code. In this case we need to reshash so use see image changed and it's not the image he probabbly expected because he didn't set anything
2. it's intended and he has an animation with the same name, then image should just be updated  and we don't even enter this codepath

AGulev: ok, tests failed. I'm digging into it

AGulev:
I went back to more conservative change sunce a test failed. I'm not sure how valid the test result, but we diffinetly don't want to introduce any breaking chnages in this area. Could you please review it again.

ekharkunov (approved):

PR #12189, opened 2026-04-10
Fixed an issue where HTTP requests could stall during app restart

AGulev:
In some cases, an application restart could not complete an ongoing HTTP request, which could result in the app freezing.

#### Issue explanatiion

The old DoDial() flow reserved a pool slot as STATE_INUSE, but kept the raw socket only in a local variable until connect and optional SSL handshake had already succeeded. 
That means shutdown did this:
- saw an in-use slot
- found c->m_Socket == INVALID
- set m_WasShutdown
- could not actually call shutdown() on the real fd
- returned, while the worker was still blocked in poll()

#### Fix

Publish the socket into the pool slot immediately after dmSocket::New(), before blocking connect or TLS handshake continues.
In this case `Shutdown()` can hit the real fd and interrupt the wait.

```mermaid
sequenceDiagram
    participant E as Engine restart
    participant H as HttpService::Delete
    participant P as ConnectionPool slot
    participant W as HTTP worker
    participant S as OS socket

    alt Before fix
        W->>P: reserve slot (STATE_INUSE)
        W->>S: create socket, connect, poll()
        Note over P: c->m_Socket still INVALID
        E->>H: finalize/restart
        H->>P: ShutdownConnectionPool()
        P-->>H: slot is in use, but no fd to shut down
        H->>W: pthread_join()
        S-->>W: wakes only on timeout/network result
        W-->>H: exits late
    else With staged change
        W->>S: create socket
        W->>P: publish fd immediately
        W->>S: connect / poll / SSL handshake
        E->>H: finalize/restart
        H->>P: ShutdownConnectionPool()
        P->>S: shutdown(fd)
        S-->>W: poll/handshake wakes
        W->>P: sees shutdown, aborts
        W-->>H: exits quickly
        H-->>E: join returns
    end

```

### DIFF
--- a/engine/dlib/src/dlib/connection_pool.cpp
+++ b/engine/dlib/src/dlib/connection_pool.cpp
@@ -251,12 +251,12 @@ namespace dmConnectionPool
         return false;
     }
 
-    static bool PublishInProgressSocket(HPool pool, Connection* connection, dmSocket::Socket socket)
+    static bool PublishInProgressSocket(HPool pool, Connection* connection, dmSocket::Socket* socket)
     {
         DM_MUTEX_SCOPED_LOCK(pool->m_Mutex);
         if (connection->m_State == STATE_INUSE && !connection->m_WasShutdown)
         {
-            connection->m_Socket = socket;
+            connection->m_Socket = *socket;
             return true;
         }
         return false;
@@ -292,7 +292,7 @@ namespace dmConnectionPool
         // an in-flight TCP connect and a subsequent SSL handshake. If shutdown already claimed
         // the slot before we could publish, abort immediately instead of continuing with an
         // untracked socket.
-        if (!PublishInProgressSocket(pool, connection, *socket))
+        if (!PublishInProgressSocket(pool, connection, socket))
         {
             *sr = dmSocket::RESULT_CONNABORTED;
             CloseInProgressSocket(pool, connection, socket);
--- a/engine/dlib/src/dlib/connection_pool.cpp
+++ b/engine/dlib/src/dlib/connection_pool.cpp
@@ -406,7 +406,10 @@ namespace dmConnectionPool
             }
         }
 
-        dmSSLSocket::Delete(sslsocket);
+        if (sslsocket != dmSSLSocket::INVALID_SOCKET_HANDLE)
+        {
+            dmSSLSocket::Delete(sslsocket);
+        }
         if (delete_socket)
         {
             dmSocket::Delete(socket);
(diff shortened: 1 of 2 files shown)


AGulev:
the fix explained in the PR description applied to the connection with proxy and to the ssl connection as well

engine/dlib/src/dlib/connection_pool.cpp
```diff
@@ -251,24 +251,65 @@ namespace dmConnectionPool
         return false;
     }
 
-    static Result ConnectSocket(HPool pool, dmSocket::Address address, uint16_t port, int timeout, dmSocket::Socket* socket, dmSocket::Result* sr)
+    static bool PublishInProgressSocket(HPool pool, Connection* connection, dmSocket::Socket socket)
```
JCash: Nit pick, we should avoid mixing types between functions.
In this case, the file already uses `Socket*`, so let's stick to that

engine/dlib/src/dlib/connection_pool.cpp
```diff
@@ -323,19 +364,62 @@ namespace dmConnectionPool
...
+            if (c->m_Socket != socket)
+            {
+                // The slot no longer owns the raw socket (typically because Shutdown()
+                // interrupted the in-flight handshake), so the borrowed local handle
+                // must delete it after we leave the mutex.
+                delete_socket = true;
+            }
+        }
+
+        dmSSLSocket::Delete(sslsocket);
```
JCash: I don't understand this?
We just assigned it to 
Shouldn't the sslsocket be kept open?

engine/dlib/src/dlib/connection_pool.cpp
```diff
@@ -425,14 +512,31 @@ namespace dmConnectionPool
...
+                    r = RESULT_SHUT_DOWN;
+                    *sock_res = dmSocket::RESULT_CONNABORTED;
+                }
+                c->Clear();
             }
         }
 
+        // If commit succeeded, ownership was transferred to the pool slot and these locals were
+        // invalidated above. Otherwise they still own the uncommitted handles and must clean up.
+        dmSSLSocket::Delete(sslsocket);
```
JCash: I think you missed this step in the code I asked about above

JCash (changes requested):
I think need some clarification about sslsocket ownership/deletion

engine/dlib/src/dlib/connection_pool.cpp
AGulev: it stay open, see the first result OK branch
>  c->m_SSLSocket = sslsocket;
>  sslsocket = dmSSLSocket::INVALID_SOCKET_HANDLE;

I just tried to reduce branches for readability

but if you think it's not clear, I'll rewrite it to make more explicit with a bool flag

engine/dlib/src/dlib/connection_pool.cpp
AGulev: I have the same step bu no comment, maybe I should add a comment there?

engine/dlib/src/dlib/connection_pool.cpp
JCash: Yeah, it's not super clear imho. But perhaps the missing comment will help.

JCash (approved):

PR #12192, opened 2026-04-11
Support morph targets for model components

### LINKED ISSUE #8024: Support for shape blending in model animations and as a parameter for Lua code
Let me cite Blender's manual:

> Shape keys are used to deform objects into new shapes for animation. In other terminology, shape keys may be called “**morph targets**” or “**blend shapes**”.
> The most popular use cases for shape keys are in character facial animation and in tweaking and refining a skeletal rig. They are particularly useful for modeling organic soft parts and muscles where there is a need for more control over the resulting shape than what can be achieved with combination of rotation and scale.

For example, in our game, the main character has to blink his eyes to make him look natural. Unfortunately, for now the character will always look straight ahead and not blink like an unalive, and I will wait for the implementation of this feature 😢 

It would be cool to support this in 3D model animations and allow this parameter to be controlled from Lua code and via go.animate.

No ideas.

I added a demo project where the model should blink its eyes (such an animation). But it doesn't do that.

You can load the `shapes.glb` model on [playcanvas viewer](https://playcanvas.com/viewer) and see that the animation works. And if you stop playing the animation, you can control the eyes manually through the Morph Target option.

[issue_morph_target.zip](https://github.com/defold/defold/files/12675370/issue_morph_target.zip)

Jhonnyg:
glTF files with morph / blend target data are now supported for model components in the engine. Two new shader constructs are now exposed that can be used to achieve morph target rendering:

```glsl
// 1) a uniform / constant that holds the weights
uniform vs_uniforms
{
	vec4 morph_targets_weights;

	// if you need to support more than 4 weights, use an array instead:
	// vec4 morph_targets_weights[x];
};

// 2) an array sampler that holds the morph target data
uniform sampler2DArray morph_targets;
```

Demo: 
[morph-target-testing.zip](https://github.com/user-attachments/files/28347210/morph-target-testing.zip)

Future improvements:
* Instancing - a new vertex attribute semantic is needed
* Mix skinning and morph animations - currently an animation can drive either a skinned animation or a morph target animation, not both
* Play multiple morph animations at the same time (also applies to bone animations)

Fixes #8024
Fixes #3619


engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -371,23 +376,6 @@ namespace dmGameSystem
         return dmGameObject::CREATE_RESULT_OK;
     }
 
-    static bool GetSender(ModelComponent* component, dmMessage::URL* out_sender)
```
Jhonnyg, line 374: Unused.

engine/rig/src/test/test_rig.cpp
```diff
@@ -865,14 +889,14 @@ void SetUpSimpleRig(dmArray<dmRig::RigBone>& bind_pose, dmHashTable64<uint32_t>&
 
         // Meshes / skins
         mesh_set->m_Models.m_Count = 2;
-        mesh_set->m_Models.m_Data = new dmRigDDF::Model[mesh_set->m_Models.m_Count];
+        mesh_set->m_Models.m_Data = new dmRigDDF::Model[mesh_set->m_Models.m_Count]();
```
Jhonnyg, line 892: Maybe overzealous but windows tests kept crashing because of uninitialized memory.

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -1757,11 +1874,18 @@ namespace dmGameSystem
...
-            dmGameSystem::EnableRenderObjectConstants(&ro, component->m_RenderConstants);
+            constants = GetScratchConstantBuffer(world);
+        }
+
+        if (constants)
+        {
+            dmGameSystem::EnableRenderObjectConstants(&ro, constants);
         }
 
+        ApplyMorphToRenderObject(world, &ro, render_material, component, render_item, component->m_Instance);
```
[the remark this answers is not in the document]
JCash: Good question?

engine/gamesys/src/gamesys/resources/res_model.cpp
```diff
@@ -237,6 +239,169 @@ namespace dmGameSystem
...
+        set_params.m_Format = dmGraphics::TEXTURE_FORMAT_RGBA32F;
+        set_params.m_Width = width;
+        set_params.m_Height = height;
+        set_params.m_LayerCount = num_morph_targets * 3;
+        set_params.m_MinFilter = dmGraphics::TEXTURE_FILTER_NEAREST;
+        set_params.m_MagFilter = dmGraphics::TEXTURE_FILTER_NEAREST;
+
+        // TODO: Perhaps we could organize/pack the data in the asset pipeline like this instead of doing it here?
+        float* data_f = (float*)data;
+        for (uint32_t i = 0; i < num_morph_targets; ++i)
```
JCash, line 330: Yes, we should do that.
I'd rather have it in this PR tbh, but as it likely entails producing another .texturesetc, we should create a separate task as the MVP2 to do that.

I mention a separate texture because otherwise we'd keep the data around unnecessarily, due to the fact that we keep the DDF data around.

engine/gamesys/src/gamesys/resources/res_model.cpp
```diff
@@ -237,6 +239,169 @@ namespace dmGameSystem
...
+
+        max_count = dmMath::Max(max_count, base_vertex_count);
+
+        uint32_t width;
+        uint32_t height;
+        ComputeMorphTextureSize(max_count, max_tex_w, max_tex_h, &width, &height);
+        if (width > max_tex_w || height > max_tex_h || width * height < max_count)
+        {
+            dmLogError(
+                "Morph target texture for '%s' would need at least %d texels in a %d x %d atlas (per mesh morph target limits: %d x %d, GPU max %d). "
```
JCash, line 300: This message belongs in the pipeline (and eventually the editor)

engine/rig/src/rig.cpp
```diff
@@ -413,6 +414,210 @@ namespace dmRig
...
+                {
+                    model = &mesh_set->m_Models[mi];
+                    break;
+                }
+            }
+            if (!model)
+            {
+                memset(dest, 0, slot.m_MorphCount * sizeof(float));
+                continue;
+            }
```
JCash: There is a lot of extra handling here.
Given the small size of the `mesh.m_MorphBaseWeights`, I'd advocate having each model+mesh have that extra data (zero initialized if it's not used)

Then this whole step would become a single mem copy.

engine/rig/src/rig.cpp
```diff
@@ -413,6 +414,210 @@ namespace dmRig
...
+            sample = num_poses - 1;
+            fraction = 0.0f;
+        }
+
+        const float* data = track->m_Weights.m_Data;
+        for (uint32_t d = 0; d < dim; ++d)
+        {
+            uint32_t i0 = sample * dim + d;
+            uint32_t i1 = i0 + dim;
+            out_weights[d] = data[i0] + fraction * (data[i1] - data[i0]);
```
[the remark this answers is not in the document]
JCash: Looks like a valid point.

engine/rig/src/rig.cpp
```diff
@@ -1462,6 +1677,37 @@ namespace dmRig
...
+        if (instance == 0x0)
+        {
+            return 0x0;
+        }
+        int32_t idx = FindMorphSlotIndex(instance, model_id);
+        if (idx < 0)
+        {
+            return 0x0;
+        }
+        const MorphWeightSlot& slot = instance->m_MorphSlots[idx];
```
JCash: I see this pattern a couple of times.
Perhaps replace with `MorphWeightSlot* slot = FindMorphSlot()` instead?

JCash (approved):
Approved, with reservation:
* Add a MVP2 task regarding moving the runtime asset generation to the pipeline
* Add the task to our 3D project to be correctly prioritized

JCash (changes requested):
Actually, due to the out-of-bounds issue, I change to "request changes"

engine/gamesys/src/gamesys/components/comp_model.cpp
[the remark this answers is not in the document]
Jhonnyg: Yes. I've fixed it by including the morph weights into the hash. Two models with the same weights should be able to batch.

engine/gamesys/src/gamesys/resources/res_model.cpp
Jhonnyg: Perhaps we should do it the same way as we do for atlases, where we pack the data at the end of the protobuf so we can just memcpy it into a texture directly without any DDF involvement at all

engine/gamesys/src/gamesys/resources/res_model.cpp
Jhonnyg: But we also need to check this in runtime since it depends on the capabilities of the client (i.e max GPU texture size).

editor/test/integration/model_utility_test.clj
```diff
@@ -45,7 +45,7 @@
 (defn- load-scene [workspace project file-path]
   (let [resource (workspace/file-resource workspace file-path)
         node-id (project/get-resource-node project resource)]
-    (model-loader/load-scene node-id resource)))
+    (model-loader/load-scene node-id resource (g/node-value project :settings))))
```
vlaaad: I think it's cleaner to use `(project/settings project)` here

editor/src/clj/editor/model_loader.clj
```diff
@@ -25,12 +25,19 @@
...
-           [java.util ArrayList]
-           [java.io InputStream]))
+           [java.io InputStream]
+           [java.util ArrayList]))
 
 (set! *warn-on-reflection* true)
 
-(defn- load-collada-scene [^InputStream stream]
+(defn- morph-target-texture-limits [project-settings]
+  (mapv #(int (Integer/parseInt (string/trim (str (get project-settings %)))))
```
vlaaad: Huh? I see you define `max_morph_target_texture_width` and `max_morph_target_texture_height` as integers in `meta.properties`, so presumably those are always ints in the editor already? Why this dance with int->str->int conversion? Can it be non-int?

vlaaad (commented):
LGTM, but I have a question about property lookup, see comment. We might want to simplify that code, unless I'm misunderstanding it...

engine/gamesys/src/gamesys/resources/res_model.cpp
JCash: That would work too!

JCash (dismissed):
lgtm

vlaaad (approved):
LGTM!

PR #12193, opened 2026-04-11
Windows fix for file operations using Unicode paths

### LINKED ISSUE #12191: Game crashes in Windows if unusual characters present in their username
Windows 11 user launches the game, and it crashes immediately. Their username consists of Korean Hangul characters. 

log.txt is:
```
INFO:DLIB: Writing log to: ./log.txt
ERROR:DLIB: Unknown result code 42

WARNING:ENGINE: Unable to locate application support path for "Defold": (-1000)
INFO:ENGINE: Loading data from: archive:game.dmanifest
ERROR:DLIB: Unknown result code 42

ERROR:RESOURCE: Failed get application support path for "1736c8f1e1ab4d6c54d695cc316d6ce2614236f7", result = 0
INFO:STEAM: Registered steam Extension
INFO:DAABBCC: Registered daabbcc Extension
ERROR:DLIB: Unknown result code 42

ERROR:RESOURCE: Failed get application support path for "1736c8f1e1ab4d6c54d695cc316d6ce2614236f7", result = 0
ERROR:LIVEUPDATE: Could not determine liveupdate folder. Liveupdate disabled 
```

In #10722 @JCash linked to https://github.com/defold/defold/blob/dev/engine/dlib/src/dlib/sys_posix.cpp - but is that not related to Android? Since this is Windows, is https://github.com/defold/defold/blob/dev/engine/dlib/src/dlib/sys_win32.cpp the right file to look at?

Just in case it's a useful hint, LLM said:

_"MakePath_8_3() appears to build a cumulative segment path in testpath, but then calls FindFirstFileW(tmp, &fd) instead of FindFirstFileW(testpath, &fd). As a result, the function seems to query the full original path on each loop iteration rather than the current path segment being converted to 8.3 form. This could produce an invalid reconstructed short path, especially for user profile paths containing non-ASCII characters."_

<details>

<summary>More LLM analysis</summary>

The biggest problem is here:

```
wchar_t testpath[MAX_PATH] = { 0 };
wcscpy(testpath, out);
wcscat(testpath, L"/");
wcscat(testpath, cursor);

WIN32_FIND_DATAW fd{0};
HANDLE h = FindFirstFileW( tmp, &fd );
```

They build testpath, but then call FindFirstFileW(tmp, &fd) instead of:

```
HANDLE h = FindFirstFileW(testpath, &fd);
```

That looks like a real bug.

- out is supposed to hold the already-converted partial 8.3 path.
- cursor is the next original path component.
- testpath is clearly intended to represent “the current cumulative path up to this directory”.
- But FindFirstFileW() is being called on tmp, which is the full original normalized path, every iteration.

So instead of resolving each path segment step by step and reading that segment’s cAlternateFileName, it appears to be querying the whole original path every time. That makes the loop logic wrong.

In practice, that could mean:

- it keeps getting info for the full final path instead of the current segment
- it appends the wrong filename / alternate filename at each step
- the generated out path can become malformed or nonsensical
- for non-ASCII paths, where this function is specifically supposed to save the day, it may fail to produce a valid usable path

That is exactly the kind of thing that could explain your user’s issue.

There are a few other problems too.

First, this line is also suspicious:

```
if (sep)
    *sep = L'/';
```

Earlier they searched for either \\ or /, but here they always restore /, not the original separator. Since tmp was normalized earlier, maybe that is intentional, but it is still sloppy. On Windows wide paths, I would expect consistent \\ handling, or at least restoring the original character.

Second, the function comment says:

```
// Is a path contains unicode characters, we need to make it 8.3 ...
```

That matches the old workaround mentality rather than true Unicode-safe handling. Modern Windows code should generally not need to collapse Unicode paths into 8.3 names just to use them internally.

Third, it uses unsafe concatenation everywhere:

```
wcscpy(...)
wcscat(...)
```

with fixed MAX_PATH buffers and no bounds checking. That is not necessarily your current bug, but it is fragile.

The key takeaway is this:

Yes, this function contains something concretely suspicious: it computes testpath but never uses it, and instead passes tmp to FindFirstFileW().
That looks like either a copy-paste mistake or a logic bug.

</details>

JCash:
This fixes issues when a folder or a filename contain Unicode characters.
E.g. you could see `Failed get application support path` due to invalid conversions.

Fixes https://github.com/defold/defold/issues/12191

### Technical changes

Instead of keeping all strings as `char`, we use Windows functions to convert to `wchar_t` and use the `W` (wide) variants of the file operation functions.

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```

### DIFF
--- a/engine/dlib/src/dlib/sys_win32.cpp
+++ b/engine/dlib/src/dlib/sys_win32.cpp
@@ -54,6 +54,15 @@
 
 namespace dmSys
 {
+    static bool Utf8ToWide(const char* path, wchar_t* out, uint32_t out_len)
+    {
+        int required = MultiByteToWideChar(CP_UTF8, 0, path, -1, NULL, 0);
+        if (required == 0 || required > (int)out_len)
+            return false;
+
+        return MultiByteToWideChar(CP_UTF8, 0, path, -1, out, required) != 0;
+    }
+
     char* GetEnv(const char* name)
     {
         return getenv(name);
@@ -71,7 +80,12 @@ namespace dmSys
 
     Result Rename(const char* dst_filename, const char* src_filename)
     {
-        bool rename_result = MoveFileExA(src_filename, dst_filename, MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH) != 0;
+        wchar_t wdst_filename[MAX_PATH];
+        wchar_t wsrc_filename[MAX_PATH];
+        if (!Utf8ToWide(dst_filename, wdst_filename, MAX_PATH) || !Utf8ToWide(src_filename, wsrc_filename, MAX_PATH))
+            return RESULT_UNKNOWN;
+
+        bool rename_result = MoveFileExW(wsrc_filename, wdst_filename, MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH) != 0;
         if (rename_result)
         {
             return RESULT_OK;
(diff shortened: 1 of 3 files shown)


engine/dlib/src/dlib/sys_win32.cpp
```diff
@@ -70,7 +80,12 @@ namespace dmSys
 
     Result Rename(const char* dst_filename, const char* src_filename)
     {
-        bool rename_result = MoveFileExA(src_filename, dst_filename, MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH) != 0;
+        wchar_t wdst_filename[MAX_PATH];
+        wchar_t wsrc_filename[MAX_PATH];
+        if (!Utf8ToWide(dst_filename, wdst_filename, MAX_PATH) || !Utf8ToWide(src_filename, wsrc_filename, MAX_PATH))
+            return RESULT_UNKNOWN;
+
+        bool rename_result = MoveFileExW(wsrc_filename, wdst_filename, MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH) != 0;
```
JCash: Moving from `A` (ascii) to `W` (wide) functions.

engine/dlib/src/dlib/sys_win32.cpp
```diff
@@ -92,92 +107,27 @@ namespace dmSys
...
-        {
-            ++path;
-        }
-        return path;
+        return GetApplicationSupportPath(application_name, path, path_len);
     }
 
-    // Is a path contains unicode characters, we need to make it 8.3 in order to properly use char* functions
-    static bool MakePath_8_3(const wchar_t* wpath, wchar_t* out)
+    Result GetApplicationSupportPath(const char* application_support_path, const char* application_name, char* path, uint32_t path_len)
```
JCash: Using the standard functions, means we don't need MakePath_8_3 anymore.

AGulev (approved):

AGulev:
@codex review

PR #12195, opened 2026-04-11
Editor: Updated scene view icon for lights

matgis:
Added a scene-view specific light icon.

Since it is currently drawn as a billboard rather than a screen-aligned rectangle, I enabled linear filtering to avoid flickering when the icon is at the edge of the viewport.


Jhonnyg (approved):
Thanks!

PR #12199, opened 2026-04-13
Apply filter params when uploading texture data

### LINKED ISSUE #12187: Bitmap fonts has unwanted antialiasing
- bitmap fonts started showing unwanted antialiasing on ios devices with existing projects.
- before it rendered like how I intended. it seems to render correctly when I playtest on mac.
- result is same with builtin `font` or `font_fnt` materials.

**To Reproduce (REQUIRED)**
1. create a gui label
2. assign a bitmap (pixelart) font
3. select nearest filtering in game.project
4. deploy to an ios device

bitmap fonts respect aa setting.

  - Version 1.12.3

**Platforms (REQUIRED):**
 - Platforms: IOS (didn't test with others)
 - OS: Macos

can provide if needed.

Jhonnyg:
Fixed an issue where updating font data in OpenGL changes texture filtering parameters but didn't update the sampler dirty state. This caused a mismatch between the expected filtering state since the sampler settings were never applied.

Fixes #12187

### DIFF
--- a/engine/graphics/src/opengl/graphics_opengl.cpp
+++ b/engine/graphics/src/opengl/graphics_opengl.cpp
@@ -4617,9 +4617,9 @@ static void LogFrameBufferError(GLenum status)
         }
         free(gl_texture_ids);
 
-        OpenGLTexture* tex    = new OpenGLTexture();
+        OpenGLTexture* tex           = new OpenGLTexture();
         tex->m_Base.m_Type           = texture_type;
-        tex->m_TextureIds     = static_ids;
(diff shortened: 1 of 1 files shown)


JCash (approved):
Looks good.
Is it testable with our geaphics tests?

Jhonnyg:
> Looks good. Is it testable with our geaphics tests?

Hm not with the current implementation - I think we need to keep making the graphics adapters more data-driven to test it. Or perhaps we can add some tests into your screenshot based test frameworks.

PR #12213, opened 2026-04-14
Use arch as a part of unpack dir name

### LINKED ISSUE #10843: Issue switching from Intel<->arm editor, if some subpath(s) already exist
If the sha path `d8e6e73` exists, the editor seems to assume it doesn't need to unpack anything, and stalls indefinitely on the loading screen.

This makes it difficult for users to switch from an Intel version to an Arm version of the editor.

### Workaround

Remove the folder `~/Library/Application\ Support/Defold`

vlaaad:
Fix #10843


vlaaad:
Tested on Mac with x86_64 and arm64 builds.

AGulev (approved):

PR #12220, opened 2026-04-15
Removed support for the `js-web` (asmjs) build from the engine

AGulev:
In this version, support for the js-web (asmjs) platform in the engine has been removed,[ as previously announced](https://forum.defold.com/t/removing-collada-support-and-the-asm-js-platform/82506).

This means `bob.jar` no longer supports `js-web` as a platform or architecture. Make sure your build scripts and CI have been updated accordingly.

Fix https://github.com/defold/defold/issues/11622


editor/resources/bundle.editor_script
```diff
@@ -60,18 +60,51 @@ end
 
 -- HTML5
 
+local function sanitize_html5_config(config)
```
AGulev: user may have `js-web` picked in his saved settings, let's sanitize it

editor/src/clj/editor/engine/native_extensions.clj
```diff
@@ -56,8 +56,10 @@
                                      :library-paths #{"android" "armv7-android"}}
    (.getPair Platform/Arm64Android) {:platform      "arm64-android"
                                      :library-paths #{"android" "arm64-android"}}
-   (.getPair Platform/JsWeb)        {:platform      "js-web"
-                                     :library-paths #{"web" "js-web"}}
+   (.getPair Platform/WasmWeb)      {:platform      "wasm-web"
+                                     :library-paths #{"web" "wasm-web"}}
+   (.getPair Platform/WasmPthreadWeb) {:platform      "wasm_pthread-web"
+                                       :library-paths #{"web" "wasm_pthread-web"}}
```
AGulev, line 62: we test all in bob, do we need bth here?

editor/src/clj/editor/app_manifest.clj
```diff
@@ -523,6 +523,16 @@
     :both webgpu-toggles
     :web-gl))
 
+(defn- normalize-manifest [manifest]
```
AGulev: app manifest cleanup

editor/src/clj/editor/prefs.clj
```diff
@@ -839,30 +839,58 @@
...
+      prefs
+      []
+      (->> legacy-key->path
+           (e/keep (fn [[legacy-key path]]
+                     (when (fresh-migration-target? prefs path)
+                       (when-some [v (legacy-pref-value legacy-prefs legacy-key)]
+                         (coll/pair path v)))))
+           (reduce #(assoc-in %1 (key %2) (val %2)) {})))))
+
+(defn- migrate-project-html5-architecture-prefs! [project-prefs]
```
AGulev: legacy options migration

editor/resources/bundle.editor_script
vlaaad: We have a schema for sanitization. If the user persisted `js-web` in preferences, but the prefs schema no longer has it, the default (`wasm-web`) will be returned. This code should be deleted.

editor/src/clj/editor/app_manifest.clj
```diff
@@ -560,30 +570,33 @@
...
                   [:wasm-web platform-pattern]
                   [:wasm_pthread-web platform-pattern]]]]))
 
+(defn serialize-manifest [manifest indent-type]
+  (util/split-lines
+    (yaml/dump manifest
+               :order-pattern app-manifest-key-order-pattern
+               :indent (case indent-type
+                         :two-spaces 2
+                         4))))
```
vlaaad: I don't think this is useful for tests, since you no longer test the actual behavior - you should inline this fn back and instead test `lines` output.

editor/src/clj/editor/app_manifest.clj
```diff
@@ -523,6 +523,16 @@
     :both webgpu-toggles
     :web-gl))
 
+(defn- normalize-manifest [manifest]
+  (if-some [platforms (and (map? manifest)
+                           (map? (:platforms manifest))
+                           (:platforms manifest))]
```
vlaaad: This line is unnecessary.

editor/src/clj/editor/app_manifest.clj
```diff
@@ -523,6 +523,16 @@
...
+    (let [platforms (dissoc platforms :js-web)]
+      (cond-> (dissoc manifest :platforms)
+        (seq platforms)
+        (assoc :platforms platforms)))
```
vlaaad: Use `editor.util/dissoc-in` instead.

editor/src/clj/editor/app_manifest.clj
```diff
@@ -560,30 +570,33 @@
...
+
 (g/defnode AppManifestNode
   (inherits r/CodeEditorResourceNode)
   (output parsed-manifest g/Any :cached (g/fnk [lines _node-id resource]
                                           (resource-io/with-error-translation resource _node-id :manifest
                                             (yaml/with-error-translation _node-id :manifest resource
                                               (with-open [reader (data/lines-reader lines)]
-                                                (yaml/load reader keyword))))))
+                                                (some-> (yaml/load reader keyword)
+                                                        normalize-manifest))))))
```
vlaaad: We don't sanitize in graph nodes — this applies sanitization on every manifest modification instead of doing it once on load. Instead, we should sanitize on load, which we typically do using `:read-fn`, though it seems it might be more involved in the case of code resources which have additional logic around read-fn identities...

editor/src/clj/editor/prefs.clj
```diff
@@ -942,16 +970,14 @@
...
                                "bundle-windows-platform" [:bundle :windows :platform]}
                               ;; these prefs already used project scope
                               (e/map #(coll/pair (str (key %) suffix) (val %))
                                      {"bundle-output-directory" [:bundle :output-directory]
                                       "recent-files-by-workspace-root" [:workflow :recent-files]
                                       "instance-count" [:run :instance-count]
                                       "simulate-rotated-device" [:run :simulate-rotated-device]
-                                      "simulated-resolution" [:run :simulated-resolution]})))))
+                                      "simulated-resolution" [:run :simulated-resolution]})))
+    (migrate-project-html5-architecture-prefs! project-prefs)))
```
vlaaad: Please roll back all changes to the migration code, and instead either delete the whole migration code and all its uses completely, or only remove the migration of old js-web keys. There is a comment on line 838:
```
;; TODO: remove migration from legacy prefs after sufficient time has passed (e.g. after 2025-10-15)
```
This is a temporary migration code; we don't maintain/improve it.

editor/test/editor/prefs_test.clj
```diff
@@ -313,6 +323,53 @@
...
+        (let [project-file (fs/create-temp-file! "project" "test.editor_settings")
+              p (prefs/make :scopes {:project project-file}
+                            :schemas [::bundle-html5-migration])]
+          (prefs/migrate-project-prefs! p)
+          (is (false? (prefs/get p [:bundle :html5 :architecture :wasm-web])))))
+      (finally
+        (.remove legacy-prefs legacy-js-web-key)
+        (.remove legacy-prefs legacy-wasm-web-key)
+        (.flush legacy-prefs)))))
+
```
vlaaad: remove these tests, the old prefs are long gone.

editor/test/integration/app_manifest_test.clj
```diff
@@ -390,3 +390,15 @@
       (g/set-property! manifest :exclude-record false)
       (is (false? (string/includes? (text) "record_null")))
       (is (false? (g/node-value manifest :exclude-record))))))
+
+(deftest legacy-js-web-removal-test
+  (test-util/with-loaded-project
+    (let [manifest (test-util/resource-node project "/app_manifest/legacy_js_web.appmanifest")
+          manifest-data (g/node-value manifest :manifest)
+          text #(slurp (data/lines-reader (app-manifest/serialize-manifest manifest-data (g/node-value manifest :indent-type))))]
```
vlaaad: See comment on app_manifest.clj about testing the right thing.

vlaaad (changes requested):

JCash (dismissed):

editor/src/clj/editor/app_manifest.clj
```diff
@@ -523,6 +524,35 @@
     :both webgpu-toggles
     :web-gl))
 
+(defn- normalize-manifest [manifest]
+  (if (map? manifest)
+    (util/dissoc-in manifest [:platforms :js-web])
+    manifest))
+
+(declare ^:private app-manifest-key-order-pattern)
```
vlaaad: This is unnecessary, move `sanitize-manifest-lines` and `read-app-manifest-fn` below `app-manifest-key-order-pattern`.

editor/src/clj/editor/code/resource.clj
```diff
@@ -204,36 +206,45 @@
...
                                        (loaded-unmodified-lines _node-id)))))
 
 (defn code-resource-type? [resource-type]
-  (and (:textual? resource-type)
-       (identical? read-fn (:read-fn resource-type))
-       (identical? write-fn (:write-fn resource-type))))
+  (or (:code-resource? resource-type)
+      (and (:textual? resource-type)
+           (identical? read-fn (:read-fn resource-type))
+           (identical? write-fn (:write-fn resource-type)))))
```
vlaaad: Not sure we want to support BOTH ways...

editor/src/clj/editor/app_manifest.clj
```diff
@@ -523,6 +524,35 @@
     :both webgpu-toggles
     :web-gl))
 
+(defn- normalize-manifest [manifest]
+  (if (map? manifest)
+    (util/dissoc-in manifest [:platforms :js-web])
```
vlaaad: Actually, now I understand that `dissoc-in` doesn't work here either, because since the file is user-editable, it's not guaranteed that `:platforms` and `:js-web` are maps. You need to do a special recursive function that checks if map, updates/dissocs, while optionally removing the empty map...

vlaaad (commented):
Maybe Mats should overtake this...

editor/resources/templates/template.appmanifest
```diff
@@ -1,12 +1,4 @@
 platforms:
-  armv7-ios:
```
AGulev, line 2: it's removed years ago (1.2.193)

AGulev:
@vlaaad no more migration logic at all!

vlaaad (approved):

PR #12227, opened 2026-04-15
Add careful ownership rules for TextLayout to avoid use-after-free

AGulev:


engine/engine/src/test/test_profilerext_lua.cpp
```diff
@@ -187,7 +187,7 @@ TEST_F(ProfilerExtLuaTest, CoroutineScopesAreAutoClosedOnPreRender)
         "    profiler.scope_begin(\"coroutine_outer\")\n"
         "    profiler.scope_begin(\"coroutine_inner\")\n"
         "    coroutine.yield()\n"
-        "    profiler.scope_end()\n"
```
AGulev, line 190: Unrealated fix. 
Related to ASAN

engine/font/src/text_layout.h
```diff
@@ -18,11 +18,12 @@
 #include <dmsdk/dlib/array.h>
 #include <dmsdk/font/text_layout.h>
 
-typedef void (*FTextLayoutFree)(HTextLayout layout);
+typedef void (*FTextLayoutDestroy)(HTextLayout layout);
```
AGulev, line 21: just semantics change: Create/Destroy
Acquire/release

engine/profiler/src/profiler.cpp
```diff
@@ -1097,6 +1099,9 @@ static dmExtension::Result AppFinalizeProfiler(dmExtension::AppParams* params)
     }
 
     dmProfiler::SetEnabled(false);
+    // The profiler UI is process-static, so app-finalize must also tear it
+    // down to avoid carrying prepared text layouts across engine reboots.
+    DeleteProfilerUI();
```
AGulev, line 1104: To make sure nothing survies reboot

engine/gui/src/gui.cpp
```diff
@@ -3761,9 +3761,13 @@ namespace dmGui
...
         InternalNode* n = GetNode(scene, node);
         TextLayout& current_text_layout = n->m_Node.m_TextLayout;
+        if (text_layout.m_Handle && current_text_layout.m_Handle != text_layout.m_Handle)
+        {
+            TextLayoutAcquire(text_layout.m_Handle);
+        }
         if (current_text_layout.m_Handle && current_text_layout.m_Handle != text_layout.m_Handle)
         {
-            TextLayoutFree(current_text_layout.m_Handle);
+            TextLayoutRelease(current_text_layout.m_Handle);
```
JCash, line 3770: 👍

engine/profiler/src/test/test_profilerext_render.cpp
```diff
@@ -0,0 +1,305 @@
...
+#include <font/font.h>
+#include <font/fontcollection.h>
+#include <platform/window.hpp>
+
+#include "../../../graphics/src/graphics_private.h"
+#include "../../../graphics/src/test/test_graphics_util.h"
+#include "../../../render/src/render/render.h"
+#include "../../../render/src/render/render_private.h"
+#include "../../../render/src/render/font/fontmap.h"
+#include "../../../render/src/render/font/font_glyphbank.h"
```
JCash: Should use regular includes:
```c++
#include <render/render.h>
#include <render/font/fontmap.h>
#include <render/font/fontmap_glyphbank.h>
```

engine/profiler/src/test/test_profilerext_render.cpp
```diff
@@ -0,0 +1,305 @@
...
+#include <script/script.h>
+#include "../../../font/src/text_layout.h"
+#include <font/font.h>
+#include <font/fontcollection.h>
+#include <platform/window.hpp>
+
+#include "../../../graphics/src/graphics_private.h"
+#include "../../../graphics/src/test/test_graphics_util.h"
+#include "../../../render/src/render/render.h"
+#include "../../../render/src/render/render_private.h"
```
JCash: We avoid ".." as it's cumbersome to maintain, and signifies a bit of a code smell.
Especially as it refers to _other_ libraries, potentially side stepping any build order rules.

Also, "private" is generally just that, private.
What parts are you using from those files? We generally comment such things next to the include line.

engine/profiler/src/test/test_profilerext_render.cpp
```diff
@@ -0,0 +1,305 @@
...
+#include "../../../render/src/render/font/fontmap.h"
+#include "../../../render/src/render/font/font_glyphbank.h"
+#include <render/font_ddf.h>
+
+#include "../profiler_render.h"
```
JCash, line 18: Here, relative paths are ok, as it's a file from the same library.

engine/render/src/test/test_render.cpp
```diff
@@ -22,9 +22,9 @@
 #include <dlib/math.h>
 #include <dlib/dstrings.h>
 #include <script/script.h>
+#include "../../../font/src/text_layout.h"
 #include <font/font.h>
 #include <font/fontcollection.h>
-#include <font/text_layout.h>
```
JCash: I guess telling the llm to not "fix" our includes:

```c++
#include <dmsdk/font/text_layout.h>
```

engine/render/src/test/wscript
```diff
@@ -49,3 +49,11 @@ def build(bld):
                 includes = ['../../src', '../../proto'],
                 target = 'test_render_buffer')
 
+    bld.program(features = 'cxx cprogram test',
+                source = ['../../../profiler/src/test/test_profilerext_render.cpp',
+                          '../../../profiler/src/profiler_render.cpp'],
```
JCash: This is not how we create tests?
This breaks the build order, where the profiler depends on render.

JCash (changes requested):
The reference stuff is good.
But the LLM went nuts with the dependency order and includes :/

engine/render/src/test/test_render.cpp
AGulev: fixed

engine/profiler/src/test/test_profilerext_render.cpp
AGulev: this mess has been fixed

engine/profiler/src/test/test_profilerext_render.cpp
AGulev: fixed

engine/profiler/src/test/test_profilerext_render.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+#include <dlib/hash.h>
+#include <script/script.h>
+#include <font/font.h>
+#include <font/fontcollection.h>
+#include <font/text_layout.h>      // HTextLayout internals and refcount helpers.
+#include <graphics/graphics.h>
+#include <platform/window.hpp>
+#include <render/render.h>
+#include <render/render_private.h> // TextContext access for lifetime assertions.
```
JCash: Two things:
1) we don't use "_private.h" headers in other libraries
2) this _should_ produce a build error, as each library only should get the include folders of A) this library and B) dynamo_home

engine/profiler/src/test/wscript
```diff
@@ -1,8 +1,24 @@
 #! /usr/bin/env python
 
 def build(bld):
+    libs = 'TESTMAIN SOCKET APP GRAPHICS_NULL DDF SCRIPT EXTENSION RESOURCE DLIB PROFILE_NULL PLATFORM_NULL LUA FONT RENDER profilerext'
+    exported_symbols = ['GraphicsAdapterNull']
+
+    if 'web' in bld.env['PLATFORM']:
+        lib_dirs = {}
+        lib_dirs['library_render.js'] = '../../render/src/js'
```
JCash: Using files directly from other libraries isn't good from a "buld system" perspective, nor us understanding the code flow.

engine/profiler/src/test/wscript
```diff
@@ -1,8 +1,24 @@
...
                 use = 'TESTMAIN DLIB profilerext_null',
                 includes = ['../../../src'],
                 target = 'test_profilerext_null')
+
+    bld.program(features = 'cxx cprogram test',
+                source = 'test_profilerext_render.cpp',
+                use = libs,
+                exported_symbols = exported_symbols,
+                web_libs = ['library_sys.js', 'library_script.js', 'library_render.js'],
+                includes = ['../../../src', '../../../render/src', '../../../graphics/src'],
```
JCash: Here it's adding includes from a different library which also isn't how we do it.

JCash (changes requested):
still a little cleanup todo.

JCash (approved):

PR #12231, opened 2026-04-16
Editor camera overlay in scene view

Jhonnyg:
The scene preview window now has an overlay in the bottom of the corner that shows the scene from the currently selected camera component.

Screen Recording 2026-04-16 at 14 27 14 (1)


editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,149 @@
...
 (defonce ^:private cached-buf-img-ref (atom nil))
+(defonce ^:private cached-camera-inset-buf-img-ref (atom nil))
 
 ;; Replacement for Screenshot/readToBufferedImage but without expensive y-axis flip.
 ;; We flip in JavaFX instead
-(defn- read-to-buffered-image [^long w ^long h]
-  (let [^BufferedImage image (let [^BufferedImage image @cached-buf-img-ref]
+(defn- read-to-buffered-image
+  ([^long w ^long h]
+   (read-to-buffered-image cached-buf-img-ref w h))
```
matgis, line 199: No need to introduce a new arity here. Instead, just supply the desired `cache-ref` at each call site.

Although personally I would probably change this to a function that takes an existing `BufferedImage` or `nil` and use it with `swap!` on the desired `cache-ref` at the call site instead.

editor/src/clj/editor/defold_project.clj
```diff
@@ -1628,6 +1628,11 @@
                                  (double (or (get settings ["display" "width"]) 0))))
   (output display-height g/Num (g/fnk [settings]
                                   (double (or (get settings ["display" "height"]) 0))))
+  (output render-clear-color g/Any (g/fnk [settings]
+                                     [(double (or (get settings ["render" "clear_color_red"]) 0.0))
+                                      (double (or (get settings ["render" "clear_color_green"]) 0.0))
+                                      (double (or (get settings ["render" "clear_color_blue"]) 0.0))
+                                      (double (or (get settings ["render" "clear_color_alpha"]) 1.0))]))
```
matgis, line 1635: Maybe simpler to use a primitive vector, and the `not-found` argument to get? Unless we expect there to be `nil` values present? 🤔
```suggestion
  (output render-clear-color g/Any (g/fnk [settings]
                                     (vector-of :double
                                       (get settings ["render" "clear_color_red"] 0.0)
                                       (get settings ["render" "clear_color_green"] 0.0)
                                       (get settings ["render" "clear_color_blue"] 0.0)
                                       (get settings ["render" "clear_color_alpha"] 1.0))))
```

editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,149 @@
...
+
+(defn- camera-renderable? [renderable]
+  (contains? (:tags renderable) :camera))
+
+(defn- camera-inset-clear-color [selected-camera-renderable]
+  (let [[clear-r clear-g clear-b clear-a] (:render-clear-color (:user-data selected-camera-renderable))]
+    [(double (or clear-r 0.0))
+     (double (or clear-g 0.0))
+     (double (or clear-b 0.0))
+     (double (or clear-a 1.0))]))
```
matgis, line 221: There's no real need for this function to exist. We should just assume that the clear color is a well-formed RGBA vector of doubles and default to opaque black if not present at the call site.

editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,149 @@
...
+
+(defn- valid-camera-inset-orthographic-zoom? [selected-camera-renderable]
+  (let [{:keys [is-orthographic orthographic-mode orthographic-zoom]} (:user-data selected-camera-renderable)]
+    (or (not is-orthographic)
+        (not= orthographic-mode :ortho-mode-fixed)
+        (> (double (or orthographic-zoom 0.0)) 0.0))))
```
matgis, line 227:
```suggestion
        (pos? (double (or orthographic-zoom 0.0))))))
```

editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,149 @@
...
+      (when (and (pos? project-width) (pos? project-height))
+        (/ project-width project-height))
+      (let [camera-aspect-ratio (double (or aspect-ratio 0.0))]
+        (when (pos? camera-aspect-ratio)
+          camera-aspect-ratio)))))
+
+(defn- make-camera-inset-camera
+  ([selected-camera-renderable]
+   (make-camera-inset-camera selected-camera-renderable camera-inset-width camera-inset-height))
+  ([selected-camera-renderable _inset-width _inset-height]
```
matgis, line 243: There appears to be no need for multiple arities here. We don't even seem to use the arguments? 🤔

editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,149 @@
...
+   (when (valid-camera-inset-orthographic-zoom? selected-camera-renderable)
```
matgis, line 244: We've already called this, or we can't even get here. Move this check to a `{:pre [(valid-camera-inset-orthographic-zoom? selected-camera-renderable)]}` assert instead?

editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,149 @@
...
+             camera-type (if is-orthographic :orthographic :perspective)
+             camera-fov-x (if is-orthographic ow fov-x)
+             camera-fov-y (if is-orthographic oh fov-y)
+             camera (c/make-camera camera-type identity {:fov-x camera-fov-x
+                                                         :fov-y camera-fov-y})]
+         (assoc camera
+           :position (Point3d. (.x world-translation) (.y world-translation) (.z world-translation))
+           :rotation (Quat4d. world-rotation)
+           :z-near (double (or near-z 1.0))
+           :z-far (double (or far-z 1000.0))))))))
```
matgis, line 267: Casting to `double` here will just unbox and then immediately box the value again, since `assoc` takes `Object` arguments.
```suggestion
           :z-near (or near-z 1.0)
           :z-far (or far-z 1000.0)))))))
```

editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,149 @@
...
+(defn- camera-inset-aspect-ratio [selected-camera-renderable]
+  (let [{:keys [is-orthographic auto-aspect-ratio aspect-ratio display-width display-height]} (:user-data selected-camera-renderable)
+        project-width (double (or display-width 0.0))
+        project-height (double (or display-height 0.0))]
+    (if (or is-orthographic auto-aspect-ratio)
+      (when (and (pos? project-width) (pos? project-height))
+        (/ project-width project-height))
+      (let [camera-aspect-ratio (double (or aspect-ratio 0.0))]
+        (when (pos? camera-aspect-ratio)
+          camera-aspect-ratio)))))
```
matgis, line 238: The whole process of determining if we can actually render the camera inset feels quite convoluted. There is a check in `valid-camera-inset-orthographic-zoom?`, but that is then followed by various functions like this possibly returning `nil` on top of that initial check.

I can't help but feel this will be hard to maintain, and could lead to subtle bugs. I suggest we have a single function that returns a well-formed map of all the data we need to proceed, or `nil`. Unless that function returns `nil`, we should simply assume the data is well-formed.

editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,149 @@
...
+     (when-let [aspect-ratio (camera-inset-aspect-ratio selected-camera-renderable)]
+       (let [{:keys [is-orthographic near-z far-z fov display-height orthographic-zoom orthographic-mode]} (:user-data selected-camera-renderable)
+             world-translation ^Vector3d (:world-translation selected-camera-renderable)
+             world-rotation ^Quat4d (:world-rotation selected-camera-renderable)
+             fov-y (double (or fov c/fov-y-35mm-full-frame))
+             fov-x (* fov-y aspect-ratio)
+             zoom (double (or orthographic-zoom 1.0))
+             zoom (if (= orthographic-mode :ortho-mode-fixed)
+                    zoom
+                    1.0)
```
matgis, line 254:
```suggestion
             zoom (double (if (and orthographic-zoom
                                   (= orthographic-mode :ortho-mode-fixed))
                            orthographic-zoom
                            1.0))
```

editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,149 @@
...
+    (doseq [[x y z] border-line-points]
+      (vtx/buf-push-floats! border-buf [x y z border-r border-g border-b border-a]))
+    (vtx/flip! border-vbuf)
+    (.glDisable gl GL/GL_DEPTH_TEST)
+    (.glLineWidth gl 2.0)
+    (gl/with-gl-bindings gl render-args [border-shader border-binding]
+      (gl/gl-draw-arrays gl GL2/GL_LINES 0 (count border-vbuf)))
+    (.glLineWidth gl 1.0)))
+
+(defn- produce-selected-camera-renderable [scene-render-data selection]
```
matgis, line 323: The `produce-` prefix should exclusively be used by `g/defnk` functions. Rename to `find-selected-camera-renderable`?

editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,149 @@
...
+        selected-cameras
+        (into []
+              (comp cat
+                    (filter camera-renderable?)
+                    (filter (fn [renderable]
+                              (or (contains? selection-set (:node-id renderable))
+                                  (some selection-set (:node-id-path renderable))))))
+              (vals (:renderables scene-render-data)))]
+    (when (= 1 (count selected-cameras))
+      (first selected-cameras))))
```
matgis, line 334: It feels a bit expensive to iterate over all the renderables just to find the selected camera. Wouldn't it work to just look at the ones in `(:selected-renderables scene-render-data)`? 🤔

editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,149 @@
...
+         :display-height display-height
+         :render-width (* display-width camera-inset-render-scale)
+         :render-height (* display-height camera-inset-render-scale)}))))
+
+(defn- remove-camera-outline-renderables [all-renderables]
+  (into {}
+        (map (fn [[render-pass renderables]]
+               [render-pass
+                (into [] (remove #(or (contains? (:tags %) :camera)
+                                      (contains? (:tags %) :outline)))
```
matgis, line 284: Filtering out all the `:outline` objects seems OK, but instead of targeting the `:camera` specifically, I think we should introduce a `:gizmo` tag that we apply to Collision Objects, Cameras and Lights so all of those are hidden from the camera inset view.

editor/src/clj/editor/scene.clj
```diff
@@ -288,6 +421,24 @@
...
+                                     index (:index renderable)
+                                     topmost? (:topmost? renderable)
+                                     selection-state (:selected renderable)]
+                                 (cond-> (assoc renderable :render-key (render-key view-matrix world-translation index topmost?))
+                                   (contains? (:pass-overrides renderable) pass/outline)
+                                   (assoc-in [:pass-overrides pass/outline :render-key]
+                                             (outline-render-key view-matrix world-translation index topmost? selection-state))))))
+                       render-sort)]))
+          renderables-by-pass)))
+
```
matgis, line 441: Hmm. I think we probably shouldn't render all the passes for the camera inset. Just `pass/opaque` and `pass/transparent`, and maybe `pass/background` should cover all the objects in the scene. That gets rid of all the `:outline`-tagged renderables in one swoop, and also lets us skip this `:pass-overrides` part here. We can also probably get by on just updating the sort order of `pass/transparent` and leave the opaque ordering as is.

editor/src/clj/editor/scene.clj
```diff
@@ -340,23 +491,65 @@
...
-                  pass-renderables (-> (get renderables pass)
-                                       (assoc-updatable-states updatable-states))]]
-      (setup-pass gl pass pass-render-args)
-      (if (= render-mode :aabbs)
-        (batch-render gl pass-render-args (make-aabb-renderables pass-renderables) batch-key)
-        (batch-render gl pass-render-args pass-renderables batch-key)))))
+(defn render!
+  ([^GLContext context render-mode renderables updatable-states viewport pass->render-args]
+   (render! context render-mode renderables updatable-states viewport pass->render-args [0.0 0.0 0.0 1.0]))
+  ([^GLContext context render-mode renderables updatable-states viewport pass->render-args [clear-r clear-g clear-b clear-a]]
```
matgis, line 497: We prefer to not introduce a new arity unless absolutely necessary. This function is only called from within this file (and should probably be made private, TBH). It's much better to simply update the call sites and supply explicit arguments.

editor/src/clj/editor/scene.clj
```diff
@@ -340,23 +491,65 @@
...
+                                      (picking-render-args viewport @last-picking-rect))
+                   pass-renderables (-> (get renderables pass)
+                                        (assoc-updatable-states updatable-states))]]
+       (setup-pass gl pass pass-render-args)
+       (if (= render-mode :aabbs)
+         (batch-render gl pass-render-args (make-aabb-renderables pass-renderables) batch-key)
+         (batch-render gl pass-render-args pass-renderables batch-key))))))
+
+;; NOTE: camera-inset-frame-version is not used but we need it as a dependency
+;;       so that preview animations are stepped in the preview window (e.g particle effects).
```
matgis, line 516: Comments like these belong inside the function, not outside of it.

editor/src/clj/editor/scene.clj
```diff
@@ -340,23 +491,65 @@
...
+(g/defnk produce-camera-inset-data [scene-render-data selection updatable-states camera-inset-frame-version ^GLAutoDrawable camera-inset-drawable]
```
matgis, line 517: I don't really understand why we need to store `camera-inset-frame-version` in a property and depend on it here. Surely we have the same exact problem for the main view? Can't we solve this the same way? I dislike having a graph property that needs to update every frame.

editor/src/clj/editor/scene.clj
```diff
@@ -340,23 +491,65 @@
...
+  (let [hidden-result {:visible false
+                       :image nil
+                       :width camera-inset-width
+                       :height camera-inset-height}]
```
matgis, line 521: Do we really need the dimensions if it ends up hidden? Could we simplify and return `nil` instead?

editor/src/clj/editor/scene.clj
```diff
@@ -863,6 +1056,25 @@
...
+               :image camera-inset-image
+               :fit-width camera-inset-width
+               :fit-height camera-inset-height
+               :preserve-ratio false
+               :smooth true
+               :scale-y -1.0}]
+   :mouse-transparent true
+   :anchor-pane/bottom camera-inset-margin
+   :anchor-pane/right camera-inset-margin})
+
```
matgis, line 1077: Rather than rendering the scene to an offscreen buffer, then transferring the pixel data out of VRAM so that we can pass it along to an `ImageView`, Wouldn't it be more efficient to render a quad with the offscreen buffer texture directly into the viewport?

editor/src/clj/editor/scene.clj
```diff
@@ -340,23 +491,65 @@
...
+                                                    pass/all-passes)
+                scene-renderables (-> (:renderables scene-render-data)
+                                      remove-camera-outline-renderables
+                                      (resort-renderables-for-camera camera-inset-camera))
+                camera-inset-frame (when camera-inset-drawable
+                                     (gl/with-drawable-as-current camera-inset-drawable
+                                       (.setSurfaceSize ^GLOffscreenAutoDrawable camera-inset-drawable (int render-width) (int render-height))
+                                       (scene-cache/process-pending-deletions! gl)
+                                       (render! gl-context :normal scene-renderables updatable-states camera-inset-viewport camera-inset-pass->render-args clear-color)
+                                       (render-camera-inset-border! gl camera-inset-viewport)
```
matgis, line 538: We need to call `(scene-cache/prune-context! gl)` after we're done rendering to ensure the `scene-cache` context associated with our drawable does not fill up with GL objects that are no longer used.

matgis (changes requested):
It's a really nice feature! ❤️

I found a few issues we probably want to address before merging, though!

editor/src/clj/editor/scene.clj
matgis: I realize we may not be able to render the quad directly to the viewport since the render target is in a different GL context than the main view. Another approach might be to render the camera inset directly into the main view as a sub-region. but we may need to either clear the depth buffer of at least the inset region before rendering the inset, or render the inset first and max out the depth values of the inset region using a separate draw call before rendering the main viewport.

The benefits would be that we do not need to upload meshes and textures to the separate OpenGL context, although this might be better solved by introducing a shared parent context for GPU resources, something we've discussed in the past.

Anyway, I think we can probably merge this feature without doing any of this and revisit it in the future.

editor/src/clj/editor/scene.clj
Jhonnyg: The camera renderable is only in the outline pass and not in a selection pass, so that's why we do this extra dance. I've added a selection pass now to the camera so that it also can be selected in the viewport.

Jhonnyg: Hm, adding a selectable pass to the camera only solved it halfways it seems. It only works to look at the selection set if the camera component itself it selected, but not if you select the camera parent object or anything above in the hierarchy. Can we even solve it in some nicer way with the current API?

editor/src/clj/editor/scene.clj
Jhonnyg: I think we would need to store the entire GL state for that to work, I'm not sure we can trust that everything that's rendering to the view isn't leaking state (e.g extensions with viewport rendering).

editor/src/clj/editor/scene.clj
```diff
@@ -92,6 +96,11 @@
    :rotation default-rotation
    :scale default-scale})
 
+(def ^:private camera-inset-margin 18.0)
+(def ^:private camera-inset-width 240.0)
+(def ^:private camera-inset-height 135.0)
+(def ^:private camera-inset-render-scale 2.0)
```
matgis: Declaring these as compile-time constants solves some boxed math warnings.
```suggestion
(def ^:private ^:const camera-inset-margin 18.0)
(def ^:private ^:const camera-inset-width 240.0)
(def ^:private ^:const camera-inset-height 135.0)
(def ^:private ^:const camera-inset-render-scale 2.0)
```

editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,136 @@
...
+        project-width (double (or display-width 0.0))
+        project-height (double (or display-height 0.0))]
+    (if (or is-orthographic auto-aspect-ratio)
+      (when (and (pos? project-width) (pos? project-height))
+        (/ project-width project-height))
+      (let [camera-aspect-ratio (double (or aspect-ratio 0.0))]
+        (when (pos? camera-aspect-ratio)
+          camera-aspect-ratio)))))
+
+(defn- make-camera-inset-camera [selected-camera-renderable aspect-ratio]
```
matgis: Fix boxed math:
```suggestion
(defn- make-camera-inset-camera [selected-camera-renderable ^double aspect-ratio]
```

editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,136 @@
...
+      :z-near (or near-z 1.0)
+      :z-far (or far-z 1000.0))))
 
 (defn vp-dims [^Region viewport]
   (types/dimensions viewport))
 
+(defn- render-camera-inset-border! [^GL2 gl ^Region viewport]
+  (let [border-shader shaders/basic-color-local-space
+        vertex-description (shaders/vertex-description border-shader)
+        [viewport-width viewport-height] (vp-dims viewport)
```
matgis: Fix boxed math:
```suggestion
        [^double viewport-width ^double viewport-height] (vp-dims viewport)
```

editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,136 @@
...
+                                       (filter (comp types/selection? key)
+                                               (:renderables scene-render-data))))]
+    (when (= 1 (count selected-cameras))
+      (first selected-cameras))))
+
+(defn- make-camera-inset-render-data [scene-render-data]
+  (when-let [selected-camera-renderable (find-selected-camera-renderable scene-render-data)]
+    (when (valid-camera-inset-orthographic-zoom? selected-camera-renderable)
+      (when-let [aspect-ratio (camera-inset-aspect-ratio selected-camera-renderable)]
+        (let [display-height camera-inset-height
```
matgis: Fix boxed math:
```suggestion
        (let [aspect-ratio (double aspect-ratio)
              display-height camera-inset-height
```

editor/src/clj/editor/scene.clj
```diff
@@ -358,6 +499,32 @@
         (batch-render gl pass-render-args (make-aabb-renderables pass-renderables) batch-key)
         (batch-render gl pass-render-args pass-renderables batch-key)))))
 
+(g/defnk produce-camera-inset-data [scene-render-data updatable-states ^GLAutoDrawable camera-inset-drawable]
+  (when-some [{:keys [camera clear-color display-width display-height render-width render-height]} (make-camera-inset-render-data scene-render-data)]
+    (when camera-inset-drawable
+      (let [camera-inset-viewport (types/->Region 0 render-width 0 render-height)
+            camera-inset-pass->render-args (into {}
+                                                (map (juxt identity (partial pass-render-args camera-inset-viewport camera)))
+                                                camera-inset-passes)
```
matgis: Style: Fix formatting.
```suggestion
            camera-inset-pass->render-args (into {}
                                                 (map (juxt identity (partial pass-render-args camera-inset-viewport camera)))
                                                 camera-inset-passes)
```

editor/src/clj/editor/scene.clj
```diff
@@ -358,6 +499,32 @@
...
+                                 (.glUseProgram ^GL2 gl 0)
+                                 (let [[w h] (vp-dims camera-inset-viewport)
+                                       buffered-image (read-to-buffered-image cached-camera-inset-buf-img-ref w h)]
+                                   (scene-cache/prune-context! gl)
+                                   buffered-image))]
+        (when-some [camera-inset-image (some-> camera-inset-frame
+                                               (SwingFXUtils/toFXImage nil))]
+          {:image camera-inset-image
+           :width display-width
+           :height display-height})))))
```
matgis: Doesn't look like `camera-inset-frame` can be `nil`? Let's remove the checks.
```suggestion
                                   buffered-image))
            camera-inset-image (SwingFXUtils/toFXImage camera-inset-frame nil)]
        {:image camera-inset-image
         :width display-width
         :height display-height}))))
```

editor/src/clj/editor/scene.clj
```diff
@@ -1703,13 +1906,24 @@
                              (let [async-copy-state-atom (g/node-value view-id :async-copy-state)]
                                (reset! async-copy-state-atom (scene-async/request-resize! @async-copy-state-atom width height))))
                            (let [drawable (gl/offscreen-drawable width height)
-                                 picking-drawable (gl/offscreen-drawable picking-drawable-size picking-drawable-size)]
+                                picking-drawable (gl/offscreen-drawable picking-drawable-size picking-drawable-size)
+                                camera-inset-drawable (when (supports-camera-inset-drawable? view-id)
+                                                        (gl/offscreen-drawable camera-inset-width camera-inset-height))]
```
matgis: Style: Fix formatting.
```suggestion
                                 picking-drawable (gl/offscreen-drawable picking-drawable-size picking-drawable-size)
                                 camera-inset-drawable (when (supports-camera-inset-drawable? view-id)
                                                         (gl/offscreen-drawable camera-inset-width camera-inset-height))]
```

editor/src/clj/editor/camera_editor.clj
```diff
@@ -285,14 +288,16 @@
...
   (let [vertex-buffer (gen-outline-vertex-buffer render-args renderables renderable-count)
         outline-vertex-binding (vtx/use-with ::frustum-outline vertex-buffer outline-shader)]
     (gl/with-gl-bindings gl render-args [outline-shader outline-vertex-binding]
       (gl/gl-draw-arrays gl GL/GL_LINES 0 (* renderable-count camera-preview-mesh-vertices-count)))))
 
 (g/defnk produce-camera-scene
-  [_node-id fov aspect-ratio near-z far-z orthographic-projection orthographic-zoom orthographic-mode project-display-width project-display-height]
+  [_node-id fov aspect-ratio auto-aspect-ratio near-z far-z orthographic-projection orthographic-zoom orthographic-mode
+   project-display-width project-display-height
+   project-render-clear-color]
```
matgis: Style: Use a single line.
```suggestion
[_node-id fov aspect-ratio auto-aspect-ratio near-z far-z orthographic-projection orthographic-zoom orthographic-mode project-display-width project-display-height project-render-clear-color]
```

editor/src/clj/editor/scene.clj
```diff
@@ -181,25 +190,136 @@
...
+      (gl/gl-draw-arrays gl GL2/GL_LINES 0 (count border-vbuf)))
+    (.glLineWidth gl 1.0)))
+
+(defn- find-selected-camera-renderable [scene-render-data]
+  (let [selected-cameras (into []
+                               (comp (filter camera-renderable?)
+                                     (filter :selected))
+                               (mapcat val
+                                       (filter (comp types/selection? key)
+                                               (:renderables scene-render-data))))]
```
matgis: We can perform the initial `RenderPass` `filter` and `mapcat` as part of the transducer here:
```suggestion
                               (comp (filter (comp types/selection? key))
                                     (mapcat val)
                                     (filter camera-renderable?)
                                     (filter :selected)
                                     (take 2))
                               (:renderables scene-render-data))]
```
We can also stop once we've found two cameras using `(take 2)`, since we'll only show the inset if a single camera is selected.

editor/src/clj/editor/scene.clj
matgis: Yeah, it appears you're right. We could maybe do more work to find the inset camera as we're flattening the scene to do it in one pass, but it's a larger change. Let's keep it this way for now.

matgis (dismissed):
I'm happy with this for now. We can always try to improve the efficiency later.

I added some minor suggestions, but feel free to merge!

PR #12235, opened 2026-04-17
Camera axis alignment cube in editor scene view

Jhonnyg:


Jhonnyg:
We'll do another take on it at some point, but not now.

PR #12237, opened 2026-04-17
Ensure the Editor loads only bundled libraries, ignoring those installed on the system

### LINKED ISSUE #11695: Apple MacBook Pro M1 Max - Defold will not start - opening splash screen only.
I have an Apple MacBook Pro M1 Max.  I've installed the temurin v25 javac as instructed.  I have tried the Defold installer (11.2) from the website, I've tried to build Defold, and I've tried to install Defold using Homebrew.  The Homebrew installation worked once and I thought all was well, but it only worked once.  I've also tried installing the 12.x beta and alpha installers with no success.  I've also tried the microsoft Javac version 25.x... no change.

Anyone else had issues with the Apple M1 platform?  I'm stumped.

As I indicated, Defold DID work once after the Homebrew installation and I was able to run the Shadertoy demo.  However, after closing Defold I have been unable to run the program.

Thanks for any tips/help

AGulev:
Fix an issue where the editor could load libraries from `/Library/Java/Extensions` instead of the bundled ones, which could break loading or otherwise affect the editor.

Fix https://github.com/defold/defold/issues/11695

### Technical changes

The editor now loads native libraries directly from Defold’s unpacked lib folder, not from system library paths.

It finds the libs automatically, loads them by exact path in parallel, fails fast if one is broken, and uses the same exact-path lookup for JNA too. Also, tests added for discovery, loading, and failure cases.

### DIFF
--- a/editor/src/java/com/defold/libs/ResourceUnpacker.java
+++ b/editor/src/java/com/defold/libs/ResourceUnpacker.java
@@ -160,19 +160,11 @@ public class ResourceUnpacker {
                         Files.writeString(unpackShaPath, sha1);
                     }
                 }
-                if (unpackPath.getParent().startsWith(Editor.getSupportPath())) {
-                    // Prevent from deletion by deleteOldUnpackDirs
-                    Executors.newSingleThreadScheduledExecutor().scheduleAtFixedRate(() -> {
-                        try {
-                            Files.setLastModifiedTime(unpackPath, FileTime.from(Instant.now()));
-                        } catch (IOException e) {
-                            logger.warn("Couldn't set lastModifiedTime for {}", unpackPath, e);
-                        }
-                    }, 0, 1, TimeUnit.DAYS);
-                }
 
                 Path unpackedLibDir = unpackPath.resolve(platform.getPair() + "/lib").toAbsolutePath();
                 System.setProperty("jna.nosys", "true");
+                // Exact-path preloading is authoritative, but JOGL still resolves some bundled
+                // natives by logical name, and JNA-by-name callers still expect these paths.
                 System.setProperty("java.library.path", unpackedLibDir.toString());
                 System.setProperty("jna.library.path", unpackedLibDir.toString());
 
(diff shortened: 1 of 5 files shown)


editor/src/java/com/defold/libs/ResourceUnpacker.java
```diff
@@ -148,12 +174,189 @@ public static void unpackResources() throws IOException, URISyntaxException {
...
+            for (NativeLibraryLoadFailure failure : failures) {
+                aggregatedException.addSuppressed(failure.cause);
+            }
+            throw aggregatedException;
+        }
+
+        return immutableLinkedHashMap(discoveredLibraries);
+    }
+
+    private static NativeLibraryLoadFailure preloadNativeLibrary(String logicalName, Path libraryPath, NativeLibraryLoader loader) {
```
britzl: Looks a bit odd to return an error instead of throwing one?

editor/src/java/com/defold/libs/ResourceUnpacker.java
```diff
@@ -148,12 +174,189 @@ public static void unpackResources() throws IOException, URISyntaxException {
...
+            return platform.getLibSuffix();
+        }
+        throw new IllegalArgumentException("Unsupported bundled native library filename: " + fileName);
+    }
+
+    private static Map<String, Path> immutableLinkedHashMap(Map<String, Path> libraries) {
+        return Collections.unmodifiableMap(new LinkedHashMap<>(libraries));
+    }
+
+    private static void startUnpackDirTouchScheduler(Path unpackPath) {
```
britzl, line 347: 👍

matgis (dismissed):
Looks good to me! ❤️

I've only tested on macOS, though!

editor/src/java/com/defold/libs/ResourceUnpacker.java
AGulev: I rewrote it using exception throwing to make it more conventional but still accumulate errors to make sure all the errors reproted properly

AGulev:
@britzl I made a fixed according to your review

PR #12238, opened 2026-04-17
Improve the Lua parser to make it less error-prone

### LINKED ISSUE #12234: Deleting 'y' value in go.property(vector3(x,y,z)) is causing an error
### Expected behaviour
If I press backspace on the Y, it should just remove similar to the X and Z property in go.property(vmath.vector3(X,Y,Z))

### Actual behaviour
Throwing an error if y is a single digit and I press backspace on it

### Steps to reproduce
1) Open a script file
2)  on top, write - go.property(vmath.vector3(10,8,0))
3) place cursor after 8
4) press backspace

Image

<hr/>

<table>
<tr><td>Build time</td><td>2026-04-09T14:36:31.051881</td></tr>
<tr><td>Defold channel</td><td>stable</td></tr>
<tr><td>Defold editor sha</td><td>0ad9c86fa0a9f7ac19bc468b1a67ee06bb2578b5</td></tr>
<tr><td>Defold engine sha</td><td>0ad9c86fa0a9f7ac19bc468b1a67ee06bb2578b5</td></tr>
<tr><td>Defold version</td><td>1.12.3</td></tr>
<tr><td>Error</td><td><a href='https://sentry.io/organizations/defold/issues/?query=id%3A"6339626ca88345dcbd0374e802369e23"'>6339626ca88345dcbd0374e802369e23</a></td></tr>
<tr><td>GPU</td><td>NVIDIA GeForce GTX 1650/PCIe/SSE2</td></tr>
<tr><td>GPU Driver</td><td>4.6.0 NVIDIA 560.94</td></tr>
<tr><td>Java version</td><td>25+36-LTS</td></tr>
<table>

AGulev:
Fix an exception that can occur in the editor while typing `go.property()`

Fix https://github.com/defold/defold/issues/12234

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
@@ -193,8 +193,10 @@ public class LuaScannerTest {
     public void testPropsFail() throws Exception {
         List<Property> properties = getPropertiesFromFile("test_props_fail.lua");
 
-        assertEquals(1, properties.size());
-        assertEquals(properties.getFirst().status(), Status.INVALID_VALUE);
+        assertTrue(properties.size() >= 1);
+        for (Property property : properties) {
+            assertEquals(Status.INVALID_VALUE, property.status());
+        }
     }
 
     @Test
@@ -619,5 +621,13 @@ public class LuaScannerTest {
         errors = LuaScanner.parse("go.property('local', resource.atlas(atlas_path))").errors();
         assertEquals(1, errors.size());
         assertEquals("expected a string literal resource argument", errors.get(0).message());
+
+        errors = LuaScanner.parse("go.property('hex', vmath.vector3(0x1, 2, 3))").errors();
+        assertEquals(1, errors.size());
+        assertEquals("wrong number format: '0x1'", errors.get(0).message());
+
+        errors = LuaScanner.parse("go.property('vec', vmath.vector3(1, 2, 3, 4))").errors();
+        assertEquals(1, errors.size());
+        assertEquals("wrong number of numeric arguments", errors.get(0).message());
     }
 }
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
@@ -187,13 +187,22 @@ public class LuaScannerTest {
         assertProperty(properties, "prop2", Double.valueOf(1.0), 1);
         assertProperty(properties, "prop3", Double.valueOf(.1), 2);
         assertProperty(properties, "prop4", Double.valueOf(-1.0E2), 3);
+
+        properties = getPropertiesFromString(
+                "go.property(\"hex\", 0x1)\n" +
+                "go.property(\"hex_float\", 0x1.8)\n" +
+                "go.property(\"negative_hex\", -0x1)");
+        assertEquals(3, properties.size());
+        assertProperty(properties, "hex", Double.valueOf(1.0), 0);
+        assertProperty(properties, "hex_float", Double.valueOf(1.5), 1);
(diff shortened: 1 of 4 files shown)


com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
```diff
@@ -193,8 +193,10 @@ public void testPropsNumber() throws Exception {
     public void testPropsFail() throws Exception {
         List<Property> properties = getPropertiesFromFile("test_props_fail.lua");
 
-        assertEquals(1, properties.size());
-        assertEquals(properties.getFirst().status(), Status.INVALID_VALUE);
+        assertTrue(properties.size() >= 1);
```
vlaaad: Should it be `assertEquals(2, properties.size());` instead?

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
```diff
@@ -193,8 +193,10 @@ public void testPropsNumber() throws Exception {
...
+        for (Property property : properties) {
+            assertEquals(Status.INVALID_VALUE, property.status());
```
vlaaad, line 207: Should we assert specific error messages?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/LuaScanner.java
```diff
@@ -556,11 +556,19 @@ private static boolean getNumArgs(LuaParser.ArgsContext argsCtx, double[] result
                 LuaParser.ExpContext exp = val.exp(0);
                 int firstTokenType = val.getStart().getType();
                 if (num != null || (exp != null && exp.number() != null && firstTokenType == LuaParser.MINUS)) {
-                    resultArgs[count] = Double.parseDouble(val.getText());
```
vlaaad, line 559: It seems we previously incorrectly rejected valid hex lua numbers (and now we still do it by using `Double.parseDouble`). Lua hex numbers can be parsed if we add a `p0` suffix, e.g.:
```java
switch (tokenType) {
    case LuaParser.HEX -> Double.parseDouble(text + "p0");
    case LuaParser.INT, LuaParser.FLOAT, LuaParser.HEX_FLOAT -> Double.parseDouble(text);
    default -> throw new NumberFormatException(text);
};
```

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/LuaScannerTest.java
```diff
@@ -619,5 +621,13 @@ public void testErrorMessages() {
         errors = LuaScanner.parse("go.property('local', resource.atlas(atlas_path))").errors();
         assertEquals(1, errors.size());
         assertEquals("expected a string literal resource argument", errors.get(0).message());
+
+        errors = LuaScanner.parse("go.property('hex', vmath.vector3(0x1, 2, 3))").errors();
```
vlaaad: That's a valid Lua number though?

vlaaad (commented):
Maybe it's fine, maybe it's not. I think the issue is that we are encountering a syntax error in the `go.property` declaration. We already have a code that skips parsing if we have seen a syntax error (see `encounteredError` use), but the current syntax checking is not enough if the error is inside the function call declaration. It feels like the proper fix would be to avoid `Double.parseDouble` invocation completely when we encounter the real syntax issue.

As I understand, what's happening there is that, given a string `vmath.vector3(1,,2)`, after parsing `1,`, the parser sees an unexpected `,`, skips over it, and then parses `2` as a next expression, but the error token (unexpected comma) is part of the next token's visible text. So, in this PR, `Double.parseDouble` happens to work out "as expected" by masking a syntax error with a number parse error. I'm not sure this is the right approach, though. We probably want to be more careful with parsing args, e.g., by checking that they contain error nodes. What do you think?

AGulev:
I think you are right, and it's much better to check for errors before parsing. I didn't put too much thought into it and just did it after the previous fix.
But I kept the exception-throwing mechanism here anyway

vlaaad (approved):
LGTM! I didn't try it locally.

PR #12240, opened 2026-04-18
Use Vulkan as the default graphics API on Android

### LINKED ISSUE #7649: Make Vulkan the default on platforms where Vulkan is supported
Make Vulkan (and MoltenVK) the default on platforms where Vulkan is supported and use OpenGL as a fallback.

Jhonnyg:
Android will from now on use Vulkan as the primary graphics adapter and OpenGL as a fallback if Vulkan is not supported on the running device. This can be changed by using an app manifest where you can select either OpenGL+Vulkan (default), only OpenGL or only Vulkan as the primary graphics adapter.

Fixes #7649


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderCompilers.java
```diff
@@ -47,10 +47,8 @@ private Set<ShaderDesc.Language> getPlatformShaderLanguages(boolean isComputeTyp
             if (platform == Platform.Arm64MacOS ||
                 platform == Platform.X86_64MacOS) {
 
-                outputSpirv = true;
-
-                // Vulkan is default on OSX since 1.9.9, meaning OpenGL is optional.
-                if (!isComputeType && outputGLSL) {
+                boolean hasExplicitOutput = outputSpirv || outputGLSL || outputMsl;
+                if (!isComputeType && (outputGLSL || !hasExplicitOutput)) {
```
Jhonnyg, line 51: I think this was wrong before, we don't need to build spir-v on osx if someone has only selected opengl in the manifest file.

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/ShaderProgramBuilderTest.java
```diff
@@ -363,30 +363,21 @@ public void testShaderReflection() throws Exception {
...
 
         expectedLanguages = new ShaderDesc.Language[] { firstLanguage, secondLanguage };
 
-        if (spirvIsDefault) {
-            getProject().getProjectProperties().putBooleanValue("shader", "output_glsl", true);
-        } else {
-            getProject().getProjectProperties().putBooleanValue("shader", "output_spirv", true);
-        }
+        getProject().getProjectProperties().putBooleanValue("shader", "output_glsl", true);
+        getProject().getProjectProperties().putBooleanValue("shader", "output_spirv", true);
```
Jhonnyg, line 376: Host platforms will always support both glsl and spirv, might as well test both.

share/extender/build_input.yml
```diff
@@ -439,13 +439,13 @@ platforms:
...
         platformIncludes:   ["{{env.NDK_PATH}}/sources/android/native_app_glue", "{{env.NDK_PATH}}/sources/android/cpufeatures"]
         flags:      ["-fno-exceptions", "-fvisibility=hidden"]
         linkFlags:  ["-Wl,--build-id=uuid"]
         libs:       []
-        dynamicLibs: ["EGL","GLESv1_CM","GLESv2","OpenSLES","m","log","c"]
-        symbols:    ["AudioDecoderTremolo", "GraphicsAdapterOpenGL", "ProfilerRemotery"]
+        dynamicLibs: ["OpenSLES","m","log","c","EGL","GLESv1_CM","GLESv2"]
+        symbols:    ["AudioDecoderTremolo", "GraphicsAdapterOpenGLES", "GraphicsAdapterVulkan", "ProfilerRemotery"]
```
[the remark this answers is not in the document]
Jhonnyg: Hm yes valid point I guess, but we should not fix it in this PR 🤔

engine/graphics/src/vulkan/android/graphics_vulkan_android.cpp
```diff
@@ -169,11 +269,6 @@ namespace dmGraphics
 
     void LoadVulkanFunctions(VkInstance vk_instance)
     {
-        if (g_functions_loaded)
-        {
-            return;
-        }
```
Jhonnyg, line 175: The pointers are based on the instance, so we need to be able to support re-entry here.

engine/graphics/src/vulkan/android/graphics_vulkan_android_surface.cpp
```diff
@@ -1,44 +0,0 @@
...
-#include <dlib/array.h>
-#include <dmsdk/dlib/android.h>
-#include <android_native_app_glue.h>
-
-#include "../graphics_vulkan_defines.h"
-#include "../graphics_vulkan_private.h"
-
-namespace dmGraphics
-{
-    VkResult CreateWindowSurface(HWindow window, VkInstance vkInstance, VkSurfaceKHR* vkSurfaceOut, const bool enableHighDPI)
```
Jhonnyg, line 25: Moved this into graphics_vulkan_android.cpp instead.

external/glfw/lib/android/android_init.c
```diff
@@ -448,21 +459,7 @@ GLFWAPI int32_t glfwAndroidWindowOpened()
...
-        if (!result)
-        {
-            destroy_gl_surface(&_glfwWinAndroid);
-            _glfwWinAndroid.should_recreate_surface = 1;
-            _glfwWin.iconified = 1;
-            return result;
-        }
-    }
-    return 1; // surface is ok
+    return _glfwAndroidPlatformVerifySurface();
```
Jhonnyg, line 462: The android glfw implementation has been split into two now, one that uses the EGL/GL functions and one that is the "No api" implementation (used for vulkan). Otherwise we would have to link EGL libraries when only using vulkan, which felt wrong.

JCash (approved):
lgtm

editor/src/clj/editor/app_manifest.clj
```diff
@@ -481,16 +481,63 @@
...
+            (generic-contains-toggles android :symbols ["GraphicsAdapterOpenGLES" "GraphicsAdapterVulkan"])
+            (generic-contains-toggles android :excludeDynamicLibs ["vulkan"])
+            (generic-contains-toggles android :dynamicLibs ["EGL" "GLESv1_CM" "GLESv2"]))
+    :open-gl (concat
+               open-gl-android-toggles
+               (exclude-libs-toggles android ["graphics_vulkan"])
+               (generic-contains-toggles android :excludeDynamicLibs ["vulkan"])
+               (generic-contains-toggles android :excludeSymbols ["GraphicsAdapterVulkan"]))
+    :vulkan vulkan-android-toggles
+    :both))
```
vlaaad, line 539: Isn't it a bit strange that we have 2 `:both` cases here?

PR #12245, opened 2026-04-20
Fix tests stability

AGulev:


engine/dlib/src/test/test_mdns.cpp
```diff
@@ -2231,6 +2231,7 @@ TEST(MDNS, ZeroTtlAnnouncements)
     Pump(mdns, 0, 20, 5 * 1000);
 
     ASSERT_EQ(dmMDNS::RESULT_OK, dmMDNS::RegisterService(mdns, &service));
+    dmTime::Sleep(100 * 1000);
```
AGulev, line 2234: it fails sometimes on windows CI, I hope it iwll help

engine/profiler/src/basic/profiler_basic.cpp
```diff
@@ -976,47 +976,52 @@ static void ProfilePropertyReset(void* ctx, ProfileIdx idx)
...
+    ProfilePropertySetU64,
+    ProfilePropertySetF64,
+    ProfilePropertyAddS32,
+    ProfilePropertyAddU32,
+    ProfilePropertyAddF32,
+    ProfilePropertyAddS64,
+    ProfilePropertyAddU64,
+    ProfilePropertyAddF64,
+    ProfilePropertyReset
+};
```
AGulev, line 1020: fixes TSAN issue in nightly builds (avoid rewriting g_Listener on every app start)

engine/render/src/test/test_render_script.cpp
```diff
@@ -525,7 +525,11 @@ TEST_F(dmRenderScriptTest, TestLuaRenderTargetSetSizeInvalid)
     "        v_wrap = graphics.TEXTURE_WRAP_MIRRORED_REPEAT\n"
     "    }\n"
     "    self.rt = render.render_target({[graphics.BUFFER_TYPE_COLOR0_BIT] = params_color})\n"
-    "    render.set_render_target_size(self.rt, %s, 4)\n"
+    "    local ok, err = pcall(render.set_render_target_size, self.rt, %s, 4)\n"
+    "    render.delete_render_target(self.rt)\n"
+    "    self.rt = nil\n"
+    "    assert(not ok, \"expected render.set_render_target_size to fail\")\n"
+    "    error(err, 0)\n"
```
AGulev, line 532: fixes asan in nightly build

engine/profiler/src/basic/profiler_basic.cpp
```diff
@@ -976,47 +976,52 @@ static void ProfilePropertyReset(void* ctx, ProfileIdx idx)
 }
 
 static const char* g_ProfilerName = "ProfilerBasic";
-static ProfileListener g_Listener = {};
+static ProfileListener g_Listener =
+{
```
JCash: We should avoid patterns like this unless it's a small POD style struct.
Here it's way too easy to introduce a change, e.g. adding/switching members without getting an actual error.

In the future (c++20 on all platforms), we could use `designated initializers`.

```
WARNING: ThreadSanitizer: data race (pid=98920)
  Write of size 8 at 0x000100606878 by main thread:
    #0 ProfilerBasic_AppInitialize(ExtensionAppParams*) profiler_basic.cpp:987 (test_profilerext_lua:x86_64+0x100221c41)
```

What part of the shutdown is actually causing the issue?
If we don't have the mechanics to wait for our systems to finish, I'd say we have an actual issue when rebooting?

JCash: To modify this PR (to treat the symptom), create a helper struct with the initialization code.
Add a variable at global scope, and you'll get the same effect.

```
static ProfileListener g_Listener = {};
struct ProfileListenerInitializer
{
    ProfileListenerInitializer()
    {
        g_Listener.m_Create = CreateListener;
        g_Listener.m_Destroy = DestroyListener;
        g_Listener.m_SetThreadName = SetThreadName;
        ... and so on
    }
} g_ProfileListenerInitializer;

JCash (changes requested):
The PR tries to remove the symptom, whereas the actual cause isn't addressed.
Also, the symptom fix needs cleanup.

engine/profiler/src/basic/profiler_basic.cpp
AGulev: we have a special APP_EXIT_CODE_REBOOT, so the profiler listener satay registered across the next app lifecycle
So the next `LogInitialize()` starts and call `ProfileSetThreadName()` before `ProfilerBasic_AppInitialize()` which used to always override `g_Listener`.

so it's kinda the fix for this case based on how reboot works.

I'll implement it the way you requested, thank you

JCash: Ok, seems like we have another issue in code reboot flow then.
But that's for another task.

JCash (approved):

PR #12250, opened 2026-04-21
Editor: Fix manual-size-mode sprite not displaying

### LINKED ISSUE #12224: Manual sprite size shown in editor but not set in collection file.
**Describe the bug**
In case of a fresh sprite component, setting the size mode to manual and then choosing an animation results in the size property being initialized in the editor but not being set to the collection file. (Because of this the sprite is not displayed in the game.)

**To Reproduce**
1. Create a game object and add a sprite component.
2. Set sprite's size mode to manual.
3. Set sprite's image and default animation.
4. See error: Size property is initialized in editor but  when running the game, the sprite is not displayed.

**Expected behavior**
The size value shown in editor should also be set to the collection file.

**Defold version:**
  - Version 1.12.3
  - Editor SHA1 0ad9c86fa0a9f7ac19bc468b1a67ee06bb2578b5

**Platforms**
 - OS: Linux Mint

JosephFerano:
Fixed an edge case where if the the size mode was set to manual for a sprite before having selected a default animation, the size wouldn't get serialized and then the sprite would not render at runtime.

Fixes #12224

### DIFF
--- a/editor/src/clj/editor/sprite.clj
+++ b/editor/src/clj/editor/sprite.clj
@@ -508,6 +508,21 @@
 (g/defnode SpriteNode
   (inherits resource-node/ResourceNode)
   (property default-animation g/Str ; Required protobuf field.
+            ;; NOTE: There's an edge case where if the user sets :sprite-mode-manual before having selected the default animation,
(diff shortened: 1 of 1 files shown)


matgis (approved):
Nice find!

PR #12253, opened 2026-04-21
Unify deps download between editor and bob

vlaaad:
Now, both editor and command-line bob use the same parallelized code to fetch dependencies. Since the editor now also parallelizes deps download, this change comes with an open project time improvement. If a project uses high-speed CDN for serving deps, the perf improvement, as measured on an example large project, is 5% (1m14s->1m10s). For typical, smaller projects that use GitHub for deps download, the improvement can reach 28%, since most of the time is spent on deps checking when opening projects.

Related to #11989


com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/util/LibraryTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Test
+    public void testReadArchiveParsesIncludeDirs() throws Exception {
+        assertIncludeDirs("", Set.of());
+        assertIncludeDirs("foo bar baz", Set.of("foo", "bar", "baz"));
+        assertIncludeDirs("foo,bar,baz", Set.of("foo", "bar", "baz"));
+        assertIncludeDirs("  foo  bar  baz  ", Set.of("foo", "bar", "baz"));
+        assertIncludeDirs("  ,, foo , bar ,,, baz  ", Set.of("foo", "bar", "baz"));
+        assertIncludeDirs("\tfoo\t\tbar\tbaz", Set.of("foo", "bar", "baz"));
+        assertIncludeDirs("im\\ possible", Set.of("im\\", "possible"));
```
vlaaad, line 37: Moved from editor tests

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/fs/ZipMountPoint.java
```diff
@@ -102,18 +102,23 @@ public boolean isFile() {
     public ZipMountPoint(IFileSystem fileSystem, String archivePath) {
         this.fileSystem = fileSystem;
         this.archivePath = archivePath;
+        this.projectArchive = null;
+        this.includeDirs = Collections.emptySet();
+        this.includeBaseDir = "";
     }
 
-    public ZipMountPoint(IFileSystem fileSystem, String archivePath, boolean projectZip) {
```
vlaaad, line 107: We no longer have a `projectZip` flag. Instead, the `projectArchive` field is used for checking if it's a project or not. Benefit: no unnecessary reparsing of the zip file here to extract base dir and include dirs.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -559,18 +563,16 @@ private static void setupProject(Project project, boolean resolveLibraries, IPro
             project.loadProjectFile(false);
             BobProjectProperties projectProperties = project.getProjectProperties();
             String[] dependencies = projectProperties.getStringArrayValue("project", "dependencies");
-            List<URL> libUrls = new ArrayList<>();
-            for (String val : dependencies) {
-                libUrls.add(new URL(val));
-            }
-
-            project.setLibUrls(libUrls);
+            project.setLibUrls(Arrays.stream(dependencies).map(URI::create).toList());
```
vlaaad, line 566: We now use URIs, not URLs.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/Library.java
```diff
@@ -0,0 +1,497 @@
...
+            progress.message(new IProgress.Message.DownloadingArchive(sanitizedUri));
+            var host = hostKey(uri);
+            if (host == null) {
+                throw new IllegalArgumentException("Invalid host URI " + uri);
+            }
+            var permit = hostLimits.computeIfAbsent(host, _ -> new Semaphore(FETCHES_PER_HOST));
+            permit.acquire();
+            try {
+                var cachedArchive = cachedResult.result().archive();
+                var request = HttpRequest.newBuilder(uri).GET().header("Accept", "application/zip");
```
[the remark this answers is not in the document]
vlaaad: We never supported files to begin with.

AGulev (dismissed):

editor/src/clj/editor/progress.clj
```diff
@@ -46,6 +47,30 @@
...
+       IProgress$Message$GeneratingReport (localization/message "progress.generating-report")
+       IProgress$Message$Working (localization/message "progress.working")
+       IProgress$Message$ReadingClasses (localization/message "progress.reading-classes")
+       IProgress$Message$DownloadingArchives (localization/message "progress.downloading-archives" {"count" (.count ^IProgress$Message$DownloadingArchives message)})
+       IProgress$Message$DownloadingArchive (localization/message "progress.downloading-archive" {"uri" (str (.uri ^IProgress$Message$DownloadingArchive message))})
+       localization/empty-message)
+     1000
+     (long (* 1000.0 fraction))
+     cancellable)))
+
```
matgis, line 73: Perhaps it would be nicer to use either a multimethod or a protocol for dispatch?

Although I did see we have a `switch` statement on the Java side, so maybe it is desirable to leave it like this. 🤷

editor/src/clj/editor/workspace.clj
```diff
@@ -536,68 +533,62 @@ ordinary paths."
...
+                                    (let [problem (.problem result)]
+                                      (or (instance? Library$Problem$Missing problem)
+                                          (instance? Library$Problem$FetchFailed problem)
+                                          (instance? Library$Problem$InstallFailed problem))))
+                                  problem-results)
+            show-open-project (coll/any? (fn [^Library$Result result]
+                                           (let [problem (.problem result)]
+                                             (or (instance? Library$Problem$InvalidArchive problem)
+                                                 (instance? Library$Problem$DefoldMinVersion problem))))
+                                         problem-results)]
```
matgis, line 550: Another situation where a few multimethods or a protocol might be nicer?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -761,17 +754,18 @@ public void mount(IResourceScanner resourceScanner) throws IOException, CompileE
...
         boolean missingFiles = false;
+        Map<String, File> libFiles = new HashMap<>();
 
-        for (String url : libFiles.keySet() ) {
-            File file = libFiles.get(url);
-
+        for (var dependency : dependencies) {
+            var archive = dependency.archive();
+            var file = archive == null ? null : archive.path().toFile();
+            libFiles.put(dependency.uri().toString(), file);
```
matgis, line 765: If `archive` is `null` here, it gets reported as `"Some libraries could not be found locally, use the resolve command to fetch them."`, but this could be misleading?

Shouldn't we base the error message on the associated `Problem` instead so users are not led astray when trying to fix it?

Or we could at least hint in the error message that the libraries could be invalid as well as missing?

editor/src/clj/editor/app_view.clj
```diff
@@ -3201,30 +3202,20 @@
...
+  (let [library-uris (project/project-dependencies project)]
+    (future
+      (error-reporting/catch-all!
+        (ui/with-progress [render-fetch-progress! (make-render-task-progress :fetch-libraries)]
+          (let [lib-results (library/fetch! (workspace/project-directory workspace) library-uris render-fetch-progress!)
+                render-install-progress! (make-render-task-progress :resource-sync)]
+            (render-install-progress! (progress/make (localization/message "progress.installing-updated-libraries")))
+            (ui/run-later
+              (workspace/set-project-dependencies! workspace lib-results)
+              (disk/async-reload! render-install-progress! workspace [] changes-view
```
matgis: This is a great opportunity to introduce a line break to decrease indentation!

editor/src/clj/editor/workspace.clj
```diff
@@ -536,68 +533,62 @@ ordinary paths."
...
+                                       {"required" (.required ^Library$Problem$DefoldMinVersion problem)
+                                        "current" EngineVersion/version})
+    Library$Problem$InstallFailed (localization/message "notification.fetch-libraries.problem.install-failed")
+    (localization/message "notification.fetch-libraries.problem.unknown" {"problem" problem})))
+
+(defn library-result-message [^Library$Result result]
+  (localization/message
+    "notification.fetch-libraries.dependency-problem"
+    {"uri" (.uri result)
+     "problem" (library-result-problem-message (.problem result))}))
```
matgis: Do these maybe belong in `library.clj`?

editor/test/integration/library_test.clj
```diff
@@ -181,8 +153,32 @@
...
+          original-uri (URI/create "https://example.com/local-extension-dir.zip")
+          property-prefix "defold.extension.test-local"
+          local-extension-dir (io/file "test/resources/lib_resource_project")
+          game-project-resource (workspace/resolve-workspace-resource workspace "/game.project")
+          expected-library-file (test-support/library-file project-directory original-uri "")]
+      (System/setProperty (str property-prefix ".url") (str original-uri))
+      (System/setProperty (str property-prefix ".path") (.getCanonicalPath local-extension-dir))
+      (try
+        (write-deps! game-project-resource "{{defold.extension.test-local.url}}")
+        (let [results (library/fetch! (workspace/project-directory workspace) (project/read-dependencies game-project-resource) progress/null-render-progress!)
```
matgis: The `project-directory` is already bound in the surrounding `let`.

editor/test/integration/library_test.clj
```diff
@@ -44,81 +45,53 @@
...
-              update-statuses (group-by :status update-states)
-              files           (library/library-files project-directory)
-              states          (library/current-library-state project-directory uris)]
-          (is (= 3 (count (update-statuses :up-to-date))))
-          (is (= 1 (count (update-statuses :error))))
-          (is (= (str (:uri (first (update-statuses :error)))) "file:/bogus"))
-          (is (every? (fn [state] (= (:tag state) "tag")) (update-statuses :up-to-date)))
-          (is (= 3 (count files)))
-          (is (= 4 (count states)))
-          (is (every? (fn [state] (= (:tag state) "tag")) (filter :file states))))))))
```
matgis, line 121: Did we lose some test coverage by removing this test?

editor/test/integration/resource_watch_test.clj
```diff
@@ -65,19 +63,6 @@
...
-(deftest include-parsing
-  (testing "sane dirs"
-    (is (not (seq (resource-watch/parse-include-dirs ""))))
-    (is (= (resource-watch/parse-include-dirs "foo bar baz") (list "foo" "bar" "baz")))
-    (is (= (resource-watch/parse-include-dirs "foo,bar,baz") (list "foo" "bar" "baz"))))
-  (testing "various spacing and commas allowed"
-    (is (= (resource-watch/parse-include-dirs "  foo  bar  baz  ") (list "foo" "bar" "baz")))
-    (is (= (resource-watch/parse-include-dirs "  ,, foo , bar ,,, baz  ") (list "foo" "bar" "baz")))
-    (is (= (resource-watch/parse-include-dirs "\tfoo\t\tbar\tbaz") (list "foo" "bar" "baz")))
-    (is (= (resource-watch/parse-include-dirs "\r\nfoo\rbar\nbaz") (list "foo" "bar" "baz"))))
```
matgis, line 77: This newline case is not covered by the corresponding test in `LibraryTest.java` now. Was the removal intentional?

editor/test/integration/test_util.clj
```diff
@@ -574,9 +579,6 @@
 
 (def load-system-and-project (fn/memoize load-system-and-project-raw))
 
-(defn clear-cached-libraries! []
-  (fn/clear-memoized! (var-get #'editor.library/fetch-library!)))
-
```
matgis, line 579: I think there might have been a misunderstanding around why we were memoizing the `fetch-library` function before. 🤔

This was done so that the various integration tests that create scratch project directories do not have to download large libraries over and over again when the entire test suite is run.

The new caching mechanism in `Library/cached` appears to only reuse libraries that were already downloaded to the `.internal/lib` directory below the project directory. But that directory won't exist when we initialize new scratch project directories.

I ran the tests, and they do in fact appear to run slightly faster than on `dev`, but if we're re-downloading large libraries like `extension-rive` to multiple scratch project directories now, things might get even quicker!

Maybe I'm misunderstanding the new caching mechanism, though?

editor/test/support/test_support.clj
```diff
@@ -77,6 +81,18 @@
 (defn write-until-new-mtime [f content]
   (do-until-new-mtime (fn [f] (fs/create-file! f content)) f))
 
+(defn library-directory ^File [project-directory]
+  (io/file (io/as-file project-directory) ".internal/lib"))
+
+(defn library-files [project-directory]
+  (seq (.listFiles (library-directory project-directory))))
+
+(defn library-file ^File [project-directory ^URI library-uri tag]
```
matgis: Style: Newline before return type tags.

editor/test/support/test_support.clj
```diff
@@ -77,6 +81,18 @@
...
+  (let [hash (DigestUtils/sha1Hex (str library-uri))
+        ^String tag (or tag "")
+        encoded-tag (.encodeToString (Base64/getUrlEncoder) (.getBytes tag "UTF-8"))]
+    (io/file (library-directory project-directory) (str hash "-" encoded-tag ".zip"))))
+
```
matgis, line 97: Is this code we should really be sharing with Bob? 🤔

matgis (dismissed):
Seems to work great!

I did have one concern around caching of the downloaded libraries we might want to address, though.

editor/src/clj/editor/progress.clj
vlaaad: I think it's fine, protocol or multimethod will lead to even more code.

editor/test/integration/resource_watch_test.clj
vlaaad: Yes, it was wrong to test newlines because they're not possible in the data model, since newlines in `game.project` are used as value terminators, e.g.:
```
[library]
include_dirs=foo
bar
baz
```
bar and baz are not part of the `include_dirs` field at all because of the `game.project` file syntax.

editor/test/integration/library_test.clj
vlaaad: Looks like we did lose a validation that invalid deps don't prevent valid ones from being downloaded/installed. I'll add a test to bob.

PR #12255, opened 2026-04-21
using a scaled (pixel-converted) height instead of font units

ivasilyeu:
It looks like, for dynamic fonts, the text layout calculation was using font-units instead of pixels, when returning the result of resource.get_text_metrics(font, ...).
This PR makes sure a scaled (pixel-converted) height gets returned, instead of font units, for the font resource metrics calculation output.

a demo project that reproduces the issue: 
[text_metrics_demo.zip](https://github.com/user-attachments/files/26932899/text_metrics_demo.zip)


JCash:
Thanks!
As we want pull requests to `dev` branch, I prepared the same fix here: https://github.com/defold/defold/pull/12258
(as a way to speed up the fix landing in beta)

PR #12256, opened 2026-04-21
Editor: Prevent editor from making invalid collision shapes

### LINKED ISSUE #8957: Physics crashes when adding collision shape: box. and set scale 0
When adding collisionobject with shape box to game object and create this object through factory we will get a bug.

**To Reproduce (REQUIRED)**
1. Create GO in Assets
2. add #collisionobject with collision shape: box
3. Create a GO through the factory

No crash

  - Version 1.8.0

**Platforms (REQUIRED):**
 - Platforms: Windows
 - OS: Windows 11

**Logs (OPTIONAL):**

```
Assertion failed: area > 1.192092896e-07F, file ..\src\box2d\Box2D\Collision\Shapes\b2PolygonShape.cpp, line 123
INFO:CRASH: Successfully wrote Crashdump to file: C:\Users\cdart\AppData\Roaming\Defold/_crash
ERROR:CRASH: CALL STACK:

ERROR:CRASH:  0 0x7FF79792AFD0 dmCrash::GenerateCallstack D:\a\defold\defold\engine\crash\src\backtrace_win32.cpp:144
ERROR:CRASH:  1 0x7FF797CE7B54 raise /tmp/job2174380374303029695/minkernel/crts/ucrt/src/appcrt/misc/signal.cpp:547
ERROR:CRASH:  2 0x7FF797CD8250 abort /tmp/job2174380374303029695/minkernel/crts/ucrt/src/appcrt/startup/abort.cpp:71
ERROR:CRASH:  3 0x7FF797CD7534 common_assert_to_stderr<wchar_t> /tmp/job2174380374303029695/minkernel/crts/ucrt/src/appcrt/startup/assert.cpp:186
ERROR:CRASH:  4 0x7FF797CD704C _wassert /tmp/job2174380374303029695/minkernel/crts/ucrt/src/appcrt/startup/assert.cpp:443
ERROR:CRASH:  5 0x7FF797B714C0 ComputeCentroid D:\a\defold\defold\engine\physics\src\box2d\Box2D\Collision\Shapes\b2PolygonShape.cpp:124
ERROR:CRASH:  6 0x7FF797B6FEE0 b2PolygonShape::Set D:\a\defold\defold\engine\physics\src\box2d\Box2D\Collision\Shapes\b2PolygonShape.cpp:180
ERROR:CRASH:  7 0x7FF797AD1F40 dmPhysics::TransformCopyShape D:\a\defold\defold\engine\physics\src\physics\physics_2d.cpp:834
ERROR:CRASH:  8 0x7FF797ACD8E0 dmPhysics::NewCollisionObject2D D:\a\defold\defold\engine\physics\src\physics\physics_2d.cpp:988
ERROR:CRASH:  9 0x7FF797A043D0 dmGameSystem::CreateCollisionObject D:\a\defold\defold\engine\gamesys\src\gamesys\components\comp_collision_object.cpp:420
ERROR:CRASH: 10 0x7FF797A01B00 dmGameSystem::CompCollisionObjectCreate D:\a\defold\defold\engine\gamesys\src\gamesys\components\comp_collision_object.cpp:470
ERROR:CRASH: 11 0x7FF797985240 dmGameObject::CreateComponents D:\a\defold\defold\engine\gameobject\src\gameobject\gameobject.cpp:829
ERROR:CRASH: 12 0x7FF7978F6E70 dmGameObject::AcquireResources D:\a\defold\defold\engine\gameobject\src\gameobject\res_collection.cpp:143
ERROR:CRASH: 13 0x7FF7978F7810 dmGameObject::ResCollectionCreate D:\a\defold\defold\engine\gameobject\src\gameobject\res_collection.cpp:262
ERROR:CRASH: 14 0x7FF79796E4A0 dmResource::DoCreateResource D:\a\defold\defold\engine\resource\src\resource.cpp:628
ERROR:CRASH: 15 0x7FF79796E9E0 dmResource::CreateAndLoadResource D:\a\defold\defold\engine\resource\src\resource.cpp:755
ERROR:CRASH: 16 0x7FF797969430 dmResource::Get D:\a\defold\defold\engine\resource\src\resource.cpp:830
ERROR:CRASH: 17 0x7FF797A8E5F0 dmGameSystem::DoLoad D:\a\defold\defold\engine\gamesys\src\gamesys\components\comp_collection_proxy.cpp:92
ERROR:CRASH: 18 0x7FF797A8DF20 dmGameSystem::CompCollectionProxyOnMessage D:\a\defold\defold\engine\gamesys\src\gamesys\components\comp_collection_proxy.cpp:427
ERROR:CRASH: 19 0x7FF79798BD70 dmGameObject::DispatchMessagesFunction D:\a\defold\defold\engine\gameobject\src\gameobject\gameobject.cpp:2323
ERROR:CRASH: 20 0x7FF7979F69B0 dmMessage::InternalDispatch D:\a\defold\defold\engine\dlib\src\dlib\message.cpp:573
ERROR:CRASH: 21 0x7FF797987ED0 dmGameObject::InitCollection D:\a\defold\defold\engine\gameobject\src\gameobject\gameobject.cpp:1780
ERROR:CRASH: 22 0x7FF797A0A080 dmEngine::Init D:\a\defold\defold\engine\engine\src\engine.cpp:1347
ERROR:CRASH: 23 0x7FF797A08FC0 dmEngineCreate D:\a\defold\defold\engine\engine\src\engine.cpp:2159
ERROR:CRASH: 24 0x7FF797A0ECD0 dmEngine::RunLoop D:\a\defold\defold\engine\engine\src\engine_loop.cpp:68
ERROR:CRASH: 25 0x7FF7979188A0 engine_main D:\a\defold\defold\engine\engine\src\engine_main.cpp:152
ERROR:CRASH: 26 0x7FF797C91034 __scrt_common_main_seh D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288
ERROR:CRASH: 27 0x7FF9C7222560 BaseThreadInitThunk <unknown>:0
ERROR:CRASH: 28 0x7FF9C89AAA20 RtlUserThreadStart <unknown>:0
ERROR:CRASH: 
```

JosephFerano:
Collision shape improvements:
- Scrubbers in the properties UI no longer allow negative dimensions.
- Manually entered invalid values now cause the build to fail instead of crashing at runtime.
- Clearer error message when a shape's computed area is too small due to tiny dimensions and/or scale.

Fixes #12111
Fixes #8957


JosephFerano:
I discovered that the scrubber had a `:min` property I could pass in, so I used that. The values that I selected as the min values above 0 are all kind of semi arbitrary, I just found tried a small number that wouldn't crash the runtime, please feel free to suggest something different if we don't like these min values I selected. Also, now that this will have build errors when the dimensions are wrong, I didn't feel the need to find a way to clamp when manually entering numbers for the properties, since, since I feel it's better UX to scream at the user with red and helpful error messages than to silently clamp.

matgis (commented):
Code changes look good to me!

My only concern is the choice of minimum values. I was kind of hoping the engine would be fine with any non-negative values or at least any positive value, but if it really crashes immediately for very small values as well we might need to do this.

The problem as I see it is that we don't really know the scale that the game content is authored in. A `min-sphere-diameter` of `0.01` might seem generous for a 2D game where shapes are typically several hundred units across, but it really is just a millimeter in a 3D game that uses SI units. Having a 1-millimeter sphere is not unreasonable if you make a game about ants, but I suppose you could run into precision issues.

Nevertheless, it might be better to just enforce the values to be non-negative (or positive if `0.0` causes a crash) unless it results in an immediate runtime error. To me, it is reasonable to be able to create very small shapes, whereas negative values are clearly an error. Zero might be considered an error too, but I haven't tested it.

JosephFerano:
> Code changes look good to me!
> 
> My only concern is the choice of minimum values. I was kind of hoping the engine would be fine with any non-negative values or at least any positive value, but if it really crashes immediately for very small values as well we might need to do this.
> 
> The problem as I see it is that we don't really know the scale that the game content is authored in. A `min-sphere-diameter` of `0.01` might seem generous for a 2D game where shapes are typically several hundred units across, but it really is just a millimeter in a 3D game that uses SI units. Having a 1-millimeter sphere is not unreasonable if you make a game about ants, but I suppose you could run into precision issues.
> 
> Nevertheless, it might be better to just enforce the values to be non-negative (or positive if `0.0` causes a crash) unless it results in an immediate runtime error. To me, it is reasonable to be able to create very small shapes, whereas negative values are clearly an error. Zero might be considered an error too, but I haven't tested it.

Ok, great this was the feedback I was hoping for. I will go and search for even smaller numbers. In my testing for BoxSphere, it cannot be too small, but it's complicated by the fact that it's an area calculation, so perhaps what I can do is actually perform the same area calculation editor side, this would give a little bit more wiggle room for one of the dimensions to be a really small number if the others are reasonably sized. Another question I just thought of is how does the 3D physics engine handle this, since this is affecting box2d it seems.

matgis:
Honestly, we might not even want to validate this on the editor side unless the value is clearly wrong (i.e. negative). Like you said, what constitutes a valid value might differ between physics engines. Negative is obviously wrong. Zero *might* be wrong for certain values, like the box dimensions reported in #12111, but it could also vary between physics engines.

For this reason, I don't think we should be searching for smaller acceptable values as you put it. I think we should validate that the values are non-negative. And if there are properties that must be above zero, we should validate that they are indeed positive, but not impose a minimum value above zero.

In the end, the dimensions of the physics shapes are also affected by the runtime scaling of the object and its parents, like in #8957. The runtime needs to handle that internally. It's not something we can guard against in the editor.

editor/src/clj/editor/collision_object.clj
```diff
@@ -372,13 +372,23 @@
             (default 0.0) ; Used to prevent validation errors during node initialization from editor scripts
             (dynamic label (properties/label-dynamic :collision-object.shape :diameter))
             (dynamic tooltip (properties/tooltip-dynamic :collision-object.shape :diameter))
+            (dynamic edit-type (g/constantly {:type g/Num :min 0.0}))
+
             (dynamic error (g/fnk [_node-id diameter]
-                             (validation/prop-error :fatal _node-id :diameter validation/prop-zero-or-below? diameter diameter-message))))
+                             (validation/prop-error
+                               :fatal _node-id
+                               :diameter validation/prop-zero-or-below? diameter diameter-message))))
```
matgis: Style: It feels a bit strange to break the arguments across multiple lines as if they are key-value pairs when they aren't?

editor/src/clj/editor/collision_object.clj
```diff
@@ -372,13 +372,23 @@
...
   (display-order [Shape :diameter])
 
   (output scene g/Any produce-sphere-shape-scene)
 
+  (output shape-errors g/Any (g/fnk [_node-id id id-counts diameter]
+                               (g/package-errors _node-id
+                                 (validate-image-id _node-id id id-counts)
+                                 (validation/prop-error
+                                   :fatal _node-id
+                                   :diameter validation/prop-zero-or-below? diameter diameter-message))))
```
matgis: Style: It feels a bit strange to break the arguments across multiple lines as if they are key-value pairs when they aren't?

editor/src/clj/editor/collision_object.clj
```diff
@@ -397,6 +407,14 @@
...
   [:scale-uniform])
 
+(defn- prop-error-box-dimensions [node-id dimensions]
+  (validation/prop-error :fatal node-id :dimensions
+                         (fn [dimensions _]
+                           (when (some #(<= ^double % 0.0) dimensions)
+                             (localization/message "error.collision-object-shape-dimensions-must-be-greater-than-zero")))
+                         dimensions
+                         dimensions-message))
+
```
matgis, line 412: You're supplying `dimensions-message` as an argument to the validation `fn`, but you're ignoring the argument and returning your own message instead?

editor/src/clj/editor/collision_object.clj
```diff
@@ -442,19 +458,33 @@
...
   (property height g/Num ; Always assigned in load-fn.
             (default 0.0) ; Used to prevent validation errors during node initialization from editor scripts
             (dynamic label (properties/label-dynamic :collision-object.shape :height))
             (dynamic tooltip (properties/tooltip-dynamic :collision-object.shape :height))
+            (dynamic edit-type (g/constantly {:type g/Num :min 0.0}))
             (dynamic error (g/fnk [_node-id height]
-                             (validation/prop-error :fatal _node-id :height validation/prop-zero-or-below? height height-message))))
+                             (validation/prop-error
+                               :fatal _node-id
+                               :height validation/prop-zero-or-below? height height-message))))
```
matgis: I think it might be legal for a Capsule to have zero height? That would make it a sphere, but it could be scripted to adjust its height at runtime.

matgis: Style: It feels a bit strange to break the arguments across multiple lines as if they are key-value pairs when they aren't?

editor/src/clj/editor/collision_object.clj
```diff
@@ -442,19 +458,33 @@
...
-
+  (output shape-errors g/Any (g/fnk [_node-id id id-counts diameter height]
+                               (g/package-errors _node-id
+                                 (validate-image-id _node-id id id-counts)
+                                 (validation/prop-error
+                                   :fatal _node-id
+                                   :diameter validation/prop-zero-or-below? diameter diameter-message)
+                                 (validation/prop-error
+                                   :fatal _node-id
+                                   :height validation/prop-zero-or-below? height height-message))))
```
matgis: Style: It feels a bit strange to break the arguments across multiple lines as if they are key-value pairs when they aren't?

editor/resources/localization/en.editor_localization
```diff
@@ -2274,7 +2274,7 @@ error.property-contains-url-prohibited-characters = "{property}" should not cont
 error.property-cannot-start-or-end-with-space = "{property}" should not start or end with a space character.
 error.property-cannot-be-negative = "{property}" cannot be negative.
 error.property-must-be-greater-than-zero = "{property}" must be greater than zero.
-error.property-must-be-at-least = "{property}" must be at least {min}.
+error.property-must-be-at-least = "{property}" must be at least {min,number,0.0#}.
```
matgis: It seems you didn't end up using this, so you probably shouldn't change it? Also not sure we should be imposing a precision limit here given that it could be relevant for the thing it's used for. And finally, if we change it for one language we should probably change it for all of them?

Better to revert this change I feel.

editor/src/clj/editor/collision_object.clj
```diff
@@ -442,19 +458,33 @@
             (default 0.0) ; Used to prevent validation errors during node initialization from editor scripts
             (dynamic label (properties/label-dynamic :collision-object.shape :diameter))
             (dynamic tooltip (properties/tooltip-dynamic :collision-object.shape :diameter))
+            (dynamic edit-type (g/constantly {:type g/Num :min 0.0}))
             (dynamic error (g/fnk [_node-id diameter]
-                             (validation/prop-error :fatal _node-id :diameter validation/prop-zero-or-below? diameter diameter-message))))
+                             (validation/prop-error
+                               :fatal _node-id
+                               :diameter validation/prop-zero-or-below? diameter diameter-message))))
```
matgis: Style: It feels a bit strange to break the arguments across multiple lines as if they are key-value pairs when they aren't?

matgis (changes requested):
We should check if it really is illegal for capsules to have zero height. Rejecting negative values should be OK, though.

editor/src/clj/editor/collision_object.clj
JosephFerano: Good catch, I didn't notice it when I extracted this function. I think originally it was used as a more generic error message but then later changes made the error more specific. I cleaned up the variable as well.

editor/src/clj/editor/collision_object.clj
JosephFerano: I ran it with 0.0 height and indeed it seems to behave just fine, like a sphere.

matgis (approved):
Editor changes look good to me! I'll defer to Mathias regarding the changes to the engine runtime.

JCash (approved):

PR #12257, opened 2026-04-21
Parallelize directory traversal when syncing resources

vlaaad:
This change improves open load time in big projects by 5% (1m12s -> 1m08s)

Fix #11989

### DIFF
--- a/editor/src/clj/util/coll.clj
+++ b/editor/src/clj/util/coll.clj
@@ -1088,9 +1088,17 @@
    (pmapv #(apply f %) (apply mapv vector coll colls))))
 
 (defn ptree
-  "Builds a tree in parallel. `children-fn` returns the ordered children for a
-  node, while `build-fn` combines the node with the vector of built child
-  results."
+  "Build a tree in parallel
+
+  Args:
+    children-fn    a function that returns the ordered child nodes for a node;
+                   will receive 1 arg: node; should return a collection of
+                   child nodes or nil
+    build-fn       a function that builds the result for a node; will receive 2
+                   args: node and built-children (vector of results returned by
+                   build-fn for the node's children, in child order); should
+                   return the built result for the node
+    root           the root node to build from"
   [children-fn build-fn root]
   (let [budget (Semaphore. (default-parallelism))]
     (letfn [(parallel-child-results [child-nodes]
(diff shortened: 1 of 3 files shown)


editor/src/clj/util/coll.clj
```diff
@@ -1083,3 +1086,41 @@
            (LazilyPersistentVector/createOwning results))))))
   ([f coll & colls]
    (pmapv #(apply f %) (apply mapv vector coll colls))))
+
+(defn ptree
+  "Builds a tree in parallel. `children-fn` returns the ordered children for a
+  node, while `build-fn` combines the node with the vector of built child
+  results."
```
matgis: Include description of what `build-fn` will be called with and is expected to return?

editor/src/clj/util/coll.clj
```diff
@@ -1083,3 +1086,41 @@
...
+                                    0 []
+                                    1 [(visit (child-nodes 0))]
+                                    (if (.tryAcquire budget)
+                                      (try
+                                        (parallel-child-results child-nodes)
+                                        (finally
+                                          (.release budget)))
+                                      (mapv visit child-nodes)))]
+                (build-fn node child-results)))]
+      (visit root))))
```
matgis, line 1133: I ran the changes through Codex code review, and it came back with some concerns that seem reasonable to me. What do you think?

• The new parallel tree builder can spawn an unbounded number of child tasks for a single wide node, and `make-file-tree` now uses it for filesystem traversal. That introduces a real scalability regression on large directories, so the patch is not fully correct.

  Review comment:

  - [P2] Keep `ptree` from forking one task per child in wide nodes.
    When a node has many children, `parallel-child-results` still calls `.fork` once per child after acquiring only a single semaphore permit. In a project tree with a wide directory (for example, an imported asset folder with thousands of files), `make-file-tree` will now create thousands of virtual threads at once, which defeats the bounded-parallelism goal from `pmapv` and can noticeably increase memory use or stall resource scanning. Limiting the number of workers per node, as `pmapv` does, would avoid this regression.

vlaaad:
Nice catch. Fixed, and even got a tiny perf improvement for the tree build (2.3s -> 2.1s, 9%)

PR #12262, opened 2026-04-22
Hotfix: Only warp when a dolly/track/tumble are active

JosephFerano:
Fix cursor warping around screen edges when trying to select or using move/rotate/scale manips.


vlaaad (approved):
LGTM!

PR #12268, opened 2026-04-23
Fix `liveupdate.is_built_with_excluded_files()` were broken by old liveupdate functions deprecation

AGulev:
`liveupdate.is_built_with_excluded_files()` should work even if user decided to exclude liveupdate entries from manifest

### DIFF
--- a/engine/liveupdate/src/liveupdate.cpp
+++ b/engine/liveupdate/src/liveupdate.cpp
@@ -114,7 +114,7 @@ namespace dmLiveUpdate
         dmResourceProvider::HArchive    m_LiveupdateArchive;
         dmResource::HManifest           m_LiveupdateArchiveManifest;
         bool                            m_IsEnabled;
-        bool                            m_MainConttextHasExcludedEntries;
+        bool                            m_MainContextHasExcludedResources;
 
     } g_LiveUpdate;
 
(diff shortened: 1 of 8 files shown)


engine/liveupdate/src/liveupdate.cpp
```diff
@@ -114,7 +114,7 @@ namespace dmLiveUpdate
         dmResourceProvider::HArchive    m_LiveupdateArchive;
         dmResource::HManifest           m_LiveupdateArchiveManifest;
         bool                            m_IsEnabled;
-        bool                            m_MainConttextHasExcludedEntries;
```
AGulev, line 117: typo: `tt`

engine/resource/proto/resource/liveupdate_ddf.proto
```diff
@@ -114,11 +114,15 @@ message ResourceEntry {
...
  * - resources                : The resources that are part of the manifest.
+ * - has_excluded_resources   : True when the build produced one or more
+ *                              resources marked EXCLUDED, regardless of whether
+ *                              those entries are present in this manifest.
  */
 message ManifestData {
     required ManifestHeader     header = 1;
     repeated HashDigest         engine_versions = 2;
     repeated ResourceEntry      resources = 3;
+    optional bool               has_excluded_resources = 4 [default = false];
```
AGulev, line 125: a new optional flag instead of relying on liveupdate manifest

JCash (approved):

PR #12269, opened 2026-04-23
Vulkan dynamic rendering support

Jhonnyg:


Jhonnyg:
It has grown stale, no point in continuing on this branch right now.

PR #12272, opened 2026-04-23
Metal graphics backend support

### LINKED ISSUE #12271: Metal graphics adapter
Add a native metal graphics adapter for OSX and iOS.

Jhonnyg:
A Metal adapter has been added to the engine that can be used for both OSX and iOS devices. At the moment, this is only available when building the engine locally by passing the `--with-metal` build system switch:

./script/build.py make_solution --platform=arm64-ios -- --with-metal
./script/build.py build_engine --platform=arm64-macos -- --with-metal

Fixes #12271

#### Technical changes

* new Metal graphics adapter, graphics_metal.mm, enabled with --with-metal on macOS/iOS targets
* Metal C++ wrapper support via Metal.hpp / metal_cpp.cpp, allowing the backend to use Apple Metal APIs from C++/Obj-C++
* build-system plumbing for Metal in waf, CMake, and scripts/build.py, including Metal framework linking and C++17 where needed
* new graphics adapter family/API entries for Metal in graphics/platform enums and adapter registration
* GLFW window creation path updated to use a no-API window for Metal, similar to Vulkan/DX12
* Metal shader pipeline support through MSL generation in shaderc/Bob
* ShaderDesc additions for Metal resource mappings and compute workgroup sizes
* SPIRV-Cross reflection support for Metal argument-buffer resource indices
* Metal argument-buffer based resource binding for uniforms, textures, samplers, and storage resources
* Metal render pipeline and compute pipeline creation/caching
* Metal texture support, including 2D, 2D array, 3D, cubemap, sub-updates, compressed formats, and async upload handling
* Metal render target support, including depth/stencil, MSAA, clears, viewport/scissor handling, and adjusted NDC/flip-Y behavior
* Metal compute shader dispatch support, including workgroup size metadata
* test-app updates with MSL shader assets and Metal packaging/build coverage
* renderer integration updates so Metal uses the same multi-buffering and adjusted-NDC paths as Vulkan/WebGPU/DX12


build_tools/waf_dynamo.py
```diff
@@ -470,7 +476,9 @@ def default_flags(self):
     for f in ['CFLAGS', 'CXXFLAGS']:
         self.env.append_value(f, flags)
 
-    if not use_cl_exe:
+    if Options.options.with_metal or platform_supports_feature(build_util.get_target_platform(), 'metal', {}):
+        self.env.append_value('CXXFLAGS', ['-std=c++17']) # Due to metal-cpp library
```
Jhonnyg, line 480: Notable change here. This is because of the "Metal.hpp" shim we use as the C++ -> OBJ-C wrapper. For now the metal adapter is just a dev feature, there's no extender/manifest support yet so no other changes required atm.

engine/graphics/proto/graphics/graphics_ddf.proto
```diff
@@ -441,13 +441,30 @@ message ShaderDesc
...
+        optional uint32 set       = 3;
+        optional uint32 msl_index = 4;
+    }
+
+    message WorkGroupSize
+    {
+        optional uint32 x = 1;
+        optional uint32 y = 2;
+        optional uint32 z = 3;
+    }
```
Jhonnyg, line 457: This is typically embedded in spir-v, but not in other languages (msl mostly)

engine/graphics/src/test/test_app_graphics.cpp
```diff
@@ -202,6 +202,73 @@ struct ClearBackbufferTest : ITest
     }
 };
 
+struct MslArgumentBuffersTest : ITest
```
Jhonnyg, line 205: The additions in the test_app_graphics were added last year so I'd like to refactor this file AFTER we have merged this PR. I would like to separate this into two parts:

* strip all of the rendering test code so this file is more of an explicit test (don't remember exactly what it did in the beginning, but I'll figure it out)
* add a new test app that we can use to test small explicit graphics/rendering features (UBO, SSBO and so on). 
* make sure the new test app works with all graphics adapters, move the shader building into the pipeline so they get embedded into the binary instead of using the python script.

engine/graphics/src/graphics.h
```diff
@@ -83,6 +83,7 @@ namespace dmGraphics
         ADAPTER_FAMILY_PRIORITY_VENDOR   = 0,
         ADAPTER_FAMILY_PRIORITY_WEBGPU   = 0,
         ADAPTER_FAMILY_PRIORITY_DIRECTX  = 0,
+        ADAPTER_FAMILY_PRIORITY_METAL    = 0,
```
Jhonnyg, line 86: Maybe these should be called ADAPTER_FAMILY_PRIORITY_NATIVE or something like that?

Jhonnyg:
@codex review pr

engine/graphics/src/metal/graphics_metal.mm
```diff
@@ -0,0 +1,5029 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#include <type_traits>
+
+// Metal.hpp is Taken from https://github.com/bkaradzic/metal-cpp/blob/metal-cpp_macOS15.2_iOS18.2/SingleHeader/Metal.hpp
+#include <Metal.hpp>
```
JCash, line 18: Please add the license/dependency in the COMPLYING_WITH_LICENSES.md

JCash (dismissed):
lgtm

PR #12279, opened 2026-04-26
Avoid building of the engine during `waf install` process

### LINKED ISSUE #10799: Investigate why WAF triggers build step twice
**Task (REQUIRED):**
For `engine` module WAF triggers build twice if `waf build install` called. That leads to https://github.com/defold/defold/issues/10785

Steps to reproduce:
1. Build engine with `./scripts/build.py build_engine`
2. `cd engine/engine`
3. Run `waf install`
4. See result.

During installation WAF calls recompiling.

`install` command executes only file installation if `build` step was done before it.

AGulev:
Submodules building and re-building of separate modules should be much faster now

Fix https://github.com/defold/defold/issues/10799

### Technical changes

* Force standalone builtins archive generation to use one Bob thread to avoid .arcd offset races.
* Set deterministic ZIP entry timestamps for generated archives.
* Replace non-deterministic shader GENERATOR_ID with a stable material-path hash.
* Make Waf-generated helper files signature-based instead of always-running.
* Make bob-engine.jar incremental with explicit bob-light.jar and shader plugin jar dependencies.
* Avoid rebuilding engine test setup during the install pass.

### DIFF
--- a/build_tools/waf_dynamo.py
+++ b/build_tools/waf_dynamo.py
@@ -17,7 +17,6 @@ from waflib.Configure import conf
 from waflib import Utils, Build, Options, Task, Logs
 from waflib.TaskGen import extension, feature, after, before, task_gen
 from waflib.Logs import error
-from waflib.Task import RUN_ME
 from BuildUtility import BuildUtility, BuildUtilityException, create_build_utility
 from waf_tests import get_test_harness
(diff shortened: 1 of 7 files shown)


ekharkunov (approved):
👍

PR #12281, opened 2026-04-26
Cleaned up WAF warnings found in `waf -v`

AGulev:
* Removed stale task ordering constraints: `before='c cxx'`, `after='proto_gen_py'`,  `after='apply_java'`, `after='jar_files'`
* Made generated proto/DDF outputs single-owner, with secondary targets consuming generated .cpp nodes: `engine`, `ddf`, `graphics`, `particle`, `rig`, `script`, `gui`.
* Fixed duplicate generated test assets: 
  * GUI embedded Lua / test DDF
  * Script test DDF / test_types.luac
  * Resource `.cont`, `.foo`, `.adc`, `.scriptc`, archive embeds
* Resource build cleanup:
  * `barchive` now accepts already compiled resource nodes
  * `luresources` now builds via `luresources_base`, then copies/updates outputs without duplicate ownership
  * Added shared `embedded_resources` test lib
* Fixed unsafe `ant_glob` usage:
  * `dlib` no longer globs `../dlib/testutil.cpp`
  * `graphics` no longer globs generated files from `build/proto`
* Added explicit stamp outputs for engine Bob test build tasks.
* Removed invalid `glslang` ordering in shaderc tests.


engine/dlib/src/test/wscript
```diff
@@ -5,6 +5,14 @@ import waf_dynamo
 from waf_dynamo import is_platform_private, platform_supports_feature
 from BuildUtility import create_build_utility
 
+def copy_testutil(task):
+    task.outputs[0].parent.mkdir()
+    with open(task.inputs[0].abspath(), 'r') as src:
+        content = src.read()
+    content = content.replace('#include "testutil.h"', '#include <dlib/testutil.h>')
+    with open(task.outputs[0].abspath(), 'w') as dst:
+        dst.write(content)
```
JCash: What's this for?
Looks like a "fix" for include paths?
But that is what a build system is for, to specify include paths in a clear fashion, so it seems something is strange/broken if this is needed?

engine/dlib/src/test/wscript
```diff
@@ -74,10 +82,16 @@ def create_test(bld, name, extra_libs = None, extra_features = None, embed_sourc
     if extra_source is None:
         extra_source = []
 
+    testutil_source = bld.path.parent.find_resource('dlib/testutil.cpp')
+    testutil_copy = bld.path.get_bld().make_node('testutil/%s_testutil.cpp' % (target_name or name))
+    bld(rule = copy_testutil,
+        source = testutil_source,
+        target = testutil_copy)
```
JCash: This also looks very strange.

engine/engine/src/test/wscript
```diff
@@ -30,6 +30,9 @@ def run_bob(task):
 
     args = task.generator.run_args
     output = run.env_command(env, args.split())
+    for out in task.outputs:
+        out.parent.mkdir()
+        out.write('')
```
JCash: What's this?
Also, does work the second time, as mkdir() should fail on a preexisting folder?

engine/engine/src/wscript
```diff
@@ -50,6 +50,11 @@ def build(bld):
     if bld.env['PLATFORM'] in ('wasm-web', 'wasm_pthread-web'):
         engine_defines.append('DM_NO_HTTP_CACHE')
 
+    engine_ddf_proto = bld.path.find_resource('../proto/engine/engine_ddf.proto')
+    engine_release_source = [
+        engine_ddf_proto.change_ext('.cpp') if src == '../proto/engine/engine_ddf.proto' else src
```
JCash: Looks a bit strange to compare with the relative path here, especially since you already have the engine_ddf_proto on the line above.

JCash (changes requested):
some concerns regarding the include paths and source files.

build_tools/waf_dynamo.py
```diff
@@ -1377,6 +1375,15 @@ def embed_file(self):
 
     # Add dependency on generated embed source files to the task gen
     self.source = source_nodes + embed_out_nodes
+    self.embed_tasks = embed_tasks
+
+@feature('embed')
+@after('process_source')
+def order_embed_file(self):
```
AGulev, line 1382: Keep generated embed headers available before compilation.

JCash (approved):

PR #12285, opened 2026-04-27
Add `/eval` endpoint to editor server

vlaaad:
Now, the editor exposes a local, authenticated `/eval` endpoint for evaluating Lua code in the editor Lua runtime.

Simplest example:

```sh
curl \
  "http://localhost:$(cat .internal/editor.port)/eval" \
  -H "Authorization: Bearer $(cat .internal/editor.token)" \
  --data "print('Hello from Lua') return 1, 'two', true"
# Hello from Lua
# => 1
# => two
# => true
```

Because `/eval` can run commands through `editor.execute`, requests must include the token from `.internal/editor.token`. This limits access to clients that also have filesystem access.

```sh
curl \
  "http://localhost:$(cat .internal/editor.port)/eval" \
  -H "Authorization: Bearer $(cat .internal/editor.token)" \
  --data "editor.execute('git', 'status', '--short', { reload_resources = false })"
#  M game.project
```

Security is documented in `openapi.json`, so the only URL your coding agent needs is still `http://localhost:$(cat .internal/editor.port)/openapi.json`. For example:
```md
› use /eval endpoint and make a short report about the available functionality
• Thinking...
• Short Report

  `/eval` evaluates raw Lua in the Defold editor extension runtime.

  Available runtime functionality includes:

  - Standard Lua globals: `assert`, `pcall`, `load`, `require`, `io`, `os`, 
    `math`, `string`, `table`, `coroutine`, etc.
  - Defold editor API via `editor`, including:
      - resource inspection: `editor.resource_attributes`,
        `editor.properties`, `editor.get`
      - resource/project mutation APIs: `editor.transact`, 
        `editor.tx.set`/`add`/`remove`/`reorder`/`reset`/`clear`, `editor.save`
      - filesystem helpers: `editor.create_directory`, `editor.delete_directory`, 
       `editor.create_resources`
      - external command execution: `editor.execute`
      - browser/file helpers: `editor.browse`, `editor.open_external_file`,
        `editor.external_file_attributes`
      - UI construction/dialog APIs under `editor.ui`
      - preferences via `editor.prefs.get`/`set`/`is_set`
  - HTTP APIs via `http`, including `http.request` and `http.server.route`/`response`
  - Utility modules: `json.encode`/`decode`, `zip.pack`/`unpack`, 
    `zlib.deflate`/`inflate`, `localization`.
```

Fix #11995

### DIFF
--- a/editor/src/clj/editor/boot_open_project.clj
+++ b/editor/src/clj/editor/boot_open_project.clj
@@ -31,6 +31,7 @@
             [editor.dialogs :as dialogs]
             [editor.disk :as disk]
             [editor.editor-extensions :as extensions]
+            [editor.editor-extensions.server :as ext.server]
             [editor.engine-profiler :as engine-profiler]
             [editor.git :as git]
             [editor.hot-reload :as hot-reload]
@@ -210,10 +211,12 @@
                                                       localization)
 
           breakpoints-view (breakpoints-view/make-breakpoints-view workspace project open-resource *view-graph* prefs (.lookup root "#breakpoints-container"))
+          token (web-server/make-token)
           server-handler (web-server/make-dynamic-handler
                            (into []
                                  cat
                                  [(web-server/built-in-routes project)
+                                  (ext.server/routes project token)
                                   (engine-profiler/routes)
                                   (console/routes console-view)
                                   (hot-reload/routes workspace)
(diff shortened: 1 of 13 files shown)


JosephFerano (approved):
Works on Linux, the token security seems to be working as well, in between restarts.

PR #12291, opened 2026-04-27
Fix for calculating multiline text height

ivasilyeu:
Legacy text metrics calculation is now using the correct descent value.
It fixes wrongly spaced lines.

### DIFF
--- a/engine/font/src/text_layout_legacy.cpp
+++ b/engine/font/src/text_layout_legacy.cpp
@@ -224,7 +224,7 @@ TextResult TextLayoutLegacyCreate(HFontCollection collection,
     float scale = FontGetScaleFromSize(font, settings->m_Size);
 
     uint32_t ascent = (uint32_t)FontGetAscent(font, 1.0f);
-    uint32_t descent = (uint32_t)FontGetDescent(font, 1.0f); // positive value
+    uint32_t descent = (uint32_t)fabsf(FontGetDescent(font, 1.0f)); // positive value
     uint32_t line_height = ascent + descent;
     float line_height_scaled = line_height * scale;
     float tracking = line_height_scaled * settings->m_Tracking;


ihar-va-MA:
Attaching a small test project demonstrating the problem. Try commenting out the `metrics_height_to_px` invocation and using the height from `resource.get_text_metrics()` directly, and you'll see that the height is calculated incorrectly.
[text_metrics_demo.zip](https://github.com/user-attachments/files/27135831/text_metrics_demo.zip)

engine/font/src/text_layout_legacy.cpp
```diff
@@ -224,7 +224,7 @@ TextResult TextLayoutLegacyCreate(HFontCollection collection,
     float scale = FontGetScaleFromSize(font, settings->m_Size);
 
     uint32_t ascent = (uint32_t)FontGetAscent(font, 1.0f);
-    uint32_t descent = (uint32_t)FontGetDescent(font, 1.0f); // positive value
+    uint32_t descent = (uint32_t)fabsf(FontGetDescent(font, 1.0f)); // positive value
```
britzl, line 227: This means we could also fix this at build time?

ivasilyeu: We can artificially figure out the correct height, by calling resource.get_text_metrics(singleline_test_string), and calculating a "conversion coefficient" by comparing the sum of ascend and descent with returned height, that allows to then calculate proper height of the actual multiline text. This is what I do in the attached test project.

britzl: What I meant by my comment is that maybe the fix could be applied at build time in bob/editor instead of doing it at runtime. But I don't know enough about the problem to fully understand if this is feasible or not.

JCash: To answer Björn's question: This is a runtime loaded .ttf font, and we get the descent once when it is loaded.
And the FontGetDescent() is (nowadays) returning the value as-is, which is easier to reason about.
However, our code that uses this value expects the value to be positive.
We do the similar abs, in res_font.cpp

JCash (approved):
Thanks for this fix.
Next time I'd like to see a unit test around it as well.

PR #12293, opened 2026-04-28
Speedup engine build on CI

AGulev:
Build time now takes around 33m (for a feature branch) + here I speedup Release saving about 1:30m (release runs only on dev/beta/master)

### Technical changes

- Cache Leiningen and Maven dependencies in editor/docs CI jobs, and skip fetching git tags for editor checkouts.
- Split editor tests into a separate Linux CI job.
- Allow editor builds to skip external-dependency tests via CI/build script flags.
- Pack only target-platform editor resources and add uberjar exclusions to reduce bundled cross-platform artifacts.
- Add release-specific Python dependency installation and simplify the release workflow dependency setup.
- Improve editor bundle jar rewrite logging while preserving the rewritten jar layout.


.github/workflows/main-ci.yml
```diff
@@ -186,7 +186,7 @@ jobs:
...
-      { name: 'Cache lein installation for docs', uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae, with: { path: 'editor/build/lein', key: "${{ runner.os }}-${{ matrix.platform || 'default' }}-lein-${{ hashFiles('editor/scripts/lein', 'editor/project.clj', 'editor/profiles.clj') }}", restore-keys: "${{ runner.os }}-${{ matrix.platform || 'default' }}-lein-\n${{ runner.os }}-lein-" } }, # v5.0.5
+      { name: 'Cache lein installation and Maven dependencies for docs', uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae, with: { path: "editor/build/lein\n~/.m2/repository", key: "${{ runner.os }}-${{ matrix.platform || 'default' }}-lein-maven-docs-${{ hashFiles('editor/scripts/lein', 'editor/project.clj', 'editor/profiles.clj') }}", restore-keys: "${{ runner.os }}-${{ matrix.platform || 'default' }}-lein-maven-docs-\n${{ runner.os }}-lein-maven-docs-\n${{ runner.os }}-${{ matrix.platform || 'default' }}-lein-docs-\n${{ runner.os }}-lein-docs-" } }, # v5.0.5
```
AGulev, line 189: better caching of maven deps. Shpuld improve stability as well

.github/workflows/main-ci.yml
```diff
@@ -440,19 +440,18 @@ jobs:
         name: 'Checkout',
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd, # v6.0.2
         with: {
-          ref: '${{env.BUILD_BRANCH}}',
-          fetch-tags: true
+          ref: '${{env.BUILD_BRANCH}}'
```
AGulev, line 443: I didn't find any evidence that current pipiline uses tags but it makes scheckout slower

.github/workflows/main-ci.yml
```diff
@@ -440,19 +440,18 @@ jobs:
...
       { name: 'Install Python', uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405, with: { python-version: 3.13} }, # v6.2.0
       { name: 'Install Java', uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654, with: { java-version: '25', distribution: 'temurin', cache: 'gradle'} }, # v5.2.0
       { name: 'Install dependencies', run: sudo apt-get install -y libgl1-mesa-dev libglw1-mesa-dev freeglut3-dev },
       {
         name: 'Build editor',
         if: (github.event_name != 'repository_dispatch') || ((github.event_name == 'repository_dispatch') && (github.event.client_payload.skip_editor != true)),
-        run: 'ci/ci.sh build-editor --platform=x86_64-linux'
+        run: 'ci/ci.sh build-editor --platform=x86_64-linux --skip-external-dependency-tests'
```
AGulev: all the extensions now tested in a separate job

.github/workflows/main-ci.yml
```diff
@@ -594,14 +624,6 @@ jobs:
...
       { name: 'Install Python', if: *should_release, uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405, with: { python-version: 3.13} }, # v6.2.0
       { name: 'Setup git user', if: *should_release, run: 'git config --global user.email "services@defold.se" && git config --global user.name "Defold Services"' },
-      { name: 'Install dependencies', if: *should_release, run: 'ci/ci.sh install --platform=${{ matrix.platform }}' },
-      {
-        # only needed for some python dependencies (should have a separate python setup function?)
-        name: 'Install ext dependencies',
-        if: *should_release,
-        shell: bash,
-        run: 'ci/ci.sh install_ext'
-      },
```
AGulev, line 604: we install_Ext as part of the next step, don't need it twice. Also I implemented a separate command as this comment suggested. 

I didn't find any evidence we need 

>{ name: 'Install dependencies', if: *should_release, run: 'ci/ci.sh install --platform=${{ matrix.platform }}' },

ci/ci.py
```diff
@@ -382,7 +397,7 @@ def test_bob(channel):
 
 
 def release(channel):
-    args = "python scripts/build.py install_ext release".split()
+    args = "python scripts/build.py install_release_dependencies release".split()
```
AGulev, line 420: new command for a short list of release dependencies to speedup the process

editor/scripts/bundle.py
```diff
@@ -394,20 +395,31 @@ def remove_platform_files_from_archive(platform, jar):
...
+    # a Python-rewritten jar, and keeping that ZIP layout stable avoids changing
+    # macOS signing/notarization inputs while moving most filtering to uberjar.
+    scan_time = time.perf_counter() - start_time
+    if files_to_remove:
+        log("Warning: removing %d platform-specific files from %s after %.3fs scan" %
+            (len(files_to_remove), jar, scan_time))
+    else:
+        log("Rewriting %s after %.3fs scan; no platform-specific files to remove" %
+            (jar, scan_time))
+    rewrite_start_time = time.perf_counter()
```
AGulev, line 422: Now this script mostly for validation. The idea is that we don't include files into uberjar when copy files. It helps to save time on the jar packing and on this script as well.
But, because of the signing process we have to touch archive (otherwise signing doesn't work), see comment

editor/tasks/leiningen/pack.clj
```diff
@@ -26,8 +26,6 @@
 
 (defn dynamo-home [] (get (System/getenv) "DYNAMO_HOME"))
 
-;; this is universal packing for all the platforms, but we clenup it in bundle.py -> remove_platform_files_from_archive()
```
AGulev, line 29: isn't true anymore

editor/project.clj
```diff
@@ -170,9 +170,41 @@
...
+                                                        integration.extension-lua-preprocessor-test
+                                                        integration.extension-rive-test
+                                                        integration.extension-simpledata-test
+                                                        integration.extension-spine-test
+                                                        integration.extension-teal-test
+                                                        integration.extension-texturepacker-test}
+                                                     test-ns-symbol))
+                                         (fn [test-var-metadata & _symbol-args]
+                                           (or (:external-dependency test-var-metadata)
+                                               (:external-dependency (meta (:ns test-var-metadata)))))]
```
vlaaad: Should we only use metadata here? Why also list the tests here?

vlaaad (commented):
I don't understand why this particular distinction — external deps vs no external deps. Why editor extensions test considered external? It mostly tests editor scripts' behaviors. Is it just to split tests "in half" in terms of time to test and run them in parallel?

editor/project.clj
AGulev: true, I forgot to remove it (I made it this way initially, and then replaced with a tag )

vlaaad:
Wait.. do we now skip "external dependency" tests altogether?

AGulev:
these particular tests depends on external dependencies services (extensions and extender). If somthing goes wrong when fetching data or on extender they will fail. It's easier to restart just them now since they in a separate run.

In the future I think such tests should be moved out of the editor and intead call hooks into its repos to call them. So now they all are extracted as separate piece of tests. If you don't like it, I can extract just half of the tests or something like that. But I see sense in this splitting this way

vlaaad:
My opinion on this PR: the caching changes are good. The test separation feels like it introduces too much complexity for a gain that can probably be achieved in a different, but much simpler way. I find the separation confusing, even with your explanation. Why don't we run **all** editor tests in a separate job instead? Semantically, this is simpler, and should also improve perf, since we don't have a single job that both builds the editor and tests it...

AGulev:
Sounds good, I'll run all the test as a separate job, totally makes sense

AGulev:
This is much better 
CleanShot 2026-04-28 at 15 32 00@2x

But I also changed job dependencies, so only tests depends on defold sdk

editor/scripts/bundle.py
vlaaad: I still don't understand it, even with the comment. What is going on with Python, zip, and macOS notarization? Is this about Python replacing symlinks with actual symlinked files? Does that break signing?

AGulev: Read this log: https://github.com/defold/defold/actions/runs/25012368502/job/73254355026 :
```
      "code": null,
      "path": "Defold-arm64-macos.dmg/Defold.app/Contents/Resources/packages/defold-8f32239cff92dbed55e4d90299b5cc5f05e02961.jar/com/sun/jna/darwin-x86-64/libjnidispatch.jnilib",
      "message": "The binary is not signed.",
      "docUrl": "https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087721",
      "architecture": "x86_64"
    },
    {
      "severity": "error",
      "code": null,
      "path": "Defold-arm64-macos.dmg/Defold.app/Contents/Resources/packages/defold-8f32239cff92dbed55e4d90299b5cc5f05e02961.jar/com/sun/jna/darwin-x86-64/libjnidispatch.jnilib",
      "message": "The signature does not include a secure timestamp.",
      "docUrl": "https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087733",
      "architecture": "x86_64"
    },
    {
      "severity": "error",
      "code": null,
      "path": "Defold-arm64-macos.dmg/Defold.app/Contents/Resources/packages/defold-8f32239cff92dbed55e4d90299b5cc5f05e02961.jar/com/sun/jna/darwin/libjnidispatch.jnilib",
      "message": "The binary is not signed.",
      "docUrl": "https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087721",
      "architecture": "i386"
    },
    {
      "severity": "error",
      "code": null,
      "path": "Defold-arm64-macos.dmg/Defold.app/Contents/Resources/packages/defold-8f32239cff92dbed55e4d90299b5cc5f05e02961.jar/com/sun/jna/darwin/libjnidispatch.jnilib",
      "message": "The signature does not include a secure timestamp.",
      "docUrl": "https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087733",
      "architecture": "i386"
    },
    {
      "severity": "error",
      "code": null,
      "path": "Defold-arm64-macos.dmg/Defold.app/Contents/Resources/packages/defold-8f32239cff92dbed55e4d90299b5cc5f05e02961.jar/com/sun/jna/darwin/libjnidispatch.jnilib",
      "message": "The binary is not signed.",
      "docUrl": "https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087721",
      "architecture": "x86_64"
    },
    {
      "severity": "error",
      "code": null,
      "path": "Defold-arm64-macos.dmg/Defold.app/Contents/Resources/packages/defold-8f32239cff92dbed55e4d90299b5cc5f05e02961.jar/com/sun/jna/darwin/libjnidispatch.jnilib",
      "message": "The signature does not include a secure timestamp.",
      "docUrl": "https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087733",
      "architecture": "x86_64"
    }
  ]
}
```

AGulev: Don't fix what's not broken. Signing is hard to test, to debug and to fix. There is no benefit from fixing it except estetical. I suggest to keep it as is

vlaaad (approved):
Shipit!

PR #12298, opened 2026-04-28
Add a LUA reference for constant buffers involved in a render.draw command

Jhonnyg:
Constant buffers created by render scripts are more robust in terms of garbage collection. Previously, a constant buffer could be deallocated between the `render.draw` call and its consumption by the renderer.

Fixes #3557


JCash (approved):

AGulev:
>Issue 3557 constant buffer gc fix


please rename the pr since it will endup in the release notes

PR #12308, opened 2026-04-29
Fix editor Scene View manipulator preview during drag

### LINKED ISSUE #12295: Issues when rotating using gizmo in Editor (1.12.4 beta)
When rotating a collection or a game object containing a 3D model in the Editor the objects also scales up (when dragging) and then down (when releasing).

https://github.com/user-attachments/assets/1ac0c521-33cf-4888-b489-b61a400ae308

**To Reproduce (REQUIRED)**
Open main.collection in the project and rotate 3D objects using gizmo.

Not scaling while rotating.

  - 1.12.4 beta

**Platforms (REQUIRED):**
 - Linux (Ubuntu 24.04)

I have this issue in this project: https://github.com/paweljarosz/space-base-maintenance

matgis:
Fixed visual discrepancies during Scene View manipulator operations caused by faster drag preview code path.

Fixes #12295

### Technical changes
* Scene transforms are now propagates as Pose records that retain separated translation, rotation, and scale instead of trying to derive these from `Matrix4d` objects.
* Renderables are now expected to specify a `:pose` for the scene rather than a `Matrix4d` `:transform`.
* When flattening the scene, we now use the local transform properties from the `:pose`, and we throw an exception in case we encounter a `scene` with a `:transform` to catch anything that hasn't been ported yet.
* Added pose helpers for matrix conversion and assignment to vecmath objects like `Vector3d`.
* Updated `make-outlined-textured-quad-scene` to take a `Pose`. Since it was also used in `extension-texturepacker`, we still accept a `Matrix4d` and convert it to a `Pose` for the time being.
* Moved scene-related schema types from `extension-texturepacker` to `editor.types`.

See counterpart PR in defold/extension-texturepacker#25.

### DIFF
--- a/editor/project.clj
+++ b/editor/project.clj
@@ -322,7 +322,7 @@
                                                     "-Ddefold.extension.simpledata.url=https://github.com/defold/extension-simpledata/archive/refs/tags/v1.2.0.zip"
                                                     "-Ddefold.extension.spine.url=https://github.com/defold/extension-spine/archive/refs/tags/4.4.1.zip"
                                                     "-Ddefold.extension.teal.url=https://github.com/defold/extension-teal/archive/refs/tags/v1.4.zip"
-                                                    "-Ddefold.extension.texturepacker.url=https://github.com/defold/extension-texturepacker/archive/refs/tags/2.6.0.zip"
+                                                    "-Ddefold.extension.texturepacker.url=https://github.com/defold/extension-texturepacker/archive/refs/heads/editor-use-pose.zip"
                                                     "-Ddefold.unpack.path=tmp/unpack"
                                                     "-Ddefold.nrepl=true"
                                                     "-Ddefold.log.dir="
--- a/editor/project.clj
+++ b/editor/project.clj
@@ -317,12 +317,12 @@
                                 :repl-options      {:init-ns user}
                                 :proto-paths       ["test/proto"]
                                 :resource-paths    ["test/resources"]
-                                :jvm-opts          ["-Ddefold.extension.lua-preprocessor.url=https://github.com/defold/extension-lua-preprocessor/archive/refs/tags/1.1.3.zip"
-                                                    "-Ddefold.extension.rive.url=https://github.com/defold/extension-rive/archive/refs/heads/editor-plugin-test-fixes.zip"
+                                :jvm-opts          ["-Ddefold.extension.lua-preprocessor.url=https://github.com/defold/extension-lua-preprocessor/archive/refs/tags/1.2.0.zip"
+                                                    "-Ddefold.extension.rive.url=https://github.com/defold/extension-rive/archive/refs/tags/11.0.4.zip"
                                                     "-Ddefold.extension.simpledata.url=https://github.com/defold/extension-simpledata/archive/refs/tags/v1.2.0.zip"
-                                                    "-Ddefold.extension.spine.url=https://github.com/defold/extension-spine/archive/refs/tags/4.4.1.zip"
+                                                    "-Ddefold.extension.spine.url=https://github.com/defold/extension-spine/archive/refs/tags/4.5.1.zip"
                                                     "-Ddefold.extension.teal.url=https://github.com/defold/extension-teal/archive/refs/tags/v1.4.zip"
-                                                    "-Ddefold.extension.texturepacker.url=https://github.com/defold/extension-texturepacker/archive/refs/heads/editor-use-pose.zip"
+                                                    "-Ddefold.extension.texturepacker.url=https://github.com/defold/extension-texturepacker/archive/refs/tags/2.7.0.zip"
                                                     "-Ddefold.unpack.path=tmp/unpack"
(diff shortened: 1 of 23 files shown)


editor/src/clj/editor/atlas.clj
```diff
@@ -692,11 +691,11 @@
...
-   x        :- types/Int32
-   y        :- types/Int32
-   width    :- types/Int32
-   height   :- types/Int32
-   page     :- types/Int32
+   x        :- types/TInt32
+   y        :- types/TInt32
+   width    :- types/TInt32
+   height   :- types/TInt32
+   page     :- types/TInt32
```
matgis, line 698: Renamed for consistency with other schemas.

editor/test/integration/non_editable_resources_test.clj
```diff
@@ -370,9 +369,9 @@
 
       (doto chair-embedded-sprite
         (tu/prop! :id "chair-embedded-sprite")
-        (tu/prop! :position [1.1 1.2 1.3])
-        (tu/prop! :rotation (math/vecmath->clj (math/euler-z->quat 1.0)))
-        (tu/prop! :scale [1.4 1.5 1.6])
+        (tu/prop! :position (vector-of :float 1.1 1.2 1.3))
+        (tu/prop! :rotation (into (vector-of :float) (math/vecmath->clj (math/euler-z->quat 1.0))))
+        (tu/prop! :scale (vector-of :float 1.4 1.5 1.6))
```
matgis, line 374: Needed to change these from vectors-of-doubles to vectors-of-float, because `Pose` retains the type it was constructed from. These end up in the `scene` value we compare against the `scene` output of the `non-editable` variant, but the `non-editable` `Pose` becomes populated by the values read from the Protobuf messages, where these fields are `float`.

editor/test/integration/test_util.clj
```diff
@@ -122,11 +123,11 @@
...
-  (mapv #(System/getProperty %)
-        ["defold.extension.rive.url"
-         "defold.extension.simpledata.url"
-         "defold.extension.spine.url"
-         "defold.extension.texturepacker.url"]))
+  (mapv local-extensions/inject-jvm-properties
+        ["{{defold.extension.rive.url}}"
+         "{{defold.extension.simpledata.url}}"
+         "{{defold.extension.spine.url}}"
+         "{{defold.extension.texturepacker.url}}"]))
```
matgis, line 130: Unrelated: Respect local-extensions.

editor/project.clj
```diff
@@ -322,7 +322,7 @@
                                                     "-Ddefold.extension.simpledata.url=https://github.com/defold/extension-simpledata/archive/refs/tags/v1.2.0.zip"
                                                     "-Ddefold.extension.spine.url=https://github.com/defold/extension-spine/archive/refs/tags/4.4.1.zip"
                                                     "-Ddefold.extension.teal.url=https://github.com/defold/extension-teal/archive/refs/tags/v1.4.zip"
-                                                    "-Ddefold.extension.texturepacker.url=https://github.com/defold/extension-texturepacker/archive/refs/tags/2.6.0.zip"
+                                                    "-Ddefold.extension.texturepacker.url=https://github.com/defold/extension-texturepacker/archive/refs/heads/editor-use-pose.zip"
```
matgis: Once we've made a new `extension-texturepacker` release, we can update this. None of the other extensions use `Matrix4d` `:transform` values internally - they all inherit the behavior from `SceneNode`.

I've made sure that the previous version of `extension-texturepacker` will continue to work with the new version of editor, providing schema checks are disabled.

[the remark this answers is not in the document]
matgis: This change is temporary - I will update the ref once we've made a new release of `extension-texturepacker`.

editor/src/clj/editor/types.clj
```diff
@@ -338,3 +340,40 @@
...
+   (s/optional-key :render-fn) (s/pred fn?)
+   (s/optional-key :select-batch-key) s/Any
+   (s/optional-key :tags) #{s/Keyword}
+   (s/optional-key :topmost?) s/Bool
+   (s/optional-key :user-data) s/Any
+   s/Keyword s/Any})
+
+;; SDK api
+(def TSceneUpdatable
+  {:initial-state s/Any
```
vlaaad: I see it was copied from https://github.com/defold/extension-texturepacker/pull/25/changes#diff-5a65fbbd31d0325890f5c5c86811e1a49206ab94b108cb222e09b76bc5aea92aL158
Is the intention that this will eventually be used by other scene updateables? Right now, it's fine that `initial-state` is required, but if we use it, e.g., for particlefx, then it will fail because `produce-scene-updatable` doesn't provide `initial-state`.

vlaaad (dismissed):
LGTM! See a comment with a question. 😀

editor/src/clj/editor/types.clj
matgis: Good point! I'll make the field optional to reflect this.

PR #12315, opened 2026-04-30
Images supported by the engine now open directly in the editor

AGulev:
The Editor now opens `jpg`, `jpeg`, and `png` images for preview without using external applications.

Fix https://github.com/defold/defold/issues/11495

### DIFF
--- a/editor/test/editor/image_test.clj
+++ b/editor/test/editor/image_test.clj
@@ -18,9 +18,11 @@
             [clojure.test.check.generators :as gen]
             [clojure.test.check.properties :as prop]
             [clojure.test :refer :all]
+            [editor.gl.texture :as texture]
             [editor.image :refer :all]
             [editor.image-util :refer :all]
             [editor.geom :refer :all]
+            [editor.workspace :as workspace]
             [schema.test])
   (:import [java.awt.image BufferedImage]))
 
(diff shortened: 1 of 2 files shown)


vlaaad (dismissed):
LGTM! Though I'd prefer if @matgis also had a look at the rendering parts.

editor/test/editor/image_test.clj
```diff
@@ -30,3 +32,22 @@
...
     (is (= 192 (.height img)))))
+
+(deftest image-resource-type-opens-in-scene-view
+  (let [registrations (atom [])]
+    (with-redefs [workspace/register-resource-type (fn [_workspace & args]
+                                                     (swap! registrations conj (apply hash-map args))
+                                                     [])]
+      (doall (register-resource-types ::workspace)))
+    (is (= ["jpg" "jpeg" "png"] (:ext (first @registrations))))
+    (is (= [:scene :default] (:view-types (first @registrations))))))
```
matgis: I think we should maybe remove this test. The proper way to test this would be to actually perform the registration with a workspace and request the view type. This test makes too many assumptions around the internal workings and ends up not testing anything particularly interesting.

We typically don't have corresponding tests for this for other resource types, so I think we should just remove it.

matgis (dismissed):
Works great! I don't think we need the `image-resource-type-opens-in-scene-view` test, though.

editor/test/editor/image_test.clj
AGulev: I reverted the test, thank you

AGulev:
I need someone to approve it again, so I don't bypass the rules

vlaaad (commented):
Bypass the rules, YOLO 😈

AGulev:
```
              |
          \   |   /
     .     \  |  /    .
      `-.__|\/_\/|_.-'
    .__  \ /     `./
       `-
      .-'`.   `     -
     '     `     __.'
           _)___(
          //    \\
         | |    | \
         | |    |  \

PR #12323, opened 2026-05-02
Ensure that builds from the editor include the `DM_DEBUG` define

### LINKED ISSUE #8669: `DM_DEBUG` isn't defined when build from the editor
`DM_DEBUG` isn't defined when the engine builds from the editor (`Cmd+B`) and works fine when bundle

**To Reproduce (REQUIRED)**
1. Download test project
2. Build
3. There is no `DM_*:DM_DEBUG` in logs

`DM_DEBUG ` defined

  - 1.7.0

**Platforms (REQUIRED):**
 - Platforms: macOS
 - OS: macOS
 - Device: mac

[dm_debug.zip](https://github.com/defold/defold/files/14571285/dm_debug.zip)

**Logs (OPTIONAL):**
```
INFO:DLIB: Log server started on port 64056
INFO:ENGINE: Target listening with name: Alexeys-MacBook-Pro-2.local - 192.168.1.67 - Darwin
INFO:ENGINE: Engine service started on port 64057
INFO:GRAPHICS: Installed graphics device 'ADAPTER_FAMILY_OPENGL'
INFO:ENGINE: Defold Engine 1.7.0 (bf4dc66)
INFO:DLIB: Initialized Remotery (ws://127.0.0.1:17815/rmt)
INFO:ENGINE: Loading data from: build/default
ERROR:MYEXTENSION: Check for DM_*:
INFO:MYEXTENSION: Registered myextension Extension
INFO:ENGINE: Initialised sound device 'default'
INFO:DLIB: SSDP: Started on address 192.168.1.67
INFO:DLIB: SSDP: Done on address 192.168.1.67
```

**Workaround (OPTIONAL):**
Bundle using `bob.jar`

AGulev:
Fixed an issue where the `DM_DEBUG` define did not work in debug builds from the editor.

Fix https://github.com/defold/defold/issues/8669

### DIFF
--- a/editor/src/clj/editor/engine/native_extensions.clj
+++ b/editor/src/clj/editor/engine/native_extensions.clj
@@ -26,6 +26,7 @@
             [editor.shared-editor-settings :as shared-editor-settings]
             [editor.system :as system]
             [editor.workspace :as workspace]
+            [editor.yaml :as yaml]
             [util.coll :as coll])
   (:import [com.defold.extender.client ExtenderClient ExtenderClientCache ExtenderResource]
            [com.dynamo.bob Platform]
@@ -161,18 +162,9 @@
 
 (def ^:private app-manifest-upload-path "_app/app.manifest")
 
-(defn- app-manifest-context-prefix ^bytes []
-  (let [line-separator (System/lineSeparator)]
-    (.getBytes (str "context:" line-separator
-                    "    baseVariant: debug" line-separator
-                    "    withSymbols: true" line-separator)
-               StandardCharsets/UTF_8)))
-
-(defn- concat-byte-arrays ^bytes [^bytes a ^bytes b]
-  (let [result (byte-array (+ (alength a) (alength b)))]
-    (System/arraycopy a 0 result 0 (alength a))
-    (System/arraycopy b 0 result (alength a) (alength b))
-    result))
+(def ^:private editor-build-app-manifest-context
+  {"baseVariant" "debug"
+   "withSymbols" true})
 
 (defn- resource-node-content ^bytes [resource-node evaluation-context]
   (if-let [content (some-> (g/node-value resource-node :save-data evaluation-context)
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/engine/native_extensions.clj
```diff
@@ -159,6 +159,44 @@
...
+(defn- resource-node-last-modified [resource-node evaluation-context]
+  (.lastModified (io/file (g/node-value resource-node :resource evaluation-context))))
+
+(defn- make-app-manifest-resource [app-manifest-resource-node evaluation-context]
+  (reify ExtenderResource
+    (getPath [_] app-manifest-upload-path)
+    (getContent [_]
+      (let [context-prefix (app-manifest-context-prefix)]
+        (if app-manifest-resource-node
+          (concat-byte-arrays context-prefix (resource-node-content app-manifest-resource-node evaluation-context))
```
vlaaad: I don't think this is reliable. This is a valid app manifest YAML that we support (JSON syntax):
```yml
{"platforms": {"wasm-web": {"context": {}}}}
```
This is not a valid app manifest YAML:
```yml
context:
  baseVariant: debug
  withSymbols: true
{"platforms": {"wasm-web": {"context": {}}}}
```

We can't simply concatenate strings here.

vlaaad (changes requested):

editor/src/clj/editor/engine/native_extensions.clj
AGulev: ah, ok. I added it into test as well

I rewrote it so parse both, merge then convert back into yaml

editor/src/clj/editor/engine/native_extensions.clj
```diff
@@ -159,6 +160,45 @@
...
+
+(defn- app-manifest-content ^bytes [app-manifest-resource-node evaluation-context]
+  (let [manifest (if app-manifest-resource-node
+                   (or (yaml/load (String. (resource-node-content app-manifest-resource-node evaluation-context)
+                                           StandardCharsets/UTF_8))
+                       {})
+                   {})
+        context (merge (get manifest "context")
+                       editor-build-app-manifest-context)]
+    (let [^String content (yaml/dump (assoc manifest "context" context)
```
vlaaad: This is not enough either. Unfortunately, since the app manifest is currently fully editable by the user, we should be ready for ANY text, including, e.g., `true` or `42`. While this content is nonsensical and will (probably?) lead to build failures, they should be validation errors, not editor exceptions. `app-manifest` ns has a special macro to work with this exact situation — `update-in-fixing`. We should use it here, e.g. something like:
```clj
(update-in-fixing manifest map? {} "context" map? {} #(assoc % context))
```
Or maybe you want to use `get-in-guarded` to avoid updating invalid app manifests...



Style: let nesting, use single let instead.

vlaaad (changes requested):
Sorry for this back and forth, but I only remembered about guarded lookups/updates for the app manifest after you re-requested the review.

editor/src/clj/editor/engine/native_extensions.clj
```diff
@@ -159,6 +161,58 @@
...
+  (.lastModified (io/file (g/node-value resource-node :resource evaluation-context))))
+
+(defn- app-manifest-with-editor-build-context [manifest]
+  (when (and (map? manifest)
+             (or (not (contains? manifest "context"))
+                 (map? (get manifest "context"))))
+    (app-manifest/update-in-fixing
+      manifest map? {}
+      "context" map? {}
+      #(merge % editor-build-app-manifest-context))))
```
vlaaad: Now that you verified that both manifest and its "context" key are maps, you don't really need update-in-fixing...

vlaaad (dismissed):
LGTM, see a nitpicking comment if you want to tidy up the code, but it's optional.

vlaaad (approved):
LGTM!

PR #12324, opened 2026-05-02
Ensure there is no menu bar in the Search and About dialogs

AGulev:
Fixes macOS dialog windows showing an in-window menu bar when the main menu contains grid-backed items.

Fix https://github.com/defold/defold/issues/12156

### Technical changes

System menu bars now avoid `CustomMenuItem` fallback rendering, while context menus keep the existing grid layout. Added a regression test for grid-backed commands in system menu bars.

Joe, I have a feeling I'm fixing a symptom here, not the root cause, since I don't fully understand why this was implemented based on having or not having a grid. So I’ll leave it for you to decide if this approach is okay, or if you have a better idea. I know you mentioned it's a macOS-only issue, and without access to a Mac it might be hard to fix, so I'm suggesting this PR as a possible fix, or at least as additional information about what’s wrong (with this fix, there are no menus in the Search and About windows - so technically it is a fix). But if you think there is a better way to fix it, feel free to close this one and open a new one.

### DIFF
--- a/editor/src/clj/editor/ui.clj
+++ b/editor/src/clj/editor/ui.clj
@@ -1639,38 +1639,45 @@
 
 (declare make-menu-items)
 
-(defn- make-menu-item [^Scene scene item command-contexts keymap localization evaluation-context]
-  (let [{:keys [id icon style children label command user-data check on-submenu-open]} item]
-    (cond
-      (= label :separator)
-      (SeparatorMenuItem.)
+(defn- make-menu-item
+  ([^Scene scene item command-contexts keymap localization evaluation-context]
+   (make-menu-item scene item command-contexts keymap localization evaluation-context true))
+  ([^Scene scene item command-contexts keymap localization evaluation-context allow-custom-menu-items?]
+   (let [{:keys [id icon style children label command user-data check on-submenu-open]} item]
+     (cond
+       (= label :separator)
+       (SeparatorMenuItem.)
 
-      children
-      (let [items (make-menu-items scene children command-contexts keymap localization evaluation-context)]
-        (make-submenu id label localization icon style items on-submenu-open))
+       children
+       (let [items (make-menu-items scene children command-contexts keymap localization evaluation-context allow-custom-menu-items?)]
+         (make-submenu id label localization icon style items on-submenu-open))
 
-      :else
-      (when-let [handler-ctx (handler/active command command-contexts user-data evaluation-context)]
-        ;; NOTE: This label is *not* updated on every menu refresh. Can't do "Show X" <-> "Hide X".
-        (let [label (or (handler/label handler-ctx evaluation-context) label)
-              enabled? (handler/enabled? handler-ctx evaluation-context)
-              key-combo (first (keymap/shortcuts keymap command))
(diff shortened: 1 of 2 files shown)


JosephFerano:
Thank you for looking into this. I have a mac mini actually, so I'll try and replicate there and go over this fix, try to make sense of it.

JosephFerano:
Ok, I did some investigating. Your gut feeling is correct. This is a symptom of another problem. Here is a screenshot from 1.12.1, but this branch has the same behavior.

Screenshot 2026-05-08 at 5 37 43 PM

The real issue is that when creating a new window, we're hitting a mac-os specific codepath that tries to create menu items again, using the mac native system menu.

```clojure

(defn- show-dialog-stage [^Stage stage localization show-fn]
  (.setOnShown stage (event-handler _ (slog/smoke-log "show-dialog")))
  (user-data! (.getScene stage) :localization localization)
  (if (and (os/is-mac-os?)
           (= (.getOwner stage) (main-stage)))
    (let [scene (.getScene stage)
          root-pane ^Pane (.getRoot scene)
          menu-bar (doto (MenuBar.)
                     (.setUseSystemMenuBar true))
          refresh-timer (->timer 3 "refresh-dialog-ui" (fn [_ _ _] (refresh scene)))]
      (.. root-pane getChildren (add menu-bar))
      (register-menubar scene menu-bar (main-menu-id))
      (refresh scene)
      (timer-start! refresh-timer)
      (timer-stop-on-closed! stage refresh-timer)
      (show-fn stage))
    (show-fn stage)))
```

After digging through the commit history, the original intent behind the code was made clear. Here's the commit;

```
7139fb53da DEFEDIT-732 Pasting in search dialog also pastes in text editor

Dialogs get their own copy of the main menu with the
system-menu-workaround thingy, a timer for updates etc just like the
main window. We use the contexts of the dialog + global contexts of
the parent scenes for determining what menu items to show, and the
contexts of only the dialog for updating the menu item state (this way
only relevant items will be enabled).
```

This makes sense, as it's trying to reset the native system menu to disable all items that are not relevant to the current popup window, which is good UX. However, the "File > New..." that has the new `CustomMenuItem`s is part of the "Search in Files" and "About" window's context, and it shouldn't be. So the "proper" fix would be to filter the window's context, so only the relevant menu items are shown, and "New..." should definitely be excluded. ~~Moreover, this is a mac only fix, but it doesn't have to be and I would argue that both Windows and Linux should have the same behavior of disabling any menu items that are not relevant to a new window's context, excluding the `setUseSystemMenuBar` part.~~

Edit: I explored the behaviors of all 3 platforms; the difference is in how the system menu keeps focus on the dialog window, whereas on Win/Linux because there's no system menu, in order to interact with the top bar menu, you lose focus on the dialog window so there's no need to do the same on these platforms. So it makes sense to keep it as a macos only fix.

In regards to this PR, we could merge this in as a temporary fix, but if there's no rush, we can hold off and I'll start a new PR to get the proper fix in. If we were to merge this immediately though, the only feedback is to remove the `?` from `allow-custom-menu-items?` as the style of the editor project is to reserve question marks for predicate functions and avoid using them for boolean values.

Let me know what your thoughts are.

PR #12327, opened 2026-05-02
Ensure `UpdateTransforms` runs only after actual transform changes

AGulev:
Improve game object transform dirty tracking.

Previously script/message updates conservatively marked collection transforms dirty every frame. This changes transform invalidation to happen only when game object transform state actually changes, including direct setters, property writes, animations, parenting changes, bone transforms, and delete/reparent paths.

### Technical changes

1. I'm testing inexample which creates 65530 objects splited into trees by 5 objects.
2. Update sends a message
3. This message just prints in `on_message`
4. As result this creates up to 4 `UpdateTransforms` which take about 6.5ms on M1 Max
CleanShot 2026-05-02 at 22 57 40@2x

After optimisation the same exact example shows only one `UpdateTransforms` call which takes  1.5ms
CleanShot 2026-05-02 at 23 01 53@2x

Example: 
[PerfTestForTransforms.zip](https://github.com/user-attachments/files/27308472/PerfTestForTransforms.zip)

### DIFF
--- a/engine/gameobject/src/gameobject/comp_anim.cpp
+++ b/engine/gameobject/src/gameobject/comp_anim.cpp
@@ -21,6 +21,7 @@
 #include <script/script.h>
 
 #include "component.h"
+#include "gameobject_private.h"
 #include "gameobject_script.h"
 #include "gameobject_props.h"
 #include "gameobject_props_lua.h"
@@ -66,6 +67,7 @@ namespace dmGameObject
         uint16_t            m_Composite : 1;
         uint16_t            m_Backwards : 1;
         uint16_t            m_FirstUpdate : 1;
+        uint16_t            m_TransformsUpdated : 1;
     };
 
     struct AnimWorld
@@ -184,7 +186,7 @@ namespace dmGameObject
         AnimWorld* world = (AnimWorld*)params.m_World;
         world->m_InUpdate = 1;
         uint32_t size = world->m_Animations.Size();
-        uint32_t orig_size = size;
+        bool transforms_updated = false;
 
         DM_PROPERTY_ADD_U32(rmtp_ComponentsAnim, size);
 
@@ -322,6 +324,11 @@ namespace dmGameObject
                     AddPropertyOptionsIndex(&property_opt, 0);
                     SetProperty(anim.m_Instance, anim.m_ComponentId, anim.m_PropertyId, property_opt, PropertyVar(v));
                 }
+                if (anim.m_TransformsUpdated)
+                {
+                    MarkTransformDirty(anim.m_Instance);
+                    transforms_updated = true;
+                }
             }
             if (completed)
             {
@@ -386,7 +393,7 @@ namespace dmGameObject
             }
         }
         world->m_InUpdate = 0;
-        update_result.m_TransformsUpdated = orig_size != 0;
+        update_result.m_TransformsUpdated = transforms_updated;
         return result;
     }
 
@@ -480,6 +487,7 @@ namespace dmGameObject
         animation.m_Next = INVALID_INDEX;
         animation.m_Playing = 1;
         animation.m_Composite = composite ? 1 : 0;
+        animation.m_TransformsUpdated = IsGameObjectTransformProperty(component_id, property_id) ? 1 : 0;
         if (animation.m_Playback == PLAYBACK_ONCE_BACKWARD || animation.m_Playback == PLAYBACK_LOOP_BACKWARD)
             animation.m_Backwards = 1;
         animation.m_FirstUpdate = 1;
--- a/engine/gameobject/src/gameobject/comp_anim.cpp
+++ b/engine/gameobject/src/gameobject/comp_anim.cpp
@@ -324,11 +324,7 @@ namespace dmGameObject
                     AddPropertyOptionsIndex(&property_opt, 0);
                     SetProperty(anim.m_Instance, anim.m_ComponentId, anim.m_PropertyId, property_opt, PropertyVar(v));
                 }
-                if (anim.m_TransformsUpdated)
-                {
-                    MarkTransformDirty(anim.m_Instance);
-                    transforms_updated = true;
-                }
+                transforms_updated |= anim.m_TransformsUpdated;
             }
             if (completed)
             {
@@ -487,7 +483,7 @@ namespace dmGameObject
         animation.m_Next = INVALID_INDEX;
         animation.m_Playing = 1;
         animation.m_Composite = composite ? 1 : 0;
-        animation.m_TransformsUpdated = IsGameObjectTransformProperty(component_id, property_id) ? 1 : 0;
+        animation.m_TransformsUpdated = component_id == 0 && IsGameObjectTransformProperty(property_id) ? 1 : 0;
         if (animation.m_Playback == PLAYBACK_ONCE_BACKWARD || animation.m_Playback == PLAYBACK_LOOP_BACKWARD)
             animation.m_Backwards = 1;
         animation.m_FirstUpdate = 1;
--- a/engine/gameobject/src/gameobject/comp_anim.cpp
+++ b/engine/gameobject/src/gameobject/comp_anim.cpp
@@ -67,7 +67,7 @@ namespace dmGameObject
         uint16_t            m_Composite : 1;
         uint16_t            m_Backwards : 1;
         uint16_t            m_FirstUpdate : 1;
-        uint16_t            m_TransformsUpdated : 1;
+        uint16_t            m_IsGameObjectTransformProperty : 1;
     };
 
     struct AnimWorld
@@ -324,7 +324,7 @@ namespace dmGameObject
                     AddPropertyOptionsIndex(&property_opt, 0);
                     SetProperty(anim.m_Instance, anim.m_ComponentId, anim.m_PropertyId, property_opt, PropertyVar(v));
                 }
-                transforms_updated |= anim.m_TransformsUpdated;
+                transforms_updated |= anim.m_IsGameObjectTransformProperty;
             }
             if (completed)
             {
@@ -483,7 +483,7 @@ namespace dmGameObject
         animation.m_Next = INVALID_INDEX;
         animation.m_Playing = 1;
         animation.m_Composite = composite ? 1 : 0;
-        animation.m_TransformsUpdated = component_id == 0 && IsGameObjectTransformProperty(property_id) ? 1 : 0;
+        animation.m_IsGameObjectTransformProperty = component_id == 0 && IsGameObjectTransformProperty(property_id) ? 1 : 0;
         if (animation.m_Playback == PLAYBACK_ONCE_BACKWARD || animation.m_Playback == PLAYBACK_LOOP_BACKWARD)
             animation.m_Backwards = 1;
         animation.m_FirstUpdate = 1;
--- a/engine/gameobject/src/gameobject/comp_script.cpp
+++ b/engine/gameobject/src/gameobject/comp_script.cpp
@@ -224,7 +224,7 @@ namespace dmGameObject
     }
 
 
-    static UpdateResult CompScriptUpdateInternal(const ComponentsUpdateParams& params, ScriptFunction function, ComponentsUpdateResult& update_result)
+    static UpdateResult CompScriptUpdateInternal(const ComponentsUpdateParams& params, ScriptFunction function, ComponentsUpdateResult&)
     {
         lua_State* L = GetLuaState(params.m_Context);
         int top = lua_gettop(L);
@@ -246,9 +246,6 @@ namespace dmGameObject
                 }
             }
         }
-
-        // TODO: Find out if the scripts actually sent any transform events
-        update_result.m_TransformsUpdated = true;
 
         assert(top == lua_gettop(L));
         return result;
--- a/engine/gameobject/src/gameobject/gameobject.cpp
+++ b/engine/gameobject/src/gameobject/gameobject.cpp
@@ -84,20 +84,8 @@ namespace dmGameObject
     PROP_VECTOR3(EULER, euler);
     PROP_VECTOR3(SCALE, scale);
 
-    void MarkTransformDirty(HInstance instance)
+    bool IsGameObjectTransformProperty(dmhash_t property_id)
     {
-        instance->m_Collection->m_DirtyTransforms = true;
-    }
-
-    bool IsGameObjectTransformProperty(dmhash_t component_id, dmhash_t property_id)
-    {
-        // Component id 0 targets the game object itself. Non-zero ids target component
-        // properties, which may have overlapping names such as "scale".
-        if (component_id != 0)
-        {
-            return false;
-        }
-
         return property_id == PROP_POSITION ||
                property_id == PROP_POSITION_X ||
                property_id == PROP_POSITION_Y ||
@@ -2285,26 +2273,25 @@ namespace dmGameObject
         return instance->m_Bone;
     }
 
-    static uint32_t DoSetBoneTransforms(HCollection hcollection, dmTransform::Transform* component_transform, uint16_t first_index, dmTransform::Transform* transforms, uint32_t transform_count)
+    static uint32_t DoSetBoneTransforms(Collection* collection, dmTransform::Transform* component_transform, uint16_t first_index, dmTransform::Transform* transforms, uint32_t transform_count)
     {
         if (transform_count == 0)
             return 0;
         uint16_t current_index = first_index;
         uint32_t count = 0;
-        Collection* collection = hcollection->m_Collection;
         while (current_index != INVALID_INSTANCE_INDEX)
         {
             HInstance instance = collection->m_Instances[current_index];
             if (instance->m_Bone)
             {
                 instance->m_Transform = transforms[count++];
-                if (component_transform && count == 1) {
+                if (component_transform && count == 1)
+                {
                     instance->m_Transform = dmTransform::Mul(*component_transform, instance->m_Transform);
                 }
-                MarkTransformDirty(instance);
                 if (count < transform_count)
                 {
-                    count += DoSetBoneTransforms(hcollection, 0x0, instance->m_FirstChildIndex, &transforms[count], transform_count - count);
+                    count += DoSetBoneTransforms(collection, 0x0, instance->m_FirstChildIndex, &transforms[count], transform_count - count);
                 }
                 if (transform_count == count)
                 {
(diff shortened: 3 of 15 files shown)


engine/gameobject/src/gameobject/comp_anim.cpp
```diff
@@ -480,6 +487,7 @@ namespace dmGameObject
         animation.m_Next = INVALID_INDEX;
         animation.m_Playing = 1;
         animation.m_Composite = composite ? 1 : 0;
+        animation.m_TransformsUpdated = IsGameObjectTransformProperty(component_id, property_id) ? 1 : 0;
```
AGulev: conditinally check if property is a GameObject property, and return `update_result.m_TransformsUpdated` based on it

engine/gameobject/src/gameobject/comp_anim.cpp
```diff
@@ -184,7 +186,7 @@ namespace dmGameObject
         AnimWorld* world = (AnimWorld*)params.m_World;
         world->m_InUpdate = 1;
         uint32_t size = world->m_Animations.Size();
-        uint32_t orig_size = size;
```
AGulev, line 187: unused here anymore

engine/gameobject/src/gameobject/comp_script.cpp
```diff
@@ -247,9 +247,6 @@ namespace dmGameObject
             }
         }
 
-        // TODO: Find out if the scripts actually sent any transform events
-        update_result.m_TransformsUpdated = true;
```
AGulev, line 251: it's done in `gameobject.cpp‎` covering all the possible transform changes

engine/gameobject/src/gameobject/test/hierarchy/test_gameobject_hierarchy.cpp
```diff
@@ -87,6 +87,44 @@ class HierarchyTest : public jc_test_base_class
     dmHashTable64<void*> m_Contexts;
 };
 
+static void SetCachedWorldPosition(dmGameObject::HCollection collection, dmGameObject::HInstance instance, const Point3& position)
+{
+    Matrix4 world = Matrix4::identity();
+    world.setCol3(Vector4(position.getX(), position.getY(), position.getZ(), 1.0f));
```
JCash: While tis is a test, it's good to keep it performant for educational reasons.

`world = Matrix4::translation(Vector3(position));`

engine/gameobject/src/gameobject/test/hierarchy/test_gameobject_hierarchy.cpp
```diff
@@ -87,6 +87,44 @@ class HierarchyTest : public jc_test_base_class
...
+    collection->m_Collection->m_WorldTransforms[instance->m_Index] = world;
```
JCash: So, the caching of the world position overwrites any other transform data?
I suggest simply storing the position, as currently, it's ambiguous and untangling it after-the-fact will be much harder.

engine/gameobject/src/gameobject/test/script/test_gameobject_script.cpp
```diff
@@ -228,6 +228,189 @@ static void CreateScriptFile(const char* file_name, const char* contents)
...
+    CreateGoFile(go_file_name, script_resource_name);
+
+    dmGameObject::HInstance go = dmGameObject::New(m_Collection, go_resource_name);
+    ASSERT_NE((dmGameObject::HInstance) 0, go);
+
+    ASSERT_TRUE(dmGameObject::Init(m_Collection));
+    ASSERT_FALSE(m_Collection->m_Collection->m_DirtyTransforms);
+
+    Matrix4 sentinel = Matrix4::identity();
+    sentinel.setCol3(Vector4(42, 43, 44, 1));
```
JCash: Same here

engine/gameobject/src/gameobject/test/script/test_gameobject_script.cpp
```diff
@@ -228,6 +228,189 @@ static void CreateScriptFile(const char* file_name, const char* contents)
...
+               "end\n");
+    CreateGoFile(go_file_name, script_resource_name);
+
+    dmGameObject::HInstance go = dmGameObject::New(m_Collection, go_resource_name);
+    ASSERT_NE((dmGameObject::HInstance) 0, go);
+
+    ASSERT_TRUE(dmGameObject::Init(m_Collection));
+
+    Matrix4 sentinel = Matrix4::identity();
+    sentinel.setCol3(Vector4(42, 43, 44, 1));
```
JCash: same here

engine/gameobject/src/gameobject/test/script/test_gameobject_script.cpp
```diff
@@ -228,6 +228,189 @@ static void CreateScriptFile(const char* file_name, const char* contents)
...
+
+    const char* go_resource_name = "/__no_transform_update__.goc";
+    char go_file_name[512];
+    dmTestUtil::MakeHostPathf(go_file_name, sizeof(go_file_name), "%s%s", m_Path, go_resource_name);
+
+    CreateScriptFile(script_file_name,
+               "function update(self)\n"
+               "    self.counter = (self.counter or 0) + 1\n"
+               "end\n");
+    CreateGoFile(go_file_name, script_resource_name);
```
JCash: Just double check with the arm64-android tests that this type of creation outside of the build system works.
(arm64-android is easier for you to test than our consoles)

engine/gameobject/src/gameobject/test/script/test_gameobject_script.cpp
```diff
@@ -228,6 +228,189 @@ static void CreateScriptFile(const char* file_name, const char* contents)
...
+
+    dmGameObject::HInstance go = dmGameObject::New(m_Collection, go_resource_name);
+    ASSERT_NE((dmGameObject::HInstance) 0, go);
+    ASSERT_EQ(dmGameObject::RESULT_OK, dmGameObject::SetIdentifier(m_Collection, go, "message_go"));
+
+    ASSERT_TRUE(dmGameObject::Init(m_Collection));
+    ASSERT_FALSE(m_Collection->m_Collection->m_DirtyTransforms);
+
+    Matrix4 sentinel = Matrix4::identity();
+    sentinel.setCol3(Vector4(42, 43, 44, 1));
```
JCash: Same here

engine/gameobject/src/gameobject/test/hierarchy/test_gameobject_hierarchy.cpp
```diff
@@ -87,6 +87,44 @@ class HierarchyTest : public jc_test_base_class
     dmHashTable64<void*> m_Contexts;
 };
 
+static void SetCachedWorldPosition(dmGameObject::HCollection collection, dmGameObject::HInstance instance, const Point3& position)
```
JCash: I recommend renaming it to "SetCachedWorldTransform" as it overwrites the full transform.

engine/gameobject/src/gameobject/test/script/test_gameobject_script.cpp
```diff
@@ -228,6 +228,189 @@ static void CreateScriptFile(const char* file_name, const char* contents)
...
+
+    const char* go_resource_name = "/__transform_message__.goc";
+    char go_file_name[512];
+    dmTestUtil::MakeHostPathf(go_file_name, sizeof(go_file_name), "%s%s", m_Path, go_resource_name);
+
+    CreateScriptFile(script_file_name,
+               "function on_message(self, message_id, message, sender)\n"
+               "    go.set_position(vmath.vector3(1, 2, 3))\n"
+               "end\n");
+    CreateGoFile(go_file_name, script_resource_name);
```
JCash: There's a lot of boiler plate for these test setups, which distracts from the actual test..
I suggest putting the path management inside of the function, and then return the path:
```c++
char script_path[512];
CreateScriptFile(script_path, m_Path, "/__transform_message__.scriptc", ...payload...);
...
dmSys::Unlink(script_path);
```

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -84,6 +84,40 @@ namespace dmGameObject
     PROP_VECTOR3(EULER, euler);
     PROP_VECTOR3(SCALE, scale);
 
+    void MarkTransformDirty(HInstance instance)
```
JCash: -> "MarkCollectionTransformsDirty"

The function sets the collection as dirty, and it's important to note that, as the only argument is an instance, and so the programmer will know that it has to "do some work" to actually set the boolean.

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -2262,6 +2301,7 @@ namespace dmGameObject
                 if (component_transform && count == 1) {
                     instance->m_Transform = dmTransform::Mul(*component_transform, instance->m_Transform);
                 }
+                MarkTransformDirty(instance);
```
JCash: Don't dereference the `instance` pointer, when you already have the `collection` pointer.
Also, no need dereference the pointer in the loop itself?
Store as a boolean and set it after the loop!

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -2357,18 +2397,24 @@ namespace dmGameObject
...
-                {
-                    Matrix4& world = collection->m_WorldTransforms[instance->m_Index];
-                    world = parent_t * dmTransform::ToMatrix4(instance->m_Transform);
-                }
-                else
-                {
-                    instance->m_Transform = dmTransform::ToTransform(inverse(parent_t) * collection->m_WorldTransforms[instance->m_Index]);
-                }
+                Matrix4 world_transform = parent_t * dmTransform::ToMatrix4(instance->m_Transform);
+                dmTransform::Transform local_transform = dmTransform::ToTransform(inverse(parent_t) * collection->m_WorldTransforms[instance->m_Index]);
```
JCash: While it's not the most time critical thing, we _did_ have the calculations separated by an if-statement.
Previously, we only calculated one.

Now, we always calculate two transforms, even if the new `if (result == dmGameObject::RESULT_OK && old_parent != instance->m_Parent)` isn't passed

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -2556,7 +2601,7 @@ namespace dmGameObject
         {
             uint16_t index = root_level[i];
             Instance* instance = collection->m_Instances[index];
-            CheckEuler(instance);
+            CheckEuler(instance, false);
```
JCash: This needs explaining in the PR.
Why is a change of transform not _always_ a valid "dirty" of the transforms?

JCash: Also, I'd restructure the code to not do the instance->collection dereference.
```c++
bool dirty_transforms = false;
for (instance in ...)
{
    dirty_transforms |= CheckEuler(instance);
}

collection->m_DirtyTransforms = dirty_transforms
```

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -2683,7 +2728,10 @@ namespace dmGameObject
 
                 // Mark the collections transforms as dirty if this component has updated
                 // them in its update function.
-                collection->m_DirtyTransforms |= update_result.m_TransformsUpdated;
+                if (update_result.m_TransformsUpdated)
+                {
+                    collection->m_DirtyTransforms = true;
+                }
```
JCash, line 2721: This is an odd change, which I normally would fail.
I suspect it changed it because of the types being mixed (bitfield vs boolean)
However, being curious I checked with godbolt and the new code was also shorter, so it seems like a good change:
https://godbolt.org/z/Po6e6qEv5

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -3316,6 +3369,7 @@ namespace dmGameObject
             }
         }
 
+        MarkTransformDirty(child);
```
JCash: same, don't dereference instance to set a collection property

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -3372,10 +3426,19 @@ namespace dmGameObject
...
 
-    static void UpdateEulerToRotation(HInstance instance)
+    static void UpdateEulerToRotation(HInstance instance, bool mark_dirty)
     {
         instance->m_PrevEulerRotation = instance->m_EulerRotation;
         instance->m_Transform.SetRotation(dmVMath::EulerToQuat(instance->m_EulerRotation));
+        if (mark_dirty)
+        {
+            MarkTransformDirty(instance);
+        }
```
JCash: Keep separation clear.
Setting the collection property should be handled outside of the instance property function.

engine/gameobject/src/gameobject/comp_anim.cpp
```diff
@@ -322,6 +324,11 @@ namespace dmGameObject
                     AddPropertyOptionsIndex(&property_opt, 0);
                     SetProperty(anim.m_Instance, anim.m_ComponentId, anim.m_PropertyId, property_opt, PropertyVar(v));
                 }
+                if (anim.m_TransformsUpdated)
+                {
+                    MarkTransformDirty(anim.m_Instance);
+                    transforms_updated = true;
+                }
```
JCash: This is one of our hot paths.
Store the `transforms_updated |= anim.m_TransformsUpdated` and use the dirty flag at the end of the function (like you do now).

The call to MarkTransformDirty() is trashing the cache, by dereferencing both the collection and the instance. And it's not even needed as the function returns this same boolean for this exact use case.

JCash (changes requested):
Good idea.
Needs changes though.

AGulev:
Great review, thank you! I missed a few things when I initially tried to implement a per-object dirty flag, which turned out not to be the best approach.
Now I made a full cleanup removeing all the leftlovers and your review helped me a lot with that, thank you!

engine/gameobject/src/gameobject/test/script/test_gameobject_script.cpp
AGulev: I replaced it with a regular resource system approach

engine/gameobject/src/gameobject/comp_anim.cpp
```diff
@@ -480,6 +483,7 @@ namespace dmGameObject
         animation.m_Next = INVALID_INDEX;
         animation.m_Playing = 1;
         animation.m_Composite = composite ? 1 : 0;
+        animation.m_TransformsUpdated = component_id == 0 && IsGameObjectTransformProperty(property_id) ? 1 : 0;
```
JCash: 👍

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -2278,7 +2305,13 @@ namespace dmGameObject
...
     uint32_t SetBoneTransforms(HInstance instance, dmTransform::Transform& component_transform, dmTransform::Transform* transforms, uint32_t transform_count)
     {
-        return DoSetBoneTransforms(instance->m_Collection->m_HCollection, &component_transform, instance->m_Index, transforms, transform_count);
+        Collection* collection = instance->m_Collection;
+        uint32_t count = DoSetBoneTransforms(collection, &component_transform, instance->m_Index, transforms, transform_count);
+        if (count > 0)
+        {
+            collection->m_DirtyTransforms = true;
+        }
+        return count;
```
JCash: I prefer using the "boolean" directly 8also, it's not technically a bolean:
```
collection->m_DirtyTransforms |= count > 0 ? 1 : 0;
````

File: engine/gameobject/src/gameobject/gameobject.cpp [hunk 22]
```diff
@@ -2246,25 +2273,25 @@ namespace dmGameObject
         return instance->m_Bone;
     }
 
-    static uint32_t DoSetBoneTransforms(HCollection hcollection, dmTransform::Transform* component_transform, uint16_t first_index, dmTransform::Transform* transforms, uint32_t transform_count)
+    static uint32_t DoSetBoneTransforms(Collection* collection, dmTransform::Transform* component_transform, uint16_t first_index, dmTransform::Transform* transforms, uint32_t transform_count)
```
JCash, line 2276: This change is redundant.

engine/gameobject/src/gameobject/gameobject.cpp
```diff
@@ -2350,24 +2383,27 @@ namespace dmGameObject
                         dmLogWarning("Could not find parent instance with id '%s'.", dmHashReverseSafe64(sp->m_ParentId));
 
                 }
-                Matrix4 parent_t = Matrix4::identity();
+                uint16_t old_parent = instance->m_Parent;
```
JCash, line 2383: Q: When does a DeleteBones happen?
I presume when a GO is deleted.
Are we not doing the "dirty transform" check elsewhere then anyways? Is this extra mechanics just extra work for the same result?

engine/gameobject/src/gameobject/comp_anim.cpp
```diff
@@ -66,6 +67,7 @@ namespace dmGameObject
         uint16_t            m_Composite : 1;
         uint16_t            m_Backwards : 1;
         uint16_t            m_FirstUpdate : 1;
+        uint16_t            m_TransformsUpdated : 1;
```
JCash: I think perhaps better is something like `m_IsTransformProperty`. 
The `m_TransformsUpdated` sounds like it may change more than it actually does.

JCash (changes requested):
Overall good.
Some smaller changes requested

engine/gameobject/src/gameobject/gameobject.cpp
AGulev: Why?
it's an internal static function, all the other internal static functions operate over Collection instead of HCollection
also it is recursive funciton which has to do 

>Collection* collection = hcollection->m_Collection;

on each iteration

engine/gameobject/src/gameobject/gameobject.cpp
AGulev: I'm not sure what do you mean here. This line of code you commented to is in `DispatchMessagesFunction` and this is the case when parent has been changed using a message - not that frequently used, but still should be taken into account. I don't understand what do you mean by Delete bones related question here.

engine/gameobject/src/gameobject/gameobject.cpp
JCash: Yeah, don't ask me how, but for some reason I missed the removal of line 2255 there.
So please disregard this comment.

engine/gameobject/src/gameobject/gameobject.cpp
JCash: Lol, yet another misread 🤦 
pls disregard

JCash (approved):

PR #12344, opened 2026-05-06
Use local Maven repository for editor dependencies

matgis:
As part of the editor build process, we commit some custom-built `.jar` files (such as Bob) to the Maven repository under a hard-coded `1.0` version so they can be picked up by `project.clj`. This is problematic when working with multiple Git worktrees since there may be incompatibilities between different source versions of these `.jar` files.

As a workaround, this PR changes `lein init` and `lein pack` to use a local Maven repository under `defold-checkout/editor/.m2/repository` for dependencies rather than the global `~/.m2/repository`. 

This makes it easier to have processes running concurrently from multiple Git worktrees, since you no longer have to match the Bob sources across all worktrees.

It also removes an unused dependency on `openmali` from the editor.

### DIFF
--- a/editor/.gitignore
+++ b/editor/.gitignore
@@ -5,6 +5,7 @@ pom.xml
 pom.xml.asc
 *.jar
 *.class
+/.m2
 /.lein-*
 /.nrepl-port
 /doc/api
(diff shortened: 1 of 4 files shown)


vlaaad (approved):
LGTM!

PR #12345, opened 2026-05-06
Make bob.jar host tool unpacking more robust

JCash:
This is a stability fix for intermittent issues when bob.jar unpacks and loads tools for building content.

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -639,6 +639,10 @@ public static InvocationResult invoke(ClassLoader classLoader, IProgress progres
             if (cmd == null) { // nothing to do: requested to print help
                 return new InvocationResult(true, Collections.emptyList());
             }
+            if (cmd.hasOption("version")) {
+                System.out.println(String.format("bob.jar version: %s  sha1: %s  built: %s", EngineVersion.version, EngineVersion.sha1, EngineVersion.timestamp));
+                return new InvocationResult(true, Collections.emptyList());
+            }
```
JCash, line 654: Return version before unpacking any tools

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/tools/AndroidTools.java
```diff
@@ -85,6 +85,52 @@ private static String getAapt2Name() {
         return "aapt2";
     }
 
+    private static String getAapt2BundletoolPlatformName(Platform hostPlatform) throws IOException {
```
JCash, line 88: Centralize aapt2 lookup
Make bundletool fallback reachable, include host platform and attempted paths in errors, and reject unsupported host setups instead of defaulting them to macos.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/PackedResources.java
```diff
@@ -19,7 +19,7 @@ public class PackedResources {
     private static volatile boolean unpackStarted = false;
     private static volatile boolean unpackDone = false;
     private static final Object unpackLock = new Object();
-    private static JarFile jarFile;
+    private static final AtomicReference<Throwable> unpackError = new AtomicReference<>();
```
JCash, line 22: Unpack tools first, then load JNI libs
Store async unpack errors and rethrow them from the waiting caller with a clearer message.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/PackedResources.java
```diff
@@ -206,14 +209,16 @@ public static void waitForuUpackAllLibsAsync() {
...
         }
     }
 
     public static void reset() throws IOException {
         unpackStarted = false;
         unpackDone = false;
-        if (jarFile != null) {
-            jarFile.close();
-        }
+        unpackError.set(null);
```
JCash, line 231: avoid closing a cached jar handle and include source/destination paths in copy/extract failures.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/Library.java
```diff
@@ -47,14 +47,20 @@
 public final class Library {
     private static final Logger logger = Logger.getLogger(Library.class.getName());
     private static final int FETCHES_PER_HOST = 4;
-    private static final HttpClient HTTP_CLIENT = HttpClient.newBuilder()
-            .connectTimeout(Duration.ofMillis(2000))
-            .followRedirects(HttpClient.Redirect.NORMAL)
-            .build();
+    private static long HOST_PROBE_TIMEOUT = 5000;
+    private static HttpClient HTTP_CLIENT = httpClient(2000);
```
JCash: Allow Bob to override the editor's default values.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Bob.java
```diff
@@ -540,6 +540,8 @@ private static Project createProject(ClassLoader classLoader, String rootDirecto
         Project project = new Project(classLoader, new DefaultFileSystem(), rootDirectory, buildDirectory);
         project.setOption("email", email);
         project.setOption("auth", auth);
+        Library.setConnectTimeout(15000);
+        Library.setHostProbeTimeout(15000);
```
JCash: Set the Bob timeouts at startup

vlaaad: Bundling from the editor will suddenly change the editor's library timeout. We should find another way to set timeouts, preferably in an immutable way, to avoid such surprising behaviors. For example, the Library class could read a timeout during initialization from a Java property, and the editor config could override it.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/Library.java
```diff
@@ -122,7 +128,7 @@ public static List<Result> fetch(Collection<URI> uris, Path libDir, String email
                         results.add(tasks.get(i).get());
                     } catch (CancellationException e) {
                         var uri = uniqueUris.get(i);
-                        results.add(new Result(uri, cachedByUri.get(uri).result().archive(), new Problem.FetchFailed()));
+                        results.add(new Result(uri, cachedByUri.get(uri).result().archive(), new Problem.FetchFailed("cancelled while checking host availability")));
```
vlaaad: We avoid string messages and use different Problem classes for a reason: the editor needs to localize the messages in various languages, and using a class is a way to select the right localization message and arguments. This means we should not use details at all, and instead, use structured Problem data. For example, for HTTP errors, if you want the error info, you need to create a class like `Problem.FailedHTTPRequest(int status)`.

vlaaad (changes requested):
We should not use the detail strings at all in problems. We should not expose a mutable API for changing timeouts that can be triggered from the editor.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1957,13 +1957,25 @@ public List<Library.Result> resolveLibUrls(IProgress progress) throws LibraryExc
...
+        var detail = libraryProblemDetail(dependency.problem());
+        return detail == null || detail.isBlank() ? message : message + ": " + detail;
+    }
+
+    private static String libraryProblemDetail(Library.Problem problem) {
+        return switch (problem) {
+            case Library.Problem.FailedHTTPRequest(var status) -> "HTTP " + status;
+            case Library.Problem.HttpConnectTimeout _ -> "HTTP connect timed out";
+            default -> null;
+        };
```
vlaaad: Seems unnecessarilty complicated? We can keep it a single `libraryResultMessage` with a single switch.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/Library.java
```diff
@@ -438,6 +464,26 @@ private static int parseLeadingInt(String value) {
...
+        return switch (problem) {
+            case Problem.Missing _ -> "missing";
+            case Problem.FetchFailed _ -> "fetch failed";
+            case Problem.FailedHTTPRequest(var status) -> "fetch failed: HTTP " + status;
+            case Problem.HttpConnectTimeout _ -> "fetch failed: HTTP connect timed out";
+            case Problem.InvalidArchive _ -> "invalid archive";
+            case Problem.DefoldMinVersion(var required) -> "requires Defold " + required + " or newer";
+            case Problem.InstallFailed _ -> "install failed";
+        };
+    }
```
vlaaad, line 485: Seems unnecessary? It's used in readArchive, which is only for tests, so we don't care about user-facing messages...

editor/src/clj/editor/library.clj
```diff
@@ -41,6 +41,12 @@
                    Library$Problem$FetchFailed
                    (localization/message "notification.fetch-libraries.problem.fetch-failed")
 
+                   Library$Problem$FailedHTTPRequest
+                   (localization/message "notification.fetch-libraries.problem.fetch-failed")
+
+                   Library$Problem$HttpConnectTimeout
+                   (localization/message "notification.fetch-libraries.problem.fetch-failed")
+
```
vlaaad, line 51: We could add localization messages to en.editor_localization here...

editor/project.clj
```diff
@@ -161,6 +161,8 @@
 
   :jvm-opts          ["-Djna.nosys=true"
                       "-Djava.net.preferIPv4Stack=true"
+                      "-Ddefold.library.connectTimeoutMillis=2000"
+                      "-Ddefold.library.hostProbeTimeoutMillis=2000"
```
vlaaad: This is dev-only. We should add these properties to `editor/bundle-resources/config` so the editor distribution keeps smaller timeouts. (host should be 5 seconds...)

vlaaad (commented):
Much better! Let's at least ensure the editor bundle uses proper timeouts (2s, 5s in config file).

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
JCash: YEah, I try to keep code style same/similar as before, but sometimes it fails :/

vlaaad (approved):
LGTM!

PR #12350, opened 2026-05-07
Improve descriptor set caching for Vulkan

Jhonnyg:
### Technical changes

When producing / commiting uniform data in the Vulkan graphics adapter, we do so by generating objects that specify the resources to be bound to the shaders:

Draw call 1 resources:
* texture A
* texture B
* uniform buffer A

Draw call 2 resources:
* texture A
* texture C
* uniform buffer B

These two set of resources were previously generated on the fly each frame and is a quite expensive operation in Vulkan. Now instead, we store a small descriptor cache per shader program that we use to try to avoid frequent updates of the same sets - for most scenes the resource inputs to the programs are identical between frames, only the data in them change (i.e camera/object matrices, tint and so on).


engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -504,24 +504,32 @@ namespace dmGraphics
...
-        // 3. Otherwise we use the RT's default render pass (honoring the user-configured load ops).
+        // 4. Otherwise use the RT's default render pass (honoring user-configured load ops).
         VkRenderPass vk_render_pass = rt->m_Handle.m_RenderPass;
         const bool is_main_rt = (render_target == context->m_MainRenderTarget);
-        if (rt->m_HasPendingClearColor && rt->m_Handle.m_RenderPassClear != VK_NULL_HANDLE)
+        if (rt->m_HasPendingClearColor && rt->m_HasPendingClearDepth && rt->m_Handle.m_RenderPassClearColorDepth != VK_NULL_HANDLE)
+        {
+            vk_render_pass = rt->m_Handle.m_RenderPassClearColorDepth;
+        }
+        else if (rt->m_HasPendingClearColor && rt->m_Handle.m_RenderPassClear != VK_NULL_HANDLE)
```
Jhonnyg, line 532: We do this dance here because of how OpenGL works. In Vulkan (and the others), there's a concept of "render passes" where you specify load and store operations on a render targets attachments (clearing, storing and so on). But in a lot of cases in the render scripts there are these scenarios:

render.set_render_target(some_target)
render.clear(buffers)

In these scenarios, we try to fold the clear into the render pass so that we don't clear the buffers twice (which is quite costly). To do this we need to keep track of what buffers are requested to be cleared by the render script and then store variants of render passes in the render target that supports these combinations.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -1824,17 +1840,15 @@ namespace dmGraphics
...
-                all_colors_in_flags = false;
-                break;
-            }
-        }
+        // COLOR0_BIT signals that all color attachments should be cleared to the same value.
+        // Individual COLOR1/2/3 bits are ignored for the fast-path check; the slow path also
+        // treats COLOR0_BIT as "clear all" to keep the semantics consistent.
+        const bool all_colors_in_flags = current_rt->m_ColorAttachmentCount > 0 &&
+                                         (flags & BUFFER_TYPE_COLOR0_BIT) != 0;
+
```
Jhonnyg, line 1848: OpenGL based behaviour.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -1866,17 +1887,10 @@ namespace dmGraphics
...
-        // render.clear(COLOR | DEPTH) on a RT that has both: color folds, depth doesn't.
-        // We detect "was pending" by the fact that we just cleared m_HasPendingClearColor in
-        // BeginRenderPass; re-checking here would require an extra flag, so for simplicity we
-        // always emit the color clear too when taking this slow path. The cost is one extra
-        // attachment entry which is negligible.
+        const bool clear_all_colors = (flags & BUFFER_TYPE_COLOR0_BIT) != 0;
         for (int i = 0; i < current_rt->m_ColorAttachmentCount; ++i)
         {
-            if (flags & color_buffers[i])
+            if (clear_all_colors || (flags & color_buffers[i]))
```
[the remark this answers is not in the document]
Jhonnyg: Our APIs are explicit about this, clearing BUFFER_TYPE_COLOR0_BIT will clear all attachments, since BUFFER_TYPE_COLOR0_BIT is an alias for BUFFER_TYPE_COLOR_BIT

JCash (approved):

PR #12359, opened 2026-05-08
Offload font map generation to a later build phase

vlaaad:
On large projects with many fonts (that don't use runtime generation), build times improved by 7-15%.

Technical notes:

Build time measurements are generally noisy, but for me, the build time of a clean build improved by ~15%, warm build improved by 14%, and hot build by 7%.

Here are 2 example profiling runs of before vs after changes:

| Phase | Baseline run 1 | Font changes run 1 | Delta | 
|---|---:|---:|---:|
| clean-build | 7:28 | 5:51 | -1:37, -21.7% |
| warm-build | 4:56 | 4:00 | -0:56, -18.9% |
| hot-build | 28.37 s | 25.03 s | -3.34 s, -11.8% |

| Phase | Baseline run 2 | Font changes run 2 | Delta | 
|---|---:|---:|---:|
| clean-build | 8:03 | 6:57 | -1:06, -13.7% |
| warm-build | 4:59 | 4:28 | -0:31, -10.4% |
| hot-build | 28.97 s | 27.25 s | -1.72 s, -5.9% |

As a reminder:
- Clean build: nothing in disc cache
- Warm build: disc cache exists
- Hot build: disc and in-memory caches exist

To be honest, I expected to only get improvements for the wam build (we now get the result from the cache), but I guess it makes sense that the clean build also improved. Getting the tree of initial build targets uses node-value, so essentially sequential, while build fns are invoked in parallel, so this change also parallelizes font-map building.

Related to #11988

### DIFF
--- a/editor/src/clj/editor/font.clj
+++ b/editor/src/clj/editor/font.clj
@@ -41,7 +41,7 @@
             [editor.workspace :as workspace]
             [schema.core :as schema]
             [service.log :as log])
-  (:import [com.dynamo.bob.font BMFont]
+  (:import [com.dynamo.bob.font BMFont Fontc]
            [com.dynamo.render.proto Font$FontDesc Font$FontMap Font$FontRenderMode Font$FontTextureFormat Font$GlyphBank]
            [com.google.protobuf ByteString]
            [com.jogamp.opengl GL GL2]
@@ -652,7 +652,8 @@
     (make-font-map _node-id font type font-desc (make-font-resource-resolver font font-resource-map))))
 
 (defn- build-glyph-bank [resource _dep-resources user-data]
-  (let [{:keys [font-map]} user-data]
+  (let [{:keys [font-desc font font-resource-map type digest-ignored/node-id]} user-data
+        font-map (make-font-map node-id font type font-desc (make-font-resource-resolver font font-resource-map))]
     (g/precluding-errors
       [font-map]
       (let [compressed-font-map (font-gen/compress font-map)]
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/font.clj
```diff
@@ -666,47 +667,64 @@
...
+                             :render-mode (:render-mode font-desc)
+                             :all-chars (:all-chars font-desc)
+                             :characters (:characters font-desc)
+                             :cache-width (:cache-width font-desc)
+                             :cache-height (:cache-height font-desc))
+
+                           is-distance-field
+                           (assoc :sdf-spread (Fontc/GetFontMapSdfSpread font-desc-pb)
+                                  :sdf-outline (Fontc/GetFontMapSdfOutline font-desc-pb)
+                                  :sdf-shadow (Fontc/GetFontMapSdfShadow font-desc-pb)))]
```
matgis: There might be a slight discrepancy here around stripping out default values from the built PB message. I think using the PB-enum value for `:layer-mask` might not allow `protobuf/make-map-without-defaults` to correctly identify it as a default value, and the `assoced` fields in the `is-distance-field` branch will obviously always be set in the output binaries even though they may have the default value.

It probably doesn't really matter in this case, but if we wanted to we could convert the `:layer-mask` value into its keyword representation and `merge` in `(protobuf/make-map-without-defaults Font$FontMap :sdf-spread ...)` instead of associng in the values directly.

matgis (dismissed):

vlaaad:
@matgis FYI, I used merge/default stripping for sdf keys, but I left the layer mask as is — Fontc/GetFontMapLayerMask returns int, not an enum.

PR #12372, opened 2026-05-11
Editor: Fix macOS dialog windows showing in-window menu bars

JosephFerano:
The menu bars no longer show up. Also the native system menu on macOS better reflects what options are available.

Fix #12156

### Technical changes

Original finding explaining why the menu bars were showing up to begin with is here;
https://github.com/defold/defold/pull/12324#issuecomment-4405978657

This PR also fixes the following;

image

The context system cannot find a key-combo for command.file.new when a dialog window is opened, which is likely correct behavior. However, there is an edge case when generating menus where if we evaluate whether to make a sub-menu or to create a command entry, if the item _has_ options populated, then it depends on it having a key-combo assigned in order to render as a command. This is unique to command.file.new since it can be both a single command and a submenu.

I first tried adding an `:expand false` descriptor to the command, and it worked, however, it was still generating options, so as a small optimization, we avoid computing them by using `:user-data` instead and checking for `:no-expand` when deciding whether to populate options.

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -1879,7 +1879,7 @@
     :id ::file
     :children [{:label (localization/message "command.file.new")
                 :id ::new
-                :expand false
+                :user-data {:no-expand true}
                 :command :file.new}
                {:label (localization/message "command.file.open")
                 :id ::open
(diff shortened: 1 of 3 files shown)


vlaaad:
Should we instead use the `:expand` functionality, and fix it so we don't even invoke `handler/options` if we have expand false?

JosephFerano:
> Should we instead use the `:expand` functionality, and fix it so we don't even invoke `handler/options` if we have expand false?

I'll investigate this codepath and see if `:expand` is available for handler/options. If it is, then yes, likely that's an even better fix.

vlaaad (approved):
Good stuff!

PR #12377, opened 2026-05-12
Component primitive order unification

### LINKED ISSUE #5666: Sprite, tiles and label winding inconsistent with models
Sprites, tiles and labels (haven't tested GUI, spine and particles) have a different winding than models and meshes. This means that they need to be rendered with either face culling disabled (the default), either with face culling set to cull front faces.

Here's a minimal repro case that showcases this behaviour (see the render script): 
[Archive.zip](https://github.com/defold/defold/files/6167105/Archive.zip)

This isn't normally an issue, since most of the times you have face culling disabled for sprites, but this means that you can't render sprites and models in the same render call. This is especially important if you want to render transparent 3D objects after you've rendered your solids with depth writing on. It's important for transparent objects to be sorted back-to-front relative to each other and you can't do that if you have to split it between two different render predicates.

I think that flipping the winding will be of no consequence for most projects (this is a breaking change only for people that render sprites with `render.FACE_FRONT` and I think people prefer to just disable backface culling altogether).

This would spare the confusion for newcomers by using the intuitive behaviour (I worked with Defold for many years and assumed the winding is consistent and today I lost a few good debugging hours because of this). To ease the transition, it can be done in steps, spaced a few weeks/months apart:

1. Introduce the new behaviour as a `game.project` property and deprecate the old one with a warning (but still keep it default so as not to break projects). Update all template projects to use the new prop.
2. Turn the warning into an error and remove the old codepath. Still keep the old behaviour as default so that old projects get an error and are forced to upgrade.
3. Remove the warning and the prop and make the new behaviour default.

### LINKED ISSUE #9072: [Bug] Default Sprite rotation

I have noticed that sprites by default rotated from camera. It can be issue if you want to render at the same time sprites and models and you want to enable STATE_CULL_FACE.

Sample project:
[My Project.zip](https://github.com/user-attachments/files/15822000/My.Project.zip)

**To Reproduce (REQUIRED)**
In sample project are two sprites. One of them rotated for 180 degrees.
Снимок экрана 2024-06-13 в 17 27 00

1) In my.render_script uncomment 194 line
Снимок экрана 2024-06-13 в 17 29 03
2) Run game
2) You will see that one of sprites not rendered
Снимок экрана 2024-06-13 в 17 30 22

I suppose it is better to fix rotation for sprites by default to face a camera
Also I have check quads (default models) and it is faced ok

  - Version 181

**Platforms (REQUIRED):**
 - Platforms: macOS
 - OS: MacOS 13.0 (22A380)
 - Device: Macbook Pro

**Workaround (OPTIONAL):**
Rotate each sprite by 180 by Y

Jhonnyg:
⚠️ BREAKING CHANGE ⚠️

The vertices for the `sprite`, `gui`, `tilegrid` and `particle` components are now unified with the rest of the components. The engine is now using counter-clockwise ordering of the vertices so that the normal of the primitive is pointing towards the camera. This means that you can use the same face culling parameters for all components.

Regarding breaking change:
This should only have an impact if you are using face culling for your components. In the default render script, we have enabled backface culling only for models but if your project is using backface culling for other components you might have to switch the front/back face using `render.set_cull_face(graphics.FACE_TYPE_BACK)`. The default cull face mode is FACE_TYPE_BACK, so you can also just remove the `render.set_cull_face` call

Fixes https://github.com/defold/defold/issues/3543
Fixes https://github.com/defold/defold/issues/5666
Fixes https://github.com/defold/defold/issues/9072

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/tile/test/TileSetGeneratorTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/tile/test/TileSetGeneratorTest.java
@@ -208,7 +208,7 @@ public class TileSetGeneratorTest {
         // bottom right, bottom left, top left, top right
         List<Float> expectedVertices = Arrays.asList(0.5f, -0.5f, -0.5f, -0.5f, -0.5f, 0.5f, 0.5f, 0.5f);
         List<Float> expectedUVs = Arrays.asList(maxU, minV, minU, minV, minU, maxV, maxU, maxV);
-        List<Integer> expectedIndices = Arrays.asList(0, 1, 2, 0, 2, 3);
+        List<Integer> expectedIndices = Arrays.asList(0, 2, 1, 0, 3, 2);
 
         assertThat(geometry.getVerticesList(), is(expectedVertices));
         assertThat(geometry.getUvsList(), is(expectedUVs));
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/TextureSetGenerator.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/TextureSetGenerator.java
@@ -239,7 +239,7 @@ public class TextureSetGenerator {
 
         // We must respect the upper limit of the vertex hull count
         if (points == null || points.length > hullVertexCount) {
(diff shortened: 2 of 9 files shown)


JCash (approved):

AGulev:
@Jhonnyg Could you please add a few lines with ⚠️ why it is a breaking change wehre user should pay attention to, what exactly might break
(as I understand some setups in render script?)

PR #12378, opened 2026-05-12
Relax default animation validation in editor scripts

vlaaad:
Now it's possible to set a default animation without setting the right atlas first.

Technical notes:

I considered another approach, but it turned out to be much more complex. I thought, ideally, we would create a special setter with `g/expand-ec` that would perform the validation inside the transaction. There were 2 issues with this approach:
1. It would require establishing an ordering for sprite properties. Currently, when we add a new sprite component using editor scripts, we use a lua table, i.e., an unordered hash map. So we would need to add sorting or something like that.
2. Sprites can be used as embedded components, so we would need to add the same custom property + ordering logic for embedded components as well.

In the end, I think simply allowing to set a string value is fine, since it's allowed in the data model, and animation validation is a separate build error check.

### DIFF
--- a/editor/src/clj/editor/editor_extensions/graph.clj
+++ b/editor/src/clj/editor/editor_extensions/graph.clj
@@ -255,17 +255,24 @@
       property
       (string/replace property "-" "_"))))
 
+(defn- available-outline-property [prop-kw outline-property]
+  (when outline-property
+    (let [outline-property (cond-> outline-property
+                                   (:ext-edit-type outline-property)
+                                   (assoc :edit-type (:ext-edit-type outline-property))
+
+                                   (not (contains? outline-property :prop-kw))
+                                   (assoc :prop-kw prop-kw))]
+      (when (and (properties/visible? outline-property)
+                 (edit-type-id->value-converter (properties/property-edit-type-id outline-property)))
+        outline-property))))
+
 (defn- outline-property [node-id property evaluation-context]
   (let [prop-kw (property->prop-kw property)
         outline-property (-> node-id
                              (g/node-value :_properties evaluation-context)
                              (get-in [:properties prop-kw]))]
-    (when (and outline-property
-               (properties/visible? outline-property)
-               (edit-type-id->value-converter (properties/property-edit-type-id outline-property)))
-      (cond-> outline-property
-              (not (contains? outline-property :prop-kw))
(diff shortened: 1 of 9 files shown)


editor/test/resources/editor_extensions/embedded_sprite_default_animation_project/test.editor_script
```diff
@@ -0,0 +1,48 @@
...
+            locations = {"Edit"},
+            run = function()
+                local collection = "/main/main.collection"
+                local go_resource = "/test.go"
+
+                editor.transact({
+                    editor.tx.add(go_resource, "components", {
+                        type = "sprite",
+                        id = "direct-sprite",
+                        __sampler__texture_sampler__0 = "/logo.atlas",
```
matgis, line 17: Unrelated: I never noticed this before, but it's kind of icky that the key mangling we do for sampler properties broke containment and became part of the script API. We might want to come up with a better way to do this at some point. 🤔

editor/src/clj/editor/editor_extensions/graph.clj
```diff
@@ -255,17 +255,24 @@
...
+    (let [outline-property (cond-> outline-property
+                                   (:ext-edit-type outline-property)
+                                   (assoc :edit-type (:ext-edit-type outline-property))
+
+                                   (not (contains? outline-property :prop-kw))
+                                   (assoc :prop-kw prop-kw))]
+      (when (and (properties/visible? outline-property)
+                 (edit-type-id->value-converter (properties/property-edit-type-id outline-property)))
+        outline-property))))
+
```
matgis, line 270: This seems like the sort of function that will be called a lot? If so, we could make it a bit more efficient:
* Move the `(properties/visible? outline-property)` check earlier - It doesn't look like the `outline-property` augmentation can affect it.
* Extract `(:ext-edit-type outline-property)` to a `let` binding so that we don't perform a hash lookup twice.

matgis (dismissed):

editor/test/resources/editor_extensions/embedded_sprite_default_animation_project/test.editor_script
vlaaad: I agree. I thought, ideally, we should present it as a list of maps, i.e., `samplers`, but our property system is not there yet.

PR #12384, opened 2026-05-13
Add debug SPIR-V optimizer flags

Jhonnyg:
New debug options have been added to bob.jar for passing optimization flags to bob for performance measurements:

```
java -jar bob.jar --debug-spirv-opt=skip|-O|-Os|-Oconfig=<file>

skip - no optimization
-O - optimize for performance
-Os - optimize for size
-Oconfig=<file> - pass in a file that specifies specific performance switches
```

The -O and -Os flags combine multiple flags to achieve its goal, where as with the config file case you can specify specific optimization flags to diagnose.


Jhonnyg:
I don't think this is relevant right now.

PR #12385, opened 2026-05-13
Vulkan push constant support

Jhonnyg:


Jhonnyg:
It's a good feature, but not right now.

PR #12394, opened 2026-05-16
Fix Base64Decode to handle input without padding

### LINKED ISSUE #11197: dmCrypt::Base64Decode return nil if input string lenght is not multiple 3
Then please report your issue in the corresponding extension repository!_

If input string length (encoded in base64) has length which is not multiple of 3 - `dmCrypt::Base64Decode` returns nil.

Example string:
Returns nil
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9eyJ0aWQiOiI5YWI3ZGYyZS1iOGM2LTRiYTQtYWRlYi04OTM0NmE3YTMwODUiLCJ1aWQiOiI5YzFjYmYwOS00MDI4LTRjN2ItYmRkYi1mNjIyNzEzYzQyMjQiLCJ1c24iOiI3ZjNhNTFkYi01MmIzLTQ0NTUtYzRhOS05ZWUyNzUzNzNjODUiLCJleHAiOjE3NTQwODU0MTQsImlhdCI6MTc1NDA4NTM1NH0

Return decoded string
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9eyJ0aWQiOiI5YWI3ZGYyZS1iOGM2LTRiYTQtYWRlYi04OTM0NmE3YTMwODUiLCJ1aWQiOiI5YzFjYmYwOS00MDI4LTRjN2ItYmRkYi1mNjIyNzEzYzQyMjQiLCJ1c24iOiI3ZjNhNTFkYi01MmIzLTQ0NTUtYzRhOS05ZWUyNzUzNzNjODUiLCJleHAiOjE3NTQwODU0MTQsImlhdCI6MTc1NDA4NTM1NH0=
(look at padding in the end)

**To Reproduce (REQUIRED)**
1. Open example project
2. Run game

If input encoded string is valid - function should return decoded string.

  - Version [e.g. 1.2.155]

**Platforms (REQUIRED):**

[base64-repro.zip](https://github.com/user-attachments/files/22189493/base64-repro.zip)

**Workaround (OPTIONAL):**
Control string length on client side and if needed - add padding symbol "="

Initially I faced that issue when use https://github.com/defold/extension-crypt as dependency for Nakama client. When I receive token from server it can not be decoded and I got nil instead of json string with tokens. Than I removed crypt extension to firce use of Lua implementation of base64 decode. And it worked. After some research I found that if I passed encoded string with additional padding - engine implementation also works. Engine implementation used functions from mbedTLS.
According to RFC https://datatracker.ietf.org/doc/html/rfc4648#section-3.2 

```
In some circumstances, the use of padding ("=") in base-encoded data
   is not required or used.  In the general case, when assumptions about
   the size of transported data cannot be made, padding is required to
   yield correct decoded data.

   Implementations MUST include appropriate pad characters at the end of
   encoded data unless the specification referring to this document
   explicitly states otherwise.

   The base64 and base32 alphabets use padding, as described below in
   sections [4](https://datatracker.ietf.org/doc/html/rfc4648#section-4) and [6](https://datatracker.ietf.org/doc/html/rfc4648#section-6), but the base16 alphabet does not need it; see
   [section 8](https://datatracker.ietf.org/doc/html/rfc4648#section-8).
```

Also I checked implementation of base64 decoding in newer version of mbedTLS. But no changes in that area.

singhvishalkr:
When decoding base64 strings that lack padding characters, `dmCrypt::Base64Decode` returns nil. This happens because mbedTLS's base64 decoder strictly requires padding, but many systems (particularly JWT implementations) omit padding per RFC 4648 section 3.2 which states that padding is not always required.

This issue can be observed while using the crypt extension with Nakama's JWT tokens. The server sends tokens without padding, and the decode fails silently.

The fix adds the missing `=` characters before passing the input to mbedTLS when the input length is not a multiple of 4. In addition, a test cases covering strings with and without padding was added to prevent regression.

The change should not affect consoles since it only modifies the mbedTLS path.

Fixes #11197

### DIFF
--- a/engine/dlib/src/dlib/crypt_mbedtls.cpp
+++ b/engine/dlib/src/dlib/crypt_mbedtls.cpp
@@ -74,7 +74,23 @@ namespace dmCrypt
     bool Base64Decode(const uint8_t* src, uint32_t src_len, uint8_t* dst, uint32_t* dst_len)
     {
         size_t out_len = 0;
-        int r = mbedtls_base64_decode(dst, *dst_len, &out_len, src, src_len);
+        int r;
+
+        uint32_t padding_needed = (4 - (src_len % 4)) % 4;
+        if (padding_needed > 0 && src_len > 0)
+        {
+            uint32_t padded_len = src_len + padding_needed;
+            uint8_t* padded_src = new uint8_t[padded_len];
+            memcpy(padded_src, src, src_len);
+            memset(padded_src + src_len, '=', padding_needed);
+            r = mbedtls_base64_decode(dst, *dst_len, &out_len, padded_src, padded_len);
+            delete[] padded_src;
+        }
+        else
+        {
+            r = mbedtls_base64_decode(dst, *dst_len, &out_len, src, src_len);
+        }
+
         if (r != 0)
         {
             if (*dst_len == 0)
--- a/engine/dlib/src/dlib/crypt_mbedtls.cpp
+++ b/engine/dlib/src/dlib/crypt_mbedtls.cpp
@@ -71,26 +71,26 @@ namespace dmCrypt
         return true;
     }
 
-    bool Base64Decode(const uint8_t* src, uint32_t src_len, uint8_t* dst, uint32_t* dst_len)
+    bool Base64Decode(const uint8_t* src_original, uint32_t src_len_original, uint8_t* dst, uint32_t* dst_len)
     {
         size_t out_len = 0;
 
-        uint32_t padding_needed = src_len > 0 ? ((4 - (src_len % 4)) % 4) : 0;
-
+        uint32_t padding_needed = (4 - (src_len_original % 4)) % 4;
         uint8_t* padded_src = 0;
-        uint32_t decode_src_len = src_len;
 
-        if (padding_needed != 0)
+        const uint8_t* src = src_original;
+        uint32_t src_len = src_len_original;
+
+        if (padding_needed > 0 && src_len_original > 0)
         {
-            decode_src_len = src_len + padding_needed;
-            padded_src = new uint8_t[decode_src_len];
-            memcpy(padded_src, src, src_len);
-            memset(padded_src + src_len, '=', padding_needed);
+            src_len = src_len_original + padding_needed;
+            padded_src = new uint8_t[src_len];
+            memcpy(padded_src, src_original, src_len_original);
+            memset(padded_src + src_len_original, '=', padding_needed);
+            src = padded_src;
         }
 
-        const uint8_t* decode_src = padded_src ? padded_src : src;
-
-        int r = mbedtls_base64_decode(dst, *dst_len, &out_len, decode_src, decode_src_len);
+        int r = mbedtls_base64_decode(dst, *dst_len, &out_len, src, src_len);
 
         if (padded_src)
         {
--- a/engine/dlib/src/dlib/crypt_mbedtls.cpp
+++ b/engine/dlib/src/dlib/crypt_mbedtls.cpp
@@ -71,22 +71,19 @@ namespace dmCrypt
         return true;
     }
 
-    bool Base64Decode(const uint8_t* src_original, uint32_t src_len_original, uint8_t* dst, uint32_t* dst_len)
+    bool Base64Decode(const uint8_t* src, uint32_t src_len, uint8_t* dst, uint32_t* dst_len)
     {
         size_t out_len = 0;
 
-        uint32_t padding_needed = (4 - (src_len_original % 4)) % 4;
+        uint32_t padding_needed = (4 - (src_len % 4)) % 4;
         uint8_t* padded_src = 0;
 
-        const uint8_t* src = src_original;
-        uint32_t src_len = src_len_original;
-
-        if (padding_needed > 0 && src_len_original > 0)
+        if (padding_needed > 0 && src_len > 0)
         {
-            src_len = src_len_original + padding_needed;
-            padded_src = new uint8_t[src_len];
-            memcpy(padded_src, src_original, src_len_original);
-            memset(padded_src + src_len_original, '=', padding_needed);
+            padded_src = new uint8_t[src_len + padding_needed];
+            memcpy(padded_src, src, src_len);
+            memset(padded_src + src_len, '=', padding_needed);
+            src_len = src_len + padding_needed;
             src = padded_src;
         }
 
--- a/engine/dlib/src/dlib/crypt_mbedtls.cpp
+++ b/engine/dlib/src/dlib/crypt_mbedtls.cpp
@@ -74,21 +74,27 @@ namespace dmCrypt
     bool Base64Decode(const uint8_t* src, uint32_t src_len, uint8_t* dst, uint32_t* dst_len)
     {
         size_t out_len = 0;
-        int r;
 
-        uint32_t padding_needed = (4 - (src_len % 4)) % 4;
-        if (padding_needed > 0 && src_len > 0)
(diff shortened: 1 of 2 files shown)


ekharkunov (dismissed):

engine/dlib/src/dlib/crypt_mbedtls.cpp
```diff
@@ -74,7 +74,23 @@ namespace dmCrypt
     bool Base64Decode(const uint8_t* src, uint32_t src_len, uint8_t* dst, uint32_t* dst_len)
     {
         size_t out_len = 0;
-        int r = mbedtls_base64_decode(dst, *dst_len, &out_len, src, src_len);
+        int r;
+
+        uint32_t padding_needed = (4 - (src_len % 4)) % 4;
```
JCash: We should cerate a define for this later in dalign.h (e.g. `DM_PADDING(x, a)`

engine/dlib/src/dlib/crypt_mbedtls.cpp
```diff
@@ -74,7 +74,23 @@ namespace dmCrypt
...
+            uint8_t* padded_src = new uint8_t[padded_len];
+            memcpy(padded_src, src, src_len);
+            memset(padded_src + src_len, '=', padding_needed);
+            r = mbedtls_base64_decode(dst, *dst_len, &out_len, padded_src, padded_len);
+            delete[] padded_src;
+        }
+        else
+        {
+            r = mbedtls_base64_decode(dst, *dst_len, &out_len, src, src_len);
+        }
```
JCash: We want to keep a single call site:
I.e. 
```
uint8_t* padded_src = 0;

if (padding_needed)
{
    uint32_t padded_len = src_len + padding_needed;
    uint8_t* padded_src = new uint8_t[padded_len];
    memcpy(padded_src, src, src_len);
    memset(padded_src + src_len, '=', padding_needed);
}

int r = mbedtls_base64_decode(dst, *dst_len, &out_len, padded_src, padded_len);

if (padded_src)
{
    delete[] padded_src;
}
```

JCash (changes requested):
Thanks!
A small cleanup needed.

singhvishalkr:
Pushed \eb94d92\ addressing @JCash review: \mbedtls_base64_decode\ is now a single call site with an optional padded buffer + cleanup, matching the structure you sketched. Commit message and author metadata are clean (no tooling co-trailer).

singhvishalkr:
@JCash \HEAD\ is still **\eb94d92\** (single \mbedtls_base64_decode\ path + no Cursor co-trailer). The GitHub review object is pinned to an older \commit.oid\; kindly re-run review when you have a minute so the **CHANGES_REQUESTED** gate can clear.

engine/dlib/src/dlib/crypt_mbedtls.cpp
```diff
@@ -74,21 +74,27 @@ namespace dmCrypt
     bool Base64Decode(const uint8_t* src, uint32_t src_len, uint8_t* dst, uint32_t* dst_len)
     {
         size_t out_len = 0;
-        int r;
 
-        uint32_t padding_needed = (4 - (src_len % 4)) % 4;
-        if (padding_needed > 0 && src_len > 0)
+        uint32_t padding_needed = src_len > 0 ? ((4 - (src_len % 4)) % 4) : 0;
```
JCash: Not sure where this change came from, but it's less readable now.
Please use the previous version.

engine/dlib/src/dlib/crypt_mbedtls.cpp
```diff
@@ -74,21 +74,27 @@ namespace dmCrypt
...
+
+        uint8_t* padded_src = 0;
+        uint32_t decode_src_len = src_len;
```
JCash: I'd like to reuse the current variable here, not a new one...

engine/dlib/src/dlib/crypt_mbedtls.cpp
```diff
@@ -74,21 +74,27 @@ namespace dmCrypt
...
+
+        if (padding_needed != 0)
         {
-            uint32_t padded_len = src_len + padding_needed;
-            uint8_t* padded_src = new uint8_t[padded_len];
+            decode_src_len = src_len + padding_needed;
+            padded_src = new uint8_t[decode_src_len];
```
JCash: let's reuse src/src_len:
```c++
uint32_t decode_src_len = src_len + padding_needed;
padded_src = new uint8_t[decode_src_len;
src = padded_src;
src_len = decode_src_len;
```

engine/dlib/src/dlib/crypt_mbedtls.cpp
```diff
@@ -74,21 +74,27 @@ namespace dmCrypt
...
             memcpy(padded_src, src, src_len);
             memset(padded_src + src_len, '=', padding_needed);
-            r = mbedtls_base64_decode(dst, *dst_len, &out_len, padded_src, padded_len);
-            delete[] padded_src;
         }
-        else
+
+        const uint8_t* decode_src = padded_src ? padded_src : src;
+
+        int r = mbedtls_base64_decode(dst, *dst_len, &out_len, decode_src, decode_src_len);
```
JCash: Simply use `src` and `src_len`.

JCash (changes requested):
Some more cleanup required.

singhvishalkr:
Updated to **87f16b5**: restored the readable padding_needed line and rewrote the padding branch so **src** / **src_len** are the progressing buffer + length handed to a single **mbedtls_base64_decode** call (src/src_len naming from your earlier sketch; parameters renamed in the .cpp definition only).

engine/dlib/src/dlib/crypt_mbedtls.cpp
```diff
@@ -71,10 +71,32 @@ namespace dmCrypt
         return true;
     }
 
-    bool Base64Decode(const uint8_t* src, uint32_t src_len, uint8_t* dst, uint32_t* dst_len)
+    bool Base64Decode(const uint8_t* src_original, uint32_t src_len_original, uint8_t* dst, uint32_t* dst_len)
```
JCash: There is no need to rename these variables.
Keep the old names.

engine/dlib/src/dlib/crypt_mbedtls.cpp
```diff
@@ -71,10 +71,32 @@ namespace dmCrypt
...
     {
         size_t out_len = 0;
+
+        uint32_t padding_needed = (4 - (src_len_original % 4)) % 4;
+        uint8_t* padded_src = 0;
+
+        const uint8_t* src = src_original;
+        uint32_t src_len = src_len_original;
```
JCash: There is no need to create a new variable.
As I mentioned, reuse the `src_len` with a new value as I suggested in my previous review.

engine/dlib/src/dlib/crypt_mbedtls.cpp
```diff
@@ -71,10 +71,32 @@ namespace dmCrypt
...
+
+        if (padding_needed > 0 && src_len_original > 0)
+        {
+            src_len = src_len_original + padding_needed;
```
JCash: Assign it later, as I suggested in my review.

JCash (changes requested):

JCash:
@singhvishalkr Hi!
Are you manually inspecting the resulting code?

singhvishalkr:
Follow-up **786ba5e**: parameters are **src** / **src_len** again; when padding applies we **memcpy/memset**, then **src_len += padding_needed** and **src = padded_src**, single **mbedtls_base64_decode**. No shadow names or **_original** suffix.

JCash (approved):
great!

PR #12399, opened 2026-05-18
Streamline initializing Git worktrees for editor work

matgis:
* Relocate the `openjfx` cache directory so it resides below the worktree.
* Use Maven central for transitive JBoss dependency to avoid non-HTTPS dependency restrictions.
* Exclude inherited environment variables from the output of `/scripts/build.py save_env`.
* Add a `editor` `environment.toml` for use with Codex.


scripts/build.py
```diff
@@ -2877,8 +2866,8 @@ def wait_uploads(self):
             f()
         self.futures = []
 
-    def _form_env(self):
-        env = os.environ.copy()
+    def _form_env(self, inherit = True):
+        env = os.environ.copy() if inherit else {}
```
matgis, line 2870: The map returned by the `_form_env` function is supplied to subprocesses, but is also the basis for the `save_env` task that outputs a shell script with the environment variables. Previously this was only used for CI.

This change makes it so that the environment variables written to `ci/env.sh` no longer include all the environment variables inherited from the shell that called `ci/env.sh`.

scripts/build.py
```diff
@@ -2233,7 +2222,7 @@ def save_env(self):
             self._log("No --save-env-path set when trying to save environment export")
             return
 
-        env = self._form_env()
+        env = self._form_env(inherit = False)
```
matgis, line 2225: This is the only place where we set `inherit = False`. It is in the `save_env` task.

editor/repository-overrides.clj
```diff
@@ -0,0 +1,2 @@
+{:mirrors {"repository.jboss.org" {:name "repository.jboss.org-via-central"
+                                   :url "https://repo1.maven.org/maven2/"}}}
```
matgis, line 2: We have a transitive dependency on `jboss` that refers to a non-HTTPS version of the library. Override it to use the version from Maven Central over HTTPS instead.

editor/.codex/environments/environment.toml
```diff
@@ -0,0 +1,13 @@
+# THIS IS AUTOGENERATED. DO NOT EDIT MANUALLY
+version = 1
+name = "editor"
+
+[setup]
+script = '''
+cd "$CODEX_WORKTREE_PATH/.."
+./scripts/build.py save_env --save-env-path ci/env.sh
+source ci/env.sh
```
matgis, line 9: Setting this up to use `build.py shell` was complicated, but we don't really need it (unless we're on Windows maybe?). Instead, we just make sure the environment variables are set up.

scripts/build.py
```diff
@@ -2914,13 +2903,13 @@ def _form_env(self):
                                       '%s/ext/bin' % self.dynamo_home,
                                       '%s/ext/bin/%s' % (self.dynamo_home, host)])
 
-        env['PATH'] = paths + os.path.pathsep + env['PATH']
+        env['PATH'] = paths + os.path.pathsep + os.environ['PATH']
```
matgis, line 2906: The code was relying on the variables being inherited into the `env` map. Now that we're no longer doing that in all code paths, it reads from the invoking environment instead.

AGulev:
@codex check if changes affect cache in github actions for the editor

scripts/build.py
```diff
@@ -2903,13 +2903,13 @@ def _form_env(self):
...
 
         # This trickery is needed for the bash to properly inherit the PATH that we've set here
         # See /etc/profile for further details
-        is_mingw = env.get('MSYSTEM', '') in ('MINGW64',)
+        is_mingw = os.environ.get('MSYSTEM', '') in ('MINGW64',)
         if is_mingw:
-            env['ORIGINAL_PATH'] = env['PATH']
+            env['ORIGINAL_PATH'] = os.environ['PATH']
```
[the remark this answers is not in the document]
matgis: This is a good catch. I've addressed the issue in a follow-up commit.

scripts/build.py
```diff
@@ -2925,8 +2925,8 @@ def _form_env(self):
 
         # XMLHttpRequest Emulation for node.js
         xhr2_path = os.path.join(self.dynamo_home, NODE_MODULE_LIB_DIR, 'xhr2', 'package', 'lib')
-        if 'NODE_PATH' in env:
-            env['NODE_PATH'] = xhr2_path + os.path.pathsep + env['NODE_PATH']
+        if 'NODE_PATH' in os.environ:
+            env['NODE_PATH'] = xhr2_path + os.path.pathsep + os.environ['NODE_PATH']
```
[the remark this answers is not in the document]
matgis: This is not a concern, as the written environment variables are not supposed to be portable, and are typically injected into the environment using the `source` command immediately after being written.

matgis:
Hmm, the previously merged #12344 would presumably affect the GitHub caching as well, seeing as how we no longer use the global Maven directory `~/.m2` for the editor?

AGulev:
When you merge it, I'll recheck all the caches we currently use

AGulev (approved):

editor/project.clj
```diff
@@ -343,6 +343,7 @@
                                                     "-Ddefold.nrepl=true"
                                                     "-Ddefold.log.dir="
                                                     "-Ddefold.dev=true"
+                                                    "-Djavafx.cachedir=.openjfx/cache"
```
vlaaad, line 346: Huh, aren't those versioned? Why do we need to relocate the cache dir?

vlaaad (approved):
LGTM!

editor/project.clj
matgis: It appears they are in fact placed under versioned directories, but not if you specify `javafx.cachedir`. However, in this case we're also interested in eliminating access to directories outside of the worktree for permission reasons.

PR #12402, opened 2026-05-18
Introduce base render target struct

Jhonnyg:
Graphics adapters now use a shared base struct for Render Targets (similar to textures and shader program).


engine/graphics/src/graphics_adapter.h
```diff
@@ -324,8 +320,6 @@ namespace dmGraphics
         DM_REGISTER_GRAPHICS_FUNCTION(tbl, adapter_name, NewRenderTarget); \
         DM_REGISTER_GRAPHICS_FUNCTION(tbl, adapter_name, DeleteRenderTarget); \
         DM_REGISTER_GRAPHICS_FUNCTION(tbl, adapter_name, SetRenderTarget); \
-        DM_REGISTER_GRAPHICS_FUNCTION(tbl, adapter_name, GetRenderTargetTexture); \
-        DM_REGISTER_GRAPHICS_FUNCTION(tbl, adapter_name, GetRenderTargetSize); \
         DM_REGISTER_GRAPHICS_FUNCTION(tbl, adapter_name, SetRenderTargetSize); \
```
Jhonnyg, line 323: Unfortunately can't remove this function (and others) from the adapter callback table yet since it's not fully data driven struct. So right now we end up in a sort of situation where some functionality are in the adapters and some are not, which is a bit annoying. We can defer some of the functionality to later (like resizing the render target), but some functions require that the inner API backed objects are updated immediately (like SetTexture) since there is no guarantee that the data is still alive after this call.

In a future API, we should probably think about it a bit so we can have a cleaner API between "functions that set/get data from the base struct" and "functions that operate on API backed objects (such as a texture)".

engine/graphics/src/dx12/graphics_dx12.cpp
```diff
@@ -1242,7 +1243,7 @@ namespace dmGraphics
     static inline bool IsRenderTargetbound(DX12Context* context, HRenderTarget rt)
     {
         DX12RenderTarget* current_rt = GetAssetFromContainer<DX12RenderTarget>(context->m_BaseContext.m_AssetHandleContainer, rt);
-        return current_rt ? current_rt->m_IsBound : 0;
+        return current_rt ? current_rt->m_Base.m_IsBound : 0;
```
JCash: I think in cases with acessing the "base", we can do so directly, by requesting the "RenderTarget" instead?
By doing that, we don't have to dereference two pointers:
```c++
RenderTarget* rt = GetAssetFromContainer<RenderTarget>(...);
return rt ? rt->m_IsBound : 0;
```

engine/graphics/src/dx12/graphics_dx12.cpp
```diff
@@ -568,10 +568,11 @@ namespace dmGraphics
         DX12RenderTarget* rt = GetAssetFromContainer<DX12RenderTarget>(context->m_BaseContext.m_AssetHandleContainer, context->m_MainRenderTarget);
         assert(rt == 0x0);
 
-        rt               = new DX12RenderTarget();
-        rt->m_Id         = DM_RENDERTARGET_BACKBUFFER_ID;
-        rt->m_Format     = DX12GetBackBufferFormat();
-        rt->m_SampleDesc = sample_desc;
+        rt = new DX12RenderTarget();
+        rt->m_Base.m_Id                   = DM_RENDERTARGET_BACKBUFFER_ID;
+        rt->m_Base.m_ColorAttachmentCount = 1;
```
JCash: SImilarly, I'd make use of a pointer to the RenderTarget:
```c++
RenderTarget* brt = &rt->m_Base;
```

JCash (changes requested):
Overall good.
I'd like to  tighten the pointer chasing when possible.

PR #12407, opened 2026-05-19
Add project build time metrics to load-project dev module

matgis:
* You can now set the `DM_DEV_LOAD_PROJECT_PATH` environment variable to an absolute directory path to change which project is loaded by `load_project.clj`.
* Refactored the `build/build-project!` function into separate phases so that they can be called and measured from `load_project.clj`.
* Exclude the `:setup-workspace` and `:fetch-libraries` phases from the total measurements. `:setup-workspace` is trivial, and `:fetch-libraries` varies too much based on network conditions.
* Removed disabled `:store-post-load-system` phase. This just stored a snapshot of the populated `system` in a `var`.
* Removed disabled `:calculate-scene-deps` phase. This was used for a while to calculate graph dependencies for the `scene` output of various resources, but doesn't really belong here.
* Added `:evaluate-build-targets` phase. In this phase we evaluate the `build-targets` outputs for all nodes connected to the `build-targets` input on `/game.project`.
* Added `:resolve-build-target-deps` phase. In this phase we evaluate `build-targets` for any `dynamic-deps` listed in the prior `build-targets`.
* Added `:build-build-targets` phase. In this phase we call the `:build-fns` in the `build-targets`, which builds the data compiled in the `build-targets` and writes the resulting binaries to disk.

Related to #11988.

### DIFF
--- a/editor/src/clj/editor/build.clj
+++ b/editor/src/clj/editor/build.clj
@@ -29,14 +29,8 @@
             [util.fn :as fn])
   (:import [java.util ArrayDeque HashMap HashSet]))
 
-(defn- batched-pmap [f batches]
-  (->> batches
-       (pmap (fn [batch] (doall (map f batch))))
-       (reduce concat)
-       doall))
-
-(defn- available-processors []
-  (.availableProcessors (Runtime/getRuntime)))
+(set! *warn-on-reflection* true)
+(set! *unchecked-math* :warn-on-boxed)
 
 (defn- compiling-progress-message [node-id->resource-path node-id]
   (if (nil? node-id)
@@ -78,9 +72,9 @@
          (cond-> ret (not (g/error-value? ret)) flatten-build-targets)))))
   ([^HashMap seen stack build-targets project proj-path->dynamic-build-targets report-node-id-progress! evaluation-context]
    (let [ret
-         (into
-           []
-           (comp
+         (if (g/error-value? build-targets)
+           build-targets
+           (coll/into-> build-targets []
              coll/flatten-xf
              (map (fn [build-target]
                     (if (g/error-value? build-target)
@@ -129,8 +123,7 @@
                                           (assoc :deps resolved-deps)
                                           (dissoc :dynamic-deps)))))]
                             (.put seen content-hash result)
-                            result)))))))
-           build-targets)]
+                            result))))))))]
      (or (g/flatten-errors ret) ret))))
 
 (defn resolve-dependencies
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/build.clj
```diff
@@ -29,14 +29,8 @@
...
   (:import [java.util ArrayDeque HashMap HashSet]))
 
-(defn- batched-pmap [f batches]
-  (->> batches
-       (pmap (fn [batch] (doall (map f batch))))
-       (reduce concat)
-       doall))
-
-(defn- available-processors []
-  (.availableProcessors (Runtime/getRuntime)))
```
matgis, line 39: Removed unused code.

editor/src/clj/editor/build.clj
```diff
@@ -78,9 +72,9 @@
          (cond-> ret (not (g/error-value? ret)) flatten-build-targets)))))
   ([^HashMap seen stack build-targets project proj-path->dynamic-build-targets report-node-id-progress! evaluation-context]
    (let [ret
-         (into
-           []
-           (comp
+         (if (g/error-value? build-targets)
+           build-targets
+           (coll/into-> build-targets []
```
matgis, line 77: Guard against `build-targets` itself being an `ErrorValue`.

editor/src/clj/editor/build.clj
```diff
@@ -165,58 +158,73 @@
...
 
-        render-progress!
-        (if (= fn/constantly-false task-cancelled?)
-          supplied-render-progress!
-          (fn render-progress! [progress]
-            (if (task-cancelled?)
-              (throw-build-cancelled-exception!)
-              (supplied-render-progress! progress))))
+(defn make-node-id->resource-path [project evaluation-context]
+  (set/map-invert (g/node-value project :nodes-by-resource-path evaluation-context)))
```
matgis, line 173: Also used by `load_project.clj`.

editor/src/clj/editor/build.clj
```diff
@@ -165,58 +158,73 @@
...
         progress-tracer (project/make-progress-tracer :build-targets step-count progress-message-fn (progress/nest-render-progress render-progress! (progress/make localization/empty-message 10 0 true) 5))
-        evaluation-context-with-progress-trace (assoc evaluation-context :tracer progress-tracer)
-        _ (doseq [node-id (rseq @steps)]
-            (try
-              (g/node-value node-id :build-targets evaluation-context-with-progress-trace)
-              (catch Throwable error
-                (throw (pipeline/decorate-build-exception error :compile node-id nil evaluation-context)))))
-        #_#_#_#_
-        prewarm-partitions (partition-all (max (quot step-count (+ (available-processors) 2)) 1000) (rseq @steps))
-        _ (batched-pmap (fn [node-id] (g/node-value node-id :build-targets evaluation-context-with-progress-trace)) prewarm-partitions)
```
matgis, line 210: I removed these commented-out lines. Parallelized builds was disabled due to high memory usage a while back, and running it in serial ended up being slightly faster anyway. We probably want to do a better take on parallelizing this if we end up taking that route.

editor/src/dev/load_project.clj
```diff
@@ -269,40 +305,7 @@
 
 (defonce load-metrics
   (du/when-metrics
-    {:new-nodes-by-path (g/node-value project :nodes-by-resource-path)
+    {:new-nodes-by-path (some-> project (g/node-value :nodes-by-resource-path))
```
matgis, line 308: Prevent exception if we're stopping after an earlier phase.

editor/src/clj/editor/build.clj
vlaaad: Hmm, I'm sure previous semantics expected that the value passed here is not an error value. For example, `resolve-node-dependencies` checks if build targets are error values before calling this function. Perhaps new semantics are better? In this case, we should update `resolve-node-dependencies` so it does not guard unnecessarily.

vlaaad (dismissed):
LGTM, but maybe we can align the semantics around error values a bit more?

matgis:
Good point. I prefer the new semantics, so I'll remove the redundant error check in `resolve-node-dependencies`.

PR #12408, opened 2026-05-19
Fix cache issues

AGulev:
- Scope Gradle cache dependency paths to Bob Gradle files.
- Cache editor-local Maven repo instead of `~/.m2`.
- Reuse the built `bob.jar` artifact in editor CI jobs.
- Remove generated Bob Maven artifact before cache save.
- Centralize editor JDK version in `build_tools/sdk.py`.


matgis (approved):
I'm not comfortable reviewing the `.yml` changes, but the changes to `.py` and `.clj` files look good to me!

PR #12414, opened 2026-05-20
Fixes model transform inconsistencies between the editor and runtime for imported glTF/glb models

Jhonnyg:
#### Model re-centering during import ⚠️ BREAKING CHANGE ⚠️
We have removed the automatic model re-centering step from the Bob model import pipeline so imported glTF model positions are not altered. It also adjusts model transform handling so skinned/bone-parented meshes preserve local transforms while rigid meshes keep their flattened world placement. Both of these changes takes Defold closer to a "what you see is what you get" approach for 3D content.

The re-centering means that If your 3D content suddenly changes positions / orientation, this means that the content needs to be modified in an authoring package (e.g Blender) and then re-exported to .glb again. Most projects should be unaffected by this change, but there _is_ a change in behaviour when importing models now in order to fix other incorrectly placed objects.

#### Editor vs runtime preview inconsitency
The editor preview now mirrors the runtime transform behaviour more closely: models with assigned skeletons use skeleton-relative transforms, while models without skeleton resources use their local model transforms, with bounds updated to match the selected transform path.

#### Runtime GPU skinning fixes
Runtime rig/model handling was also adjusted so parentless root bones keep the inherited ancestor transform from the imported skeleton, and skinning matrices are only applied when the model is actually animating.


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ModelUtil.java
```diff
@@ -1045,10 +1043,11 @@ private static Rig.Model loadModel(Node node, Model model, ArrayList<Modelimport
...
         modelBuilder.setId(MurmurHash.hash64(node.name)); // the node name is the human readable name (e.g Sword)
-        // Handle GLTF hierarchy correctly based on whether the model is skinned:
-        // - If model is skinned: use node.local to preserve bone hierarchy system
-        // - If model is not skinned: use node.world to flatten transform hierarchy into model transform
-        if (skeleton.size() > 0) {
+        // Preserve local transforms only for meshes that rely on the bone hierarchy at runtime.
+        // Other rigid meshes in a skinned scene should keep their flattened world placement
+        // to match the authored scene preview.
+        boolean preserveLocalTransform = node.skin != null || model.parentBone != null;
+        if (preserveLocalTransform) {
```
Jhonnyg, line 1050: This makes sure the root node of a scene is applied to the internal hierarchy, otherwise we can't support root orientation or scale.

editor/src/clj/editor/model_scene.clj
```diff
@@ -470,9 +470,26 @@
...
          :renderable-buffers renderable-buffers}))))
 
-(defn- make-renderable-model [model model-request-id mesh-set mesh-material-index->material-name]
+(defn- make-bone-id->world-transform [skeleton]
+  (into {}
+        (map (fn [bone]
+               [(:id bone)
+                (let [{:keys [translation rotation scale]} (:world bone)]
+                  (math/clj->mat4 translation rotation scale))]))
+        (:bones skeleton)))
```
Jhonnyg, line 479: Previously, objects anchored to skeleton bones (e.g a hat anchored to a head/neck bone of a skinned character) were not transform correctly. We now first create a bone-id->transform map and then transform the model accordingly.

editor/src/clj/editor/model_scene.clj
```diff
@@ -488,14 +505,16 @@
                          geom/null-aabb
                          renderable-meshes)]
         {:pose model-pose
+         :pose-without-skeleton local-pose
```
Jhonnyg: The "pose-without-skeleton" is a bit of an editor-only custom fix to match the runtime. it's used when the model doesn't have a skeleton resource assigned, but the mesh file _does_. Previously, the editor would always pose the model based on the skeleton since we will generate both the pose and mesh data from the input resource if they exist. 

In the runtime case we don't apply any pose if the model doesn't have a skeleton resource, regardless if the input has this data or not.

engine/rig/src/rig.cpp
```diff
@@ -682,17 +682,27 @@ namespace dmRig
...
             else
-                bp.m_World = bp.m_Local;
+            {
+                // Apply the skeleton bone transform when the there is no parent.
+                const dmRigDDF::Bone* sk_bone = &skeleton->m_Bones[bi];
+                // Separate the anscestor (everything "above" this joint hierarchy-wise) from the local transform of this joint.
+                const Matrix4 world_bind = dmTransform::ToMatrix4(sk_bone->m_World);
+                const Matrix4 local_bind = dmTransform::ToMatrix4(sk_bone->m_Local);
+                const Matrix4 ancestor = world_bind * dmVMath::Inverse(local_bind);
+                bp.m_World = dmTransform::ToTransform(ancestor * dmTransform::ToMatrix4(bp.m_Local));
```
Jhonnyg, line 707: UpdatePoseTransforms used to treat a parentless bone as if its local transform was already in model space, which would lose any static transform inherited from nodes above the root bone in the imported glTF hierarchy. 

But as soon as runtime animation updates the pose, UpdatePoseTransforms recomputes m_World and would drop ancestor. The new code applies that static ancestor transform so parentless bones are in the same space as animations. 

There's a bit of work here that is done which potentially could be cached. But since this is only run on "root" nodes which typically are just a few, I'm not sure it's warranted to introduce more code changes just for this?

editor/src/clj/editor/model_scene.clj
```diff
@@ -627,23 +648,43 @@
...
   (if (g/error-value? scene)
     scene
     (let [old-node-id (:node-id scene)
-          model-scenes (:children scene)]
-      (assoc scene
+          model-scenes (:children scene)
+          augmented-model-scenes (mapv #(augment-model-scene % old-node-id new-node-id new-node-outline-key material-name->material-scene-info use-skeleton-transforms?)
+                                       model-scenes)
+          scene-aabb (when-not use-skeleton-transforms?
+                       (model-scenes-aabb augmented-model-scenes))
```
Jhonnyg: When there is no skeleton assigned, we need to update the AABB to use the non-posed transforms since we are not using the skeleton pose to transform the model. Otherwise picking, framing and selection will be incorrect.

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -2024,7 +2026,9 @@ namespace dmGameSystem
 
                 dmVMath::Vector4 animation_data(0.0f, 0.0f, 0.0f, 0.0f);
 
-                if (dmRig::GetPoseMatrixCacheDataOffset(world->m_RigContext, component->m_RigInstance) != dmRig::INVALID_POSE_MATRIX_CACHE_ENTRY)
+                if (component->m_Resource->m_RigScene->m_SkeletonRes &&
+                    dmRig::IsAnimating(component->m_RigInstance) &&
+                    dmRig::GetPoseMatrixCacheDataOffset(world->m_RigContext, component->m_RigInstance) != dmRig::INVALID_POSE_MATRIX_CACHE_ENTRY)
```
Jhonnyg, line 2031: This guard keeps skinned models on the static/non-skinned render path until an animation is actually playing.
Some imported skeletons carry root/ancestor transforms, and applying bind-pose skinning while idle can visibly change placement/rotation compared to the authored model transform.
When animation starts, we pass the pose matrix cache data so the shader applies the animated skeleton as expected.

JCash (dismissed):

editor/src/clj/editor/model_scene.clj
```diff
@@ -627,23 +648,43 @@
...
+(defn- model-scenes-aabb [model-scenes]
+  (transduce
+    (keep (fn [{:keys [aabb pose]}]
+            (when pose
+              (geom/aabb-transform aabb (pose/matrix pose)))))
+    geom/aabb-union
+    geom/null-aabb
+    model-scenes))
+
+(defn augment-scene [scene new-node-id new-node-outline-key material-name->material-scene-info use-skeleton-transforms?]
```
vlaaad: Code style: we use `?` suffix for predicate functions, e.g., `even?`, `pos?`, but not for boolean values, e.g.:
```clj
(let [even (even? num)] ...)
```

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -2024,7 +2026,9 @@ namespace dmGameSystem
 
                 dmVMath::Vector4 animation_data(0.0f, 0.0f, 0.0f, 0.0f);
 
-                if (dmRig::GetPoseMatrixCacheDataOffset(world->m_RigContext, component->m_RigInstance) != dmRig::INVALID_POSE_MATRIX_CACHE_ENTRY)
+                if (component->m_Resource->m_RigScene->m_SkeletonRes &&
+                    dmRig::IsAnimating(component->m_RigInstance) &&
```
vlaaad: Codex complained about this line; feel free to ignore if irrelevant. In its words:
gating `m_AnimationData` on `dmRig::IsAnimating()` can drop the final sampled pose for one-shot animations. `UpdatePlayer` clears `m_Playing` when `PLAYBACK_ONCE_*` completes, but `DoAnimate` still samples and commits that final pose in the same update. Render then sees `IsAnimating() == false` and sends zero animation data, so the final frame can render bind/unskinned instead of the completed pose.

vlaaad (dismissed):
Editor changes look good to me. Minor code style comment.

engine/gamesys/src/gamesys/components/comp_model.cpp
Jhonnyg: Yeah it makes sense. Thanks!

PR #12415, opened 2026-05-20
Fix `gui.set_screen_position()` for nodes with scale

### LINKED ISSUE #12412: gui.get_screen_position() return not expected value
**Describe the bug**
The gui.get_screen_position() method does not return the same coordinates that were set using the gui.set_screen_position() method (because of the scale of the parent node)

**To Reproduce**
run project and click to green box to execute simple code:
```lua
local pos = gui.get_screen_position(node)
gui.set_screen_position(node, pos)
```

**Expected behavior**
The value that we set, we can get back. In the attached example, all the green squares would have to stay in place when you click on them.

**Defold version :**
  - 1.12.4 (402218d5544666871f07ecdfa21032a7fb59413f)

**Platforms :**
 - Platforms: any
 
**Example Project :**
[get_screen_pos_issue.zip](https://github.com/user-attachments/files/28028477/get_screen_pos_issue.zip)

AGulev:
Fixed an issue where `gui.set_screen_position()` and `gui.screen_to_local()` calculated incorrect positions for nodes with scaled parents.

Fix https://github.com/defold/defold/issues/12412

### DIFF
--- a/engine/gui/src/gui.cpp
+++ b/engine/gui/src/gui.cpp
@@ -4474,6 +4474,8 @@ namespace dmGui
         Vector4 adjust_scale;
         Vector4 offset(0.0f);
 
+        Vector3 position = screen_position;
+
         // Calculate the new parents scene transform
         // We also need to calculate values relative to adjustments and offset
         // corresponding to the values that would be used in AdjustPosScale (see reasoning below).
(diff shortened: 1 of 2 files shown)


engine/gui/src/gui.cpp
```diff
@@ -4495,10 +4509,6 @@ namespace dmGui
             offset.setY(offset.getY() + scene->m_AdjustOffsetY);
         }
 
-        // We calculate a new position that will be the relative position once
-        // the node has been childed to the new parent.
-        Vector3 position = screen_position - parent_m.getCol3().getXYZ();
```
AGulev, line 4500: we can't compensate position like `screen_position - parent_translation` it doesn't include scale

so now it is `inverse(parent_m) * Vector4(screen_position, 1.0f)`

britzl (approved):

PR #12424, opened 2026-05-21
Add Lua stack index arguments to dmSDK functions LuaToJson and JsonToLua

### LINKED ISSUE #9446: `LuaToJson` uses the first object in the stack, not the top
https://defold.com/ref/stable/dmScript/#dmScript::LuaToJson:L-json-json_len

The linked documentation for the `dmScript::LuaToJson` function doesn't state what value it uses to transform a table to string: from the top of the stack or not. Lua C API usually uses the object at the top of the stack and I assumed that `dmScript::LuaToJson` works the same way. But it didn't, because the function [gets the first value on the stack](https://github.com/defold/defold/blob/dev/engine/script/src/luacjson/lua_cjson.c#L975). It's very inconvenient and strange, in my view.

Any variant is great:
- the function takes a value from the top of the stack
- or add a parameter that specifies where to take the value from

  - Version: 1.9.3

**Workaround (OPTIONAL):**
```c++
int LuaToJson(lua_State* L, int index, char** json, size_t* json_len)
{
    lua_pushvalue(L, index);
    lua_insert(L, 1);
    int result = dmScript::LuaToJson(L, json, json_len);
    lua_remove(L, 1);
    return result;
}
```

britzl:
It is now possible to provide a Lua stack index to the dmSDK function `LuaToJson()` to specify which stack value to encode, as well as the index at which the option table is located. Previously the default was to always encode the value at stack index 1 and to read the options from index 2, which required workarounds for most use-cases. Also `JsonToLua()` now has a version which takes the stack index of the options table.

```cpp
int JsonToLua(lua_State* L, const char* json, size_t json_len, int options_index);
int LuaToJson(lua_State* L, int index, int options_index, char** json, size_t* json_len);
```

Fixes #9446

### DIFF
--- a/engine/script/src/luacjson/lua_cjson.c
+++ b/engine/script/src/luacjson/lua_cjson.c
@@ -471,7 +471,9 @@ static void json_create_config(lua_State *l)
 #endif // DEFOLD
 
 // DEFOLD
-static void json_initialize_config(lua_State *l, json_config_t* cfg, int encode_keep_buffer, int protected_mode, char* errbuf, size_t errbuf_len)
+static void json_initialize_config(lua_State *l, json_config_t* cfg,
+                                   int encode_keep_buffer, int protected_mode,
+                                   int options_index, char* errbuf, size_t errbuf_len)
 {
     int i;
 
@@ -545,9 +547,9 @@ static void json_initialize_config(lua_State *l, json_config_t* cfg, int encode_
     cfg->escape2char['u'] = 'u';          /* Unicode parsing required */
 
     // read additional config values from options table
-    if (lua_istable(l, 2))
+    if (lua_istable(l, options_index))
     {
-        lua_pushvalue(l, 2); // push config table to top of stack
+        lua_pushvalue(l, options_index); // push config table to top of stack
 
         lua_getfield(l, -1, "decode_null_as_userdata");
         if (!lua_isnil(l, -1))
@@ -970,13 +972,14 @@ static int json_encode(lua_State *l)
 // a default config here to avoid passing the config as
 // a user data object.
 // Also, we return the generated string, and let the caller delete it
-int lua_cjson_encode(lua_State *l, char** json_str, size_t* json_length)
+int lua_cjson_encode(lua_State *l, int index, int options_index,
+                     char** json_str, size_t* json_length)
 {
     json_config_t cfg;
     strbuf_t local_encode_buf;
     strbuf_t *encode_buf;
 
-    json_initialize_config(l, &cfg, DEFAULT_ENCODE_KEEP_BUFFER, 1, 0, 0);
+    json_initialize_config(l, &cfg, DEFAULT_ENCODE_KEEP_BUFFER, 1, options_index, 0, 0);
     if (!cfg.encode_keep_buffer) {
         /* Use private buffer */
         encode_buf = &local_encode_buf;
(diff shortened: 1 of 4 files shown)


AGulev (dismissed):
I like it.

But initially we talked about this issue as about a breaking change we will do at some point. 

He you decided to keep it back compatible with I personally like (it will help to avoid preaiimg changes in almost ALL our extensions), but as I remember @JCash was against it.

So, I approve it and if Mathias is fine with it and you merge it, please don't forget to remove the Breaking Change label from the issue

britzl:
Sure, I could remove the functions. Or at the very least stop documenting them. And perhaps also emit a warning.

AGulev:
not sure about warning, but deprecate them by removing from documentation is indeed a good idea, then we can create a task to cleanup all our extension to use new one and remove old one in some time

engine/script/src/luacjson/lua_cjson.c
```diff
@@ -471,7 +471,9 @@ static void json_create_config(lua_State *l)
 #endif // DEFOLD
 
 // DEFOLD
-static void json_initialize_config(lua_State *l, json_config_t* cfg, int encode_keep_buffer, int protected_mode, char* errbuf, size_t errbuf_len)
+static void json_initialize_config(lua_State *l, json_config_t* cfg,
```
JCash: By convention, we keep the Lua stack index directly after the Lua state.

engine/script/src/luacjson/lua_cjson.c
```diff
@@ -1627,13 +1630,15 @@ static int DefoldError(json_config_t* cfg, const char* format, ...)
 // a default config here to avoid passing the config as
 // a user data object.
 // Note: We also have a boolean which decides upon the error handling
-int lua_cjson_decode(lua_State *l, const char* json_string, size_t json_len, int protected_mode, char* errbuf, size_t errbuf_len)
+int lua_cjson_decode(lua_State *l, const char* json_string, size_t json_len,
```
JCash, line 1633: Same with the index here.

engine/script/src/script_json.cpp
```diff
@@ -72,12 +72,22 @@ namespace dmScript
...
+        return JsonToLuaInternal(L, json, json_len, options_index, 0);
     }
 
     int LuaToJson(lua_State* L, char** json, size_t* json_len)
     {
-        return lua_cjson_encode(L, json, json_len);
+        return lua_cjson_encode(L, 1, 2, json, json_len);
+    }
+
+    int LuaToJson(lua_State* L, int index, int options_index, char** json, size_t* json_len)
```
JCash, line 88: 👍

engine/script/src/script_json.cpp
```diff
@@ -72,12 +72,22 @@ namespace dmScript
 
     int JsonToLua(lua_State* L, const char* json, size_t json_len)
     {
-        return JsonToLuaInternal(L, json, json_len, 0);
+        return JsonToLuaInternal(L, json, json_len, 2, 0);
+    }
+
+    int JsonToLua(lua_State* L, const char* json, size_t json_len, int options_index)
```
JCash: Same here

JCash (changes requested):
overall good.
However, I'd like to keep the convention of having the Lua indices next to the Lua state.

JCash (dismissed):

PR #12425, opened 2026-05-21
Rename internal library to dmbedtls to avoid naming collisions

JCash:
## Summary
- Renames the packaged mbedtls build artifact to `dmbedtls`
- Updates Waf, CMake, and extender build references to link against the new library name
- Keeps the feature flag, source layout, and upstream `mbedtls` API references unchanged

Related issue https://github.com/defold/defold/issues/10939#issuecomment-4506312390


AGulev (approved):
👍❤️

PR #12443, opened 2026-05-24
Fix runtime font job teardown crash

### LINKED ISSUE #12437: EXC_BAD_ACCESS in JobPostProcessGlyph
**Description**

`EXC_BAD_ACCESS` (address `0x8`) crash in `JobSystemUpdate` → `ProcessFinishedJobs` → `JobPostProcessGlyph` → `ResFontAddGlyph` caused by a use-after-free of `FontGenJobData` in `jobsystem.cpp`.

**Root cause - AI speculation:** `CancelJobInternal` (jobsystem.cpp ~line 337) has a special case for `JOBSYSTEM_STATUS_FINISHED` that returns `RESULT_OK` without changing the job status or removing it from the done queue. This means:

1. `CancelPendingJobs` calls `CancelJobInternal` on a RUNNING glyph job.
2. The job completes on the background thread during the spin-wait → status becomes `FINISHED`.
3. `CancelJobInternal` returns `RESULT_OK` — **but leaves the job in the done queue with status `FINISHED`**.
4. `DeallocateJobResourceInfo` frees `FontGenJobData` (the job's context pointer).
5. On the next frame, `JobSystemUpdate` → `ProcessFinishedJobs` sees the job as `FINISHED` and calls `JobPostProcessGlyph(freed_ptr)`.
6. `jobdata->m_FontResource` is NULL (freed memory zeroed by iOS) → `font->m_FontMap` at offset 8 → dereference of address `0x8` → `EXC_BAD_ACCESS`.

**Proposed fix direction - AI speculation (two changes to `jobsystem.cpp`):**

The fix requires two coordinated changes. The exact implementation depends on the surrounding code structure, but the intent is:

1. **`CancelJobInternal`** — when the job is found in `FINISHED` state, set its status to `CANCELED` before returning `RESULT_OK`. This prevents the done queue from retaining a job whose context has been freed.

2. **`ProcessFinishedJobs`** — before invoking a job's callback, skip jobs whose status is `CANCELED` (instead of invoking the callback unconditionally for all `FINISHED` jobs).

**To Reproduce**

The attached minimal project (`defold-font-crash-demo.zip`) reproduces the crash in ~5 seconds, 100% of the time on an iPad 9th gen (Defold 1.12.4 debug build).

**How the repro works:**

1. A collection proxy (`proxy_a`) is loaded. It contains 5 labels all rendering with a 512pt SDF font (`font_a`). A script cycles each label through a different character every frame.
2. 512pt SDF glyph rasterization takes ~80–120ms per glyph on a single background thread. Five labels submit 5 sequential jobs — total rasterization time ~500ms, guaranteed to span multiple frames.
3. After frame 5, the script sends `disable` + `unload` to `proxy_a`.
4. This triggers `CancelPendingJobs` (via `DeleteFontResource`) while multiple glyph jobs are still `RUNNING`.
5. `CancelJobInternal` spin-waits until each job becomes `FINISHED`, returns `OK` (bug), leaves jobs in the done queue, and `DeallocateJobResourceInfo` frees their `FontGenJobData`.
6. The next `JobSystemUpdate` calls `ProcessFinishedJobs` → `JobPostProcessGlyph(freed_ptr)` → crash.

**Steps to reproduce:**

1. Open the attached project in Defold 1.12.4.
2. Build and deploy to an iOS device (arm64).
3. Launch the app — it runs automatically with no user interaction required.
4. Observe `EXC_BAD_ACCESS` within ~5 seconds.

**Expected behavior**

The app should not crash

**Defold version:**

- Version: **1.12.4** (`402218d5544666871f07ecdfa21032a7fb59413f`)

**Platforms:**

- Platforms: **iOS** (confirmed crash), others are unconfirmed
- OS: iOS 26.4 and iOS 26.5 are confirmed
- Devices: iPhone 12 mini (iPhone13,2), iPhone 13 mini (iPhone14,4), iPhone 13 Pro Max (iPhone14,3), iPhone 16 (iPhone16,1), iPad 9th gen (demo project repro device)

Please see attached [defold-font-crash-demo.zip](https://github.com/user-attachments/files/28160188/defold-font-crash-demo.zip)

Project structure:

```text
defold-font-crash-demo/
  fonts/font_a.font          # 512pt SDF font
  main/
    driver.script            # orchestrates load → render → unload
    label_cycler.script      # cycles 5 labels through new characters each frame
    scene_a.collection       # 5 label game objects using font_a
    main.collection          # collection proxy wrapper
  appmanifests/game.appmanifest
  game.project
```

**Logs:**

Console output from iPad 9th gen — crash occurs ~594ms after the unload is issued (the spin-wait duration for 5 queued SDF glyph jobs):

```text
DEBUG:SCRIPT: [font-uaf] Step 1: Loading Scene A (Font A 512pt SDF, prewarm chars: 'ABC')...
DEBUG:SCRIPT: [font-uaf] Step 2: Scene A loaded. Enabling labels...
DEBUG:SCRIPT: [font-uaf] Step 3: Labels rendering. Font A is 512pt SDF — each glyph job
DEBUG:SCRIPT: [font-uaf]          takes ~80-120ms, longer than one frame. The per-frame
DEBUG:SCRIPT: [font-uaf]          JobSystemUpdate will see the job as RUNNING, not FINISHED.
DEBUG:SCRIPT: [font-uaf] Step 4: Unloading proxy_a while a glyph job is still RUNNING.
DEBUG:SCRIPT: [font-uaf]         CancelJobInternal will spin-wait, then hit the FINISHED bug.
DEBUG:SCRIPT: [font-uaf] Step 5: EXPECT EXC_BAD_ACCESS at 0x8 in the next JobSystemUpdate
INFO:CRASH: Successfully wrote Crashdump to file: .../_crash
ERROR:CRASH: CALL STACK:
ERROR:CRASH: 0   DefoldFontUAFCrashDemo              0x0000000104ac9324 _ZN7dmCrash7OnCrashEi + 108
ERROR:CRASH: 1   libsystem_platform.dylib            0x00000002b68cd158 352B728E-E013-32F4-A54C-C221ED498CDC + 16728
ERROR:CRASH: 2   DefoldFontUAFCrashDemo              0x000000010498037c _ZN12dmGameSystemL19JobPostProcessGlyphEP10JobContexty15JobSystemStatusPvS3_i + 76
ERROR:CRASH: 3   DefoldFontUAFCrashDemo              0x000000010498037c _ZN12dmGameSystemL19JobPostProcessGlyphEP10JobContexty15JobSystemStatusPvS3_i + 76
ERROR:CRASH: 4   DefoldFontUAFCrashDemo              0x0000000104aa47b8 JobSystemUpdate + 504
ERROR:CRASH: 5   DefoldFontUAFCrashDemo              0x000000010490300c _ZN8dmEngine4StepEPNS_6EngineE + 788
ERROR:CRASH: 6   DefoldFontUAFCrashDemo              0x0000000104903854 _Z14dmEngineUpdatePN8dmEngine6EngineE + 28
ERROR:CRASH: 7   DefoldFontUAFCrashDemo              0x0000000104aad6c8 -[AppDelegate appUpdate] + 36
ERROR:CRASH: 8   QuartzCore                          0x00000001a0e6caf4 740507EF-68D4-3950-ABCE-8F6FF6EA12A1 + 781044
ERROR:CRASH: 22  DefoldFontUAFCrashDemo              0x0000000104904460 _ZN8dmEngine7RunLoopEPKNS_13RunLoopParamsE + 44
ERROR:CRASH: 23  DefoldFontUAFCrashDemo              0x000000010490441c _Z11engine_mainiPPc + 104
```

Symbolicated production stack trace (matches exactly):

```text
dmGameSystem::ResFontAddGlyph        res_font.cpp:813
dmGameSystem::JobPostProcessGlyph    fontgen.cpp:325
ProcessFinishedJobs                  jobsystem.cpp:594
JobSystemUpdate                      jobsystem.cpp:709
dmEngine::Step                       engine.cpp
dmEngineUpdate
-[AppDelegate appUpdate]
```

**Additional context:**

The crash is also triggered by the production app termination path on iOS:

```text
  → ShutdownEngine (AppDelegate.m:<...>)
    → dmEngine::Delete (engine.cpp:398)
      → dmGameObject::DeleteCollection
        → dmGameSystem::CompCollectionProxyDestroy
          → dmResource::DeletePreloader
            → dmResource::UpdatePreloader / PostCreateUpdateOneItem
              → dmGameSystem::ResFontPostCreate / ResFontAddGlyph
                → JobSystemUpdate → ProcessFinishedJobs
                  → JobPostProcessGlyph(freed_ptr) → EXC_BAD_ACCESS
```

The bug requires runtime font generation to be active (`[font] runtime_generation = 1` in `game.project`) and a font large enough that rasterization outlasts one frame (~16ms at 60fps). SDF fonts at 256pt or larger are reliably affected.

AGulev:
Prevents pending dynamic font glyph jobs from using freed font job data during resource reload or destruction. Adds a regression test for canceling pending dynamic font jobs.

Fix https://github.com/defold/defold/issues/12437


AGulev:
@codex review

engine/dlib/src/dlib/jobsystem.cpp
```diff
@@ -324,26 +326,10 @@ void* JobSystemGetData(HJobContext context, HJob hjob)
...
-    if (item->m_Status == JOBSYSTEM_STATUS_FINISHED)
-    {
-        return JOBSYSTEM_RESULT_OK;
-    }
-
-    // Can only cancel queued/created items directly, but still wait on children when already canceled
-    assert(item->m_Status == JOBSYSTEM_STATUS_CREATED || item->m_Status == JOBSYSTEM_STATUS_QUEUED || item->m_Status == JOBSYSTEM_STATUS_CANCELED);
-
-    JobSystemResult result = JOBSYSTEM_RESULT_CANCELED;
+static JobSystemResult CancelJobInternal(JobThreadContext* ctx, HJob hjob);
```
JCash, line 329: This forward declaration is redundant, as noone is calling it between this line and and line 356.

engine/dlib/src/dlib/jobsystem.cpp
```diff
@@ -364,6 +350,29 @@ static JobSystemResult CancelJobInternal(JobThreadContext* ctx, HJob hjob)
...
+    return result;
+}
+
+static JobSystemResult CancelJobInternal(JobThreadContext* ctx, HJob hjob)
+{
+    JobItem* item = GetJobItem(ctx, hjob);
+    if (!item)
+        return JOBSYSTEM_RESULT_INVALID_HANDLE;
+
+    if (item->m_CallbackRunning || item->m_Status == JOBSYSTEM_STATUS_PROCESSING)
```
JCash: I believe m_CallbackRunning should be a state, i.e. a `JOBSYSTEM_STATUS_CALLBACK`.

engine/dlib/src/dlib/jobsystem.cpp
```diff
@@ -324,26 +326,10 @@ void* JobSystemGetData(HJobContext context, HJob hjob)
...
-        return JOBSYSTEM_RESULT_OK;
-    }
-
-    // Can only cancel queued/created items directly, but still wait on children when already canceled
-    assert(item->m_Status == JOBSYSTEM_STATUS_CREATED || item->m_Status == JOBSYSTEM_STATUS_QUEUED || item->m_Status == JOBSYSTEM_STATUS_CANCELED);
-
-    JobSystemResult result = JOBSYSTEM_RESULT_CANCELED;
+static JobSystemResult CancelJobInternal(JobThreadContext* ctx, HJob hjob);
 
+static JobSystemResult CancelChildJobsInternal(JobThreadContext* ctx, JobItem* item, JobSystemResult result)
```
JCash: I'm not sure why we need a "child" specialization?
Two code paths makes it more confusing imo.

Also, the child jobs may very well be calling their callbacks, which will happen before the parent call, but the CancelChildJobsInternal function doesn't have any code changes?

engine/dlib/src/dlib/jobsystem.cpp
```diff
@@ -584,10 +594,15 @@ static void ProcessFinishedJobs(HJobContext context, jc::RingBuffer<HJob>& items
             }
 
             item = *_item;
+            run_callback = item.m_Job.m_Callback && !item.m_SkipCallback;
```
JCash: I believe it should be enough with checking if the state is `CANCELED` here and if so, set `item.m_Job.m_Callback = 0`

engine/dlib/src/dlib/jobsystem.cpp
```diff
@@ -55,6 +55,8 @@ struct JobItem
     int32_atomic_t  m_NumChildrenCompleted;
 
     JobSystemStatus m_Status;
+    bool            m_SkipCallback;
+    bool            m_CallbackRunning;
```
JCash: This is a "hot struct", which we want to keep as small as possible. E.g. we'd use a bitfield from a `uint8_t`.

However, I feel we already have all the states we need to keep track of this. See comments below.

JCash (changes requested):
I think we can reuse the current state flag, and also avoid two separate code paths (which I also feel isn't covered in the tests) relating to cancelling parent vs children.

engine/dlib/src/dlib/jobsystem.cpp
AGulev: I simlified function to one codepath

AGulev:
@codex review

engine/dlib/src/dlib/jobsystem.cpp
```diff
@@ -584,6 +592,10 @@ static void ProcessFinishedJobs(HJobContext context, jc::RingBuffer<HJob>& items
             }
 
             item = *_item;
+            if (item.m_Job.m_Callback)
+            {
+                _item->m_Status = JOBSYSTEM_STATUS_CALLBACK;
```
JCash: Use the `item` we just stored the variable into (for easier reading)

engine/dlib/src/dlib/jobsystem.cpp
```diff
@@ -330,19 +330,22 @@ static JobSystemResult CancelJobInternal(JobThreadContext* ctx, HJob hjob)
...
+
+    // Finished jobs can still be waiting in the done queue. Clear the callback
+    // before cancellation reports complete so owners can release callback state.
+    assert(item->m_Status == JOBSYSTEM_STATUS_CREATED || item->m_Status == JOBSYSTEM_STATUS_QUEUED || item->m_Status == JOBSYSTEM_STATUS_CANCELED || item->m_Status == JOBSYSTEM_STATUS_FINISHED);
 
-    JobSystemResult result = JOBSYSTEM_RESULT_CANCELED;
+    if (was_finished)
+    {
+        item->m_Job.m_Callback = 0;
+    }
```
JCash: While I suggested setting Callback = 0, I also think this becomes confusing imho.

E.g. if you get a callback with "cancelled", you should also get a callback with "finished". I think the inconsistency is troublesome.

It _is_ a new system (and already in dmSDK!), so I don't want to rush into things we may regret later.

I can see two flavors:
* Caller wishes to cleanup immediately, without callback(), and do it's own manual cleanup.
* Caller simply wants to cancel, and let the callback() do the cleanup.

Perhaps we should go with a "setting" like you mentioned, but as a separate function `JobSystemCancelJobNoCallback(HJobContext context, HJob job)´, that sets the callback to 0 (and all its children).
(I don't want to introduce an "struct Options" for this)

It would then call `CancelJobInternal(JobThreadContext* ctx, HJob hjob, bool clear_callback)`.

engine/dlib/src/dlib/jobsystem.cpp
JCash: Also, the if-statement can be removed.
I think it's fine to set status as "callback" even if there is no pointer.

engine/dlib/src/dmsdk/dlib/jobsystem.h
```diff
@@ -67,6 +68,7 @@ enum JobSystemStatus
     JOBSYSTEM_STATUS_PROCESSING       = 3,
     JOBSYSTEM_STATUS_FINISHED         = 4,
     JOBSYSTEM_STATUS_CANCELED         = 5,
+    JOBSYSTEM_STATUS_CALLBACK         = 6,
```
JCash: They are ordered chronologically, so "callback" should be after "finished".

engine/dlib/src/dlib/jobsystem.cpp
```diff
@@ -330,19 +330,22 @@ static JobSystemResult CancelJobInternal(JobThreadContext* ctx, HJob hjob)
...
 
-    if (item->m_Status == JOBSYSTEM_STATUS_PROCESSING)
+    if (item->m_Status == JOBSYSTEM_STATUS_PROCESSING || item->m_Status == JOBSYSTEM_STATUS_CALLBACK)
     {
         return JOBSYSTEM_RESULT_PENDING;
     }
-    if (item->m_Status == JOBSYSTEM_STATUS_FINISHED)
-    {
-        return JOBSYSTEM_RESULT_OK;
-    }
```
JCash, line 340: GIven the suggestions below, we should be able to put this block back, to make it easier to read.

JCash (changes requested):
While it is what we talked about, I think we can simplify things further.

engine/dlib/src/dlib/jobsystem.cpp
AGulev: fixed
but status set into the original value, not the copy

engine/dlib/src/dlib/jobsystem.cpp
AGulev: sounds good, fixed

engine/dlib/src/dmsdk/dlib/jobsystem.h
AGulev: fixed

engine/dlib/src/dlib/jobsystem.cpp
AGulev: We can t restore that early return because a finished parent may still have child callbacks pending/running.

engine/dlib/src/dlib/jobsystem.cpp
JCash, line 421: Are these changes necessary now?
I.e. the "if (!was_finished)" etc?
To me it still looks like we can early out if the task is finished?
```c++
    if (item->m_Status == JOBSYSTEM_STATUS_FINISHED)
    {
        return JOBSYSTEM_RESULT_OK;
    }
```

JCash (changes requested):
Still simplifications that can be made imho.

engine/dlib/src/dlib/jobsystem.cpp
JCash: Only if we're about to reset the callback then?

AGulev:
@codex review

engine/dlib/src/dmsdk/dlib/jobsystem.h
```diff
@@ -218,10 +225,29 @@ enum JobSystemResult JobSystemPushJob(HJobContext context, HJob job);
  *     dmTime::Sleep(1000);
  *     jr = JobSystemCancelJob(job_context, hjob);
  * }
+ * // If the loop observed PENDING before this point, INVALID_HANDLE means the
+ * // job was flushed by JobSystemUpdate() between polls.
```
AGulev, line 232: now we return pending while callback running. but right after that we remove job so it immidiatly become `INVALID_HANDLE` the next poll

engine/dlib/src/dlib/jobsystem.cpp
```diff
@@ -365,17 +354,88 @@ static JobSystemResult CancelJobInternal(JobThreadContext* ctx, HJob hjob)
...
     return result;
 }
 
+static JobSystemResult CancelJobNoCallbackInternal(JobThreadContext* ctx, HJob hjob)
+{
+    JobItem* item = GetJobItem(ctx, hjob);
+    if (!item)
+        return JOBSYSTEM_RESULT_INVALID_HANDLE;
+
+    item->m_Job.m_Callback = 0;
```
AGulev, line 394: drop callbacks before freeing the data

engine/gamesys/src/gamesys/resources/res_font.cpp
```diff
@@ -111,24 +118,60 @@ namespace dmGameSystem
...
+                continue;
             }
 
+            QueueCancelledCallback(job_info, callbacks);
             DecRefJobResourceInfo(font->m_Factory, font, job_info);
             DeallocateJobResourceInfo(job_info);
         }
-        font->m_PendingJobs.SetSize(0);
+
+        for (uint32_t i = 0; i < callbacks.Size(); ++i)
```
AGulev, line 163: It keeps the user-facing callback for canceled `font.prewarm_text()` requests, but runs it only after the pending jobs are canceled

engine/gamesys/src/gamesys/resources/res_font.cpp
```diff
@@ -212,18 +256,27 @@ namespace dmGameSystem
...
 
-    dmResource::Result ResFontPrewarmText(FontResource* resource, const char* text, FPrewarmTextCallback cbk, void* cbk_ctx)
+    static dmResource::Result ResFontPrewarmTextInternal(FontResource* resource, const char* text, FPrewarmTextCallback cbk, void* cbk_ctx, bool callback_on_cancel)
     {
         if (!resource->m_IsDynamic)
         {
             return dmResource::RESULT_NOT_SUPPORTED;
         }
 
+        if (resource->m_CancelingPendingJobs)
```
AGulev: block new font jobs during teardown

AGulev (commented):
I think it's ready for review now

engine/dlib/src/dmsdk/dlib/jobsystem.h
```diff
@@ -218,10 +228,29 @@ enum JobSystemResult JobSystemPushJob(HJobContext context, HJob job);
...
  * ```
  */
 enum JobSystemResult JobSystemCancelJob(HJobContext context, HJob job);
 
+/*# cancel a job (and its children) without invoking callbacks
+ * @note Cancelled jobs will be flushed at the next JobSystemUpdate()
```
JCash: This note doesn't serve a purpose for the user and may be confusing?

engine/gamesys/src/gamesys/resources/res_font.cpp
```diff
@@ -111,24 +118,60 @@ namespace dmGameSystem
...
+        }
+
+        dmArray<PendingCancelCallback> callbacks;
+        callbacks.SetCapacity(font->m_PendingJobs.Size());
+
+        font->m_CancelingPendingJobs = 1;
+        while (!font->m_PendingJobs.Empty())
         {
-            FontJobResourceInfo* job_info = font->m_PendingJobs[i];
+            FontJobResourceInfo* job_info = font->m_PendingJobs[font->m_PendingJobs.Size() - 1];
```
JCash: 1) Why are we looping backwards? I.e. why not keep the for-loop?
2) I see no call to SetSize() here to "step" the loop. We shouldn't disguise it in other functions.

engine/gamesys/src/gamesys/resources/res_font.cpp
JCash: I think the name "canceling" is not same as "destroying" or similar. And canceling a prewarm request doesn't warrant "RESULT_NOT_LOADED".

I suggest something like "destroying" or similar name.

engine/gamesys/src/gamesys/scripts/script_font.cpp
```diff
@@ -173,8 +182,7 @@ static void PrewarmTextCallback(void* _ctx, int result, const char* errmsg)
 
         dmScript::TeardownCallback(cbk);
     }
-    dmScript::DestroyCallback(cbk); // only do this if you're not using the callback again
-    delete ctx;
+    DestroyPrewarmTextCallbackContext(ctx); // only do this if you're not using the callback again
```
JCash: Was there an actual problem here?
To me it looks like defensive codeing by an llm?
Remember, we check data early, not at each possible usage (like llm's often want to)

And, the previous code was jsut fine, no need for an extra function for that.

All in all, we should keep an eye on code size.

JCash (changes requested):
Overall good, but I'd like some cleanup.

engine/gamesys/src/gamesys/resources/res_font.cpp
AGulev: it's because we are removing elements from the array here using `RemovePendingJob`

AGulev: I'm fixing it without an attempt to reuse `RemovePendingJob`

engine/gamesys/src/gamesys/scripts/script_font.cpp
AGulev: you're right, I missed that. there is only one additional place where it's needed

engine/gamesys/src/gamesys/resources/res_font.cpp
```diff
@@ -111,24 +117,64 @@ namespace dmGameSystem
...
             while (JOBSYSTEM_RESULT_PENDING == jr)
             {
                 dmTime::Sleep(1000);
-                jr = JobSystemCancelJob(font->m_Jobs, hjob);
+                jr = JobSystemCancelJobNoCallback(font->m_Jobs, hjob);
+            }
+
+            // If JobSystemUpdate() ran a callback while we were waiting, the
+            // callback already removed and destroyed this job info.
+            if (i == font->m_PendingJobs.Size() || font->m_PendingJobs[i] != job_info)
```
JCash: This is a strange mistmatch though. I see two cases:
1) JobSystemUpdate() is called don main thread, and so is this function.
2) if they are called on different threads, I see no synchronization points (i.e. mutex) to make it safe to read/write the same set of data (m_PendingJobs)

engine/gamesys/src/gamesys/scripts/script_font.cpp
```diff
@@ -217,36 +216,39 @@ static int PrewarmText(lua_State* L)
...
     {
         return DM_LUA_ERROR("Failed to get font %s: %d", dmHashReverseSafe64(fontc_path_hash), r);
     }
 
+    dmGameSystem::FPrewarmTextCallback callback = 0;
+    CallbackContext* cbk_ctx = 0;
+    if (top > 2 && lua_isfunction(L, 3))
+    {
+        dmScript::LuaCallbackInfo* luacbk = dmScript::CreateCallback(L, 3);
+        if (luacbk)
```
JCash: Open question: Not sure we need to check the result here, as I'd imagine failing to create the callback should raise a luaL_error() ?

JCash (changes requested):
a new threading question needs to be either be cleared up or simplified.

JCash (approved):

PR #12445, opened 2026-05-25
Editor: Skip disabled menu items in context menus

### LINKED ISSUE #12439: Skip disabled context menu items when navigating with keyboard
### The problem
When navigating in context menus using keyboard (arrow keys), the focus currently moves to disabled/grayed-out items. This creates friction in keyboard-only workflows, as users must press additional keys to skip items they cannot interact with.
For example, when navigating a context menu with 10 items where 3 are disabled, users waste time cycling through non-actionable options.

### Solution
When navigating context menus with arrow keys (Up/Down), the focus should automatically skip over disabled items and move directly to the next enabled menu item. This matches the behavior found in most modern applications like VS Code.

### Alternatives
1. Visual indicators only - keeping current behavior, but making disabled items more obviously non-interactive (doesn't solve the navigation issue)
2.  Adding a modifier key (e.g., Ctrl+Arrow) to skip disabled items - adds complexity and isn't discoverable

JosephFerano:
Fixes #12439

### DIFF
--- a/editor/src/clj/editor/ui.clj
+++ b/editor/src/clj/editor/ui.clj
@@ -1476,40 +1476,67 @@
   {:control control
    :menu-id menu-id})
 
-(defn- next-focusable-menu-item [^ObservableList items ^EventTarget target-node forward]
+(defn- focusable-menu-item? [^MenuItem menu-item]
+  (and (.isVisible menu-item)
+       (not (.isDisable menu-item))
+       (not (instance? CustomMenuItem menu-item))))
+
+(defn- current-menu-item? [^MenuItem menu-item ^EventTarget target-node]
+  (let [node (.getStyleableNode menu-item)]
+    (or (and node (.isFocused node))
+        (and node (nodes-along-path? target-node node nil)))))
+
+(defn- next-focusable-menu-item [^ObservableList items ^EventTarget target-node]
   (let [item-count (.size items)]
     (loop [index 0
            first-focusable nil
-           last-focusable nil
-           previous-focusable nil
-           current-found false
-           next-focusable nil]
+           current-found false]
       (if (< index item-count)
         (let [menu-item ^MenuItem (.get items index)
-              node (.getStyleableNode menu-item)
-              target-item (and node (nodes-along-path? target-node node nil))
-              current (or target-item (and node (.isFocused node)))
-              focusable (and (.isVisible menu-item)
-                             (not (.isDisable menu-item))
-                             (not (instance? CustomMenuItem menu-item)))]
-          (if (and target-item (instance? CustomMenuItem menu-item))
+              current (current-menu-item? menu-item target-node)
+              focusable (focusable-menu-item? menu-item)]
+          (cond
+            (and current (instance? CustomMenuItem menu-item))
             ::custom-menu-item
+
+            (and current-found focusable)
+            menu-item
+
+            :else
             (recur (inc index)
                    (or first-focusable (when focusable menu-item))
-                   (if focusable menu-item last-focusable)
-                   (if current last-focusable previous-focusable)
-                   (or current-found current)
-                   (or next-focusable (when (and current-found focusable) menu-item)))))
-        (if current-found
-          (if forward
-            (or next-focusable first-focusable)
-            (or previous-focusable last-focusable))
-          (if forward first-focusable last-focusable))))))
+                   (or current-found current))))
+        first-focusable))))
+
+(defn- previous-focusable-menu-item [^ObservableList items ^EventTarget target-node]
+  (let [item-count (.size items)]
+    (loop [index (dec item-count)
+           last-focusable nil
+           current-found false]
+      (if (<= 0 index)
+        (let [menu-item ^MenuItem (.get items index)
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/ui.clj
```diff
@@ -1572,70 +1634,73 @@
...
-                                                                (:style child))))
-                                     :graphic {:fx/type image-icon
-                                               :path (:icon child)
-                                               :size 16.0}}))))
-                            category-items))}))
-                   column))}))}}))))
+                              (update-vals #(localization/natural-sort-by-label @localization %)))
+        grid-menu
+        (fx/instance
+          (fx/create-component
```
JosephFerano: Indentation so I can do the doto user-data! at the end.

vlaaad (commented):
Looks good, a couple of comments:
1. I find it strange that we special-case `::grid-menu`. What if we add another custom menu? Should we also call it `::grid-menu` even if it's not using grid? Or should we then add one more special case here? I'm not sure if it's feasible, but perhaps we could replace `::grid-menu` checks with `CustomMenuItem` checks somehow?
2. `next-focusable-menu-item` looks like it could be a bit simpler. Just a feeling that we might not need this many variables in `loop`. Maybe? Not sure.
3. We install the focus filter to context menu items, but not to menu buttons (see `register-button-menu` below). As a result, the context menus of menu buttons still traverse disabled menu items:

https://github.com/user-attachments/assets/cbcb31f9-c1cf-462e-8acf-3ea3193b2db1

(not clear from the video, but I'm pressing Down multiple times)

JosephFerano:
> 1. I find it strange that we special-case `::grid-menu`. What if we add another custom menu? Should we also call it `::grid-menu` even if it's not using grid? Or should we then add one more special case here? I'm not sure if it's feasible, but perhaps we could replace `::grid-menu` checks with `CustomMenuItem` checks somehow?

Yes, a CustomMenuItem check was sufficient, I didn't realize the item I was tagging was just a CustomMenuItem, I thought it was some other type :facepalm: 

> 2. `next-focusable-menu-item` looks like it could be a bit simpler. Just a feeling that we might not need this many variables in `loop`. Maybe? Not sure.

I think it came out complicated because I tried to unify the traversal logic. It's much cleaner now as two separate functions for next/prev.

> 3. We install the focus filter to context menu items, but not to menu buttons (see `register-button-menu` below). As a result, the context menus of menu buttons still traverse disabled menu items:

Ah great, didn't know about this one. Added!

JosephFerano:
> I find it strange that we special-case ::grid-menu. What if we add another custom menu? Should we also call it ::grid-menu > even if it's not using grid? Or should we then add one more special case here? I'm not sure if it's feasible, but perhaps we > could replace ::grid-menu checks with CustomMenuItem checks somehow?

Actually turns out it was installing the filters, so instead I'm adding a tag that better explains the semantics.

editor/src/clj/editor/ui.clj
```diff
@@ -1475,13 +1476,112 @@
...
+            (and current-found focusable)
+            menu-item
+
+            :else
+            (recur (dec index)
+                   (or last-focusable (when focusable menu-item))
+                   (or current-found current))))
+        last-focusable))))
+
+(defn- focus-next-menu-item! [^ContextMenu context-menu ^KeyEvent event forward]
```
vlaaad: I'd name it differently, without mentioning a specific direction (next), because it's not necessary next. `focus-traverse-menu-item!`?

editor/src/clj/editor/ui.clj
```diff
@@ -1475,13 +1476,112 @@
...
+            ::custom-menu-item
+
+            (and current-found focusable)
+            menu-item
+
+            :else
+            (recur (dec index)
+                   (or last-focusable (when focusable menu-item))
+                   (or current-found current))))
+        last-focusable))))
```
vlaaad: They are nearly identical! 🤔

vlaaad (dismissed):
LGTM! See some minor code style comments

editor/src/clj/editor/ui.clj
JosephFerano: Fair enough

editor/src/clj/editor/ui.clj
```diff
@@ -1475,13 +1476,86 @@
...
+              (.requestFocusOnIndex context-menu-content (.indexOf (.getItems context-menu) menu-item))))
+          true)))))
+
+(defn- skip-disabled-menu-item-focus-filter? [^ContextMenu context-menu]
+  (let [items (.getItems context-menu)
+        item-count (.size items)]
+    (loop [index 0]
+      (when (< index item-count)
+        (or (true? (user-data (.get items index) ::skip-disabled-menu-item-focus-filter))
+            (recur (inc index)))))))
```
vlaaad: Looks like a simple iteration, maybe we could use `coll/any?` or something here?

vlaaad (dismissed):

editor/src/clj/editor/ui.clj
JosephFerano: Yes, much nicer

vlaaad (approved):
Ship it!

PR #12450, opened 2026-05-25
Added graphics.webgl_version_hint

### LINKED ISSUE #12397: Allow host pages to opt out of dmloader.js defaults via pre-set Module.* properties
## Problem

`dmloader.js` is a black box from the host page's perspective: any property the host pre-sets on `Module` is silently discarded when `dmloader.js`'s own `var Module = {...}` redefines the global. Six different settings that a WebView wrapper has a legitimate need to control are therefore unreachable without rewriting the bundle byte-for-byte on serve.

The settings, all currently unreachable from the host:

1. **`Module.isWASMPthreadSupported`** — `dmloader.js` derives this from `crossOriginIsolated` + `SharedArrayBuffer` and overwrites unconditionally. A WebView wrapper that serves the bundle from a custom URL scheme can pass that probe but still can't construct same-origin Workers, so pthread WASM is selected but then fails to load.
2. **WebGL2 vs WebGL1** — the embedded GLES driver under a WebView may advertise WebGL2 capabilities it can't actually implement; the engine's VAO / instancing init then trips inside the driver. The host can't ask for a WebGL1 fallback today.
3. **WebGL context attributes** — `preserveDrawingBuffer` is off by default. Hosts behind a flaky compositor (which can fail `eglSwapBuffers` between renders) need it on so the canvas content survives the failed swap.
4. **WebGL extension filtering** — some embedded GLES drivers falsely advertise compressed-texture / float-texture / depth-texture extensions and then reject the actual upload, crashing the engine's basis-universal transcoder. The host can't strip the advertised names.
5. **Footer button strip** — the engine_template emits `<div class="buttons-background">` for the Fullscreen / Made-with-Defold anchors. `[html5] show_fullscreen_button = 0` + `show_made_with_defold = 0` suppresses the inner anchors but leaves a 42 px white bar behind that breaks true-fullscreen layouts.
6. **WebGL context-loss recovery** — WebView hosts can drop the WebGL backbuffer when the host pauses or re-attaches the view; the engine renderer has no recovery path so frames render black. The host has no hook to install loss/restore handlers.

## Concrete reproducer

A WebView wrapper that serves the bundle from a custom URL scheme (`myscheme://`) hits at least three of these at once on the OHOS emulator's software-GLES stack:

* **pthread Worker construction**: `crossOriginIsolated` + `SharedArrayBuffer` are present, so dmloader picks the pthread WASM. Chromium's separate Worker origin check then rejects `new Worker('myscheme://.../*_pthread_wasm.js')`:
  ```
  Uncaught SecurityError: Failed to construct 'Worker': Script at 'myscheme://localhost/...' cannot be accessed from origin 'null'.
  ```
  Reproduced regardless of how permissively the scheme is registered (standard + local + secure + CORS-enabled).

* **WebGL2 VAO crash**: with the pthread issue worked around, the engine reaches WebGL2 init and aborts inside `dmArray::operator[]` on the first frame — the GLES driver advertised WebGL2 but the actual VAO binding path fails. Falling back to WebGL1 lets the engine continue past first frame.

* **Compositor blink**: with both above worked around, the canvas renders but `eglSwapBuffers` fails on every frame at the OS Render Service level (`EGL_BAD_SURFACE, g_handle is null`). The host can't fix the compositor, but `preserveDrawingBuffer:true` would keep the WebGL backbuffer stable between attempts.

These are not Defold bugs — the engine is correct in all three cases. They're driver / compositor weaknesses that the host knows about and would handle if dmloader gave it a way to.

## Current workaround

Rewrite the served `dmloader.js` byte-for-byte (e.g. via an `onInterceptRequest` hook) to flip the relevant lines. Fragile, has to be redone every Defold release, and gets unmaintainable fast as the number of toggles grows.

## Proposal

Let the host page pre-set the relevant properties on `Module` (the standard Emscripten pattern of seeding `Module` before the loader script tag). `dmloader.js` captures each into a top-level `DEFOLD_HOST_*` variable before its own `var Module = {...}` redefinition runs, then honours each at the matching apply site:

| Host-set property                                | dmloader behaviour                                                                                   |
| ------------------------------------------------ | ---------------------------------------------------------------------------------------------------- |
| `Module.isWASMPthreadSupported = false`          | Force the single-threaded WASM variant; skip the `crossOriginIsolated` + `SharedArrayBuffer` probe. |
| `Module.isWebGL2Supported = false`               | Wrap `HTMLCanvasElement.getContext` so any `'webgl2'` request returns `'webgl'`.                     |
| `Module.webGLContextAttributes = { ... }`        | `Object.assign`'d into the attrs argument of the GL context creation call.                           |
| `Module.webGLExtensionFilter = (n) => boolean`   | Wrap `WebGLRenderingContext.prototype.{getSupportedExtensions, getExtension}` to filter.             |
| `Module.showButtonStrip = false`                 | Inject a `<style>` tag that hides `.buttons-background` at `DOMContentLoaded`.                       |
| `Module.autoReloadOnWebGLContextRestore = true`  | Attach `webglcontextlost` / `webglcontextrestored` listeners on the canvas; reload on restored.      |

All six share the same plumbing: capture-before-redefinition, apply-at-IIFE / apply-at-probe-site. Default behaviour is byte-identical when nothing is pre-set — each capture yields `undefined` / `null` / `false` and the matching apply block skips.

PR #12396 implements all six.

## Latest status (2026-05-25)

* For hosts that control the build, bundling HTML5 without pthread support is now confirmed as the cleanest fix for the original Worker-selection problem. With stock `bob.jar --platform js-web --architectures wasm-web`, the official Defold build path produces a single-threaded bundle (`<game>.wasm` + `<game>_wasm.js`, no `*_pthread*`, and `dmloader.js` emits `Module["isWASMPthreadSupported"] = false ...`). My OHOS HAP path now uses that and no longer rewrites `dmloader.js` for the pthread case.
* The runtime opt-out in this issue is still useful for black-box bundle consumers and for teams shipping one generic HTML5 bundle to multiple hosts, but for a controlled OHOS build the build-time `wasm-web` choice is preferred.
* I also tested engine-side mitigation in a Defold fork: a real WebGL2 probe before committing the renderer, plus runtime validation of reported texture capabilities. Those reduce Defold-owned failure modes, but on the OHOS emulator they were not sufficient by themselves.
* The current OHOS emulator still requires the host-side equivalents of `Module.isWebGL2Supported = false` and `Module.webGLExtensionFilter = ...`. Without that host shim, the emulator could go fully offline after launch; the emulator-side logs showed compressed-texture / ASTC failure immediately before the drop (`decode_compressed_texture: astc format` / `decode astc image error`) and the crash monitor recorded `Client heartbeat lost, exit` with Windows code `3221226505`.
* After reintroducing the host-side WebGL1 downgrade + extension filter in the OHOS wrapper, the HAP stayed up through repeated multi-minute soaks and `hdc` remained connected. That is the strongest current evidence that items 2 and 4 in this issue are still load-bearing in practice even with better engine-side probing.
* The remaining `DGLES` spam (`bind external with nullptr gbuffer`, `createimage with ctx null`) still appears to be below Defold, in ArkWeb/Chromium GPU plus the OHOS emulator's DGLES/compositor stack. I do not currently have evidence that Defold alone can eliminate those messages.

## Affected platforms

Any WebView host whose underlying capabilities differ from what `dmloader.js` assumes at compile time. The concrete cases that motivated each opt-out are documented inline in the PR; the OHOS WebView wrapper hits the most of them at once, but the same pattern occurs on other custom-scheme embeddings (Android `WebView.loadData`, iOS `WKWebView` with `loadFileURL`, embedded WebView2 hosts on Windows).

Default browser embedding via `http(s)://` is unaffected — that path already worked and continues to work unchanged.

## Scope

This issue covers **the full set of host-facing opt-outs that follow the capture-before-redefinition pattern**: pthread, WebGL2, context attrs, extension filter, button strip, context-loss recovery. Each is independently motivated and additively implemented; reviewers can accept or reject individual opt-outs without affecting the others.

Out of scope here:

* A general-purpose `Defold.config(...)` host API — would be a much larger redesign and is its own conversation.
* Auto-detection of any of these conditions at runtime (Worker constructability, WebGL2 stability, swap-buffer health). The host knows its environment better than dmloader can probe for it; explicit opt-outs are simpler and more predictable.
* Cumulative changes to `engine_template.html` to expose new game.project settings for the same toggles. The button-strip case in particular could alternatively become a `[html5] show_button_strip` toggle; the loader-side opt-out is a strict superset (works for hosts that can't change game.project).

britzl:
Added `graphics.webgl_version_hint` to be able to select a WebGL 1 context even when the device supports WebGL 2. This may be useful in scenarios when the WebGL 2 implementation or reported extensions contain bugs or inconsistencies.

Fixes #12397


engine/platform/src/dmsdk/platform/window.h
```diff
@@ -223,7 +223,7 @@ typedef struct WindowCreateParams
     uint8_t                 m_Fullscreen                : 1;
     uint8_t                 m_PrintDeviceInfo           : 1;
     uint8_t                 m_HighDPI                   : 1;
-    uint8_t                                             : 4;
+    uint8_t                 m_WebGLVersionHint          : 4;
```
JCash: I think we should reuse the m_OpenGLVersionHint here, and not create more graphics driver dependencies for the WindowCreationParams

JCash (dismissed):
Overall good
Would like to avoid the ne field in the window creation parameters though.

engine/platform/src/dmsdk/platform/window.h
britzl: I was thinking about this myself. The m_OpenGLVersionHint could be renamed to something more generic such as m_GraphicsApiVersionHint. It might be confusing though since we have `m_GraphicsApiVersionMajorHint` in the `dmGraphics::ContextParams` struct?

JCash (dismissed):

PR #12453, opened 2026-05-26
Support configuring mouse bindings for scene view controls

### LINKED ISSUE #11338: Rotate around pivot/center when holding down right mouse button and moving mouse
When you are in scene view in 3d view , you pan with middle mouse button and the same operation happens with right mouse button. 

I would make right mouse button rotate around pivot/center when holding right button down and moving mouse position.

I don't know any alternatives

### LINKED ISSUE #11506: Editor keymap should support mouse buttons and the missing actions that use mouse added to it
I hate the CTRL+M1 editor camera rotation. Mostly because M1 often triggers before CTRL, which leads to no rotating the camera and thus irritation. I'd prefer to have it on mouse scroll click.

The editor keymap should support mouse buttons and then the missing actions, that use mouse buttons, be added to it. The two I'm aware of are CTRL+M1 and ALT+M1.

*Head empty..*

### LINKED ISSUE #12405: Customizable scene mouse gesture controls
Currently, there is no way for users to customize keybindings for manipulating the scene camera (orbit, pan, zoom, free camera). This task will introduce a new "Scene" tab to the Preferences dialog, that initially will provide a way for users to modify which mouse buttons and modifiers trigger which camera actions. Further configuration for the Scene View can be added later on.

JosephFerano:
The Preferences Dialog Keymap tab now includes bindings for mouse related operations. This allows users to create their own custom mouse + keyboard modifiers (ctrl/alt/shift) for scene related operations, namely mouse controls like zoom/orbit/pan/free-cam. Scenes like Tilesource, Atlas, and Curve View can override default camera behaviors, while Curve Editor and Tilemap allow customization of their special actions like adding control points on curves and copying/erasing tiles.

image
image

Fixes #12405
Fixes #11506
Fixes #11338

### Technical changes

A new `mouse-binding` was added to avoid messing with the internals of `keymap` and `handler`. For `keymap`, a lot of the logic was tied to `KeyCombination` and other keyboard centric data structures that didn't handle mouse + modifiers all that well. `handler` also supported many features not needed for these mouse-bindings, and was missing functionality to register additional sub-contexts.

The mouse-bindings have a fallback system that improves UX for the user, allowing them to configure camera controls to be applied globally if they change `Scene 2D Camera` but still allow scene specific camera controls to override the defaults. In other words, Tilemap can have it's own custom "Pan" binding, but all other scenes would still have the default. But also the default "Pan" on `Scene 2D camera` can be changed to something else and everyone else would inherit.

A smaller feature was the "modifiers" binding, which is slightly different, in particular how it's presented to the user, as it only allows a single modifier to be pressed. This can be enhanced by also allowing any keyboard registration, but it was kept simple for now.

There was one area of investigation that was started and that's re-ordering the camera/tools during input handling dispatch. Right now, camera goes first then tools, but having tools then camera solves certain issues (while potentially introducing others not yet discovered). Currently, it's having to handle special edge cases with the camera to make sure the camera handles custom camera controls used in tools. The changes were stashed but this can still be investigated in the future as it might simplify the dispatch handling and reduce "camera aware" logic.


editor/src/clj/editor/prefs_dialog.clj
```diff
@@ -308,42 +418,304 @@
...
     (let [^TableView table-view (.getSource e)]
-      (when-let [command (.getSelectedItem (.getSelectionModel table-view))]
-        (show-new-shortcut-dialog! swap-state command (.getWindow (.getScene table-view)))))))
+      (when-let [row (.getSelectedItem (.getSelectionModel table-view))]
+        (show-binding-dialog-for-row! swap-state row (.getWindow (.getScene table-view)))))))
+
+(defn- handle-new-shortcut-action [swap-state command ^ActionEvent e]
+  (show-binding-dialog-for-row! swap-state {:kind :keyboard
+                                            :command command}
+                              (.getOwnerWindow (.getParentPopup ^MenuItem (.getSource e)))))
```
matgis: Whitespace fix:
```suggestion
                                (.getOwnerWindow (.getParentPopup ^MenuItem (.getSource e)))))
```

editor/src/clj/editor/camera.clj
```diff
@@ -966,8 +956,20 @@
...
+          movement (cond
+                     (and mouse-binding-command (not override-camera))
+                     (camera-command->movement mouse-binding-command)
+                     (= type :mouse-pressed)
+                     (let [command (mouse-binding/command-for-action (camera-mouse-binding-context local-cam)
+                                                                     action)
+                           movement (camera-command->movement command)]
+                       (or (and movement (movements-enabled movement) movement)
+                           :idle))
+                     :else
```
matgis, line 976: Style: This would read nicer if there were newlines between each pair of condition + result.

editor/src/clj/editor/input.clj
```diff
@@ -189,3 +189,25 @@
...
            (trackable-key? (:key-code action)))
       (update :pressed-keys disj (:key-code action)))))
+
+(defn- active-modifiers [x]
+  (into #{}
+        (filter x)
+        [:shift :alt :control]))
+
+(defn- modifiers-match? [expected actual]
+  (= (set expected)
```
matgis: It seems a bit inefficient to create a `set` from `expected` here each time. Could we maybe convert them to sets earlier, I.e. in `mouse-binding/register!` and `mouse-binding/set-user-overrides!`?

editor/src/clj/editor/mouse_binding.clj
```diff
@@ -0,0 +1,333 @@
...
+
+(set! *warn-on-reflection* true)
+
+(def buttons [:primary :middle :secondary])
+(def modifiers [:shift :control :alt])
+
+(def button->label
+  {:primary "Left"
+   :middle "Middle"
+   :secondary "Right"})
```
matgis: Do we need to localize these?

editor/src/clj/editor/mouse_binding.clj
```diff
@@ -0,0 +1,333 @@
...
+;; https://www.defold.com/license
+;;
+;; Unless required by applicable law or agreed to in writing, software distributed
+;; under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+;; CONDITIONS OF ANY KIND, either express or implied. See the License for the
+;; specific language governing permissions and limitations under the License.
+
+(ns editor.mouse-binding
+  (:require [clojure.string :as string]
+            [editor.input :as i]
```
matgis: Style: `i` seems a bit short. We tend to reserve really short names for things that are used in *a lot* of places in the code. I think we should import `[editor.input :as input]` whenever we need it.

editor/styling/stylesheets/components/_radio-button.scss
```diff
@@ -23,7 +29,7 @@
         -fx-border-radius: 50%;
         -fx-background-image: string.unquote("null");
         -fx-background-color: -df-background;
-        -fx-border-width: 1;
+        -fx-border-width: 2;
```
matgis: I don't think we use any radio buttons anywhere else now, but this change could potentially mess up other UIs that rely on the radio buttons being a certain size, since it applies everywhere.

In general, I think the look of the radio buttons in the new Preference popup is a bit large - they should match the size of the Checkbox control. They are also a bit weird in that they have some sort of 3D-looking bevel when unselected, which is lost when they are selected.

I think we could maybe leave it like this for now and do a proper styling pass on them later.

editor/src/clj/editor/prefs_dialog.clj
```diff
@@ -308,42 +418,304 @@
...
+                :children
+                [{:fx/type fxui/button
+                  :text (localization-state (localization/message "dialog.button.cancel"))
+                  :on-action (fn [_] (swap-state dissoc :mouse-binding-popup))}
+                 {:fx/type fxui/button
+                  :text (localization-state (localization/message "prefs.keymap.mouse-binding.button.apply"))
+                  :on-action (fn [_]
+                               (update-mouse-bindings
+                                 #(mouse-binding/update-command-binding % context command binding-index draft-binding))
+                               (swap-state dissoc :mouse-binding-popup))}]}))}))
```
matgis, line 502: I noticed the Cancel and Apply buttons blended in with the background in the new Mouse Binding popup. It doesn't look bad, but generally we've tried to have buttons appear distinct from the background. I think we can ship it like this, but we should introduce a button style that is suitable for this sort of bright panel background at some point.

matgis (changes requested):
Very nice! Some comments:

Naming: The operation of moving the camera forward / backward is typically referred to as Dolly, not Zoom. In 2D views, it makes sense to call it Zoom, because it is equivalent to adjusting the Zoom ring of the camera lens, but in 3D, where you will typically use a perspective projection, we should call it Dolly since it actually moves the camera along the view axis. It's also possible we would like to be able to adjust the Zoom of the 3D camera (i.e. its FOV) separately from the Dolly in the future. So I would suggest changing the term to Dolly in the preference key as well as in the localized strings. In code, I think Dolly is the better term in general, if I had to pick between naming things `zoom` or `dolly` as a rule.

I think we should maybe unbind the Orbit mode from the 2D modes by default? It's frequently something I do by mistake, and I have to choose Realign Camera from the menu to restore the view and get the rulers back. I don't mind having the binding target there in case someone really loves to observe their guis, atlases and tile maps from an oblique angle, but I think now's an excellent opportunity to fix this minor annoyance.

Related to the above, there doesn't seem to be a way for me to completely unbind a mouse binding from the Preferences dialog?

Regression: It's now possible to Orbit the Curve Editor View using the fallback "Scene 2D Camera -> Orbit" binding. This was blocked before, and is a bit problematic since it messes up the Curve Editor View and can't be restored. We should respect the `movements-enabled` property of the `CameraController` associated with the view.

Niggle: If I've right-clicked without moving the mouse, the Context Menu shows us as you'd expect. However, I have to left-click to close the Context Menu before I can use RMB to pan again. If it's a simple fix, it would be neat if I could dismiss the Context Menu and initiate panning with a single press of the right mouse button.

I would like us to address the renaming of the actions from Zoom to Dolly and fix the Curve Editor View Orbit regression before merging.

JosephFerano:
> I think Dolly is the better term in general, if I had to pick between naming things zoom or dolly as a rule.

:+1: 

> I think we should maybe unbind the Orbit mode from the 2D modes by default?

:+1: 

> Related to the above, there doesn't seem to be a way for me to completely unbind a mouse binding from the Preferences dialog?

You can right click on the list of pills and from the context menu you can "Remove X".

> Regression: It's now possible to Orbit the Curve Editor View using the fallback "Scene 2D Camera -> Orbit" binding.

Ah good catch, I missed this. Fixed.

JosephFerano:
> Niggle: If I've right-clicked without moving the mouse, the Context Menu shows us as you'd expect. However, I have to left-click to close the Context Menu before I can use RMB to pan again. If it's a simple fix, it would be neat if I could dismiss the Context Menu and initiate panning with a single press of the right mouse button.

Ok, it wasn't hard to implement. Done.

JosephFerano:
@matgis 

I did some additional changes, I fixed a regression that was introduced in the original free camera branch where right clicking on an object in the scene view wasn't selecting the object in the scene. I tried to investigate the fact that the context menu wasn't enabling the option to show/hide the selected object, likely because the right click context menu launches before the graph has been able to detect that there's a selected object. Also right clicking around other objects doesn't work, if one is selected it stays selected. So now I'm thinking, was the original behavior that I reverted to desired? I think it's good because it avoids extra clicks, but to make it work nicely I would create a bug report/task to improve this a bit more. Thoughts?

Also did a refactor to swap modifiers everywhere to use a set. The original `action` map was modeling modifiers as boolean entries, which was causing some of these conversion issues. I think a set is a better data structure to model actively pressed modifiers so I changed everything to use that instead and avoid conversions.

Finally, the fallback system for the inheritance created an undesired behavior that I didn't like (free cam mode wasn't activating in atlas/tile map). The mouse-binding runtime didn't really need the concept of the fallbacks because the input dispatching already handled it, so removed it from mouse-binding resolving and it's not a view-only thing so that the prefs_dialog can show the user what is being inherited.

editor/src/clj/editor/tile_map.clj
```diff
@@ -1312,8 +1336,12 @@
...
 
              nil)]
-    (when (seq tx)
-      (g/transact tx)
+    (if (coll/not-empty tx)
+      (do
+        (g/transact tx)
+        nil)
+      ;; If we pass an action instead, we enable right-click context menu
+      ;; action)))
```
matgis: Codex might have found something here. Perhaps we shouldn't be returning `true` here, but the `action` instead?

### Codex:

`scene/dispatch-input` passes each handler's return value to the next handler. Returning `true` here means an unhandled tile-map click, including secondary clicks intended for selection/context menus, reaches `SelectionController` as a boolean instead of the original action, so right-click selection/context menus in tile-map views are blocked. Return `action` on the fallthrough path.

matgis (dismissed):
I'm really happy with how it works!

Regarding the remaining issues around right-clicks you mentioned, I think we can treat them as a separate issue. However, Codex did find something that could perhaps be related to it. I'll let you investigate, and maybe create a separate issue for it.

My only remaining concern was around still not being able to figure out how to remove a mouse binding:

> Mats: here doesn't seem to be a way for me to completely unbind a mouse binding from the Preferences dialog?

> Joe: You can right click on the list of pills and from the context menu you can "Remove X".

I tried this, but for me, the "Remove X" menu item only appears for keyboard bindings, not mouse bindings. Maybe I'm doing something wrong?

Anyway, I'm approving the PR for now!

JosephFerano:
> I'm really happy with how it works!

:heart: 

> Regarding the remaining issues around right-clicks you mentioned, I think we can treat them as a separate issue.

Sure, let me include a screen recording, I should've done that previously

https://github.com/user-attachments/assets/be2a27c0-0b52-4889-8e10-0631e5aaf1e2

If I right click the first time, the context system hasn't been updated it seems. However `Ctrl + e` works. So it just seems like a race condition where the graph hasn't updated the current selection context when we try to populate the context menu.

> My only remaining concern was around still not being able to figure out how to remove a mouse binding:
> 
> I tried this, but for me, the "Remove X" menu item only appears for keyboard bindings, not mouse bindings. Maybe I'm doing something wrong?

Oh, you might be trying to click on the inherited context, which has a "ghost" pill and the tool tip. If that's it, then I should consider improving the discover-ability of how this works. I have a couple of ideas;

- Change the colors to make it more obvious, maybe the ghost pill isn't obvious enough
- Add a disabled tooltip entry, maybe something like "Remove (Inherited)", so the mechanism is even more explicit
- I could remove showing the inheritance altogether if you think it's confusing, so there would be no pill. I like the inheritance system but if you think it would be more confusing, I can remove it, should be quick.

Here's a screen capture;

https://github.com/user-attachments/assets/80effaaa-836d-48d3-aed5-160d57b04777

editor/src/clj/editor/tile_map.clj
JosephFerano: Oh right, so in this particular case, I just added a note, maybe it can be made clearer then. On `dev`, right click is disabled in the tile map editor. So I left it as a boolean to preserve existing behavior and left a short note explaining what would be needed to get right click context menus working. I can see why, there's really no "selection" in this scene, so it doesn't make as much sense to use a right click context menu. I'll update the comment.

matgis (approved):

PR #12457, opened 2026-05-26
Make the editor GUI system independent of Spine

vlaaad:
This PR replaces the hard-coded Spine GUI node handling with a generic custom GUI node pipeline used by [extension-spine](https://github.com/defold/extension-spine/pull/277).

Main changes:
- Adds `TYPE_CUSTOM` GUI node support for named custom types and typed `custom_properties`.
- Adds Bob-side registration and migration for GUI custom node types.
- Migrates legacy Spine GUI node fields into custom properties during build/load paths.
- Adds generic GUI resource-kind registration in the editor so extension-defined GUI resources can participate in outlines, properties, validation, and build output (e.g., `Spine Scenes`).

Main internal editor SDK additions: 
1. [`register-gui-resource-kind`](https://github.com/defold/defold/blob/9736f3ae90772b18ef7d93481314234e38eec7eb/editor/src/clj/editor/gui.clj#L5056) — to define resource types (like Spine Scenes list)
2. [`register-custom-node-type-info`](https://github.com/defold/defold/blob/9736f3ae90772b18ef7d93481314234e38eec7eb/editor/src/clj/editor/gui.clj#L5056) — to define new GUI nodes (like Spine nodes that use Spine Scene resources)


engine/gamesys/src/gamesys/components/comp_gui.cpp
```diff
@@ -682,7 +791,11 @@ namespace dmGameSystem
 
     static void SetNodeCallback(const dmGui::HScene scene, dmGui::HNode n, const void* node_desc)
     {
-        SetNode(scene, n, (const dmGuiDDF::NodeDesc*) node_desc);
+        const dmGuiDDF::NodeDesc* node_desc_ddf = (const dmGuiDDF::NodeDesc*) node_desc;
+        if (SetCustomProperties(scene, n, node_desc_ddf))
```
vlaaad: From Codex: `SetNodeCallback` calls `SetCustomProperties` for every layout descriptor. Layout nodes without `custom_properties` overrides have an empty list, so switching layouts can clear the node’s default custom properties at runtime.

engine/gui/src/gui.cpp
```diff
@@ -2741,6 +2837,62 @@ namespace dmGui
...
+        if (!property)
+        {
+            return RESULT_RESOURCE_NOT_FOUND;
+        }
+        if (property->m_Property.m_Type != prop->m_Type)
+        {
+            return RESULT_WRONG_TYPE;
+        }
+
+        FreeCustomProperty(&property->m_Property);
```
vlaaad, line 2913: From Codex: `SetNodeCustomProperty` frees the old string before assigning the new property. Passing back the same scene-owned string stores a dangling pointer.

engine/gui/src/dmsdk/gui/gui.h
```diff
@@ -325,6 +368,32 @@ namespace dmGui
...
+     * @param node [type: dmGui::HNode] node
+     * @param key [type: dmhash_t] property name hash
+     * @param prop [type: dmGui::CustomProperty*] output property
+     * @return result [type: dmGui::Result] RESULT_OK on success
+     */
+    Result GetNodeCustomProperty(HScene scene, HNode node, dmhash_t key, CustomProperty* prop);
+
+    /*#
+     * Set a custom property on a GUI node.
+     * String values are owned by the GUI scene after a successful call and will
```
vlaaad: From Codex: The public setter takes `const CustomProperty*` but transfers and later `free()`s string ownership. This makes string literals, stack buffers, or non-`malloc` memory unsafe.

vlaaad (commented):
To @JCash — here are some findings from Codex that you might want to validate.

engine/gamesys/src/gamesys/components/comp_gui.cpp
JCash: it's probably worth looking into, yes.
I can do that once I take over for the next phase (moving world transforms into C code)

engine/gui/src/dmsdk/gui/gui.h
JCash: I mean, not sure why it complains, when the comment clearly states that gui takes ownership.

engine/gui/src/gui.cpp
JCash: Not really sure what it's talking about here.
The incoming "prop" must have a new string.

JCash (dismissed):

engine/gamesys/proto/gamesys/gui_ddf.proto
```diff
@@ -15,6 +15,35 @@ option java_outer_classname = "Gui";
...
+    optional string id          = 1; // for the editor
+    optional uint64 id_hash     = 2; // for runtime
+    required PropertyType type  = 3;
+
+    oneof value
+    {
+        float          number       = 4;
+        bool           boolean      = 5;
+        uint64         hash         = 6;
+        string         string_value = 7;
```
matgis: We should fix this inconsistent field name so it is just `string`. It's legal. we use it in `ddf_struct.proto`.

engine/gamesys/proto/gamesys/gui_ddf.proto
```diff
@@ -15,6 +15,35 @@ option java_outer_classname = "Gui";
...
+        dmMath.Vector3 vector3      = 8;
+        dmMath.Vector4 vector4      = 9;
+        dmMath.Quat    quat         = 10;
```
matgis, line 43: Would it be better to merge all of these into a single `dmMath.Vector4` `vector` field and have the `PropertyType` decide how to interpret it? 🤔

editor/src/clj/editor/gui.clj
```diff
@@ -435,19 +439,139 @@
...
+                                                              id actual-pb-type expected-pb-type))))))
+                                       (custom-property-entry->pb type-info
+                                                                 [(:id custom-property-info)
+                                                                  (custom-property-prop-value custom-property-info custom-property)]
+                                                                 include-custom-property-defaults))))
+                                 (sort-by :id)
+                                 vec)]
+      (if (coll/empty? custom-properties)
+        (dissoc node-desc :custom-properties)
+        (assoc node-desc :custom-properties custom-properties)))))
```
matgis: We have `protobuf/assign-repeated` for this sort of thing.

editor/src/clj/editor/gui.clj
```diff
@@ -435,19 +439,139 @@
...
+        regular-prop->value (apply dissoc prop->value (keys custom-properties))]
+    (cond-> regular-prop->value
+            (coll/not-empty custom-properties)
+            (assoc :custom-properties (custom-properties-prop-value->pb type-info custom-properties include-custom-property-defaults)))))
+
+(defn- sanitize-custom-properties [node-desc type-info]
+  (if-not (contains? node-desc :custom-properties)
+    node-desc
+    (let [include-custom-property-defaults (contains? (set (:overridden-fields node-desc))
+                                                      custom-properties-pb-field-index)
```
matgis: This seems wrong. We shouldn't include the defaults if the `node-desc` represents a template override node or a layout override node.

Only original `node-descs` - ones where `(:template-node-child node-desc)` is `false` among the top-level `nodes` (i.e. not `layouts` -> `nodes`) should include default values. For any other node-desc, the presence of an entry should signal an override of that property.

I don't really think we should be doing any sanitization of the `custom_properties` at all.

What we *should* be doing during the sanitation step is to migrate the legacy `spine_*` field values into `custom_properties`. However, the place to do this is in the `editor.spineguiext/fixup-spine-node` function along with the other node-specific sanitation that is already going on.

Also, the fact that this code was doing this and didn't trip any test failures suggests we need to add coverage for this scenario.

editor/src/clj/editor/gui.clj
```diff
@@ -666,6 +777,56 @@
...
+    (fn [custom-properties {:keys [id]}]
+      (let [old-value (get custom-properties id ::not-found)]
+        (case old-value
+          ::not-found custom-properties
+          (let [new-value (rename-fn old-value old-name new-name)]
+            (cond-> custom-properties
+                    (not= old-value new-value)
+                    (assoc id new-value))))))
+    custom-properties
+    custom-property-infos))
```
matgis: It would be cool to limit the rename to resource properties only.

editor/src/clj/editor/gui.clj
```diff
@@ -666,6 +777,56 @@
...
+                        {} layout->prop->override
+                        (let [new-value (rename-custom-gui-resource-references-in-properties rename-fn custom-property-infos old-name new-name old-value)]
+                          (cond-> layout->prop->override
+                                  (not= old-value new-value)
+                                  (update layout-name merge new-value))))))
+                  old-layout->prop->override
+                  old-layout->prop->override)]
+            (when-not (identical? old-layout->prop->override new-layout->prop->override)
+              (g/set-property node-id :layout->prop->override new-layout->prop->override))))))))
+
```
matgis: I don't think we should be doing this. If a node needs to respond to gui resource renames, it should implement the `update-gui-resource-reference` method for the specific node type and gui resource type. So in the case of `editor.spineguiext/SpineNode` our case it would be for `[::SpineNode :spine-scene]`. All other resource properties such as `:texture` and `:material` are inherited, so should already be covered.

editor/src/clj/editor/gui.clj
```diff
@@ -799,19 +963,22 @@
...
-       (vec)))
+(defn- prop->overridden-pb-field-indices [prop->override custom-property-id->info]
+  (cond-> (->> prop->override
+               (keys)
+               (remove custom-property-id->info)
+               (keep prop-key->pb-field-index)
+               (sort)
+               (vec))
+          (coll/any? custom-property-id->info (keys prop->override))
+          (conj custom-properties-pb-field-index)))
```
matgis: I make this point in the sister PR to `extension-spine` as well, but as far as the file formats are concerned, I think we should exclude `custom_properties` from the `overridden_fields`. For the legacy field values, an entry in the `overridden_fields` decides whether the value is overridden because field values that equal the Protobuf-declared default will be stripped out. For `custom_properties`, the presence of an entry should be the deciding factor.

However, we need to make sure this rule set applies to the runtime and Bob as well.

editor/src/clj/editor/gui.clj
```diff
@@ -799,19 +963,22 @@
...
 
 (def ^:private override-retained-pb-fields
   ;; These pb-fields will always be kept on override nodes. Other fields will be
   ;; stripped out unless their values deviate from the original node.
   [:type ; Not overridden or necessary, but improves readability in files.
-   :custom-type ; Not overridden or necessary, but improves readability in files.
+   :custom-type ; Legacy numeric custom type retained while loading old files.
```
matgis: We should probably remove `:custom-type` now, right? And include `:custom-properties` as an always-retained field, so we don't have to list it among the `overridden_fields`.

editor/src/clj/editor/gui.clj
```diff
@@ -828,29 +995,36 @@
...
-      (assoc
-        :type type ; Explicitly include the type (pb-field is optional, so :type-box would be stripped otherwise).
-        :child-index child-index))) ; Used to order sibling nodes in the SceneDesc.
+  (let [type-info (get-registered-node-type-info gui-node-type-registry (g/node-type _this))]
+    (-> (protobuf/make-map-without-defaults Gui$NodeDesc
+          :custom-type-name custom-type-name
+          :custom-properties (custom-properties-prop-value->pb
+                               type-info
+                               (select-keys prop->value (keys (:custom-property-id->info type-info)))
+                               false)
```
matgis: This seems wrong. You can't read from `prop->value` here, because it will have overrides from the currently selected layout. Note how we explicitly read `^:raw` property values for all other properties to avoid getting the layout override values.

editor/src/clj/editor/gui.clj
```diff
@@ -1320,47 +1497,48 @@
...
+                                         prop->value))))))))
+
+                    ;; We've reached the override root. Anything not
+                    ;; overridden for the current-layout by this point will
+                    ;; use values from our default layout.
+                    (let [node (g/node-by-id basis node-id)]
+                      (coll/merge
+                        (make-prop->value-for-default-layout node)
+                        prop->value))))
+                (make-recursive-prop->value-for-default-layout basis _node-id)))))
```
matgis: What changed here? Are we sure about this? Maybe a remnant that should be reverted?

editor/src/clj/editor/gui.clj
```diff
@@ -1369,32 +1547,33 @@
...
+                                    (let [layout-override-value (get prop-kw->layout-override-value prop-kw)]
+                                      (pair prop-kw
+                                            (cond-> (assoc prop-info :assoc-original-value? false) ; Disable automatic assoc in OverrideNode.produce-value. We want to manage it ourselves.
+
+                                                    (some? layout-override-value)
+                                                    (assoc :original-value layout-override-value) ; Any :original-value is fine. The key just needs to be present.
+
+                                                    (nil? layout-override-value)
+                                                    (dissoc :original-value))))))
+                                prop-kw->prop-info)))))))
```
matgis: What changed here? Another remnant that should be reverted?

editor/src/clj/editor/gui.clj
```diff
@@ -1441,7 +1620,7 @@
   [target-node-id target-layout-name evaluation-context]
   {:pre [(string? target-layout-name)]}
   (let [layout->prop->value (g/node-value target-node-id :layout->prop->value evaluation-context)
-        prop-kw->prop-info (:properties (g/node-value target-node-id :_declared-properties evaluation-context))
+        prop-kw->prop-info (:properties (g/node-value target-node-id :_properties evaluation-context))
```
matgis: Another remnant that should be reverted?

editor/src/clj/editor/gui.clj
```diff
@@ -3153,6 +3341,58 @@
...
+(defn- add-gui-resource-kind-handler [project {:keys [scene parent resource-kind-info]} select-fn]
+  (query-and-add-resources!
+    (:label resource-kind-info)
+    (:exts resource-kind-info)
+    (g/node-value parent :name-counts)
+    project
+    select-fn
+    (partial add-gui-resource-kind-entry scene parent resource-kind-info)))
+
+(g/defnode GuiResourceKindNode
```
matgis, line 3274: Should we migrate the built-in gui resource type container nodes such as `TexturesNode` to this system as well?

editor/src/clj/editor/gui.clj
```diff
@@ -3203,20 +3443,22 @@
       clipping/setup-states
       sort-scene)))
 
-(defn- make-layout-desc [layout-name decorated-node-msgs]
+(defn- strip-custom-type [node-desc]
+  (dissoc node-desc :custom-type))
```
matgis: It should have already been removed during sanitization, right?

editor/src/clj/editor/gui.clj
```diff
@@ -3203,20 +3443,22 @@
...
                (fn [{:keys [layout->prop->override] :as decorated-node-msg}]
                  {:pre [(map? layout->prop->override)]}
                  (when-some [prop->override (coll/not-empty (layout->prop->override layout-name))]
-                   (let [overridden-fields
-                         (->> prop->override
-                              (map (comp prop-key->pb-field-index key))
-                              (sort))]
+                   (let [type-info (node-desc->node-type-info gui-node-type-registry decorated-node-msg)
+                         custom-property-id->info (:custom-property-id->info type-info)
+                         overridden-fields (prop->overridden-pb-field-indices prop->override custom-property-id->info)]
```
matgis: We shouldn't include `custom_properties` among the `overridden_fields`.

editor/src/clj/editor/gui.clj
```diff
@@ -3203,20 +3443,22 @@
...
                      (->> prop->override
-                          (e/map prop-entry->pb-field-entry)
+                          (prop->pb-field-entries-for-type type-info true)
```
matgis: This looks wrong. `prop->pb-field-entries-for-type` looks at `:custom-properties` in `prop->override`, which shouldn't exist there?

editor/src/clj/editor/gui.clj
```diff
@@ -3254,16 +3497,17 @@
...
         (coll/into-> node-msgs []
           (map #(dissoc % :layout->prop->override :layout->prop->value))
           (map (fn [node-desc]
-                 (if (:template-node-child node-desc)
-                   (strip-unused-overridden-fields-from-node-desc node-desc)
-                   (strip-redundant-size-from-node-desc node-desc)))))]
+                 (strip-custom-type
+                   (if (:template-node-child node-desc)
+                     (strip-unused-overridden-fields-from-node-desc node-desc)
+                     (strip-redundant-size-from-node-desc node-desc))))))]
```
matgis: There shouldn't be a `:custom-type` - it should have been converted into `:custom-type-name` by the `sanitize-fn`.

editor/src/clj/editor/gui.clj
```diff
@@ -3305,7 +3549,30 @@
...
+        (dissoc :id)
+        (cond-> hash-property
+          (assoc :hash (murmur/hash64 (:string-value custom-property))))
+        (cond-> hash-property
+          (dissoc :string-value)))))
+
+(defn- node-desc->rt-custom-properties [node-desc]
+  (cond-> node-desc
+          (contains? node-desc :custom-properties)
+          (update :custom-properties #(mapv custom-property->rt-custom-property %))))
```
matgis: We have `protobuf/sanitize-repeated` for this sort of thing.

editor/src/clj/editor/gui.clj
```diff
@@ -3376,36 +3643,44 @@
...
-                               (some->> (-> decorated-node-msg
-                                            (dissoc :layout->prop->override :layout->prop->value)
-                                            (into (map prop-entry->pb-field-entry)
-                                                  prop->value)
-                                            (node-desc->rt-node-desc layout-name))
+                               (some->> (let [type-info (node-desc->node-type-info gui-node-type-registry decorated-node-msg)
+                                              include-custom-property-defaults (some? prop->override)]
+                                          (-> decorated-node-msg
+                                              (dissoc :layout->prop->override :layout->prop->value)
+                                              (into (prop->pb-field-entries-for-type type-info include-custom-property-defaults prop->value))
```
matgis: Again, seems possibly wrong. `prop->pb-field-entries-for-type` reads from `:custom-properties`, which does not exist in `prop->value`.

editor/src/clj/editor/gui.clj
```diff
@@ -3709,12 +3996,13 @@
                   (coll/into-> (:material-infos aux-basic-gui-scene-info) (sorted-set)
                     (map key)
                     (remove coll/empty?))]
-              {:font (:font-names aux-basic-gui-scene-info)
-               :layer (:layer-names aux-basic-gui-scene-info)
-               :material material-names
-               :particlefx (:particlefx-resource-names aux-basic-gui-scene-info)
-               :spine-scene (:spine-scene-names aux-basic-gui-scene-info)
-               :texture (:texture-resource-names aux-basic-gui-scene-info)})))
+              (merge
```
matgis: Use `coll/merge`?

editor/src/clj/editor/gui.clj
```diff
@@ -3798,37 +4086,47 @@
...
-(defn add-gui-node-with-props! [scene parent node-type custom-type props select-fn]
+(defn add-gui-node-with-props! [scene parent node-type-info props select-fn]
   (-> (g/with-auto-evaluation-context evaluation-context
         (let [node-tree (g/node-value scene :node-tree evaluation-context)
+              def-node-type (:node-type node-type-info)
+              id-base (or (some-> (node-types/->name def-node-type) extension-type-name->id)
+                          (subs (name (:type node-type-info)) 5))
               id (or (:id props)
-                     (id/resolve (subs (name node-type) 5)
+                     (id/resolve id-base
```
matgis: Inline `id-base` here so we don't evaluate it when there is an `:id`?

editor/src/clj/editor/gui.clj
```diff
@@ -3798,37 +4086,47 @@
...
               next-index (gui-attachment/next-child-index parent evaluation-context)
               node-properties (assoc props
                                 :id id
                                 :child-index next-index
-                                :custom-type custom-type
-                                :type node-type)]
+                                :type (:type node-type-info))
+              node-properties (cond-> node-properties
+                                (coll/not-empty (:custom-type-name node-type-info))
+                                (assoc :custom-type-name (:custom-type-name node-type-info)))]
```
matgis: Could use `protobuf/assign` here:
```clj
(protobuf/assign node-properties :custom-type-name (not-empty (:custom-type-name node-type-info)))
```

editor/src/clj/editor/gui.clj
```diff
@@ -3925,18 +4228,30 @@
...
-                (let [prop-value (pb-value->prop-value pb-value)]
-                  (-> node-properties
-                      (dissoc! pb-field)
-                      (assoc! prop-key prop-value)))
-                node-properties))
-            (transient node-desc)
-            node-desc)))
+(defn- node-desc->node-properties
+  ([gui-node-type-registry node-desc]
+   (node-desc->node-properties gui-node-type-registry node-desc false))
```
matgis: Remove information-hiding arity.

editor/src/clj/editor/gui.clj
```diff
@@ -3925,18 +4228,30 @@
...
+                                   (assoc! prop-key prop-value)))
+                             node-properties)))
+                       (transient node-desc)
+                       node-desc))
+             (= :type-custom (:type node-desc))
+             (assoc :custom-type-name (:custom-type-name type-info))
+             (and include-custom-property-defaults
+                  (not (contains? node-desc :custom-properties))
+                  (not (coll/empty? (:custom-properties type-info))))
+             (coll/merge (:custom-property-defaults type-info))))))
```
matgis: It seems to me we could have ended up in a simpler place with the custom property handling if we just let plugin gui nodes such as `editor.spineguiext/SpineNode` implement a multimethod we call here, and then have `SpineNode` write these property values to the `custom_properties` field in its `node-msg` output?

There would be a little bit of `rt-node-msg` conversion around the `custom_properties`, but we would avoid a lot of complexity. 🤔

editor/src/clj/editor/gui.clj
```diff
@@ -3945,44 +4260,51 @@
...
-(defn- extract-overrides [node-properties]
+(def ^:private node-property-defaults
+  (persistent!
+    (reduce-kv
+      (fn [node-property-defaults pb-field pb-value]
+        (if-some [[prop-key pb-value->prop-value] (pb-field-to-node-property-conversions pb-field)]
+          (assoc! node-property-defaults prop-key (pb-value->prop-value pb-value))
+          (assoc! node-property-defaults pb-field pb-value)))
+      (transient {})
+      default-node-desc-pb-field-values)))
```
matgis: Do we need this? The node `(property ...)` declares defaults, so if there isn't a value for a custom property we should still be fine?

editor/src/clj/editor/gui.clj
```diff
@@ -3945,44 +4260,51 @@
...
-                               prop-value))))))
+              (mapcat (fn [prop-key]
+                        (if (= :custom-properties prop-key)
+                          (coll/not-empty
+                            (select-keys node-properties (keys (:custom-property-id->info type-info))))
+                          (let [prop-value (node-properties prop-key ::not-found)]
+                            [(pair prop-key
+                                   (if (= ::not-found prop-value)
+                                     (node-property-defaults prop-key)
+                                     prop-value))])))))
```
matgis: This seems like it could be combined with `node-desc->node-properties`, or be an additional extension point for `SpineNode` to implement. 🤔

editor/src/clj/editor/gui.clj
```diff
@@ -3945,44 +4260,51 @@
...
         (:overridden-fields node-properties)))
 
 (def ^:private default-material-proj-path (protobuf/default Gui$SceneDesc :material))
 
 (defn load-gui-scene [project self resource scene]
   {:pre [(map? scene)]} ; Gui$SceneDesc in map format.
   (let [graph-id           (g/node-id->graph-id self)
+        workspace          (project/workspace project)
```
matgis: Use `resource/workspace` instead?

Style: Don't align indentation.

editor/src/clj/editor/gui.clj
```diff
@@ -435,19 +439,139 @@
...
+                                   (coll/not-empty (:custom-type-name node-desc)))]
+      (when-not (= type (:type type-info))
+        (throw (IllegalArgumentException.
+                 (format "GUI node custom_type_name '%s' belongs to node type %s, expected %s."
+                         custom-type-name (:type type-info) type))))
+      (when (and (contains? node-desc :custom-type)
+                 (not= (:custom-type type-info) (:custom-type node-desc)))
+        (throw (IllegalArgumentException.
+                 (format "GUI node custom_type_name '%s' resolves to custom_type %s, but the node specifies custom_type %s."
+                         custom-type-name (:custom-type type-info) (:custom-type node-desc))))))))
```
matgis: `IllegalArgumentException` seems like the wrong exception type?

editor/src/clj/editor/gui.clj
```diff
@@ -435,19 +439,139 @@
...
+
+(defn- sanitize-node-custom-type-name [node-desc type-info]
+  (if (= :type-custom (:type node-desc))
+    (-> node-desc
+        (assoc :custom-type-name (:custom-type-name type-info))
+        (dissoc :custom-type))
+    (dissoc node-desc :custom-type-name :custom-type)))
```
matgis: I don't think the keys will be present in the `false` branch, since defaults are stripped when reading?

editor/src/clj/editor/gui.clj
```diff
@@ -828,29 +995,36 @@
...
+          :id (if (g/node-override? _this) generated-id id)
+          :color color ; TODO: Not used by template (forced to [1.0 1.0 1.0 1.0]). Move?
+          :alpha alpha
+          :inherit-alpha inherit-alpha
+          :enabled enabled
+          :layer layer
+          :parent parent
+          :overridden-fields (prop->overridden-pb-field-indices
+                                (gt/overridden-properties _this)
+                                (:custom-property-id->info type-info)))
```
matgis: Style: Indentation.

editor/src/clj/editor/gui.clj
```diff
@@ -871,8 +1045,8 @@
 
 (defn- layout-property-prop-def? [prop-def]
   (let [value-dependencies (-> prop-def :value :dependencies)]
-    (and (= 2 (count value-dependencies))
-         (contains? value-dependencies :prop->value))))
+    (and (contains? value-dependencies :prop->value)
+         (= 2 (count value-dependencies)))))
```
matgis: The original order seems better to me, since a lookup is more expensive than `count`.

editor/src/clj/editor/gui.clj
```diff
@@ -1000,12 +1172,12 @@
...
-          [prop-kw]
-          (g/with-auto-evaluation-context evaluation-context
-            (keys (changes-fn evaluation-context node-id prop-kw nil nil))))]
-    (layout-property-clears-tx-data layout-name node-id cleared-prop-kws)))
+  (g/with-auto-evaluation-context evaluation-context
+    (let [cleared-prop-kws
+          (if (nil? changes-fn)
+            [prop-kw]
+            (keys (changes-fn evaluation-context node-id prop-kw nil nil)))]
+      (layout-property-clears-tx-data evaluation-context layout-name node-id cleared-prop-kws))))
```
matgis: Another remnant from an earlier take? We don't seem to use the `evaluation-context`, so let's revert these changes if that's the case.

editor/src/clj/editor/gui.clj
```diff
@@ -4602,171 +5112,244 @@
...
-                        :node-type-infos (:node-type->custom-type->type-info @node-type-info-registry)})))))
+   (or (-> gui-node-type-registry
+         :type->custom-type->type-info
+         (get type)
+         (get custom-type))
+     (throw (IllegalStateException.
+              (format "Unable to locate GUI node type info. Extension not loaded? (type=%s, custom-type=%s, type-infos=%s)"
+                      type
+                      custom-type
+                      (:type->custom-type->type-info gui-node-type-registry)))))))
```
matgis: I dislike having two arities that take different kinds of arguments. Let's split this into two functions and get rid of the `get-` prefix while we're at it.

editor/src/clj/editor/gui.clj
```diff
@@ -4602,171 +5112,244 @@
...
+
+(defn- update-gui-resource-type-tx-data [workspace f & args]
+  (concat
+    (apply g/update-property workspace :resource-types update-gui-resource-type-map f args)
+    (apply g/update-property workspace :resource-types-non-editable update-gui-resource-type-map f args)))
 
-(defn- get-registered-node-type-infos []
-  (:type-infos @node-type-info-registry))
+(defn- get-registered-node-type-infos [gui-node-type-registry]
+  (:type-infos gui-node-type-registry))
```
matgis: This seems redundant, let's get rid of it.

editor/src/clj/editor/gui.clj
```diff
@@ -4250,50 +4601,61 @@
...
-        (protobuf/sanitize-repeated :nodes sanitize-scene-node)
-        (protobuf/sanitize-repeated :layouts sanitize-layout)
+        (protobuf/sanitize-repeated :nodes (partial sanitize-scene-node gui-node-type-registry))
+        (protobuf/sanitize-repeated :layouts (partial sanitize-layout gui-node-type-registry))
         (protobuf/assign-repeated :resources merged-resource-descs)
         (update :material fn/or default-material-proj-path))))
 
+(defn- make-gui-read-fn [gui-node-type-registry]
+  (comp (partial sanitize-scene gui-node-type-registry)
+        (partial protobuf/read-map-without-defaults (:pb-class pb-def))))
```
matgis: We don't need this? Just specify `:sanitize-fn (partial sanitize-scene gui-node-type-registry)` to `register-ddf-resource-type`.

editor/src/clj/editor/gui.clj
```diff
@@ -4319,11 +4681,22 @@
...
-                 target-node+name (when-not (neg? (coll/index-of atlas-exts ext))
-                                    (drop-target-node-id+unique-name gui-scene :textures-node base-name evaluation-context))]
-        (when-some [[target-node name] target-node+name]
-          (add-texture gui-scene target-node resource name))))))
+      (g/with-auto-evaluation-context evaluation-context
+        (or
+          (when-some [[resource-kind resource-kind-info]
+                      (coll/some
+                        (fn [[resource-kind {:keys [exts] :as resource-kind-info}]]
+                          (when (some #{ext} exts)
```
matgis: More efficient? `(when-not (neg? (coll/index-of exts ext)) ...)`

editor/src/clj/editor/gui.clj
```diff
@@ -4333,24 +4706,30 @@
...
+      (concat
+        (resource-node/register-ddf-resource-type workspace
+          :ext ext
+          :label (:label def)
+          :build-ext (:build-ext def)
+          :node-type GuiSceneNode
+          :ddf-type (:pb-class def)
+          :load-fn load-gui-scene
+          :allow-unloaded-use false ; Sort of works, but disabled until we can fix the file formats to not include all nodes imported from templates.
+          :read-fn (make-gui-read-fn base-node-type-registry)
```
matgis: Supply `:sanitize-fn` from `partial` instead, constructed outside the loop.

editor/src/clj/editor/gui.clj
```diff
@@ -4432,6 +4811,137 @@
     ;; make the node type a branch
     (attachment/alias workspace node-type :nodes GuiNode)))
 
+;; SDK api
+(defn gui-resource-kind-node [basis gui-scene-node resource-kind]
+  (coll/some
+    (fn [arc]
+      (let [id (gt/source-id arc)]
```
matgis: Use `node-id`, since `id` is an overloaded term.

editor/src/clj/editor/gui.clj
```diff
@@ -4432,6 +4811,137 @@
...
+             (format "GUI resource kind does not specify :attach-fn. (resource-kind=%s, attach-fn=%s)" resource-kind attach-fn))))
+  (let [exts (cond
+               (string? exts) [exts]
+               (vector? exts) exts
+               :else nil)]
+    (when-not (and (seq exts)
+                   (every? string? exts))
+      (throw (IllegalArgumentException.
+               (format "GUI resource kind does not specify string or vector :exts. (resource-kind=%s, exts=%s)" resource-kind exts))))
+    (assoc info :exts exts)))
```
matgis: Seems we could use `ensure/argument` for these?

editor/src/clj/editor/gui.clj
```diff
@@ -4432,6 +4811,137 @@
...
+
+;; SDK api
+(defn register-gui-resource-kind
+  "Create transaction steps that register a GUI resource kind
```
matgis: Missing `.`.

editor/src/clj/editor/gui.clj
```diff
@@ -4602,171 +5112,244 @@
...
-  (-> state
-      (assoc-in [:node-type->custom-type->type-info node-type custom-type] type-info)
-      (update :type-infos conj type-info)
-      (cond-> (not (:deprecated type-info)) (update :node-cls->type-info assoc node-cls type-info))))
-
-(defonce ^:private node-type-info-registry
-  (atom (reduce register-node-type-info empty-node-type-registry base-node-type-infos)))
+(defn- add-node-type-info [state {:keys [node-type type custom-type-name deprecated] :as type-info}]
+  (let [custom-type (or (:custom-type type-info)
+                        (murmur/hash32 custom-type-name))
```
matgis, line 5100: Are we reading the `:custom-type` here to maintain backwards compatibility with earlier versions of `extension-spine`, or did we forget to update it to `:custom-type-name` in `extension-spine`?

It seems we should be using `:custom-type-name` everywhere, right?

editor/src/clj/editor/gui.clj
```diff
@@ -4602,171 +5112,244 @@
...
+        (assoc-in [:type->custom-type->type-info type custom-type] type-info)
+        (update :type-infos conj type-info)
+        (cond-> (not deprecated)
+          (update :node-type->type-info assoc node-type type-info)
+          (and (not deprecated) custom-type-name) (assoc-in [:custom-type-name->type-info custom-type-name] type-info)))))
+
+(defn- gui-resource-type-from-resource-types [resource-types]
+  (or (get resource-types (:ext pb-def))
+      (throw (IllegalStateException.
+               (format "Unable to locate GUI resource type. (ext=%s)" (:ext pb-def))))))
```
matgis: This should never happen. I think we could just `assert`?

editor/src/clj/editor/gui.clj
```diff
@@ -4602,171 +5112,244 @@
...
+(defn- update-gui-resource-type [resource-type f & args]
+  (when-not (map? resource-type)
+    (throw (IllegalStateException.
+             (format "Unable to update GUI resource type before it has been registered. (ext=%s)" (:ext pb-def)))))
+  (let [resource-type (apply f resource-type args)
+        gui-node-type-registry (:gui-node-type-registry resource-type)]
+    (assoc resource-type
+      ;; GUI read-time sanitization resolves custom node types, so the read-fn
+      ;; must be rebuilt whenever extensions update the node type registry.
+      :read-fn (make-gui-read-fn gui-node-type-registry))))
```
matgis: Updating the `:read-fn` like this is unconventional. Elsewhere, we simply supply the `workspace` to the `:sanitize-fn` and query resource types from the `workspace` inside the `:sanitize-fn`. See `editor.collection/sanitize-collection`.

It might be slightly slower to perform the lookup each time we sanitize a resource, but I think we should take that approach here as well to keep things simple.

Later on, we could do a refactoring that supplies some prepared data to every `:read-fn` and `:sanitize-fn` so they do not have to repeat these lookups.

editor/src/clj/editor/gui.clj
```diff
@@ -4602,171 +5112,244 @@
...
-          node-type (g/node-type node)
-          last-colon-index (str/last-index-of property \:)
-          property-name (if last-colon-index (subs property (inc last-colon-index)) property)]
-      (when-let [prop-kw ((node-type->layout-property-names node-type) property-name)]
-        (let [layout-name (if last-colon-index (subs property 0 last-colon-index) "")
-              layout->prop->value (g/node-value node-id :layout->prop->value evaluation-context)]
+    (let [[layout-name property-name] (ext-property->layout-name+property-name property)]
+      (when-let [prop-kw (ext-property-name->prop-kw node-id property-name evaluation-context)]
+        (let [layout->prop->value (g/node-value node-id :layout->prop->value evaluation-context)
+              prop-info (ext-property-info node-id prop-kw evaluation-context)]
```
matgis: I don't understand why we need to change anything here now that everything is backed by declared node properties using `layout-property-getter` and so on? The original code should work, no?

editor/src/clj/editor/gui.clj
```diff
@@ -4602,171 +5112,244 @@
...
+  (let [script-name->custom-property-info (:custom-property-script-name->info type-info)]
+    (reduce-kv
+      (fn [[attachment custom-properties] property lua-value]
+        (if-let [custom-property-info (script-name->custom-property-info property)]
+          (let [value (attachment-custom-property-value rt project evaluation-context custom-property-info lua-value)]
+            [(dissoc attachment property)
+             (assoc custom-properties (:id custom-property-info) value)])
+          [attachment custom-properties]))
+      [attachment nil]
+      attachment)))
```
matgis: I don't think we should need any of this if we just treat `custom_properties` as storage?

editor/src/clj/editor/gui.clj
```diff
@@ -4790,32 +5373,153 @@
...
+    (throw (IllegalArgumentException.
+             (format "GUI custom property does not specify :protobuf-type. (node-type=%s, property=%s, custom-property-static=%s)"
+                     (:name @node-type) prop-kw custom-property-static))))
+  (when-not (contains? custom-property-pb-type->value-field protobuf-type)
+    (throw (IllegalArgumentException.
+             (format "Unsupported GUI custom property protobuf type. (type=%s)" protobuf-type))))
+  (when-not (contains? layout-property-keys prop-kw)
+    (throw (IllegalArgumentException.
+             (format "GUI custom property %s on %s is not a layout property."
+                     prop-kw (:name @node-type)))))
```
matgis: We could use `ensure/argument` here?

Although I'm not sure we need this at all?

editor/src/clj/editor/resource_node.clj
```diff
@@ -288,15 +288,16 @@
...
-        read-raw-fn (if read-defaults
-                      (partial protobuf/read-map-with-defaults ddf-type)
-                      (partial protobuf/read-map-without-defaults ddf-type))
-        read-fn (cond->> read-raw-fn
-                         (some? sanitize-fn) (comp sanitize-fn))
+        read-fn (or read-fn
+                    (cond->> (if read-defaults
+                               (partial protobuf/read-map-with-defaults ddf-type)
+                               (partial protobuf/read-map-without-defaults ddf-type))
+                             sanitize-fn (comp sanitize-fn)))
```
matgis: We might want to revert this and just supply the `workspace` to the `:sanitize-fn`.

editor/src/clj/internal/node.clj
```diff
@@ -150,6 +150,15 @@
...
 
+(defn property-statics
+  [nt label]
+  (into {}
+        (keep (fn [[property-label property-info]]
+                (let [static-value (get (:statics property-info) label ::not-found)]
+                  (when-not (identical? ::not-found static-value)
+                    (pair property-label static-value)))))
+        (all-properties nt)))
+
```
matgis, line 161: We might not end up needing the `(static)` declarations, but I think it's a rather neat addition. 👍

editor/test/integration/extension_spine_test.clj
```diff
@@ -259,9 +294,6 @@
       (is (not (g/error? (g/node-value main-collection :node-outline)))))))
 
 (deftest legacy-spine-project-user-migration-test
-  ;; Clear custom gui scene loaders to ensure a clean test.
-  (gui/clear-custom-gui-scene-loaders-and-node-types-for-tests!)
-
```
matgis, line 264: Yay! We finally got rid of this hack! 🥳

editor/test/integration/extension_spine_test.clj
```diff
@@ -77,6 +88,30 @@
...
+                                     dirty-save-data)]
+          (is (= #{} dirty-proj-paths)
+              (str "Legacy Spine migration should not dirty resources on load.\n"
+                   (->> dirty-save-data
+                        (map (fn [save-data]
+                               (str (resource/proj-path (:resource save-data))
+                                    "\n"
+                                    (test-util/save-data-diff-message save-data))))
+                        (string/join "\n")))))))))
+
```
matgis: This is already covered by `integration.save-data-test/no-unsaved-changes-after-load-test`, providing we include a `legacy_spine_fields.gui` in the `silently_migrated` project folder. This is the place to put silent migration tests. We should also check that the legacy field values end up as `custom_properties` in the `"gui"` section of `integration.save-data-test/silent-migrations-test`.

editor/test/integration/extension_spine_test.clj
```diff
@@ -432,6 +465,69 @@
...
 
+(defn- property-value-choices [node-id prop-kw]
+  (->> (g/node-value node-id :_properties)
+       :properties
+       prop-kw
+       :edit-type
+       :options
+       (mapv first)))
+
+(deftest new-spine-node-custom-property-edit-test
```
matgis: `custom-property` -> `property`

editor/test/integration/extension_spine_test.clj
```diff
@@ -456,10 +552,26 @@
         (gui-test/with-visible-layout! gui-scene "Portrait"
           (test-util/prop! spine-node :spine-default-animation "jump")))
 
-      (testing "After overriding a property, the override NodeDesc includes all properties from the default layout."
+      (testing "After overriding a property, the runtime override NodeDesc includes the effective custom properties."
```
matgis: Clarify so it is evident that we're testing all overridden properties, not just the `custom_properties`.

editor/src/clj/editor/gui.clj
```diff
@@ -4602,171 +5112,244 @@
...
-       (throw (ex-info (format "Unable to locate GUI node type info. Extension not loaded? (node-cls=%s)" (:k node-cls))
-                       {:node-cls node-cls
-                        :node-type-infos (keys (:node-cls->type-info @node-type-info-registry))}))))
-  ([node-type custom-type]
-   {:pre [(keyword? node-type)
+  ([gui-node-type-registry node-type]
+   {:pre [(g/node-type? node-type)]}
+   (or (-> gui-node-type-registry
+         :node-type->type-info
+         (get node-type))
```
matgis: Does this work when the `node-type` is redefined from the REPL?

editor/test/integration/gui_test.clj
```diff
@@ -84,12 +219,228 @@
...
     (let [node-id (test-util/resource-node project "/logic/main.gui")
           _gui-node (ffirst (g/sources-of node-id :child-outlines))]
       (is (some? _gui-node)))))
 
+(deftest custom-gui-extension-registration
+  (test-util/with-scratch-project "test/resources/empty_project"
+    (let [custom-type (murmur/hash32 "TestCustom")
+          test-quat [(float 0.0) (float 0.0) (float 0.707) (float 0.707)]
+          test-vector3 [(float 1.0) (float 2.0) (float 3.0)]
+          test-vector4 [(float 4.0) (float 5.0) (float 6.0) (float 7.0)]
```
matgis: A bit noisy, perhaps? Would `(vector-of :float)` work? Do we even need them to be floats?

editor/test/integration/gui_test.clj
```diff
@@ -482,8 +837,8 @@
         user-data {:scene scene :parent parent :display-profile name :handler-fn gui/add-layout-handler}]
     (test-util/handler-run :edit.add-embedded-component [{:name :workbench :env {:selection [parent] :project project :user-data user-data :app-view app-view}}] user-data)))
 
-(defn- run-add-gui-node! [project scene app-view parent node-type custom-type]
-  (let [user-data {:scene scene :parent parent :node-type node-type :custom-type custom-type :handler-fn gui/add-gui-node-handler}]
+(defn- run-add-gui-node! [project scene app-view parent node-type-info]
```
matgis: Maybe cleaner for this function to take a `node-type` instead of a `node-type-info`?

editor/test/integration/gui_clipping_test.clj
```diff
@@ -28,8 +30,12 @@
                    (into {}))]
     (id->node id)))
 
+(defn- gui-node-type-info [project node-type]
+  (get-in (get (workspace/get-resource-type-map (project/workspace project) :editable) "gui")
+          [:gui-node-type-registry :node-type->type-info node-type]))
```
matgis: We seem to be repeating this helper. Maybe add it to `editor.gui`?

editor/test/integration/gui_test.clj
```diff
@@ -1105,6 +1460,185 @@
...
+      (let [gui-scene (test-util/resource-node project gui-resource)
+            custom-node (gui-node gui-scene "custom")]
+        (testing "Properties panel exposes custom properties as real graph properties."
+          (is (= 2.0 (prop custom-node :test-number)))
+          (is (= 2.0 (:test-number (g/node-value custom-node :prop->value))))
+          (is (contains? (get-in (g/node-value custom-node :_properties) [:properties]) :test-number))
+          (is (contains? (get-in (g/node-value custom-node :_properties) [:properties]) :test-dash))
+          (is (contains? (get-in (g/node-value custom-node :_properties) [:properties]) :test-underscore))
+          (is (contains? (get-in (g/node-value custom-node :_properties) [:properties]) :test-slash))
+          (is (contains? (get-in (g/node-value custom-node :_properties) [:properties]) :test-colon))
```
matgis: No need for `get-in` when the path has a single key.

editor/test/integration/gui_test.clj
```diff
@@ -1105,6 +1460,185 @@
...
+          (is (not (contains? (get-in (g/node-value custom-node :_properties) [:properties]) :custom-properties))))
+
+        (testing "Editor scripts use graph property names with regular snake_case conversion."
+          (g/with-auto-evaluation-context evaluation-context
+            (is (= 0.0 ((ext-graph/ext-value-getter custom-node "test_slash" project evaluation-context))))
+            (is (= 0.0 ((ext-graph/ext-value-getter custom-node "test_colon" project evaluation-context))))
+            (is (= 0.0 ((ext-graph/ext-value-getter custom-node "Landscape:test_slash" project evaluation-context))))
+            (is (= 0.0 ((ext-graph/ext-value-getter custom-node "Landscape:test_colon" project evaluation-context))))))
```
matgis, line 1583: This all seems to stem from the decision to make `custom_properties` a system-level thing rather than just storage. But since these are real properties, we don't need them to be a system-level thing.

editor/test/integration/save_data_test.clj
```diff
@@ -214,9 +215,79 @@
     {"data_type" :allowed-default
      "long_values" :unused}}
 
+   'dmGuiDDF.Property
+   {:default
+    {"id_hash" :runtime-only}}
```
matgis: It should be marked as such in the `.proto` file. Then we don't need this here.

editor/test/integration/save_data_test.clj
```diff
@@ -251,7 +322,9 @@
 
    ['dmGuiDDF.NodeDesc "[TYPE_CUSTOM]"]
    {:default
-    {"custom_type" :non-overridable
+    {"custom_type" :deprecated ; Project files use custom_type_name. Runtime numeric custom_type coverage is tested in GUI build tests.
+     "custom_properties" :non-editable
```
matgis: I don't think we should list `custom_properties` as a `pb-ignored-field`? I can't recall exactly why the `:non-editable` exclusion exists, but these values are editable in the sense that the user can perform actions that affect the value of the `custom_properties` field.

editor/test/integration/save_data_test.clj
```diff
@@ -266,6 +339,10 @@
      "size" :unused
      "size_mode" :unused
      "slice9" :unused
+     "spine_create_bones" :deprecated
+     "spine_default_animation" :deprecated
+     "spine_scene" :deprecated
+     "spine_skin" :deprecated
```
matgis: Whenever you tag a field as `:deprecated`, you must also add a comment detailing where silent migration is tested. For example:
```clj
'dmGuiDDF.SceneDesc
{:default
 {"background_color" :deprecated ; Migration tested in integration.save-data-test/silent-migrations-test.
  "spine_scenes" :deprecated}} ; Migration tested in integration.save-data-test/silent-migrations-test.
```

editor/test/resources/save_data_project/checked01.gui
```diff
@@ -711,20 +782,33 @@ nodes {
...
   overridden_fields: 29
   overridden_fields: 30
-  overridden_fields: 39
-  overridden_fields: 40
-  overridden_fields: 41
   overridden_fields: 45
   overridden_fields: 46
   overridden_fields: 47
-  overridden_fields: 48
+  overridden_fields: 50
```
matgis: We probably shouldn't list the `custom_properties` field among the `overridden_fields`.

matgis (changes requested):
I've only reviewed the editor changes and the `.proto` changes. 

The custom Gui resource types refactoring is pretty much what I had in mind, but it seems to me that a lot of unnecessary complexity was introduced by treating `custom_properties` as a system-level thing in the editor code, rather than simply a place in the file to store property values.

I would suggest reverting most of the editor-side changes related to `custom_properties` and instead concentrating on delegating the reading of the `prop-kw->value` map from a `node-desc` to a multimethod that can be customized for the `SpineNode`, and then emitting the `^:raw` property values as `custom_properties` it its `node-msg` output. There is more details among my PR comments.

I would also like to see some changes to the `.proto` field names, and found a lot of issues around how overrides are isolated across several override layers. The fact that these findings did not trip any test failures suggests we need more coverage as well.

matgis:
Here are some additional concerns reported by Codex. I'll let you decide if they are valid concerns:

1. [P1] Bob now rejects existing numeric custom GUI nodes without a Bob-side registry entry. `GuiBuilder.migrateCustomNode` now runs for every `TYPE_CUSTOM` node and throws when `custom_type` cannot be resolved in `GuiCustomTypeRegistry`. Existing native GUI custom-node extensions can save only a numeric runtime `custom_type`, so those projects would fail to build even though they worked before.

2. [P2] Editor build targets omit default custom properties on default-layout nodes. The editor runtime build path uses sparse node messages, while Bob fills registered custom-property defaults. That can make editor-built content miss valid defaults such as Spine’s `spine_skin` / `spine_create_bones`, causing runtime `GetNodeCustomProperty` to report not found where Bob-built content works.

editor/src/clj/editor/gui.clj
vlaaad: Would be cool, but I'm anxious about expanding the scope of this PR more.

editor/src/clj/editor/gui.clj
vlaaad: We don't do it any more!

editor/src/clj/editor/gui.clj
vlaaad: This is for earlier extension spine compatibility; otherwise, we use a custom type name everywhere, and we save with strings.

editor/src/clj/editor/gui.clj
vlaaad: I think rebuilding the read-fn is what keeps things simpler. We keep the read-fn a pure function of its inputs instead of depending on the whole workspace. Perhaps we should change `editor.collection/sanitize-collection` instead?

editor/src/clj/editor/resource_node.clj
vlaaad: I don't think it's going to be simpler that way.

editor/test/integration/gui_test.clj
vlaaad: The test is now fixed, we use custom properties only when saving/loading/building, otherwise we use graph properties.

editor/test/integration/gui_test.clj
```diff
@@ -3186,17 +3782,10 @@
...
                          "disabled_on/box" {:enabled false}
                          "enabled_on/box" {}}
-              "Unaltered" {"disabled_off/box" {:enabled false}
-                           "enabled_off/box" {:enabled false}
-                           "disabled_on/box" {:enabled false}
-                           "enabled_on/box" {}}
-              "Altered Parent" {"disabled_off/box" {:enabled false}
-                                "enabled_off/box" {:enabled false}
-                                "disabled_on/box" {}
+              "Unaltered" {}
```
vlaaad, line 3808: In `make-rt-layout-desc`, we previously had a code that omitted producing layout nodes if the layout's node description was equal to the base node description. It was broken because we compared a layout node desc with cleared defaults to a non-layout node desc without defaults clearing, so semantically equal desc were not equal in Clojure terms.

vlaaad:
> Bob still errors on unknown numeric `TYPE_CUSTOM` hashes

This is fine, we expect all GUI extensions to use registered names

> Collapse `vector3`/`vector4`/`quat` proto fields into one vector field

I don't have a strong opinion, but I think it's a bit clearer with the current straightforward mapping.

You had a question whether it's okay to use node types as keys (e.g., what if gets redefined in the REPL). It's okay, and we have other places in the code that use node types as keys, it's basically a small record wrapper around a keyword key.

@JCash, there was an issue with animation playback. It didn't play. I naively switched from `update-vertices!` to `spineext/plugin-update-vertices` in `produce-vertices` fn, but it caused all instances of spine node to play the same animation simultaneously. Weird stuff... It seems we need some state per spine node for playback in the editor, while we currently use a single shared handle per scene?..

@JCash, Mats left some comments about your changes. In particular: 
- [stencil support](https://github.com/defold/extension-spine/pull/277#discussion_r3361550952) — we no longer have it it seems
- [do we need to cover for absent index buffer?](https://github.com/defold/extension-spine/pull/277#discussion_r3363362143) — probably not? it seems when there is no index buffer, count it 0, so it would be a noop...
- [empty skin/animation](https://github.com/defold/extension-spine/pull/277/changes/BASE..3f1e4ff077e7a0bf266275f652d93c9656c341ef#r3363381190) — we skipped `plugin-set-skin` and `plugin-set-animation` before too, but perhaps we shouldn't?

vlaaad:
Everything is addressed as far as I'm aware, though @matgis, you had some comments about @JCash's changes that I think he would be better at addressing, the list is in the comment above

JCash:
> It seems we need some state per spine node for playback in the editor, while we currently use a single shared handle per scene?

Yes, I must have broken something along the way. SHouldn't be too hard to fix.

JCash (dismissed):

editor/src/clj/editor/gui.clj
matgis: The `:read-fn` will never be pure, since its result depends on the state of the resource on disk.

If we wanted to take this approach with `sanitize-collection`, we would need to update the `:read-fn` of both `.component` and `.go` whenever a component `resource-type` is updated. It all starts to feel a bit icky.

editor/src/clj/editor/gui.clj
```diff
@@ -700,18 +794,26 @@
...
+  (let [basis (:basis evaluation-context)
+        layout->prop->value (g/node-value node-id :layout->prop->value evaluation-context)
+        layout-property (contains? (get layout->prop->value "") prop-kw)
+        current-layout (get (g/node-value node-id :trivial-gui-scene-info evaluation-context) :current-layout "")]
     (assert (g/property-value-origin? basis node-id :layout->prop->override))
     (concat
-      (when (g/property-value-origin? basis node-id prop-kw)
+      (when (and (or (not layout-property)
+                     (str/blank? current-layout))
+                 (g/property-value-origin? basis node-id prop-kw))
```
matgis: This seems wrong? The `current-layout` shouldn't have any bearing here. Did we patch around a specific unit test case?

editor/src/clj/editor/gui.clj
```diff
@@ -4790,32 +5304,133 @@
...
+                   "Invalid %s argument: GUI custom property does not specify :protobuf-type. (node-type=%s, property=%s, custom-property-static=%s)"
+                   (:name @node-type) prop-kw custom-property-static)
+  (ensure/argument protobuf-type custom-property-pb-type->value-field
+                   "Invalid %s argument: Unsupported GUI custom property protobuf type. (type=%s)"
+                   protobuf-type)
+  (ensure/argument prop-kw layout-property-keys
+                   "Invalid %s argument: GUI custom property %s on %s is not a layout property."
+                   prop-kw (:name @node-type))
+  (assoc custom-property-static
+    :id prop-kw
```
matgis: Another place where `id` is a bit confusing. Maybe it should be `:prop-kw` instead? We use the term `custom-property-id->info` in a bunch of places, but it is really not an `id`, but a `prop-kw`.

editor/src/clj/editor/gui.clj
```diff
@@ -811,46 +921,59 @@
...
+   :custom-type-name ; Not overridden or necessary, but improves readability in files.
    :id ; Possibly overridden to reflect new position in scene hierarchy, but not listed among :overridden-fields. Used to locate the original node.
    :parent ; Possibly overridden to reflect new position in scene hierarchy, but not listed among :overridden-fields. No property exists for field.
    :template-node-child ; Signals that the node is overriding a node from a template scene. No property exists for field.
    :child-index ; Extra key sneaked into the node-desc to control order. No field exists for property.
+   :custom-properties ; The presence of an entry signals that the custom property is overridden.
    :overridden-fields]) ; Controls which pb-fields we apply override properties from when loading. No property exists for field.
 
+(defn- strip-custom-properties-from-overridden-fields [node-desc]
+  (if-not (contains? node-desc :overridden-fields)
```
matgis: No need for this check since we're using `assign-repeated`, right?

editor/src/clj/editor/gui.clj
```diff
@@ -1192,31 +1314,35 @@
             (dynamic tooltip (properties/tooltip-dynamic :gui :layer))
             (value (layout-property-getter layer))
             (set (layout-property-setter layer)))
+  (property custom-type-name g/Str (default (protobuf/default Gui$NodeDesc :custom-type-name))
+            (dynamic visible (g/constantly false)))
```
matgis: This probably shouldn't be a property, but an output that is overridden by e.g. `SpineNode`?

editor/src/clj/editor/gui.clj
```diff
@@ -435,19 +443,118 @@
...
+   :type-quat :quat
+   :type-string :string
+   :type-vector3 :vector3
+   :type-vector4 :vector4})
+
+(defn- custom-property-prop-value [custom-property-info custom-property]
+  (let [value-field (custom-property-pb-type->value-field (:protobuf-type custom-property-info))]
+    (if (contains? custom-property value-field)
+      (get custom-property value-field)
+      (:default custom-property-info))))
```
matgis: Use `get` with `::not-found` to avoid two lookups?

editor/src/clj/editor/gui.clj
```diff
@@ -799,10 +909,10 @@
       (when emit-warnings?
         (validate-contains-localized :warning error-gui-layer-from-template-not-found-in-scene-message-fn :layer node-id layer-names layer)))))
 
-(defn- overridden-pb-field-indices [gui-node]
-  (->> gui-node
-       (gt/overridden-properties)
+(defn- prop->overridden-pb-field-indices [prop->override custom-property-id->info]
```
matgis: This isn't really `custom-property-id->info`, but `custom-prop-kw->info`, right?

I think `id` should only refer to the string value we write to for the `id` field in the `Gui$Property`, and we should use the term `prop-kw` to refer to node properties.

We should clean up the use of the term `id` around custom properties before merging.

editor/src/clj/editor/gui.clj
```diff
@@ -3203,30 +3381,33 @@
...
+                         (into (prop->pb-field-entries-for-type type-info true prop->override))
+                         (clear-node-desc-defaults)
+                         (protobuf/assign-repeated :overridden-fields overridden-fields)
+                         (strip-unused-overridden-fields-from-node-desc)
+                         (cond-> node-id
+                           (assoc :id node-id)
+                           (not (str/blank? node-parent))
+                           (assoc :parent node-parent)
+                           (:template-node-child decorated-node-msg)
+                           (assoc :template-node-child true))))))))))
```
matgis, line 3389: Style: Group conditions and clauses in `cond->`?

editor/src/clj/editor/gui.clj
```diff
@@ -3254,16 +3435,28 @@
...
         node-descs
-        (coll/into-> node-msgs []
-          (map #(dissoc % :layout->prop->override :layout->prop->value))
-          (map (fn [node-desc]
-                 (if (:template-node-child node-desc)
-                   (strip-unused-overridden-fields-from-node-desc node-desc)
-                   (strip-redundant-size-from-node-desc node-desc)))))]
+        (mapv (fn [decorated-node-msg]
+                (let [type-info (node-desc->node-type-info gui-node-type-registry decorated-node-msg)
+                      node-msg (strip-node-msg-decorations decorated-node-msg)
```
matgis: This could be moved to the `false` branch of `if (:template-node-child node-msg)` below, since we use `select-keys` in the `true` branch.

editor/src/clj/editor/gui.clj
```diff
@@ -3636,6 +3863,13 @@
   (input project-settings g/Any)
   (input default-tex-params g/Any)
   (input exclude-gles-sm100 g/Any)
+  (input resource-types g/Any)
+  (output gui-node-type-registry g/Any :cached
+          (g/fnk [resource-types]
+            (gui-node-type-registry-from-resource-types resource-types)))
+  (output gui-resource-kind-registry g/Any :cached
+          (g/fnk [resource-types]
+            (gui-resource-kind-registry-from-resource-types resource-types)))
```
matgis: These outputs are trivial and don't really need to be `:cached`.

In fact, I'm leaning towards removing them altogether and instead including them in the `trivial-gui-scene-info` in the interests of keeping the number of connections down.

editor/src/clj/editor/gui.clj
```diff
@@ -3938,51 +4192,77 @@
...
+                          node-properties
+                          (persistent!
+                            (reduce
+                              (fn [node-properties {:keys [id] :as custom-property}]
+                                (if-let [custom-property-info (protobuf-id->custom-property-info id)]
+                                  (assoc! node-properties
+                                          (:id custom-property-info)
+                                          (custom-property-prop-value custom-property-info custom-property))
+                                  node-properties))
+                              (transient node-properties)
```
matgis, line 4188: Shouldn't we `dissoc` `:custom-properties` from `node-properties` here?

editor/src/clj/editor/gui.clj
```diff
@@ -3938,51 +4192,77 @@
...
-                       (pair prop-key
-                             (if (= ::not-found prop-value)
-                               (node-property-defaults prop-key)
-                               prop-value))))))
-        (:overridden-fields node-properties)))
+(defn- extract-overrides [type-info node-properties]
+  (let [custom-property-overrides
+        (when (contains? node-properties :custom-properties)
+          (coll/not-empty
+            (select-keys node-properties (keys (:custom-property-id->info type-info)))))
```
matgis: `custom-property-id->info` -> `custom-prop-kw->info`?

editor/src/clj/editor/gui.clj
```diff
@@ -3938,51 +4192,77 @@
...
+        regular-overrides (into {}
+                                (comp (remove #{custom-properties-pb-field-index})
+                                      (map pb-field-index->prop-key)
+                                      (remove non-overridable-properties)
+                                      (map (fn [prop-key]
+                                             (pair prop-key
+                                                   (if (contains? node-properties prop-key)
+                                                     (node-properties prop-key)
+                                                     (node-property-defaults prop-key))))))
+                                (:overridden-fields node-properties))]
```
matgis: Would be nicer to use `coll/into->` here.

editor/src/clj/editor/gui.clj
```diff
@@ -3938,51 +4192,77 @@
...
 
 (defn load-gui-scene [project self resource scene]
   {:pre [(map? scene)]} ; Gui$SceneDesc in map format.
-  (let [graph-id           (g/node-id->graph-id self)
+  (let [graph-id (g/node-id->graph-id self)
+        workspace (resource/workspace resource)
+        basis (g/now)
+        resource-types (resource/resource-types-by-type-ext basis workspace :editable)
+        gui-node-type-registry (gui-node-type-registry-from-resource-types resource-types)
+        gui-resource-kind-registry (gui-resource-kind-registry-from-resource-types resource-types)
```
matgis, line 4232: None of these really need the full map of `resource-types` from the `workspace`, since they ultimately only need information from the `.gui` `resource-type`. Why not use `(resource/resource-type resource)` instead?

This would also avoid the need for connecting the `workspace` `resource-types` to each `GuiSceneNode`, right?

editor/src/clj/editor/gui.clj
```diff
@@ -4112,6 +4388,39 @@
                                                                      :particlefx (resolve-resource (:particlefx particlefx-desc))]]
                                       (attach-particlefx-resource self particlefx-resources-node particlefx-resource)))))
 
+      (for [[resource-kind {:keys [label icon exts node-type resource-property attach-fn] :as resource-kind-info}] gui-resource-kind-registry
+            :let [resource-descs (filterv (fn [{:keys [path]}]
+                                            (contains? (set exts) (FilenameUtils/getExtension ^String path)))
```
matgis: Construct `(set exts)` outside of loop.

editor/src/clj/editor/gui.clj
```diff
@@ -4144,6 +4453,7 @@
                                      [:template-build-targets :template-build-targets]]]
                       (g/connect node-tree from self to))
                     (for [[from to] [[:trivial-gui-scene-info :trivial-gui-scene-info]
+                                     [:gui-node-type-registry :gui-node-type-registry]
```
matgis: If we really need it, we could pass it along with the `trivial-gui-scene-info` and save a bunch of connections.

editor/src/clj/editor/gui.clj
```diff
@@ -4250,50 +4559,80 @@
...
+      (when (and (contains? node-desc :custom-type)
+                 (not= (:custom-type node-type-info) (:custom-type node-desc)))
+        (throw (IllegalStateException.
+                 (format "GUI node custom_type_name '%s' resolves to custom_type %s, but the node specifies custom_type %s."
+                         custom-type-name (:custom-type node-type-info) (:custom-type node-desc))))))
+    (-> (if (not= :type-custom (:type node-desc))
+          (dissoc node-desc :custom-type-name :custom-type)
+          (-> node-desc
+              (assoc :custom-type-name (:custom-type-name node-type-info))
+              (dissoc :custom-type)))
```
matgis, line 4560: Should we fix the `:custom-type-name` before handing the `node-desc` to to the `convert-fn`?

editor/src/clj/editor/gui.clj
```diff
@@ -4250,50 +4559,80 @@
...
-
-(defn- sanitize-layout-node [node-desc]
-  (-> node-desc
-      (sanitize-node-fields)
+      (let [node-desc'' (-> node-desc'
+                            (strip-custom-properties-from-overridden-fields)
+                            (strip-redundant-size-from-node-desc))]
+        (cond-> node-desc''
+                (coll/empty? (:overridden-fields node-desc''))
+                (dissoc :overridden-fields))))))
```
matgis: We shouldn't have to deal with an empty `:overridden-fields` here. Instead, let's wrap the return value of `strip-size-from-overridden-fields` with `coll/not-empty`.

editor/src/clj/editor/gui.clj
```diff
@@ -4432,6 +4788,133 @@
     ;; make the node type a branch
     (attachment/alias workspace node-type :nodes GuiNode)))
 
+;; SDK api
+(defn gui-resource-kind-node [basis gui-scene-node resource-kind]
+  (coll/some
+    (fn [arc]
+      (let [node-id (gt/source-id arc)]
+        (when (and (= GuiResourceKindNode (g/node-type* basis node-id))
```
matgis: `(g/node-instance? basis GuiResourceKindNode node-id)`

editor/src/clj/editor/gui.clj
matgis: It looks like we're incurring quite a bit of complexity from maintaining this backwards compatibility?

For example, we maintain separate maps of `:custom-type-name->type-info` and `:type->custom-type->type-info`, maintain the old `register-node-type-info` alongside the new `register-custom-node-type-info`.

I don't think there's any shame in forcing users to update `extension-spine` by removing the legacy `register-node-type-info` if we wanted to get rid of this additional complexity.

editor/src/clj/editor/gui.clj
```diff
@@ -4759,14 +5270,17 @@
...
                                 (gui-attachment/scene-node->node-tree basis parent-node-id)
                                 parent-node-id)
-        props (assoc defaults
-                :child-index (gui-attachment/next-child-index node-tree-or-gui-node evaluation-context)
-                :custom-type custom-type
-                :type node-type)]
+        props (cond-> (assoc defaults
+                        :child-index (gui-attachment/next-child-index node-tree-or-gui-node evaluation-context)
+                        :type type)
+                custom-type-name (assoc :custom-type-name custom-type-name))]
```
matgis: `protobuf/assign` works like `assoc`, but will `dissoc` keys that supply `nil` values.

editor/src/clj/editor/gui.clj
```diff
@@ -4790,32 +5304,133 @@
...
+                (value (layout-property-getter spine-scene))
+                (set (layout-property-setter spine-scene))))
+
+    (register-custom-node-type-info
+      workspace
+      {:node-type SpineNode
+       :custom-type-name \"Spine\"
+       :display-name \"Spine\"
+       :icon spine-scene-icon
+       :defaults visual-base-node-defaults})"
```
matgis: Maybe clearer to be explicit in the example?
```
:icon "/defold-spine/editor/resources/icons/spine_scene.png"
:defaults gui/visual-base-node-defaults
```

matgis (changes requested):
It looks mostly good now, however there are some concerns and comments that I think we want to address before merging. Specifically,
* We use the term `id` to mean `prop-kw` or Protobuf string `id` in different places when used with custom properties.
* Introducing `gui-node-type-registry` as an additional connection between the `GuiSceneNode` and every `GuiNode` instead of passing it along with `trivial-gui-scene-info`.
* Having the outcome of `update-gui-resource-reference-impl` depend on the `current-layout` seems wrong?

editor/src/clj/editor/gui.clj
vlaaad: I guess we did, the problem is that a line below uses `(g/node-value node-id prop-kw evaluation-context)` which is current-layout aware, so we had an issue before too. I'll fix it!

editor/src/clj/editor/gui.clj
vlaaad: This can be completely deleted, actually, since `gui-node-type-registry` already has a mapping from node type to custom type name. I'll fix.

editor/src/clj/editor/gui.clj
vlaaad: It is needed for `extract-overrides` since it uses presence in custom properties as a sign that a property is overridden. `node-properties` is not the same thing as a `prop-kw->value` bundle; it has some extra properties too, like `template-node-child` and `parent`.

editor/src/clj/editor/gui.clj
vlaaad: I don't think so. `convert-fn`'s contract is to sanitize the description close to how it was read from the file. For example, spine extension handles legacy `:type-spine` and then migrates it to `:type-custom`, so its responsibility is to take an "old shape" and, if necessary, update it to "new shape". Fixing the custom type name before invoking the converter will be an obstacle in this case.

editor/src/clj/editor/gui.clj
vlaaad: Custom type here is needed for reading older formats with `type: TYPE_CUSTOM` and `custom_type: 405028931`. We don't force the re-save of such files unless they meaningfully change, so forcing people to update the extension-spine version is not enough here. We have such GUI nodes in refold games BTW, see https://github.com/search?q=org%3Arefold+%22custom_type%3A+405028931%22&type=code

I think it's normal to expect that such nodes will stay for a while longer, since they will be actually re-saved only when their gui files meaningfully change...

editor/src/clj/editor/gui.clj
vlaaad: I'd prefer to have a connection for proper invalidation, and then it's better to have `resource-types` consistently used as a source of registries.

matgis (approved):
LGTM! ❤️

PR #12468, opened 2026-05-27
Updated ruleset for how shader languages/versions are selected

Jhonnyg:
#### Technical changes

Centralizes bob’s shader language selection around the graphics adapters that are actually enabled for the target build. This fixes macOS default builds to emit SPIR-V only, keeps Android emitting SPIR-V + GLES, and preserves console/plugin behavior.

Rules for selecting shader languages to include in a build:

1. Start with platform defaults:
* macOS: Vulkan -> SPIR-V
* Android: Vulkan + OpenGLES -> SPIR-V + GLES
* other non-console platforms: OpenGL/OpenGLES

2. Apply manifest changes:
* symbols/libs add adapter languages
* excludeSymbols/excludeLibs remove adapter languages
* removals win

3. Apply explicit additions:
* bob debug flags force-add languages (mostly for tests)

4. Apply existing omissions
* shader.exclude_gles_sm100 removes only GLES 2.0/WebGL 1.0

Consoles:
* PS4/PS5/Xbox use plugin compilers
* Switch keeps SPIR-V through common compiler


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderCompilers.java
```diff
@@ -252,9 +416,18 @@ public ShaderProgramBuilder.ShaderCompileResult compile(ArrayList<ShaderCompileP
     }
 
     public static IShaderCompiler GetCommonShaderCompiler(Platform platform) {
+        if (platform == Platform.X86_64PS4 ||
+            platform == Platform.X86_64PS5 ||
+            platform == Platform.X86_64XBone) {
+            return null;
+        }
```
Jhonnyg, line 279: These platforms will install their own shader compiler plugin to do the shader compilation.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ExtenderUtil.java
```diff
@@ -1252,6 +1252,13 @@ public static boolean hasSymbol(String symbolName, Map<String, Object> platformS
         Map<String, Object> yamlPlatformContext = (Map<String, Object>) platformSettings.getOrDefault("context", null);
         if (yamlPlatformContext != null) {
             boolean symbolFound = false;
+            List<String> excludedSymbols = (List<String>) yamlPlatformContext.getOrDefault("excludeSymbols", new ArrayList<String>());
+            for (String symbol : excludedSymbols) {
+                if (symbol.equals(symbolName)) {
+                    return false;
+                }
+            }
```
Jhonnyg, line 1260: If the symbol we are looking for is excluded (e.g GraphicsAdapterVulkan is default on osx but excluded by a manifest file), we need to factor this in when parsing the manifest.

Jhonnyg:
@codex review pr

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/ShaderProgramBuilder.java
```diff
@@ -620,7 +623,9 @@ public static void main(String[] args) throws IOException, CompileExceptionError
 
             IShaderCompiler.CompileOptions compileOptions = new IShaderCompiler.CompileOptions();
             compileOptions.forceIncludeShaderLanguages.add(ShaderDesc.Language.LANGUAGE_GLSL_SM120);
+            compileOptions.forceIncludeShaderLanguages.add(ShaderDesc.Language.LANGUAGE_GLSL_SM330);
             compileOptions.forceIncludeShaderLanguages.add(ShaderDesc.Language.LANGUAGE_GLES_SM100);
+            compileOptions.forceIncludeShaderLanguages.add(ShaderDesc.Language.LANGUAGE_GLES_SM300);
```
Jhonnyg, line 628: Needed because we are now a bit stricter with which shaders we bundle, but we however need to include all shader languages here because we build a couple of resources directly into the binaries (e.g some debug shaders I think). I'm not sure if we can avoid compiling resources into the binary, but maybe we could compile them into separate libraries or something.. Not sure.

JCash (approved):

Issue #11421, opened 2025-10-25
[SUGGESTION] Preferences keymap.


dprogrb:
Здравствуйте!

Предлагаю добавить возможность переключать фокус ввода с диалогового окна Preferences на редактор Defold. Чтобы можно было комфортно пробовать keymap в редакторе.
На данный момент, чтобы воспользоваться текстовым редактором при открытом Preferences, необходимо закрыть это окно. Это усложняет возможность знакомство с горячими клавишами редактора.

Hello
I propose adding the ability to switch input focus from the Preferences . This will make it easier to select the keyboard layout in the editor.
Currently, to use the text editor with the settings window open, you need to close this window Preferences . This makes it difficult to learn the editor's hotkeys.

<img width="1273" height="661" alt="Image" src="https://github.com/user-attachments/assets/bddf7b04-37d2-4b2f-a32e-7e5dd13ba864" />

JosephFerano:
@dprogrb 

We discussed it briefly. There are some potential features we can consider. The first would be more involved and it would be introducing a general "window dock/pop out" feature where windows can be made into OS windows. The other is to introduce a reminder system whenever you performance an action without a shortcut. Jetbrains' IDEs have a plugin called [Key Promoter](https://plugins.jetbrains.com/plugin/9792-key-promoter-x) that creates a notification to assist you in learning editor shortcuts.

However, because both those would take time, a workaround in the meantime I can suggest is to copy [the keybindings in the windows section](https://github.com/defold/defold/blob/dev/editor/src/clj/editor/keymap.clj#L171-L285) I highlighted, and paste it into a separate file, this way you can at least still search and while it's not as pretty, at least all the defaults are there.

PR #12470, opened 2026-05-27
Add non-modal option to dialogs created by editor scripts

### LINKED ISSUE #11439: Add possibility to have interactive editor reacting to input when Editor UI popup is opened
Current Editor UI popup is gathering all inputs, so Editor beneath is not receiving/reacting to any input. This doesn't allow to make an action "in the Editor" without closing the popup, so doesn't allow to make complex popups that would require also some action to be taken meanwhile in the Editor. This could allow to make interactive tutorials or tools that interact with the Editor, enhancing it's possibilities, like some cool "Auto-tiling configurator" while Tilemap Editor is in the background or some "Atlas enhanced tools", when atlas is beneath.

I would love to have a possibility to have a Popup that is allowing user to also still operate in the Editor freely, while it being itself on top and still visible to user.

Somehow pass input to Editor? Or invoke certain action in Editor from the Popup?

Example use case:
Check if user opened a file in Editor, while popup is still up:

Image

vlaaad:
We added a `modal` option to `editor.ui.dialog` in editor scripts. Set `modal = false` to keep the dialog open while users continue working in the editor behind it.

This is useful for editor-script tools, tutorials, and assistants that need to stay visible while the user selects files, edits scenes, or changes properties in the main editor.

Fix #11439

### DIFF
--- a/editor/src/clj/editor/editor_extensions/ui_components.clj
+++ b/editor/src/clj/editor/editor_extensions/ui_components.clj
@@ -835,10 +835,11 @@
       (fx/run-later
         (future/complete!
           f
-          (fxui/show-dialog-and-await-result!
-            :event-handler #(assoc %1 ::fxui/result (:result %2))
-            :error-handler #(future/fail! f %)
-            :description desc)))
+          (rt/and-refresh-context
+            (fxui/show-dialog-and-await-result!
+              :event-handler #(assoc %1 ::fxui/result (:result %2))
+              :error-handler #(future/fail! f %)
+              :description desc))))
       f)))
 
 ;; endregion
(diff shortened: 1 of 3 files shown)


JosephFerano:
There was an issue from a while back ago that requested something similar but applied to the Preferences dialog.
#11421
This is for editor scripts but I wonder if this can be applied there in regards to the user's feature request. Would we want such a change?

vlaaad:
I think it would be good if the Preferences dialog were non-modal, though it would require refactoring it to apply the keymap changes as soon as shortcuts are edited. I think we should do that as a separate task from this.

JosephFerano (approved):
It works!

PR #12474, opened 2026-05-28
Allow setting width, height, and resizable properties on dialogs

### LINKED ISSUE #11967: Allow to define width and height of the popup editor UI
The width of Editor UI popups is fixed. Thei height is detemined by its content.

Full control over the layout of the window, allowing to set width and/or height specific to the popup.

Add options to resize the popup window to user's liking.

vlaaad:
Now it's possible to create resizable dialogs with predefined width and/or height using editor scripts!

```lua
editor.ui.dialog({title = "Dialog", width = 720, height = 600})
editor.ui.dialog({title = "Width-only dialog", width = 720})
editor.ui.dialog({title = "Height-only dialog", height = 600})
```

Fixes #11967


JosephFerano:
I noticed that on Linux I can set the size of the window to 1x1 or 2x2 (testing user error) and it will maintain a minimum area of something in the order of 100x100. However, if I set it to 3x3 or higher, these dimensions are respected and I get a tiny little box. This might be my window manager's behavior (X11 QTile). Would it make sense to clamp at some reasonably sized minimum?

vlaaad:
Heh, strange, on macOS I can create 1x1 dialogs. While it's obviously a nonsensical size, as well as 100000x100000, I think it's fine to allow it, as I expect this error will be very easy to see and fix.

PR #12477, opened 2026-05-28
Add instancing support for morph targets

### LINKED ISSUE #12219: Morph target MVP2
There are some things we need to polish and fix to make this a more solid feature:

* move the array texture packaging to the asset pipeline. right now we produce weights from the gltf files, pack these into float arrays and then copy them into array texture slices. all of that can be moved to the pipeline instead
* instancing support. morph targets and instancing does not work currently. a hash of the weights are included into the render item hash for instanced render items, which means that models with different weights (e.g animation state) will break batches. we should introduce a new semantic (similar to bone weights) so that morphed animations can leverage instancing as well

Jhonnyg:
Models using morph targets can now be instanced by using the new `SEMANTIC_TYPE_MORPH_TARGET_WEIGHTS` vertex attribute semantic type. Similar to `mtx_world`, `mtx_normal`, specifying a vertex attribute in the shader with the name `morph_targets_weights` will signal the engine that the morph target weights will be produced in a per-instance buffer:

```glsl
#version 140
// Per-vertex attributes
in highp vec4 position;
in mediump vec2 texcoord0;
in mediump vec3 normal;
// Per-instance
in mediump mat4 mtx_world;
in mediump mat4 mtx_normal;
in mediump mat4 morph_targets_weights;
// Morph target delta values
uniform sampler2DArray morph_targets;
```

Fixes #12219


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/MeshsetBuilder.java
```diff
@@ -69,6 +78,27 @@ public Task create(IResource input) throws IOException, CompileExceptionError {
...
+
+            boolean split_meshes = this.project.option("model-split-large-meshes", "false").equals("true");
+            if (split_meshes) {
+                ModelUtil.splitMeshes(scene);
+            }
+
+            int morphTargetTextureCount = ModelUtil.getNumMorphTargetTextures(scene);
+            for (int i = 0; i < morphTargetTextureCount; ++i) {
+                taskBuilder.addOutput(input.changeExt(String.format("_morph_%d.texturec", i)));
+            }
```
Jhonnyg, line 119: Each sub-mesh can potentially require a morph target texture resource, which we can't know until we have actually loaded the scene. I don't think there's a way around this problem unfortunately 🤔

engine/render/src/render/material.cpp
```diff
@@ -133,6 +133,14 @@ namespace dmRender
...
             info->m_CoordinateSpace = dmGraphics::COORDINATE_SPACE_LOCAL;
         }
+        else if (name_hash == VERTEX_STREAM_MORPH_TARGET_WEIGHTS)
+        {
+            info->m_SemanticType = dmGraphics::VertexAttribute::SEMANTIC_TYPE_MORPH_TARGET_WEIGHTS;
+            if (instancing_supported)
+            {
+                info->m_StepFunction = dmGraphics::VERTEX_STEP_FUNCTION_INSTANCE;
+            }
+        }
```
Jhonnyg, line 143: Default settings for the stream

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/AtlasBuilder.java
```diff
@@ -97,8 +97,7 @@ public void build(Task task) throws CompileExceptionError, IOException {
         TextureSetResult result            = AtlasUtil.generateTextureSet(this.project, task.firstInput(), builder);
         TextureImage.Type textureImageType = getTexureType(builder);
 
-        int buildDirLen         = project.getBuildDirectory().length();
-        String texturePath      = task.output(1).getPath().substring(buildDirLen);
+        String texturePath      = BuilderUtil.getResourcePathFromOutput(project, task.output(1));
```
Jhonnyg: There was a pattern here that we use here and there, so it's been refactored into a helper function instead.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/MeshsetBuilder.java
```diff
@@ -27,15 +27,46 @@
...
+        "model-split-large-meshes",
+        "model-max-morph-target-texture-width",
+        "model-max-morph-target-texture-height"})
 public class MeshsetBuilder extends Builder  {
+    private static class MorphTargetTextureOutput {
+        IResource resource;
+        ModelUtil.PackedMorphTargetTexture texture;
+    }
+
+    private static class BobMorphTargetTextureCollector implements ModelUtil.MorphTargetTextureCollector {
```
Jhonnyg: Editor and bob have their own implementation of this interface. The idea here is that this class will act as a container that gathers all the compiled morph target textures when a model is loaded. A model can contain multiple such morph target textures, so we need to store a list of the resources somewhere.

An alternative solution here would be if the ModelUtil.loadScene returns the DDF rig and a list of morph target textures?

Jhonnyg:
@codex review pr

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1367,18 +1367,40 @@ private Future buildRemoteEngine(IProgress progress, ExecutorService executor, A
      *  Options from the `game.project` file that may affect build outputs.
      */
     private static class GameProjectBuildOption {
+        private enum ValueType {
+            BOOLEAN,
+            INTEGER
+        }
```
Jhonnyg, line 1373: Added since I want to use the same GameProjectBuildOptions class for both the true/false options and other data types.

editor/src/clj/editor/rig.clj
```diff
@@ -45,24 +48,74 @@
...
+                      (update model :meshes
+                              (fn [meshes]
+                                (mapv (fn [mesh]
+                                        (if-let [morph-target-texture (coll/not-empty (:morph-target-texture mesh))]
+                                          (if-not (string/starts-with? morph-target-texture "__morph_target_texture_")
+                                            mesh
+                                            (assoc mesh :morph-target-texture (token->texture-resource-path morph-target-texture)))
+                                          mesh))
+                                      meshes))))
+                    models)))))
```
Jhonnyg: Dear editor devs: can this be rewritten in some simpler way? The point of the code is to set the resource output for the morph target textures for the meshes that require them. But it feels like the function could be rewritten in some way to simplify this.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/BuilderUtil.java
```diff
@@ -49,4 +49,18 @@ public static IResource checkResource(Project project, IResource owner, String f
...
+     * For example, "build/default/main/generated.texturec" becomes "/main/generated.texturec".
+     */
+    public static String getResourcePathFromOutput(Project project, IResource outputResource) {
+        String path = outputResource.getPath();
+        String buildDirectory = project.getBuildDirectory();
+        String buildDirectoryPrefix = buildDirectory + "/";
+        if (!path.startsWith(buildDirectoryPrefix)) {
+            throw new IllegalArgumentException(String.format("Expected output resource path '%s' to start with build directory '%s'", path, buildDirectoryPrefix));
+        }
+        return path.substring(buildDirectory.length());
```
JCash: I think using common names, like `Path::relativize`, is good, to help devs find their way in our code.
Suggestions:
* relativize
* getRelativePath

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/AtlasBuilder.java
JCash: While I'm not arguing against it in principle, I also feel the `IResource::getPath()` (i.e. the relative path) should suffice here? Presumably we should be able to use that, depending on what the output resources think of as their root?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/Project.java
```diff
@@ -1507,22 +1529,24 @@ public static String getShaderAdaptersOption(Platform platform, List<Map<String,
...
+        options.add(new GameProjectBuildOption("output-glsl430", "output-glsl430", "shader", "output_glsl430", GameProjectBuildOption.ValueType.BOOLEAN));
+        options.add(new GameProjectBuildOption("exclude-gles-sm100", "exclude-gles-sm100", "shader", "exclude_gles_sm100", GameProjectBuildOption.ValueType.BOOLEAN));
+
+        options.add(new GameProjectBuildOption("sound-stream-enabled", "sound-stream-enabled", "sound", "stream_enabled", GameProjectBuildOption.ValueType.BOOLEAN));
+        options.add(new GameProjectBuildOption("model-split-large-meshes", "model-split-large-meshes", "model", "split_meshes", GameProjectBuildOption.ValueType.BOOLEAN));
+        options.add(new GameProjectBuildOption("model-max-morph-target-texture-width", "model-max-morph-target-texture-width", "model", "max_morph_target_texture_width", GameProjectBuildOption.ValueType.INTEGER));
+        options.add(new GameProjectBuildOption("model-max-morph-target-texture-height", "model-max-morph-target-texture-height", "model", "max_morph_target_texture_height", GameProjectBuildOption.ValueType.INTEGER));
```
JCash, line 1547: 👍

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -818,13 +825,38 @@ namespace dmGameSystem
...
+        }
+        return false;
+    }
+
+    static void FillMorphWeightsFloatSlots(const float* weights, uint32_t weight_count, float* out, uint32_t slot_count)
+    {
+        for (uint32_t i = 0; i < slot_count; ++i)
+        {
+            out[i] = i < weight_count ? weights[i] : 0.0f;
+        }
```
JCash, line 851: If the memory is 0 initialized (and persistent), shouldn't it be enough with a `memcpy()`?

engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -837,6 +869,32 @@ namespace dmGameSystem
 
         uint32_t uv_channels_count = (uv_channels[0] ? 1 : 0) + (uv_channels[1] ? 1 : 0);
 
+        if (GetMorphTargetWeightsAttributeVectorType(attribute_infos, step_function, &morph_target_weights_vector_type))
+        {
+            const uint32_t weight_capacity = dmGraphics::VectorTypeToElementCount(morph_target_weights_vector_type);
+            const uint32_t vec4_slots      = (weight_capacity + 3) / 4;
+            const float* weights           = 0;
```
JCash, line 879: When looking at our code, I realize that our dmArray<float> uses `new uint8_t[]` to allocate, which only gives 1-byte alignment. That is likely not good for any loops/memcopy performance.

We should fix dmArray (etc) to make sure we pass the `alignof(T)` to the `SetCapacity()` function, and handle it there.

engine/gamesys/src/gamesys/resources/res_meshset.cpp
```diff
@@ -41,6 +41,19 @@ namespace dmGameSystem
...
+            const dmRigDDF::Model* model = &MeshSet->m_Models[mi];
+            for (uint32_t mesh_i = 0; mesh_i < model->m_Meshes.m_Count; ++mesh_i)
+            {
+                const dmRigDDF::Mesh* mesh = &model->m_Meshes[mesh_i];
+                if (mesh->m_MorphTargetTexture && mesh->m_MorphTargetTexture[0])
+                {
+                    dmResource::PreloadHint(params->m_HintInfo, mesh->m_MorphTargetTexture);
+                }
+            }
+        }
```
JCash, line 55: Off topic: Sometimes I also wish our file formats/resource system could do this instead.
E.g. we wrap the resource in a "dmResourceDDF", which has fields for dependencies, and the resource system itself could call PreloadHint on resources.

JCash (dismissed):

editor/src/clj/editor/rig.clj
vlaaad: Perhaps something like this?
```clj
#_(let [token->texture-resource-path #(str "<<" % ">>")
        mesh-set {:models [{:meshes [{:morph-target-texture "__morph_target_texture_0"}]}]}]
    (update mesh-set :models coll/mapv-> 
            update :meshes coll/mapv-> 
            (fn [mesh]
              (if-let [morph-target-texture (coll/not-empty (:morph-target-texture mesh))]
                (if-not (string/starts-with? morph-target-texture "__morph_target_texture_")
                  mesh
                  (assoc mesh :morph-target-texture (token->texture-resource-path morph-target-texture)))
                mesh))))
  ;;=> {:models [{:meshes [{:morph-target-texture "<<__morph_target_texture_0>>"}]}]}
```

editor/src/clj/editor/rig.clj
```diff
@@ -45,24 +48,74 @@
      :build-fn build-animation-set
      :user-data {:animation-set animation-set}}))
 
+(defn- replace-morph-target-texture-tokens
+  [mesh-set morph-target-textures dep-resources]
+  (let [token->texture-resource-path
+        (into {}
+              (map (fn [{:keys [token resource]}]
+                     [token (resource/proj-path (get dep-resources resource resource))]))
+              morph-target-textures)]
```
vlaaad: Could use `coll/pair-map-by`

editor/test/integration/model_utility_test.clj
```diff
@@ -13,18 +13,44 @@
...
+  (:import [com.dynamo.bob.util TextureUtil]
+           [com.dynamo.rig.proto Rig$MeshSet]
+           [java.nio.file Files]))
 
 (defn- load-scene [workspace project file-path]
   (let [resource (workspace/file-resource workspace file-path)
         node-id (project/get-resource-node project resource)]
     (model-loader/load-scene node-id resource (project/settings project))))
 
+(defn- build-target! [project build-target]
```
vlaaad: Worth investigating: I think this function might be unnecessary. Could we use `test-util/build-node!` here?

vlaaad (commented):
LGTM, see comments

editor/src/clj/editor/rig.clj
```diff
@@ -51,23 +51,18 @@
...
-                                            (assoc mesh :morph-target-texture (token->texture-resource-path morph-target-texture)))
-                                          mesh))
-                                      meshes))))
-                    models)))))
+        (coll/pair-map-by :token
+                          (fn [{:keys [resource]}]
+                            (resource/proj-path (get dep-resources resource resource)))
+                          morph-target-textures)]
+    (update mesh-set :models coll/mapv->
+            (update :meshes coll/mapv->
```
vlaaad: I think there should be no `(` on this line...

editor/test/integration/test_util.clj
```diff
@@ -1389,16 +1389,16 @@
...
         workspace (project/workspace project)
         old-artifact-map (workspace/artifact-map workspace)]
     (g/with-auto-evaluation-context evaluation-context
-      (build/build-project! project resource-node old-artifact-map nil evaluation-context))))
+      (build/build-project! project resource-node old-artifact-map opts evaluation-context))))
 
-(defn build-node! [resource-node]
-  (let [build-result (build-node-result! resource-node)]
-    (when-some [error (:error build-result)]
+(defn build-node! [resource-node opts]
```
vlaaad, line 1399: I'm surprised the CI succeeded... `build-node!` is used in other places. You need to update the callers of `build-node!` to provide a second argument. We use it in dev.clj during development, so it didn't fail the build, but it breaks the dev tool.

vlaaad (changes requested):
LGTM, but please update all `build-nodes!` users to supply the needed args.

vlaaad (approved):

PR #12480, opened 2026-05-28
Fix POSIX crash signal chaining

AGulev:
Restores previous signal handlers correctly and fixes macOS crash reraising so crashes keep the original signal behavior.


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/bundle/MacOSBundler.java
```diff
@@ -192,6 +192,13 @@ public void bundleApplication(Project project, Platform platform, File bundleDir
 
         BundleHelper.throwIfCanceled(canceled);
 
+        // Copy PrivacyInfo.xcprivacy
+        // According to https://developer.apple.com/documentation/bundleresources/adding-a-privacy-manifest-to-your-app-or-third-party-sdk
+        // the PrivacyInfo.xcprivacy on macOS should be in the `Resources` folder.
+        BundleHelper.copyPrivacyManifest(project, platform, resourcesDir);
+
+        BundleHelper.throwIfCanceled(canceled);
+
```
AGulev, line 201: copy privacy manifest before signing

engine/crash/src/backtrace_signal_posix.h
```diff
@@ -0,0 +1,213 @@
...
+        struct sigaction* previous_signal_action = GetPreviousSignalAction(signum, previous_signal_actions);
+        if (!previous_signal_action)
+        {
+            return;
+        }
+
+        sigaction(signum, previous_signal_action, 0);
+    }
+
+    static inline bool ChainSignalToPreviousHandler(int signum, siginfo_t* si, void* sc, struct sigaction* previous_signal_actions, FSignalAction current_handler)
```
AGulev: If another handler was installed before Defold, call it after Defold writes the crash dump.

engine/crash/src/backtrace_signal_posix.h
```diff
@@ -0,0 +1,213 @@
...
+            return false;
+        }
+
+        if ((previous_signal_action->sa_flags & SA_SIGINFO) &&
+            previous_signal_action->sa_sigaction &&
+            previous_signal_action->sa_handler != SIG_DFL &&
+            previous_signal_action->sa_handler != SIG_IGN &&
+            previous_signal_action->sa_sigaction != current_handler)
+        {
+            RestoreSignalHandler(signum, previous_signal_actions);
```
AGulev: Restore the old handler before calling it, so we do not loop back into Defold’s handler.

engine/crash/src/backtrace_signal_posix.h
```diff
@@ -0,0 +1,213 @@
...
+        return false;
+    }
+
+    static inline bool WasPreviousSignalIgnored(int signum, struct sigaction* previous_signal_actions)
+    {
+        struct sigaction* previous_signal_action = GetPreviousSignalAction(signum, previous_signal_actions);
+        return previous_signal_action && previous_signal_action->sa_handler == SIG_IGN;
+    }
+
+    static inline bool ShouldReturnToOriginalFault(int signum, const siginfo_t* si)
```
AGulev: On Apple, only return to the original fault for real crash signals. User-raised signals still use raise.

engine/crash/src/backtrace_signal_posix.h
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    static inline void RaiseDefaultSignalHandler(int signum, const siginfo_t* si)
+    {
+        if (!IsValidSignal(signum))
+        {
+            return;
+        }
+
+#if defined(__linux__)
```
AGulev: On Linux, try to re-send the same signal info to keep the original crash details.

engine/crash/src/backtrace_signal_posix.h
```diff
@@ -0,0 +1,213 @@
...
+        if (WasPreviousSignalIgnored(signum, previous_signal_actions))
+        {
+            RestoreSignalHandler(signum, previous_signal_actions);
+            return;
+        }
+
+        RaiseDefaultSignalHandler(signum, si);
+    }
+
+    static inline bool InstallSignalHandler(int signum, FSignalAction handler, struct sigaction* previous_signal_actions)
```
AGulev: If Defold’s handler is already installed, do not overwrite the saved previous handler.

engine/crash/src/backtrace_signal_posix.h
AGulev: moved into a separate file to avoid defines

engine/crash/src/backtrace_signal_posix.h
AGulev: moved into a separate file to avoid defines

engine/crash/src/backtrace_execinfo.cpp
```diff
@@ -100,25 +101,25 @@ namespace dmCrash
...
+            return;
+        }
+
+        // The default behavior is restored for the signal.
         // Unless this is done first thing in the signal handler we'll
         // be stuck in a signal-handler loop forever.
-        sigaction(signum, &sigdfl[signum], NULL);
+        ResetToDefaultSignalHandler(signum);
         OnCrash(signum);
+        ChainSignalOrRaiseDefault(signum, si, sc, g_PreviousSignalActions, Handler);
```
JCash: This looks like it can be simplified:
```
ResetToDefaultSignalHandler(...);

if (g_CrashDumpEnabled)
{
  OnCrash(signum
}

ChainSignalOrRaiseDefault(...);
```

engine/crash/src/backtrace_libunwind.cpp
```diff
@@ -196,36 +199,27 @@ namespace dmCrash
...
+            return;
+        }
+
         // The default behavior is restored for the signal.
         // Unless this is done first thing in the signal handler we'll
         // be stuck in a signal-handler loop forever.
-        ResetToDefaultHandler(signum);
+        ResetToDefaultSignalHandler(signum);
         OnCrash(signum);
+        ChainSignalOrRaiseDefault(signum, si, sc, g_PreviousSignalActions, Handler);
```
JCash: same comment here

engine/crash/src/backtrace_libunwind_ndk.cpp
```diff
@@ -94,20 +96,14 @@ namespace dmCrash
...
-        sa.sa_handler = SIG_DFL;
-        sa.sa_flags = 0;
-        sigaction(signum, &sa, NULL);
-    }
-
     static void Handler(const int signo, siginfo_t* const si, void *const sc)
     {
         if (!g_CrashDumpEnabled)
+        {
+            ResetToDefaultSignalHandler(signo);
```
JCash: The `ResetToDefaultSignalHandler()` can be moved outside of this if-statement.

engine/crash/src/backtrace_signal_posix.h
```diff
@@ -0,0 +1,160 @@
...
+namespace dmCrash
+{
+    static const int MAX_SIGNAL_COUNT = 64;
+
+    typedef void (*FSignalAction)(int, siginfo_t*, void*);
+
+    static inline bool IsValidSignal(int signum)
+    {
+        return signum > 0 && signum < MAX_SIGNAL_COUNT;
+    }
```
JCash: While small functions like this can be inlined like this, I don't think the rest of the functions here are ever going to be inlined. Also, that's not how we write our headers.

Create a backtrace_signal_posix.cpp for the implementation.

engine/crash/src/backtrace_signal_posix.h
JCash: I think such comments should be code comments.

JCash (changes requested):
Overall good I guess. Hard to know until we get the correct behavior.
Some code cleanup needed though.
Also missing a unit test for this?

engine/crash/src/backtrace_execinfo.cpp
AGulev: fixed!

engine/crash/src/backtrace_libunwind.cpp
AGulev: fixed!

engine/crash/src/backtrace_libunwind_ndk.cpp
AGulev: fixed!

engine/crash/src/backtrace_signal_posix.h
AGulev: fixed

engine/crash/src/backtrace_signal_posix.h
AGulev: added

engine/crash/src/test/test_crash_posix_signal.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+static volatile sig_atomic_t g_PreviousSignalActionCalls = 0;
+static volatile sig_atomic_t g_PreviousSignalActionWasRestored = 0;
+
+static void TestCurrentSignalAction(int signum, siginfo_t* si, void* sc)
+{
+    (void)signum;
+    (void)si;
+    (void)sc;
+}
```
JCash, line 78: I kind of expected this to also store some state, for the test to make sure they occur in the correct order?

engine/crash/src/test/test_crash_posix_signal.cpp
```diff
@@ -0,0 +1,163 @@
...
+    sigaction(signum, 0, &current);
+
+    g_PreviousSignalActionCalls++;
+    g_PreviousSignalActionWasRestored = (current.sa_flags & SA_SIGINFO) && current.sa_sigaction == TestPreviousSignalAction;
+}
+
+static void TestPreviousSignalHandler(int signum)
+{
+    (void)signum;
+}
```
JCash, line 103: Same here, shouldn't we check that it's either called or _not_ called?

JCash (dismissed):
Tentatively approved.
I think the test should verify all callbacks?

AGulev:
I added all the missing tests

PR #12487, opened 2026-05-29
Add Code location for editor script commands

### LINKED ISSUE #11529: Add 'Editor' view to list of allowed locations for editor's scripts
Enable to add editor scripts in the scene context, available from a context menu opened with `Right Mouse Button`:
Image 

Image

I want to have a quick access to add custom elements to the scene (like adding player, Cube, ground, camera, primitives, and so on - everything that might help designing levels in Defold Editor). It might be also useful for the Code Editor context menu, etc.

I want add custom command in a context menu that is available in "Editor" pane (for collection scene editor, code editor, etc).

```
function M.get_commands()
    return {
        {
            label = "Add Cube 3D",
            locations = { "Editor" },  -- add here possibility to use "Editor"?
            run = function (opts)
            end
        }
    }
end
```

Similar: https://github.com/defold/defold/issues/11528

Code view context menu: https://github.com/defold/defold/pull/10768

vlaaad:
Now commands that target `Code` location will show up in the Code editor context menu.

Fix #11529

### DIFF
--- a/editor/.clj-kondo/hooks/clojure_core.clj
+++ b/editor/.clj-kondo/hooks/clojure_core.clj
@@ -1,0 +1,42 @@
+;; Copyright 2020-2026 The Defold Foundation
+;; Copyright 2014-2020 King
+;; Copyright 2009-2014 Ragnar Svensson, Christian Murray
+;; Licensed under the Defold License version 1.0 (the "License"); you may not use
+;; this file except in compliance with the License.
+;;
+;; You may obtain a copy of the License, together with FAQs at
+;; https://www.defold.com/license
+;;
+;; Unless required by applicable law or agreed to in writing, software distributed
+;; under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+;; CONDITIONS OF ANY KIND, either express or implied. See the License for the
+;; specific language governing permissions and limitations under the License.
+
+(ns hooks.clojure-core
+  (:refer-clojure :exclude [map])
+  (:require [clj-kondo.hooks-api :as api]))
+
+(defn- map-fn-node [map-node arity]
+  (let [args (mapv #(api/token-node (symbol (str "_x" % "__"))) (range arity))]
+    (api/list-node
+      [(api/token-node 'fn)
+       (api/vector-node args)
+       (api/list-node
+         (list* (api/token-node 'get)
+                map-node
+                args))])))
+
+(defn map [{:keys [node]}]
+  (let [[map-symbol-node f-node & coll-nodes] (:children node)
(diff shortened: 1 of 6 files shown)


editor/.clj-kondo/hooks/clojure_core.clj
```diff
@@ -0,0 +1,42 @@
...
+                  1 1
+                  2 2
+                  nil)]
+    {:node
+     (if (and f-arity (= :map (:tag f-node)))
+       (api/list-node
+         (list* map-symbol-node
+                (map-fn-node f-node f-arity)
+                coll-nodes))
+       node)}))
```
vlaaad, line 42: Weird, but clj-kondo complained about the following form in commands.clj: `(into #{} (map {...}) locations)`, that map literal is invalid arity. This fixes the false-positive.

editor/src/clj/editor/editor_extensions/commands.clj
```diff
@@ -164,17 +163,19 @@
         contexts (into #{}
                        (map {"Assets" :asset-browser
                              "Bundle" :global
+                             "Code" :code-view
                              "Debug" :global
                              "Edit" :global
                              "Outline" :outline
                              "Project" :global
-                             "Scene" :global
+                             "Scene" :workbench
```
vlaaad, line 171: No change in behavior, I tried commands with various selections, and global and workbench contexts work the same in the editor. It's just a bit clearer since scene commands register themselves on a workbench.

vlaaad:
@codex review

editor/src/clj/editor/editor_extensions/commands.clj
```diff
@@ -164,17 +163,19 @@
         contexts (into #{}
                        (map {"Assets" :asset-browser
                              "Bundle" :global
+                             "Code" :code-view
```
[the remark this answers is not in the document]
vlaaad: I carefully considered this feedback. The thing is, we might want to introduce selection as selected text or text ranges. And you can query only 1 selection, because semantically, in a particular context, there is only 1 thing called selection. In the Assets pane, you can query the selected resource, because we have a tree view of resources with selection. In the scene view, you can query the selected resource, but it's not the same as the opened file — you might have a referenced go in a collection view, and selecting this go will report the go file as a selected resource. So the concept of "selected resource" is not applicable to the Code view — there are no resources to select, and if there is a selection, it's probably not a resource at all. 

A way to get the current resource is to introduce a new query, called `owner_resource` or something similar, which would report the current resource. Let's do it if there is a demand for it.

JosephFerano (approved):
Image

Works!

PR #12489, opened 2026-05-29
Show status bar message on editor script reload

### LINKED ISSUE #11182: No feedback for Reload Editor Scripts
Currently, when I press Shift+Cmd+R, the editor gives no feedback if the action succeeds.  
It would be helpful to see confirmation - either a console log saying the script was reloaded successfully or a message in the status bar.

vlaaad:
The editor now provides visible feedback in the status bar whenever editor scripts are reloaded.

Fix #11182


AGulev (approved):

JosephFerano (approved):
Works! Nice and subtle.

PR #12497, opened 2026-06-01
Editor: Color dropper from grid popup

### LINKED ISSUE #12519: Color Picker refactor to suppor Color Dropper in Grid Settings Popup
The color dropper icon does not show up for the color picker in the grid settings popup. It was avoided because there is some tight coupling to the graph that can be decoupled by not using a graph system node and instead use a `defrecord` stored on UI user data. Then the settings popup can create its own color dropper popup.

- [X] Refactor color dropper
- [x] Fix color dropper fighting for focus with the popup control.

Current
Image

What it should have

Image

JosephFerano:
You can now use the color dropper from the grid settings popup in the scene view toolbar.

image

Fixes #12519

### Technical changes

This PR changes ColorDropper from being a graph node to a defrecord to avoid the need to pass around graph dependencies in UI code. The color picker's jfx component then handles the clean up. Also needed to add some special handling for the picker in the settings popup to get it to work.


editor/src/clj/editor/color_dropper.clj
```diff
@@ -27,11 +27,12 @@
 (set! *warn-on-reflection* true)
 (set! *unchecked-math* :warn-on-boxed)
 
-(g/defnode ColorDropper
-  (property color Color)
-  (property dropper-area StackPane)
-  (property size-listener ChangeListener)
-  (property image WritableImage))
+(defrecord ColorDropper [^Color color
```
vlaaad: defonce/record

vlaaad (commented):
Can we simplify the lifecycle if we don't hide the popup? Or is this unavoidable? Also, for some reason, the cursor in the toolbar does not change to ☝️ when hovering over the color dropper icon, unlike in the properties view. Sorry for the terrible video from the phone, but somehow, it doesn't change at all when I do screen recording...

https://github.com/user-attachments/assets/d2001a6f-6cc1-4841-917b-13faef29c362

JosephFerano:
NOTE: Something weird happened, where my reply replaced the original message, so I restored it and separated the comment.

> Can we simplify the lifecycle if we don't hide the popup? Or is this unavoidable?

Yeah, this lifecycle stuff came out kinda funky. I tried to simplify it and there might be better ways. This is what happens when the popup is not hidden;

https://github.com/user-attachments/assets/c56a9e79-725d-45f5-9992-e06318320f24

There are additional issues. If you use the color picker from the Properties View, notice that the scene view toolbar disappears. This is because the color dropper requires focus in order for escape key press to work. I did not investigate if that in particular was fixable. So that's also another reason for this lifecycle code.

I felt like this wasn't good UX (keeping it open) and hiding it is nicer for the user, but I can see how the code is also kind of ugly now. One possible option that sits in the middle is to just let the popup close without reopening it. It's not as nice, but at least it'll hide it which I think is nicer and it would also clean up some of the code. Would love your feedback, just let me know and I will fix it. :)

Edit: Another possible solution is to do the same WritableImage trick but grab the pixels from the popup window and render them on top of the main window, but that seemed a bit over the top, but it's also an option, but I don't know if it'll fix the lifecycle issues, might stay the same if we try to relaunch.

> Also, for some reason, the cursor in the toolbar does not change to ☝️ when hovering over the color dropper icon, unlike in the properties view. Sorry for the terrible video from the phone, but somehow, it doesn't change at all when I do screen recording...

I'll investigate

JosephFerano:
@vlaaad

I found a workaround, but it requires a small hack where we pretend that the color dropper overlay is within the `:workbench` context, this way the toolbar doesn't disappear. A lot of the funky lifetime stuff goes away, the toolbar remains in place and we can more easily hide and show the popup again. Also found the CSS fix to make the hand cursor appear on hover.

It makes me question though, is there a reason to hide the toolbar in the first place? Is that intentional? I've noticed that clicking around the editor, the only place now where the toolbar hides is when clicking in the Assets Browser. Was it perhaps an optimization in the past that is now no longer needed now that most of the code moved to cljfx?

editor/src/clj/editor/ui/settings_popup.clj
```diff
@@ -35,7 +35,7 @@
            [javafx.css Styleable]
            [javafx.event ActionEvent Event]
            [javafx.geometry HPos Point2D VPos]
-           [javafx.scene Cursor Node Parent]
+           [javafx.scene Cursor Node Parent Scene]
```
vlaaad: Unused import

vlaaad (dismissed):
LGTM!

Regarding toolbar hiding when focusing on the Assets pane. If I'm not mistaken, it worked like this before me and @matgis joined the project. It seems that when the editor was originally designed, it was envisioned that the editor UI/UX would have 2 distinct contexts, "global", available anywhere, and "workbench", which I think is about editing a particular resource. TBH, I find this distinction a bit weird, and particularly, toolbar behavior too. I'd expect such a toolbar to be just a ui element of a scene view, always visible in the scene view, but it seems it's somehow something else... We also want to make the editor layout freely configurable, which will probably require us to get rid of this global/workbench distinction... I dunno... stuff's complicated, maybe we can dig something up in the git history, but I don't have the needed context, unfortunately.

JosephFerano:
Ok, understood, I added a TODO comment explaining what you just said. Might be useful in the future.

vlaaad (approved):
Shipit!

PR #12499, opened 2026-06-01
Improve error handling around using broken resources in editor scripts

### LINKED ISSUE #11183: No errors when invalid file has been created in Editor Scripts
When I create an invalid go file, for example:
```lua
 editor.create_resources({"/my.go", "\"name\":\"my_go\""})
```
the editor doesn’t report an error.

If I then try to add a component to this invalid go using `editor.transact(txs)`, no error is shown, but nothing is added either.

It’s only when I attempt to open the go file in the editor that I see it’s invalid.
Ideally, the editor should raise an error immediately when the resource is created incorrectly, or at the very least when an operation (like adding a component) is attempted on it.

vlaaad:
Editor scripts now report invalid resources created with `editor.create_resources()` instead of silently accepting them. For example:
```lua
editor.create_resources({{"/test/invalid.go", "\"name\":\"invalid\""}})
--> Created resources are invalid: /test/invalid.go
```

Trying to edit a broken resource also fails:
```lua
editor.tx.add("/test/invalid.go", "components", { type = "label" })
--> Cannot edit defective resource: /test/invalid.go
```

Fix #11183

### DIFF
--- a/editor/src/clj/editor/editor_extensions.clj
+++ b/editor/src/clj/editor/editor_extensions.clj
@@ -148,14 +148,17 @@
 
 ;; region script API
 
-(defn- make-ext-get-fn [project]
+(defn- make-ext-get-fn [project localization]
   (rt/lua-fn ext-get [{:keys [rt evaluation-context]} lua-node-id-or-path lua-property]
     (let [unresolved-editor-lookup (rt/->clj rt graph/unresolved-editor-lookup-coercer lua-node-id-or-path)
           property (rt/->clj rt coerce/string lua-property)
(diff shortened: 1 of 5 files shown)


JosephFerano (approved):
It reports errors for both now. It will still create the invalid resource, should it allow the creation?

vlaaad:
I thought about it, and I think the answer is yes, it should create an invalid resource; the semantics of `create_resources` is batched file IO. To some extent, we already allow editor scripts to work and produce states that are considered errors by the build system (e.g., duplicate IDs of gos in the same collection).

PR #12514, opened 2026-06-03
Add clear console escape sequences

### LINKED ISSUE #12431: `/console/stream` doesn't expose a safe way to distinguish console clear from an regular `\n`
@selimanac and I are trying to consume the Editor console via `/console/stream` for external code editors but we can't figure out how to clear the console between sessions, which happens in the Editor.

In the Editor code, `clear-console!` sets an internal `:clear true` flag and sends `\n`, but that state is not visible to `/console/stream` consumers.

This means an external client cannot reliably tell whether a received blank line means:
- a real empty log line from the Editor/game output, or
- a console clear between sessions

### Suggested fix:
- append an control marker `U+001[С/D/E/F]` to the `\n` message for clear events, or
- switch the stream to a framed format where control messages are distinguishable from ordinary log lines

### About U+001[C/D/E/F]
These are four C0 control characters from ASCII/Unicode, originally intended for structuring plain-text data streams before formats like CSV, JSON, XML, and databases became common.

```
FS separates files
GS separates groups within a file
RS separates records within a group
US separates units/fields within a record
```

The best candidate is Record Separator (U+00E) I think.

vlaaad:
Now we emit a clear marker when clearing the console stream.

The marker is a leading newline followed by the same xterm-style escape sequence emitted by `clear`:

- `\n` preserves the previous behavior for simple line-oriented clients, which still observe an empty line on clear.
- `ESC [ 3 J` clears scrollback.
- `ESC [ H` moves the cursor to the home position.
- `ESC [ 2 J` clears the screen.

This allows `/console/stream` users to get a proper clear notification instead of just a newline. For example, this automatically works in shell console streaming:
```sh
curl -N http://localhost:$(cat .internal/editor.port)/console/stream
```

Fix #12431

### DIFF
--- a/editor/src/clj/editor/console.clj
+++ b/editor/src/clj/editor/console.clj
@@ -98,8 +98,8 @@
   "Clear the console. Callable from a background thread."
   []
   (let [subscribers (:stream-subscribers (swap! pending-atom assoc :clear true :entries [] :index 0))]
-    ;; ESC [ 2 J clears the screen, ESC [ H moves the cursor to the home position.
-    (run! #(% "\u001b[2J\u001b[H\n") subscribers)))
+    ;; ESC [ 3 J clears scrollback, ESC [ H moves the cursor to the home position, ESC [ 2 J clears the screen.
+    (run! #(% "\n\u001b[3J\u001b[H\u001b[2J") subscribers)))
 
 (def ^:private remote-log-pump-thread (atom nil))
 (def ^:private console-stream (atom nil))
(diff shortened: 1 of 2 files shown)


JosephFerano (dismissed):
It works, but some things I noted while testing;

In the VSCode terminal, clearing from the defold side doesn't clear any history at all, it just moves the output up to give the appearance of having cleared, which I guess is fine, you can still clear on the VSCode side.

fish is more aggressive about clearing screen where it also clears previous history. I was a bit surprised at first and started investigating this, only to realize it  bash and zsh work fine, it was fish behaving this way. On the latter two it clears fine.

I tried to get VSCode to attach to the running game editor so I could inspect the logs from the Output tab instead of the terminal but it failed on Linux . This led me to a little bit of yak shaving so I might continue investigating later. I might try testing this on either Windows to see how it behaves.

vlaaad:
Good point! I changed the sequence, so now `curl -N ...` also clears the scrollback, so now clearing works like in the editor's console.

PR #12517, opened 2026-06-03
 Fix ASTC textures using non-block-aligned dimensions

AGulev:
Fix ASTC-compressed textures whose width or height is not evenly divisible by the ASTC block size. These textures now keep their expected size in the engine and editor, avoiding issues where sprites, atlases, or texture dimensions could appear slightly larger than intended. WebGPU support remains compatible with its block-alignment requirements.

Technical notes:

ASTC compression now keeps logical mip dimensions in `TextureImage` metadata while computing compressed byte sizes with ceil-divided block counts. The encoder no longer pads source pixels before calling `astcenc`, so edge blocks are encoded from the real image dimensions. WebGPU converts logical dimensions to block-rounded storage/copy extents inside the backend for now.
In the future we can move these calculations into Bob and add an explicit `storage_size` field to the DDF.


com.dynamo.cr/com.dynamo.cr.bob/src/com/defold/extension/pipeline/texture/TextureCompressorASTC.java
```diff
@@ -105,27 +105,42 @@ public boolean supportsTextureCompressorPreset(TextureCompressorPreset preset) {
...
-
-        if (blockSizes == null) {
+        if (!supportsTextureFormat(format)) {
             System.err.println("Format " + format + " is not an ASTC format.");
             return 0;
         }
 
-        int blockSizeX = blockSizes[0];
-        return ((width + blockSizeX - 1) / blockSizeX) * blockSizeX;
+        return width;
```
AGulev, line 113: now it is basically the same as default implementation. Maybe we need a cleanup here and get rid from these functions, or keep it and use if/when we decide adding `storage_size` into DDF

com.dynamo.cr/com.dynamo.cr.bob/src/com/defold/extension/pipeline/texture/TextureCompressorASTC.java
```diff
@@ -105,27 +105,42 @@ public boolean supportsTextureCompressorPreset(TextureCompressorPreset preset) {
...
         int blockSizeY = blockSizes[1];
-        return ((height + blockSizeY - 1) / blockSizeY) * blockSizeY;
+
+        // ASTC byte length follows the Khronos formula, generalized by block size:
+        // floor((width + block_width - 1) / block_width) *
+        // floor((height + block_height - 1) / block_height) * 16.
+        // For ASTC 6x6 this matches:
+        // floor((width + 5) / 6) * floor((height + 5) / 6) * 16.
+        // https://registry.khronos.org/webgl/extensions/WEBGL_compressed_texture_astc/
+        // https://registry.khronos.org/OpenGL/extensions/KHR/KHR_texture_compression_astc_hdr.txt
```
AGulev, line 142: I added this comment in a few places where it's important to know the context of this

engine/graphics/src/webgpu/graphics_webgpu.cpp
```diff
@@ -337,59 +337,20 @@ static WGPUTextureFormat WebGPUFormatFromTextureFormat(TextureFormat format)
...
-    case TEXTURE_FORMAT_RGBA_ASTC_10X8:     return 10;
-    case TEXTURE_FORMAT_RGBA_ASTC_10X10:    return 10;
-    case TEXTURE_FORMAT_RGBA_ASTC_12X10:    return 12;
-    case TEXTURE_FORMAT_RGBA_ASTC_12X12:    return 12;
-    case TEXTURE_FORMAT_RGB_BC1:            return 4;
-    case TEXTURE_FORMAT_RGBA_BC3:           return 4;
-    case TEXTURE_FORMAT_RGBA_BC7:           return 4;
-    default:                                return 0;
-    };
-}
```
AGulev, line 366: this isn't unique for the webgpu backend, moved into graphics.cpp

engine/graphics/src/webgpu/graphics_webgpu.cpp
```diff
@@ -490,19 +457,49 @@ static void WebGPUSetTextureInternal(WebGPUTexture* texture, const TextureParams
...
+    uint32_t new_storage_height;
+    const uint32_t new_texture_width  = params.m_MipMap == 0 ? params.m_Width  : texture->m_Base.m_Width;
+    const uint32_t new_texture_height = params.m_MipMap == 0 ? params.m_Height : texture->m_Base.m_Height;
+
+    WebGPUGetTextureStorageDimensions(texture->m_Base.m_Format, texture->m_Base.m_Width, texture->m_Base.m_Height, &current_storage_width, &current_storage_height);
+    WebGPUGetTextureStorageDimensions(params.m_Format, new_texture_width, new_texture_height, &new_storage_width, &new_storage_height);
+
+    if (new_storage_width > max_texture_dimension_2d || new_storage_height > max_texture_dimension_2d)
+    {
+        dmLogError("Unable to upload texture data, texture size %ux%u (%ux%u block-aligned) exceeds maximum supported texture size (%ux%u).",
```
AGulev, line 490: opengl backend use 1px alpha texture in such cases, but I se we don't do that in webgpu, so I keep it as is

com.dynamo.cr/com.dynamo.cr.bob/src/com/defold/extension/pipeline/texture/TextureCompressorASTC.java
Jhonnyg: These functions were added specifically for the ASTC compressor, so if we don't need them any longer we can remove them (and add at a later time).

Jhonnyg (approved):

PR #12523, opened 2026-06-04
Disable Collision Object "Group" property when it's a tilemap or tilegrid

### LINKED ISSUE #12371: Disable Group when using collisions from tilemap
Currently, when using a Tilemap with Collision Groups defined in its tilesource, as a Collision Shape in the Collisionobject component the "Group" property is not taken into account. It could have a correct string, incorrect string or no string at all and it doesn't change the behaviour of the collisiobobject until the tilemap is assigned as the shape.

Image

This is a correct behaviour and is noted in the manual: 
https://defold.com/manuals/physics-shapes/#tilemap-collision-shape

"Note that the Group property is not used here since the collision groups are defined in the tile map’s tile source."

But is a bad UX.

Disable the input field for the "Group" property when Tilemap is used as Collision Shape enable back, when other shape is used.

Remove this "Group" property at all, when Tilemap is used as Collision Shape, but show it, when other shape is used.

JosephFerano:
The editor will now disable the `Group` property of a collision object component when the `Collision Shape` property is set to a tilemap resource. The collision object groups are instead controlled by the groups defined within the tile source used by the assigned tilemap.

Fixes #12371

### DIFF
--- a/editor/resources/localization/en.editor_localization
+++ b/editor/resources/localization/en.editor_localization
@@ -1213,6 +1213,7 @@ property.collision-object.event-contact.tooltip = If disabled, filters out any c
 property.collision-object.event-trigger = Generate Trigger Events
 property.collision-object.event-trigger.tooltip = If disabled, filters out any trigger events involving this collision object
 property.collision-object.group = Group
+property.collision-object.group.tilemap.tooltip = This property is not used when Collision Shape is a tilemap. Collision groups are defined in the tilemap's tile source.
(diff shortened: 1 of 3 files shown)


paweljarosz:
Works as expected, thanks!

Zrzut ekranu z 2026-06-04 14-03-27

Zrzut ekranu z 2026-06-04 14-03-19

matgis (approved):

PR #12527, opened 2026-06-04
Attempt to create a Vulkan adapter when checking support

### LINKED ISSUE #12510: Improve graphics adapter selection for Vulkan
Currently, we detect support by creating a dummy VkInstance during bootstrapping. This only checks that the API is supported, but not that the device can actually create a logical/physical device needed for Vulkan backed rendering. Instead, we should do the entire chain of creating a vkinstance, physical device and a logical device so that we know for sure we can use Vulkan on the given device.

To fix it, we likely don't want to create all of these support structures multiple times, so some changes in the startup flow is needed:

When checking support, we create a new VulkanContext and attempt to go through all the hoopla of device creation. If everything is OK, we can store the pointer in the "global" context pointer (i.e g_VulkanContext) and reuse/return it later in the NewContext function. This piggybacks on the current setup where all graphics adapters use a global context, which is not want we want eventually. But for now maybe it's fine? We can always change the bootstrapping flow later, since it's all internal functions anyway!

Jhonnyg:
The way we check for Vulkan support has been refactored so that we now check that the device actually has a vulkan compatible GPU. Previously we would signal support if the API was available and that we could create a context, but we never checked if we could actually create a Vulkan ready device.

Fixes #12510


engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -1552,6 +1642,21 @@ namespace dmGraphics
...
             LoadVulkanFunctions(inst);
         #endif
+            PhysicalDevice physical_device;
+            LogicalDevice logical_device;
+            QueueFamily queue_family;
+            memset(&physical_device, 0, sizeof(physical_device));
+            memset(&logical_device, 0, sizeof(logical_device));
+
+            res = CreateVulkanLogicalDevice(inst, VK_NULL_HANDLE, false, false, false,
+                &physical_device, &logical_device, &queue_family, 0x0, 0x0);
```
Jhonnyg, line 1629: This is mostly the new is-supported-check: the CreateVulkanLogicalDevice will only return VK_SUCCESS if there are any actual devices available that we can use.

engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -1201,245 +1201,307 @@ namespace dmGraphics
         }
     }
 
-    bool InitializeVulkan(HContext _context)
+    static void AddRequiredDeviceExtensions(dmArray<const char*>& device_extensions)
     {
-        VulkanContext* context = (VulkanContext*) _context;
-        VkResult res = CreateWindowSurface(context->m_BaseContext.m_Window, context->m_Instance, &context->m_WindowSurface, dmPlatform::GetWindowStateParam(context->m_BaseContext.m_Window, WINDOW_STATE_HIGH_DPI));
-        if (res != VK_SUCCESS)
+        device_extensions.SetCapacity(2);
```
JCash, line 1206: Perhaps OffsetCapacity() in case the order of the AddRequiredDeviceExtensions() call is changed.

JCash (approved):

PR #12529, opened 2026-06-04
Add missing editable semantic types

Jhonnyg:
Minor bugfix for the editor when using custom material with vertex attributes that has explicitly specified semantic types.


vlaaad (approved):
Shipit!

PR #12530, opened 2026-06-04
Add `image` module to editor extensions

### LINKED ISSUE #10994: Ability to read image pixel data from editor scripts
Many types of editor script features would be unlocked by reading pixels from images, but that doesn't appear to currently be possible: https://forum.defold.com/t/is-it-impossible-to-read-images-from-an-editor-script/81011

Examples include building click colliders that care about the shape of an object, not just its outer rect, and building convex hulls for tilemaps to work around some of the built-in limitations (https://forum.defold.com/t/how-can-i-modify-tilesource-collision-shapes-via-script/80965/4).

I would like to be able to read the pixels of an image from within an editor script, just like you can at runtime via `image.load_buffer(bytes)` or similar.

I originally tried using the `image` API as well as https://github.com/britzl/defold-png, but neither runtime APIs nor native extensions seem to be available in editor scripts.

As far as I can tell, the only viable alternative is to bundle executables for various platforms to read pngs and then to `.execute()` the platform-specific external tool. I haven't found a better route so far, so this is probably what I'll have to do for now.

vlaaad:
Editor scripts can now inspect image files. The new `image` module can load image files, read their size, and iterate over all pixels.

Example:

```lua
local img = image.load_file("assets/source.png")
local w, h = image.size(img)
local r, g, b, a = image.pixel(img, 1, 1)

for x, y, r, g, b, a in image.pixels(img) do
    print(x, y, r, g, b, a)
end
```

Fix #10994


vlaaad:
@JosephFerano I decided to simplify the `image.size` API — removed string overload that had a performant path for loading image size. I did it for 2 reasons:
1. We can add it later if it gets requested, so far it hasn't
2. We might want to add `image.load_resource` in the future, which will use graph for image data, looking up size will be even faster since it will be tracked/persisted in the graph... but it won't work with the `image.size` fn since it treats strings as file paths — absolute or relative, while resource paths look like absolute paths on Linux and macOS.

So, if there will be demand for faster image size lookup, I think it will be better to implement using `image.file_size`, `image.resource_size`, etc.

Just some things that popped into my mind after sleeping on this PR...

JosephFerano (approved):
I saw what you mean in the diff, makes sense to make it a separate call. Works well.

PR #12535, opened 2026-06-05
Add prebuilt datac support for gamesys test data

JCash:
- Add `.prebuilt_<suffix>` handling to the gamesys test-data builder so prebuilt fixtures are copied into the build output as `.<suffix>` files.
- Keep CMake aware of `.prebuilt_` sources so the content build reruns when those fixtures change.
- Add a `PrebuiltData` test and a `data/prebuilt.prebuilt_datac` fixture that loads as `/data/prebuilt.datac`.

### DIFF
--- a/engine/gamesys/src/gamesys/test/build_test_data.py
+++ b/engine/gamesys/src/gamesys/test/build_test_data.py
@@ -8,8 +8,27 @@ import tempfile
 from pathlib import Path
 
 
+PREBUILT_MARKER = ".prebuilt_"
+
+
+def prebuilt_output_path(path):
+    marker_index = path.name.rfind(PREBUILT_MARKER)
+    if marker_index <= 0:
+        return None
+
+    suffix = path.name[marker_index + len(PREBUILT_MARKER):]
+    if not suffix:
+        return None
+
+    return path.with_name(path.name[:marker_index] + "." + suffix)
+
+
+def is_prebuilt_source_file(path):
+    return prebuilt_output_path(path) is not None
+
+
 def ignored_source_file(path):
-    return path.name == ".DS_Store" or path.name.startswith(".prebuilt_")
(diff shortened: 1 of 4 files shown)


engine/gamesys/src/gamesys/test/data/prebuilt.prebuilt_datac
```diff
@@ -0,0 +1,3 @@
+
+tag-one
+tag-twohello
```
JCash, line 3: Some dummy test data

engine/gamesys/src/gamesys/test/test_gamesys.cpp
```diff
@@ -464,6 +464,24 @@ TEST_F(ResourceTest, DataResourceContents)
     dmResource::Release(m_Factory, (void*)resource);
 }
 
+TEST_F(ResourceTest, PrebuiltData)
```
JCash, line 467: It is up to each test to handle its own data, no assumptions made, as it's copied as-is.

engine/gamesys/src/gamesys/test/build_test_data.py
```diff
@@ -201,29 +235,37 @@ def main():
...
+        copy_optional_output_tree(bob_output_abs, output_root, "lua")
+        copy_immediate_outputs(bob_output_abs, output_root)
+        copy_raw_outputs(stage_root, output_root, build_inputs)
+    else:
+        print("  No Bob inputs; copying prebuilt outputs only", flush=True)
+        if dst.exists():
+            shutil.rmtree(dst)
+        dst.mkdir(parents=True, exist_ok=True)
+
+    copy_prebuilt_outputs(root / folder, dst)
```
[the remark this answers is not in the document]
JCash: I don't want to patch Waf workflow unless really necessary. I want os to more forward with CMake support.

Jhonnyg (approved):

Issue #12516, opened 2026-06-03
Crash in device_avaudio.mm, iOS


ihar-va-MA:
**Description**

Fatal crash in Defold's iOS audio system (`device_avaudio.mm`). The sound thread calls `[AVAudioPlayerNode play]` via `DeviceAVAudioReconfigureIfNeeded`, which triggers an NSException from Apple's AVAudio framework: `required condition is false: _engine->IsRunning()` — meaning the `AVAudioEngine` is not running at the time of the call.

In our case, the crash consistently follows a third-party video ad SDK (`UADSViewController`) becoming active. Whether that is the root cause or merely a trigger for a pre-existing engine state issue is unclear without further investigation.

**To Reproduce**

1. Build an iOS app with Defold 1.12.4 that uses the sound system (playing background music or sound effects).
2. Integrate a third-party video ad SDK (e.g. Unity Ads) that takes control of `AVAudioSession` when it plays a video ad.
3. Trigger a video ad while the Defold sound system is active.
4. Observe a fatal crash in the sound thread with `_engine->IsRunning()`.

**Expected behavior**

Defold's sound system should gracefully handle external `AVAudioSession` interruptions (such as those caused by video ad SDKs) without crashing. The sound thread should either skip reconfiguration while the engine is stopped, or properly wait for the interruption to end before attempting to restart audio nodes.

**Defold version**

- Version 1.12.4 (commit `402218d5544666871f07ecdfa21032a7fb59413f`)

**Platforms**

- Platforms: iOS
- OS: iOS 17.6.1 (confirmed), likely affects other iOS 17.x versions
- Device: iPhone 13 mini (confirmed); likely not device-specific

**Logs**

Crash stacktrace (NSException, unhandled, fatal):

```
Exception: com.apple.coreaudio.avfaudio: required condition is false: _engine->IsRunning()

thread_start
_pthread_start
dmThread::ThreadStartProxy          (thread_posix.cpp:40)
dmSound::SoundThread                (sound.cpp:2100)
dmSound::UpdateInternal             (sound.cpp:2024)
dmDeviceAVAudio::DeviceAVAudioFreeBufferSlots      (device_avaudio.mm:316)
dmDeviceAVAudio::DeviceAVAudioReconfigureIfNeeded  (device_avaudio.mm:128)
-[AVAudioPlayerNode play]
AVAudioNodeImplBase::Start
AVAudioPlayerNodeImpl::StartImpl
AVAE_RaiseException
[...]

PR #12545, opened 2026-06-08
Fix rare AVAudio related crash on iOS

AGulev:
Fix a crash that might occur while switching between applications or showing ads.

https://github.com/defold/defold/issues/12516


engine/sound/src/test/CMakeLists.txt
```diff
@@ -166,7 +166,11 @@ endif()
 set(_sound_export_file "${SOUND_TEST_GEN_DIR}/test_sound_exported_symbols.cpp")
 set(_sound_exports ${SOUND_TEST_EXPORTS} LoopBackDevice)
 defold_create_exported_symbols_file("${_sound_export_file}" "${_sound_exports}")
-defold_add_executable(test_sound "${SOUND_TEST_ROOT}/test_sound.cpp" "${_sound_export_file}")
+set(_sound_test_sources "${SOUND_TEST_ROOT}/test_sound.cpp")
+if(TARGET_PLATFORM MATCHES "^(x86_64-macos|arm64-macos)$")
```
JCash: Why only macOS?
We do want to test on iOS eventually as well

engine/sound/src/test/test_sound.cpp
```diff
@@ -115,6 +115,10 @@ extern uint32_t MUSIC_ADPCM_WAV_SIZE;
 extern unsigned char AMBIENCE_ADPCM_WAV[];
 extern uint32_t AMBIENCE_ADPCM_WAV_SIZE;
 
+#if defined(DM_PLATFORM_MACOS)
```
JCash: Same here, better to designe for APPLE than only macOS.

JCash (changes requested):
overall good, would like to see some iOS preparation for future running tests on device

engine/sound/src/test/CMakeLists.txt
AGulev: fixed

engine/sound/src/test/test_sound.cpp
AGulev: fixes

JCash (approved):

PR #12551, opened 2026-06-08
Use borderless windowed fullscreen mode by default for desktop platforms

Jhonnyg:
When enabling the `display->fullscreen` setting in the game.project, we now create a borderless fullscreen window instead of the old "explicit" fullscreen mode.

Fixes #4752

#### Technical changes

* Borderless fullscreen is now the de-factor mode when using fullscreen in a project
* There were a few places where we #ifdef __PLATFORM__ in the platform-glfw windowing module that is now moved into the platform specific implementation(s).


engine/platform/src/platform_window_glfw3.cpp
```diff
@@ -224,28 +256,27 @@ namespace dmPlatform
 
             if (params.m_OpenGLUseCoreProfileHint)
             {
-                bool can_set_profile_and_fc = true;
-
-            #ifdef _WIN32
-                // Not supported on windows when requesting the default OpenGL version (which will equate to the highest available)
-                can_set_profile_and_fc = !use_highest_version;
-            #endif
```
Jhonnyg, line 232: I've tried to move as much as possible of these platform-specific codepaths into their respective _win32/linux/osx.cpp implementations.

JCash (dismissed):

PR #12560, opened 2026-06-09
Improve `lift-target-arcs` performance

vlaaad:
This change improves build performance of big projects by 3% (2m26s -> 2m20s on a measured run).

What's interesting is that this work started as a way to reduce allocations (too much mapv uses), but the thing that moved the needle was extracting and reusing a `graph` local. So, the change technically changes the behavior by adding an assumption that wasn't in the function before (the assumption — all overrides are in the same graph), but we already had the same assumption in practice, e.g., in `g/override-predecessors`.

Related to #11988

### DIFF
--- a/editor/src/clj/internal/graph.clj
+++ b/editor/src/clj/internal/graph.clj
@@ -799,18 +799,20 @@
                    (into result propagated-arcs))))))))
 
 (defn- lift-target-arcs [basis target-id target-override-chain arcs]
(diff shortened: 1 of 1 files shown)


matgis (approved):
Looks good to me!

I wonder if there is a potential further improvement to be seen if we can early-out when there are no overrides? 🤔

vlaaad:
Perhaps!

PR #12564, opened 2026-06-10
Improve warm build performance of large projects by 19%

vlaaad:
On a large project the measured warm build time (when there is a build cache on disc) improved from 3m16s to 2m37s.

Technical notes:
I did 2 measurement runs, numbers above are averages of these runs:
```
;; before
"Elapsed time: 205720.703959 msecs"
"Elapsed time: 186601.907958 msecs"
;; after
"Elapsed time: 153019.124916 msecs"
"Elapsed time: 161656.230708 msecs"
```
We no longer use sorted maps for content hashing. They are very slow to build. Here are microbenchmark results for building a sorted map (old approach) vs sorting an array of keys (new approach):

| map size | old | new | improvement |
|---:|---:|---:|---:|
| 8 | 2.04 µs | 1.08 µs | 47.07% |
| 32 | 8.16 µs | 6.16 µs | 24.50% |
| 128 | 46.73 µs | 31.67 µs | 32.21% |
| 512 | 249.86 µs | 169.14 µs | 32.31% |

Time is microseconds per call; lower is better.

Fix #11988

### DIFF
--- a/editor/src/clj/util/digestable.clj
+++ b/editor/src/clj/util/digestable.clj
@@ -129,35 +129,58 @@
   (let [tag-sym (symbol (.getSimpleName (class resource)))]
     (digest-tagged! tag-sym (resource/resource-hash resource) writer opts)))
 
-(defn- to-sorted-map [map]
-  {:pre [(map? map)]}
-  (if (or (sorted? map)
-          (< (count map) 2))
-    map
-    (into (sorted-map)
-          map)))
-
 (defn- digest-map! [coll writer opts]
-  (let [sorted-map (to-sorted-map coll)
-        last-index (dec (count sorted-map))]
-    (digest-raw! "{" writer)
-    (reduce-kv (fn digest-map-entry! [^long index key value]
-                 (if (ignored-key? key)
-                   index
-                   (try
-                     (digest! key writer opts)
-                     (digest-raw! " " writer)
-                     (if (node-id-entry? key value)
-                       (digest-tagged! 'Node (persistent-node-id-value value opts) writer opts)
-                       (digest! value writer opts))
-                     (when (< index last-index)
-                       (digest-raw! ", " writer))
-                     (inc index)
-                     (catch Throwable error
-                       (throw (augment-throwable-with-path error coll key))))))
-               0
-               sorted-map)
-    (digest-raw! "}" writer)))
+  {:pre [(map? coll)]}
+  (digest-raw! "{" writer)
+  (let [n (count coll)
+        last-index (dec n)]
+    (if (or (sorted? coll)
+            (< n 2))
+      (reduce-kv
+        (fn digest-map-entry! [^long index key value]
+          (if (ignored-key? key)
+            index
+            (try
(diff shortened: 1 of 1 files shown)


editor/src/clj/util/digestable.clj
```diff
@@ -129,35 +129,58 @@
...
+              (if (node-id-entry? key value)
+                (digest-tagged! 'Node (persistent-node-id-value value opts) writer opts)
+                (digest! value writer opts))
+              (when (< index last-index)
+                (digest-raw! ", " writer))
+              (inc index)
+              (catch Throwable error
+                (throw (augment-throwable-with-path error coll key))))))
+        0
+        coll)
```
vlaaad, line 155: Still using old code for already sorted maps (or maps that don't require sorting).

editor/src/clj/util/digestable.clj
```diff
@@ -129,35 +129,58 @@
...
+               output-index 0]
+          (when (< array-index n)
+            (let [key (aget keys-array array-index)]
+              (if (ignored-key? key)
+                (recur (unchecked-inc-int array-index) output-index)
+                (do
+                  (try
+                    (digest! key writer opts)
+                    (digest-raw! " " writer)
+                    (let [value (get coll key)]
```
matgis, line 174: Since this is a hot-spot, we might also want to get rid of this lookup? Do we get even better performance if we put the entire `MapEntry` in the `object-array` and sort it by key?

editor/src/clj/util/digestable.clj
```diff
@@ -129,35 +129,58 @@
...
+                      (if (node-id-entry? key value)
+                        (digest-tagged! 'Node (persistent-node-id-value value opts) writer opts)
+                        (digest! value writer opts)))
+                    (when (< output-index last-index)
+                      (digest-raw! ", " writer))
```
matgis, line 179: I think this bug might have been present prior to this change, but it looks like we're not correctly excluding the last comma separator if there are ignored entries, since we're comparing `output-index` against a `last-index` derived from the entire length of the input `coll`. I think it makes sense to fix this issue as part of this PR.

matgis (approved):

editor/src/clj/util/digestable.clj
vlaaad: I tried it, but it was a bit slower because we had to create an entry. With the current code, we don't create any entries at all.

editor/src/clj/util/digestable.clj
vlaaad: Yes, this behavior is also present in the old version. You can't solve it by always incrementing an index during iteration. For example, if the ignored key is last, you would still add `", "` before it, e.g., `{:a 1, }`. You need filter + interpose for "proper" solution, but that's probably going to be slower.

matgis: Maybe we should just remove this logic, since we're in a hot-spot. It could just always digest a separator regardless if there are more elements.

PR #12587, opened 2026-06-14
iOS: use Apple's devicectl instead of ios-deploy for install & launch

TrulyBright:
## Problem

The editor's "Install on device" / "Launch" step for iOS shells out to `ios-deploy`. That project is effectively abandoned (latest release 1.12.2, 2022; last commit ~2 years ago) and its device discovery relies on the legacy `usbmuxd` `_apple-mobdev2._tcp` Bonjour service, which modern iOS no longer advertises. On current macOS/iOS it silently fails with `Failed to list devices`, even when the device is correctly paired with "Connect via network" enabled. `README_IOS.md` already notes ios-deploy is outdated and recommends the Xcode command-line tools.

## Change

Replaces the `ios-deploy` backend with Apple's first-party `devicectl` (CoreDevice), which is actively maintained and works over both USB and WiFi:

- `list devices --json-output -` for detection (picks the first `paired` device)
- `device install app --device <id> <app> --json-output -`, parsing the returned `bundleID`
- `device process launch --device <id> <bundleID>` (devicectl launches by bundle id, not path)

`devicectl` is resolved via `xcrun --find devicectl` with a fallback to the default `Xcode.app` path (handles `xcode-select` pointing at the bare Command Line Tools). The existing `tools.ios-deploy-path` preference is still honored as an explicit override.

No `ios-deploy` fallback is kept — it's non-functional on current OS versions, so this is a clean switch.

## Testing

Verified the underlying `devicectl` commands and their JSON output shapes directly against a physical iPhone 13 (iOS 26-era) on macOS 26 — list/install/launch all succeed over WiFi where `ios-deploy` returns "Failed to list devices". Lua syntax validated with `luac -p`.

## Open question for maintainers

The `tools.ios-deploy-path` preference key/label (and its localized strings in 5 languages) still says "ios-deploy"; happy to rename it to something tool-neutral if you'd prefer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/editor/resources/bundle.editor_script
+++ b/editor/resources/bundle.editor_script
@@ -185,7 +185,10 @@ local function get_ios_device(devicectl)
     local devices = json.decode(output).result.devices
     for i = 1, #devices do
         local d = devices[i]
-        if d.connectionProperties and d.connectionProperties.pairingState == "paired" then
+        -- `list devices` returns every paired CoreDevice target (Apple Watch, Apple TV,
+        -- Mac, other iPads/iPhones), so restrict the pick to deployable iOS devices.
+        if d.connectionProperties and d.connectionProperties.pairingState == "paired"
+            and d.hardwareProperties and d.hardwareProperties.platform == "iOS" then
             return {id = d.identifier, label = (d.deviceProperties and d.deviceProperties.name) or d.identifier}
         end
     end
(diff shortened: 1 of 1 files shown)


vlaaad (changes requested):
1. Please address codex comments
2. Fix prefs: `tools.ios-deploy-path` is now an incorrect preference. If users had it configured, we would use ios-deploy instead of devicectl. Rename preference, make sure the preference dialog doesn't mention ios-deploy. 
3. We should have no mentions of ios-deploy in the repo anymore — currently it has many more usages in other files, like some internal docs, localization, and tests.

PR #12595, opened 2026-06-15
Clean up Bob temporary files

### LINKED ISSUE #12591: Repeated Windows bundling from the Editor consumes several GB of drive space per build
## Description

When repeatedly creating Windows bundles from the Defold Editor, a large amount of free space on my system drive (`C:`) is consumed after each bundling operation.

In my current project, every Windows bundle operation appears to consume approximately **4 GB** of disk space on the `C:` drive. If I create bundles many times during the day, which is a normal workflow when testing a large project, the accumulated temporary disk usage can easily reach **tens of gigabytes**. I have seen cases where around **70 GB** of free space disappeared during one work session.

The important part is that this disk space is released after I close the Defold Editor. This suggests that the temporary data is either not cleaned up after each bundling operation, or some files/handles/caches remain alive for as long as the Editor process is running.

This is a serious workflow issue for large projects because I often need to bundle the game many times per day for testing. As the free space on the system drive runs out, other applications running in parallel start to suffer as well. The only practical workaround is to regularly close the Defold Editor, wait for the disk space to be released, then start the Editor again and reopen the project.

## Expected behaviour

Temporary files or intermediate bundle data created during the bundling process should be cleaned up after each bundle operation is finished.

Repeated bundling from the Editor should not continuously consume additional disk space until the Editor is closed.

## Actual behaviour

Each Windows bundle operation consumes approximately 4 GB of free space on the system drive.
The consumed disk space is not released after the bundle operation completes.
The space is released only after closing the Defold Editor.

## Steps to reproduce

I do not currently have a minimal reproduction project because the issue happens on a large production project, and preparing a repro case would require sharing a project/archive of several gigabytes.

The general workflow is:

1. Open a large Defold project on Windows.
2. Create a Windows bundle from the Editor.
3. Check free space on the system drive (`C:`).
4. Repeat the bundling operation multiple times.
5. Observe that free disk space decreases after each bundling operation.
6. Close the Defold Editor.
7. Observe that the previously consumed disk space is released.

## Environment

* Defold version: [1.12.4 - 1.13.0]
* Platform: Windows 10 - 11
* Bundle target: Windows
* Project size: large production project, several GB of assets

## Workaround

The only current workaround is to close the Defold Editor regularly in order to release the disk space, then start the Editor again and reopen the project.
This is very inconvenient when frequently testing Windows bundles during development.

## Additional context

This may be related to issue #8483 (“Bob uses deleteOnExit() but it doesn't delete non-empty folders”), but it does not look exactly the same from the user side.

In this case, the disk space is released when the Editor is closed, which suggests that the temporary data may still be referenced or kept alive by the Editor process after the bundle operation has finished.

Even without a minimal repro project, I hope this can be investigated by checking what temporary files, directories, file handles, or in-memory/external caches are created during repeated Windows bundling from the Editor.

AGulev:
Bob now removes temporary files created during builds, archiving, bundling, and shader compilation when the build finishes. This prevents stale temporary files from accumulating across repeated Bob runs while still allowing temporary files to be kept explicitly for debugging.

Fix https://github.com/defold/defold/issues/12591

### Technical changes
Better to review with Hide whitespace option because of big try blocks
CleanShot 2026-06-15 at 12 46 00@2x

- Added `BobTempScope`, a per-project temporary directory scope with `DM_BOB_KEEP_TEMP` / `defold.bob.keepTemp` opt-out support.
- Made `Project` `AutoCloseable`, added `createTempFile()` / `createTempDirectory()`, and dispose the Bob CLI project in a `finally` block (I know, you Björn, not a huge fan of these new features, but in this case it helps to avoid all these `finally()` blocks all around especially taking into account that many tests bypass `invoke()`)
- Routed temporary file creation through `Project` in archive publishing, resource archives, Android/iOS/macOS bundling, Lua bytecode, Ogg/Opus handling, and zip publishing.
- Updated shader compilation pipelines to track temporary files and destroy shader pipeline resources in `finally` paths.
- Removed `FileUtil.deleteOnExit()` usage and cleaned up tests to use explicit cleanup, `TemporaryFolder`, or try-with-resources project disposal.
- Updated archive extraction helpers now that extracted files are not registered with JVM-wide delete-on-exit cleanup.
- Added `BobTempScopeTest` coverage for scoped cleanup, closed-scope behavior, and `Project.dispose()` cleanup.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/test/util/FontTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/test/util/FontTest.java
@@ -57,10 +57,6 @@ public class FontTest {
     @Rule
     public TemporaryFolder temporaryFolder = new TemporaryFolder();
 
-    private File newTempFile(String prefix, String suffix) throws IOException {
-        return File.createTempFile(prefix, suffix, temporaryFolder.getRoot());
-    }
-
     private String copyResourceToDir(String tmpDir, String resName) throws IOException {
         String outputPath = Paths.get(tmpDir, resName).toString();
 
@@ -237,7 +233,7 @@ public class FontTest {
             .build();
 
         // temp output file
-        File outfile = newTempFile("glyph-bank-output", ".glyph_bankc");
+        File outfile = temporaryFolder.newFile("glyph-bank-output.glyph_bankc");
 
         // compile font
         Fontc fontc = new Fontc();
@@ -283,7 +279,7 @@ public class FontTest {
             .build();
 
         // temp output file
-        File outfile = newTempFile("glyph-bank-output", ".glyph_bankc");
+        File outfile = temporaryFolder.newFile("glyph-bank-output.glyph_bankc");
 
         // compile font
         Fontc fontc = new Fontc();
@@ -329,7 +325,7 @@ public class FontTest {
             .build();
 
         // temp output file
-        File outfile = newTempFile("glyph-bank-output", ".glyph_bankc");
+        File outfile = temporaryFolder.newFile("glyph-bank-output.glyph_bankc");
 
         // compile font
         Fontc fontc = new Fontc();
@@ -374,7 +370,7 @@ public class FontTest {
             .build();
 
         // temp output file
-        File outfile = newTempFile("glyph-bank-output", ".glyph_bankc");
+        File outfile = temporaryFolder.newFile("glyph-bank-output.glyph_bankc");
 
         // compile font
         Fontc fontc = new Fontc();
@@ -457,7 +453,7 @@ public class FontTest {
                 .setSize(24)
                 .build();
 
-        File outfile = newTempFile("font-output", ".fontc");
+        File outfile = temporaryFolder.newFile("font-output.fontc");
 
         // compile font
         boolean success = true;
@@ -504,7 +500,7 @@ public class FontTest {
             .build();
 
         // temp output file
-        File outfile = newTempFile("font-output", ".fontc");
+        File outfile = temporaryFolder.newFile("font-output.fontc");
 
         // compile font
         Fontc fontc = new Fontc();
@@ -546,7 +542,7 @@ public class FontTest {
             .build();
 
         // temp output file
-        File outfile = newTempFile("font-output", ".fontc");
+        File outfile = temporaryFolder.newFile("font-output.fontc");
 
         // compile font
         Fontc fontc = new Fontc();
--- a/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/test/util/FontTest.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/test/util/FontTest.java
@@ -30,14 +30,15 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.HashMap;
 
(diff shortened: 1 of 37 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/BobTempScope.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    private static boolean shouldKeepTemp() {
+        String propertyValue = System.getProperty(KEEP_TEMP_PROPERTY);
+        if (propertyValue != null) {
+            return Boolean.parseBoolean(propertyValue);
+        }
+
+        String envValue = System.getenv(KEEP_TEMP_ENV);
+        return envValue != null && !envValue.isEmpty() && !"0".equals(envValue) && !"false".equalsIgnoreCase(envValue);
```
britzl: What's the point of checking envValue.isEmpty()? It should be enough to check if it's not null and then "1" or "true"

`return envValue != null && (envValue.equals("1") || envValue.equalsIgnoreCase("true"));`

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/BobTempScope.java
```diff
@@ -0,0 +1,87 @@
...
+package com.dynamo.bob.util;
+
+import com.dynamo.bob.logging.Logger;
+import org.apache.commons.io.FileUtils;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+
+public class BobTempScope implements AutoCloseable {
```
britzl: It would help with a short explainer of the purpose of this class.

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/test/util/FontTest.java
```diff
@@ -231,8 +237,7 @@ public void testTTF() throws Exception {
             .build();
 
         // temp output file
-        File outfile = File.createTempFile("glyph-bank-output", ".glyph_bankc");
-        FileUtil.deleteOnExit(outfile);
+        File outfile = newTempFile("glyph-bank-output", ".glyph_bankc");
```
britzl: You should be able to do:

```
File outfile = temporaryFolder.newFile("foo.glyph_bankc");
```

https://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html#newFile(java.lang.String)

The file should be guaranteed to clean up after the test has finished

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/BobTempScope.java
britzl: Not sure I like the BobTempScope name. For all intents and purposes this class wraps a temporary directory and cleanup. It could just as well be called BobTempDirectory.java ?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/util/BobTempScope.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private final File rootDirectory;
+    private final boolean keepTemp;
+    private final Thread shutdownHook;
+    private boolean closed = false;
+
+    public BobTempScope() throws IOException {
+        this.rootDirectory = Files.createTempDirectory("defold-bob-invoke-").toFile();
+        this.keepTemp = shouldKeepTemp();
+        this.shutdownHook = new Thread(this::cleanupOnShutdown, "bob-temp-scope-cleanup");
```
AGulev: I also added shutdown hook, if user exit in the middle of build process

britzl (approved):

PR #12600, opened 2026-06-15
Support `arcd` archives over 4 GiB

### LINKED ISSUE #12592: Windows bundle crashes at startup when game resource archive exceeds 2 GB
## Description

I have a large Defold desktop project where the Windows bundled build crashes immediately at startup when the final game resource archive becomes larger than approximately **2 GB**.

The same project runs correctly from the Defold Editor. The Windows bundle is created successfully, but when I launch the bundled executable, it opens the game window and then immediately fails during startup with resource loading errors.

I reproduced this on at least:

* Defold **1.12.4**
* Defold **1.13.0**

The crash happens in both debug and release Windows bundles.

## Actual behaviour

The bundled executable fails during startup with errors like:

```text
WARNING:RESOURCE: Unable to create resource: /_generated_358d8fb7d80c5b08.goc: FORMAT_ERROR
ERROR:GAMEOBJECT: Could not instantiate game object from prototype /_generated_358d8fb7d80c5b08.goc.
...
WARNING:RESOURCE: Unable to create resource: /windows/__main/__main.collectionc: FORMAT_ERROR
```

The error looks like missing or invalid resources, but after many tests it appears to be connected to the final archive/resource size boundary around **2 GB**.

## Expected behaviour

A Windows bundle should start normally even when the game resource archive is larger than 2 GB, or the build step should report a clear error if this size is not supported.

## Important isolation result

After reducing the amount of bundled resources so that the final resource archive became smaller than 2 GB, the same project started working correctly again.
In my case I removed several large unused atlases. After that, the Windows bundle launched normally and worked without problems.
This strongly suggests that the crash is not caused by one specific corrupted PNG, collection, atlas, or game object, but by crossing a size limit in the bundled game resources.

## Notes

I cannot currently provide a small reproduction project, because reproducing this requires a very large project with several gigabytes of assets. The issue was found in a private commercial project.

I can provide private project access to Defold developers if needed, although the current project version has already been reduced below the problematic size to continue development.

This may be related to previous large archive / large resource / 2 GB boundary issues in Defold, but I can still reproduce a startup crash on Windows when the final bundled resource archive exceeds this size.

## Environment

* Platform: Windows 10 - 11
* Bundle target: Windows
* Defold versions tested: 1.12.4, 1.13.0
* Build types tested: debug and release bundles
* Project type: large desktop game project with many atlases and image resources

AGulev:
Windows bundles with large game resource archives can now load resources at startup when archive data crosses the 2 GiB boundary, instead of failing with resource `FORMAT_ERROR` messages. The archive format also gains room for offsets beyond 4 GiB.

Fix https://github.com/defold/defold/issues/12592

### Technical changes
- Bump archive indexes to version 6 and pack entry flags into the high 4 bits of a 64-bit offset word.
- Change Bob archive offsets to `long`, reject v5 indexes, and add v6 read/write validation.
- Update runtime archive loading, mmap/file-backed reads, Android large-file handling, and provider error propagation for 64-bit offsets.
- Move generated archive temp files into filesystem outputs to avoid streaming very large `.arci`/`.arcd` files through `setContent()`.
- Update `unpack_arc.py` and archive format docs for the v6 packed entry layout.
- Add Bob and engine resource archive tests for v6, v5 rejection, sparse large offsets, mounted archives, wrapped indexes, and replacement indexes.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ArchiveBuilder.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ArchiveBuilder.java
@@ -60,9 +60,15 @@ public class ArchiveBuilder {
 
     private static Logger logger = Logger.getLogger(ArchiveBuilder.class.getName());
 
-    public static final int VERSION = 5;
+    public static final int VERSION_5 = 5;
+    public static final int VERSION_6 = 6;
+    public static final int VERSION = VERSION_6;
     public static final int HASH_MAX_LENGTH = 64; // 512 bits
     public static final int MD5_HASH_DIGEST_BYTE_LENGTH = 16; // 128 bits
+    public static final long MAX_UNSIGNED_INT_OFFSET = 0xffffffffL;
+    public static final int V6_FLAGS_SHIFT = 60;
+    public static final int V6_MAX_FLAGS = 0xf;
+    public static final long V6_OFFSET_MASK = 0x0fffffffffffffffL;
 
     private List<ArchiveEntry> entries = new ArrayList<ArchiveEntry>();
     private List<ArchiveEntry> excludedEntries;
@@ -218,7 +224,7 @@ public class ArchiveBuilder {
                 // do not write to it at the same time
                 synchronized (archiveData) {
                     alignBuffer(archiveData, this.resourcePadding);
-                    entry.setResourceOffset((int) archiveData.getFilePointer());
+                    entry.setResourceOffset(archiveData.getFilePointer());
                     archiveData.write(buffer, 0, buffer.length);
                 }
             }
@@ -236,8 +242,10 @@ public class ArchiveBuilder {
     private void writeArchiveIndex(RandomAccessFile archiveIndex, List<ArchiveEntry> archiveEntries) throws IOException {
         TimeProfiler.start("writeArchiveIndex");
 
+        int version = archiveIndexVersion(archiveEntries);
+
         // INDEX
-        archiveIndex.writeInt(VERSION); // Version
+        archiveIndex.writeInt(version); // Version
         archiveIndex.writeInt(0); // Pad
         archiveIndex.writeLong(0); // UserData, used in runtime to distinguish between if the index and resources are memory mapped or loaded from disk
         archiveIndex.writeInt(0); // EntryCount
@@ -260,12 +268,18 @@ public class ArchiveBuilder {
         int entryOffset = (int) archiveIndex.getFilePointer();
         alignBuffer(archiveIndex, 4);
 
-        ByteBuffer indexBuffer = ByteBuffer.allocate(4 * 4 * archiveEntries.size());
+        ByteBuffer indexBuffer = ByteBuffer.allocate(getArchiveEntryDataSize(version) * archiveEntries.size());
         for (ArchiveEntry entry : archiveEntries) {
-            indexBuffer.putInt(entry.getResourceOffset());
-            indexBuffer.putInt(entry.getSize());
-            indexBuffer.putInt(entry.getCompressedSize());
-            indexBuffer.putInt(entry.getFlags());
+            if (version == VERSION_5) {
+                indexBuffer.putInt((int) entry.getResourceOffset());
+                indexBuffer.putInt(entry.getSize());
+                indexBuffer.putInt(entry.getCompressedSize());
+                indexBuffer.putInt(entry.getFlags());
+            } else {
+                indexBuffer.putLong(packArchiveEntryOffsetAndFlags(entry));
+                indexBuffer.putInt(entry.getSize());
+                indexBuffer.putInt(entry.getCompressedSize());
+            }
         }
         archiveIndex.write(indexBuffer.array());
 
@@ -284,7 +298,7 @@ public class ArchiveBuilder {
 
         // Update index header with offsets
         archiveIndex.seek(0);
-        archiveIndex.writeInt(VERSION);
+        archiveIndex.writeInt(version);
         archiveIndex.writeInt(0); // Pad
         archiveIndex.writeLong(0); // UserData
         archiveIndex.writeInt(archiveEntries.size());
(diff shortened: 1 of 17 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/GameProjectBuilder.java
```diff
@@ -290,118 +291,127 @@ static public void transformGameProjectFile(BobProjectProperties properties) {
...
+                parent.mkdirs();
+            }
+            Files.move(sourceFile.toPath(), destination.toPath(), StandardCopyOption.REPLACE_EXISTING);
+            return;
+        }
+
+        try (FileInputStream inputStream = new FileInputStream(sourceFile)) {
+            output.setContent(inputStream);
+        }
+    }
```
AGulev, line 311: it's not directly related, bbut I noticed how slow test it, so I replaced it with a simple move (as we do for liveupdate)

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/GameProjectBuilder.java
```diff
@@ -290,118 +291,127 @@ static public void transformGameProjectFile(BobProjectProperties properties) {
...
+
     @Override
     public void build(Task task) throws CompileExceptionError, IOException {
-        FileInputStream archiveIndexInputStream = null;
```
AGulev, line 295: this change is actually small, use 

Image

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ArchiveBuilder.java
JCash: I got a review comment:
"
[P1] ArchiveBuilder still silently truncates offsets above 4 GiB: [ArchiveBuilder.java#L221](https://github.com/defold/defold/blob/6ca566f9b52b5ee3a41e28deb19ae00c47eef452/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ArchiveBuilder.java#L221) casts the long file pointer to int, then writes it as the 32-bit archive offset. The runtime format is still uint32_t ([resource_archive.h#L72](https://github.com/defold/defold/blob/6ca566f9b52b5ee3a41e28deb19ae00c47eef452/engine/resource/src/resource_archive.h#L72)), so this PR fixes 2-4 GiB unsigned offsets but a larger .arcd will wrap and build a corrupt archive. Add an explicit > 0xffffffffL build error unless the format is being upgraded to 64-bit offsets."

engine/resource/src/resource_archive.cpp
```diff
@@ -318,7 +366,10 @@ namespace dmResourceArchive
         {
             // we need to read from the file on disc
             FILE* resource_file = afi->m_FileResourceData;
-            fseek(resource_file, resource_offset, SEEK_SET);
+            if (SeekResourceData(resource_file, resource_offset) != 0)
+            {
+                return dmResourceArchive::RESULT_IO_ERROR;
+            }
```
JCash: "[P2] The new seek/read failures are still dropped by the provider layer. ReadEntry() and ReadEntryPartial() now return RESULT_IO_ERROR on failed large-file seeks ([resource_archive.cpp#L369](https://github.com/defold/defold/blob/6ca566f9b52b5ee3a41e28deb19ae00c47eef452/engine/resource/src/resource_archive.cpp#L369), [#L494](https://github.com/defold/defold/blob/6ca566f9b52b5ee3a41e28deb19ae00c47eef452/engine/resource/src/resource_archive.cpp#L494)), but provider_archive.cpp ignores those return values and reports success ([#L230](https://github.com/defold/defold/blob/6ca566f9b52b5ee3a41e28deb19ae00c47eef452/engine/resource/src/providers/provider_archive.cpp#L230), [#L243](https://github.com/defold/defold/blob/6ca566f9b52b5ee3a41e28deb19ae00c47eef452/engine/resource/src/providers/provider_archive.cpp#L243)). That can preserve the old “successful read, invalid resource data later” failure mode for unsupported/failed seeks. Propagate the archive read result to a provider error."

JCash (changes requested):
Overall good
But seems like there are two areas worth looking at?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ArchiveBuilder.java
AGulev: do we want to support 4GB+ archives?

AGulev:
@codex review

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ArchiveBuilder.java
```diff
@@ -60,9 +60,15 @@ public class ArchiveBuilder {
 
     private static Logger logger = Logger.getLogger(ArchiveBuilder.class.getName());
 
-    public static final int VERSION = 5;
+    public static final int VERSION_5 = 5;
+    public static final int VERSION_6 = 6;
+    public static final int VERSION = VERSION_6;
```
JCash: While I understand the rationale here, we'd have to support two versions throughout.
I.e. next actual version would them bump the current version _two_ steps, VERSION_7 (32 bit) and VERSION_8 (64 bit).
One variable now means two things, which I think will become troublesome.

I think we should instead, have a header flag (in Version 6), like `IS64BIT`, and set it to false by default.

JCash: Question is: Is it a problem if we update directly to version 6?
While it's "easy" to keep a version 5 around, it feels wrong.

engine/resource/src/resource_archive.cpp
```diff
@@ -43,6 +57,160 @@ namespace dmResourceArchive
...
+        if (archive->m_ArchiveFileIndex && archive->m_ArchiveFileIndex->m_Entries)
+        {
+            return archive->m_ArchiveFileIndex->m_Entries;
+        }
+
+        uint32_t entry_offset = dmEndian::ToNetwork(archive->m_ArchiveIndex->m_EntryDataOffset);
+        return (EntryData*)((uintptr_t)archive->m_ArchiveIndex + entry_offset);
+    }
+
+    static FILE* OpenResourceData(const char* path)
```
JCash: I recommend calling it `FileOpen64()`.
We probably eventually want to move it to `dmSys::` namespace, but that's for later.

engine/resource/src/resource_archive.cpp
```diff
@@ -43,6 +57,160 @@ namespace dmResourceArchive
...
+        setvbuf(file, 0, _IONBF, 0);
+        return file;
+#elif defined(__linux__)
+        return fopen64(path, "rb");
+#else
+        return fopen(path, "rb");
+#endif
+    }
+
+    static int SeekResourceData(FILE* file, uint64_t offset)
```
JCash: -> `FileSeek64()`

JCash (changes requested):
Overall good.
I think we can move to version 6 though, and support both 32/64 bit archives.

AGulev:
@codex review

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ArchiveBuilder.java
AGulev: Fixed, but I still support v5 reading. Now when I think about it I'm not sure anymore we need this back compatability (an ability to read v5 archives).
It is kinda a breaking change, but since it happens in both pipeline and engine it's more like internal implementation detail.

But I can imagine some systems use one engine with many assets e.g. our Examples page.

What do you think?

JCash: If it's only our examples, we should be able to rebuild them all with the same defold version?

> but I still support v5 reading. Now when I think about it I'm not sure anymore we need this back compatability (an ability to read v5 archives).

I'm confused, do you want v5 or not? 🤔 

I vote removing V5 support, as it _should_ be an internal detail.

JCash (dismissed):

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/archive/ArchiveBuilder.java
AGulev: My initial intention was to keep it.
but after the latest refactoring I think we don't really need it and I'm asking for you opinion here

if you vote for removing I'll remove it. Thank you!

AGulev:
@codex review

PR #12601, opened 2026-06-15
Use VkQuality to determine if Vulkan or OpenGLES should be used on Android

### LINKED ISSUE #12586: Use VkQuality
Integrate https://github.com/android/vkquality to detect if Vulkan och OpenGL should be used.

Jhonnyg:
We now use the VkQuality library (https://github.com/android/vkquality) to gatekeep which Android devices are allowed to run Vulkan (or fallback to OpenGLES).

Fixes #12586 

#### Technical changes

* the VkQuality library is added as an "external" dependency. It's built with CMake instead of waf
* the build.py script can build specific platforms by passing the `--package=<package>` flag
* if there are multiple graphics adapters being linked, the DefoldActivity will run the VkQuality check before the engine starts and then cache this result
* when checking for vulkan support in the graphics adapter selection, we now query this result and then either block or allow the app to create a device and test for if any compatible GPUs are available


Jhonnyg:
@codex review

engine/engine/content/builtins/manifests/android/AndroidManifest.xml
```diff
@@ -27,6 +27,7 @@
 
         <meta-data android:name="android.max_aspect" android:value="2.1" />
         <meta-data android:name="android.notch_support" android:value="true"/>
+        <meta-data android:name="com.defold.vkquality.enabled" android:value="{{defold.vkquality.enabled}}"/>
```
Jhonnyg, line 30: Maybe this introduces a new naming scheme in contrast to the "dynamo" name we use for similar things. Do we want to stick to this or something similar to what we have elsewhere?

e.g "com.dynamo.android.DefoldActivity"

I'm personally not that fond of the dynamo name, so I'd rather stick to Defold but it's up for debate!

Jhonnyg: Also, the point of this setting is that the java side is responsible for calling into the VkQuality library during bootstrapping (before there is an engine), so we use this switch for controlling that behaviour.

com.dynamo.cr/com.dynamo.cr.bob/build.gradle
```diff
@@ -292,6 +292,10 @@ task compileBobLight(dependsOn: [compileSrc, createLuaJitShareZip]) {
         exclude '**/*dSYM.zip'
     })
     inputs.file("${project.dynamoHome}/ext/share/java/bundletool-all.jar").optional()
+    inputs.files(fileTree("${project.dynamoHome}/ext/lib") {
+        include 'armv7-android/libvkquality.so',
+                'arm64-android/libvkquality.so'
+    })
```
Jhonnyg, line 301: I don't think/know if we do this for other libraries, we mostly build them as static libraries, but in this case we need to bundle the actual .so file as well since the vkquality.jar is responsible for that part, we just need to make sure it is included in the bundle.

engine/graphics/src/dx12/graphics_dx12_private.h
```diff
@@ -274,7 +274,7 @@ namespace dmGraphics
     void            SetupMainRenderTarget(DX12Context* context, DXGI_SAMPLE_DESC sample_desc);
     HRESULT         CreateShaderModule(DX12Context* context, const char* target, void* data, uint32_t data_size, DX12ShaderModule* shader);
 
-    bool            DX12IsSupported();
+    bool            DX12IsSupported();
```
Jhonnyg: Not sure what this is about, line ending changes perhaps..

engine/graphics/src/vulkan/android/graphics_vulkan_android.cpp
```diff
@@ -148,6 +152,239 @@ namespace dmGraphics
...
+        }
+
+        va_list args;
+        va_start(args, format);
+        vsnprintf(error_buffer, error_buffer_size, format, args);
+        va_end(args);
+        error_buffer[error_buffer_size - 1] = 0;
+    }
+
+    static void AndroidSetVkQualityJavaExceptionError(JNIEnv* env, const char* message, char* error_buffer, uint32_t error_buffer_size)
```
Jhonnyg: There's some JNI glue here to call java functions from the activity class, but I wonder if we don't already have such functionality somewhere else? This feels a bit like re-implementing the wheel..

Jhonnyg:
@codex review

engine/engine/content/builtins/manifests/android/AndroidManifest.xml
JCash: I think defold is clearer for 3rd party users

engine/graphics/src/vulkan/android/graphics_vulkan_android.cpp
JCash: we have `android.cpp` in dlib, but all this seem very specialized, and only used for the graphics+android+vulkan, so I guess it's in a good place?

engine/graphics/src/vulkan/android/graphics_vulkan_android.cpp
```diff
@@ -148,6 +152,239 @@ namespace dmGraphics
...
+        {
+            AndroidSetVkQualityJavaExceptionError(env, "Could not load com.dynamo.android.DefoldActivity", error_buffer, error_buffer_size);
+            return false;
+        }
+
+        jmethodID method_id = env->GetStaticMethodID(activity_class, method_name, "()I");
+        if (env->ExceptionCheck() || !method_id)
+        {
+            char message[128];
+            snprintf(message, sizeof(message), "Could not find DefoldActivity.%s()I", method_name);
```
JCash: -> `dmSnPrintf`

engine/graphics/src/vulkan/android/graphics_vulkan_android.h
```diff
@@ -26,6 +26,8 @@ namespace dmGraphics
     void     AndroidVulkanBeginFrame(VulkanContext* context);
     bool     AndroidVulkanHandleWindowSurfaceChange(VulkanContext* context, uint32_t window_width, uint32_t window_height);
     void     AndroidVulkanInitializeContext(VulkanContext* context);
+    bool     AndroidGetVkQualityResult(int32_t* init_result, int32_t* recommendation, char* error_buffer, uint32_t error_buffer_size);
```
JCash: I think some code comment about usage and expected values of `init_result` and `recommendation` would be good.

Also, it's not used outside of the .cpp file, so we can remove it from here.

external/glfw/java/com/dynamo/android/DefoldActivity.java
```diff
@@ -154,6 +165,82 @@ public boolean isAppInstalled(String appPackageName) {
 
     private boolean keyboardActive;
 
+    private static synchronized void runVkQualityPreflight(Context context) {
```
JCash: I wasn't expecting VkQuality stuff in two places?
Could it be put into one place?

E.g. 1) could the `graphics_vulkan_android.cpp` do this part as well?

or 2) Could it all be done here, and we set a boolean flag (+ message), and the c++ part only gets this part via JNI?

In short, I'd like to see one owner of the code.

external/vkquality/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
JCash, line 1: I don't see it added to COMPLYING_WITH_LICENSES.md where we keep the full list of our dependencies and their licenses.

scripts/build.py
```diff
@@ -2447,12 +2457,99 @@ def build_engine(self):
...
+        build_dir = join(source_dir, 'build', platform)
+        install_dir = join(self.dynamo_home, package_name)
+        package_dir = join(self.defold_root, 'packages')
+        package_path = join(package_dir, '%s-%s.tar.gz' % (package_name, platform))
+
+        ndk_root = join(self.dynamo_home, 'ext', 'SDKs', PACKAGES_ANDROID_NDK)
+        toolchain_file = join(ndk_root, 'build', 'cmake', 'android.toolchain.cmake')
+        android_jar = join(self.dynamo_home, 'ext', 'share', 'java', 'android.jar')
+        android_abi = 'arm64-v8a' if platform == 'arm64-android' else 'armeabi-v7a'
+        android_api = sdk.ANDROID_64_NDK_API_VERSION if platform == 'arm64-android' else sdk.ANDROID_NDK_API_VERSION
```
JCash: I feel this is a lot of platform+library specific knowledge that should be kept out of the build.py.

Is there a way to reuse our settings from the ./script/cmake/sdk_* and functions here?
E.g. by inserting some .cmake files on the command line or something?

JCash (changes requested):
Overall good.
Some questions regarding code ownership, and code build abstractions.

Jhonnyg:
@codex review pr

JCash (dismissed):

PR #12607, opened 2026-06-16
Allow removing GUI runtime texture mappings

### LINKED ISSUE #12049: Add "nil" support for gui.set
Support `nil` for `gui.set` - to be able manage created resources usage manually 

No alternatives 

I want to create texture resources then set it in `gui`, after some time be able to release it fully manually.

AGulev:
GUI scripts can now remove a named runtime texture mapping by calling `gui.set(msg.url(), "textures", nil, {key = ...})`. This lets projects release runtime-created atlas resources after they are no longer used by a GUI scene.

Fix https://github.com/defold/defold/issues/12049

### Technical changes

- Added `nil` handling to `gui.set()` for the GUI component `textures` property, restricted to `msg.url()` and keyed texture mappings.
- Releases component-owned resource property references when dynamic GUI texture mappings are deleted.
- Ensures failed texture mapping insertion releases the newly acquired resource reference.
- Allows same-key dynamic texture replacement when the dynamic texture table is already full, preventing stale entries after the old backing resource is released.
- Added gamesys coverage for replacing and removing runtime atlas mappings before `resource.release()`.
- Added GUI coverage for replacing an existing dynamic texture at full table capacity.

### DIFF
--- a/engine/gui/src/gui.cpp
+++ b/engine/gui/src/gui.cpp
@@ -745,10 +745,21 @@ namespace dmGui
 
     static Result AddTexture(HScene scene, dmHashTable64<TextureInfo>& info_array, dmhash_t texture_name_hash, HTextureSource texture_source, NodeTextureType texture_type, uint32_t original_width, uint32_t original_height, uint32_t image_type)
     {
-        if (info_array.Full())
-            return RESULT_OUT_OF_RESOURCES;
+        TextureInfo texture_info(texture_source, texture_type, original_width, original_height, image_type);
+        TextureInfo* existing_texture_info = info_array.Get(texture_name_hash);
 
-        info_array.Put(texture_name_hash, TextureInfo(texture_source, texture_type, original_width, original_height, image_type));
+        if (existing_texture_info)
+        {
+            *existing_texture_info = texture_info;
+        }
+        else
+        {
+            if (info_array.Full())
+                return RESULT_OUT_OF_RESOURCES;
+
+            info_array.Put(texture_name_hash, texture_info);
+        }
+
         UpdateTexture(scene, texture_name_hash, texture_source, texture_type);
 
         return RESULT_OK;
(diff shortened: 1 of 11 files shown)


engine/gui/src/gui.cpp
```diff
@@ -745,10 +745,21 @@ namespace dmGui
 
     static Result AddTexture(HScene scene, dmHashTable64<TextureInfo>& info_array, dmhash_t texture_name_hash, HTextureSource texture_source, NodeTextureType texture_type, uint32_t original_width, uint32_t original_height, uint32_t image_type)
     {
-        if (info_array.Full())
```
AGulev, line 748: it's not directly related, but even if all slots are busy we should be able to replace an atlas

engine/gui/src/gui_script.cpp
```diff
@@ -893,19 +907,55 @@ namespace dmGui
...
+            dmMessage::URL sender;
+            dmScript::GetURL(L, &sender);
+            dmMessage::URL target;
+            dmScript::ResolveURL(L, 1, &target, &sender);
+            bool is_self = (sender.m_Socket == target.m_Socket) &&
+                           (sender.m_Path == target.m_Path) &&
+                           (sender.m_Fragment == target.m_Fragment);
+            if (!is_self)
+            {
+                return DM_LUA_ERROR("'gui.set()' can only be used to change a property of the GUI component itself, use 'msg.url()'");
```
britzl, line 933: What do we mean by the last bit of the message: `use 'msg.url()'`

AGulev: that's how gui.set for resources work : 
CleanShot 2026-06-17 at 09 34 13@2x
https://defold.com/ref/stable/gui-lua/#gui.set:node-property-value-options

it's very unusual for `gui.set()` to change itself as a component not as internals of the comonent. Only `go.set()` can do it, so  it was made this way

AGulev:
@britzl I answered your question and PR waits for your review

britzl (approved):

PR #12613, opened 2026-06-17
Add sample count setting for multisampled render target attachments

### LINKED ISSUE #11036: Enable FSAA for individual attachments in custom render targets

FSAA (enabled via **Project -> Display -> Samples**) only works when drawing to the default render target (i.e., directly to the screen). However, it's common to draw to intermediate RTs first in order to do post-processing, but FSAA will not be applied in this case. It would be very useful to allow FSAA for any render target so that games that do post-processing can look more polished as well.

Instead of a global setting that only applies to the default render target, allow FSAA to be **enabled / disabled for individual attachments** in any render target. This level of granularity is necessary because, for example, attachments that contain colors should be anti-aliased, but attachments that contain other data such as normals, depth, material properties, etc, should not be anti-aliased.

A project can use custom anti-aliasing shaders, but these will incur a performance cost, and it's suboptimal to do this when a solution is already part of the engine.

Brief discussion here: https://forum.defold.com/t/fsaa-only-works-on-default-render-target/81107

Jhonnyg:
Added configurable FSAA sample counts to render targets and implemented multisampled render targets with automatic color resolving. The sample count applies to all color, depth, and stencil attachments in the render target.

Example

```lua
local target = render.render_target({
    sample_count = 4,

    [graphics.BUFFER_TYPE_COLOR0_BIT] = {
        format = graphics.TEXTURE_FORMAT_RGBA,
        width = 1280,
        height = 720
    },

    [graphics.BUFFER_TYPE_DEPTH_BIT] = {
        format = graphics.TEXTURE_FORMAT_DEPTH,
        width = 1280,
        height = 720
    }
})
```

The same render-target-level sample_count setting is available for .render_target resources in the editor.
Requested sample counts are normalized to a value supported by the active graphics adapter. The effective value is available through `resource.get_render_target_info()`.

Fixes #11036

#### Technical changes

* Adds a configurable FSAA sample count per render target in Lua, resources, and the editor.
* Applies the selected sample count uniformly to all color, depth, and stencil attachments.
* Adds shared validation, normalization, fallback behavior, and effective sample-count reporting.
* Implements multisampled render targets with automatic color resolves for Vulkan, Metal, and OpenGL.
* Updates render-target creation, resizing, deletion, render-pass, clear-pipeline, and graphics-pipeline handling for multisampling.
* Falls back to single-sample render targets on adapters that do not yet support multisampled off-screen render targets.


Jhonnyg:
@codex review

vlaaad (dismissed):
Editor changes look good 👍

engine/graphics/src/metal/graphics_metal.mm
```diff
@@ -1633,6 +1647,14 @@ static void BeginRenderPass(MetalContext* context, HRenderTarget render_target)
                 colorAttachment->setResolveTexture(tex->m_Texture);
                 colorAttachment->setStoreAction(MTL::StoreActionMultisampleResolve);
             }
+            else if (rt->m_Base.m_SampleCount > 1)
+            {
+                MetalTexture* resolve_tex = GetAssetFromContainer<MetalTexture>(context->m_BaseContext.m_AssetHandleContainer, rt->m_Base.m_TextureColorResolve[i]);
+                assert(resolve_tex && resolve_tex->m_Texture);
```
JCash, line 1653: Since user can set sample count now, is there a risk they abort their game with this assert()?

"[assert(resolve_tex && resolve_tex->m_Texture)](https://github.com/defold/defold/blob/6265652a8f1f76249f310ce99adf763a6f47a2d0/engine/graphics/src/metal/graphics_metal.mm#L1650-L1656): real Metal risk. The native texture creation result is [stored without checking for nil](https://github.com/defold/defold/blob/6265652a8f1f76249f310ce99adf763a6f47a2d0/engine/graphics/src/metal/graphics_metal.mm#L4058-L4064), even though Metal explicitly permits texture creation to fail and return nil ([Apple documentation](https://developer.apple.com/documentation/metal/mtldevice/maketexture%28descriptor%3A%29)). If the MSAA allocation succeeds but the following resolve allocation fails—most plausibly under GPU-memory pressure—the first clear/draw hits this assert. Resizing has the same unchecked sequence."

JCash (dismissed):

PR #12620, opened 2026-06-19
Show release notes during update process and bundle with the editor

### LINKED ISSUE #7186: Show recent changes before updating the editor
It is not clear to the user what changes an update of the editor brings with it both in terms of changes to the editor but also to the runtime engine.

When clicking "Update Available" in the editor a popup with a change log should be shown to help the user make an informed decision about whether to update or not.

Link to release notes on the forum

JosephFerano:
After pressing "Update Available", you will have a list of changes available so you know what you are getting when updating. Release notes can also be viewed now from the editor from `Help > Release Notes`.

Screenshot from 2026-07-10 19-07-49

Fixes #7186

### Technical changes

Right now we're using a throwaway channel (release-notes-view) so we can test the full pipeline without touching stable/beta/alpha, so pushing on this branch will publish release notes json/md/manifest to S3 on a channel for this branch. I tagged the dev only stuff DEV-ONLY / REVERT ME. I'll delete these as soon as things are approved. Feel free the delete the S3 files after this branch is merged.

Made some changes to `releasenotes_github_projectv2.py` to make it more robust. I also made release notes gate beta and stable releases if there are problems finding issues or anything else. If this is not desired and it should only be a warning, we can relax that part. Also added a flag for the CI that helps it check issues using the github API, avoiding the shallow clone problem. I also kept it where you can generate the release notes manually and it'll use the local repo instead of the web API checks. There's also commands in `build.py` to upload to S3 if we want to generate then deploy manually (like fixing a type or something).

Release notes are bundled in the editor and shown to the user when they first launch a new version, so prefs now keeps track of which versions have been opened. The markdown file goes into the cache directory, so on linux for me that's
`~/.local/state/Defold/unpack/<hash>/_defold/releasenotes`

To test locally, I've been using these helpers;

From `updater.clj`, this will fetch the correct notes into memory
```clojure
(defonce dev-updater (atom nil))
(with-redefs [system/defold-channel (constantly "release-notes-view")
                start-timer! (fn [updater _ _]
                               (check! updater)
                               (reset! dev-updater updater)
                               updater)]
    (start!))
```

To enable the "Update Availabler" button;
```clojure
(require 'dev)
  (editor.ui/run-now
    (init-pending-update-indicator! (editor.ui/main-stage)
                                    (.lookup @the-root "#update-link")
                                    (dev/project)
                                    (-> @the-root editor.ui/context :env :changes-view)
                                    @editor.updater/dev-updater
                                    (-> @the-root editor.ui/context :env :localization)))
```

To have it download a published build from the CI, use this
```clojure
(swap! (:state-atom @dev-updater) assoc
         :server-sha1 "122fe2ab49f46dbdcfedb5eeb5634b1deb98848f"
         :downloaded-sha1 nil :installed-sha1 nil :current-download nil)
```

Note that running the restart and update process will just dump the files to the `editor/` directory of the Defold source code. Install it manually and launch it, you should see the release notes popup on first launch as a markdown tab.


vlaaad:
@codex review

vlaaad:
Before I start properly reviewing, 2 quick questions that popped into my mind after a brief skim:
1. Why bundle the notes into the editor? Isn't the idea to show them BEFORE installing an update to see what's new? And if that's the case, what do we bundle if a new update by definition is not known to old version?
2. Why make release notes a graph node at all?

JosephFerano:
@vlaaad 

> 1. Why bundle the notes into the editor? Isn't the idea to show them BEFORE installing an update to see what's new? And if that's the case, what do we bundle if a new update by definition is not known to old version?

The notes are the same version as the release they are bundled in. Those are in markdown format. The ones that are fetched are newer releases and those are in json format. These markdown bundled notes are shown to the user after they've updated and restarted on first launch of the new version and they can view them with `Help > Release Notes`. The new CI steps automate generating and bundling the notes with the editor. The idea was to give users the opportunity to view the release notes in a nicer format, with more details, after updating, while also automating the generation step.

> 2. Why make release notes a graph node at all?

I believe you are referring to the `make-releasenotes-snapshot`.  After bundling the release notes, in order to make them openable as a tab, so they render nicely, I found that I could piggy back off of the `make-debugger-snapshot` code and put the release notes in markdown format there. If there's a better way to do it, please let me know.

scripts/releasenotes_github_projectv2.py
```diff
@@ -531,18 +623,29 @@ def check_issue_commits(issues):
...
-    yellow("  %d issue(s) present on beta via reference commits" % merge_beta_count)
+    yellow("  %d issue(s) present on dev via reference commits" % reference_dev_count)
+    yellow("  %d issue(s) present on beta via reference commits" % reference_beta_count)
     red("  %d issue(s) not present on dev" % missing_dev_count)
     red("  %d issue(s) not present on beta" % missing_beta_count)
     red("  %d issue(s) not present on dev+beta" % missing_dev_beta_count)
 
+    if required_missing_count > 0:
+        red("\nRelease notes audit FAILED: %d issue(s) missing from required branch(es): %s" % (required_missing_count, ", ".join(required_branches)))
+        sys.exit(1)
```
vlaaad: I'm not sure this should fail the release notes gen. I tried the old vs new script on 1.13.0, both report that 83 issues are not present in beta. 

Old says:
```
  Checking #12365 'Resolve game.project resource defaults before rendering the form' (https://github.com/defold/defold/pull/12370)
    dev  OK via merge commit (a3497356ed6c82f7dd2b1c6356444d81cf353a6b)
    Missing from beta
```

New says:
```
  Checking #12365 'Resolve game.project resource defaults before rendering the form' (https://github.com/defold/defold/pull/12370)
    Missing from dev+beta
```

First, the older version is more informative. But also, this change exists in beta! Here: https://github.com/defold/defold/commits/beta/?after=936a93ad121b9c7c5601608941703935683638c4+104 Ctrl+f 12370

So we should either not fail the release notes gen, or not emit a false-positive error.

scripts/releasenotes_github_projectv2.py
```diff
@@ -472,8 +552,19 @@ def parse_github_project(version):
...
-    print("\nChecking issue commits for dev and beta presence")
+def audit_branches_for_channel(channel):
+    # Which branches a note's fix has to be on, per release channel.
+    # beta/stable ship from beta with everything merged forward -> require both.
+    # alpha (dev track) and custom validation channels build off dev / a feature
+    # branch, so requiring beta would fail them for fixes that just haven't
+    # reached beta yet -> require dev only. Unknown channel (manual run with no
+    # --channel) keeps the strict default.
+    if channel in ("beta", "stable", None):
+        return ["dev", "beta"]
```
vlaaad: Sometimes we commit something into beta without committing it to dev. see, e.g., https://github.com/defold/defold/commit/936a93ad121b9c7c5601608941703935683638c4.

editor/src/clj/editor/ui/updater.clj
```diff
@@ -13,8 +13,14 @@
 ;; specific language governing permissions and limitations under the License.
 
 (ns editor.ui.updater
-  (:require [editor.dialogs :as dialogs]
```
vlaaad, line 16: We should use a styled dialog for consistency.

editor/src/clj/editor/ui/updater.clj
```diff
@@ -45,6 +51,59 @@
...
+   :scene {:fx/type fx.scene/lifecycle
+           :stylesheets [(str (io/resource "dialogs.css"))
+                         (str (io/resource "editor.css"))]
+           :root
+           {:fx/type fx.v-box/lifecycle
+            :style-class "release-notes-dialog"
+            :children [{:fx/type markdown/view
+                        :v-box/vgrow :always
+                        :content content
+                        :project nil
```
vlaaad: Hmm. Pretty sure it's required.

editor/src/clj/editor/resource_watch.clj
```diff
@@ -205,6 +220,7 @@
       {:snapshot (combine-snapshots (list* (make-builtins-snapshot workspace)
                                            (make-directory-snapshot workspace project-directory)
                                            (make-debugger-snapshot workspace)
+                                           (make-releasenotes-snapshot workspace)
```
vlaaad: Let's avoid treating release notes as a resource for now.

editor/src/clj/editor/updater.clj
```diff
@@ -40,6 +41,129 @@
 (defn- update-url [archive-domain channel]
   (format (get-in connection-properties [:updater :update-url-template]) archive-domain channel))
 
+(defn- release-notes->markdown
+  ^String [release-notes]
+  (let [issue-types ["BREAKING CHANGE" "NEW" "FIX"]
+        issue->closed-issues (fn [{:keys [closed_issues repository]}]
+                               (string/join ","
+                                            (into []
+                                                  (map (fn [issue-number]
```
vlaaad: Perf nitpicking: `coll/join-to-string` + `e/map` instead of `into` to avoid intermediate collections. BTW, I merged a clj-kondo lint config to dev that suggests using coll/join-to-string

JCash:
> Isn't the idea to show them BEFORE installing an update

I believe we identified two use cases:
* show the upcoming changes, to allow user to not update
* show the release notes of the _current_ version, reminding the user to know what's new in this release

scripts/build.py
```diff
@@ -3212,6 +3212,82 @@ def push_tag(self, tag):
         cmd = 'git push -f origin %s' % tag
         run.shell_command(cmd)
 
+    def _build_editor_release_notes(self):
```
JCash: Nowadays, we try to avoid build.py from growing.
Please move the bulk of this code to e.g. `./build_tools/releasenotes.py`

vlaaad (changes requested):
## Dialog styling
The dialog doesn't seem to be styled like the rest of the editor dialogs:
Image

For comparison, styled dialog:
Image

## Hidden resource

Image

I think it's really neat that we show it as a tab. Unfortunately, I think we are not yet ready to do it. Resources are a part of a game pipeline. Other `_defold` resources are resources because they are built and used in the game. Here, you made it a resource to work around an issue that we can't open tabs that are not resources. I think we should implement a way to show tabs that are not resources. Other use cases for this functionality: showing the asset portal as a tab, showing diffs as a tab. But I think this PR is not the right place to implement support for non-resource tabs. I added a feature request for this: https://github.com/defold/defold/issues/12691 Here, I'd suggest for now to make release notes a java classpath entry instead of a project resource, and use a dialog that shows a markdown view for the release notes.

## Resource md vs dialog contents

I noticed that the update dialog shows a simple list of one-liner notes, while Help -> Release Notes shows them with descriptions. Should/can we show the same thing?

## Forum as a source of truth?

The issue description (https://github.com/defold/defold/issues/7186) says we should treat forum posts as a source of truth. Doesn't seem like we do it?

## Alpha channel and scope

A question somewhat related to the previous section. I haven't seen any design docs for the feature, BTW... But I remember we talked that we want to also have release notes for the alpha channel, but we don't have different version numbers there. Doesn't seem like the current design supports this use-case? I see that `manifest.json` only includes a public semver version number, so it can't work for the alpha channel? What we have:
1. `manifest.json` — semver versions https://d.defold.com/editor2/channels/release-notes-view/release-notes/manifest.json
2. `<semver>.json`, e.g. https://d.defold.com/editor2/channels/release-notes-view/release-notes/1.13.1.json — a list of release notes per semantic version

TBH, I was not expecting us to have this design. Instead, I'd expect something like this:
1. manifest.json — a list of the latest released shas, perhaps limited to some number like the latest 256.
2. updater checks for the udpate, sees that latest sha is "abc123", it's current is "000aaa". It checks manifest.json, records all shas between 000aaa and abc123, and fetches:
3. `<sha>.json` — a list of release notes for a particular sha.

I'd expect this approach because it seems to be capable to show e.g. smaller updates on the alpha/beta branch too.

I would also expect to have a specialized script that on every update of a channel (when we write to `update-v4.json`), before doing it, would always update manifest.json with new sha and <sha>.json with release notes for all commits between the previous sha in `update-v4.json` and a published one. 

But the approach I'm expecting is not the same as "treat forum as a source of truth" stated in the issue, and is not the same as what's implemented here. Seems like we (I?) missed a design doc and some discussion about the approach we should take?

scripts/build.py
```diff
@@ -3852,6 +3934,7 @@ def _form_env(self, inherit = True):
 build_builtins   - Build builtin content archive
 bump             - Bump version number
 release          - Release editor
+upload_editor_release_notes - Manual: upload editor release-notes.{md,json} for --version to S3 for --channel
```
JCash: Perhaps "upload_release_notes"? GIven that the release notes are about a lot more than the editor

JCash (changes requested):
Overall good
Some minor reorganization and renameing requested.

editor/src/clj/editor/ui/updater.clj
JosephFerano: I threaded the project reference, but welcome.clj calls this and does not have a project so it would be nil as well.

JosephFerano:
@vlaaad 

> But I remember we talked that we want to also have release notes for the alpha channel, but we don't have different version numbers there. Doesn't seem like the current design supports this use-case?

The current design can support it. Currently updates are still triggered by SHA, not by version. It was not implemented because I think Mathias and I settled on getting beta/stable merged in first and leave the "diffing" for later. My plan was to to write the current version's release note's as a json to the editor's cache directory. Then, when we detect an update, we download all versions the user will be updating to including the current one (in case there are newer updates to the current), and then from the editor client we dedupe the release notes with the one on disk. This will effectively show the diff of notes.

> I would also expect to have a specialized script that on every update of a channel (when we write to update-v4.json), before doing it, would always update manifest.json with new sha and .json with release notes for all commits between the previous sha in update-v4.json and a published one.

There is an `update_manifest` function that updates the manifest if there's a new version. It has been moved to a new `build_tools/releasenotes.py` per request. But I guess you mean the design below.

> TBH, I was not expecting us to have this design. Instead, I'd expect something like this:
> 
> 1. manifest.json — a list of the latest released shas, perhaps limited to some number like the latest 256.
> 2. updater checks for the udpate, sees that latest sha is "abc123", it's current is "000aaa". It checks manifest.json, records all shas between 000aaa and abc123, and fetches:
> 3. <sha>.json — a list of release notes for a particular sha.

After analyzing this approach, I have several concerns about this. The main one is now potentially downloading dozens of release notes at a time on alpha, depending on when the last time was that the user opened their editor. Even if we cache as we go, there might still be times where we have to burst download quite a few. This was probably the main reason I went with the current approach. It's only ever 1 download per version, even on alpha, and then once implemented, the client does the work of deduping. In either scenario we should still be caching. Still feels like quite a lot of IO.

There's also a question of how the CI would handle this. I think you implied that a new SHA would just have the diff from the previous release notes. On alpha that would just be 1 release note per json? We would have to make modifications to the release notes generator script to avoid all the extra work (and avoid potential rate limiting by github) so we only fetch the one release note, because currently the way the script works is just to grab all issues tagged with the current VERSION. The other alternative which is probably not much better is each alpha <sha>.json contain all release notes for that VERSION up until that point. It's immutable which is a good, but it's a lot of extra data and the clients have to dedupe still.

I guess this is why I suggested some simple http endpoint, since these sorts of calculations can be done server side and we can have more flexibility while still keeping fetches low.

Another slight issue is the loss of human friendliness with the <sha>.json naming, it might be nicer to have <version-sha>.json if we were to do that.

Finally, there's a question that perhaps neither the current approach nor the SHA based one might be addressing and that's how "skip release notes" would be handled in the alpha channel. The user might get an update with no release notes. Less of an issue in beta/stable because those will have more user facing messages. This is where perhaps there's an alternative approach to alpha and that's using commit messages instead of release notes. Maybe that can just be an edge case we handle for "skip release notes" where if we detect this, then use the commit message instead.

Is there a scenario you've considered that I didn't take into account that favors the SHA based approach? I lean towards keeping my approach, but I would love to continue the discussion, perhaps there's something I haven't thought of yet.

vlaaad:
@codex review

editor/src/clj/editor/app_view.clj
```diff
@@ -2582,6 +2585,51 @@
          (make-open-resource-plan app-view prefs project resource opts evaluation-context))
        (perform-open-resource-plan! localization))))
 
+(defn- release-notes-resource
+  "Classpath URL of the running version's bundled release notes, staged into
+  resources/release-notes/ at build time (nil for dev builds, which ship none)."
+  []
+  (io/resource (str "release-notes/" (system/defold-version) ".md")))
```
vlaaad: Make it a delay instead of a function so we don't constantly do IO on ui thread when checking if `help.open-release-notes` is enabled.

editor/src/clj/editor/ui/updater.clj
```diff
@@ -56,20 +104,17 @@
     (ui/on-action! link
       (fn [_]
         (let [can-install? (updater/can-install-update? updater)
-              can-download? (updater/can-download-update? updater)]
+              can-download? (updater/can-download-update? updater)
+              can-get-new? (and can-download? (updater/platform-supported? updater))]
```
vlaaad: booleans with `?` suffix (would be nice to also fix `can-install?`)

editor/src/clj/editor/ui/updater.clj
```diff
@@ -45,7 +48,52 @@
...
+                       {:fx/type fxui/legacy-button
+                        :text (localization (localization/message "updater.release-notes-dialog.button.update-now"))
+                        :on-action {:result true}}]}})
+
+(defn- show-release-notes-update-dialog!
+  "Shows the release notes dialog, blocking the current thread until the user
+  dismisses it. Must be called on the JavaFX application thread. Returns true if
+  the user chose to update now, false otherwise."
+  [content project localization]
+  (fxui/show-dialog-and-await-result!
```
vlaaad: We should avoid using `fxui/show-dialog-and-await-result!` in favor of `fxui/show-stateless-dialog-and-await-result!`, which works much better with the modern cljfx. For example, you can just keep the stage `:showing true` instead of looking up props, and the lifecycle is simpler (call fn to close). It also avoids cljfx renderer abstraction (generally unnecessary)

editor/src/clj/editor/ui/updater.clj
```diff
@@ -45,7 +48,52 @@
     (updater/restart! updater)
     (Platform/exit)))
 
-(defn init! [^Stage stage link updater install-and-restart! render-progress! localization]
+(defn- release-notes-update-dialog [{:keys [content project localization] :as props}]
+  {:fx/type dialogs/dialog-stage
+   :showing (fxui/dialog-showing? props)
+   :on-close-request {:result false}
+   :title (localization (localization/message "updater.release-notes-dialog.title"))
```
vlaaad: fxui/defc with localization watcher for nicer semantics

editor/src/clj/editor/app_view.clj
```diff
@@ -2582,6 +2585,51 @@
...
+                        :text (localization (localization/message "release-notes-dialog.button.close"))
+                        :cancel-button true
+                        :on-action {:result nil}}]}})
+
+(defn show-release-notes-dialog!
+  "Shows the running version's bundled release notes in a dialog. No-op when no
+  notes shipped. Must be called on the JavaFX application thread."
+  [localization project]
+  (when-let [url (release-notes-resource)]
+    (fxui/show-dialog-and-await-result!
```
vlaaad: Similarly, prefer `show-stateless-dialog-and-await-result!`

editor/src/clj/editor/boot_open_project.clj
```diff
@@ -412,6 +413,15 @@
               (open-resource readme-resource))
             (app-view/restore-tabs-from-prefs! app-view prefs localization workspace project))
 
+          ;; The first time a given editor version is opened, surface its bundled
+          ;; release notes (once per version; the set tracks every opened version).
```
vlaaad, line 416: Why not track the last opened version?

editor/src/clj/editor/updater.clj
```diff
@@ -40,6 +42,129 @@
...
+  5)
+
+(defn- parse-version
+  "Turns \"1.13.2\" into [1 13 2] (nil for anything that isn't a plain dotted
+  number). Comparing these vectors with `compare` gives normal version ordering."
+  [s]
+  (when (string? s)
+    (let [parts (string/split s #"\.")]
+      (when (and (seq parts) (every? #(re-matches #"\d+" %) parts))
+        (mapv #(Long/parseLong %) parts)))))
```
vlaaad: Would it be preferable to use `util/natural-order` for sorting? Not sure if that comparator has the same behavior when it comes to a differing number of dots, but it intuitively compares strings with numbers...

editor/src/clj/editor/updater.clj
```diff
@@ -40,6 +42,129 @@
...
+  version strings are skipped. If we don't know the current version (dev builds),
+  everything counts, so you get the most recent few."
+  [manifest-versions current-version]
+  (let [current (parse-version current-version)]
+    (-> (coll/into-> manifest-versions []
+          (keep (fn [v]
+                  (when-some [parsed (parse-version v)]
+                    (when (or (nil? current) (pos? (compare parsed current)))
+                      [v parsed])))))
+        (->> (sort-by second #(compare %2 %1)))
```
vlaaad: `coll/descending-order`

editor/src/clj/editor/welcome.clj
```diff
@@ -587,7 +587,9 @@
                                (ui/visible! progress-bar (not (progress/done? progress)))
                                (ui/render-progress-bar! progress progress-bar))))
         install-and-restart! #(ui.updater/install-and-restart! stage updater localization)]
-    (ui.updater/init! stage update-link updater install-and-restart! render-progress! localization)))
+    ;; No project is open on the welcome screen; release-notes links are external
+    ;; (forum) urls that don't need one.
+    (ui.updater/init! stage update-link nil updater install-and-restart! render-progress! localization)))
```
vlaaad, line 592: `project` is still a required argument in `markdown/view`. If any release note text includes a link with `file://` scheme, clicking it will throw an exception. This would be our bug because we supplied a `nil` project to a view that explicitly requires a non-nil `project` while documenting it so. The solution here is to make `project` an optional prop to the markdown viewer to make it behave nicely in a non-project context such as the welcome screen, and check if it exists before attempting to do `g/with-auto-evaluation-context`

scripts/releasenotes_github_projectv2.py
```diff
@@ -275,10 +311,13 @@ def get_closing_issue(pr):
     for node in reversed(pr["closingIssuesReferences"]["nodes"]):
         issue_number = node["number"]
         repository = node["repository"]["name"]
-        return get_issue(issue_number, repository)
+        issue = get_issue(issue_number, repository)
+        if issue is not None:
+            return issue
```
vlaaad: Huh, isn't it the same code as before?

ci/ci.py
```diff
@@ -440,6 +441,48 @@ def release(channel, platform=None):
...
+        if not os.path.exists(notes_json):
+            raise Exception("%s already exists, but matching %s is missing" % (notes_md, notes_json))
+        print("%s already exists - using manually-authored notes as-is" % notes_md)
+        return
+
+    # The generator audits every issue's fix for presence on the branch(es) this
+    # channel ships from and exits non-zero - failing the job - if any are missing
+    # or generation itself errors. --use-github-compare does that check via the
+    # GitHub compare API since CI runs on a shallow clone with no branch history.
+    call('"%s" scripts/releasenotes_github_projectv2.py --version %s --channel %s --token %s --use-github-compare generate' % (
```
vlaaad, line 477: From codex (use your own judgement): alpha is described as best effort, but `call` may throw here, filing the job.

.github/workflows/main-ci.yml
```diff
@@ -502,6 +539,15 @@ jobs:
         }
       },
       { name: 'Install dependencies', run: sudo apt-get update && sudo apt-get install -y libgl1-mesa-dev libglw1-mesa-dev freeglut3-dev },
+      {
+        name: 'Download release notes',
+        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131, # v7.0.0
+        continue-on-error: true,
```
vlaaad, line 545: From codex (use your own judgement): we continue on error for stable and beta, while we should tolerate only alpha errors.

build_tools/releasenotes.py
```diff
@@ -0,0 +1,83 @@
...
+        versions = json.loads(obj.get()['Body'].read())
+    except botocore.exceptions.ClientError as e:
+        # First release on this channel -> no manifest yet. Any other error
+        # (auth, throttling, network) must NOT be ignored, or we'd publish a
+        # manifest with only this version and lose the channel's history.
+        code = e.response.get('Error', {}).get('Code')
+        if code not in ('NoSuchKey', 'NoSuchBucket', '404'):
+            raise
+    if not isinstance(versions, list):
+        versions = []
```
vlaaad: From codex (use your own judgement): invalid manifest silently destroys history here. If the manifest is somehow invalid, it's better if we fail instead of dropping history.

editor/styling/stylesheets/components/_markdown.scss
```diff
@@ -131,4 +131,12 @@
       }
     }
   }
-}
\ No newline at end of file
+}
+
+.release-notes-dialog {
```
vlaaad: `release-notes-dialog` style class seems unused.

scripts/build.py
```diff
@@ -3336,6 +3337,12 @@ def _release_web_pages(self, releases):
                                                  'sha1' : release_sha1})
         new_obj.put(Body=new_obj_content, ContentType='application/json')
 
+        # The editor release notes (the update dialog's source) must land before
+        # update-v4.json points users at the new sha1.
+        # DEV-ONLY (issue-7186 validation): 'release-notes-view' requires notes so
+        # the disposable channel exercises the full path. Remove before merge.
+        releasenotes.upload(self, bucket, required = self.channel in ('beta', 'stable', 'release-notes-view'))
```
vlaaad: From codex (use your judgement): missing required notes are known before any upload is necessary, yet info.json is updated first. Failure can leave channel metadata split between old/new release, and we can avoid it by performing this check before any s3 writes.

scripts/releasenotes_github_projectv2.py
```diff
@@ -531,18 +636,37 @@ def check_issue_commits(issues):
                 red("    Missing from beta")
                 missing_beta_count = missing_beta_count + 1
 
+        # A fix must be present on every branch this channel ships from to be
+        # listed, otherwise the notes would advertise a change that isn't in the
+        # release. The dev/beta tallies above are diagnostics only.
+        branch_ok = {"dev": dev_ok, "beta": beta_ok}
+        if all(branch_ok.get(b, False) for b in required_branches):
```
vlaaad: From codex (use your judegement): if PR in another repo fixes an issue in defold/defold, it seems it will be skipped here.

editor/src/clj/editor/ui/updater.clj
```diff
@@ -45,7 +48,52 @@
...
+             :content content
+             :project project
+             :stylesheets [(str (io/resource "editor.css"))]
+             :root-props {:style-class "md-page-root"}}
+   :footer {:fx/type dialogs/dialog-buttons
+            :children [{:fx/type fxui/legacy-button
+                        :text (localization (localization/message "updater.release-notes-dialog.button.later"))
+                        :cancel-button true
+                        :on-action {:result false}}
+                       {:fx/type fxui/legacy-button
```
vlaaad, line 78: I think this button should also be:
```clj
 :variant :primary
 :default-button true
```

vlaaad (changes requested):
Question.. I don't quite understand our approach now. Did we decide to do release notes on every push? I think this was my suggestion, but then we discussed it, and I remember we also talked about doing it manually since the release notes can change when we edit PR descriptions, so there might be no push at all when we want to udpate the release notes.


## Dialog content
Image

1: should we put version into header instead of uninformative "What's new"?
2. should we group  issues into lists with headers, so we can see what belongs to engine, to editor, and to other? like in the forum, where we have Engine and Editor sections?
3. Should we show more info, like in the forum? Not just the headline, but an explanation (body) too? Perhaps using expandable view (see our support of &lt;details> and &lt;summary> tags)

scripts/releasenotes_github_projectv2.py
JosephFerano: The reason is that `get_issue` now returns None instead of exiting the script. So this gave the function a chance to find other potential issues being closed. However, upon further analyzing this code, I feel like it's tolerating bad issue numbers and perhaps we should fail the build, as we did before, loudly, if we linked to an issue that does not exist in the org.

editor/src/clj/editor/boot_open_project.clj
JosephFerano: The intention is to show the user the release notes, even if it's an older version. The reasoning is that the purpose of the bundled release notes is to enhance user discovery of features, so even if it's an older version, they would be able to see what was added in that version, and potentially discover new functionality.

Although admittedly, this was added when the markdown was shown as a tab, slightly less nice now that it's a dialog window.

JosephFerano:
> Did we decide to do release notes on every push? I think this was my suggestion, but then we discussed it, and I remember we also talked about doing it manually since the release notes can change when we edit PR descriptions, so there might be no push at all when we want to update the release notes.

I was originally bundling/publishing notes only for beta/stable (for now), but I didn't see a reason to not do the same for alpha (dev). For manual deployment, I had removed a command from build.py that allows manual S3 upload, I should've added to releasenotes.py. It's there now.

> 1: should we put version into header instead of uninformative "What's new"?

Yes we can, I guess I didn't include it for the new release notes since it would include the notes from many versions,
not just the newest. However, for the bundled release notes, it can be added.

> 2. should we group issues into lists with headers, so we can see what belongs to engine, to editor, and to other? like in the forum, where we have Engine and Editor sections?

The purpose of this was to keep these notes simple, while the current release notes would show the full descriptions,
but with your next suggestion...

> 3. Should we show more info, like in the forum? Not just the headline, but an explanation (body) too? Perhaps using expandable view (see our support of <details> and <summary> tags)

I think if we use an expandable view, then yes it makes sense. The reason I wanted to avoid adding more info was because
we're not just showing the newest releae notes but all versions in between. But if we can do proper information hiding,
then we can include everything then. I'll investigate.

editor/src/clj/editor/updater.clj
JosephFerano: Currently differing number of dots will not sort correctly;
```clojure
  (def versions
    ["1.9.0" "1.10.0" "1.13.2" "1.13" "1.13.10" "1.13.9" "1.013.0" "1.13.2-beta" "banana"])

  (sort-by parse-version coll/descending-order (filter parse-version versions))
  ;; => ("1.13.10"
  ;;     "1.13.9"
  ;;     "1.13.2"
  ;;     "1.013.0"
  ;;     "1.10.0"
  ;;     "1.9.0"
  ;;     "1.2.1"
  ;;    "1.13")  <-- Wrong

  (sort #(.compare util/natural-order %2 %1) versions)
  ;; => ("banana"
  ;;     "1.13.10"
  ;;     "1.13.9"
  ;;     "1.13.2-beta" <-- But this one could be wrong
  ;;     "1.13.2"
  ;;     "1.013.0"
  ;;     "1.13"  <-- Got it right
  ;;     "1.10.0"
  ;;     "1.9.0"
  ;;    "1.2.1")
  ```
`util/natural-order` would handle differing dots, but if labels ever got introduced, would not handle them well. Not sure which one is more likely to happen in the future (missing dots or added labels), but we can use a regex to eliminate that possibility and just validate that it's a series of numbers and dots.

File: .github/workflows/main-ci.yml (hunk 20 above)
Comment (JosephFerano) [reply .github/workflows/main-ci.yml:545]: There are guards further along that will fail the build on some kind of error. But if we don't introduce this, then alpha errors would fail as well.

File: editor/src/clj/editor/ui/updater.clj [hunk 26]
```diff
@@ -45,7 +49,61 @@
...
+       :localization localization
+       :content content
+       :versions versions
+       :project project})))
+
+(defn- prompt-and-download! [stage project updater localization download-confirmed?]
+  (if-let [{:keys [markdown versions]} (updater/release-notes updater)]
+    (when (show-release-notes-update-dialog! markdown versions project localization)
+      (updater/download-and-extract! updater))
+    (when (or download-confirmed?
```
vlaaad: boolean with `?` suffix

editor/src/clj/editor/markdown.clj
```diff
@@ -399,14 +401,31 @@
...
+(def ^:private prop-image-width-cap
+  (fx/make-binding-prop
+    (fn [^HBox hbox ^Image image]
+      (.bind (.maxWidthProperty hbox) (.widthProperty image))
+      #(.unbind (.maxWidthProperty hbox)))
+    fx.lifecycle/scalar))
+
+(def ^:private ext-with-image-width-cap
+  (fx/make-ext-with-props
+    {::image-width-cap prop-image-width-cap}))
```
vlaaad: Unnecessary indirection, prop objects can be used directly.

scripts/releasenotes_github_projectv2.py
```diff
@@ -472,8 +536,19 @@ def parse_github_project(version):
...
-    print("\nChecking issue commits for dev and beta presence")
+def audit_branches_for_channel(channel):
+    # Which branches a note's fix has to be on, per release channel.
+    # beta/stable ship from beta with everything merged forward -> require both.
+    # alpha (dev track) and custom validation channels build off dev / a feature
+    # branch, so requiring beta would fail them for fixes that just haven't
+    # reached beta yet -> require dev only. Unknown channel (manual run with no
+    # --channel) keeps the strict default.
+    if channel in ("beta", "stable", None):
+        return ["dev", "beta"]
```
vlaaad: Not sure this was addressed? I think we don't need to require a commit to be in dev when the channel is beta?

vlaaad:
The code looks good; I found only a couple of minor nitpicks. 

I have 2 questions/concerns:
1. When checking for updates, we download json and display it as expandable items. At the same time, we bundle a markdown file with a different formatting and display it as a big, long document. So, the same release notes are displayed differently if they are bundled or fetched. It would be nice to make it consistent. Maybe bundle a json instead, and use the same json->markdown pipeline?
2. I still don't quite get it, but why do we generate these notes on push? Their source of truth is PR descriptions that update without commits/pushes, so why do it on push? But maybe it's fine? e.g., when we do a manual pass through PRs when starting a beta, we currently run a release notes gen locally, and then create a commit/push... Are any workflows changing here?

JosephFerano:
> 1. When checking for updates, we download json and display it as expandable items. At the same time, we bundle a markdown file with a different formatting and display it as a big, long document. So, the same release notes are displayed differently if they are bundled or fetched. It would be nice to make it consistent. Maybe bundle a json instead, and use the same json->markdown pipeline?

Yes, given this new design change, it does make sense to now switch to json since this interface is nicer. I'll make the change. The markdown will likely still be uploaded to S3, for the reasons I'll explain below.

> 2. I still don't quite get it, but why do we generate these notes on push? Their source of truth is PR descriptions that update without commits/pushes, so why do it on push? But maybe it's fine? e.g., when we do a manual pass through PRs when starting a beta, we currently run a release notes gen locally, and then create a commit/push... 

The upside of automating this on push is for several reasons;

- It would be tedious for the team to have to update release notes manually for alpha releases. Beta less tedious, but still requires manual intervention that could be otherwise automated.
- By doing it in the CI, we can bundle the notes. As a preference, I think it's nicer to avoid hitting the network to get something we can package in the binary.
- We also automate the S3 upload for the "new" release notes fetch.

> Are any workflows changing here?

From what I understood from my discussion with Mathias, we don't necessarily want to keep doing this manually and we were open to stop checking the release notes into the repo. That they are checked in seems to be more of a "no better place to put them" rather than a preference. Hopefully I'm not misremembering this part of the conversation. If they live on S3 instead of the repo, that gives us a few more options, like perhaps a website build can list the contents of the buckets somewhere. When it comes time to grab the markdown for the forum, it can be grabbed from S3 similar to how we grab dev editor builds. Perhaps a backup of the buckets every once in a while will suffice to ensure they don't get lost.

editor/src/clj/editor/app_view.clj
```diff
@@ -2586,7 +2608,17 @@
        (perform-open-resource-plan! localization))))
 
 (def ^:private release-notes-resource-delay
-  (delay (io/resource (str "release-notes/" (system/defold-version) ".md"))))
+  (delay (io/resource (str "release-notes/" (system/defold-version) ".json"))))
+
+(defn- bundled-release-notes-markdown [url]
```
vlaaad: > `markdown`

It's json now, right?

build_tools/releasenotes.py
```diff
@@ -70,22 +67,41 @@ def update_manifest(bucket, version, channel):
     obj.put(Body=json.dumps(versions), ContentType='application/json')
 
 
+# Channels that also publish the human-readable notes. Alpha releases on every
+# push, and nobody reads its markdown, so it only ships what the editor needs.
```
vlaaad: I think, for simplicity, we can always publish both md and json?

vlaaad (dismissed):
I think it's good! Great work! Don't forget to remove dev-only code before merging, and maybe rename a function (see comment) — no need to re-request a review for that.

vlaaad (approved):
🚀 🚀 🚀

PR #12632, opened 2026-06-22
Beta build fixes from vk quality merge

Jhonnyg:


build_tools/clang_compilation_database.py
```diff
@@ -5,7 +5,11 @@
 # Based on https://github.com/weka-io/waf/blob/8ac4646d0966fe22f8ae8955c7e9b9d56ae8288e/waflib/extras/clang_compilation_database.py
 # Original is BSD 3-Clause License
 
-import sys, os, json, shlex, pipes
+import sys, os, json
+try:
+	from pipes import quote
+except ImportError:
+	from shlex import quote
```
Jhonnyg, line 12: Pipes doesn't seem to be available in 3.14, but it's still used in dev. Not sure if we need this guardrail or not?

engine/engine/CMakeLists.txt
```diff
@@ -1,644 +0,0 @@
-cmake_minimum_required(VERSION 4.0)
```
Jhonnyg, line 1: This file was merge in from the vkquality branch, but beta doesn't use cmake through-and-through yet.

JCash (approved):

PR #12633, opened 2026-06-22
Fix Lua timer self-cancel crashes

### LINKED ISSUE #12436: Intermittent macOS runtime crash in LuaTimerCallback / lua_settop

The Defold macOS runtime crashes intermittently with a native `EXC_BAD_ACCESS` / `SIGSEGV`.

I don’t have a minimal reproduction project yet because the crash appears random in a larger project. I have not been able to reproduce the same crash in the HTML5/Web build of the same project.

The crash log points to `lua_settop` called from `dmScript::LuaTimerCallback`, so this may be related to timer callback lifetime, script unloading, or collection proxy unloading/updating.

**To Reproduce (REQUIRED)**

I don’t have reliable minimal reproduction steps yet.

Current observed behavior:

1. Run the project from the Defold editor on macOS.
2. Play for some time / navigate through the game.
3. At some point, the macOS runtime crashes.
4. The same project has not reproduced this crash in the HTML5/Web build so far.

I understand that this is not ideal for debugging, but the crash is native and happens before any useful Lua error is reported.

The engine should not crash natively.

If there is an invalid Lua callback, an expired timer callback, or an issue caused by unloaded scripts/collections, I would expect either a Lua error, safe callback cancellation, or a clear engine error message instead of a native crash.

  - Version: 1.12.4

**Platforms (REQUIRED):**
 - Platforms: macOS
 - OS: macOS 26.5 (25F71)
 - Device: MacBook Pro, Apple M4 Pro
 - CPU architecture: ARM64 / Apple Silicon

Also tested:
 - Platforms: HTML5/Web
 - Result: unable to reproduce the crash so far

No minimal repro project yet.

The issue currently happens intermittently in a larger project, and I have not been able to isolate the exact timer/script/collection involved.

**Logs (OPTIONAL):**

I can attach:

- Full macOS crash report
[mac os error.txt](https://github.com/user-attachments/files/28153139/mac.os.error.txt)

- Defold `_crash` output (I had to add a .txt, because github wouldn't let me attach a file named just “_crash”)
[_crash.txt](https://github.com/user-attachments/files/28153037/_crash.txt)

Key macOS crash summary:

```text
Exception Type:    EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS
Termination Reason: Namespace SIGNAL, Code 11, Segmentation fault: 11

Thread 0 Crashed:: engine_main Dispatch queue: com.apple.main-thread
0   dmengine   lua_settop + 112
1   dmengine   dmScript::LuaTimerCallback(...) + 92 (script_timer.cpp:530)
2   dmengine   dmScript::UpdateTimers(...) + 248 (script_timer.cpp:253)
3   dmengine   dmScript::UpdateScriptWorld(...) + 88
4   dmengine   dmGameObject::CompScriptUpdate(...)
...
8   dmengine   dmGameSystem::CompCollectionProxyUpdate(...)
```

Defold console output:

```text
--Any printouts from the game, no errors
INFO:CRASH: Successfully wrote Crashdump to file: /Users/nanaminer/Library/Application Support/Defold/_crash
ERROR:CRASH: CALL STACK:

ERROR:CRASH: 0   dmengine                            0x0000000102d5ed18 _ZN7dmCrash7OnCrashEi + 108
ERROR:CRASH: 1   libsystem_platform.dylib            0x0000000182a37744 _sigtramp + 56
ERROR:CRASH: 2   dmengine                            0x0000000102c59180 _ZN8dmScript14InvokeCallbackEPNS_15LuaCallbackInfoEPFvP9lua_StatePvES4_ + 272
ERROR:CRASH: 3   dmengine                            0x0000000102c64968 _ZN8dmScriptL16LuaTimerCallbackEPNS_10TimerWorldENS_14TimerEventTypeEjfmm + 92
ERROR:CRASH: 4   dmengine                            0x0000000102c63654 _ZN8dmScript12UpdateTimersEPNS_10TimerWorldEf + 248
ERROR:CRASH: 5   dmengine                            0x0000000102c57cac _ZN8dmScript17UpdateScriptWorldEPNS_11ScriptWorldEf + 88
ERROR:CRASH: 6   dmengine                            0x00000001028a32d8 _ZN12dmGameObject16CompScriptUpdateERKNS_22ComponentsUpdateParamsERNS_22ComponentsUpdateResultE + 40
ERROR:CRASH: 7   dmengine                            0x00000001028af780 _ZN12dmGameObjectL23UpdateComponentFunctionEPNS_10CollectionEjNS_18UpdateFunctionTypeERNS_22ComponentsUpdateParamsE + 224
ERROR:CRASH: 8   dmengine                            0x00000001028ac178 _ZN12dmGameObject6UpdateEPNS_16CollectionHandleEPKNS_13UpdateContextE + 516
ERROR:CRASH: 9   dmengine                            0x00000001028da654 _ZN12dmGameSystem25CompCollectionProxyUpdateERKN12dmGameObject22ComponentsUpdateParamsERNS0_22ComponentsUpdateResultE + 448
ERROR:CRASH: 10  dmengine                            0x00000001028af780 _ZN12dmGameObjectL23UpdateComponentFunctionEPNS_10CollectionEjNS_18UpdateFunctionTypeERNS_22ComponentsUpdateParamsE + 224
ERROR:CRASH: 11  dmengine                            0x00000001028ac178 _ZN12dmGameObject6UpdateEPNS_16CollectionHandleEPKNS_13UpdateContextE + 516
ERROR:CRASH: 12  dmengine                            0x000000010284d540 _ZN8dmEngine4StepEPNS_6EngineE + 1456
ERROR:CRASH: 13  dmengine                            0x000000010284db80 _Z14dmEngineUpdatePN8dmEngine6EngineE + 28
ERROR:CRASH: 14  dmengine                            0x000000010284e7a8 _ZN8dmEngine7RunLoopEPKNS_13RunLoopParamsE + 92
ERROR:CRASH: 15  dmengine                            0x000000010284e734 _Z11engine_mainiPPc + 104
ERROR:CRASH: 16  dyld                                0x000000018266fe00 start + 6992
```

Additional context (OPTIONAL):

The crash seems intermittent/random and only reproduced on macOS so far.

The stack suggests the crash happens while processing a Lua timer callback during script world update, with collection_proxy also present in the stack.

Any advice on how to identify the specific Lua timer callback involved would be helpful. I can test a debug build or add additional logging if you can point me to the right place.

AGulev:
Lua timers can now safely cancel or restart themselves from inside their own callbacks without freeing callback state while it is still running. This prevents crashes in timer-heavy callback flows and keeps cancelled timers immediately invisible to `timer.get_info()`, `timer.trigger()`, and repeated cancellation attempts.

Fix https://github.com/defold/defold/issues/12436

### Technical changes

- Tracks Lua timer callback invocation depth in `TimerWorld` and defers freeing dead timers until callbacks and timer updates have unwound.
- Adds shared deferred cleanup for dead timers and version updates when cleanup happens outside the normal update pass.
- Wraps both scheduled timer callbacks and `timer.trigger()` callbacks with begin/end callback guards.
- Treats timers marked not alive as invalid in `timer.trigger()` and `timer.get_info()`.
- Adds regression coverage for repeating and one-shot timers that cancel/restart inside callbacks, `timer.trigger()` cancellation, normal cancellation outside update, and script finalization killing live timers.

### DIFF
--- a/engine/script/src/script_timer.cpp
+++ b/engine/script/src/script_timer.cpp
@@ -88,8 +88,10 @@ namespace dmScript
         dmSet<uint16_t>     m_Instances; // the pool indices
         dmArray<uint16_t>   m_ScratchBuffer; // When doing operations on the timer instances, we need a copy to avoid self modification of m_Instances
         uint16_t            m_Version;   // Incremented to avoid collisions each time we push timer indexes back to the m_IndexPool
+        uint16_t            m_CallbackInvocationDepth;
         uint16_t            m_InUpdate : 1;
         uint16_t            m_IsDirty : 1;
+        uint16_t            m_HasDeferredDeadTimers : 1;
     };
 
     dmArray<TimerWorld*> g_Worlds;
@@ -167,14 +169,17 @@ namespace dmScript
         timer_world->m_Instances.SetCapacity(INITIAL_TIMER_CAPACITY);
 
         timer_world->m_Version = 0;
+        timer_world->m_CallbackInvocationDepth = 0;
         timer_world->m_InUpdate = 0;
         timer_world->m_IsDirty = 0;
+        timer_world->m_HasDeferredDeadTimers = 0;
         return timer_world;
     }
 
     void DeleteTimerWorld(HTimerWorld timer_world)
     {
         assert(timer_world->m_InUpdate == 0);
+        assert(timer_world->m_CallbackInvocationDepth == 0);
         delete timer_world;
     }
 
@@ -212,6 +217,46 @@ namespace dmScript
             tgt[i] = src[i];
         }
         return size;
+    }
+
+    static uint32_t FreeDeadTimers(HTimerWorld timer_world)
+    {
+        assert(timer_world != 0x0);
+
+        uint32_t size = CopyIndices(timer_world->m_Instances, timer_world->m_ScratchBuffer);
+        uint32_t free_count = 0;
+        for (uint32_t i = 0; i < size; ++i)
+        {
+            Timer* timer = GetTimerFromIndex(timer_world, timer_world->m_ScratchBuffer[i]);
+            if (timer && timer->m_IsAlive == 0)
+            {
+                FreeTimer(timer_world, timer);
+                ++free_count;
+            }
+        }
+        return free_count;
+    }
+
+    static void BeginLuaTimerCallback(HTimerWorld timer_world)
+    {
+        assert(timer_world != 0x0);
+        ++timer_world->m_CallbackInvocationDepth;
+    }
+
+    static void EndLuaTimerCallback(HTimerWorld timer_world)
+    {
+        assert(timer_world != 0x0);
+        assert(timer_world->m_CallbackInvocationDepth > 0);
+        --timer_world->m_CallbackInvocationDepth;
+
+        if (timer_world->m_CallbackInvocationDepth == 0 && timer_world->m_InUpdate == 0 && timer_world->m_HasDeferredDeadTimers)
+        {
+            timer_world->m_HasDeferredDeadTimers = 0;
+            if (FreeDeadTimers(timer_world) > 0)
+            {
+                ++timer_world->m_Version;
+            }
+        }
     }
 
     void UpdateTimers(HTimerWorld timer_world, float dt)
@@ -301,6 +346,12 @@ namespace dmScript
             }
         }
 
+        if (timer_world->m_HasDeferredDeadTimers)
+        {
+            timer_world->m_HasDeferredDeadTimers = 0;
+            FreeDeadTimers(timer_world);
+        }
+
         if (timer_world->m_IsDirty)
         {
             ++timer_world->m_Version;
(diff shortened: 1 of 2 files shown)


engine/script/src/script_timer.cpp
```diff
@@ -88,8 +88,10 @@ namespace dmScript
         dmSet<uint16_t>     m_Instances; // the pool indices
         dmArray<uint16_t>   m_ScratchBuffer; // When doing operations on the timer instances, we need a copy to avoid self modification of m_Instances
         uint16_t            m_Version;   // Incremented to avoid collisions each time we push timer indexes back to the m_IndexPool
+        uint16_t            m_CallbackInvocationDepth;
```
JCash, line 91: Why is the "depth" needed? Isn't it enough to know if we're in the call or not?

JCash (approved):
lgtm

engine/script/src/script_timer.cpp
AGulev: it's needed because we have `timer.trigger()` which can be called from a callback

Issue #12608, opened 2026-06-16
Metal graphics adapter support MVP2


Jhonnyg:
**Task (REQUIRED):**
* Add manifest options to select the metal adapter for iOS & OSX separately
* Benchmark on iOS + make sure it's production ready for iOS
* Make the metal adapter default on OSX + iOS

**Expected outcome (REQUIRED):**
Describe the expected outcome of the task (new issue, documentation, code change etc)

selimanac:
First run. 🙂
Probably expected, or maybe you already know about this.
Should I open a new issue?

Edit: It works as long as I only have sprites. 🤘 It turns out this happens when I have some models in the collection. I’ll look into it. I only have a few very simple shaders, so they might be causing this.

```
Thread 0 Crashed:: engine_main Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x18c56e5e8 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x18c5a98d8 pthread_kill + 296
2   libsystem_c.dylib             	       0x18c4b0644 abort + 148
3   libsystem_c.dylib             	       0x18c4af8a0 __assert_rtn + 284
4   dmengine                      	       0x100ec5028 dmGraphics::DrawSetup(dmGraphics::MetalContext*) + 3936 (graphics_metal.mm:2922)
5   dmengine                      	       0x100ebe5cc dmGraphics::MetalDrawElements(dmGraphics::GraphicsContext*, dmGraphics::PrimitiveType, unsigned int, unsigned int, dmGraphics::Type, unsigned long, unsigned int) + 112 (graphics_metal.mm:3039)
6   dmengine                      	       0x100f9d1ec dmRender::Draw(dmRender::RenderContext*, dmRender::Predicate*, dmRender::NamedConstantBuffer*) + 1176 (render.cpp:1270)
7   dmengine                      	       0x100f9ccd0 dmRender::DrawRenderList(dmRender::RenderContext*, dmRender::Predicate*, dmRender::NamedConstantBuffer*, dmRender::FrustumOptions const*, dmRender::SortOrder) + 3312 (render.cpp:1099)
8   dmengine                      	       0x100f9fe38 dmRender::ParseCommands(dmRender::RenderContext*, dmRender::Command*, unsigned int) + 488 (render_command.cpp:197)
[...]

PR #12639, opened 2026-06-23
Make the Metal graphics adapter selectable in the app manifest

Jhonnyg:
The Metal graphics adapter is now selectable in the app manifest tool for OSX and iOS.

Part of #12608


Jhonnyg:
@codex review

editor/src/clj/editor/app_manifest.clj
```diff
@@ -660,20 +664,152 @@
...
+   [:both "OpenGL & Vulkan"]])
+
+(def ^:private apple-adapter-ids [:open-gl :metal :vulkan])
+
+(def ^:private apple-graphics-choice-selections
+  [[:open-gl #{:open-gl}]
+   [:metal #{:metal}]
+   [:vulkan #{:vulkan}]
+   [:open-gl-metal #{:open-gl :metal}]
+   [:both #{:open-gl :vulkan}]])
```
Jhonnyg: Question: should we remove vulkan support for osx/ios? or keep it until we are more confident about the metal adapter? it's barely been used for ios, but it's actively used for osx

engine/engine/CMakeLists.txt
```diff
@@ -137,7 +137,7 @@ endif()
 if("GraphicsAdapterDX12" IN_LIST ENGINE_GRAPHICS_SYMBOLS)
   list(APPEND ENGINE_SHADER_OUTPUT_ARGS --shader-output hlsl)
 endif()
-if("GraphicsAdapterMetal" IN_LIST ENGINE_GRAPHICS_SYMBOLS)
+if("GraphicsAdapterMetal" IN_LIST ENGINE_GRAPHICS_SYMBOLS OR TARGET_PLATFORM MATCHES "^(x86_64-macos|arm64-macos|x86_64-ios|arm64-ios)$")
```
Jhonnyg, line 140: We need to compile msl shaders as well for apple platforms since internal shaders will be embedded into the binary (e.g debug font) and at this stage we don't know which adapters will be linked so we must compile all variants "just in case". I think we should do it in a different way and separate these embedded resources from the engine library (which i think is the issue) in some way.

editor/src/clj/editor/app_manifest.clj
```diff
@@ -660,20 +664,152 @@
...
+     :metal (concat legacy-metal-osx-toggles exclude-open-gl-osx-toggles exclude-platform-osx-toggles exclude-vulkan-osx-toggles)
+     :open-gl (concat open-gl-osx-toggles exclude-vulkan-osx-toggles)
+     :both (concat open-gl-osx-toggles explicit-vulkan-osx-toggles exclude-metal-osx-toggles)
+     :both (concat open-gl-osx-toggles explicit-vulkan-osx-toggles)
+     :both (concat open-gl-osx-toggles exclude-metal-osx-toggles)
+     :both open-gl-osx-toggles
+     :vulkan (concat explicit-vulkan-osx-toggles exclude-open-gl-osx-toggles exclude-platform-osx-toggles)
+     :vulkan (concat explicit-vulkan-osx-toggles exclude-open-gl-osx-toggles)
+     :vulkan (concat explicit-vulkan-osx-toggles exclude-platform-osx-toggles)
+     :vulkan explicit-vulkan-osx-toggles]))
```
Jhonnyg: This block will make sure the editor for manifests will show the correct adapter even for older manifest files since the shape of the variants have changed now. Do we want this or not? I imagine that we will remove Vulkan / MoltenVK once Metal is default, so maybe we don't need this?

editor/src/clj/editor/app_manifest.clj
```diff
@@ -660,20 +664,144 @@
...
+   [:open-gl-metal #{:open-gl :metal}]
+   [:both #{:open-gl :vulkan}]])
+
+(defn- adapter-selection-toggles [adapter-toggles exclude-toggles selection]
+  (let [selected? (set selection)]
+    (concat
+      (mapcat adapter-toggles selection)
+      (mapcat exclude-toggles (remove selected? apple-adapter-ids)))))
+
+(defn- make-adapter-selection-setting [default adapter-toggles exclude-toggles compatibility-id+toggles]
```
vlaaad: Function name implies it's generic, but it specifically refers to `apple-graphics-choice-selections`. Should either be generic or have an Apple-specific name.

editor/src/clj/editor/app_manifest.clj
```diff
@@ -660,20 +664,144 @@
...
+     :vulkan exclude-vulkan-osx-toggles}
+    [:open-gl (concat open-gl-osx-toggles exclude-vulkan-osx-toggles)
+     :both (concat open-gl-osx-toggles explicit-vulkan-osx-toggles exclude-metal-osx-toggles)
+     :both (concat open-gl-osx-toggles explicit-vulkan-osx-toggles)
+     :both (concat open-gl-osx-toggles exclude-metal-osx-toggles)
+     :both open-gl-osx-toggles
+     :vulkan (concat explicit-vulkan-osx-toggles exclude-open-gl-osx-toggles exclude-platform-osx-toggles)
+     :vulkan (concat explicit-vulkan-osx-toggles exclude-open-gl-osx-toggles)
+     :vulkan (concat explicit-vulkan-osx-toggles exclude-platform-osx-toggles)
+     :vulkan explicit-vulkan-osx-toggles]))
```
vlaaad: Do the old variants still make sense in their old shape? We haven't done "compatibility" to preserve old states as selected check boxes so far. I'm not sure this "compatibility" is useful, especially if the old, manually specified, state no longer represents a valid selection of libs and exclusions.

editor/src/clj/editor/app_manifest.clj
```diff
@@ -660,20 +664,144 @@
...
+
+(def ^:private apple-graphics-choice-selections
+  [[:open-gl #{:open-gl}]
+   [:metal #{:metal}]
+   [:vulkan #{:vulkan}]
+   [:open-gl-metal #{:open-gl :metal}]
+   [:both #{:open-gl :vulkan}]])
+
+(defn- adapter-selection-toggles [adapter-toggles exclude-toggles selection]
+  (let [selected? (set selection)]
```
vlaaad: Selection is already a set, as I understand (from `apple-graphics-choice-selections`)

editor/src/clj/editor/app_manifest.clj
```diff
@@ -660,20 +664,144 @@
...
+      (mapcat exclude-toggles (remove selected? apple-adapter-ids)))))
+
+(defn- make-adapter-selection-setting [default adapter-toggles exclude-toggles compatibility-id+toggles]
+  (apply make-choice-setting
+         (concat
+           (mapcat (fn [[choice-id selection]]
+                     [choice-id (adapter-selection-toggles adapter-toggles exclude-toggles selection)])
+                   apple-graphics-choice-selections)
+           compatibility-id+toggles
+           [default])))
```
vlaaad: I think this function should just not exist at all. Like, it's going to be simpler if we just use `make-choice-setting` directly.

editor/src/clj/editor/app_manifest.clj
```diff
@@ -660,20 +664,144 @@
...
+         (concat
+           [:open-gl (concat exclude-metal-ios-toggles exclude-vulkan-ios-toggles)
+            :metal (concat metal-ios-toggles exclude-open-gl-metal-ios-toggles exclude-vulkan-ios-toggles)
+            :vulkan (concat explicit-vulkan-ios-toggles exclude-open-gl-vulkan-ios-toggles exclude-metal-ios-toggles)
+            :open-gl-metal (concat open-gl-ios-toggles metal-ios-toggles exclude-vulkan-ios-toggles)
+            :both (concat open-gl-ios-toggles explicit-vulkan-ios-toggles exclude-metal-ios-toggles)]
+           [;; Compatibility with older manifests where iOS Vulkan came from the generic graphics selector.
+            :vulkan (concat explicit-vulkan-ios-toggles exclude-open-gl-vulkan-ios-toggles)
+            :vulkan explicit-vulkan-ios-toggles]
+           [:open-gl])))
```
vlaaad: Does it make sense to introduce compatibility now when we have never done it before? Is it still useful? Should it be then, at least, called something like "vulkan pre 1.13.1" or something?

vlaaad (changes requested):
Could be simplified

editor/src/clj/editor/app_manifest.clj
vlaaad: Also, don't use `apply` here, there is just no point in doing it. Just call `make-choice-setting` directly.

Jhonnyg: Yeah that was my question, I was also leaning towards skipping the compatibility code but wanted a second pair of eyes on that part. I decided to remove the compat check.

editor/src/clj/editor/app_manifest.clj
```diff
@@ -660,21 +664,99 @@
     :vulkan vulkan-android-toggles
     :both))
 
+(def apple-graphics-choice-options
+  [[:open-gl "OpenGL"]
+   [:metal "Metal"]
+   [:vulkan "Vulkan"]
+   [:open-gl-metal "OpenGL & Metal"]
+   [:both "OpenGL & Vulkan"]])
```
vlaaad: Maybe call it `:open-gl-vulkan`? It's strange to name it "both" when we have 3 options, and "both" means 2 out of 3.

vlaaad (dismissed):
Editor changes look good 👍

editor/src/clj/editor/app_manifest.clj
Jhonnyg: Right, yes that makes sense. Thanks!

JCash (dismissed):

PR #12640, opened 2026-06-23
Fix atlas packing regression

### LINKED ISSUE #12593: Atlas packing regression: images are packed very inefficiently, producing unnecessarily large atlases
## Description

Defold’s automatic atlas packer appears to have regressed at some point.

In older Defold versions, atlas packing worked well and produced compact atlas layouts. In recent versions, the packer can generate very inefficient layouts with a large amount of empty space.

I do not know the exact version where this changed, but the current behaviour looks like a regression.

## Actual behaviour

The atlas packer sometimes produces atlases that are much larger than necessary.

For example, even with only two rectangular images in an atlas, Defold may create an atlas with roughly twice the area actually needed to fit those images.

There are also cases where removing images from an atlas can unexpectedly increase the final atlas size instead of reducing it.

I have attached two short animations showing the issue visually:

https://github.com/user-attachments/assets/a2d7a7c7-2554-4176-8efa-2c83cdca5fee

https://github.com/user-attachments/assets/f8ca9e83-7cac-4a3e-ab3d-427ec890be32

## Expected behaviour

The atlas packer should place images efficiently and choose the smallest reasonable atlas size that can contain all source images, taking padding and other atlas settings into account.

Removing images from an atlas should not normally increase the final atlas area unless there is a clear technical reason.

## Environment

- Platform: Windows
- Defold versions tested: 1.12.4, 1.13.0

JosephFerano:
A regression was fixed that would not allow certain images to be rotated, requiring extra space. Also improved the packing algorithm with extra attempts which helped reduce atlas sizes under certain cases with extrusion settings.

Fixes #12593

### Technical changes

Added an occupancy getter for layouts.

### DIFF
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/TextureSetLayout.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/TextureSetLayout.java
@@ -261,6 +261,13 @@ public class TextureSetLayout {
         public int getHeight() {
             return height;
         }
+        public float getOccupancy() {
+            long used = 0;
+            for (Rect r : rectangles) {
+                used += (long) r.getWidth() * r.getHeight();
+            }
+            return (float) used / ((long) width * height);
+        }
 
         @Override
         public String toString() {
--- a/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/TextureSetLayout.java
+++ b/com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/TextureSetLayout.java
@@ -340,14 +340,9 @@ public class TextureSetLayout {
      * @param rotate
      * @return
      */
-    // Seed orderings for the greedy MaxRects packer. The packer is a greedy
-    // heuristic and is non-monotonic: a single fixed sort lands on a good packing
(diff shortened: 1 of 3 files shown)


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/textureset/TextureSetLayout.java
```diff
@@ -340,21 +340,124 @@ public static Layout gridLayout(int margin, List<Rect> rectangles, Grid gridSize
      * @param rotate
      * @return
      */
+    // Seed orderings for the greedy MaxRects packer. The packer is a greedy
+    // heuristic and is non-monotonic: a single fixed sort lands on a good packing
+    // for some inputs and a needlessly large one for others (issue #12593, e.g. the
+    // same image set packs to 2048x2048 with one ordering but 4096x2048 with
+    // another). We pack with area-descending first - the historical order, kept as
+    // the fast path so existing layouts are byte-for-byte unchanged - and only when
```
JCash: Since we don't store the layout on disc, "so existing layouts are byte-for-byte unchanged" isn't exactly true?

JCash (dismissed):

AGulev (dismissed):
I think it is good fix

JCash (approved):

AGulev (approved):
please add a one-liner with user facing changes for release notes

PR #12641, opened 2026-06-23
 Fix glitchy HTML5 sound on Samsung Internet Browser and some other cases

### LINKED ISSUE #10041: Glitchy audio at low FPS when played in a single thread
On single-threaded platforms (is it just HTML5?) we hear glitches in the audio if FPS fluctuates or becomes too low (10-15).

We get the same on desktop platforms if we turn off `sound.use_thread` and set update frequency to 30 or lower - sound buffer size becomes short, we hear glitches.

https://github.com/user-attachments/assets/ee026bab-7632-405a-8172-9523525a3dac

**To Reproduce (REQUIRED)**

**- HTML5**
1. Set `display.update_frequency` to 10.
2. Run HTML5 build and play sounds.

**- Windows** (and probably other platforms)
1. Turn off `sound.use_thread`.
2. Set `display.update_frequency` to 30 or lower.
3. Run the build and play sounds.

The audio system should play sounds clearly at any FPS.

  - Version: 1.9.6, 1.9.7

**Platforms (REQUIRED):**
 - Platforms: HTML5 and all other if `sound.use_thread` is disabled.

[test_audio.zip](https://github.com/user-attachments/files/18425243/test_audio.zip)

I think [the 768 byte constant for sound frame count](https://github.com/defold/defold/commit/f59eb246a40e16d3353c30bcc38b5bb5a28c228e) should be calculated dynamically from frame to frame, if it's possible.

This is especially important for HTML5 for mobile, where the framerate can be dropped to low values, especially when loading collections (tip: we mute sounds when doing this!).

Related:
- https://github.com/defold/defold/issues/9445

### LINKED ISSUE #10889: Play sound issues in web browser for sounds with fast attack
Sounds play clearly when app starting in defold ide, while in web browser "sound start" sounds dirty.
I checked ogg, wav, opus - the same problem. Chrome and FireFox last versions - the same.
Not all sounds are affected, looks like sounds with very fast start. Issue isn't 100% repro for every play.
I found workaround: added 20ms silence at the beginning of each sound.

**To Reproduce (REQUIRED)**
In code - simple sound.play()
Start from IDE - clear sound
Start in browser - dirty start

Clear sound on all platforms

  - Version 1.10.2

**Platforms (REQUIRED):**
 - Platforms: Windows, HTML5
 - OS: Windows 10
 - Device: PC
 - Chrome 137.0.7151.104 (64-bit)
 - FireFox 139.0.4 (64-bit)

**Workaround (OPTIONAL):**
20ms of silence for sound at the beginning

[magic1.zip](https://github.com/user-attachments/files/20906483/magic1.zip)

Note (13 Jan 2025):

Looks like 20ms silence solution corresponds to default "Sample Frame Count" which is 1024 for 48000Hz rate.
Recently, we had to change sample frame count to 2048 to avoid another issue (heavy sound play issues on some devices/browsers). Then - problems related to this issue returned.
Increasing silence to 43ms helps.

Looks like sound engine race at the beginning of sound play. Sometimes first update occurs earlier than first "Sample Frame Count" is loaded.

### LINKED ISSUE #10973: Audio/Sound stuttering on low end devices(android) for HTML5 (empty project)
"I create an empty project with with just a small ogg file playing in the background but on low end devices android the sound stutters when looped. I lowered the sample rate."

**To Reproduce (REQUIRED)**
create an audio clip that loops in init function using ogg or wave that is longer than 1 minute. 

`function init(self)
	sound.play("/go#sound")
end`

On higher end devices stuttering doesn't happen but on lower end devices stuttering happens.

  - Version [1.10.3]

**Platforms (REQUIRED):**
Built in html5 but tested on
android 12

[test project on itch](https://gelzo.itch.io/test) password "test"

**Screenshots (OPTIONAL):**

Image

AGulev:
HTML5 builds now keep browser audio playback buffered enough to avoid glitchy music and sound effects on Samsung Internet. The sound queue adapts to browser-reported latency and detected underruns instead of relying on a fixed buffer count, so affected playback has more room to stay continuous.

Fix https://github.com/defold/defold/issues/9445
Fix https://github.com/defold/defold/issues/10041
Fix https://github.com/defold/defold/issues/10889 
Fix https://github.com/defold/defold/issues/10973

### Technical changes

- Adds HTML5 sound device start/idle callbacks from `device_js.cpp` into `library_sound.js`.
- Tracks the number of actually playing sound instances in `sound.cpp`, so Emscripten builds can stop the device when instances remain allocated but none are playing.
- Replaces the fixed WebAudio buffer count with an adaptive queue target that grows on real underruns, uses browser latency as a floor, and decays after stable playback.
- Suppresses queue learning after suspend/resume, idle restarts, and intentional silence so stale gaps are not treated as underruns.
- Avoids an extra audio sample copy for non-shared WASM memory while preserving the `SharedArrayBuffer` copy path required by threaded builds.


JCash (approved):
lgtm

PR #12645, opened 2026-06-24
Added dependency meta data to the bundle

### LINKED ISSUE #11405: Design an extension versioning system
In order to better reproduce debuggable builds, using the same code that was used at the time of release of a game, we'd like to design a way for extensions to include version numbers to the build.

Design:
https://docs.google.com/document/d/14NnEdPw2dn3kSpVYQd4rhLDQel3KIZknQmK-i5NRqCw/edit?usp=sharing

JCash:
We have added support for project dependency info in a crash dump. The information has been added in order to help when debugging issues, and replicating a debuggable build from the dependencies.

The information is not included by default, but can be included by setting the game.project setting `project.dependencies_metadata = 1`.

The information contains:
* "url" - anonymized url (e.g "https://github.com/defold/extension-spine/archive/refs/tags/4.6.0.zip")
* "commit-sha1 - the commit sha1 of the .zip (e.g. "da782a14b149238d7d30527c6816254cac1b829a") _if available!_
* "payload-sha1" - the sha1 checksum of the .zip itself (e.g. "921C7AE8158669E4E6DE37CD0597529A")

_Note_: The commit sha1 is read from the metadata of the .zip file, as GitHub adds the sha1 there. This is not necessarily true for all dependencies, e.g if they are served from other hosts.

You can also use `sys.load_resource("/.internal/dependencies.json")` at runtime in case you need to store your own bread crumbs.

Fixes https://github.com/defold/defold/issues/11405

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


AGulev:
We always exclude this data by default because it may contain sensitive information that developers don't want to include in the build. I think it shouldn't be enabled by default.

ekharkunov:
It would be nice if we can provide user a way to define URL patterns because it can be the case when extensions aren't downloaded from Github. It also can be downloaded from Gitlab or even from custom storage

JCash:
> It would be nice if we can provide user a way to define URL patterns because it can be the case when extensions aren't downloaded from Github. It also can be downloaded from Gitlab or even from custom storage

@ekharkunov what do you mean? It's still a url (anonymized) even if it's form GitLab?

ekharkunov:
> > It would be nice if we can provide user a way to define URL patterns because it can be the case when extensions aren't downloaded from Github. It also can be downloaded from Gitlab or even from custom storage
> 
> @ekharkunov what do you mean? It's still a url (anonymized) even if it's form GitLab?

Let me show some examples:
Gitlab:
```
https://<gitlab_host>/defold-extensions/lang/-/archive/master/lang-master.zip

https://<gitlab_host>/defold-extensions/lang/-/archive/bbc5e71ccef607a9c216edec17e5deed0d909d00/lang-bbc5e71ccef607a9c216edec17e5deed0d909d00.zip
```
Also I use custom endpoint to download package to download prepackaged extensions from Gitlab package registry or S3 bucket behind reverse proxy. Example:
```
https://defold-packages.<host>/nakama.0c593d29.zip
```
My initial question was about providing place where I as a developer can define url pattern to extract extension name and extension name. By default we can provide Github-like patterns.

AGulev (dismissed):
Looks good, bur please change PR description since we don't turn on this option by default anymore

JCash:
> define url pattern to extract extension name

I think think can be solved by the user in the Lua script, if they want to store the information differently.
Imo, the url holds the truth, and repeating the extension name doesn't give much extra in terms of usability.

editor/src/clj/editor/defold_project.clj
```diff
@@ -1889,10 +1902,15 @@
 
 (defn open-project! [graph extensions workspace-id game-project-resource render-progress!]
   (let [dependencies (read-dependencies game-project-resource)
+        include-dependencies-metadata (read-dependencies-metadata-setting game-project-resource)
```
vlaaad: Let's only read `game-project-resource` once when opening the project. We should merge `read-dependencies` and `read-dependencies-metadata-setting`

ekharkunov:
> I think think can be solved by the user in the Lua script, if they want to store the information differently. Imo, the url holds the truth, and repeating the extension name doesn't give much extra in terms of usability.

Ofc, user can do it himself. But why he can't do it before this changes? I think we misunderstand each other. I'm talking about how to extract info from the url, not about how to store information.

editor/src/clj/editor/build.clj
```diff
@@ -209,12 +213,37 @@
...
+          (DependencyMetadata/metadataPath
+            (library/directory (workspace/project-directory workspace))))
+
+        build-metadata-file (io/file build-dir DependencyMetadata/OUTPUT_PATH)]
+    (if (and (project/include-dependencies-metadata? (project/settings project evaluation-context))
+             (.exists source-metadata-file))
+      (DependencyMetadata/minifyJson (.toPath source-metadata-file) (.toPath build-metadata-file))
+      (Files/deleteIfExists (.toPath build-metadata-file)))))
+
+(defn build-build-targets! [build-targets project workspace old-artifact-map render-progress! evaluation-context]
```
vlaaad: Arity changed, but not every use was updated. See `editor/src/dev/load_project.clj`.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/GameProjectBuilder.java
```diff
@@ -291,6 +287,13 @@ private ResourceGraph createResourceGraph(Project project) throws CompileExcepti
                 graph.add(resource);
             }
         }
+        for (String path : CopyCustomResourcesBuilder.getCustomResourcePaths(project)) {
+            IResource resource = project.getResource(path);
+            graph.add(resource);
+        }
```
vlaaad: From codex: this code now adds every `custom_resources` path to ResourceGraph with `graph.add(resource)`. Custom resources were previously leaf archive entries via r.output(), but ResourceWalker parses registered proto source extensions like `.collection`, `.go`, and `.data`; custom resources with those extensions can now fail graph generation or pull in unintended dependencies.

editor/src/clj/editor/library.clj
```diff
@@ -70,14 +71,21 @@
...
+          nil
+          (Progress.
+            (reify Progress$Reporter
+              (report [_ message fraction] (render-progress! (progress/bob message fraction)))
+              (close [_] (render-progress! progress/done))
+              (isCanceled [_] false))))]
+    (if (and include-dependencies-metadata
+             (coll/not-empty library-results))
+      (DependencyMetadata/saveAsJson lib-dir library-results)
+      (DependencyMetadata/deleteJson lib-dir))
```
vlaaad: This is wrong. `editor.library` is a small wrapper around Library.java whose only responsibility is calling corresponding Library methods. Library.java does not save/delete the dep metadata, so library.clj shouldn't either.

editor/src/clj/editor/library.clj
```diff
@@ -70,14 +71,21 @@
...
-    (or library-uris [])
-    (directory project-directory)
-    nil
-    nil
-    (Progress.
-      (reify Progress$Reporter
-        (report [_ message fraction] (render-progress! (progress/bob message fraction)))
-        (close [_] (render-progress! progress/done))
-        (isCanceled [_] false)))))
+(defn fetch! [project-directory library-uris render-progress! include-dependencies-metadata]
```
vlaaad: Arity of `library/fetch!` was changed, but not every use of it was updated. See `editor/src/dev/load_project.clj`.

vlaaad (changes requested):
I think the editor implementation is generally wrong. There is a comment in library.clj why it's wrong to do the thing this PR makes it do there, but that's not even the main problem with it. The code decides whether to produce dependency metadata JSON when **fetching** libraries, which does not make sense. What if you set a `game.project` setting to false, then do fetch, then change it back to true, then build? 

Could it be that we always want to produce this source file, and only minify/delete it in the build output when actually building? Do we need the source file at all? Should we instead only produce the minified version when building with the setting enabled?

Also, should we, perhaps, design an extensible way for users to define regexes that extract version information from dep URLs? Not everyone uses GitHub as a deps CDN. I started prototyping the idea of parsing versions from URLs for the asset portal (if we know a version, we can check if there is an update) — https://docs.google.com/document/d/1vRoIvbVE5ET2AasqURmo4vCahfe9BnsOVx5XB3rNeAE/edit 

The parsing of URIs to deps is a thing that might be useful outside of runtime metadata, but it should be extensible somehow... Maybe we should reconsider the design and scope?

JCash:
@vlad 

> Could it be that we always want to produce this source file

Yes, that is definitely the intention (and how Bob works)

> Also, should we, perhaps, design an extensible way for users to define regexes that extract version information from dep URLs? Not everyone uses GitHub as a deps CDN. I started prototyping the idea of parsing versions from URLs for the asset portal (if we know a version, we can check if there is an update) —

I'm now more inclined to simply _skip_ the version altogether, since there is no standard, and we don't want to add any "magic" code. I feel adding a plugin API for this makes little sense.
I'll leave "version" out for now, and we can have another ticket for adding it (with a separate design) later on if we need it.

vlaaad (changes requested):
In bob, after every call to `Library.fetch`, we do `syncDependencyMetadata(resolvedLibs)`. In editor, after every call to `library/fetch!`, we call equivalent `project/sync-dependencies-metadata!`. To me, it looks like writing dependencies metadata is the responsibility of `Library.fetch` method in bob. This will simplify both editor and bob.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/GameProjectBuilder.java
vlaaad: This problem is still unaddressed. You can reproduce it with this test (in ProjectBuildTest.java):
```java
    @Test
    public void testArchiveBuildWithLuaCustomResourceTreatsItAsLeaf() throws IOException, CompileExceptionError, MultipleCompileException {
        createDefaultFiles();
        createFile(contentRoot, "game.project",
                "[project]\n" +
                "custom_resources = /custom\n" +
                "\n" +
                "[display]\n" +
                "width=640\n" +
                "height=480\n");
        createFile(contentRoot, "custom/payload.lua",
                "function init(self)\n" +
                "end\n");

        buildArchive(false);

        assertTrue(new File(contentRoot, "build/custom/payload.lua").isFile());
        assertFalse(new File(contentRoot, "build/custom/payload.luac").exists());
    }
```

vlaaad:
@JCash I left a comment on a still-unaddressed comment

JCash:
Thanks @vlaaad !
Not sure how my local review missed that (It has no extra comments currently)

@codex review please

editor/test/integration/library_test.clj
```diff
@@ -142,9 +148,16 @@
         (game-project/set-setting! game-project ["project" "dependencies"] [uri])
         (fetch-libraries! workspace (project/project-dependencies project) identity)
         (is (= 1 (count (project/find-resources project "lib_resource_project/simple.gui"))))
+        (is (true? (.isFile dependency-metadata-file)))
```
vlaaad: `(true? ...)` is redundant here (and below), can be safely omitted, e.g.:
```clj
(is (.isFile dependency-metadata-file))
```

vlaaad (dismissed):
I think the changes are fine. One thing that worries me a bit is that we have a build pipeline that was built around transforming sources/inputs into outputs in the build folder, and dependency metadata building just... ignores it. It just has its own parallel code that ignores our general approach. I wonder if it's possible to actually build this file inside our build system? e.g., as an extra output / build target of `/game.project`? What do you think?

vlaaad:
From my minion:
> For Bob: remove `.internal/lib/dependencies.json` entirely and keep dependency metadata in memory as `Library.Result` values from `resolve`/`cached`/`mount`. Add a dedicated dependency-metadata task wired from `GameProjectBuilder` as a normal `/game.project` subtask. It should generate `build/.internal/dependencies.json` directly from the current mounted dependency results when `project.dependencies_metadata` is enabled. If disabled, Bob still needs targeted stale-output cleanup or general pruning of outputs that were present in the previous state but are no longer declared.

> For the editor: use the workspace’s current dependency results instead of reading a file from `.internal/lib`. Add a normal build target from `GameProjectNode` when `dependencies_metadata` is enabled, outputting `.internal/dependencies.json` into the build directory. Since `pipeline/build!` already prunes files not represented by current build targets, disabling the setting should naturally remove the build metadata output.

I think it's comment is very useful about the editor — it does delete build files when they are not a part of the build, which is what we want here, right?

JCash:
> It just has its own parallel code that ignores our general approach.

@vlaaad I agree with the potential improvement of using a more standard way to include it. I'll look into it.
However, it's not completely ignoring the build mechanisms. It becomes part of the "custom resources" step, which is a normal process. Explicitly _not_ transforming, but simply copying (although that may be considered a transform of sorts). The new part is that it originate from _outside_ of the project.

> .internal/lib/dependencies.json entirely and keep dependency metadata in memory as Library.Result

This comment ignores the fact that the `resolve` command is an explicit command, that can be run apart from the actual build step. There is no "memory" to be shared, other than the file on disc.

We could of course recalculate that information for each build, at some larger cost (calculating a sha1 for the entire extension-rive for instance), due to the fact that all information is available "offline".

I might also try to modify the `custom_resources` value, to inject the file there. That should further simplify code imho.

vlaaad:
When building, we use either `Library.fetch`, or `Library.cached`; in both cases, we have a list of `Library.Result`, which is all you need.

vlaaad:
Hmm, but I see, it's expensive to calculate sha if extension-rive...

editor/src/clj/editor/game_project.clj
```diff
@@ -74,6 +80,47 @@
...
+  (make-writer [this opts] (io/make-writer (io/make-output-stream this opts) opts)))
+
+(defn- make-dependency-metadata-build-target [node-id workspace dependencies]
+  (let [dependency-metadata (DependencyMetadata/collect dependencies)]
+    (when (coll/not-empty dependency-metadata)
+      (bt/with-content-hash
+        {:node-id node-id
+         :resource (workspace/make-build-resource (->DependencyMetadataResource workspace))
+         :build-fn build-dependency-metadata
+         :user-data (.writeValueAsBytes dependency-metadata-object-mapper dependency-metadata)}))))
```
vlaaad: We *build* the thing when producing *input* to the builder, which wastes CPU. The point of the build system is to skip rebuilds when inputs didn't change, so we should try to do more work in the build fn. I think we should put `dependencies` as a userdata, and then collect + serialize in the build fn instead. We might want to add some custom dgestable for the `Library.Result` class in this case.

editor/src/clj/editor/game_project.clj
```diff
@@ -153,15 +200,20 @@
...
+                                          (assoc ["input" "gamepads"] (:resource gamepads-build-target)))
+          game-project-build-target (bt/with-content-hash
+                                      {:node-id _node-id
+                                       :resource (workspace/make-build-resource resource)
+                                       :build-fn build-game-project
+                                       :user-data {:settings-map settings-map
+                                                   :meta-settings (:settings meta-info)
+                                                   :path->built-resource-settings path->built-resource-settings}
+                                       :deps dep-build-targets})
+          dependency-metadata-build-target (when (true? (get settings-map dependencies-metadata-setting-path))
```
vlaaad: I'd suggest doing a pass of `$Clojure Code Style` skill here, particularly about trivial single-use code — this can move to `cond->`.

editor/test/integration/library_test.clj
```diff
@@ -146,9 +152,16 @@
...
+        (is (true? (.contains (slurp dependency-metadata-file) uri)))
+        ;; disable dependency metadata, fetch libraries, we should still update the metadata file
+        (game-project/set-setting! game-project ["project" "dependencies_metadata"] false)
+        (fetch-libraries! workspace (project/project-dependencies project) identity)
+        (is (true? (.isFile dependency-metadata-file)))
```
vlaaad: I think my comment about redundant `true?` wasn't addressed?

vlaaad (commented):
Looks good, I think we can improve things a bit in the editor build code, but otherwise I'm happy with the PR 👍

editor/src/clj/editor/game_project.clj
JCash: The style skill _is_ loaded, so perhaps it's lacking some instructions for this? 🤔

editor/src/clj/util/digestable.clj
```diff
@@ -245,6 +247,27 @@
   (digest! [value writer opts]
     (digest-sequence! "[" value "]" writer opts))
 
+  Library$Archive
+  (digest! [value writer opts]
+    (digest-tagged! 'LibraryArchive
+                    {:path (str (.path value))
+                     :size (Files/size (.path value))
+                     :modified-time (str (Files/getLastModifiedTime (.path value) (make-array LinkOption 0)))
```
vlaaad: Not sure this is necessary? Hmm. The idea with hashing here is that it hashes in-memory inputs to functions. Now I see my suggested approach was wrong since `DependencyMetadata/collect` is not a pure function of its inputs. I was wrong. We need to always call `DependencyMetadata/collect` in `produce-build-targets` and pass the result of that function as user data to the builder. THen, we should delete these digestables. Sorry for misleading you 😅

vlaaad (changes requested):

editor/src/clj/editor/game_project.clj
```diff
@@ -80,10 +81,26 @@
...
+(defn- dependency-content-hash-data [dependencies]
+  (into []
+        (map (fn [^Library$Result dependency]
+               (let [archive ^Library$Archive (.archive dependency)]
+                 (cond-> {:uri (.uri dependency)
+                          :problem (some-> (.problem dependency) str)}
+                   archive
+                   (assoc :archive {:path (str (.path archive))
+                                    :size (Files/size (.path archive))
+                                    :modified-time (str (Files/getLastModifiedTime (.path archive)
```
vlaaad: Nit: there is `path/last-modified-ms`

editor/src/clj/editor/game_project.clj
```diff
@@ -80,10 +81,26 @@
...
+
+(defn- dependency-content-hash-data [dependencies]
+  (into []
+        (map (fn [^Library$Result dependency]
+               (let [archive ^Library$Archive (.archive dependency)]
+                 (cond-> {:uri (.uri dependency)
+                          :problem (some-> (.problem dependency) str)}
+                   archive
+                   (assoc :archive {:path (str (.path archive))
+                                    :size (Files/size (.path archive))
```
vlaaad: Nit: there is `path/byte-size`

vlaaad (dismissed):
LGTM!

PR #12649, opened 2026-06-25
Move PBR related code from comp_model into comp_model_pbr

Jhonnyg:
#### Technical changes

* Moved PBR render-constant setup out of comp_model.cpp into the new comp_model_pbr.cpp / comp_model_pbr.h files.
* Fixed a crash when rendering PBR materials for meshsets with no embedded material entries.
* Added default glTF/PBR material fallback values when material_index is outside the meshset material table.
* Fixed PBR texture detection for generated models where the mesh .glb has no material data, but the Defold .model has texture bindings.


engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -42,6 +42,7 @@
 #include "../resources/res_skeleton.h"
 #include "../resources/res_texture.h"
 #include "comp_private.h"
+#include "comp_model_pbr.h"
```
Jhonnyg: In an effort to reduce some clutter in this file, I decided to move out the PBR related code into a separate module.

engine/gamesys/src/gamesys/components/comp_model_pbr.cpp
```diff
@@ -0,0 +1,401 @@
...
+            * };
+            *******************************************/
+            bool has_base_color_texture = ddf_material.m_Pbrmetallicroughness.m_Basecolortexture.m_Texture.m_Index != -1;
+            bool has_metallic_roughness_texture = ddf_material.m_Pbrmetallicroughness.m_Metallicroughnesstexture.m_Texture.m_Index != -1;
+
+            if (using_default_material)
+            {
+                has_base_color_texture = has_base_color_texture || HasModelTexture(material_info, PBR_METALLIC_ROUGHNESS_BASE_COLOR_TEXTURE_SAMPLER);
+                has_metallic_roughness_texture = has_metallic_roughness_texture || HasModelTexture(material_info, PBR_METALLIC_ROUGHNESS_METALLIC_ROUGHNESS_TEXTURE_SAMPLER);
+            }
```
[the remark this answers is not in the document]
Jhonnyg: I attempted to fix this, but it only made things worse. I _think_ it should be up to the content itself to declare if they should be using textures or not. The way we currently have to do it is to bind textures to all slots anyway otherwise the editor will cause GL errors, so we need a more holistic approach to this problem later on. For now it works as before and looks good, so I say we postpone this work a bit.

engine/graphics/src/dx12/graphics_dx12.cpp
```diff
@@ -1668,12 +1669,7 @@ namespace dmGraphics
         ubo->m_BaseUniformBuffer.m_BoundBinding = binding;
         ubo->m_BaseUniformBuffer.m_BoundSet     = set;
 
-        if (context->m_CurrentUniformBuffers[set][binding])
-        {
-            DX12DisableUniformBuffer(_context, (HUniformBuffer) context->m_CurrentUniformBuffers[set][binding]);
-        }
-
-        context->m_CurrentUniformBuffers[set][binding] = ubo;
+        context->m_CurrentUniformBuffers[set][binding] = ubo;
```
Jhonnyg, line 1672: There was a logic issue here that caused the light buffer to first unbind itself and then force bind itself while the UBO thought it was unbound 🤔 

Now we allow a UBO to be bound to multiple units (like textures) and then let the calling code (light.cpp) handle the bind/unbind more explicitly.

JCash (approved):

PR #12652, opened 2026-06-26
Switched to explicit code signing and added macOS entitlements

britzl:
Code signing is now an explicit option (changed from`--skip-codesign` to `--codesign`). The macOS release, debug and headless engines plus texc, modelc and shaderc libs are signed and relevant entitlements are added. The entitlements are also added when bundling for macOS.


JCash (approved):
lgtm!

PR #12656, opened 2026-06-27
Align the texture buffer for morph targets to 4

Jhonnyg:
Regression fix for 1.13.1:
While moving the texture generation process to the pipeline step, it caused WebGL to behave erratically while all other platforms behaved correctly. After a ton of investigation, it boiled down to misaligned memory for the texture buffer.

We now always align the first image data appended to texture sets by 4, so it's future proof when we introduce HDR textures and so forth.


JCash (approved):
lgtm!

PR #12657, opened 2026-06-28
Draft: add editor search exclusion patterns

### LINKED ISSUE #11267: Exclude files from Searches | Mark directories as excluded
When using "Search in Files" or "Open Asset" it often includes files I am not interested in. When I develop I usually have external libraries, test, and other data that I know I don't want to show up in any kind of search.

I would like to be able to mark these directories as excluded to stop them from ending up in any "search" dialog.

Doing it manually

Common feature in other IDEAs however there they would also be excluded from any Language Server feature which I don't think I want here.

This is an editor only feature and would infer no change on any builds.

softwaresat:
Adds per-project editor search exclusion patterns for Open Asset and Search in Files. In the Open Asset dialog, users can open a filter dropdown, add directory or path patterns to exclude, and see each added exclusion as an entry in the dropdown. Each exclusion can be enabled or disabled with a checkbox, removed when no longer needed, and filtering can also be turned on or off globally. The exclusions are saved per project in .editor_settings, so they persist between editor sessions without affecting game.project, builds, bundling, or runtime behavior.

Fixes #11267

This is opened as a draft PR to make the proposed approach easier to review before final polishing.

### Technical Changes

- Adds [:search :exclude-patterns] as an array of [pattern enabled] pairs and [:search :filtering] as a boolean to the prefs schema, persisted per project in .editor_settings.
- Extends make-select-list-dialog with :extra-initial-state, :extra-event-handler, :refilter-atom, and :header-extra-desc-fn so callers can add sidebar/header UI and custom events without changing the dialog internals.
- Updates resource-dialog/make to optionally build an exclusion filter UI when supplied with exclusion-pattern atoms. The UI uses a cljfx popup with an enabled checkbox list and add-pattern text field.
- Applies exclusion patterns inside the existing resource dialog filter-fn, reading from the exclusion-patterns atom on each filter call.
- Uses a refilter-atom so changes made in the popup immediately rerun filtering.
- Normalizes added exclusions by storing the pattern directly, without a leading !, since all stored patterns are exclusions by definition.
- Adds compile-exclude-pred as a simple predicate over enabled pattern pairs.
- Creates and saves exclusion-pattern prefs through query-and-open!, then passes the required atoms and callbacks through to resource-dialog/make.
- Applies the same exclusion patterns to Search in Files by accepting exclude patterns in make-search-data-future and reading them from prefs in search_results_view.
- Adds tests for exclusion predicate behavior and Search in Files filtering.

One question I had while implementing this: should these exclusion patterns behave like general text substring filters, or should they be directory/path-oriented only? For example, should adding input exclude any file whose path contains input, or should the UI/logic guide users toward excluding directories only?

## PR checklist

* [x] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


editor/src/clj/editor/app_view.clj
```diff
@@ -3227,18 +3146,35 @@
 
 (def ^:private open-assets-term-prefs-key [:open-assets :term])
 
+(defn- search-pref-key [key]
```
JosephFerano: This doesn't need its own function. Please inline. This pattern is elsewhere in this PR as well, I would inline those as well.

editor/src/clj/editor/resource_dialog.clj
```diff
@@ -29,10 +35,69 @@
...
+          badge (filter-popup/badge-count state)
+          show-counter (pos? badge)
+          anchor (filter-popup/anchor-point node)]
+      {:fx/type fxui/with-popup-window
+       :desc {:fx/type fx/ext-on-instance-lifecycle
+              :on-created (fn [^Node n] (reset! button-node-atom n))
+              :desc {:fx/type fx.button/lifecycle
+                     :focus-traversable false
+                     :mnemonic-parsing false
+                     :style (str "-fx-background-color: #393c41;"
```
JosephFerano: CSS shouldn't go here, this should go in scss files. But also, there's a broader issue here;

I believe the CSS already exists, because `console.clj` already has a filter button with CSS applied. It seems like we duplicated a lot of the logic that's there for the console rather than refactoring the console filter button so that it's more generally available, the UI that is.

editor/src/clj/editor/filter_popup.clj
```diff
@@ -0,0 +1,214 @@
...
+                                          :max-width ##Inf
+                                          :mnemonic-parsing false
+                                          :selected (boolean (get state key))
+                                          :on-selected-changed {:event-type :filter-popup/toggle-filter :key key}
+                                          :text (localization (localization/message label))}))
+                                  resource/exclude-filters)
+                            true
+                            (conj {:fx/type fx.separator/lifecycle
+                                   :style-class "console-filter-popup-separator"})
+                            true
```
JosephFerano: If you have a bunch of `cond->` branches with `true` then perhaps `cond->` is not the right function to use here, in particular since they use `conj`, meaning the first element should have these already. But perhaps you're trying to conditionally place things in between other elements, I would suggest a `concat` or something else.

JosephFerano (changes requested):
Hi Satvik, I'll continue the discussion here in the PR that way I can also start reviewing the code itself.

I'm a little surprised by how much code this is requiring. I haven't been able to review everything but the amount gives me the impression that this is doing a bit too much.

`make-select-list-dialog` now has a new contract for some of the parameters that expect functions. This broke "Jump to Symbol", as an example.

The team is currently out on vacation, I'll slowly keep reviewing in the meantime, so hang in there.

editor/src/clj/editor/app_view.clj
softwaresat: Made change

editor/src/clj/editor/resource_dialog.clj
softwaresat: Removed the duplicated CSS by adding a new shared CSS file

editor/src/clj/editor/filter_popup.clj
softwaresat: I removed the repeated true branches and replaced them with into for the generated exclusion toggles, conj for the unconditional controls, and a cond-> for optional pattern list.

JosephFerano:
Hi Satvik, I'll take a look at some point this week, there was a commit I made that touches some of the code you're working on and there are now merge conflicts, could you please address those before I take a look?

softwaresat:
I resolved all merge conflicts

editor/src/clj/editor/filter_popup.clj
```diff
@@ -0,0 +1,227 @@
...
+                                                                         :filter-popup/remove)}}))
+                              (conj {:fx/type fx.text-field/lifecycle
+                                     :v-box/margin 4
+                                     :min-width 220
+                                     :text filter-popup-text
+                                     :prompt-text (localization (localization/message "dialog.open-assets.filter.prompt"))
+                                     :on-text-changed {:event-type :filter-popup/set-text}
+                                     :on-action {:event-type :filter-popup/add}}))}]}]})
+
+(defn event-handler
```
JosephFerano, line 180: We have stopped using the event-handler pattern for cljfx, which I'm not sure if it's documented anywhere. There is a discussion about it in the following PR

https://github.com/defold/defold/pull/11441#discussion_r2538550969

Vlad gives some reasoning and there is probably further discussion here. But also you can take a look at the breakpoints-view ns to get an idea of the patterns. The general idea is to just inline the functions directly. Notice the use of `fn/partial` for partial application. So this should make a good reference.

PR #12661, opened 2026-06-29
Added lit materials with skinning taking into account normal.

### LINKED ISSUE #12624: Built-in skinned lit material does not take skinned normals into account for shading
| Left (without skinned normals) | Right (with skinned normals) |
|---|---|
|Image|Image|
|Image|Image|

**To Reproduce (REQUIRED)**
1. Use built-in skinned lit material with a directional light in the setup

Normals taken into account

  - 1.13.1 (alpha)

**Platforms (REQUIRED):**
 - Platforms: Linux
 - OS: Ubuntu 24.04
 - Device: Acer Nitro 16

**Workaround (OPTIONAL):**

**Screenshots (OPTIONAL):**

paweljarosz:
Modified the built-in lit skinned materials and added vertex programs that take into account normal when applying lighting.

Fixes #12624


engine/engine/content/builtins/materials/model_skinned_instanced_lit.vp
```diff
@@ -0,0 +1,48 @@
+#version 140
```
Jhonnyg: We shouldn't be needing a new .vp file for this, there's not model_lit.vp so presumably we shouldn't need new variants for the skinning.

engine/engine/content/builtins/materials/model_skinned_instanced_lit.vp
```diff
@@ -0,0 +1,48 @@
...
+    mediump mat4 mtx_proj;
+    mediump vec4 light;
+};
+
+#include "/builtins/materials/skinning.glsl"
+
+void main()
+{
+    vec4 skinned_position = get_skinned_position(position);
+    vec3 skinned_normal = get_skinned_normal(normal);
```
Jhonnyg: I'm guessing this is the only new piece needed - you can add this line into the existing skinned .vp files instead and remove the variants.

Jhonnyg (changes requested):
We should not add new .vp file if we can avoid it.

engine/engine/content/builtins/materials/model_skinned_instanced_lit.vp
paweljarosz: I didn't want to affect the other vp cause maybe there was a reason not to take them into account? But if not, I can ofc update the existing ones

Jhonnyg (approved):

Issue #8420, opened 2024-01-17
Add support for string values to go.property


drajjon:
**Is your feature request related to a problem? Please describe (REQUIRED):**
I'd like support for normal "string" properties on scripts, in addition to the existing "hash" support. I want it currently for unit test configuration (e.g. "run tests matching this regex") and in the future I expect to need it for other strategy/card-game related configuration.

**Describe the solution you'd like (REQUIRED):**
I can do this and it works. (no errors, I can go edit this string property in the editor)
`go.property('pattern', '.*') -- default to all tests`

**Describe alternatives you've considered (REQUIRED):**
Hashes work for most Defold-related use cases, but you cannot turn a hash back into a string in a release-mode way. (and in debug mode, it's also blocked due to a bug: #7422 ) So if you actually need the STRING (to display to user e.g. character names, to serialize/save data, to use as a regex string...) you cannot use go.property at all right now.

**Additional context (OPTIONAL):**
<img width="500" alt="image" src="https://github.com/defold/defold/assets/133168946/702fd629-fe48-4bda-8cf9-aad62e330713">

ekharkunov:
You can use custom data component to store any type of data. Here is an example how to implement custom data component https://github.com/defold/extension-simpledata.
After implementation you can add custom data component to any game object and get data from it via `go.get("url_of_component", "field_name")

davebalmer:
This is the sort of thing that can hamper bonding with Defold as a new experience. Seems supporting string properties would be the most intuitive and elegant. Especially for the case where we're using `factory.create()` etc. It's nice to let the `go` handle its own init at the time of init, and passing the data necessary for that at the time it is created would be great. 

I know strings are expensive internally, but this would be a big quality of life win. Especially for new devs coming from other engines. I ask that you consider this instead of having us split string data into special cases separate from the other properties (eg. index into a table in a separate lua module or using subsequent calls to `msg.post()`). Thanks!

AGulev:
I think I should post my opinion from discord conversation we had as well
https://discord.com/channels/250018174974689280/817015172987486208/1265385389896630484

From my point of view, the string type in game logic isn't particularly useful.

In 99% of cases, it cannot be used directly without extra preprocessing. In the remaining 1%, such as showing text on the screen, it's still not very useful. Most of the time, you want an abstraction that can return localized text by key, not a direct value. So, I don't see any useful applications for a string as a GO property because if you need to use it as a key, you can use a hash instead.

[...]

PR #12675, opened 2026-07-01
Text support for script properties deliverable 1

Jhonnyg:
First deliverable of issue #8420 - text support in script properties.

#### Technical changes

* Adds PROPERTY_TYPE_TEXT to the game object property data model and DDF property declarations.
* Extends Bob/Lua scanning so go.property("name", "text") is parsed as a text property, including escaped strings, long Lua strings, and UTF-8 cntent.
* Serializes text properties into PropertyDeclarations.text_entries, reusing the existing string_values pool for storage.
* Adds pipeline support for text property overrides in .go and .collection files via PROPERTY_TYPE_TEXT.
* Extends the engine-side PropertyVar variant with borrowed const char* text values.
* Extends property containers to own copied text strings and return them as PROPERTY_TYPE_TEXT.
* Wires static text properties into script initialization so self.<property> receives the declared or overridden text value.
* Adds test coverage for Lua scanning, Bob serialization, game object/collection property overrides, property containers, and script default/override initialization.


engine/gameobject/src/gameobject/gameobject_script.cpp
```diff
@@ -2692,6 +2703,16 @@ namespace dmGameObject
             lua_pushboolean(L, var.m_Bool);
             lua_settable(L, index - 2);
         }
+        count = declarations->m_TextEntries.m_Count;
```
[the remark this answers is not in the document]
Jhonnyg: This is fine, first delivery does not add get/set support.

com.dynamo.cr/com.dynamo.cr.bob.test/src/com/dynamo/bob/pipeline/test_props.lua
```diff
@@ -18,7 +18,7 @@ go.property  ('prop5', 0)
 
 go.property(invalid_string, 2)
 go.property("three_args", 1, 2)
-go.property("unknown_type", "hello")
+go.property("text_type", "hello")
```
JCash, line 21: Perhaps test more variations, e.g. UTF8, Lua code, xml/json?

engine/gameobject/src/gameobject/gameobject_props.cpp
```diff
@@ -354,6 +357,16 @@ namespace dmGameObject
...
 
+    void PropertyContainerPushText(HPropertyContainerBuilder builder, dmhash_t id, const char* value)
+    {
+        uint32_t entry_offset = AllocateEntry(builder, id, PCV_TYPE_TEXT);
+        uint32_t string_offset = builder->m_StringOffset;
+        builder->m_PropertyContainer->m_Indexes[entry_offset] = string_offset;
+        size_t string_length = strlen(value) + 1;
+        memcpy(&builder->m_PropertyContainer->m_StringData[string_offset], value, string_length);
+        builder->m_StringOffset += string_length;
+    }
```
JCash: I think an improvement is that the callers passes in the string length.
They either have the length already, or can call strlen() themselves.

engine/gameobject/src/dmsdk/gameobject/gameobject.h
```diff
@@ -369,6 +374,7 @@ namespace dmGameObject
             float m_V4[4];
             float m_M4[16];
             bool m_Bool;
+            const char* m_Text;
```
JCash, line 382: Seems the caller own the memory.
Perhaps make a note of it here?

engine/gameobject/src/gameobject/gameobject_props_ddf.cpp
```diff
@@ -37,6 +39,10 @@ namespace dmGameObject
         {
             params.m_URLStringSize += strlen(prop_descs->m_StringValues.m_Data[prop_descs->m_UrlEntries[i].m_Index]) + 1;
         }
+        for (uint32_t i = 0; i < prop_descs->m_TextEntries.m_Count; ++i)
+        {
+            params.m_TextSize += strlen(prop_descs->m_StringValues.m_Data[prop_descs->m_TextEntries[i].m_Index]) + 1;
```
JCash: It seems a bit strange that we don't know the sizes of the strings here, as they just came from the content pipeline?
If it's not stored in the data format, perhaps we can use the prev/current indices to do it?

JCash (changes requested):
Overall good.

Could use some documentation and some strlen() cleanup

JCash (approved):

PR #12679, opened 2026-07-02
Improve idle CPU usage

### LINKED ISSUE #12548: High CPU usage on macOS
The editor often has surprisingly high cpu usage even when idle and even when the window isn't focused.

**To Reproduce (REQUIRED)**
1. Open a Defold project
2. Put the editor in the background
3. Observe CPU usage in Activity Monitor

CPU usage should be down in low single digits when idle and not having focus.

  - 1.12.4

**Platforms (REQUIRED):**
 - Platforms: macOS
 - OS: Sequoia 15.7.7
 - Device: M3 Pro

**Screenshots (OPTIONAL):**

Image

vlaaad:
Reduce idle editor CPU usage when the editor does not have focus. This drops CPU from around 8-12% to 3-6% on macOS.

Technical notes:

Measurements show that most of the time is spent in various refreshes we run when drawing:
Screenshot 2026-07-02 at 09 58 18
Particularly, in the repaint of the console view (mostly node value and cache updates), right split refresh, and scene cache pruning.

This PR adds some micro-optimizations to the hot spots and additionally adds throttling to timers so all running timers are limited to 15 fps when the app is not focused.

Here is a CPU diff graph of before (red) vs after (blue):
Screenshot 2026-07-02 at 10 04 43

Fix #12548

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -386,6 +386,7 @@
   (output open-sidebar-panes g/Any :cached (g/fnk [open-sidebar-panes] (into {} open-sidebar-panes)))
   (output open-views g/Any :cached (g/fnk [open-views] (into {} open-views)))
   (output open-dirty-views g/Any :cached (g/fnk [open-dirty-views] (into #{} (keep #(when (second %) (first %))) open-dirty-views)))
+  (output scene-view-ids g/Any :cached (gu/passthrough scene-view-ids))
   (output hidden-renderable-tags types/RenderableTags (gu/passthrough hidden-renderable-tags))
   (output hidden-node-outline-key-paths types/NodeOutlineKeyPaths (gu/passthrough hidden-node-outline-key-paths))
   (output active-tab-pane TabPane (g/fnk [^Tab active-tab ^SplitPane editor-tabs-split]
--- a/editor/src/clj/editor/code/view.clj
+++ b/editor/src/clj/editor/code/view.clj
@@ -865,6 +865,12 @@
 (g/defnk produce-cursor-repaint-info [canvas color-scheme cursor-opacity layout lines repaint-trigger visible-cursors :as cursor-repaint-info]
   cursor-repaint-info)
 
+(g/defnk produce-tick-info [cursor-opacity elapsed-time-at-last-action gesture-start layout lines :as tick-info]
+  tick-info)
+
+(g/defnk produce-repaint-info [canvas-repaint-info completions-showing cursor-repaint-info hover-showing-regions rename-cursor-range resource-node :as repaint-info]
+  repaint-info)
+
 (defn- make-cursor-rectangle
   ^Rectangle [^Paint fill opacity ^Rect cursor-rect]
   (doto (Rectangle. (.x cursor-rect) (.y cursor-rect) (.w cursor-rect) (.h cursor-rect))
@@ -1629,7 +1635,9 @@
   (output minimap-cursor-range-draw-infos CursorRangeDrawInfos :cached produce-minimap-cursor-range-draw-infos)
   (output execution-markers r/Regions :cached produce-execution-markers)
   (output canvas-repaint-info g/Any :cached produce-canvas-repaint-info)
-  (output cursor-repaint-info g/Any :cached produce-cursor-repaint-info))
+  (output cursor-repaint-info g/Any :cached produce-cursor-repaint-info)
+  (output tick-info g/Any :cached produce-tick-info)
+  (output repaint-info g/Any :cached produce-repaint-info))
 
 (defn- mouse-button [^MouseEvent event]
   (condp = (.getButton event)
@@ -2145,7 +2153,7 @@
 
 (def ^:private ^:const completion-icon-text-spacing 4.0)
 
-(defn- completion-list-cell-view [completion]
+(defn- completion-list-cell-view [view-node completion]
   (if completion
     (let [text (:display-string completion)
           matching-indices (fuzzy-choices/matching-indices completion)]
@@ -2162,14 +2170,15 @@
                             (fuzzy-choices/make-matched-text-flow-cljfx
                               text matching-indices
                               :deprecated (contains? (:tags completion) :deprecated))]}
-       :on-mouse-clicked {:event :accept :completion completion}})
+       :on-mouse-clicked (fn [_]
+                           (accept-suggestion! view-node (code-completion/insertion completion)))})
     {}))
 
 (defn- completion-popup-view
   [{:keys [canvas-repaint-info font font-name font-size
            visible-completion-ranges query screen-bounds completions-showing
            completions-doc project completions-combined completions-selected-index
-           completions-shortcut-text]}]
+           completions-shortcut-text view-node]}]
   (let [item-count (count completions-combined)]
     (if (or (not completions-showing) (zero? item-count))
       {:fx/type fxui/ext-value :value nil}
@@ -2244,14 +2253,26 @@
                   :props {:items [refresh-key completions-combined]}
                   :desc
                   {:fx/type fx.ext.list-view/with-selection-props
-                   :props {:on-selected-index-changed {:event :select}}
+                   :props {:on-selected-index-changed (fn [index]
+                                                        (set-properties! view-node nil {:completions-selected-index index})
+                                                        (resolve-selected-completion! view-node))}
                    :desc
                    {:fx/type fx.list-view/lifecycle
                     :style {:-fx-font-family (str \" font-name \") :-fx-font-size font-size}
                     :id "fuzzy-choices-list-view"
                     :style-class ["flat-list-view" "completion-popup-list-view"]
                     :fixed-cell-size cell-height
-                    :event-filter {:event :completion-list-view-event-filter}
+                    :event-filter (fn [e]
+                                    (when (instance? KeyEvent e)
+                                      (let [^KeyEvent e e
+                                            code (.getCode e)]
+                                        ;; redirect everything except arrows to canvas
+                                        (when-not (or (= KeyCode/UP code)
+                                                      (= KeyCode/DOWN code)
+                                                      (= KeyCode/PAGE_UP code)
+                                                      (= KeyCode/PAGE_DOWN code))
+                                          (ui/send-event! (get-property view-node :canvas) e)
+                                          (.consume e)))))
                     :min-width completions-width
                     :pref-width completions-width
                     :max-width completions-width
@@ -2259,7 +2280,7 @@
                     :pref-height (* cell-height item-count)
                     :max-height (min (* cell-height max-visible-completions-count) max-completions-height)
                     :cell-factory {:fx/cell-type fx.list-cell/lifecycle
-                                   :describe completion-list-cell-view}}}}}]
+                                   :describe (fn/partial completion-list-cell-view view-node)}}}}}]
                completions-shortcut-text
                (conj {:fx/type fx.label/lifecycle
                       :style-class "completion-popup-hint"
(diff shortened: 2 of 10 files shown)


editor/src/clj/editor/code/view.clj
```diff
@@ -1629,7 +1635,9 @@
   (output minimap-cursor-range-draw-infos CursorRangeDrawInfos :cached produce-minimap-cursor-range-draw-infos)
   (output execution-markers r/Regions :cached produce-execution-markers)
   (output canvas-repaint-info g/Any :cached produce-canvas-repaint-info)
-  (output cursor-repaint-info g/Any :cached produce-cursor-repaint-info))
+  (output cursor-repaint-info g/Any :cached produce-cursor-repaint-info)
+  (output tick-info g/Any :cached produce-tick-info)
+  (output repaint-info g/Any :cached produce-repaint-info))
```
vlaaad, line 1640: To minimize `node-value` calls, in `repaint-view!`, we now bundle all accessed data into 2 cached outputs.

editor/src/clj/editor/code/view.clj
```diff
@@ -2346,41 +2374,6 @@
...
-            (ui/send-event! (get-property view-node :canvas) e)
-            (.consume e)))))
-
-    :select
-    (do
-      (set-properties! view-node nil {:completions-selected-index (:fx/event e)})
-      (resolve-selected-completion! view-node))
-    :accept
-    (accept-suggestion! view-node (code-completion/insertion (:completion e)))))
-
```
vlaaad, line 2383: Inlined "map event handling" to simpler fn callback event handling above. This allowed us to get rid of cljfx `renderer` abstraction in favor of `advance-component!` in `repaint-view!`, which improved performance (no thread scheduling overhead)

editor/src/clj/editor/app_view.clj
```diff
@@ -386,6 +386,7 @@
   (output open-sidebar-panes g/Any :cached (g/fnk [open-sidebar-panes] (into {} open-sidebar-panes)))
   (output open-views g/Any :cached (g/fnk [open-views] (into {} open-views)))
   (output open-dirty-views g/Any :cached (g/fnk [open-dirty-views] (into #{} (keep #(when (second %) (first %))) open-dirty-views)))
+  (output scene-view-ids g/Any :cached (gu/passthrough scene-view-ids))
```
vlaaad, line 389: Though the output is trivial, it's pulled every frame, and checking the cache is much faster than pulling inputs.

editor/src/clj/editor/defold_project.clj
```diff
@@ -1837,11 +1837,13 @@
 (defn cache-retain?
   ([endpoint]
    (case (g/endpoint-label endpoint)
+     (:repaint-info :right-split-desc :scene-view-ids :tick-info) true
      (:build-targets) (project-resource-node? (g/now) (g/endpoint-node-id endpoint))
      (:save-data :save-value) (project-file-resource-node? (g/now) (g/endpoint-node-id endpoint))
      false))
   ([basis endpoint]
    (case (g/endpoint-label endpoint)
+     (:repaint-info :right-split-desc :scene-view-ids :tick-info) true
```
vlaaad, line 1846: Pulled every frame — guaranteed to be retained.

editor/src/clj/editor/scene_cache.clj
```diff
@@ -83,39 +83,23 @@
...
+          acc
+          (pair (assoc object-caches
+                       cache-id
+                       (assoc cache-meta
+                              :caches
+                              (assoc caches-by-context
+                                     context
+                                     (vary-meta cache-for-context dissoc :pending-deletions))))
+                (into pending-deletions pending-deletions-from-cache)))))
+    (pair object-caches
```
vlaaad, line 102: I did 2 things:
1. inlined `consume-pending-deletions-in-cache-meta` and `consume-pending-deletions-in-caches-by-context` to avoid fn call and pair construction overhead
2. added perf optimization to skip assocs when there are no changes.

editor/src/clj/editor/scene_cache.clj
```diff
@@ -191,49 +175,29 @@
...
+                                                         cached-object+request-data))
+                                                     cached-object+request-data-by-request-id)
+                                               (coll/not-empty)
+                                               (make-deletion destroy-batch-fn)
+                                               (vector)))
+                pruned-caches-by-context (assoc caches-by-context context pruned-cached-object+request-data-by-request-id)
+                pruned-cache-meta (assoc cache-meta :caches pruned-caches-by-context)]
+            (pair (assoc object-caches cache-id pruned-cache-meta)
+                  (into deletions deletions-from-cache))))))
+    (pair object-caches
```
vlaaad, line 200: Similar change here: inlined private single-use helpers, added a check to skip unnecessary work.

editor/src/clj/editor/ui.clj
```diff
@@ -2437,13 +2437,15 @@
...
+         unfocused-interval (max interval (long (* 1e9 (/ 1.0 15.0))))]
      {:last last
       :timer (proxy [AnimationTimer] []
                (handle [^long now]
                  (profiler/profile "timer" name
-                   (let [elapsed (- now start)
-                         delta (- now (long @last))]
+	                   (let [elapsed (- now start)
+	                         delta (- now (long @last))
+	                         interval (if (:focused @focus-state) interval unfocused-interval)]
```
vlaaad: throttling

britzl:
> This PR adds some micro-optimizations to the hot spots and additionally adds throttling to timers so all running timers are limited to 15 fps when the app is not focused.

This still seems quite high for an app that doesn't have focus. What are these timers?

editor/src/clj/internal/node.clj
```diff
@@ -128,7 +128,7 @@
 (defn input-dependencies       [nt]        (some-> nt deref :input-dependencies))
 (defn property-display-order   [nt]        (some-> nt deref :property-display-order))
 (defn cascade-deletes          [nt]        (some-> nt deref :cascade-deletes))
-(defn behavior                 [nt label]  (some-> nt deref (get-in [:behavior label])))
+(defn behavior                 [nt label]  (some-> nt deref :behavior (get label)))
```
vlaaad, line 131: microoptimization, not sure it was useful...

editor/src/clj/internal/node.clj
```diff
@@ -1611,8 +1611,8 @@
          (let [~node-id-sym (gt/node-id ~node-sym)]
            ~(check-jammed-form description label node-sym node-id-sym label-sym evaluation-context-sym
               (apply-default-property-shortcut-form description label node-sym node-id-sym label-sym evaluation-context-sym
-                (mark-in-production-form node-id-sym label-sym evaluation-context-sym
-                  (check-caches-form description label node-id-sym label-sym evaluation-context-sym
+                (check-caches-form description label node-id-sym label-sym evaluation-context-sym
+                  (mark-in-production-form node-id-sym label-sym evaluation-context-sym
```
vlaaad, line 1615: Microoptimization: if we have a thing in the cache, we only check the cache without wasting cpu/memory on marking output in production. If a thing is not in the cache, everything still works correctly.

editor/src/clj/editor/code/view.clj
```diff
@@ -3684,73 +3676,61 @@
...
+      ;; Show rename popup if appropriate
+      (fxui/advance-graph-user-data-component!
+        view-node :rename-popup
+        (when rename-cursor-range
+          (g/with-auto-evaluation-context evaluation-context
+            (g/node-value view-node :rename-view evaluation-context))))
+      ;; Show completion suggestions if appropriate.
+      (fxui/advance-graph-user-data-component!
+        view-node :completion-popup
+        (when completions-showing
```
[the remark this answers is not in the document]
vlaaad: This is fine; the fact that renderer exceptions didn't stop the timer was accidental.

vlaaad:
> This still seems quite high for an app that doesn't have focus. What are these timers?

Refresh/redraw timers, for example, if you have a console view, you want it to update when your game has focus, since you still might want to read console output as you interact with the game.

AGulev:
Do we want all these graphs in release notes where we usually have user facing changes?

editor/src/clj/editor/code/view.clj
vlaaad: Note: we do 2 outputs instead of 1 because `repaint-view!` may transact. 1 node-value call for before tx, 1 for after.

vlaaad:
@AGulev no, that's why they are after "technical notes" marker

editor/src/clj/editor/scene_cache.clj
```diff
@@ -191,49 +175,29 @@
...
-
 (defn- prune-context-in-object-caches [object-caches context]
   (reduce-kv
-    (fn [[object-caches deletions] cache-id cache-meta]
-      (let [[pruned-cache-meta deletions-from-cache]
-            (prune-context-in-cache-meta cache-meta context)]
-        (pair (assoc object-caches cache-id pruned-cache-meta)
-              (into deletions deletions-from-cache))))
-    (pair (empty object-caches)
+    (fn [[object-caches deletions :as acc] cache-id cache-meta]
```
matgis: We could defer unpacking `acc` until after the `(nil? cached-object+request-data-by-request-id)` check?

editor/src/clj/editor/scene_cache.clj
```diff
@@ -83,39 +83,23 @@
...
-
 (defn- consume-pending-deletions-in-object-caches [object-caches context]
   (reduce-kv
-    (fn [[object-caches pending-deletions] cache-id cache-meta]
-      (let [[cache-meta-without-pending-deletions pending-deletions-from-cache]
-            (consume-pending-deletions-in-cache-meta cache-meta context)]
-        (pair (assoc object-caches cache-id cache-meta-without-pending-deletions)
-              (into pending-deletions pending-deletions-from-cache))))
-    (pair (empty object-caches)
+    (fn [[object-caches pending-deletions :as acc] cache-id cache-meta]
```
matgis: We could defer unpacking `acc` until after the `(nil? cached-object+request-data-by-request-id)` check?

editor/src/clj/editor/scene_cache.clj
```diff
@@ -83,39 +83,23 @@
...
+            pending-deletions-from-cache (some-> cache-for-context meta :pending-deletions)]
+        (if (coll/empty? pending-deletions-from-cache)
+          acc
+          (pair (assoc object-caches
+                       cache-id
+                       (assoc cache-meta
+                              :caches
+                              (assoc caches-by-context
+                                     context
+                                     (vary-meta cache-for-context dissoc :pending-deletions))))
```
matgis: Style: `assoc` should be `[[:inner 0]]`. Let's update `.cljfmt.edn` while we're at it!

editor/src/clj/editor/ui.clj
```diff
@@ -2437,13 +2437,15 @@
...
+         unfocused-interval (max interval (long (* 1e9 (/ 1.0 15.0))))]
      {:last last
       :timer (proxy [AnimationTimer] []
                (handle [^long now]
                  (profiler/profile "timer" name
-                   (let [elapsed (- now start)
-                         delta (- now (long @last))]
+	                 (let [elapsed (- now start)
+	                       delta (- now (long @last))
+	                       interval (if (:focused @focus-state) interval unfocused-interval)]
```
matgis: Style: Indentation was messed up?

editor/src/clj/editor/ui.clj
```diff
@@ -2437,13 +2437,15 @@
          last (atom start)
          interval (if fps
                     (long (* 1e9 (/ 1 (double fps))))
-                    0)]
+                    0)
+         unfocused-interval (max interval (long (* 1e9 (/ 1.0 15.0))))]
```
matgis: Define constants for the two refresh rates?

matgis (dismissed):

PR #12687, opened 2026-07-03
Platform header cleanup

JCash:
This is a cleanup of our reliance of `*_vendor.h` headers.

### Technical changes

For these api's, we'll use a single public type definition.

Drawback is that the spin lock implementation isn't inlined anymore.
However I compared it to e.g. SDL, and they do the same.

Along with this, we need to make some cleanup on the private repos as well.

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


engine/dlib/src/dmsdk/dlib/spinlock.h
```diff
@@ -25,28 +25,46 @@
...
      *
      */
+    struct Spinlock
+    {
+        union
+        {
+            void*          m_Handle;
+            int32_atomic_t m_Lock;
+        };
+    };
```
Jhonnyg, line 49: Do we need to define the full struct here or can we hide it?

engine/dlib/src/dmsdk/dlib/file_descriptor.h
```diff
@@ -34,6 +27,13 @@
  */
 namespace dmFileDescriptor
 {
+    struct PollFD
+    {
+        int m_Fd;
+        int m_Events;
+        int m_REvents;
+    };
```
Jhonnyg, line 35: Do we need the full struct here or is a typedef enough and then hide it elsewhere?

Jhonnyg (approved):
LGTM, just two minor comments!

engine/dlib/src/dmsdk/dlib/spinlock.h
JCash: While I'd like to it's a bit dependent on how "breaking" we want it to be.

* We currently create the data on the stack, and then call `Init(&lock)`.
* There is no dynamic allocation (on our part)

Also, the fields aren't documented, which "says" they're private.
I feel fixing this type of api may be a later improvement

engine/dlib/src/dmsdk/dlib/file_descriptor.h
JCash: Same notes as above.
+ The typedefs was the part that was problematic with many different platform headers.

engine/dlib/src/dmsdk/dlib/spinlock.h
JCash: Hmm, I take that back. I'll see what I can do

JCash: I think changing this api is a separate task, for another time.

PR #12696, opened 2026-07-06
Jump to symbol dialog with fuzzy search

### LINKED ISSUE #12084: Shortcut to jump to symbol withint the same file
The new function list on the right sidebar is a great addition.
It would be helpful to have a shortcut to quickly jump to a symbol within the same file, with a fuzzy search, much like when you want to open a file.
This would reduce the amount of mouse motion (and strain) and ease the search when the number of functions becomes too high to be handled quickly in the sidebar.

Image

JosephFerano:
You can now press `Ctrl + Shift + O` on Windows/Linux and `Cmd + Shift + O` on Mac to open the jump-to-symbol dialog. You can fuzzy search symbols in the current script you are working on and press enter to jump to them, for quick navigation.

output

Fixes #12084


editor/src/clj/editor/code/view.clj
```diff
@@ -3088,6 +3094,76 @@
...
+  (run [view-node user-data open-resource-fn]
+    (g/with-auto-evaluation-context evaluation-context
+      (let [resource-node (get-property view-node :resource-node evaluation-context)
+            lsp (lsp/get-node-lsp (:basis evaluation-context) resource-node)
+            resource (g/node-value resource-node :resource evaluation-context)
+            localization (get-property view-node :localization evaluation-context)]
+        (if (lsp/has-language-servers-running-for-language? lsp (resource/language resource))
+          (let [document-symbols (get-property view-node :document-symbols evaluation-context)
+                items (mapv #(select-keys % [:name :kind :selection-range :detail :tags])
+                            document-symbols)
```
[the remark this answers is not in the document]
JosephFerano: Yes, I'm aware. That was my next task.

JosephFerano:
image

A few notes;

- This handles eliding super long function names manually because `TextFlow` does not support it.
- I right aligned the "details", if you prefer to keeping everything left aligned, I'll remove it
- I also replaced the word "function" with the unicode ƒ, both because I think it looks nice and to condense things. If you think it's too fancy, let me know and I'll remove that.
- Ctrl + Shift + O seems to be the default in VSCode and a couple of other editors
- I borrowed the VSCode UX for the most part, the UX I chose was to not make any jumps initially, as I find that jarring, VSCode has an initial jump to the nearest symbol, but I don't know if I like that. Instead here we wait for two things; as you start typing, it will jump to the first selection or if you use arrow keys, it will also jump to symbol. And it'll also return back if you cancel.
- The Structure Pane is nice because of the Tree structure, however, for a flat list, I ended up being more selective about what gets shown. It seems like the Lua LSP returns a lot of low value symbols as children, things like if statements and locals. Good for "structure" but it really pollutes the search space. Here I'll include the search space in VSCode, this is what would be shown to our users if we just flattened all children. I tested various scripts and I feel like I got all the important bits but if there's anything I missed, please let me know.
- I'm passing in the deprecated flag thing, but I don't know if LSP is passing deprecated into the tags set. I wasn't able to see it.

image

editor/src/clj/editor/code/view.clj
```diff
@@ -3088,6 +3093,94 @@
...
         (show-no-language-server-for-resource-language-notification! resource)))))
 
+(defn- flatten-document-symbols [document-symbols]
+  (letfn [(walk [acc {:keys [kind children] :as symbol}]
+            (let [acc (conj acc (dissoc symbol :children))]
+              (if (and (coll/not-empty children)
+                       (contains? #{:object :class :enum :struct :namespace} kind))
+                (reduce walk acc children)
+                acc)))]
+    (reduce walk [] document-symbols)))
```
vlaaad: Could we use `coll/flatten-xf` here?

editor/src/clj/editor/code/view.clj
```diff
@@ -3088,6 +3093,94 @@
                 (show-goto-popup! view-node open-resource-fn results)))))
         (show-no-language-server-for-resource-language-notification! resource)))))
 
+(defn- flatten-document-symbols [document-symbols]
+  (letfn [(walk [acc {:keys [kind children] :as symbol}]
+            (let [acc (conj acc (dissoc symbol :children))]
+              (if (and (coll/not-empty children)
+                       (contains? #{:object :class :enum :struct :namespace} kind))
```
vlaaad: Should we go through the whole tree?

editor/src/clj/editor/code/view.clj
```diff
@@ -3088,6 +3093,94 @@
...
+                   ;; filters; the dialog stays quiet on open so the cursor only
+                   ;; moves once the user engages.
+                   :preview-item-fn #(navigate-to-document-symbol! view-node %)
+                   :cell-fn (fn [{:keys [name kind detail tags] :as item} _localization]
+                              ;; The dialog is a fixed width, so a row must never grow wider than
+                              ;; it and cause a horizontal scrollbar. A very long name gets cut
+                              ;; short with a "…"; the signature (further down) trims the same way.
+                              (let [max-name-length 56
+                                    indices (:matching-indices (meta item))
+                                    elide? (> (count name) max-name-length)
```
vlaaad: `?` suffix on a boolean

editor/src/clj/editor/dialogs.clj
```diff
@@ -610,7 +639,10 @@
              :props {:selection-mode selection
                      :on-selected-indices-changed {:event-type :set-selected-indices}}
              :desc {:fx/type fx/ext-on-instance-lifecycle
-                    :on-created select-first-list-item-on-items-changed!
+                    :on-created (fn [view]
+                                  (select-first-list-item-on-items-changed! view)
+                                  (when preview-item-fn
+                                    (install-selection-preview! preview-item-fn preview-on-items-changed? view)))
```
vlaaad: I would rather define a binding prop here, see `fx/make-binding-prop` uses. It's better to keep `:on-created` a stable fn ref, not a closure, since `select-list-dialog` is used as a cljfx component — currently, if we change `preview-item-fn` or `preview-on-items-changed?` values, the changes will not be picked up.

editor/src/clj/editor/dialogs.clj
```diff
@@ -725,7 +757,13 @@
...
       :prompt         filter text field's prompt, string or MessagePattern,
                       defaults to \"dialog.select-item.prompt\" message
-      :owner          the owner window, defaults to main stage"
+      :owner          the owner window, defaults to main stage
+      :preview-item-fn
+                      optional side-effecting fn of one arg, called with the
+                      item to preview when the list gains focus, its focused
+                      item changes, or filtering changes the item list while
+                      the filter term is non-empty; when opened without an
+                      initial filter, it is not called during setup"
```
vlaaad: Also document `:preview-on-items-changed?` option?

Also, to me, it seems we introduced `preview-on-items-changed?` only for a single specific use case: skip preview on dialog open, when we have no text and we get our first selection. Maybe we could get rid of this fn completely? For example, by passing an additional arg to preview-fn, something like "preview-source" with values like `:opened`, `:focus-changed`? Just feels that current semantics of `preview-item-fn` and `preview-on-items-changed?` are too... I don't know... too oddly specific for the use case?

vlaaad (commented):
Great work! Feels great to use! See some comments about our dialogs changes.

editor/src/clj/editor/code/view.clj
JosephFerano: I used `coll/tree-xf` instead, not sure if it's much better.

editor/src/clj/editor/code/view.clj
JosephFerano: I'm not sure I understand your question. My intent was to prune and avoid too much noise, as I pointed out in the PR comment above.

Is there a scenario you're thinking of where we're missing a symbol that can be shown but isn't? Are you saying that rather than skipping recursion into certain children, we should still explore all children in case we find something we can use? In the case of functions, I found that it's hard to distinguish between genuinely useful symbols and noise, so that's why I was more aggressive with the early prunning and gave preferential treatment to top level symbols, which is what I like to use this sort of tool for in my editor; jumps to significant symbols.

Let me know what your thoughts are.

editor/src/clj/editor/dialogs.clj
```diff
@@ -592,37 +592,11 @@
...
-                    (changed [_ _ _ focused]
-                      (when focused
-                        (preview! (.getFocusedItem focus-model))))))
-    (.addListener (.getItems view)
-                  (reify ListChangeListener
-                    (onChanged [_ _]
-                      (when (preview-on-items-changed?)
-                        (preview! (first (.getItems view)))))))))
+(def ^:private ext-with-selection-preview-prop
+  (fx/make-ext-with-props {:selection-preview fxui/list-view-selection-preview-prop}))
```
vlaaad: No need for this, `fxui/list-view-selection-preview-prop` can be used as a key, i.e., as a prop, directly. Also `list-view-selection-preview-prop` should be defined here since it belongs to select list dialog

editor/src/clj/editor/code/view.clj
vlaaad: Hmm, maybe I misunderstood something here... I somehow thought that the idea was to take the whole tree, but only limit it to items of a particular type, but I think it works well now, so feel free to disregard my comment :)

editor/src/clj/editor/code/view.clj
```diff
@@ -3094,20 +3094,20 @@
...
 (defn- flatten-document-symbols [document-symbols]
-  (letfn [(walk [acc {:keys [kind children] :as symbol}]
-            (let [acc (conj acc (dissoc symbol :children))]
-              (if (and (coll/not-empty children)
-                       (contains? #{:object :class :enum :struct :namespace} kind))
-                (reduce walk acc children)
-                acc)))]
-    (reduce walk [] document-symbols)))
+  (coll/into-> document-symbols []
+    (comp (coll/tree-xf
```
vlaaad: I think you don't need `comp` when you use `coll/into->`

vlaaad (dismissed):
LGTM!

vlaaad (approved):
LGTM! Don't forget to update the PR description with user-facing notes/screenshots :)

AGulev:
Should we also add info about this shortcut here maybe? 
CleanShot 2026-07-07 at 13 11 34@2x


and also in doc?

vlaaad:
@AGulev, not sure it belongs there, it's not as fundamental to editor use as opening assets or building...

AGulev:
@vlaaad you're right
but would be nice to update docs maybe (we should have some manuals where it fits there, right?)

JosephFerano:
@AGulev Yes, I will create a PR now for some docs

PR #12706, opened 2026-07-08
Keep toolbar when moving focus from Outline to Toolbar dropdown

### LINKED ISSUE #12674: Scene menu  dissapear when click on it for the first time
https://github.com/user-attachments/assets/7edbeed1-7cf3-42e6-b634-f7dc6e9a05b1

vlaaad:
This fixes an issue where the scene view toolbar disappeared when focus changed from Outline to a dropdown in a toolbar.

Technical notes:

The problem is that after we showed the toolbar dropdown, it refreshed, recreating the nodes, which caused the focused node to lose its parent, which caused no selection, which caused the scrollbar to disappear. This PR changes the toolbar impl to use cljfx, which preserves the nodes on updates. There are a lot of changes, but most of them are mechanical due to a need to move some code from fxui to ui ns.

Fix #12674


editor/src/clj/editor/ui.clj
```diff
@@ -2132,140 +2135,243 @@
...
+             (if (neg? i)
+               acc
+               (let [def-name (if (zero? i) name (symbol (str name "$phase-" i)))]
+                 (recur
+                   (dec i)
+                   def-name
+                   (let [ext (compose i)]
+                     (conj acc
+                           `(defn ~def-name [~'props]
+                              ~(assoc ext :desc `(assoc ~'props :fx/type ~acc-name)))))))))))))
```
vlaaad, line 2229: This code simply moved here from fxui ns.

editor/src/clj/editor/ui.clj
```diff
@@ -2132,140 +2135,243 @@
...
+(defn advance-graph-user-data-component! [view-node key desc]
+  (advance-user-data-component! view-node g/user-data g/user-data! key desc))
+
+(defn advance-ui-user-data-component! [javafx-node key desc]
+  (advance-user-data-component! javafx-node user-data user-data! key desc))
+
+(def ^:private prop-toolbar-choice-box-value-changed
+  (fx/make-binding-prop
+    (fn bind-toolbar-choice-box-value-changed [^ChoiceBox choice-box on-value-changed]
+      (let [^ChangeListener listener #(when %3 (on-value-changed choice-box %3))]
```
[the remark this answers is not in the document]
vlaaad: No, it doesn't — cljfx skips invoking callbacks during advance

editor/src/clj/editor/ui.clj
```diff
@@ -2312,17 +2418,14 @@
...
 (defn- refresh-toolbars!
   [^Scene scene localization evaluation-context]
-  (let [visible-command-contexts (visible-command-contexts scene evaluation-context)
-        current-command-contexts (current-command-contexts scene evaluation-context)
+  (let [command-contexts (current-command-contexts scene evaluation-context)
         root (.getRoot scene)
-        app-view (-> current-command-contexts first :env :app-view)
-        active-tab (g/maybe-node-value app-view :active-tab evaluation-context)]
+        app-view (-> command-contexts first :env :app-view)
+        ^Tab active-tab (g/maybe-node-value app-view :active-tab evaluation-context)]
```
JosephFerano, line 2424: Nitpick: Is there a style preference as to where the type hint goes? For let bindings I've usually put them sandwiched. There's another one above like this in `refresh-toolbar`.

JosephFerano (approved):
One thing I noticed is that when I use the color dropper in both the previous version and this commit, the scene toolbar still disappears. I guess this is something I can investigate in the color picker branch, I'm guessing it has to do with the `refresh-toolbars!` check where it's checking which is the active-tab?

editor/src/clj/editor/ui.clj
vlaaad: Hmm, AFAIK we don't have a preference, but we could introduce it.

PR #12710, opened 2026-07-09
Improve toolbar refresh performance

vlaaad:
After implementing #12706, I decided to measure the performance of toolbar updates. 

Baseline before cljfx: 0.86-0.96 ms per noop update, 4.10-5.88ms per modifying update
After #12706: 1.46-1.61ms per noop update, 5.64-6.95 per modifying update.

Here is how the refresh toolbars CPU profile looked after cljfx migration:
Screenshot 2026-07-09 at 10 05 22
cljfx is a big part, but so are Node lookups in scene visibility refresh

So, using cljfx introduced a perf regression. This PR fixes it, with the following numbers: ~0.68 ms per noop update, 2.29-2.47ms per modifying update. Here is the new profile:

Screenshot 2026-07-09 at 10 08 03
Node lookups disappeared, and cljfx became a smaller piece of the pie too.

### DIFF
--- a/editor/project.clj
+++ b/editor/project.clj
@@ -88,7 +88,7 @@
 
                      [com.github.ben-manes.caffeine/caffeine "3.1.2"]
 
-                     [cljfx "1.10.8"
+                     [cljfx "1.10.9"
                       :exclusions [org.clojure/clojure
                                    org.openjfx/javafx-base
                                    org.openjfx/javafx-graphics
--- a/editor/src/clj/editor/ui.clj
+++ b/editor/src/clj/editor/ui.clj
@@ -57,7 +57,7 @@
            [java.awt Desktop Desktop$Action]
            [java.io File IOException]
            [java.net URI]
-           [java.util Collection]
+           [java.util ArrayDeque Collection]
            [javafx.animation AnimationTimer KeyFrame KeyValue Timeline]
            [javafx.application Platform]
            [javafx.beans InvalidationListener Observable]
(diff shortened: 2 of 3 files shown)


editor/project.clj
```diff
@@ -88,7 +88,7 @@
 
                      [com.github.ben-manes.caffeine/caffeine "3.1.2"]
 
-                     [cljfx "1.10.8"
+                     [cljfx "1.10.9"
```
vlaaad, line 91: I went through a lot of cljfx code to do some optimizations; it's all mechanical, e.g., replace metadata-based protocol impls with reifications of protocol impls:
https://github.com/cljfx/cljfx/commit/d6beef6144c6f3633b0e350ff0359348fbb263c3

editor/src/clj/editor/ui.clj
```diff
@@ -286,8 +286,27 @@
            (.initOwner stage owner)))
      stage)))
 
+(defn lookup-by-id
+  "Find node by id, significantly faster than Node/.lookup"
```
vlaaad, line 290: 9.8x faster, or 89% improvement:
```clj
(.lookup root "#visibility-settings-graphic")
;; mean 4.230833 us

(lookup-by-id root "visibility-settings-graphic")
;; mean 431.785282 ns
```

JosephFerano (approved):
I profiled with `clj-async-profiler` (it finally works for me!) and got similar flamegraph shapes to yours. Would it make sense to at some point do a pass over all uses of .lookup or maybe add it as some kind of skill or clj-kondo lint so that we can update usages of .lookup whenever we touch them again?

vlaaad:
I had a brief look at our uses of .lookup, many of these used css classes instead of css IDs, but I think we could still try to add a lint, yes!

PR #12715, opened 2026-07-09
Add `/preview` endpoint that returns scenes as PNGs

### LINKED ISSUE #12690: Add a way to preview files as PNGs for Agents
Agents create popups by editing files on disk in our protobuf text format.
The only way to validate how they look for the agent is by running the game.

Maybe we can provide a way to generate a preview as an image right from the editor, similar to what we do here (see screenshot), but without outlines—or with outlines as an optional setting. (edited)

Image

vlaaad:
Now, the editor's HTTP server can return PNGs of scenes in the project. This is useful for agentic workflows that need to see what the scene looks like when an agent edits it. As usual, the endpoint is documented in the OpenAPI spec, so the only thing your agent needs to use it is still the URL `http://localhost:$(cat .internal/editor.port)/openapi.json`.

Fix #12690

### DIFF
--- a/editor/src/clj/editor/camera.clj
+++ b/editor/src/clj/editor/camera.clj
@@ -45,6 +45,8 @@
 (def fov-x-35mm-full-frame 54.4)
 (def fov-y-35mm-full-frame 37.8)
 
+(def ^:const orthographic-framing-margin 1.1)
+
 (def vector3-up (Vector3d. 0.0 1.0 0.0))
 
 (defn camera-forward-vector
@@ -547,14 +549,14 @@
               factor-y (/ proj-height h)
               fov-x-prim  (* factor-x (.fov-x camera))
               fov-y-prim  (* factor-y (.fov-y camera))]
-          [(* 1.1 fov-x-prim) (* 1.1 fov-y-prim)])))))
+          [(* orthographic-framing-margin fov-x-prim) (* orthographic-framing-margin fov-y-prim)])))))
 
 (defn camera-orthographic-frame-aabb
   ^Camera [^Camera camera ^Region viewport ^AABB aabb]
   {:pre [(= :orthographic (:type camera))]}
   (let [aspect (/ ^double (:fov-x camera) ^double (:fov-y camera))
         [^double fov-x ^double fov-y] (camera-orthographic-fov-from-aabb camera viewport aabb)
-        [fov-x fov-y] (if (> (/ fov-x aspect) (* aspect fov-y))
+        [fov-x fov-y] (if (> (/ fov-x aspect) fov-y)
                         [fov-x (/ fov-x aspect)]
                         [(* aspect fov-y) fov-y])
         filter-fn (or (:filter-fn camera) identity)]
--- a/editor/src/clj/editor/scene.clj
+++ b/editor/src/clj/editor/scene.clj
@@ -20,10 +20,12 @@
             [cljfx.fx.text-area :as fx.text-area]
             [clojure.set :as set]
             [clojure.spec.alpha :as s]
+            [clojure.string :as string]
             [dynamo.graph :as g]
             [editor.background :as background]
             [editor.camera :as c]
             [editor.colors :as colors]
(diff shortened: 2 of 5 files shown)


editor/src/clj/editor/camera.clj
```diff
@@ -547,14 +549,14 @@
...
-          [(* 1.1 fov-x-prim) (* 1.1 fov-y-prim)])))))
+          [(* orthographic-framing-margin fov-x-prim) (* orthographic-framing-margin fov-y-prim)])))))
 
 (defn camera-orthographic-frame-aabb
   ^Camera [^Camera camera ^Region viewport ^AABB aabb]
   {:pre [(= :orthographic (:type camera))]}
   (let [aspect (/ ^double (:fov-x camera) ^double (:fov-y camera))
         [^double fov-x ^double fov-y] (camera-orthographic-fov-from-aabb camera viewport aabb)
-        [fov-x fov-y] (if (> (/ fov-x aspect) (* aspect fov-y))
+        [fov-x fov-y] (if (> (/ fov-x aspect) fov-y)
```
vlaaad, line 559: Fixed a frame rect calculation, seems it was like this since 2016. It unnecessarily zoomed in and clipped some wide scenes.

editor/src/clj/editor/scene.clj
```diff
@@ -2030,6 +2052,13 @@
         (scene-cache/prune-context! gl)
         buf-image))))
 
+(g/defnk produce-preview-renderables [aux-render-data scene-render-data]
+  (:renderables
+    (merge-render-datas
+      {:renderables (select-keys (:renderables aux-render-data) [pass/background])}
```
vlaaad, line 2058: Previews had black backgrounds because they ignored all aux renderables. I now render the background.

editor/src/clj/editor/scene.clj
```diff
@@ -2136,7 +2165,8 @@
                   (g/connect view-id         :scene-aabb                    camera          :scene-aabb)
                   (g/connect view-id         :viewport                      camera          :viewport)
 
-                  (g/connect app-view-id     :selected-node-ids             view-id         :selection)
+                  (when (:inherit-selection opts true)
+                    (g/connect app-view-id   :selected-node-ids             view-id         :selection))
```
vlaaad, line 2169: Kinda funny, but initial impl would generate previews for selected nodes, which is unexpected if you open a scene and select something small...

editor/src/clj/editor/scene.clj
```diff
@@ -2191,15 +2221,24 @@
...
-(defn dispose-preview [node-id]
-  (when-some [^GLAutoDrawable drawable (g/node-value node-id :drawable)]
-    (gl/with-drawable-as-current drawable
-      (scene-cache/drop-context! gl))
-    (.destroy drawable)
-    (g/set-property! node-id :drawable nil)))
+(defn dispose-preview
+  ([node-id]
+   (g/with-auto-evaluation-context evaluation-context
+     (dispose-preview node-id evaluation-context)))
```
vlaaad, line 2230: I thought it was okay to add an arity in this case since `dispose-preview` doesn't seem to imply ec...

vlaaad:
@codex review

editor/src/clj/editor/scene.clj
```diff
@@ -2334,3 +2373,72 @@
...
+(defn- render-preview-response [project app-view request]
+  (g/let-ec [params (coll/into-> (string/split (:query request "") #"&") {}
+                      (map
+                        (fn [s]
+                          (let [[k v] (string/split s #"=" 2)]
+                            [(keyword k) (or v "")]))))
+             width (or (some-> (:width params) parse-long)
+                       (g/node-value project :display-width evaluation-context))
+             height (or (some-> (:height params) parse-long)
+                        (g/node-value project :display-height evaluation-context))
```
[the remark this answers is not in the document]
vlaaad: `parse-long` does not throw; it returns nil on non-int strings.

matgis (approved):

PR #12716, opened 2026-07-09
Close Add Component context menu on selecting a component

### LINKED ISSUE #12638: Context menu not disappearing after selecting a component
The context menu was always disappearing as soon as you click a component to add it to the game object. It is now not disappearing and you need to manually click outside the context menu to close it.

**To Reproduce (REQUIRED)**
1. In main.collection add a game object - the context menu disappears, when you click `Add Game Object` and the object is added.
2. Right click on the game object and click `Add Component` and select any component to add - click on it and the menu won't be closed.

Context menu closes on component selection.

  - 1.13.1 alpha (dev)

**Platforms (REQUIRED):**
 - Linux
 - OS: Ubuntu 24.04
 - Device: Acer Nitro 16

https://github.com/user-attachments/assets/d12a7c98-6fc4-4e7f-b9bd-2e86eab06f52

vlaaad:
This fixes an issue where the Add Component context menu remained open after selecting a component in the scene view.

Fix #12638

### DIFF
--- a/editor/src/clj/editor/ui.clj
+++ b/editor/src/clj/editor/ui.clj
@@ -1620,7 +1620,9 @@
       (some-> content
               (.lookup ".grid-menu-item-enabled")
               (.requestFocus)))))
-
+;; NOTE: make-grid-menu sets :hide-on-click to false because a CustomMenuItem can have headers and
+;; empty space that we don't want dismissing the context menu when clicked on. So manually walk up
+;; the PopupWindow and hide them
 (defn- hide-popup-window-chain! [^Event event]
   (let [node ^Node (.getSource event)]
     (loop [window (some-> node .getScene .getWindow)]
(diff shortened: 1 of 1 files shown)


editor/src/clj/editor/ui.clj
```diff
@@ -1621,6 +1621,13 @@
               (.lookup ".grid-menu-item-enabled")
               (.requestFocus)))))
```
JosephFerano:
```suggestion
;; NOTE: make-grid-menu sets :hide-on-click to false because a CustomMenuItem can have headers and
;; empty space that we don't want dismissing the context menu when clicked on. So manually walk up
;; the PopupWindow and hide them
```

JosephFerano (dismissed):
I investigated this a bit to understand what the underlying issue was, and looks like the fix is the right one. I added a NOTE that explains the _why_ this function is needed, which I think would be nice to add.

PR #12718, opened 2026-07-10
Fix dependency fetch timeout issue

### LINKED ISSUE #12603: Fetch Libraries fails with "Unreachable host" error for GitHub and custom hosts
When attempting to use the "Fetch Libraries" feature in a project, the editor intermittently throws an error stating that the fetch was aborted because a host could not be reached. 

This happens even though the internet connection is perfectly stable and active. For instance, the target hosts (such as `github.com` or custom private servers) are fully accessible via a web browser at the exact same moment.

**To Reproduce (REQUIRED)**
1. Open a Defold project with dependencies configured in `game.project`.
2. Click on 'Project' -> 'Fetch Libraries'.
3. Wait for the process to start.
4. See the "Fetch Failed" error dialog with "Fetch was aborted because a host could not be reached".

The editor should successfully connect to the hosts, download the library archives, and update dependencies without network connection false-positives.

 - Version 1.12.4

**Platforms (REQUIRED):**
 - OS: Windows 11
 - Device:  Desktop PC

The issue is environmental or network-handling related within the editor rather than project-specific, but it occurs on projects utilizing external dependencies.

**Workaround (OPTIONAL)**
No reliable workaround inside the editor. Repeatedly triggering 'Fetch Libraries' or restarting the editor sometimes helps.

**Screenshots (OPTIONAL)**
Image
Image

**Additional context (OPTIONAL)**
The issue doesn't seem to be tied to a specific domain: it reproduces both with standard GitHub URLs (`github.com`) and custom self-hosted endpoints. Browser access to these domains works flawlessly while Defold claims they are unreachable.

vlaaad:
In projects with many dependencies it is sometimes observed that fetching dependencies timeout for no reason, even when the internet connection is both stable and fast. This change increases the connection timeout to avoid timeouts.

Fix #12603


AGulev (dismissed):

vlaaad:
@AGulev, this is the version you were testing (except it also had more debug logs)

vlaaad:
(I forgot to stage 3->4 fetches per host when committing)

AGulev (approved):

PR #12720, opened 2026-07-10
Implement HTTP file download in editor scripts

vlaaad:
Now, you can download files using the `path` option of the `http.request` fn:

```lua
local response = http.request(url, {
    path = "downloads/archive.zip"
})
```


Insality:
The best!

Before I had to pack the zip inside JSON to be able to download assets in the asset store :D

vlaaad:
@Insality if only someone expressed the demand by creating an issue...

JosephFerano (approved):
Image

It would be really nice if there was some sort of progress update, I tried downloading the editor which is 300+ MB and wasn't sure if things were working, until I opened the file explorer to see the tmp file in progress.

PR #12725, opened 2026-07-13
Make sure async load of a collection with more than 1024 resources do not block the main thread

### LINKED ISSUE #12684: Increase MAX_PRELOADER_REQUESTS limit — async_load stalls game thread when collection exceeds 1024 preloader nodes
**Description**

When a collection proxy is loaded via `async_load`, the engine builds a preloader tree limited to `MAX_PRELOADER_REQUESTS = 1024` nodes (`engine/resource/src/resource_preloader.cpp:89`). From what our investigation showed, any resource that does not fit in the tree is silently dropped from async loading and instead loaded **synchronously on the main thread** during `ResCollectionCreate`, causing a hard frame freeze proportional to the number of overflow resources. There is no log warning, no error, and no API for the caller to detect or avoid this, as far as we're aware.

In a production mobile game with a collection containing **4292 resources**, this caused a **~900 ms main-thread freeze** measured via the Defold profiler:

| Profile scope | Calls | Total ms | ms/call |
|---|---|---|---|
| `collectionproxyc.UpdatePreloader` | 2 | 1783 | **891** |
| `UpdatePreloader.PreloaderUpdateOneItem` | 23 | 1779 | — |
| `DoPreloaderUpdateOneReq.CreateAndLoadResource` | **4292** | 1768 | **0.41** |
| `CreateAndLoadResource.LoadResourceToBufferWithOffset` | 4292 | 880 | 0.20 |
| `LoadResourceToBufferWithOffset.GetRaw` | 4273 | 869 | **0.20** |

4273 archive reads (`GetRaw`) completed synchronously in 2 frames, all on the main thread. The entire `~900 ms` was user-visible jank.

**Root cause, speculative**

`resource_preloader.cpp:337` — when the 1024-slot tree is full, `PreloadHint` returns `RESULT_OUT_OF_MEMORY` and the hint is silently discarded:

```cpp
if (!preloader->m_FreelistSize)
{
    // Preload queue is exhausted; this is not fatal, it just means the resource will be loaded
    // inside the main thread which may cause stuttering
    return RESULT_OUT_OF_MEMORY;
}
```

The per-frame 10 ms budget in `comp_collection_proxy.cpp:325` is irrelevant for overflow resources: the synchronous loads happen **inside** a single `DoPreloaderUpdateOneReq` → `ResCollectionCreate` → `dmResource::Get()` chain, within one iteration of the `do...while` loop, so the budget check never fires until the entire burst has already completed.

**Full call chain:**

```
UpdatePreloader  (10 ms budget checked between iterations)
  └─ PreloaderUpdateOneItem
       └─ DoPreloaderUpdateOneReq         ← one iteration, no budget check inside
            └─ FinishLoad
                 └─ CreateResource
                      └─ ResCollectionCreate
                           └─ dmResource::Get  (× N_overflow, synchronous)
                                └─ CreateAndLoadResource
                                     └─ LoadResourceToBufferWithOffset
                                          └─ GetRaw  (~0.2 ms per resource, main thread)
```

**Reproduction**

A minimal self-contained reproduction project: [defold-preloader-demo.zip](https://github.com/user-attachments/files/29604784/defold-preloader-demo.zip).

*Note*: The `resources/` directory is not included in the repo. Run `python3 generate.py` once before opening the project in the editor — it creates the 18,073 resource files that trigger the overflow.

It generates a collection with 3012 game objects × 5 unique script components = 18,073 preloader tree nodes, overflowing the 1024-node cap by 17,049 nodes. On a Samsung Galaxy A35 (fast NVMe-backed APK reads), the measured freeze is **~190 ms in a single frame**:

```
GOs in collection       : 3012
Scripts per GO          : 5  (each is a distinct tree node)
Preloader nodes needed  : 18073  (1 root + 3012 GOs + 15060 scripts)
MAX_PRELOADER_REQUESTS  : 1024  (resource_preloader.cpp:89)
Overflow nodes          : 17049  (loaded synchronously on main thread)
Sending async_load ...
--- proxy_loaded ---
  Wall time              : 355.1 ms
  Frames elapsed         : 13
  Wall-clock intervals between consecutive update() calls (socket.gettime()):
    after frame  1:    15.2 ms
    after frame  2:    16.7 ms
    ...
    after frame 12:   193.7 ms  <-- SYNC BURST (ResCollectionCreate)
  Finalisation frame     : 3.4 ms
  RESULT: BUG REPRODUCED
  Single-frame sync burst: 193.7 ms
  ResCollectionCreate loaded 17049 overflow nodes synchronously
```

---

**Possible solution:**

**Option A (minimal change):** Increase `MAX_PRELOADER_REQUESTS` to a larger value, e.g. 4096 or 8192.

```cpp
// engine/resource/src/resource_preloader.cpp:89
// Before
static const uint32_t MAX_PRELOADER_REQUESTS = 1024;

// After
static const uint32_t MAX_PRELOADER_REQUESTS = 4096;
```

**Option B (expose via `game.project`):** Make the limit configurable without a rebuild, so projects can tune it to their content:

```ini
[resource]
max_preloader_requests = 4096
```

**In addition, it would be nice to add a warning log when overflow occurs:** At minimum, emit a `dmLogWarning` at the `RESULT_OUT_OF_MEMORY` return site so developers are informed at runtime that synchronous loading will occur. Currently the overflow is completely silent — there is no way to know it is happening without a profiler.

---

**Additional context**

- Engine version tested: Defold 1.12.x
- `async_load` is explicitly documented as the non-blocking alternative to `load`; the silent synchronous fallback violates that contract from the caller's perspective

britzl:
This change makes sure that async load of a collection with more than 1024 resources do not block the main thread.

Fixes #12684


britzl:
@codex review

britzl:
@codex review

engine/resource/src/resource_preloader.cpp
```diff
@@ -377,21 +516,22 @@ namespace dmResource
...
             {
-                ++new_hint_count;
+                continue;
             }
+            PushPendingHint(preloader, hint);
+            preloader->m_Request[hint.m_Parent].m_QueuedChildCount++;
         }
-        return new_hint_count != 0;
+        return AdmitPendingHint(preloader);
     }
```
AGulev, line 578: Bulk hint transfer is quadratic and bypasses the time budget. Each new hint linearly scans all pending hints, and PopHints() processes the entire batch inside one update iteration. A resource with 300 direct dependencies performs roughly 45,000 comparisons before the soft-time-limit check. Consider using indexed deduplication or processing the batch incrementally.

AGulev (changes requested):
See my comment + my PR into your pr: https://github.com/defold/defold/pull/12821 - in case we have recursion which isn't in the initial 1024 resources - it will not be detected

engine/resource/src/resource_preloader.cpp
britzl: Good catch! Fixed it.

engine/resource/src/resource_preloader.cpp
```diff
@@ -437,13 +646,63 @@ namespace dmResource
...
+                }
+                RetainedResource retained = { req->m_Resource, req->m_Parent };
+                preloader->m_RetainedResources.Push(retained);
+            }
+            req->m_Resource = 0;
+        }
+        PreloaderRemoveLeaf(preloader, index);
+    }
+
+    static void ReleaseRetainedResources(HPreloader preloader, TRequestIndex parent)
```
AGulev, line 673: I think we have a similar issue here to the one we fixed earlier.
`ReleaseRetainedResources()` searches the entire `m_RetainedResources` array every time a parent finishes, even though it only needs the resources belonging to that one parent.
With many retained resources this becomes:
parent A scans 1000 entries
parent B scans 1000 entries
parent C scans 1000 entries
...even if most of those entries belong to other parents.
Overall this can become quadratic for large collections.
I think we should store retained resources per parent (or otherwise index them by parent), so releasing them becomes O(number of that parent's children) instead of scanning the whole array every time. It's similar to the optimisation we made previously where we replaced repeated global searches with direct indexing.

engine/resource/src/resource_preloader.cpp
AGulev: The code limits how many hints are added to the request tree per frame (64), but it still processes every new hint in one go before that limit is applied.
If a preload callback suddenly returns 5,000 hints, we still have to process all 5,000 in a single update. Only after that do we start adding them to the tree 64 at a time.
That means we can still get a long frame, because the expensive work happens before the limit.
I think the limit should be applied earlier: process only a small number of new hints each update as well, so the work is spread across multiple frames.

PR #12732, opened 2026-07-14
Add pipeline support for loading .hdr files

Jhonnyg:
First deliverable of supporting .hdr files that can be used in the engine. 

#### Technical changes

* Added HDR .hdr texture pipeline support in Bob, including .hdr as a TextureBuilder input extension.
* Added texc float texture formats for RGBA16F and RGBA32F, with generated Java enum support.
* Added stb-backed HDR detection/loading in a dedicated texc_stbi.cpp module, exposed through texc/JNI as IsHDR and LoadHDR.
* Updated TextureGenerator to detect HDR input from bytes, load HDR data via texc, then use the normal CreateImage flow for texture generation.
* Added HDR texture generation paths for float texture profiles, defaulting HDR inputs to RGBA32F when no profile is provided.
* Added texc support for resizing and flipping RGBA32F image data.
* Added default/uncompressed encoding support for RGBA16F and RGBA32F texture outputs.
* Updated texture compression handling so uncompressed float texture formats can be emitted through the existing compressor path.


engine/texc/src/java/com/dynamo/bob/pipeline/Texc.java
```diff
@@ -54,7 +54,9 @@ public enum PixelFormat {
         PF_RGBA_ASTC_10x8(28),
         PF_RGBA_ASTC_10x10(29),
         PF_RGBA_ASTC_12x10(30),
-        PF_RGBA_ASTC_12x12(31);
+        PF_RGBA_ASTC_12x12(31),
+        PF_RGBA16F(32),
+        PF_RGBA32F(33);
```
Jhonnyg, line 59: We'll start with these and then add more formats later when we have everything else in place.

engine/texc/src/texc.h
Jhonnyg: I don't think this is true? 🤔  this table doesn't have all of the formats the graphics_ddf.proto one has.

engine/texc/src/texc_stbi.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <string.h>
+
+#include <dlib/log.h>
+
+#define STB_IMAGE_STATIC
+#define STBI_ONLY_HDR
+#define STBI_NO_STDIO
+#define STBI_NO_THREAD_LOCALS
+#define STB_IMAGE_IMPLEMENTATION
+#include "../../dlib/src/stb/stb_image.h"
```
Jhonnyg: texc now uses stbi for the HDR loading part of the pipeline support. We already use stbi for other things, and there is no built-in .hdr loading in java.

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/TextureGenerator.java
```diff
@@ -64,6 +65,35 @@ public static class GenerateResult {
...
 
+    private static class DecodedImage {
+        public String path;
+        public int width;
+        public int height;
+        public int componentCount;
+        public int texcPixelFormat;
+        public int texcColorSpace;
+        public byte[] data;
+    }
```
Jhonnyg, line 77: There's quite a lot of changes in this file so the generate flow is more data driven so that we can treat HDR files a bit different from 8 bit inputs.

engine/texc/src/texc.h
JCash: Possibly not true (anymore?)
remove the comment

engine/texc/src/texc_stbi.cpp
JCash: Directly Including a file from another library folder , is not the way we should do it.
I see some options:
1) support loading HDR images in our dmImage namespace (thus using dlib)
2) having a duplicate stb_image.h in this folder, only for this use case

engine/texc/src/texc_stbi.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+        if (!image->m_Path || !image->m_Data)
+        {
+            stbi_image_free(hdr_data);
+            DestroyLoadedImage(image);
+            return false;
+        }
+
+        memcpy(image->m_Data, hdr_data, (size_t)data_count);
+        stbi_image_free(hdr_data);
```
JCash: can we use `hdr_data` as-is?
The caller owns it, and it seems to be what we want, so do we need to make a full copy?

JCash (changes requested):
Overall good.
I'd like to avoid the direct include dependency on dlib though.
Also, I think we should support compressed HDR textures via ASTC
I feel it's best to do it while the PR is "warm" and it's in the current context.

(BasisU may not support it until next version, so that can wait)

engine/dlib/src/dlib/image.h
```diff
@@ -82,6 +94,9 @@ namespace dmImage
             return 1;
         case dmImage::TYPE_LUMINANCE_ALPHA:
             return 2;
+        case dmImage::TYPE_RGBA32F:
+            // Float images are not exposed through the byte-oriented image consumers yet.
+            return 0;
```
Jhonnyg, line 99: The first deliverable is for the pipeline only, this will change in the next deliverable where we add the engine pieces.

Jhonnyg:
@codex review

JCash (approved):

PR #12734, opened 2026-07-14
Fix collection proxy load stalls when a collection has more than 1024 resources

### LINKED ISSUE #12684: Increase MAX_PRELOADER_REQUESTS limit — async_load stalls game thread when collection exceeds 1024 preloader nodes

britzl:
This change fixes a long standing issue when loading collections with more than 1024 resources. When such a collection is loaded asynchronously only 1024 resources will be loaded async over multiple frames. Remaining resources will be loaded on the main thread in a single frame which causes a noticeable "hitch". This change makes sure the resource preloader will allocate the exact amount of space required to load the resources needed. No more, no less.

Fixes #12684

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/graph/ResourceCounter.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Set;
+
+import com.dynamo.bob.CompileExceptionError;
+import com.dynamo.bob.Project;
+import com.dynamo.bob.fs.IResource;
+import com.dynamo.bob.pipeline.graph.ResourceWalker.IResourceVisitor;
+
+import com.google.protobuf.Message;
+
+public class ResourceCounter implements IResourceVisitor {
```
britzl, line 32: A specialized resource visitor with the sole purpose of counting resources as a resource walker traverses a resource graph

AGulev:
@codex review

AGulev:
Current apprach assume the loader deduplicate resource nodes in runtime BEFORE loading starts, but it is actually does it while loading is in progress. 

So the three
-R
--A.go
---A.script
--B.go
---A.script

in java end up couning it as 4 nodes needed
but in runtime it will require 5 nodes to be created 

It's not a directly related issue, but same nature https://github.com/defold/defold/issues/3736

AGulev:
When you build tree you count all the resource types, but I don't see any gates which will not count `load_dynamically` factories, `load_dynamically` collection factories, and internal collection proxies. Maybe I missunderstood code, but it seems like it will just count all its resounces in the root collection without takene these into account

AGulev (changes requested):
See my comments above

britzl:
> When you build tree you count all the resource types, but I don't see any gates which will not count `load_dynamically` factories, `load_dynamically` collection factories, and internal collection proxies. Maybe I missunderstood code, but it seems like it will just count all its resounces in the root collection without takene these into account

Good point. I've adressed this.

engine/resource/src/resource_preloader.cpp
```diff
@@ -446,16 +487,22 @@ namespace dmResource
         assert(req->m_PendingChildCount == 0);
     }
 
-    HPreloader NewPreloader(HFactory factory, const dmArray<const char*>& names)
+    HPreloader NewPreloader(HFactory factory, const dmArray<const char*>& names, uint32_t request_count)
     {
-        ResourcePreloader* preloader = new ResourcePreloader();
-        // root is always allocated so we don't add index zero in the free list
-        for (uint32_t i = 0; i < MAX_PRELOADER_REQUESTS - 1; i++)
+        dmLogInfo("NewPreloader() %d", request_count);
```
AGulev: Is this really something we want to keep?

engine/gamesys/src/gamesys/components/comp_collection_proxy.cpp
```diff
@@ -222,6 +222,37 @@ namespace dmGameSystem
         return (DoLoad(params->m_Factory, (CollectionProxyComponent *) params->m_UserData) == dmGameObject::UPDATE_RESULT_OK);
     }
 
+    static dmResource::Result GetCollectionPreloaderRequestCount(dmResource::HFactory factory, const char* path, uint32_t* request_count)
+    {
+        void* message_data = 0;
+        uint32_t message_size = 0;
+        dmResource::Result result = dmResource::GetRaw(factory, path, &message_data, &message_size);
```
AGulev, line 229: We parse this collection synchronously, which shouldn’t really be a problem in most cases. It should be a fast operation, but I decided to mention it.

AGulev (dismissed):
Now it looks good, just a couple of questions I found

PR #12750, opened 2026-07-17
Notification when editor behavior is disabled after error

### LINKED ISSUE #11299: Editor exceptions shouldn't disable functionality permanently

JosephFerano:
On rare occasions whenever the editor encountered an error, the editor might lose functionality. There is now a notification in such a case that will let you attempt to re-enable the disabled parts and optionally restart the editor as well. There's also a function available for coding agents to attempt the same recovery.

image

Fixes #11299


editor/.codex/skills/clojure-code-style/SKILL.md
```diff
@@ -44,6 +44,15 @@ Prefer `if-let`/`when-let` over `if-some`/`when-some`. Latter can be used only w
...
 
+## Editor recovery
+
+If your change causes the running editor to disable handlers or timers after an exception, fix or revert the cause first. Then re-enable disabled functionality from the REPL:
+
+```clojure
+(editor.ui/run-now
+  (editor.ui/enable-disabled-functionality!))
+```
+
```
JosephFerano, line 55: Figured it might be good to add it as a SKILL, but not sure if it's desired because this is mostly about style.

matgis (approved):

PR #12766, opened 2026-07-18
Add runtime support for text properties

Jhonnyg:
Second deliverable for #8420 - dynamic runtime support via go.get/set:

``` lua
-- update a player label with the player name
local my_text = go.get("#script", "player_name")
my_text = "Name: " .. my_text
go.set("#player_label", "text", my_text)
```

#### Technical changes

* Added runtime text property support through go.get() and go.set().
* Added Lua string conversion and dynamic property-container support for factory/spawn overrides.
* Added native GetPropertyAsText() and SetPropertyFromText() APIs with documented ownership rules.
* Added validation for embedded null bytes and property-container size overflow.
* Added coverage for UTF-8, multiline and empty strings, runtime replacement, and factory overrides.
* Updated the go.property(), go.get(), and go.set() API documentation.


JCash (changes requested):
lgtm!
I think we should add a "text" property to comp_label.cpp, and deprecate "label.set_text()" ?

JCash (approved):

PR #12770, opened 2026-07-19
Allow running games without taking focus

### LINKED ISSUE #12707: Add option to keep the window on the same screen as the editor, while also avoiding focus
For agentic workflows, it would be useful if the game did not switch monitors and instead ran on the same screen as the editor is on (or remembered the last one), and also optionally ran without taking focus.

An agent may run and re-run the game many times, and if the user is working on something else, it can be annoying.

AGulev:
The editor’s `run` HTTP command can now launch or reboot a game without taking focus by passing `focus=false`. Existing workflows remain unchanged because games continue to start focused by default.

Fix https://github.com/defold/defold/issues/12707

### Technical changes

- Added a validated `focus` query parameter to `/command/run`, including OpenAPI documentation and a `400` response for invalid values.
- Propagated the focus option through editor build, launch, and reboot paths.
- Added the `display.start_unfocused` engine configuration and exposed it through `WindowCreateParams`.
- Configured GLFW to avoid focusing the window when it is created or shown, including windowed-fullscreen handling.
- Preserved the existing default behavior and `WindowCreateParams` structure size.
- Added editor argument-generation, command parsing, web-server integration, and platform initialization tests.


vlaaad (commented):
Please do a `$Clojure Code Style` pass over your changes, particularly single-use code and boolean with `?` suffix.

AGulev:
Is it some kind of skill you are talking about? Where can I get it?

vlaaad:
It's checked in

AGulev:
@vlaaad I found it and applied changes

editor/test/editor/engine_test.clj
```diff
@@ -0,0 +1,44 @@
...
+;; this file except in compliance with the License.
+
+(ns editor.engine-test
+  (:require [clojure.test :refer :all]
+            [editor.engine :as engine]))
+
+(def ^:private start-unfocused-argument "--config=display.start_unfocused=1")
+
+(deftest launch-arguments-focus-test
+  (let [launch-arguments #'engine/launch-arguments]
```
vlaaad: Tests set in stone a random subset of the impl.

editor/test/editor/engine_test.clj
```diff
@@ -0,0 +1,44 @@
...
+            "--config=display.start_unfocused=0"
+            "--config=custom.value=1"]
+           (launch-arguments nil
+                             "--config=display.start_unfocused=0\n--config=custom.value=1"
+                             false
+                             2
+                             false)))))
+
+(deftest reboot-arguments-focus-test
+  (let [reboot-arguments #'engine/reboot-arguments
```
vlaaad: Tests set in stone a random subset of the impl.

editor/src/clj/editor/command_requests.clj
```diff
@@ -244,9 +257,15 @@
...
-                                               command->help))}}]
+                                               command->help))}}
+                  {:name "focus"
+                   :in "query"
+                   :description "Whether the launched game should take focus. Only applies to the `run` command."
+                   :schema {:type "boolean"
+                            :default true}}]
      :responses {"200" {:description "Command completed and returned a response body"}
                  "202" {:description "Accepted"}
+                 "400" {:description "Invalid command option"}
```
vlaaad: Remove this. Agents read the OpenAPI spec, so it should take as little context as possible, with only relevant details. If they get an error, they will see it and fix.

editor/src/clj/editor/command_requests.clj
```diff
@@ -244,9 +257,15 @@
                                     (reduce-kv (fn [acc command _]
                                                  (conj! acc command))
                                                (transient [])
-                                               command->help))}}]
+                                               command->help))}}
+                  {:name "focus"
+                   :in "query"
+                   :description "Whether the launched game should take focus. Only applies to the `run` command."
```
vlaaad: Maybe reword to keep it shorter?
```
"Whether the launched game takes focus; only applies to `run`"
```

editor/src/clj/editor/engine.clj
```diff
@@ -101,20 +102,25 @@
...
 
-(defn reboot! [target local-url debug?]
-  (let [uri (URI. (format "%s/post/@system/reboot" (:url target)))
-        conn ^HttpURLConnection (get-connection uri)
-        instance-index (:instance-index target)
-        instance-index? (some? instance-index)
-        args (cond-> [(str "--config=resource.uri=" local-url)]
-                     debug?
-                     (conj (str "--config=bootstrap.debug_init_script=/_defold/debugger/start.luac"))
+(defn- reboot-arguments [target local-url debug focus]
```
vlaaad: Did you run `$Clojure Code Style` on this file? This `reboot-arguments` and `launch-arguments` are single use in production code.  Meanwhile, test that exercise them are wrong because they exercise a random pure subset of the production code instead of the real production code. Test should not use private fns, this makes the tests encode random parts of the code as a behavior that is set in stone, instead of testing the actual expected behavior of the system. For example, if I were designing a "pure" version of engine launcher, I would define its pure contract as input -> map of command args + env + cwd for the process start. This can be defensible to test as a thing on its own, since it leaves  the launcher a trivial "just launch the process".

vlaaad (changes requested):

editor/src/clj/editor/engine.clj
AGulev: >Did you run $Clojure Code Style on this file?

yes

vlaaad (dismissed):
Ship it 👍

engine/platform/src/platform_window_glfw3.cpp
```diff
@@ -328,6 +328,7 @@ namespace dmPlatform
         glfwWindowHintString(GLFW_COCOA_FRAME_NAME, params.m_Title);
         glfwWindowHint(GLFW_VISIBLE, GLFW_FALSE);
         glfwWindowHint(GLFW_FOCUSED, GLFW_FALSE);
+        glfwWindowHint(GLFW_FOCUS_ON_SHOW, params.m_StartUnfocused ? GLFW_FALSE : GLFW_TRUE);
```
britzl: Shouldn't it be m_FocusOnShow or m_StartFocused to better match GLFW_FOCUS_ON_SHOW?

britzl (changes requested):
I suggest to switch from StartUnfocused to StartFocused or FocusOnShow/FocusOnStart

britzl (approved):

PR #12777, opened 2026-07-20
Decouple editor undo system from graph state

### LINKED ISSUE #12340: Decouple editor undo system from graph state
**Task (REQUIRED):**
Currently any change made to the project graph becomes undoable because the undo system keeps snapshots of previous graph states. This tight coupling between the project graph state and undoable state prevents us from making changes to the graph outside of the undo system, like lazy-loading resource nodes into the project graph on demand. This is required functionality for partial project loading.

The task is to move the undo system over to a more traditional command pattern that can be automatically constructed from the existing `g/transact` calls that are spread across the code base.

Once decoupled, we can interleave non-undoable changes to the project graph with undoable ones. This will open up the possibility of partial project loading, as well as enable us to work on a host of other undo-related issues.

Part of #3844, #6392, #7999, #8000, #10166, #11408.

matgis:
The internals of the editor undo system have been rewritten for added flexibility. This will eventually result in various user-facing improvements in future releases, but right now the changes are mostly under the hood.

Fixes #12340.

### Technical changes

:warning: We should merge defold/extension-teal#17 and update the ref in `project.clj` before merging this.

The editor undo system has been completely overhauled. Rather than storing a series of snapshots of the edited graph state, we now use a traditional command-based undo system. Whenever a transaction is performed on the graph, the `TransactionSteps` are `realized` into atomic `TransactionChanges` that are `performed` on the graph and appended to a `undoable-changes` vector. Each `TransactionChange` contains all the information required to `perform` and `revert` the change. To undo, we `revert` the `undoable-changes` of an undo step in reverse order.

The distinction between a `TransactionStep` and a `TransactionChange` is subtle but important. The `TransactionSteps` express the intent of an atomic change to the graph, such as deleting a node or updating a property, whereas the `TransactionChange` contains the information of what actually happened as a result. For example, a `DeleteNodeTXS` `TransactionStep` might be `realized` into many `TransactionChanges` that delete override nodes, nodes connected to `:cascade-delete` inputs, and remove connections from the graph. An `UpdatePropertyTXS` `TransactionStep` might be `realized` into `TransactionChanges` that change the raw property from the value it was at the time into the new value, along with the `TransactionChanges` resulting from `realizing` the `TransactionSteps` returned by the `set` function associated with the property.

Transactions create undo steps by default, but you can make changes to the graph without registering undo steps by supplying `{:undoable false}` as `opts` to `g/transact` or wrapping `TransactionSteps` in `g/non-undoable`. The code has been updated in many places to make all the changes to nodes outside the `project` graph non-undoable.

### Undo architecture

- Replaces per-graph snapshot history with system-level undo stacks.
- Introduces reversible `TransactionChange` records alongside high-level `TransactionStep` records.
- Undo reverts recorded changes in reverse order; redo performs them in their original order.
- Replaces `internal.history` with a bounded `paper-tape` implementation that tracks undo-stack revisions.
- Preserves operation labels, sequence coalescing, cancellation, and the 60-step limit.
- Recomputes invalidation and transaction effects during undo and redo.

### Transaction API

- Adds `{:undoable false}` to make an entire transaction non-undoable.
- Adds `g/non-undoable` for excluding selected steps from an otherwise undoable transaction.
- Adds `:undo-key`, defaulting to `:undo/global`, to support independent undo stacks.
- Changes undo APIs to accept an undo key instead of a graph ID:
  - `g/undo!`
  - `g/redo!`
  - `g/has-undo?`
  - `g/has-redo?`
  - `g/reset-undo!`
  - `g/cancel!`
  - `g/undo-stack-count`
- Adds `g/undo-stack-revision` for detecting any undo/redo state change, including changes that do not affect stack depth.
- Removes the `:history` option from `g/make-graph!`.
- Decouples `:full-invalidation` from undoability.
- Adds batched node addition and deletion transaction steps.

The API documentation now also describes the ownership constraints for operations distributed across different undo stacks.

### Graph mutation and replay

- Makes `IBasis` query-only.
- Moves mutation into explicit `basis-plan-*`, `basis-perform-*`, and `basis-revert-*` functions.
- Captures the state needed to replay node, property, connection, override, and graph-value changes.
- Makes replay resilient when subjects have been removed or changed by another undo stack.
- Preserves granular `node->overrides` relationships during node and override operations.
- Recomputes property and override state against the current basis where required.
- Uses stable arc identifiers so deleting and restoring nodes preserves connection order and handles duplicate array-input arcs. This is achieved by storing the `Arcs` in the new `util.pkid-vector` collection. It is an extension of `PersistentVector` that supports stable addressing with stable primary key ids (PKIDs) while using the same compact linear storage as `PersistentVector`.
- Still enforces graph volatility when establishing connections. Graph `:volatility` is no longer a technical constraint, but serves as a sanity check to prevent project graph node outputs from depending on view graph state. It would be bad if a view setting affected the `save-value`, for example.
- The code is now resilient against "dangling arcs", a situation where an `Arc` remains in `:sarcs` or `:tarcs` when its source or target node-ids are no longer present in the graph. This can happen if a node is created in one undo step, followed by a connection in a later undo step, followed by undoing the creation of the node.

### Editor migration

- Marks initial project loading, resource synchronization, graph setup, and graph disposal as non-undoable.
- Excludes transient UI state such as selections, cameras, visibility, tabs, previews, console state, code-view state, and scene-tool state from undo.
- Uses `g/non-undoable` where model edits and view updates occur in the same transaction.
- Updates project Undo and Redo actions to use the `:undo/global` undo stack.
- Still resets global undo after structural resource-sync changes because earlier undo steps might operate on nodes that were deleted as a result.
- Adds assertions ensuring plugin loading and view/preview lifecycle functions do not create undo steps.

### Supporting changes

- Adds a simulated reload phase to the `load_project.clj` dev module.
- Adds `reduce=>` and `reduce-kv=>` to `coll`. These mirror `reduce->` and `reduce-kv->`, but takes the `init` value as the first argument, followed by the input sequence.
- Adds `removing-assoc`, `removing-update`, and corresponding `-in` variants to `coll`. These will `dissoc` the key from the map instead of inserting a `nil` value.
- Adds `coll/transform-non-empty->`. This mirrors `transform->`, but returns `nil` if the result is empty. This is of particular use with the new `removing-` functions described above.
- Expands schema validation during node construction.
- Validates declared property default values when evaluating `g/defnode` forms.
- Adds extensive unit test coverage for every transaction-step category, cross-stack replay, override handling, invalidation, connection ordering, and undo-stack behavior.

### Bug fixes

- Fixes an issue where property setters for inherited properties were not being run with the default values as new nodes were introduced to the graph.
- Fixed validation of multi-value property values.
- Arcs from another graph can now shadow inherited arcs on override nodes.
- The `:_declared-properties` output is now properly `:cached`, not just declaring itself so.
- The `g/node-value` `:tracer` now traces the `:_properties` and `:_declared-properties` outputs correctly.


editor/src/clj/dynamo/graph.clj
```diff
@@ -317,10 +352,15 @@
...
-                history explicitly.
+                must not be written back into the cache.
+
+              :undoable
+                Defaults to true. When false, commits graph changes without
+                appending the realized TransactionChanges to the undo stack.
+
+              :undo-key
+                Defaults to :undo/global. The undo stack where realized
+                TransactionChanges are appended.
```
vlaaad, line 363: Just a comment, an idea. We currently use a single undo key per transaction. I think it's fine for now, and even if we decide to change it, it is out of scope for this PR.

But I have a scenario in mind that doesn't seem to be satisfied with the undo key approach, though it may work if we supply an undo key **set** per tx. Imagine we implement local undo for code files, so undo queues are separate for each code file. There may be a situation, for example, a symbol rename refactoring, where a single undoable action actually touches 2 otherwise independent files. Implementing functional undo for this scenario could work if we use a set of undo keys and, when undoing, check that the change that is undoable is on top of the undo stack for every affected undo key. Something like that... 

Just a thought about something we might want to change a bit later. Still, this is out of scope for this PR.

editor/src/clj/internal/transaction.clj
```diff
@@ -105,162 +197,196 @@
...
 
-(defonce/type OverrideNodeTXS [original-node-id override-node-id]
+(defn- realize-add-override [ctx undoable-changes override-id root-node-id traverse-fn init-props-fn]
+  (if-let [{:keys [override-id
+                   override]}
+           (ig/basis-plan-add-override (:basis ctx) override-id root-node-id traverse-fn init-props-fn)]
+    (perform-and-conj-change ctx undoable-changes (->AddOverrideTXC override-id override))
+    (pair ctx undoable-changes)))
+
+(defonce/type AddOverrideTXS [override-id root-id traverse-fn init-props-fn]
```
vlaaad: Not sure if you are still working on it here, but `AddOverrideTXS` appears unused. Perhaps it makes sense to ask an agent to go over the changes and report actionable clj-kondo lints?

editor/test/internal/txsteps/connect_test.clj
```diff
@@ -0,0 +1,1387 @@
...
+          (g/transact
+            {:full-invalidation true
+             :undoable false}
+            (g/connect replacement-source-node-id :property-output target-node-id :regular-input))]
+
+      (is (= [] (:undoable-changes tx-result)))
+      (is (= 0 (g/undo-stack-count :undo/global)))
+      (is (= [] (g/targets basis initial-source-node-id :property-output)))
+      (is (= [[replacement-source-node-id :property-output]] (g/sources basis target-node-id :regular-input)))
+      (is (= :replacement-value (g/node-value target-node-id :regular-output))))))
```
vlaaad: 🙀

editor/test/internal/txsteps/expand_test.clj
```diff
@@ -0,0 +1,91 @@
...
+;;
+;; You may obtain a copy of the License, together with FAQs at
+;; https://www.defold.com/license
+;;
+;; Unless required by applicable law or agreed to in writing, software distributed
+;; under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+;; CONDITIONS OF ANY KIND, either express or implied. See the License for the
+;; specific language governing permissions and limitations under the License.
+
+(ns internal.txsteps.expand-test
```
vlaaad, line 15: Should we also add expand-ec test?

editor/src/java/clojure/lang/PkidVector.java
```diff
@@ -0,0 +1,631 @@
...
+ * realized indexes, and vector equality and hashing consider only realized
+ * values.</p>
+ *
+ * <p>Clojure's transient vector implementation is final and cannot retain the
+ * additional fields, so generic transient conversion is rejected.
+ * Consequently, this class cannot be used as the target of
+ * transient-optimized operations such as {@code into}. This class also does
+ * not define a Java serialization format for its pkid state.</p>
+ */
+public final class PkidVector extends PersistentVector {
```
vlaaad, line 43: Alternatively, we could implement `IPersistentVector` and related interfaces.

editor/test/util/pkid_vector_test.clj
```diff
@@ -0,0 +1,592 @@
...
+                  :next-pkid 0}
+           table (pkid-vector/pkid-vector)]
+      (if (= operation-index (count operations))
+        true
+        (let [operation (operations operation-index)
+              model (apply-model-operation model operation)
+              table (apply-table-operation table operation)]
+          (if (table-matches-model? model table)
+            (recur (inc operation-index) model table)
+            false))))))
```
vlaaad, line 762: I think it's useful to use generative/property-based testing for custom collection types so the test runner can find edge cases on its own. I asked Codex to add more property-based tests, and it found an implementation bug:
```clj
(require '[util.pkid-vector :as pkid])
(-> (pkid/pkid-vector)
    (pkid/assoc-pkids [129] :anchor)
    (pkid/assoc-pkids [128] :restored)
    (pkid/dissoc-pkids [129]))
;; throws: IllegalStateException
;; Unsupported IntSet container: null 
```


For reference, this is the patch that Codex produced to find the bug: 

[changes.patch](https://github.com/user-attachments/files/30224088/changes.patch)

vlaaad (commented):
Good stuff 👍

editor/src/clj/dynamo/graph.clj
matgis: Yes, I'm not entirely clear around how it should work myself. My initial thought was that each action would be placed in the undo stack associated with the resource in which the change was made. I.e. If I perform a symbol rename in `utils.lua`, I need to undo it from `utils.lua`, but I suspect we also may have actions that operate on multiple files without the user viewing one of them.

It might be that your suggestion of having any of the files involved undo the change across all of the involved files is the way to go, but it might also be confusing if I undo from a file I didn't realize was involved in a broader change and have it revert changes elsewhere.

Anyway, this PR was not really intended to address this problem yet. There are similar operations you can perform in other IDEs, so we should look at how it works there and figure out if it makes sense for us.

At the moment, none of the production code uses the `:undo-key` option so far. I think we should keep it that way until we figure out how it should work.

editor/src/clj/internal/transaction.clj
matgis: Ah, you're right! Nice catch! I'll be sure to run a `clj-kondo` check across the touched code to ensure this stuff doesn't happen.

editor/test/internal/txsteps/connect_test.clj
matgis: Yeah, it's a lot of stuff to cover. 😮‍💨

editor/test/internal/txsteps/expand_test.clj
matgis: Yes we should! And `callback-ec` should be covered too! I think I missed it because they both share the same `TransactionStep` implementation.

editor/src/java/clojure/lang/PkidVector.java
matgis: I did it like this because I wanted all the fields of `PersistentVector` inline in `PkidVector` rather than having a reference to a `PersistentVector` inside. We have a lot of `arc-tables` that use these `PkidVectors` for storage (millions of them in the big synthetic project) and I wanted to make it as efficient as possible.

My first attempt was to just have a `deftype` that wrapped these, but having the `PkidVector` actually be the vector saves memory and pointer indirections.

We could leverage the internal classes of `PersistentVector` and implement `IPersistentVector` ourselves, but there would still be a substantial chunk of code we'd need to copy from `PersistentVector` to mirror its behavior since everything happens inside instance methods.

In the end, we ended up having to override only the six public methods that returned a `PersistentVector` instance with our own quite simple implementations. The rest of the code in `PkidVector` relates to the internals of managing the taken and next PKIDs.

Overall, I think subclassing `PersistentVector` is the right call here.

editor/test/util/pkid_vector_test.clj
matgis: Interesting! I'm surprised that the existing `operations-match-sorted-map-model` generative test didn't uncover this issue. I'll incorporate this added coverage!

editor/src/clj/internal/node.clj
```diff
@@ -307,14 +314,64 @@
...
+
+(defn- node-type-deref-ordered-property-setter-infos-raw [node-type-deref]
+  {:pre [(node-type-deref? node-type-deref)]}
+  ;; :property-order-decl is not merged by the merge-supertypes function, so we
+  ;; compile a declaration-order that includes inherited properties ourselves.
+  (let [prop-kws-in-declaration-order
+        (reduce (fn [prop-kws supertype]
+                  (into prop-kws
+                        (remove (set prop-kws))
+                        (:property-order-decl @supertype)))
```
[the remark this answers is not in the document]
matgis: The concern is not valid.

Although `:property-order-decl` is not merged, `:supertypes` is flattened during node-type construction. A leaf’s supertype list already includes its parent and grandparent (tested in `internal.defnode-test/inheritance`). Therefore, this `reduce` call visits every ancestor and reads each ancestor’s own `:property-order-decl`.

PR #12782, opened 2026-07-20
Remember scene settings and camera state, better camera and grid defaults

### LINKED ISSUE #12472: Per-resource scene view camera and toolbar persistent settings
Currently, all settings in the scene view toolbar (grid, camera, visibility) persist globally per project instead of per resource. Some camera settings, like ortho vs perspective, don't even persist, and every time you re-open a resource, even if it was set to perspective camera, opening it will have reset both the camera view mode, but also the camera position.

This task will persist as many settings per resource as possible, including camera position, view mode, fov, visibility settings, grid settings, etc;. The defaults will be used whenever a new collection is created, however, it would be beneficial to the user if they can also have a custom global setting and per resource settings, but we would have to design that.

JosephFerano:
This change persists scene settings and camera state:

- Camera state (2D/3D, position, rotation, zoom, etc;) is now remembered for every resource. If a resource tab is closed (like collection, atlas, go, etc;) or the editor is restarted, the camera will restored to the state it was in.
- Scene visibility settings are also now per scene and remembered. The visibility for one collection can be different from another.
- 3D resources (such as model, gltf/glb,mesh, etc;) now start in perspective mode
- Grid settings now remember if working in a 2D or 3D scene. When the 2D button is toggled the grid settings will change to whatever was configured for that specific mode.

Fixes #12472

### Technical changes

Additional Note:

Scene visibility had a few issues; it needed some clean up and apparently the light toggle wasn't working. There was also some poor evaluation-context uses and some dead code. So a good chunk of changes are just cleaning stuff up.

### DIFF
--- a/editor/resources/localization/en.editor_localization
+++ b/editor/resources/localization/en.editor_localization
@@ -1133,6 +1133,7 @@ scene-popup.scene-visibility.gui-shapes = GUI Shapes
 scene-popup.scene-visibility.gui-particle-effects = GUI Particle Effects
 scene-popup.scene-visibility.gui-spine-scenes = GUI Spine Scenes
 scene-popup.scene-visibility.gui-text = GUI Text
+scene-popup.scene-visibility.lights = Lights
 scene-popup.scene-visibility.models = Models
 scene-popup.scene-visibility.particle-effects = Particle Effects
 scene-popup.scene-visibility.skeletons = Skeletons
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -523,6 +523,7 @@
              new-resource-node-id (some-> new-active-tab (editor-tab/resource-node-id evaluation-context))
              new-view-node-id (some-> new-active-tab editor-tab/view-node-id)
              should-load-scene-visibility (and is-in-active-tab-pane
+                                               (not= old-active-tab new-active-tab)
                                                (= :scene (some-> new-active-tab editor-tab/view-type-id)))
              path-key (when should-load-scene-visibility
                         (resource/resource->proj-path
@@ -2227,9 +2228,9 @@
   (g/let-ec [open-views (g/node-value app-view :open-views evaluation-context)
              scene-view-ids (g/node-value app-view :scene-view-ids evaluation-context)]
     (doseq [view-id scene-view-ids]
-      (when-let [resource (:resource (get open-views view-id))]
-        (save-scene-camera-prefs! prefs view-id resource))
       (try
+        (when-let [resource (:resource (get open-views view-id))]
+          (save-scene-camera-prefs! prefs view-id resource))
         (scene/dispose-scene-view! view-id)
         (catch Throwable error
           (error-reporting/report-exception! error)))))
@@ -2400,8 +2401,7 @@
       (get scene-default-camera-projection-by-ext ext :orthographic))))
 
 (defn- camera-prefs [camera]
-  {:2d-mode (camera/mode-2d? camera)
-   :projection (:type camera)
+  {:projection (:type camera)
    :position (math/vecmath->clj (:position camera))
    :rotation (math/vecmath->clj (:rotation camera))
    :fov-x (:fov-x camera)
@@ -2418,18 +2418,16 @@
                                 (camera-prefs camera)))))
 
 (defn- try-load-camera-from-prefs [prefs path-key]
-  (when-let [{:keys [projection position rotation fov-x fov-y focus-point]}
-             (prefs/get-pref-entry-in prefs [:scene :resource-settings] path-key [:camera] nil)]
-    (let [[position-x position-y position-z] position
-          [rotation-x rotation-y rotation-z rotation-w] rotation
-          [focus-x focus-y focus-z focus-w] focus-point]
-      (assoc (camera/make-camera (or projection :orthographic)
-                                 identity
-                                 {:fov-x (or fov-x 1000.0)
-                                  :fov-y (or fov-y 1000.0)})
-        :position (Point3d. (double position-x) (double position-y) (double position-z))
-        :rotation (Quat4d. (double rotation-x) (double rotation-y) (double rotation-z) (double rotation-w))
-        :focus-point (Vector4d. (double focus-x) (double focus-y) (double focus-z) (double focus-w))))))
+  (let [{:keys [projection position rotation fov-x fov-y focus-point]}
+        (prefs/get-pref-entry-in prefs [:scene :resource-settings] path-key [:camera] nil)]
+    (when (and projection position rotation fov-x fov-y focus-point)
+      (let [[position-x position-y position-z] position
+            [rotation-x rotation-y rotation-z rotation-w] rotation
+            [focus-x focus-y focus-z focus-w] focus-point]
+        (assoc (camera/make-camera projection identity {:fov-x fov-x :fov-y fov-y})
+          :position (Point3d. (double position-x) (double position-y) (double position-z))
+          :rotation (Quat4d. (double rotation-x) (double rotation-y) (double rotation-z) (double rotation-w))
+          :focus-point (Vector4d. (double focus-x) (double focus-y) (double focus-z) (double focus-w)))))))
 
 (defn- make-tab! [app-view prefs localization resource-node view-type ^ObservableList tabs opts]
   (let [basis (g/now)
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -367,6 +367,7 @@
   (property manip-space g/Keyword)
   (property keymap g/Any)
   (property localization g/Any)
+  (property scene-visibility g/Any)
 
   (input outline-pane-desc g/Any)
   (input properties-pane-desc g/Any)
@@ -486,8 +487,12 @@
              ^SplitPane editor-tabs-split (g/node-value app-view :editor-tabs-split evaluation-context)
              ^Scene app-scene (g/node-value app-view :scene evaluation-context)
              properties-view (g/node-value app-view :properties-view evaluation-context)
+             scene-visibility (g/node-value app-view :scene-visibility evaluation-context)
              new-resource-node-id (some-> new-active-tab (editor-tab/resource-node-id evaluation-context))
              new-view-node-id (some-> new-active-tab editor-tab/view-node-id)
+             resource (resource-node/resource basis new-resource-node-id)
+             prefs (g/node-value scene-visibility :prefs evaluation-context)
+             path-key (resource/resource->proj-path resource)
 
              tx-data
              (when (and is-in-active-tab-pane
@@ -509,6 +514,17 @@
     ;; The remaining steps should always be performed, even if we didn't end up
     ;; updating the graph connections.
     (recent-files/save-tab-selections! prefs app-view)
+
+    ;; Update scene-visibility settings
+    (g/with-auto-evaluation-context evaluation-context
+      (if-let [settings (prefs/get-pref-entry prefs [:scene :resource-settings] path-key)]
+        (scene-visibility/update-settings scene-visibility settings evaluation-context)
+        (scene-visibility/update-settings scene-visibility
+                                          ;; TODO JOE: This doesn't belong here
+                                          {:filters-enabled true :filtered-renderable-tags (cond-> #{}
+                                                                                             (system/defold-dev?)
+                                                                                             (conj :dev-visibility-bounds))}
+                                          evaluation-context)))
 
     (g/let-ec [active-tab-pane (g/node-value app-view :active-tab-pane evaluation-context)]
       (doseq [^TabPane tab-pane (.getItems editor-tabs-split)]
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -2238,7 +2238,14 @@
     (refresh-scene-view! view-id dt))
   (scene-cache/prune-context! nil))
 
-(declare save-scene-camera-prefs!)
+(defn- save-scene-camera-prefs! [prefs view resource]
+  (g/let-ec [camera (some-> (:basis evaluation-context)
+                            (scene/view->camera view)
+                            (g/node-value :local-camera evaluation-context))
+             path-key (resource/resource->proj-path resource)]
+    (when (and camera path-key)
+      (prefs/set-pref-entry-in! prefs [:scene :resource-settings] path-key [:camera]
+                                (camera/camera->prefs-value camera)))))
 
 (let [TabHeaderSkin (Class/forName "javafx.scene.control.skin.TabPaneSkin$TabHeaderSkin")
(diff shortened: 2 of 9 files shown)


editor/resources/localization/en.editor_localization
```diff
@@ -1143,6 +1143,7 @@ scene-popup.scene-visibility.gui-shapes = GUI Shapes
 scene-popup.scene-visibility.gui-particle-effects = GUI Particle Effects
 scene-popup.scene-visibility.gui-spine-scenes = GUI Spine Scenes
 scene-popup.scene-visibility.gui-text = GUI Text
+scene-popup.scene-visibility.lights = Lights
```
JosephFerano, line 1146: There was code to handle the Lights toggle in scene visibility, but it wasn't working

editor/src/clj/editor/app_view.clj
```diff
@@ -142,6 +144,36 @@
...
+   "png" :orthographic
+   "atlas" :orthographic
+   "tilesource" :orthographic
+   "tileset" :orthographic
+   "tilemap" :orthographic
+   "tilegrid" :orthographic
+   "gui" :orthographic
+   "font" :orthographic
+   "sprite" :orthographic
+   "label" :orthographic})
```
JosephFerano: Kept the orthographic maps here for explicitness, could be removed since the default is :orthographic anyway, but I like it like this.

editor/src/clj/editor/app_view.clj
```diff
@@ -2195,12 +2240,18 @@
...
-        (error-reporting/report-exception! error))))
+(declare save-scene-camera-prefs!)
+
+(defn- dispose-scene-views! [app-view prefs]
+  (g/let-ec [open-views (g/node-value app-view :open-views evaluation-context)
+             scene-view-ids (g/node-value app-view :scene-view-ids evaluation-context)]
+    (doseq [view-id scene-view-ids]
+      (try
+        (when-let [resource (:resource (get open-views view-id))]
+          (save-scene-camera-prefs! prefs view-id resource))
```
JosephFerano: Save when we close the editor. If there's a better place please let me know.

editor/src/clj/editor/boot_open_project.clj
```diff
@@ -106,6 +106,29 @@
 (defn- find-tab [^TabPane tabs id]
   (some #(and (= id (.getId ^Tab %)) %) (.getTabs tabs)))
 
+(defn- clean-up-resource-prefs [prefs changes]
```
JosephFerano, line 109: Hopefully not overkill, wanted to clean up as we go. Hopefully I covered all the scenarios in the tests.

editor/src/clj/editor/grid.clj
```diff
@@ -268,37 +273,41 @@
...
-          :on-value-changed (partial value-changed-fn :color)}
-         {:key :opacity :type :slider :label "scene-popup.grid.opacity" :min 0.0 :max 1.0
-          :value (prefs/get prefs [:scene :grid :opacity])
-          :on-value-changed (partial value-changed-fn :opacity)
-          :slider-value->string (fn [^double v]
-                                  (str (Math/round (* v 100)) "%"))}]
-        descriptors (filterv #(not (contains? ignored-keys (:key %))) all-descriptors)
-        initial-state (into {} (keep #(when-let [k (:key %)] [k (:value %)])) descriptors)]
-    (settings-popup/show! owner keymap localization initial-state 240 descriptors)))
+  (g/let-ec [scene-view-id (g/node-value app-view :active-view evaluation-context)
```
JosephFerano, line 281: Fixing evaluation-context uses

editor/src/clj/editor/prefs.clj
```diff
@@ -78,6 +78,12 @@
 
 #_[:boolean :string :password :locale :keyword :integer :number :one-of :array :set :object :object-of :enum :tuple]
 
+(defn- vec3-schema [x y z]
```
JosephFerano, line 81: I don't know if this is not desired, if we prefer to inline i.

editor/src/clj/editor/prefs.clj
```diff
@@ -249,19 +255,52 @@
...
+                                                                          :fov-x {:type :number :default 1000.0}
+                                                                          :fov-y {:type :number :default 1000.0}
+                                                                          :focus-point {:type :tuple
+                                                                                        :items [{:type :number :default 0.0}
+                                                                                                {:type :number :default 0.0}
+                                                                                                {:type :number :default 0.0}
+                                                                                                {:type :number :default 1.0}]}}}}}}
+             ;; NOTE: We also track whether the grid button is active or not, however, not here. Because the grid visibility
+             ;; is controlled by the SceneVisibilityNode, we ended up piggy-backing a per-resource grid button active setting
+             ;; through [:scene :resource-settings <resource-proj-path> :scene-visibility]
```
JosephFerano, line 286: Interesting wrinkle to this, left a note

editor/src/clj/editor/app_view.clj
```diff
@@ -142,6 +144,36 @@
...
+   "atlas" :orthographic
+   "tilesource" :orthographic
+   "tileset" :orthographic
+   "tilemap" :orthographic
+   "tilegrid" :orthographic
+   "gui" :orthographic
+   "font" :orthographic
+   "sprite" :orthographic
+   "label" :orthographic})
+
```
matgis: We don't really want to take this hardcoded approach, since editor plugins can register new resource types that the main editor repo doesn't know about. We could let users override this using `:view-opts` when they register resource types, but I think a better approach might be to evaluate the `:scene` output of the resource and traverse the hierarchy. If you encounter any `:aabb` that has a positive volume, it is likely a 3D scene and should default to perspective projection.

editor/src/clj/editor/app_view.clj
```diff
@@ -367,6 +399,7 @@
   (property manip-space g/Keyword)
   (property keymap g/Any)
   (property localization g/Any)
+  (property scene-visibility g/Any)
```
matgis: This should be an input of type `g/NodeID`, not a property.

editor/src/clj/editor/app_view.clj
```diff
@@ -2195,12 +2240,18 @@
...
+(defn- dispose-scene-views! [app-view prefs]
+  (g/let-ec [open-views (g/node-value app-view :open-views evaluation-context)
+             scene-view-ids (g/node-value app-view :scene-view-ids evaluation-context)]
+    (doseq [view-id scene-view-ids]
+      (try
+        (when-let [resource (:resource (get open-views view-id))]
+          (save-scene-camera-prefs! prefs view-id resource))
+        (scene/dispose-scene-view! view-id)
+        (catch Throwable error
+          (error-reporting/report-exception! error)))))
```
matgis: This stuff is not really about disposing scene views, making the function name misleading. Instead, call `save-scene-camera-prefs!` from the `ui/on-closed!` handler.

editor/src/clj/editor/app_view.clj
```diff
@@ -2355,6 +2406,47 @@
...
+  (let [ext (some-> resource resource/resource-type :ext)]
+    (case ext
+      "collisionobject"
+      (g/let-ec [game-project (project/get-resource-node project "/game.project" evaluation-context)
+                 physics-type (game-project/get-setting game-project ["physics" "type"] evaluation-context)]
+        (if (= "2D" physics-type)
+          :orthographic
+          :perspective))
+
+      (get scene-default-camera-projection-by-ext ext :orthographic))))
```
matgis: Basing it on the `:scene` `:aabbs` as suggested above should solve this.

editor/src/clj/editor/app_view.clj
```diff
@@ -2355,6 +2406,47 @@
...
+
+(defn- camera-prefs [camera]
+  {:projection (:type camera)
+   :position (math/vecmath->clj (:position camera))
+   :rotation (math/vecmath->clj (:rotation camera))
+   :fov-x (:fov-x camera)
+   :fov-y (:fov-y camera)
+   :focus-point (math/vecmath->clj (:focus-point camera))})
```
matgis: Move to `editor.camera`?

editor/src/clj/editor/app_view.clj
```diff
@@ -2355,6 +2406,47 @@
...
+        (prefs/get-pref-entry-in prefs [:scene :resource-settings] path-key [:camera] nil)]
+    (when (and projection position rotation fov-x fov-y focus-point)
+      (let [[position-x position-y position-z] position
+            [rotation-x rotation-y rotation-z rotation-w] rotation
+            [focus-x focus-y focus-z focus-w] focus-point]
+        (assoc (camera/make-camera projection identity {:fov-x fov-x :fov-y fov-y})
+          :position (Point3d. (double position-x) (double position-y) (double position-z))
+          :rotation (Quat4d. (double rotation-x) (double rotation-y) (double rotation-z) (double rotation-w))
+          :focus-point (Vector4d. (double focus-x) (double focus-y) (double focus-z) (double focus-w)))))))
+
```
matgis, line 2415: Move prefs deserialization to `editor.camera` too?

editor/src/clj/editor/app_view.clj
```diff
@@ -2406,6 +2504,9 @@
     (let [close-handler (.getOnClosed tab)]
       (.setOnClosed tab (ui/event-handler event
                           (recent-files/add! prefs resource view-type)
+                          (when (= :scene (:id view-type))
+                            (save-scene-camera-prefs! prefs view resource))
+
```
matgis, line 2476: Both this and the `recent-files/add!` call should actually call `(resource-node/resource basis resource-node)` to get the updated `resource` from the `resource-node` itself. Otherwise we will not be picking up the updated resource path if the user renames the resource from the Assets panel while the editor tab is open.

editor/src/clj/editor/boot_open_project.clj
```diff
@@ -369,6 +393,7 @@
         (ui/context! workbench :workbench context-env (app-view/->selection-provider app-view) dynamics))
       (g/transact
         (concat
+          (g/set-property app-view :scene-visibility scene-visibility)
```
matgis: As stated above, this would be better if connected to an input on the `app-view`, and moved down to the other `scene-visibility` connections.

editor/src/clj/editor/prefs.clj
```diff
@@ -249,19 +255,52 @@
...
+                                                                                             {:type :number :default 0.0}
+                                                                                             {:type :number :default 1.0}]}
+                                                                          :fov-x {:type :number :default 1000.0}
+                                                                          :fov-y {:type :number :default 1000.0}
+                                                                          :focus-point {:type :tuple
+                                                                                        :items [{:type :number :default 0.0}
+                                                                                                {:type :number :default 0.0}
+                                                                                                {:type :number :default 0.0}
+                                                                                                {:type :number :default 1.0}]}}}}}}
```
matgis: We don't appear to use the `w` component? If we're not using it, we shouldn't be storing it.

matgis (changes requested):
Works great! I would like to get rid of the hard-coded defaults per file extension, and found some minor issues.

editor/src/clj/editor/boot_open_project.clj
```diff
@@ -278,10 +301,11 @@
 
       (workspace/add-resource-listener! workspace 0
                                         (reify resource/ResourceListener
-                                          (handle-changes [_ _ _]
+                                          (handle-changes [_ changes _]
                                             (let [open-views (g/node-value app-view :open-views)
                                                   panes (.getItems ^SplitPane editor-tabs-split)]
-                                              (handle-resource-changes! scene panes open-views changes-view)))))
+                                              (handle-resource-changes! scene panes open-views changes-view)
+                                              (clean-up-resource-prefs prefs changes)))))
```
matgis, line 314: Should we also call some version of `clean-up-resource-prefs` that prunes out resources that no longer exist when we load the project as well?

editor/src/clj/editor/app_view.clj
JosephFerano: I agree, but in that case, I inlined the function because it's `scene/dispose-scene-view!` and `scene-cache/drop-context!` that's really doing the heavy lifting. The reason I was calling `save-scene-camera-prefs!` inside `dispose-scene-views!` was because it had the dependencies and iteration I needed to invoke it. By inlining, we can avoid duplicating this machinery.

editor/src/clj/editor/boot_open_project.clj
JosephFerano: Oh good call.

editor/src/clj/editor/app_view.clj
JosephFerano: I see what you mean but I'm a little concerned with this approach. We would be doing a full scene hierarchy traversal while also potentially triggering false positives. I asked the agents to try and find false positives in a few projects, and it supposedly found some, apparently because of cameras in collections;
```
Across 6 2D projects (36 scene resources), 4 came back as 3D:

| project        | misclassified                                        |
|----------------+------------------------------------------------------|
| Tinyswords     | `/main/main.collection`                              |
| TopDownShooter | `/main/Level1.collection`, `/main/Level2.collection` |
| war-battles    | `/main/main.collection`                              |
```
It's also a marginal gain because this will only happen once when opening an existing resource for the first time. If we did nothing, then everything will behave as it currently does (until they change the camera).

My preference would be to add a `:view-opts` entry to only the few resources that are clearly 3D (assuming none of the lights are meant to work in 2D mode), and any nil defaults to ortho;

`"model" "mesh" "cubemap" "gltf" "glb" "directional_light" "point_light" "spot_light" "ambient_light"`

I would also keep the `collisionobject` physics type check.

Let me know what your thoughts are. I'll trust your judgement if you think the scene aabb is the better approach.

matgis: Even if we do allow :view-opts to override the behavior, I still think the approach I proposed is a good default. You're right to point out that the camera AABBs would need to be skipped. We can use the tags for this. Scene traversal is itself not a costly operation - we do it every frame. And evaluating the :scene output is also "free" in that it will be cached for the immediate render call that follows once the view opens.

editor/src/clj/editor/app_view.clj
```diff
@@ -2195,13 +2238,7 @@
...
   (scene-cache/prune-context! nil))
 
-(defn- dispose-scene-views! [app-view]
-  (doseq [view-id (g/node-value app-view :scene-view-ids)]
-    (try
-      (scene/dispose-scene-view! view-id)
-      (catch Throwable error
-        (error-reporting/report-exception! error))))
-  (scene-cache/drop-context! nil))
+(declare save-scene-camera-prefs!)
```
matgis: We could probably just move the implementation up to here in the file now?

matgis (changes requested):
Changes look good, but the issue with the hardcoded resource types remains to be addressed.

editor/src/clj/editor/app_view.clj
JosephFerano: Ok, so I made this the default fallback, I still hardcoded well-known extensions to skip the scene traversal. If any of those are debatable, please let me know and I'll remove them (or add).

matgis (approved):
LGTM!

PR #12787, opened 2026-07-20
Improve test reliability

### LINKED ISSUE #10694: Throw exception if editor tests fail to download dependencies
**Task (REQUIRED):**
It seems the editor tests will silently ignore errors when fetching libraries, which can result in confusing test failures later on. We should throw an exception if a dependency fails to install for some reason.

And maybe it would make sense to retry downloads in some cases?

Editor tests that rely on library dependencies should throw an exception if one of the required libraries fail to install for some reason.

vlaaad:
Fix #10694


vlaaad:
@codex review

JosephFerano (approved):
LGTM

PR #12805, opened 2026-07-22
Quit Defold if the project folder was removed or moved

vlaaad:
If a project directory is removed or renamed, the Defold editor can no longer be used. Previously, the editor raised an exception in such cases. Now, we show a dialog informing the user that the editor will quit:

Screenshot 2026-07-22 at 14 31 37

Fix #10913

### DIFF
--- a/editor/src/clj/editor/app_view.clj
+++ b/editor/src/clj/editor/app_view.clj
@@ -904,10 +904,28 @@
 
 (defn async-reload!
   [app-view changes-view workspace moved-files]
-  (let [render-reload-progress! (make-render-task-progress :resource-sync)]
-    (disk/async-reload! render-reload-progress! workspace moved-files changes-view
-                        (fn [_success]
-                          (ui/user-data! (g/node-value app-view :scene) ::ui/refresh-requested? true)))))
+  (if-not (.isDirectory (workspace/project-directory workspace))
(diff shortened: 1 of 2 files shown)


editor/src/clj/editor/app_view.clj
```diff
@@ -904,10 +904,28 @@
...
+        localization
+        {:title (localization/message "dialog.project-directory-unavailable.title")
+         :icon :icon/triangle-error
+         :header (localization/message "dialog.project-directory-unavailable.header")
+         :content {:wrap-text true
+                   :text (localization/message "dialog.project-directory-unavailable.content")}
+         :buttons [{:text (localization/message "dialog.button.quit")
+                    :cancel-button true
+                    :default-button true}]})
+      (.setOnCloseRequest stage nil)
```
vlaaad, line 920: Disable save checks.

matgis (approved):
This is better than it was before, but the "unsaved changes will be lost" situation is still a bit dire. Perhaps we could let the user salvage the situation by allowing them to click something like a "Sorry, I moved it back!" button in the modal dialog? 🤔

vlaaad:
Good idea for improvement, though I'm not sure this scenario is realistic — if users decide to move the project to a different folder, chances are, they saved the changes beforehand. In this change, we have not taken away the ability to save into a deleted directory, but simply changed an error into a blocking notification. We could implement unsaved changes restoration if users request it though!

PR #12811, opened 2026-07-22
Fix callbacks created from Lua coroutines

AGulev:
Particle effects, asynchronous buffer loading, and asynchronous texture creation now accept callbacks when called from a Lua coroutine. Callbacks remain valid after the originating coroutine finishes and is garbage collected, avoiding callback creation errors and unsafe texture cleanup.

### Technical changes

- Create callbacks from the calling Lua state in `particlefx.play()`, `sys.load_buffer_async()`, and `resource.create_texture_async()`.
- Store the main Lua state for asynchronous texture buffer cleanup so the request never retains a collected coroutine state.
- Add coroutine regression tests for all three APIs, including forced garbage collection before asynchronous completion.
- Add the new test resources to the Waf/Bob `build.inputs` files.
- Verify the new coroutine tests and existing async buffer and texture tests with `test_gamesys`.

### DIFF
--- a/engine/gamesys/src/gamesys/scripts/script_particlefx.cpp
+++ b/engine/gamesys/src/gamesys/scripts/script_particlefx.cpp
@@ -250,7 +250,7 @@ namespace dmGameSystem
 
         if (top > 1 && !lua_isnil(L, 2))
         {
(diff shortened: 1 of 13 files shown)


britzl (approved):
Nice one!

PR #12820, opened 2026-07-23
Fix morph-target corruption when one model has meshes with different target counts

### LINKED ISSUE #12819: Morph targets corrupt when one model has meshes with different target counts

`model.set_blend_weights()` stores one weight array on the model *component*, which is shared by every mesh the component renders. When a single component contains **two or more meshes with different morph-target counts** sharing **one material**, the meshes with fewer targets render with garbage deformation instead of ignoring the extra weights.

The docs for `set_blend_weights` state: "At most as many weights are applied as each mesh has morph targets; extra entries in the table are ignored." The render path never performs that per-mesh clamp.

Image

**To Reproduce (REQUIRED)**

Minimal repro project: https://github.com/LaP0573/morph-target-bug-repro

1. One `.model` component references a glTF with two meshes (dense morph accessors): mesh A has 2 morph targets, mesh B has 1. Both share **one** material (the morph vertex shader from the official morph-target example).
2. A script calls `model.set_blend_weights("#model", {0, 1})`, a length-2 vector with a nonzero entry at index 2, beyond mesh B's single target.
3. Build and run. (The project auto-cycles weights off and on; or launch with `--config=morph.mode=1` to hold weights-on.)

Per the documented contract, mesh B has 1 target, so the 2nd weight is ignored and mesh B stays at rest. Mesh A applies its 2nd target.

Actual: mesh A morphs correctly, but mesh B deforms into garbage (see image). The render path fills mesh B's shader weights from the script array length (2), not its own target count (1); its vertex shader then samples morph-texture array layers past mesh B's range, the GPU clamps the out-of-range layer to the last valid one, and mesh B applies the wrong slice.

  - 1.13.0 (verified present on `dev` as well)

**Platforms (REQUIRED):**
 - Platforms: Windows (this is shared render logic in `comp_model.cpp`, not backend-specific)
 - OS: Windows 11

https://github.com/LaP0573/morph-target-bug-repro

**Root cause (OPTIONAL):**

`engine/gamesys/src/gamesys/components/comp_model.cpp`: `ApplyMorphToRenderObject` (uniform `morph_targets_weights` path) and `WriteMeshAttributes` (instanced `SEMANTIC_TYPE_MORPH_TARGET_WEIGHTS` path) both fill the shader weight slots from the script `weights_count`, never clamped to `render_item->m_Mesh->m_MorphTargetCount`. A morph texture has `target_count * 3` layers, so a mesh with fewer targets samples layers past its own range. A two-line clamp resolves it (PR to follow).

LaP0573:
Morph targets (blend shapes) could render corrupted when a single model component contained two or more meshes with **different morph-target counts** sharing one material. `model.set_blend_weights()` stores one weight array on the component, and the render path applied its full length to every mesh instead of clamping to each mesh's own target count, so a mesh with fewer targets sampled out-of-range morph-texture layers and deformed into garbage. It now clamps per mesh, matching the documented `set_blend_weights` behavior ("extra entries in the table are ignored"). Meshes with matching target counts are unaffected.

image

Fixes #12819 

### Technical changes
* `comp_model.cpp`: clamp `weights_count` to `render_item->m_Mesh->m_MorphTargetCount` before filling the shader weight slots, in both `ApplyMorphToRenderObject` (uniform `morph_targets_weights`) and `WriteMeshAttributes` (instanced `SEMANTIC_TYPE_MORPH_TARGET_WEIGHTS` path). `mesh_morph_count` was already computed at the uniform site (used only for the shader-capacity warning); the clamp reuses it.
* Verified by building the engine from source (`x86_64-win32`) and running the repro as an A/B: with the clamp reverted mesh B is corrupt; with it applied mesh B rests, and mesh A (2 targets) is correct in both.


JCash:
Hi @LaP0573 !

For this issue, I'd like to see a unit test catching this issue, with a test model that triggers it.

LaP0573:
@JCash done!

Jhonnyg (approved):

PR #12823, opened 2026-07-24
Filtered "Search in Files" search speed improvement

JosephFerano:
When preparing search results, the editor would grab all textual resources without considering excluded file types from the search. Now the search dialog will only prepare the included file types for scanning, improving the speed at which results can start.

Fixes #11794

### Technical changes

Here's a repro project in case you want one that has a ton of files you can search through. One search term is "NEEDLE" .

[issue-11794-filter-search-speed.zip](https://github.com/user-attachments/files/30402558/issue-11794-filter-search-speed.zip)

### DIFF
--- a/editor/src/clj/editor/defold_project_search.clj
+++ b/editor/src/clj/editor/defold_project_search.clj
@@ -31,42 +31,55 @@
   (some-> entry :resource resource/proj-path))
 
 (defn make-search-data-future
-  "Returns a future yielding search data for every project resource that passes
-  search-resource?, each with its :search-value built and ready to match against."
+  "Returns a map of :data-future and :cancel!. Deref :data-future for search data
+  covering every project resource that passes search-resource?, each with its
+  :search-value built and ready to match against. Call :cancel! to abandon the
+  work without reporting an error."
   [report-error! project search-resource?]
-  (let [evaluation-context (g/make-evaluation-context)]
-    (future
-      (try
-        (let [search-data
-              (->> (g/node-value project :node-id+resources evaluation-context)
-                   (into []
-                         (keep (fn [[node-id resource]]
-                                 (thread-util/throw-if-interrupted!)
-                                 (when (and (resource/loaded? resource)
-                                            (not (resource/internal? resource))
-                                            (= :file (resource/source-type resource))
-                                            (search-resource? resource))
-                                   (let [resource-type (resource/resource-type resource)
-                                         search-value-fn (when (:search-fn resource-type)
-                                                           (:search-value-fn resource-type))]
-                                     (when (and search-value-fn
-                                                (resource/exists? resource)
-                                                (resource/textual? resource))
-                                       (let [search-value (search-value-fn node-id resource evaluation-context)]
-                                         (when-not (g/error? search-value)
-                                           {:resource resource
-                                            :search-value search-value}))))))))
-                   (sort-by search-data-sort-key))]
-          (ui/run-later
-            (project/update-system-cache-save-data! evaluation-context)
-            (project/log-cache-info! (g/cache) "Cached searched save values in system cache."))
-          search-data)
-        (catch InterruptedException _
-          ;; future-cancel was invoked because the filter changed; discard.
-          nil)
-        (catch Throwable error
-          (report-error! error)
-          nil)))))
+  (let [evaluation-context (g/make-evaluation-context)
+
+        ;; Interrupting the thread can surface as any exception, not just
+        ;; InterruptedException, so we record that we cancelled rather than
+        ;; guessing from the exception type.
+        cancelled (volatile! false)
+
+        data-future
+        (future
+          (try
+            (let [search-data
+                  (->> (g/node-value project :node-id+resources evaluation-context)
+                       (into []
+                             (keep (fn [[node-id resource]]
+                                     (thread-util/throw-if-interrupted!)
+                                     (when (and (resource/loaded? resource)
+                                                (not (resource/internal? resource))
+                                                (= :file (resource/source-type resource))
+                                                (search-resource? resource))
+                                       (let [resource-type (resource/resource-type resource)
+                                             search-value-fn (when (:search-fn resource-type)
+                                                               (:search-value-fn resource-type))]
+                                         (when (and search-value-fn
+                                                    (resource/exists? resource)
+                                                    (resource/textual? resource))
+                                           (let [search-value (search-value-fn node-id resource evaluation-context)]
+                                             (when-not (g/error? search-value)
+                                               {:resource resource
+                                                :search-value search-value}))))))))
+                       (sort-by search-data-sort-key))]
+              (ui/run-later
+                (project/update-system-cache-save-data! evaluation-context)
+                (project/log-cache-info! (g/cache) "Cached searched save values in system cache."))
+              search-data)
+            (catch Throwable error
+              (when-not @cancelled
+                (report-error! error))
+              nil)
+            (finally
+              (Thread/interrupted))))]
+    {:data-future data-future
+     :cancel! (fn cancel! []
+                (vreset! cancelled true)
+                (future-cancel data-future))}))
 
 (defn- resource-matches-library-setting? [resource include-libraries?]
   (or include-libraries?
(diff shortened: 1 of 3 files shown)


editor/src/clj/editor/defold_project_search.clj
```diff
@@ -147,38 +173,64 @@
...
+                          (swap! pending-search-atom start-search! search-data-future search-string))
                         (catch Throwable error
                           (report-error! error)))
                       nil)
      :abort-search! (fn []
                       (try
                         (swap! pending-search-atom abort-search!)
+                        (when-let [current @search-data-atom]
+                          (reset! search-data-atom nil)
+                          ((:cancel! current)))
```
matgis: As stated above, you shouldn't separate reads and writes to an `atom`. Use `swap!` instead to "take" the value from the atom so that a retried concurrent `swap!` will see that the value is `nil` and not attempt to cancel it again.

editor/src/clj/editor/defold_project_search.clj
```diff
@@ -23,42 +23,67 @@
...
+                (project/update-system-cache-save-data! evaluation-context))
+              (when-not @cancelled
+                (report-error! error))
+              nil)
+            (finally
+              (Thread/interrupted))))]
+    {:data-future data-future
+     :cancel! (fn cancel! []
+                (vreset! cancelled true)
+                (future-cancel data-future))}))
```
matgis: I don't understand what prompted this change? What do you mean by "Interrupting the thread can surface as any exception"? You added a call to `thread-util/throw-if-interrupted!`, so if you catch `InterruptedException`, you can treat that as the thread having been cancelled, no?

editor/src/clj/editor/defold_project_search.clj
```diff
@@ -85,51 +110,52 @@
...
 
-(defn- make-search-resource? [searched-exts search-libraries]
+(defn- parse-searched-exts [searched-exts]
+  (let [searched-exts (some-> searched-exts
+                     (string/replace #" " "")
+                     (string/split #","))]
+    (coll/into-> searched-exts []
+                 (remove empty?)
+                 (map #(string/replace % #"\*?\." ""))
+                 (distinct))))
```
matgis: Style: Indentation is messed up.

editor/src/clj/editor/defold_project_search.clj
```diff
@@ -85,51 +110,52 @@
...
   (let [file-ext-patterns
-        (into []
-              (comp (remove empty?)
-                    (distinct)
-                    (map #(text-util/search-string->re-pattern (string/replace % #"\*?\." "") :case-insensitive)))
-              (some-> searched-exts
-                      (string/replace #" " "")
-                      (string/split #",")))]
+        (coll/into-> searched-ext-strings []
+                     (map #(text-util/search-string->re-pattern % :case-insensitive)))]
```
matgis: This can be moved into `parse-searched-exts`?

Style: We indent `coll/into->` using two spaces. We have a rule for it in `.cljfmt.edn`.

editor/src/clj/editor/defold_project_search.clj
```diff
@@ -147,38 +173,64 @@
...
+        ;; changes, the now-obsolete future is cancelled so its whole-project
+        ;; work does not keep running in the background. Reading the atom and
+        ;; resetting it is not atomic, but every caller arrives on the UI
+        ;; thread, so there is no interleaving to guard against.
+        search-data-atom (atom nil)
+        prepare-search-data!
+        (fn [searched-exts search-libraries]
+          (let [searched-ext-strings (parse-searched-exts searched-exts)
+                filter-key [searched-ext-strings search-libraries]
+                current @search-data-atom]
```
matgis: In a multithreaded environment, you can't safely `deref` an `atom` and `reset!` it based on the value. You have to call `swap!` and operate on the value within the `swap!` function. Otherwise a concurrently running thread can overwrite the atom with a stale value.

editor/src/clj/editor/defold_project_search.clj
```diff
@@ -147,38 +173,64 @@
...
+              (:data-future current)
+              (let [{:keys [data-future cancel!]}
+                    (make-search-data-future
+                      report-error! project
+                      (make-search-resource? searched-ext-strings search-libraries))]
+                (reset! search-data-atom {:key filter-key :data-future data-future :cancel! cancel!})
+                (when current
+                  ((:cancel! current)))
+                data-future))))
+
```
matgis, line 176: It looks like you should be able to use `(fn/memoize {:limit 1} ...)` here instead of managing the `search-data-atom` and `filter-key` yourself?

editor/src/clj/editor/search_results_view.clj
```diff
@@ -312,12 +311,21 @@
...
+        (init-search-in-files-tree-view! resources-tree)
+
+        (let [term (prefs/get prefs search-in-files-term-prefs-key)
+              exts (prefs/get prefs search-in-files-exts-prefs-key)
+              include-libraries? (prefs/get prefs search-in-files-include-libraries-prefs-key)]
+          (start-search! term exts include-libraries?)
+          (ui/text! search term)
+          (ui/text! types exts)
+          (ui/value! include-libraries-check-box include-libraries?))
+
```
matgis, line 323: Why did this code move up?

editor/test/editor/defold_project_search_test.clj
```diff
@@ -136,7 +136,7 @@
 
 (defn- try-make-search-data-future [project]
   (let [report-error! (fn/make-call-logger)
-        search-data-future (project-search/make-search-data-future report-error! project)]
+        search-data-future (:data-future (project-search/make-search-data-future report-error! project (constantly true)))]
```
matgis: Use `fn/constantly-true`?

editor/test/editor/defold_project_search_test.clj
matgis: This is only used for the first `testing` block now. I realize you might have left it this way to avoid a huge indentation change, but I think it is better to move this into that `testing` block now that it is no longer needed elsewhere.

matgis (changes requested):

editor/src/clj/editor/search_results_view.clj
JosephFerano: It's a small timing optimization, just to kick things off earlier. If it's not worth it, I'll revert this change.

editor/src/clj/editor/defold_project_search.clj
JosephFerano: Ah, this was a review suggestion from an agent. After investigating more, I think the suggestion comes from `InterruptibleChannel`

```
If a thread is blocked in an I/O operation on an interruptible channel then another thread may invoke the 
blocked thread's interrupt method. This will cause the channel to be closed, the blocked thread to receive a
ClosedByInterruptException, and the blocked thread's interrupt status to be set.
```

But it seems like it was an overly defensive suggestion, after investigating the code, there are no interruptible NIO channel operations. I'll revert.

matgis (changes requested):
You didn't fix anything. Did you forget to push?

JosephFerano:
> You didn't fix anything. Did you forget to push?

Yes I forgot :man_facepalming:

editor/src/clj/editor/defold_project_search.clj
```diff
@@ -176,27 +163,16 @@
   [workspace project start-consumer! stop-consumer! report-error!]
   (let [pending-search-atom (atom nil)
 
-        ;; Cached prep future, keyed by the ext/library filter. When the filter
-        ;; changes, the now-obsolete future is cancelled so its whole-project
-        ;; work does not keep running in the background. Reading the atom and
-        ;; resetting it is not atomic, but every caller arrives on the UI
-        ;; thread, so there is no interleaving to guard against.
-        search-data-atom (atom nil)
+        active-search-data-atom (atom nil)
```
matgis: Rename to `active-search-data-future-atom`?

editor/src/clj/editor/defold_project_search.clj
```diff
@@ -147,38 +160,57 @@
...
+                                thread (start-search-thread report-error! search-data-future resource-type->matches-fn produce-fn)
                                 consumer (start-consumer! consume-fn)]
                             {:thread thread
                              :consumer consumer})
                           (do (start-consumer! (constantly [::done]))
                               nil)))]
     {:start-search! (fn [search-string searched-exts include-libraries?]
                       (try
-                        (swap! pending-search-atom start-search! search-string searched-exts include-libraries?)
+                        (let [search-data-future (prepare-search-data! searched-exts include-libraries?)
```
matgis, line 199: There's a potential future improvement to be made here. The replacement `future` snapshots its `evaluation-context` before the previous preparation is cancelled. The previous future publishes its partial cache afterward via `run-later`, so the replacement cannot reuse those values and recomputes them during rapid filter edits. Potentially we could wait for these partial results to enter the system cache so the replacement `evaluation-context` can benefit from them. However, doing so could affect responsiveness, so let's keep it like this for now.

editor/test/editor/defold_project_search_test.clj
```diff
@@ -161,94 +160,147 @@
...
+
+      (testing "A search runs correctly after a prior search with the same filter was aborted"
+        (let [report-error! (fn/make-call-logger)
+              consumer (make-consumer report-error!)
+              start-consumer! (partial consumer-start! consumer)
+              stop-consumer! consumer-stop!
+              {:keys [start-search! abort-search!]}
+              (project-search/make-file-searcher workspace project start-consumer! stop-consumer! report-error!)]
+          (start-search! "255" nil true)
+          (is (true? (test-util/block-until true? timeout-ms consumer-finished? consumer)))
```
matgis: It looks like this test waits for the first search to finish before aborting, so it never exercises cancellation of an in-flight preparation? I suspect the test would pass even if a cancelled cached `future` were incorrectly reused?

matgis (dismissed):
I had a concern around one of the tests that we may want to address before merging. Looks good otherwise!

PR #12825, opened 2026-07-24
Improve native gamepad mapping and HID fallback support

JCash:
GameController.framework exposes semantic controls rather than the physical button and axis indices described by SDL’s `gamecontrollerdb.txt`. This caused valid SDL mappings to produce incorrect runtime input on macOS, notably for 8BitDo, Nintendo Switch Pro, and Xbox controllers.

Controllers not recognized by GameController.framework were also unavailable, and some controllers without an exact database row incorrectly fell back to legacy name-based mappings.

On Windows, we now load xinput1_4.dll in order to get the correct GUID information.

### Technical changes

- Updated the bundled SDL GameController database from the upstream SDL_GameControllerDB project.
- Preserved the complete SDL mapping row as `raw_mapping` in the compiled input data.
- Added native-driver mapping capabilities so the input system can:
  - supply the selected SDL database row to the driver at runtime;
  - use physical database mappings when available;
  - use a stable automatic semantic mapping when no database row matches.
- Updated the macOS GameController driver to synthesize the physical SDL packet described by the selected database row.
- Added direct IOKit support for controllers not handled by GameController.framework.
- Matched SDL’s macOS GUID construction, including transport-independent USB-style GUIDs and product-name CRCs.
- Added SDL-compatible Nintendo face-button handling and GameController stick-axis normalization.
- Added automatic mapping fallback for semantic controllers without an exact database entry, including newer Xbox wireless variants.
- Removed ad-hoc controller-name aliases in favor of device names and mappings supplied by the platform and SDL database.
- Improved `gdc` trigger detection by preferring axis input when mapping triggers.
- Updated Windows XInput capability detection to prefer `xinput1_4.dll`, allowing the extended product information required for accurate controller GUIDs.
- Fixed a macOS shutdown crash by releasing retained `GCController` and `IOHIDDeviceRef` objects before closing `IOHIDManager`, matching SDL’s IOKit teardown order.

- HID tests: 10/10 passed.
- Input tests: 18/18 passed.
- Manually tested GamePadTester with:
  - 8BitDo Ultimate 2C Wireless
  - Nintendo Switch Pro Controller
  - Xbox Wireless Controller
  - Sony Dualsense
  - XBox 360 Controller
- Verified the 8BitDo controller selects its SDL GUID mapping and reports correctly mapped controls.
- Verified Nintendo face buttons and stick directions.
- Verified GamePadTester exits normally with all three controllers connected.
- Confirmed no new shutdown crash report was generated.

The protobuf representation intentionally replaces the extracted 16-byte `guid` with the complete SDL `raw_mapping`. The runtime derives the GUID from the first 32 hexadecimal characters and passes the same row to semantic native drivers, avoiding duplicated or platform-specific mapping tables.


engine/engine/content/builtins/input/gamecontrollerdb.txt
```diff
@@ -546,7 +546,7 @@
 03000000250900000500000000000000,PS3 Controller,a:b2,b:b1,back:b9,dpdown:h0.8,dpleft:h0.4,dpright:h0.2,dpup:h0.1,leftshoulder:b6,leftstick:b10,lefttrigger:b4,leftx:a0,lefty:a1,rightshoulder:b7,rightstick:b11,righttrigger:b5,rightx:a2,righty:a3,start:b8,x:b0,y:b3,platform:Windows,
```
JCash, line 546: Updated with the latest from SDL
I removed my old XBox additions, figuring it should already have worked (and it does)

engine/hid/src/darwin/gamepad_darwin.mm
```diff
@@ -141,6 +135,12 @@
     AppleGamepadLegacyElement m_LegacyButtonMap[APPLE_GAMEPAD_SEMANTIC_BUTTON_COUNT];
     AppleGamepadLegacyElement m_LegacyAxisMap[APPLE_GAMEPAD_SEMANTIC_AXIS_COUNT];
     int           m_Id;
+#if TARGET_OS_OSX
+    IOHIDDeviceRef    m_HidDevice;
+    CFMutableArrayRef m_HidAxes;
+    CFMutableArrayRef m_HidButtons;
+    CFMutableArrayRef m_HidHats;
+#endif
```
JCash, line 147: Not all controllers are picked up by the GameController framework, so we still fallback on IOKit if that's the case.

engine/hid/src/hid_guid.cpp
```diff
@@ -174,63 +172,7 @@ GamepadGuid CreateGUID(uint16_t bus, uint16_t vendor, uint16_t product, uint16_t
...
-    }
-
-    if (identity.m_Vendor == USB_VENDOR_BACKBONE)
-    {
-        if (identity.m_Product == USB_PRODUCT_BACKBONE_ONE_IOS_PS5)
-            return "Backbone One PlayStation Edition";
-        if (identity.m_Product == USB_PRODUCT_BACKBONE_ONE_IOS)
-            return "Backbone One";
-    }
+    (void) identity;
```
JCash, line 175: We use the names from the raw SDL mappings instead now.

engine/input/proto/input/input_ddf.proto
```diff
@@ -227,7 +227,7 @@ message GamepadMapRuntime
     required string device = 1;         // The name of the mapping, found in the .gamepads or gamecontrollerdb.txt
     optional float dead_zone = 2;       // for the legacy mappings only
     repeated GamepadMapEntry map = 3;
-    optional bytes guid = 4;            // Single 16-byte SDL GUID. If set, it's an SDL mapping
+    optional string raw_mapping = 5;    // Canonical SDL database row, supplied to native drivers at runtime
```
JCash, line 230: We keep the physical mappings so that we can send it to the gamepad driver to figure out what needs to be done.

engine/tools/src/gdc/main.cpp
```diff
@@ -65,7 +65,7 @@ struct Driver
     Trigger m_Triggers[dmInputDDF::MAX_GAMEPAD_COUNT];
 };
 
-void GetDelta(dmHID::HGamepad gamepad, dmHID::GamepadPacket* zero_packet, dmHID::GamepadPacket* input_packet, dmInputDDF::GamepadType* gamepad_type, uint32_t* index, float* value, float* delta);
+void GetDelta(dmHID::HGamepad gamepad, dmHID::GamepadPacket* zero_packet, dmHID::GamepadPacket* input_packet, bool prefer_axis, dmInputDDF::GamepadType* gamepad_type, uint32_t* index, float* value, float* delta);
```
JCash, line 68: We should see if we can't build+distribute the SDL Gamepad Tool instead!

engine/hid/src/darwin/gamepad_darwin.mm
```diff
@@ -1546,6 +1420,7 @@ static bool SupportsController(GCController* controller, AppleGamepadDevice* dev
         device->m_ButtonCount += device->m_HasGuideButton ? 1 : 0;
         device->m_HasCaptureButton = GetCaptureButton(controller) != nil;
         device->m_ButtonCount += device->m_HasCaptureButton ? 1 : 0;
+        device->m_HasNintendoButtons = IsControllerSwitchPro(controller) || IsControllerSwitchJoyConPair(controller);
```
britzl: It feels a bit dirty to have this special Nintendo controller handling here, but maybe there is no way around it?

britzl (dismissed):

JCash:
@codex review please

engine/hid/src/darwin/gamepad_darwin.mm
JCash: it's an SDL thing that is still needed.

PR #12847, opened 2026-07-29
Cleanup callbacks on reboot() and exit()

### LINKED ISSUE #10161: Calling sys.exit() in HTML5 does not remove event listeners
Calling `sys.exit(0)` in html5 the engine will set `engine->m_Alive = false;` which means that the engine mostly is considered shut down. Event listeners set in glFW are not removed and will try to send focus events:

```
Aborted(Assertion failed: g_Window, at: ../src/platform_window_glfw.cpp,89,OnWindowFocus) 
```

**To Reproduce (REQUIRED)**
1. Call `sys.exit()` in an HTML5 build
2. Click on and outside of the canvas
3. Notice that the console will show errors

The event listeners should be removed. Possibly also other things? Maybe the whole canvas should be removed/destroyed?

  - 1.9.6

**Platforms (REQUIRED):**
 - Platforms: HTML5

ekharkunov:
This change makes sure that Javascript event listeners are removed on engine reboot and exit.

Fixes https://github.com/defold/defold/issues/10161


ekharkunov:
@codex review

AGulev (approved):

PR #12858, opened 2026-07-30
Support text shaping preview in Editor

### LINKED ISSUE #12671: Support engine text layout in the editor
We want to be able to have a WYSIWYG experience with our fonts in the editor, in particular when it comes to ligatures and layout.

Expected goals:
* Same appearance in editor as in engine
* Only use Fontc.java from bob, no more editor compilation

JCash:
We now use engine runtime to render fonts in editor. This helps us keep the same visuals in both editor and at runtime, e.g. for RTL text (arabic) or similar.

Fixes https://github.com/defold/defold/issues/12671
Fixes https://github.com/defold/defold/issues/8919

### Technical changes

* Single code path for building and rendering fonts
* Moved font glyph/vertex generation to font library
* Added Java FFM bindings `fontrender.jar` to bob/editor
* Editor uses fontrender.jar to produce vertices/texture
* We check app manifest to use legacy/full text shaping, to make it same in editor and engine
* Added `fontviewer` tool to help more rapidly develop our font system
  * After building with tests, run `./engine/font/build/arm64-macos/src/test/bin/fontviewer`

image

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


JCash:
@codex review

engine/font/src/text_layout_skribidi.cpp
```diff
@@ -189,16 +192,39 @@ static bool LayoutText(LayoutContext* ctx,
...
         l.m_Length  = glyph_index - prev_glyph_index;
         layout->m_Lines.Push(l);
     }
 
+    // SkriBidi includes an empty line after a terminal newline. The legacy
+    // layout does not, so discard it here as well.
+    if (!layout->m_Lines.Empty() &&
+        layout->m_Lines.Back().m_Length == 0 &&
+        (codepoints[num_codepoints - 1] == dmUtf8::UTF_WHITESPACE_NEW_LINE ||
+         codepoints[num_codepoints - 1] == dmUtf8::UTF_WHITESPACE_CARRIAGE_RETURN))
```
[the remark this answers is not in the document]
JCash: it's already handled in TextLayoutSkribidiCreate

JCash:
@codex review

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/font/Fontc.java
```diff
@@ -14,1198 +14,552 @@
 
 package com.dynamo.bob.font;
 
-import java.awt.BasicStroke;
```
JCash, line 17: Relying solely on a single code path (engine) for rendering fonts now.

editor/src/clj/editor/app_manifest.clj
```diff
@@ -981,4 +981,4 @@
     :node-type AppManifestNode
     :view-types [:code :default]
     :view-opts {:code {:use-custom-editor false}}
-    :lazy-loaded true))
+    :lazy-loaded false))
```
JCash, line 1024: Apparently needed in order to check the info from the app manifest.

editor/src/clj/editor/defold_project.clj
```diff
@@ -952,6 +952,7 @@
         (g/connect script-intelligence :build-errors game-project :build-errors)
         (g/connect game-project :display-profiles-data project :display-profiles)
         (g/connect game-project :texture-profiles-data project :texture-profiles)
+        (g/connect game-project :use-font-layout project :use-font-layout)
```
JCash, line 955: We want to alter behavior of the text shaping, depending on what the engine will use (wysiwyg).
Similar to texture profiles.

editor/src/clj/editor/font.clj
```diff
@@ -13,7 +13,8 @@
 ;; specific language governing permissions and limitations under the License.
 
 (ns editor.font
```
JCash, line 15: The font_gen.clj/fontc.clj were removed, as editor doesn't generate or compile any parts of the font now, and little code remained there.

editor/src/clj/editor/font.clj
```diff
@@ -56,6 +56,51 @@
 
 (set! *warn-on-reflection* true)
 
+(def ^String default-characters-string (String. Fontc/ASCII_7BIT))
```
JCash, line 59: Moved from the old files.

editor/test/integration/build_test.clj
```diff
@@ -549,7 +549,7 @@
             glyph-bank-bytes (content-bytes {:resource glyph-bank-build-path})
             glyph-bank (protobuf/bytes->map-with-defaults Font$GlyphBank glyph-bank-bytes)]
         (is (= 1024 (:cache-width glyph-bank)))
-        (is (= 256 (:cache-height glyph-bank))))))
+        (is (= 512 (:cache-height glyph-bank))))))
```
JCash, line 552: The old code and engine code is not exactly 1:1, so some tests need adjustments.

engine/font/scripts/generate_font_renderer_bindings.py
```diff
@@ -0,0 +1,173 @@
+#!/usr/bin/env python3
```
JCash, line 1: Generate the Java FFM files

engine/font/src/java/com/dynamo/bob/font/FontRenderer.java
```diff
@@ -0,0 +1,538 @@
+// Copyright 2020-2026 The Defold Foundation
```
JCash, line 1: The public FontRenderer Java api

engine/render/src/render/font/font_renderer.cpp
```diff
@@ -31,10 +28,11 @@
 #include "fontmap_private.h"
 #include "font_renderer_private.h"  // for FontMap, RenderLayerMask
 
-#include "font_renderer.h"       // for FontGlyphCompression
-#include "font_renderer_api.h"   // for the font renderer backend api
+#include "font_renderer.h"          // for FontGlyphCompression
```
JCash, line 31: The old api was never used, and we'll use the font library as the api instead anyways.

engine/font/src/render/font_render.h
```diff
@@ -0,0 +1,27 @@
...
+#define DM_FONT_RENDER_H
+
+#include <dmsdk/graphics/graphics.h>
+
+/**
+ * Creates a graphics vertex declaration matching FontGlyphVertex.
+ * The caller owns the returned declaration and must delete it with
+ * dmGraphics::DeleteVertexDeclaration().
+ */
+dmGraphics::HVertexDeclaration FontCreateGlyphVertexDeclaration(dmGraphics::HContext context);
```
JCash, line 25: In the future, we can make sure it supports custom vertex formats

scripts/build.py
```diff
@@ -478,7 +478,7 @@ def get_default_target_platforms():
     if 'WD' in os.environ:
         SHELL= '%s\\bash.exe' % os.environ['WD'] # the binary directory
 
-ENGINE_LIBS = "testmain dlib jni texc modelc shaderc ddf platform font graphics particle lua hid input physics resource extension script render rig gameobject gui sound liveupdate crash gamesys tools record profiler engine sdk".split()
+ENGINE_LIBS = "testmain dlib jni texc modelc shaderc ddf platform graphics font particle lua hid input physics resource extension script render rig gameobject gui sound liveupdate crash gamesys tools record profiler engine sdk".split()
```
JCash, line 481: the font render library depends on the vertex declaration, and also we use it for the new fontviewer tool

CMakeLists.txt
```diff
@@ -176,6 +185,10 @@ set(DEFOLD_HOST_LIBS
   jni
   texc
   modelc
+  ddf
+  platform
+  graphics
+  font
```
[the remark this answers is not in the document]
JCash: I'll look into this.

vlaaad (changes requested):
I found 2 issues.

### Static font preview ignores character set

The editor preview ignores the character set compiled into static fonts.

`native-layout-text` measures text using a renderer created from the full font file:

https://github.com/defold/defold/blob/899f250ce7af2d8d741d1e6c047518b304595396/editor/src/clj/editor/font.clj#L432-L446

`apply-native-entry-state!` sends the same unrestricted text for rendering:

https://github.com/defold/defold/blob/899f250ce7af2d8d741d1e6c047518b304595396/editor/src/clj/editor/font.clj#L720-L723

The renderer receives the complete TTF/OTF bytes here:

https://github.com/defold/defold/blob/899f250ce7af2d8d741d1e6c047518b304595396/editor/src/clj/editor/font.clj#L977-L982

Repro project: 

[static-font-character-preview.zip](https://github.com/user-attachments/files/30891857/static-font-character-preview.zip)

The attached repro contains a static font whose character set only includes `A`, while the label contains `A Ж`. The editor displays both characters, but the running game can only display `A` because the static glyph bank does not contain `Ж`. Measurement and wrapping can disagree for the same reason.

Using the complete font is correct for runtime-generated fonts, but static font previews need to be restricted to the compiled glyph set.

### Performance regression

`fontc.clj` was much faster than the Java-based font compiler. In the same repro project, if you open a `repro.font` and click on the `All chars` toggle:
- the old editor recompiles the font in around 1 second
- this branch recompiles the font in around 1m 10s - **70 times slower**!

This has repercussions throughout the editor — any scene open that uses fonts gets very slow. Can we do a perf improvement pass over the Java font compiler as part of this task? Or should we perhaps make a separate high-priority issue for this?

engine/font/src/fontc/fontc.cpp
```diff
@@ -0,0 +1,905 @@
...
+    TextLayoutRelease(layout);
+    return FONT_RENDERER_RESULT_OK;
+}
+
+FontRendererResult FontcGenerateGlyph(HFontRenderer renderer, uint32_t codepoint, FontcGlyph* output)
+{
+    if (!renderer || !output)
+        return FONT_RENDERER_RESULT_INVALID_ARGUMENT;
+    memset(output, 0, sizeof(*output));
+    const uint32_t     glyph_index = FontGetGlyphIndex(renderer->m_Font, codepoint);
```
AGulev, line 555: that's what Sol Max found: [P1] Skip missing glyphs before generation

When `FontGetGlyphIndex` returns 0, this still rasterizes and copies the `.notdef` glyph before Java discards it. The `all_chars` path iterates roughly 1.1 million code points, so most iterations repeatedly generate the same missing glyph. Return an empty result with glyph index 0 before calling `FontGenerateGlyph`.

engine/font/src/java/com/dynamo/bob/font/FontRenderer.java
```diff
@@ -0,0 +1,538 @@
...
+    GeneratedGlyph generateGlyphMetrics(int codepoint) {
+        return generateGlyph(codepoint, false);
+    }
+
+    private synchronized GeneratedGlyph generateGlyph(int codepoint, boolean copyPixels) {
+        if (!Character.isValidCodePoint(codepoint))
+            throw new IllegalArgumentException("Invalid Unicode codepoint");
+        try (Arena arena = Arena.ofConfined()) {
+            MemorySegment result = FontcGlyph.allocate(arena);
+            int status = FontcGenerateGlyph(requireHandle(), codepoint, result);
```
AGulev, line 339: that's what Sol Max found: [P2] Make the metrics-only path skip rasterization

`copyPixels=false` only suppresses the final Java copy. This call still generates the complete SDF, allocates a second native pixel buffer in `FontcGenerateGlyph`, copies it, and immediately frees it. Consequently editor metadata-only compilation still rasterizes every supported glyph. Add a native metrics-only entry point or generation flag that uses `m_GenerateImage=false` and derives the required dimensions without creating pixels.

engine/font/src/render/glyph_gen.cpp
```diff
@@ -0,0 +1,133 @@
...
+        {
+            const uint8_t  value = glyph->m_Bitmap.m_Data[y * width + x];
+            const uint32_t offset = (y * width + x) * channels;
+            if (params->m_OutputBitmap)
+            {
+                rgb[offset + 0] = SdfCoverage(value, params->m_SdfEdgeValue, pixel_dist_scale, params->m_Antialias);
+                if (channels == 3)
+                {
+                    rgb[offset + 1] = params->m_HasOutline ? SdfCoverage(value, outline_edge, pixel_dist_scale, params->m_Antialias) : 0;
+                    rgb[offset + 2] = params->m_HasShadow ? SdfCoverage(value, shadow_edge, pixel_dist_scale, true) : 0;
```
AGulev: that's what Sol Max found: [P2] Preserve bitmap shadow blur coverage

For `TYPE_BITMAP`, the previous implementation formed the shadow from the face plus outline and applied `shadow_blur` convolution passes. This instead produces a thresholded SDF expansion; `SdfCoverage` provides only edge antialiasing, and `shadow_edge` omits the outline width. Bitmap fonts therefore get a hard expanded shadow, potentially narrower than their outline. Restore graduated blur coverage based on the face-plus-outline silhouette and add a pixel-level regression test.

AGulev (commented):
that's what Sol Max found: three inline findings.

AGulev:
it should probabbly fix https://github.com/defold/defold/issues/8919 as well?

JCash:
Updated with fixes and added extra performance boost when getting metrics/sdf bitmaps from native code.
1+ second -> ~0.5s

servers.html
```diff
@@ -0,0 +1,549 @@
....queue.length===0&&Object.keys(O.impl).length>0)){if(v.includes(r)){N(r,{code:"capability_disabled",message:"capability not available in this session"});continue}N(r,{code:"not_granted",message:"capability not available in this session"})}await _,parent.postMessage({__frame_ready:!0},R);let E=-1,A=0;const j=()=>{A=0;const r=Math.max(document.documentElement.scrollHeight,document.body?document.body.scrollHeight:0);r!==E&&(E=r,parent.postMessage({__frame_size:!0,h:r},a))},b=()=>{A||(A=setTimeout(j,200))},U=new ResizeObserver(b);U.observe(document.documentElement),document.body&&U.observe(document.body),document.addEventListener("load",b,!0),document.fonts?.ready.then(b,()=>{}),addEventListener("beforeprint",()=>{E=-1,j()}),b(),addEventListener("message",r=>{if(!(r.source!==parent||r.origin!==a)&&Y(r.data)){try{e?.promoted()}catch{}E=-1,b()}})})}})();</script><!-- /frame-runtime --><meta charset=utf8><meta name=viewport content="width=device-width,initial-scale=1"><title>Current build server costs</title><style>:root{color-scheme:light}body{margin:0;padding:0;font:14px -apple-system,BlinkMacSystemFont,sans-serif;background:#faf9f5;color:#141413}img{max-width:100%}</style></head><body>
```
AGulev: what is this? tdo we need this file?

com.dynamo.cr/com.dynamo.cr.bob/src/com/dynamo/bob/pipeline/GlyphBankBuilder.java
```diff
@@ -64,38 +77,31 @@ public void build(Task task) throws CompileExceptionError, IOException {
...
         FontDesc fontDesc = builder.build();
 
-        final IResource inputFontFile = BuilderUtil.checkResource(this.project, task.firstInput(), "font", fontDesc.getFont());
-        BufferedInputStream fontStream = new BufferedInputStream(new ByteArrayInputStream(inputFontFile.getContent()));
+        IResource inputFontFile = BuilderUtil.checkResource(this.project, task.firstInput(), "font", fontDesc.getFont());
+        byte[] fontBytes = inputFontFile.getContent();
+        String bitmapPath = null;
+        byte[] bitmapBytes = null;
+        if (fontDesc.getFont().toLowerCase(Locale.ROOT).endsWith(".fnt")) {
+            bitmapPath = getBitmapPath(fontBytes);
```
vlaaad: This line may throw IOException (which is really a `BMFontFormatException`). should it be caught here and converted to `CompileExceptionError`?

vlaaad (dismissed):
LGTM!

servers.html
JCash: wow, yeah that's not part of the pr :/

engine/font/src/fontc/fontc.cpp
```diff
@@ -0,0 +1,987 @@
...
+    CachedGlyph new_glyph;
+    memset(&new_glyph, 0, sizeof(new_glyph));
+    new_glyph.m_Font = font;
+    new_glyph.m_Frame = session->m_Frame;
+    new_glyph.m_GlyphIndex = glyph_index;
+
+    FontGlyphGenParams params;
+    GetGlyphGenParams(session, font, &params);
+    if (FontGenerateGlyph(font, glyph_index, &params, &new_glyph.m_Glyph) != FONT_RESULT_OK)
+        return 0;
```
AGulev, line 328: [P1] Reject glyphs without bitmap data

FontGenerateGlyph can return success for mapped but non-rendering glyphs such as U+200D. The glyph is cached and counted as visible, then FontPackGlyphVertices asserts because its bitmap width is zero. I reproduced this with Noto Sans: vertex sizing returned 6 vertices and 576 bytes, followed by an abort in glyph_vertex.cpp:69. Reject and free zero-sized glyphs before caching, or consistently exclude them from vertex generation.

editor/src/clj/editor/font.clj
```diff
@@ -615,7 +961,54 @@
...
 
-(defn- make-font-map [_node-id font type pb-msg font-resource-resolver]
+(defn- make-native-renderer-spec
+  ^NativeRendererSpec [font font-desc font-map use-font-layout runtime-generation]
+  (let [render-params (FontRenderer$Params.)
+        measure-params (FontRenderer$Params.)
+        shadow-blur (double (if (and (pos? ^double (:shadow-alpha font-desc))
+                                     (pos? ^double (:alpha font-desc)))
+                              (:shadow-blur font-desc)
+                              0.0))
```
AGulev, line 970: P2] Preserve blur for shadow-only single-layer fonts

The editor disables shadow blur whenever face alpha is zero, while Bob conditions blur only on shadow alpha. In single-layer mode the shader applies face and shadow colors independently, so alpha=0 with a visible blurred shadow is valid. The built game renders blurred shadow-only text, but the editor generates different padding and a hard shadow. Gate blur only on shadow-alpha to match compilation.

editor/src/clj/editor/font.clj
```diff
@@ -615,7 +961,54 @@
...
+    (set! (.-sdfOutline render-params) (float (native-sdf-limit sdf-padding outline-width)))
+    (set! (.-sdfShadow render-params) (float (if (zero? shadow-blur)
+                                              1.0
+                                              (native-sdf-limit sdf-padding shadow-blur))))
+    (set! (.-outlineWidth render-params) (float outline-width))
+    (set! (.-shadowBlur render-params) (float shadow-blur))
+    (set! (.-shadowX render-params) (float (:shadow-x font-desc)))
+    (set! (.-shadowY render-params) (float (:shadow-y font-desc)))
+    (set! (.-layerMask render-params) (int (:layer-mask font-map)))
+    (set! (.-useTextShaping render-params) (boolean use-font-layout))
```
AGulev: [P1] Only shape fonts generated at runtime

useTextShaping ignores runtime-generation. Offline fonts are built as glyph banks and force legacy layout at runtime, while the editor retains the TTF and enables full shaping from the app manifest. This makes static fonts preview kerning, ligatures, bidi, and potentially different fallback behavior that the built game cannot reproduce. Gate this with both use-font-layout and runtime-generation, and test static and dynamic configurations separately.

AGulev (dismissed):

PR #12860, opened 2026-07-30
Restore graphics resources

### LINKED ISSUE #8200: Handle WebGL context loss
WebGL context can be lost when not enough GPU resources are available. 

I don't know how the game can restore it, or if it should. But at the very least, it should have a handler for this error. Because in the logs of the error reporting system I see _indirect_ errors that there is no WebGL context and consequently players see a frozen game:

- `TypeError: Argument 1 ('shader') to WebGL2RenderingContext.shaderSource must be an instance of WebGLShader`
- `TypeError: Failed to execute 'getAttribLocation' on 'WebGL2RenderingContext': parameter 1 is not of type 'WebGLProgram'`
- `TypeError: WebGLRenderingContext.shaderSource: Argument 1 is not an object.`

etc.

No alternatives.

https://www.khronos.org/webgl/wiki/HandlingContextLost

ekharkunov:
Recovering from a lost graphics context (WebGL / HTML5)

On HTML5 the browser may take the WebGL context away at any moment (tab in background, GPU reset, driver update) and later give it back. Previously Defold survived the loss (rendering was paused) but the scene never came back: every texture, shader, buffer and render target on the GPU was gone for good.

With this change the engine can now fully recreate all GPU resources after a context restore, driven by a single call from your render script:

* `render.reload_resources(finish_callback, [progress_callback])` - recreates, in chunks spread across frames (so the page stays responsive), every GPU resource that was created from resource files: textures, shaders/programs, buffers, meshes, models. Works in both debug and release builds.
* Render targets are restored automatically, including ones created at runtime with render.render_target(). Their contents are transient and are simply re-rendered on the next frame.
* Fonts work: the glyph cache detects the loss and rebuilds itself lazily - text reappears without any action on your part.
* Sprites, particles, streamed geometry self-heal because they re-upload their vertex data every frame.
* Rendering stays paused during the reload and resumes automatically when it finishes.
* Textures created at runtime from memory (resource.create_texture) have no file to reload from: their handles stay valid, but your app must upload the pixels again (resource.set_texture) - the finish callback is a good place for that.
* The builtins archive is now kept in memory by default (`dmengine.unload_builtins` defaults to 0), so builtin materials/shaders/fonts can be restored too. If you set it back to 1 to save memory, builtin resources cannot be reloaded after a context loss.

## How to use it (Lua)

Subscribe to context events with render.set_listener and start the reload when the context comes back:

```Lua
-- custom.render_script
function init(self)
    -- ... your usual render setup ...

    render.set_listener(function(self, event_type)
        if event_type == render.CONTEXT_EVENT_CONTEXT_LOST then
            -- Rendering is paused, all GPU resources are invalid.
            -- Don't issue any draw calls; just note the state if you need to.
            print("Graphics context lost")
        elseif event_type == render.CONTEXT_EVENT_CONTEXT_RESTORED then
            -- The context is back; start recreating GPU resources.
            render.reload_resources(
                function(self)
                    print("All resources recreated, rendering resumed")
                    -- Good place to re-upload textures you created with
                    -- resource.create_texture / resource.set_texture.
                end,
                function(self, done, total)
                    print("Restoring " .. done .. "/" .. total)
                end)
        end
    end)
end
```

That is the entire integration for a typical game: two callbacks, no resource lists, no per-asset code.

## Reacting to context loss/restore in native extensions

Native extensions get two new lifecycle events in their `OnEvent` callback (declared in `dmsdk/extension/extension.h`):

* `EXTENSION_EVENT_ID_GRAPHICS_CONTEXT_LOST` - sent right after the engine has paused rendering and invalidated all graphics handles. Mark your GPU state as dead and stop touching the graphics API.
* `EXTENSION_EVENT_ID_GRAPHICS_CONTEXT_RESTORED` - sent when the context is usable again: the new context is e been re-enabled and render targets have been restored. This is the safe point to recreate any GPU state of your extension owns outside Defold's resource system (own FBOs, custom renderer contexts, etc.). Note that at this moment the engine's resource work list has not been re-uploaded yet - if you need "after all resources are back", wait for the first rendered frame (rendering resumes when the reload finishes).

```cpp
#include <dmsdk/sdk.h>

static void OnEvent(dmExtension::Params* params, const dmExtension::Event* event)
{
    switch (event->m_Event)
    {
    case EXTENSION_EVENT_ID_GRAPHICS_CONTEXT_LOST:
        // GPU objects are gone. Flag them as dead; do NOT call into the graphics API.
        MyRenderer_MarkDeviceLost();
        break;
    case EXTENSION_EVENT_ID_GRAPHICS_CONTEXT_RESTORED:
        // Context is live again: recreate any private GPU state (FBOs, contexts, ...).
        MyRenderer_RecreateDeviceObjects();
        break;
    default:
        break;
    }
}

DM_DECLARE_EXTENSION(MyExtension, "MyExtension", AppInitialize, AppFinalize,
                     Initialize, Update, OnEvent, Finalize)
```

## Making an extension's resource type restorable

If your extension registers its own resource type that owns GPU objects (the spine/rive pattern), opt the tources with `ResourceTypeSetGraphicsRestoreOrder` (`dmsdk/resource/resource.h`). The order value is the recreation bucket - lower runs first: `RESOURCE_GRAPHICS_RESTORE_ORDER_TEXTURES (0)`, `..._PROGRAMS (1)`, `..._BUFFERS (2)`.

The contract: during a context restore your recreate function is called with `ResourceRecreateParams::m_IsContextRestore == 1`. The source bytes are unchanged - restore the GPU objects in place, keeping all handles and pointers stable (components may cache raw pointers into your resource). The engine's zero-guards regenerate the underlying GL handles, so re-uploading content through the normal `dmGraphics::Set*Data` calls is all you need.

```cpp
static ResourceResult MyResourceRecreate(const ResourceRecreateParams* params)
{
    MyResource* resource = (MyResource*) ResourceDescriptorGetResource(params->m_Resource);

    if (params->m_IsContextRestore)
    {
        // Context restore: bytes are unchanged, GL names were invalidated.
        // Re-upload into the SAME buffer handles — the engine regenerates the
        // underlying GL objects in place. Do not delete/recreate handles here.
        dmGraphics::SetVertexBufferData(resource->m_VertexBuffer,
                resource->m_VertexDataSize, resource->m_VertexData,
                dmGraphics::BUFFER_USAGE_STATIC_DRAW);
        dmGraphics::SetIndexBufferData(resource->m_IndexBuffer,
                resource->m_IndexDataSize, resource->m_IndexData,
                dmGraphics::BUFFER_USAGE_STATIC_DRAW);
        return RESOURCE_RESULT_OK;
    }

    // Normal hot-reload: bytes changed, do your usual full recreate.
    return RecreateFromBytes(resource, params->m_Buffer, params->m_BufferSize);
}

static ResourceResult MyResourceTypeRegister(HResourceTypeContext ctx, HResourceType type)
{
    ResourceTypeSetContext(type, ...);
    ResourceTypeSetCreateFn(type, MyResourceCreate);
    ResourceTypeSetDestroyFn(type, MyResourceDestroy);
    ResourceTypeSetRecreateFn(type, MyResourceRecreate);
    // Opt in to render.reload_resources: recreated in the "buffers" bucket.
    ResourceTypeSetGraphicsRestoreOrder(type, RESOURCE_GRAPHICS_RESTORE_ORDER_BUFFERS);
    return RESOURCE_RESULT_OK;
}

DM_DECLARE_RESOURCE_TYPE(ResourceTypeMyRes, "myresc", MyResourceTypeRegister, 0);
```

## Component-owned GPU buffers (upload-once / dirty-flag pattern)

Components that generate vertices every frame (the sprite/particle pattern) need nothing - the per-frame upload heals them. Components that upload once and then only on a dirty flag must notice the loss themselves. Use
`dmGraphics::GetInvalidationGeneration` (`dmsdk/graphics/graphics.h`): the counter increments on every context your world and compare each update. The buffer handle stays valid - just re-upload the CPU-side data into it:

```cpp
struct MyComponentWorld
{
    dmGraphics::HContext      m_GraphicsContext;
    dmGraphics::HVertexBuffer m_VertexBuffer;    // uploaded once at creation
    dmArray<uint8_t>          m_VertexData;      // keep the CPU copy!
    uint32_t                  m_ObservedInvalidationGeneration;
};

// On world creation:
world->m_ObservedInvalidationGeneration =
        dmGraphics::GetInvalidationGeneration(graphics_context);

// In the component's Update:
uint32_t generation = dmGraphics::GetInvalidationGeneration(world->m_GraphicsContext);
if (generation != world->m_ObservedInvalidationGeneration)
{
    world->m_ObservedInvalidationGeneration = generation;
    // The GPU-side storage was wiped by a context loss. The handle is still
    // valid — re-upload the retained CPU data into the same buffer.
    dmGraphics::SetVertexBufferData(world->m_VertexBuffer,
            world->m_VertexData.Size(), world->m_VertexData.Begin(),
            dmGraphics::BUFFER_USAGE_STATIC_DRAW);
}
```

The key requirement: keep a CPU-side copy (or be able to regenerate the data).

Fixes #8200 

### Technical details
Restoration is two-phasing:

- **Phase A** - regenerate names in place ("zero-guards"). `EnsureGLTexture` / `EnsureGLBuffer` / `EnsureGLRenderbuffer`: at every bind/upload site, if the slot's GL name is 0, glGen* a new name into the same slot. Same pattern for programs/shaders in `ReloadShader`/`OpenGLReloadProgram` (`glCreateShader`/`glCreateProgram` into the existing slot).
- **Phase B** - re-upload content. New production-safe resource path `dmResource::RecreateResource(factory, name or hash)` (no filename table - works in release, where RELOAD_SUPPORT is off), reads the source bytes by hash from the mounts (`LoadResourceByHash`), calls the type's `m_RecreateFunction` with `ResourceRecreateParams::m_IsContextRestore = 1`, and doesn't fire `ResourceReloadedCallbacks` - correctness depends on handle indirection, not notifications.

**Chunked reload driver** (render.cpp): `render.reload_resources` builds a work list via `dmResource::IterateResources`, and `UpdateResourceReload` (called from `engine.cpp:StepFrame`, outside the do_render block so it runs while rendering is paused) processes items under a **~4 ms/frame** time budget, invoking the Lua progress/finish callbacks and resume rendering on completion. Individual item failures are logged and skipped, never abort the whole restore.

#### Per-subsystem highlights

Work list bucketing is type-driven, not hardcoded. `ResourceType` structure stores `m_GraphicsRestoreOrder`; set via dmsdk `ResourceTypeSetGraphicsRestoreOrder`; get via `ResourceTypeGetGraphicsRestoreOrder`. Engine types set restore order from their own register functions. `BuildReloadWorklist` queries the type through dmResource::GetResourceTypeGraphicsRestoreOrder. materialc/fontc are excluded (they own no GPU objects / manage a runtime cache).

Render targets are restored wholesale in the graphics layer, not through the resource system - because the most common RTs are runtime-created from render scripts and aren't resources at all, and RT contents are transient anyway. The shared `GraphicsContext` keeps an explicit `m_RenderTargets` list maintained in the `NewRenderTarget`/`DeleteRenderTarget`. `OpenGLRestoreRenderTargets` called from `OpenGLRecreateGraphicsHandles`, regenerates each dead FBO into its original slot, clears the `m_Attached` flags and re-runs `ApplyRenderTargetAttachments`, which re-allocates attachment storage into the existing texture/renderbuffer handles via `Ensure` functions.

**Synthetic resources are skipped**. `render_targetc` creates attachment texture wrappers via `CreateResource` from their bytes, so recreate-by-hash can never succeed. `ResourceDescriptor` stores an `m_CreatedFromMemory` bit. The work list checks `dmResource::IsResourceCreatedFromMemory` and skip id it's true.

**Sampler-state cache**. A freshly regenerated GL texture has default parameters (mipmapped min-filter). On name regeneration, the cached `ApplySamplerState` is re-applied to the fresh object.

**Shader pipeline fixes (pre-existing bugs exposed by restore)**: `SetupGraphicsProgram` never stored `m_VertexShader` program (broke reload and leaked shaders); `ReloadShader` now checks GL_COMPILE_STATUS instead of failing silently.

**WebGL/Emscripten specifics**.
* Extension objects from `getExtension()` are invalidated with the context. So all `emscripten_webgl_enable_extension` calls were extracted into `OpenGLEnableWebGLExtensions()` and re-run it as the first step of `OpenGLRecreateGraphicsHandles` (VAO/instancing/compressed-texture extensions must be live before any re-upload).
* `loseContext()` kills GL immediately but the DOM event arrives a tick later, so one engine frame always run_GL_ERROR/CHECK_GL_FRAMEBUFFER_ERROR now tolerate that frame (`IsNativeContextLost()` via `emscripten_is_webgl_context_lost()`.

**Invalidation generation**. `GraphicsContext::m_InvalidationGeneration` is bumped in the `InvalidateGraphicsHandles` dispatcher and exposed as dmsdk `dmGraphics::GetInvalidationGeneration`. Consumers cache-and-compare: `comp_model` marks all dynamic per-instance attribute buffers dirty (existing dirty path re-uploads into the same handles), and the font `UpdateCacheTexture`) forces a full `ResetCache` - storage is re-allocated into the same `HTexture` and glyphs re-upload lazily as text is drawn.

**Event wiring**. render.cpp got `SetRenderContextEventListener`; engine.cpp registers `OnRenderContextEvent`, which forwards to extensions via `dmExtension::DispatchEvent`. `GRAPHICS_CONTEXT_LOST` is dispatched from `OnContextEvent` after pause+invalidate; `GRAPHICS_CONTEXT_RESTORED` is dispatched from `StartResourceReload`.

**Builtins**. `dmengine.unload_builtins` default flipped 1 -> 0: the builtins mount must stay alive so builtin shaders/materials remain byte-addressable for recreate-by-hash (no per-type byte caching in the graphics layer - the mount is the byte source).

P.S. I was thinking about how to implement functionality which user can use to show preloader screen. Right now I don't have good idea so I suggest discuss it separately. 

## PR checklist

* [ ] Code
* [ ] Documentation

----------

   1. `### Technical changes` 
   1. `Technical changes:`
   2. `Technical notes:`

```

Fixes #1234

### Technical changes
```


engine/gamesys/src/gamesys/components/comp_model.cpp
```diff
@@ -2344,13 +2348,37 @@ namespace dmGameSystem
         return false;
     }
 
+    // After a lost graphics context, the custom vertex attribute buffers (uploaded once and then
+    // only on attribute changes) hold no GPU data anymore. Their source data is CPU-side, so marking
+    // every render item dirty makes the existing dispatch path re-upload it into the same buffer
+    // handles the next time the items are drawn.
+    static void MarkDynamicVertexAttributesDirty(ModelWorld* world)
```
JCash, line 2355: I honestly didn't expect any changes to this type of user code.

My gut feeling is that the detection happens somewhere in the engine update loop, and then we avoid calling the gameworld update/render?

It feels strange that this particular component type's render items are singled out for special "context lost" treatment?

engine/gamesys/src/gamesys/resources/res_model.cpp
```diff
@@ -245,6 +245,45 @@ namespace dmGameSystem
         }
     }
 
+    // Graphics context restore: re-upload the existing CPU-side mesh data into the SAME buffer
+    // handles. SetVertexBufferData/SetIndexBufferData regenerate the underlying GPU objects in
+    // place, so the ModelResourceBuffers pointers and buffer handles cached by components (e.g.
+    // MeshRenderItem::m_Buffers in comp_model) stay valid — unlike the full Release+Acquire path,
+    // which would allocate new ModelResourceBuffers and leave live components with dangling pointers.
+    static void RestoreBufferContents(ModelResource* resource)
```
JCash, line 253: I don't think we really need two separate functions for this?
I feel "restore" buffer contents should be able to be the same as our current way of setting the contents in the first place?

engine/graphics/src/opengl/graphics_opengl.cpp
```diff
@@ -5504,6 +5756,36 @@ static void LogFrameBufferError(GLenum status)
         memset(context->m_GLHandlesData.m_AllGLHandles.Begin(), 0, (context->m_GLHandlesData.m_AllGLHandles.End() - context->m_GLHandlesData.m_AllGLHandles.Begin()) * sizeof(uint32_t));
     }
 
+    static void OpenGLRecreateGraphicsHandles(HContext _context)
```
JCash, line 5759: A bit similar here.
Shouldn't this part be part of the regular "init" setup?
That way we make sure that this code is maintained and working.

engine/graphics/src/webgpu/graphics_webgpu.cpp
```diff
@@ -3632,6 +3632,7 @@ static void WebGPUCloseWindow(HContext _context)
 }
 
 static void WebGPUInvalidateGraphicsHandles(HContext context) { }
+static void WebGPURecreateGraphicsHandles(HContext context) { }
```
JCash, line 3635: I would assume that WebGPU (as all backends do) suffers from the same issue.
If the users GPU evicts stuff, we need to recreate/upload things.

Are no backend changes needed in this case?

engine/render/src/render/font/fontmap.cpp
```diff
@@ -586,6 +587,17 @@ namespace dmRender
         bool texture_too_small = IsTextureTooSmall(font_map);
         bool update_cache = font_map->m_IsCacheSizeDirty || texture_too_small;
 
+        // A lost graphics context wiped both the cache texture storage and the cached glyph
+        // pixels; the glyphs are only ever uploaded via sub-updates on cache misses, so the
+        // bookkeeping must be reset too — ResetCache re-allocates the storage and makes every
+        // glyph re-upload lazily as it is drawn again.
+        uint32_t invalidation_generation = dmGraphics::GetInvalidationGeneration(font_map->m_GraphicsContext);
```
JCash, line 594: Can we not just have a dmGraphics::IsContextLost() check?
I'd like to avoid the extra book keeping on the lower level structures, when the caller already knows what to do.

```c++
// from the font_renderer.cpp
if (dmGraphics::IsContextLost(graphics_context))
{
    UpdateCacheTexture(font_map, true); // force update the cache
}
```

engine/render/src/render/render.cpp
```diff
@@ -1338,6 +1377,209 @@ namespace dmRender
...
+
+        DM_PROFILE("UpdateResourceReload");
+
+        const uint64_t budget_us = 4000; // ~4 ms/frame to keep a single-threaded (HTML5) page responsive
+        uint64_t start = dmTime::GetMonotonicTime();
+
+        while (state->m_Cursor < state->m_Total)
+        {
+            const ResourceReloadItem& item = state->m_Worklist[state->m_Cursor];
+            dmResource::Result r = dmResource::RecreateResource(state->m_Factory, item.m_NameHash);
```
JCash, line 1537: This just collects the hashes.
The user may load resources from any live update package.

There is no guarantee that you'll get the same resource as before.
And for more complex types like materials, that expose properties from the shaders, you may get a different behavior.

JCash (commented):
I have concerns/questions about
1) design around reloading live update content (which may have been unmounted)
1) some code cleanup/simplification
1) context lost handling on e.g. WebGPU?

Some llm review comments:
[P1] Uniform buffers remain invalid after restoration. [OpenGLInvalidateGraphicsHandles()](https://github.com/defold/defold/blob/93b155f013a4bf250eda5461f21c80e4e5901540/engine/graphics/src/opengl/graphics_opengl.cpp#L5751-L5756) zeros every GL handle, including HUniformBuffer handles. Textures and vertex/index buffers gained EnsureGL* paths, but [OpenGLSetUniformBuffer()](https://github.com/defold/defold/blob/93b155f013a4bf250eda5461f21c80e4e5901540/engine/graphics/src/opengl/graphics_opengl.cpp#L2329-L2347) still binds the zero handle. The render context’s light UBO is one affected engine-owned object: after restoring a WebGL2 context, lighting uploads either produce GL_INVALID_OPERATION or leave the shader without light data. OpenGLSetUniformBuffer() needs to regenerate and allocate storage when its handle is zero, or all live UBOs must be restored centrally.

[P2] A second loss during chunked reload completes against the dead context. [UpdateResourceReload()](https://github.com/defold/defold/blob/93b155f013a4bf250eda5461f21c80e4e5901540/engine/render/src/render/render.cpp#L1521-L1561) continues advancing the cursor even while rendering is paused, while [OnContextEvent(CONTEXT_LOST)](https://github.com/defold/defold/blob/93b155f013a4bf250eda5461f21c80e4e5901540/engine/render/src/render/render.cpp#L1588-L1600) leaves the active reload state untouched. If WebGL is lost again during a multi-frame reload, re-creates run as GL no-ops, the next restored listener cannot restart because a reload is already active, and the old reload eventually unpauses rendering with missing objects. A new loss should suspend/cancel the reload and reset its cursor/worklist for the next restore.

[P2] Program reflection rebuilding leaks per context restoration. [RebuildProgramUniformsAfterRestore()](https://github.com/defold/defold/blob/93b155f013a4bf250eda5461f21c80e4e5901540/engine/graphics/src/opengl/graphics_opengl.cpp#L4046-L4054) frees only m_BlockMemory. BuildUniformBuffers() subsequently memsets the existing OpenGLScratchUniformBuffer entries, losing their m_Indices/m_Offsets allocations and old GL-handle slots. OpenGLBuildUniforms() also overwrites the previous Uniform::m_Name strings. Repeated browser context losses therefore steadily leak CPU memory and handle-table entries; the old reflection state needs full cleanup before rebuilding.

JCash (changes requested):
as seen in the comment above

PR #12872, opened 2026-08-02
Add App Manifest options to exclude GUI, Particle FX, and Tilemaps

### LINKED ISSUE #10721: strip gui from engine

I appreciate the options for stripping parts of the engine in a custom `.appmanifest` file. Instead of using Defold's built-in GUI system, I prefer to stick to the game world and build my own GUI framework on top of that. I assume there's plenty of GUI-related data in the engine that could be stripped out to minimize the final build size.

I'd like to see an option added to the `.appmanifest` file for stripping out all GUI-related functionality.

No alternatives, adds some bloat to my application.

I recall talk of eventually merging the GO and GUI worlds. Perhaps this could be considered a very basic step in that direction?

### LINKED ISSUE #12869: Add option to strip particlefx from engine

I use game objects in lieu of Defold's particlefx system, for the sake of collision mechanics and other fun things.

Please add an option to the _.appmanifest_ file for stripping out all particlefx-related functionality.

Similar requests:

https://github.com/defold/defold/issues/10721
https://github.com/defold/defold/issues/12870

### LINKED ISSUE #12870: Add option to strip tilemaps from engine

I use game objects in lieu of Defold's tilemap system. In my opinion, the current implementation of tilemaps is far less flexible than simply spawning a grid of game objects.

Please add an option to the _.appmanifest_ file for stripping out all tilemap-related functionality.

Similar requests:

https://github.com/defold/defold/issues/10721
https://github.com/defold/defold/issues/12869

AGulev:
Projects can now exclude the built-in GUI, Particle FX, and Tilemap subsystems from custom engines using new App Manifest settings. Each option removes the associated resources, components, and Lua module while preserving existing behavior for projects that do not enable the exclusions. Default engine configurations remain unchanged.

Release engine sizes with each option enabled independently:

| Platform | Configuration | Size (bytes) | Saved (bytes) | Reduction |
|---|---|---:|---:|---:|
| arm64-macos | Regular | 6,673,480 | — | — |
| arm64-macos | Without GUI | 6,523,528 | 149,952 | 2.247% |
| arm64-macos | Without Particle FX | 6,623,544 | 49,936 | 0.748% |
| arm64-macos | Without Tilemaps | 6,656,856 | 16,624 | 0.249% |
| wasm-web | Regular | 2,397,812 | — | — |
| wasm-web | Without GUI | 2,223,893 | 173,919 | 7.253% |
| wasm-web | Without Particle FX | 2,325,652 | 72,160 | 3.009% |
| wasm-web | Without Tilemaps | 2,380,879 | 16,933 | 0.706% |

The macOS measurements use an arm64 release engine stripped with `strip -x`. The web measurements use the optimized, uncompressed release `.wasm` output before gzip or Brotli compression.

Fix https://github.com/defold/defold/issues/12870
Fix https://github.com/defold/defold/issues/12869
Fix https://github.com/defold/defold/issues/10721

### Technical changes

- Added `exclude-gui`, `exclude-particle-fx`, and `exclude-tilemap` App Manifest properties with localized labels and tooltips.
- Registered Particle FX and Tilemap resources and components through the existing link-time resource and component descriptors.
- Registered the `particlefx` and `tilemap` Lua modules through the existing script-extension lifecycle, allowing their registration roots to be excluded by symbol.
- Moved Particle FX and Tilemap configuration and context creation into their component descriptors, with matching descriptor-owned cleanup.
- Supplied the 2D physics context to the Tilemap resource descriptor through the resource-type context map.
- Split GUI implementation objects into `gamesys_gui` and Particle FX implementation objects into `gamesys_particle`, keeping them outside the core `gamesys` archive dependency graph.
- Removed GUI callback registration and Particle FX/Tilemap context setup from the core engine. The GUI component descriptor now installs and clears its GUI context callbacks during component-type creation and destruction.
- Added `dmGui::SetContextCallbacks()` to the real and null GUI backends so GUI integration remains owned by `gamesys_gui` instead of the core engine.
- Added GUI script support for the existing `META_TABLE_GET_USER_DATA` mechanism, allowing a GUI script to expose its component’s owning game object without a global resolver or a direct GUI dependency in core gamesys.
- Preserved the original fast path for ordinary GO scripts. Added a typed internal `dmGameObject::GetInstanceFromLua()` overload that is used only after the normal GO lookup fails, such as for GUI script instances.
- Reused the existing `gui_null` backend when GUI is excluded. No separate `gamesys_gui_null` library is required.
- Added `particle_null` for builds that retain GUI while excluding Particle FX. It contains the Particle DDF descriptors needed by GUI and safe stubs for the particle functions referenced by GUI.
- Kept the GUI particle resource callback in `comp_particlefx.cpp`; `particle_null` supplies its corresponding null implementation when the real Particle FX library is excluded.
- Updated CMake installation and link definitions, default exported roots, Extender allowed libraries and symbols, and editor platform library handling. Waf files remain unchanged.
- Added App Manifest compatibility coverage for defaults, editing, round-tripping, setting combinations, platform manifests, and Windows library spellings.
- Added tests for the typed GO instance lookup, GUI instance user-data lookup, Particle FX null backend, and GUI linked against `particle_null`.
- Built release engines for arm64-macos and wasm-web in regular, individual-exclusion, and combined-exclusion configurations.
- Verified the exclusion variants link without the removed implementation libraries and that core `gamesys` has no GUI symbol dependency.
- Verified the Particle FX null backend with 1 test and 12 assertions, the GUI/Particle FX null combination with 1 test and 5 assertions, and focused GO and GUI script-instance lookup tests.
- No Lua, DMSDK, resource-format, Bob command-line, or Waf API changes are introduced.


engine/gamesys/src/gamesys/components/comp_gui_null.cpp
```diff
@@ -0,0 +1,33 @@
+// Copyright 2020-2026 The Defold Foundation
```
JCash: Is this really necessary?
The native extension system shouldn't need a null version, as
1) Gui script should also be excluded
1) no other engine code should call these functions

engine/particle/src/test/test_particle_null.cpp
```diff
@@ -0,0 +1,62 @@
+// Copyright 2020-2026 The Defold Foundation
+// Licensed under the Defold License version 1.0 (the "License"); you may not use
+// this file except in compliance with the License.
```
JCash, line 5: Do we have some instruction somewhere for this, because the LLM's love to shorten our licenses for some reason.

engine/particle/src/wscript
```diff
@@ -29,6 +29,12 @@ def build(bld):
                          use = 'DDF DLIB SOCKET',
                          source = ['particle.cpp', particle_ddf_proto])
 
+    particle_null = bld.stlib(features = 'cxx',
```
JCash: No need to change any Waf files anymore

engine/gamesys/CMakeLists.txt
```diff
@@ -195,6 +225,49 @@ foreach(_gamesys_dependency IN ITEMS
...
+  endif()
+endforeach()
+defold_target_link_libraries(gamesys_gui "${TARGET_PLATFORM}" SCOPE PUBLIC ${GAMESYS_GUI_DEPENDENCIES})
+set_target_properties(gamesys_gui PROPERTIES DEFOLD_SDK_HEADERS_TARGET gamesys_sdk_headers)
+
+defold_add_library(gamesys_gui_null STATIC
+    "${GS_SRC_DIR}/components/comp_gui_null.cpp")
+target_include_directories(gamesys_gui_null PUBLIC "${GS_SRC_DIR}" "${GS_DMSDK_SRC_DIR}")
+defold_target_link_libraries(gamesys_gui_null "${TARGET_PLATFORM}" SCOPE PUBLIC ${GAMESYS_GUI_NULL_DEPENDENCIES})
+set_target_properties(gamesys_gui_null PROPERTIES DEFOLD_SDK_HEADERS_TARGET gamesys_sdk_headers)
```
JCash: Same here, is this really necessary?
Gui is now an extension, and we shouldn't need a "null" version for it.

JCash (changes requested):
Overall good, but we should look over the "null" versions of our extensions, as it seems we've missed a connection somehwhere?

Compare to our "normal" native extensions, there is no null version of those.

Also, gui_script.cpp should use `DM_DECLARE_EXTENSION` as well.

engine/gamesys/src/gamesys/components/comp_gui_null.cpp
AGulev: it took me some time to remove direct call left in the codebase, and now it can be removed. Thanks

engine/particle/src/test/test_particle_null.cpp
AGulev: we probabbly need one

engine/particle/src/wscript
AGulev: I reverted all the waf changes

engine/gamesys/CMakeLists.txt
AGulev: removed

engine/gameobject/src/gameobject/gameobject_script.cpp
```diff
@@ -495,6 +495,15 @@ namespace dmGameObject
...
     }
 
+    HInstance GetInstanceFromLua(lua_State* L, uint32_t script_instance_type_hash) {
+        uintptr_t user_data;
+        if (dmScript::GetUserData(L, &user_data, script_instance_type_hash)) {
+            return (HInstance)user_data;
+        } else {
+            return 0;
+        }
+    }
```
JCash, line 505: 👍

engine/gamesys/src/gamesys/gamesys_script.cpp
```diff
@@ -49,13 +47,20 @@ namespace dmScript {
...
-        return instance;
+
+        dmScript::GetInstance(L);
+        uint32_t user_type_hash = dmScript::GetUserType(L, -1);
+        lua_pop(L, 1);
+
+        if (user_type_hash != 0)
+        {
+            return dmGameObject::GetInstanceFromLua(L, user_type_hash);
+        }
```
JCash, line 62: Could this be the implementation now?
Seems we don't need two implementations now?

JCash (approved):
Lgtm
Small "unification" question regarding our Lua contexts. Might want to look into merging the code cases a little bit.

engine/gamesys/src/gamesys/gamesys_script.cpp
AGulev: Yes, that was my initial thought as well—I even implemented it that way—but it makes the GameObject case a bit slower.

I think that could be critical in cases like physics, for example.

So I decided to keep the current logic, where we avoid dmScript::GetUserType by default and only use it for GUI, which does not happen as frequently as it does for GameObjects.

PR #12879, opened 2026-08-03
Fix Vulkan crash during graphics shutdownfix crash

### LINKED ISSUE #12878: Android/Vulkan: SIGSEGV in vkWaitForFences
**Description**

A fatal SIGSEGV crash occurs on Android (Vulkan) in the `DefoldJob_0` worker thread when `vkWaitForFences` is called inside `dmGraphics::AsyncProcessCallback` (`graphics_vulkan.cpp:5122`) after the Vulkan logical device has already been destroyed.

AI conclusions, treat with a grain of salt:
The root cause is a TOCTOU (time-of-check-time-of-use) race condition in the async texture upload path:

1. `AsyncProcessCallback` checks `m_DeleteContextRequested` at the top of the function and proceeds (line ~5039).
2. Concurrently, the main thread destroys the Vulkan logical device (`m_LogicalDevice.m_Device`).
3. The job thread reaches `vkWaitForFences(context->m_LogicalDevice.m_Device, ...)` with a now-invalid device handle → SIGSEGV (`fault addr 0x10`).

The existing guard (`dmAtomicGet32(&context->m_DeleteContextRequested)`) is checked too early — long before `vkWaitForFences` is reached — leaving a window where the device can be destroyed while the job is still in-flight.

Symbolicated call stack (arm64-v8a):

```
#00 LogStack() + 200
#01 sigaction + 212
#02 dmCrash::ResetToDefaultSignalHandler(int) at backtrace_signal_posix.cpp:47
#03 dmCrash::Handler(int, siginfo*, void*) at backtrace_libunwind_ndk.cpp:109
#04 art::SignalChain::Handler(int, siginfo*, void*) + 1128
#05 __kernel_rt_sigreturn
#06 vkWaitForFences + 436
#07 dmGraphics::AsyncProcessCallback(JobContext*, unsigned long, void*, void*) at graphics_vulkan.cpp:5121
#08 ProcessOneJob(JobThreadContext*, unsigned long) at jobsystem.cpp:483
#09 JobThread(void*) at jobsystem.cpp:574
#10 dmThread::ThreadStartProxy(void*) at thread_posix.cpp:76
#11 __pthread_start(void*) + 180
#12 __start_thread + 68
```

**To Reproduce**

Our observation is that the crash happens on the app reload, ie potentially during a collection transition / sys.reboot:

1. Launch the game 
2. Main map is opened
3. Close the game
4. Open the game again
5. Repeat previous steps 2-5 times

**Expected behavior**

No crashes.

**Defold version**
- Version 1.13.1 beta (commit `e27c83e8d3b949c93f94ade0f213140aa7d5e508`)

**Platforms**
- Platform: Android
- OS: Android 16 (also likely reproducible on other Android versions using Vulkan)
- Device: Google Pixel 9 (arm64-v8a, Mali GPU)

**Logs**

Fatal signal from logcat:
```
Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x10
in tid DefoldJob_0
```

**Additional context**

The crash site is `graphics_vulkan.cpp:5122` or `graphics_vulkan.cpp:5121`:

```cpp
// We wait for the fence here so we don't keep a bunch of large buffers around.
// Waiting for the fence should be fine here anyway, since we have commited the work on a different queue than the "main" graphics queue
vkWaitForFences(context->m_LogicalDevice.m_Device, 1, &fence, VK_TRUE, UINT64_MAX);
```

AGulev:
Prevents Vulkan applications from crashing during graphics shutdown when an asynchronous texture upload is still queued or running. Graphics teardown now waits for outstanding upload work before destroying the Vulkan device.

Fix https://github.com/defold/defold/issues/12878

### Technical changes

- Stop accepting asynchronous Vulkan texture uploads once graphics shutdown begins.
- Add an idempotent worker barrier that drains previously queued graphics jobs before destroying Vulkan resources and the logical device.
- Serialize the shutdown flag and job-barrier submission with asynchronous texture job creation to prevent new work from being queued behind the barrier.
- Apply the barrier from both `VulkanCloseWindow()` and `VulkanDeleteContext()`.
- Add an opt-in `issue-12878` regression test that delays the worker until teardown and verifies that the async upload drains before device destruction.
- The barrier follows the engine's current single-worker job-system configuration and asserts that requirement.
- Verified the regression four consecutive times using Vulkan on Apple M1 Max; the normal Vulkan graphics test also passes.

### DIFF
--- a/engine/graphics/src/vulkan/graphics_vulkan.cpp
+++ b/engine/graphics/src/vulkan/graphics_vulkan.cpp
@@ -21,6 +21,7 @@
 #include <dlib/log.h>
 #include <dlib/jobsystem.h>
 #include <dlib/thread.h>
+#include <dlib/time.h>
 
 #include <dmsdk/vectormath/cpp/vectormath_aos.h>
 
@@ -1705,7 +1706,7 @@ bail:
         VulkanContext* context = (VulkanContext*) _context;
         if (context != 0x0)
         {
-            dmAtomicStore32(&context->m_DeleteContextRequested, 1);
+            VulkanStopAsyncProcessing(context);
 
             for (uint32_t i = 0; i < DM_MAX_FRAMES_IN_FLIGHT; ++i)
             {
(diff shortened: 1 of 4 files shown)


engine/graphics/src/vulkan/graphics_vulkan.cpp
```diff
@@ -5134,6 +5135,52 @@ namespace dmGraphics
...
+    {
+        if (!context->m_AsyncProcessingSupport)
+        {
+            dmAtomicStore32(&context->m_DeleteContextRequested, 1);
+            return;
+        }
+
+        int32_atomic_t barrier_pending;
+        {
+            DM_MUTEX_SCOPED_LOCK(context->m_BaseContext.m_AssetHandleContainerMutex);
```
Jhonnyg, line 5158: Maybe add a comment about why we lock the asset container here? I'm assuming it's because we don't want to add or remove anything from it?

Jhonnyg (dismissed):
I think it's a good improvement! I think we should re-think the async uploading of graphics assets in the adapter, so we don't need a perfect solution right now.

engine/graphics/src/vulkan/graphics_vulkan.cpp
AGulev: added

PR #12891, opened 2026-08-04
Add "skip this release" button

### LINKED ISSUE #12887: Add "skip this release" functionality for updates
For users that might not want to upgrade immediately, the "Update Available" persists, it would be polite to the user to allow them to ignore an update.

JosephFerano:
Currently, there's no way to skip a new release of Defold so the "Update Available" status goes away, instead it will sit there forever until the user decides to upgrade. This PR adds a "skip" button in the release notes dialog as well as a context menu entry in "Help > Check for Updates" to allow the user to upgrade if they change their minds. For stable, this will effectively skip a Defold minor version (unless there's a patch release), but for alpha/beta, new releases means you will still get prompted for new releases in the same version.

Fixes #12887


editor/src/clj/editor/ui/updater.clj
```diff
@@ -29,11 +29,11 @@
   (fn [updater]
     (ui/run-later
       (let [can-install (updater/can-install-update? updater)
-            can-download (updater/can-download-update? updater)]
-        (ui/visible! link (or can-install can-download))
+            advertised (updater/update-advertised? updater)]
+        (ui/visible! link (or can-install advertised))
```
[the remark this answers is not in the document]
JosephFerano: I was missing a commit

editor/src/clj/editor/ui/updater.clj
```diff
@@ -53,55 +53,123 @@
...
+
+      :else
+      (dialogs/make-info-dialog
+        localization
+        {:title (localization/message "updater.up-to-date-dialog.title")
+         :icon :icon/circle-info
+         :owner stage
+         :header (localization/message "updater.up-to-date-dialog.header")}))))
+
+(defn check-for-updates! [stage project updater install-and-restart! localization]
```
AGulev, line 168: not sure if it's valid :

>Every invocation starts an independent future, while the Help command remains enabled. Repeated clicks can race updates to the shared updater state and schedule duplicate dialogs. If both dialogs choose Update, the second call can violate download-and-extract!’s can-download-update? precondition after the first starts downloading. Coalesce concurrent checks or disable the command while one is active.

JosephFerano: Yeah, seems like a worthwhile fix

AGulev: not sure it worth attention, but it's better to ask for your opinion

>[P2] A stalled update request permanently disables manual checks. check-for-updates! sets the in-progress flag before calling check!, but check! reads the URL without connect/read timeouts. If the server accepts the connection and stops responding, the future never reaches the finally, leaving “Check for Updates” disabled until restart. Use a bounded request like the release-notes downloader and clear the flag on timeout/cancellation.

AGulev (approved):

editor/src/clj/editor/ui/updater.clj
JosephFerano: Yeah, doesn't seem like that big of a deal.

PR #12896, opened 2026-08-04
Fix sound exhaustion when Web Audio is unavailable

AGulev:
HTML5 games now continue advancing sounds silently when the browser cannot create or run a Web Audio context. This prevents completed sounds from occupying all configured sound instances. When audio becomes available, output recovers automatically and active sounds continue from their current logical position.

### Technical changes

- Always create a valid HTML5 sound device, using a clocked 48 kHz silent fallback when Web Audio is unavailable.
- Continue consuming mixed buffers in silent mode so completed sound instances are released normally.
- Retry Web Audio every two seconds and immediately after user interaction or media-device changes.
- Preserve each device’s initial sample rate and switch recovered output to Web Audio without resetting native sound instances.
- Catch context creation, resume, buffer, and queue failures and return safely to silent playback.
- Share context-state handling across multiple Defold devices, guard asynchronous resume bookkeeping against stale contexts, and cancel obsolete queued sources during transitions.
- Add JavaScript-side device cleanup, including retry timers, device-change listeners, active sources, and the shared engine-owned context.
- Add defensive native cleanup when device opening fails and unregister the JavaScript device during close.
- Add deterministic fake-clock and fake-Web-Audio tests covering fallback, recovery, suspension, failures, stale resume promises, multiple devices, and final cleanup.
- Verified with `node --unhandled-rejections=strict engine/sound/src/test/test_library_sound.js` and JavaScript syntax checks.

Test Project:
[SilentDeviceTest.zip](https://github.com/user-attachments/files/30704467/SilentDeviceTest.zip)

### DIFF
--- a/engine/sound/src/js/library_sound.js
+++ b/engine/sound/src/js/library_sound.js
@@ -41,6 +41,7 @@ var LibrarySoundDevice =
         NotifyContextStateChanged: function(shared) {
             var audioCtx = shared.audioCtx;
             var running = audioCtx !== undefined && audioCtx.state == "running";
+            // Retries can report the same state repeatedly. Reset device queues only on a real transition.
             if (running != shared.contextRunning) {
                 shared.contextRunning = running;
                 DefoldSoundDevice.ForEachDevice(shared, function(device) {
@@ -70,6 +71,7 @@ var LibrarySoundDevice =
             }
 
             var audioCtx = shared.audioCtx;
+            // Detach the failed context first so every device immediately switches to silent clocking.
             shared.audioCtx = undefined;
             shared.resumePending = false;
             if (audioCtx !== undefined) {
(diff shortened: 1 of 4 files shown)


engine/sound/src/js/library_sound.js
```diff
@@ -1,257 +1,535 @@
...
+            shared.audioCtx = undefined;
+            var root = DefoldSoundDevice.GetGlobal();
+            var audioCtxCtor = root !== undefined ? (root.AudioContext || root.webkitAudioContext) : undefined;
+            if (audioCtxCtor === undefined) {
+                return undefined;
+            }
+
+            var audioCtx;
+            try {
+                // The default sample rate varies by output device and can be lower than 44100 Hz.
```
britzl, line 107: What is this comment referring to and why on the line below is the sampleRate fixed to 48000?

AGulev: this isn't my comment I keep this logic as is, as I understand it that we try 48k first if it's not supported take what device suggests. It's needed to support all the needed sound formats (WebAudio downsample automatically), see https://github.com/defold/defold/pull/10493

Image

britzl (approved):

PR #12900, opened 2026-08-05
Native Apple Silicon support for the iOS Simulator

### LINKED ISSUE #11705: Add ios arm64 simulator build option to bob
**Is your feature request related to a problem?**
Ios arm builds installs on xcode simulators, but crashes on launch. After researching this a bit I noticed other engines offer dedicated platform targets for xcode simulator builds. 

**Describe the solution you'd like**
Would be great if we can add a bob platform target like: arm64-ios-simulator

**Describe alternatives you've considered**
We can test in physical devices. But simulators are useful to do quick tests for application window / layout behavior in physically not present, or newly released devices.

ThomasRopke:
Running a game in the iOS Simulator on an Apple Silicon Mac used to require Rosetta, a replaced universal simulator runtime and manual signing, because the old `x86_64-ios platform` produced an Intel binary that modern Xcode simulator runtimes cannot install. This release adds a new `arm64_sim-ios` platform that builds, bundles, installs and launches natively in the simulator with no extra setup.

In the editor, the iOS bundle dialog now has a single "Simulator" checkbox instead of the architecture checkboxes, and simulator bundles no longer ask for a signing identity or provisioning profile.

⚠️ ⚠️ ⚠️ **Important note**: the `x86_64-ios` platform is removed. If your app manifest or native extension references x86_64-ios, rename that section to `arm64_sim-ios`. The long-unsupported armv7-ios entries have also been cleaned up.

Fixes #11705

### Technical changes
Running a game in the iOS Simulator on an Apple Silicon Mac used to require manual work. The `x86_64-ios` platform produced an Intel binary, which the ARM-only simulator runtimes in current Xcode versions cannot install, so you had to install Rosetta, replace the runtime with a universal one and sign the bundle by hand. That is no longer needed.

The new `arm64_sim-ios` platform builds, bundles, installs and launches natively. In the editor, the iOS bundle dialog now has a single Simulator checkbox instead of the architecture checkboxes, and a simulator bundle no longer asks for a signing identity or a provisioning profile.

`x86_64-ios` is removed. If you name it in an app manifest, rename that section to `arm64_sim-ios`.

Technical notes:

The platform is named `arm64_sim-ios` following the `wasm_pthread-web` precedent: the simulator is a variant in the architecture field, not a new OS. Because the OS token stays `ios`, every existing `ios$` regex, every `rsplit("-", 1)[-1]` and CMake's `REGEX REPLACE "^[^-]+-"` resolve correctly with no change, and Extender's `common -> <os> -> <arch>-<os>` resolution inherits the `ios` section. `build_input.yml` is a 14 line in-place edit of the old `x86_64-ios` section rather than a new duplicated base section.

`-mios-simulator-version-min` is what makes clang and ld tag a binary with the IOSSIMULATOR platform that simctl requires. It is selected in waf, CMake, `share/ext/common.sh` and `build_input.yml`, and all four have to agree.

IOSBundler rewrites the Apple platform keys in Info.plist and ad-hoc signs the bundle inside out, so nested frameworks and app extensions stay valid. The nested signing loops moved into a shared helper; the device path runs the same codesign commands as before.

`armv7-ios` is removed from `waf_dynamo.py` and `app_manifest.clj`. It has been unsupported since 1.2.193 and the entries were dead.

`share/ext/common.sh` had no LDFLAGS of its own and leaked MACOSX_DEPLOYMENT_TARGET from the preceding host tool build. Both break the device package build on current Xcode, which I bisected against protobuf: without LDFLAGS the CMake configure fails to find Threads, and with MACOSX_DEPLOYMENT_TARGET set clang rejects the iOS version flag. The fix is outside the simulator itself but are needed to build the packages at all, and can be split out if you would rather review them separately.

Vulkan stays disabled for the simulator, as it was for `x86_64-ios`, so no MoltenVK simulator package is added. The twelve precompiled dependencies are rebuilt for the new platform and the retired `x86_64-ios` archives are deleted, the same way js-web was retired. The non-English localization files are left alone since they come from Crowdin.

`main-ci.yml` keeps `skip_tests: true` from the old `x86_64-ios` entry. Turning the simulator leg on needs a boot step or `IOS_SIMULATOR_ID` in the workflow, because `select_simulator()` in `build_tools/build_ios.py` requires a booted device when several are available, which is the state of a fresh runner. Left for a follow-up.

Verified by:

- Engine unit tests on a booted iPhone 16e simulator: 125 test apps, 158 assertions, 0 failures.
- Bob `PlatformTest` and `GamepadConverterTest`, and the editor app-manifest, native-extensions and save-data suites: 0 failures, 0 errors.
- Bundling, installing and playing a real game in the simulator, with the resulting bundle checked for `CFBundleSupportedPlatforms=iPhoneSimulator`, `DTSDKName=iphonesimulator26.5`, Mach-O platform 7 (IOSSIMULATOR) and an ad-hoc signature.


scripts/cmake/functions_test.cmake
```diff
@@ -351,7 +351,7 @@ function(_defold_stage_ios_xcode_test_app target_name run_dir_norm configfile)
 endfunction()
 
 function(_defold_configure_ios_xcode_test_app target_name run_dir_norm configfile)
-  if(NOT (TARGET_PLATFORM MATCHES "ios$" AND CMAKE_GENERATOR STREQUAL "Xcode"))
+  if(NOT (TARGET_PLATFORM MATCHES "(ios|iossim)$" AND CMAKE_GENERATOR STREQUAL "Xcode"))
```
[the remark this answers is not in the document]
ThomasRopke: Fixed https://github.com/defold/defold/compare/d45edb8197aaa677d00ccd6a64bf01565222f957..07ace7471fd8b806560f4fe21efd367499b9b3fa

britzl:
@ThomasRopke is this PR ready for review?

ThomasRopke:
@britzl Yes a review would be appreciated, thank you

britzl:
> @britzl Yes a review would be appreciated, thank you

Ok, I am assigning this to Mathias and Evgeniy

build_tools/waf_dynamo.py
```diff
@@ -585,7 +585,9 @@ def default_flags(self):
         if 'linux' in self.env['BUILD_PLATFORM']:
             self.env.append_value('LINKFLAGS', ['-target', '%s-apple-darwin19' % target_arch])
 
-    elif TargetOS.IOS == target_os and target_arch in ('armv7', 'arm64', 'x86_64'):
+    elif TargetOS.IOS == target_os and target_arch in ('armv7', 'arm64'):
```
ekharkunov: Just note: we could remove `armv7` here because it's supported for a long time

build_tools/waf_dynamo.py
```diff
@@ -1859,7 +1851,7 @@ def detect(conf):
         conf.env['AR']      = '%s/%sar' % (bin_dir, llvm_prefix)
         conf.env['RANLIB']  = '%s/%sranlib' % (bin_dir, llvm_prefix)
 
-    elif TargetOS.IOS == target_os and build_util.get_target_architecture() in ('armv7','arm64','x86_64'):
+    elif TargetOS.IOS == target_os and build_util.get_target_architecture() in ('armv7','arm64'):
```
ekharkunov: `armv7` can be removed (unsupported)

editor/src/clj/editor/app_manifest.clj
```diff
@@ -28,7 +28,7 @@
 
 (def android #{:armv7-android :arm64-android})
 
-(def ios #{:armv7-ios :arm64-ios :x86_64-ios})
+(def ios #{:armv7-ios :arm64-ios :arm64-iossim})
```
ekharkunov: `:armv7-ios` can be removed

editor/src/clj/editor/app_manifest.clj
```diff
@@ -44,11 +44,11 @@
 
 (def vulkan-ios #{:arm64-ios})
 
-(def metal-ios #{:arm64-ios :x86_64-ios})
+(def metal-ios #{:arm64-ios :arm64-iossim})
 
 (def all-platforms
   #{;; ios
-    :armv7-ios :arm64-ios :x86_64-ios
+    :armv7-ios :arm64-ios :arm64-iossim
```
ekharkunov: :armv7-ios can be removed

editor/src/clj/editor/app_manifest.clj
```diff
@@ -809,7 +809,7 @@
     [[:platforms [;; ios
                   [:armv7-ios platform-pattern]
```
ekharkunov: :armv7-ios can be removed

share/extender/build_input.yml
```diff
@@ -380,20 +380,64 @@ platforms:
 
   # armv7-ios removed in Defold 1.2.193
 
-  x86_64-ios:
+  # Mirrors "ios" above: resolution is common -> <os> -> <arch>-<os>, and the
+  # simulator OS token is "iossim", so it cannot inherit from "ios". Keep in sync.
+  iossim:
```
ekharkunov: I suggest name new platform as `arm64_sim-ios`. The same way as we named `wasm_pthread-web`. In that case we avoid duplication of common flags/env vars in build_input.yml.

ekharkunov (changes requested):
Mostly looks good to me.
Main concern is platform name. I suggest to name it `arm64_sim-ios` in build_input.yml and related places.
Also I mentioned small cleanup across the code: we can remove armv7-ios from the code because it's unsupported for a long time

share/extender/build_input.yml
JCash: Yeah, I was the one suggesting "arm64-iossim", but neither is a very good name.

I now also suggest we use "arm64_sim-ios" for now, until we can revamp all our names and code into something we like better (e.g. a 3-tuple)

JCash (commented):
Overall good, some smaller notes:

As suggested, for now, let's use "arm64_sim-ios" for legacy reasons.

Have you run the unit tests locally?
Once those works, we'd likely want to enable it in `main-ci.yml`.

See native_extensions.clj
"[P1] Editor simulator builds fail for projects with native extensions. The editor now selects arm64-iossim, but editor.engine.native-extensions/extender-platforms has no corresponding Platform/Arm64IosSim entry. Consequently, supported-platform? returns false and custom-engine builds fail before contacting Extender. Add the platform mapping, library paths, and Info.plist mappings."

See GamepadConverter.java
"[P2] Simulator builds discard iOS gamepad mappings. GamepadConverter.normalizePlatform("arm64-iossim") returns "arm64-iossim" because it only recognizes ios and names ending in -ios. Existing mappings normalize to "ios", so both .gamepads and SDL mappings are filtered out. Normalize iossim/*-iossim to ios and add coverage."

ThomasRopke:
All addressed and rebased on dev.

Renamed to `arm64_sim-ios`. The OS token stays `ios`, so the existing `ios$` regexes work
unchanged and `build_input.yml` is a 14 line edit of the old `x86_64-ios` section instead of a
duplicated base section. `armv7-ios` removed from `waf_dynamo.py` and `app_manifest.clj`.
native_extensions.clj fixed. GamepadConverter needed no change: `arm64_sim-ios` ends in `-ios`
so `normalizePlatform` already handles it, which the old name did not. Added a test to pin that.

Tests run locally on a booted iPhone 16e: 106 engine test executables, 1713 cases, 0 failures.
Bob and editor suites clean. Also bundled and played a real game in the simulator.

main-ci.yml keeps `skip_tests: true`. Enabling the simulator leg needs a boot step or
`IOS_SIMULATOR_ID` in the workflow, since `select_simulator()` raises when nothing is booted and
the runner has several available. Better as a follow-up than more surface on this one.

Two fixes are outside the simulator itself: `common.sh` (LDFLAGS and unsetting
`MACOSX_DEPLOYMENT_TARGET`, which affects the device path too) and the harfbuzz version in
`external/skribidi/wscript`. Both are needed to build the packages on current Xcode, so they land here.

ekharkunov (dismissed):

JCash (dismissed):

ThomasRopke:
@JCash anything you need from me before we can merge this?

ThomasRopke:
Run 33846092085 on 78245c0df9: the six Install dependencies failures from the previous run are gone. Those were the ncurses5 URLs fixed by [#13088](https://github.com/defold/defold/issues/13088), picked up in the rebase.

The only failure left is bld-eng-darwin (x86_64-ios, true), same step as last run. The ok-to-test mirror takes .github/ from dev, so the matrix that runs is dev's, which still contains x86_64-ios. This PR removes that platform, so the job cannot build. The rename of that matrix row to arm64_sim-ios is in this PR's main-ci.yml and is discarded by the mirror step.

For the same reason arm64_sim-ios is never built by the mirror. The other seven jobs were cancelled by fail-fast rather than failing on their own.

PR #12905, opened 2026-08-06
Skip dart sass etag check

JosephFerano:
Per a discussion on slack, this will help avoid some DNS failures we're experiencing at times when trying to get the ETag for a cached resource during CI runs. Doesn't directly fix the issue, since it's external but it'll minimize disruptions.


AGulev (changes requested):
saas.clj should be included into a cache key in github actions:
https://github.com/defold/defold/blob/41853fcf38e54a6edeb09fa4684d1f872eb66e1a/.github/workflows/main-ci.yml#L43
...
>${{ hashFiles('build_tools/sdk.py', 'editor/tasks/leiningen/sass.clj') }}

britzl (approved):

AGulev (approved):

PR #12908, opened 2026-08-06
Add a mutex for guarding the graphics queue for Vulkan

Jhonnyg:
### Technical changes

* Serializes Vulkan queue access for vkQueueSubmit and vkQueuePresentKHR through LogicalDevice::m_QueueMutex.
* Adds an async texture upload queue regression/stress test covering concurrent upload-worker queue usage while the main thread submits/presents frames.
* Fixes Vulkan dynamic vertex/index buffer overwrites by deferring old buffer storage on full buffer data uploads instead of overwriting storage that may still be referenced by recorded command buffers.
* Corrects Vulkan index buffer sub-data uploads to write at the requested byte offset.
* Tracks Vulkan buffer logical size from the requested vkCreateBuffer size instead of the padded memory allocation requirement.
* Fully initializes default-constructed Vulkan DeviceBuffer objects.
* Avoids retagging buffer frame ownership for CPU-only uploads performed outside an active frame.
* Simplifies the async texture queue test by using fixed stress-test parameters instead of environment-variable configuration.

### DIFF
--- a/engine/graphics/src/test/test_app_graphics.cpp
+++ b/engine/graphics/src/test/test_app_graphics.cpp
@@ -1270,9 +1270,9 @@ static void* EngineCreate(int argc, char** argv)
     graphics_context_params.m_DefaultTextureMinFilter = dmGraphics::TEXTURE_FILTER_LINEAR_MIPMAP_NEAREST;
     graphics_context_params.m_DefaultTextureMagFilter = dmGraphics::TEXTURE_FILTER_LINEAR_MIPMAP_NEAREST;
     graphics_context_params.m_VerifyGraphicsCalls     = 1;
-//#if defined(DM_VULKAN_VALIDATION)
+#if defined(DM_VULKAN_VALIDATION)
     graphics_context_params.m_UseValidationLayers     = 1;
-//#endif
+#endif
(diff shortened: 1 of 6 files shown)


Jhonnyg:
@codex review

JCash (dismissed):
lgtm!

JCash:
LLM:
"[P2] Restore the validation-layer build guard" body="The commented-out preprocessor directives enable Vulkan validation unconditionally. When `VK_LAYER_KHRONOS_validation` is not installed, instance creation merely disables it, but `CreateLogicalDevice()` still requests the unavailable device layer and may fail with `VK_ERROR_LAYER_NOT_PRESENT`. This can prevent `test_app_graphics` from starting on ordinary Vulkan installations. Restore the `#if defined(DM_VULKAN_VALIDATION)` guard." file="/Users/mathiaswesterdahl/work/defold/engine/graphics/src/test/test_app_graphics.cpp" start=1273 end=1275 priority=2}

The queue serialization and buffer-lifetime changes otherwise look sound under static review. I reviewed exact head `a486b26a1d`; `git diff --check` passes. CI was still mostly pending, so I did not treat it as validation evidence.

